Issue #1271, opened 2022-04-15
Docs: The "Structs and Tuples" doesn't mention Tuples


dmihal:
See title

This page should either describe tuples, or reference the Tuple description on the "Built-In Types" page

Issue #1270, opened 2022-04-15
Document how to regenerate `forc` command docs in readme


adlerjohn:
Add instructions to regenerate the documentation for `forc` subcommands to the README.md in https://github.com/FuelLabs/sway/tree/417a73c16c7926d58f53aca1f434886528244a58/docs

Ref: #1047

mitchmindtree:
Currently these docs are in the `forc-documenter` crate here:

https://github.com/FuelLabs/sway/tree/417a73c16c7926d58f53aca1f434886528244a58/scripts/forc-documenter

Perhaps we could link to these from the `docs` README.

adlerjohn:
Those are pretty good. Even a simple link to that would be sufficient than the current _nothing_ :joy:

Issue #1156, opened 2022-04-05
Document that `asm` blocks don't support jump instructions and how to work around that by using control flow outside the `asm` block


mohammadfawaz:
Basically, `ji` and `jnei` don't work in an `asm` block. This is okay because users can always use `if/else/while` and combine smaller `asm` blocks that way.

PR #1278, opened 2022-04-16
Doc Fixes - `forc-documenter`, assembly, tuples, smart-contracts link

### LINKED ISSUE #1277: 404 from Sway book
the "smart contract" link gives a 404, here: https://fuellabs.github.io/sway/latest/blockchain-development/storage.html#storage

ControlCplusControlV:
Wanted to do some quick docs fixes and decided to merge them all into one PR as 4 less than 10 line PRs felt a bit spammy. (Can break up if needed though)

This fixes
#1277 
#1271 
#1270 
#1156 

Tried to keep with docs style but if any stylistic or errors I missed are there lmk and I can get them resolved.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -13,3 +13,13 @@ To serve locally:
 ```sh
 mdbook serve
 ```
+
+## Regenerate Forc SubCommand Docs
+
+With forc installed running the command
+
+```sh
+cargo run --bin forc-documenter write-docs
+```
+
+will generate the proper docs for `forc` and it's commands based on `forc --help`. This behavior is further documented in [the Forc Documenter README](../scripts/forc-documenter/README.md).
--- a/docs/README.md
+++ b/docs/README.md
@@ -22,4 +22,4 @@ With forc installed running the command
 cargo run --bin forc-documenter write-docs
 ```
 
-will generate the proper docs for `forc` and it's commands based on `forc --help`. This behavior is further documented in [the Forc Documenter README](../scripts/forc-documenter/README.md).
+will generate the proper docs for `forc` and its commands based on `forc --help`. This behavior is further documented in [the Forc documenter README](../scripts/forc-documenter/README.md).
--- a/docs/src/advanced/assembly.md
+++ b/docs/src/advanced/assembly.md
@@ -31,7 +31,7 @@ Note that in the above example:
 - `one` is an example of a "reserved register", of which there are 16 in total. Further reading on this is linked below under "Semantics".
 - we return `r2` & specify the return type as being u32 (the return type is u64 by default).
 
-An important note is that the `ji` and `jnei` opcodes are not available with the `asm` block. For those looking to use control structure with assembly it is reccomended to use `if`, `else`, and `while` control structures in surrounding sway, to the combine multiple smaller `asm` blocks. 
+An important note is that the `ji` and `jnei` opcodes are not available within an `asm` block. For those looking to introduce control flow to `asm` blocks, it is recommended to surround smaller chunks of `asm` with control flow (`if`, `else`, and `while`).
 
 ## Helpful Links
 
--- a/docs/src/advanced/assembly.md
+++ b/docs/src/advanced/assembly.md
@@ -31,6 +31,8 @@ Note that in the above example:
 - `one` is an example of a "reserved register", of which there are 16 in total. Further reading on this is linked below under "Semantics".
 - we return `r2` & specify the return type as being u32 (the return type is u64 by default).
 
+An important note is that the `ji` and `jnei` opcodes are not available with the `asm` block. For those looking to use control structure with assembly it is reccomended to use `if`, `else`, and `while` control structures in surrounding sway, to the combine multiple smaller `asm` blocks. 
+
 ## Helpful Links
 
 For examples of assembly in action, check out the [Sway standard library](https://github.com/FuelLabs/sway/tree/master/sway-lib-std).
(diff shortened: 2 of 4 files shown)


docs/src/advanced/assembly.md
```diff
@@ -31,6 +31,8 @@ Note that in the above example:
 - `one` is an example of a "reserved register", of which there are 16 in total. Further reading on this is linked below under "Semantics".
 - we return `r2` & specify the return type as being u32 (the return type is u64 by default).
 
+An important note is that the `ji` and `jnei` opcodes are not available with the `asm` block. For those looking to use control structure with assembly it is reccomended to use `if`, `else`, and `while` control structures in surrounding sway, to the combine multiple smaller `asm` blocks.
```
sezna, line 34:
```suggestion
An important note is that the `ji` and `jnei` opcodes are not available within an `asm` block. For those looking to introduce control flow to `asm` blocks, it is recommended to surround smaller chunks of `asm` with control flow (`if`, `else`, and `while`).
```

Thanks for updating all the docs, just adding some clarity/jargon here 😄

docs/README.md
```diff
@@ -13,3 +13,13 @@ To serve locally:
...
+
+## Regenerate Forc SubCommand Docs
+
+With forc installed running the command
+
+```sh
+cargo run --bin forc-documenter write-docs
+```
+
+will generate the proper docs for `forc` and it's commands based on `forc --help`. This behavior is further documented in [the Forc Documenter README](../scripts/forc-documenter/README.md).
```
adlerjohn, line 25:
```suggestion
will generate the proper docs for `forc` and its commands based on `forc --help`. This behavior is further documented in [the Forc documenter README](../scripts/forc-documenter/README.md).
```

docs/src/basics/structs_and_tuples.md
```diff
@@ -41,6 +41,24 @@ _This information is not vital if you are new to the language, or programming in
 
 Structs have zero memory overhead. What that means is that in memory, each struct field is laid out sequentially. No metadata regarding the struct's name or other properties is preserved at runtime. In other words, structs are compile-time constructs. This is the same in Rust, but different in other languages with runtimes like Java.
 
+## Tuples
+
+Tuples are a [basic static-length type](./built_in_types.md#tuple-types) which contain multiple different types within themselves. The type of a type is defined by the values within it, and a tuple can contain basic types as well as structs and enums. An example of a tuple declaration is provided below.
```
adlerjohn, line 46:
```suggestion
Tuples are a [basic static-length type](./built_in_types.md#tuple-types) which contain multiple different types within themselves. The type of a tuple is defined by the types of the values within it, and a tuple can contain basic types as well as structs and enums. An example of a tuple declaration is provided below.
```

docs/src/basics/structs_and_tuples.md
```diff
@@ -41,6 +41,24 @@ _This information is not vital if you are new to the language, or programming in
...
+
+```sway
```
adlerjohn, line 48: I would prefer if new code snippets were put into `examples/`, but these are small so I could let it slide this time.

ControlCplusControlV: Ok, will keep that in mind for future PRs. Snippets like this hopefully won't change much with future releases (declarations and simple operations) but I get why that could be harder to maintain on larger examples such as the example programs

adlerjohn (approved):
lgtm, but would like @sezna to give another set of eyes

mitchmindtree:
> 4 less than 10 line PRs felt a bit spammy

Just for the record, I don't think there's a PR size too small for @adlerjohn :)

docs/src/basics/structs_and_tuples.md
mitchmindtree: Keep in mind we can also use ranges to lift certain slices of code out of the `test` directory if we want to - could be a nice alternative for smaller cases where we want small snippets tested, but don't want to make a whole new example for a small snippet.

ControlCplusControlV: That would be a good way to pull small snippets, would it also be possible then to pull code from the `std` library as it is filled with useful and high quality code?

mitchmindtree: Yes, it should be safe to reference anything within this repo.

Here's an example of how a sub-snippet is [referenced](https://github.com/FuelLabs/sway/blob/master/docs/src/examples/subcurrency.md) in markdown, and here it is [rendered](https://fuellabs.github.io/sway/latest/examples/subcurrency.html).

Here, only the `body` section of the example is included, described by the postfix `:body` after the filename.

```
{{#include ../../../examples/subcurrency/src/main.sw:body}}
```

If we take a look at `examples/subcurrency/src/main.sw` we can see the "anchor" provided [here](https://github.com/FuelLabs/sway/blob/master/examples/subcurrency/src/main.sw#L1). Since this was written, the "body" anchor used has come to represent the entire file, but similarly anchors can be used to include smaller subsections of code.

I think the benefit of using comment anchors over line numbers is that, in the case that the code (example or test) is modified for whatever reason in the future, the snippet should still contain the relevant section. The risk with line numbers of course is that the slice slowly becomes offset from the desired code as the code is updated in the future. All that said, I'm not even sure it's possible to use direct line number ranges in the `#include` preprocessor anyway.

sezna (approved):

docs/src/basics/structs_and_tuples.md
ControlCplusControlV: Interesting, will try toying around with in my future docs PRs then so hopefully it is easier to maintain. I didn't know Sway could use Anchor's like that, so will use it as the default approach for include smaller subsections of code in `examples` for any future docs updates

PR #1279, opened 2022-04-16
Docs: add example of asm tuple return

nfurfaro:


docs/src/advanced/assembly.md
```diff
@@ -23,6 +23,27 @@ pub fn add_1(num: u32) -> u32 {
...
+An `asm` block can only return a single register. If you really need to return more than one value, you can return a tuple. Here's an example showing how you might return a tuple of `(u64, u64)`:
+
+```sway
+pub fn adder(a: u64, b: u64, c: u64) -> (u64, u64) {
+    let empty_tuple = (0u64, 0u64);
+    asm(output: empty_tuple, r1: a, r2: b, r3: c, r4, r5) {
+        add r4 r1 r2;      // add a & b and put the result in r4
+        add r5 r2 r3;      // add b & c and put the result in r5
+        sw output r4 i0;   // store the word at r4 in output + 0 bytes
+        sw output r5 i1;   // store the word at r5 in output + 1 word
```
nfurfaro, line 35: I tested this example before opening the PR, and it is correct. What supprised me was that the `imm` offset here needs to be `i1` (1 word) and not `i8` (8 bytes).

docs/src/advanced/assembly.md
```diff
@@ -23,6 +23,27 @@ pub fn add_1(num: u32) -> u32 {
 }
 ```
 
+An `asm` block can only return a single register. If you really need to return more than one value, you can return a tuple. Here's an example showing how you might return a tuple of `(u64, u64)`:
+
+```sway
+pub fn adder(a: u64, b: u64, c: u64) -> (u64, u64) {
```
adlerjohn, line 29: Put this in an actual Sway file in `examples/` so it can be tested as part of CI.

docs/src/advanced/assembly.md
```diff
@@ -23,6 +23,27 @@ pub fn add_1(num: u32) -> u32 {
 }
 ```
 
+An `asm` block can only return a single register. If you really need to return more than one value, you can return a tuple. Here's an example showing how you might return a tuple of `(u64, u64)`:
```
mohammadfawaz, line 26: Should we mention that what we're really returning is a reference to a tuple and not the tuple itself? Someone might  say: "you just said I can only return a single register from an `asm` block and now you're telling me I can return a tuple that contains multiple registers?"

nfurfaro: 👍

docs/src/advanced/assembly.md
otrho, line 35:
```suggestion
        sw output r4 i0;   // store the word in r4 in output + 0 words
        sw output r5 i1;   // store the word in r5 in output + 1 word
```
Yep, `lw` and `sw` always deal in words... if you had a `b256` in there then you'd be `mcpi`ing 32 _bytes_... this is making me feel uneasy though.  Let's never document anything more complicated than this (a couple of `u64`s) for ASM blocks.  🙂

otrho (changes requested):

docs/src/advanced/assembly.md
nfurfaro: Sounds good to me.

otrho (approved):

PR #1282, opened 2022-04-18
Format `storage` similar to `struct`s

### LINKED ISSUE #1218: Formatter adds a trailing white space in storage declarations
Example
```rust
contract;

storage {
    balance: u64, 
}
```
The formatter adds a trailing white space after `u64,` but it really shouldn't.

amadeusluntz:
Fix #1218 
It seems that the formatter was *not* handling the special `storage` declaration like a `struct` declaration, which I assume should be the case for the purpose of formatting. This fixes that and adds some tests to check it.

Also added `.vscode` IDE config file to gitignore.


.gitignore
```diff
@@ -17,3 +17,6 @@ sway-lib-std/Forc.lock
 
 # Forc's build directory
 out
+
+# Files specific to IDEs
+.vscode
```
adlerjohn, line 22: Platform- or IDE-specific excludes should be in `.git/info/exclude` or your personal `~/.gitignore`.

eureka-cpu:
Looks like it just needs formatting!

eureka-cpu:
Will take a look again tomorrow to confirm this, unless @JoshuaBatty gets to it first as it's quite late here

JoshuaBatty (approved):

Issue #1057, opened 2022-03-28
Review approach to parsing in `sway-fmt`


mitchmindtree:
There were some concerns raised at the tooling sync today around some possible duplication in work around `sway-fmt`'s approach to parsing. We should make sure to attempt to share as much of this logic with `sway-core` as is feasible and avoid having multiple implementations.

cc @ControlCplusControlV 
also cc @canndrew - if I remember correctly you're currently working on reworking sway's parser - is that right?

canndrew:
Yeah. I've written a new parser though it's not merged into master yet. i would certainly make sense to re-use that.

eureka-cpu:
I think (if I remember correctly) one of the problems is that the current formatter isn't making use of the type knowledge from the compiler and instead has separate logic to determine how to format each type.

mitchmindtree:
Assigning @eureka-cpu as designated driver for this one after our tooling chat today :smile: :tada: 

A couple ways to move forward that come to mind:

- Evaluate the approach taken by existing formatters (like `rustfmt`) - how/why are they different? Can we take inspiration or learn lessons? What are the higher-level steps taken to get from unformatted src -> formatted src?
- Review the new parser and see if it has what we need in terms of spans and "best effort" parsing. I.e. will attempt to construct and return the rest of the tree even if a branch is malformed. If missing anything, open issues to discuss desired features / best approaches.

mitchmindtree:
Going to close this in favour of the new tracking issue for the revamped swayfmt at #1516.

PR #1283, opened 2022-04-18
Fixed issue where sway formatter was operating on a pathless AST

### LINKED ISSUE #328: Compiler does not give Path to files on imports
for a code like this
```
script;

dep math;
use math::add;

fn main() {
    let p = add();
}
```

None of the _Spans_  give me Path to the file that _add()_ is defined in

amadeusluntz:
Fixes #328 

The issue was very specific to the formatted, `get_formatted_data` did not pass a `BuildConfig` to `sway_core::parse`.

The fix is fairly trivial and shouldn't affect anything since the case in which it was detected and in which it is relevant (parsing multiple files) already implies a `Forc.toml` exists (and indeed, `forc fmt` already fails with `Error: Manifest file does not exist` if one doesn't)

---

Pedantic note:

Whether or not this behavior is desirable I do not know, I guess there might be a design philosophy in which you'd want the formatter be to intentionally unaware of between-file/between-module relationship, maybe not even needing an AST (or the AST of a "valid" program) in order to gets things ala as-you-type formatting (rather than just when the file is valid) ...

I'm trying to keep this PR as impartial as possible on the issue since it uses the `BuildConfig` optionally; Might still encourage future code in the formatter using it (=> becoming more dependant on having a valid package structure). Maybe a minor concern, but feel free to close this (and the original issue) if it goes stylistically against where you're taking the formatter.


forc-fmt/Cargo.toml
```diff
@@ -14,5 +14,6 @@ clap = { version = "3", features = ["derive"] }
 forc-util = { version = "0.10.1", path = "../forc-util" }
 prettydiff = "0.5"
 sway-fmt = { version = "0.10.1", path = "../sway-fmt" }
+sway-core = { version = "0.10.1", path = "../sway-core" }
```
adlerjohn, line 17: Alphabetize

sway-lsp/src/capabilities/formatting.rs
```diff
@@ -18,7 +18,7 @@ pub fn format_document(
 pub fn get_format_text_edits(text: Arc<str>, options: FormattingOptions) -> Option<Vec<TextEdit>> {
     // we only format if code is correct
 
-    match get_formatted_data(text.clone(), options) {
+    match get_formatted_data(text.clone(), options, None) {
```
mitchmindtree, line 21: Should this also be passed a `BuildConfig` for correctness? Otherwise it appears `forc fmt` might behave differently from `sway-lsp`'s formatting?

amadeusluntz: That would mean making the lsp aware of a) what file the text is part of b) the manifest (forc.toml) likely associated with said file. With `forc fmt` this isn't that big an issue since it seems intended to run in a filesystem-aware way anyway, with the LSP it might not be the intended behavior (?)

If you think the behavior is fine I'd implement it in an optional way (i.e. if possible to find associated file & manifest pass them to `get_format_text_edits`, if not pass `None`, so you can still format any arbitrary bit of compileable sway code)

Seems like a higher-level design question, so up to you.

mitchmindtree: I've opened up #1283 to track this - happy to see this land in the meantime :)

mitchmindtree:
> Whether or not this behavior is desirable I do not know, I guess there might be a design philosophy in which you'd want the formatter be to intentionally unaware of between-file/between-module relationship, maybe not even needing an AST (or the AST of a "valid" program) in order to gets things ala as-you-type formatting (rather than just when the file is valid) ...

Totally, there has been quite a bit of recent discussion in meetings and on slack around a possible redesign of `sway-fmt`. In particular on how to take better advantage of the parser, and how what we want out of the parser might influence its implementation, e.g. maybe some kind of best-effort parsing of all branches with indicators in the cases that certain leafs/branches are malformed or something along these lines. See https://github.com/FuelLabs/sway/issues/1057.

It's looks like a recent PR landing might have caused a merge conflict - would you mind merging master or rebasing onto it? The once CI passes I'm happy to see this land :+1:

sway-lsp/src/capabilities/formatting.rs
amadeusluntz: I assume that's #1299

mitchmindtree: Ah woops, yes

amadeusluntz:
I think everything should pass now (but you need to approve the workflows to run again)

As a side note, is there some command to run every single check locally? (or at least the one that are run-able)? Would be a 1-line script but save forgetful people like myself a lot of "push only to realize there's an extra space" scenarios. If not would you mind me adding one?

mitchmindtree:
> If not would you mind me adding one?

Sure! If you do, would you mind adding it under the `scripts` directory?

Keep in mind that depending on what you're working on you might not always need to run everything (running absolutely all CI checks locally can take a long while). It's totally fine to lean on CI in these cases, so please don't stress if you miss anything. The common gotchas are the `cargo clippy` check and `cargo fmt --all --check`. I think some folk use `cargo watch` for clippy, and editor automation or local-only git hooks for auto-formatting. I personally just run them before committing or pushing, but have been meaning to look into some local-only git hook solution as I too am forgetful.

mitchmindtree (approved):

amadeusluntz:
@mitchmindtree I opened a bikeshedding PR to add said scripts and instructions for how to run them as pre-push hooks here #1302 ... but might be a waste since everyone might just have their own preferences for doing these locally ,and realistically `act` (for running all the actions locally) is annoying to setup so only the formatting script is easy to use as a hook.

PR #1284, opened 2022-04-18
Added option to not include std implicitly

### LINKED ISSUE #1150: Add the ability to disable the implicit `std` dependency in `Forc.toml`
Follow-up from #1137, which removes the need to explicitly add `std` as a dependency.

amadeusluntz:
Resolves #1150 , implementation is pretty straightforward, as described, defaults to including the std implicitly if the field is not specified.

Added some warnings when this option is used since I assume it's an uncommon use-case outside of working on sway itself or on some hyper-specialized program, also added a warning when the option is used *and* a path to the std is specified, which clarifies the explicit std will still be used (maybe not worth mentioning though, could remove).

---

Bikesheding:

I used `println_yellow_err` for the warnings since I've seen that behavior, but it might be worth having a specific `println_yellow_warning` function instead for greater clarity, or even setting a yellow == warning standard and renaming this function to `println_yellow_warning` or just `println_warning`.

### DIFF
--- a/forc-pkg/src/manifest.rs
+++ b/forc-pkg/src/manifest.rs
@@ -254,7 +254,6 @@ impl Manifest {
     fn implicitly_include_std_if_missing(&mut self, sway_git_tag: &str) {
         const CORE: &str = "core";
         const STD: &str = "std";
-        let include_implicit_std = self.project.implicit_std.unwrap_or(true);
         // Don't include `std` if:
         // - this *is* `core` or `std`.
         // - either `core` or `std` packages are already specified.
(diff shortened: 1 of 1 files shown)


forc-pkg/src/manifest.rs
```diff
@@ -263,13 +265,20 @@ impl Manifest {
...
         // Add a `[dependencies]` table if there isn't one.
         let deps = self.dependencies.get_or_insert_with(Default::default);
         // Add the missing dependency.
-        let std_dep = implicit_std_dep(sway_git_tag.to_string());
-        deps.insert(STD.to_string(), std_dep);
+        if !include_implicit_std {
+            println_yellow_err("Warning: Project being built without an STD");
+            let std_dep = implicit_std_dep(sway_git_tag.to_string());
+            deps.insert(STD.to_string(), std_dep);
+        }
```
mitchmindtree, line 281: Rather than adding a `if` branch here, would you mind if we added the condition to the list of early return conditions above (L266)? *See the other comment with the code suggestion.*

Also just a heads up - we *do* want to include `std` implicitly if `include_implicit_std` is `true`.

forc-pkg/src/manifest.rs
```diff
@@ -263,13 +265,20 @@ impl Manifest {
             || self.pkg_dep(STD).is_some()
             || self.dep(STD).is_some()
         {
+            if !include_implicit_std {
+                println_yellow_err("Warning: You specified an STD path but set `implicit-std` to `false`.\
+                 Your custom std path will take precdence and be imported regardless.");
+            }
```
mitchmindtree, line 271: I think it should be fine not to warn about this. Eventually we should add some more extensive `Forc.toml` manifest documentation to the Sway book that describes in detail what each of the fields in the manifest mean, similar to [cargo's manifest documentation](https://doc.rust-lang.org/cargo/reference/manifest.html). Under the `implicit_std` field documentation, we can list these conditions under which `std` is not included implicitly.

forc-pkg/src/manifest.rs
```diff
@@ -263,13 +265,20 @@ impl Manifest {
             || self.pkg_dep(STD).is_some()
             || self.dep(STD).is_some()
```
mitchmindtree, line 266:
```suggestion
            || self.dep(STD).is_some()
            || !self.project.implicit_std.unwrap_or(true)
```

mitchmindtree (changes requested):
Thanks for the PR! I've added a few notes.

forc-pkg/src/manifest.rs
amadeusluntz: Ups, should have been an else there for those includes... anyway, if warnings aren't needed your alternative solution is much more elegant, implemented it that way.

mitchmindtree (approved):

PR #1285, opened 2022-04-18
Shorthand for build and test

### LINKED ISSUE #1025: Add `b` and `t` shortcuts for `build` and `test`

amadeusluntz:
Resolves #1025 -- mainly opening the PR to ask some questions though

@otrho 's suggestion there was to find a library that procedurally generates shorthand for all arguments in a disambiguous way. This doesn't seem possible if we want said library to integrate with `clap`. Even if there were, shorthand/short arguments doesn't seem to be an option for enum-based parsing, which is a fairly elegant way to handle the first argument.

That being said, I think it'd be doable to annotate the enum with a proc macro that generates shorthands, but I'd wager it's too much bother writing it for not that much gain(?)

The other option would be to have a function the full command in `run_cli`, generates the shorthands based on all subcommands (and optionally all the flags/options), then transforms the command (with this approach you could get a behavior where `forc b -pa -p -pi` becomes `forc build --print-intermediate-asm --path --print-ir --silent` becomes`)  ... this would also require a macro-based approach but it seems like I'd make generating the appropriate docs very annoying (right now clap handles this automatically)

I myself am a fan of having the shorthands written by hand in the codebase or ideally not having them at all, I find that APIs are a horrible place to cut code-volume cost via code generation.

---

Anyway, if you think the manual approach is the best feel free to close this PR since I don't want to steal too many easy issues from other newbies. If you think an automatic-generated approach is best I'd love some more details about the ideal behavior and I can go ahead and try implementing it.

### DIFF
--- a/forc/src/cli/mod.rs
+++ b/forc/src/cli/mod.rs
@@ -31,8 +31,8 @@ struct Opt {
 enum Forc {
     #[clap(name = "addr2line")]
     Addr2Line(Addr2LineCommand),
+    #[clap(alias = "b")]
     Build(BuildCommand),
-    B(BuildCommand),
     Clean(CleanCommand),
     #[clap(after_help = completions::COMPLETIONS_HELP)]
     Completions(CompletionsCommand),
@@ -40,8 +40,8 @@ enum Forc {
     Init(InitCommand),
     ParseBytecode(ParseBytecodeCommand),
     Run(RunCommand),
+    #[clap(alias = "t")]
     Test(TestCommand),
-    T(TestCommand),
     Update(UpdateCommand),
     JsonAbi(JsonAbiCommand),
     /// Find all forc plugins available via `PATH`.
@@ -64,7 +64,7 @@ pub async fn run_cli() -> Result<()> {
     let opt = Opt::parse();
     match opt.command {
         Forc::Addr2Line(command) => addr2line::exec(command),
-        Forc::Build(command) | Forc::B(command) => build::exec(command),
+        Forc::Build(command) => build::exec(command),
         Forc::Clean(command) => clean::exec(command),
         Forc::Completions(command) => completions::exec(command),
         Forc::Deploy(command) => deploy::exec(command).await,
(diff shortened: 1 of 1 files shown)


forc/src/cli/mod.rs
```diff
@@ -32,6 +32,7 @@ enum Forc {
     #[clap(name = "addr2line")]
     Addr2Line(Addr2LineCommand),
     Build(BuildCommand),
+    B(BuildCommand),
```
adlerjohn, line 35: Hmmmmm. Rather than this, I think `clap` allows you to specify a short form?

mitchmindtree: Yeah I'm currently looking for this, but haven't been able to find it:

https://github.com/clap-rs/clap/blob/v3.1.9/examples/derive_ref/README.md#command-attributes

I wonder if it's only available on arguments and not subcommands?

adlerjohn: Ah good point!

mitchmindtree: One thing it might be worth adding to these short-hand commands is a `#[clap(about = "...")]` section that mentions the command is simply short-hand for the longer form. Otherwise I think it might generate duplicate "about" text under `forc --help`.

mitchmindtree: I've opened https://github.com/clap-rs/clap/issues/3639.

adlerjohn: Can you file an issue in this repo to track improving the changes in this PR once clap merges that issue?

mitchmindtree: Looks like there's a solution already and we just didn't read deep enough!

https://github.com/clap-rs/clap/issues/3639#issuecomment-1101350736

@amadeusluntz would you mind applying the recommendation mentioned in the above linked comment? That is, rather than adding a separate `B` enum variant, add the following attribute to `Build`:

```rust
    #[clap(alias = "b")]
    Build(BuildCommand),
```

that said the comment's recommendation looks a little different to the docs. If the above doesn't work, maybe try:

```rust
    #[clap(alias("b"))]
    Build(BuildCommand),
```

mitchmindtree: Oh and the same goes for `Test` too.

mitchmindtree (changes requested):
See [this comment](https://github.com/FuelLabs/sway/pull/1285/files#r852077459).

forc/src/cli/mod.rs
amadeusluntz: My bad, I just tested long/short, was unaware of `alias` working for enums. Changed to `visible_alias`.

amadeusluntz:
locally testing it works, shows up as:
`build             Compile the current or target project [aliases: b]`
rather than
`b, build             Compile the current or target project`
But that choice seems to be down to calp with no obvious way of changing it.

mitchmindtree (approved):

eureka-cpu (approved):
lgtm!

Issue #252, opened 2021-09-20
Parser desiredata


sezna:
It has been said more than a few times now that the parser could be improved. I'm going to summarize why we used Pest up until this point, and then we can use this issue to accumulate specific requests of the parser. After sorting that out, we should be able to formulate how to best update it to suit our needs.

The advantages of pest so far have been:
1. Rapid prototyping
2. A grammar to communicate to stakeholders
3. Tooling synchronization around the grammar

Moving forward, we should probably split our parser out into a separate tokenizer and lexer. We will still need, for the sake of the project:
1. A high level grammar file for language tooling, _or_ a simple tokenizer that can help out `forc` and the VSCode server. Updating our API to these other crates should be considered a part of this issue.
    1. Some formal grammar specification is needed even if we opt for the tokenizer above for communication and documentation about the language.

The problems, as I see them, are:
1. A blurring of the lines between the tokenizer and the lexer
2. A messy API coming from Pest
3. Limited grammar expressiveness

One solution I see for incremental progress would be to limit how specific our Pest grammar is, instead using it only for tokenizing. Then we could add a separate lexer module immediately after that which converts it into a `ParseTree`. The only other solution would be to do something entirely separate from pest, I suppose.

_This is not a high priority, but when we do get to this, it would be good to have a design hashed out and agreed upon._

PR #1286, opened 2022-04-18
sway-parse initial PR

canndrew:
This PR adds implements the new parser in a separate sway-parse crate in the cargo workspace.

You can currently switch between using the new parser or the old parser by modifying [this line](https://github.com/FuelLabs/sway/commit/ad209d69edbcc7ea30375c8ded640a30bf9d1c89#diff-066e2f8870fac537ada2033c79f060beaebb91c9d17af31fc413485add581deaR132) in sway-core/src/lib.rs. This should either be turned into a configurable option or removed entirely if/when we feel we're ready to commit to the new parser.

After the new parser gets run on a file its output is transformed into the old AST using the routines in sway-core/src/convert_parse_tree.rs. This is so that I didn't have to rewrite the typechecker to work on the new AST. However converting to the old AST also involves a lot desugaring/lowering so the old AST is now effectively a another intermediate representation.


canndrew:
Something that I haven't done yet is add error recovery, so if there's any syntax errors then compilation will fail and just report the first error. I think, before adding that, we should write ui tests like the rust compiler has in order to specify what error output should look like and what errors should be recoverable.

adlerjohn (commented):
ci failures

canndrew:
CI should be fixed now.

Cargo.toml
```diff
@@ -18,7 +17,8 @@ members = [
     "sway-utils",
     "test",
     "test-sig-gen-util",
-    "scripts/forc-documenter"
+    "scripts/forc-documenter",
+    "sway-parse",
```
adlerjohn, line 21: Alphabetize

sway-parse/src/expr/mod.rs
```diff
@@ -0,0 +1,1145 @@
+use crate::priv_prelude::*;
```
sezna, line 1: I love the idea of a private prelude. We should use this more!

sway-parse/src/parser.rs
```diff
@@ -0,0 +1,252 @@
...
+        Parser {
+            token_trees: token_stream.token_trees(),
+            errors,
+        }
+    }
+
+    pub fn emit_error(&mut self, kind: ParseErrorKind) -> ErrorEmitted {
+        let span = match self.token_trees.first() {
+            Some(token_tree) => token_tree.span(),
+            None => todo!(),
```
sezna, line 19: Preferably resolve this before merging

sezna (changes requested):
This is awesome. I can only imagine how amazing it feels to be on the final stretches of this!

1. I think we should get this in ASAP before merge conflicts come in. As most of the team is in Amsterdam, this week should be a nice calm period, so hopefully we can manage this.
2. In a follow up PR, could you write some top-level module documentation strings describing the overall architecture and algorithm? I found most of the code easy to read and self-documenting, but some architectural-level descriptions from the person who wrote this would go a long way.
3. With the old AST being another IR....maybe @emilyaherbert can have fun with a four-stage pipeline like MIR, haha. This layer is definitely unnecessary but I agree that it's the only way to manage this new parser's integration for now.  Not the end of the world.

Could you change the new/old parser toggle to be a command line flag that defaults to the new parser? That way, similar to IR, we can do `--use-old-parser`? It can just be something `forc` passes in to `BuildConfig`

eureka-cpu:
Is there a reference for issues that we were running into with the current parser that led to this PR (for tracking purposes)?

sezna:
#252

sway-parse/src/parser.rs
canndrew: Ooops. Good catch :grimacing:

sway-parse/src/expr/mod.rs
canndrew: Yeah I write all my crates like this. Other than making it easier to manage imports it also forces things to have a consistent meaning throughout the crate (eg. so `Span` doesn't mean two different things in two different files).

canndrew:
> Could you change the new/old parser toggle to be a command line flag that defaults to the new parser? That way, similar to IR, we can do --use-old-parser? It can just be something forc passes in to BuildConfig

I've done this. There's now a `--use-orig-parser` option.

canndrew:
This is now rebased to fix the conflicts.

mitchmindtree:
Adding the new flag to the `forc` commands requires updating the `forc` docs. Unfortunately we don't have any way of generating these docs while running `mdbook build`, however we *do* have a new `forc-documenter` script that at least catches out-of-date docs during CI and that automates the process of updating the forc command docs. Could you run:

```
cargo run --bin forc-documenter write-docs
```

and then push a new commit with the result? That should update the `forc` docs and appease CI.

adlerjohn:
Does this close #252?

sezna:
> Does this close https://github.com/FuelLabs/sway/issues/252?
yes

canndrew:
@sezna: I've made the requested changes. Is this ready to be merged?

sezna (approved):
Reviewing a PR this large is not super useful. I did scan it, and I also checked out the branch and tested it out. I'm sure there are things we are missing in the review, but it will probably make more sense to merge this in and handle the fallout, maybe holding a week or two before releasing so we can feel more solid about it.

PR #1213, opened 2022-04-11
Remove global `Namespace` arena in favour of structural sharing. Refactor to use absolute paths.

mitchmindtree:
Closes #682

# Part 1. ~Gobal State~ Structural Sharing

This removes the global arena used for managing namespaces during type
checking in favour of reverting to the original approach of representing
new scopes by cloning the namespace.

Prior to the introduction of the global arena, cloning the namespace to
represent new scopes caused exponential memory growth as the entirety of
each of the inner collections was cloned along with the namespace
itself.

In this new approach, we aim to avoid this by using clone-friendly,
drop-in-replacement collections from the `im` crate that enable
"structural sharing" of the `Namespace` fields. *See TODO below for
whether this is done yet or not.*

You can read more about structural sharing [here][1]. The gist is that
rather than deep cloning the entire structure, we clone a reference to
the original state, and new allocations only occur upon *modifying* the
data structure with only the memory needed to record the difference.

The result is that we can avoid the side-effecting nature of maintaining
global state, making it easier to reason about how `Namespace`s are
accessed and updated within functions by reading their signatures.

It should also allow us to lean on RAII for cleaning up the namespace
contents, solving the bug where we are currently unable to compile the
same project more than once within the same process as mentioned
[here][2]. This is particularly relevant to `sway-lsp` where we will
likely want to lean on `sway-core` to type check the same code many
times within one process.

This also gets us a step toward achieving referential transparency for a
lot of our type checking functions which may enable easier caching and
parallelism in the future, though this would still currently be blocked
by the type engine's global state.

Many of the newer methods previously provided by `NamespaceWrapper` and
`NamespaceRef` have been moved to `Namespace` itself.

# Part 2. Absolute Paths

The final set of changes in this PR are a refactor of the `Namespace` type in
order to ensure access to the whole module hierarchy (as well as the init
namespace) during all of type-checking.

### Splitting `Namespace` into `Root`, `Module` and `Items`.

Previously, we used a single `Namespace` structure to represent not only the
module tree but the items within each module too.

This PR refactors the original `Namespace` type into:

- `Module`: Represents a single module. Stores the set of items available within
  the module as well as the set of submodules.
- `Items`: The set of declarations, implementations, synonyms and aliases
  available within a particular scope.

This allows for moving module-level behaviour onto the `Module` type, and
item-level behaviour to the `Items` type. This is particularly useful in
type-checking for distinguishing between whether or not we need the entire
module tree or just the set of items for a particular module (e.g. the
monomorphize methods).

[...]

PR #1287, opened 2022-04-18
Fix import of submodules within `std::context`

mitchmindtree:
I noticed these when they resulted in compilation errors while working
on the `Namespace` refactor in #1213.

It appears that on `master` we're leaking submodules into the root
namespace, and as a result these haven't been caught yet. #1213 appears
to fix this and results in a compilation error. I might add a
`should_fail` E2E test as a part of #1213 to ensure we don't
accidentally leak submodules into the root in the future.


mohammadfawaz (approved):
I think this is reasonable. Good catch!

PR #1289, opened 2022-04-18
Fix bug where `forc` panics if dependency declared with both `package` and `path`

### LINKED ISSUE #1053: `forc` panics when a dependency is declared with both `path` and `package`
```
thread 'main' panicked at 'no entry found for key', forc-pkg/src/pkg.rs:542:45
```
It looks like an issue with the `graph_to_path_map` construction, possibly an issue where a dependency name is used where a package name should be.

Likely an oversight of #1002, but not really a blocker for anything.

mitchmindtree:
A small fix that allows us to point the `forc` E2E package dependency test to the local `std`. 

This should make it easier to iterate on changes in `sway-core` that might require breaking changes in `std` without breaking this test.

Closes #1053.

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -525,18 +525,18 @@ pub fn graph_to_path_map(
                 })?
             }
             SourcePinned::Path => {
-                let parent_node = graph
+                let (parent_node, dep_name) = graph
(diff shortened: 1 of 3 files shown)


adlerjohn (approved):
utACK

Issue #1291, opened 2022-04-18
Implement `Eq` for `String`


mohammadfawaz:
It would be useful to be able to compare two strings (to be able to `assert` for example).

nfurfaro:
Duplicate of #1060

mohammadfawaz:
Ah thanks! Closing as a dup then.

PR #1292, opened 2022-04-18
Correct handling string return values

mohammadfawaz:
Two things:
* Fixing the size of a `String`.
* Treating `Strings` of size 1 word or less as copy types. We might have to do the same for structs and other complex types when they're small enough to fit in registers.

The following contract, for example, now returns the correct receipts and can be tested from the SDK.

```rust
contract;

abi MyContract {
    fn get_small_string() -> str[8];
    fn get_large_string() -> str[9];
}

    fn get_small_string() -> str[8] {
        let my_string: str[8] = "gggggggg";
        my_string
    }
    fn get_large_string() -> str[9] {
        let my_string: str[9] = "ggggggggg";
        my_string
    }
}
```

The SDK has a test similar to the above that should be re-enabled (cc @digorithm). For now, I can't really check this from a script because I can't check what was returned from the contract call against a literal `String` using an `assert`. Relevant: https://github.com/FuelLabs/sway/issues/1291


test/src/e2e_vm_tests/test_programs/should_pass/language/is_reference_type/src/main.sw
```diff
@@ -20,7 +20,7 @@ fn main() -> bool {
     assert(!is_reference_type::<byte>());
     assert(!is_reference_type::<u64>());
 
-    assert(is_reference_type::<str[1]>());
+    assert(!is_reference_type::<str[1]>());
```
mohammadfawaz, line 23: If we go with this, then we should probably make it clear in the language that types that "fit in a word" are always copy types. That includes things like `str[1]`, `struct { u64 }`, `struct { struct { u64 } }`, `array[1]`, and so on. It is a bit weird, but I think this would make the specs a bit simpler where the optimization that we've been talking about is actually baked in the language and _has_ to happen.

otrho: I think you're right.  We'd absolutely need regression tests to enforce it for all our types too.

otrho (approved):

PR #1302, opened 2022-04-19
Scripts + instructions to git hook formatting checks and/or gh-action suite

amadeusluntz:
Based on the quick discussion here #1283

Adds two one-liners for running all formatting checks and all github actions respectively + instructions for how to add these as git hooks. Part of the instructions could also included in the remote's .git instead (the short script in `pre-push` to run everything in the `pre-push.d` dir)

Not sure if it's worth adding to the repo though, might be that everyone has their own preference for these and nobody but me might end up using these.

### DIFF
--- a/scripts/local_testing/README.md
+++ b/scripts/local_testing/README.md
@@ -1,0 +1,27 @@
+# Local Testing
+Tiny scripts for running formatting checks or tests locally. [act](https://github.com/nektos/act) is required for running all github actions.
+
+## To run as git hooks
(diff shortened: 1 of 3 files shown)


adlerjohn:
Our general preference for these things has been to allow users to locally do whatever they want, and use CI as the backstop for what is allowed to be merged into master or not. There's too many different platforms and variations in development environments to enforce things through git hooks or platform-specific scripts.

PR #1276, opened 2022-04-15
Add binary_xor function

nfurfaro:
This PR just adds a missing wrapper function for the `xor` op.
Test to come in a followup PR (they're already written).

PR #1314, opened 2022-04-19
Handle removed dep in manifest during path reconstruction from graph

mitchmindtree:
This fixes a bug where removing a path dependency (e.g. core) from the
manifest where that dependency is shared by another dependency (e.g.
std) could cause path map reconstruction to fail.

Previously, we assumed that the dependency would be present within the
manifest in order to retrieve the source path. This PR removes this
assumption and will instead reconstruct the dependency graph (and in
turn the lock file) when a dependency is discovered to be missing during
path reconstruction.

First spotted by @nfurfaro [here][1].

[1]: https://github.com/FuelLabs/sway/pull/1276/files#r853131667

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -536,11 +536,19 @@ pub fn graph_to_path_map(
                 let detailed = parent_manifest
                     .dependencies
                     .as_ref()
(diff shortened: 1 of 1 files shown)


JoshuaBatty (approved):
utACK

PR #1318, opened 2022-04-20
Add appropriate test harness for `forc init --template counter`

### LINKED ISSUE #1313: Update test harness for `forc init --template counter`
Currently, we are just using the default_test_program. Ideally, we want it to reflect what is mentioned in the sway book [here](https://fuellabs.github.io/sway/latest/testing/testing-with-rust.html).

There seems to be a bug related to the whole by-copy v.s. by-ref issue that the IR introduced. The example in the book fails with the below output.

```
---- harness stdout ----
Contract deployed @ e0e4b8e32cde9b060a1fd62887a8e7024032ba85863a59cefb6e39ce39f94125
thread 'harness' panicked at 'assertion failed: `(left == right)`
  left: `42`,
 right: `0`', tests/harness.rs:38:5
```

Once this is resolved, we can update the test harness file.

JoshuaBatty:
closes #1313

### DIFF
--- a/forc/src/ops/forc_init.rs
+++ b/forc/src/ops/forc_init.rs
@@ -190,19 +190,18 @@ pub(crate) fn init_from_git_template(
     if out_dir.join("tests").exists() {
         // Change the project name and authors of the Cargo.toml file
         edit_cargo_toml(&out_dir, &project_name, &real_name)?;
+
+        if template_name == "counter" {
+            edit_counter_test_harness(&out_dir, &project_name)?;
+        }
     } else {
         // Create the tests directory, harness.rs and Cargo.toml file
         fs::create_dir_all(out_dir.join("tests"))?;
 
-        // If the counter template is used, insert the counter test harness
-        // Ohterwise, insert the default test harness
-        let test_program = if template_name == "counter" {
-            counter_test_harness(&project_name)
-        } else {
-            defaults::default_test_program(&project_name)
-        };
-
-        fs::write(out_dir.join("tests").join("harness.rs"), test_program)?;
+        fs::write(
+            out_dir.join("tests").join("harness.rs"),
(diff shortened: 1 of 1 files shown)


mitchmindtree:
I don't think we want to have this special case just for the book's counter example embedded in `forc` itself. Perhaps instead, we can we check the template code itself for whether or not it contains a `test/` harness already and if so, copy that harness?

Ahh I've just noticed that the `counter` example no longer has that test harness in the repo. I think this might have been removed mistakenly in #1040, not realising that the harness was used in the book. Can we re-add the test harness to the example itself and then do the above? This way we can also ensure the test harness code that is copied as a part of the test is valid by checking it in CI.

As a side note: Now that `std` and `core` are local to this repo and we've resolved the cyclic dependency between the `fuels-rs` and `sway` repos, I think it should be fine to include the example test harnesses under the top-level workspace so that their build artifacts can be shared with the rest of the workspace, and so that these tests are run when we run the whole workspace.

cc @adlerjohn

SilentCicero:
Was wondering if the Counter example could just be directly pulled from the Sway examples from git. This way, others can also see the example code it's pulling (for tests and contracts etc).

Thoughts?

JoshuaBatty:
> 

Hi @SilentCicero this is already the exact functionality :) 

I agree we should include the test back into the counter example and then we can use that as a base for the template. Will update now.

mitchmindtree:
Yeah my mistake! I forgot that the default behaviour **already** clones the existing test harness for that template. Our real problem is just that our examples are missing test harnesses (as of #1040) so the correct fix here is probably just to re-add those (and make sure they get tested under CI).

adlerjohn:
The test file was removed in #1040 because it wasn't tested as part of CI. Agree the right approach is to add in the test harness again and test it as part of CI.

PR #1321, opened 2022-04-20
Add test harness for counter example

JoshuaBatty:
This was removed in #1040 but I think we should put back the counter example test as we can then use that for 
`forc init  --template counter`.  We can also in a future PR, link to this example in the sway book [here](https://fuellabs.github.io/sway/latest/testing/testing-with-rust.html)


Cargo.lock
```diff
@@ -3255,24 +3923,44 @@ dependencies = [
...
+ "sway-types 0.10.3",
+ "sway-utils 0.10.3",
  "tokio",
  "tower",
  "tower-lsp",
 ]
 
+[[package]]
+name = "sway-types"
+version = "0.8.0"
```
mitchmindtree, line 3935: Ahhhh, still a bit of the old cyclic dependency between `fuels-rs` an `sway` left.

mitchmindtree (commented):
Nice, I think this is the right approach.

First, we just need to untangle the last bit of the cyclic dependency between `fuels-rs` and `sway` caused by `fuels-core` depending on `sway-types`. This is just requires moving the `JsonABI` type declaration (and associated `Function` and `Property` types) into the `fuels-rs` repo. I'll have a go at this now.

Once this is done it will be safe to depend on the Rust SDK not only within these tests, but within any of the crates in this repo.

mitchmindtree:
>This just requires moving the JsonABI type declaration (and associated Function and Property types) into the fuels-rs repo. I'll have a go at this now.

I've opened https://github.com/FuelLabs/fuels-rs/pull/219.

mitchmindtree:
Ahh, looks like the `sway-lib-std` tests might need to account for the SDK version bump? If so, maybe it's worth doing so in a separate PR first?

mitchmindtree:
Once #1337 lands I think this will finally be unblocked!

We should update this harness to use the new `fuels` crate that originally landed as `fuels-rs` in https://github.com/FuelLabs/fuels-rs/pull/168 but then was renamed to `fuels` in https://github.com/FuelLabs/fuels-rs/pull/242. Should help to clean up the imports a little as described in https://github.com/FuelLabs/fuels-rs/pull/168.

PR #1322, opened 2022-04-20
Update CI to install forc and build sway examples 

JoshuaBatty:
CI is currently failing for PR #1321. This is because the `out` directory isn't being built for the sway examples, leading to CI errors for:
`cargo-build-workspace`
`cargo-clippy`
`cargo-test-lib-std`
`cargo-test-workspace`
`cargo-unused-deps-check`

This PR now installs `forc` and builds all sway examples during each of these stages now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,6 +107,16 @@ jobs:
           profile: minimal
           toolchain: stable
       - uses: Swatinem/rust-cache@v1
+      - name: Install Forc
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: --debug --path ./forc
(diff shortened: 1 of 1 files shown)


mitchmindtree (approved):
LGTM!

I think this makes much more sense than committing `out/debug/<proj>-abi.json` files, as at least this way we can always make sure the ABI files for the examples are up-to-date.

It's annoying there isn't a simple way to re-use a sequence of steps without writing a whole github action. Maybe one day github will give us "function"s, or let jobs "branch" out to other jobs or something. Meanwhile this is probably fine.

I would be more worried about the `cargo install` of `forc` taking up more CI time, but the rust-cache should make this pretty negligible.

Issue #1058, opened 2022-03-29
Migrate the `./sway-lib-std/tests` into the E2E `stdlib` directory and get them all working


mitchmindtree:
*Follow-up to #1052.*

Previously under the `sway-lib-std` repo the tests within the `tests` directory [weren't actually run under CI](https://github.com/FuelLabs/sway-lib-std/blob/dc036471ced326e41f076700c0501efc1b442455/.github/workflows/ci.yml#L19-L48). Now that we've moved them into the `sway` repo, it would be nice if we could migrate these tests into the existing E2E test harness, I'd imagine under `should_pass/stdlib` might make sense.

@nfurfaro would you be interested in having a go at this? I haven't yet looked closely at how the `sway` E2E harness and the `sway-lib-std` harnesses differ just yet, but perhaps you're already a little more familiar and might already have an idea on how to go about this?

nfurfaro:
@mitchmindtree I could work on this, yes.
I think it's worth discussing the best approach, as the 2 test harnesses are quite a bit different.
FWIW, I was actually thinking about migrating the remaining stdlib tests out of the e2e harness and into the stdlib one (there aren't too many left as I've already ported some).

It would be good to have @sezna or @otrho chime in here as well, but I quite like the stdlib harness for working with contracts and the SDK. The e2e harness is not currently adequate for testing contracts (unless it has changed since I last worked with it) .

Could we keep the tests separate, and have CI run them both?

mitchmindtree:
> Could we keep the tests separate, and have CI run them both?

I think I'd be fine with either approach as long as it's all tested via CI.

That said, if the E2E test suite is lacking w.r.t. contract testing, maybe that's something that could be improved there? I believe there are a few contract tests in there though, so it might have indeed evolved a little since you last worked with it.

Also of note: the tests were recently split into `should_pass` and `should_fail`, each with sub-directories for the different kinds of tests (e.g. `stdlib`, `language`, `forc`, `test_contracts`, etc), so if we do have the std lib tests separate it would be a bit of an outlier.

otrho:
There's both #685 and #723 (and others probably) which are both asking these questions.  Contract tests definitely need improving and if the lib harness can address that it'd be great.

mitchmindtree:
> I'm wondering if we can start by getting the tests under sway-lib-std/tests running under CI as a first step

[...]

PR #1328, opened 2022-04-20
Migrate std-lib tests

nfurfaro:
Testing out Plan A: https://github.com/FuelLabs/sway/issues/1058#issuecomment-1104120265

If we like this direction,  a follow-up issue could be to migrate all stdlib tests (there aren't many) out of the e2e test harness.

closes: #1058


JoshuaBatty (changes requested):
There seem to be a few .DS_Store files accidentally committed, can these be removed...

nfurfaro:
@JoshuaBatty for sure! I thought I deleted all of those, will do asap.

sezna (dismissed):
LGTM, ds store should probably be in the gitignore though...

nfurfaro:
>ds store should probably be in the gitignore though...
I've added it to my local .gitignore_global so I don't make that mistake again. Maybe a separate PR can add it to the sway .gitignore

JoshuaBatty (approved):

JoshuaBatty (approved):

Issue #1239, opened 2022-04-13
`forc test` should check Rust version


adlerjohn:
Running `forc test` should check that
1. Rust exists, and
2. The version of Rust matches some supported version. Currently only 1.58 or above. Maybe warn if it's above.

SilentCicero:
Agreed, this is a great idea.

SilentCicero:
@adlerjohn what about a warning at `forc init` also?

adlerjohn:
Probably a good idea for that subcommand too.

mitchmindtree:
Looks like `cargo` supports this officially as of recently! Should hopefully make this a bit easier.

https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html#cargo-rust-version

We could possibly do something like:

1. Add our minimum supported Rust version to `forc`'s `Cargo.toml`.
2. Use `include_str!` to have `forc` include it's own cargo manifest under a `const` `&str`.
3. During `forc init`, use `toml` to deserialize the manifest and retrieve the `rust-version` field under the `[package]` table. Use the same entry under the test harness `Cargo.toml` generated by `forc init`.

That said, just because this min version works with `forc` doesn't mean it will with the SDK. Perhaps `fuels-rs` itself could expose its own min-version via a `const &str`, and then we could use that directly in `forc`? In the meantime, we should at least be sure to sync the minimum Rust version across all our repos and crates.

SilentCicero:
Yeah, it's likely best for package level requirement.

We could add something to test, but this is usually a better route as it's more higher level.

adlerjohn:
cc @digorithm for thoughts on how the SDK could provide some sort of guarantees around both Rust version and its own version compatibility

mitchmindtree:
I've opened #1345 with some related ideas that might help with this.

PR #1333, opened 2022-04-20
Add forc rust version check

ra0x3:
## Description:
This PR adds a check to `forc init` and `forc test` to ensure that Rust is installed

## Steps to test this PR:
- [ ] Try to `rustup install` and `rustup default` a few `rustc` versions
  - 1.57, 1.58, 1.60 
- [ ] Try to run `forc init` or `forc test` using some of these different versions
  - Note the warning when using a version > 1.58
- [ ] Update your `$PATH` to not include `cargo` (to simulate not having cargo installed), then try `forc init` and `forc test`
  - Note the error message

---
[Software Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552)
[Technical Design Template](https://app.asana.com/0/59792373528535/184709971311943)
[List of cached micro services](https://dub.duckduckgo.com/duckduckgo/ddg/wiki/Microservice-caching)
[Dealing with broken integration tests](https://app.asana.com/0/411841517318433/1200020138263066)

### DIFF
--- a/forc/src/cli/commands/test.rs
+++ b/forc/src/cli/commands/test.rs
@@ -1,4 +1,6 @@
 use crate::ops::forc_build;
+use crate::utils::check_rust_version;
+
 use anyhow::Result;
 use clap::Parser;
 use std::io::{BufRead, BufReader};
@@ -42,6 +44,9 @@ pub(crate) struct Command {
 
 pub(crate) fn exec(command: Command) -> Result<()> {
     // Ensure the project builds before running tests.
+
+    check_rust_version()?;
+
     forc_build::build(Default::default())?;
 
     let mut cmd = process::Command::new("cargo");
--- a/forc/src/utils/mod.rs
+++ b/forc/src/utils/mod.rs
@@ -1,8 +1,56 @@
 pub mod defaults;
 pub mod parameters;
 
+use anyhow::Result;
+use forc_util::{println_red_err, println_yellow_err};
+use rustc_version::{version, Version};
+use std::fs::File;
+use std::io::Read;
+use std::path::Path;
+
 /// The `forc` crate version formatted with the `v` prefix. E.g. "v1.2.3".
 ///
 /// This git tag is used during `Manifest` construction to pin the version of the implicit `std`
 /// dependency to the `forc` version.
 pub const SWAY_GIT_TAG: &str = concat!("v", clap::crate_version!());
+
+pub(crate) fn forc_cargo_toml_as_str() -> Result<String> {
+    let cargo_dir = env!("CARGO_MANIFEST_DIR");
+    let cargo_file = format!("{}/Cargo.toml", cargo_dir);
+    let toml_path = Path::new(&cargo_file);
+
+    let mut file = File::open(toml_path)?;
+    let mut toml = String::new();
+    file.read_to_string(&mut toml)?;
+
+    Ok(toml)
+}
+
+pub(crate) fn check_rust_version() -> Result<()> {
+    let toml = forc_cargo_toml_as_str()?;
+    let rustc_version = match version() {
+        Ok(v) => v,
+        Err(_e) => {
+            println_red_err("rustc was not found in this environment.\n\nPlease see https://www.rust-lang.org/tools/install for more details on how you can install rustc.");
+            std::process::exit(0x01);
+        }
+    };
+
+    let cargo_toml: toml::Value = toml::de::from_str(&toml)?;
+
+    if let Some(table) = cargo_toml.as_table() {
+        if let Some(version) = table.get("package").unwrap().get("rust-version") {
+            let version_str = &version.as_str().unwrap();
+            let forc_rustc_version = Version::parse(version_str)?;
+            if rustc_version > forc_rustc_version {
+                let warning = format!(
+                    "\nFound rustc version {}, which is greater than the suggested version {}\n",
+                    &rustc_version, &forc_rustc_version
+                );
+                println_yellow_err(&warning);
+            }
+        }
+    }
+
+    Ok(())
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,6 +1077,7 @@ dependencies = [
  "fuel-vm",
  "futures",
  "hex",
+ "rustc_version 0.2.3",
  "serde",
  "serde_json",
  "sway-core",
--- a/forc/Cargo.toml
+++ b/forc/Cargo.toml
@@ -7,6 +7,7 @@ homepage = "https://fuel.network/"
 license = "Apache-2.0"
 repository = "https://github.com/FuelLabs/sway"
 description = "Fuel Orchestrator."
+rust-version = "1.58.0"
 
 [lib]
 name = "forc"
@@ -43,6 +44,7 @@ url = "2.2"
 uwuify = { version = "^0.2", optional = true }
 walkdir = "2.3"
 whoami = "1.1"
+rustc_version = "0.2"
 
 [features]
 default = []
--- a/forc/src/ops/forc_init.rs
+++ b/forc/src/ops/forc_init.rs
@@ -1,5 +1,5 @@
 use crate::cli::InitCommand;
-use crate::utils::defaults;
+use crate::utils::{check_rust_version, defaults};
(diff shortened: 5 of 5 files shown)


ra0x3:
Main things I want feedback on are:
- Are these changes in the proper place?
- Are the semantics/style similar to the team style?
- How would we test something like this? Since it's using `forc/Cargo.toml`

So on and so forth...

forc/src/utils/mod.rs
```diff
@@ -1,8 +1,56 @@
...
+use std::path::Path;
+
 /// The `forc` crate version formatted with the `v` prefix. E.g. "v1.2.3".
 ///
 /// This git tag is used during `Manifest` construction to pin the version of the implicit `std`
 /// dependency to the `forc` version.
 pub const SWAY_GIT_TAG: &str = concat!("v", clap::crate_version!());
+
+pub(crate) fn forc_cargo_toml_as_str() -> Result<String> {
+    let cargo_dir = env!("CARGO_MANIFEST_DIR");
```
ra0x3, line 18: Do we have a preferred way of doing something like this?

forc/src/utils/mod.rs
```diff
@@ -1,8 +1,56 @@
...
+        Err(_e) => {
+            println_red_err("rustc was not found in this environment.\n\nPlease see https://www.rust-lang.org/tools/install for more details on how you can install rustc.");
+            std::process::exit(0x01);
+        }
+    };
+
+    let cargo_toml: toml::Value = toml::de::from_str(&toml)?;
+
+    if let Some(table) = cargo_toml.as_table() {
+        if let Some(version) = table.get("package").unwrap().get("rust-version") {
```
ra0x3, line 42: Since this reads the forc/Cargo.toml can we always expect that to be present and well-formed (i.e. not malformed)?

mitchmindtree (changes requested):
:tada: nice job on the first PR!

---

I think the primary motivation behind #1239 is to ensure that the user is aware of the minimum rustc version required for compatibility with `forc test`.

The reason why this is particularly relevant to `forc test` is that `forc test` actually uses `cargo test` under the hood to run tests on the compiled `sway` program via a Rust harness. It's important that the user's rust version is *at least* as new as what was used to build `forc` itself to make sure that the user's `rustc` version is capable of all the recent, fancy language features that we take advantage of in the harness code and its dependencies.

This harness is generated by `forc init` which includes writing the harness' `Cargo.toml` manifest file. I think we should update `utils::defaults::default_tests_manifest` to also include a matching `rust_version`. This should allow us to remove the `check_rust_version` function altogether, as `cargo` should automatically inform the user if the installed version is older than `rust_version`. Are you up for making this change?

forc/src/cli/commands/test.rs
```diff
@@ -41,6 +44,11 @@ pub(crate) struct Command {
 }
 
 pub(crate) fn exec(command: Command) -> Result<()> {
+    if let Err(e) = check_rust_version() {
+        println_red_err(&e.to_string());
+        std::process::exit(0x01);
```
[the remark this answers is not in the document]
mitchmindtree: Thinking further, we might not need this check - see [the top-level review comment](https://github.com/FuelLabs/sway/pull/1333#pullrequestreview-948166544).

ra0x3:
@mitchmindtree  Thanks for the detailed feedback 🙏 . Changes made. You might have to clue me in on how to pass the following CI tests though:
- cargo-toml-fmt-check (I do run `cargo fmt` before committing)
- cargo-clippy (I similarly run `cargo clippy` before committing)

forc/src/utils/mod.rs
```diff
@@ -1,8 +1,47 @@
...
+                        &local_rustc_version, &forc_rustc_version
+                    );
+                    println_yellow_err(&warning);
+                }
+            }
+        }
+    }
+
+    Ok(local_rustc_version)
+}
```
mitchmindtree, line 47: The idea behind my suggestion of abstracting out a `forc_rustc_version` is to separate the concerns of:

1. retrieving forc's Rust version and
3. checking that we have a valid Rust version.

E.g. 

```rust
pub(crate) fn forc_rustc_version() -> Result<Version> {
    // Just determines forc's rustc version. Does *not* check against installed rust version.
}
```

I think we can actually get rid of the version **checking** part completely, as Cargo should automatically do this for us when `forc test` runs `cargo test` under the hood and `cargo` inspects the `rust-version` field within the test harness' `Cargo.toml`. This is what I was attempting to explain in [this comment](https://github.com/FuelLabs/sway/pull/1333#pullrequestreview-948166544).

What we *should* be using `forc_rustc_version` function for is to retrieve the version that should be used in the test harness' `Cargo.toml` that is generated by `forc init`.

forc/src/utils/mod.rs
```diff
@@ -1,8 +1,47 @@
...
+    let mut file = File::open(toml_path)?;
+    let mut toml = String::new();
+    file.read_to_string(&mut toml)?;
+
+    let cargo_toml: toml::Value = toml::de::from_str(&toml)?;
+
+    if let Some(table) = cargo_toml.as_table() {
+        if let Some(project) = table.get("project") {
+            if let Some(v) = project.get("rust-version") {
+                let vs = &v.as_str().unwrap();
```
mitchmindtree, line 33: Should return an error here rather than `unwrap`. E.g. something like:

```rust
let vs = v.as_str().ok_or_else(|| anyhow!("rust-version was not a string"))?;
```

forc/src/utils/defaults.rs
```diff
@@ -28,6 +29,7 @@ edition = "2021"
 license = "Apache-2.0"
 name = "{project_name}"
 version = "0.1.0"
+rust-version = "1.58.0"
```
mitchmindtree, line 32: We should not hard-code the version here, but instead use the `forc_rustc_version` function to retrieve the compatible version.

mitchmindtree (changes requested):

ra0x3:
@mitchmindtree The forbearance is appreciated 😅 . 
- I think my confusion came from misunderstanding Forc.toml vs Cargo.toml
- Now that I see we're only adding `rust-version` to Cargo.toml - now I see the no need to manually check versions
- These little one-liner code optimizations are very much appreciated 🙏 

I do have a question however. I know that `forc` essentially `cargo`, but why do we use two manifest files? What differential value does Cargo.toml provide vs Forc.toml?

SilentCicero:
> 

I think the Cargo toml is tightly designed for specifically Rust packages. It's not something we can just add arbitrary fields do for us with Forc. Also Forc is suppose to be somewhat and eventually independent of the language Rust, for example you could write your tests in Typescript or Sway.

Having our own Forc.toml also allows for tight integration with our own cryptographic package manager, and its custom logic.

forc/src/utils/mod.rs
```diff
@@ -1,8 +1,16 @@
...
+use rustc_version::{version, Version};
+
 /// The `forc` crate version formatted with the `v` prefix. E.g. "v1.2.3".
 ///
 /// This git tag is used during `Manifest` construction to pin the version of the implicit `std`
 /// dependency to the `forc` version.
 pub const SWAY_GIT_TAG: &str = concat!("v", clap::crate_version!());
+
+pub(crate) fn forc_rustc_version() -> Result<Version> {
+    let version = version().map_err(|e| anyhow!("Failed to locate rustc: {}", e))?;
```
mitchmindtree, line 14: Sorry I should have clarified - the idea here was to keep the logic where we retrieved the rust version from `forc`'s `Cargo.toml` file, as this is the best indicator of what Rust version will remain compatible with the test harness generated by `forc`. I think we can probably remove the `rustc_version` crate dependency altogether.

mitchmindtree (changes requested):
No stress at all, I think this is a particularly confusing task as there are multiple contexts of manifests and Rust instances to distinguish between and consider!

forc/src/utils/defaults.rs
mitchmindtree: My comment here still stands. It looks like you've removed the `forc_rustc_version` function, but that's what we should be using to determine what version to use in this field.

mitchmindtree (changes requested):

PR #1339, opened 2022-04-21
Remove standard library dependency from example

SilentCicero:


adlerjohn (changes requested):
This is undesired behavior. Removing the explicit std dependency means it'll use latest `master` rather than local, i.e. CI won't test this example.

mitchmindtree:
I do agree it's a pity to have this in the book, but it's probably more important that this remains tested against master by CI :thinking:

Perhaps we could use a "range" or "anchor" somehow in the mdbook [pre-processor line](https://github.com/FuelLabs/sway/blob/505984dd97ea40519445069ba45c838699f22aa9/docs/src/introduction/forc_project.md?plain=1#L25) to ignore the last part of this (similar to as mentioned [here](https://github.com/FuelLabs/sway/pull/1278#discussion_r852600309))?

PR #1226, opened 2022-04-12
Remove `std` from the list of dependencies in the default Forc.toml

mohammadfawaz:

PR #1348, opened 2022-04-21
Remove explicit `std` in getting started example

adlerjohn:


docs/src/introduction/overview.md
```diff
@@ -85,7 +85,6 @@ license = "Apache-2.0"
 name = "wallet_contract"
 
 [dependencies]
```
mohammadfawaz, line 87: We should probably remove `[dependencies]` to match the behaviour of `forc init`.

SilentCicero (dismissed):

docs/src/introduction/overview.md
adlerjohn: `forc init` has a `[dependencies]`, as does `cargo init`.

mohammadfawaz: What's actually bad is that we actually discussed this in my PR https://github.com/FuelLabs/sway/pull/1226 10 days ago.
image

mohammadfawaz (approved):

PR #1351, opened 2022-04-21
Updated `forc-init` to accept `program_type` parameters

### LINKED ISSUE #1186: Introduce `--contract` and `--script` to `forc init`
`$ forc init --script` -> create an empty script. Question: is a Rust test harness required with this?
`$ forc init --contract` -> create two Sway files: one for the ABI and one for its implementation. Also create a test harness populated with boilerplate code. 
We could also do `--predicate` and `--library`.

The default should probably be a contract.

eureka-cpu:
Closes #1186 

This adds program type fields to the `InitCommand` struct making it possible to create different program types via `forc init --<program_type> <project_name>`. I've updated `sway-utils::constants` with two `entry_types`: `MAIN_ENTRY` & `LIB_ENTRY` so that `Forc.toml` uses the correct `entry` name and updated `forc::utils::defaults` to have a default program for each `program_type`. I add a file, `forc::utils::program_type` that contains the struct `ProgramType` to match on in `forc_init`.


forc/src/ops/forc_init.rs
```diff
@@ -82,6 +82,10 @@ fn print_welcome_message() {
 pub fn init(command: InitCommand) -> Result<()> {
     let project_name = command.project_name;
     validate_name(&project_name, "project name")?;
+    let project_type = match command.project_type {
```
sezna, line 85: [structopt has an option for a default value](https://docs.rs/structopt/latest/structopt/#how-to-derivestructopt).

You can [also make the lowercasing part of structopt as well](https://docs.rs/structopt/latest/structopt/#custom-string-parsers)

eureka-cpu:
@adlerjohn currently this command would look like: `forc init <project_name> <project_type>` and wouldn't need the `--`. Personally I think this is cleaner for internal use, and for anyone creating a `contract` you could leave out the `project_type` bit. We could also make it a flag, and this would look like `... --project-type <project_type>`.

Braqzen:
How about the order is changed to have the project name be at the end?
E.g. `forc init <project name>` defaults to a contract and for a different program type the program type would be put before the project name but after the init i.e. `forc init <program type> <project name>`.

eureka-cpu:
> How about the order is changed to have the project name be at the end? E.g. `forc init <project name>` defaults to a contract and for a different program type the program type would be put before the project name but after the init i.e. `forc init <program type> <project name>`.

This was the initially approach, but unfortunately `clap` has trouble determining which is the name of the project and which is the project type. cc @sezna for a more detailed explanation than I could give.

adlerjohn:
Since program type is optional, it should be at the end.

eureka-cpu:
> Since program type is optional, it should be at the end.

Should it be an Option, or an Arg?

mitchmindtree:
Perhaps it might be nicer to match `cargo new` behaviour here?

Some examples:

- `cargo new --lib my-lib` - creates a library called `my-lib`
- `cargo new --bin my-app` - creates a binary called `my-app`
- `cargo new my-app` - default behaviour creates a binary called `my-app`

When discussing support init with program types, I was imagine we'd have something similar for forc. For example:

- `forc init --contract my-contract`
- `forc init --library my-lib`
- `forc init --script my-script`
- `forc init --predicate my-pred`
- `forc init my-contract`

Perhaps this is a little more concise / approachable?

I think we could achieve this behaviour by adding these flags as `bool`s to the `InitCommand`, but I could be wrong. Then we can do something like:

```rust
let prog_type = match (cmd.contract, cmd.library, cmd.script, cmd.predicate) {
    (false, true, false, false) => /* library */,
    (false, false, true, false) => /* script */,
    (false, false, false, true) => /* predicate */,
    (_, false, false, false) => /* contract */,
    _ => /* return err: can only pick one program type */,
};
```

### project type -> program type

Much of the terminology in this PR and discussion has been around "project-type", however in the Sway book I believe we use the term "**program** type". I'd be open to seeing the terminology used in the book bike-shedded/discussed, but in this PR we should probably be consistent and follow the terminology that currently exists in the book.

adlerjohn:
My personal preference is to always imitate Rust and Cargo unless there's an explicit reason not to.

forc/src/ops/forc_init.rs
```diff
@@ -82,6 +85,21 @@ fn print_welcome_message() {
...
+        command.predicate,
+        command.library,
+    ) {
+        (_, false, false, false) => Contract,
+        (false, true, false, false) => Script,
+        (false, false, true, false) => Predicate,
+        (false, false, false, true) => Library,
+        _ => anyhow::bail!(
+            "Unrecognized program type, or multiple types detected: \
+        \n Possible Types:\n - contract\n - script\n - predicate\n - library"
```
adlerjohn, line 100: ?
```suggestion
        \n Possible Types:\n - contract\n - script\n - predicate\n - lib"
```

adlerjohn (commented):
Looks fine to me but I'll defer to others.

forc/src/ops/forc_init.rs
```diff
@@ -111,22 +127,50 @@ pub(crate) fn init_new_project(project_name: String) -> Result<()> {
...
         Path::new(&project_name).join(constants::TEST_MANIFEST_FILE_NAME),
         defaults::default_tests_manifest(&neat_name),
     )?;
 
-    // Insert default main function
-    fs::write(
-        Path::new(&project_name).join("src").join("main.sw"),
-        defaults::default_program(),
-    )?;
+    // Insert project based on project_type
```
mitchmindtree, line 147:
```suggestion
    // Insert src based on program_type
```

mitchmindtree (changes requested):
Great stuff! One last nit, otherwise utACK :+1:

forc/src/cli/commands/init.rs
```diff
@@ -13,6 +13,18 @@ pub struct Command {
...
+    #[clap(long)]
+    pub contract: bool,
+    /// Adding this flag creates an empty script program.
+    #[clap(long)]
+    pub script: bool,
+    /// Adding this flag creates an empty predicate program.
+    #[clap(long)]
+    pub predicate: bool,
+    /// Adding this flag creates an empty library program.
+    #[clap(long = "lib")]
```
mitchmindtree, line 26: It feels a little odd only having a short-hand name for the `library` type but not for the others. Maybe we can omit this shorthand for now and discuss whether or not we want short-hands in another issue?

forc/src/ops/forc_init.rs
```diff
@@ -82,6 +85,21 @@ fn print_welcome_message() {
...
+        command.predicate,
+        command.library,
+    ) {
+        (_, false, false, false) => Contract,
+        (false, true, false, false) => Script,
+        (false, false, true, false) => Predicate,
+        (false, false, false, true) => Library,
+        _ => anyhow::bail!(
+            "Multiple types detected, please specify only one program type: \
+        \n Possible Types:\n - contract\n - script\n - predicate\n - lib"
```
mitchmindtree, line 100: If we remove the shorthand for library, we should update this msg too.

mitchmindtree (approved):

PR #1352, opened 2022-04-21
Update readme for stdlib tests

nfurfaro:
Updated the location of the build script and linked to the docs for `forc test`

### DIFF
--- a/test/src/sdk-harness/README.md
+++ b/test/src/sdk-harness/README.md
@@ -9,36 +9,18 @@ $ cd sway
 $ cargo install --path ./forc
 ```
 
-In the root of the `sway-lib-std` is a bash build script. To run it:
+In the `sway/test/src/sdk-harness` directory is a bash build script. To run it:
 
 ```sh
-$ cd sway-lib-std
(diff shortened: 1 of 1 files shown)


Braqzen (changes requested):
I would add in the `$` at the start of each command namely lines 8, 9, 15, 16 and 23 to make a distinction between a command and comment. It might be considered pedantic but I think consistency and convention are good to follow

adlerjohn:
@Braqzen the book uses `sh` for commands with no output and `console` with `$` for commands + their output. I'm not sure where comments come into play?

nfurfaro:
@Braqzen In this case, using markdownlint to format the doc stripped out all the `$` symbols.

https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint

Braqzen (approved):

otrho:
Another thing to consider is if we are happy with users copy-n-pasting the examples into their terminal.  The `$` prevents this from being easy, which could be good or bad.

PR #1353, opened 2022-04-21
Enable ^, << & >> and add  new BitwiseOps traits.

nfurfaro:
Previously, we had an impl self( `impl u64` ) where the `binary_and`, `binary_or` and `binary_xor` functions were defined.
This PR improves on this by:
- adding 3 new traits `pub trait BitwiseAnd` , `pub trait BitwiseOr`  and `pub trait BitwiseXor` 
- refactoring to impl these traits for `u64`
- I've also enabled the `^`, `<<` and `>>` operators and added tests for them.
- this also reenables the `u64_ops` tests which were recently disabled with the new parser merge (They work when using the `&`, `|` and `^` operators).

This sets us up nicely to implement bitwise ops and Shiftable for b256.


sezna:
`Andorable` is very close to the word _adorable_, which is comedic.
 
It might be tedious but it could be worth it to split these out into separate traits: `BitwiseAnd`, `BitwiseOr`, and `BitwiseXor`?

nfurfaro:
> it could be worth it to split these out into separate traits

Yeah, I can do that, not that tedious really.

otrho (approved):
LGTM.

Issue #1347, opened 2022-04-21
Hide normal lockfile creation reason behind verbosity flag


adlerjohn:
Running `forc build` on a new project creates a lockfile and emits a scary cause that's entirely expected. The cause should be hidden behind a verbosity flag, or removed entirely.

```console
$ forc build
  Creating a new `Forc.lock` file
    Cause: failed to read /home/john/tmp/foo/Forc.lock: No such file or directory (os error 2)
```

cc @mitchmindtree for thoughts.

mitchmindtree:
Yeah totally agreed, nice job on opening this. I think I've already seen one or two peeps mistake this for an error.

Perhaps even having a more concise short-form (rather than the raw error description) but on the same line might be nicer? e.g.

```
$ forc build
  Creating a new `Forc.lock` file (cause: lock file does not yet exist)
```

Issue #1201, opened 2022-04-08
Library crates should prefer `thiserror` to `anyhow` for error handling


mitchmindtree:
The primary reason to prefer `thiserror` to `anyhow` for libraries is because `thiserror` enables users of the library to handle different types of error conditions.

The `thiserror` crate is aimed at making it trivial to declare a set of custom `*Error` `struct`s and `enum`s. Enums are particularly useful for this as they allow downstream users of your library to exhaustively check error conditions in order to handle them correctly and with confidence that they have all cases covered.

On the other hand, `anyhow` is aimed at making error handling quick and easy for application crates, and does not at all make it easy for devs to actually handle certain conditions without inspecting the contents of strings.

Edit: I'll add that both `thiserror` and `anyhow` were published by the same author around the same time, and that this use-case distinction is intentional. Both are useful, but should be used for their particular use-cases.

eureka-cpu:
Closing in favor of #4817

Issue #884, opened 2022-03-06
Consider using a more flexible logging crate in `forc` (and throughout sway) over `println`


mitchmindtree:
This should help in providing different levels of verbosity in `forc` and any other sway-related CLI tooling.

We could probably get away with something simple like `log`, but may also want to consider something that has some profiling integration so we can time passes and track compiler performance #318 (I think maybe `trace` does something like this?)

JoshuaBatty:
Was just taking a look at `rust-analyzer` and they are using the `tracing` crate for logging which is made by the `tokio` team. It has compatibility with `log` through the `tracing-log` crate. 

One main benefit to using `tracing` over `log` is we can add logs for `Spans` and `Events` on functions that return a `Future`. There also seems to be quite a rich ecosystem of other crates extending the functionality of `tracing` for doing things like timing-profiling etc..

Voxelot:
+1 to tracing. We already use tracing in sdk and fuel-core as per the code guidelines here: https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/logging.md

Voxelot:
Although, I just discovered this other library which might be useful for more interactive cli usecases: https://github.com/crossterm-rs/crossterm

mitchmindtree:
Nice, `crossterm` does look super powerful! It might be a little overkill for now as we don't have any interactive CLI stuff going on just yet, but will be good to keep in mind.

For now, starting with integrating `tracing` sounds like a great approach.

kayagokalp:
I started to integrate `tracing` into forc and I was curious about a couple of details of the desired behavior. 

We use `anyhow::Result<()>` as the return type throughout the repo. As I understand the error message is printed automatically when the main also returns an `anyhow::Result<()>`. Do we want to keep this behavior? I think we may want to switch errors to `tracing` as well. 

As for implementation, I am currently leaning towards consuming that error in `main` (or to be more general in the caller function) and logging a generic error with `tracing` (the details of the error will be already logged in the failing calls).

For example, if an error happens at [forc_init.rs:140](https://github.com/FuelLabs/sway/blob/a34792b696dac1c70984c8bca66b7412fd0482b2/forc/src/ops/forc_init.rs#L140) the output would be something like:

(output of: `cargo run --bin forc -- init --template er name`)

```
[...]

PR #1354, opened 2022-04-21
Hide Forc.lock creation warn behind verbosity

ra0x3:
### Description
This PR swaps out the the scary "File does not exist" warning for a "lock file did not exist" warning when creating a new lock file

---

### Testing steps
- [ ] `forc build` without a `Forc.lock`

### DIFF
--- a/forc/src/ops/forc_build.rs
+++ b/forc/src/ops/forc_build.rs
@@ -72,7 +72,9 @@ pub fn build(command: BuildCommand) -> Result<pkg::Compiled> {
         let string = toml::ser::to_string_pretty(&lock)
             .map_err(|e| anyhow!("failed to serialize lock file: {}", e))?;
         fs::write(&lock_path, &string).map_err(|e| anyhow!("failed to write lock file: {}", e))?;
-        println!("   Created new lock file at {}", lock_path.display());
+        if !command.silent_mode {
+            println!("   Created new lock file at {}", lock_path.display());
+        }
         Ok(plan)
     })?;
 
--- a/forc/src/ops/forc_build.rs
+++ b/forc/src/ops/forc_build.rs
@@ -57,8 +57,14 @@ pub fn build(command: BuildCommand) -> Result<pkg::Compiled> {
 
     // If necessary, construct a new build plan.
     let plan: pkg::BuildPlan = plan_result.or_else(|e| -> Result<pkg::BuildPlan> {
-        println!("  Creating a new `Forc.lock` file");
-        println!("    Cause: {}", e);
+        if !command.silent_mode {
+            let cause = if e.to_string().contains("No such file or directory") {
+                anyhow!("lock file did not exist")
+            } else {
+                e
+            };
+            println!("  Creating a new `Forc.lock` file (cause: {})", cause);
+        }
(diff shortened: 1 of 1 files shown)


forc/src/ops/forc_build.rs
```diff
@@ -57,16 +57,24 @@ pub fn build(command: BuildCommand) -> Result<pkg::Compiled> {
 
     // If necessary, construct a new build plan.
     let plan: pkg::BuildPlan = plan_result.or_else(|e| -> Result<pkg::BuildPlan> {
-        println!("  Creating a new `Forc.lock` file");
-        println!("    Cause: {}", e);
+        if !command.silent_mode {
```
ra0x3, line 60: The issue mentioned a `verbose` flag. I didn't see that but saw this `silent_mode` flag which I'm assuming is orthogonal to `verbose`

forc/src/ops/forc_build.rs
```diff
@@ -57,16 +57,24 @@ pub fn build(command: BuildCommand) -> Result<pkg::Compiled> {
 
     // If necessary, construct a new build plan.
     let plan: pkg::BuildPlan = plan_result.or_else(|e| -> Result<pkg::BuildPlan> {
-        println!("  Creating a new `Forc.lock` file");
```
mitchmindtree, line 60: I believe it's only the `Cause` line that we want to hide - I think we should always *at least* show this line to inform the user that the lock file was created.

The motivation behind #1347 is that the `Cause` line could often look a bit scary and lead the user to believe something went wrong, when really the line just describes *why* the lock file was created (or recreated).

forc/src/ops/forc_build.rs
mitchmindtree: I think we can omit the hiding behaviour altogether in favour of moving the "cause" onto the same line as you have done below.

forc/src/ops/forc_build.rs
```diff
@@ -57,16 +57,24 @@ pub fn build(command: BuildCommand) -> Result<pkg::Compiled> {
...
-        println!("    Cause: {}", e);
+        if !command.silent_mode {
+            let cause = if e.to_string().contains("No such file or directory") {
+                anyhow!("lock file did not exist")
+            } else {
+                e
+            };
```
mitchmindtree, line 65: I think this is fine for now, however there is a whole range of possible errors beyond the "No such file or directory" error so eventually we should remove this special case that only handles lock file creation in favour of a more structured approach for creating a short-form cause message. This is the kind of thing that #1201 would help a lot with.

mitchmindtree (changes requested):

forc/src/ops/forc_build.rs
mitchmindtree: Also w.r.t. `silent_mode` flag - I believe this is more about silencing output altogether (I think a `quiet` flag is a more common name in unix tooling). That said, I'm not sure we actually abide by this flag much at all yet. We should revisit the flag when we have a nicer approach to logging #884.

mitchmindtree (approved):

adlerjohn (approved):
Update the PR description to reflect the latest design. Good work!

PR #1355, opened 2022-04-21
match expressions take 2

### LINKED ISSUE #1263: Document `match` and `if let`
As per the title, `match` and `if let` should be documented in the Sway Book.

SilentCicero:
Closes #1263
- take 2
- with comments
- example in its own directory
- formatting


examples/match_statements/src/main.sw
```diff
@@ -0,0 +1,26 @@
...
+    };
+
+    // Match as an expression.
+    let b = match x {
+        5 => { 42 },
+        _ => { 24 },
+    };
+
+    // Match as control flow.
+    match 42 {
```
sezna, line 21: This is also an expression since it returns its value. A more "pure" control flow example could be:

```
fn main() {
  let boolean = true;
  match boolean {
    true => {
        foo();
    },
    false =>  {
      bar();
    }
  };
}

fn foo() {
  // do something
}

fn bar() {
  // do_something
}
```

because the returned value from `match` is truly never used and is therefore only used for control flow

SilentCicero: Ah, yeah you are right. I'll revise.

adlerjohn (commented):
Looks fine to me. Will defer to @sezna

adlerjohn (commented):
Update from `master` to pick up changes from #1384

sezna:
@SilentCicero mind fixing up the CI stuff?

mohammadfawaz:
This should be review for final review now as well.

adlerjohn (approved):

PR #1359, opened 2022-04-21
Swap core dep for implicit-std

ra0x3:
### Description

This PR swaps the `core` dependency for an `implicity-std = false` setting in all e2e test manifest files

### Testing steps
- [ ] `cargo test` should pass


mitchmindtree (approved):
Nice one!

PR #1360, opened 2022-04-21
Replace `hello_world` example with `counter` example

### LINKED ISSUE #1361: `forc init --template counter` uses the `examples/hello_world` URL
`forc init --template counter` uses the `examples/hello_world` URL, also `counter` & `hello_world` are duplicate programs.

eureka-cpu:
Closes #1361

I've removed the `hello_world` from `examples` since it is a duplicate of the `counter` example, and updated the URL in `forc_init` to `examples/counter` instead of `examples/hello_world`.


JoshuaBatty:
Nice one. There are also a few references in the book that will need updating as well. 

[here](https://github.com/FuelLabs/sway/blob/4401800b63b590862de4ea4d9321b8eb2264ae35/docs/src/introduction/forc_project.md#L25) and [here](https://github.com/FuelLabs/sway/blob/a39ec9507ed1eecf0fb4b05ae44d01d87953adb4/docs/src/testing/testing-with-rust.md#L24)

eureka-cpu:
> Nice one. There are also a few references in the book that will need updating as well.
> 
> [here](https://github.com/FuelLabs/sway/blob/4401800b63b590862de4ea4d9321b8eb2264ae35/docs/src/introduction/forc_project.md#L25) and [here](https://github.com/FuelLabs/sway/blob/a39ec9507ed1eecf0fb4b05ae44d01d87953adb4/docs/src/testing/testing-with-rust.md#L24)

Should these be updated with `counter`, or should we keep `hello_world` specifically for these examples? @adlerjohn @SilentCicero

JoshuaBatty:
I'd say we should update to reference the counter example instead since the code is identical.

Braqzen (dismissed):
Looking over the code it seems like there aren't as many tests as I thought. I was expecting a suite to exist somewhere which would have been updated too.
Nevertheless, simpler fix than I thought and the addition of tests / expanding the template to use the other repo examples can be another issue

sezna:
[It looks like some markdown in the docs needs to be updated.](https://github.com/FuelLabs/sway/runs/6120986904?check_suite_focus=true#step:4:13)

sezna (changes requested):
[It looks like some markdown in the docs needs to be updated.](https://github.com/FuelLabs/sway/runs/6120986904?check_suite_focus=true#step:4:13)

eureka-cpu:
> Nice one. There are also a few references in the book that will need updating as well.
> 
> [here](https://github.com/FuelLabs/sway/blob/4401800b63b590862de4ea4d9321b8eb2264ae35/docs/src/introduction/forc_project.md#L25) and [here](https://github.com/FuelLabs/sway/blob/a39ec9507ed1eecf0fb4b05ae44d01d87953adb4/docs/src/testing/testing-with-rust.md#L24)

So the first one I think we can actually leave how it is, since it's not using the `hello_world` example, it's actually just a walk through of how to initialize a new project and just happens to use the name `hello_world`

sezna:
🚨 merge conflicts 🚨

eureka-cpu:
> 🚨 merge conflicts 🚨

weird, I had previously removed that file...

docs/src/forc/commands/forc_init.md
```diff
@@ -15,10 +15,27 @@ forc init [OPTIONS] <PROJECT_NAME>
 
 ## OPTIONS:
 
+`--contract`
```
adlerjohn, line 18: Why is this here?

adlerjohn (changes requested):

docs/src/forc/commands/forc_init.md
eureka-cpu: There's a bug with the documentor I think. If I manually change it back, and then run the documentor again, it fills this in from my other branch...? super weird

adlerjohn: You might be using `forc` built with your other branch??

eureka-cpu: Right you are 🤦🏻

sezna (approved):

Issue #1363, opened 2022-04-22
Broken CI - `forc-documenter` job failing - suddenly `forc` is outputting to `stderr` rather than `stdout`


mitchmindtree:
The `forc-documenter` script works by running `forc --help` to retrieve available commands, and then running `forc <cmd> --help` to generate the documentation. The CI job uses this to ensure `forc`'s documentation stays up to date.

To retrieve the output from the `forc --help` command, it captures and parses `stdout` (see [here](https://github.com/FuelLabs/sway/blob/7812a88d938a9719fcbd0e7e47d44a5d30c016bb/scripts/forc-documenter/src/main.rs#L127-L130)). Until this morning it appears this has been working without issue, however after attempting to debug #1362 I noticed that the `forc` invocations (via `std::process::Command`) are now outputting everything via `stderr` instead. Not sure why!  

To investigate:

- Did we pick up a `clap` dependency update recently that might have changed this?
- Are there other reports of `clap` apps suddenly outputting to `stderr` instead of `stdout`?
- Are there reports of GitHub action workflow job `stdout`/`stderr` switching?

Potential workaround that comes to mind is to just merge the `stdout` and `stderr` output and operate on that to make sure we've captured the output we need one way or another, however it would be nice to actually understand what's going on here.

cc @binggh 
cc @otrho for your deeper unix wizardry in case there's something obvious I'm missing!

binggh:
Seems like the issue is on clap's end with 3.1.11: https://github.com/clap-rs/clap/pull/3646

mitchmindtree:
Nice find! That was a mind-boggler :exploding_head: cc @SilentCicero ^

Will close after that lands and gets published.

mitchmindtree:
Looks like that patch was merged and clap `3.1.12` published yesterday, closing.

PR #1364, opened 2022-04-22
Fix CI - Concat `forc`'s stdout and stderr output in `forc-documenter`

mitchmindtree:
This is a temporary workaround for #1363 until we work out what's really going on there.


mitchmindtree:
cc @binggh

otrho (approved):

PR #1370, opened 2022-04-22
Added examples from examples repo

### LINKED ISSUE #1369: Add additional examples into `forc init --template`
Current examples

Braqzen:
Closes #1369 

Added all examples (except "build all examples") into `forc_init.rs`


adlerjohn (commented):
Rather than `master`, should this use the same tagged version as `forc` is?

Braqzen:
@eureka-cpu It's fine. I've seen your ticket which has a bunch more code changes so I thought that I'd just merge them in instead of changing the string although it wouldn't have caused any problems if I have updated it here myself. I've got a bit more to do anyway

forc/src/ops/forc_init.rs
```diff
@@ -85,13 +85,30 @@ pub fn init(command: InitCommand) -> Result<()> {
...
-                }
+                "counter" => Url::parse(&example_url)?,
+                "fizzbuzz" => Url::parse(&example_url)?,
+                "liquidity_pool" => Url::parse(&example_url)?,
+                "msg_sender" => Url::parse(&example_url)?,
+                "native_token" => Url::parse(&example_url)?,
+                "storage_example" => Url::parse(&example_url)?,
+                "subcurrency" => Url::parse(&example_url)?,
+                "wallet_smart_contract" => Url::parse(&example_url)?,
+                "asm_return_tuple_pointer" => Url::parse(&example_url)?,
```
adlerjohn, line 99: To avoid future headaches, alphabetize.

mitchmindtree: To keep the branching lower, this could also be written as:

```rust
                "counter"
                | "fizzbuzz"
                | "liquidity_pool"
                | "msg_sender"
                | "native_token"
                | "storage_example"
                | "subcurrency"
                | "wallet_smart_contract"
                | "asm_return_tuple_pointer" => Url::parse(&example_url)?,
```

mitchmindtree (commented):
I think this is fine as a temporary improvement until more general git / github support is added for templates.

For now, we could probably even remove the `match` altogether and just try:

```rust
let template_url = Url::parse(&example_url)?;
```

Then, in the case that the following call to `init_from_git_template` fails for whatever reason (e.g. 404 or other error), then we can give an error like:

```rust
init_from_git_template(project_name, &template_url)
    .map_err(|e| anyhow!("Failed to initialize project from a template with the given name "{template}": {e}.\n  Note: If you are attempting to initialize this project from a Sway example, please ensure the template name matches one of the available examples."))?;
```

Or something along these lines (formatted correctly). This would allow us to avoid adding/removing/editing example names while we await a more comprehensive solution for sourcing templates.

mitchmindtree:
Just a related note: I've opened https://github.com/FuelLabs/sway/issues/1380 with a rough design for a more general approach to template sources.

adlerjohn:
> remove the match altogether and just try

One downside of _not_ hardcoding examples _somewhere_ (pending better fancy stuff of course) is that there wouldn't be an immediate way for the user to get a list of templates to use.

eureka-cpu:
> > remove the match altogether and just try
> 
> One downside of _not_ hardcoding examples _somewhere_ (pending better fancy stuff of course) is that there wouldn't be an immediate way for the user to get a list of templates to use.

give a current list of templates when fetching fails?

Braqzen:
> 

If a user attempts to fetch the `counter` example but accidentally types in `counte` then the initial request will fail as such a repo does not exist.
This would then present the user with an error message prompting them to
1. Make sure they know what they are trying to fetch
2. Display a list of the current examples in addition to that message.

Option 2 may obviously fail and thus the displaying of the repos should be an optional step that is appended to the end of the regular error message.
This means that if they type in an incorrect example (or one that no longer exists) it will fetch all of the repo names so that they do not have to look it up themselves.

forc/src/ops/forc_init.rs
```diff
@@ -85,22 +108,67 @@ pub fn init(command: InitCommand) -> Result<()> {
...
+                Ok(()) => Ok(()),
+                Err(error) => {
+                    // TODO: change this to use anyhow!(), possibly without a match?
+                    let mut error_message = format!("Failed to initialize project from a template with the given name \"{template}\": {error}.\n  Note: If you are attempting to initialize this project from a Sway example, please ensure the template name matches one of the available examples.\n");
+                    let examples = get_sway_examples()?;
+                    for example in examples {
+                        error_message.push_str(format!("\t- {}\n", example).as_str());
+                    }
+                    println!("{}", error_message);
+                    Ok(())
```
mitchmindtree, line 132: Rather than

```rust
                    println!("{}", error_message);
                    Ok(())
```
I think this could be (untested):

```rust
                    bail!("{}", error_message)
```

forc/src/ops/forc_init.rs
```diff
@@ -85,22 +108,67 @@ pub fn init(command: InitCommand) -> Result<()> {
...
+            // Ok(())
+
+            // If the user queried an existing example then continue otherwise attempt to fetch the examples and append them
+            // to the end of the error message so that the user can see the existing examples to choose from
+            match init_from_git_template(project_name, &template_url) {
+                Ok(()) => Ok(()),
+                Err(error) => {
+                    // TODO: change this to use anyhow!(), possibly without a match?
+                    let mut error_message = format!("Failed to initialize project from a template with the given name \"{template}\": {error}.\n  Note: If you are attempting to initialize this project from a Sway example, please ensure the template name matches one of the available examples.\n");
+                    let examples = get_sway_examples()?;
```
mitchmindtree, line 127: If this fails, rather than returning with only the `get_sway_examples` error (likely some networking error), let's still notify about the failure to initialise the project, e.g.

```rust
if let Err(err) = get_sway_examples() {
    bail!("{}\nFailed to fetch available examples: ", error_message, err);
}
```

mitchmindtree (changes requested):
Nice one! Just a couple small nits.

In the long term, we'll likely remove all the GitHub REST stuff in favour of #1380, but I think this is fine as a temporary approach for making the `forc init --template` experience much nicer in the meantime.

cc'd @JoshuaBatty as he's a bit more familiar with `forc init` generally

forc/src/ops/forc_init.rs
```diff
@@ -85,38 +109,92 @@ fn print_welcome_message() {
...
+            init_new_project(project_name, program_type)
         }
-        None => init_new_project(project_name, program_type),
     }
 }
 
+fn get_sway_examples() -> Result<Vec<String>> {
+    // Query the main repo so that we can search for the "sha" that belongs to "examples"
+    let sway_response: GithubRepoResponse =
+        ureq::get("https://api.github.com/repos/FuelLabs/sway/git/trees/master")
```
adlerjohn, line 169: Should also use `{SWAY_GIT_TAG}`

JoshuaBatty:
Nice one LGTM. Might just need to rebase onto master again first.

JoshuaBatty (approved):

mitchmindtree (approved):
Looks good!

@JoshuaBatty I don't think it needs a rebase, I think it might just be waiting on my approval.

Issue #1324, opened 2022-04-20
Update docs and init manifest for exported fuels-rs


adlerjohn:
Depends on a new release including https://github.com/FuelLabs/fuels-rs/pull/168

mohammadfawaz:
I think this has been done already. Closing.

PR #1371, opened 2022-04-22
(spike) Create forc-test lib

ra0x3:


ra0x3:
@mitchmindtree  
- I realize it was mentioned this might be a possible duplicate of #1324 - but this issue doesn't go into detail
- Was super quick to knock this out, so wanted to put this PR up in case it can be used (also wanted to get some practice regarding how we create and publish new crates)
- Did _not_ publish `forc-test` to crates.io - wanted to get some feedback before doing that (with it being possibly unnecessary) - so haven't tested that it works quite yet

mitchmindtree:
Thanks for opening this!

While #1324 is brief, it links to [a PR](https://github.com/FuelLabs/fuels-rs/pull/168) that is a pretty detailed discussion around the same thing we're trying to do there.

I think it's worth reading through that PR and to consider using the `fuels-rs` crate instead (which looks like it's yet to be published).

> Did not publish forc-test to crates.io

Generally, we publish crates using CI when a new release gets cut. You can find this CI job [here](https://github.com/FuelLabs/sway/blob/fe4f296ab9afaaf10e52f027e617f41e0945f214/.github/workflows/ci.yml#L345). I think fuel-bot is the only one with permissions to publish any of our crates :smiling_face_with_tear:

adlerjohn:
If you don't want a crate published then set its version to `0.0.0` and add the `publish = false` field to the crate's manifest file.

PR #1373, opened 2022-04-23
`forc-gm` starter plugin

### LINKED ISSUE #1200: Add a minimal `forc` plugin example, `forc-gm`, along with a chapter in the book
It would be very useful to have a minimal-as-possible forc plugin demonstration for developers who are considering writing their own plugins.

Including this in the `sway` repo would also allow us to include code snippets directly into a dedicated chapter in the book while ensuring that the code snippets remain tested and up-to-date.

We already have `forc-lsp`, `forc-explore` and soon `forc-fmt` which each act as nice demonstrations, however a more minimal case may be easier to digest and document in the book.

This could be a minimal `forc-gm` plugin that responds to `forc gm` with `gn`. cc @SilentCicero 

---

For anyone looking to pick up this issue, I recommend referring to the `forc-lsp` plugin as a reference for how to add this.

binggh:
Fixes #1200 

So, this addresses #1200, but I have a suggestion to instead make this its own repo instead so that someone wanting to write their own plugin can use [cargo generate](https://github.com/cargo-generate/cargo-generate) to instantly clone and start working on their plugin. 

Assuming we have a repo named `forc-plugin-template`:

```console
$ cargo generate https://github.com/FuelLabs/forc-plugin-template
🤷   Project Name : forc-my-plugin
🔧   Generating template ...
[1/5]   Done: .gitignore
[2/5]   Done: Cargo.toml
[3/5]   Done: README.md
[4/5]   Done: src/lib.rs
[5/5]   Done: src
🔧   Moving generated files into: `/path/to/plugin/forc-my-plugin`...
💡   Initializing a fresh Git repository
✨   Done! New project created /path/to/plugin/forc-my-plugin
```

Developer can then `cd` into the project directory and start working.

If the above sounds good, we can change this PR to just updating the docs in `/forcs/plugins.md`, remove the `forc-gm` directory and use that project in its own repo instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
     "docstrings",
     "examples/build-all-examples",
     "forc",
+    "forc-gm",
     "forc-explore",
     "forc-fmt",
     "forc-lsp",
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -35,6 +35,7 @@
   - [Calling Contracts](./blockchain-development/calling_contracts.md)
 - [Forc](./forc/index.md)
   - [Dependencies](./forc/dependencies.md)
+  - [Plugins](./forc/plugins.md)
   - [Commands](./forc/commands/index.md)
     - [forc addr2line](./forc/commands/forc_addr2line.md)
     - [forc build](./forc/commands/forc_build.md)
--- a/docs/src/forc/plugins.md
+++ b/docs/src/forc/plugins.md
@@ -1,0 +1,37 @@
+# Plugins
+
+The Fuel ecosystem has a few plugins which can be easily installed via cargo, and anyone can write their own plugins.
+
+Let's install a starter plugin, `forc-gm`, and take a look at how it works underneath:
+
+```sh
+cargo install forc-gm
+```
+
+Check that we have installed `forc-gm`:
+
+```console
+$ forc plugins
+/Users/<USER>/.cargo/bin/forc-gm
+```
+
+Underneath, `forc-gm` is a simple CLI app, with [clap](https://docs.rs/clap/latest/clap/) as the only dependency:
+
+```rust
+{{#include ../../../forc-gm/src/main.rs}}
+```
+
+You can say gm, or you can greet Fuel:
+
+```console
+$ forc gm
+gn!
+$ forc gm fuel
+gn from Fuel!
+```
+
+## Writing your own plugin
+
+We encourage anyone to write and publish their own `forc` plugin to enhance their development experience.
+
+Your plugin must be named in the format `forc-<MY_PLUGIN>` and you may use the above template as a starting point. You can use [clap](https://docs.rs/clap/latest/clap/) and add more subcommands, options and configurations to suit your plugin's needs.
--- a/forc-gm/Cargo.toml
+++ b/forc-gm/Cargo.toml
@@ -8,7 +8,5 @@ license = "Apache-2.0"
 repository = "https://github.com/FuelLabs/sway"
 description = "A sample `forc` plugin."
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
 [dependencies]
 clap = { version = "=3.1.8", features = ["derive"] }
--- a/forc-gm/Cargo.toml
+++ b/forc-gm/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "forc-gm"
+version = "0.1.0"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+homepage = "https://fuel.network/"
+license = "Apache-2.0"
+repository = "https://github.com/FuelLabs/sway"
+description = "A sample `forc` plugin."
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+clap = { version = "=3.1.8", features = ["derive"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1095,6 +1095,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "forc-gm"
+version = "0.1.0"
+dependencies = [
+ "clap 3.1.8",
+]
+
+[[package]]
 name = "forc-lsp"
 version = "0.10.3"
 dependencies = [
--- a/forc-gm/src/main.rs
+++ b/forc-gm/src/main.rs
@@ -1,0 +1,44 @@
+//! A sample `forc` plugin example.
+//!
+//! Once installed and available via `PATH`, can be executed via `forc gm`.
+
+use clap::Parser;
+
+#[derive(Debug, Parser)]
+#[clap(name = "forc-gm", about = "Sample Forc plugin", version)]
+struct App {
+    // Your name which Forc should greet!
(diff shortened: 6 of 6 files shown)


forc-gm/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "forc-gm"
+version = "0.1.0"
```
adlerjohn, line 3: If this is going to be published (not sure if we want it to?) then it needs to have the same `version` as other published crates.

forc-gm/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+homepage = "https://fuel.network/"
+license = "Apache-2.0"
+repository = "https://github.com/FuelLabs/sway"
+description = "A sample `forc` plugin."
+
+[dependencies]
+clap = { version = "=3.1.8", features = ["derive"] }
```
adlerjohn, line 12: major.minor unless there's a specific reason not to.
```suggestion
clap = { version = "3.1", features = ["derive"] }
```

forc-gm/Cargo.toml
binggh: Does seem like an unnecessary crate to publish, since it would just be a template...

Only upside would be that it allows people to do `cargo install forc-gm` easily to test it out, however on second thought, maybe it should remain unpublished since there's so few lines anyway. I'll update it with `publish = false` for now and bump the version down.

mitchmindtree: We should potentially just publish this even once, if only to stop any malicious actors from potentially using the name in the future.

mitchmindtree:
> but I have a suggestion to instead make this its own repo instead so that someone wanting to write their own plugin can use [cargo generate](https://github.com/cargo-generate/cargo-generate) to instantly clone and start working on their plugin.

Does `cargo generate` offer a way to use a project from within a subdirectory at all? I'm particularly curious as I was thinking about this w.r.t. `forc init --template` in #1380.

docs/src/SUMMARY.md
```diff
@@ -35,6 +35,7 @@
   - [Calling Contracts](./blockchain-development/calling_contracts.md)
 - [Forc](./forc/index.md)
   - [Dependencies](./forc/dependencies.md)
+  - [Plugins](./forc/plugins.md)
```
mitchmindtree, line 38: Should we maybe have the "Plugins" chapter come after "Commands" as they are kind of about extending commands?

docs/src/forc/plugins.md
```diff
@@ -0,0 +1,37 @@
+# Plugins
+
+The Fuel ecosystem has a few plugins which can be easily installed via cargo, and anyone can write their own plugins.
```
mitchmindtree, line 3: Perhaps we could start with something like:

_Plugins can be used to extend forc with new commands that go beyond the native commands mentioned in the previous chapter. While the Fuel ecosystem provides a few commonly useful plugins (`forc-fmt`, `forc-lsp`, `forc-explore`), anyone can write their own!_

docs/src/forc/plugins.md
```diff
@@ -0,0 +1,37 @@
...
+
+Let's install a starter plugin, `forc-gm`, and take a look at how it works underneath:
+
+```sh
+cargo install forc-gm
```
mitchmindtree, line 8: Re [this discussion](https://github.com/FuelLabs/sway/pull/1373#discussion_r857058919) - looks like we should publish if we want to install it like this :)

docs/src/forc/plugins.md
```diff
@@ -0,0 +1,37 @@
...
+{{#include ../../../forc-gm/src/main.rs}}
+```
+
+You can say gm, or you can greet Fuel:
+
+```console
+$ forc gm
+gn!
+$ forc gm fuel
+gn from Fuel!
```
mitchmindtree, line 30: Should this just be
```suggestion
gn Fuel!
```
? cc @SilentCicero for meme check

mitchmindtree:
Following my Q [here](https://github.com/FuelLabs/sway/pull/1373#issuecomment-1107695878), I think another motivator for keeping this in the repo is so that it stays up-to-date with the book, and that we can reliably use the mdbook #include preprocessor to include its code into the book. I'm not sure we could do this reliably if it's in another repo.

binggh:
> > but I have a suggestion to instead make this its own repo instead so that someone wanting to write their own plugin can use [cargo generate](https://github.com/cargo-generate/cargo-generate) to instantly clone and start working on their plugin.
> 
> Does `cargo generate` offer a way to use a project from within a subdirectory at all? I'm particularly curious as I was thinking about this w.r.t. `forc init --template` in #1380.

I tried it, doesn't work - seems like it looks only for the repo and not within the subdirs of the repo: https://github.com/cargo-generate/cargo-generate/blob/main/src/git/utils.rs#L70-L85

> Following my Q https://github.com/FuelLabs/sway/pull/1373#issuecomment-1107695878, I think another motivator for keeping this in the repo is so that it stays up-to-date with the book, and that we can reliably use the mdbook #include preprocessor to include its code into the book. I'm not sure we could do this reliably if it's in another repo.

Yeah, this makes sense 👍  `cargo generate` is out of the question then since it doesn't seem to pull from subdirectories within a repo. We can let plugin writers generate their own cargo project then (it's trivial anyway)

docs/src/SUMMARY.md
binggh: Makes sense!

docs/src/forc/plugins.md
binggh: i can also see why perhaps we might not want to publish this since it's just a sample for demonstration purposes, and if so we can remove this instruction

mitchmindtree: I think it's nice to get the experience of what it's like to install a custom plugin too. Though I suppose users already get this through `forc-fmt` etc. That said, I'd be happy to see us publish this along with the rest of the crates even if it is only a demo - all our publishing is automated and the occasional version bump shouldn't be much of a hassle. I think it's probably important we at least own the name on crates.io either way.

mitchmindtree (approved):

Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = [
     "docstrings",
     "examples/build-all-examples",
     "forc",
+    "forc-gm",
```
adlerjohn, line 7: Alphabetize

forc-gm/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "forc-gm"
+version = "0.10.3"
```
adlerjohn, line 3:
```suggestion
version = "0.11.0"
```

adlerjohn (changes requested):
last nits!

forc-gm/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "forc-gm"
+version = "0.11.3"
```
mitchmindtree, line 3:
```suggestion
version = "0.11.0"
```
^ This should match `forc`'s version

mitchmindtree (changes requested):

forc-gm/Cargo.toml
binggh: Oops! My bad

adlerjohn (approved):

Issue #1065, opened 2021-12-29
Standard Ethereum signing support modules (for MetaMask, Ledger, Trezor and WalletConnect)


SilentCicero:
## Background

Fuel can support Ethereum signing schemes (commonly `EIP712` and the `eth_sign` RPC endpoint) via it's `keccak256` and `ecrecover` opcode in the FuelVM in combination with either a simple `predicate` script or basic proxy `contract`. This would allow for easy UX adoption from users / projects who use common Ethereum wallets / apps such as `MetaMask`, `Ledger`, `Trezor` and `WalletConnect` without the need for additional installation, apps, plugins or special key management.

We should provide an easily accessible set of `modules` / `tools` for Sway developers to leverage pre-built secure modules for this flow.

---

## Predicate Pseudo Code

```rs
predicate;

use std::address::*;
use std::metadata::witness;
use std::hash::keccak256;
use std::serialize::serialize_packed;

const owner = Address(0);

fn main() -> bool {
    let digest = hash(
        serialize_packed("Ethereum signed message/", tx_id()
    ), HashMethod::keccak256);
    
    owner == ec_recover(digest, witness(0))
}
```

## Contract Pseudo Code

```rs
contract;

use std::address::*;
use std::metadata::witness;
use std::hash::keccak256;
use std::serialize::serialize_packed;

////////////////////////////////////////
// ABI declarations
////////////////////////////////////////

abi EthereumSignableWallet {
    fn forward(gas: u64, color: u64, amount: u64, destination: Address, data: Bytes);
}

////////////////////////////////////////
// Storage declarations
////////////////////////////////////////

const owner = Address(0); // this would be filled in at deployment.

////////////////////////////////////////
// Helper methods
////////////////////////////////////////

fn recover_ethereum_witness() -> Address {
    let digest = keccak256(
        serialize_packed("Ethereum signed message/", tx_id()
    );
    
    ec_recover(digest, witness(0))
}

////////////////////////////////////////
// ABI definitions
////////////////////////////////////////

impl EthereumSignableWallet for Contract {
    fn forward(gas: u64, color: u64, amount: u64, destination: Address, data: Bytes) {
        assert(owner == recover_ethereum_witness());
        call(gas, color, amount, desintation, data)
    }
}
```

---

Users can simply use their Ethereum wallets in our system, but will be required to either put their spendable deposits behind a restrictive predicate or put all their assets into a proxy contract.

## UX Flow Predicate Example
[...]

PR #1374, opened 2022-04-23
ECRECOVER ethereum addresses

### LINKED ISSUE #1068: Add the ability to recover an ethereum address.
We can currently recover a fuel public key or address.
We should add a module (maybe external to lib-std) to allow recovery of ethereum addresses:
ie: `ec_recover_eth_addr` (will probably require a new `B160` type as a prerequisite).

simonr0204:
Closes #1068 

- Adds `std::vm::evm` namespace to the std-lib, and implements `ecr`
- Recovers a 20-byte ethereum address. Value is returned as a left-padded Address type. 
- Currently recovers address from the literal signed data: EIP-191 domain separators (\x19Ethereum ... etc. ) are not yet taken into account. In ethereum web3 libraries these are often prepended automatically so we will probably need to add support to handle them later, e.g. for #1065


simonr0204:
Sorry -  I was a bit quick to request review, and now you all get to witness my first struggle with the sway CI

SilentCicero:
@simonr0204 run the formatter on all the sway files... QUICK... before John sees it!

simonr0204:
> @simonr0204 run the formatter on all the sway files... QUICK... before John sees it!

But... CI is passing :o

SilentCicero:
> > @simonr0204 run the formatter on all the sway files... QUICK... before John sees it!
> 
> But... CI is passing :o

We don't yet have formatting CI checks for Sway. 

The current CI pipleline check for Sway formatting is just John first waving from his desk over PR reviews.

sway-lib-std/src/lib.sw
```diff
@@ -18,5 +18,6 @@ dep block;
 dep token;
 dep ecr;
 dep reentrancy;
+dep vm/lib;
```
nfurfaro, line 21: Do we need a `lib` in the `vm` module?
I'm not saying it's wrong, but we should probably discuss the stdlib's desired structure as it grows.
cc @sezna

test/src/sdk-harness/build.sh
```diff
@@ -8,7 +8,7 @@ for f in $FILES
 do
 if [ ! -e ${f}/$MANIFEST ]
   then
-  echo "Can't build a project without a Forc manifest."
+  echo ""${f}" not built : can't build a project without a Forc manifest."
```
nfurfaro, line 11: This will conflict with the work in #1372
Maybe we can land that one first and then update this.

test/src/sdk-harness/test_projects/evm_ecr/mod.rs
```diff
@@ -0,0 +1,36 @@
+use fuel_core::service::Config;
+use fuel_tx::Transaction;
+use fuels_contract::script::Script;
+use fuels_signers::provider::Provider;
+use std::fs::read;
+
+async fn execute_script(bin_path: &str) -> u64 {
```
nfurfaro, line 7: I've used a helper like this to reduce boilerplate before too, and it makes me think we should create a small helpers lib for use in our test suite so we don't have to keep reimplementing (and maintaining) things like this in every new test. Probably something for a separate PR though.

test/src/sdk-harness/test_projects/evm_ecr/src/main.sw
```diff
@@ -0,0 +1,41 @@
...
+    // create a signature:
+    let sig_hi = 0x82115ed208d8fe8dd522d88ca77812b34d270d6bb6326ff511297766a3af1166;
+    let sig_lo = 0xc07204f554a00e49a2ee69f0979dc4feef07f7dba8d779d388fb2a53bc9bcde4;
+    let signature: B512 = ~B512::from(sig_hi, sig_lo);
+
+    // recover the address:
+    let result : Result<Address, EcRecoverError> = ec_recover_address(signature, msg_hash);
+    let recovered_address = result.unwrap();
+
+    recovered_address.value == ethereum_address.value
```
nfurfaro, line 39: Here you should be able to just do `recovered_address == ethereum_address` because I've implemented the `Eq` trait for `Address`.

nfurfaro (commented):
A few minor things, and we probably need a more general conversation about the structure of the stdlib as it matures, but looks great overall!

sway-lib-std/src/lib.sw
simonr0204: Yeah it's very possible this is just wrong! My logic was that this way when we add more stuff to the vm or evm sublibrairies, there's no external dependency to update with the new module, we just add it to lib.sw inside the sublibrary itself. Trying to avoid anything to do with the evm leaking 😂

test/src/sdk-harness/build.sh
simonr0204: Oh I forgot I left this in there. Will revert it

sway-lib-std/src/vm/evm/lib.sw
```diff
@@ -0,0 +1,3 @@
+library evm;
+
+dep ecr;
```
adlerjohn, line 3: format all files (should have a newline at the end)

test/src/sdk-harness/test_projects/evm_ecr/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["Simon"]
```
adlerjohn, line 2:
```suggestion
authors = ["Fuel Labs <contact@fuel.sh>"]
```

test/src/sdk-harness/test_projects/evm_ecr/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    receipts[0].val().unwrap()
+}
+
+#[tokio::test]
+async fn evm_ecr_implementation() {
+    let path_to_bin = "test_projects/evm_ecr/out/debug/evm_ecr.bin";
+    let return_val = execute_script(path_to_bin).await;
+    assert_eq!(1, return_val);
+}
```
adlerjohn, line 36: Format Rust files too!

sway-lib-std/src/vm/evm/ecr.sw
```diff
@@ -0,0 +1,51 @@
...
+use ::hash::{HashMethod, hash_pair};
+use ::result::*;
+
+pub enum EcRecoverError {
+    UnrecoverablePublicKey: (),
+}
+
+/// Recover the public key derived from the private key used to sign a message.
+/// Returns a `Result` to let the caller choose an error handling strategy.
+pub fn ec_recover(signature: B512, msg_hash: b256) -> Result<B512, EcRecoverError> {
```
adlerjohn, line 15: Wait why don't you use the `ec_recover` function that already exists for this?

sway-lib-std/src/vm/evm/ecr.sw
```diff
@@ -0,0 +1,51 @@
...
+
+/// Recover the address derived from the private key used to sign a message.
+/// Returns a `Result` to let the caller choose an error handling strategy.
+/// Ethereum addresses are 20 bytes long, so these are left-padded to fit in a 32 byte Address type.
+pub fn ec_recover_address(signature: B512, msg_hash: b256) -> Result<Address, EcRecoverError> {
+    let pub_key_result = ec_recover(signature, msg_hash);
+
+    if let Result::Err(e) = pub_key_result {
+        // propagate the error if it exists
+        Result::Err(e)
```
adlerjohn, line 37: Instead of this, maybe just return `pub_key_result` if `is_error()`?

sway-lib-std/src/vm/evm/ecr.sw
nfurfaro: Yes, good catch.

sway-lib-std/src/vm/evm/ecr.sw
nfurfaro: I believe this suggestion won't work. The types are different even though they appear to line up.

adlerjohn: Oh I see because `EcRecoverError` is being re-declared. In that case, just don't re-declare it and use the one in that already exists!

simonr0204: Yes this is all because I started with a copy paste of ecr.sw and forgot to tidy up the redundancies. Only the 2nd function here needs to exist.

adlerjohn: > banishes half the file to the shadow realm

test/src/sdk-harness/test_projects/evm_ecr/mod.rs
simonr0204: Yes. It would be nice to have a one-liner to execute a script and return the boolean result. We can do this directly with `forc run` so just exposing that in the SDK would be good.

sway-lib-std/src/vm/evm/ecr.sw
simonr0204: hmm the types don't line up though. `ec_recover_address` returns `Result<Address, EcRecoverError>`, whereas `ec_recover` returns `Result<B512, EcRecoverError>`

Even with the syntax as it is, i'm getting some weirdness where with `ec_recover` copy/pasted, it works, but when I import it from `std::ecr`, I get this : 

` let recovered_address = result.unwrap();
   |                                    ^^^^^^ No method named "unwrap" found for type "enum Result<struct Address<b256>, enum EcRecoverError<()>>".`

adlerjohn: > hmm the types don't line up though.

Ah I see. Well you still should be able to re-use the `EcRecoverError` from `ecr` at least.

test/src/sdk-harness/test_projects/evm_ecr/mod.rs
simonr0204: I'll mark this resolved and open an issue.

sway-lib-std/src/vm/evm/ecr.sw
simonr0204: Resolved by [39e4fba](https://github.com/FuelLabs/sway/pull/1374/commits/39e4fbaaa3481b8acbd05db4744cbb2d4b5f2073)

sway-lib-std/src/vm/evm/ecr.sw
simonr0204: Resolved by [39e4fba](https://github.com/FuelLabs/sway/pull/1374/commits/39e4fbaaa3481b8acbd05db4744cbb2d4b5f2073)

test/src/sdk-harness/test_projects/evm_ecr/mod.rs
simonr0204: Resolved by [39e4fba](https://github.com/FuelLabs/sway/pull/1374/commits/39e4fbaaa3481b8acbd05db4744cbb2d4b5f2073)

test/src/sdk-harness/test_projects/evm_ecr/Forc.toml
simonr0204: Resolved by [39e4fba](https://github.com/FuelLabs/sway/pull/1374/commits/39e4fbaaa3481b8acbd05db4744cbb2d4b5f2073)

test/src/sdk-harness/build.sh
simonr0204: Resolved by [39e4fba](https://github.com/FuelLabs/sway/pull/1374/commits/39e4fbaaa3481b8acbd05db4744cbb2d4b5f2073)

test/src/sdk-harness/test_projects/evm_ecr/src/main.sw
simonr0204: Resolved by [39e4fba](https://github.com/FuelLabs/sway/pull/1374/commits/39e4fbaaa3481b8acbd05db4744cbb2d4b5f2073)

test/src/sdk-harness/test_projects/evm_ecr/src/main.sw
```diff
@@ -0,0 +1,42 @@
+script;
+
+use std::b512::B512;
+use std::address::Address;
+use std::vm::evm::ecr::ec_recover_address;
+use std::ecr::EcRecoverError;
+use std::assert::assert;
```
adlerjohn, line 7: Sort dependencies in Sway files alphabetically.

simonr0204: Resolved : [b6c31ff](https://github.com/FuelLabs/sway/pull/1374/commits/b6c31ffd044cd0724506bfd423ee8b95edf00e9b)

I feel like if this is a hard requirement then `forc fmt` should be able to do it ? I was also concerned about import order, which in an ideal world is entirely superstitious but ...

sway-lib-std/src/vm/evm/ecr.sw
```diff
@@ -0,0 +1,32 @@
...
+use ::b512::B512;
+use ::context::registers::error;
+use ::ecr::{EcRecoverError, ec_recover};
+use ::hash::{HashMethod, hash_pair};
+use ::result::*;
+
+/// Recover the address derived from the private key used to sign a message.
+/// Returns a `Result` to let the caller choose an error handling strategy.
+/// Ethereum addresses are 20 bytes long, so these are left-padded to fit in a 32 byte Address type.
+pub fn ec_recover_address(signature: B512, msg_hash: b256) -> Result<Address, EcRecoverError> {
```
adlerjohn, line 13: cc @nfurfaro @Braqzen for thoughts here. Do we want a new type like `EVMAddress`, rather than re-using `Address`?

We could just keep it as-is and change it later if we find out by using it that it's not the best.

Braqzen: What is the additional benefit of implementing and maintaining a new type other than `Address`?
We already have `Address` and `ContractID`.
Breaking down `Address` into 2 types for namespacing alone is not something that I am fond of so if `Address` has all the functionality needed then I'd rather not have an additional type with a different name.

If there is some functionality / distinction which I am unaware of then I have no issue with an additional type for its explicit purpose.
FYI: I'm aware of ecrecover but I've never used it myself so @nfurfaro holds more weight here.

simonr0204: I think unless we implement a B160 type, making the distinction doesn't give us much. Arguably the same could be said of zeroing out the first 12 bytes (we could just do the keccak hash of the pubkey and be done with). But I suppose it does make it more obvious visually what you're dealing with if there are 24 leading zeros.

nfurfaro: I don't think another new type is warranted here... That said, I can imagine a scenario where someone sends funds to an Ethereum address mistakenly because it's an `Address`.
I'm in favour of the zeroed initial bytes for the visual distinction they add. One could also implement a helper function in the Ethereum lib like `is_eth_address() ` which checks the initial 12 bytes.

adlerjohn: > `is_eth_address()`

If we go down this route of having to distinguish between a Fuel address and an Ethereum address then we should move that to the type system.

nfurfaro: I guess functionally there's no difference between sending to an Ethereum address by mistake and sending to the wrong fuel address. Both are on the user and should be avoided.

sway-lib-std/src/lib.sw
sezna: Well, `lib.sw` is typically the entry point to a _project_, not a submodule. That being said, we don't really have a precedent yet for what to name a submodule's version of what Rust calls a `mod.rs`. We could do a quick bikeshed -- to me, it is a bit confusing to have two `lib.sw` files

sway-lib-std/src/vm/evm/ecr.sw
mohammadfawaz: I'm in favour of different types long term. It's worth bikeshedding over at some point so we may want to create an issue for it.

Semantically, Address and EthAddress are separate things with different "address spaces", just like ContractId and Address, though they share many properties (a trait maybe?)

sway-lib-std/src/lib.sw
simonr0204: How about... `mod.sw`?

sway-lib-std/src/vm/evm/ecr.sw
SilentCicero: In favor of different types for safety. An Ethereum address is going to be potentially managed and used differently.

If there is a case where you want to cast an Ethereum address to an address there should be an explicit step and method for this.

sezna (approved):
Approved from a code-quality perspective, but note that I'm not very familiar with the ECR algorithm and perhaps somebody else should ensure that the implementation looks correct.

adlerjohn (approved):

PR #1381, opened 2022-04-24
update `forc plugins` to emit file names only, with option of full paths

### LINKED ISSUE #1214: Add `forc plugins` command flag to only emit plugin file names instead of full paths
Follow up to #1199. Refer to that PR for context.

binggh:
Closes #1214

Extends `forc plugins` to emit file names only by default and with the option of adding `-v` or `--verbose` to print full paths.

@mitchmindtree i went ahead and inversed the behaviour you described in the issue, if this is undesirable I can change it.

### DIFF
--- a/forc/src/cli/commands/plugins.rs
+++ b/forc/src/cli/commands/plugins.rs
@@ -1,7 +1,6 @@
 use crate::cli::PluginsCommand;
 use anyhow::Result;
 use clap::Parser;
-use std::ffi::OsStr;
 use std::path::PathBuf;
 
 /// Find all forc plugins available via `PATH`.
@@ -10,7 +9,7 @@ use std::path::PathBuf;
 #[derive(Debug, Parser)]
 pub struct Command {
     /// Prints the absolute path to each discovered plugin.
-    #[clap(long = "verbose", short = 'v')]
+    #[clap(long = "paths", short = 'p')]
     print_full_path: bool,
 }
 
@@ -23,6 +22,12 @@ pub(crate) fn exec(command: PluginsCommand) -> Result<()> {
     Ok(())
 }
 
+/// # Panics
+///
+/// This function assumes that file names will never be empty since it is only used with
+/// paths yielded from plugin::find_all(), as well as that the file names are in valid
+/// unicode format since file names should be prefixed with `forc-`. Should one of these 2
+/// assumptions fail, this function panics.
 fn print_plugin(path: PathBuf, print_full_path: bool) {
     if print_full_path {
         println!("{}", path.display());
@@ -30,9 +35,9 @@ fn print_plugin(path: PathBuf, print_full_path: bool) {
         println!(
             "{}",
             path.file_name()
-                .unwrap_or_else(|| OsStr::new(""))
+                .expect("Failed to read file name")
                 .to_str()
-                .unwrap()
+                .expect("Failed to print file name")
         );
     }
 }
--- a/forc/src/cli/commands/plugins.rs
+++ b/forc/src/cli/commands/plugins.rs
@@ -1,8 +1,35 @@
+use crate::cli::PluginsCommand;
 use anyhow::Result;
+use clap::Parser;
+use std::path::PathBuf;
 
-pub(crate) fn exec() -> Result<()> {
+/// Find all forc plugins available via `PATH`.
+///
+/// Prints information about each discovered plugin.
+#[derive(Debug, Parser)]
+pub struct Command {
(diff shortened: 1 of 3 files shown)


forc/src/cli/commands/plugins.rs
```diff
@@ -1,8 +1,38 @@
...
+use std::path::PathBuf;
+
+/// Find all forc plugins available via `PATH`.
+///
+/// Prints information about each discovered plugin.
+#[derive(Debug, Parser)]
+pub struct Command {
+    /// Prints the absolute path to each discovered plugin.
+    #[clap(long = "verbose", short = 'v')]
+    print_full_path: bool,
```
mitchmindtree, line 14: I think `verbose` is maybe a bit too general/ambiguous for this case? Could we change it to something like `paths`, e.g. `cargo plugins --paths`?

forc/src/cli/commands/plugins.rs
```diff
@@ -1,8 +1,38 @@
...
 }
+
+fn print_plugin(path: PathBuf, print_full_path: bool) {
+    if print_full_path {
+        println!("{}", path.display());
+    } else {
+        println!(
+            "{}",
+            path.file_name()
+                .unwrap_or_else(|| OsStr::new(""))
```
mitchmindtree, line 33: I think we should just `expect` here seeing as this is only used with paths yielded from looping over `plugin::find_all()` above which should never contain a file with no file name. I'd also add a comment to `print_plugin` function stating the assumption, and that it panics if this assumption isn't met.

mitchmindtree (changes requested):

forc/src/cli/commands/plugins.rs
binggh: Makes sense! Was wondering how to best handle this to be honest

binggh:
Realised there's a flaw in `forc-documenter` while working on this:

Say we are actively working on a command locally and change something related to the command (can be a comment, option, etc.). We want to re-generate the docs for the command based on the options/subcommands that we changed locally, but `forc-documenter` calls `forc --help` on the installed `forc` and not the local `forc` that we are debugging on.

I worked around it for this PR by building forc and copying it into my `.cargo/bin` for now in order to generate the docs for the updated `plugins.md` - will open an issue and work on changing this behaviour.

mitchmindtree:
> Realised there's a flaw in forc-documenter while working on this

I think this is OK - it's not so much a flaw in `forc-documenter`, but rather that it's easy to forget to run `cargo install --path ./forc` before running `forc-documenter`. The CI should catch this however, as CI does exactly this.

> I worked around it for this PR by building forc and copying it into my .cargo/bin for now

Note that this is pretty much all `cargo install --path ./forc` does. It builds the local copy and installs it in `~/.cargo/bin`.

I guess we could change `forc-documenter` to invoke `forc` through cargo instead? I.e.

```
cargo run --bin forc -- --help
```
```
cargo run --bin forc -- <command> --help
```

This would make it harder to forget to install the local copy first. However, the problem I see with this however is that if `forc` has not been built before, the `forc-documenter write-docs` command might take a surprisingly long time as it will have to build the entirety of `forc` first, which might be strange behaviour for a script that's just meant to update the docs.

I think I'd be happy to leave it as is?

mitchmindtree (approved):

binggh:
> > Realised there's a flaw in forc-documenter while working on this
> 
> I think this is OK - it's not so much a flaw in `forc-documenter`, but rather that it's easy to forget to run `cargo install --path ./forc` before running `forc-documenter`. The CI should catch this however, as CI does exactly this.
> 
> > I worked around it for this PR by building forc and copying it into my .cargo/bin for now
> 
> Note that this is pretty much all `cargo install --path ./forc` does. It builds the local copy and installs it in `~/.cargo/bin`.
> 
> I guess we could change `forc-documenter` to invoke `forc` through cargo instead? I.e.
> 
> ```
> cargo run --bin forc -- --help
> ```
> 
> ```
> cargo run --bin forc -- <command> --help
> ```
> 
> This would make it harder to forget to install the local copy first. However, the problem I see with this is that if `forc` has not been built before, the `forc-documenter write-docs` command might take a surprisingly long time as it will have to build the entirety of `forc` first, which might be strange behaviour for a script that's just meant to update the docs.
> 
> I think I'd be happy to leave it as is?

I think I was mainly concerned about messing up the `forc` installation by installing a local copy for debugging - not sure if that should be a cause for concern... If you think it shouldn't be a big deal then I can close the issue I just opened after adding your comments to that :)

Either way, this PR should be ready!

Issue #1143, opened 2022-04-04
Compiler allows unconstrained impl type parameters.


adlerjohn:
While implementing #1118, stumbled upon a bug: https://github.com/FuelLabs/sway/runs/5819569715?check_suite_focus=true

Note: the old codegen also errors, but with a different error.

mohammadfawaz:
Another instance of generic types leaking into IR.. all generic types should be resolved before IR generation. The error in the old codegen is likely just a symptom of the same problem.

otrho:
This was the smallest repro I could come up with:
```rust
script;

struct S { }

impl<T> S {
  fn use_asm(self, value: T) {
    asm(size: size_of::<T>()) {
      aloc size;
    };
  }
}

fn main() {
    (S{}).use_asm(true);
}
```

It appears that the ASM block isn't copying the T type.  Note that the bug didn't arise when just using a generic function, it seemed to need to be in a generic `impl`.

otrho:
Could be tackled alongside #1103 and/or maybe #1116?

otrho:
I have a feeling this is where it needs to be fixed, as the problem is the trait method hasn't resolved T: https://github.com/FuelLabs/sway/blob/9fde15ff4232ad71a81bc5fa913576124a0d64a4/sway-core/src/semantic_analysis/ast_node/declaration.rs#L578

But I'm not sure.  @emilyaherbert @sezna do you have any ideas?  What does that comment mean?

sezna:
That comment is pretty vague and I feel like I probably wrote it, sorry about that. I believe what I meant was that the methods in a trait declaration are stored as un-type-checked until the trait is implemented, since they could be relatively generic and not have all the necessary type info.

otrho:
I'm still not sure what the solution is here.  I think the issue is that the struct itself doesn't have any type parameters, only the impl.  So the monomorphisation process currently starts with 'generic' type mappings which get resolved via `copy_types()` but these are all instantiated by typed function decls, typed struct decls or typed enum decls.

In the above we need the same but for typed impl decls, which isn't quite the same..?  The line I quoted above with the comment is in a `TypedTraitDeclaration` but as you say, a trait's methods aren't type checked until instantiated.  I think we instantiate them as `TypedFunctionDeclaration`s but in the case of `f()` in my repro, `T` is not resolved.  But is this true for all trait methods or am I missing something?  Let me check...

otrho:
[...]

PR #1388, opened 2022-04-24
Allow type arguments in struct call paths

sezna:
While working on #1143, I noticed that the new parser was not allowing type arguments in this position:

```rust
let a = S::<u64> { };
```

This was blocking me, anyway, so I went ahead and implemented it. The new parser worked well for this application and was easy to build upon. 🎊

### DIFF
--- a/test/src/e2e_vm_tests/test_programs/should_pass/language/primitive_type_argument/src/main.sw
+++ b/test/src/e2e_vm_tests/test_programs/should_pass/language/primitive_type_argument/src/main.sw
@@ -1,0 +1,14 @@
+script; 
+
+struct S<T> { }
+
+impl<T> S<T> {
+  fn f(self) -> u64 {
+    5
+  }
+}
+
+fn main() -> u64 {
+  let a = S::<u64> { };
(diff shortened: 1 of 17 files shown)


test/src/e2e_vm_tests/test_programs/should_pass/language/primitive_type_argument/src/main.sw
```diff
@@ -0,0 +1,14 @@
+script;
```
mohammadfawaz, line 1: Can we actually have a more elaborate test with multiple type arguments and maybe a longer call path?

sezna: I've added some more thorough test cases, let me know what you think 😄

mohammadfawaz: Nice!

mohammadfawaz (approved):
LGTM! Thanks.

Issue #1365, opened 2022-04-22
Error annotation window snippets and empty 'dummy' spans aren't compatable.


otrho:
Spans now have a 'dummy' which is empty, which replaces the original 'empty' span.  `Span::dummy()` has a start and end of 0 and an empty string "".  `Span::empty()` used to have a non empty string `"Source span not available"`, which isn't great (and has been described as cryptic elsewhere) but was still printable.

In `forc-util/src/lib.rs` there is `format_err()` which tries to construct a nice source annotation from the span.  If the start and end are equal then it just increments the end, which is a little bit broken, especially for an empty span.  In turn the new size of (end - start =) 1 doesn't match the empty span and we get a panic:
```
thread 'main' panicked at 'SourceAnnotation range `(0, 1)` is bigger than source length `0`'
```
The solution is probably to recognise empty spans and not even try to create a `SourceAnnotation`.  This might not be the only place where this is an issue though.

Or maybe the solution is not to have an empty span ever as they don't make sense.  This though would require everywhere a span can't be provided to instead use `Option` which is quite an invasive change.

canndrew:
> Or maybe the solution is not to have an empty span ever as they don't make sense.

Using `None` is definitely more correct/type-safe than using `DUMMY_SPAN`. Ideally we should always know the span of an error though. `DUMMY_SPAN` was added with the intention of it being a temporary hack that could be removed.

Adding one to the end of an empty span is still broken though. Arbitrary `usize` values aren't valid `str` indexes, even if they're within the range of the string.

canndrew:
I've made a PR to fix the crash: https://github.com/FuelLabs/sway/pull/1394

Braqzen:
Note: `forc --version` is 0.11.0

Steps to reproduce:
1. `forc init span_bug`
2. Copy and paste the code presented below into the `main.sw` and save the changes
3. `forc build`

```
contract;

abi Escrow {
    fn constructor();
}

storage {
    approved: bool
}

impl Escrow for Contract {

    fn constructor() {
        storage.approved = true;
    }

}
```

Note the warning too
![1](https://user-images.githubusercontent.com/103777923/165627895-23bc18f8-c1c9-4a96-af8c-1cb6d8d81e94.png)

adlerjohn:
I might be missing some context, but how is `This storage declaration is never accessed and can be removed.` related to this issue?

Braqzen:
[...]

PR #1394, opened 2022-04-26
Fix annotation span bug #1365

canndrew:


otrho:
This change prevents the panic, but it seems to swallow the error message.

Before this patch:
```
: ../forc build
  Compiled library "core".
  Compiled library "std".
  Compiled library "foolib".
thread 'main' panicked at 'SourceAnnotation range `(0, 1)` is bigger than source length `0`', /home/toby/.cargo/registry/src/github.com-1ecc6299db9ec823/annotate-snippets-0.9.1/src/display_list/from_snippet.rs:286:9
```
After:
```
: ../forc build
  Compiled library "core".
  Compiled library "std".
  Compiled library "foolib".
error
 |
 |
____

  Aborting due to 1 error.
Error: Failed to compile foocaller
```
Prior to the span changes it would produce a slightly malformed but not entirely empty error:
```
: ../forc build
  Compiled library "core".
  Compiled library "std".
  Compiled library "foolib".
error
--> :0:1
 |
0 | Source span not available.
 | ^ Internal compiler error: Verification failed: Attempt to insert value of incorrect type into a struct.
```

otrho:
Admittedly it's hard to reproduce.  You need the compiler to produce an error with a `Span::dummy()` deliberately.

adlerjohn:
Is there no way to reproduce this in practice?

sezna:
cc @canndrew and @otrho regarding @adlerjohn 's question?

Braqzen:
@adlerjohn @sezna see [how to reproduce](https://github.com/FuelLabs/sway/issues/1365#issuecomment-1111468435)

otrho:
Generally the only place we use dummy empty spans are for errors with no obvious code correlation, which are progressively becoming rarer as the code matures.  We should only see this sort of thing with ICEs, which are hard to deliberately repro.

otrho:
Before:
```
warning                                                                                                                                                                                                     
 --> braqzen_repro/src/main.sw:8:5                                                                                                                                                                          
  |                                                                                                                                                                                                         
6 |                                                                                                                                                                                                         
7 | storage {                                                                                                                                                                                               
8 |     approved: bool                                                                                                                                                                                      
  |     -------- This storage declaration is never accessed and can be removed.                                                                                                                             
9 | }                                                                                                                                                                                                       
  |                                                                                                                                                                                                         
____                                                                                                                                                                                                        
                                                                                                                                                                                                            
thread 'main' panicked at 'SourceAnnotation range `(0, 1)` is bigger than source length `0`', /home/toby/.cargo/registry/src/github.com-1ecc6299db9ec823/annotate-snippets-0.9.1/src/display_list/from_snipp
et.rs:286:9                                                                                           
```
After:
```
warning
 --> braqzen_repro/src/main.sw:8:5
  |
6 | 
7 | storage {
8 |     approved: bool
  |     -------- This storage declaration is never accessed and can be removed.
9 | }
  |
____

error: Internal compiler error: Verification failed: State access operation must be to a u64 pointer.
____

  Aborting due to 1 error.
Error: Failed to compile braqzen_repro
```

sezna (approved):
It would be nice to fix the ICE but this is an incremental improvement, so LGTM…

adlerjohn:
Can we get an issue filed with the repro?

sezna:
I think it could be included under #1432? cc @otrho how do we remove all potential ICE cases from verification?

otrho:
Yeah #1432 is a reaction to this issue.  They are the epitome of ICEs though -- they indicate that the compiler has created some invalid IR -- so that doesn't need to change necessarily.  So ideally we should never see them, but if a general user was to get one at the moment it isn't obvious what they should even submit in a bug report.

PR #1397, opened 2022-04-26
Return a unit type constant when a function has no return value

mohammadfawaz:
E.g. 
```rust
fn f() {
    asm ( .. ) {
        //  No return value. Just a log for example
    }
}
```
In the above, the `asm` block doesn't return anything but we still return its result (of type `()`) from `f()`. Instead, I'm now returning a separate `const ()` directly which seems to simplify a few things downstream. The current code is causing an illegal `move` to be generated (a `move` from a register that isn't assigned) which causes some issues in the register allocator. With this patch, the problem goes away.

This fixes the "not enough registers" problem that @nfurfaro was seeing.


mohammadfawaz:
There's probably a few ways of doing this but this seems to be the quickest.

otrho:
Yeah, if feels like you're treating the symptom and not the cause..?  This will need a follow up.

otrho:
The ASM for any script/contract which used to implicitly return unit is now explicitly loading a unit (zero) from the data section, but is still returning `$zero`.  Not ideal, but neither is a lot of the new post-IR ASM until we get in there and aggressively clean it up, which I'm absolutely busting to do, but isn't `p.critical` at this stage.

Looks like I was premature with just `call_basic_storage`, other contracts have changed too...

otrho (approved):
Approved with a view to #1408.

mohammadfawaz:
@otrho thank you so much for fixing the tests! 🙌

Issue #1386, opened 2022-04-24
Clean up how `build-all-examples` checks for formatting


adlerjohn:
Overall, putting formatting checks into a script called `build_all_examples` seems hacky. Even if its just CI, this is tooling/code other people will have to read and maintain. Would prefer to see the main logic abstracted into a function along the lines of `for_all_examples` that was shared between separate `build_all_examples` and `fmt_check_all_examples` scripts or something like that.

_Originally posted by @mitchmindtree in https://github.com/FuelLabs/sway/pull/1384#pullrequestreview-951182832_

binggh:
I think I can pick this up and include this in my PR #1398: was already refactoring the original script since I was moving it into the `/scripts`

PR #1398, opened 2022-04-26
Move build all examples

### LINKED ISSUE #1389: Move `build-all-examples` out of `examples` directory
Currently the `build-all-examples` test binary is located in the `examples` directory and thus requires some special-casing. Move it to the root of the repo.

binggh:
Closes #1389 and #1386

- Move build-all-examples -> examples-checker in the scripts dir
- Use `clap` in the script so we can run `forc fmt --check` and `forc build` separately
- Some refactoring in the script
- Support running with paths or all examples (`--paths <PATH_1> <PATH_2> ...`, `--all-examples`)


Cargo.toml
```diff
@@ -2,7 +2,7 @@
 resolver = "2"
 members = [
     "docstrings",
-    "examples/build-all-examples",
+    "scripts/build-all-examples",
```
adlerjohn, line 5: alphabetize

scripts/build-all-examples/src/main.rs
```diff
@@ -8,11 +8,18 @@ use std::{
...
 
+fn get_sway_path() -> PathBuf {
+    let mut curr_path = std::env::current_dir().unwrap();
+    loop {
+        if curr_path.ends_with("sway") {
+            return curr_path;
+        }
+        curr_path = curr_path.parent().unwrap().to_path_buf()
+    }
+}
```
adlerjohn, line 19: Instead, can the program accept a list of paths? That way it can even be used outside the example directory for things like building and testing the stdlib!

binggh:
I've extended the script usability such that you have to explicitly specify one of two choices:

1) if you want to build examples by paths:

`cargo run --bin build-examples build --paths <PATH_1> <PATH_2> ...`

2) or if you want to build all examples found within `/examples`:

`cargo run --bin build-examples build --all-examples`

For now I've updated the CI to do the latter by default - we can extend the usage of this script to run using paths in future, when the need to test things outside of the /examples dir arises :)

adlerjohn:
Nice. Is this PR ready for review?

binggh:
> Nice. Is this PR ready for review?

Not just yet, have some local changes that will address #1386 :)

binggh:
OK, I renamed it to `examples-checker` (couldn't think of a better name for now), and I separated the `fmt` and `build` call. Also added in either the `--paths` or the `--all-examples` options so that you can either specify paths or simply run the forc command on all examples. Should be good to go!

A few questions:

1) Is there a need to build/format examples during the clippy step in the CI? Seems unnecessary because it doesn't seem to check the examples.
2) We are also building/formatting Sway examples under `cargo-test-workspace` as well as `cargo-unused-deps-check`, was wondering if there's a reason we are doing this for those CI jobs as well, because they seem unrelated.

mitchmindtree (changes requested):
Nice work!

> Is there a need to build/format examples during the cargo-clippy step in the CI?

I believe the build part is required but not the formatting check - see #1322 for why building the examples is necessary.

Rather than checking example formatting in multiple jobs, perhaps we can create a dedicated job called something like `forc-fmt-check-examples` that just checks the formatting of the examples. This should make it clearer in the PR CI job overview if that particular step fails.

scripts/examples-checker/src/main.rs
```diff
@@ -0,0 +1,171 @@
...
+    let mut curr_path = std::env::current_dir().unwrap();
+    loop {
+        if curr_path.ends_with("sway") {
+            return curr_path;
+        }
+        curr_path = curr_path.parent().unwrap().to_path_buf()
+    }
+}
+
+fn run_forc_command(path: PathBuf, cmd_args: Vec<&str>) -> (PathBuf, bool) {
```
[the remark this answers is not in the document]
binggh: Thanks for the catch! Will change those parts as recommended.

binggh:
Made 2 changes on the suggestion of @mitchmindtree:

1) Updated the fns to accept a reference to Path, and while refactoring realized that it was unnecessary to return a `Path` as well since the path was not being used within the fns. Changed it to return the `success` bool only.

2) Updated `ci.yml` to build the examples in most places, and only running `forc fmt --check` in a dedicated job (`forc-fmt-check-sway-examples`)

scripts/examples-checker/src/main.rs
```diff
@@ -0,0 +1,168 @@
...
+
+fn run_forc_build(path: &Path) -> bool {
+    run_forc_command(path, &["build", "--path"])
+}
+
+fn run_forc_fmt(path: &Path) -> bool {
+    run_forc_command(path, &["fmt", "--check", "--path"])
+}
+
+fn print_summary(summary: &Vec<(PathBuf, bool)>, command_kind: CommandKind) {
```
mitchmindtree, line 97: Small tip - there is almost never a case where `&Vec<T>` is preferred over `&[T]` in argument position. This is because the only differentiating behaviour between `Vec<T>` and `[T]` is that `Vec` is growable, however a `Vec` can't be grown behind an immutable reference. In other words, this argument could be `&[(PathBuf, bool)]` as `&Vec<T>` automatically derefs to `&[T]`, and `&[T]` is strictly more flexible (e.g. also supports fixed-size arrays, `Box<[T]>`, and so on).

scripts/examples-checker/src/main.rs
```diff
@@ -0,0 +1,168 @@
...
+        "successes"
+    };
+    let failures_str = if failures == 1 { "failure" } else { "failures" };
+    println!(
+        "{} {}, {} {}",
+        successes, successes_str, failures, failures_str
+    );
+
+    if failures > 0 {
+        std::process::exit(1);
```
mitchmindtree, line 124: Generally I think it's better to avoid hiding early program exits like this in seemingly unrelated functions. `print_summary` doesn't sound like the kind of function that could terminate the whole program. It might be better to return a result or whether or not it was successful so that `std::process::exit(1)` can be called from some higher level.

mitchmindtree (approved):
LGTM! I left a couple small tips, but neither really matter for this script so approving anyway, but happy to approve again if you wanna address them.

scripts/examples-checker/src/main.rs
binggh: Thanks for this tip - definitely is strictly better!

mitchmindtree (approved):

PR #1400, opened 2022-04-26
Ensure well-formed and concrete types before IR gen

sezna:
The IR stage is written such that any unresolved types are an internal compiler error. This is an invariant of the IR according to @otrho and therefore should be upheld as an ICE.

This PR adds a pass before the IR stage that ensures all types are well-formed and not insufficiently specified before IR gen. Eventually we could make a _new_ tree type that uses a specified `Type` enum without any unspecified variants. That would get us compile-time proofs of correctness and allow us to drop the type engine after type checking. However, it would be a _ton_ of copy/paste as we would have to recreate our AST type, so I'm deeming it out of scope for now.

before:
image

after:
image

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/mod.rs
+++ b/sway-core/src/semantic_analysis/ast_node/mod.rs
@@ -94,6 +94,35 @@ impl std::fmt::Display for TypedAstNode {
 }
 
 impl TypedAstNode {
+    /// Returns `true` if this AST node will be exported in a library, i.e. it is a public declaration.
+    pub(crate) fn is_public(&self) -> bool {
+        use TypedAstNodeContent::*;
+        match &self.content {
+            Declaration(decl) => decl.visibility().is_public(),
+            ReturnStatement(_)
+            | Expression(_)
+            | WhileLoop(_)
+            | SideEffect
+            | ImplicitReturnExpression(_) => false,
+        }
+    }
+
+    /// Naive check to see if this node is a function declaration of a function called `main` if
+    /// the [TreeType] is Script or Predicate.
+    pub(crate) fn is_main_function(&self, tree_type: TreeType) -> bool {
+        match &self {
+            TypedAstNode {
+                content:
+                    TypedAstNodeContent::Declaration(TypedDeclaration::FunctionDeclaration(
+                        TypedFunctionDeclaration { name, .. },
+                    )),
+                ..
+            } if name.as_str() == crate::constants::DEFAULT_ENTRY_POINT_FN_NAME => {
+                matches!(tree_type, TreeType::Script | TreeType::Predicate)
+            }
+            _ => false,
+        }
+    }
     /// if this ast node _deterministically_ panics/aborts, then this is true.
     /// This is used to assist in type checking branches that abort control flow and therefore
     /// don't need to return a type.
--- a/sway-core/src/type_engine/unresolved_type_check.rs
+++ b/sway-core/src/type_engine/unresolved_type_check.rs
@@ -1,0 +1,248 @@
+//! This module handles the process of iterating through the typed AST and ensuring that all types
+//! are well-defined and well-formed. This process is run on the AST before we pass it into the IR,
+//! as the IR assumes all types are well-formed and will throw an ICE (internal compiler error) if
+//! that is not the case.
+
+use crate::{semantic_analysis::*, type_engine::*};
+
+/// If any types contained by this node are unresolved or have yet to be inferred, throw an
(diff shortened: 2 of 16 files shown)


test/src/e2e_vm_tests/test_programs/should_fail/insufficient_type_info/Forc.toml
```diff
@@ -0,0 +1,7 @@
+[project]
+authors = ["Sezna"]
```
adlerjohn, line 2:
```suggestion
authors = ["Fuel Labs <contact@fuel.sh>"]
```

adlerjohn (changes requested):

sway-core/src/semantic_analysis/ast_node/mod.rs
```diff
@@ -94,6 +94,35 @@ impl std::fmt::Display for TypedAstNode {
...
+    /// the [TreeType] is Script or Predicate.
+    pub(crate) fn is_main_function(&self, tree_type: TreeType) -> bool {
+        match &self {
+            TypedAstNode {
+                content:
+                    TypedAstNodeContent::Declaration(TypedDeclaration::FunctionDeclaration(
+                        TypedFunctionDeclaration { name, .. },
+                    )),
+                ..
+            } if name.as_str() == "main" => {
```
mohammadfawaz, line 120: Can the string `"main"` be moved the file that contains all the constants? I'm going to assume that this string is used in multiple places throughout the codebase.

test/src/e2e_vm_tests/test_programs/should_fail/insufficient_type_info/src/main.sw
```diff
@@ -0,0 +1,10 @@
+script;
+
+
+fn foo<T>() {
+  let x = size_of::<T>();
```
mohammadfawaz, line 5: Formatter :)

sway-core/src/type_engine/unresolved_type_check.rs
```diff
@@ -0,0 +1,248 @@
...
+                        .map(|x| &x.type_id)
+                        .flat_map(UnresolvedTypeCheck::check_for_unresolved_types)
+                        .collect(),
+                );
+                body
+            }
+            ConstantDeclaration(TypedConstantDeclaration { value, .. }) => {
+                value.check_for_unresolved_types()
+            }
+            StorageReassignment(TypeCheckedStorageReassignment { fields, rhs, .. }) => fields
```
mohammadfawaz, line 232: `StorageReassignment` cannot have generics either. I suppose this doesn't hurt though.

sway-core/src/type_engine/unresolved_type_check.rs
```diff
@@ -0,0 +1,248 @@
...
+/// error to signal to the user that more type information is needed.
+pub(crate) trait UnresolvedTypeCheck {
+    fn check_for_unresolved_types(&self) -> Vec<CompileError>;
+}
+
+impl UnresolvedTypeCheck for TypeId {
+    fn check_for_unresolved_types(&self) -> Vec<CompileError> {
+        use TypeInfo::*;
+        match look_up_type_id(*self) {
+            UnknownGeneric { name } => vec![CompileError::UnableToInferGeneric {
```
mohammadfawaz, line 18: Small nit regarding how this error is defined in `errors.rs`. Should we change it to be closer to Rust? Something like: 
"cannot infer type for type parameter `T`..."

mohammadfawaz:
I like this change very much. A verification pass for each major step in the compiler helps detecting and debugging errors. Also helps achieve a #beyondmonolithic compiler 😁

sway-core/src/type_engine/unresolved_type_check.rs
sezna: It was easy enough to write...maybe we will have generic storage in the future...shrug

mohammadfawaz (approved):
LGTM!

mohammadfawaz:
Wait, there's a 14 MB binary files added in this change called `nvim.appimage` 😨

otrho:
I'll all for Neovim advocacy but this is a bit much. 😉

otrho (commented):
That was more involved that I thought it would be.

If you're up for it, while removing the appimage you could also patch this to use the 'main' constant:
https://github.com/FuelLabs/sway/blob/95816e4e41aae1d3425ba6ff5e7266076d8400fa/sway-core/src/lib.rs#L481

Otherwise, looks good.

adlerjohn (approved):

PR #1407, opened 2022-04-26
Remove the old `pest` parser code

sezna:
_WIP_


sway-core/src/lib.rs
```diff
@@ -127,60 +105,32 @@ impl ParseTree {
 /// ```
 ///
 /// # Panics
-/// Panics if the generated parser from Pest panics.
+/// Panics if the parser from panics.
```
eureka-cpu, line 108: Remove from here?

eureka-cpu (commented):
LGTM, very satisfying to review, good work!

otrho (approved):
So much red.. my retinas... ahh.  We can consider doing something similar to the codegen soonish.

adlerjohn (dismissed):

sway-core/src/error.rs
```diff
@@ -115,6 +115,10 @@ impl<T> From<Result<T, TypeError>> for CompileResult<T> {
 }
 
 impl<T> CompileResult<T> {
+    pub fn is_ok(&self) -> bool {
+        self.value.is_some() && self.warnings.is_empty() && self.errors.is_empty()
```
mitchmindtree, line 119:
```suggestion
        self.value.is_some() && self.errors.is_empty()
```
We can have a successful compilation result but still get warnings - should the `self.warnings.is_empty()` part of this condition be removed?

sway-core/src/error.rs
```diff
@@ -978,42 +970,7 @@ impl TypeError {
...
-                        }
-                        buf
-                    }
-                    pest::error::ErrorVariant::CustomError { message } => message.to_string(),
-                }
-            ),
-            a => format!("{}", a),
-        }
+        format!("{}", self)
     }
```
mitchmindtree, line 974: Should we use [#[deprecated]](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute) on this method recommending that we use `format!("{}")` like its implementation? Or maybe just remove it?

test/src/e2e_vm_tests/harness.rs
```diff
@@ -147,6 +144,8 @@ pub(crate) fn test_json_abi(file_name: &str) -> Result<()> {
         fs::read_to_string(oracle_path).expect("Something went wrong reading the file.");
     let output_contents =
         fs::read_to_string(output_path).expect("Something went wrong reading the file.");
+    println!("{}", oracle_contents);
+    println!("{}", output_contents);
```
mitchmindtree, line 148: Do we still want to print these? Or were they just for debugging parser changes?

mitchmindtree (commented):
Exciting stuff!!

:red_circle: = my fav diff colour

sway-core/src/error.rs
otrho: Fair call.  I say remove it.
```
 : grep -Ir to_friendly_error_string .
./forc-pkg/src/pkg.rs:        .map(|e| e.to_friendly_error_string())
./sway-core/src/error.rs:    pub fn to_friendly_error_string(&self) -> String {
./sway-fmt/src/fmt.rs:                    e.to_friendly_error_string(),
./forc-util/src/lib.rs:    let friendly_str = maybe_uwuify(&err.to_friendly_error_string());
./sway-lsp/src/capabilities/diagnostic.rs:                message: error.to_friendly_error_string(),
```

mohammadfawaz:
image

otrho (approved):

emilyaherbert (approved):

PR #1409, opened 2022-04-27
Add build-forc-test-project to ci

ra0x3:
### Description
- This PR adds a `build-forc-test-project` CI job that simply builds, formats, and tests a new forc project to ensure basic commands aren't broken

### Testing steps
- [ ] Added `build-forc-test-project` CI job should pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -464,6 +464,7 @@ jobs:
   build-publish-master-image:
     needs:
       [
+        build-forc-test-project,
         build-sway-examples,
         build-sway-lib-core,
         build-sway-lib-std,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -373,6 +373,7 @@ jobs:
   notify-slack-on-failure:
     needs:
       [
+        build-forc-test-project,
         build-sway-examples,
         build-sway-lib-core,
         build-sway-lib-std,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,7 +127,7 @@ jobs:
       - name: Build test project
         run: forc build --path test-proj
       - name: Run test project's test suite
-        run: furc test --path test-proj
+        run: forc test --path test-proj
 
   cargo-build-workspace:
     needs: cancel-previous-runs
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -127,7 +127,7 @@ jobs:
       - name: Build test project
         run: forc build --path test-proj
       - name: Run test project's test suite
-        run: furc test --path test-proj
```
mohammadfawaz: Ok, `furc test` sounds awesome.

.github/workflows/ci.yml
```diff
@@ -102,6 +102,33 @@ jobs:
             rm tmp.txt && exit 0
           fi
 
+  build-forc-test-project:
```
mitchmindtree, line 105: Could you add this new job to the list of `needs` for each of the publish jobs?

mitchmindtree (changes requested):

.github/workflows/ci.yml
```diff
@@ -437,6 +464,7 @@ jobs:
   build-publish-master-image:
     needs:
       [
+        build-forc-test-project,
```
mitchmindtree, line 467: Could you also add this to the `needs` arrays for the `notify-slack-on-failure` and `publish` jobs?

mitchmindtree (changes requested):

.github/workflows/ci.yml
mitchmindtree: @ra0x3 we're still missing this under the `publish` job.

mitchmindtree (approved):
Nice, I think this should make for a good start!

Currently this only tests the default init type (`--contract`) but we can extend this with each of the different init types (i.e. `--contract`, `--library`, `--predicate`, `--script`) in a follow-up PR. This could be a good use-case for the CI matrix feature, e.g. https://github.com/FuelLabs/fuel-core/commit/344115aa6c6a18863496cdf2dc6bb6e3fba05e94#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR57

ra0x3:
> Nice, I think this should make for a good start!

@mitchmindtree 
- I agree we should add the other implementation as well eventually. I can create this as a `goodfirstissue` 
- I see that Github actions allows for parallel jobs via the `matrix` - we should definitely parallelize what we can to speed up CI (not super hugely important however)
- My question here is -- I see many of our CI jobs that seem to be independent of other jobs (this job I just merged for example). Is there any reason we're not using a `matrix` strategy on these jobs currently?

PR #1410, opened 2022-04-27
Update the `Forc.lock` files for the `should_fail` E2E tests.

otrho:
They were accidentally left uncommitted after the switch from an explicit `core` dependency to `implicit-std = false`.


mitchmindtree (approved):

ra0x3 (approved):
👍🏽

PR #1415, opened 2022-04-27
Some code cleanup in `register_allocator.rs`

### LINKED ISSUE #1055: Improve code quality of `liveness_analysis()` in the register allocator
Clean up the implementation of https://github.com/FuelLabs/sway/blob/b1f175d36e929f7c0e3242b91df4d03d6fb4d84a/sway-core/src/asm_generation/register_allocator.rs#L53 as per Toby's comments in https://github.com/FuelLabs/sway/pull/1041

mohammadfawaz:
Closes #1055

### DIFF
--- a/sway-core/src/asm_generation/register_allocator.rs
+++ b/sway-core/src/asm_generation/register_allocator.rs
@@ -53,22 +53,17 @@ pub type InterferenceGraph =
 pub(crate) fn liveness_analysis(ops: &[RealizedOp]) -> HashMap<usize, BTreeSet<VirtualRegister>> {
     // Hash maps that will reprsent the live_in and live_out tables. The key of each hash map is
     // simply the index of each instruction in the `ops` vector.
-    let mut live_in: HashMap<usize, BTreeSet<VirtualRegister>> = HashMap::new();
-    let mut live_out: HashMap<usize, BTreeSet<VirtualRegister>> = HashMap::new();
-    for i in 0..ops.len() {
-        live_in.insert(i, BTreeSet::new());
-        live_out.insert(i, BTreeSet::new());
-    }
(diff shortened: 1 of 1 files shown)


sway-core/src/asm_generation/register_allocator.rs
```diff
@@ -53,22 +53,17 @@ pub type InterferenceGraph =
...
-    for (ix, op) in ops.iter().enumerate() {
-        offset_to_ix.insert(op.offset, ix);
-    }
+    let offset_to_ix = HashMap::from_iter(ops.iter().enumerate().map(|(idx, op)| (op.offset, idx)));
 
-    let mut modified: bool;
-    while {
+    let mut modified = false;
+    while !modified {
         modified = false;
```
adlerjohn, line 67: By definition, if `!modified` is `true`, then `modified` is `false`.

otrho: So this line is now redundant.

But hang on, does the algorithm want to loop until nothing is newly modified, or until it's modified for the first time?

mohammadfawaz: Every outer loop needs to re-initialize `modified` to `false`. 

The algorithm needs to know if the inner loop modified anything. If it did, we need more iterations of the outer loop.

mohammadfawaz: Ok.. I updated the logic. It should be `while modified`. The first outer loop iteration has to run (like a `do while`) so I have to initialize `modifed` to `true`. Then, every outer loop iteration has to re-initialize `modified` to `false`.

otrho: Yep, that makes more sense. 👍

otrho (approved):

PR #1417, opened 2022-04-27
Support `jnzi` in the compiler

### LINKED ISSUE #1412: Use the new `jnzi` opcode in the compiler
https://github.com/FuelLabs/fuel-specs/pull/306 introduces a new `jnzi` opcode that is similar to `jnei` but uses an 18 bit immediate instead of 12. This allows larger bytecodes to be formed. The natural way of using this new opcode is to only use it if needed. That is, continue using `jnei` if the immediate needed fits in 12 bits and switch to `jnzi` otherwise. To use `jnzi`, the two registers previously used in `jnei` will need to be compared before passing the result into `jnzi`.

Of course, `jnzi` won't work before it is actually implemented in the VM.

There is no need to enable `jnzi` in `asm` blocks because we don't support jumps there anyways.

mohammadfawaz:
Closes #1412

Mostly just replacing uses of `jnei` with `jnzi`, updating IR tests, and some e2e tests. Also bumping versions of various components to the latest available including the defaults for `forc init`. Finally, `lib-std` test are also updated to compile with the latest `fuels-rs`.

Remaining tasks:
* [x] bump `fuel-core` - waiting for a new release after https://github.com/FuelLabs/fuel-core/pull/293.
* [x] bump `fuels-rs` - waiting for a new release and hopefully the `lib-std` tests will pass 🤞 after that.


forc/Cargo.toml
```diff
@@ -23,10 +23,10 @@ clap = { version = "3.1", features = ["cargo", "derive", "env"] }
 clap_complete = "3.1"
 forc-pkg = { version = "0.11.0", path = "../forc-pkg" }
 forc-util = { version = "0.11.0", path = "../forc-util" }
-fuel-asm = "0.3"
-fuel-gql-client = { version = "0.5", default-features = false }
-fuel-tx = "0.7"
-fuel-vm = "0.6"
+fuel-asm = "0.4"
+fuel-gql-client = { version = "0.6.2", default-features = false }
```
adlerjohn, line 27: These should just be major.minor throughout, along with `fuel-core`.
```suggestion
fuel-gql-client = { version = "0.6", default-features = false }
```

Voxelot:
@mohammadfawaz fuel-rs is ready to bump now

forc/Cargo.toml
mohammadfawaz: I remember things didn't work without the `.2` 🤔 but I will try again once I'm through with all the bumps.

mohammadfawaz: Yup seems to be okay now with just major.minor. Updated everywhere.

test/src/sdk-harness/test-helpers/Cargo.toml
```diff
@@ -5,7 +5,6 @@ edition = "2021"
 
 
 [dependencies]
-fuel-core = { version = "0.5", default-features = false }
-fuel-tx = "0.7"
-fuels-contract = "0.8"
-fuels-signers = { version = "0.8", features=["test-helpers"] }
+fuel-core = { version = "0.6", default-features = false }
+fuel-tx = "0.9"
+fuels = "0.10.1"
```
adlerjohn, line 10:
```suggestion
fuels = "0.10"
```

sway-core/src/asm_lang/virtual_ops.rs
```diff
@@ -473,6 +477,19 @@ impl VirtualOp {
                         .collect()
                 }
             }
+            JNZI(_, i) => {
+                // Two possible successors: the next instruction as well as the instruction
+                // indicated in the jump offset. Use `offset_to_ix` to figure out the index in
+                // `ops` that corresponds to the offset specified.
```
otrho, line 483: Is this an exact copy of `JNEI`?  If so they might as well share the same code with `JNEI(_, _, i) | JNZI(_, i) => { ... }`

sway-core/tests/ir_to_asm/let_reassign_while_loop.asm
```diff
@@ -12,10 +12,12 @@ lw   $r0 data_0               ; literal instantiation
 sw   $r2 $r0 i0               ; store value
 addi $r0 $r2 i0               ; get_ptr
 lw   $r0 $r2 i0               ; load value
-jnei $r0 $one i21
+jnzi $r0 i15
+ji   i23
```
otrho, line 16: So we're unfortunately adding an extra jump here with the benefit of 18bit addressing.  Not the end of the world, but we could one day be smarter and use `JNEI` when we know the offset is < 256K.  Can't really be done in IR, would have to be a peephole ASM optimisation.

Just a thought.  There are approximately a zillion of these things which we could add in the future and I'm guessing never will actually get around to doing.

otrho (dismissed):

sway-core/src/asm_lang/virtual_ops.rs
mohammadfawaz: It is the same code but `JNEI(_, _, i) | JNZI(_, i) => { ... }` apparently doesn't work because the `i`s have different types (`VirtualImmediate12` v.s. `VirtualImmediate18`). Is there a way to make it work? 

I suppose I could always pull the common code out into a separate generic function.

sway-core/tests/ir_to_asm/let_reassign_while_loop.asm
mohammadfawaz: Yes this is a bit unfortunate. One alternative is to add another instruction that flips the condition before doing any jumps. That would get us back the same pattern of jumps as before but at the cost of an additional comparator. Or, as you said, use `jnei` if the "size" of the label fits in 12 bits.

I created a low priority issue here https://github.com/FuelLabs/sway/issues/1439

sway-core/src/asm_lang/virtual_ops.rs
otrho: Oh right, of course.

otrho (approved):

adlerjohn (approved):

nfurfaro:
too late, but still, LGTM!

Issue #1206, opened 2022-04-08
Remove `*.in_ir` and `*.out_ir` from `.gitattributes` once `FileCheck` is finalized.


mohammadfawaz:
Once `FileCheck` is finalized, remove the two lines in `.gitattributes` that correspond to file extensions `*.in_ir` and `*.out_ir`. (See https://github.com/FuelLabs/sway/pull/1204#discussion_r846541548)

otrho:
Fair enough.  Good way not to forget. :wink:

PR #1419, opened 2022-04-28
Switch the IR unit tests to use FileCheck.

otrho:
This is both simpler and a more expressive way to test IR passes.  It uses special embedded directives to confirm that an IR transformation has been applied correctly.

I've had this branch languishing around for a few weeks, thought I'd finish it up and submit it.


mohammadfawaz:
Can we also address this now https://github.com/FuelLabs/sway/issues/1206? 😁

otrho:
Yep, once someone approves this one!

mohammadfawaz (approved):
Very nice! 
Can we somehow ask for specific passes to run in a given test like `llvm-lit` instead of having the pass be part of the test name?

otrho:
Yeah, one day. :)  One step at a time.  Considering there are currently a total of 2 passes, this will do for now.  Thanks.

Issue #1188, opened 2022-04-07
Implement `mut` arguments


mohammadfawaz:
The following shouldn't compile today but it does:
```rust
struct A {
    a: u64,
}

impl A {
    fn f(self) {
        self.a = 42;
    }
}
```
because `self` is not passed to `f` as `mut`.
The change I'm proposing to fix https://github.com/FuelLabs/sway/issues/908 rightfully breaks the above. However, there is no way to un-break it because we can't yet write:
```rust
impl A {
    fn f(mut self) {
        self.a = 42;
    }
}
```
More generally, we want to be able to annotate any argument as `mut`:
```rust
fn bla(mut a: u64, mut b: u64) {
    a = 0;
    b = 0;
}
```
Edit: the case with arbitrary arguments implies that we support passing arguments by ref, which still has to be property designed and thought of. So, for now, we should just enable `mut self`.

mohammadfawaz:
We need to establish some ground rules around how different types are passed into functions. One proposal is as follows:
```
struct A {
    a: u64,
}

struct B {
    b1: u64,
    b2: u64,
}

impl A {
    fn f1(self, x: u64, b: B) {
        // Read-only access to self, self.a, x, b.b1, b.b2
    }

    fn f2(mut self, mut x: u64, mut b: B) {
        // Read/write access to self, self.a, x, b.b1, b.b2
    }

    fn f3(ref self, ref x: u64, ref b: B) {
        // Read-only access to self, self.a, x, b.b1, b.b2
    }
   fn f4(ref mut self, ref mut x: u64, ref mut b: B) {
        // Readwrite access to self, self.a, x, b.b1, b.b2
    }
}

fn main() {
    let mut a_data = A { a: 0 };
    let mut x_data = 0;
    let mut b_data = B { b1: 0, b2: 0 };

    a_data.f1(x_data, b_data); // No changes to a_data, x_data, or b_data expected.

    a_data.f2(x_data, b_data); // No changes to a_data, x_data, or b_data expected.

    a_data.f3(x_data, b_data); // No changes to a_data, x_data, or b_data expected.

    a_data.f3(x_data, b_data); // Potential changes to a_data, x_data, or b_data.
}

```

Note: I don't think there should be any distinction between copy types (such as `u64`) and non-copy types (such as `struct`) with regards to the above, even though `structs` are really just pointers to data with our implementation.

sezna:
[...]

PR #1420, opened 2022-04-28
Disallow assigning to fields of non-mutable structs

### LINKED ISSUE #908: The compiler does not prevent assigning a data member of a non-mutable struct
Here's a simple example. The code compiles and the `assert` passes. I'm not supposed to be able to write to `S.x` here since `S` is not declared as mutable.
```rust
script;
use std::chain::*;

struct S { x: u64 }

fn main() {
    let s = S { x: 0 }; // Non-mutable!!
    s.x = 1;
    assert(s.x == 1);
}
```
Reassigning the whole struct causes an error but reassigning parts of it does not.

canndrew:
Fixes #908


mohammadfawaz:
Thanks Andrew! There seems to be some E2E tests failing. 

Have you also checked if the test I have here passes? https://github.com/FuelLabs/sway/issues/908#issuecomment-1091883879 - I think there's a similar one in our tests.

adlerjohn (commented):
failing CI

mohammadfawaz:
I believe the failing E2E test requires the `mut` keyword on function arguments which we don't have yet. Otherwise, the test will continue to fail with your change. We can either wait until `mut` for arguments is fully implemented before submitting this change, or disable the test for now and mark it as blocked by the "`mut` keyword for arguments" feature.

Either way, I'm going to convert this to a draft to reduce the number of PRs that are ready for review. I hope you don't mind!

canndrew:
@mohammadfawaz: This is the test case that's failing.

```
script;

use std::assert::assert;

struct A {
    a: u64,
    b: u64,
}

impl A {
    fn f(self) {
        self.a = 42;
        self.b = 77;
    }
}

fn main() -> bool {
    let a = A {
        a: 0,
        b: 0,
    };
    a.f();
    assert(a.a == 42);
    assert(a.b == 77);
    true
}
```

The problem here isn't the lack of `mut` on function parameters, it's that the code here doesn't make sense. Assuming that sway is meant to have rust-like semantics rather than java-like semantics then the method `f` moves (or copies) `a` and so any modifications that `f` makes to `self` still won't show-up in `main`. In order to make this work we need `&mut self` methods, but references aren't implemented at all yet.

If this test passed before without the `mut`-check then that's a bad thing. The only way to make something like this work with mutable references is to re-write the code like this:

```
script;

use std::assert::assert;

struct A {
    a: u64,
    b: u64,
}

impl A {
    fn f(mut self) -> A {
        self.a = 42;
        self.b = 77;
        self                // <- return self here
    }
}

fn main() -> bool {
    let a = A {
        a: 0,
        b: 0,
    };
    let a = a.f();          // <- rebind `a` here
    assert(a.a == 42);
    assert(a.b == 77);
    true
}
```

@adlerjohn ^^

mohammadfawaz:
> @mohammadfawaz: This is the test case that's failing.
> 
> 
> 
> ```
> 
> script;
> 
> 
> 
> use std::assert::assert;
> 
> 
> 
> struct A {
> 
>     a: u64,
> 
>     b: u64,
> 
> }
> 
> 
> 
> impl A {
> 
>     fn f(self) {
> 
>         self.a = 42;
> 
>         self.b = 77;
> 
>     }
> 
> }
> 
> 
> 
> fn main() -> bool {
> 
>     let a = A {
> 
>         a: 0,
> 
>         b: 0,
> 
>     };
> 
>     a.f();
> 
>     assert(a.a == 42);
> 
>     assert(a.b == 77);
> 
>     true
> 
> }
> 
> ```
> 
> 
> 
> The problem here isn't the lack of `mut` on function parameters, it's that the code here doesn't make sense. Assuming that sway is meant to have rust-like semantics rather than java-like semantics then the method `f` moves (or copies) `a` and so any modifications that `f` makes to `self` still won't show-up in `main`. In order to make this work we need `&mut self` methods, but references aren't implemented at all yet.
> 
> 
> 
> If this test passed before without the `mut`-check then that's a bad thing. The only way to make something like this work without mutable references is to re-write the code like this:
> 
> 
> 
> ```
> 
> script;
> 
> 
> 
> use std::assert::assert;
> 
> 
> 
> struct A {
> 
>     a: u64,
> 
>     b: u64,
> 
> }
> 
> 
> 
> impl A {
> 
>     fn f(mut self) -> A {
> 
>         self.a = 42;
> 
>         self.b = 77;
> 
>         self                // <- return self here
> 
>     }
> 
> }
> 
> 
> 
> fn main() -> bool {
> 
>     let a = A {
> 
>         a: 0,
> 
>         b: 0,
> 
>     };
> 
>     let a = a.f();          // <- rebind `a` here
> 
>     assert(a.a == 42);
> 
>     assert(a.b == 77);
> 
>     true
> 
> }
> 
> ```
> 
> 
> 
> @adlerjohn ^^

You're absolutely right. That makes a lot of sense and this is where #1188 comes in. The failing test is indeed broken because it needs mut refs (or mut + ret like you wrote in the second example). Given this, I suggest we disable the test for now and track re-enabling it in #1188. The fix in this PR is more critical to merge I think.

canndrew:
I've disabled to offending test by commenting it out with a note linking to this issue. Is this good to go now?

adlerjohn (commented):
CI failing still ❗

mohammadfawaz:
Ah yes.. this test should probably be removed as well: https://github.com/FuelLabs/sway/blob/master/sway-core/tests/sway_to_ir/impl_self_reassignment.sw

There's probably no clean way to disable it atm, so we will just have to delete it for the time being.

mohammadfawaz (approved):
LGTM! Thanks.

PR #1422, opened 2022-04-28
Span/String/Ident fixups

canndrew:
This PR is just some small clean-ups:

* Rename `{CompileError, CompileWarning}::internal_span` to just `span` and make it return an owned `Span`.
* Replace lots of uses of `String` with `Ident`.


otrho:
Seems fair enough though is there a reason to introduce all the Span cloning?

canndrew:
This is rebased now.

@otrho: It was annoying for me at one point because I wanted to store more specific data in an error and compute the span from that. But you can't do that if you have to return a reference. Also cloning `Arc`s is fairly cheap, so long as it's not in the hot-path of the compiler. And cloning errors presumably isn't something we're doing a lot of (or at least we shouldn't be).

adlerjohn:
Is this ready for review @canndrew?

otrho (approved):

mohammadfawaz:
I'm going to merge this as it has been approved. I hope that's okay @canndrew!

Issue #605, opened 2022-01-09
Formatter configuration file


adlerjohn:
Add a configuration file, similar to `rustfmt.toml`, such that the formatter can be configured. Low-hanging features:

1. Init new formatting file, with defaults.
2. Option: tabs vs spaces
3. Option indent size

otrho:
> 2. Option: tabs vs spaces

At the risk of flamewar, IMO tabs should never be an option, and the formatter should always replace them with spaces.

adlerjohn:
I wouldn't be inherently opposed to an more-opinionated formatter for some things like tabs vs spaces.

mitchmindtree:
> > 2. Option: tabs vs spaces
> 
> At the risk of flamewar, IMO tabs should never be an option, and the formatter should always replace them with spaces.

It could be worth considering taking this even further and keeping the formatting configuration options as minimal as possible generally.

I remember the `rustfmt` configurations being useful early on while everyone was still experimenting with what standards to settle on, but I think we should consider omitting it altogether if we can before too many people start relying on it. It would keep our overhead for the formatter lower, and help to maintain a clear standard for CI checks, consistent readability between different projects, etc. My intuition is that most folks are happy to sacrifice a few bits and pieces of personal preference in favour of a clear, decent-enough-looking, ecosystem-wide standard - though I could be wrong!

At the very least, we should make sure we have a strong motivation for custom formatting configuration before committing to something like this.

PR #1425, opened 2022-04-28
Added skeleton for `sway-fmt` v2

eureka-cpu:
WIP: wrapping my head around implementing the new parser and studying `rustfmt`

### DIFF
--- a/sway-fmt-v2/Cargo.toml
+++ b/sway-fmt-v2/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "sway-fmt-v2"
+version = "0.11.0"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+homepage = "https://fuel.network/"
+license = "Apache-2.0"
+repository = "https://github.com/FuelLabs/sway"
+description = "Sway sway-fmt."
+
+[dependencies]
+ropey = "1.2"
+sway-parse = { version = "0.11.0", path = "../sway-parse" }
+sway-types = { version = "0.11.0", path = "../sway-types" }
+thiserror = "1.0.30"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3218,6 +3218,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "sway-fmt-v2"
+version = "0.11.0"
+dependencies = [
+ "ropey",
+ "sway-parse",
+ "sway-types",
+ "thiserror",
+]
+
+[[package]]
 name = "sway-ir"
 version = "0.11.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ members = [
     "scripts/forc-documenter",
     "sway-core",
     "sway-fmt",
+    "sway-fmt-v2",
     "sway-ir",
     "sway-lsp",
     "sway-parse",
--- a/sway-fmt-v2/src/error.rs
+++ b/sway-fmt-v2/src/error.rs
@@ -1,0 +1,3 @@
+use thiserror::Error;
+#[derive(Debug, Error)]
+pub enum FormatterError {}
--- a/sway-fmt-v2/src/fmt.rs
+++ b/sway-fmt-v2/src/fmt.rs
@@ -1,0 +1,37 @@
+use std::sync::Arc;
+use sway_parse::Item;
+use thiserror::Error;
+
+pub use crate::error::FormatterError;
+
+#[derive(Debug, Clone, Copy)]
+pub struct Formatter {
+    pub align_fields: bool,
+    pub tab_size: u32,
+}
+impl Formatter {
+    pub fn format(&self, src: Arc<str>) -> Result<String, FormatterError> {
+        let items = sway_parse::parse_file(file.clone(), todo!())?.items;
+        Ok(items
+            .into_iter()
+            .map(|item| -> Result<String, FormatterError> {
+                use Item::*;
+                Ok(match item {
+                    Use(use_stmt) => todo!("Format me!"),
+                    // don't format if we don't have a formatter for this `Item`
+                    otherwise => otherwise.span().as_str().to_string(),
+                })
+            })
+            .collect::<Result<Vec<String>, _>>()?
+            .join("\n"))
+    }
+}
+
+impl Formatter {
+    pub fn default() -> Self {
+        Self {
+            align_fields: true,
+            tab_size: 4,
+        }
+    }
+}
--- a/sway-fmt-v2/src/lib.rs
+++ b/sway-fmt-v2/src/lib.rs
@@ -1,0 +1,5 @@
+mod error;
+mod fmt;
+
+pub use crate::fmt::Formatter;
+pub use error::FormatterError;


adlerjohn:
Will this close #605?

eureka-cpu:
> Will this close #605?

Yep! At present it's rough but already does!

adlerjohn:
The same logic that's used to check the the Forc lockfile is probably a good start.

adlerjohn:
Also, do you best to cut off the scope of this PR to something atomic, otherwise it'll be impossible to review!

eureka-cpu:
> Also, do you best to cut off the scope of this PR to something atomic, otherwise it'll be impossible to review!

So the aim of this PR is just to make the foundation of the new formatter, and then implementations of each of the aforementioned in my top comment can be tackled in separate issues. I planned on bikeshedding that with you tomorrow, if possible. To be specific, included in this PR is all of the configurations, `swayfmt.toml` and the base (unfinished, but there) implementations for each configuration. It won't actually format anything until after this PR.

sway-fmt-v2/src/config/fundamentals.rs
```diff
@@ -0,0 +1,54 @@
+//! Basic configuration options.
+use serde::{Deserialize, Serialize};
+
+use crate::constants::{DEFAULT_TAB_SPACES, DEFAULT_MAX_LINE_WIDTH};
+
+/// Basic styling preferences.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Fundamentals {
```
sezna, line 8: Maybe `WhitespaceStyle`?

sway-fmt-v2/src/config/imports.rs
```diff
@@ -0,0 +1,56 @@
...
+            imports_granularity: ImportGranularity::Preserve,
+            imports_indent: IndentStyle::Block,
+            imports_layout: ListTactic::Mixed,
+        }
+    }
+}
+
+/// Configuration for import groups, i.e. sets of imports separated by newlines.
+#[derive(Serialize, Deserialize, Debug)]
+pub enum GroupImportsTactic {
```
sezna, line 31: Maybe just `GroupImports`, since the other enums don't say `Tactic`?

sway-fmt-v2/src/config/manifest.rs
```diff
@@ -0,0 +1,260 @@
...
+                    Ok(FormatConfig::default())
+                }
+            }
+            None => Ok(FormatConfig::default()),
+        }
+    }
+    /// Check the user's settings, and replace the values of the default formatter
+    /// with those of the user's if they exist.
+    ///
+    pub fn apply_user_settings(user_settings: OptionFormatConfig) -> FormatConfig {
```
eureka-cpu, line 117: @JoshuaBatty if there's an easier/nicer looking way of doing this, I'm all ears

eureka-cpu: @mitchmindtree also feel free to weigh in, in general

mitchmindtree: I think a nicer way to do this might be to implement this as a constructor on `FormatConfig`, something like:

```rust
impl FormatConfig {
    /// Construct the set of configuration to be used from the given set of options.
    pub fn from_opts(opts: &OptionFormatConfig) -> Self {
        Self {
            whitespace: opts.whitespace.as_ref().map(Whitespace::from_opts).unwrap_or_default(),
            imports: opts.imports.as_ref().map(Imports::from_opts).unwrap_or_default(),
            // etc.
        }
    }
}
```

Then also have something for each of the subsets of options (i.e. `Whitespace`, `Imports`), like:

```rust
impl Whitespace {
    pub const DEFAULT_MAX_WIDTH: usize = 100;
    pub const DEFAULT_HARD_TABS: bool = false;
    // the rest of the defaults...

    pub fn from_opts(opts: &WhitespaceOptions) -> Self {
        Self {
            max_width: opts.max_width.unwrap_or(Whitespace::DEFAULT_MAX_WIDTH),
            hard_tabs: opts.hard_tabs.unwrap_or(Whitespace::DEFAULT_HARD_TABS),
            // etc
        }
    }
}
```

Also I noticed the top-level options are called `OptionFormatConfig`, but the subsets are named `*Options` (i.e. `WhitespaceOptions`, `ImportsOptions`, etc). Maybe we should rename the top-level type to be consistent, e.g. `FormatConfigOptions`.

Also, we might be able to omit the `Format` word from these types altogether as it's kind of implied by being in the formatter crate, e.g. `sway_fmt::Config`. Maybe `Config` and `ConfigOptions` would be fine names for these?

eureka-cpu: @mitchmindtree thanks so much, I'll get this sorted tomorrow morning!

eureka-cpu: Is it necessary to have constants for the boolean defaults?

mitchmindtree: I think it's generally good practise to publicly expose defaults in a consistent manner, however I think an exception can be made in the case that the default value is equivalent to the `Default` implementation of its type (as is the case for `hard_tabs` being a `bool` whose `Default` implementation returns `false`).

> Is it necessary to have constants for the boolean defaults?

If they're all `false` by default, then possibly no in favour of using `Default::default()`, however if any are `true` then they should have constants.

> Should I make every default a constant?

Up to you - I think I probably would just for consistency.

eureka-cpu: Thanks so much for the input, other than some warnings for unused methods or constants, this PR should be ready for review tomorrow I think!

sway-fmt-v2/src/config/manifest.rs
```diff
@@ -0,0 +1,170 @@
...
+            ordering: Ordering::default(),
+            items: Items::default(),
+            lists: Lists::default(),
+            literals: Literals::default(),
+            expressions: Expressions::default(),
+            heuristics: Heuristics::default(),
+            structures: Structures::default(),
+            comments: Comments::default(),
+        }
+    }
```
mitchmindtree, line 63: These `default() -> Self` constructors should be done with the `Default` trait instead, i.e.

```rust
impl Default for Config {
    fn default() -> Self {
        // ...
    }
}
```

However, seeing as you have already implemented `Default` for each of the sub-config types (i.e. `Whitespace`, `Imports`, etc), then you can use Rust's short-hand derive to generate the implementation instead, e.g.

```rust
#[derive(Debug, Default)]
pub struct Config {
```

sway-fmt-v2/src/config/manifest.rs
```diff
@@ -0,0 +1,170 @@
...
+impl ConfigOptions {
+    /// Given an optional path to a `swayfmt.toml`, read it and construct a `Config`.
+    /// If settings are omitted, those fields will be set to default. If `None` is provided,
+    /// the default config will be applied. If a `swayfmt.toml` exists but is empty, the default
+    /// config will be applied.
+    ///
+    /// At present, this will only return a warning if it catches unusable fields.
+    /// Upon completion, this should give errors/warnings for incorrect input fields as well.
+    ///
+    pub fn from_dir_or_default(config_path: Option<&Path>) -> Result<Config> {
```
mitchmindtree, line 143: I think we should remove the `Option` around the `&Path`. If a user doesn't have a `&Path`, they should probably just call `Config::default()` directly.

Then, let's rename this to `from_dir` and refactor out a `from_file` function constructor which takes the direct path to the `swayfmt.toml`. Similar to how our `ManifestPath` type has both `from_file` and `from_dir` constructors. We should make sure the doc comments correspond accordingly too.

Also, this function should return `Result<Self>` instead of `Config`. When reading code it can be a bit confusing to see something that looks like a constructor for one type that actually produces a different type. We should do the conversion from `ConfigOptions` to `Config` in the equivalent `Config::from_file` and `Config::from_dir` constructors instead.

mitchmindtree (commented):
I just had a closer look through - all looking **very clean** so far, really great stuff! I've just added a couple more comments to address first.

So I guess the plan is to first land this work, and then start on implementations for each of the `Format` implementations for each of the `Item*`s? If so, sounds good to me! Can't wait to see those implementations start to fill out :)

Fwiw, when it comes time to work on the `Format` implementations, if it's easier, I'd be happy to see just the default configuration cases handled first for each item and leave a `todo!()` on the alternative configurations in favour of handling those later.

sway-fmt-v2/src/config/manifest.rs
eureka-cpu: Okay, I'll try to figure this out today, thanks again so much for all the help and feedback !

eureka-cpu: > I think we should remove the Option around the &Path. If a user doesn't have a &Path, they should probably just call Config::default() directly.

I'm a bit confused on this statement. I figured that the only way we wanted the user to interact with the formatter was via a `swayfmt.toml` and if a path to it doesn't exist then the formatter would just use default. In what way would the user be calling `Config::default()`? 

I was under the assumption that we wanted the config to be finalized before being given to the formatter. This way the only command the user ever needs is `forc fmt`. My apologies for the late reply bit of a busy few days here!

eureka-cpu: Upon further thought, maybe we could have it to where if the toml is absent, we just allow it to fail silently instead of accepting a none value?

eureka-cpu: Okay, I've gone ahead and refactored according to this but my question still stands!

sway-fmt-v2/src/config/expr.rs
```diff
@@ -0,0 +1,115 @@
...
+                .unwrap_or(TypeCombinatorLayout::Wide),
+            spaces_around_ranges: opts.spaces_around_ranges.unwrap_or(false),
+            match_block_trailing_comma: opts.match_block_trailing_comma.unwrap_or(false),
+            match_arm_leading_pipe: opts
+                .match_arm_leading_pipe
+                .unwrap_or(MatchArmLeadingPipe::Never),
+            force_multiline_blocks: opts.force_multiline_blocks.unwrap_or(false),
+            fn_args_layout: opts.fn_args_layout.unwrap_or(ItemsLayout::Tall),
+            fn_single_line: opts.fn_single_line.unwrap_or(false),
+        }
```
mitchmindtree, line 77: Currently, the default values for the `Expressions` fields are manually specified in two places:

1. In the implementation of `Default` for `Expressions` and
2. In the implementation of `Expressions::from_opts`.

This makes it easy to make mistakes when updating default values in the future. E.g. a value may only get updated in one place, but not the other. To avoid this, there should be a single source of truth for each default value.

One approach would be to specify each default value as a constant, and use the constant in both implementations.

Another way of achieving this might be to leave the `Expressions::default` implementation and use it within the `Expressions::from_opts`, so `Expressions::from_opts` would look something like:

```rust
impl Expressions {
    pub fn from_opts(opts: &ExpressionsOptions) -> Self {
        let default = Self::default();
        Self {
            expr_brace_style: opts
                .expr_brace_style
                .unwrap_or(default.expr_brace_style),
            trailing_semicolon: opts.trailing_semicolon.unwrap_or(default.trailing_semicolon),
            // etc
```

Whatever solution we use, we should apply it to the other sets of configuration options too.

sway-fmt-v2/src/config/manifest.rs
```diff
@@ -0,0 +1,137 @@
...
+                .unwrap_or_default(),
+            comments: opts
+                .comments
+                .as_ref()
+                .map(Comments::from_opts)
+                .unwrap_or_default(),
+        }
+    }
+    /// Given a directory to a forc project containing a `swayfmt.toml`, read and
+    /// construct a `Config` from `ConfigOptions`.
```
mitchmindtree, line 105: How about:

> Given a path to a forc project directory or one of its subdirectories, read and
> construct a `Config` from the project's `swayfmt.toml` configuration file.

mitchmindtree (changes requested):
Getting close!

sway-fmt-v2/src/config/manifest.rs
mitchmindtree: > I think we should remove the Option around the &Path. If a user doesn't have a &Path, they should probably just call Config::default() directly.

I think the confusion might come from my use of the word "user" in this sentence. I was referring to the user of the code, (i.e. caller of the function), not the downstream user of rustfmt. Does that make more sense?

---

> and if a path to it doesn't exist then the formatter would just use default (and not return an error, since it is okay for one to be absent).

I think the following is a more elegant solution for `sway-fmt` or `forc-fmt` code that checks for the `Config` file:

```rust
let config = Config::from_dir(path).unwrap_or_default();
```
rather than:
```rust
let config = Config::from_dir_or_default(Some(path));
```
My issue with the latter is that it would require manually checking if the `swayfmt.toml` exists before deciding whether to pass it `Some(path)` or `None`, whereas that check is already done inside `Config::from_dir` anyway.

That said, `unwrap_or_default` probably isn't the right choice here either, as we only want to fallback to the default config if the file doesn't exist, otherwise we want to give some useful error.

Rather than using `anyhow` for `from_dir`'s error type, we might be better off with a custom `ConfigError` using `thiserror`, e.g. something like the following

```rust
/// Errors encountered when loading a `Config` from a `swayfmt.toml` or directory.
#[derive(Debug, Error)]
pub enum ConfigError {
    #[error("failed to parse config: {err}")]
    Deseriallize {
        err: toml::Error,
    }
    #[error("could not find a `swayfmt.toml` in the given directory or its parents")]
    NotFound,
}
```

Then we could match on the result of `from_dir` to handle it properly:

```rust
let config = match Config::from_dir(path) {
    Ok(config) => config,
    Err(ConfigError::NotFound) => Config::default(),
    Err(err) => bail!(err),
};
```

eureka-cpu: This makes much more sense, as you can see I was trying to establish some logic around allowing the absence of the `swayfmt.toml` and ended up with some weird `Option` nonsense. Thanks for clearing this up, will have a commit later today.

sway-fmt-v2/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "sway-fmt-v2"
+version = "0.12.1"
```
adlerjohn, line 3:
```suggestion
version = "0.12.2"
```

sway-fmt-v2/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
```
adlerjohn, line 1: If we're publishing this to crates.io, can you add it to https://github.com/FuelLabs/sway/blob/c253486ed03414e851a6a43c5084b52d337979a4/.github/workflows/ci.yml#L468-L479

mitchmindtree: Perhaps we should not publish `sway-fmt-v2` and instead wait until it's at feature parity with the existing `sway-fmt`, then replace the old `sway-fmt`, then publish this as `sway-fmt` instead?

In the meantime, we can always `cargo install --path ./sway-fmt-v2` if we want to use/test it locally or in CI.

Maybe we should add a `publish = false` for now instead? Thoughts @eureka-cpu @adlerjohn ?

adlerjohn: Not against that. If we're not publishing then set the version to `0.0.0` as well

.gitignore
```diff
@@ -18,3 +18,6 @@ out
 # accidental upstream patch release breakage before users do.
 examples/*/Cargo.lock
 test/src/sdk-harness/Cargo.lock
+
+# These are unneeded mac generated files
+.DS_Store
```
mitchmindtree, line 23: I think it's generally better practice to have platform/editor/user-specific files under your own local gitignore file, otherwise we open the floodgates to filling the `.gitignore` with files unrelated to the project. [This SO post discusses how you can setup a local gitignore](https://stackoverflow.com/questions/5724455/can-i-make-a-user-specific-gitignore-file). All that said, as `.DS_Store` is such a common culprit I'm not against leaving it in.

mitchmindtree (changes requested):
Only things left are:

- Change version to `0.0.0`.
- Add `publish = false` flag in favour of eventually publishing as `sway-fmt` when it's ready.

After these I'm happy to approve!

.gitignore
eureka-cpu: I'll check back in tonight, lmk if I should remove the `.DS_Store`

mitchmindtree (approved):
:partying_face:

.gitignore
otrho: Yeah, take it out.

Issue #702, opened 2022-01-25
Add `forc --list`


adlerjohn:
Lists commands, similar to `cargo --list`

adlerjohn:
@ControlCplusControlV has requested to work on this, ref: https://github.com/FuelLabs/sway/issues/482#issuecomment-1021603180

adlerjohn:
This is no longer necessary to implementation completions, though it may be independently useful.

mitchmindtree:
I think this would still be quite useful for downstream tooling. E.g. the `forc-documenter` could use this to generate all the modules (rather than having to list them in the index.md).

We should also consider including available plugin commands in the output (I think `cargo --list` does this) so that downstream tools can process these altogether. Perhaps a flag could be provided for omitting plugins from the output.

ra0x3:
- Gonna unassign myself (for now) due to this largely being covered in https://github.com/FuelLabs/sway/issues/1215. 
- Of course we can always come back and complete this if the need is there

mitchmindtree:
Labelled as low-priority as `forc-documenter` will use distinct `forc --help` and `forc plugins` commands for documenting each anyway, and this isn't really blocking anything, but will still be useful command to have in the future.

bingcicle:
Would it be a good time to prioritise this issue perhaps? I think as we start to move plugins into their own repos (#2608, #2609 ) we might get more questions of "why is XYZ feature not listed under `forc help`?", which we are already starting to get now, since `forc deploy` and `forc run` have been moved into the `forc-client` plugin. We also have [forc-wallet](https://github.com/FuelLabs/forc-wallet) which already does not show up as a command in core forc.

cargo prompts you to 'see all commands with --list', which I think we can follow as a guide here perhaps.

![Screenshot 2022-09-08 at 5 24 34 PM](https://user-images.githubusercontent.com/25565268/189086299-55286bad-1995-4b0a-a987-524744620646.png)

mitchmindtree:
@bingcicle we do have the `forc plugins` command, e.g.

```
[mindtree@minddesk:~/programming/nix/fuel]$ forc plugins
Installed Plugins:
forc-deploy
forc-run
forc-fmt
forc-explore
forc-wallet
forc-lsp

[mindtree@minddesk:~/programming/nix/fuel]$ forc plugins --describe
Installed Plugins:
  forc-deploy           No description found for this plugin.
  forc-run              Run script project. Crafts a script transaction then sends it to a running node
[...]

PR #1428, opened 2022-04-29
Add forc --list

ra0x3:
### Description

This PR adds `forc --list` which should be similar to `cargo --list`

### Testing steps
- [ ] Run `forc --list` and you should see a list and description  of all installed forc plugins

### DIFF
--- a/forc/src/cli/commands/list.rs
+++ b/forc/src/cli/commands/list.rs
@@ -1,0 +1,30 @@
+use crate::cli::ListCommand;
+use crate::utils::plugin_descriptions::plugin_description;
+use anyhow::Result;
+use clap::Parser;
+use std::path::PathBuf;
+
+
+#[derive(Debug, Parser)]
+pub struct Command {}
+
+pub(crate) fn exec(command: ListCommand) -> Result<()> {
+    let ListCommand {} = command;
+
+    for path in crate::cli::plugin::find_all() {
+        print_plugin_and_description(path);
+    }
+    Ok(())
+}
+
+fn print_plugin_and_description(path: PathBuf) {
+    let plugin_name = path
+        .file_name()
+        .expect("Failed to read file name")
+        .to_str()
+        .expect("Failed to print file name");
+
+    let description = plugin_description(plugin_name);
+
+    println!("{}\t\t{}", plugin_name, description);
+}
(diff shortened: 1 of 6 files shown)


forc/src/utils/plugin_descriptions.rs
```diff
@@ -0,0 +1,27 @@
...
+    let e = plugin_name_to_enum(n);
+    match e {
+        Plugin::ForcFormat => "Forc plugin for running the Sway code formatter.".to_string(),
+        Plugin::ForcLsp => {
+            "Forc plugin for the Sway LSP (Language Server Protocol) implementation".to_string()
+        }
+        Plugin::ForcExplore => "Forc plugin for running the Fuel Block Explorer.".to_string(),
+        Plugin::None => "Unidentified plugin".to_string(),
+    }
+}
```
adlerjohn, line 27: Rather than hardcoding these, could we instead add a `--description` option to each plugin or something? We have to consider the case that third-party plugins are installed for which the source code isn't in the Sway repo.

forc/src/cli/commands/list.rs
```diff
@@ -0,0 +1,30 @@
...
+pub(crate) fn exec(command: ListCommand) -> Result<()> {
+    let ListCommand {} = command;
+
+    for path in crate::cli::plugin::find_all() {
+        print_plugin_and_description(path);
+    }
+    Ok(())
+}
+
+fn print_plugin_and_description(path: PathBuf) {
```
ra0x3, line 20: Obviously very similar to what is used by `plugins` command -- but with a tab separated description

forc/src/utils/plugin_descriptions.rs
```diff
@@ -0,0 +1,27 @@
+pub enum Plugin {
+    ForcFormat,
+    ForcLsp,
+    ForcExplore,
+    None,
+}
+
+pub fn plugin_name_to_enum(n: &str) -> Plugin {
```
ra0x3, line 8: Would obviously need an entire list of all `forc` plugins and their respective descriptions

forc/src/utils/plugin_descriptions.rs
ra0x3: @adlerjohn  I'm sure that's definitely possible. But the original issue doesn't mention this additional context. Could you add this context to https://github.com/FuelLabs/sway/issues/702. Obviously being newer I'm not familiar with anything relating to 3rd party stuff and the issue doesn't mention it

Issue #1017, opened 2022-03-24
Parse all Token's from document for LSP


JoshuaBatty:
Currently, when a document is opened in VScode, the editor sends the text to the server which then parses's it to collect the Tokens. In `sway-lsp` we have a `Vec<Token>` that is populated by matching on the parsed result returned from the `parse()` fn in `sway-core` which is a `SwayParseTree`. We can then access each individual `AstNode` by looping over the `sway_parse_tree.tree.root_nodes`.

The `content` field in `AstNode` can be either: 
```
pub enum AstNodeContent {
    UseStatement(UseStatement),
    ReturnStatement(ReturnStatement),
    Declaration(Declaration),
    Expression(Expression),
    ImplicitReturnExpression(Expression),
    WhileLoop(WhileLoop),
    IncludeStatement(IncludeStatement),
}
```

At the moment, we are only matching on the `Declaration` and `Expression` variants. There are further variants within those enums that are also not getting matched on. For example, we are not parsing the following `Declaration` variants.

```
Declaration::ImplTrait,
Declaration::ImplSelf,
Declaration::AbiDeclaration,
Declaration::ConstantDeclaration,
Declaration::StorageDeclaration,
```

In relation to `Expression`, we are matching on `CodeBlock` and `FunctionApplication` but are ignoring all of the following.

```
Expression::Literal,
Expression::LazyOperator,
Expression::VariableExpression,
Expression::Tuple,
Expression::StructExpression,
Expression::IfExp,
Expression::MatchExp,
Expression::AsmExpression,
Expression::MethodApplication,
Expression::SubfieldExpression,
Expression::DelineatedPath,
Expression::AbiCast,
Expression::ArrayIndex,
Expression::DelayedMatchTypeResolution,
Expression::IfLet,
Expression::SizeOfVal,
Expression::SizeOfType
``` 

I ended up looking at this section of the code when trying to understand why rust-analyzer was providing richer information regarding the same test file. It seems a good starting point before working on testing the current capabilities is to catch more of these unimplemented variants. 

@leviathanbeak did you have any thoughts on this?

leviathanbeak:
I started from the most basic ones and many of them were not even offered at that time.

Each of these could now be a ticket in itself - this issue could offer a starting point for the next feature implementation.

JoshuaBatty:
This has mostly been solved in #2121. Going to close for now and will reopen smaller issues if we notice that certain tokens aren't being collected by the server.

PR #1430, opened 2022-04-29
Collect tokens from missing SwayParseTree variants

JoshuaBatty:
This PR mostly addresses the missing `Declaration` and `Expression` variants of the `AstNode`s returned by the `SwayParseTree` as mentioned in this issue #1017.

Note, going forward I also plan on using the `TypedParseTree` for collecting the tokens. This is essential for gathering type information about some tokens, but also for accessing the `namespace` associated with each token.

### DIFF
--- a/sway-core/src/type_engine/engine.rs
+++ b/sway-core/src/type_engine/engine.rs
@@ -350,7 +350,7 @@ pub fn insert_type(ty: TypeInfo) -> TypeId {
     TYPE_ENGINE.insert_type(ty)
 }
 
-pub(crate) fn look_up_type_id(id: TypeId) -> TypeInfo {
+pub fn look_up_type_id(id: TypeId) -> TypeInfo {
     TYPE_ENGINE.look_up_type_id(id)
 }
 
--- a/sway-core/src/type_engine/engine.rs
+++ b/sway-core/src/type_engine/engine.rs
@@ -350,7 +350,7 @@ pub fn insert_type(ty: TypeInfo) -> TypeId {
     TYPE_ENGINE.insert_type(ty)
 }
 
-pub fn look_up_type_id(id: TypeId) -> TypeInfo {
+pub(crate) fn look_up_type_id(id: TypeId) -> TypeInfo {
     TYPE_ENGINE.look_up_type_id(id)
 }
 
--- a/sway-lsp/src/capabilities/hover.rs
+++ b/sway-lsp/src/capabilities/hover.rs
@@ -41,7 +41,7 @@ pub fn get_hover_data(session: Arc<Session>, params: HoverParams) -> Option<Hove
 
 fn get_hover_format(token: &Token, documents: &Documents) -> Hover {
     let value = match &token.token_type {
-        TokenType::Variable(var_details) => {
+        TokenType::VariableDeclaration(var_details) => {
             let var_type = match &var_details.var_body {
                 VarBody::FunctionCall(fn_name) => get_var_type_from_fn(fn_name, documents),
                 VarBody::Type(var_type) => var_type.clone(),
@@ -56,17 +56,21 @@ fn get_hover_format(token: &Token, documents: &Documents) -> Hover {
             )
         }
         TokenType::FunctionDeclaration(func_details) => func_details.signature.clone(),
-        TokenType::Struct(struct_details) => format!(
+        TokenType::StructDeclaration(struct_details) => format!(
             "{}struct {}",
             extract_visibility(&struct_details.visibility),
             &token.name
         ),
-        TokenType::Trait(trait_details) => format!(
+        TokenType::TraitDeclaration(trait_details) => format!(
             "{}trait {}",
             extract_visibility(&trait_details.visibility),
             &token.name
         ),
-        TokenType::Enum => format!("enum {}", &token.name),
+        TokenType::EnumDeclaration(enum_details) => format!(
+            "{}enum {}",
+            extract_visibility(&enum_details.visibility),
+            &token.name
+        ),
         _ => token.name.clone(),
     };
 
--- a/sway-lsp/src/capabilities/semantic_tokens.rs
+++ b/sway-lsp/src/capabilities/semantic_tokens.rs
@@ -71,28 +71,39 @@ fn create_semantic_token(next_token: &Token, prev_token: Option<&Token>) -> Sema
     }
 }
 
-// these values should reflect indexes in `token_types`
-static FUNCTION: u32 = 1;
-static LIBRARY: u32 = 3;
-static PARAMETER: u32 = 5;
-static VARIABLE: u32 = 9;
-static ENUM: u32 = 10;
-static STRUCT: u32 = 11;
-static TRAIT: u32 = 12;
+/// these values should reflect indexes in `token_types`
+#[repr(u32)]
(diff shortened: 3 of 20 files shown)


sway-lsp/src/core/token.rs
```diff
@@ -101,81 +116,340 @@ pub fn traverse_node(node: AstNode, tokens: &mut Vec<Token>) {
...
+            tokens.push(token);
+
+            handle_expression(*address, tokens);
+        }
+        Expression::ArrayIndex { prefix, index, .. } => {
+            handle_expression(*prefix, tokens);
+            handle_expression(*index, tokens);
+        }
+        Expression::DelayedMatchTypeResolution { .. } => {
+            //Should we handle this since it gets removed during type checking anyway?
```
mitchmindtree, line 426: Good Q: Is this how the parser represents the `_` syntax? I.e. `let foo: Vec<_> = ...` where `_` explicitly indicates that the type should be inferred? If so, it would be cool if `sway-lsp` could show what type is inferred when you hover over it. That said, that would all have to wait til we introduce type checking anyways.

sway-lsp/src/utils/function.rs
```diff
@@ -1,6 +1,6 @@
-use sway_core::FunctionDeclaration;
+use sway_types::Span;
 
-pub(crate) fn extract_fn_signature(func_dec: &FunctionDeclaration) -> String {
-    let value = func_dec.span.as_str();
+pub(crate) fn extract_fn_signature(span: &Span) -> String {
```
mitchmindtree, line 3: If we change this to take a span directly instead of a `FunctionDeclaration`, we should add a doc comment to the function that states what assumptions we make about the given `Span`. E.g. it expects a span from either a `FunctionDeclaration`, trait function, or whatever other declarations this is used for. Should probably do the same for `get_function_details` too.

mitchmindtree (commented):
Epic, looks like you're getting into the swing of it!

sway-core/src/type_engine/engine.rs
```diff
@@ -350,7 +350,7 @@ pub fn insert_type(ty: TypeInfo) -> TypeId {
     TYPE_ENGINE.insert_type(ty)
 }
 
-pub(crate) fn look_up_type_id(id: TypeId) -> TypeInfo {
+pub fn look_up_type_id(id: TypeId) -> TypeInfo {
```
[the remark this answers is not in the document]
JoshuaBatty: Agh good spot, was experimenting with the `TypedParseTree` and must have missed this.

sway-lsp/src/capabilities/hover.rs
```diff
@@ -56,17 +56,21 @@ fn get_hover_format(token: &Token, documents: &Documents) -> Hover {
...
         ),
-        TokenType::Trait(trait_details) => format!(
+        TokenType::TraitDeclaration(trait_details) => format!(
             "{}trait {}",
             extract_visibility(&trait_details.visibility),
             &token.name
         ),
-        TokenType::Enum => format!("enum {}", &token.name),
+        TokenType::EnumDeclaration(enum_details) => format!(
+            "{}enum {}",
```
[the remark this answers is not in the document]
JoshuaBatty: Yeah, `extract_visibility` produces a string with trailing space. Which isn't very intuitive, will change this now.

sway-lsp/src/capabilities/semantic_tokens.rs
```diff
@@ -74,19 +74,23 @@ fn create_semantic_token(next_token: &Token, prev_token: Option<&Token>) -> Sema
 // these values should reflect indexes in `token_types`
 static FUNCTION: u32 = 1;
 static LIBRARY: u32 = 3;
+static PARAMETER: u32 = 5;
 static VARIABLE: u32 = 9;
 static ENUM: u32 = 10;
 static STRUCT: u32 = 11;
 static TRAIT: u32 = 12;
```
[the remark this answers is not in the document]
JoshuaBatty: I like the idea of having an `enum` with `#[repr(u32)]`. Easy change.

sway-lsp/src/core/token.rs
JoshuaBatty: Yeah this would be cool, might need to wait for type checking as you say.

sway-lsp/src/utils/function.rs
JoshuaBatty: Added comments

JoshuaBatty:
Thanks @mitchmindtree, have addressed each of your comments.

sway-lsp/src/capabilities/semantic_tokens.rs
```diff
@@ -71,24 +71,39 @@ fn create_semantic_token(next_token: &Token, prev_token: Option<&Token>) -> Sema
...
+    _Operator,      // 4
+    Parameter,      // 5
+    _String,        // 6
+    _Type,          // 7
+    _TypeParameter, // 8
+    Variable,       // 9
+    Enum,           // 10
+    Struct,         // 11
+    Interface,      // 12
+}
```
mitchmindtree, line 90: Rather than including variants for the currently unsupported types, I think you could also do something like:

```rust
#[repr(u32)]
enum TokenTypeIndex {
    Function = 1,
    Namespace = 3,
    Parameter = 5,
    Variable = 9,
    Enum = 10,
    Struct = 11,
    Interface = 12,
}
```

JoshuaBatty: Agh nice, didn't know that. Updating now.

mitchmindtree (approved):
Giving this a utACK tick, but I'll cc @sezna just to check if its OK to make these parse-tree type fields `pub`?

I feel like I remember some talk of potentially removing `sway-core`'s `parse-tree` module in favour of the new `sway-parse` stuff at some point, but even if that's the case I don't know if that should block this as I guess we could just replace this at the same time.

adlerjohn (commented):
Looks fine to me, will defer to @sezna

sezna:
>  I remember some talk of potentially removing sway-core's parse-tree module in favour of the new sway-parse stuff at some point

The way things currently work is:
1. `sway-parse` produces a [`Vec<Item>`](https://github.com/FuelLabs/sway/blob/master/sway-parse/src/item/mod.rs#L15) where the items are potentially nested top level things
2. We convert the output of `sway-parse` to the old `ParseTree`
3. The old pipeline continues as normal, unaware of the fact that we didn't use the old parser to generate the parse tree.

Eventually we'd want to skip step 2 above, but it isn't a priority since it works. It'll probably get prioritized after all major features are stabilized, then we can start cleaning up stale/non-optimal code etc.


It is totally fine to make these parse tree items public, as this PR does -- `sway-core`'s external API is more or less just whatever the tooling needs. @JoshuaBatty should look into using `sway-parse::Item` instead, though, as it would be faster and the items are a bit more well-formed.

sezna (approved):
LGTM but see above comment about potentially switching the lsp over to `sway-parse`.

JoshuaBatty:
Awesome, thanks for that @sezna! I hadn't looked into `sway-parse::Item` yet. Appreciate the feedback, will look at integrating this into `sway-lsp`. Thanks :)

mitchmindtree (approved):
Re-approving after successful rebase :+1:

PR #1435, opened 2022-04-29
Script test helper

nfurfaro:
This adds a simple test helper used to run scripts via the SDK.
As the project grows we'll have more an d more tests running scripts, and every time the SDK releases breaking changes we'll have to update the syntax in numerous places.
This adds the `script_runner()` function which can be used in SDK tests like this:
```
use test_helpers::script_runner;

#[tokio::test]
async fn contract_id_eq_implementation() {
    let path_to_bin = "test_projects/contract_id_type/out/debug/contract_id_type.bin";
    let return_val = script_runner(path_to_bin).await;
    assert_eq!(1, return_val);
}
```

Curent limitations: ATM, it just returns a  `u64` value to compare against, which works well for scripts that return `bool` values or `u64`s.
A future improvement would be to have `script_runner` return an array of `Receipt`s perhaps.

### DIFF
--- a/test/src/sdk-harness/Cargo.toml
+++ b/test/src/sdk-harness/Cargo.toml
@@ -19,6 +19,9 @@ hex = "0.4.3"
 rand = "0.8"
 tokio = { version = "1.12", features = ["rt", "macros"] }
 
+[dev-dependencies]
+test-helpers = { path = "./test-helpers"}
+
 [[test]]
 harness = true
 name = "integration_tests"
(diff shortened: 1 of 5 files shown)


simonr0204 (approved):

otrho:
My first thoughts seeing this was 'if you have a script oriented test then just add it to the E2Es' but this is more to have some supplementary tests attached to other std-lib tests, and to keep them all localised?

nfurfaro:
@otrho yeah, some of the stdlib tests don't require contracts, and running a script is perfectly adequate.
I'd like to keep all the stdlib tests in one place though, and as the SDK allows testing scripts as well as contracts we've been doing these kind of tests under the sdk-harness. Basically organizing the tests by _what_ we're testing rather than _how_ we're testing if that makes sense.

PR #1438, opened 2022-04-29
Rust 1.60 docker update

Voxelot:
- update docker image to rust 1.60
- use cargo-chef to improve image building performance
- removed redundant binary target definition (due to https://github.com/LukeMathWalker/cargo-chef/issues/128)


adlerjohn (commented):
If we're going to bump the Rust version in the docker file, shouldn't we also do it for CI?

Voxelot:
> If we're going to bump the Rust version in the docker file, shouldn't we also do it for CI?

@adlerjohn it looks like CI automatically uses the latest rust version: https://github.com/FuelLabs/sway/runs/6236542491?check_suite_focus=true#step:3:15

adlerjohn (commented):
Looks fine to me, but would like @sezna and @mitchmindtree / @JoshuaBatty to give a look.

mitchmindtree:
Do we document anywhere what this Dockerfile is for or how we use it? I've never used Docker and am currently unaware of how we use this in the Fuel ecosystem.

JoshuaBatty:
I'm in the exact same boat as Mitch around Docker, unfortunately. Would love to read up on how it's used in Fuel if such a document exists?

sezna:
> Do we document anywhere what this Dockerfile is for or how we use it? 

I don't think we document it, but this is the base image that our Github runners use when running our CI. So this change makes our CI use Rust 1.60.

Voxelot:
All I know is that it creates a docker image with a compiled version of forc. @sezna is this currently being used?

sezna (approved):

PR #1440, opened 2022-04-30
Make all string types non-copy

### LINKED ISSUE #1424: Pass small strings (<= 8bytes) in an out of a contract using a "pointer" (and `retd` instead of `ret`)
That is, treat small strings like large strings. This was actually the case until recently.. We probably just have to return https://github.com/FuelLabs/sway/pull/1292 as per the discussion around https://github.com/FuelLabs/sway/issues/1368

mohammadfawaz:
This was the old behavior which was changed in #1292. We agreed to bring back the old behavior for consistency. So, the compiler now does not differentiate between small strings (<= 8 bytes) and large strings.

Closes #1424.

### DIFF
--- a/sway-core/tests/ir_to_asm/takes_string_returns_string.asm
+++ b/sway-core/tests/ir_to_asm/takes_string_returns_string.asm
@@ -1,0 +1,27 @@
+.program:
+ji   i4
+noop
+DATA_SECTION_OFFSET[0..32]
+DATA_SECTION_OFFSET[32..64]
+lw   $ds $is 1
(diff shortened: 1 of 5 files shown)


sezna (approved):

mohammadfawaz:
There are some CI failures that I need to address, which I will do asap.

sezna:
Maybe make it a draft until CI is ready to go?

sezna:
Per our slack convo, I pushed a commit here, hope you don't mind 😄

sezna (approved):

PR #1441, opened 2022-04-30
Fix `forc run` default URL

mohammadfawaz:
The default url needs to start with `http://` now.


binggh:
The commit was me - took the liberty to fetch your branch and run the script instead (hope you don't mind it) :p Turns out it was indeed an issue of running the script with the local `forc` installed. I will open a PR to update the CI message with clearer instructions.

mitchmindtree (approved):

PR #1443, opened 2022-05-01
Clearer failure message for forc-documenter

binggh:
Related: #1441 

The CI gives an [unclear action message](https://github.com/FuelLabs/sway/runs/6241495673?check_suite_focus=true#step:6:27) with regards to running `forc-documenter` after local changes to forc commands. The right action should be to install `forc` from path so that the developer actively working on local changes receives the latest `forc --help`, before running the script.


sezna (approved):
Seems reasonable. This seems like a common case that would be hit when using `forc-documenter`

mitchmindtree (approved):

PR #1445, opened 2022-05-02
Updated toml file to exclude all examples using glob operator

### LINKED ISSUE #1444: The Cargo.toml file should update the `exclude` section for examples to use the glob operator to exclude all of the examples
Currently there are a few explicitly specified examples in the `exclude` section however more examples have been added and will continue to be added.
It's better to catch them all automatically instead of excluding them 1-by-1.

Braqzen:
Closes #1444


JoshuaBatty:
I can't remember exactly why this was added but I'm not sure we still need to have this exclude block. Vaguely remember @mitchmindtree saying it could probably be removed. Perhaps we can merge this for now and remove this in a separate PR once @mitchmindtree can confirm this or not.

JoshuaBatty (approved):

sezna (approved):

PR #1450, opened 2022-05-02
Add --locked to cargo CI

ra0x3:
### Description
This PR adds `--locked` to CI `cargo` commands 

### Testing Steps
- [ ] CI should pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -336,6 +336,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
+          args: --locked
 
   forc-documenter-dry-run:
     needs: cancel-previous-runs
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -183,7 +183,7 @@ jobs:
         name: Cargo Build Workspace
         with:
           command: build
-          args: --workspace --all-features --all-targets
+          args: --locked --workspace --all-features --all-targets
         env:
           RUSTFLAGS: '-D warnings'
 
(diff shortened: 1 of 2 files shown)


ra0x3:
`cargo-test-lib-std` fails on this with:
> Run actions-rs/cargo@v1
/home/runner/.cargo/bin/cargo test --locked --manifest-path ./test/src/sdk-harness/Cargo.toml -- --test-threads=1 --nocapture
    Updating crates.io index
error: the lock file /home/runner/work/sway/sway/test/src/sdk-harness/Cargo.lock needs to be updated but --locked was passed to prevent this
If you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
Error: The process '/home/runner/.cargo/bin/cargo' failed with exit code 101

- Which seems to be a good thing? 
- Basically trying to figure out whether or not the lock file generated for the `cargo-test-lib-std` job 
- I see there's a comment in the `ci.yml` file
>  TODO: Remove this upon merging std tests with the rest of the E2E tests.
- Which might possibly explain why this job fails?

CC @JoshuaBatty  @mitchmindtree

mitchmindtree:
> error: the lock file /home/runner/work/sway/sway/test/src/sdk-harness/Cargo.lock needs to be updated but --locked was passed to prevent this

Yeah I think it's working as intended! It looks like we need to update the `test/src/sdk-harness` project to include its `Cargo.lock`. It looks like it hasn't been included as we have `Cargo.lock` in our gitignore, which is surprising considering our repo still tracks the workspace `Cargo.lock` :thinking: I've opened #1497.

mitchmindtree (approved):
Nice, LGTM!

JoshuaBatty (approved):
Nice one, looks good.

PR #1452, opened 2022-05-02
Adds ethAddress type

simonr0204:
Adds ethAddress type


adlerjohn (commented):
Would `EvmAddress` be preferred to `EthAddress`?

simonr0204:
> Would `EvmAddress` be preferred to `EthAddress`?

Agreed. Although actually I wanted to just use `Address` since it's already namespaced to `vm::evm` but it seemed to be colliding with the base `Address` from the std-lib.

mohammadfawaz:
> > Would `EvmAddress` be preferred to `EthAddress`?
> 
> Agreed. Although actually I wanted to just use `Address` since it's already namespaced to `vm::evm` but it seemed to be colliding with the base `Address` from the std-lib.

The collision _should_ only happen if both are imported in the same scope. Otherwise, this is probably a bug.

simonr0204:
hmm they're not both imported, but I couldnt work out how to get the one I was after unambiguously. Possibly my Rust not being up to scratch. If anyone can see a way to it that would be awesome.

mohammadfawaz:
> hmm they're not both imported, but I couldnt work out how to get the one I was after unambiguously. Possibly my Rust/Sway not being up to scratch. If anyone can see a way to it that would be awesome.

Writing the full path directly (without importing either `Address`s) should work... as in: `path::to::evm::Address` and `path::to::Address`.

Braqzen:
Since there is a distinction being made between an address for Fuel / Ethereum I think there should also be an addendum to the recovery function.
Currently there is `ec_recover_address(...)` however the aforementioned distinction is now ambiguous / misleading and thus should there should be an additional function `ec_recover_evm_address(...)`.

There may be an issue here since the caller may not know which function to call and thus they may need to call both and perform checks on the result.
Alternatively, having only 1 function which returns a super type of `Address` and `EVMAdress` (or an error) allows for easy match statements (when implemented).

If I'm ignorant and you can easily determine which of the 2 functions would be called then I'd prefer to have two separate functions.
If you cannot determine ahead of the call then I prefer the single function with a super type since matching will make control flow easy.

adlerjohn:
In favor of different function names, against a sum type for both Fuel and EVM addresses in the stdlib. Enshrining EVM addresses across the stdlib rather than self-contained within the `evm` namespace isn't a good idea.

simonr0204:
There currently are two separate address types, and two separate corresponding ecrecover functions. The EVM stuff is all under the `vm::evm` namespace. That would allow `std::address` and `std::vm::evm::Address` to both exist. In my mind calling the latter `std::vm::evm::EVMAddress` was a bit redundant.

SilentCicero:
@simonr0204 please also update the relevant sections in the Sway book for this PR. I.e. anything it effects on the Crypto side.

nfurfaro:
>In my mind calling the latter std::vm::evm::EVMAddress was a bit redundant.

Kind of... although when scanning code which instantiates an address, you would just be seeing `~Address::from(...` and it won't be obvious which type of address this is without also looking at all the `use` statements.

adlerjohn:
I would also vote for two different names. There's lots of times where both addresses and EVM addresses will be used in the same code.

docs/src/blockchain-development/hashing_and_cryptography.md
```diff
@@ -13,3 +13,5 @@ The Sway standard library provides easy access to a selection of cryptographic h
 ```sway
 {{#include ../../../examples/signatures/src/main.sw}}
 ```
+
+Note: Recovery of consensus layer addresses (e.g. 20-byte Ethereum addresses) is also supported via modules in `std::vm`
```
adlerjohn, line 17:
```suggestion
> **Note** Recovery of EVM addresses is also supported via `std::vm::evm`.
```

docs/src/introduction/standard_library.md
```diff
@@ -2,7 +2,7 @@
 
 Similar to Rust, Sway comes with its own standard library.
 
-The Sway Standard Library is the foundation of portable Sway software, a set of minimal shared abstractions for the broader Sway ecosystem. It offers core types, like `Result<T, E>` and `Option<T>`, library-defined operations on language primitives, native asset management, blockchain contextual operations, access control and storage management, among many other things.
+The Sway Standard Library is the foundation of portable Sway software, a set of minimal shared abstractions for the broader Sway ecosystem. It offers core types, like `Result<T, E>` and `Option<T>`, library-defined operations on language primitives, native asset management, blockchain contextual operations, access control, storage management, and support for consensus layer VM types, among many other things.
```
adlerjohn, line 5:
```suggestion
The Sway Standard Library is the foundation of portable Sway software, a set of minimal shared abstractions for the broader Sway ecosystem. It offers core types, like `Result<T, E>` and `Option<T>`, library-defined operations on language primitives, native asset management, blockchain contextual operations, access control, storage management, and support for types from other VMs, among many other things.
```

sway-lib-std/src/vm/evm/ecr.sw
```diff
@@ -1,16 +1,15 @@
...
-use ::address::Address;
 use ::b512::B512;
 use ::context::registers::error;
 use ::ecr::{EcRecoverError, ec_recover};
+use ::vm::evm::evm_address::EvmAddress;
 use ::hash::keccak256;
 use ::result::*;
 
-/// Recover the address derived from the private key used to sign a message.
+/// Recover the ethereum address derived from the private key used to sign a message.
```
adlerjohn, line 10:
```suggestion
/// Recover the EVM address derived from the private key used to sign a message.
```

sway-lib-std/src/vm/evm/ecr.sw
```diff
@@ -20,13 +19,8 @@ pub fn ec_recover_address(signature: B512, msg_hash: b256) -> Result<Address, Ec
         let pub_key = pub_key_result.unwrap();
 
         // Note that Ethereum addresses are derived from the Keccak256 hash of the pubkey (not sha256)
```
adlerjohn, line 21:
```suggestion
        // Note that EVM addresses are derived from the Keccak256 hash of the pubkey (not sha256)
```

sway-lib-std/src/vm/evm/evm_address.sw
```diff
@@ -0,0 +1,40 @@
...
+} {
+    fn into(addr: EvmAddress) -> b256 {
+        addr.value
+    }
+}
+
+/// Functions for casting between the b256 and Address types.
+impl From for EvmAddress {
+    fn from(bits: b256) -> EvmAddress {
+        // An ethereum address is only 20 bytes, so the first 12 are set to zero
```
adlerjohn, line 31:
```suggestion
        // An EVM address is only 20 bytes, so the first 12 are set to zero
```

PR #1453, opened 2022-05-02
Add --describe -d opt to forc plugins

ra0x3:
### Description
- This PR adds flags to include long descriptions for `forc plugins`

### Testing steps
- [ ] Try passing `-d` and/or `--describe` to `forc plugins`

Output should resemble
```text
➜  sway git:(rashad/1215-add-forc-plugins-opt) ✗ forc plugins --describe
Installed Plugins:
  forc-explore          Forc plugin for running the Fuel Block Explorer.
  forc-lsp              Forc plugin for the Sway LSP (Language Server Protocol) implementation.
  forc-fmt              Forc plugin for running the Sway code formatter.
````
### Notes
- Every `forc plugins` description starts with `Forc plugin for...` which is a little redundant. 
  - We know it's a plugin because it's listed via `forc plugin`
- I see `cargo` has something similar to:
```text
➜  sway git:(rashad/1215-add-forc-plugins-opt) ✗ cargo --list
Installed Commands:
    add
    audit
    b                    alias: build
    bench                Execute all benchmarks of a local package
    build                Compile a local package and all of its dependencies
    c                    alias: check
    check                Check a local package and all of its dependencies for errors
    clean                Remove artifacts that cargo has generated in the past
    clippy               Checks a package to catch common mistakes and improve your Rust code.
    config               Inspect configuration values
    d                    alias: doc
    doc                  Build a package's documentation
```
- If there is buy in, we could potentially either (1) update these descriptions to remove the redundant prefix, or (2) Add a separate type of description

### DIFF
--- a/docs/src/forc/commands/forc_run.md
+++ b/docs/src/forc/commands/forc_run.md
@@ -12,7 +12,7 @@ forc run [OPTIONS] [NODE_URL]
 URL of the Fuel Client Node
 
 [env: FUEL_NODE_URL=]
-[default: http://127.0.0.1:4000]
+[default: 127.0.0.1:4000]
 
 
 ## OPTIONS:
--- a/docs/src/forc/commands/forc_run.md
+++ b/docs/src/forc/commands/forc_run.md
@@ -12,7 +12,7 @@ forc run [OPTIONS] [NODE_URL]
 URL of the Fuel Client Node
 
 [env: FUEL_NODE_URL=]
-[default: 127.0.0.1:4000]
+[default: http://127.0.0.1:4000]
 
 
 ## OPTIONS:
--- a/forc/src/cli/commands/plugins.rs
+++ b/forc/src/cli/commands/plugins.rs
@@ -42,12 +42,18 @@ fn parse_description_for_plugin(plugin: &str) -> String {
         .expect("Could not get plugin description.");
 
     let stdout = String::from_utf8_lossy(&proc.stdout);
-    stdout
-        .split('\n')
-        .nth(1)
-        .map(|x| if x.is_empty() { default_description } else { x })
-        .unwrap()
-        .to_owned()
+
+    // If the plugin doesn't support a -h flag
+    match stdout.split('\n').nth(1) {
+        Some(x) => {
+            if x.is_empty() {
+                default_description.to_owned()
+            } else {
+                x.to_owned()
+            }
+        }
+        None => default_description.to_owned(),
+    }
 }
 
 /// # Panics
--- a/forc/src/cli/commands/plugins.rs
+++ b/forc/src/cli/commands/plugins.rs
@@ -11,15 +11,57 @@ pub struct Command {
     /// Prints the absolute path to each discovered plugin.
     #[clap(long = "paths", short = 'p')]
     print_full_path: bool,
+    /// Prints the long description associated with each listed plugin
+    #[clap(long = "describe", short = 'd')]
+    describe: bool,
 }
 
 pub(crate) fn exec(command: PluginsCommand) -> Result<()> {
-    let PluginsCommand { print_full_path } = command;
+    let PluginsCommand {
+        print_full_path,
+        describe,
+    } = command;
 
+    println!("Installed Plugins:");
     for path in crate::cli::plugin::find_all() {
-        print_plugin(path, print_full_path);
+        print_plugin(path, print_full_path, describe);
     }
     Ok(())
+}
+
+fn parse_description_for_plugin(plugin: &str) -> String {
+    use std::process::Command;
+    let proc = Command::new(plugin)
+        .arg("-h")
+        .output()
+        .expect("Could not get plugin description.");
+
+    let stdout = String::from_utf8_lossy(&proc.stdout);
+    stdout
+        .split('\n')
+        .nth(1)
+        .unwrap_or("No description found for this plugin.")
+        .to_owned()
+}
+
+fn format_print_description(path: PathBuf, print_full_path: bool, describe: bool) -> String {
+    let display = if print_full_path {
+        path.display().to_string()
+    } else {
+        path.file_name()
+            .expect("Failed to read file name")
+            .to_str()
+            .expect("Failed to print file name")
+            .to_string()
+    };
+
+    let description = parse_description_for_plugin(&display);
+
+    if describe {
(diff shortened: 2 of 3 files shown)


forc/src/cli/commands/plugins.rs
```diff
@@ -11,33 +11,68 @@ pub struct Command {
...
     for path in crate::cli::plugin::find_all() {
-        print_plugin(path, print_full_path);
+        print_plugin(path, print_full_path, describe);
     }
     Ok(())
 }
 
+fn parse_description_for_plugin(plugin: &str) -> String {
+    use std::process::Command;
+    let proc = Command::new(plugin)
```
ra0x3, line 34: I'm not sure if there's a more idiomatic way to grab plugin descriptions

forc/src/cli/commands/plugins.rs
```diff
@@ -11,33 +11,68 @@ pub struct Command {
...
+
+    let stdout = String::from_utf8_lossy(&proc.stdout);
+    stdout
+        .split('\n')
+        .nth(1)
+        .unwrap_or("No description found for this plugin.")
+        .to_owned()
+}
+
+fn format_print_description(path: PathBuf, print_full_path: bool, describe: bool) -> String {
```
ra0x3, line 47: Not sure which functions need docstrings vs which don't. Didn't add one to these since they're private - but happy to go back and add a docstring

JoshuaBatty: Yeah for non public functions then it's at the discretion of the dev if they do this. I'd say in general its good practice to document each function so it's easier for other developers to understand when they come across it.

forc/src/cli/commands/plugins.rs
JoshuaBatty: Not that i'm aware of currently. There is a `print_help` method in `Clap` but I don't think that helps us here. https://docs.rs/clap/latest/clap/struct.App.html#method.print_help

JoshuaBatty:
Thanks for making those changes, looking really good. Just pulled down your changes and tried running `forc plugins --describe`. My last nit is that (at least for me) the process seems to be hanging somewhere after printing the installed plugins to the terminal. It does seem to exist and allow for typing commands again but seems to take around 20 seconds or so for this to happen. 

Is this happening for you also? If so, would you mind seeing what the culprit is and if you can find a way to gracefully exit after the plugins have printed?

ra0x3:
@JoshuaBatty 
- That's super interesting 🤔
- I'm attaching a screen-recording of the behavior I get
- I don't see the delay - or any delay at all. 
- I'm wondering if you have many more plugins installed than I do? I only have 4

https://user-images.githubusercontent.com/17253182/166612143-4e665964-0844-40ec-976f-59f82edb42fd.mov

ra0x3:
@JoshuaBatty 
- So I tried to pursue the "many plugins cause something to hang" angle with no luck
- Created several fake plugins, but didn't notice anything hanging or taking a long time
- However, I did find a little (possible) bug
  - Where if the plugin had no `-h` flag, the call would `panic`. Not sure if that's even a bug cause not sure if we require a `-h` flag. Either way, I addressed that in the latest commit

https://user-images.githubusercontent.com/17253182/166615501-adbb69dd-e705-405e-8fcd-e38ec63c27f4.mov

JoshuaBatty:
hmm that's very strange. I was running your branch in WSL on Ubuntu. Just tried on my macbook pro and I'm now seeing that everything is returning instantly as well. Seems to be a secondary issue going on unrelated to this PR.

forc/src/cli/commands/plugins.rs
```diff
@@ -11,33 +11,88 @@ pub struct Command {
...
+    let default_description = "No description found for this plugin.";
+    let proc = Command::new(plugin)
+        .arg("-h")
+        .output()
+        .expect("Could not get plugin description.");
+
+    let stdout = String::from_utf8_lossy(&proc.stdout);
+
+    // If the plugin doesn't support a -h flag
+    match stdout.split('\n').nth(1) {
```
ra0x3, line 47: Someone feel free to suggest a cool one-liner for this

docs/src/forc/commands/forc_run.md
```diff
@@ -12,7 +12,7 @@ forc run [OPTIONS] [NODE_URL]
 URL of the Fuel Client Node
 
 [env: FUEL_NODE_URL=]
-[default: http://127.0.0.1:4000]
+[default: 127.0.0.1:4000]
```
mohammadfawaz, line 15: Is this intentional? It was changed recently due to a change to `forc run`.

ra0x3: @mohammadfawaz 
- This is the result of `cargo run --bin forc-documenter write-docs`. 
- Was this a recent (as of a few hours ago today) change?

JoshuaBatty:
I'll need to look into why forc documenter isn't passing CI, perhaps @binggh might be able to provide some faster insight?

docs/src/forc/commands/forc_run.md
binggh: @ra0x3 you have to have the latest `forc` in your path, or have installed your local `forc` (if you have made changes to forc)

mohammadfawaz: This is the relevant change: https://github.com/FuelLabs/sway/pull/1441

ra0x3: @binggh 
- I'm using the latest version from crates.io (I believe). Should I be using a different version? 

```bash
➜  foo $(which forc) --version
forc 0.11.0
```

binggh: The script works by pulling the output from `forc --help` to generate documentation - since you've made changes to `forc plugins` command locally, the version on crates.io would not have your desired changes. You have to install the local forc (`cargo install --path ./forc`), before running the documenter.

ra0x3: @mohammadfawaz @binggh Ah,  _local forc vs cargo forc_. Thanks 👌

binggh (approved):
LGTM!

adlerjohn (commented):
lgtm, but will defer to one of @JoshuaBatty or @mitchmindtree

JoshuaBatty (approved):
Nice LGTM!

PR #1454, opened 2022-05-03
Adding a `bitcast` to make sure that the IR generated for storage accesses is legal

### LINKED ISSUE #1451: Booleans in `storage` blocks don't seem to work
Example:
```rust
contract;

abi Escrow {
    fn constructor();
}

storage {
    approved: bool
}

impl Escrow for Contract {
    fn constructor() {
        storage.approved = true;
    }
}
```
Error:
```
error: Internal compiler error: Verification failed: State access operation must be to a u64 pointer.
```
Same code with `u64` doesn't have the issue.

mohammadfawaz:
closes #1451 

Even though we're changing how storage is implemented, this is a trivial fix and the test that I keep expanding is going to be useful for coverage in the future.

### DIFF
--- a/sway-core/src/optimize.rs
+++ b/sway-core/src/optimize.rs
@@ -2102,13 +2102,29 @@ impl FnCompiler {
                     Type::Uint(_) | Type::Bool => {
                         // These types fit in a word. use state_store_word/state_load_word
                         match access_type {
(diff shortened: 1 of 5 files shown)


otrho (approved):
Awesome.

Issue #1128, opened 2022-04-02
Investigate potential bugs arising from apparent compromises in #1115.


otrho:
Take a look at the comments in the #1115 PR about stack overflow and out of gas.

otrho:
Also the bit where we need nested `if let`s to match a particular pattern.

mohammadfawaz:
I had someone at the workshop face a functional issue with an early return for a simple primarily checker function.. likely related to this.

I will try to reproduce if I can remember.

sezna:
Nested `if let`  was not implemented in #625 as we need a richer pattern matching syntax in order to allow for that. There was a chicken/egg problem where desugaring enums required something to desugar to and we weren't sure what the right path there was.

The plan now is to figure out how desugaring should work in an IR world.

otrho:
I'm getting `RESERV00` errors when investigating this still, though a new release of the SDK and associates tools is coming which should help out there.

PR #1456, opened 2022-05-03
Streamline the `is_reentrant()` library function.

otrho:
I had some strange errors I think related to trying to get the previous call frame with the current one is zero.  Otherwise this change isn't much different to the existing code, just a little simpler.


otrho:
Closes #1128

sezna (approved):

nfurfaro (approved):
Looks good @otrho .
This has been a surprisingly tricky feature to land in a clean way.

Issue #1458, opened 2022-05-03
Fix the build-forc-test-project CI job to correct test the local forc


mohammadfawaz:
See failure in https://github.com/FuelLabs/sway/pull/1457

The failure is related to the fact that building the project created with `forc init` requires the standard library version `0.12.0` which doesn’t exist yet.

Potentially, that job should change the manifest file of the generated project to add an explicit std dep that's local.

ra0x3:
<img width="1678" alt="Screen Shot 2022-05-03 at 7 18 12 AM" src="https://user-images.githubusercontent.com/17253182/166470882-4dab165f-eabf-4889-b23b-003e5854ec79.png">

adlerjohn:
This could be as simple as `echo "std = { path = \"../sway-lib-std/\" }" >> test-proj/Forc.toml` before the `forc build`.

ra0x3:
@adlerjohn Ah, yes. There are no actual dependencies in the file anyway 👍

PR #1459, opened 2022-05-03
Update manifest for build-forc-test-project

ra0x3:
### Description
- This PR updates the `build-forc-test-project` CI job to use a local `sway-lib-std` dependency

### Testing Steps
- [ ] CI should pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,6 +149,8 @@ jobs:
           args: --debug --path ./forc-fmt
       - name: Initialize test project
         run: forc init test-proj
+      - name: Update project manifest to use local sway-lib-std
+        run: echo "std = { path = \"../sway-lib-std/\" }" >> test-proj/Forc.toml
       - name: Build test project
         run: forc build --path test-proj
       - name: Run test project's test suite


.github/workflows/ci.yml
```diff
@@ -149,6 +149,15 @@ jobs:
...
         run: forc init test-proj
+      - name: Update project manifest to use local sway-lib-std
+        run: echo '[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "test-proj"
+
+[dependencies]
+std = { path = "../../sway-lib-std" }' > ./Forc.toml
```
adlerjohn, line 160: Hmmm. Rather than nuking the entire manifest, I would prefer just appending the local `std` dep: https://github.com/FuelLabs/sway/issues/1458#issuecomment-1116162972

mohammadfawaz: I agree.. that minimizes the need to change this `echo` when `forc init` changes in the future.

adlerjohn (approved):

adlerjohn:
@ra0x3 can you file an issue to add `[workspace]` to the test project's Cargo manifest as per https://github.com/FuelLabs/sway/runs/6275224635?check_suite_focus=true#step:10:18? (Do it in a separate PR since it's not blocking.)

mohammadfawaz (approved):

PR #1467, opened 2022-05-04
Update parse bytecode example

### LINKED ISSUE #1406: `forc` README needs updating
README for `forc` doesn't have the new formatting for `forc parse-bytecode` which pretty-prints the hex of the opcodes now

binggh:
Closes #1406 

- Updated `forc parse-bytecode` documentation example by using the `counter` template, with the pretty printed output


docs/src/forc/commands/forc_parse-bytecode.md
```diff
@@ -19,23 +19,80 @@ Print help information
...
+counter$ forc parse-bytecode obj
+
+  half-word   byte   op                   raw           notes
+          0   0      JI(4)                90 00 00 04   conditionally jumps to byte 16
+          1   4      NOOP                 47 00 00 00
+          2   8      Undefined            00 00 00 00   data section offset lo (0)
+          3   12     Undefined            00 00 00 c8   data section offset hi (200)
+          4   16     LW(63, 12, 1)        5d fc c0 01
+          5   20     ADD(63, 63, 12)      10 ff f3 00
+          6   24     LW(17, 6, 73)        5d 44 60 49
```
eureka-cpu, line 40: Since these will be examples in the book, you can shorten this by adding `// --snip--` here

scripts/forc-documenter/src/constants.rs
```diff
@@ -56,24 +56,81 @@ You can find an example within our [fuels-rs book](https://fuellabs.github.io/fu
...
+counter$ forc parse-bytecode obj
+
+  half-word   byte   op                   raw           notes
+          0   0      JI(4)                90 00 00 04   conditionally jumps to byte 16
+          1   4      NOOP                 47 00 00 00
+          2   8      Undefined            00 00 00 00   data section offset lo (0)
+          3   12     Undefined            00 00 00 c8   data section offset hi (200)
+          4   16     LW(63, 12, 1)        5d fc c0 01
+          5   20     ADD(63, 63, 12)      10 ff f3 00
+          6   24     LW(17, 6, 73)        5d 44 60 49
```
eureka-cpu, line 77: and here as well

eureka-cpu (changes requested):
looking good, just a few small things to keep it short and sweet

docs/src/forc/commands/forc_parse-bytecode.md
binggh: Is this utility supported natively? Doesn't seem to work, unless we have to add [this](https://lib.rs/crates/mdbook-snips)?

eureka-cpu: Ah, you know what, this is my mistake. I was thinking in Rust's documentation that translates to md, and this is just pure md. Disregard my previous comments unless you've found some other reasonable solution, I think this is fine!

eureka-cpu: For reference: https://doc.rust-lang.org/rustdoc/what-to-include.html#what-to-exclude

eureka-cpu: If using the `snip` works with that crate, that would be much cleaner

binggh: I realized that crate doesn't work either because it only works on `rust` code blocks. Perhaps I can simply manually truncate the example output in the middle and show something like this?

```
counter$ forc parse-bytecode obj

  half-word   byte   op                   raw           notes
          0   0      JI(4)                90 00 00 04   conditionally jumps to byte 16
          1   4      NOOP                 47 00 00 00
          2   8      Undefined            00 00 00 00   data section offset lo (0)
          3   12     Undefined            00 00 00 c8   data section offset hi (200)
          4   16     LW(63, 12, 1)        5d fc c0 01
          5   20     ADD(63, 63, 12)      10 ff f3 00
          ...
          ...
         61   244    Undefined            fa f9 0d d3
         62   248    Undefined            00 00 00 00
         63   252    Undefined            00 00 00 c8

```

adlerjohn: If you can't figure out how to snip after a bit of exploring, you can always defer it to a future PR. Just file an issue.

binggh: Yep - I think I will just simply manually truncate it for now to shorten the example output

adlerjohn (approved):

eureka-cpu (approved):

PR #1469, opened 2022-05-04
Hide use orig asm and parser in the help text for forc build, deploy, run

### LINKED ISSUE #1462: `forc build --help` should not display `--use-orig-asm` and `--use-orig-parser`
Until we completely remove the original parser and codegen, these options should be hidden from the end user. That is, the flags should still work but they should not be displayed in the help menu.

binggh:
Closes #1462 

- Add `hide = true` to `use_orig_parser` and `use_orig_asm` to hide them from the help text found in `build`, `deploy` and `run` while maintaining functionality
- Remove corresponding documentation about the options

### DIFF
--- a/forc/src/cli/commands/build.rs
+++ b/forc/src/cli/commands/build.rs
@@ -14,10 +14,10 @@ pub struct Command {
     #[clap(short, long)]
     pub path: Option<String>,
     /// Whether to compile using the original (pre- IR) pipeline.
-    #[clap(long)]
+    #[clap(long, hide = true)]
     pub use_orig_asm: bool,
     /// Whether to compile using the original (pest based) parser.
-    #[clap(long)]
+    #[clap(long, hide = true)]
     pub use_orig_parser: bool,
     /// Whether to compile to bytecode (false) or to print out the generated ASM (true).
     #[clap(long)]
--- a/docs/src/forc/commands/forc_build.md
+++ b/docs/src/forc/commands/forc_build.md
@@ -81,18 +81,6 @@ Whether to compile to bytecode (false) or to print out the generated IR (true)
 
 Silent mode. Don't output any warnings or errors to the command line
 
-
-`--use-orig-asm` 
-
-
-Whether to compile using the original (pre- IR) pipeline
-
-
-`--use-orig-parser` 
-
-
-Whether to compile using the original (pest based) parser
-
 ## EXAMPLES:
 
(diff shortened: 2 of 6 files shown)


forc/src/cli/commands/build.rs
```diff
@@ -14,10 +14,10 @@ pub struct Command {
     #[clap(short, long)]
     pub path: Option<String>,
     /// Whether to compile using the original (pre- IR) pipeline.
-    #[clap(long)]
+    #[clap(long, hide = true)]
     pub use_orig_asm: bool,
     /// Whether to compile using the original (pest based) parser.
-    #[clap(long)]
+    #[clap(long, hide = true)]
```
otrho, line 20: I think you need to do exactly the same thing in `run.rs` and `deploy.rs` too.  The issue only mentioned `build` but it makes sense to hide them everywhere.

otrho (changes requested):

forc/src/cli/commands/build.rs
binggh: makes sense, thanks for the catch!

adlerjohn: Wait, does `forc test` not have options to turn off these flags?? 😂

mohammadfawaz: `forc test` only builds the Rust test harness and not the Sway code, right? It can't tell the compiler what to do.

adlerjohn: https://github.com/FuelLabs/sway/blob/07d36690e6478894fc271b8a49cb9a771ddb552a/forc/src/cli/commands/test.rs#L44-L45

adlerjohn: Maybe leave for a subsequent PR @binggh

mohammadfawaz: TIL. I always thought `forc test` was just an alias for `cargo test` (and I've always used `cargo test`) but I guess it's doing more things now.

binggh: I believe the default values used in Default::default() are `false`, so the flags won't be set by default i think: https://doc.rust-lang.org/src/core/default.rs.html#187

adlerjohn (approved):

otrho (approved):

sezna (approved):

PR #1480, opened 2022-05-04
Format long use stmnts + tests

ra0x3:
### Description
- This PR formats long `use` statements by adding newline `\n` chars when the line extends beyond a certain length

### Testing steps
- [ ] Tests were added so `cargo test -p sway-fmt` should pass
- [ ] Create a simple `forc` project and try using this one-line, long `use` statement
```text
std::{address::*, assert::assert, block::*, chain::auth::*, context::{* , call_frames::*}, contract_id::ContractId, hash::* , panic::panic, storage::* , token::*};
```

### Here are some examples
```rust
// This...
use std::{address::*, assert::assert, block::*, chain::auth::*, context::{* , call_frames::*}, contract_id::ContractId, hash::* , panic::panic, storage::* , token::*};

// Formats to...
use std::{
    address::*,
    assert::assert,
    block::*,
    chain::auth::*,
    context::{*, call_frames::*},
    contract_id::ContractId,
    hash::*,
    panic::panic,
    storage::*,
    token::*,
};

// And this...
use std::{address::*, context::{*,text::{call_frames::*, dial_frames::{Transaction, TransactionParameters}, token_storage::{CallData, Parameters}}}, assert::assert, block::*, chain::auth::*, context::{*, call_frames::*}, contract_id::ContractId, hash::*, panic::panic, storage::*, token::*};

// Formats to...
use std::{
    address::*,
    assert::assert,
    block::*,
    chain::auth::*,
    context::{*, call_frames::*},
    context::{
        *,
        text::{
            call_frames::*,
            dial_frames::{Transaction, TransactionParameters},
            token_storage::{CallData, Parameters}
        }
    },
    contract_id::ContractId,
    hash::*,
    panic::panic,
    storage::*,
    token::*,
};
```


sway-fmt/src/traversal_helper.rs
```diff
@@ -85,35 +87,36 @@ pub fn format_delineated_path(line: &str) -> String {
...
-                // is needed to ignore trailing commas with newlines.
-                let to_push: String = line[current..index]
-                    .to_string()
-                    .chars()
-                    .filter(|c| !c.is_whitespace())
-                    .collect();
-                if !to_push.is_empty() {
-                    buffer.push(to_push);
-                }
+/// Tokenizes the line on separators keeping the separators.
```
ra0x3, line 90: Broke this out since it's being use in multiple places now

sway-fmt/src/traversal_helper.rs
```diff
@@ -278,4 +334,31 @@ mod tests {
...
+        let expected_result = "std::{address::*, assert::assert, block::*, chain::auth::*, context::{*, call_frames::*}, contract_id::ContractId, \n    hash::*, panic::panic, storage::*, token::*};";
+        let lines = format_use_stmnt_length(long_use_stmt);
+
+        assert_eq!(lines.len(), 2);
+        assert_eq!(lines.concat(), expected_result);
+
+        // We can't guarantee _every time_ that line length will be <= MAX_LINE_LENGTH
+        // but we can guarantee that line.len() will be within a certain tolerance
+        for line in lines {
+            let tolerance: usize = (0.3 * MAX_LINE_LENGTH as f32) as usize;
```
ra0x3, line 360: TODO: This can be calculated

mohammadfawaz:
I see that the approach you've taken here is to split the `use` statement into multiple lines with a max text width of 100 in mind. I worry the new output is not much more readable than before. Here's an example of what `cargo fmt` does:
https://github.com/FuelLabs/sway/blob/38e68b8644b9abd8aa5ac062e85795c1d674d9b4/sway-core/src/convert_parse_tree.rs#L1

For something like:
```
std::{address::*, assert::assert, block::*, chain::auth::*, context::{* , call_frames::*}, contract_id::ContractId, hash::* , panic::panic, storage::* , token::*};
```
I expect to see the following:
```
use std::{
    address::*, 
    assert::assert, 
    block::*, 
    chain::auth::*, 
    context::{* , call_frames::*}, 
    contract_id::ContractId, 
    hash::*, 
    panic::panic, 
    storage::* , 
    token::*};
```

ra0x3:
@mohammadfawaz
- I agree that is _much_ better looking (and preferred IMO). 
- Didn't wanna go _too_ far before getting some eyes on the initial work.
- More than happy to use the approach you just pasted

adlerjohn:
Yeah the approach should be if a line exceeds some length (for now it can be a constant, just don't hard-code it), each element is placed on a new line, recursively.

JoshuaBatty:
I think the approach that @mohammadfawaz suggested is the way we want to go here. 

Would you mind converting this PR to a draft until the output is close to what Mohammad suggested? Thanks!

forc-fmt/src/main.rs
```diff
@@ -28,6 +28,9 @@ pub struct App {
     /// Path to the project, if not specified, current working directory will be used.
     #[clap(short, long)]
     pub path: Option<String>,
+    /// Maximum length of line
+    #[clap(short = 'l', long = "length")]
+    pub line_length: Option<u32>,
```
adlerjohn, line 33: Rather than adding an option that will be removed later, I would hard-code the length to a constant `80` or `100` for now. Then later we can make it a configurable parameter in `swayfmt.toml` #605.

eureka-cpu: Checkout my recent commit! #1425 nearly have the `swayfmt.toml` ready I believe. If possible, anything relating to the current formatter we may want to hold off on. The foundation is almost ready, and we can start building there.

sway-fmt/src/traversal_helper.rs
```diff
@@ -278,4 +402,36 @@ mod tests {
...
+---    address::*,
+---    assert::assert,
+---    block::*,
+---    chain::auth::*,
+---    context::{
+---        *,
+---        text::{
+---            call_frames::*,
+---            dial_frames::{Transaction,TransactionParameters},
+---                token_storage::{CallData,Parameters,
```
adlerjohn, line 425: 1. Why is there an extra indentation here?
2. Can you use the `ALREADY_FORMATTED_LINE_PATTERN` constant here? I know it'll be kind of cumbersome, but the alternative is hard-coding the literal 😂

sway-fmt/src/code_builder_helpers.rs
```diff
@@ -192,9 +192,16 @@ pub fn get_new_line_pattern(line: &str) -> Option<&str> {
...
-        let end = line.find(';').unwrap();
+    if line.starts_with(ALREADY_FORMATTED_LINE_PATTERN) {
+        let char_idxs = vec![
+            line.find(';').unwrap_or(0),
+            line.rfind(',').unwrap_or(0),
+            line.rfind('}').unwrap_or(0),
+            line.rfind('{').unwrap_or(0),
+        ];
+
+        let end = char_idxs.iter().max().unwrap_or(&0);
```
ra0x3, line 203: - This will panic if (for whatever reason) that `0` default is used. 
  - Not sure how we would prefer to handle this
- While testing I did not see a panic when using properly formatted input (only if someone accidentally has a typo or something)

adlerjohn: If it will panic on `0`, why not just `unwrap()`?

sway-fmt/src/traversal_helper.rs
```diff
@@ -278,4 +407,38 @@ mod tests {
...
+{ALREADY_FORMATTED_LINE_PATTERN}    assert::assert,
+{ALREADY_FORMATTED_LINE_PATTERN}    block::*,
+{ALREADY_FORMATTED_LINE_PATTERN}    chain::auth::*,
+{ALREADY_FORMATTED_LINE_PATTERN}    context::{{
+{ALREADY_FORMATTED_LINE_PATTERN}        *,
+{ALREADY_FORMATTED_LINE_PATTERN}        text::{{
+{ALREADY_FORMATTED_LINE_PATTERN}            call_frames::*,
+{ALREADY_FORMATTED_LINE_PATTERN}            dial_frames::{{Transaction, TransactionParameters}},
+{ALREADY_FORMATTED_LINE_PATTERN}            token_storage::{{CallData, Parameters,
+{ALREADY_FORMATTED_LINE_PATTERN}        }}}}
```
adlerjohn, line 432: Hmmm. Should this not be
```suggestion
{ALREADY_FORMATTED_LINE_PATTERN}            token_storage::{{CallData, Parameters}},
{ALREADY_FORMATTED_LINE_PATTERN}        }}
```

ra0x3: @adlerjohn You are correct (oversight on my part). Just pushed an update

JoshuaBatty (approved):
Nice LGTM

adlerjohn (approved):
utACK

Issue #1107, opened 2022-03-30
`forc` has substantial duplicate code


eureka-cpu:
As mentioned by @mitchmindtree in #863. Please feel free to add instances here so we can track progress:

- [x] create abstraction for checking Sway `tree_types` in `forc_abi_json`, `forc_deploy` & `forc_run` @eureka-cpu #1105 
- [x] update the `Manifest::from_dir` function to automatically check parent directories in the case that the immediate directory doesn't contain a manifest. This might save us from always having to call `find_manifest_dir` everywhere before we can call `Manifest::from_dir`. @eureka-cpu #1232 @mitchmindtree #1249

eureka-cpu:
@mitchmindtree @JoshuaBatty if you find any other instances of dup code feel free to add them here and I'll tackle them regularly until we manage it to the point of closing the issue

mitchmindtree:
Closing now that these have been solved. Future areas of code cleanup/refactoring can get highlighted in dedicated issues.

PR #1487, opened 2022-05-05
Manifest files use env variable to set author or default back to the …

### LINKED ISSUE #1484: The manifest files should set the `author` key from an environment variable for dev purposes otherwise default back to the `realname()`
When creating a new project with `forc init <name>` the created manifest files set the author to be the `realname()` i.e. the user.
This is desirable behaviour for external use however for development purposes we would like to have the author set to "Fuel Labs \<contact@fuel.sh>".

There should be an environment variable that is checked and used if it is set otherwise default back to the `realname()` for the author.

Braqzen:
…realname

Closes #1484


forc/src/utils/defaults.rs
```diff
@@ -1,11 +1,14 @@
 /// We intentionally don't construct this using [serde]'s default deserialization so we get
 /// the chance to insert some helpful comments and nicer formatting.
 pub(crate) fn default_manifest(project_name: &str, entry_type: &str) -> String {
-    let real_name = whoami::realname();
+    let author = match std::env::var("FORC_INIT_MANIFEST_AUTHOR") {
```
adlerjohn, line 4: Abstract this string into a constant since it's used more than once and is part of the public API.

adlerjohn: Fixed in [afe7cd6](https://github.com/FuelLabs/sway/pull/1487/commits/afe7cd62ff64983632cf2c9bf4b17be7074c47ad)

adlerjohn (dismissed):
Looks fine to me, but will defer to tooling team.

eureka-cpu (commented):
If possible, since we're using the same code twice here, could you make a function possibly called `get_author` and call it for those variables? Just trying to tackle code bloat when I see it so we don't have to go back and fix it later, if this is a blocker it could also just be added to #1107 for a separate PR

forc/src/utils/defaults.rs
```diff
@@ -1,11 +1,14 @@
 /// We intentionally don't construct this using [serde]'s default deserialization so we get
 /// the chance to insert some helpful comments and nicer formatting.
 pub(crate) fn default_manifest(project_name: &str, entry_type: &str) -> String {
-    let real_name = whoami::realname();
+    let author = match std::env::var(sway_utils::FORC_INIT_MANIFEST_AUTHOR) {
+        Ok(author) => author,
+        _ => whoami::realname(),
+    };
```
otrho, line 7: FYI a simpler way to do this pattern is:
```rust
let author =
    std::env::var(sway_utils::FORC_INIT_MANIFEST_AUTHOR).unwrap_or_else(|| whoami::realname());
```

otrho (dismissed):

JoshuaBatty (changes requested):
Thanks @Braqzen! Can you please make the suggested edits that were made by @otrho and @eureka-cpu first before landing this?

eureka-cpu (approved):
looking clean, thanks!

JoshuaBatty (approved):

JoshuaBatty:
Looks good, thanks @Braqzen

PR #1494, opened 2022-05-06
Changes `panic` to `revert` in `sway-lib-std`

### LINKED ISSUE #1470: Rename `panic` to `revert`
While `panic` is closer to Rust, it doesn't capture the semantics of the FuelVM as closely as `revert` would. A panic in the VM is when an instruction can't be executed, while a revert is when the `RVRT` instruction is executed.

gr00vytvniks:
Closes #1470 

I changed every instance of `panic` to `revert` and some of the comments. I also ran the test suite on a node and made corrections to the test where applicable.


sway-lib-std/src/revert.sw
```diff
@@ -1,9 +1,9 @@
-library panic;
+library revert;
 
 /// Context-dependent:
-/// will panic if used in a predicate
+/// will revert if used in a predicate
```
adlerjohn, line 4: This is a case where you want "panic", as it refers to a VM panic.
```suggestion
/// will panic if used in a predicate
```

sway-lib-std/src/token.sw
```diff
@@ -54,7 +54,7 @@ pub fn transfer_to_output(amount: u64, asset_id: ContractId, recipient: Address)
     let mut output_found = false;
 
     // If an output of type `OutputVariable` is found, check if its `amount` is
-    // zero. As one cannot transfer zero coins to an output without a panic, a
+    // zero. As one cannot transfer zero coins to an output without a revert, a
```
adlerjohn, line 57: Ditto
```suggestion
    // zero. As one cannot transfer zero coins to an output without a panic, a
```

nfurfaro (commented):
Looks good once existing comments are addressed.

adlerjohn (commented):
CI failing.

eureka-cpu:
Looks like the `use` statements in the examples needs to be alphabetized to satisfy the formatter check.

sezna:
Nice contribution! Thank you!

The CI (continuous integration -- our testing infrastructure) is currently failing for these changes. Here is what you need to do to fix them:
1. [Failure 1](https://github.com/FuelLabs/sway/runs/6336224456?check_suite_focus=true#step:9:112). Update the import in the test.
2. [Failure 2](https://github.com/FuelLabs/sway/runs/6336223120?check_suite_focus=true#step:7:189). The examples with an `x` here need to have the command `forc fmt` run on them, to format the code.

sezna (approved):
Looks good to me, will wait for a once-over from @adlerjohn before merging 😄 thanks for the contribution!

adlerjohn (approved):

PR #1497, opened 2022-05-07
Remove `Cargo.lock` from top-level `.gitignore`. Track `Cargo.lock` for std lib test harness.

mitchmindtree:
I was just investigating https://github.com/FuelLabs/sway/pull/1450 when I noticed `Cargo.lock` was declared in the gitignore.

Now I'm wondering how our repo has been tracking the workspace-level `Cargo.lock` the whole time while this was the case? I guess it might have been added to tracking before being added to the gitignore?

This also adds the sway std lib test harness lock file which was ignored due to this, unblocking #1450.


nfurfaro (approved):
LG. No idea how this got added to the .gitignore!

adlerjohn (changes requested):
Can we ignore the lockfile for the examples directory? And any other subdirectory of Sway tests.

adlerjohn:
Wait, why do we want the test harness lockfile? Shouldn't we want that to use the latest version of things, as that's more realistic?

mitchmindtree:
> Can we ignore the lockfile for the examples directory? And any other subdirectory of Sway tests.

What `examples` lock file? Do you possibly have one left over in your local repo that's no longer used or tracked?

> Wait, why do we want the test harness lockfile? Shouldn't we want that to use the latest version of things, as that's more realistic?

I don't think we lose that by adding the lockfile for reproducibilty. We should indeed always make sure to update the _minor_ releases in the `Cargo.toml` when the fuel/fuels crates are updated upstream, but we'll need to do that either way.

adlerjohn:
There are no examples lockfiles because the lockfiles are currently ignored 😂. If we stop ignoring the lockfiles, then whenever anyone tests examples locally they'll get a bunch of files that can be accidentally committed.

It's not about reproducibility, but rather realistic testing conditions. Users actually writing tests won't have the versions of their dependencies pinned when they run `forc test` for the first time. If we use a lockfile for tests then we'll never be able to catch issues before the user.

adlerjohn:
See: https://github.com/FuelLabs/fuels-rs/pull/177

mitchmindtree:
> There are no examples lockfiles because the lockfiles are currently ignored joy. If we stop ignoring the lockfiles, then whenever anyone tests examples locally they'll get a bunch of files that can be accidentally committed.

I tested this before opening the PR. There are no example lock files, because there are no tests in the examples yet :smiling_face_with_tear:  Now that you mention it though, I'll pre-emptively add `examples/*/Cargo.lock` as hopefully those examples will get some tests as of #1321.

> It's not about reproducibility, but rather realistic testing conditions. Users actually writing tests won't have the versions of their dependencies pinned when they run forc test for the first time. If we use a lockfile for tests then we'll never be able to catch issues before the user.

Fair call. To clarify, I think for testing generally the big benefit of reproducibility is knowing *why* things start breaking, and not having to hunt down unrelated CI breakage on master. That said, I understand preferring to always test against the latest available patch release as there's a better chance of hitting issues before users, and that it's better we find out about issues earlier via seemingly random CI breakage rather than user reports. I don't really have a better solution, so will remove.

adlerjohn (approved):
lgtm, but would like @JoshuaBatty and @sezna to confirm

sezna (approved):
the discussion makes sense to me

ra0x3 (approved):
Thanks @mitchmindtree

JoshuaBatty (approved):
LGTM. I agree we should make sure all of the sway examples have tests, especially now that they can be used as templates.

PR #1500, opened 2022-05-08
Add `require` function.

nfurfaro:
A simple wrapper around `assert` which allows passing in a custom `u64` value to the inner `revert`, + a few simple tests.


sway-lib-std/src/assert.sw
```diff
@@ -10,3 +10,12 @@ pub fn assert(a: bool) {
         ()
     }
 }
+
+/// wrapper for `assert` that allows passing a custom revert value `v` if condition `c` is not true.
+pub fn require(c: bool, v: u64) {
```
adlerjohn, line 15: Make this generic, and use `LOGD` followed by a revert. You can use `is_reference_type` and `size_of` to know how many bytes to log. For simplicity, just use `LOGD` all the time even if it's a word.

otrho: What would get logged exactly?  This function looks to me like an `assert()` but instead of `revert(0)` it's `revert(x)` which makes tracking down which assert failed a bit easier.

You're suggesting a third argument to add some verbosity?  Or to remove `v` and log something else and still `revert(0)`?

adlerjohn: Similar to Solidity custom errors (https://blog.soliditylang.org/2021/04/21/custom-errors/), `require`'s second parameter should be a struct which can eventually be part of the JSON ABI. In the meantime, at least it can be logged even though the logged value can't be decoded by the SDK.

nfurfaro: hmm. If we make `require` generic, I think we need to also be able to add the constraint that the generic value either is a `u64` or is a struct/enum with a field/variant with a `u64` type which can be passed to the inner call to `revert`.

nfurfaro: Perhaps something along these lines:
`require`'s second param is a struct type (ie: `CustomError` provided by the `assert` module in the stdlib. Users can instantiate as many CustomErrors as they need in their contract, and there will always be a  `u64` value available to pass to `revert`:
```
pub struct CustomError<T> {
    Id: ContractId, // use to log the current contract id 
    RevertValue: u64, // value to pass to `revert`
    AdditionalData: T // user can log out generic data to keep this flexible
}
    
    
```

adlerjohn: There's no need for a `u64` value, just set that to a hard-coded standard value for indicating a custom error. See Solidity panic codes as an example: https://docs.soliditylang.org/en/v0.8.4/control-structures.html#panic-via-assert-and-error-via-require

`LOGD` already logs the contract ID in the receipt: https://github.com/FuelLabs/fuel-specs/blob/84cff9a590a28d7c5e6cb62d91b52e8e30c6ef67/specs/vm/opcodes.md#logd-log-data-event

adlerjohn: This can be as simple as `fn require<T>(c: bool, v: T)`, then just `LOGD` `v`, followed by a `RVRT` with some error code that's documented to mean "look at the previous logd receipt".

sway-lib-std/src/assert.sw
```diff
@@ -10,3 +11,17 @@ pub fn assert(a: bool) {
         ()
     }
 }
+
+/// A wrapper for `assert` that allows logging a custom value `v` if condition `c` is not true.
+/// This will then revert with the value `42`, which indicates that you should look at the previous logd receipt for further debugging clues. Note that the SDK will not currently decode this logd receipt for you.
```
nfurfaro, line 16: I'm not attached to the value `42` here as a signal value, but we do need to use something.

sway-lib-std/src/assert.sw
```diff
@@ -10,3 +11,17 @@ pub fn assert(a: bool) {
...
+pub fn require<T>(c: bool, v: T) {
+    if !c {
+        let size = size_of::<T>();
```
nfurfaro, line 19: As I've simply used `logd` for all types, I don't think there's a need to call `is_reference_type()` if we just need the size to pass to the `logd` op.

nfurfaro: Note: we can replace the logging here with the generic `log` func in the stdlib (coming soon).

otrho: `LOGD` expects `r1` below to be a pointer so if it isn't a reference type it'll try to read memory from somewhere weird and not make sense.  The VM spec doesn't say what happens if you try to log invalid memory but it might panic?

So I think `LOG` for non-ref and `LOGD` for ref really should be done here, unless `log()` is very imminent and will Do The Right Thing?

sway-lib-std/src/assert.sw
otrho: Is it worth making some special constants, only used internally within the stdlib but still semi-documented so when we see 42 we know where to look?

otrho (changes requested):

sway-lib-std/src/assert.sw
nfurfaro: Makes sense. I'll fix this to do the right thing, and swap it out to use the generic `log()` when it lands.
Otherwise this is one more PR on hold. This way at least `require` will be usable.

nfurfaro:
>Is it worth making some special constants, only used internally within the stdlib but still semi-documented so when we see 42 we know where to look?

I think so, but not sure what approach to take for this. For now I've made the value `42` into a const:
`const FAILED_REQUIRE_SIGNAL = 42;`
with some inline documentation. This can be extracted into a central location when/if we create one.
cc @otrho

otrho (approved):
Yep, maybe as we make more constants we can put them somewhere central, this is a good start I reckon.

adlerjohn (commented):
Looks fine.

PR #1503, opened 2022-05-09
Publish Diagnostics from language server even when no errors or warnings detected

### LINKED ISSUE #1315: LSP Diagnostics are

JoshuaBatty:
Incorrect warnings and errors were being reported in VS-Code after updating the document. It appears that we have only been publishing diagnostics back to the client if any errors or warnings were reported by the parser. This meant that if an error was rectified, then we were not clearing the previously reported diagnostic information and the Client would display the incorrect warnings and errors. 

Also fixes an off-by-one error when reporting diagnostic ranges to the Client. 

closes #1315

### DIFF
--- a/sway-lsp/src/server.rs
+++ b/sway-lsp/src/server.rs
@@ -481,16 +481,13 @@ fn main() {
 
     #[tokio::test]
     async fn did_change() {
-        let (mut service, mut messages) = LspService::new(|client| Backend::new(client, config()));
+        let (mut service, _) = LspService::new(|client| Backend::new(client, config()));
 
         // send "initialize" request
         let _ = initialize_request(&mut service).await;
 
         // send "initialized" notification
         initialized_notification(&mut service).await;
-
-        // ignore the "window/logMessage" notification: "Initializing the Sway Language Server"
-        messages.next().await.unwrap();
 
         let uri = Url::parse("inmemory:///test").unwrap();
         let text = r#"script;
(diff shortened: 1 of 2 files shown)


sway-lsp/src/server.rs
```diff
@@ -131,20 +138,22 @@ impl LanguageServer for Backend {
...
         let diagnostics = capabilities::text_sync::handle_change_file(self.session.clone(), params);
-        self.publish_diagnostics(uri, diagnostics).await;
+        self.publish_diagnostics(uri, diagnostics, version).await;
     }
 
     async fn did_save(&self, params: DidSaveTextDocumentParams) {
         let uri = params.text_document.uri.clone();
         let diagnostics = capabilities::text_sync::handle_save_file(self.session.clone(), &params);
-        self.publish_diagnostics(uri, diagnostics).await;
+        self.publish_diagnostics(uri, diagnostics, None).await;
```
mitchmindtree, line 156: Is there a reason we don't know the `version` here? Is there any chance `DidSaveTextDocumentParams` should be updated to provide the version?

mitchmindtree (commented):
Nice find!

sway-lsp/src/server.rs
JoshuaBatty: There isn't a `version` within the `DidSaveTextDocumentParams` struct. I actually think it might be better to just stick with no version support like it was previously. Makes it simpler and less of our own state to manage and sync with the client. Will update that section now. We can address this in a separate PR if necessary.

mitchmindtree (dismissed):
utACK!

adlerjohn (commented):
looks fine to me, but will defer to @mitchmindtree

mitchmindtree (approved):

PR #1504, opened 2022-05-09
Extract forc documentation examples from constants.rs into examples.rs

binggh:
Some minor code quality related `forc-documenter` refactoring - moving examples found in documentation out of `constants.rs` into `examples.rs` for clearer semantic separation


JoshuaBatty (approved):
Nice, makes sense to me!

PR #1507, opened 2022-05-09
Updated example to show how to call a private function

### LINKED ISSUE #1506: Update the `Calling Contracts` section of the Sway Book to show how to define and call non-ABI functions within the `impl` of the contract
### Issue

My expectation, based on Solidity, was that I can write a function inside the `impl <name of abi> for Contract` and then call that function within the same contract.

E.g.
```
contract;

abi Issue {
    fn method_a() -> bool;
    fn method_b() -> bool;
}

impl Issue for Contract {
    fn method_a() -> bool {
        method_b()
    }

    fn method_b() -> bool {
        true
    }
}
```

This does not work because the ABI is for `external` calls only.
Removing `method_b()` from the ABI and leaving the rest of the code unaltered creates another issue which is that `method_b()` is not defined in the ABI (and it must be).

The solution is

```
contract;

abi Issue {
    fn method_a() -> bool;
}

impl Issue for Contract {
    fn method_a() -> bool {
        method_b()
    }
}

fn method_b() -> bool {
    true
}
```

however that has not been commented in the `Calling Contracts` section which is the first place that I thought to look.
This may be a non-issue to those who are proficient in Rust syntax and thus may be self-evident however that is not the case for all developers, especially Solidity specialists.

### Solution

Update the Sway Book to include an example with a footnote explaining how additional functions can be used to keep the code DRY without defining them in the ABI.

Braqzen:
Closes #1506

### DIFF
--- a/docs/src/blockchain-development/calling_contracts.md
+++ b/docs/src/blockchain-development/calling_contracts.md
@@ -49,7 +49,7 @@ impl ContractB for Contract {
 }
 ```
 
-> **NOTE** The ABI is for `external` calls only therefore you cannot define a method in the ABI and call it in the same contract. If you want to define a function for a contract, but keep it private so that only your contract can call it, you can define it outside of the `impl` and call it inside the contract similar to the `return_45()` function.
+> **NOTE** The ABI is for external calls only therefore you cannot define a method in the ABI and call it in the same contract. If you want to define a function for a contract, but keep it private so that only your contract can call it, you can define it outside of the `impl` and call it inside the contract, similar to the `return_45()` function above.
 
 ## Advanced Calls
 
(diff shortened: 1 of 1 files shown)


docs/src/blockchain-development/calling_contracts.md
```diff
@@ -45,6 +49,8 @@ impl ContractB for Contract {
 }
 ```
 
+> **NOTE** The ABI is for `external` calls only therefore you cannot define a method in the ABI and call it in the same contract. If you want to define a function for a contract, but keep it private so that only your contract can call it, you can define it outside of the `impl` and call it inside the contract similar to the `return_45()` function.
```
adlerjohn, line 52:
```suggestion
> **NOTE** The ABI is for external calls only therefore you cannot define a method in the ABI and call it in the same contract. If you want to define a function for a contract, but keep it private so that only your contract can call it, you can define it outside of the `impl` and call it inside the contract, similar to the `return_45()` function above.
```

adlerjohn (approved):

Issue #610, opened 2022-01-10
Add associated items to `impl`s and `trait`s.


otrho:
We should be able to add types and consts to `trait` and `impl` declarations, as per:

https://doc.rust-lang.org/reference/items/associated-items.html

Useful for:
```rust
impl u64 {
    pub const MIN: u64 = 0;
    pub const MAX: u64 = 18446744073709551615;
    pub const BITS: u32 = 64;
}
```
and obviously:
```rust
pub trait Iterator {
    type Item;
    fn next(&mut self) -> Option<Self::Item>;
}
```

PR #1515, opened 2022-05-10
Add numeric constants.

adlerjohn:
Add numeric associated constants (like https://doc.rust-lang.org/std/primitive.u64.html#associatedconstant.MAX).

To avoid being blocked by #610, uses function.


Braqzen:
I'm wondering if it makes more sense to have this code live in [lib-core/ops.sw](https://github.com/FuelLabs/sway/blob/master/sway-lib-core/src/ops.sw). Thoughts?

adlerjohn:
Co you're actually correct, not sure why I put it in `std`.

sway-lib-std/src/num.sw
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /// The largest value that can be represented by this integer type,
+    /// 2<sup>64</sup> - 1.
+    pub fn MAX() -> u64 {
+        18446744073709551615
+    }
+
+    /// The size of this integer type in bits.
+    pub fn BITS() -> u32 {
```
nfurfaro, line 16: Any reason the return type here is different than MIN and MAX? Ditto for each of the other types' BITS methods.

adlerjohn: That's just how Rust does it for whatever reason: https://doc.rust-lang.org/std/primitive.u64.html#associatedconstant.BITS

nfurfaro: I see . That's weird.
You call a method on a u64 and get a u32 back, seems like needless mental overhead. Must be some reason though...

adlerjohn:
Ready for review @Braqzen @otrho

sway-lib-std/src/num.sw
adlerjohn: 🤷. I vote for keeping it in line with Rust unless we have a reason not to. I'm sure they had their reasons.

nfurfaro (approved):
LGTM!

otrho:
In C++ this file is called `limits`.  `num` is fine though. 🙂

Issue #1145, opened 2022-04-05
Annotations


adlerjohn:
- [ ] Add annotations
- [ ] Add annotation for function statefulness
- [ ] Add annotation for ABI method payability

mitchmindtree:
@adlerjohn perhaps we can close this in favour of a dedicated ABI method payability issue. Would you mind writing one up with your thoughts on how this should behave? Or link to somewhere it's already explained?

adlerjohn:
I would prefer to keep this as a tracking issue and make sub-issues.

PR #1518, opened 2022-05-11
Replace the `impure` keyword with `#[storage(...)]` annotations.

### LINKED ISSUE #470: AST item annotation system (pure/read/write/chain for storage access)
We should support an annotation system for AST items. This is blocked by #252, as the annotation system will inherently be coupled with the parser.

Generically, we should be able to tag items with `#[annotation]`. This annotation system will then be used to implement storage usage annotations regarding purity and impurity, as the comments describe.

old issue text:
---

Do we want to keep `impure`? Do we want to do something else? See #466 for more discussion and context.

---

Options so far:
1. `fn impure` instead of `impure fn`
2. `ifn`
3. `imp fn` or `fn imp`
4. `fn foo() -> impure u64`
5. `fn_impure` or `fn_with_effects`

Additionally, the default right now is pure for everything, impure if specified. We should also bikeshed if we want to keep that or switch to impure by default.

cc @nfurfaro @adlerjohn @SilentCicero 

---

### LINKED ISSUE #1447: Remove `impure` keyword
In anticipation of annotations provided a better solution (#1145), remove the `impure` keyword to avoid people using it then having to change a bunch of things.

otrho:
Closes #1447.
Closes #470. Except for the chain bit. Do we want a new issue for chain attributes?


sway-core/src/parse_tree/declaration/function.rs
```diff
@@ -50,7 +50,7 @@ impl FunctionDeclaration {
             .unwrap_or(false)
         {
             let _ = signature.next();
-            Purity::Impure
+            Purity::Reads // TODO: add attributes to parser instead of `impure` keywords.  Or not.
```
adlerjohn, line 53: Wait why is this TODO here?

sway-core/src/error.rs
```diff
@@ -876,8 +877,12 @@ pub enum CompileError {
         missing_patterns: String,
         span: Span,
     },
-    #[error("Impure function called inside of pure function. Pure functions can only call other pure functions. Try making the surrounding function impure by prepending \"impure\" to the function declaration.")]
-    PureCalledImpure { span: Span },
+    #[error(
+        "Impure function called inside of pure function. Pure functions can only call \
+        other pure functions. Try making the surrounding function impure by adding \
+        \"#[{STORAGE_PURITY_ATTRIBUTE_NAME}({attrs})]\" to the function declaration."
```
adlerjohn, line 883: Ideally this error message would more precisely reflect the use annotation usage rather than referring to "impure" and "pure."

adlerjohn (commented):
Can you also add documentation to The Sway Book?

sway-core/src/parse_tree/declaration/function.rs
otrho: This is about the old Pest parser, which Alex already has a branch started for removing.  So this is an almost definitely WONTDO.

adlerjohn: Ahhh got it

otrho:
Have updated the docs though I'm not sure the terms I've used are OK.  Like using 'privilege'.  And maybe the page needs more structure, it only has a single section.

docs/src/blockchain-development/purity.md
```diff
@@ -2,6 +2,23 @@
...
+#[storage(read)]
+fn get_amount() -> u64 {
+    ...
+}
+
+#[storage(read, write)]
+fn increment_amount(increment: u64) -> u64 {
+    ...
+}
+
```
adlerjohn, line 17:
```suggestion
}
```

docs/src/blockchain-development/purity.md
```diff
@@ -2,6 +2,23 @@
 
 A function is _pure_ if it does not access any [persistent storage](./storage.md). Conversely, the function is _impure_ if it does access any storage. Naturally, as storage is only available in smart contracts, impure functions cannot be used in predicates, scripts, or libraries. A pure function cannot call an impure function.
 
-In Sway, functions are pure by default but can be opted in to impurity via function annotations which are currently work-in-progress.
+In Sway, functions are pure by default but can be opted into impurity via the `storage` function attribute.  The `storage` attribute may take `read` and/or `write` arguments indicating which type of access the function requires.
+
+```sway
+#[storage(read)]
+fn get_amount() -> u64 {
+    ...
```
adlerjohn, line 10: While we're here, could you file an issue to make a `[...]` annotation? Would be very useful for docs!

docs/src/blockchain-development/purity.md
```diff
@@ -2,6 +2,23 @@
...
+#[storage(read, write)]
+fn increment_amount(increment: u64) -> u64 {
+    ...
+}
+
+```
+
+Impure functions which call other impure functions must have at least the same storage privileges or a superset of those for the function called.  For example, to call a function with write access a caller must also have write access, or both read and write access.  To call a function with read and write access the caller must also have both privileges.
+
+The `storage` attribute may also be applied to [methods and associated functions](../basics/methods_and_associated_functions.md) but are not required nor accepted for [trait](../advanced/traits.md) or [ABI declarations](../sway-program-types/smart_contracts.md#the-abi-declaration).
```
adlerjohn, line 22: This will change in the future, right?

otrho: Um, no?  AFAICT we decided that attributes should only be applied to implementations and not declarations.

docs/src/blockchain-development/purity.md
otrho: I'm not sure what you mean, sorry. :slightly_smiling_face: An annotation for `mdbook` to markup `...`?

otrho:
Closes #1447.
Closes #470.  Except for the `chain` bit.  Do we want a new issue for `chain` attributes?

docs/src/blockchain-development/purity.md
adlerjohn: I should have said attribute, and should have written `#[...]`. Am I imagining things or does Rust not have something like that, which allows something that looks like ellipses in documentation and such to actually be valid Rust code?

docs/src/blockchain-development/purity.md
adlerjohn: Looping in @sezna. That was the _initial_ decision, but it was revised (and clearly not documented 😂) to add attributes to the ABI declarations at the very least to enable method payability: https://github.com/FuelLabs/fuel-specs/issues/271. Ref: #https://github.com/FuelLabs/sway/issues/1145

> Add annotation for ABI method payability

docs/src/blockchain-development/purity.md
otrho: I have no idea, not something I've seen.  Again, maybe @sezna knows..?

docs/src/blockchain-development/purity.md
mitchmindtree: I'd imagine we'd want to at least support it for trait/ABI method declarations that provide default implementations, in case those default implementations need access to storage?

Otherwise my impression is that the storage access attribute is more about whether or not *implementations* of functions, trait methods or ABI methods interact with storage. It's not quite clear to me what it would mean to use the attribute with a trait or ABI method _declaration_ (that has no default implementation)?

@adlerjohn in the case of method payability, would it be enough to instead require that implementations of ABI methods that access received coins require the payable attribute, rather than requiring that the ABI declarations themselves do?

adlerjohn: > would it be enough to instead require that implementations of ABI methods that access received coins require the payable attribute, rather than requiring that the ABI declarations do?

No, because the _caller_ doesn't know the implementation; they only know the ABI declaration.

mitchmindtree: > No, because the caller doesn't know the implementation; they only know the ABI declaration.

For the caller to be able to call the ABI method, there must be an implementation of the ABI for the callee somewhere, no? My understanding is that if there's an ABI declaration, for it to actually be used it must be implemented for `Contract` at some point - is that right?

If so, and if `Contract`'s ABI method implementation requires payable, then in order to call it the user would have to add the payable attribute to whatever function it is they are calling the payable method from. This would be the case whether or not the declaration has a payable attribute, right?

adlerjohn: > For the caller to be able to call the ABI method, there must be an implementation of the ABI for the callee somewhere, no? My understanding is that if there's an ABI declaration, for it to actually be used it must be implemented for Contract at some point - is that right?

That's correct, but just because you're correct doesn't mean you're right. The model for cross-contract calls is through an ABI, where the implementation of that ABI is unknown at compile time. This is the same model that Ethereum uses and Move (Libra's thing) does not use. It allows your contract to interact with contracts that are deployed later, where obviously it's impossible to know their implementation at compile time since they don't exist yet.

Put another way: cross-contract calls can _only_ use the ABI, not the implementation of the called contract.

mitchmindtree: Ahhhhh, cross-contract calls was the context I was missing!

Do we have any sketches of what cross-contract calls might look like in Sway code yet?

adlerjohn: Yes. https://fuellabs.github.io/sway/v0.12.2/blockchain-development/calling_contracts.html

otrho: Can I get some clarification on what still needs to be done here please?  I can add annotations to ABI declarations and update the docs now if that's what we want.

adlerjohn: Annotations (generally) allowed for ABI and trait method declarations. Storage annotations specifically too. And need to enforce that storage annotations for implementations match declarations.

canndrew:
The parsing side of this looks okay to me. We should really have a `Spanned` trait that we can use rather than having a macro for adding a `span` method to `Annotated<$ty_name>` but that's something that can be done in a follow-up PR.

mohammadfawaz (commented):
LGTM. I'll let @sezna make the final approval.

mitchmindtree (approved):
>  Ugh, that's the last time I try to resolve a merge conflict in the GitHub web editor. :(

:joy:

PR #1522, opened 2022-05-11
When resolving custom types, recurse into Arrays to be sure the element type is resolved.

otrho:


sway-core/src/type_engine/engine.rs
```diff
@@ -227,24 +227,12 @@ impl Engine {
...
             (Array(a_elem, a_count), Array(b_elem, b_count)) if a_count == b_count => {
-                let mut warnings = vec![];
+                let (warnings, new_errors) = self.unify(a_elem, b_elem, span, help_text.clone());
+
+                // If there was an error then we want to report the array types as mismatching, not
+                // the elem types.
                 let mut errors = vec![];
-                if a_count == b_count {
-                    let (new_warnings, new_errors) =
-                        self.unify(a_elem, b_elem, span, help_text.clone());
```
otrho, line 234: This isn't a part of the fix, but I noticed that it's checking if `a_count == b_count` but this arm won't even match due to line 229 so this is all redundant.

mohammadfawaz (approved):
Presumably we should be able to have any combination of arrays and structs now (e.g. arrays of structs that contain other arrays etc.)

PR #1524, opened 2022-05-11
Fix formatting aliases

ra0x3:
### Description
- This PR fixes an issue where `forc-fmt` would not respect aliases and would turn `use foo::{bar as Bar};` into `use foo::{barasBar};`
  - The fix is done by swapping `.match_indices()` with a similar function `match_indices_str()` which can match the full `as` token

### Testing steps
- [ ] A test was added so `cargo test -p sway-fmt` should pass
- [ ] Create a new `forc init` project and try out some of the following use statements:

```rust
// This...
use a::b::{c as d, e as f};

// Formats to...
use a::b::{c as d, e as f};

// And this...
use std::{address::*, assert::assert as LocalAssert, block::*, chain::auth::*, context::{*,text::{call_frames::*, dial_frames::{Transaction as DialFrameTransaction, TransactionParameters}, token_storage::{CallData, Parameters}}}, contract_id::ContractId, hash::*, panic::panic, storage::*, token::*};

// Formats to...
use std::{
    address::*,
    assert::assert as LocalAssert,
    block::*,
    chain::auth::*,
    context::{
        *,
        text::{
            call_frames::*,
            dial_frames::{Transaction as DialFrameTransaction, TransactionParameters},
            token_storage::{CallData, Parameters}
        }
    },
    contract_id::ContractId,
    hash::*,
    panic::panic,
    storage::*,
    token::*,
};
```

### DIFF
--- a/sway-fmt/src/traversal_helper.rs
+++ b/sway-fmt/src/traversal_helper.rs
@@ -96,7 +96,10 @@ fn match_indices_str(s: &str) -> Vec<(usize, &str)> {
 
     while start < s.len() {
         // Try to match the 'as' token first then fallback to single chars
-        if start <= s.len() - 4 && &s[start..start + as_token.len()] == as_token {
+        if start <= s.len() - as_token.len()
+            && s.len() >= as_token.len()
+            && &s[start..start + as_token.len()] == as_token
+        {
             res.push((start + 1, as_token.trim()));
             start += as_token.len();
             continue;
--- a/sway-fmt/src/traversal_helper.rs
+++ b/sway-fmt/src/traversal_helper.rs
@@ -85,11 +85,47 @@ pub fn format_delineated_path(line: &str) -> String {
     line.chars().filter(|c| !c.is_whitespace()).collect()
 }
 
+// Same as s.match_indices(|ch| {}) but allows to match by checking &str vs char
+fn match_indices_str(s: &str) -> Vec<(usize, &str)> {
+    let mut res: Vec<(usize, &str)> = Vec::new();
+
+    // Match the as token with spaces so as to avoid imprperly matching an 'as' substring
+    // in another type of token
+    let as_token = " as ";
+    let mut start = 0;
+
+    while start < s.len() {
+        // Try to match the 'as' token first then fallback to single chars
+        if start <= s.len() - 4 && &s[start..start + as_token.len()] == as_token {
+            res.push((start + 1, as_token.trim()));
+            start += as_token.len();
+            continue;
+        }
+
+        match &s[start..start + 1] {
+            "," => {
+                res.push((start, ","));
+            }
+            "{" => {
+                res.push((start, "{"));
+            }
+            "}" => {
+                res.push((start, "}"));
+            }
+            _ => {}
+        };
+
+        start += 1;
+    }
+
+    res
+}
+
 /// Tokenizes the line on separators keeping the separators.
 fn tokenize(line: &str) -> Vec<String> {
     let mut buffer: Vec<String> = Vec::new();
     let mut current = 0;
-    for (index, separator) in line.match_indices(|c: char| c == ',' || c == '{' || c == '}') {
+    for (index, separator) in match_indices_str(line) {
         if index != current {
             // Chomp all whitespace including newlines, and only push
             // resulting token if what's left is not an empty string. This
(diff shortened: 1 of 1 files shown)


adlerjohn (commented):
lgtm, but will defer to @JoshuaBatty and @mitchmindtree

sway-fmt/src/traversal_helper.rs
```diff
@@ -85,11 +85,47 @@ pub fn format_delineated_path(line: &str) -> String {
...
+    let mut res: Vec<(usize, &str)> = Vec::new();
+
+    // Match the as token with spaces so as to avoid imprperly matching an 'as' substring
+    // in another type of token
+    let as_token = " as ";
+    let mut start = 0;
+
+    while start < s.len() {
+        // Try to match the 'as' token first then fallback to single chars
+        if start <= s.len() - 4 && &s[start..start + as_token.len()] == as_token {
```
mitchmindtree, line 99: Can `s.len()` ever be less than `4`? If so, I think the `s.len() - 4` expression may underflow and `panic!` or wrap around as a result.

mitchmindtree: Also, where does `4` come from? Can we get this from a const or retrieving the length directly from what this refers to?

sway-fmt/src/traversal_helper.rs
```diff
@@ -85,11 +85,47 @@ pub fn format_delineated_path(line: &str) -> String {
...
+            "{" => {
+                res.push((start, "{"));
+            }
+            "}" => {
+                res.push((start, "}"));
+            }
+            _ => {}
+        };
+
+        start += 1;
```
mitchmindtree, line 118: Generally, it's a good idea to avoid iterating through `str`'s one `byte` at a time. This is because characters encoded in `UTF8` do not always have a consistent byte length.

This is a common source of confusion with Rust string handling, particularly w.r.t. use of `len`, indexing and slicing! The [`str::len`](https://doc.rust-lang.org/std/primitive.str.html#method.len) docs briefly demonstrate how the byte-length and character-length can differ.

One alternative approach for this loop might be to use:

```rust
for ch in s.chars() {
```

and then bump the `start` byte slice position with
```rust
    start += ch.len_utf8();
}
```

mitchmindtree (changes requested):

sway-fmt/src/traversal_helper.rs
ra0x3: I can make this change 👍

sway-fmt/src/traversal_helper.rs
ra0x3: @mitchmindtree 
- I see what you mean here, and this is a good thing to keep in mind
- However, for this particular use case, don't `use` statements guarantee that our `str` is limited to characters from the set `abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789;}{,` ? which means that 1 char is guaranteed to be one byte?

adlerjohn: Looks like identifiers can only be ASCII, so it might be fine to assume that. Morever, this will be replaced in the future with using the parse tree instead of manual string examination, so doesn't have to be completely perfect.

mitchmindtree: > Morever, this will be replaced in the future with using the parse tree instead of manual string examination, so doesn't have to be completely perfect.

Yeah agreed, happy to leave this as is pending the new formatter.

> Looks like identifiers can only be ASCII, so it might be fine to assume that

Just to address this for future cases: Even if identifiers *should* only be ASCII in sway, it doesn't mean users won't try inputting non-ascii text and end up with LSP or formatter `panic!`ing. Any time we use `String` or `str` (everywhere), we make the implicit assumption that we're handling UTF8. Unless we validate that we're loading pure-ASCII when loading Sway files in `sway-core` itself and use ASCII string types everywhere to assure this at the type-level, we shouldn't make assumptions like this.

mitchmindtree (approved):

JoshuaBatty (approved):

sway-fmt/src/traversal_helper.rs
ra0x3: @mitchmindtree Thanks for that context. Even though this has been approved, let me see if I can update it to match our convention :)

adlerjohn: @ra0x3 Maybe merge in as-is (since at least the current code _works_), then if there's anything to improve it can be done in a follow-up PR.

PR #1534, opened 2022-05-12
Implement `Display` for `TypeId`s;  Add type parameters to the definition of custom types

sezna:
I've been working on unblocking vectors and allowing unused generics. I'm blocked on some changes in #1433 . For the sake of atomic PR's, I figured I'd wrap up the changes I've made thus far into a PR:

1. Include `type_parameters` in the definition of a struct or enum in the type engine. This allows for handling of generic types where the type is not necessarily used in the fields or variants.
2. Change `TypeId` from a type alias to a newtype. This provides greater type safety but also allowed me to implement `Display` for it. Now when you print a `TypeId`, instead of the numeric value in the type engine, you'll see the actual type it represents.
3. Do a bit of refactoring/code cleanup in places.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -102,9 +102,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.57"
+version = "1.0.56"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "08f9b8508dccb7687a1d6c4ce66b2b0ecef467c94667de27d8d7fe1f8d2a9cdc"
+checksum = "4361135be9122e0870de935d7c439aef945b9f9ddd4199a553b5270b49c82a27"
 
 [[package]]
 name = "arrayref"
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -102,9 +102,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.56"
+version = "1.0.57"
```
adlerjohn, line 105: Why the change to lockfile?

adlerjohn (commented):
Also, CI is failing

sway-core/src/type_engine.rs
```diff
@@ -13,8 +16,33 @@ pub use type_info::*;
...
+pub struct TypeId(usize);
 
+impl std::ops::Deref for TypeId {
+    type Target = usize;
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Display for TypeId {
```
mohammadfawaz, line 29: image

Cargo.lock
mohammadfawaz: So that's been happening to me sometimes and not sure why (especially when I'm cutting a release). I usually just unstage `Cargo.lock`, restore it, and then rebuild.

mitchmindtree:
Might be particularly good for @emilyaherbert to review as there's potentially quite a bit of overlap in monomorphisation / declaration checking with #1433?

otrho (approved):
I like it.

mitchmindtree (approved):
> potentially quite a bit of overlap in monomorphisation / declaration checking with https://github.com/FuelLabs/sway/pull/1433?

I'm looking again and I think I must have been mistaken or looking at a different PR :woozy_face: The only real overlapping change is moving the struct declaration stuff, but that's easy enough to address. Lgtm too!

PR #1539, opened 2022-05-13
Remove `rand` dependency from default Cargo.toml

adlerjohn:
It's only used for a random salt, which is 1) hidden under the hood, and 2) not needed since you can just use zero.


vnepveu (approved):
Good atomic PR haha

mitchmindtree (approved):

PR #1540, opened 2022-05-13
Do not overwrite preceeding numeric type info on cast

### LINKED ISSUE #1512: abigen assigning u32 type to u64 function parameters
I have a function which takes two `u64` as parameters:  

```
fn multiply_u64(a: u64, b: u64) -> (u64, u64) {
        let result_u128: U128 = mul64(a, b);
        (result_u128.upper, result_u128.lower)
        }
```

but abigen is assigning `u32` type to both `a` and `b`. I have been unable to reproduce this with a minimal example, and there are plenty of functions with a similar signature in the std lib which don't have this problem. Confirmation that someone else can reproduce would be a good start. Code here:
https://github.com/FuelLabs/sway/tree/simon-u128/test/src/sdk-harness/test_projects/u128

Possibly a subject for a separate issue, but if I try to return a `U128` (which is a struct, containing two `u64` values) directly  instead of a tuple of its values, I get the following error from the abigen macro:

```
7 | abigen!(U128Contract, "test_projects/u128/out/debug/u128-abi.json");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `fuels_core`
  |
  = note: this error originates in the macro `abigen` (in Nightly builds, run with -Z macro-backtrace for more info)
help: there is a crate or module with a similar name
  |
7 | fuel_core;
```

sezna:
Closes #1512

The inference engine currently replaces old type info with new, more accurate type info when more constraining type information is found. For example, if we know a type is `Numeric`, and it is then used in the context of a `u64`, we can replace all the type info for that type with `u64`.

We are mistakenly performing this update on integer casts. If a number `x` is a `u32`, for example, and it is used in the context of a `u64`, we are updating all the type information up to that point with `u64`. This is incorrect, as casting only updates the type information from the cast site onward.

### DIFF
--- a/test/src/e2e_vm_tests/test_programs/should_pass/test_contracts/issue_1512_repro/Forc.lock
+++ b/test/src/e2e_vm_tests/test_programs/should_pass/test_contracts/issue_1512_repro/Forc.lock
@@ -4,9 +4,4 @@ dependencies = []
 
 [[package]]
 name = 'issue_1512_repro'
-dependencies = ['std git+https://github.com/fuellabs/sway?tag=v0.12.2#2b6e9384f06692ec627293ae5db5e2f748fe2c30']
-
-[[package]]
-name = 'std'
-source = 'git+https://github.com/fuellabs/sway?tag=v0.12.2#2b6e9384f06692ec627293ae5db5e2f748fe2c30'
 dependencies = ['core']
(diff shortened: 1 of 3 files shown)


mohammadfawaz (approved):
Subtle!

Very nice :)

PR #1541, opened 2022-05-13
Fix asm for `get_ptr` when the offset is > 4096

mohammadfawaz:
This fixes a functional issue due to a missing `add` in the asm generated for `get_ptr` when the offset is higher than 4096.


mohammadfawaz:
> Is there an issue/bug that exposed this? Impressive find

@nfurfaro was seeing an intermittent failure in his `b256_ops` test. Turns out there were two problems:
* Non-determinism in the bytecode caused by the non-deterministic tuple name generated using the `nanoid` crate.
* The actual functional issue that this PR fixes. The failure triggered only when the randomly generated tuple names started with the last few letters of the alphabet.

sway-core/src/asm_generation/from_ir.rs
```diff
@@ -1058,6 +1058,15 @@ impl<'ir> AsmBuilder<'ir> {
                             &instr_reg,
                             instr_val.get_span(self.context),
                         );
+                        self.bytecode.push(Op {
+                            opcode: either::Either::Left(VirtualOp::ADD(
+                                instr_reg.clone(),
+                                self.stack_base_reg.as_ref().unwrap().clone(),
+                                instr_reg.clone(),
+                            )),
+                            comment: "get_ptr".into(),
```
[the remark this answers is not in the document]
mohammadfawaz: Done! Let me know if the comment is better now.

test/src/e2e_vm_tests/test_programs/should_pass/language/is_prime/src/main.sw
```diff
@@ -35,8 +35,7 @@ fn main() -> bool {
     assert(check_prime(11) == check_prime(17));
     assert(check_prime(12) == false);
     assert(check_prime(18) == false);
-    // The assert below currently fails. I think this is a compiler bug
-    // assert(check_prime(12) == check_prime(18));
+    assert(check_prime(12) == check_prime(18));
```
mohammadfawaz, line 38: This is unrelated to this change. Just a random test fix.

sway-core/src/asm_generation/from_ir.rs
```diff
@@ -1058,6 +1058,15 @@ impl<'ir> AsmBuilder<'ir> {
                             &instr_reg,
                             instr_val.get_span(self.context),
                         );
+                        self.bytecode.push(Op {
+                            opcode: either::Either::Left(VirtualOp::ADD(
+                                instr_reg.clone(),
+                                self.stack_base_reg.as_ref().unwrap().clone(),
+                                instr_reg.clone(),
+                            )),
+                            comment: "get offset reg for get_ptr".into(),
```
otrho, line 1067: OK, so until now we'd never actually ever run this code and it turned out to be buggy.  An argument to introduce coverage tracking to our tests.

PR #1543, opened 2022-05-13
Docker Fixes

AlicanC:
I've been trying to run Forc with Docker. I believe something like `docker run ghcr.io/fuellabs/forc build` should work, but we have some issues:

[fix: add missing Docker container dependency](https://github.com/FuelLabs/sway/commit/f0d00f94a2c937fa5bead5f43a8dbad151d83061)

I was getting the following error so I added `libssl-dev` as a dependency:
```
> docker run ghcr.io/fuellabs/sway:latest ./forc build
./forc: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
```

[fix: publish Docker image as "forc" and not "sway"](https://github.com/FuelLabs/sway/commit/a8481b62c6a9423af91bac8b0914edcb3efd6d9b)

Our image is called `sway` which I believe should be called `forc`. This commit adds image labels for Forc during the build process.

[fix: add default Docker image command](https://github.com/FuelLabs/sway/commit/95f161f6e6cae3707e90d0ecc76e323f6c201654)

The image doesn't have a default command so you need to specify the `forc` binary yourself. This commit adds that.


AlicanC:
@Voxelot @rfuelsh this isn't a blocker for me and I have other priorities, so I haven't tested this at all!

It would be great if someone could do that.

rfuelsh (approved):

PR #1544, opened 2022-05-13
Added generic sha256

### LINKED ISSUE #1511: Generic stack type hashing
Related: #823

Outline:

Check if type is a reference type with `is_reference_type()`.

If not. then it's in the register:
1. Grow stack with [`CFEI`](https://github.com/FuelLabs/fuel-specs/blob/84cff9a590a28d7c5e6cb62d91b52e8e30c6ef67/specs/vm/opcodes.md#cfei-extend-call-frame-immediate) by one word.
2. Save value in register to memory with [`SW`](https://github.com/FuelLabs/fuel-specs/blob/84cff9a590a28d7c5e6cb62d91b52e8e30c6ef67/specs/vm/opcodes.md#sw-store-word).
3. Hash with [`S256`](https://github.com/FuelLabs/fuel-specs/blob/84cff9a590a28d7c5e6cb62d91b52e8e30c6ef67/specs/vm/opcodes.md#s256-sha-2-256).

If yes:
1. Get size with `size_of()`.
2. Hash with [`S256`](https://github.com/FuelLabs/fuel-specs/blob/84cff9a590a28d7c5e6cb62d91b52e8e30c6ef67/specs/vm/opcodes.md#s256-sha-2-256).

Braqzen:
Closes #1511


sway-lib-std/src/hash.sw
```diff
@@ -124,3 +124,31 @@ pub fn hash_pair(value_a: b256, value_b: b256, method: HashMethod) -> b256 {
...
     }
 }
+
+// Registers are 64 bits (8 bytes) wide. Words are the same width as registers.
+pub fn sha256<T>(param: T) -> b256 {
+    let result_buffer: b256 = 0x0000000000000000000000000000000000000000000000000000000000000000;
+    if !is_reference_type::<T>() {
+        asm(r1: param, r2, r3: 8, r4: result_buffer) {
+            // Move to register r2
+            move r2 sp;
```
nfurfaro, line 134: minor nit, but maybe format with comments after the op to align with pre-existing functions in the module. ie: `move r2 sp; // Move stack pointer to register r2`

sway-lib-std/src/hash.sw
```diff
@@ -124,3 +124,31 @@ pub fn hash_pair(value_a: b256, value_b: b256, method: HashMethod) -> b256 {
         }
     }
 }
+
+// Registers are 64 bits (8 bytes) wide. Words are the same width as registers.
```
adlerjohn, line 128:
```suggestion
/// Returns the SHA-2-256 hash of `param`.
```

sway-lib-std/src/hash.sw
```diff
@@ -124,3 +124,31 @@ pub fn hash_pair(value_a: b256, value_b: b256, method: HashMethod) -> b256 {
...
+        let size = size_of::<T>();
+        asm(r1: result_buffer, r2: param, r3: size) {
+            // Hash
+            s256 r1 r2 r3;
+            // Return
+            r1: b256
+        }
+    }
+}
+
```
adlerjohn, line 154: Format file, there's an extra newline here.

sway-lib-std/src/hash.sw
nfurfaro: Use triple backslash `///`for public comments as they will eventually be used for auto-generated docs.
Give some description of what this function is for.

nfurfaro:
The implementation looks good :)
I'd like to see some tests for hashing different types added to the stdlib's `test_projects/hashing` before merging though.

sway-lib-std/src/hash.sw
```diff
@@ -124,3 +124,23 @@ pub fn hash_pair(value_a: b256, value_b: b256, method: HashMethod) -> b256 {
         }
     }
 }
+
+/// Returns the SHA-2-256 hash of `param`.
+pub fn sha256<T>(param: T) -> b256 {
+    let result_buffer: b256 = 0x0000000000000000000000000000000000000000000000000000000000000000;
```
adlerjohn, line 130: Hmmmm. This might not be the most important, but would making `result_buffer` `mut` be more semantically correct? The compiler can't really analyze what's in asm blocks yet, but it would just be more info to both the reader and potentially later the compiler.

Also, can you use `ZERO` instead of the literal `0x0000000000000000000000000000000000000000000000000000000000000000`? Not sure if it actually works. https://github.com/FuelLabs/sway/blob/afcc5211d3e6bbcebecdb1b977ae4a44dc7e9c2b/sway-lib-std/src/constants.sw#L4

Braqzen: The `mut` does make more sense to me but then I doubt that `const` would work if I tried to mix them both so it's probably one or the other

Braqzen:
> The implementation looks good :) I'd like to see some tests for hashing different types added to the stdlib's `test_projects/hashing` before merging though.

Sure, I already have some assertions that can be compared to `hash_u64`. I don't know what a good test would be for the generic types of structs / enums. In my testbed I have the `digest` and output to compare against but I cannot actually tell if they are correct

sway-lib-std/src/hash.sw
nfurfaro: Agree we should be using `mut` here even if it doesn't seem to make a difference now.

adlerjohn:
The unfortunate only way to test hashing is to go on some online hashing site and put in a few values, then use those values as test vectors.

sway-lib-std/src/hash.sw
adlerjohn: Ah I meant

```rust
let mut result_buffer: b256 = ZERO;
```

which should work just as well as using the literal `0x0000000000000000000000000000000000000000000000000000000000000000`.

sway-lib-std/src/hash.sw
```diff
@@ -124,3 +124,23 @@ pub fn hash_pair(value_a: b256, value_b: b256, method: HashMethod) -> b256 {
...
+    if !is_reference_type::<T>() {
+        asm(r1: param, r2, r3: 8, r4: result_buffer) {
+            move r2 sp; // Move to register r2
+            cfei i8; // Grow stack by 1 word
+            sw r2 r1 i0; // Save value in register r1 to memory at r2
+            s256 r4 r2 r3; // Hash
+            r4: b256 // Return
+        }
```
otrho, line 138: More hot tips for the aspiring ASM block author:
* The argument names don't have to resemble register names, they can be any identifier.  So you could use `asm(ptr: param, buf, eight: 8, hash: result_buffer) { ... }` and then the ASM would be slightly more readable.
* Although it isn't 100% necessary, I would really like to see a corresponding `CSEI` instruction after the hash is done.
* `// Move to register r2` is a bit misleading, as is `MOV` in general, since it's really just copying `$sp` to `r2`.  Just to be clear.

That second point might be contentious.  Taking some stack and then just forgetting about it is strictly leaking memory, which :triggered: me.  But OTOH, adding an extra instruction uses gas and the Fuel VM has enormous memory reserves to perhaps it's better just to leak.  @adlerjohn what do you say?

Hrmm, and this is absolutely assuming that structs and other aggregates have deterministically ordered elements and are packed tightly with no garbage bytes between elements.  This might be something we need to codify in the specs if we haven't already.  Or we make `T` belong to a `Serialized` trait which makes those guarantees for us.

mohammadfawaz:
> The unfortunate only way to test hashing is to go on some online hashing site and put in a few values, then use those values as test vectors.

Can we not use a hashing library such as https://docs.rs/sha2/latest/sha2/ in the [hashing test](https://github.com/FuelLabs/sway/tree/master/test/src/sdk-harness/test_projects/hashing)

sway-lib-std/src/hash.sw
adlerjohn: Yeah there's a balance to be struck between leaking memory (which is free so long as you don't run out of memory!), and reclaiming, which is strictly not-free (as in free beer). I'm leaning towards the standard library should be leak-free since it's expected to be used very heavily and there's a real possibility of running out of memory.

adlerjohn: > Hrmm, and this is absolutely assuming that structs and other aggregates have deterministically ordered elements and are packed tightly with no garbage bytes between elements. This might be something we need to codify in the specs if we haven't already. Or we make T belong to a Serialized trait which makes those guarantees for us.

Definitely something to work on in a follow-up PR, or several!

adlerjohn:
> Can we not use a hashing library such as

Ah, I meant if you wanted to test using only a Sway script and no Rust harness.

otrho (commented):
The library code is looking OK.  Is it possible to add tests to check that the hashes are returning the right thing, using a separate or online tool?  It could be a bit laborious -- each of the non-reference values are big-endian 64-bit values which wouldn't be that hard to reproduce.  The reference types except enum are pretty straight forward too.  Enums are a bit trickier as you need to treat them as tuples with a u64 tag field before the value field.

Braqzen:
> The library code is looking OK. Is it possible to add tests to check that the hashes are returning the right thing, using a separate or online tool? It could be a bit laborious -- each of the non-reference values are big-endian 64-bit values which wouldn't be that hard to reproduce. The reference types except enum are pretty straight forward too. Enums are a bit trickier as you need to treat them as tuples with a u64 tag field before the value field.

I was thinking about using [this hashing crate](https://github.com/RustCrypto/hashes#usage) however I am unsure of what is expected. Personally I thought to write a script that generates the hashes and have an assertion in our test to make sure that they are equal but then the reader does not know how to produce those hashes and still whether they are valid. I could add a comment into the tests so that they can pull the crate and reproduce the work but that is prone to errors and not user friendly which leaves another rust file lingering around somewhere.
What are the recommendations here?
Do you want a separate script somewhere in the sway repo which produces these specific hashes?

otrho:
Generating the correct hashes in the harness is optimal, yeah.  I'm thinking you could recreate the memory layout in Rust byte arrays and hash them with the crate.  Each array would be pretty explicit and obvious as to what it's supposed to be representing.  E.g, Sway `true` is Rust `[0, 0, 0, 0, 0, 0, 0, 1]` as a `[u8; 8]`.  Gets fiddly when converting strings and other weird numerics, but not impossible.

examples/hashing/src/main.sw
```diff
@@ -1,11 +1,79 @@
...
+}
+
+struct Person {
+    name: str[4],
+    age: u64,
+    alive: bool,
+    location: Location,
+    stats: Stats,
+    some_tuple: (bool,
+    u64), some_array: [u64;
```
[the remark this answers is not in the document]
Braqzen: I have mentioned this before and I believe someone has mentioned that this will get fixed with the new parser. I believe it was @eureka-cpu

test/src/sdk-harness/test_projects/hashing/src/main.sw
```diff
@@ -1,13 +1,101 @@
...
+    fn sha256_u8(value: u8) -> b256;
+    fn sha256_u16(value: u16) -> b256;
+    fn sha256_u32(value: u32) -> b256;
+    fn sha256_u64(value: u64) -> b256;
+    fn sha256_str(value: str[4]) -> b256;
+    fn sha256_bool(value: bool) -> b256;
+    fn sha256_b256(value: b256) -> b256;
+    fn sha256_tuple(value: (bool, u64)) -> b256;
+    fn sha256_array(value1: u64, value2: u64) -> b256;
+    fn sha256_struct(location: bool) -> b256;
```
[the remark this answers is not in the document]
mohammadfawaz: Sounds good! No reason to block this PR for that. We can always revisit once that issue is resolved.

mohammadfawaz:
LGTM! I'd like another set of eyes to review this and I'll approve after.

sway-lib-std/src/hash.sw
```diff
@@ -1,5 +1,7 @@
 library hash;
 
+use ::constants::ZERO;
+
 // Should this be a trait eventually? Do we want to allow people to customize what `!` does?
 // Scala says yes, Rust says perhaps...
 pub fn not(a: bool) -> bool {
```
adlerjohn, line 7: Aside to this PR...but what is `not` doing here 😂. Can you file an issue to investigate?

sway-lib-std/src/hash.sw
```diff
@@ -124,3 +126,24 @@ pub fn hash_pair(value_a: b256, value_b: b256, method: HashMethod) -> b256 {
         }
     }
 }
+
+/// Returns the SHA-2-256 hash of `param`.
+pub fn sha256<T>(param: T) -> b256 {
+    let mut result_buffer: b256 = ZERO;
+    if !__is_reference_type::<T>() {
+        asm(buffer, ptr: param, eight_bytes: 8, hash: result_buffer) {
+            move buffer sp; // Copy stack pointer to memory at "buffer"
```
adlerjohn, line 135:
```suggestion
            move buffer sp; // Make `buffer` point to the current top of the stack
```

adlerjohn (commented):
nit

sway-lib-std/src/hash.sw
mohammadfawaz: So close...
image

adlerjohn (approved):

sway-lib-std/src/hash.sw
Braqzen: Issue: #1593

mohammadfawaz (approved):

PR #1545, opened 2022-05-14
Use forc-documenter as an mdbook preprocessor instead

### LINKED ISSUE #1510: Investigate using custom mdbook preprocessor to generate forc command documentation
This section in the mdbook docs about writing custom preprocessors looks really interesting:

https://rust-lang.github.io/mdBook/for_developers/preprocessors.html

I wonder if we could take advantage of this to build the forc command and plugin documentation at mdbook build time rather than using `forc-documenter` to regenerate and replace all the text separately?

If we can work out how to include the command content at `mdbook build` time, we could also remove the whole forc-documenter CI check and save on some diff noise related to keeping the forc command docs up to date.

cc @binggh it could be worth investigating this before diving deeper into more `forc-documenter` issues in case we can come up with a solution that removes the need for `forc-documenter` altogether? Edit: Or alternatively `forc-documenter` could be refactored toward _being_ a mdbook preprocessor plugin?

binggh:
Closes #1510 

Looked into [using custom preprocessors for mdbook](https://rust-lang.github.io/mdBook/for_developers/preprocessors.html) instead of a manual `forc-documenter` script, managed to port it over as a preprocessor.

Points to take note:
- With this change, we no longer need to run `forc-documenter` manually to update the `Forc` related documentation - this step is automatically ran within the `mdbook build` step instead. This will save us time when trying to iterate quickly over PRs
- The only time  that someone has to manually do something is if a forc command is being deleted or added - the change has to be made within `SUMMARY.md`
- The internals still work more or less the same, but with cleaner code since there's no longer a mess of creating/deleting files. Like before, the preprocessor is running `forc <SUBCOMMAND> --help`, parsing and formatting the output and writing it into a [chapter's content](https://docs.rs/mdbook/latest/mdbook/book/struct.Chapter.html#structfield.content).
- Examples now no longer need to be stored in ugly raw strings as we did using `forc-documenter`, instead we store them within a folder `examples/` in Markdown, which makes editing a lot easier.

There is a small downside to this approach, namely that we can't see the diffs of the documentation between changes when we make PRs. This means that if we made changes to forc commands it would be ideal to run `mdbook build` locally just to double check if the documentation was generated correctly - however, I think this should be a non-issue since the `forc <SUBCOMMAND> --help` output behaves in a predictable manner.


adlerjohn:
> The only time that someone has to manually do something is if a forc command is being deleted or added - the change has to be made within SUMMARY.md

At the cost of dumpstering on your hard work 😂...this seems like a regression, no? Could at the very least this part be checked by CI?

binggh:
> At the risk of dumpstering on your hard work 😂...this seems like a regression, no? Could at the very least this part be checked by CI?

No worries 😂. 

I did think about that quite a bit, you're right in that it seems regressive because there's some manual action to be taken now if we want to add/remove commands, but after some thought, changing each command's details (options, comments, help texts, etc.) would occur a lot more frequently versus removing/adding new commands at the top-level of `forc` (correct me if I'm wrong). 

In that case, with the previous script the workflow of running `forc-documenter` each time _might_ slow down devs. With the usage of `forc-documenter` we also had to make sure we had the right forc installed (local? latest released version?) to generate the docs, which can be confusing.

You're right though, that we can perhaps check if we left out any commands in `SUMMARY.md` and remind a dev to include the entry if they added/removed a command but did not remove it from the book.

binggh:
@adlerjohn actually, i didn't know why i didn't implement it but i could just manually write over the SUMMARY.md like with forc-documenter

binggh:
Added a way to make mdbook build fail with an error if there are inconsistencies between the book and forc toolchain's commands.

**Scenario 1** - Command removed from Forc toolchain, but exists within the book:

This was tested by removing `forc addr2line` from `forc`, rebuilding it locally and running `mdbook build` with the preprocessor.

```
2022-05-15 14:12:53 [INFO] (mdbook::book): Book building has started

Some commands were removed from the Forc toolchain, but still exist in SUMMARY.md:

forc addr2line

To fix this, remove the above command(s) from SUMMARY.md.

2022-05-15 14:12:53 [ERROR] (mdbook::utils): Error: The "forc-documenter" preprocessor exited unsuccessfully with exit status: 1 status
```

**Scenario 2** - Command added to Forc toolchain, but not documented in the book:

This was tested by removing a command entry from `SUMMARY.md` and running `mdbook build`. The preprocessor picks up the command via `forc --help` but does not find the corresponding command documentation page within the book.

```
2022-05-15 14:21:18 [INFO] (mdbook::book): Book building has started

Some commands were missing from SUMMARY.md:

forc addr2line

To fix this, add the above command(s) in SUMMARY.md.

2022-05-15 14:21:19 [ERROR] (mdbook::utils): Error: The "forc-documenter" preprocessor exited unsuccessfully with exit status: 1 status
```

binggh:
Might be a dumb question but: how do I remove the `forc-documenter-dry-run` job from the workflow? I removed it from `ci.yml`, not sure if I need to remove it elsewhere.

adlerjohn:
Those yaml files are indeed the only place where CI workflows are defined. So if it's removed from there then it's removed. You can confirm that it no longer exists in the list of checks!

adlerjohn:
> also think there might be room for improvement here by perhaps automating the editing of SUMMARY.md within GitHub actions (not sure if this is possible?). I'll investigate if this is possible and maybe add it in a new PR, though I am not sure if we want to defer every part to the CI

Editing files in CI is a nono because it's unexpected. There's no way for users to really test the behavior of CI locally. Note that this PR doesn't change CI, but rather changes the build process (which can be reproduced locally!), which CI then uses.

Overall I think the state of things after this PR is merged is pretty good w.r.t. forc documenter.

adlerjohn:
I've removed `forc-documenter-dry-run` as a required status check.

scripts/mdbook-forc-documenter/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "mdbook-forc-documenter"
+version = "0.1.0"
+edition = "2021"
```
adlerjohn, line 4:
```suggestion
version = "0.0.0"
edition = "2021"
publish = false
```

scripts/mdbook-forc-documenter/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+
+[dependencies]
+anyhow = "1"
+clap = { version = "3.1", features = [ "derive" ]  }
+mdbook = "=0.4.15"
+semver = "1.0"
+serde = "1.0"
+serde_json = "1.0"
+
```
adlerjohn, line 22:
```suggestion
serde_json = "1.0"
```

scripts/mdbook-forc-documenter/examples/forc_completions.md
```diff
@@ -100,7 +20,7 @@ Run the command:
 
 ```sh
 mkdir -p ~/.local/share/bash-completion/completions
-forc completions --shell=bash >> ~/.local/share/bash-completion/completions/forc
+forc competions --shell=bash >> ~/.local/share/bash-completion/completions/forc
```
adlerjohn, line 23: Hmmm
```suggestion
forc completions --shell=bash >> ~/.local/share/bash-completion/completions/forc
```

binggh: Oopsie, probably accidentally removed it while copying the whole chunk over

adlerjohn (commented):
lgtm, good work. Will defer to @JoshuaBatty and @mitchmindtree for final review.

scripts/mdbook-forc-documenter/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+[lib]
+name = "mdbook_forc_documenter"
+path = "src/lib.rs"
+
+
+[dependencies]
+anyhow = "1"
+clap = { version = "3.1", features = [ "derive" ]  }
+mdbook = "=0.4.15"
```
mitchmindtree, line 19: Do we need to pin to `=0.4.15` in particular? Or could we specify `0.4.15` to allow for using that version and above?

mitchmindtree: Also are we able to disable default features here? E.g.

```toml
mdbook = { version = "0.4.15", default-features = false }
```

My guess is we don't need [these features](https://github.com/rust-lang/mdBook/blob/bf258eeb9b245b9c329e147dd1884307c5ef7644/Cargo.toml#L60) to build our forc documenter preprocessor, as the actual `mdbook` tool will have those features anyway?

Hopefully disabling default features might help to cut down on compilation time a bit by removing the need for those extra crates. That said, they'll probably still show up in our `Cargo.lock` either way.

scripts/mdbook-forc-documenter/README.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+## Usage
+
+### Adding a new forc command
+
+Enter a new entry under the `Commands` section within `SUMMARY.md`, in this format:
+
+```md
+- [forc gm](./forc_gm.md)\n
```
mitchmindtree, line 18:
```suggestion
- [forc gm](./forc_gm.md)
```
Does it need the `\n` on the end like this?

mitchmindtree (commented):
Very cool! This mdbook preprocessor business seems like a seriously handy trick to have up our sleeve :)

Just a couple small questions but otherwise looking good!

scripts/mdbook-forc-documenter/README.md
binggh: Nope! Thanks for the catch

scripts/mdbook-forc-documenter/Cargo.toml
binggh: > Do we need to pin to `=0.4.15` in particular? Or could we specify `0.4.15` to allow for using that version and above?

Hmm nope, realized that was a mistake on my part because I had the wrong version installed and it was throwing a warning, just had to update it

> My guess is we don't need [these features](https://github.com/rust-lang/mdBook/blob/bf258eeb9b245b9c329e147dd1884307c5ef7644/Cargo.toml#L60) to build our forc documenter preprocessor, as the actual mdbook tool will have those features anyway?

Thanks for this good point! Yup, seems like we only require the `build` feature, so we can go ahead and remove the features.

mitchmindtree: Great news, looks like that did slim down `Cargo.lock` quite a bit too!

mitchmindtree (approved):
Great stuff!

It's a pity we can't generate all the command files too, but those errors look really nice and clear so I don't see that being a big deal.

Very happy to approve :)

Issue #905, opened 2022-03-08
Directly writing a tuple element of a tuple struct member doesn't work


mohammadfawaz:
I expect the following to compile but it doesn't (parsing error):
```rust
script;

struct S1 {
    t: (u64, u64), 
}

fn foo() {
    let s1 = S1 {
        t: (0, 1)
    };
    let t1 = s1.t.0; // Parsing error
}
```

Instead, I have to do this:
```rust
struct S1 {
    t: (u64, u64), 
}

fn foo() {
    let s1 = S1 {
        t: (0, 1)
    };
    let t = s1.t;  // This is okay
    let t1 = t.0; 
}
```

Edit. Reading works fine now but writing still fails. E.g.
```rust
let mut s1 = S1 {
    t: (0, 1)
};
s1.t.0 = 5;
```
Error: 
```bash
10 |     };
11 |     s1.t.0  = 5;
   |     ^^^^^^ this expression cannot be assigned to
12 | }
```

mohammadfawaz:
Actually, `let t1 = (s1.t).0` works so I guess this is not a big deal. Though I think Rust allows `s1.t.0`.

otrho:
There is still a problem with arrays too, needing e.g., `(rec.ary)[0]` when `rec.ary[0]` should work.  Most parsers treat array, tuple and struct field accesses as peers which can be chained with left associativity.

@canndrew Is this the case with the new parser?  Will this issue be fixed already?

mohammadfawaz:
Reading is annoying but not a huge deal. However, writing is quite problematic. I'd like to be able to do:
```rust
    let mut s1 = S1 {
        t: (0, 1)
    };
    s1.t.0 = 5;
```
but I can't. Even `(s1.t).0 = 5` doesn't work. So, the only way we have right now is to update the whole tuple which, of course, is bad.

mohammadfawaz:
We don't have this problem with nested structs. See test in https://github.com/FuelLabs/sway/pull/911/files

mohammadfawaz:
Structs seem to be handled separately from everything else. Basically the only thing that works without parenthesis is chaining struct accesses. I think we need a general solution here.

mohammadfawaz:
This seems to work now thanks to the new parser 😃

mohammadfawaz:
Edit: it seems that reading is still fine now but writing continues to fails, i.e. something like:
```
let mut s1 = S1 {
    t: (0, 1)
};
[...]

PR #1558, opened 2022-05-16
Parse tuple field reassignments

canndrew:
This is a partial fix for #905. Tuple field reassignments are now accepted by the parser and the parse tree converter, however they are still rejected by the semantic analysis phase of the compiler.


canndrew:
> but why is it called a `Projection`?

That's the terminology I'm familiar with for the operation of extracting a field from a tuple. eg. see here: https://ncatlab.org/nlab/show/dependent+sum+type#as_a_negative_type

PR #1559, opened 2022-05-16
Add a Spanned trait to the parser

canndrew:
This is useful for abstracting over types that have a `Span`, for instance in #1518.


otrho:
I've moved that trait to `sway-types/src/span.rs` where I think it makes more sense.  This way it could be used anywhere a `Span` is.  Sound OK @canndrew , @mohammadfawaz , @sezna ?

mohammadfawaz (approved):

Issue #1368, opened 2022-04-22
Handling of complex types at the interfaces between internal and external contexts.


mohammadfawaz:
(The below is as of `forc v0.12.1`)
## Overview
I'm creating this issue to keep track of the current state of the interfaces between internal and external contexts. Specifically, this includes two types of interfaces:
* Contract method arguments
* Contract method return data

## Contract Method Arguments
In an internal context (the contract itself), the method arguments are loaded from the call frame. Specifically `param2` [here](https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/main.md#call-frames) which may contain the actual data passed into the contract or a pointer to the data. To load `param2`, we use `lw $rA $fp i74`. 

The agreed on implementation of the above is as follows: if the data passed into the contract method is an integer or a boolean, then the data itself should be stored in `param2`. Otherwise, `param2` should simply contain a pointer to the data in memory. Below is a summary of the possible cases in Swayland, the desired behavior, and the current behavior. In the table, "value" means `param2` contains the actual data while "pointer" means `param2` contains a pointer to the data.

#### Case 1: Single Argument
| Type        | Desired Behavior | Compiler Behavior | Scripts Behavior | Rust SDK Behavior |
| ----------- | ----------- | ----------- | ----------- | ----------- |
| `u8`, `u16`, `u32`, `u64`, `bool`  | value | value | | value |
| `str[1]`, ..., `str[8]`  | pointer | pointer | | value (to fix: https://github.com/FuelLabs/fuels-rs/issues/239)|
| `struct` of size <= 8 bytes  | pointer| pointer | | pointer |
| `tuple` of size <= 8 bytes  | pointer | pointer | | pointer |
| `array` of size <= 8 bytes | pointer |  pointer | | value (to fix: https://github.com/FuelLabs/fuels-rs/issues/239) |
| `enum` of size <= 8 bytes | pointer | pointer | | pointer |
| `b256` | pointer | pointer | | pointer |
| `str[9]`, ... | pointer | pointer | | value (to fix: https://github.com/FuelLabs/fuels-rs/issues/239) |
| `struct` of size > 8 bytes  |pointer | pointer | | pointer |
| `tuple` of size > 8 bytes  | pointer | pointer | | pointer |
| `array` of size > 8 bytes| pointer | pointer | | value (to fix: https://github.com/FuelLabs/fuels-rs/issues/239) |
| `enum` of size > 8 bytes | pointer | pointer | | pointer |

[...]

PR #1566, opened 2022-05-16
Fixing the handling of constant strings by properly accounting for their size

mohammadfawaz:
As discussed https://github.com/FuelLabs/sway/issues/1368, strings are reference types regardless of their size. Treat constant strings as such. This fixes a situation where returning a struct that contains a string fails.


sway-core/tests/ir_to_asm/ret_string_in_struct.asm
```diff
@@ -0,0 +1,37 @@
...
+lw   $r0 data_4               ; load fn selector for comparison
+eq   $r0 $r1 $r0              ; function selector comparison
+jnzi $r0 i14                  ; jump to selected function
+lw   $r0 data_5               ; load fn selector for comparison
+eq   $r0 $r1 $r0              ; function selector comparison
+jnzi $r0 i22                  ; jump to selected function
+rvrt $zero                    ; revert if no selectors matched
+move $r2 $sp                  ; save register for temporary stack value
+cfei i8                       ; allocate 8 bytes for temporary struct
+lw   $r1 data_0               ; literal instantiation
```
otrho, line 18: I'm surpised this works.  I would've thought it's loading `"foob"` into `$r1`.

otrho (approved):

sway-core/tests/ir_to_asm/ret_string_in_struct.asm
mohammadfawaz: This is similar to what we do with `b256` I believe. You basically just get back the pointer.

Issue #1343, opened 2022-04-21
Very occasional CI breakage when landing older PRs. Make PR merge conditions stricter?


mitchmindtree:
Occasionally issues like #1342 slip through in the case that newer PRs have landed since CI ran on the most recent commit before landing an existing older PR.

We are implicitly assuming that if an older PR passed CI in its last commit, that it should be compatible with all changes between following its last commit and the moment it is merged with master, which is of course not always the case.

If we really want `master`'s CI to never fail, we could consider adding a merge condition that requires the PR to be based on (or have merged) the current `HEAD` of `master`.

That said, I'm unsure how to go about this, and whether or not avoiding the occasional breakage is worth the time sink required to constantly rebase or merge `master` for PRs everytime a newer PR lands might be questionable.

otrho:
I think this is a bit of an edge case and unless there's some existing magic we can employ, it's probably not going to be worth it.

adlerjohn:
We could enforce that the branch is up-to-date before merging, but that would require a re-review for every PR that is approved but not yet merged and another PR is merged before it, which would be unscalable as we add more people working on the same codebase.

Generally speaking these breaks have been few and far in between, that a notification on `master` CI breaking is sufficient. If the specific fix is not immediately obvious, the offending PR can simply be reverted.

mitchmindtree:
Gitlab have a "Merge Trains" feature that addresses this issue:

https://docs.gitlab.com/ee/ci/pipelines/merge_trains.html

Looks like GitHub have something similar called "Merge Queues":

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue

Could be worth looking into this soon as we seem to be running into this more frequently as activity picks up.

PR #1569, opened 2022-05-17
Fix CI not installing mdbook-forc-documenter preprocessor. Fix out-of-date `Cargo.lock` file missing pest dependencies.

binggh:
Fixes CI not installing `mdbook-forc-documenter` as a preprocessor before trying to do `mdbook build`, which ended up creating empty Forc command pages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,6 +113,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+      - uses: Swatinem/rust-cache@v1
+      - name: Install mdbook-forc-documenter
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: --debug --path ./scripts/mdbook-forc-documenter
       - name: Run mdbook build
         uses: peaceiris/actions-mdbook@v1
         with:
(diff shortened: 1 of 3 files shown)


mitchmindtree:
Tests are breaking, it looks like this is because `mdbook` crate used by the preprocessor depends on pest, but those pest dependencies were removed in #1407, so seeing as the preprocessor was branched from master before that landed we missed that change. Another case of #1343 :smiling_face_with_tear: Should be fine to run `cargo build` locally to build the workspace, then push the `Cargo.lock` changes that occur as a result.

mitchmindtree:
Oh, we'll also want to install the preprocessor here too:

https://github.com/FuelLabs/sway/blob/master/.github/workflows/gh-pages.yml

^ That's the action that runs for the actual publishing of the book. Apologies, I should have caught this in #1545!

mitchmindtree:
Final thing to solve is the `build-mdbook` job.

My guess is that this is due to the `build-mdbook` job using an older version of `mdbook` (`0.4.7`), while the preprocessor uses a much more recent version of the `mdbook` crate (lock file lists `0.4.15`). It could be worth trying to update the `mdbook-version` on the "Run mdbook build" action of both the `build-mdbook` CI job and the gh-pages workflow to use `0.4.15`?

It's strange we don't get any logs whatsoever out of this "Run mdbook build" action that we're using.

mitchmindtree:
Oh, we just needed to install `forc` as the preprocessor uses it :facepalm: Good catch!

mitchmindtree (approved):

binggh:
> Oh, we just needed to install `forc` as the preprocessor uses it 🤦 Good catch!

😵‍💫

Issue #409, opened 2021-11-19
Explicit ordering of `dep` statements is required to resolve dependencies.


otrho:
Found by @nfurfaro:

> In this case,
> ```
> library std;
>
> dep ops;
> dep hash;
> dep storage;
> dep constants;
> dep chain;
> dep b512;
> dep ecr;
> ```
> ecr uses b512, so b512 must be declared first.

PR #1572, opened 2022-05-17
Adding a generic `log` function

### LINKED ISSUE #823: Add `println!` and `print!`
Per discussion on Slack, it would be nice to have `println!` and `print!` for devs 😄 

Per @nfurfaro:

> Would be nice to integrate this with the tooling, so that println! in a contract would work for debugging on a local node (like hardhat’s console.sol , and would function like the Rust println! in a script.
>
> It was such a win for debugging Solidity when console.sol was introduced.
>
> ideally, forc could strip out these println! s when compilling for release, and leave them in for debugging
>
> no, and that’s why solidity doesn’t have it. Hardhat added a special console.sol contract you could inherit from which added the capability, but only when running against a local hardhat node.

### LINKED ISSUE #1358: Should `logging` be in it's own file in the `std` lib
Right now we have logging in `chain`, but I think it should be its own module, i.e. `logging` or something along those lines.

Thoughts @nfurfaro ?

nfurfaro:
This PR 
- uses the `__is_reference_type` and `__size_of` intrinsics to use either `log` or `logd` internally.
- deprecates the type-specific log functions, ie: `log_u64`, `log_b256`, etc
- Adds a `logging` module and reorgs the lib a little to bypass this blocker: https://github.com/FuelLabs/sway/issues/409
- preserves a bare `chain` module to prevent breaking everything that uses `chain/auth` (this feels a bit hacky... probably better to keep it clean, remove the chain mod and force users to update the path.)

close #1358
close #823


nfurfaro:
Ahh. I should really add more tests, there's only one existing atm :)

nfurfaro:
After #1573 lands I'll merge `master` re-push this.

PR #1573, opened 2022-05-17
Adding an intrinsics module

### LINKED ISSUE #1532: Include an `intrinsics` module in the standard library that simply wraps existing intrinsics.

nfurfaro:
All in the title.

close #1532


sway-lib-std/src/intrinsics.sw
```diff
@@ -0,0 +1,13 @@
+//! Exposes compiler intrinsics as stdlib wrapper functions.
+library intrinsics;
+
+
```
adlerjohn, line 4:
```suggestion

```

sway-lib-std/src/intrinsics.sw
```diff
@@ -0,0 +1,13 @@
...
+/// Returns whether a generic type `T` is a reference type or not.
+pub fn is_reference_type::<T>() -> bool {
+    __is_reference_type::<T>()
+}
+
+/// Returns the size of a generic type `T` in bytes.
+pub fn size_of::<T>() -> u64 {
+    __size_of::<T>()
+}
```
adlerjohn, line 13: newline

adlerjohn (commented):
Does this actually work?? I'd like to see some tests.

adlerjohn:
Converting to draft until tested and CI passes.

sway-lib-std/src/intrinsics.sw
```diff
@@ -0,0 +1,12 @@
+//! Exposes compiler intrinsics as stdlib wrapper functions.
+library intrinsics;
+
+/// Returns whether a generic type `T` is a reference type or not.
+pub fn is_reference_type<T>(value: T) -> bool {
```
mohammadfawaz, line 5: Do we need to pass a `value` here?

nfurfaro: we don't for the intrinsic itself... I can test without the param for this wrapper function.

mohammadfawaz: Yeah and then you'd have to call the intrinsics using `is_reference_type::<u64>()` and `size_of::<u64>()` for example.

nfurfaro: @mohammadfawaz That seems to work if I use it like this for example:
```
pub fn is_reference_type<T>() -> bool {
    __is_reference_type::<T>()
}

assert(!is_reference_type::<u64>());
```

But then I can't use `is_reference_type<T>()` to check an instance of a type, only the parent type itself...

mohammadfawaz: Hmmm yeah that wouldn't be possible, but is that something we want? 

We do have `size_of_value` that actually takes a value so we have that distinction between `size_of` and `size_of_value` (Rust has the same distinction). If we really want a version of `is_reference_type` that takes a value, then I think we would need a separate function here `is_reference_value` or something.

adlerjohn: > I want to be able to do this: assert(!is_reference_type(42));

Why do you want to do that? You know `42` isn't a reference type.

adlerjohn (commented):
Remove the use of value since that's not how `is_reference_type()` and `size_of()` are used.

sway-lib-std/src/intrinsics.sw
nfurfaro: 42 was just an example. :)
Curently, the way we;ve been using these intrinsic is to have an outer function with a generic param like so:
```
/// Returns the SHA-2-256 hash of `param`.
pub fn sha256<T>(param: T) -> b256 {
    let mut result_buffer: b256 = ZERO;
    if !__is_reference_type::<T>() {
        asm(buffer, ptr: param, eight_bytes: 8, hash: result_buffer) {
            move buffer sp; // Copy stack pointer to memory at "buffer"
            cfei i8; // Grow stack by 1 word
            sw buffer ptr i0; // Save value in register at "ptr" to memory at "buffer"
            s256 hash buffer eight_bytes; // Hash the next eight bytes starting from "buffer" into "hash"
            cfsi i8; // Shrink stack by 1 word
            hash: b256 // Return
        }
    } else {
        let size = __size_of::<T>();
        asm(hash: result_buffer, ptr: param, bytes: size) {
            s256 hash ptr bytes; // Hash the next "size" number of bytes starting from "ptr" into "hash"
            hash: b256 // Return
        }
    }
}
```
I thought this module's wrapper functions would follow the same pattern, but perhaps I misunderstood.
As far as I can tell, in the sha256 example above, the only way `__is_reference_type::<T>()` knows what `T` is is because of `param`.

nfurfaro: I think I see where my issue lies.
I was treating `is_reference_type<T>()` as the outer function (analogous to the `sha256<T>()` function above.
But really, for `is_reference_type<T>()` to know what `T` is without hardcoding the type, it needs to be used inside another generic function.

test/src/sdk-harness/test_projects/intrinsics/src/main.sw
```diff
@@ -0,0 +1,53 @@
...
+    field_1: bool,
+    field_2: u64,
+}
+fn is_ref_type<T>(param: T) -> bool {
+    is_reference_type::<T>()
+}
+
+fn get_size_of<T>(param: T) -> u64 {
+    size_of::<T>()
+}
```
otrho, line 17: I wonder if the actual library functions in the `intrinsics` module should do this?  They seem a bit redundant otherwise.  What is the motivation around creating the wrappers in the new module -- to hide the `__` prefix?

adlerjohn: > to hide the `__` prefix?

Yes, because the alternative is to make the specific names of the intrinsics part of the public API of the language. Which we could certainly do, but then it's one more set of keywords people have to account for.

>  But I don't anticipate needing to put anything else in this module particularly..?

Can't really think of anything either, but that's fine! Not every intrinsic needs to be available to the end user, as you said.

nfurfaro: I'm in favor of these wrappers to hide the `__` prefix. It helps keep a nice uniform api surface for users.

otrho (approved):

PR #1581, opened 2022-05-18
Fix bug where trait/abi method argument type mismatch went uncaught

### LINKED ISSUE #1327: Compiler should error when the argument types of a trait/abi method implementation do not match the declaration.
The contract below compiles without any errors even though the definition of `foo` in the ABI has a different signature than its implementation in the `impl` block (note the `u64` argument v.s. `str[7]`).

```rust
contract;

abi MyContract {
    fn foo(x: u64) -> str[7];
}

struct S {
    x: u64,
}

    fn foo(s: str[7]) -> str[7] {
        s
    }
}
```

mitchmindtree:
Closes #1327.

Turns out unification was already catching this, but the produced error
was never returned from the trait impl checking function :D

It looks like a new `Vec` was introduced at one point to collect errors
in order to work around ownership issues involved with some fancy uses
of `iter().find_map()`, however these errors were never collected from
the outer scope. I've refactored this section to use basic for loops and
control flow instead which makes it easier for Rust to reason about
ownership over the top-level `errors` Vec and reduces the need for the
extra error-collecting `Vec`s.

Also adds two new `should_fail` tests to verify that we produce errors
when argument types differ between implementation and trait/abi
declaration.


mitchmindtree:
Apologies for the extra commit, I spotted a chance to simplify further and couldn't help myself :smiling_face_with_tear: 

OK, ready for review!

otrho (approved):
Ha, I literally need this right now, as I'm creating impurity mismatch errors in this file and they're not getting reported. :metal:

Issue #1220, opened 2022-04-11
Add list of native `forc plugins` to the book


eureka-cpu:
There should be a list of forc plugins and an install example in the book

Issue #1577, opened 2022-05-18
Change Sway Book `latest` to point to latest semver release, not master branch.


mitchmindtree:
@binggh raised a valid concern in a recent chat that currently the book hosted under `latest` is always the current state of `master`, and does not necessarily match the state of `forc` that you get when you `cargo install forc`.

We should do something like the following:

1. Rename the URL suffix of the `master` branch book to `master`, e.g. `https://fuellabs.github.io/sway/master` (rather than `latest`).

2. Add some CI automation where, on each new version release of `forc`, we also host the current state of the book under a URL for that version, e.g. `https://fuellabs.github.io/sway/0.12.2`.

3. Change the behaviour of the `https://fuellabs.github.io/sway/latest` URL to point to the latest versioned release, i.e. the version you get when  you do `cargo install forc`, **not** the latest master release.

binggh:
In this way it also assures us that we know for sure that code examples and forc command documentation are accurate for a certain version.

PR #1582, opened 2022-05-18
document native plugins within the Sway book

### LINKED ISSUE #1295: `mdbook-forc-documenter`: support documentation for native plugins
Related discussions about `forc --list`:
https://github.com/FuelLabs/sway/pull/1047#discussion_r846969076
https://github.com/FuelLabs/sway/issues/702

Currently, `forc-documenter` uses `forc --help` to generate documentation for forc commands. When/if we have `forc --list` ready to support listing forc commands + forc plugins, we should use that instead.

binggh:
Closes #1295 #1220

This PR documents Forc native plugins within the Sway book: namely `fmt`, `explore`, and `lsp`

Points to note:
- Includes some refactoring from the previous `mdbook-forc-documenter` PR to have better separation of concerns
- Support parsing of `subcommand` help text - previously within `forc-documenter`, native commands had no subcommands. `forc-explore` has the `clean` subcommand and this PR includes code to handle parsing subcommands.

P.S.: Note that the new .md files will be file stubs due to mdbook link preprocessor requiring them to work - to see the actual html output, please pull from this branch and do `cargo install --path scripts/mdbook-forc-documenter/ && mdbook build docs --open`

The plugins will appear as new sub-chapters under `6.4 Plugins`.


.github/workflows/gh-pages.yml
```diff
@@ -23,6 +23,11 @@ jobs:
         with:
           command: install
           args: --debug --path ./forc
+      - name: Install Forc plugins
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: forc-fmt forc-lsp forc-explore
```
adlerjohn, line 30: If we don't need release builds here (don't think we do), then debug builds will be faster.
```suggestion
          args: --debug forc-fmt forc-lsp forc-explore
```

binggh: that makes sense - thanks for the catch!

scripts/mdbook-forc-documenter/src/plugins.rs
```diff
@@ -0,0 +1,3 @@
+pub fn plugin_commands() -> Vec<String> {
+    vec!["fmt".to_string(), "explore".to_string(), "lsp".to_string()]
```
mitchmindtree, line 2: It might be worth adding a note about updating this to the README when new plugins are added?

Alternatively, we could create a `forc-plugins` directory and move these plugins into this directory. Then we could programmatically discover each of these without having to be aware of updating this list.

---

Also, for lists of known constant strings like this, I think something like the following would be more Rusty:

```rust
pub const PLUGIN_COMMANDS: &[&str] = ["fmt", "explore", "lsp"];
```

You might need to specify the `'static` lifetime on the references, but I think the Rust crew were working towards removing the need for specifying the `'static` lifetime in `const`s.

This saves the 4 dynamic allocations required for the 3 `String`s and the `Vec` that contains them. In our case there's no way this would have any noticable impact on our workflow, but might be worth being mindful of.

mitchmindtree (changes requested):
Tested locally, looking good!

Currently the `mdbook-forc-documenter` README only mentions commands, but could be worth mentioning that plugins should be updated in a similar manner.

.github/workflows/ci.yml
```diff
@@ -123,6 +123,11 @@ jobs:
         with:
           command: install
           args: --debug --path ./forc
+      - name: Install Forc plugins
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: --debug forc-fmt forc-lsp forc-explore
```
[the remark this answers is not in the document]
binggh: would we want to document changes from the latest release or from the local crates that we get from `--path`, or is this a non-issue?

scripts/mdbook-forc-documenter/src/plugins.rs
binggh: Hmm the problem is when we're calling `forc --help` to get native commands we're getting a variable number of [`possible_forc_commands()`](https://github.com/FuelLabs/sway/pull/1582/files#diff-68d5c1aed8bd2b9759a0a6f05a0b40fda007ae93bb9172c14bbf86c40735c9c5R7) sometimes this includes invalid commands that we call later when we do `forc <COMMAND> --help`, and we discard invalid commands in the script. The reason why I did it like this is so we do not have to manually add/remove commands to a const slice.

Then we're using that as input into `get_contents_from_commands()` which accepts `&Vec<String>` - that's why I chose to return `Vec<String>` for the `plugin_commands()` as well, unless there's a more idiomatic Rust way to do this(?)

binggh: > Alternatively, we could create a forc-plugins directory and move these plugins into this directory. Then we could programmatically discover each of these without having to be aware of updating this list.

This sounds like a good idea, a new `forc-plugins` in the proj root?

mitchmindtree: > This sounds like a good idea, a new forc-plugins in the proj root?

Yeah that sounds good! Might also have to update the top-level `[workspace]` for the new paths, and perhaps update any references to the old paths in CI.

.github/workflows/ci.yml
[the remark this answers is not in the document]
mitchmindtree: I think we should do the same as we do for `forc`, i.e. from `--path`. We can worry about the behaviour of `latest` w.r.t. selecting the latest published release rather than `master` in #1577.

scripts/mdbook-forc-documenter/src/commands.rs
```diff
@@ -0,0 +1,108 @@
...
+        }
+        if line == "SUBCOMMANDS:" {
+            has_parsed_subcommand_header = true;
+        }
+    }
+
+    possible_commands
+}
+
+pub fn get_contents_from_commands(commands: &Vec<String>) -> HashMap<String, String> {
```
mitchmindtree, line 32:
```suggestion
pub fn get_contents_from_commands(commands: &[String]) -> HashMap<String, String> {
```

Whenever you have a `&Vec<T>`, 99% of the time you want a `&[T]` instead.

This is because the only distinction between a `Vec` and a slice is that a `Vec` is growable, however it cannot be grown behind an immutable reference. Arguments that take a slice are strictly more flexible, as they support not just `&Vec`s via auto-deref, but other slices, fixed-size arrays or other abstractions over slices.

This isn't super important for this PR, but just worth keeping in mind and is related to your comment [here](https://github.com/FuelLabs/sway/pull/1582#discussion_r876589051).

scripts/mdbook-forc-documenter/src/plugins.rs
mitchmindtree: @binggh ah that makes sense if you're using this with another API that is more dynamic!

`get_contents_from_commands` taking a `&Vec` should be addressed (I [left a comment](https://github.com/FuelLabs/sway/pull/1582/files#r876675992) on how and why), however it will still need to take a slice of `String`, and `String` must be dynamically allocated either way so I think it's fine to disregard my suggestion here (or go with the `forc-plugins` directory instead).

binggh: Ah yeah that comment makes sense, I'll update that + move the plugins into a `forc-plugins` directory as well

adlerjohn (commented):
merge conflicts

scripts/mdbook-forc-documenter/src/plugins.rs
```diff
@@ -0,0 +1,28 @@
+use std::fs;
+use std::path::PathBuf;
+
+fn get_sway_path() -> PathBuf {
+    let mut curr_path = std::env::current_dir().unwrap();
+    loop {
+        if curr_path.ends_with("sway") {
```
mitchmindtree, line 7: Just because the repo's name is `sway` does not mean that a local clone of the repo is. E.g. while less common, there are many git users out there that do stuff like `git clone https://github.com/fuellabs/sway Sway` or give it some other name that suits them.

Instead, can we make this a `find_forc_plugins_dir` where, instead of looking for a path ending with `sway`, we keep looping through parents (as we currently do) until the directory at `curr_path` contains a `forc-plugins` subdirectory and return that?

scripts/mdbook-forc-documenter/src/plugins.rs
```diff
@@ -0,0 +1,28 @@
...
+
+pub fn plugin_commands() -> Vec<String> {
+    let plugins_dir = get_sway_path().join("forc-plugins");
+    let mut plugins: Vec<String> = Vec::new();
+
+    for entry in fs::read_dir(&plugins_dir)
+        .expect("Failed to read plugins directory")
+        .flatten()
+    {
+        let path = entry.path();
```
mitchmindtree, line 22: We should at least check that the `path` is 1. a directory and 2. it starts with `forc-`. If it doesn't meet these criteria, we should `continue` onto the next entry (and not `unwrap` or `panic!`).

Otherwise we will run into issues if e.g. we decide to put a `README.md` in this directory to explain what crates are allowed in here, or a developer's editor puts some temporary file in here, etc.

mitchmindtree (changes requested):

scripts/mdbook-forc-documenter/src/plugins.rs
```diff
@@ -0,0 +1,48 @@
...
+            return Ok(curr_dir.join("forc-plugins"));
+        }
+
+        curr_dir = curr_path_buf.parent().unwrap();
+    }
+
+    Err(anyhow!(
+        "Could not find Cargo.toml in the project directory"
+    ))
+}
```
mitchmindtree, line 26: I think we can simplify this a bit further and reduce the need to loop over each dir twice.

I haven't tested this, but maybe something like:

```rust
fn find_forc_plugins_dir() -> PathBuf {
    let mut curr_path = std::env::current_dir().unwrap();
    loop {
        if let Ok(entries) = fs::read_dir(&curr_path) {
            if !curr_path.join("Cargo.toml").exists() {
                continue;
            }
            for entry in entries.filter_map(Result::ok) {
                let path = entry.path();
                if path.is_dir() && path.ends_with("forc-plugins") {
                    return Ok(path);
                }
            }
        }
        curr_path = curr_path
            .parent()
            .ok_or_else(|| anyhow!("Could not find Cargo.toml in the project directory"))?;
    }
}
```

mitchmindtree (changes requested):

scripts/mdbook-forc-documenter/src/plugins.rs
binggh: Oh yeah this def looks cleaner, going to use this instead

mitchmindtree:
I've opened #1618 with some final tweaks for this PR.

scripts/mdbook-forc-documenter/src/formatter.rs
```diff
@@ -37,14 +40,15 @@ pub fn is_arg(token: &str) -> bool {
 }
 
 pub fn format_header_line(header_line: &str) -> String {
-    "\n# ".to_owned() + header_line + "\n"
+    return "\n# ".to_owned() + header_line.split_whitespace().next().unwrap() + "\n";
```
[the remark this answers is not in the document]
binggh: oh nice catch, think I was trying to manipulate the string in a line above and added the return, then realised I could do it in-line and forgot to remove the `return`. But you're right about it being strange `clippy` didn't complain

mitchmindtree (approved):
:tada:

mitchmindtree:
Ah, looks like there are a few conflicts

PR #1588, opened 2022-05-18
Improvements to namespace API and declaration ABI.

emilyaherbert:
This is built off of #1586

PR #1589, opened 2022-05-18
Introduce the `CopyTypes` trait and the `TypeMapping` type.

emilyaherbert:
This is built off of #1588 

This replaces all instances of the manual implementation of `copy_types` with the `CopyTypes` trait.


otrho (approved):
This seems OK except it doesn't really change anything?  If monomorphization took a `T: CopyTypes` maybe?  Or does it, and I missed it?   I guess if you try to call `copy_types()` on a type that isn't `CopyTypes` then it should complain specifically about the missing method from the trait... 🤔

emilyaherbert:
>  after this goes in, I will take some of that refactoring and apply it here as well so we have the best of both worlds

Great! Also note: #1591 and #1592, which refactor the type engine significantly. But, both of those are ready to go in ASAP, if they would be helpful to your changes 😄

PR #1591, opened 2022-05-18
Introduce the `MonomorphizeHelper` trait.

emilyaherbert:
This is built off of #1590 

This change replaces all of the instances of `monomorphize_inner` with the `MonomorphizeHelper` trait.

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/declaration/monomorphize.rs
+++ b/sway-core/src/semantic_analysis/ast_node/declaration/monomorphize.rs
@@ -1,0 +1,12 @@
+use sway_types::{Ident, Span};
+
+use crate::{namespace::Items, semantic_analysis::TypeMapping, TypeParameter};
+
+pub(crate) trait MonomorphizeHelper {
+    type Output;
+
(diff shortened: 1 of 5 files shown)


sway-core/src/semantic_analysis/ast_node/declaration/monomorphize.rs
```diff
@@ -0,0 +1,12 @@
+use sway_types::{Ident, Span};
+
+use crate::{namespace::Items, semantic_analysis::TypeMapping, TypeParameter};
+
+pub(crate) trait MonomorphizeHelper {
+    type Output;
+
+    fn type_parameters(&self) -> &[TypeParameter];
+    fn name(&self) -> &Ident;
+    fn span(&self) -> &Span;
```
mitchmindtree, line 10: Are the `type_parameters`, `name` or `span` methods used anywhere?

emilyaherbert: They are used in the PR following this one. I added them here to make that change more clear.

otrho (approved):

PR #1595, opened 2022-05-18
Finish parsing `where` clauses

mohammadfawaz:
Parsing `where` clauses was not complete. Finishing that and adding some error handling.
Also, changing the "where clauses are not supported" to an actual compiler error as the current internal error asks users to file an issue. This is all temporary anyways. We will start removing the errors one by one as we introduce support for `where`.

Tests will be added with a follow up change handling semantic analysis.


sway-core/src/parse_tree/declaration/type_parameter.rs
```diff
@@ -59,5 +59,5 @@ impl TypeParameter {
 
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub(crate) struct TraitConstraint {
-    pub(crate) name: Ident,
+    pub(crate) name: CallPath,
```
emilyaherbert, line 62: Would you be able to change this from `name` to `call_path`?

emilyaherbert (changes requested):

sway-core/src/parse_tree/declaration/type_parameter.rs
mohammadfawaz: Done!

otrho (dismissed):

emilyaherbert (approved):

PR #1599, opened 2022-05-18
Cleanup constants.sw & add impl b256 to core/num

### LINKED ISSUE #1597: Cleaning up constants.sw from the standard library
Looking at https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/constants.sw:
* `ZERO` should probably be moved into https://github.com/FuelLabs/sway/blob/master/sway-lib-core/src/num.sw into an `impl` block for `b256`. it is a bit odd having a global constant named `ZERO` without any indication as to what its type is.
* `MAX_U64` can be removed now in favour of `u64::max()`.
* `MAX_B256` can also be moved into an `impl` block for `b256` in `num.sw`.

nfurfaro:
Close #1597 

I opted to follow the established naming of `min` & `max`, rather than something like `zero` & `full` / `saturated`.


sway-lib-core/src/num.sw
```diff
@@ -71,3 +71,21 @@ impl u8 {
         8
     }
 }
+
+impl b256 {
+     /// The smallest value that can be represented by this type.
```
adlerjohn, line 76: format file

sway-lib-core/src/num.sw
```diff
@@ -71,3 +71,21 @@ impl u8 {
...
+    pub fn min() -> b256 {
+        0x0000000000000000000000000000000000000000000000000000000000000000
+    }
+
+    /// The largest value that can be represented by this type,
+    /// 2<sup>8</sup> - 1.
```
adlerjohn, line 82: this isn't right

examples/hashing/src/main.sw
```diff
@@ -2,10 +2,10 @@ script;
 
 use std::{
     chain::log_b256,
-    constants::ZERO,
     core::num::*,
     hash::{HashMethod, hash_pair, hash_u64, hash_value, sha256}
 };
+use core::num::*;
```
adlerjohn, line 8: Hmm @sezna would it be a good idea to implicitly import `core::num::*` in the predule too?

sway-lib-core/src/num.sw
nfurfaro: weird! will fix, no idea where that came from.

adlerjohn: No, the `<sup>` was needed. See the Rust source code. Please revert. What's not right is that this line only here says `8` instead of `256`.

otrho (dismissed):

otrho:
@nfurfaro Still failing...

otrho (approved):

Issue #1578, opened 2022-05-18
Parse all modules in a project before beginning the type-checking stage


mitchmindtree:
Currently, we parse and type-check each module in a sway project one at a time in a rough depth-first order.

We should separate the concerns of parsing and and type-checking into separate stages both for code clarity and to unlock solutions to many issues related to inter-module definitions and the order in which we type-check items.

From a quick scan, I believe this is a blocker (or at least a first-stage) for addressing the following:

- #201
- #409
- #870
- #1267
- #1557.
- @emilyaherbert's recent idea w.r.t. introducing a new "collection" stage prior to type-checking (no issue yet).

## Potential Solution

1. The existing `sway_parse::Program` type more closely matches what we currently conceive of as a parsed "module" in Sway. We should rename what is currently referred to as `program` to `module`. I.e. `sway_parse::program` becomes `sway_parse::module`, `sway_parse::Program` becomes `sway_parse::Module`.

2. Re-introduce a new `sway_parse::Program` type that is a small wrapper around a tree of `sway_parse::Module`s.

3. Add a high-level function the `sway_parse` lib.rs called `parse_project` that begins by calling `parse_file` for the root module, then parses loads and parses all submodules based on the parent module's parsed `dependencies`.

4. Refactor the `type_check` step in terms of the new `sway_parse::Program` type. More specifically, refactor the `import_new_file` function to `import_new_dependency` where rather than loading and parsing the file, we look up the pre-parsed module in the module tree.

---

Marking as `P: Critical` as this is a blocker for #409 (critical) as well as a number of other high priority issues.

canndrew:
I'm not sure that the logic for hunting down files for sub-modules is something that belongs in the parser. The parser is, in my mind at least, just something that converts a block of text to an AST (where that AST represents the structure of the original text as literally as possible). The old parser already had *far* too much logic built into it which is why the old AST still exists and we now have another AST layer on top of it. So I'm wary of heading in that direction again. I think it's important to keep the stages of the compilation pipeline as simple and well-defined as possible.

I think the general idea of what you're proposing is right though, just that it belongs in sway-core. Consider, for instance, that if we ever decide to add rust-like macros then it won't be possible to know about sub-modules until after macro expansion.

mitchmindtree:
Thanks for the input @canndrew, that makes a lot of sense.

Revising the "Proposed Solution" above with this in mind, I'm thinking we:

- Keep step 1.
[...]

Issue #1548, opened 2022-05-14
function composition in impl, method not found


simonr0204:
Attempting to compose functions in an impl block results in `no method found`. E.g. 

```
pub struct A {
    value: u64,
}

impl A {
    fn a(self, a: u64) -> A {
        A {
            value: self.value + a
        }
    }

    fn b(self, b: u64) -> A {
        self.a(b)
    }
}
```

:

```
46 |         self.a(b)
   |              ^ No method named "a" found for type "struct A<u64>".
```

Full repro here: https://github.com/FuelLabs/sway/blob/simon-impl-function-composition-bug/test/src/sdk-harness/test_projects/temp/src/main.sw

adlerjohn:
This is blocking implementation of more complex stdlib types, so moving to critical priority.

otrho:
I have taken a look at this and although the problem is quite straight forward, I'm not sure what the solution is.

The problem is while we're type checking the trait methods (i.e., converting them from function declarations to _typed_ function declarations) none of them are in a namespace yet, so they can't refer to each other.

https://github.com/FuelLabs/sway/blob/82ad97ac10560f2339c13fcca9126602da3fc70f/sway-core/src/semantic_analysis/ast_node/mod.rs#L504-L541

The call which fails (or generates the errors) is on line 536.

Initially I thought the solution might be to sort the methods based on dependency like we do for top level functions.  But they're not added to the namespace individually so that wouldn't help.  They get added as a collection here:

https://github.com/FuelLabs/sway/blob/82ad97ac10560f2339c13fcca9126602da3fc70f/sway-core/src/semantic_analysis/ast_node/mod.rs#L547-L551

So then I thought we could add them to the temporary namespace (`impl_namespace` in the above code) so that they could type check.  But that involves creating `TypedFunctionDeclaration`s for each of the methods first, which is exactly what we're trying to do in the above code anyway.  Chicken and egg.

I thought 'dummy' `TypedFunctionDeclaration`s could work, but they still need to have all the `TypeInfo` values in the `FunctionDeclaration`s converted to `TypeId` which seems wasteful, plus faking the body and spans and other meta all seems very _hacky_. 

But unless I'm missing a simple path here (I hope I am TBH!) then that may be the only way to fix this.  That, or to add the ability to inject very lean non-typechecked function signatures into the namespace temporarily so these guys can type check properly afterwards.

[...]

Issue #609, opened 2022-01-10
Match expression arms don't support simple expressions.


otrho:
E.g.,
```
fn main() -> bool {
    match false {
        true => false,
        foo => foo,
    }
}
```
This produces errors:
```
warning
 --> main.sw:0:17
  |
...
4 |         true => false,
  |                 ----- This returns a value of type bool, which is not assigned to anything and is ignored.
  |
warning
 --> main.sw:0:17
  |
...
5 |         foo =>  foo ,
  |                 --- This returns a value of type bool, which is not assigned to anything and is ignored.
  |
error
 --> main.sw:0:11
  |
...
3 |       match false {
  |  ___________^
4 | |         true => false,
  | |____________^ No method named "eq" found for type "bool".
  |
error
 --> main.sw:0:17
  |
...
4 |         true => false,
  |                 ^^^^^ Mismatched types.
expected: bool
found:    ().
```
This final error is repeated a few more times.

All of the type errors go away if the match arm expressions are bracket code blocks, e.g., `true => { false },`.

The warning about `no method named "eq" for type "bool"` is separate; looks like we need to implement `Ord` for `bool` still.

otrho:
I've made https://github.com/FuelLabs/sway-lib-core/issues/8 for the `eq` thing.

emilyaherbert:
This has to do with how the parser interprets implicit return statements. The fix is probably in changing how we handle implicit returns.

PR #1600, opened 2022-05-18
Add `U128` type

### LINKED ISSUE #1478: Add new type `U128`
We need a type to contain the result of multiplying 2 `u64`s for example.
This should be opt-in, as in the default when performing math ops with a type should be to return the same type.

Bikeshedding discussion here: https://fuellabs.slack.com/archives/C01KBNGSWGJ/p1651596324523349

adlerjohn:
Fixes #1478


sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    pub fn lt(self, other: Self) -> bool {
+        self.upper < other.upper || self.upper == other.upper && self.lower < other.lower
+    }
+}
+
+// TODO this doesn't work?
+// impl core::ops::OrdEq for U128 {
+// }
```
adlerjohn, line 50: If I uncomment these lines I get the following error:
```console
error
  --> sway-lib-std/src/u128.sw:49:1
   |
47 |   
48 |   // TODO this doesn't work?
49 | / impl core::ops::OrdEq for U128 {
50 | | }
   | |_^ The trait "Ord" is not implemented for type "struct U128<u64, u64>"
51 |   
52 |   fn disable_overflow() {
   |
```

even though `Ord` is clearly defined above. I don't think if this is because it's specifically expecting trait `Ord` instead of `core::ops::Ord`, or because or the order of definitions.

adlerjohn: cc @otrho @emilyaherbert

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+        // If shifting by at least the number of bits, then saturate with
+        // zeroes.
+        if (other >= 128) {
+            return ~Self::new();
+        };
+
+        // If shifting by at least half the number of bits, then upper word can
+        // be discarded.
+        if (other >= 64) {
+            return ~Self::from(self.lower << (other - 64), 0);
```
adlerjohn, line 167: The compiler emits an incorrect warning that this line is unreachable. cc @otrho

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+}
+
+fn enable_overflow() {
+    asm(r1) {
+        movi r1 i0;
+        flag r1;
+    }
+}
+
+impl u64 {
```
Braqzen, line 66: This is the `u128` library.
This should be moved into either `core::num` or `core::ops`

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+    pub fn overflowing_add(a: u64, b: u64) -> U128 {
+        disable_overflow();
+        let mut v = U128 {
+            upper: 0,
+            lower: 0,
+        };
+        asm(r1, r2, a: a, b: b, v_ptr: v) {
```
Braqzen, line 73: Also, name the variables please. Anyone learning should have an easier time reading accurately named variables

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+    asm(r1) {
+        movi r1 i0;
+        flag r1;
+    }
+}
+
+impl u64 {
+    pub fn overflowing_add(a: u64, b: u64) -> U128 {
+        disable_overflow();
+        let mut v = U128 {
```
Braqzen, line 69: Should this have a `default()` or `new()` which achieves the same result?
Oh, it has a `new()` below. Why not use it here?

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+    }
+}
+
+// TODO this doesn't work?
+// impl core::ops::OrdEq for U128 {
+// }
+
+fn disable_overflow() {
+    asm(r1) {
+        movi r1 i3;
```
Braqzen, line 54: We do have the opcodes page but adding a short comment per line would really help (rather than going back and forth and figuring out how each opcode works)

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+        U128 {
+            upper: 0,
+            lower: 0,
+        }
+    }
+
+    /// Downcast to `u64`. Err if precision would be lost, Ok otherwise.
+    pub fn to_u64(self) -> Result<u64, ()> {
+        match self.upper {
+            0 => {
```
Braqzen, line 112: Are the brackets required here or can we do
```
0 => Result::Ok(self.lower),
_ => Result::Err(()),
```

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+                Result::Ok(self.lower)
+            },
+            _ => {
+                Result::Err(())
+            },
+        }
+    }
+
+    /// The smallest value that can be represented by this integer type.
+    pub fn min() -> U128 {
```
Braqzen, line 122: Can we be cheeky and use the `new()` function from above?

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+            lower = ~u64::max() - (other.lower - self.lower - 1);
+            upper = upper - 1;
+        } else {
+            lower = self.lower - other.lower;
+        };
+
+        U128 {
+            upper: upper,
+            lower: lower,
+        }
```
Braqzen, line 253:
```suggestion
        U128 {
            upper, lower,
        }
```

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+}
+
+impl core::ops::Multiply for U128 {
+    // Multiply a U128 with a U128. Panics of overflow.
+    pub fn multiply(self, other: Self) -> Self {
+        let zero = ~U128::from(0, 0);
+        let one = ~U128::from(0, 1);
+
+        let mut total = ~U128::new();
+        let mut i = 128 - 1 + 1;
```
Braqzen, line 264:
```suggestion
        let mut i = 128;
```

Code quality, you can add a comment if you want to explain the `+1-1`

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+    // Divide a U128 by a U128. Panics if divisor is zero.
+    pub fn divide(self, divisor: Self) -> Self {
+        let zero = ~U128::from(0, 0);
+        let one = ~U128::from(0, 1);
+
+        assert(divisor != zero);
+
+        let mut quotient = ~U128::new();
+        let mut remainder = ~U128::new();
+        let mut i = 128 - 1 + 1;
```
Braqzen, line 291:
```suggestion
        let mut i = 128;
```
Code quality, you can add a comment if you want to explain the `+1-1`

test/src/e2e_vm_tests/mod.rs
```diff
@@ -121,6 +121,9 @@ pub fn run(filter_regex: Option<regex::Regex>) {
             ProgramState::Return(1), // true
         ),
         ("should_pass/stdlib/ge_test", ProgramState::Return(1)), // true
+        ("should_pass/stdlib/u128_test", ProgramState::Return(1)), // true
```
Braqzen, line 124: Why have 3 tests instead of 1?

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u128_div_test/src/main.sw
```diff
@@ -0,0 +1,25 @@
+script;
+
+use std::assert::assert;
+use std::result::*;
+use std::u128::*;
```
Braqzen, line 5: The `u128` coming from the `std` just looks wrong. I think the additional `types` (like `B512` etc.) should be moved somewhere into `core::`

Braqzen (changes requested):

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+        while i > 0 {
+            // Workaround for not having break keyword
+            let shift = i - 1;
+            total = total << 1;
+            if (other & (one << shift)) != zero {
+                total = total + self;
+            }
+
+            i = i - 1;
+        }
```
otrho, line 275: This is obviously pretty expensive and is why it's running out of gas?

Couldn't the algorithm be reduced to `((lhs.upper * rhs.upper) << 64) + (lhs.upper * rhs.lower + lhs.lower * rhs.upper) + lhs.lower * rhs.lower`?  The `*` and `+` operations would need to be able to grab the overflow and use it, so it's a bit trickier than just that, but it would still be cheaper than looping 128 times... maybe?

It would only work if `$of` actually contains the overflowed bits from a multiply, and not just the flag to indicate an overflow... is that how it works?

sway-lib-std/src/u128.sw
otrho: They're returning `U128`s though, so you'd be introducing a dependency on this file into `num` which I think is less desirable.

sway-lib-std/src/u128.sw
otrho: #609

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+    lower: u64,
+}
+
+pub trait From {
+    /// Function for creating U128 from its u64 components.
+    pub fn from(h: u64, l: u64) -> Self;
+} {
+    fn into(v: U128) -> (u64, u64) {
+        (v.upper, v.lower)
+    }
```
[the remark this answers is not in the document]
simonr0204: hmm - a variable called variable ?  Maybe 'value'  ?

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u128_div_test/src/main.sw
simonr0204: I'm not sure if they belong in `core`, but we did think about factoring them out into `std::types` or something like that. There will be more and more of these abstractions and it would be good to organise them. That might be better handled in a separate PR though.

sway-lib-std/src/u128.sw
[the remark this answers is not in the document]
otrho: `value` or `variable` aren't particularly more expressive than `v` in a one line function like this IMO.

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u128_div_test/src/main.sw
nfurfaro: agree that we could create a subdirectory in `std` to collect these types. There's already an open issue about reorganizing std where I belive this is discussed, will take a look.
`core` is for language-defined primitive types, this fits in std nicely.

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+        flag r1;
+    }
+}
+
+fn enable_overflow() {
+    asm(r1) {
+        movi r1 i0;
+        flag r1;
+    }
+}
```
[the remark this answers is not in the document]
nfurfaro: hmm.  My understanding was that the `flag` op had to be used in the same asm block as the op you're trying to allow overflow for, otherwise, the bit is unset after the block returns. 
Is this working?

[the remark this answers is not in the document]
nfurfaro: Also, perhaps rather than `movi r1 i3;` here you could initialize a second register with the value as a named variable (or even better, as a `const`) to:
   a.) help document the function better, and
   b.) to follow the general guideline of "if it can be done outside of asm, do it outside of asm".
   
   ie:
   ```
const WRAPPING_FLAG_VALUE = 3;
asm(r1: WRAPPING_FLAG_VALUE) {
    flag r1;
}
```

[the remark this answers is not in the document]
Comment (nfurfaro) [reply sway-lib-std/src/u128.sw:64]: And finally, it seems that the flag values for `enable` and `disable` are reversed.
An unset flag (ie: `0`) should be the default state in which panic occurs on overflow.
Setting to a non-zero value should enable overflow.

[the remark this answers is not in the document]
Comment (Braqzen) [reply sway-lib-std/src/u128.sw:64]: > Also, perhaps rather than `movi r1 i3;` here you could initialize a second register with the value as a named variable (or even better, as a `const`) to: a.) help document the function better, and b.) to follow the general guideline of "if it can be done outside of asm, do it outside of asm".
> 
> ie:
> 
> ```
> const WRAPPING_FLAG_VALUE = 3;
> asm(r1: WRAPPING_FLAG_VALUE) {
>  flag r1;
> }
> ```

What about
```
asm(WRAPPING_FLAG_VALUE: 3) {
    flag WRAPPING_FLAG_VALUE;
}
```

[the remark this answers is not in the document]
Comment (nfurfaro) [reply sway-lib-std/src/u128.sw:64]: I think a combination of a descriptively-named register and a const value would be best, and in line with `tx.sw` and other modules where we use named const values to initialize registers with.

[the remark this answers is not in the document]
Comment (nfurfaro) [reply sway-lib-std/src/u128.sw:64]: >Do these belong somewhere more general as pub definitions? Like num or somewhere?

We had talked about a stdlib module for wrapping VM opcodes, this would probably fit the bill. Will look for the open issue.

File: test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u128_div_test/src/main.sw (hunk 13 above)
Comment (adlerjohn) [reply test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u128_div_test/src/main.sw:5]: `core`'s only reason for existence is to define operators on primitive types, so that `std` can use them.

File: sway-lib-std/src/u128.sw [hunk 17]
```diff
@@ -0,0 +1,310 @@
...
+/// The 128-bit unsigned integer type.
+/// Represented as two 64-bit components: `(upper, lower)`, where `value = (upper << 64) + lower`.
+pub struct U128 {
+    upper: u64,
+    lower: u64,
+}
+
+pub trait From {
+    /// Function for creating U128 from its u64 components.
+    pub fn from(h: u64, l: u64) -> Self;
```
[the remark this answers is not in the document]
adlerjohn: Fixed in 3aaa6fd

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,310 @@
...
+}
+
+impl core::ops::BitwiseOr for U128 {
+    pub fn binary_or(self, other: Self) -> Self {
+        ~U128::from(self.upper | other.upper, self.lower | other.lower)
+    }
+}
+
+impl core::ops::Shiftable for U128 {
+    pub fn lsh(self, other: u64) -> Self {
```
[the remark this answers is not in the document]
adlerjohn: Fixed in 3aaa6fd

sway-lib-std/src/u128.sw
adlerjohn: It can't be used because parsing is order-dependent. This issue has been an enormous sore spot that will hopefully be fixed soon. More info: https://github.com/FuelLabs/sway/issues/1578

sway-lib-std/src/u128.sw
adlerjohn: No it can't. #1548

sway-lib-std/src/u128.sw
adlerjohn: Fixed in 36f5ef7

sway-lib-std/src/u128.sw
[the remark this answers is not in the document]
adlerjohn: @otrho given that no other part of the stdlib needs these functions, can we defer it to a future PR?

sway-lib-std/src/u128.sw
adlerjohn: I implemented this and it works for small multiplications that fit in 64 bits, but fails for `~u64::max() * ~u64:;max()`. It looks like a variation of what @simonr0204 had originally implemented for u64*u64 multiplication: https://github.com/FuelLabs/sway/blob/ce5534d6b179203d7efc68e9a550754b3cf24d03/sway-lib-std/src/u128.sw#L120-L146. Can we defer optimizing this to a future PR?

Braqzen (dismissed):

sway-lib-std/src/u128.sw
```diff
@@ -0,0 +1,327 @@
...
+        asm(product, overflow, left: self, right: right, result_ptr: result) {
+            mul product left right;
+            move overflow of;
+            sw result_ptr overflow i0;
+            sw result_ptr product i1;
+        };
+        enable_overflow();
+        result
+    }
+}
```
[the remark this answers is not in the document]
adlerjohn: Fixed in 0602360.

Braqzen (approved):
Tuturu

sway-lib-std/src/u128.sw
otrho: Yep, sure.  I may take a look at the IR for a mul and see if there's an easyish optimisation we can employ to get the gas cost down.

otrho (approved):

sway-lib-std/src/u128.sw
adlerjohn: #1682

sway-lib-std/src/u128.sw
adlerjohn: #1667

PR #1604, opened 2022-05-19
Fix Shiftable trait to shift by u64.

adlerjohn:
`Shiftable` shouldn't be restricted to only shifting by the same type. This seems fine for primitive unsigned integers, but e.g. for a `U128` type or potentially some string type then it doesn't make sense to shift by that type.

### DIFF
--- a/sway-lib-core/src/ops.sw
+++ b/sway-lib-core/src/ops.sw
@@ -201,18 +201,18 @@ impl Mod for u8 {
 }
 
 pub trait Shiftable {
-    fn lsh(self, other: Self) -> Self;
-    fn rsh(self, other: Self) -> Self;
+    fn lsh(self, other: u64) -> Self;
+    fn rsh(self, other: u64) -> Self;
 }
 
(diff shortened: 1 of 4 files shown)


otrho (approved):

nfurfaro (approved):
LGTM. This is how I implemented Shiftable for b256 as well.

Braqzen:
Why is shifting a smaller type by `u64` not a problem?
What about overflows / underflows / confusion from user about why they can shift by a larger type?

Is there no problem because all of the integers are treated / padded to be `u64`?

otrho:
Yeah, I think so.  We can keep it simple.  Many CPU architectures will use just a `u8` as the shift amount and/or ignore the bits which aren't needed.  E.g., you only need 5 bits to shift a `u32` so only the bottom  5 bits of the shift amount might be used.

We might want to shift a `b512` or `b1024` one day, so a `u8` isn't big enough.

Braqzen (approved):

PR #1602, opened 2022-05-19
add failing test for issue

AlicanC:
This following script is failing to compile and I've added an e2e_vm_test for it:
```rs
script;

fn main() -> u64 {
    while false {

    };
    42
}
```

Error:
```
  Compiled library "core".
  Compiled library "std".
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', sway-core/src/asm_lang/virtual_ops.rs:461:57
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/panicking.rs:116:14
   2: core::panicking::panic
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/panicking.rs:48:5
   3: sway_core::asm_lang::virtual_ops::VirtualOp::successors
   4: sway_core::asm_generation::RealizedAbstractInstructionSet::allocate_registers
   5: sway_core::asm_generation::JumpOptimizedAsmSet::allocate_registers
   6: sway_core::asm_generation::from_ir::compile_ir_to_asm
   7: sway_core::ast_to_asm
   8: forc_pkg::pkg::compile
   9: forc_pkg::pkg::build
  10: forc::ops::forc_build::build
  11: forc::cli::commands::build::exec
  12: forc::cli::run_cli::{{closure}}
  13: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  14: tokio::park::thread::CachedParkThread::block_on
  15: tokio::runtime::thread_pool::ThreadPool::block_on
  16: forc::main
```

image

PR #1611, opened 2022-05-19
Fixing how the number of predecessors of an IR block is calculated

mohammadfawaz:
Fixes the issue flagged by @AlicanC in https://github.com/FuelLabs/sway/pull/1602

The symptom is an issue in the final asm generator. The reason being a missing `ret` for an implicit return in the user code. Basically, the check in https://github.com/FuelLabs/sway/blob/6eef7ab750cd3282f08b6014960cbc02afae717a/sway-core/src/optimize.rs#L337 fails, and so, we don't create a `ret`. Turns out that the number of predecessor blocks is incorrectly calculated (should be 1 instead of 0 for the example that JC provided).

### DIFF
--- a/sway-ir/src/block.rs
+++ b/sway-ir/src/block.rs
@@ -259,10 +259,26 @@ impl Block {
 #[doc(hidden)]
 impl BlockContent {
     pub(super) fn num_predecessors(&self, context: &Context) -> usize {
-        match &context.values[self.instructions[0].0].value {
-            ValueDatum::Instruction(Instruction::Phi(list)) => list.len(),
-            _ => unreachable!("First value in block instructions is not a phi."),
-        }
+        self.function.block_iter(context).fold(0, |mut acc, b| {
+            acc += b.instruction_iter(context).fold(0, |mut acc, i| {
+                acc += match context.values[i.0].value {
+                    ValueDatum::Instruction(Instruction::ConditionalBranch {
+                        true_block,
+                        false_block,
+                        ..
+                    }) => {
+                        true_block.get_label(context) == self.label
+                            || false_block.get_label(context) == self.label
+                    }
+                    ValueDatum::Instruction(Instruction::Branch(block)) => {
+                        block.get_label(context) == self.label
+                    }
+                    _ => false,
+                } as usize;
+                acc
+            });
+            acc
+        })
     }
 }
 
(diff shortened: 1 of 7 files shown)


sway-ir/src/block.rs
```diff
@@ -259,10 +259,26 @@ impl Block {
...
+                    }
+                    ValueDatum::Instruction(Instruction::Branch(block)) => {
+                        block.get_label(context) == self.label
+                    }
+                    _ => false,
+                } as usize;
+                acc
+            });
+            acc
+        })
```
otrho, line 281: I think it might be simpler to just iterate over all instructions in the function, filter for `br` and `cbr`, gather all the labels, filter for `self.label`, and take the length.  Not sure about perf complexity but it'd be easier to read.

Also, `fold` should very rarely have a mutable integer accumulator.  Instead you can just add to it.
```
iter().fold(0, |acc, i|) {
    acc + if blah { 0 } else { 1 }
})
```

otrho (approved):
Yep, it was pretty naive in hindsight to think that using the `phi` to calculate the number of predecessors would work.  :facepalm:   Thanks for working this out!

sway-ir/src/block.rs
otrho: I pushed a fix for some merge conflicts below, and then since I was already on the branch, I  refactored the above:
```rust
        self.function
            .instruction_iter(context)
            .filter(|(_block, ins_value)| match &context.values[ins_value.0].value {
                ValueDatum::Instruction(Instruction::ConditionalBranch {
                    true_block,
                    false_block,
                    ..
                }) => {
                    true_block.get_label(context) == self.label
                        || false_block.get_label(context) == self.label
                }
                ValueDatum::Instruction(Instruction::Branch(block)) => {
                    block.get_label(context) == self.label
                }
                _otherwise => false,
            })
            .count()
```
It's using `Function::instruction_iter()` which is a little shortcut I made exactly for this sort of thing.  And then it takes your match expression and puts it in a filter and then just counts the number of matches with `Iterator::count()`.  A bit simpler and no mutable accumulators.

otrho (dismissed):

sway-ir/src/block.rs
otrho: Changed my mind.  That cast is just a little too janky. :slightly_smiling_face: Have pushed my change.

mohammadfawaz: Yup this is much better. I knew you'd hate that cast... just wanted a shortcut. Anyways, thank you for making the changes and for all the tips 😄

PR #1613, opened 2022-05-19
Fix issue with how scoping is handled in the compiler.

emilyaherbert:


otrho (approved):
Thanks Emily.  I'm surprised this hasn't caused issues prior to now, TBH.

PR #1614, opened 2022-05-19
forc template implementation

### LINKED ISSUE #1380: Extending `forc init --template <source>` with support for arbitrary `git` sources (+ short-hand for Sway examples and GitHub)
## Current State

Currently, `forc init --template <source>` assumes that the specified `<source>` is a name of a sway example, and will attempt to fetch a forc project from the sway repo's `examples/` directory.

## Goal

Our goal is to not only support sway examples, but to enable 3rd-parties to also provide their own custom templates, inspired by [the `cargo generate` command](https://github.com/cargo-generate/cargo-generate).

## How

The plan is to implement this on top of `git` and our existing support for [finding forc projects within subdirectories of git repositories](https://github.com/FuelLabs/sway/blob/fe4f296ab9afaaf10e52f027e617f41e0945f214/forc-pkg/src/pkg.rs#L1104).

Curiously, even though the current `forc init --template` implementation only supports sway examples, it *is* implemented by fetching the sway repo (rather than shipping the examples in the binary) via an ad-hoc GitHub REST API approach.

Depending on GitHub's REST API is unsustainable, as it omits more general `git` support and centralises our tooling around GitHub. Further, if we were to extend it, we would likely end up duplicating existing work in our more generalised `git` support for forc dependencies.

Instead, we should remove the GitHub REST API approach in favour of more general git support. E.g.

```
forc init --template <git-repo-url> <proj-name>
```

The above assumes that a template forc project (indicated by a `Forc.toml`) exists at the root of `<git-repo-url>`.

In order to support pulling an example from the subdirectory of a repo (e.g. the `examples/` directory within the `sway` repo), we could allow for indicating the name of the template forc project to search for with something like:

```
forc init --template <git-repo-url> --template-name <template-proj-name> <proj-name>
```

This still makes the assumption that we only care about whatever commit happens to be at the HEAD of the repository. We could possibly also support specifying particular git references (e.g. `--tag "v1.0.0"`), but we could keep this simple to start by assuming HEAD.

This is of course quite verbose, but could act as the foundation for some short-hand commands.

## Short-hand

While we should avoid *depending* on github's REST API in our tooling, I think it's fine to add some high-level sugar seeing as it's by far the most common way of sharing repositories. Likewise, while we don't want to *only* 

- `forc init --template github:owner/repo proj-name` expands to:
   - `forc init --template https://github.com/owner/repo proj-name`
- `forc init --template github:owner/repo:template-name proj-name` expands to:
   - `forc init --template https://github.com/owner/repo --template-name template-name proj-name`
- `forc init --template example:counter proj-name` expands to:
   - `forc init --template github:fuellabs/sway:counter proj-name`

Very open to suggestions/improvements on the above syntax, but more wanted to show that we can add friendly support for things like github repositories or sway examples as short-hand on top of a more general git support - a lot of which we already support through our support for forc git dependencies.

## `forc init` implies template?

Perhaps we should have a dedicated command for initialising from a template? We could do something like:

- `forc new <proj-name>` - Creates a new empty project called `<proj-name>`.
- `forc init <template-source> <proj-name>` - Using `init` instead of `new` could imply we're initialising from some template, removing the need for a separate `--template` option.

By having `forc init` imply the use of a template, we could simplify the short-hand commands further too, e.g.

- `forc init example:counter proj-name`
- `forc init github:fuellabs/swayswap proj-name`

And so on.

---

cc @SilentCicero @adlerjohn @Braqzen @JoshuaBatty I've noticed lots of fuzzy discussion around `init` and `template` stuff lately so thought I'd get some thoughts down and try lay out a rough design. Thoughts?

kayagokalp:
closes #1380 

This PR introduces `forc template` and removes `forc init --template`. `forc template` currently takes the following inputs:

- `--url` : git repo URL
- `--project-name` : name of the project
- ` --template-name` : (optional) if specified provided name is searched in the repo. (for example `counter` example can be searched in [sway repo](https://github.com/FuelLabs/sway))

The current implementation fetches the HEAD of the default branch of the repo URL. 

**Example Usage** 

`forc template --project-name test_project --url https://github.com/FuelLabs/sway --template-name counter`

`forc template --project-name test_project --url https://github.com/owner/repo`

---

**Testing Steps**

`cargo run --bin forc template --project-name test_project --url https://github.com/FuelLabs/sway --template-name counter` : should result in a new folder created "test_project" with `counter` example inside it. Also, the forc.toml will be edited.

`cargo run --bin forc template --project-name test_project --url https://github.com/FuelLabs/sway` : should result in an error since there is no `Forc.toml` in the root. (Because there is no `Forc.toml` in the root and no template name is provided which means the template should be in the root directory).

I am leaving this as a draft since there might be some other desired behavior.


mitchmindtree:
I haven't looked closely at this PR just yet, but just a heads up that you'll need to add a file stub to the forc commands section of the book for the new `template` command. See [here](https://github.com/FuelLabs/sway/tree/master/scripts/mdbook-forc-documenter#adding-a-new-forc-command). The `mdbook-forc-documenter` plugin should then be able to automatically build the docs for the command at `mdbook build` time.

cc @binggh just a heads-up it looks like even though we have nice error messages when building the book locally, they're still hidden in CI: https://github.com/FuelLabs/sway/runs/6515239006. I've opened an issue https://github.com/FuelLabs/sway/issues/1617.

scripts/mdbook-forc-documenter/examples/forc_template.md
```diff
@@ -0,0 +1,16 @@
+
+## EXAMPLE:
+
+
+```console
+$ forc template --url https://github.com/owner/template/ --project_name my_example_project
```
adlerjohn, line 6:
````suggestion
```sh
forc template --url https://github.com/owner/template/ --project_name my_example_project
````

File: scripts/mdbook-forc-documenter/examples/forc_template.md [hunk 2]
```diff
@@ -0,0 +1,16 @@
...
+```
+
+The command above fetches the HEAD of the `template` repo and searchs for `Forc.toml` at the root of the fetched repo. It will fetch the repo and preapare a new `Forc.toml` with new project name. Outputs everthing to `current_dir/project_name`.
+
+
+```console
+$ forc template --url https://github.com/FuelLabs/sway/ --template_name counter --project_name my_example_project
```
adlerjohn, line 13:
````suggestion
```sh
forc template --url https://github.com/FuelLabs/sway/ --template_name counter --project_name my_example_project
````

**Review by adlerjohn — COMMENTED** [review]
lgtm, but will defer to @mitchmindtree and @JoshuaBatty

File: forc/src/ops/forc_template.rs [hunk 3]
```diff
@@ -0,0 +1,169 @@
...
+use std::fs::File;
+use std::io::{Read, Write};
+use std::path::{Path, PathBuf};
+use std::{env, fs};
+use sway_utils::constants;
+use url::Url;
+
+pub fn init(command: TemplateCommand) -> Result<()> {
+    validate_name(&command.project_name, "project name")?;
+    let template_name = match command.template_name.clone() {
```
mitchmindtree, line 18:
```suggestion
    // The name used for the temporary local repo directory used for fetching the template.
    let local_repo_name = match command.template_name.clone() {
```
Perhaps something like this is a better name for this binding as it's only used for the local git repository, and isn't necessarily the name of the template project.

forc/src/ops/forc_template.rs
```diff
@@ -0,0 +1,169 @@
...
+        &git_source.source.repo,
+        &git_source.commit_hash,
+    );
+    if !repo_path.exists() {
+        println!("  Fetching {}", git_source.to_string());
+        fetch_git(fetch_id, &template_name, &git_source)?;
+    }
+
+    let from_path = match command.template_name {
+        Some(_) => find_dir_within(&repo_path, &template_name, SWAY_GIT_TAG).ok_or_else(|| {
```
mitchmindtree, line 47:
```suggestion
        Some(ref template_name) => find_dir_within(&repo_path, template_name, SWAY_GIT_TAG).ok_or_else(|| {
```

forc/src/ops/forc_template.rs
```diff
@@ -0,0 +1,169 @@
...
+use std::path::{Path, PathBuf};
+use std::{env, fs};
+use sway_utils::constants;
+use url::Url;
+
+pub fn init(command: TemplateCommand) -> Result<()> {
+    validate_name(&command.project_name, "project name")?;
+    let template_name = match command.template_name.clone() {
+        Some(temp_name) => temp_name,
+        None => "DEFAULT_TEMP_NAME".to_string(),
```
mitchmindtree, line 20:
```suggestion
        None => format!("{}-template-source", commands.project_name),
```
While this is only used as a temporary directory name and for `.forc/git/checkouts`, it might be useful to derive a name from the project name in case we have to debug `.forc/git/` in the future.

mitchmindtree (changes requested):
@kayagokalp great stuff! Always great to see improvements / extra features with a diff containing more :red_circle: than :green_circle: :)

Also, very happy to see you managed to make sense of the existing git functions in `forc-pkg` to get this working!

I've left a few comments, otherwise it also looks like this introduces a few `println!`s. Now that your `tracing` work has landed we should use that instead.

JoshuaBatty:
Nice work @kayagokalp. Pulled down and tested and seems to work nicely. Agree with @mitchmindtree comments above, especially with shortening the cli arguments down.

kayagokalp:
@mitchmindtree @JoshuaBatty Thanks for the great comments! With current changes, the following commands are valid:

`forc template --template-name counter test_counter_project`  (Automatically fetches sway repo and searches for counter)

`forc template --url custom_url_here --template-name counter test_counter_project`  (Automatically fetches the `custom_url_here` repo and searches for counter)

forc/src/ops/forc_template.rs
```diff
@@ -15,39 +15,40 @@ use url::Url;
...
+        fetch_git(fetch_id, &local_repo_name, &git_source)?;
     }
 
     let from_path = match command.template_name {
-        Some(_) => find_dir_within(&repo_path, &template_name, SWAY_GIT_TAG).ok_or_else(|| {
+        Some(ref template_name) => find_dir_within(&repo_path, template_name, SWAY_GIT_TAG).ok_or_else(|| {
             anyhow!(
                 "failed to find a template `{}` in {}",
-                template_name,
+                local_repo_name,
```
mitchmindtree: I think we want to use `template_name` here

mitchmindtree (changes requested):
One last fix then I think we're good to go!

forc/src/ops/forc_template.rs
kayagokalp: Yep missed that one, :) thanks

mitchmindtree (approved):
Pre-emptively approving before CI complete after one-line change :)

Again nice work!

JoshuaBatty (approved):
LGTM, great work @kayagokalp !

PR #1618, opened 2022-05-20
Fix bug where `mdbook-forc-documenter` would panic if `mdbook build` called not in Sway root

mitchmindtree:
While testing, I noticed that the build would panic if you attempted to
`mdbook build` the book while in the `docs` directory. I then realised
we had a misuse of the `current_dir` function which must have slipped
through in a previous PR review. I've fixed this by adding a
`find_sway_repo_root` fn and implementing functions to find the plugins
and examples directories in terms of it.

cc @binggh this PR is against #1582 - landing this will apply the fixes there, then we can land #1582.


binggh (approved):

binggh:
LGTM - thanks for this catch!

Issue #1399, opened 2022-04-26
Investigate the new trait-based approach for storage.


mohammadfawaz:
From @sezna:

What if Storage was exposed by a trait Store? When you implement Store for a type T, you define how it can be stored in state. Then the compiler can just refer to the store() method with its syntax instead of doing lots of custom work.
e.g.:
```rust
impl Store for u64 {
   fn store(self, domain_separator: b256, key: b256) {
      asm(r1: self, r2: domain_separator, r3: key) {
         // do the storage for just a u64; too lazy to write asm rn
      }
   }
}
impl Store for b256 {
   fn store(self, domain_separator: b256, key: b256) {
      asm(r1: self, r2: domain_separator, r3: key) {
         // do the storage for just a b256; too lazy to write asm rn
      }
   }
}
```
then impling Store for more complex types becomes easier:
```rust
impl Store for Foo {
   fn store(self, domain_sep: b256, key: b256) {
     self.field_1.store(domain_sep, key);
     self.field_2.store(domain_sep, key + 1);
     self.field_3.store(domain_sep, key + 2);
    }
}
```
the above interface is just an initial thought, we could change it w/ regards to how we handle domain separators and stuff. But it could be similar to rust's Hash interface which uses a stateful hasher: https://doc.rust-lang.org/std/hash/trait.Hasher.html

This eliminates the need to ban heap types -- we instead just require that the trait Store is implemented for T (edited) 

And then storage hash maps fall out of that trivially!

We'd want to include impls of Store for all stdlib types. Since users will likely make new types out of our stdlib building blocks, it's unlikely they'd ever need to write asm. They can just do the thing I did in the Foo example above (call .store() on contained values). Additionally  (once we have a derive system) this would be derivable trivially.

The new approach would have the same syntax as it does currently, that would just use the trait under the hood:
```rust
storage {
  field_1: Foo,
}

storage.field_1 = 'yo'
```
The trait will need to be more complex than what I wrote above to handle writing and partial writes

nfurfaro:
My initial reaction is, I like it!
A few questions that immediately come to mind:
1.) Cost. In the example, storing `Foo` seems like it would be writing to storage 3 x (but maybe this has to happen anyway, regardless how we expose this to the user...)
[...]

Issue #1159, opened 2022-04-05
Handle `where` clause for struct.


mohammadfawaz:

emilyaherbert:
@mohammadfawaz  I propose that we combine this PR, #1160 , and #1161. I believe that from a design perspective, there is a solution that could tackle these three items together (or, at least, somewhat together, minus anticipated duplicated code in various parts)

They should all fall under "where clause for data structures."

mohammadfawaz:
If that's the case, then I will consolidate #1160 and #1161 into this issue.

emilyaherbert:
Blocked by #1821.

PR #1622, opened 2022-05-20
`Storable` trait and `Mapping`

mohammadfawaz:
🚨  THIS IS A DRAFT FOR EARLY FEEDBACK ONLY AND IS FAR FROM DONE 🚨 

Meant to close https://github.com/FuelLabs/sway/issues/1399 and also implement `Mapping`.

Blockers:
* [ ] https://github.com/FuelLabs/sway/issues/1159 to be able to see `impl<K, V> Mapping<K, V> where V: Storable`
* [ ] https://github.com/FuelLabs/sway/issues/1143 to be able to have `Mapping` to not need to use `K` and `V` in its data members.

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
+++ b/sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
@@ -6,8 +6,9 @@ use crate::{
     semantic_analysis::ast_node::*,
     type_engine::{insert_type, AbiName, IntegerBits},
 };
+use fuel_crypto::Hasher;
 
-mod method_application;
+pub mod method_application;
 use crate::type_engine::TypeId;
 use ast_node::declaration::CreateTypeId;
 use method_application::type_check_method_application;
@@ -86,6 +87,7 @@ impl TypedExpression {
             | Literal(_)
             | StorageAccess { .. }
             | TypeProperty { .. }
+            | GenerateB256Seed { .. }
             | VariableExpression { .. }
             | FunctionParameter
             | TupleElemAccess { .. } => false,
@@ -191,6 +193,7 @@ impl TypedExpression {
             | TypedExpressionVariant::VariableExpression { .. }
             | TypedExpressionVariant::AbiName(_)
             | TypedExpressionVariant::StorageAccess { .. }
+            | TypedExpressionVariant::GenerateB256Seed { .. }
             | TypedExpressionVariant::FunctionApplication { .. } => vec![],
         }
     }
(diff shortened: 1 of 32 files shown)


sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -516,6 +519,16 @@ impl TypedExpression {
                 },
                 span,
             ),
+            Expression::BuiltinGenerateB256Seed { span } => ok(
```
emilyaherbert, line 522: What about writing the equivalent of "GenerateB256Seed" as a function in the standard library? That way we don't have to special case it in the compiler.

emilyaherbert (changes requested):

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
mohammadfawaz: The compiler passes through the intrinsic `__generate_b256_seed` (name is WIP) all the way to codegen where the actual generation happen. I'm not sure we can do that with pure library code and without a new intrinsic, right?

otrho:
Initial thoughts: I don't mind generating the unique keys in an intrinsic, but I don't think it actually needs to go all the way to IR?  It looks like everything that is new in `from_ir.rs` could just go in `optimize.rs`, i.e., the keys are generated during IR generation, and just stored in the data section.  Is this not true?

And so I'm understanding correctly, the `Mapping` is essentially a hash map (with `insert(key, value)` and `get(key)`)
and a wrapper around the storage key management, so that the user doesn't have to think about the key?  And it would be generic if we had support in the compiler, but for now it's just `u64` -> `u64`?  I'm not really a fan of the name `Mapping`, even just `Map` or `HashMap` would be better IMO.

Otherwise, looking good! :+1:

mohammadfawaz:
> Is this not true?
So this is the part that gets ugly... `new()` needs to be inlined for this to work because before inlining, we only have a single copy of this intrinsic (by looking at the IR before inlining for example). Or are you saying we can do traversal of the AST to figure out when to increment the counter by chasing `FunctionApplication`s?

>  so that the user doesn't have to think about the key
Yup!

> And it would be generic if we had support in the compiler, but for now it's just `u64` -> `u64`?
Yeah! there are a few blockers with generics right now. Just thought I'd showcase `u64 -> u64` first to get feedback on the overall approach, though I don't expect any of this to go in before we make it generic.

> I'm not really a fan of the name `Mapping`, even just `Map` or `HashMap` would be better IMO.
Potentially yeah. `Mapping` is from Solidity really and maybe `Map` is a better option. We can bikeshed over the name in the coming days.

otrho:
> So this is the part that gets ugly... `new()` needs to be inlined for this to work because before inlining, we only have a single copy of this intrinsic

I see.  Yuck.  I would be _slightly_ more comfortable calling it `generate_uid()` since that sounds more general and low level, like something that belongs in IR. :wink:

PR #1632, opened 2022-05-22
 Fixing JSON ABI and the function selector for arrays

### LINKED ISSUE #1631: JSON ABI doesn't include custom type `T` when defined in an array `[T; n]`
Let the following Sway contract:

```Rust
contract;

struct Person {
    name: str[4]
}

abi TestContract {
    fn return_array_of_structs(param: [Person; 2]) -> [Person; 2];
}


impl TestContract for Contract {
    fn return_array_of_structs(param: [Person; 2]) -> [Person; 2] {
        param
    }
}
```

After building this contract, I would expect the JSON ABI to include the definition of that struct `Person`. But this is what is outputted:

```
[
  {
    "type": "function",
    "inputs": [
      {
        "name": "param",
        "type": "[struct Person; 2]",
        "components": null
      }
    ],
    "name": "return_array_of_structs",
    "outputs": [
      {
        "name": "",
        "type": "[struct Person; 2]",
        "components": null
      }
    ]
  }
]
```

And it doesn't tell me what's inside `Person`. This breaks the SDK because it tries to `abigen` `Person` but it doesn't have enough information.

mohammadfawaz:
Closes #1631 

* Update the JSON ABI file to handle arrays: insert their element types in the "components" section. Example:
If `[Person; 2]` is an argument to a contract method where `Person` is defined as `struct Person { name: str[4] }`, then the corresponding JSON ABI entry looks like:
```json
{
        "name": "param1",
        "type": "[struct Person; 2]",
        "components": [
          {
            "name": "__array_element",
            "type": "struct Person",
            "components": [
              {
                "name": "name",
                "type": "str[4]",
                "components": null
              }
            ]
          }
        ]
      },
```
* Update the function selector to take arrays into account. Arrays are represented as `a[<type>;<size>]`.

Both of these things should be documented in the specs. Created an issue here: https://github.com/FuelLabs/fuel-specs/issues/319


sway-core/src/type_engine.rs
```diff
@@ -75,6 +75,11 @@ impl ToJsonAbi for TypeId {
                     .map(|x| x.generate_json_abi())
                     .collect(),
             ),
+            TypeInfo::Array(type_id, _) => Some(vec![Property {
+                name: "r#array_element".to_string(),
```
mohammadfawaz, line 79: I didn't know what name to choose specifically, but I guess it doesn't matter. Also, there should only ever be a single entry in this vector of `Property`s.

adlerjohn (commented):
CI failing

otrho (approved):
Looks good.  Approving.  This includes the removal of the whitespace, right?

Issue #1628, opened 2022-05-22
First use of a generic method fixes its generic type


AlicanC:
Here, the first use of `fn write<T>(self, val: T, offset: u64)` with a `bool` sets its `T` to `bool` and prevents it from being called with a `u64` (or something else) later.
```rs
script;

use std::mem::*;
use std::intrinsics::*;
use std::assert::assert;

struct TestStruct {
    boo: bool,
    uwu: u64
}

fn main() -> bool {
    // Create a struct
    let foo = TestStruct { boo: true, uwu: 42 };
    let foo_len = size_of::<TestStruct>();
    assert(foo_len == 16);

    // Create a clone of the struct
    let buf = ~Buffer::alloc(foo_len);
    buf.write(true, 0);
    buf.write(42, size_of::<bool>());
//            ^^ This parameter was declared as type bool, but argument of type u64 was provided.
    let foo: TestStruct = buf.into_unchecked();
    assert(foo.boo == true);
    assert(foo.uwu == 42);

    true
}
```

The `Buffer` in question:
```rs
pub struct Buffer {
    ptr: Pointer,
    len: u64
}

impl Buffer {
    // ... snip ...

    /// Writes the given value into the buffer
    pub fn write<T>(self, val: T, offset: u64) {
        if is_reference_type::<T>() {
            let ptr = ~Pointer::from(val);
            let len = size_of::<T>();
            
            // We can't reference `self.write_unchecked` like this:
            // self.write_unchecked(ptr, len, offset);

            // Instead we inline
            copy(self.ptr.with_offset(offset), ptr, len);
        } else {
            let dst_ptr = self.ptr.with_offset(offset);
            asm(ptr: dst_ptr.val, val: val) {
                sw ptr val i0;
            };
        }
    }

    // ... snip ...
}
```
You can find the full code of `Buffer` (and `std::mem`) here: https://github.com/FuelLabs/sway/blob/522bf95b7f06e679161712ae4a0821c74f318a2c/sway-lib-std/src/mem.sw

PR #1633, opened 2022-05-22
Improve spans in error messages around generic types

### LINKED ISSUE #1401: Improve spans for unresolved generics
In some cases, the spans we associate with the idents for generic types are at the declaration site, when it'd be better if they were at the call site. See the pictures in #1400 as an example. We'd want those errors to point at the calling of `foo()` in `fn main()` in that PR. 

This will most likely involve adding a `Span` to `Ref` types in the type engine. Currently, when a type `Ref`s to another type, the span we use is the type being referred to. The referrer loses its span info. Adding an (optional?) `Span` to `Ref` could give us this flexibility.

sezna:
This PR adds a `Span` to every `TypeInfo::Ref` to improve the error UX. Currently, if a type refers to another type via `Ref`, we use that underlying type as the span for errors and warnings. However, in the case of inference or monomorphized types, we want the call site to be shown. This PR uses spans from the call site preferentially and falls back to declaration spans if there is no `Ref` type preceding.

Tl;dr, before, the span pointed to the decl:
image

After:
image

This helps me a lot with #1628 and #1143, which is why I went ahead and knocked it out. 

Closes #1401

### DIFF
--- a/sway-core/src/type_engine.rs
+++ b/sway-core/src/type_engine.rs
@@ -162,8 +162,8 @@ fn chain_of_refs() {
     let sp = Span::dummy();
     // numerics
     let id = engine.insert_type(TypeInfo::Numeric);
-    let id2 = engine.insert_type(TypeInfo::Ref(id));
-    let id3 = engine.insert_type(TypeInfo::Ref(id));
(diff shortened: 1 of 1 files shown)


mohammadfawaz (approved):
I like it. Looks good!

PR #1634, opened 2022-05-22
Prevent method parameters from overly inferring

### LINKED ISSUE #1628: First use of a generic method fixes its generic type

sezna:
Closes #1628 

It turns out we forgot to call `.monomorphize()` on method applications in general, so only functions have been getting monomorphized. This PR fixes that.

edit: Some other things are falling out of this PR, putting it into a draft while I do some more construction. 🚧


sway-core/src/semantic_analysis/ast_node/expression/typed_expression/method_application.rs
```diff
@@ -201,6 +239,10 @@ pub(crate) fn type_check_method_application(
 
     // type check all of the arguments against the parameters in the method declaration
     for (arg, param) in args_buf.iter().zip(method.parameters.iter()) {
+        /*
+        let mut param = param.clone();
+        param.r#type = insert_type(look_up_type_id(param.r#type));
+        */
```
mohammadfawaz, line 245: Commented code.

mohammadfawaz: Oups sorry!.. just noticed that this is a draft. Shouldn't have commented.

emilyaherbert:
@sezna  I recommend that you wait for #1592 to go in, because this diff will be super simplified. It should reduce most of these changes to just this. This call to `monomorphize`:
https://github.com/FuelLabs/sway/blob/885a2bad873b2ebb43d613f3498157a8e98cf5bc/sway-core/src/semantic_analysis/ast_node/expression/func_app_instantiation.rs#L27

Add that call here:
https://github.com/FuelLabs/sway/blob/885a2bad873b2ebb43d613f3498157a8e98cf5bc/sway-core/src/semantic_analysis/ast_node/expression/typed_expression/method_application.rs#L291

And also here:
https://github.com/FuelLabs/sway/blob/885a2bad873b2ebb43d613f3498157a8e98cf5bc/sway-core/src/semantic_analysis/ast_node/expression/typed_expression/method_application.rs#L364

This new `monomorphize` function is mutually recursive with `resolve_type_{with|without}_name` eliminating cases in which we had been not-fully-monomorphizing.

In fact, my newest PR for if-lets merges the ides of method applications and function applications:
https://github.com/FuelLabs/sway/blob/c5b20419b6bbb7644f905cb56aa7f1bd0c73fde0/sway-core/src/semantic_analysis/ast_node/expression/typed_expression/method_application.rs#L197

PR #1640, opened 2022-05-23
Compile predicates in IR

### LINKED ISSUE #1641: Compile predicates in IR
Predicates are subsets of scripts, but are stubbed out in IR currently. We actually have everything we need to compile them, though.

sezna:
Predicates are the same as scripts, with the unique characteristic that the type system checks that the main function returns a boolean. There are also some opcode checks at the end. IR doesn't need to know about any of this, so we can reuse `compile_script`, as predicates are essentially a subset of scripts with some special properties.

Closes #1641

### DIFF
--- a/sway-core/src/optimize.rs
+++ b/sway-core/src/optimize.rs
@@ -24,13 +24,15 @@ pub(crate) fn compile_ast(ast: TypedParseTree) -> Result<Context, CompileError> 
             main_function,
             declarations,
             all_nodes: _,
-        } => compile_script(&mut ctx, main_function, &namespace, declarations),
-        TypedParseTree::Predicate {
-            namespace: _,
-            main_function: _,
-            declarations: _,
+        }
+        | TypedParseTree::Predicate {
+            namespace,
+            main_function,
+            declarations,
             all_nodes: _,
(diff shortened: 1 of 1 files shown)


adlerjohn (approved):
utACK

sway-core/src/optimize.rs
```diff
@@ -24,13 +24,15 @@ pub(crate) fn compile_ast(ast: TypedParseTree) -> Result<Context, CompileError>
...
+        }
+        | TypedParseTree::Predicate {
+            namespace,
+            main_function,
+            declarations,
             all_nodes: _,
-        } => unimplemented!("compile predicate to ir"),
+            // predicates and scripts have the same codegen, their only difference is static
+            // type-check time checks.
+        } => compile_script(&mut ctx, main_function, &namespace, declarations),
```
mitchmindtree, line 35: Could be nice to name `compile_script` to something that indicates it is expected to handle both, like `compile_script_or_predicate` or `compile_executable`, but I won't block landing this because of it :)

mitchmindtree (approved):

otrho (approved):

PR #1646, opened 2022-05-23
Add `break` and `continue` keywords

### LINKED ISSUE #587: `break` and `continue` keywords
`break` should abort a while loop, and `continue` should jump back to evaluating the condition for another loop. Those jump labels already exist, so this issue involves new keywords and jumping back to those jump labels.

I do _not_ think we should be able to `break` on a value like in Rust. That adds a lot of type-checking complexity for not a lot of benefit, IMO. Maybe someday...but just vanilla `break` and `continue` should be good for now.

emilyaherbert:
- Closes #587


otrho:
This might be a bit tricky to convert to IR, let me know if you want guidance or even for me to co-author.

emilyaherbert:
Closing because this branch is significantly out-of-date.

PR #1647, opened 2022-05-23
Add booleans to the keywords in the parser

emilyaherbert:
Simple PR to add booleans as a keyword in the parser


otrho:
This looks a bit out of place.  `true` and `false` aren't intrinsic functions really.

How does this improve things though?  Maybe we need the concept of predefined axiomatic values in the compiler, perhaps slightly higher level than as intrinsics, but I don't quite know why.

adlerjohn:
+1 on it's not immediately obvious why this is needed. Also weird that they're not prefixed with `__` (which might hint that they shouldn't be intrinsics).

emilyaherbert:
The word "intrinsics" is often used as a short-hand way of referring to "intrinsic functions." I believe that, philosophically, most compilers also contain "intrinsic expressions". For example, in the case of Sway, unsigned ints are lifted to the compiler level inside of `TypeInfo::Str` and `TypeInfo::Array`, thus making those unsigned ints intrinsic expressions.

In the context of the parser, I was mistakenly under the impression that `Intrinsic` meant """ a special case that has intrinsic value to the compiler """. If at parse time `Intrinsic` is meant to represent just intrinsic _functions_, then I recommend that we change the name to `IntrinsicFunctionName` or add doc strings to this type.

I think that the `"true"` and `"false"` tokens are hooks into compile-time knowledge of the conversion from the `"true"` token to `true` and from the `"false"` token to `false` (without which they would be parsed as variables). Introducing `IntrinsicFunctionName` would allow us to introduce `IntrinsicExpression`, which would just contain `"true"` and `"false"`.

But I think I am just splitting hairs about this now 😄 

>  it's not immediately obvious why this is needed

This change is used in the following PR https://github.com/FuelLabs/sway/pull/1646, but I broke it into its own PR for readability.

> Could you speak to how this follows those conventions and what makes it future-proofing?

I don't know specifically how Rust and LLVM handle the conversion between the `"true"` token to `true` and from the `"false"` token to `false`. But it terms of future-proofing, this change makes it a little more clear where the `"true"` and `"false"` tokens are parsed.

mohammadfawaz:
`true` and `false` are really just keywords: https://doc.rust-lang.org/reference/keywords.html

emilyaherbert:
Works for me @mohammadfawaz !

mohammadfawaz:
@otrho are you okay with moving towards making `true` and `false` keywords as Emily has done is her latest commit?

otrho (approved):
Oh, right, I didn't notice this latest change, sorry!  `true` and `false` should definitely be keywords, not variable names.

Issue #1580, opened 2022-05-18
Update E2E test harness to verify specific failures.


otrho:
I think we've talked about this but I can't find a specific issue tracking it -- the `should_fail` tests should only 'pass' when the failures (errors and/or warnings) are specifically what we expect.  At the moment the projects could fail to build for any reason and the tests would still pass.

mohammadfawaz:
Slightly related https://github.com/FuelLabs/sway/issues/723

But absolutely.. this would be extremely helpful. I'd like to be able to write one negative test that checks for many error messages. Preferably, those checks would live inside the test itself and not in `mod.rs`. (e.g. from Clang: https://github.com/llvm/llvm-project/blob/47258ffc5c7aa9ae2017bdde58a88e0f928d8f07/clang/test/Parser/brackets.cpp#L10`

PR #1658, opened 2022-05-24
Add missing dependency in 2 test cases to supress unnecessary warnings

emilyaherbert:
This was generating a bunch of additional warnings about how it can't find `core::ops::eq` and I forgot to add the dependency in the test cases where failure is expected.


otrho (approved):
Argh, we need #1580. :slightly_smiling_face:

adlerjohn (approved):

PR #1660, opened 2022-05-25
Parsing precedence fixes

canndrew:
This PR fixes two places in the parser where precedence was wrong, or at least didn't match what rust does.

Firstly, brace-postfixed expressions (ie. `{ ... }`/`if x { .. }`/`match x { ... }`/`while x { ... }`)  have low precedence if they appear at the start of a statement. This means that is no longer accepted:

```
fn foo() -> u32 {
    if true { 1 } else { 2 } + 3
}
```

but this still is:

```
fn foo() {
    let bar = if true { 1 } else { 2 } + 3;
}
```

Secondly, comparison operators can longer be chained together. This means that `x < y < z` is now a parse error rather than being parsed as `(x < y) < z`.


adlerjohn:
Is this one ready for review @canndrew?

PR #1662, opened 2022-05-25
Updated welcome message colors

### LINKED ISSUE #1659: `forc init` needs more coloured output
There's a little bit of green here but we can make this look much better with more colours!

I think this is a _critical priority_

Braqzen:
Closes #1659 

New colors

### DIFF
--- a/forc-util/src/lib.rs
+++ b/forc-util/src/lib.rs
@@ -207,12 +207,12 @@ pub fn println_green(txt: &str) {
     println_std_out(txt, TermColor::Green);
 }
 
-pub fn println_light_blue(txt: &str) {
+pub fn print_light_blue(txt: &str) {
     let stdout = StandardStream::stdout(ColorChoice::Always);
     print_with_color(txt, TermColor::Rgb(40, 200, 255), stdout);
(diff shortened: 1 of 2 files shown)


forc-util/src/lib.rs
```diff
@@ -207,6 +207,16 @@ pub fn println_green(txt: &str) {
     println_std_out(txt, TermColor::Green);
 }
 
+pub fn println_light_blue(txt: &str) {
```
kayagokalp, line 210: Should we name this as `print_light_blue` rather than `println_light_blue` as it might be confusing later on?

forc-util/src/lib.rs
```diff
@@ -207,6 +207,16 @@ pub fn println_green(txt: &str) {
...
+    let stdout = StandardStream::stdout(ColorChoice::Always);
+    print_with_color(txt, TermColor::Rgb(40, 200, 255), stdout);
+}
+
+pub fn println_light_green(txt: &str) {
```
kayagokalp, line 215: Same thing with the `println_light_blue`

Issue #1526, opened 2022-05-12
Windows Support (tracking issue)


mitchmindtree:
Opening this as a more general response to @ra0x3's [question](https://github.com/FuelLabs/sway/pull/1513/files/f3f1ef72e6bb345dcadf7dc1b90373e96510d794#r870688423) in #1513.

## Current State

Windows support for Sway and its tooling is largely untested.

**If you are a Windows user, you are currently recommended to use WSL**. That is, at least until we start testing and publishing Windows binaries.

## Moving Forward

If we wish to support Windows we should do the following for each of the `fuel-core`, `fuels-rs` and `sway` repos and their crates:

- [ ] **Check all path handling** and that no code makes the assumption we're using unix-style paths. Primary culprits are the use of concatentating strings or using `format!` to construct paths, e.g.. `format!("{}/{}/{}", foo, bar, baz)`. Windows does not support "foo/bar/baz" style paths. Instead, `Path::join` or `PathBuf::collect` should be used.
      Another example: until [#1213](https://github.com/FuelLabs/sway/pull/1213), Sway directly appended the path that you see in `dep foo/bar/baz;` verbatim, meaning that nested modules could never be successfully compiled on Windows.
- [ ] **Check that all I/O heavy dependencies are Windows-friendly**. Well implemented crates will give compile errors on platforms they don't support, but that's not always the case.
- [ ] **Add a CI job** that at least compiles and runs some basic tests on Windows.
- [ ] **Update docs** w.r.t. any windows-specific needs in regards to system dependencies or setup.

Only once we've addressed these should we consider publishing and distributing Windows native binaries. Otherwise for the most part, Rust and the wider ecosystem makes it easy to write cross-platform code.

### Related

- #1191.
- Old issue #413.

Voxelot:
Needs to be broken down into repo-level issues.

PR #1669, opened 2022-05-25
Refactor forc init + add forc new

ra0x3:
### Description
- This PR changes `forc init` to `forc new`, and updates `forc init` functionality to be more similar to `cargo init`

### Testing steps
- [ ] Try pulling down the PR and creating a `forc init` project with combinations of the different flags
----

Read the Docs:
- Sway Book: https://fuellabs.github.io/sway/latest
- Rust SDK Book: https://fuellabs.github.io/fuels-rs/latest
- TypeScript SDK: https://github.com/FuelLabs/fuels-ts

Join the Community:
- Follow us @SwayLang: https://twitter.com/SwayLang
- Ask questions in dev-chat on Discord: https://discord.com/invite/xfpK4Pe

Report Bugs:
- Sway Issues: https://github.com/FuelLabs/sway/issues/new add


docs/src/SUMMARY.md
```diff
@@ -44,6 +44,7 @@
     - [forc clean](./forc/commands/forc_clean.md)
     - [forc completions](./forc/commands/forc_completions.md)
     - [forc deploy](./forc/commands/forc_deploy.md)
+    - [forc new](./forc/commands/forc_new.md)
```
adlerjohn, line 47: Should this be alphabetical?

forc/src/cli/commands/init.rs
```diff
@@ -5,20 +5,27 @@ use clap::Parser;
...
-    /// Adding this flag creates an empty library program.
+    /// Create a package with a library target (src/lib.sw).
     #[clap(long)]
     pub library: bool,
-    /// The name of your project
-    pub project_name: String,
+    /// Create a package with a contract target (src/contract.rs).
+    #[clap(long)]
+    pub contract: bool,
+    /// Create a package with a contract target (src/predicate.rs).
```
mitchmindtree, line 20:
```suggestion
    /// Create a package with a predicate target (src/predicate.rs).
```

forc/src/cli/commands/init.rs
```diff
@@ -5,20 +5,27 @@ use clap::Parser;
 /// Create a new Forc project.
 #[derive(Debug, Parser)]
 pub struct Command {
-    /// The default program type, excluding all flags or adding this flag creates a basic contract program.
+    /// The path at which to create the manifest
```
mitchmindtree, line 8:
```suggestion
    /// The directory in which the forc project will be initialized.
```

forc/src/cli/commands/init.rs
```diff
@@ -5,20 +5,27 @@ use clap::Parser;
 /// Create a new Forc project.
```
mitchmindtree, line 5:
```suggestion
/// Create a new Forc project in an existing directory.
```

forc/src/cli/commands/init.rs
```diff
@@ -5,20 +5,27 @@ use clap::Parser;
...
 #[derive(Debug, Parser)]
 pub struct Command {
-    /// The default program type, excluding all flags or adding this flag creates a basic contract program.
+    /// The path at which to create the manifest
     #[clap(long)]
-    pub contract: bool,
-    /// Adding this flag creates an empty script program.
+    pub path: Option<String>,
+    /// Create a package with a binary target (src/main.sw). This is the default behavior
```
mitchmindtree, line 11:
```suggestion
    /// Create a package with a script target (src/main.sw).
```

I believe our default program type should be the `contract` program type. It looks like this is correct under the `forc new` CLI command, but not here.

forc/src/cli/commands/new.rs
```diff
@@ -0,0 +1,35 @@
+use crate::ops::forc_new;
+use anyhow::Result;
+use clap::Parser;
+
+const TEMPLATE_HELP: &str = r#"Initialize a new project from a template.
+
+Example Templates: 
+ - counter"#;
+
+/// Create a new Forc project.
```
mitchmindtree, line 10:
```suggestion
/// Create a new Forc project at <path>.
```

forc/src/cli/commands/new.rs
```diff
@@ -0,0 +1,35 @@
...
+    #[clap(long)]
+    pub script: bool,
+    /// Adding this flag creates an empty predicate program.
+    #[clap(long)]
+    pub predicate: bool,
+    /// Adding this flag creates an empty library program.
+    #[clap(long)]
+    pub library: bool,
+    /// The name of your project
+    pub project_name: String,
```
mitchmindtree, line 29: Let's rename this field to `path` and update the comment. E.g. the following should be valid:

```
forc new my-project
```
```
forc new foo/bar/my-project
```

The name of the project should be taken from the final directory in the path (i.e. `my-project` in the above examples).

If we want to match Rust's behaviour, we could also include a `--name <NAME>` option that allows the user to specify a forc project name that doesn't necessarily match the final directory name. We can leave that for a future PR if you'd prefer though.

forc/src/cli/commands/new.rs
```diff
@@ -0,0 +1,35 @@
...
+
+Example Templates: 
+ - counter"#;
+
+/// Create a new Forc project.
+#[derive(Debug, Parser)]
+pub struct Command {
+    /// Initialize a new project from a template
+    #[clap(short, long, help = TEMPLATE_HELP)]
+    pub template: Option<String>,
```
mitchmindtree, line 15: The `template` option can be removed as #1614 has introduced a dedicated command for templates.

forc/src/ops/forc_new.rs
```diff
@@ -0,0 +1,190 @@
...
+struct GithubTree {
+    mode: String,
+    // We only care about the "path" which are files / directory names
+    path: String,
+    sha: String,
+    size: Option<usize>,
+    #[serde(rename = "type")]
+    data_type: String,
+    url: String,
+}
```
mitchmindtree, line 69: We should be able to remove all types and any logic related to GitHub requests/responses in this module after removing the unnecessary `--template` option. The need to add `#[allow(dead_code)]` here is a good indicator these are no longer needed :)

mitchmindtree (changes requested):
Great to see this happening!

---

I think we can remove a *lot* of duplicated code/logic between `forc new` and `forc init`.

We can think of
```
forc new <path>
```
very roughly as a tiny short-hand wrapper around:
```
mkdir -p <path>
forc init --path <path>
```

Let's refactor the `forc new` logic here to take advantage of the `forc init` functionality to make this easier to maintain and keep behaviour consistent.

adlerjohn:
Converting to draft both because will a few comments to address, and so as not to merge it during the hackathon by accident since it's a breaking change.

docs/src/introduction/standard_library.md
```diff
@@ -4,13 +4,7 @@ Similar to Rust, Sway comes with its own standard library.
...
 
-The entire Sway standard library is a Forc project called `std`, and is available directly here: <https://github.com/FuelLabs/sway/tree/master/sway-lib-std> (navigate to the appropriate tagged release if the latest `master` is not compatible).
-
-## Using the Standard Library
-
-The standard library is made implicitly available to all Forc projects created using [`forc init`](../forc/commands/forc_init.md). In other words, it is not required to manually specify `std` as an explicit dependency. Forc will automagically use the version of `std` that matches its version.
-
-Importing items from the standard library can be done using the `use` keyword, just as importing items from any Sway project. For example:
+The standard library is made implicitly available to all Forc projects created using [`forc new`](../forc/commands/forc_new.md). Importing items from the standard library can be done using the `use` keyword. Example:
```
mitchmindtree, line 7: Was removing this section intentional?

forc-util/src/restricted.rs
```diff
@@ -62,7 +62,7 @@ pub fn contains_invalid_char(name: &str, use_case: &str) -> Result<()> {
         }
     }
     for ch in chars {
-        if !(unicode_xid::UnicodeXID::is_xid_continue(ch) || ch == '-') {
+        if !(unicode_xid::UnicodeXID::is_xid_continue(ch) || ch == '-' || ch == '/') {
```
mitchmindtree, line 65: This function checks for valid package names, and package names shouldn't contain `'/'` so I don't think we want this.

forc/src/cli/commands/init.rs
```diff
@@ -2,23 +2,30 @@ use crate::ops::forc_init;
...
-    /// Adding this flag creates an empty script program.
+    pub path: Option<String>,
+    /// Create a package with a script target (src/main.sw).
     #[clap(long)]
     pub script: bool,
-    /// Adding this flag creates an empty predicate program.
-    #[clap(long)]
-    pub predicate: bool,
-    /// Adding this flag creates an empty library program.
+    /// Create a package with a library target (src/lib.sw). This is the default behavior.
```
mitchmindtree, line 14: The default behaviour creates a contract, not a library.

forc/src/ops/forc_init.rs
```diff
@@ -1,97 +1,65 @@
...
-    sha: String,
-    size: Option<usize>,
-    #[serde(rename = "type")]
-    data_type: String,
-    url: String,
-}
+pub fn init(command: InitCommand) -> Result<()> {
+    let project_dir = match &command.path {
+        Some(p) => PathBuf::from(p),
+        None => std::env::current_dir().unwrap(),
```
mitchmindtree, line 50:
```suggestion
        None => std::env::current_dir().context("failed to get current directory for forc init")?,
```

^ this might require importing `anyhow::Context`.

forc/src/ops/forc_init.rs
```diff
@@ -1,97 +1,65 @@
...
-
-    print_light_blue("Report Bugs:\n");
-    print_light_green("- Sway Issues: ");
-    print_light_blue("https://github.com/FuelLabs/sway/issues/new\n");
-}
+    let project_name = command.name.unwrap_or_else(|| {
+        project_dir
+            .to_str()
+            .unwrap()
+            .split('/')
```
mitchmindtree, line 57: This assumes Unix paths, which we should avoid whenever trivial to do so (particularly for #1526).

To retrieve the project name in the case one was not given, we should instead do something like:

```rust
project_dir.file_stem().context("failed to infer project name from directory name")?
```

https://doc.rust-lang.org/std/path/struct.Path.html#method.file_stem

forc/src/ops/forc_init.rs
```diff
@@ -1,97 +1,65 @@
...
-    size: Option<usize>,
-    #[serde(rename = "type")]
-    data_type: String,
-    url: String,
-}
+pub fn init(command: InitCommand) -> Result<()> {
+    let project_dir = match &command.path {
+        Some(p) => PathBuf::from(p),
+        None => std::env::current_dir().unwrap(),
+    };
```
mitchmindtree, line 52: Let's do some checks on this directory here before going on:

1. Check the directory exists. If not, return an error.
2. Check that it does not already contain a `Forc.toml` for an existing project to avoid accidentally overriding an existing project. If it does, return an error.

forc/src/ops/forc_init.rs
```diff
@@ -106,82 +74,92 @@ pub fn init(command: InitCommand) -> Result<()> {
...
     // Make a new directory for the project
-    fs::create_dir_all(Path::new(&project_name).join("src"))?;
+    fs::create_dir_all(Path::new(&project_dir).join("src"))?;
+
+    if command.verbose {
+        println_light_blue(&format!(
+            "\nUsing project directory at {}",
+            canonicalize(&project_dir)?
+        ))
+    }
```
mitchmindtree, line 89: Perhaps we should move this output earlier, right after we retrieve `project_dir` and check that it's valid?

forc/src/ops/forc_init.rs
```diff
@@ -106,82 +74,92 @@ pub fn init(command: InitCommand) -> Result<()> {
...
-    )?;
+    let harness_path = Path::new(&project_dir).join("tests").join("harness.rs");
+    fs::write(&harness_path, defaults::default_test_program(&project_name))?;
+
+    if command.verbose {
+        println_light_blue(&format!(
+            "\nCreating test harness at {}",
+            canonicalize(&harness_path)?
+        ));
+    }
```
mitchmindtree, line 148: Maybe we should move this earlier to before we write to `harness_path` above?

forc/src/ops/forc_init.rs
```diff
@@ -106,82 +74,92 @@ pub fn init(command: InitCommand) -> Result<()> {
...
-    )?;
+    let gitignore_path = Path::new(&project_dir).join(".gitignore");
+    fs::write(&gitignore_path, defaults::default_gitignore())?;
+
+    if command.verbose {
+        println_light_blue(&format!(
+            "\nCreating .gitignore at {}",
+            canonicalize(&gitignore_path)?
+        ));
+    }
```
mitchmindtree, line 159: Maybe we should move this earlier to before we write to `gitignore_path` above?

forc/src/ops/forc_init.rs
```diff
@@ -1,97 +1,65 @@
...
 
-#[allow(dead_code)]
-#[derive(serde::Deserialize)]
-struct GithubRepoResponse {
-    sha: String,
-    url: String,
-    // We only care about the tree here
-    tree: Vec<GithubTree>,
-    truncated: bool,
+pub fn canonicalize(path: &PathBuf) -> Result<String> {
```
mitchmindtree, line 37: It looks like this `canonicalize` function is intended to be a wrapper around `Path::canonicalize` but returns a `String` in order to more easily be "print"able.

Instead of introducing this function, consider using the `Path::display` method as necessary, e.g. the following is valid:

```rust
println!("Created {}", path.canonicalize()?.display());
```

mitchmindtree (changes requested):

docs/src/introduction/standard_library.md
ra0x3: Bad merge. Will fix

forc-util/src/restricted.rs
ra0x3: @mitchmindtree Is that not what you mentioned [here](https://github.com/FuelLabs/sway/pull/1669#pullrequestreview-987011928) and [here](https://github.com/FuelLabs/sway/pull/1669#discussion_r883256924)?

forc/src/ops/forc_init.rs
ra0x3: `.file_stem()` Nice 👍

forc/src/ops/forc_init.rs
ra0x3: @mitchmindtree Super minor this. I guess I can change to past tense (i.e., `Created`), since I think putting this rather ugly `println_light_blue` in between the `let harness_path` and the`fs::write` makes this a little harder to read

forc/src/ops/forc_init.rs
ra0x3: Same as comment above regarding past tense

forc-util/src/restricted.rs
mitchmindtree: In [this comment](https://github.com/FuelLabs/sway/pull/1669#discussion_r883256924) my intention was to suggest that the new command should support full paths (i.e. `forc new foo/bar/proj-name`) rather than only supporting a project name (i.e. `forc new proj-name`).

That suggestion is only related to this function in the sense that the default project name used for `forc new` is retrieved from the directory name. In the command `forc new foo/bar/baz`, the inferred project name is not `foo/bar/baz`, but rather `baz` (taken from the directory name, not the full path). As a result, there should be no need for this project name validation function to support `/` characters.

Hopefully this makes a bit more sense?

forc/src/ops/forc_init.rs
mitchmindtree: Yeah using past tense instead SGTM.

forc-util/src/restricted.rs
ra0x3: @mitchmindtree Ah, I see. Change incoming

forc/src/ops/forc_init.rs
```diff
@@ -34,29 +34,38 @@ fn print_welcome_message() {
...
+        None => {
+            std::env::current_dir().context("Failed to get current directory for forc init.")?
+        }
     };
 
+    if !project_dir.is_dir() {
+        anyhow::bail!("'{}' is not a valid directory.", project_dir.display());
+    }
+
+    if Path::exists(&Path::new(&project_dir).join("Forc.toml")) {
```
mitchmindtree, line 49:
```suggestion
    if project_dir.join(sway_utils::constants::MANIFEST_FILE_NAME).exists() {
```

Hot-tip: `PathBuf` automatically dereferences to `Path` (just like `String` derefs to `str`, or `Vec<T>` derefs to `[T]`), so there's no need to explicitly create a `Path` here.

forc/src/ops/forc_init.rs
```diff
@@ -34,29 +34,38 @@ fn print_welcome_message() {
...
-            .to_str()
-            .unwrap()
-            .split('/')
-            .last()
+            .file_stem()
+            .context("Failed to infer project name from directory name.")
             .unwrap()
+            .to_string_lossy()
             .to_string()
     });
```
mitchmindtree, line 70: We should return the error here rather than panicking, e.g. something like this:

```rust
let project_name = command.name.ok_or_else(|| {
    let dir_name = project_dir
        .file_stem()
        .context("Failed to infer project name from directory name.")?
        .to_string_lossy();
    Ok(dir_name)
});

**Review by mitchmindtree — CHANGES_REQUESTED** [review]
Just a couple more small things, then I'll test locally and we should be good to go!

**Review by JoshuaBatty — APPROVED** [review]
Nice work. Pulled down and tested locally. All looks good.

File: forc/src/ops/forc_init.rs [hunk 22]
```diff
@@ -60,14 +60,20 @@ pub fn init(command: InitCommand) -> Result<()> {
...
+            let dir = project_dir
+                .file_stem()
+                .context("Failed to infer project name from directory name.");
+
+            match dir {
+                Ok(d) => d.to_string_lossy().to_string(),
+                Err(e) => anyhow::bail!("Failed to infer get current directory name: '{}'", e),
+            }
+        }
+    };
```
mitchmindtree, line 76: It looks like we have some double-up in the error message:
```
"Failed to infer get current directory name: 'Failed to infer project name from directory name.'),
```
Good point about the `Cow` btw! I forgot `to_string_lossy()` returns a `Cow` and not a `String`.

How about something like this?
```rust
let project_name = match command.name {
    Some(name) => name,
    None => {
        project_dir
            .file_stem()
            .context("Failed to infer project name from directory name.")?
            .to_string_lossy()
            .into_owned()
    }
};
```

mitchmindtree (changes requested):
Just opened #1841 to cut down on some of the duped logic.

forc-util/src/lib.rs
```diff
@@ -207,6 +207,20 @@ pub fn println_green(txt: &str) {
...
+
+pub fn print_light_blue(txt: &str) {
+    let stdout = StandardStream::stdout(ColorChoice::Always);
+    print_with_color(txt, TermColor::Rgb(40, 200, 255), stdout);
+}
+
+pub fn print_light_green(txt: &str) {
+    let stdout = StandardStream::stdout(ColorChoice::Always);
+    print_with_color(txt, TermColor::Rgb(90, 255, 40), stdout);
+}
```
mitchmindtree, line 222: It was recently discussed in #tooling on slack that we should switch from using light colours in the `forc init` / `forc new` output back to using default `tracing` log output as the output is difficult to read on some terminals and lighter themes.

Could you:

1. Remove these new functions.
2. Change the verbose output sections to use `tracing::debug!` instead.
3. Remove the `--verbose` output flags from `forc new` and `forc init` commands in favour of allowing users to specify the level of verbosity using the `RUST_LOG` environment variable (already provided by the `tracing` subscriber).

We can possibly revisit colouring output in a future dedicated PR, however when we do we should be more comprehensive in testing different terminals and colour themes (light, dark, etc) before moving ahead with it.

mitchmindtree (changes requested):

forc-util/src/lib.rs
adlerjohn: This one is still unresolved @ra0x3

ra0x3: @adlerjohn Doh! Forgot to push up the commit 🤦

forc/src/ops/forc_init.rs
```diff
@@ -97,7 +35,38 @@ fn print_welcome_message() {
...
+        None => {
+            std::env::current_dir().context("Failed to get current directory for forc init.")?
+        }
+    };
+
+    if !project_dir.is_dir() {
+        anyhow::bail!("'{}' is not a valid directory.", project_dir.display());
+    }
+
+    if project_dir.join("Forc.toml").exists() {
```
adlerjohn, line 49: Use the constant `constants::MANIFEST_FILE_NAME` here and below instead of `"Forc.toml"`.

PR #1670, opened 2022-05-25
Improve grammar for type argument error

sezna:
The pluralization when there was only 1 argument was bothering me...


mohammadfawaz (approved):
Nice.

PR #1671, opened 2022-05-25
Monomorphize method applications

### LINKED ISSUE #1628: First use of a generic method fixes its generic type

sezna:
We were not monomorphizing method applications at all. This PR fixes that. This change was made much easier by #1589  and #1666 

Closes #1628 

This PR works by monomorphizing declarations when methods are instantiated against them.  This now means that you can annotate types for a parent type and a function, like Rust.

So, this example works:

```rust

struct Foo<T> {
  a: T
}
impl <T> Foo {
   fn bar<O>(o: O) -> u64 {
     5
  }
}
fn main() {
      // note the annotation `u64` for `Foo` and `bool` for `bar()`
    Foo::<u64>::bar::<bool>(true)
}

```

This involved changes to the parser to allow generic arguments in the new position and calling `monomorphize` on declarations when methods are instantiated. All the other changes are just fallout from those two things.

edit: some interaction with recent PRs has broken some tests, moving to draft while that gets sorted out...


AlicanC (commented):
I tried this on my local while working on the Multicall script and it indeed solves my issues! Can do things a lot more conveniently now 👍 👍 👍 👍

sway-core/src/convert_parse_tree.rs
```diff
@@ -1238,53 +1238,60 @@ fn expr_to_expression(ec: &mut ErrorContext, expr: Expr) -> Result<Expression, E
...
+                match suffix.pop() {
+                    Some((_double_colon_token, call_path_suffix)) => match suffix.pop() {
+                        Some((_double_colon_token, maybe_method_segment)) => {
+                            let PathExprSegment {
+                                fully_qualified,
+                                name,
+                                generics_opt,
+                            } = maybe_method_segment;
+                            if let Some((_double_colon_token, generic_args)) = generics_opt {
+                                println!("here?1");
```
otrho, line 1251:
```suggestion
```

otrho (changes requested):
LGTM, though I'm not the type checker expert.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression/method_application.rs
```diff
@@ -63,6 +64,21 @@ pub(crate) fn type_check_method_application(
         None
     };
 
+    // monomorphize the parent type
```
emilyaherbert, line 67: remove

sway-core/src/semantic_analysis/ast_node/expression/typed_expression/method_application.rs
```diff
@@ -375,10 +391,64 @@ fn find_method(
         }
         (ty, true) => ty.clone(),
     };
+    let ty = (|| -> CompileResult<TypeId> {
```
emilyaherbert, line 394: Is it necessary to use a function closure here? This makes it harder to read and maintain IMO.

test/src/e2e_vm_tests/test_programs/should_pass/language/generic_impl_self/src/main.sw
```diff
@@ -54,7 +54,7 @@ impl DoubleIdentity<u8, u8> {
 }
 
 fn double_identity2<T, F>(x: T, y: F) -> DoubleIdentity<T, F> {
-  ~DoubleIdentity::new::<T, F>(x, y)
+  ~DoubleIdentity::<T, F>::new(x, y)
```
emilyaherbert, line 57: Can you explain the reason for changing the syntax?

emilyaherbert (changes requested):

emilyaherbert:
Based on our convo @sezna, I'm going to close this PR, copy your parser changes to a new PR, then take over https://github.com/FuelLabs/sway/issues/1628.

PR #1676, opened 2022-05-26
tiny perf gain by bailing early when finding methods for an error recovery type

sezna:
I noticed that we iterate through every implemented trait and method if the type is in error recovery, which is guaranteed to fail. This is just a small performance boost for the error recovery case.


mohammadfawaz (dismissed):
LGTM! Can this change be done in other places as well? e.g. in `get_methods_for_type_by_trait`?

mohammadfawaz (approved):

PR #1677, opened 2022-05-26
Remove collecting operators as tokens in language server

JoshuaBatty:
The language server was collecting operators such as `+` `-` `/` and `*`. However, these were getting stored as `plus` `divide` etc.. leading to an `attempt to subtract with overflow` bug when calculating the character length of the collected token.

This PR removes the collection of operators which fixes a bug where the language server was crashing in the `swayswap` and `liquidity_pool` examples.


sway-lsp/src/core/token.rs
```diff
@@ -366,9 +366,26 @@ fn handle_expression(exp: Expression, tokens: &mut Vec<Token>) {
...
+                            tokens.push(token);
+                        }
+                    }
+                }
+                MethodName::FromModule { method_name, .. } => {
+                    let ident = method_name.clone();
+                    let token = Token::from_ident(&ident, TokenType::MethodApplication);
+                    tokens.push(token);
+                }
+            };
```
mitchmindtree, line 388: It looks like the idea behind this PR is to only collect the token in the case that this method call is not one of the methods that are desugared from a short-hand operator like `+`, `/`, `!`, etc. However, it looks like this change might also unintentionally stop collecting all `FromType` method names that don't have at least two prefixes in their call path? I think it might also omit all non-`core` packages that happen to have an `ops` module too?

Maybe we want something like this instead?

```rust
// Check if the given method is a `core::ops` application desugared from short-hand syntax like / + * - etc.
fn desugared_op(method_name: &MethodName) -> bool {
    if let MethodName::FromType { ref call_path, .. } = method_name {
        let prefix0 = call_path.prefixes.get(0).map(|ident| ident.as_str());
        let prefix1 = call_path.prefixes.get(1).map(|ident| ident.as_str());
        if let (Some("core"), Some("ops")) = (prefix0, prefix1) {
            return true;
        }
    }
    false
}

// Don't collect applications of desugared operators due to mismatched ident lengths. 
if !desugared_op(&method_name) {
    let ident = method_name.easy_name();
    let token = Token::from_ident(&ident, TokenType::MethodApplication);
    tokens.push(token);
}
```

Let me know if I'm missing the point :joy:

JoshuaBatty: No, that's exactly the point! :) 

Thanks, this solution is definitely the preferred method! Will update this now.

mitchmindtree (approved):
My intuition is that there's probably a better solution to this than just not collecting the tokens at all (e.g. should we have a match that goes from `"divide" => /,` etc instead?), but happy to see this land for now if it helps to avoid LSP `panic!`ing during app dev.

adlerjohn (approved):
utACK

Issue #1472, opened 2022-05-04
Publish `forc-gm` once the crate name is available


mohammadfawaz:
#1471 stops `forc-gm` from being published. Once the name `forc-gm` is available on `crates.io`, revert #1471

PR #1680, opened 2022-05-26
Remove `forc-gm` in the docs for now

binggh:
Let's remove this section first until we get the crate name back or if we have an alternative solution.


SilentCicero:
Instead of removing, i'd just switch it to the `forc-explorer` or something for now. It's a good section! Just not valid atm.

mitchmindtree:
Related https://github.com/FuelLabs/sway/issues/1472

binggh:
> Instead of removing, i'd just switch it to the `forc-explorer` or something for now. It's a good section! Just not valid atm.

wanted to remove it asap in case of a malicious crate, I can do a proper write-up for an alternative plugin in a separate PR?

mitchmindtree (approved):

mitchmindtree (approved):

PR #1684, opened 2022-05-26
Introduce FromTrait method name

sezna:
This is another PR coming out of #1671 

I noticed that we have two modes of method name, `FromType` and `FromModule`. In the `FromType` case, a provided type is optional as is a span. So, it looks like this:

```
enum MethodName { 
  FromType { type_info: Option<TypeInfo>, type_span: Option<Span>, method_name: ... }
}
```

Those two options either must _both_ be defined or _neither_ should be defined. This means that our type system is not defending us from the case of _one_ being undefined. 

To remove this invalid state from the type, I introduced a third variant, which is `FromTrait`. That allows you to reference a trait function directly from the module it was implemented in. This is what the behavior was previously if those two `Option`s were `None`, but now it is made explicit with a third variant. This variant is primarily used for `core::ops::` stuff.

Also, when I ran the test suite, a lockfile got updated. I went ahead and included that as it needs to be done anyway...can put in a separate PR if y'all feel that's better, though.

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/expression/typed_expression/method_application.rs
+++ b/sway-core/src/semantic_analysis/ast_node/expression/typed_expression/method_application.rs
@@ -286,7 +286,7 @@ pub(crate) fn resolve_method_name(
                 type_arguments,
                 namespace,
                 &arguments,
-                &call_path,
+                call_path,
                 self_type
             ),
             return err(warnings, errors),
@@ -305,7 +305,7 @@ pub(crate) fn resolve_method_name(
                     type_arguments,
                     namespace,
                     &arguments,
-                    &call_path,
+                    call_path,
                     self_type
                 ),
                 return err(warnings, errors),
@@ -380,5 +380,5 @@ fn find_method(
     } else {
         namespace.find_module_path(call_path.full_path())
     };
-    namespace.find_method_for_type(insert_type(ty), &abs_path, self_type, &arguments)
+    namespace.find_method_for_type(insert_type(ty), &abs_path, self_type, arguments)
 }


adlerjohn (commented):
CI failing

sway-core/src/parse_tree/expression/method_name.rs
```diff
@@ -8,18 +8,27 @@ pub enum MethodName {
...
+        type_name: TypeInfo,
+        type_name_span: Span,
     },
     /// Represents a method lookup that does not contain any types in the path
+    /// something like a.b(c)
+    /// in this case, the first argument defines where to look for the method
     FromModule { method_name: Ident },
+    /// something like a::b::c()
+    /// in this case, the path defines where the fn symbol is defined
+    /// used for things like core::ops::add(a, b).
```
emilyaherbert, line 20: From my understanding, it is not possible to call `core::ops::add(a, b)` in user-facing Sway. Can you either add more explanation about how `core::ops::add` is not user-facing code, or provide an example of user-facing Sway code that would trigger this case?

emilyaherbert (changes requested):
If `FromType` means "a method that is implemented _on a type_", then I think we should rename from `FromModule` to be `FromObject` meaning "a method that is implemented _on an object_ (which also has a type)".

So, given this Sway code:

```rust
struct Data {
  x: u64
}

impl Data {
  fn new(v: u64) -> Data {
    Data { x: v }
  }

  fn value(self) -> u64 {
    self.x
  }
}
```

Then calling `let d = ~Data::new(0)` would be a `FromType` method, and calling `d.value()` would be a `FromObject` method.

Given a different example `1 + 2`, this resolves to `1.add(2)`, which can be considered "a method _from a trait_ that is implemented _on an object_ (which also has a type)". When I think of things from this perspective, I feel that `FromObject` should be modified to include a `call_path` and `FromTrait` should be dropped in favor of using `FromObject`.

This would then allow us to cover this case:

```rust
struct Data {
  x: u64
}

trait One {
  fn one() -> u64;
}

impl One for Data {
  fn one() -> u64 {
    1
  }
}

trait Value {
  fn value(self) -> u64;
}

impl Value for Data {
  fn value(self) -> u64 {
    self.x
  }
}

impl Data {
  fn new(v: u64) -> Data {
    Data { x: v }
  }
}
```

`let d = ~Data::new(0)` and `~Data::one()` would resolve to a `FromType` method, and `d.value()` would resolve to a `FromObject` method.

emilyaherbert:
> As for the naming, `FromModule` means that the method should exist in the immediate surrounding module, i.e. `a.b(c)` should mean there's a `fn b()` for the type of `a` in the current module

Ah I see. Just to be clear---given a case in which 1) the module with the decl, 2) the module with the decl's impl, and 3) the module with the call `a.b(c)`, this would fall into `FromModule`?

> That sounds reasonable for a long term goal, and I'd be happy to bikeshed it in an issue, but are you happy with this as a way of improving type safety for the way things currently are?

After this one clarification, SGTM!

sway-core/src/parse_tree/expression/method_name.rs
emilyaherbert: 👍

emilyaherbert:
> So if they're all separated, the impls must be imported accordingly.

This is what I meant! I just wanted some clarification on what this would look like.

I'll approve!

emilyaherbert (approved):

PR #1686, opened 2022-05-26
On Forc.toml change, forc should try to update only package nodes that appear in the lock::Diff

### LINKED ISSUE #900: On `Forc.toml` change, `forc` should try to update only package nodes that appear in the `lock::Diff`
## Problem

Currently, `forc`'s approach to updating the lock file in response to changes in the manifest is pretty simple. It:

1. Loads the `BuildPlan` from the `Forc.lock`.
2. Validates the `Lock` by retrieving all dependencies of the project node from the graph into a set, and compares them to the set of dependencies loaded by the manifest.
3. If the sets don't match, recreate the `Lock` from scratch.

Step 3 involves fetching each package's source in order to determine the current head for the specified git branch or tag. While this is simple in implementation and should always result in a correct package graph, this can result in a couple issues:

- We end up doing more fetching than is necessary if only one or a subset of dependencies were changed.
- We may unintentionally update the pinned HEAD of deps unrelated to the dep that was added/removed/edited in the manifest.

## Solution

To solve this, we should improve the strategy for handling detected differences between the `Forc.lock` and `Forc.toml`. Steps 1 and 2 should remain the same, but the following should be something closer to:

3. If the sets don't match, determine the packages that have been removed, added or edited.
4. Handle each change accordingly:
    - For removed packages, remove the node and all its dependencies that aren't used elsewhere in the graph. We can do so by removing nodes in breadth-first-search order to make sure parents are always removed before children. If a visited node still has an edge to a parent, then it means it has a dependent elsewhere in the graph and should be skipped.
    - For added packages, `fetch_deps` only for that sub-graph and merge it into the main graph.
    - For edited packages, perform a "remove" of the old node and an "add" as above.

## Offline

The changes above should also allow us to achieve `offline` updates to the lock file in the case that packages have only been *removed*. However, we'll still require fetching repositories for package edits or additions. After landing the above changes, we could open another issue around potentially building a local index of known dependencies so that, in the case that `--offline` is specified and we need to edit or add nodes to a lock file, we can lookup our local index in a best-effort approach that doesn't require fetching from remotes.

kayagokalp:
~~I will be simplifying and refactoring the code a little bit more so leaving this as a draft for now.~~ 

closes #900.

## About the PR
Before this PR, forc was re-fetching all the dependencies once there is a mismatch between the lock file and the manifest.
 
For more context about the implementation please take a look into the discussion at the #900 and particularly [this](https://github.com/FuelLabs/sway/issues/900#issuecomment-1138644198) comment.

### Basic Overview of What is Happening

1. Removed and added dependencies are detected
2. Removed dependencies handled by doing a BFS from the project node. Searching for detected removed dependencies and children of the removed dependencies (given that no other node depends on them) in the order of the breadth-first traversal.
3. Added dependencies handled by first connecting the added dependency to the root node and then recursively fetching_children for the dependencies of the newly added dependency while checking if any dependency of the newly added dependency is already fetched.

### To test:

1. `forc init --library child`
2. `forc init parent`
3. `forc build` in the parent directory to build it before depending on the child.
4.  add the "child" library as a dependency for the "parent"
5. `forc build` in the "parent" directory

This test procedure will be resulting in the following dependency graph:

after_step_5


forc-pkg/src/pkg.rs
```diff
@@ -905,7 +1029,7 @@ fn dep_to_source(pkg_path: &Path, dep: &Dependency) -> Result<Source> {
         Dependency::Detailed(ref det) => match (&det.path, &det.version, &det.git) {
             (Some(relative_path), _, _) => {
                 let path = pkg_path.join(relative_path);
-                Source::Path(path)
+                Source::Path(fs::canonicalize(path)?)
```
kayagokalp, line 1032: We can insert relative paths to the `Forc.toml` in the background they are also stored like that meaning the following:

Adding std as a dependency from `context_testing_abi` would result in  a path as the following:

`/Users/kayagokalp/sway/test/src/e2e_vm_tests/test_programs/should_pass/require_contract_deployment/caller_context_test/../../test_abis/context_testing_abi/../../../../../../../sway-lib-std`

which is basically 

`/Users/kayagokalp/sway/sway-lib-std`

Before this change, this was detected as a change which is actually not so I converted these paths to absolute paths before storing them as a pkg field.

kayagokalp:
There are still some unnecessary calls to functions and duplicate code pieces, will be cleaning up and marking this as ready to review 🎉

kayagokalp:
Realized a bug while removing the packages. The `BuildPlan` creation from `lock` fails and `BuildPlan` is always created from `new` once there is a package removal. Converting the PR into draft until that is solved.

forc-pkg/src/pkg.rs
```diff
@@ -193,6 +193,148 @@ impl BuildPlan {
...
+            .compilation_order
+            .last()
+            .ok_or_else(|| anyhow!("Invalid Graph"))?;
+        let proj_id = graph[proj_node].id();
+        let (added, removed) =
+            self.get_difference_with_manifest(proj_node, manifest_file, sway_git_tag)?;
+
+        use petgraph::visit::{Bfs, Walker};
+        //find root_node_id
+        let root_node_id = *self.compilation_order().last().unwrap();
```
mitchmindtree, line 270: This is the same as `proj_node` that we retrieved earlier.

forc-pkg/src/pkg.rs
```diff
@@ -193,6 +193,148 @@ impl BuildPlan {
...
+            // to remove a either it will be in removed list or it needs to have 0 incoming edge.
+            if graph.edges_directed(node, Direction::Incoming).count() == 0
+                || removed
+                    .clone()
+                    .into_iter()
+                    .any(|removed_dep| removed_dep.name == graph[node].name)
+            {
+                graph.remove_node(node);
+            }
+        }
```
mitchmindtree, line 287: I was thinking on package removal a little more, and I think it might be more efficient and a bit cleaner if we start by removing the edges between the root and the removed nodes. I'm thinking something like this:

```rust
/// Remove the given set of packages from `graph` along with any dependencies that are no
/// longer required as a result.
fn remove_deps(
    graph: &mut Graph,
    path_map: &PathMap,
    proj_node: NodeIx,
    to_remove: &BTreeSet<Pkg>,
) {
    // Find the edges between the root and the removed packages.
    let edges_to_remove: Vec<_> = graph
        .edges_directed(proj_node, Direction::Outgoing)
        .filter_map(|e| {
            let dep_pkg = graph[e.target()].unpinned(path_map);
            if to_remove.contains(&dep_pkg) {
                Some(e.id())
            } else {
                None
            }
        })
        .collect();

    // Remove the edges.
    for e in edges_to_remove {
        graph.remove_edge(e);
    }

    // Do a BFS from the root and remove all nodes that are no longer connected to the root.
    let mut bfs = Bfs::new(graph, proj_node);
    bfs.next(graph); // Skip the root node (aka project node).
    while let Some(node) = bfs.next(graph) {
        if graph.edges_directed(node, Direction::Incoming).next().is_none() {
            graph.remove_node(node);
        }
    }
}
```

forc-pkg/src/pkg.rs
```diff
@@ -193,6 +193,148 @@ impl BuildPlan {
...
+
+        let mut visited_map: HashMap<Pinned, NodeIx> = HashMap::new();
+        // After removing some nodes the index of project node can change.
```
mitchmindtree, line 290: Ahhh that's a great point - sorry I should have mentioned this quirk with the `petgraph::Graph` type earlier!

A much easier way of working around this (rather than recalculating the compilation order and reconstructing the path map, etc) is to use petgraph's `StableGraph` type rather than its regular `Graph` type.

The benefit of the `StableGraph` type over the `Graph` type is that you can remove nodes and edges without invalidating the indices of the remaining nodes and edges. When adding new nodes or edges, it will re-use free slots before extending the internal node or edge vectors. You can read a little more about `StableGraph` [here](https://docs.rs/petgraph/latest/petgraph/stable_graph/struct.StableGraph.html).

Knowing this, perhaps we should change our `Graph` type to:
```rust
pub type Graph = petgraph::stable_graph::StableGraph<Node, Edge, Directed, GraphIx>;
```
*This might require enabling a stable graph feature, I don't quite remember.*

Let me know once you've done this and simplified the following code as a result and I'll do another review :)

mitchmindtree (commented):
Great work @kayagokalp! I've left a few comments but for the most part I think this is on the right track.

As a side-note: looks like there's a lot of duplicated code between `BuildPlan::validate` and `BuildPlan::get_difference_with_manifest` :thinking: Seeing as the only role of `validate` was originally to check if the package sets match, perhaps we can just rename `validate` to `diff_manifest` and alter it to return the difference in packages?

forc-pkg/src/pkg.rs
```diff
@@ -557,6 +706,7 @@ pub fn graph_to_path_map(
                     .ok_or_else(|| anyhow!("missing path info for dependency: {}", dep.name))?;
                 let path = parent_path.join(rel_dep_path);
                 if !path.exists() {
+                    println!("{:?}", path);
```
mitchmindtree, line 709:
```suggestion
```

forc/src/ops/forc_build.rs
```diff
@@ -125,3 +127,19 @@ pub fn build(command: BuildCommand) -> Result<pkg::Compiled> {
 
     Ok(compiled)
 }
+
+fn create_new_lock(
+    plan: &pkg::BuildPlan,
+    old_lock: &Lock,
+    manifest: &ManifestFile,
+    lock_path: &PathBuf,
```
mitchmindtree, line 135:
```suggestion
    lock_path: &Path,
```

forc/src/ops/forc_build.rs
```diff
@@ -69,18 +69,20 @@ pub fn build(command: BuildCommand) -> Result<pkg::Compiled> {
...
         let plan = pkg::BuildPlan::new(&manifest, SWAY_GIT_TAG, offline)?;
-        let lock = Lock::from_graph(plan.graph());
-        let diff = lock.diff(&old_lock);
-        lock::print_diff(&manifest.project.name, &diff);
-        let string = toml::ser::to_string_pretty(&lock)
-            .map_err(|e| anyhow!("failed to serialize lock file: {}", e))?;
-        fs::write(&lock_path, &string).map_err(|e| anyhow!("failed to write lock file: {}", e))?;
-        info!("   Created new lock file at {}", lock_path.display());
+        info!("  Creating a new `Forc.lock` file. (Cause: {})", cause);
+        create_new_lock(&plan, &old_lock, &manifest, &lock_path)?;
```
mitchmindtree, line 74: Rather than creating the lock file here and then potentially a second time after checking the diff, perhaps we can instead keep track of the new lock file cause and create it in one place? E.g. something like this:

```rust
    let mut new_lock_cause = None;
    let mut plan = plan_result.or_else(|e| -> Result<pkgs::BuildPlan> {
        if locked {
            bail!(
                "The lock file {} needs to be updated but --locked was passed to prevent this.",
                lock_path.to_string_lossy()
            );
        }
        new_lock_cause = if e.to_string().contains("No such file or directory") {
            Some(anyhow!("lock file did not exist"))
        } else {
            Some(e)
        };
        let plan = pkg::BuildPlan::new(&manifest, SWAY_GIT_TAG, offline)?;
        Ok(plan, cause)
    });

    let diff = plan.validate(&manifest, SWAY_GIT_TAG)?;
    if !diff.added_is_empty() || !diff.removed.is_empty() {
        plan = plan.apply_new_manifest(diff, SWAY_GIT_TAG, offline)?;
        new_lock_cause = Some(anyhow!("lock file did not match manifest `[dependencies]`"));
    }

    if let Some(cause) = new_lock_cause {
        info!("  Creating a new `Forc.lock` file. (Cause: {})", cause);
        create_new_lock(&plan, &old_lock, &manifest, &lock_path)?;
        info!("   Created new lock file at {}", lock_path.display());
    }
```

forc/src/ops/forc_build.rs
```diff
@@ -42,7 +42,10 @@ pub fn build(command: BuildCommand) -> Result<pkg::Compiled> {
     let lock_path = lock_path(manifest.dir());
 
     // Load the build plan from the lock file.
-    let plan_result = pkg::BuildPlan::from_lock_file(&lock_path, SWAY_GIT_TAG);
+    // This BuildPlan would be unsafe (true as passed as unsafe_plan) meaning
+    // there might be some pinned dependency whose parent does not include itself,
+    // i.e: a removed dependency. For more information, please see https://github.com/FuelLabs/sway/issues/1778
+    let plan_result = pkg::BuildPlan::from_lock_file(&lock_path, SWAY_GIT_TAG, true);
```
mitchmindtree, line 48: I can see where you're coming from with this change w.r.t. attempting to fix #1778, but I don't think we should enable the creation of a potentially unsafe build plan. We should always uphold all invariants for each of our types to avoid ever ending up in a position where we're unsure if they're valid or not, otherwise our types lose meaning.

Would you mind removing the `unsafe_plan` changes from this PR and we can discuss the solution in #1778?

kayagokalp: That makes perfect sense, but without fixing #1778, `apply_new_manifest` will not be able to handle removed dependencies as we wanted (once there is a removal, BuildPlan will get generated from scratch). Should we discuss and solve it in #1778 and merge it here or once this PR is ready should we merge this one and merge solution for the #1778 later?

mitchmindtree (approved):
Looking great to me! Awesome work @kayagokalp :)

JoshuaBatty (approved):
Yeah this is really great @kayagokalp, nice work

PR #1687, opened 2022-05-26
Move unnecessary usage of SDK for tests to e2e tests

adlerjohn:
Tests that weren't actually using the SDK for anything don't need to be in the SDK harness.


nfurfaro:
This is the opposite direction that I had been heading in, which was to centralize all stdlib tests under the stdlib test harness, basically grouping tests by **_what_** we're testing.
This is grouping them by _**how**_ we're testing instead.
I'm not attached to the previous direction, so if this has concrete benefits (ie: resolving weird dependency spaghet) then I'm on board.

adlerjohn:
@nfurfaro the issue right now is that we're depending on the SDK for tests, and the SDK is depending on `forc` for tests. This creates a circular dependency issue. Minimizing unnecessary use of the SDK here means minimizing any potential issues with relying on particular functionality from the SDK.

adlerjohn:
poking @nfurfaro for final review

nfurfaro (approved):

PR #1688, opened 2022-05-26
New intrinsic for generating a unique `b256` key

mohammadfawaz:
This is a simple intrinsic that hands out `b256` keys. This is useful for mappings as shown in https://github.com/FuelLabs/sway/pull/1622 as mappings need a seed key that is hashed with user keys when calling `insert()` and `get()`.

The intrinsic needs to be in IR because it can only be resolved _after_ function inlining, that is, after the IR pipeline.

Long term, I imagine these intrinsics to be in their own category, both in semantic analysis and in IR. So, that should be cleaned up eventually.


sway-core/src/optimize.rs
```diff
@@ -2518,7 +2529,7 @@ mod tests {
         let output = path_converter.replace_all(output.as_str(), "$1/path/to/$2");
 
         if output != expected {
-            tracing::error!("{}", prettydiff::diff_lines(&expected, &output));
+            println!("{}", prettydiff::diff_lines(&expected, &output));
```
mohammadfawaz, line 2532: @kayagokalp I've had to do this here similarly to https://github.com/FuelLabs/sway/pull/1624 to be able to see the actual diffs when I was building the tests.

kayagokalp: I think we should go like that 👍. I will look into places we had to come back to `println!` and think about what we can do about those places (or whether we should do something for them)

sway-core/src/asm_generation/from_ir.rs
```diff
@@ -2300,7 +2329,7 @@ mod tests {
 
         let asm_script = format!("{}", asm);
         if asm_script != expected {
-            tracing::error!("{}", prettydiff::diff_lines(&expected, &asm_script));
+            print!("{}", prettydiff::diff_lines(&expected, &asm_script));
```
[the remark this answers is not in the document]
kayagokalp: I think the tracing subscriber is not initialized at the time of this line getting executed. This happened with the testing suite before (#1624), so `tracing::error` is not printed out, for forc-* parts of the repo this is not a problem since every part of the forc initializes the tracing subscriber but I think it is not clear where to initialize the subscriber for places like test-suite and sway-core so that is missed. I am opening up an issue for correcting this and will work on this quickly.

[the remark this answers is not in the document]
mohammadfawaz: That being said, I'm not sure why I changed it to `print!` instead of `println!` 😆 .. I'll change it.

sway-ir/src/parser.rs
```diff
@@ -351,7 +357,7 @@ mod ir_builder {
                     *c.as_bytes().get(0).unwrap()
                 }
                 / "\\x" h:hex_digit() l:hex_digit() {
-                    (dbg!(h) << 4) | dbg!(l)
```
otrho, line 354: Ooh!

otrho (approved):
I prefer this name, thanks. :slightly_smiling_face:

These UIDs are obviously deterministic and only unique per compile unit (of which we only have one ATM) but as the compiler becomes more modular we'll need to keep in mind that be may get collisions one day.

PR #1700, opened 2022-05-26
Remove sdk-harness script lib.

adlerjohn:
Follow-up of #1687.

Also removes unneeded Forc project in root of SDK harness.


nfurfaro (approved):
UTACK.

PR #1703, opened 2022-05-26
Remove fuels-signers from SDK harness.

adlerjohn:


nfurfaro (approved):
assuming CI passes, looks good.

PR #1704, opened 2022-05-26
Remove rand from SDK harness.

adlerjohn:


nfurfaro (approved):
LG. Can always add back if actually needed again.

Issue #1074, opened 2022-02-14
Add module to the stdlib with wrapper functions for common (all?) opcodes


nfurfaro:
Like FuelLabs/sway-lib-std#38 , the goal here would be to minimize the usage of assembly needed while building out the stdlib (and other libraries) by providing utilities for calling opcodes in a simpler way (in terms of writing , reviewing and reading/maintaining).
wrapper functions could be chained together to achieve whatever is needed for the current task at hand.
Idea by @otrho .

- [ ] extract overflow FLAG wrapper functions  `enable_overfow()` and `disable_overflow()` from `U128` type module.

mohammadfawaz:
I'm thinking we should start by transforming the implementation of all core ops into intrinsics (instead of asm blocks) that become IR instructions such as `add`, `mul`, etc. These would make writing optimization passes like "instruction combining" so much easier.

vaivaswatha:
> I'm thinking we should start by transforming the implementation of all core ops into intrinsics (instead of asm blocks) that become IR instructions such as `add`, `mul`, etc. These would make writing optimization passes like "instruction combining" so much easier.

https://github.com/FuelLabs/sway/pull/1948#issue-1268611016

mohammadfawaz:
I'm going to close this now given that we have such wrappers in the form of intrinsics for most of the useful opcodes. We are now adding new intrinsics as the need arises.

PR #1707, opened 2022-05-27
Implement Add for b256

nfurfaro:
poor performance due, `OutOfGas` trying to increment anything over 1000,000,000,000,000,000
Next iteration will use `U128::overflowing_add()`

- New implementation uses `overflowing_add` modelled on the similar function in `std::U128`
- I had to make `compose` / `decompose` functions public for this, but opened a related issue to clean that up here: #1718 
- ideally, `disable_overflow()` and `enable_overflow()` would be extracted into a central location where they can be reused by other modules. This can be tackled in another PR, and is already tracked as a task in https://github.com/FuelLabs/sway/issues/1074


sway-lib-core/src/ops.sw
```diff
@@ -659,7 +656,7 @@ fn get_word_from_b256(val: b256, offset: u64) -> u64 {
 }
 
 /// Build a single b256 value from 4 64 bit words.
-fn compose(word_1: u64, word_2: u64, word_3: u64, word_4: u64) -> b256 {
+pub fn compose(word_1: u64, word_2: u64, word_3: u64, word_4: u64) -> b256 {
```
adlerjohn, line 659: We're not really pressed for time for this feature, since it can be worked around with hash onion. I would prefer _not_ adding things that aren't ops for primitive types to `core::ops`, especially not at the public interface.

sway-lib-std/src/u128.sw
```diff
@@ -49,7 +49,7 @@ impl core::ops::Ord for U128 {
 // impl core::ops::OrdEq for U128 {
 // }
 
-fn disable_overflow() {
+pub fn disable_overflow() {
```
adlerjohn, line 52: Refactor these into another modular in a separate PR, that can be merged into `master` first.

adlerjohn (commented):
We're not really in a hurry for this feature, so let's do it cleanly instead of introducing tech debt.

sway-lib-core/src/ops.sw
nfurfaro: fair enough. I linked an issue in the description which will clean this up.

sway-lib-std/src/u128.sw
nfurfaro: yep, will do as mentioned in the description. going to convert this to a draft and land some cleanup PRs first then.

sway-lib-core/src/ops.sw
nfurfaro: This is now being addressed via the addition of a new `core-utils` lib which can be used by both `core` and `std`, while keeping `compose` and `decompose` out of the public API of `std`.

nfurfaro:
@otrho It looks like you had pushed a commit to this branch but I steamrolled it when I force pushed.
That said, I think @mitchmindtree 's recent PR will fix this, CI will need to wait till the next forc release and should then pass.

otrho:
> @otrho It looks like you had pushed a commit to this branch

Yeah, I pushed something which resolved the lock file issue but it was still failing CI for other reasons, so :+1: , all good.

otrho (approved):
Man, this would've been much easier to review if all the `forc.lock` stuff was in a different PR!

nfurfaro:
>Man, this would've been much easier to review if all the forc.lock stuff was in a different PR!

Sorry about that :( 
I'm not sure I could have done that though, as the only reason I had to nuke all the forc locks was because of the real changes in this PR (adding the new dependency).
This mess should be unnecessary in the future after this is fixed:

PR #1716, opened 2022-05-27
Add a full recursive check of the AST for purity mismatches.

### LINKED ISSUE #1151: Validate impurity and that no storage opcodes are generated in a pure context

### LINKED ISSUE #1183: Warn on unnecessary storage annotations
Example:
```rust
contract;

abi MyContract {
    #[storage(write)] // Or any other storage annotation
    fn foo() -> u64;
}

    #[storage(write)] // Or any other storage annotation
    fn foo() -> u64 {
        0
    }
}
```
should emit a warning saying that the annotations are not necessary.

otrho:
In particular, locate storage ASM operations and confirm the enclosing function has the correct `storage` attributes.

I spent/wasted a couple of days trying to make this more efficient by annotating the AST as we build it so we don't need to search it like this does, but I simply couldn't make it work in a non-hacky way with our current data structures.  Perhaps the proposed big refactor we have planned might allow it to happen, dunno.

Closes #1151.
Closes #1183.

### Update:

This has taken me five different attempts:
1. Recurse through the typed tree gathering errors.  Was almost working but needed data structure tweaks to get the errors correct.
1. Simplify by iterating over IR instructions from the compiled program.  Was ugly and didn't fit -- this is a semantic check which belongs more in the front end.
2. Try again but with the untyped parse tree which is a less complex data structure.  Was unable to follow function calls because they aren't inlined yet.
3. Back to 1) with the typed tree, but found that parts of the tree e.g., impl methods, are still untyped parse tree nodes, so would require recursion code for both types of tree, which seemed ridiculous.
4. Back to 2) but from within the front-end as a semantic check, but performed after IR is compiled. i.e., the purity check itself is not in the IR library but back in `sway-core`.

### Also...

~~Two remaining issues: I had to allow storage ops in libraries as mentioned in #1151.  I've also disabled a couple of E2E tests because they try to run contracts which have storage ops, so similarly to the library issue, it complains about storage ops in a script.~~

~~These restrictions need to be loosened somehow.~~

Also the formatter can't handle attributes.

### DIFF
--- a/examples/counter/src/main.sw
+++ b/examples/counter/src/main.sw
@@ -1,8 +1,8 @@
 contract;
 
 abi TestContract {
-    fn initialize_counter(value: u64) -> u64;
-    fn increment_counter(amount: u64) -> u64;
+    #[storage(write)]fn initialize_counter(value: u64) -> u64;
+    #[storage(read, write)]fn increment_counter(amount: u64) -> u64;
 }
 
 storage {
@@ -10,12 +10,12 @@ storage {
 }
 
 impl TestContract for Contract {
-    fn initialize_counter(value: u64) -> u64 {
+    #[storage(write)]fn initialize_counter(value: u64) -> u64 {
         storage.counter = value;
         value
     }
 
-    fn increment_counter(amount: u64) -> u64 {
+    #[storage(read, write)]fn increment_counter(amount: u64) -> u64 {
         let incremented = storage.counter + amount;
         storage.counter = incremented;
         incremented
--- a/examples/storage_example/src/main.sw
+++ b/examples/storage_example/src/main.sw
@@ -3,18 +3,18 @@ contract;
 use std::storage::{get, store};
 
 abi StorageExample {
-    fn store_something(amount: u64);
-    fn get_something() -> u64;
+    #[storage(write)]fn store_something(amount: u64);
+    #[storage(read)]fn get_something() -> u64;
 }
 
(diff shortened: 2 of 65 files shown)


mohammadfawaz:
> I've also disabled a couple of E2E tests because they try to run contracts which have storage ops, so similarly to the library issue, it complains about storage ops in a script.

Hmmm the scripts themselves shouldn't really have any storage instructions right? They simply call ABI methods that themselves do the actual work. Whether the implementations of those methods access storage or not is not something the script should be concerned about. Is the issue here that the ABI interface itself has storage annotations and those are visible in the script because the ABI has to be imported?

otrho:
Yeah, the problem is, the compiler gets one thing to compile, a script, contract, library or predicate and then everything is inlined.  This code is checking for storage and finding it.  At the moment it can't tell where a function call is going to, whether it's into a library or contract, or whether it's local.

Should be fixed with the refactor (@emilyaherbert's plan, what are we calling it?) which should provide us with the explicit modularity we can check.

otrho:
This code is making it so the storage purity 'colourisation' is ubiquitous -- if some code somewhere could ever use storage eventually then it must be annotated.  So scripts are inevitably going to need to have annotations.  I think it's obviously fair to block the `storage` decls themselves from non contract source files but annotations should be allowed anywhere.

mohammadfawaz (approved):
LGTM! Everything here is reasonable.

PR #1732, opened 2022-05-27
Change any instance of `if let .. { .. } else { .. }` in the standard library to use match expressions

emilyaherbert:
This PR changes any instance of a `if let .. { .. } else { .. }` in the standard library to use match expressions. This is done for readability, maintainability, and to make it future-proof.

To be clear, this change is not necessary, nor will cause any user-facing changes to the standard library. This is simple a refactoring to improve code quality---the `_` branch now clearly exists visually, for if it needs to be used.

### DIFF
--- a/sway-lib-std/src/option.sw
+++ b/sway-lib-std/src/option.sw
@@ -29,19 +29,17 @@ impl<T> Option<T> {
 
     /// Returns `true` if the result is [`Some`].
     fn is_some(self) -> bool {
-        if let Option::Some(t) = self {
-            true
-        } else {
-            false
+        match self {
+            Option::Some(_) => true,
+            _ => false,
         }
     }
 
     /// Returns `true` if the result is [`None`].
     fn is_none(self) -> bool {
-        if let Option::Some(t) = self {
-            false
-        } else {
-            true
+        match self {
+            Option::Some(_) => false,
+            _ => true,
         }
     }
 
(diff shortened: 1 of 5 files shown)


sway-lib-std/src/option.sw
```diff
@@ -51,10 +49,11 @@ impl<T> Option<T> {
     /// Instead, prefer to use pattern matching and handle the [`None`]
     /// case explicitly.
     fn unwrap(self) -> T {
-        if let Option::Some(inner_value) = self {
-            inner_value
-        } else {
-            revert(0);
+        match self {
+            Option::Some(inner_value) => inner_value,
+            _ => {
```
adlerjohn, line 54: Is there a need to use `{  }` here? Does `revert` not half control flow analysis properly? If not, should an issue be filed, or is that expected behavior?

emilyaherbert: There is no reason to use `{ }`, this was just my personal code style habit. I updated it to match the style of the other branch.

sway-lib-std/src/result.sw
```diff
@@ -52,10 +50,11 @@ impl<T, E> Result<T, E> {
...
     fn unwrap(self) -> T {
-        if let Result::Ok(inner_value) = self {
-            inner_value
-        } else {
-            revert(0);
+        match self {
+            Result::Ok(inner_value) => inner_value,
+            _ => {
+                revert(0);
+            }
```
Braqzen, line 57:
```suggestion
            _ => revert(0),
```

Braqzen (changes requested):

adlerjohn (approved):

Braqzen (approved):

PR #1735, opened 2022-05-27
Adding an `else` in the U128 library to bypass a bogus warning

### LINKED ISSUE #1730: Implement a workaround for #1682
Implement a workaround for https://github.com/FuelLabs/sway/issues/1682 in the `U128` library to avoid the bogus unreachable code warnings until the root problem is fixed.

mohammadfawaz:
Closes #1730 
Also, a `;` after `if` expression are no longer needed if the `if` expression is not a RHS.


adlerjohn (approved):
Technically, the `else` can be kept in this particular case!

PR #1741, opened 2022-05-28
Add instructions to use fuelup.

adlerjohn:
gm
~~This works, although without #1742 users will still have to install openssl (they would have to do that anyways if compiling from source so it's not a regression in that sense).~~


adlerjohn:
Should be good to go now

JoshuaBatty (approved):

Issue #1605, opened 2022-05-19
Remove Inline Assembly Examples from the Docs


ControlCplusControlV:
The libraries example in the Sway book uses inline assembly, [Libraries](https://fuellabs.github.io/sway/latest/sway-program-types/libraries.html), other sections may also use inline assembly, which should be removed (exception being the section on assembly in the Sway Book).

This is confusing to people learning Sway, and examples should be rewritten in Sway entirely to focus on the language not it's assembly.

PR #1745, opened 2022-05-28
Improve the libraries section

mohammadfawaz:
Also closes https://github.com/FuelLabs/sway/issues/1605


docs/src/sway-program-types/libraries.md
```diff
@@ -1,75 +1,85 @@
...
+```
 
-impl Add for u64 {
-    fn add(self, other: Self) -> Self {
-        asm(r1: self, r2: other, r3) {
-            add r3 r2 r1;
-            r3: u64
-        }
-    }
+* An `use` statement that imports `revert` from another library _inside_ the standard library:
```
mohammadfawaz, line 23:
```suggestion
* A `use` statement that imports `revert` from another library _inside_ the standard library:
```

Braqzen (approved):

adlerjohn (approved):

PR #1748, opened 2022-05-28
refactor(forc-pkg): replace `git2` with `gitoxide`

### LINKED ISSUE #920: Evaluate the `gitoxide` pure-Rust git implementation crate as a potential replacement for `git2`
Currently `forc` uses the `git2` crate for providing general git dependency support.

`git2` is a wrapper around the well-vetted but un-rusty C library `libgit2`. `git2` also currently requires `openssl` to be installed on downstream systems.

The [`gitoxide` crate](https://github.com/Byron/gitoxide) is a pure-Rust implementation of git. We should investigate whether or not its feasible to use a subset of its subcrates (like `git-repository`, `git-transport`) as an alternative to `git2` in order to take advantage of the more Rust-esque API (particularly favouring `async` and `Result`s over callbacks). I haven't dug deep enough yet to know what TLS approaches are supported, but if it were to support `rustls` that might be yet another motivator and allow for dropping the `openssl` requirement for `forc`.

Br1ght0ne:
Close #920.

- [x] resolving `GitReference`s
- [ ] fetching
- [ ] `HEAD` manipulation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,6 +13,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "addr2line"
+version = "0.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b9ecd88a8c8378ca913a680cd98f0f13ac67383d35993f86c90a70e3f137816b"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
+name = "adler"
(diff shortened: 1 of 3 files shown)


sezna:
closing as this is outdated and the issue was solved in other ways, according to @Br1ght0ne

PR #1118, opened 2022-03-31
stdlib `Vec`

adlerjohn:
Fixes #1077

Based on #82

Blocked by:
- [x] #1121
- [x] #1143
- [x] #1956

PR #1750, opened 2022-05-28
demonstrating that the vec library builds now

emilyaherbert:
Untested! But it builds

This code is copied from the previously failing case: https://github.com/FuelLabs/sway/blob/906903aa247d21f1fdf6334e49f16f54c6b40372/sway-lib-std/src/vec.sw


sway-lib-std/src/vec.sw
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /// Gets the capacity of the allocation.
+    fn capacity(self) -> u64 {
+        self.cap
+    }
+
+    fn grow(self) {
+        let mut self = self;
+        let new_cap = if self.cap == 0 {
```
Braqzen, line 47: no match statement?

emilyaherbert:
Oopsies build errors.. one sec

sway-lib-std/src/vec.sw
```diff
@@ -0,0 +1,189 @@
...
+    pub fn capacity(self) -> u64 {
+        self.buf.cap
+    }
+
+    /// Clears the vector, removing all values.
+    ///
+    /// Note that this method has no effect on the allocated capacity
+    /// of the vector.
+    pub fn clear(self) {
+        let mut self = self;
```
Braqzen, line 140: Why is line 140 needed?

sway-lib-std/src/vec.sw
emilyaherbert: This code is copied from John's implementation here: https://github.com/FuelLabs/sway/blob/906903aa247d21f1fdf6334e49f16f54c6b40372/sway-lib-std/src/vec.sw

sway-lib-std/src/vec.sw
emilyaherbert: It was failing to build because it was complaining that self wasn't mutable. I tried adding `mut` in front of `self` in the parameters but it was giving me a type error, so I did this as a hack to get it to build.

Braqzen: I mean, I don't understand how assigning self back to self does anything. What is the point of this line?

adlerjohn:
@emilyaherbert can you upstream any fixes you find into #1118? Or you can leave them here and I can since there will be merge conflicts.

sway-lib-std/src/vec.sw
emilyaherbert: Sorry, I mistyped! Putting mut in front of self in the parameters was a _parse_ error. So basically all this line is doing is reassigning self to self, but with the mut keyword, so that it can avoid the parse error.

Braqzen: But why do you _need_ to assign self back to self? Isn't that basically a no-op?

emilyaherbert: Yes, and no. Normally, in Sway, and in Rust, you would add the `mut` keyword in front of `self` in the function parameters of `clear`. However doing that on `master` right now is either not supported or there is a bug, and doing that generates a parse error. So, this _is_ a no-op in that there are no changes to self, but it's also not a no-op because it allows the compiler to see that `self` is mutable, which is needs to do, in order to assign to it in the `clear` function.

I don't think this code should be merged! We should address the parse error first. I just wanted to demonstrate that this library does indeed compile on `master`, so I threw this hacky line in.

mohammadfawaz: Relevant: https://github.com/FuelLabs/sway/issues/1188

emilyaherbert:
Update: the test builds, but it returns the wrong value. I think that this is because of the call to `let mut self = self;`

mohammadfawaz:
I merge with master and pushed a small change to vec:
* Use `mut self` directly in all the methods that needs it.
* Only call `mcp` if the size is > 0. There are other places where this may be needed.

With this and https://github.com/FuelLabs/sway/pull/1956, the test added here passes. It's a pretty basic use case but this is a good sign nonetheless. Once https://github.com/FuelLabs/sway/pull/1956 goes in, we can do more extensive testing on `Vec<T>`.

PR #1754, opened 2022-05-28
`StorageMap` implementation and test

mohammadfawaz:
The implementation is actually quite simple and relies on the `__generate_uid()` intrinsics which returns a unique `b256` seed for every `StorageMap` that gets initialized via `new()`. Note that `new()` has to be called for each map once as shown in the test.

Remaining tasks:
* [ ] Docs: https://github.com/FuelLabs/sway/pull/1760


SilentCicero:
Needs docs :)

test/src/sdk-harness/test_projects/storage_map/src/main.sw
```diff
@@ -0,0 +1,281 @@
...
+    z: b256,
+}
+
+pub enum Enum {
+    V1: b256,
+    V2: u64,
+    V3: b256,
+}
+
+storage {
```
Braqzen, line 17: So we have native types that can be added into storage without the use of the std library. Why can a mapping not be one of those types that is built into ... the compiler? What makes a `u64` able to be put directly into storage but a `mapping` has to be in the library?

test/src/sdk-harness/test_projects/logging/src/main.sw
```diff
@@ -17,7 +17,7 @@ fn main() {
     let k: b256 = 0xef86afa9696cf0dc6385e2c407a6e159a1103cefb7e2ae0636fb33d3cb2a9e4a;
     let a: str[4] = "Fuel";
     let b: [u8;
-    3] = [1, 2, 3];
+    3] = [1u8, 2u8, 3u8];
```
Braqzen, line 20: Why do these 3 numbers need a type after them? The array is declared as `u8`. If anything I would expect possibly only declaring the first value as a `u8` similar to a vec in rust.

test/src/sdk-harness/test_projects/exponentiation/mod.rs
```diff
@@ -1,19 +1,17 @@
...
+use fuels::signers::LocalWallet;
 use fuels::tx::ContractId;
 use fuels_abigen_macro::abigen;
-use fuels::signers::LocalWallet;
 
-abigen!(
-    TestPowContract,
-    "test_artifacts/pow/out/debug/pow-abi.json"
-);
+abigen!(TestPowContract, "test_artifacts/pow/out/debug/pow-abi.json");
```
Braqzen, line 6: Shouldn't the exponentiation code be another issue?

test/src/sdk-harness/test_projects/storage_map/mod.rs
```diff
@@ -0,0 +1,477 @@
...
+        TxParameters::default(),
+    )
+    .await
+    .unwrap();
+
+    TestStorageMapContract::new(id.to_string(), wallet)
+}
+
+#[tokio::test]
+async fn can_insert_and_get() {
```
Braqzen, line 23: This test is way too large and difficult to follow. Split it into multiple functions. I would also consider assigning variables and passing them in into the "set" and "get" so that you can change something in 1 place and have the test continue to work rather than remembering to change it in both calls. If the functions are small enough then this may not be needed

Braqzen (changes requested):

test/src/sdk-harness/test_projects/storage_map/mod.rs
mohammadfawaz: Fully agree. The reason I decided to go with one big function is to show that you can defined many storage maps, insert into all of them separately, and then be able to read from all of them with no collision. That being said, I can probably showcase that with a much simpler large (but much smaller than this) test and let some other smaller tests check for individual case.

test/src/sdk-harness/test_projects/exponentiation/mod.rs
mohammadfawaz: All I did was run `cargo fmt`. I will revert this file for now.

test/src/sdk-harness/test_projects/logging/src/main.sw
mohammadfawaz: We just get warnings at the moment. We should be able to inver the type directly through the array but I guess we currently don't. I will revert this file for now as it has nothing to do with this PR.

test/src/sdk-harness/test_projects/storage_map/src/main.sw
mohammadfawaz: Well, `StorageMap` is just a `struct` so it's no different than custom types that users can define and declare inside a `storage` block. The compiler handles it exactly as it handles any user-defined type. We can make `StorageMap` a primitive type but I think that just adds bloat to the compiler and it's really not considered good compiler practice. `HashMap` in Rust is a library for example.

test/src/sdk-harness/test_projects/storage_map/.gitignore
```diff
@@ -0,0 +1,2 @@
+out
```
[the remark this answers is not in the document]
mohammadfawaz: You mean the `.gitignore` file? All the `sdk-harness` tests seem to have them so I kept it for consistency. We can decide to remove all of them though!

test/src/sdk-harness/test_projects/storage_map/src/main.sw
```diff
@@ -0,0 +1,281 @@
...
+    u64>, map18: StorageMap<str[33],
+    u64>, map19: StorageMap<[b256;
+    3],
+    u64>, 
+}
+
+abi StorageMapTest {
+    fn init();
+
+    fn into_u64_to_bool(key: u64, value: bool);
```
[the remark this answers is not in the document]
mohammadfawaz: In my head a map is "key _to_ value" so I went with `insert_into_<type1>_to_<type2>_map`. Let me know if that's clear!

Braqzen (approved):
Personally, I would categorise the tests into their own modules so that you can have cleaner names and a hierarchy. 

For example, all of the `test_u64_to_` tests would go under a `mod u64` (or `mod u64_to`) and each test name would be renamed accordingly e.g. `test_u64_to_bool_map` to `to_bool_map` or `bool_map`

test/src/sdk-harness/test_projects/storage_map/src/main.sw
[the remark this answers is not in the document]
Braqzen: Ah, I just go by "key, value pair" or similarly "key to value". I would have still gone with `insert_<type2>_into_<type1>` since I primarily go by the "pair" thought and thus I think of storing a value under a specific key

test/src/sdk-harness/test_projects/storage_map/.gitignore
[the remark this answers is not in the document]
nfurfaro: Yeah, I don't think we need to commit the build artifacts as CI builds them already.

test/src/sdk-harness/test_projects/storage_map/src/main.sw
```diff
@@ -0,0 +1,281 @@
...
+        storage.map14 = ~StorageMap::new::<u32, u64>();
+        storage.map15 = ~StorageMap::new::<(b256, u8, bool), u64 >();
+        storage.map16 = ~StorageMap::new::<Struct, u64>();
+        storage.map17 = ~StorageMap::new::<Enum, u64>();
+        storage.map18 = ~StorageMap::new::<str[33], u64>();
+        storage.map19 = ~StorageMap::new::<[b256;
+        3], u64>();
+    }
+
+    fn insert_into_u64_to_bool_map(key: u64, value: bool) {
```
nfurfaro, line 133: what about just: `fn insert_into_map1(...`

The function signature gives insight into the type of mapping this is.

nfurfaro (dismissed):
Minor nit about test function names, but it's non critical.
Congratulations on this, seriously!

test/src/sdk-harness/test_projects/storage_map/src/main.sw
mohammadfawaz: I thought it would be easier to know how to call these from the test if they have the types in the signature instead of having to look up what `map1` is and so on, but we can bikeshed over this!

Braqzen (approved):

test/src/sdk-harness/test_projects/storage_map/src/main.sw
nfurfaro: agreed that it's easier for the tester the way you have it.

nfurfaro (approved):

PR #1758, opened 2022-05-29
Adding a new `flags.sw` module

nfurfaro:
This PR just adds a new module `flags.sw` and moves the `enable_overflow` & `disable_overflow` functions into it.
One note: the function naming seems slightly off to me.

- `disable_overflow()` is actually used to disable the default "panic on overflow" VM behaviour (which in turn **enables** overflowing to occur and be recorded in `$of`.
- `enable_overflow()` is actually used to re-enable the default "panic on overflow" VM behaviour (which **disables** overflow from occurring.

Maybe I'm just thinking about this wrong 🤔 
An easy fix would be to simply swap the names.


mohammadfawaz:
> Maybe I'm just thinking about this wrong

I agree with you. So really the most descriptive names are `enable_panic_on_overflow()` and `disable_panic_on_overflow()`.

nfurfaro:
>Are those reasonable?

I think so; very clear what is happening with each.

sway-lib-std/src/flags.sw
```diff
@@ -0,0 +1,30 @@
+//! Functionality for setting & unsetting Fuel-VM flags to modify behavior related to the $err and $of registers
+library flags;
+
+use ::context::registers::flags;
+
+pub fn disable_overflow() {
```
adlerjohn, line 6: Add documentation and format file.

sway-lib-std/src/flags.sw
```diff
@@ -0,0 +1,33 @@
+//! Functionality for setting & unsetting Fuel-VM flags to modify behavior related to the $err and $of registers
+library flags;
+
+use ::context::registers::flags;
+
+/// Call this function to allow overflowing operations to occur without a FuelVM panic.
+/// IMPORTANT !!! Don't forget to call enable_panic_on_overflow() after performing the operations for which you disabled the default panic-on-overflow behaviour in the first place !
```
adlerjohn, line 7: American English everywhere.

nfurfaro: I need an american english vs vode linter :(

adlerjohn: Set your IDE language ❤️

sway-lib-std/src/flags.sw
```diff
@@ -0,0 +1,33 @@
+//! Functionality for setting & unsetting Fuel-VM flags to modify behavior related to the $err and $of registers
```
adlerjohn, line 1:
```suggestion
//! Functionality for setting and unsetting FuelVM flags to modify behavior related to the `$err` and `$of` registers
```

otrho (approved):

adlerjohn (approved):

PR #1760, opened 2022-05-29
Add StorageMap documentation in the Sway Book

mohammadfawaz:
This is only the first pass. I will do a second pass and clean up the rest of the book and examples as needed in a second PR.

### DIFF
--- a/docs/src/blockchain-development/storage.md
+++ b/docs/src/blockchain-development/storage.md
@@ -33,8 +33,44 @@ let v = storage.var1;
 Notes:
 
 * The only types currently supported by the syntax above are integers, Booleans, and structs.
-* The `storage` syntax cannot be used for mappings. Mappings need to be handled manually for now as shown in the [Subcurrency](../examples/subcurrency.md) example.
 * Storage, in general, is still work-in-progress and so, its use model may change in the future.
+
+## Storage Maps
+
+Generic storage maps are available in the standard library as `StorageMap<K, V>` which have to be defined inside a `storage` block and allow you to call `insert()` and `get()` to insert values at specific keys and get those values respectively. Storage maps also have to be initialized using `new()`. For example:
+
+```sway
+{{#include ../../../examples/storage_map/src/main.sw}}
+```
+
+There are three important components to correctly using a storage map:
+
+* Declaring the storage map with your desired data types inside a `storage` block:
+
+```sway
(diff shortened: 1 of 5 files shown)


docs/src/blockchain-development/storage.md
```diff
@@ -33,9 +33,45 @@ let v = storage.var1;
...
+* Declaring the storage map with your desired data types inside a `storage` block:
+
+```sway
+storage {
+    map1: Storage<u64, u64>,
+    // Other storage items
+}
+```
+
+* Making sure to initialize the storage map using `new()` in some contract method before using it:
```
adlerjohn, line 57: It's not clear if this needs to be called once and only exactly once, or if it should be called every time the storage variable will be used.

adlerjohn (changes requested):

docs/src/blockchain-development/storage.md
mohammadfawaz: Yeah only once. I'll make that clear.

adlerjohn: If it should only be called exactly once then it would be nice if the docs also included some sort of initialization logic.

mohammadfawaz: > if the docs also included some sort of initialization logic

I do have `fn init()` that calls `new()` or are you thinking of something else?

adlerjohn: As in, logic to ensure 1) init is only callable once, and 2) other methods require the contract to be initialized.

mohammadfawaz: I'm sorry I'm still a bit confused 🤔 ... By "logic" you mean having a few sentences in the docs saying those two things are required? Or is this actually logic in the code itself?

mohammadfawaz: Actually `new()` may be gone sooner than I thought. I'll clean up this section when I'm done.

adlerjohn: I meant actual code. But if this can be banished to the shadow realm before it even sees the light of day, even better!

SilentCicero: Banish it to the shadow realm. Agree.

adlerjohn (approved):
utACK ya love to see it

Issue #1757, opened 2022-05-29
Improve the section on `match` expressions in the Sway Book


mohammadfawaz:
Now that we have `enum` matching, it would be nice to include an example of that in https://fuellabs.github.io/sway/latest/basics/control_flow.html?highlight=match#match-expressions

PR #1762, opened 2022-05-29
Add an `enum` matching example

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/1757

### DIFF
--- a/examples/match_statements/src/main.sw
+++ b/examples/match_statements/src/main.sw
@@ -7,6 +7,12 @@ fn bar() {
     // do something
 }
 
+enum SomeEnum {
+    A: u64,
+    B: bool,
+    C: b256,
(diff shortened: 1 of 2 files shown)


examples/match_statements/src/main.sw
```diff
@@ -36,6 +42,23 @@ fn main() -> u64 {
         },
     };
 
+    // Match an enum
+    let e = SomeEnum::A(42);
+    let v = match e {
+        SomeEnum::A(val) => {
```
adlerjohn, line 48: Add a TODO here that the `{ }` aren't needed, but are only there because of <link to PR> a bug in the formatter.

adlerjohn (approved):

PR #1774, opened 2022-05-31
Make tuple annotation more specific

### LINKED ISSUE #906: Tuples with a single element cause some issues
Simple example:
```rust
script;
use std::chain::*;

struct S {
    t: (u64),
}

fn main() {
    let s = S {
        t: (2)
    };
}
```
I get the following compile error:
```
 8 |
 9 |     let s = S {
10 |         t2: (2)
   |              ^ Mismatched types.
expected: (u64)
found:    numeric.

11 |     };
12 | }
```
Can't get it to work even if I change `(2)` to `2`, `(2u64)`, or `2u64`.

emilyaherbert:
This PR makes the syntax for creating tuples more specific, where before it was ambiguous is some cases. There are now 4 concrete scenarios:

For example, this Sway code passes:

```rust
script;

use std::assert::assert;

struct S {
    t: (u64,)
}

fn main() -> u64 {
    let a = S {
        t: (2,)
    };
    let b = match a {
        S { t } => t,
    };
    assert(b.0 == 2);

    1
}
```

And this Sway code passes:

```rust
script;

use std::assert::assert;

struct S {
    t: (u64)
}

fn main() -> u64 {
    let a = S {
        t: (2)
    };
    let b = match a {
        S { t } => t,
    };
    assert(b == 2);

    1
}
```

But this Sway code fails:

```rust
script;

use std::assert::assert;

struct S {
    t: (u64,)
}

fn main() -> u64 {
    let a = S {
        t: (2)
    };
    let b = match a {
        S { t } => t,
    };
    assert(b == 2);

    1
}
```

```bash
  --> tuple_single_element/src/main.sw:11:13
   |
 9 | 
10 |     let a = S {
11 |         t: (2)
   |             ^ Mismatched types.
expected: (u64)
found:    numeric.
help: Struct field's type must match up with the type specified in its declaration.
12 |     };
13 |     let b = match a {
   |
____

error
  --> tuple_single_element/src/main.sw:16:14
   |
14 | 
15 |     };
16 |     assert(b == 2);
   |              ^^ No method named "eq" found for type "(u64)".
17 | 
18 |     1
   |
```

And this Sway code fails:

```rust
script;

use std::assert::assert;

struct S {
    t: (u64)
}

fn main() -> u64 {
    let a = S {
        t: (2,)
    };
    let b = match a {
        S { t } => t,
    };
    assert(b == 2);

    1
}
```

```bash
  --> tuple_single_element/src/main.sw:11:12
   |
 9 | 
10 |     let a = S {
11 |         t: (2,)
   |            ^^^^ Mismatched types.
expected: u64
found:    (u64).
help: Struct field's type must match up with the type specified in its declaration.
12 |     };
13 |     let b = match a {
   |
```

- Closes #906


test/src/e2e_vm_tests/test_programs/should_pass/language/tuple_single_element/Forc.toml
```diff
@@ -0,0 +1,6 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+license = "Apache-2.0"
+name = "tuple_single_element"
+entry = "main.sw"
+implicit-std = true
```
mohammadfawaz, line 6: The test should use a path to `std`. Or, as Alex said, maybe avoid using `assert` altogether and set this to `false`.

emilyaherbert: > The test should use a path to std

Why is that?

test/src/e2e_vm_tests/test_programs/should_pass/language/tuple_single_element/src/main.sw
```diff
@@ -0,0 +1,19 @@
...
+fn main() -> u64 {
+    let a = S {
+        t: (2,)
+    };
+    let b = match a {
+        S { t } => t,
+    };
+    assert(b.0 == 2);
+
+    1
```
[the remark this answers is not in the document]
emilyaherbert: Is there a reason to make this change, other than speeding up compile times for this one test? Using assert here is beneficial because it forces the test to include the standard library---including any tuples in the standard library. Assert also adds readability and maintainability---this test can easily be expanded in the future by adding additional asserts and is agnostic to the testing infra used (i.e. to change the test you don't also need to change the test harness)

test/src/e2e_vm_tests/test_programs/should_pass/language/tuple_single_element/Forc.toml
emilyaherbert: Ah I see, makes sense! But instead of pointing it to GitHub at all, its probably better to require that all tests use the local version of `core` and `std`.

canndrew:
If I understand this correctly, this PR is a bug fix which makes the type `(T)` get parsed as `T` in parenthesis, rather than as the 1-element tuple `(T,)`. Is that right? If so then :+1:

mohammadfawaz (approved):
Looks reasonable!

Issue #1773, opened 2022-05-30
Struct references do not contain unique Spans


JoshuaBatty:
I've run into an issue when traversing the `TypedParseTree` to collect tokens for the language server. I'm storing the `TypedDeclaration` and `TypedExpression` tokens in a `HashMap`, with the tokens `Ident` and `Span` used as the keys. This is all working great so far and i'm just about ready to put up a PR later today that demonstrates this. 

In the image below, VSCode is drawing squiggly lines under all of the tokens that were successfully collected during the traversal step outlined above. You can see that the only tokens that haven't been collected are all references to `Particle` other than the original declaration. This is because the references do not report their unique `Span`s, and instead, only have information about the original `Span`. 

We need to have the unique `Span` for each token in a file in order to get the token position in the file. Without this, we are unable to perform any language server functionality on that token. Interestingly, not only is the `Span` information incorrect, but the name reported on `TypedDeclaration::ImplTrait{trait_name, ..} trait_name.suffix` is `r#Self` instead of `Particle`. 

![Tpt traversal ](https://user-images.githubusercontent.com/1289413/171068498-a13ec815-e69a-4466-a618-5f5c0d1816f9.jpg)

I will note that the correct names and spans of each token are there during the initial `SwayParseTree` stage, but for some reason, this information is being thrown away during type checking. Was hoping to request from the compiler team for this information to be made persistent so the language server can access it on the `TypedParseTree`.

JoshuaBatty:
#1901 mostly fixes this issue for now. The only token in the above example that I'm unable to collect is the last line. 
`let p = ~Particle::new(position, velocity, acceleration, mass);`

From what I can see, this is because the `~Particle` token is a `MethodName::FromType` during the parsing stage. The `type_name_span` on `FromType` contains the correct span information that we need to collect this token. However, due to the inlining happening in the `type_check_method_application` fn, that information is lost and a `TypedExpression` is returned without knowledge of the span carried over. 

I think for now we are just going to have to ignore these token types and wait for the new collection context as proposed by @emilyaherbert. Open to other suggestions though.

JoshuaBatty:
This last issue was solved recently in #2389. 

There are other areas that still need the original Span information carried over during type checking but we can open individual issues for them as we come across them.

PR #1775, opened 2022-05-31
Collect tokens from the `TypedParseTree` for the language server.

JoshuaBatty:
This PR introduces the traversal of the `TypedParseTree` in order to collect tokens for the language server, but now with type information. We are extracting nearly all the tokens apart from a few struct reference tokens. The reason we are not collecting those currently is described in [this issue](https://github.com/FuelLabs/sway/issues/1773).

We aren't currently using any of the collected typed tokens, that work will be done in future PR's. Just wanted to put up the traversal and token collection work as it stands to keep the PR size down.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3674,6 +3674,7 @@ dependencies = [
  "tokio",
  "tower",
  "tower-lsp",
+ "tracing",
 ]
 
 [[package]]
--- a/sway-core/src/parse_tree.rs
+++ b/sway-core/src/parse_tree.rs
@@ -5,7 +5,7 @@ pub mod declaration;
 mod expression;
 pub mod ident;
 mod include_statement;
-mod literal;
+pub mod literal;
 mod return_statement;
 mod use_statement;
 mod visibility;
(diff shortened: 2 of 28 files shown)


otrho (commented):
LGTM in general, seems fairly straight forward.  There are a bunch of commented out code around the place and functions like `test_typed_parse()` which I don't think are called but are there for debugging?  So it could do with some tidying up, but I'm not sure how strict we are with that in `sway-lsp`. :slightly_smiling_face:

JoshuaBatty:
> LGTM in general, seems fairly straight forward. There are a bunch of commented out code around the place and functions like `test_typed_parse()` which I don't think are called but are there for debugging? So it could do with some tidying up, but I'm not sure how strict we are with that in `sway-lsp`. 🙂

Thanks Toby, yeah I decided to leave those in for now as I need them to continue working on this but don't want those functions getting called in the language server plugin. Also since I'm the only one touching this code currently perhaps it's ok?

otrho (approved):

PR #1776, opened 2022-05-31
Eliminating the need for `new()` to use `StorageMap`

mohammadfawaz:
The idea is as follows:
* Method applications look for a storage index (if applicable): for example `storage.map1.insert()` will become a `FunctionApplication` that also knows the index of `map1` in the `storage` block (in `method_application.rs`).
* Add a new metadata to function calls if they have an associated `self_state_idx` value (in `optimize.rs`). 
* The inliner moves that metadata to an intrinsic that (in `inliner.rs`)
* The intrinsic uses the index attached to it as metadata to re-generate the hash and return a `b256` as expected from it (in `from_ir.rs`).

The old intrinsic `__generate_uid()` is no longer needed. I just replaced it with `__get_storage_key()`.


sway-ir/src/value.rs
```diff
@@ -23,6 +23,7 @@ pub struct Value(pub generational_arena::Index);
 pub struct ValueContent {
     pub value: ValueDatum,
     pub span_md_idx: Option<MetadataIndex>,
+    pub state_idx_md_idx: Option<MetadataIndex>,
```
mohammadfawaz, line 26: These two should really be replaced with a `Vec` but I'm planning to do that in a separate PR because a lot of code would have to change which would be distracting.

otrho (approved):

PR #1783, opened 2022-05-31
Add additional operators as syntactic sugar

### LINKED ISSUE #1574: Introduce the `+=` and `-=` operators
In Sway, I'd like to be able to do:

```
let mut counter = 0;
while counter < 10 {
  counter += 1;
}
```

instead of what you currently have to do:

```
let mut counter = 0;
while counter < 10 {
  counter = counter + 1;
}
```

For ease of implementation, `+=` and `-=` could be syntactic sugar.

emilyaherbert:
This introduces the operators `+=`, `-=`, `*=`, `/=`, `<<=`, and `>>=`. They are implemented as syntactic sugar.

For example this:

```rust
let mut counter = 0;
while counter < 10 {
  counter += 1;
}
```

is desugared to this:

```rust
let mut counter = 0;
while counter < 10 {
  counter = counter + 1;
}
```

- Closes #1574

### DIFF
--- a/sway-core/src/convert_parse_tree.rs
+++ b/sway-core/src/convert_parse_tree.rs
@@ -1,3 +1,5 @@
+use sway_parse::expr::{ReassignmentOp, ReassignmentOpVariant};
+
 use {
     crate::{
         constants::{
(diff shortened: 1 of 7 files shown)


otrho (approved):
Looks good!

nfurfaro:
Looks good!

canndrew:
Is this what we want long-term? In rust these are (for better or worse) separate operators. There are cases where `x += y` will compile but `x = x + y` won't. For instance if `x` is `*z`, `z` has type `&mut T` and `T` is `!Copy`. These are kind-of edge cases though and you could argue that people shouldn't be overloading operators for `!Copy` types. But still, this does diverge us from rust a bit more.

otrho:
Maybe not?  But we don't have references yet, let alone `Copy` traits and giving the application developers a little bit more ergonomics is definitely worth it for now.

emilyaherbert:
@canndrew If we want to add that functionality in the future then we could simply change this case: https://github.com/FuelLabs/sway/blob/5030e155fed6204437f397ddc2c8eec55a43174d/sway-core/src/convert_parse_tree.rs#L1088

Issue #1648, opened 2022-05-24
Expand the scope of the IR unit tests to allow for use of the core library and standard library


emilyaherbert:
Currently, the IR unit tests are unable to use the core and standard libraries, due to a limitation in how the unit test infra is set up. I propose that we change this so that we can expand the scope of the IR unit tests.

Unit tests are here:
https://github.com/FuelLabs/sway/blob/53361ee726dacdb5d1ffa2f4f5354b1476b02f05/sway-core/src/optimize.rs#L2546

otrho:
Since the IR tests are using external files, they kinda feel more like integration tests anyway.  Doing this makes sense, especially considering the `core` dependency.

I would also like to streamline these tests a bit.  The way they currently work involves a lot of work, and maybe using `FileCheck` rather than just a straight diff here might be better.

mohammadfawaz:
+1 to using `FileCheck`. Really each test should look for specific things instead of diff-ing against some result that can change for all kinds of reasons that are unrelated to the purpose of the test.

mohammadfawaz:
I also think many `should_fail` E2E tests should actually be `FileCheck` tests in `sway-core/tests`. These would check the `stdout/stderr` for specific error messages (potentially many of them in one test).

otrho:
> +1 to using `FileCheck`. Really each test should look for specific things instead of diff-ing against some result that can change for all kinds of reasons that are unrelated to the purpose of the test.

Yeah, I originally just wanted to get something going quickly, but now all you have to do is add some new metadata somewhere and bang! the diff is pretty much every line.

> I also think many should_fail E2E tests should actually be FileCheck

Hrmm.  I think perhaps having a separate file per test with `FileCheck` annotations might be painful when we could just embed some checks for errors and warnings in the harness like we do now for results.

OTOH, I still really want to make #723 happen, and that would involve `FileCheck` most likely, and `lit` too maybe.

PR #1788, opened 2022-06-01
Tuple field reassignments

canndrew:
This PR allows tuple field projections to be on the left-hand-side of a reassignment.


canndrew:
@sezna: Re: the internal errors. That code is run after type-checking succeeds, and if the code is well-typed then those errors are impossible. Also they replace an internal error that was already being raised there, so I'm not introducing any new internal errors.

emilyaherbert (approved):

otrho:
Some new tests under `sway-core/tests` would be nice, but at the moment it's a pain and we have #1648.

PR #1790, opened 2022-06-01
Disambiguate `path` dependencies using the ID of the package at the path root

### LINKED ISSUE #1637: Only include git source string in `Forc.lock` `dependencies` lists if necessary for disambiguation
Currently, the `dependencies` list for each `[[package]]` within the `Forc.lock` always specifies the full git source string for each git dependency, even if there is only one instance of that dependency within the entire graph.

We should only include the git string in the case that it is necessary to disambiguate between multiple packages within the graph that have the same name. Otherwise, we should just specify the dependency's name and fetch the git source from that dependency node as necessary.

This would be similar to how `Cargo.lock` only specifies the semver version of each `crates.io` dependency in the case that there are multiple instances of that dependency within the graph. Otherwise, it just uses the crate name.

This behaviour is useful for quickly identifying whether or not a project may unintentionally contain duplicates of a dependency (e.g. at multiple different versions).

### LINKED ISSUE #1789: `[dependencies]` sourced by `path` should include their parent package in their `Forc.lock` file entry for disambiguation
## Problem

Currently, `[dependencies]` that are sourced via `path` always have a single entry within the graph and in turn within the lock file.

As a result, odd behaviour can arise if more than one package depends on a `path` dependency with the same name.

For example, consider the following `[dependencies]` table:

```toml
[dependencies]
foo = { path = "/path/to/foo" }
bar = { git = "https://url.com/to/bar" }
```

If `bar` also has a dependency named `foo` that is sourced via a `path`, we have a problem as only either the top-level `foo`, or `bar`'s `foo` dependency will be stored in the graph - not both.

It's possible that we might have ran into similar issues with `core` and `std`. #1778 is possibly related, but also maybe not.

## Solution

We should disambiguate dependencies that are sourced via `path`. We can do so by including whether or not the `path` is relative to some other package, and if so which package it is relative to. This should be included for each path-sourced package in the `Forc.lock` as well the in-memory package graph by adding a field to `SourcePinned::Path`, e.g. `SourcePinned::Path(SourcePathPinned)` where:

```rust
/// A pinned instance of a path source.
#[derive(Clone, Debug, Eq, Hash, PartialEq, Deserialize, Serialize)]
pub struct SourcePathPinned {
    /// If this path dependency was relative to another package, this is that package's ID.
    pub relative_to: PinnedId,
}
```

mitchmindtree:
Implements and closes #1789. Also closes #1637.

TODO:

- [x] Disambiguate `path` dependencies using parent package ID.
- [x] Impl `Display` and `FromStr` for `PinnedId` to convert it to and
  from hex for path source string in lock file.
- [x] Address #1637 now to reduce this PR adding even more noise in the
  lock file `dependencies` lists.
- [x] Update lock files of all examples and tests in repository.

### DIFF
--- a/examples/asm_return_tuple_pointer/Forc.lock
+++ b/examples/asm_return_tuple_pointer/Forc.lock
@@ -1,11 +1,14 @@
 [[package]]
 name = 'asm_return_tuple_pointer'
+source = 'root'
 dependencies = ['std']
 
 [[package]]
 name = 'core'
+source = 'path+from-root-25602139F18DA3F2'
 dependencies = []
 
 [[package]]
 name = 'std'
+source = 'path+from-root-25602139F18DA3F2'
 dependencies = ['core']
--- a/examples/counter/Forc.lock
+++ b/examples/counter/Forc.lock
@@ -1,11 +1,14 @@
 [[package]]
 name = 'core'
+source = 'path+from-root-D8BC155605882939'
 dependencies = []
 
 [[package]]
 name = 'counter'
+source = 'root'
 dependencies = ['std']
 
 [[package]]
 name = 'std'
+source = 'path+from-root-D8BC155605882939'
 dependencies = ['core']
--- a/examples/enums/Forc.lock
+++ b/examples/enums/Forc.lock
@@ -1,12 +1,4 @@
 [[package]]
-name = 'core'
-dependencies = []
-
-[[package]]
 name = 'enums'
+source = 'root'
 dependencies = []
-
-[[package]]
-name = 'std'
-source = 'git+https://github.com/fuellabs/sway?tag=v0.14.5#60c626cf486d5c53c44d84db1ec81cb90174cad3'
-dependencies = ['core']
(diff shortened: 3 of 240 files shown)


mitchmindtree:
## Problem

In attempting to update the E2E lock files, I realised that there's a problem with the approach of disambiguating path dependencies based on the immediate parent.

For example, in the following graph where *a -> b* means *a* depends on *b* via a path dependency:

- *foo -> bar*
- *foo -> std*
- *bar -> core*
- *std -> core*

We have an issue where even though *bar* and *std* both refer to the same *core* dependency, *core* ends up being introduced twice as it is disambiguated via the immediate parent. This exact pattern occurs multiple times within the E2E tests, and would no likely occur downstream in projects with multiple inter-related path dependencies within a single forc workspace.

## Possible Solution

Rather than only using the immediate parent's package ID to disambiguate path dependencies, we should instead use the ID of the closest non-path ancestor (i.e. git or root package ID).

mitchmindtree:
OK, I've implemented the solution above and all seems to be working nicely, including diamond subgraphs of path dependencies as in the given example.

The forc lock files of all the examples and tests have also been updated with their new `source` fields, necessary to disambiguate path deps.

JoshuaBatty (dismissed):
Nice work, utACK.

mitchmindtree:
Just rebased again to ensure I've updated all the recent lock file additions since CI last ran (see final commit).

otrho (approved):
I'll just approve again...

PR #1791, opened 2022-06-01
Add `Identity` type

### LINKED ISSUE #1309: Make the usage of msg_sender() a bit more ergonomic.
In Ethereum development, `msg.sender` is used everywhere for authentication. We simply call msg.sender and know that we’ll get an `Address` back.
We can’t  do that in Sway because:
- `Address` and `ContractId` are different types
- either type may be returned by `msg_sender()`
- an error may also be returned (if coins are spent from more than one address, then a unique sender can't be determined)

We can at least abstract most of the boilerplate needed to unwrap a nested Result away from the user for the most common cases. 

Open to input on how we do this. Probably the most helpful thing would be simple examples (even pseudocode) of how you would like to be able to use `msg_sender()` in practice, from which we can derive the best way forward.

## Ideas for helper functions
- similar to `is_err()` and `is_ok()`, we could have `is_address()` & `is_contract()` which return `bool` values.
- we can have an `unwrap_address()` and `unwrap_contract()` which either resolve to the thing or panic, or return Options. 
- We could even have a `unwrap_bytes()` which returns the inner `b256` value of either an `Address` or a `ContractId`

The problem with the Options is that it just defers the actual unwrapping and pushes it back into the users code, which is sort of what I want to avoid. That said, I think we should maybe do both so users can decide how to handle the absence of the value they want (ie: panic (revert) or something else). 

For many use cases, panicking when an expected address isn't there makes sense. We encode these sorts of checks everywhere when developing on Ethereum, ie: `Require(msg.sender != 0x234.., "Wrong Address!");`
so this would be the equivalent of saying `Assert(msg_sender().is_address);` before proceeding to use the Address value.

Note that enum matching (coming soon) will make this much cleaner, but those should mostly be implementation details most users don't need to care/know about.

cc @SilentCicero @Braqzen

nfurfaro:
This PR adds the new type `Identity`. 

From the docs:
>//! A wrapper type with 2 variants (`Address` or `ContractId`) sed to represent either in a generic way. The use of this type allows for handling interactions with contracts and addresses in a consistent manner.

### It:
- refactors `msg_sender()` to use the `Identity` type in place of the `Sender` type
- deprecates `Sender`
- adds a new `mint_to()` function to `token.sw`.
- adds tests for `mint_to`
- updates examples to use `Identity`

closes #1309 (in conjunction with the new matching on enums syntax in the language).


nfurfaro:
I need to update the sway-examples first.

sway-lib-std/src/vm/evm/ecr.sw
```diff
@@ -13,8 +13,7 @@ pub fn ec_recover_evm_address(signature: B512, msg_hash: b256) -> Result<EvmAddr
     let pub_key_result = ec_recover(signature, msg_hash);
 
     match pub_key_result {
-        Result::Err(e) => Result::Err(e),
-        _ => {
+        Result::Err(e) => Result::Err(e), _ => {
```
Braqzen, line 16: yikes, this formatting...

test/src/sdk-harness/test_artifacts/auth_testing_contract/src/main.sw
```diff
@@ -6,45 +6,38 @@ use std::contract_id::ContractId;
...
-                ret = true;
-            } else {
-                ret = false;
-            };
-        };
-
+        }
+        let unwrapped = result.unwrap();
+        match unwrapped {
+            Identity::ContractId(v) => {ret = true},
```
Braqzen, line 24: should the assertion be removed?

test/src/sdk-harness/test_artifacts/auth_testing_contract/src/main.sw
```diff
@@ -6,45 +6,38 @@ use std::contract_id::ContractId;
...
-                ret = true;
-            } else {
-                ret = false;
-            }
-        };
-
+        }
+        let unwrapped = result.unwrap();
+        match unwrapped {
+            Identity::Address(v) => {ret = true},
```
Braqzen, line 38: should the assertion be removed?

test/src/sdk-harness/test_artifacts/reentrancy_attacker_contract/src/main.sw
```diff
@@ -1,20 +1,20 @@
...
-        } else {
-            revert(0);
-        }
-    } else {
-        revert(0);
+fn get_msg_sender_id_or_panic(result: Result<Identity, AuthError>) -> ContractId {
+    match result {
+        Result::Ok(s) => {
+            match s {
+                Identity::ContractId(v) => {v},
```
Braqzen, line 13: Pretty sure you do not need the brackets around the **v** etc., see it done in multiple places

sway-lib-std/src/token.sw
```diff
@@ -6,6 +6,21 @@ use ::contract_id::ContractId;
...
+
+/// Mint `amount` coins of the current contract's `asset_id` and send them to `destination` by calling either force_transfer() or transfer_to_output(), depending on the type of `Identity`.
+pub fn mint_to(amount: u64, recipient: Identity) {
+    mint(amount);
+    match recipient {
+        Identity::Address(addr) => {
+            transfer_to_output(amount, contract_id(), addr);
+        },
+        Identity::ContractId(id) => {
+            force_transfer(amount, contract_id(), id);
```
Braqzen, line 19: `force_transfer()` really needs to be renamed to `transfer_to_contract()` to follow the similar name of transferring to an output

sway-lib-std/src/vm/evm/ecr.sw
nfurfaro: I'm tired of fighting the formatter, as long as CI is happy with it. ( I know a new formatter version is coming which hopefully will clean this up).

test/src/sdk-harness/test_artifacts/auth_testing_contract/src/main.sw
nfurfaro: This is an assignment.

test/src/sdk-harness/test_artifacts/auth_testing_contract/src/main.sw
nfurfaro: same as above.

test/src/sdk-harness/test_projects/token_ops/src/main.sw
```diff
@@ -10,6 +10,8 @@ abi TestFuelCoin {
     fn get_balance(target: ContractId, asset_id: ContractId) -> u64;
     fn mint_and_send_to_contract(amount: u64, destination: ContractId);
     fn mint_and_send_to_address(amount: u64, recipient: Address);
```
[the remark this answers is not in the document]
nfurfaro: We decided to keep `mint_to_contract` and `mint_to_address` (in addition to the new `mint_to` function) in the `token` module.

The test contract tests all 3 of these. 
However, the SDk currently can't handle passing in an `Identity` as a parameter to an abi function, so I exposed 2 different functions on the contract to test `mint_to`.
`Address` and the other a `ContractId`.
 Both just wrap their param in an `Identity` and call the lib function `mint_to` with an `Identity` as it's meant to be called, hence the similarly named functions.
 
I will update the names to make this more clear.

sway-lib-std/src/token.sw
Braqzen: `transfer_to_output()` can be stopped?

nfurfaro: no, but forwarding coins to a contract via a call could be rejected. This is a way to force coins directly to a contract without the receiving contract being able to do about it.
The danger here is that the receiving contract may not be equipped to ever withdraw/move those coins so they can potentially be lost.

adlerjohn: `transfer_to_contract` alone isn't the safest name because it doesn't capture the semantics that it's _only_ transferring to a contract, with no possibility of logic execution by the receiver. This doesn't really exist in Ethereum land, so we have to be careful. Regardless, a name change can be bikeshed in a subsequent PR.

sway-lib-std/src/vm/evm/ecr.sw
adlerjohn: If you want to avoid this, you can enclose each arm in `{ }`! https://github.com/FuelLabs/sway/blob/60c626cf486d5c53c44d84db1ec81cb90174cad3/examples/wallet_smart_contract/src/main.sw#L41-L46

sway-lib-std/src/chain/auth.sw
```diff
@@ -37,24 +33,23 @@ pub fn caller_contract_id() -> ContractId {
...
+/// Returns a `Result::Ok(Identity)`, or `Result::Err(AuthError)` if an identity cannot be determined.
+pub fn msg_sender() -> Result<Identity, AuthError> {
     if caller_is_external() {
-        let sender_res = get_coins_owner();
-        match sender_res {
-            Result::Ok(sender) => Result::Ok(sender),
-            sender_res => sender_res,
+        let owner_res = get_coins_owner();
+        match owner_res {
+            Result::Ok(owner) => Result::Ok(owner), owner_res => owner_res,
```
adlerjohn, line 42: You can just return `owner_res` here directly, no? Saves an object creation.
```suggestion
            Result::Ok(owner) => owner_res, owner_res => owner_res, 
```

sway-lib-std/src/identity.sw
```diff
@@ -0,0 +1,10 @@
+//! A wrapper type with 2 variants (`Address` or `ContractId`) sed to represent either in a generic way. The use of this type allows for handling interactions with contracts and addresses in a consistent manner.
```
adlerjohn, line 1: Also wrap at 80 chars pls
```suggestion
//! A wrapper type with 2 variants, `Address` an `ContractId`. The use of this type allows for handling interactions with contracts and addresses in a unified manner.
```

sway-lib-std/src/chain/auth.sw
otrho: You could remove the `match` expression entirely, if both arms are returning the same thing, and that thing is what is being matched...

sway-lib-std/src/token.sw
nfurfaro: >a name change can be bikeshed in a subsequent PR.
Agreed. I almost changed it here, but reverted the commit.

sway-lib-std/src/chain/auth.sw
nfurfaro: The changes to `msg_sender` were introduced in commit 307d91cb1a0260027c5256b9086254a7b47eb189, which is not part of this PR.
That said, there were a few issues last time we tried to clean this up.
- `let owner_res = get_coins_owner();` this call to `get_coins_owner` returns a result which looks like it should be able to be directly returned by `msg_sender`, but they're actually different types (or they were... I'll have to retest this).
- the reason we have `Result::Ok(owner) => Result::Ok(owner)` is because the first `Result` is from `get_coins_owner()` so we can't return it directly. So we extract the inner `owner` value and cast it as a `Result` of the correct type. Will check this out tomorrow and retest to see what might have changed.
If possible I'll remove the `match` entirely. It looks like it should work...

otrho: OK, we need first class functions so we can add `map` and `map_err` etc.  Carry on.

sway-lib-std/src/vm/evm/ecr.sw
Braqzen: I think it's better to wait till formatter v2 is out than to `{}`

sway-lib-std/src/identity.sw
```diff
@@ -0,0 +1,10 @@
...
+library identity;
```
Braqzen, line 2: What's the reason for moving `Identity` out of `auth`?

nfurfaro: I had originally defined `Sender` in the `auth` module because that's where I first needed it.
Now that the usage of `Identity` is going to be more widespread I thought it could stand on its own.
I don't feel strongly about this though; it could really go either way.

sway-lib-std/src/chain/auth.sw
nfurfaro: I stand corrected. It seems I was able to successfully remove the `match` altogether. This is only surprising because previously I was not able to do this. Nice to see things progressing on this front!

adlerjohn (approved):

PR #1799, opened 2022-06-01
refactor: make use of intrinsics module in storage

nfurfaro:
All in the title.


mohammadfawaz (approved):
LGTM! Thanks.

adlerjohn (commented):
Both in the interest of being lazy and not having to update contract IDs, and in the interest of keeping things consistent: I don't think we actually need this change. Not all intrinsics will be exposed to the end user, e.g. ones that get the storage key might be be included in the `std::intrinsics` modular. Because of that, it would be more consistent if the stdlib used intrinsics  directly instead of the wrapper functions.

nfurfaro:
Closing as unneeded.

PR #1800, opened 2022-06-01
Introduce forc build [profile]s in anticipation of introduction of compiler optimisation passes

### LINKED ISSUE #1785: Introduce `forc` build `[profile]`s in anticipation of introduction of compiler optimisation passes
Currently, `forc`'s support for different build profiles is non-existent. We *do* currently anticipate their support within the compiler output artifact directory by placing artifacts in a profile subdirectory which is currently assumed to be "debug", e.g. `<project>/out/debug/` where `debug` is the default profile.

## Motivation

Build profiles should allow for describing a set of options for guiding compilation. This could include an "optimisation level" as a short-hand for a particular list of optimisation passes to run, or debug options for including extra kinds of annotations/symbols in the output that might be required for debugging in certain cases.

There has recently been some discussion around introducing optimisation passes into `sway-core`. In order to easily enable/disable these, we should prepare support for build profiles so that devs can easily enable/disable certain paths and set the precedent for feeding project-specific build parameters into compilation in general. This could also be a useful place for introducing flags for experimental compiler features (rather than needing to add CLI flags to `forc` on every invocation).

## Cargo's `[profile]`

Cargo uses a `[profile]` table for this which might act as a good reference for the kinds of options that we might want available. It's worth keeping in mind that many (most?) of these options are specific to Rust, and that upon introducing this feature it's OK to start with a very minimal set of support (e.g. optimisation level).

https://doc.rust-lang.org/cargo/reference/profiles.html

## Forc `[build-profile]`?

Seeing as the `Forc.toml` manifest will likely also include support for certain kinds of *deployment* configuration, we may wish to name this `[build-profile]` to clarify that this table of options is specifically designed to be received as an input to *compilation*. The CLI option flag given to `forc build` could still be `--profile` as the command makes it clear we are referring to the project's `build` process.

## Default build profiles

To be consistent with cargo and the vast majority of languages out there, we should provide two profiles by default:

- `debug` - little to know optimisations applied and any useful debugging options enabled.
- `release` - all stable optimisations applied and stripped of all debugging options that would affect performance/size in any way.

If no `--profile <profile>` option is specified, we should likely default to `debug` as is currently case and consistent with cargo behaviour.

CLI shorthands could be provided for these common default profiles so that rather writing `forc build --profile release`, users can instead write `forc build --release` (similar to cargo).

## Implementation

1. Introduce a `[build-profile]` table to the `Manifest` within `forc-pkg`.
2. Add some `const` `BuildProfile`s for the default debug and release profiles.
3. Add a `forc build` `--profile <profile>` for specifying which profile to use by its name.
4. In `forc build`, determine which build profile was specified.
5. Forward `BuildProfile` options to compilation via `sway_core::BuildConfig`. This will involve ensuring that the `BuildConfig` has the necessary params.
6. Ensure the output artifact directory used actually matches the name of the build profile.
7. Document the feature in the forc manifest reference within The Sway Book.

kayagokalp:
## About this PR

- [x] Documentation
- [x] ~Remove flags from the CLI commands as we will have them in the profile.~ Decided to have them override the selected profile, can also add a warning while doing so. 

closes #1785.

- `[build-profile]` is introduced which can be used to pass optimizations settings to the compiler. 
- There are default `debug` and `release` profiles introduced. These profiles are included in the manifest file *implicitly* if they are not already present. 
- `--release` flag is introduced to `forc build`, `forc deploy`.
- `--build-profile` option is introduced to `forc build`, `forc deploy`.

## To Test
Testing steps will be added before marking this ready to review

### Testing `--release` flag and default build profile (`debug`)
  * Since the default build profile (`debug`) has silent = `false`, ~setting `--release` should silence a warning/error producing build process~ After https://github.com/FuelLabs/sway/pull/1800/commits/11833301e6e6b8b0bfaecfb81448ea9361fb1f88 `--release` does not silence. A `should_fail` test from `e2e_vm_tests` can be used for testing this. 
  1. `forc build` inside a `should_fail` `e2e_vm_test` (an example can be `abi_impl_purity_mismatch`) -> should print full error
  2. `forc build --release` inside a `should_fail` `e2e_vm_test` (an example can be `abi_impl_purity_mismatch`) -> should just print  "Aborting due to 1 error."

### Testing `--build-config` with a user-defined build profile
  1. run `forc init test_project`
  2. add the following section to `Forc.toml`:
      ```
     [build-profile.test_profile]
     print-finalized-asm = true
     print-intermediate-asm = false
     print-ir = false
     silent = true
      ```
  3. run `forc build --build-profile test_profile` -> finalized asm should be printed.

### Testing `--build-config` with a user-defined build profile and `--release` at the same time
  1. run `forc init test_project`
  2. add the following section to `Forc.toml`:
      ```
     [build-profile.test_profile]
     print-finalized-asm = true
     print-intermediate-asm = false
     print-ir = false
     silent = true
      ```
  3. run `forc build --build-profile test_profile --release`  => you should get a warning stating that the release will be used and no finalized asm should be printed as `release` build profile is used.

### Testing overriding a build profile with CLI options
  1. run `forc init test_project`
  2. add the following section to `Forc.toml`:
      ```
     [build-profile.test_profile]
     print-finalized-asm = false
     print-intermediate-asm = false
     print-ir = false
     silent = true
      ```
  3. run `forc build --build-profile test_profile --print-finalized-asm`  => since you provided an option (`--print-finalized-asm`) forc will omit `--build-profile test_profile` part.


ra0x3:
Just chiming in to say these PR testing steps and the descriptions are great, and extremely helpful for me as a reviewer. 👍

docs/src/forc/manifest_reference.md
```diff
@@ -15,6 +15,8 @@ The `Forc.toml` (the _manifest_ file) is a compulsory file for each package and
 * `[network]` — Defines a network for forc to interact with.
   * `url` — URL of the network.
 
+* [`[build-profiles]`](#the-build-profiles-section) - Defines the build profiles.
```
adlerjohn, line 18:
```suggestion
* [`[build-profiles]`](#the-build-profiles--section) - Defines the build profiles.
```

kayagokalp:
@sezna That is a good point! I will be changing it. Should there be any difference between `debug` and `release` with the current parameter set we have for the profiles?

ra0x3 (approved):
- Pulled down the PR and followed the testing steps ✅ 
 - All testing steps seem to have produced expected outputs ✅ 
 - Approving so as to not be a blocker, but will defer to @JoshuaBatty  and @mitchmindtree 

@kayagokalp  One comment (and not a blocker) is that when passing 2 profiles to `build` I go the message:

> "Both test_profile and release provided as build profile. Using release!"

Which took me a second to comprehend. Was thinking something a little more explicit might be a bit better (again not a blocker)

> "You specified both 'test_profile' and 'release' profiles. Using the 'release' profile"

Or maybe even more verbose/explicit than that ☝️. So as to make it impossible for someone to misunderstand what might be going on :)

adlerjohn (approved):
utACK, but deferring final approval to @JoshuaBatty or @mitchmindtree

JoshuaBatty (approved):
Pulled down and tested locally and can confirm it works. Great work.

forc/src/cli/commands/build.rs
```diff
@@ -48,6 +48,15 @@ pub struct Command {
     /// needs to be updated, Forc will exit with an error
     #[clap(long)]
     pub locked: bool,
+    /// Name of the build profile to use.
+    /// If it is not specified, forc will use debug build profile.
+    #[clap(long)]
+    pub build_profile: Option<String>,
```
mitchmindtree, line 54: I think we can name this option `profile`, as the `build` prefix is already implied by the command. E.g.

```
forc build --profile <profile>
```
rather than
```
forc build --build-profile <profile>
```

mitchmindtree (commented):
I've left a few comments that we should address in followup work, but overall good stuff @kayagokalp :)

Issue #1516, opened 2022-05-10
[Tracking] `swayfmt-v2` MVP


eureka-cpu:
Tracking for the new formatter:

>The design of `rustfmt` is to leave code functionality in tact, but also do more than just move white space. For instance,
>long generic parameters can be cleaned up with a `where` clause addition and formatting of those parameters after it. I
>think that's a really great characteristic that keeps code uniform and even teaches users the standard practices.
>
>It uses an `AST` for most of the formatting, but for macros it uses tokens or otherwise leaves them untouched. Most macros aren't formatted, unless necessary like the `format!` macro since it takes in variables that result in a `String`.
>
>`rustfmt` can recognize when something won't benefit from being formatted and opts that portion of the code out of the formatting process. If code doesn't fit the conditions of what would need formatting, e.g. a user created diagram, then it will
>just leave it alone.
>
Please feel free to hop in and take on some of this if it interests you:

- husk of the formatter, config settings & `swayfmt.toml` #1425 #1803 @eureka-cpu 
- imports #2295 @eureka-cpu 
- literals @eureka-cpu 
- whitespace #1888 #1894 #1911 @eureka-cpu 
- expressions #2338 @eureka-cpu 
- collections @eureka-cpu 
- user defined @kayagokalp 
- items
  - abi #2043 #2044 @eureka-cpu 
  - const #2005 @kayagokalp 
  - enum #1873 @kayagokalp 
  - fn #2173  @eureka-cpu 
  - impl #2403 @eureka-cpu 
  - storage #2069 @kayagokalp 
  - struct #2053 @kayagokalp 
  - trait #2294 @eureka-cpu 
  - use #2295 @eureka-cpu 
  - comments ( blocker: #1517 ) @mitchmindtree & @kayagokalp (#2229, #2311)
  - `break` and `continue` https://github.com/FuelLabs/sway/issues/2118 (pending https://github.com/FuelLabs/sway/pull/2112) @eureka-cpu 
- attributes #2012 #2061 @eureka-cpu 
- Contract call arguments: https://github.com/FuelLabs/sway/issues/992 #2338 @eureka-cpu 
- long conditionals #2597 @eureka-cpu 

`swayfmt`'s goal is to be somewhat transparent. It should format Sway code to our standards, and if the user so decides it, they can add a `swayfmt.toml` to their root dir and it will assume only the changes present, and leave the rest to default.

Once completed add to CI documentor: #1806

mitchmindtree:
[...]

PR #1803, opened 2022-06-01
Adds `forc-fmt-v2` plugin

eureka-cpu:
With this it should be easier for anyone wanting to work on the new formatter to test their logic.

```
cargo build --manifest-path ~/sway/forc-plugins/forc-fmt-v2/Cargo.toml && mv ~/sway/target/debug/forc-fmt-v2 ~/.cargo/bin
```

I have mine set as an alias `forcmat` to speed up the process.


eureka-cpu:
For some context regarding the versions, I had some trouble getting it to work just keeping them at `0.0.0` as instructed per my last PR, so I updated them to current and it seems to work now. LMK if there's a change I need to make there!

forc-plugins/forc-fmt-v2/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "forc-fmt-v2"
+version = "0.14.4"
```
adlerjohn, line 3: To avoid having to keep up versioning an unpublished create, just set to 0
```suggestion
version = "0.0.0"
```

eureka-cpu: I left a comment in anticipation of this, but I'll look into it further!

adlerjohn: Oh shoot I missed that haha

binggh:
If this is something official that should be documented in the Sway book, please consider installing the plugin within the CI so that the mdbook preprocessor can install and document the commands! This can be done in this PR or a separate PR.

More in the [preprocessor README](https://github.com/FuelLabs/sway/blob/master/scripts/mdbook-forc-documenter/README.md)

eureka-cpu:
> If this is something official that should be documented in the Sway book, please consider installing the plugin within the CI so that the mdbook preprocessor can install and document the commands! This can be done in this PR or a separate PR.
> 
> More in the [preprocessor README](https://github.com/FuelLabs/sway/blob/master/scripts/mdbook-forc-documenter/README.md)

This PR is only to make it more effective for other devs to contribute to #1516 but it will eventually replace the current `forc-fmt` plugin. I don't think that updating the documentation at this point is necessary but will need to be before the official change happens!

forc-plugins/forc-fmt-v2/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+                        manifest_path.clone(),
+                    );
+                    match Formatter::format(
+                        formatting_options,
+                        file_content.clone(),
+                        Some(&build_config),
+                    ) {
+                        Ok(formatted_content) => {
+                            if app.check {
+                                if *file_content != *formatted_content {
```
JoshuaBatty, line 83: `formatted_content` should be fine here, no need to dereference `*formatted_content`

JoshuaBatty (commented):
Nice one looks good apart from that one small nit.

JoshuaBatty (approved):
LGTM

PR #1804, opened 2022-06-01
Create installation.md

sdankel:
Provide installation instructions for forc.

Is there anything I need to do to include this file in the docs website besides creating the file?


adlerjohn (commented):
Aren't installation instructions available here? https://github.com/FuelLabs/sway/blob/master/docs/src/introduction/installation.md

We use [mdBook](https://github.com/rust-lang/mdBook) for the book. Full instructions on how to add pages are there.

Issue #1813, opened 2022-06-02
Add support for `enums` in `storage {}`


Braqzen:
As a user I would really like to store an `enum` inside `storage {}`.

Currently I must use another variable type, `u64`, with some number to perform the same state checks etc.
This isn't the worst thing in the world however when it comes to an `enum` that uses the common `Address` and `ContractId` that becomes a different story as now I must use additional variables or add additional code to figure out which type I am using.

mohammadfawaz:
I have a branch with the fix. I'll put up a PR today.

PR #1826, opened 2022-06-02
Allow `enums` and `strings` in a `storage` blocks

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/1813

Again, this is temporary until we migrate everything to use the `Storable` trait and have a derive system that enables automatic inference of `impl` blocks for arbitrary structs and enums.

Basically, we have full coverage now for all types except for arrays. The plan for arrays is similar to how we handle structs, but we don't really support chaining well (example: `storage.some_struct.array[i].some_field`). This may not matter if we can support arrays with `Storable`.

The idea with this change is to basically load/store using `state_load_quad_word`/`state_store_quad_word` as many times as needed to cover the full `union` or `string`. We use an array of `b256` for full coverage and then use up as much as needed from it. For example, if the size of the `string` is 6 words, then we create an array `[b256; 2]`, and only use up an element and half of it (effectively).


sway-core/src/optimize.rs
```diff
@@ -2148,6 +2148,124 @@ impl FnCompiler {
...
+                        // Use state_load_quad_word/state_store_quad_word as many times as needed
+                        // using sequential keys
+
+                        // First, create a name for the value to load from or store to
+                        let mut value_name = format!("{}{}", "val_for_", ix.to_usize());
+                        for ix in &indices {
+                            value_name = format!("{}_{}", value_name, ix);
+                        }
+                        let alias_value_name =
+                            self.lexical_map.insert(value_name.as_str().to_owned());
```
otrho, line 2161: This ends up being more lines, even though it's a 'one-liner'.  Well, at least the obsolete `.as_str().to_owned()` should go.
```suggestion
                        let value_name = format!(
                            "val_for_{}{}",
                            ix.to_usize(),
                            indices
                                .iter()
                                .map(|idx| format!("_{idx}"))
                                .collect::<Vec<_>>()
                                .join("")
                        );
                        let alias_value_name = self.lexical_map.insert(value_name);
```

sway-core/src/optimize.rs
```diff
@@ -2148,6 +2148,124 @@ impl FnCompiler {
...
+
+                        // Create an array of `b256` that will hold the value to store into storage
+                        // or the value loaded from storage. The array has to fit the whole type.
+                        let b256_array_type = Type::Array(Aggregate::new_array(
+                            context,
+                            Type::B256,
+                            ir_type_size_in_bytes(context, r#type) / 32 + 1,
```
otrho, line 2168:
```suggestion
                            (ir_type_size_in_bytes(context, r#type) + 31) / 32,
```

otrho (changes requested):

sway-core/src/optimize.rs
```diff
@@ -2148,6 +2148,124 @@ impl FnCompiler {
...
+
+                        fn add_to_b256(x: fuel_types::Bytes32, y: u64) -> fuel_types::Bytes32 {
+                            let x = bigint::uint::U256::from(*x);
+                            let y = bigint::uint::U256::from(y);
+                            let res: [u8; 32] = (x + y).into();
+                            fuel_types::Bytes32::from(res)
+                        }
+
+                        let mut iter = 0;
+                        loop {
```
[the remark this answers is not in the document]
mohammadfawaz: Ok cleaned up all of that... It was ugly for sure. Does the new version look better?

otrho (approved):

Issue #1795, opened 2022-06-01
Unable to access fields on `self`


nfurfaro:
In `sway-lib-std::b512` we have this impl trait which works fine:
```
impl core::ops::Eq for B512 {
    fn eq(self, other: Self) -> bool {
        (self.bytes)[0] == (other.bytes)[0] && (self.bytes)[1] == (other.bytes)[1]
    }
}
```

However, trying to do this fails:
```
pub trait From {
    fn from(h: b256, l: b256) -> Self;
} {
    fn into(self) -> (b256, b256) {
        ((self.bytes)[0], (self.bytes)[1])
    }
}
```

The error is:
```
sway/sway-lib-std on  furnic/bug-repro [$!]
❯ forc build
  Compiled library "core".
error
  --> /Users/furnic/dev/fuel/sway/sway-lib-std/src/b512.sw:16:11
   |
14 |
15 |     fn into(self) -> (b256, b256) {
16 |         ((self.bytes)[0], (self.bytes)[1])
   |           ^^^^ This is a Self, not a struct. Fields can only be accessed on structs.
17 |     }
18 | }
   |
____

error
  --> /Users/furnic/dev/fuel/sway/sway-lib-std/src/b512.sw:16:28
   |
14 |
15 |     fn into(self) -> (b256, b256) {
16 |         ((self.bytes)[0], (self.bytes)[1])
   |                            ^^^^ This is a Self, not a struct. Fields can only be accessed on structs.
17 |     }
18 | }
   |
____

  Aborting due to 2 errors.
Error: Failed to compile std
```

To reproduce, checkout the `furnic/bug-repro` branch and build the stdlib with `forc build`.

cc @emilyaherbert

emilyaherbert:
Hi @nfurfaro, taking a look at this again, this isn't actually a bug at all. You are trying to add this trait:

```rust
pub trait From {
    fn from(h: b256, l: b256) -> Self;
} {
    fn into(self) -> (b256, b256) {
        ((self.bytes)[0], (self.bytes)[1])
    }
}
```

which has the method `from` to be defined for any `impl From for X`, and the method `into` which is automatically defined for any `impl From for X`. However, based on this implementation, how does `into` know what `X` is? i.e. how would we know whether `self` of type `X` has the field `bytes`, or is even a struct at all?

[...]

PR #1827, opened 2022-06-02
Impl `into` for `B512`

### LINKED ISSUE #1792: Add into() for B512
As per the comments:
>pub trait From {
    fn from(h: b256, l: b256) -> Self;
} {
    // @todo add into() when tuples land, as it would probably return 2 b256 values
    // fn into() {...}
}

nfurfaro:
All in the title. 

closes #1792 
ref #1795


sway-lib-std/src/storage.sw
```diff
@@ -92,7 +92,8 @@ pub fn get<T>(key: b256) -> T {
     }
 }
 
-pub struct StorageMap<K, V> { }
+pub struct StorageMap<K, V> {
```
adlerjohn, line 95: Remove drive-by formatting

otrho (approved):

Issue #1825, opened 2022-06-02
Swayswap contract ID nondeterminism when switching to use `U128`


mohammadfawaz:
This version of Swayswap: https://github.com/FuelLabs/swayswap/blob/d754d348d8baee491640f00209a66ee8f7f4b495/contracts/exchange_contract/src/main.sw

Likely nondeterminism in the bytecode.

PR #1829, opened 2022-06-02
Make match var return name deterministic

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/1825

This was causing some nondeterminism in Swayswap.

The fix is similar to https://github.com/FuelLabs/sway/pull/1542/files


otrho:
Just thinking -- you've made it atomic, for the future when we might have threads?  In that case it would no longer be deterministic, different threads might use it at different times.  So, having a static counter for now is good, but not guaranteed to be OK in a threaded compiler.

mohammadfawaz:
> Just thinking -- you've made it atomic, for the future when we might have threads? In that case it would no longer be deterministic, different threads might use it at different times. So, having a static counter for now is good, but not guaranteed to be OK in a threaded compiler.

Yeah this is second time we do this. I didn't like too much the first and I don't like it here, but I don't really know what else to do that would also work in a multi-threaded setting 🤔

PR #1830, opened 2022-06-02
Report type name in the error when a type is unknown

### LINKED ISSUE #1740: Confusing error message when you use a `struct` or `enum` that hasn't been declared to imported
There is a confusing error message when you use a `struct` or `enum` that hasn't been declared to imported.

For example this Sway code:

```rust
use std::chain::auth::*;
//use std::result::*;

fn bogus() -> Sender {
  let sender: Result<Sender, AuthError> = msg_sender();
  sender.unwrap()
}
```

generates this error message:

```bash
error
  --> match_expressions_inside_generic_functions/src/main.sw:9:15
   |
 7 | 
 8 | fn bogus() -> Sender {
 9 |   let sender: Result<Sender, AuthError> = msg_sender();
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^ Unknown type name.
10 |   sender.unwrap();
11 | }
   |
```

A better error message would be one that specifies that `Result` is an unknown type name and that has the underline just under the `Result` symbol.

AlicanC:
Resolves #1740.

Before:
```bash
error
  --> match_expressions_inside_generic_functions/src/main.sw:9:15
   |
 7 | 
 8 | fn bogus() -> Sender {
 9 |   let sender: Result<Sender, AuthError> = msg_sender();
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^ Unknown type name.
```

After:
```bash
error
  --> match_expressions_inside_generic_functions/src/main.sw:9:15
   |
 7 | 
 8 | fn bogus() -> Sender {
 9 |   let sender: Result<Sender, AuthError> = msg_sender();
   |               ^^^^^^ Unknown type name "Result".
```


emilyaherbert:
It looks like this found an error in how imports are done in the Sway examples---good work!

adlerjohn:
@emilyaherbert is there a follow-up on fixing the bug?

emilyaherbert:
> is there a follow-up on fixing the bug?

Yes---@AlicanC on this PR, can you add the appropriate import for `Address`, here? https://github.com/FuelLabs/sway/blob/4790c907c5cfd153df41bfb0cb8e7a73fd8981ad/examples/signatures/src/main.sw#L8 and then resubmit the PR. Thanks!

AlicanC:
Even more was caught after. Fixed all and marked ready.

emilyaherbert (approved):

test/src/sdk-harness/test_artifacts/reentrancy_target_contract/src/main.sw
```diff
@@ -23,7 +23,7 @@ impl Target for Contract {
         if is_reentrant() {
             true
         } else {
-            let result: Result<Sender, AuthError> = msg_sender();
+            let result: Result<Identity, AuthError> = msg_sender();
```
adlerjohn, line 26: Why is this change needed? Is the SDK harness not testing this?

adlerjohn (approved):

test/src/sdk-harness/test_artifacts/reentrancy_target_contract/src/main.sw
AlicanC: Forgot to mention: as a side effect this PR also exposes some errors that were being thrown out before. This should be one of those.

PR #1831, opened 2022-06-02
Actually use some of the existing traits

emilyaherbert:
Previously, we had these traits:
https://github.com/FuelLabs/sway/blob/7ba15502fe5df004994063eab8a716f3dba6e8cf/sway-core/src/types/friendly_type_string.rs#L1
https://github.com/FuelLabs/sway/blob/7ba15502fe5df004994063eab8a716f3dba6e8cf/sway-core/src/types/json_abi_string.rs#L1
https://github.com/FuelLabs/sway/blob/7ba15502fe5df004994063eab8a716f3dba6e8cf/sway-core/src/types/to_json_abi.rs#L1

that we were only using for `TypeId`s. But, there are many types in the compiler that had the same function, just hardcoded. This PR refactors the `type_engine` module to expose these traits and implements these traits in replacement of the previously hardcoded functions. This PR produces no change functionally, but makes the compiler more readable and maintainable---i.e. if we want multiple types to have the same method with the same signature, it should be a method from a trait so that there is no code drift.

In implementing this, I ended up refactoring a number of files to expose dependencies. I don't think the changes to how dependencies are exposed in individual files is, like, the best way to do it ever (neither was the previous way dependencies were exposed). BUT for some reason, organizing the dependencies this way allows clippy to reveal that certain variants of `ResolvedType` are unused!
https://github.com/FuelLabs/sway/blob/7ba15502fe5df004994063eab8a716f3dba6e8cf/sway-core/src/type_engine/resolved_type.rs#L16

If we intend for these to be unused, we should remove them from the variants. If we do _not_ intend for them to be unused, we should figure out why that is and fix it!


sway-core/src/type_engine/resolved_type.rs
```diff
@@ -13,29 +13,35 @@ pub enum ResolvedType {
     Unit,
     Byte,
     B256,
+    #[allow(dead_code)]
```
otrho, line 16: What's going on with all these `dead_code`s?

otrho (commented):
utACK, looks like a nice cleanup!  Reminds me that we should use `Spanned` wherever we can in `sway-core` too.

sway-core/src/type_engine/resolved_type.rs
emilyaherbert: That's what I thought! The way in which I reorganized the dependencies allowed clippy to reveal that these cases were unused! We should either decide to use them, or remove them, IMO.

otrho: Oh, wow, OK.  So, hrmm, do we think we'll need in the future?  Either way, if we do, it'll be easy to just reintroduce them, they'd not that complicated.  I vote they just go now.

PR #1835, opened 2022-06-03
Bump to v0.15.0

mohammadfawaz:
Waiting on:
* [x] https://github.com/FuelLabs/sway/pull/1834
* [x] https://github.com/FuelLabs/sway/pull/1829
* [x] https://github.com/FuelLabs/sway/pull/1826

Deferring https://github.com/FuelLabs/sway/pull/1669 to the next release as it's still being worked on.


Braqzen:
Can we add #1852 to the list?

mohammadfawaz:
> Can we add #1852 to the list?

Would certainly be nice to have, but I don't think it's being worked on just yet? I'm thinking we release `0.15.0` now because it has some critical bug fixes and I don't want to delay it further. We can always cut a new release once this is done.

PR #1836, opened 2022-06-03
Add forc feature for overriding packages in the package graph, akin to cargo's [patch] feature

### LINKED ISSUE #1784: Add forc feature for overriding packages in the package graph, akin to cargo's `[patch]` feature
## Motivation

Here's a great list of motivations from [the cargo book](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html):

> The desire to override a dependency can arise through a number of scenarios. Most of them, however, boil down to the ability to work with a crate before it's been published to [crates.io](https://crates.io/). For example:
> 
>    - A crate you're working on is also used in a much larger application you're working on, and you'd like to test a bug fix to the library inside of the larger application.
>    - An upstream crate you don't work on has a new feature or a bug fix on the master branch of its git repository which you'd like to test out.
>    - You're about to publish a new major version of your crate, but you'd like to do integration testing across an entire package to ensure the new major version works.
>    - You've submitted a fix to an upstream crate for a bug you found, but you'd like to immediately have your application start depending on the fixed version of the crate to avoid blocking on the bug fix getting merged.

I'd imagine having something like this would be a big win for app devs who could more conveniently patch in development branches of `core`, `std` or other dependencies when wanting to test WIP upstream changes in their apps.

## Approach

The implementer should first ensure they are familiar with cargo's `[patch]` feature and its capabilities and the motivations above.

Work on this will for the most part be contained within the `forc-pkg` crate. This will include:

1. Updating the `Manifest` with support for a `[patch]` section.
2. Handling for the `[patch]` table during construction of the `BuildPlan`. This will involve looking at the `BuildPlan` constructors while considering the `forc::ops::forc_build::build` function where the `BuildPlan` gets created.
3. Replacement should likely occur *while* constructing the package graph. This will include when the graph is constructed from a lock file, or while we're fetching dependencies and constructing the graph from scratch.
4. Document new feature in the forc manifest reference within The Sway Book.

kayagokalp:
~This is still WIP, opening up for visibility.~

- [x] Support for patching while creating the `BuildPlan` from the manifest file.
- [x] Documentation.
- [x] Support for patching while creating the `BuildPlan` from the lock file.
- [x] Address #1869

## About this PR
implements and closes #1784.

- Introduces `[package]` table to the manifest file.
- While creating a `source` from a `dependency` check for patches and applies them if there are any.

### Testing

**Patching `std` with a different branch**

1. run `forc build` 
2. add patch to the Forc.toml
3. run `forc build`
4. remove patch from the Forc.toml
5. run `forc build`

- After step 1 in the Forc.lock you should be seeing the master branch of the sway repo (for std)
- After step 3 in the Forc.lock you should be seeing the kayagokalp/1620 branch of the sway repo (for std)
- After step 5 in the Forc.lock you should be seeing the master branch of the sway repo (for std)

You can use the Forc.toml provided below.

```
[project]
authors = ["user"]
entry = "main.sw"
organization = "Fuel_Labs"
license = "Apache-2.0"
name = "parent"
implicit-std = false

[dependencies]
std = { git = "https://github.com/fuellabs/sway", branch = "master"} 
```

and to add patch in step2, you can use:

```
[patch.'https://github.com/fuellabs/sway']
std = { git = "https://github.com/fuellabs/sway", branch = "kayagokalp/1620"} 
```

or if you would like to try patching with a path:

```
[patch.'https://github.com/fuellabs/sway']
std = { path = "/path/to/std" } 
```


kayagokalp:
Currently, if there is a `patch` supplied for a `dependency` in the manifest file, it creates a diff between the lock and the manifest file since the lock was created without (if it is previously created) the patches. So the patches are automatically applied by `apply_pkg_diff()`. I marked `Support for patching while creating the BuildPlan from the lock file.` completed in the PR explanation but if we explicitly need the patches before `apply_pkg_diff()` I can take a look to implement it like that.

docs/src/forc/manifest_reference.md
```diff
@@ -99,3 +101,32 @@ Note that providing the corresponding cli options (like `--print-finalized-asm`)
...
+[project]
+authors = ["user"]
+entry = "main.sw"
+organization = "Fuel_Labs"
+license = "Apache-2.0"
+name = "wallet_contract"
+implicit-std = false
+
+[dependencies]
+std = { git = "https://github.com/fuellabs/sway", branch = "master"}
```
adlerjohn, line 119:
```suggestion
std = { git = "https://github.com/fuellabs/sway", branch = "master" }
```

docs/src/forc/manifest_reference.md
```diff
@@ -99,3 +101,32 @@ Note that providing the corresponding cli options (like `--print-finalized-asm`)
...
+organization = "Fuel_Labs"
+license = "Apache-2.0"
+name = "wallet_contract"
+implicit-std = false
+
+[dependencies]
+std = { git = "https://github.com/fuellabs/sway", branch = "master"} 
+
+[patch.'https://github.com/fuellabs/sway']
+std = { git = "https://github.com/fuellabs/sway", branch = "test"}
```
adlerjohn, line 122:
```suggestion
std = { git = "https://github.com/fuellabs/sway", branch = "test" }
```

docs/src/forc/manifest_reference.md
```diff
@@ -99,3 +101,32 @@ Note that providing the corresponding cli options (like `--print-finalized-asm`)
...
+
+[patch.'https://github.com/fuellabs/sway']
+std = { git = "https://github.com/fuellabs/sway", branch = "test"} 
+```
+
+In the example above, `std` is added as a git dependency and then patched with the same repo's `test` branch. You can also patch git dependencies with dependencies defined with a path.
+
+```toml
+[patch.'https://github.com/fuellabs/sway']
+std = { path = "/path/to/local_std_version"}
```
adlerjohn, line 129:
```suggestion
std = { path = "/path/to/local_std_version" }
```

forc-pkg/src/pkg.rs
```diff
@@ -1034,6 +1069,7 @@ fn dep_to_source(pkg_path: &Path, dep: &Dependency) -> Result<Source> {
                             either `branch`, `tag` or `rev`"
                     ),
                 };
+                //We don't have any patch for this dependency
```
adlerjohn, line 1072:
```suggestion
                // We don't have any patch for this dependency
```

kayagokalp:
Realized a bug when the lock file is removed will be taken care of shortly in the meantime converting this to a draft.

kayagokalp:
As it turns out, the bug was not with the code but with myself...

I was providing the wrong path for the patch but couldn't understand what is happening because lack of a good error message there. So I improved the error message. Also removed unused `Patch` enum. I changed it into a type declaration

docs/src/forc/manifest_reference.md
```diff
@@ -99,3 +101,32 @@ Note that providing the corresponding cli options (like `--print-finalized-asm`)
...
+[project]
+authors = ["user"]
+entry = "main.sw"
+organization = "Fuel_Labs"
+license = "Apache-2.0"
+name = "wallet_contract"
+implicit-std = false
+
+[dependencies]
+std = { git = "https://github.com/fuellabs/sway", branch = "master" }
```
adlerjohn, line 119: Does patching work with implicit `std` too?

kayagokalp: Yes, it works with implicit `std` too. To be more explicit and descriptive about what we are patching I created the doc like this. (i.e, explicitly added the dependency that I am patching). But I also realize that in this way the example is not in its most "default" form. 🤔

adlerjohn: I would prefer two subsections, one for patching a library that's not `std` or `core`, and one specifically for patching the implicit `std`.

forc-pkg/src/pkg.rs
```diff
@@ -732,6 +733,35 @@ pub(crate) fn fetch_deps(
...
 
+fn apply_patch(
+    name: &String,
+    source: &Source,
+    manifest: &Manifest,
+    parent_path: &Path,
+) -> Result<Source> {
+    let mut patches = manifest.patches();
+    match source {
+        //Check if the patch is for a git dependency.
```
mitchmindtree, line 744:
```suggestion
        // Check if the patch is for a git dependency.
```

forc-pkg/src/pkg.rs
```diff
@@ -732,6 +733,35 @@ pub(crate) fn fetch_deps(
...
+        Source::Git(git) => {
+            //Check if we got a patch for the git dependency.
```
mitchmindtree, line 746:
```suggestion
            // Check if we got a patch for the git dependency.
```

forc-pkg/src/pkg.rs
```diff
@@ -732,6 +733,35 @@ pub(crate) fn fetch_deps(
...
+            if let Some(source_patches) = patches.find(|patch| *patch.0 == git.repo.to_string()) {
```
mitchmindtree, line 747: Rather than iterating over all of the entries, `manifest.patch` is already a `BTreeMap` with log(n) lookup. We should use something like `manifest.patch.as_ref().and_then(|patches| patches.get(&git.repo.as_str()))` here.

mitchmindtree (changes requested):
Haven't tested but looking good - nice work! Just a few small things to address.

mitchmindtree:
@kayagokalp would you mind merging master? It's just been a couple days is all - then will happily approve :)

mitchmindtree:
@kayagokalp actually that said, this could probably also use a test. Would you mind adding a small test for this under `should_pass/forc`? It could be as simple as patching the implicit std URL with the local copy of `std` or something along these lines.

kayagokalp:
The introduction of #1790 seems to have an impact on this one, as there is a check for the dependency from its parent manifest file. If the dependency is coming from the patch it is not in the dependency section. I will be fixing this shortly and pushing the solution with the test I added.

forc-pkg/src/pkg.rs
```diff
@@ -827,10 +828,25 @@ pub fn graph_to_path_map(
...
                     })?;
-                let rel_dep_path = detailed
-                    .path
-                    .as_ref()
-                    .ok_or_else(|| anyhow!("missing path info for dependency: {}", dep.name))?;
+                // Check if there is a patch for this dep
+                let patch = parent_manifest
+                    .patches()
+                    .find(|patches| patches.1.contains_key(&dep_name))
+                    .and_then(|patches| patches.1.get(&dep_name));
```
mitchmindtree, line 835: `find` and `and_then` used in succession can often be replaced with `find_map` - not important, just a tip!

test/src/e2e_vm_tests/test_programs/should_pass/forc/dependency_patching/.gitignore
```diff
@@ -0,0 +1,2 @@
+out
+target
```
mitchmindtree, line 2: I think these should be covered by the top-level `.gitignore`?

mitchmindtree (commented):
Nice one kaya, that test is perfect! Almost there

forc-pkg/src/pkg.rs
kayagokalp: Oh, thanks :) that looks better

test/src/e2e_vm_tests/test_programs/should_pass/forc/dependency_patching/.gitignore
kayagokalp: yep, couldn't decide what was our general approach here while adding tests under e2e, deleting it 👍

mitchmindtree (approved):
Nice one @kayagokalp, this will be super handy :rocket:

kayagokalp:
There was a conflict so I had the merge manually, once again I am asking for your review 😄

adlerjohn (commented):
There should be a `test.toml` file I think: https://github.com/FuelLabs/sway/blob/bdfb7449cac1e80ee4258abf1882c3caec99111c/test/src/e2e_vm_tests/test_programs/README.md#testtoml

kayagokalp:
Oh sorry, added that locally but forget to include it in the commit. Adding rn.

test/src/e2e_vm_tests/test_programs/should_pass/forc/dependency_patching/test.toml
```diff
@@ -0,0 +1,3 @@
+category = "run"
+expected_result = { action = "return", value = 0 }
+validate_abi = true
```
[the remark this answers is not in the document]
kayagokalp: Hmm, that's correct. Maybe just providing `category = "compile"` would be enough for the test.toml.

mitchmindtree (approved):

adlerjohn (approved):
utACK

PR #1838, opened 2022-06-03
Generate ABI JSONs for scripts

### LINKED ISSUE #1656: Generate ABI from scripts
We currently only generate ABIs when building contracts. We should generate them when building scripts too.

If the generated JSON looked like a contract with just a `main` fn, then it would work really well with the code we already have on SDKs.

Blocked by: https://github.com/FuelLabs/sway/issues/845

AlicanC:
- [Use assert_matches to show errors during tests](https://github.com/FuelLabs/sway/pull/1838/commits/3f5ad93442236d6590757f18fde4d9614f784fc0)

  The previous `assert!` was hiding the actual error so I switched to `assert_matches!`.

  Before:
  ```
  assertion failed: crate::e2e_vm_tests::harness::test_json_abi(name).is_ok()
  ```

  After:
  ```
  assertion failed: `Err(Mismatched ABI JSON output.)` does not match `Ok(_)`
  ```

- [Generate ABI JSONs for scripts](https://github.com/FuelLabs/sway/pull/1838/commits/bf5602ff2bd779f7c858e7a078afb6826d6c470f)

  Fixes #1656.


emilyaherbert (approved):

adlerjohn (approved):
utACK :shipit:

Issue #723, opened 2022-01-29
Refactor the E2E test driver.


otrho:

otrho:
@m-fawaz and I have used the LLVM testing tools `lit` and `FileCheck`, both of which have Rust crates and might be a good fit for our testing.

`lit` is a driver, not dissimilar to what I was describing above, automatically discovering tests and their metadata.

`FileCheck` is a tool for checking changes in files, by inlining what we expect the output to be within the input itself.  This is particularly useful for IR, where the input and output are both IR... it might not be quite as useful for our general E2E testing though.  Need to confirm.

https://crates.io/crates/lit
https://docs.rs/filecheck/latest/filecheck/

PR #1840, opened 2022-06-03
Error on multiple function definitions

### LINKED ISSUE #751: Compiler should error out when the same function is defined multiple times
Example:
```
script;
fn foo() {}
fn foo() {}
fn main() {}
```
Even this should error out:
```
script;
fn foo() {}
fn foo<T>() {}
fn main() {}
```

vaivaswatha:
Closes #751


otrho:
Looks good!  I'm thinking it might be an idea to create a `should_pass` test which puts the same named function into very close but distinct scopes.  Like putting one in a local library with `dep` & `use`, and `main.sw`, or nested functions (although I'm not sure if Sway supports nested functions yet), or in an ABI or `impl` and global scope.

vaivaswatha:
> Looks good! I'm thinking it might be an idea to create a `should_pass` test which puts the same named function into very close but distinct scopes. Like putting one in a local library with `dep` & `use`, and `main.sw`, or nested functions (although I'm not sure if Sway supports nested functions yet), or in an ABI or `impl` and global scope.

:+1: I'll do that.

vaivaswatha:
> Looks good! I'm thinking it might be an idea to create a `should_pass` test which puts the same named function into very close but distinct scopes. Like putting one in a local library with `dep` & `use`, and `main.sw`, or nested functions (although I'm not sure if Sway supports nested functions yet), or in an ABI or `impl` and global scope.

Thanks @otrho . I've added this now.

emilyaherbert (changes requested):
Awesome job 👍  can you add a test case where the contract imports a function? So something like:

```rust
contract;

dep testlib::foo;

abi TestContr {
    fn foo();
}

fn foo() {}

impl TestContr for Contract {
    fn foo() {
       foo();
    }
}

fn main() {

}
```

vaivaswatha:
> Awesome job +1 can you add a test case where the contract imports a function? So something like:
> 
> ```rust
> contract;
> 
> dep testlib::foo;
> 
> abi TestContr {
>     fn foo();
> }
> 
> fn foo() {}
> 
> impl TestContr for Contract {
>     fn foo() {
>        foo();
>     }
> }
> 
> fn main() {
> 
> }
> ```

Thank you @emilyaherbert . 
> `dep testlib::foo;`

:arrow_up:  didn't work, but :arrow_down: did, so I added that.
```
dep testlib;
use testlib::foo
```

otrho (approved):

mohammadfawaz:
Merge conflicts! Looks good otherwise :)

mohammadfawaz (changes requested):
The change looks great! Just a small nit on the tests.

emilyaherbert:
> starting to wonder if we should refactor this file so it gets less conflicts, haha. Not sure how we'd do that, though.

Instead of comparing the result of the test in the `mod.rs` file, we could compare it to the contents of a different file. Then we could just blindly iterate through the contents of the test dir. This is similar to how the JSON ABI output is tested: https://github.com/FuelLabs/sway/blob/e8f10e7d66118e2dcc8f0fdcef5ecf3c4a30c830/test/src/e2e_vm_tests/harness.rs#L125

vaivaswatha:
All suggested changes done.

otrho:
> Instead of comparing the result of the test in the `mod.rs` file, we could compare it to the contents of a different file.

This is exactly what I proposed months ago in #723 @emilyaherbert and as `DEITY` is my witness I'm going to start working on it.  These conflicts are just too annoying.

mohammadfawaz (approved):

emilyaherbert:
> This is exactly what I proposed months ago in https://github.com/FuelLabs/sway/issues/723

Great proposal! I am in favor of doing this 😄

mohammadfawaz:
@emilyaherbert I think this needs your approval to go through :)

emilyaherbert (approved):

PR #1842, opened 2022-06-03
Fix how the state index metadata is inlined

mohammadfawaz:
Calling functions that contain calls to `insert()` and `get()` from `StorageMap` because of a bug of how inlining of the state index metadata was being inlined.

### DIFF
--- a/sway-ir/src/optimize/inline.rs
+++ b/sway-ir/src/optimize/inline.rs
@@ -203,6 +203,7 @@ fn inline_instruction(
     if let ValueContent {
         value: ValueDatum::Instruction(old_ins),
         span_md_idx,
+        state_idx_md_idx: current_state_idx_md_idx,
         ..
     } = context.values[instruction.0].clone()
     {
@@ -240,7 +241,7 @@ fn inline_instruction(
                     .collect::<Vec<Value>>()
                     .as_slice(),
                 span_md_idx,
-                state_idx_md_idx,
+                current_state_idx_md_idx,
             ),
             Instruction::Cmp(pred, lhs_value, rhs_value) => new_block.ins(context).cmp(
                 pred,
--- a/sway-ir/tests/inline/get_storage_key.ir
+++ b/sway-ir/tests/inline/get_storage_key.ir
@@ -3,13 +3,13 @@
 // regex: LABEL=[[:alpha:]0-9]+:
 
 script {
-    fn anon_0() -> b256 {
+    fn return_storage_key_wrapper() -> b256 {
         entry:
-        v0 = call anon_1(), !2, !3
+        v0 = call return_storage_key(), !2, !3
         ret b256 v0
     }
 
-    fn anon_1() -> b256 {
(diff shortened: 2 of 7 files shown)


sway-ir/src/optimize/inline.rs
```diff
@@ -203,6 +203,7 @@ fn inline_instruction(
     if let ValueContent {
         value: ValueDatum::Instruction(old_ins),
         span_md_idx,
+        state_idx_md_idx: current_state_idx_md_idx,
```
Braqzen, line 206: This is my first time looking at this code so I have no idea what I am looking at.
These abbreviations may be something completely obvious to someone who is not ignorant in this area however that is not the case for newbies.

I generally dislike abbreviations for this reason and think that alternative names should be selected if they are too long when verbose.

sway-ir/tests/inline/get_storage_key.ir
```diff
@@ -4,6 +4,12 @@
 
 script {
     fn anon_0() -> b256 {
```
Braqzen, line 6: I can kind of tell what is going on here but the function name could be improved. Similar to the other `anon` functions

test/src/sdk-harness/test_projects/storage_map/src/main.sw
```diff
@@ -102,13 +103,21 @@ abi StorageMapTest {
     3]) -> u64;
 }
 
+fn insert_into_u64_to_bool_map_inner(key: u64, value: bool) {
+    storage.map1.insert(key, value);
+}
+
+fn get_from_u64_to_bool_map_inner(key: u64) -> bool {
+    storage.map1.get(key)
+}
```
Braqzen, line 112: When using functions outside of the trait I prefer the convention of "underscore then name of the function" e.g. `_insert_into_u64_to_bool_map`

test/src/e2e_vm_tests/test_programs/should_pass/require_contract_deployment/get_storage_key_caller/src/main.sw
```diff
@@ -26,5 +26,11 @@ fn main() -> u64 {
 
     assert(f3 != f4);
 
+    let(cf1, cf2, cf3, cf4) = caller.from_callers();
+    assert(f1 == cf1);
+    assert(f2 == cf2);
+    assert(f3 == cf3);
+    assert(f4 == cf4);
```
Braqzen, line 33: The names could certainly be a little more descriptive for anyone wondering why they should be equal

Braqzen (commented):
I will defer to anyone that understands the `sway-ir` code. 
My comments are only regarding code quality otherwise I'd approve.

sway-ir/tests/inline/get_storage_key.ir
mohammadfawaz: So the IR generator chooses those, but you're absolutely right: there is no need for a test to also have non-descriptive names like that.

emilyaherbert (approved):

PR #1843, opened 2022-06-03
Adds `from_dir` to new formatter

eureka-cpu:
This just cleans up the previous `from_opts` on `Formatter` and the `match` statement that was in the `run` function of the `forc-fmt-v2` plugin into a `from_dir` function on `Formatter` instead.


sway-fmt-v2/src/lib.rs
```diff
@@ -6,7 +6,7 @@
 #![allow(dead_code)]
 pub mod config;
 mod constants;
-mod error;
+pub mod error;
```
adlerjohn, line 9: Is this needed to be public?

eureka-cpu: got leftover from when I was fleshing out the change!

adlerjohn (commented):
looks fine to me, will defer to @mitchmindtree or @JoshuaBatty

mitchmindtree (approved):

PR #1845, opened 2022-06-03
forc run simulate txs

ra0x3:


ra0x3:
@adlerjohn 
- Not really sure how to test this
- Would I just `forc init --contract` a new project and `forc run --simulate-tx` ?
  - Getting a `Error: Response errors; Unknown argument "utxoValidation" on field "dryRun" of type "Mutation".` error when taking this route
- Wondering basically if I have to do anything special  test this and ensure it's working

adlerjohn:
That's how you would test it, yes. I think that error is popping up because your version of `fuel-core` is outdated.

forc/src/ops/forc_run.rs
```diff
@@ -51,27 +51,29 @@ pub async fn run(command: RunCommand) -> Result<Vec<fuel_tx::Receipt>> {
         TxParameters::new(command.byte_price, command.gas_limit, command.gas_price),
     );
 
+    let node_url = match &manifest.network {
+        Some(network) => &network.url,
+        _ => &command.node_url,
+    };
```
mitchmindtree, line 57: I think we want the `command.node_url` to have priority here. E.g. I can imagine the user having a URL specified in their manifest, but wanting to temporarily test another by passing a flag. It also might be strange to have the flag just be ignored in the case that a URL is specified in the manifest.

mitchmindtree (commented):
Looking good! Just one small suggestion.

forc/src/ops/forc_run.rs
JoshuaBatty: Yeah I agree. We want the flag to take priority here if one was passed in.

ra0x3: @mitchmindtree @JoshuaBatty 
- I see the `command.node_url` has a `default_value`, so it will always be present. How are we handling that in this case? 
- Always just use `command.node_url` when `command.simulate_tx` is passed? 
- Or maybe use `manifest.network.url` in the case that `command.simulate_tx` is passed and `command.node_url` is the default value?

mitchmindtree: Ahh good spot!

I think we should make `command.node_url` an `Option`, move that default URL out into a constant, and then work out the target node URL by:

1. Checking if `command.node_url` is `Some`,
2. if `None`, use `manifest.network.url`,
3. If that's also `None`, fallback to the default URL (the one that is currently specified [here](https://github.com/FuelLabs/sway/blob/98f261fcc81b14fbbc742e1974460517d11bc653/forc/src/cli/commands/run.rs#L26)).

How does that sound?

forc/src/cli/commands/run.rs
```diff
@@ -90,6 +90,12 @@ pub struct Command {
     /// needs to be updated, Forc will exit with an error
     #[clap(long)]
     pub locked: bool,
+
+    /// Execute the transaction and return the final mutated transaction along with receipts
+    /// (which includes whether the transaction reverted or not). The transaction is not inserted
+    /// in the node's view of the blockchain, (i.e. it does not affect the chain state).
+    #[clap(long = "simulate-tx")]
```
mitchmindtree, line 97:
```suggestion
    #[clap(long)]
```
^ I think this means the same

forc/src/cli/commands/run.rs
```diff
@@ -90,6 +90,12 @@ pub struct Command {
...
+    pub simulate_tx: bool,
```
adlerjohn, line 98: Instead of `simulate_tx`, which is redundant (since you're doing something with a tx anyways), how about `simulate`?

adlerjohn (approved):

mitchmindtree (approved):
:man_dancing:

PR #1853, opened 2022-06-05
Made transfer_to a little more generic

### LINKED ISSUE #1851: Make the `transfer_to()` from `token.sw` in the `std-lib` non-specific to the `contract_id()`
Currently the function uses the `contract_id()` to automatically transfer the asset - being itself.

I find myself transferring variable assets frequently and thus [transfer_to()](https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/token.sw#L46) should be deprecated to allow any assets to be transferred rather than just the contract that it is being called from.

Here is the new function to take its place.

```rust
fn transfer(to: Identity, asset: ContractId, amount: u64) {
    match to {
        Identity::Address(address) => {
            transfer_to_output(amount, asset, address);
        },
        Identity::ContractId(address) => {
            force_transfer_to_contract(amount, asset, address);
        }
    }
}
```

Braqzen:
# Summary

Changes made:

- Renamed `transfer_to()` to `transfer()`
- `transfer()` now takes an `asset` of type `ContractId` instead of having a hardcoded `contract_id()` embedded

# Related

Closes #1851

### DIFF
--- a/sway-lib-std/src/token.sw
+++ b/sway-lib-std/src/token.sw
@@ -11,7 +11,7 @@ use ::identity::Identity;
 /// Mint `amount` coins of the current contract's `asset_id` and transfer them to `destination` by calling either force_transfer_to_contract() or transfer_to_output(), depending on the type of `Identity`.
 pub fn mint_to(amount: u64, recipient: Identity) {
     mint(amount);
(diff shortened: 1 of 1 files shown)


nfurfaro (approved):
LGTM.

PR #1854, opened 2022-06-05
Updated section regarding enums

### LINKED ISSUE #1812: Edit the sway book to show how to use an `enum` of `enums`
# Info

In #1811 I have used the incorrect syntax to use an `enum` from inside of an `enum`.
The [sway book](https://fuellabs.github.io/sway/latest/basics/structs_and_tuples.html#enums) shows examples of an `enum` containing a `struct` but not an `enum` containing another `enum`.

Braqzen:
# Summary

Changes made:

- Renamed file to reflect usage of enums (not just structs and tuples)
- Split enums into clear sections
- Added enum of enums section

# Related

Closes #1812

### DIFF
--- a/docs/src/basics/index.md
+++ b/docs/src/basics/index.md
@@ -8,7 +8,7 @@ Sway basics.
 - [Built-in Types](./built_in_types.md)
 - [Blockchain Types](./blockchain_types.md)
 - [Functions](./functions.md)
-- [Structs and Tuples](./structs_and_tuples.md)
+- [Structs, Tuples & Enums](./structs_tuples_and_enums.md)
 - [Methods and Associated Functions](./methods_and_associated_functions.md)
 - [Control Flow](./control_flow.mdz)
 - [Comments and Logging](./comments_and_logging.md)
--- a/docs/src/basics/structs_tuples_and_enums.md
+++ b/docs/src/basics/structs_tuples_and_enums.md
@@ -1,4 +1,4 @@
-# Structs and Tuples
+# Structs, Tuples & Enums
 
 ## Structs
 
@@ -59,6 +59,7 @@ let y: bool = my_tuple.1;
 Tuples can also contain tuples within themselves, and be used in destructing syntax to declare multiple values at once.
 
 Common usecases for tuples are returning multiple values from a function, packing parameters into a function, or storing a series of related values.
+
 ## Enums
 
 _Enumerations_, or _enums_, are also known as _sum types_. An enum is a type that could be one of several variants. To declare an enum, you enumerate all potential variants. Let's look at _enum declaration syntax_:
@@ -73,7 +74,23 @@ enum Color {
 }
 ```
 
-Here, we have defined five potential colors. Each enum variant is just the color name. As there is no extra data associated with each variant, we say that each variant is of type `()`, or unit. It is also possible to have an enum variant contain extra data. Take a look at this more substantial example, which combines struct declarations with enum variants:
+Here, we have defined five potential colors. Each enum variant is just the color name. As there is no extra data associated with each variant, we say that each variant is of type `()`, or unit.
+
+To instantiate a variable with the value of an enum the syntax is
+
+> **Note**
+> enum instantiation does not require the `~` tilde syntax
+
+```sway
+fn main() {
+    let blue = Color::Blue;
+    let silver = Color::Silver;
+}
+```
+
+### Enums of Structs
+
+It is also possible to have an enum variant contain extra data. Take a look at this more substantial example, which combines struct declarations with enum variants:
 
 ```sway
 use std::collections::Vec;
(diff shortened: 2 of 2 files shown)


docs/src/basics/structs_tuples_and_enums.md
```diff
@@ -1,4 +1,4 @@
-# Structs and Tuples
+# Structs, Tuples & Enums
```
adlerjohn, line 1:
```suggestion
# Structs, Tuples, and Enums
```

docs/src/basics/structs_tuples_and_enums.md
```diff
@@ -73,7 +74,23 @@ enum Color {
 }
 ```
 
-Here, we have defined five potential colors. Each enum variant is just the color name. As there is no extra data associated with each variant, we say that each variant is of type `()`, or unit. It is also possible to have an enum variant contain extra data. Take a look at this more substantial example, which combines struct declarations with enum variants:
+Here, we have defined five potential colors. Each enum variant is just the color name. As there is no extra data associated with each variant, we say that each variant is of type `()`, or unit.
+
+To instantiate a variable with the value of an enum the syntax is
+
+> **Note**
+> enum instantiation does not require the `~` tilde syntax
```
adlerjohn, line 82: Wasn't the need for this completely removed? cc @mohammadfawaz

docs/src/basics/structs_tuples_and_enums.md
```diff
@@ -103,34 +120,56 @@ enum InventoryEvent {
...
+
 ```sway
-enum Color {
-    Blue: (),
-    Green: (),
-    Red: (),
-    Silver: (),
-    Grey: (),
+fn main() {
+  let event = InventoryEvent::ItemLoss(Claim {
```
adlerjohn, line 127: Weird indentation. While here, can you move these to `examples/`?

docs/src/basics/structs_tuples_and_enums.md
```diff
@@ -103,34 +120,56 @@ enum InventoryEvent {
...
+    let error2 = Error::UserError(UserError::Unauthorized);
 }
+```
 
-let event = InventoryEvent::ItemLoss(Claim {
-    insurance_company: ~Insurer::default(),
-    item_number: 42,
-    item_cost: 1_000,
-});
+The syntax above is not the most ergonomic but it is possible. It is advisable to not use the `enum Error` structure which nests `enums` and instead use the `enums` directly
```
adlerjohn, line 166: What's the use of `Error` then if it's never used?

Braqzen: Initially I thought that I could use the syntax `Error::StateError::Void`.
I was thinking about a hierarchy of enums and "sub-classing" them in a way which made me get to that nasty syntax.
Since the issue was around I thought to mention it plus since enums containing structs is a thing then someone might wonder what enum of enums is like.
That being said, it's nasty so I'm pointing out that it's better to not use it and instead have simple enums.

docs/src/basics/structs_tuples_and_enums.md
Braqzen: By "move" do you mean add the same examples into contract form under `examples/` or to completely remove everything to do with enums and move that under `examples/`?

docs/src/basics/structs_tuples_and_enums.md
mohammadfawaz: Not yet. The tilde is still needed for structs. The new parser should make its removal easier.

docs/src/basics/structs_tuples_and_enums.md
adlerjohn: Completely move all inline code in this page into the examples directory, and include it with `{{#include ...}}`.

docs/src/basics/structs_tuples_and_enums.md
adlerjohn: Ah I see. Then maybe clarify that a bit more, that it's easier to not use deeply-nested enums.

docs/src/basics/index.md
```diff
@@ -8,7 +8,7 @@ Sway basics.
 - [Built-in Types](./built_in_types.md)
 - [Blockchain Types](./blockchain_types.md)
 - [Functions](./functions.md)
-- [Structs and Tuples](./structs_and_tuples.md)
+- [Structs, Tuples & Enums](./structs_tuples_and_enums.md)
```
adlerjohn, line 11:
```suggestion
- [Structs, Tuples, and Enums](./structs_tuples_and_enums.md)
```

docs/src/basics/structs_tuples_and_enums.md
```diff
@@ -0,0 +1,104 @@
...
+
+Key points to note:
+
+- You must import all of the enums for use instead of just the `Error`
+- The lines may get unnecessarily long (depending on the names)
+- The syntax is not the most ergonomic
+
+### Enum Memory Layout
+
+> **Note**: This information is not vital if you are new to the language, or programming in general.
```
adlerjohn, line 102: For consistency
```suggestion
> **Note**
> This information is not vital if you are new to the language, or programming in general.
```

adlerjohn (commented):
nits

docs/src/basics/index.md
Braqzen: Oh, I thought I have changed that

mohammadfawaz (changes requested):

mohammadfawaz (approved):
Excellent, looks great!

PR #1861, opened 2022-06-06
Fix bug where forc_pkg::add_deps ignores that dependency name might differ from package name

### LINKED ISSUE #1858: Fix bug where `forc_pkg::add_deps` ignores that dependency name might differ from package name
As of #1002 `forc` support having a dependency name that differs from the package name.

While rebasing #1790 against master, I noticed that the `forc_pkg::add_deps` function introduced in #1686 incorrectly assumes the dependency name is always equal to the package name when adding the edges.

cc @kayagokalp

kayagokalp:
closes #1858.

## About This PR

- `forc_pkg::add_deps` now takes into account that dependency name and package name might differ.


forc-pkg/src/pkg.rs
```diff
@@ -181,8 +181,8 @@ pub enum SourceGitPinnedParseError {
 pub type DependencyName = String;
 
 pub struct PkgDiff {
-    pub added: Vec<Pkg>,
-    pub removed: Vec<Pkg>,
+    pub added: Vec<(String, Pkg)>,
+    pub removed: Vec<(String, Pkg)>,
```
mitchmindtree, line 185:
```suggestion
    pub removed: Vec<(DependencyName, Pkg)>,
```
^ Do you mind making these changes to the `added` field, as well as the `remove_deps` and `add_deps` fn signatures? `DependencyName` is just an alias for `String`, but I think it'll be useful for readability.

forc-pkg/src/pkg.rs
```diff
@@ -400,7 +405,7 @@ fn add_deps(
     let fetch_id = fetch_id(&path_map[&graph[*proj_node].id()], fetch_ts);
     let proj_node_after_delete = compilation_order.last().unwrap();
     for added_package in to_add {
```
mitchmindtree, line 407:
```suggestion
    for (added_dep_name, added_dep_pkg) in to_add {
```
Might be better to avoid the awkward tuple indexing?

mitchmindtree (commented):
Couple tiny nits, otherwise looks perfect :+1:

forc-pkg/src/pkg.rs
JoshuaBatty: Hi @kayagokalp are you able to make the change @mitchmindtree suggested above? Happy to approve this once that's been made.

kayagokalp: Oh, I guess I missed this one, sorry :typing

JoshuaBatty (approved):

Issue #1869, opened 2022-06-06
remove_deps cannot handle removed dependencies correctly when there are multiple roots after edge removal


kayagokalp:
Currently, `remove_deps` first removes edges and then removes the nodes with no incoming edge, in parent to child direction, which can be seen from [here](https://github.com/FuelLabs/sway/pull/1686#discussion_r885190272). Because of #1778, `remove_deps` wasn't executed in package removal scenarios. But with #1784, after patching a dependency and removing the patch from `Forc.toml`, I realized that the `BuildPlan` generation from the lock file succeeds but packages are not removed correctly.

Since the edges are removed first, we cannot reach subgraphs that are now unconnected to the original root of the graph while doing a BFS from the root node. Which means they are not deleted.

### At the beginning 
<img width="524" alt="Screen Shot 2022-06-06 at 18 01 46" src="https://user-images.githubusercontent.com/20915464/172188312-2bb3f9e1-d9ea-4420-a922-e154a93b1f47.png">

### After edge removal
<img width="699" alt="Screen Shot 2022-06-06 at 18 00 39" src="https://user-images.githubusercontent.com/20915464/172188357-54dbf6bc-df3f-4cc6-bfa8-faeea0c61e95.png">

## Proposed Solution
We can remove the nodes (that are either one of the removed ones or have no incoming edge) while doing the BFS rather than removing the edges first. I think removing the node also invokes the removal of the edges for the removed node.

Or we can keep track of the new roots that emerged while removing the edges and do separate BFS from each of the roots we found.

cc @mitchmindtree

mitchmindtree:
> Since the edges are removed first, we cannot reach subgraphs that are now unconnected to the original root of the graph while doing a BFS from the root node

Good spot!

Yeah I think your proposal makes sense. To clarify: after edge removal, rather than doing a BFS we can instead iterate over all nodes in reverse compilation order (that is, the compilation order *before* edge removal occurred, no need to recalculate it first) and remove every node that has no parent (i.e. no incoming edges).

PR #1870, opened 2022-06-06
remove_deps cannot handle removed dependencies correctly when there are multiple roots after edge removal

kayagokalp:
This is WIP opening up for visibility. 

## About this PR
This is one of the proposed solutions for #1869


mitchmindtree (approved):
Ahh just spotted this following my response [here](https://github.com/FuelLabs/sway/issues/1869#issuecomment-1148198439). This approach looks fine too! It involves iterating over the `to_remove` deps more often during node removal, but there will never be enough deps across a forc workspace for it to be noticable.

LGTM!

PR #1878, opened 2022-06-06
Move the impls for UnresolvedTypeCheck to match the location of the other impls for each type

emilyaherbert:
Simple refactoring change


mohammadfawaz (approved):
LGTM!

PR #1881, opened 2022-06-06
Use the `Spanned` trait everywhere

emilyaherbert:
This PR removes the manual implementations of `span(&self)` in favor of implementing `Spanned`.


mohammadfawaz (approved):
Quickly looking through it and everything seems reasonable. Glad to see all those `clone()` calls going away 😄

mohammadfawaz (approved):

sway-core/src/semantic_analysis/ast_node/declaration/monomorphize.rs
```diff
@@ -58,7 +58,7 @@ pub(crate) trait Monomorphize {
 
 impl<T> Monomorphize for T
 where
-    T: MonomorphizeHelper<Output = T>,
+    T: MonomorphizeHelper<Output = T> + Spanned,
```
otrho, line 61: Yay, I was wondering if we would actually ever need it in a `where`.

PR #1883, opened 2022-06-06
Replace manual string methods with `Display`

emilyaherbert:
This PR:

1) replaces the manual string methods with `Display`
2) removes the `FriendlyTypeString` trait because it is no longer needed


mohammadfawaz (approved):
Much cleaner!

mohammadfawaz (approved):

Issue #1888, opened 2022-06-07
sway-fmt-v2 should be able to handle indentations correctly.


kayagokalp:
Since we are building the formatted code from AST items we may need to have a common way of building the formatted code (something like a `CodeBuilder` that keeps track of the current indentation level). So that we can build the code with correct indentations. 

I can see this kind of logic is present in the current formatter but would love to discuss if there are other ways we can reconstruct the code with correct indentations. 

If this is indeed something we need I can implement it before #1873 

cc @eureka-cpu

eureka-cpu:
Going to add some handling of newlines today

eureka-cpu:
Now just need to handle block indentation handling and this should be good to go

eureka-cpu:
Take a look at `shape.rs` from `rustfmt` I think that will give you the answer for handling indents!

eureka-cpu:
I think with #1894 & #1911 this should be closed but feel free to reopen if not

PR #1894, opened 2022-06-07
Adds associated functions for `NewlineStyle`

eureka-cpu:
The is nearly straight from `rustfmt` with some minor adjustments. It aids in formatting newlines depending on system type. Should help with #1888

### DIFF
--- a/sway-fmt-v2/src/items/item_use.rs
+++ b/sway-fmt-v2/src/items/item_use.rs
@@ -4,5 +4,7 @@ use sway_parse::ItemUse;
 impl Format for ItemUse {
     fn format(&self, _formatter: &Formatter) -> FormattedCode {
         todo!()
+        // reorder use statements alphabeticaly
+        // break into new lines for long use statements
     }
 }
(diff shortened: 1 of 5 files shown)


sway-fmt-v2/src/items/item_use.rs
```diff
@@ -4,5 +4,7 @@ use sway_parse::ItemUse;
 impl Format for ItemUse {
     fn format(&self, _formatter: &Formatter) -> FormattedCode {
         todo!()
+        // reorder use statements alphabeticaly
```
kayagokalp, line 7: Maybe we can remove this diff? (assuming it is not related to this PR)

eureka-cpu: Ah, you're right! Thank you

kayagokalp (commented):
lgtm, but will defer to @mitchmindtree or @JoshuaBatty

JoshuaBatty (approved):
Looks good.

PR #1896, opened 2022-06-07
Make TypedImplTrait its own thing

emilyaherbert:
This bug I'm working on right now requires `TypedImplSelf` to be its own object, so I broke that out into it's own PR for readability.


mohammadfawaz (approved):
This is a non-functional change right? Looks great!

Issue #1872, opened 2022-06-06
Incorrect padding for enums with different sized variants in a `ReturnData` receipt


mohammadfawaz:
Consider the following example (from https://github.com/segfault-magnet/sway_encoding_bug):
```rust
library lib;

abi MyContract {
    fn test_function() -> TopLevelEnum;
}

pub enum LowerLevelEnum {
    first: u32,
    second: b256,
}

pub struct ThenAStruct {
    first: LowerLevelEnum,
    second: u32
}

pub enum TopLevelEnum {
	first: ThenAStruct
}
```
and the contract
```rust
impl MyContract for Contract {
    fn test_function() -> TopLevelEnum {
            TopLevelEnum::first(ThenAStruct { first: LowerLevelEnum::first(1), second: 2 })
    }
}
```
The `ReturnData` receipt looks like:
```
  {
    "ReturnData": {
      "data": "0000000000000000000000000000000000000000000000010000000000000002000000000000000000000000000000000000000000000000",
      "digest": "bc1b6da4984d4ab324cf3da4575928cb5323a3e30fc4f914a0874e99cf72d11d",
      "id": "2eed3ca7238e5294661e6baa492acaa4ffbfcc7a9111751da2482d975db55a09",
      "is": 11496,
      "len": 56,
      "pc": 11584,
      "ptr": 11632
    }
  },
```
The padded zeros at the end are in the wrong spot. From @segfault-magnet :
```
[
// TopLevelEnum discriminant
0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,
// ThenAStruct begins here
        // LowerLevelEnum discriminant
        0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,
        // LowerLevelEnum::first u32
        0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x1,

        // Missing padding 3 WORDS of padding here because the biggest variant of LowerLevelEnum is b256

        // ThenAStruct::second u32 here
        0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x2,
        // Then a struct ends here

// Here are the 3 missing words of padding! TopLevelEnum shouldn't have any
   padding due to it having only one variant
0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,
0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,
0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0
]
```
[...]

PR #1897, opened 2022-06-07
Fix how `enum` constants are initialized in codegen

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/1872

Basically some padding has to be added in situations where the variant that's being initialized is smaller than the total size of the enum. The rest of this change is working around this idea and cleaning up the reliance on the constant value type (the constant variant) instead of the type of the stack variable itself (the whole enum).

The padding is added _between_ the tag of the enum and the variant. E.g. 
```rust
enum E {
    X: b256,
    Y: u64,
}
```
Storing `E::Y(42)` into looks like:
```rust
word n:     0000000000000001  // This is the tag
word n+1:   0000000000000000  // First word of padding
word n+2:   0000000000000000  // Second word of padding
word n+3:   0000000000000000  // Third word of padding 
word n+4:   000000000000002a  // The actual data
```
Also fixes parsing of `const` in IR. Previously, some consts were mistyped because there was no reliance on the actual type provided, so a union type was occasionally being missed (`ConstantValue` itself doesn't known union types).


sway-core/src/asm_generation/from_ir.rs
```diff
@@ -2090,98 +2102,195 @@ impl<'ir> AsmBuilder<'ir> {
...
+                // If the literal we're trying to initialise with is smaller than than the actual
+                // size of the union, then a padding of zeros is required. Calculate the size of
+                // the padding and set the appropriate bytes to zero.
+                let constant_size = ir_type_size_in_bytes(self.context, &constant.ty);
+                assert!(constant_size <= value_size);
+                let padding_size = value_size - constant_size;
+                let padding_size_in_words = size_bytes_in_words!(padding_size);
+
+                if padding_size > 0 {
+                    // Store padding of zeros and then store the value itself
```
otrho, line 2245: So, do you want to put the padding first?  While the values _are_ big endian, when the variant is smaller we'll expect it to butt up against the tag won't we?  Or have you changed the way that works elsewhere?

sway-core/src/asm_generation/from_ir.rs
```diff
@@ -2287,10 +2396,17 @@ pub fn aggregate_idcs_to_field_layout(
             Type::Union(aggregate) => {
                 let idx = *idx as usize;
                 let field_type = context.aggregates[aggregate.0].field_types()[idx];
+                let union_size_in_bytes = ir_type_size_in_bytes(context, &ty);
                 let field_size_in_bytes = ir_type_size_in_bytes(context, &field_type);
 
-                // The union fields are all at offset 0.
-                ((offs, field_size_in_bytes), field_type)
+                // The union fields are at offset (union_size - variant_size) due to left padding.
```
otrho, line 2402: Ahh, OK, not sure if it's the best approach but if it works then it doesn't make that much difference.

sway-core/src/asm_generation/from_ir.rs
mohammadfawaz: I feel like left padding is more intuitive? But also makes the math a bit more annoying to follow so I don't really have a strong opinion here. This should definitely be in the spec somewhere because it affects external contexts.

mohammadfawaz: Sadly, the example in the spec does not tell us much about enums with enums that have different sizes: https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#sum-types-enums

mohammadfawaz: Created an issue here: https://github.com/FuelLabs/fuel-specs/issues/339

sway-core/src/asm_generation/from_ir.rs
otrho: Argh, catching up on the Slack discussion - they need to be left padded for the ABI.  All good, carry on.

otrho (approved):
So there was a subtle bug in parsing constants into the correct type?

mohammadfawaz:
> So there was a subtle bug in parsing constants into the correct type?

Tiny issue but basically it is possible for the initializer to be, say, `u64` but the real type is `(u64 | b256)` and that information was being occasionally dropped.

PR #1900, opened 2022-06-08
Add VS Code config

AlicanC:
This PR adds some essential VSC config to the project. This config will be picked up by VSC and it will offer to install the recommended extensions. It also has a launch configuration that will debug the tests with in-editor breakpoint, etc. support.

Another popular option is to gitignore the `.vscode` directory, but choosing the right extensions and writing the correct launch configurations might be a real hassle so it is nice to provide some examples using this directory.

If a developer wants their own config they can create their own VSC workspace and personalize that as usual.

Non-VSC devs won't be affected by this PR, and VSC devs will see a recommendation dialog if they don't have all the extensions installed which can be dismissed forever.

Note: Helps with https://github.com/FuelLabs/pm/issues/24 as a side effect.

### DIFF
--- a/.vscode/README.md
+++ b/.vscode/README.md
@@ -1,0 +1,9 @@
+# VS Code Configuration
+
+This directory contains essential VS Code configuration that should be useful to most developers planning to work on this project.
+
+If you want to personalize these, you can create your own workspace as always.
+
+## Contributing
+
+Make sure you are adding things that will be useful to almost all developers.
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,8 @@
+{
+  "recommendations": [
+    // Rust language support
+    "rust-lang.rust-analyzer",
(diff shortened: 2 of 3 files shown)


AlicanC:
Reviewers, please add more if you know any VSC devs that I might've missed.

otrho:
No, sorry, we have a policy of not adding editor specific config to the repository.  This has been tried before and always gets shot down. :slightly_smiling_face:

AlicanC:
Not really surprised to hear that :D

I am pushing VSC specifically because it's what GitHub uses so it also configures the experience at `https://github.dev/FuelLabs/sway/`. So I'd say while it's editor config, it's also GitHub/repo config.

If a discussion was already had though, I don't feel this PR is beneficial enough to reopen it. This was some progress towards https://github.com/FuelLabs/pm/issues/24, but I can do it on another repo.

PR #1901, opened 2022-06-08
Fix incorrect `CallPath` suffix on `TypedImplTrait`

JoshuaBatty:
Previously, the `trait_name.suffix` on `ImplTrait` was being assigned `Ident::new_with_override("r#Self", block_span.clone())`. This meant in the language server, the name and span for the token was in-correct. See issue #1773 

This PR creates the correct `Ident` if the type_implementing_for matches `TypeInfo::Custom`, otherwise the original behavior is retained. I'm not 100% sure why the original Ident was created with the name `r#Self` and with the whole block as the span but maybe @canndrew has a better idea?

This is my first PR to `sway-core` so please let me know if I have done anything incorrectly.

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/declaration/function.rs
+++ b/sway-core/src/semantic_analysis/ast_node/declaration/function.rs
@@ -18,7 +18,7 @@ pub struct TypedFunctionDeclaration {
     pub(crate) type_parameters: Vec<TypeParameter>,
     /// Used for error messages -- the span pointing to the return type
     /// annotation of the function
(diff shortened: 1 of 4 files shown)


otrho (approved):

otrho:
A bunch of PRs are getting the same semantic analysis conflicts today...

JoshuaBatty:
Ok, should be good now, have removed most of the duplicate code from #1896 and just has the relevant changes remaining.

otrho (approved):

Issue #1904, opened 2022-06-08
Add a test to ensure we produce a nice error when detecting a dependency cycle


mitchmindtree:
This could be two or three packages which depend on each other under the E2E `should_fail` tests. We want to ensure that 1. we don't attempt to infinitely fetch (I don't think we do) and 2. we produce a decent error.

kayagokalp:
I just tried to build a package with cyclic dependency. It fails as it should but maybe we can touch the error produced a little.

Current output:

```
Error: dependency cycle detected: Cycle(NodeIndex(2))
```

I am not sure if NodeIndex(2) part means something to the user but maybe something like the following would be easier to understand if that ever happens to a user:

```
Error: dependency cycle detected between 'dependency_a' and `dependency_b` 
```

kayagokalp:
I am quickly adding two libraries that depend on each other so that we do not forget about https://github.com/FuelLabs/sway/pull/1910#issuecomment-1150801323

PR #1910, opened 2022-06-08
Dependency cycles are explicitly printed

### LINKED ISSUE #1908: Error message can be more descriptive when there is a dependency cycle
I just tried to build a package with cyclic dependency. It fails as it should but maybe we can touch the error produced a little.

Current output:

```
Error: dependency cycle detected: Cycle(NodeIndex(2))
```

I am not sure if NodeIndex(2) part means something to the user but maybe something like the following would be easier to understand if that ever happens to a user:

```
Error: dependency cycle detected between 'dependency_a' and `dependency_b` 
```

_Originally posted by @kayagokalp in https://github.com/FuelLabs/sway/issues/1904#issuecomment-1149732183_

kayagokalp:
## About this PR
closes #1908.

Dependency cycles were detected but not explicitly printed. 

With this PR, we are computing strongly connected components using [Kosaraju's algorithm](https://en.wikipedia.org/wiki/Kosaraju%27s_algorithm). If the resulting list of nodes has more than 1 element, they are members of a cycle in the graph.  

### Test

1. `forc init --library dependency_a`
2. `forc init --library dependency_b`
3. add `dependency_b` as a dependency for `dependency_a`
4. add`dependency_a` as a dependency for `dependency_b`
5. `forc init --library dependency_c`
6. `forc init --library dependency_d`
7. add `dependency_c` as a dependency for `dependency_d`
8. add `dependency_d` as a dependency for `dependency_c`
10. add `dependency_c` as a dependency for `dependency_a` 
11. `forc build` in dependency_a

The resulting dependency graph would be something like the following:

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -723,9 +723,31 @@ impl Default for GitReference {
 /// dependencies are always compiled before their dependents.
 pub fn compilation_order(graph: &Graph) -> Result<Vec<NodeIx>> {
     let rev_pkg_graph = petgraph::visit::Reversed(&graph);
-    petgraph::algo::toposort(rev_pkg_graph, None)
-        // TODO: Show full list of packages that cycle.
-        .map_err(|e| anyhow!("dependency cycle detected: {:?}", e))
+    petgraph::algo::toposort(rev_pkg_graph, None).map_err(|_| {
+        // Find strongly connected components
+        // If the vector has an element with length more than 1, it contains a cyclic path.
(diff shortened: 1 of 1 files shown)


JoshuaBatty (approved):
Nice, looks good.

mitchmindtree:
Wicked stuff @kayagokalp :) Would be sweet to have a test, but we can get around to it.

kayagokalp:
Thanks, @mitchmindtree :), for testing, is addressing #1904 would be enough or should we have a dedicated test for this section of the code?

PR #1911, opened 2022-06-08
Adds associated functions for handling of indentation

eureka-cpu:
Associated to issue #1888 

This should give us the methods we need to handle indentation in the new formatter.

This also adds a `utils` folder that will house the associated functions for organizational purposes.


kayagokalp (commented):
this looks good to me, i.e I think we have everything we need for unblocking the implementation of the logic for say `enums` here.

So I think this will give us a strong base (if not the whole thing) for indentation handling.  I will defer to @mitchmindtree or @JoshuaBatty for the final approval.

JoshuaBatty (approved):
LGTM

eureka-cpu:
> this looks good to me, i.e I think we have everything we need for unblocking the implementation of the logic for say `enums` here.
> 
> 
> 
> So I think this will give us a strong base (if not the whole thing) for indentation handling.  I will defer to @mitchmindtree or @JoshuaBatty for the final approval. 

Does this close #1888 ?

Issue #1893, opened 2022-06-07
Bytecode hash for predicates and scripts on compile


simonr0204:
It would be nice for the predicate/script hash (in addition to the bytecode) to be available in `out/debug/` after compiling.

When building systems which depend on these hashes, it's a bit of a pain to have to open up the sdk just to load in the binary file and hash it, and then come back and copy/paste the result into the relevant Sway code.

Currently on `forc build` the bytecode size is displayed. Maybe it could also spit out the hash?

PR #1913, opened 2022-06-09
Store/show bytecode hash

seem-less:
Closes [#1893](https://github.com/FuelLabs/sway/issues/1893)


mohammadfawaz:
* The formatter is failing. You may run the formatter first locally by running `cargo fmt` in the main `sway` directory.
* With this change, we're going to also display the bytecode hash for contracts even when deploying them. So, when running `forc deploy`, we will see something like:
```consol
  Bytecode size is 68 bytes.
  Bytecode hash is: 0x455fd1b532c4f411b5c6b4c5bc4b9ef9d8e4b50e9e48221e7da56507878ea1de
Contract id: 0xb9bd3a82c2a32802da245dc01e6fa9df25c3226287f8b279c0be1895d7122418
Logs:
TransactionId(HexFormatted(2c6c13378aa58fc8e2621e965728116f3e63bbb2189cf70632db00939243a5ef))
```
@simonr0204 is this desirable? The issue you opened only talks about predicates and scripts.

simonr0204:
I can't think of a use case for contracts, only scripts and predicates. Confusing the hash with the contract Id would indeed be the obvious mistake this could induce... Maybe we don't show it for contracts?

seem-less:
ok no worries, I will only make it display for scripts/predicates

seem-less:
I have updated the PR to only generate the hash if the json_abi is empty (indicating that the program type is a contract or library as intended [here](https://github.com/FuelLabs/sway/blob/2f9ae54a59df88eaa1b9986f621e00ebdcb6a7ab/forc-pkg/src/pkg.rs#L1313)). But the build command is currently generating the *-abi.json for scripts/predicates as well which I don't think is intended as seen [here](https://github.com/FuelLabs/sway/blob/2f9ae54a59df88eaa1b9986f621e00ebdcb6a7ab/forc-pkg/src/pkg.rs#L1432).

Shall this be made as a new issue or should I make the fix in this PR?

adlerjohn:
Ah, so scripts with an ABI declaration now will output a non-empty JSON ABI file. You'll need to explicitly check the type of program.

forc/Cargo.toml
```diff
@@ -25,6 +25,7 @@ forc-pkg = { version = "0.15.1", path = "../forc-pkg" }
 forc-util = { version = "0.15.1", path = "../forc-util" }
 fs_extra = "1.2"
 fuel-asm = "0.5"
+fuel-crypto = "0.5"
```
mohammadfawaz, line 28: It looks like `fuel-tx` already has `fuel-crypto` as a dependency, so there should be no need to add `fuel-crypto` as an explicit dependency here!

adlerjohn (commented):
There should be some documentation on what specifically this bytecode hash is. Some users might confuse it for a bytecode root, which is how the contract ID is computed.

forc/Cargo.toml
seem-less: I am not entirely sure how to reference a dependency of a dependency without modifying the `fuel-tx` crate to expose `fuel-crypto`. It seems that `fuel-tx` also has `fuel-asm` as a dependency, yet it is also listed as a separate dependency here.

How should I go about this without adding `fuel-crypto` explicitly here?

mohammadfawaz: Hmmm good point. This may not be possible after all? @sezna, @adlerjohn thoughts?

adlerjohn: Huh, you actually appear to be correct @seem-less, `fuel-crypto` is not re-exported by `fuel-tx`! Nevermind then, you can leave `fuel-crypto` as a dependency.

As for why `fuel-asm` is a dependency...that's code debt that was never fixed 😂. Feel free to open another PR to clean it up.

mitchmindtree (commented):
Thanks for the contribution @seem-less! I've left a couple small suggestions, but otherwise looks really good!

forc-pkg/src/pkg.rs
```diff
@@ -1421,9 +1431,14 @@ pub fn build(
...
         json_abi.extend(compiled.json_abi);
         bytecode = compiled.bytecode;
+        tree_type = compiled.tree_type;
         source_map.insert_dependency(path.clone());
     }
-    let compiled = Compiled { bytecode, json_abi };
+    let compiled = Compiled {
+        bytecode,
+        json_abi,
+        tree_type,
```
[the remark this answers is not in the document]
seem-less: WIZARDRY

seem-less:
> There should be some documentation on what specifically this bytecode hash is. Some users might confuse it for a bytecode root, which is how the contract ID is computed.

I added a little bit to the forc_build doc explaining what is used to hash the binary file. Is this enough or does it require more explanation?

forc/src/ops/forc_build.rs
```diff
@@ -160,6 +161,15 @@ pub fn build(command: BuildCommand) -> Result<pkg::Compiled> {
 
     info!("  Bytecode size is {} bytes.", compiled.bytecode.len());
 
+    if let TreeType::Script | TreeType::Predicate = compiled.tree_type {
+        // hash the bytecode for scripts/predicates and store it in a file in the output directory
+        let bytecode_hash = format!("0x{}", fuel_crypto::Hasher::hash(&compiled.bytecode));
+        let hash_file_name = format!("{}-bin-hash", &manifest.project.name);
+        let hash_path = output_dir.join(hash_file_name);
+        fs::write(hash_path, &bytecode_hash)?;
```
adlerjohn, line 169: Can you file an issue to explore how best to handle this? It's not necessarily the case that we want this file emitting by default, as usually users won't actually need to script or predicate hash manually.

forc/src/ops/forc_build.rs
```diff
@@ -160,6 +161,15 @@ pub fn build(command: BuildCommand) -> Result<pkg::Compiled> {
 
     info!("  Bytecode size is {} bytes.", compiled.bytecode.len());
 
+    if let TreeType::Script | TreeType::Predicate = compiled.tree_type {
+        // hash the bytecode for scripts/predicates and store it in a file in the output directory
+        let bytecode_hash = format!("0x{}", fuel_crypto::Hasher::hash(&compiled.bytecode));
+        let hash_file_name = format!("{}-bin-hash", &manifest.project.name);
```
adlerjohn, line 167: This `-bin-hash` should probably be a constant.

adlerjohn (commented):
Generally looks fine to me, will defer to @mitchmindtree for final review.

forc/src/ops/forc_build.rs
mitchmindtree: I think this should be fine - it's a tiny file and will be enclosed in the `out/<profile>/` directory, so hopefully not too intrusive. I'm happy to leave it in for this PR at least.

mitchmindtree (commented):
Looks great! Happy to approve after John's `const` comment addressed.

adlerjohn (approved):
Looks fine to me, but will defer to @mitchmindtree for final review.

mitchmindtree (approved):
Looking good. Thanks again @seem-less!!

Issue #1916, opened 2022-06-09
`unit` type in IR has size 8 bytes which seems wasteful


mohammadfawaz:
Specifically here: https://github.com/FuelLabs/sway/blob/0587e82b107a32dae851654b9ec207ea077483fb/sway-core/src/asm_generation/from_ir.rs#L2332

Not sure about the implications of changing the size to 0. Soliciting @otrho input.

This comes up in situations like the following:
```rust
script;

enum SomeEnum {
    A: (),
    B: (),
}

fn main() -> SomeEnum {
    SomeEnum::B()
}
```
In the above, I'd expect a `retd` receipt of size 8 because the `enum` should really have size 8 (only the tag!). Instead, we get the following:
```json
  {
    "ReturnData": {
      "data": "00000000000000010000000000000000",
      "digest": "783825822a6f9e62da2190e828e4c9d2576e5977e3a0b3620b092dfb9e9996fa",
      "id": "0000000000000000000000000000000000000000000000000000000000000000",
      "is": 10352,
      "len": 16,
      "pc": 10396,
      "ptr": 10424
    }
```
because the `enum` is essentially a struct of two elements: a tag and a value. Because the size of `unit` is 8 bytes, the maximum size of the `enum` above is `8 + 8 = 16 bytes`

mohammadfawaz:
Here's my attempt to fixing this: https://github.com/FuelLabs/sway/pull/1922

otrho:
The solution needs to be careful, there could be problematic edge cases.

If we never actually read a unit value, since it's 'empty', then in aggregates it could just be a typed empty space; in memory a pointer to zero bytes (i.e., actually pointing to the next thing, but never used so it's OK) and in a register or return value, just zero.

But this only works if we never ever actually use a unit value for anything, even 'accidentally' in some general code which maybe just copies values around.  Off the top of my head, it probably just works, if size_of is 0 then nothing will be copied, aggregate sizes will still be correct, etc.  🤔

mohammadfawaz:
I'm looking into whether I can start by making enums not have the union at all if all variants are unit types.

PR #1922, opened 2022-06-09
Enums with unit type variants don't need a union

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/1916 

As per https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#sum-types-enums, enums where all variants have unit types should have size 8 bytes (only the tag).


otrho (approved):
I wonder if this could instead have been an IR pass.  It might be trickier to _remove_ the insert than to never have it in the first place, but still...

PR #1925, opened 2022-06-09
sway-fmt-v2 Formatter struct should have a Shape field that can keep track of the current indentation

### LINKED ISSUE #1924: sway-fmt-v2 Formatter struct should have a Shape field that can keep track of the current indentation
There should be a common way between different item formatters so that they can all reach the current indentation state of the formatter. We can do this by simply having a shape field in the Formatter struct and getting shape from self in Format trait implementation. 

Related to #1911 & #1888 

Part of #1516

kayagokalp:
closes #1924 

This PR adds a shape field to the Formatter so that from each implementation of the format trait we can reach a shared shape instance. This way we can keep track of the last indentation level before adding a newly formatted code result.

### DIFF
--- a/forc-plugins/forc-fmt-v2/src/main.rs
+++ b/forc-plugins/forc-fmt-v2/src/main.rs
@@ -49,12 +49,12 @@ fn run() -> Result<()> {
         Some(path) => PathBuf::from(path),
         None => std::env::current_dir()?,
     };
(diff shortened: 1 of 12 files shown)


eureka-cpu:
is this ready for review?

eureka-cpu (approved):
good work, looks good!

PR #1927, opened 2022-06-09
sway-fmt-v2 should actually use the newline style utils implemented before

### LINKED ISSUE #1926: sway-fmt-v2 should actually use the newline style implemented in #1894
Currently, we have nice utils around applying newline styles but we are not using them. As we start the implementation of individual formatting logics for different items (enums, use, etc.) we want to apply the newline styling after we are done with the formatting. 

Related to #1888 & #1894.

Part of #1516

kayagokalp:
closes #1926.

After formatting is done and the result is collected, we are now applying the newline style to the resulting `FormattedCode`.

### DIFF
--- a/sway-fmt-v2/src/fmt.rs
+++ b/sway-fmt-v2/src/fmt.rs
@@ -2,6 +2,7 @@ use std::{path::Path, sync::Arc};
 use sway_core::BuildConfig;
 use sway_parse::ItemKind;
 
+use crate::{utils::newline_style::apply_newline_style, config::whitespace::NewlineStyle};
 pub use crate::{
     config::manifest::Config,
     error::{ConfigError, FormatterError},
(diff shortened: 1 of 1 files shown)


eureka-cpu:
ready for review?

sway-fmt-v2/src/fmt.rs
```diff
@@ -6,6 +6,7 @@ pub use crate::{
     config::manifest::Config,
     error::{ConfigError, FormatterError},
 };
+use crate::{config::whitespace::NewlineStyle, utils::newline_style::apply_newline_style};
```
mitchmindtree, line 9: There's already a set of crate imports above - we should merge these imports into that set.

mitchmindtree (commented):
Just one small nit, otherwise will defer to @eureka-cpu for approval once ready for review.

Also, don't forget to switch from "Draft" to "Ready for Review" when the time comes :)

sway-fmt-v2/src/fmt.rs
eureka-cpu: I was also going to say something about this, but the above is public, I don't think these need to be necessarily since they just get used in a single method here, but I digress haha

mitchmindtree: Ah good spot! My apologies, kaya disregard my comment :)

In general, it's good to try and be mindful of the public API and to take the more conservative choice when we're unsure (i.e. private, or `pub(crate)`). Technically, anything that changes in the public API is a breaking semver change, so if we're not careful it can make it trickier to reach a level of "stability".

PR #1928, opened 2022-06-10
Add `std::tx::get_script_data`

AlicanC:
Adds `std::tx::get_script_data` to allow accessing script data conveniently:

```rs
script;

use std::assert::assert;
use std::tx::get_script_data;

fn main() {
    // Reference type
    let received: b256 = get_script_data();
    let expected: b256 = 0xef86afa9696cf0dc6385e2c407a6e159a1103cefb7e2ae0636fb33d3cb2a9e4a;
    assert(received == expected);

    // Copy type
    let received: u64 = get_script_data();
    let expected: u64 = 17259675764097085660;
    assert(received == expected);
}
```

### DIFF
--- a/sway-lib-std/src/tx.sw
+++ b/sway-lib-std/src/tx.sw
@@ -4,6 +4,7 @@ library tx;
 
 use ::address::Address;
 use ::contract_id::ContractId;
+use ::intrinsics::is_reference_type;
 
 ////////////////////////////////////////
 // Transaction fields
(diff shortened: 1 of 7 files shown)


otrho (approved):
A reminder: you can name arguments to an asm block any identifier, they don't need to be `rX`.  Often this can make them a bit more readable.

edit: Ah it's still a draft.  More coming?

AlicanC:
> A reminder: you can name arguments to an asm block any identifier, they don't need to be `rX`. Often this can make them a bit more readable.

I do this in new code, but the file I was editing never did this so I went for consistency.

> edit: Ah it's still a draft. More coming?

Whoops. Forgot to mark ready.

sway-lib-std/src/tx.sw
```diff
@@ -139,24 +141,30 @@ pub fn tx_script_start_offset() -> u32 {
...
-//         move to_return sp;
-//         // copy script data into above buffer
-//         mcp to_return script_data_ptr script_data_len;
-//         to_return: T
-//     }
-// }
+pub fn get_script_data<T>() -> T {
+    // TODO some safety checks on the input data? We are going to assume it is the right type for now.
+    let ptr = tx_script_data_start_offset();
+    if is_reference_type::<T>() {
```
mohammadfawaz, line 157: Can we replace this `if/else` with a call to `mem::read()`? 😄

AlicanC: Yes!

The same can also be done for: https://github.com/FuelLabs/sway/pull/1984

cc @camsjams

AlicanC: @mohammadfawaz created an issue: https://github.com/FuelLabs/sway/issues/2015

PR #1935, opened 2022-06-10
Allow the transfer_to function to transfer any native asset

### LINKED ISSUE #1851: Make the `transfer_to()` from `token.sw` in the `std-lib` non-specific to the `contract_id()`

nfurfaro:
This PR cleans up the `token` module api by:
- renaming `transfer_to(...)` to just `transfer(...)`
- adding a third param `asset_id: ContractId` to `transfer(...)` to make it general purpose
- renames the `recipient` and `destination` params both to `to` for consistency.
- adds new tests
- refactors 2 old tests now that SDK can handle `Identity`
- improves documentation 

closes #1851


adlerjohn (commented):
Update the comments too

sway-lib-std/src/token.sw
```diff
@@ -8,24 +8,32 @@ use ::tx::*;
 use ::context::call_frames::contract_id;
 use ::identity::Identity;
 
-/// Mint `amount` coins of the current contract's `asset_id` and transfer them to `destination` by calling either force_transfer_to_contract() or transfer_to_output(), depending on the type of `Identity`.
-pub fn mint_to(amount: u64, recipient: Identity) {
+/// Mint `amount` coins of the current contract's `asset_id` and transfer them
+/// to `to` by calling either force_transfer_to_contract() or transfer_to_output
```
adlerjohn, line 12: The newlines also need a `///`

sway-lib-std/src/token.sw
```diff
@@ -8,24 +8,32 @@ use ::tx::*;
...
+/// (), depending on the type of `Identity`.
```
adlerjohn, line 13: This `()` should be on the previous line.

sway-lib-std/src/token.sw
nfurfaro: Not sure what you mean by the newlines...

adlerjohn: Ahhh wait my bad, GitHub was rending things funny. Carry on.

sway-lib-std/src/token.sw
```diff
@@ -42,29 +50,43 @@ pub fn burn(amount: u64) {
     }
 }
 
-/// Transfer `amount` coins of the current contract's `asset_id` and send them to `destination` by calling either force_transfer_to_contract() or transfer_to_output(), depending on the type of `Identity`.
-pub fn transfer_to(amount: u64, recipient: Identity) {
-    match recipient {
+/// Transfer `amount` coins of the current contract's `asset_id` and send them
+/// to `to` by calling either force_transfer_to_contract() or transfer_to_output
+/// (), depending on the type of `Identity`.
```
adlerjohn, line 55:
```suggestion
/// to `to` by calling either force_transfer_to_contract() or transfer_to_output(),
/// depending on the type of `Identity`.
```

sway-lib-std/src/token.sw
```diff
@@ -42,29 +50,43 @@ pub fn burn(amount: u64) {
...
+///
+/// CAUTION !!!
+///
+/// This may transfer coins to a contract even with no way to retrieve them
+/// ( ie: no withdraw() function on receiving contract), possibly leading to
```
adlerjohn, line 60:
```suggestion
/// (i.e. no withdrawal functionality on receiving contract), possibly leading to
```

sway-lib-std/src/token.sw
```diff
@@ -42,29 +50,43 @@ pub fn burn(amount: u64) {
...
-/// Use of this function can lead to irretrievable loss of coins if not used with caution.
-pub fn force_transfer_to_contract(amount: u64, asset_id: ContractId, destination: ContractId) {
-    asm(r1: amount, r2: asset_id.value, r3: destination.value) {
+/// UNCONDITIONAL transfer of `amount` coins of type `asset_id` to
+/// the contract at `to`.
+///
+/// CAUTION !!!
+///
+/// This will transfer coins to a contract even with no way to retrieve them
+/// ( ie: no withdraw() function on receiving contract), possibly leading to
```
adlerjohn, line 79:
```suggestion
/// (i.e. no withdrawal functionality on receiving contract), possibly leading to
```

adlerjohn (commented):
Looks fine to me, but will defer to @otrho for final review.

otrho (approved):
This is probably obvious but I'm curious about the use of 'coins' in the docs.  Most of the time I see the language being 'asset' and 'token' rather than 'coin', though I guess the 'mint' analogy is already in place too.

nfurfaro:
I don't know if there any hard and fast rules around this, but I tend to think of native assets when I hear `coins`, as compared to `tokens` which makes me think of ledger-based tokens (ie: erc-20 tokens on Ethereum ).

adlerjohn (approved):

PR #1937, opened 2022-06-10
Replace r#type with `type_id` and `type_info`.

emilyaherbert:
Before there were two ways of referring to a `TypeId` member in a struct: `r#type` and `type_id`, and two ways of referring to `TypeInfo` in a struct: `r#type` and `type_info`. This PR standardizes that.


otrho (approved):
utACK

PR #1944, opened 2022-06-11
Test for cyclic dependency added to should_fail

kayagokalp:
This adds a test to the `should_fail` part of the suite so that once there is a cyclic dependency we are indeed not building (more precisely not trying to build, i.e infinite fetching) the project.

I added a `dependency_a` which depends on `dependency_b`. But `dependency_b` depends on `dependency_a`. So trying to build `dependency_a` (or `dependency_b`) should fail.


test/src/e2e_vm_tests/test_programs/should_fail/cyclic_dependency/dependency_a/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "lib.sw"
+license = "Apache-2.0"
+name = "dependency_a"
```
mitchmindtree, line 5:
```suggestion
name = "dependency_a"
implicit-std = false
```

test/src/e2e_vm_tests/test_programs/should_fail/cyclic_dependency/dependency_b/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "lib.sw"
+license = "Apache-2.0"
+name = "dependency_b"
```
mitchmindtree, line 5:
```suggestion
name = "dependency_b"
implicit-std = false
```

mitchmindtree (changes requested):
Looks good, just missing a couple of `implicit-std = false` (you can spot this by seeing the `git` dependency show up in the lock files).

mitchmindtree:
@kayagokalp would you mind also rebuilding those two packages locally in order to update their lock files and push the updated lock files too? Otherwise they'll end up needing to be updated in some future unrelated PR. #1951 aims at addressing this.

kayagokalp:
Oh right, my bad. Pushing those as well

mitchmindtree:
I think it's fine to include it still, and we'll need to as of #1951 anyway.

The error is showing up in the E2E test CI pass, but it looks like it's `panic!`ing rather that just exiting with the error. I'll do up a quick PR for this.

kayagokalp:
Hmm, I might be missing something but I think cannot add the lock files since the lock file of `dependency_a` or `dependency_b` is not actually being generated at the moment (because we cannot build them). 

The old lock file that was present in this PR was generated before I added `dependency_b` as a dependency for `dependency_a` (i.e, before creating the cyclic dependency) and I accidentally pushed that file.

mitchmindtree:
> Hmm, I might be missing something but I think cannot add the lock files since the lock file of dependency_a or dependency_b is not actually being generated at the moment (because we cannot build them).

:laughing: that's a really good point! Turns out the aforementioned CI error was unrelated to this and solved in #1955 anyways.

mitchmindtree (approved):

Issue #407, opened 2021-11-19
Constant value evaluation in the AST needed.


otrho:
Constant expressions are used in the language in a couple of areas, two I can think of are the RHS initialisers of `const` declarations and also the length specifier in a static array type declaration or initialiser.

```rust
const FOO: u32 = 0x12345678;
const MASK: u32 = 0x00FFFF00;
const FOO_MIDDLE: u32 = FOO & MASK;

const X_SIZE: u64 = 4;
const Y_SIZE: u64 = 2;
let ary = [false; X_SIZE - Y_SIZE + 1];
```

So for us to be able to perform semantic analysis these constant expressions need to be evaluated first.  It's also complicated a bit by the fact that we de-sugar our binary operators into trait methods in `std::ops`.

mohammadfawaz:
Adding the "language feature" label even though this doesn't introduce changes to the language itself.

PR #1948, opened 2022-06-12
Allow more than just a `literal` in `const` initializer expressions 

### LINKED ISSUE #996: Allow the use of `const` with non-primitive types
Currently, when the `const` keyword is used, a literal is expected on the RHS of the declaration. It would be nice if we also allow things like:
```rust
const ETH_ID = ~ContractId::from(0x00..00);
```
I think this requires constant evaluation in the frontend due to the `from()` method call.

vaivaswatha:
Resolves #996.

This does not cover #407. That's because `&` (the example in the Issue) and other operations on values are encoded as `AsmExpression` AST nodes and we'll need to go deeper. That's better done as a separate PR. Ideally we should have first class representation for these operations in the AST than hidden behind just an `op_name`, but that's discussion for another time and place.


sway-core/src/semantic_analysis/const_fold.rs
```diff
@@ -0,0 +1,264 @@
...
+    pub fn get(&self, k: &K) -> Option<&V> {
+        self.container.get(k).and_then(|val_vec| val_vec.last())
+    }
+    pub fn pop(&mut self, k: &K) {
+        match self.container.get_mut(k) {
+            Some(val_vec) => {
+                val_vec.pop();
+                if val_vec.is_empty() {
+                    self.container.remove(k);
+                }
```
otrho, line 47: It wouldn't hurt to just leave the empty vec?

otrho (changes requested):
This is a great start!  Just shows how tricky and large compile time evaluation is, when this is still fairly limited but a significant change.

I have a couple of nits:
* I would prefer everywhere you've used 'fold' to instead be 'eval'.  But this could just be me -- I think of 'constant folding' as an optional middle end optimisation pass.  I think of 'constant evaluation' as front end reduction of declarations, maybe because that's what langs like C++ call it.
* `const_fold.rs` doesn't belong in the `semantic_analysis` directory.  I was going to reorganise the code very soon around IR in the front-end, but now I think this has forced the issue. 🙂 When we introduced IR it was in parallel with the existing ASMgen, so to keep it simple I put everything in `optimize.rs` and `asm_generation/from_ir.rs`.  Both these files are way to big and need to be refactored... I was planning to create a new `ir_generation` module to replace `optimize.rs` and split `optimize.rs` up in there, and this new `const_fold.rs` (or maybe `const_eval.rs` :wink:) would go in there.  We could merge this as is and immediately do the refactor in a new PR though, rather than in this PR.

Actually, in that case, maybe these nits could just be addressed in a subsequent PR.

sway-core/src/semantic_analysis/const_fold.rs
vaivaswatha: > It wouldn't hurt to just leave the empty vec?

That depends. I was thinking that this data structure could someday be useful elsewhere too (where its size would matter), and in which case it's better for us to remove non-existing keys (and reclaim unused memory). I don't have a strong opinion on this, so if you think that removing this code (i.e, not deleting the key) is better, I'll do that.

sway-core/src/error.rs
```diff
@@ -973,11 +973,8 @@ pub enum CompileError {
     Parse { error: sway_parse::ParseError },
     #[error("\"where\" clauses are not yet supported")]
     WhereClauseNotYetSupported { span: Span },
-    #[error(
-        "Initializing a constant value via `const` is currently limited to primitive values \
-        such as numbers, bools, strings or b256s."
-    )]
-    NonLiteralConstantDeclValue { span: Span },
+    #[error("Could not evaluate initializer to a const value.")]
```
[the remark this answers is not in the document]
vaivaswatha: > 

Done.

vaivaswatha:
> This is a great start! Just shows how tricky and large compile time evaluation is, when this is still fairly limited but a significant change.
> 
> I have a couple of nits:
> 
> * I would prefer everywhere you've used 'fold' to instead be 'eval'.  But this could just be me -- I think of 'constant folding' as an optional middle end optimisation pass.  I think of 'constant evaluation' as front end reduction of declarations, maybe because that's what langs like C++ call it.
> * `const_fold.rs` doesn't belong in the `semantic_analysis` directory.  I was going to reorganise the code very soon around IR in the front-end, but now I think this has forced the issue. slightly_smiling_face When we introduced IR it was in parallel with the existing ASMgen, so to keep it simple I put everything in `optimize.rs` and `asm_generation/from_ir.rs`.  Both these files are way too big and need to be refactored... I was planning to create a new `ir_generation` module to replace `optimize.rs` and split `optimize.rs` up in there, and this new `const_fold.rs` (or maybe `const_eval.rs` wink) would go in there too.  We could merge this as is and immediately do the refactor in a new PR though, rather than in this PR.
> 
> Actually, in that case, maybe these nits could just be addressed in a subsequent PR.

Thank you for the review @otrho . I've renamed it to const-eval now. I've kept the renamed file at the original location with the assumption that we'll move it all to the right place later on.

otrho (approved):

PR #1949, opened 2022-06-12
Fix sway formatting turns `op` + `=` into `op` + ` ` + `=` which breaks the build command because it is not a valid expression

### LINKED ISSUE #1945: Sway formatting turns `+=` into `+ =` which breaks the `build` command because it is not a valid expression
# Reproduce

So, here is a fun error.
The following will build however you should use `forc fmt` to ensure that everything is formatted correctly.

```rust
contract;

abi MyContract {
    fn test_function() -> bool;
}

    fn test_function() -> bool {
        let mut a = 1;
        a += 1;
        true
    }
}
```

The formatted file looks as follows

```rust
contract;

abi MyContract {
    fn test_function() -> bool;
}

    fn test_function() -> bool {
        let mut a = 1;
        a + = 1;
        true
    }
}
```

Notice that the `+=` has turned into `+ =`.
This means that you can format the code however you will break the `build` command because `+ =` is not a valid expression.

You have 3 options.

1. Pass the formatting check but break the build so it won't compile
2. Do not format your code and the CI will fail (but it will build if it were to get that far)
3. Do not use `+=` and instead use the long form.

# Screenshots

cc @eureka-cpu

kayagokalp:
closes #1945 

Until we are ready to switch to formatter v2, I am pushing a quick fix as this seems fairly easy to fix and the bug is breaking the build process.

### DIFF
--- a/sway-fmt/src/code_builder.rs
+++ b/sway-fmt/src/code_builder.rs
@@ -10,9 +10,9 @@ use crate::code_builder_helpers::{
 use super::{
     code_builder_helpers::{
         clean_all_whitespace, handle_ampersand_case, handle_assignment_case, handle_colon_case,
-        handle_dash_case, handle_logical_not_case, handle_multiline_comment_case, handle_pipe_case,
-        handle_plus_case, handle_string_case, handle_whitespace_case, is_comment,
-        is_multiline_comment,
+        handle_dash_case, handle_forward_slash_case, handle_logical_not_case,
+        handle_multiline_comment_case, handle_multiply_case, handle_pipe_case, handle_plus_case,
+        handle_string_case, handle_whitespace_case, is_comment, is_multiline_comment,
     },
     code_line::{CodeLine, CodeType},
 };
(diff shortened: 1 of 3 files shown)


Braqzen (dismissed):
Shouldn't this have a test case?

kayagokalp:
I wasn't sure whether we should have a test or not since: 

 - Handling of `=` sign part does not have one too
 - We are going to be removing the v1 formatter soon (wasn't sure if we should add more noise)
 
But I think we can add one so that we are sure it does not touch `+=` as it used to do (if you also think that would be a better approach to follow)

Does something like this seem good?:

```
    #[test]
    fn test_plus_equal() {
        let correct_sway_code = r#"script;
fn main() {
    let mut a = 1;
    a += 1;
}
"#;
        let sway_code = r#"script;
fn main() {
    let mut a = 1;
    a += 1;
}
"#;
        let result = get_formatted_data(sway_code.into(), OPTIONS, None);
        assert!(result.is_ok());
        let (_, formatted_code) = result.unwrap();
        assert_eq!(correct_sway_code, formatted_code);
    }
```

mohammadfawaz:
Shouldn't we also do the same for other operators: `-=`, `*=`, etc.? I suspect hey suffer from the same problem.

kayagokalp:
Yes, that seems correct, I can see that this happens with:

- '-='
- '*='
- '/='

but `<<=` and `>>=` seem to be formatted correctly.

Braqzen (approved):

PR #1951, opened 2022-06-13
Use `--locked` option when building tests to ensure up-to-date lock file

mitchmindtree:
Every now and then the lock files for examples fall out-of-date or are
accidentally omitted from PRs/commits. This means that they sometimes
accidentally slip into other unrelated PRs when someone else runs the
test suites locally.

Using `--locked` causes the command to fail if the existing lock file is
out of date, or if no lock file exists yet. This will help to ensure
lock files must be updated when their associated manifest dependencies
are updated.


kayagokalp (approved):
LGTM, I wondered what was happening with lock file changes that pop out randomly. hopefully, this will clean those commits nicely 🎉

PR #1956, opened 2022-06-13
Allow `self` to be mutable

mohammadfawaz:
This the first step to closing https://github.com/FuelLabs/sway/issues/1188. Allowing `self` to be `mut` unblocks the implementation of `Vec<T>`. We can then enable more complex scenarios involving other `mut` args.

I had to make a small change to the IR verifier to allow storing into a reference type argument directly. In reality, because we inline everything, this change is not really functional. All those arguments go away by the time we get to codegen. Also, mut arguments are disallowed for contract methods anyways, so nothing to worry about there. All of this should be cleaned up as part of the rest of #1188.

With this, we can use `mut self` in the `Vec` library and it seems to be working well so far.


sway-core/src/convert_parse_tree.rs
```diff
@@ -952,10 +952,12 @@ fn fn_args_to_function_parameters(
             .collect::<Result<_, _>>()?,
         FnArgs::NonStatic {
             self_token,
+            mutable_self,
             args_opt,
         } => {
             let mut function_parameters = vec![FunctionParameter {
                 name: Ident::new(self_token.span()),
+                is_mutable: matches!(mutable_self, Some(_mut_token)),
```
otrho, line 960: `mutable_self.is_some()` is probably more readable here.  But not a necessary change.  I was looking at it thinking 'what is `_mut_token` there for?'

sway-core/src/optimize.rs
```diff
@@ -1113,6 +1113,7 @@ impl FnCompiler {
                 .iter()
                 .map(|(name, expr)| TypedFunctionParameter {
                     name: name.clone(),
+                    is_mutable: false,
```
otrho, line 1116: This will need to be consistent with the original function.  I hate this code so much now, often I'm wishing we just inlined the callee here rather than later because of things like this (originally the amount of work to do one or the other was 50/50 and so I just did it this way semi-arbitrarily).

So, if we call a function with a mutable self in Sway, then recreate that function here, and then inline it later, does it still work?

Looking at the tests below, I guess it must. :slightly_smiling_face: Maybe I'm missing something here, but I'd assume just setting mutable to false here would break things.

sway-core/src/convert_parse_tree.rs
mohammadfawaz: Ah yes silly me.

sway-core/src/optimize.rs
mohammadfawaz: `mut` for `self` only matters in type checking which is why this works. That being said, it is odd and I cringed writing it. Mutability here should probably be propagated to arguments in IR and validated in the verifier.

otrho: OK.  We need `Vec` more than we need this to be perfect. :smile:

otrho (dismissed):

otrho (dismissed):

otrho (approved):

PR #1958, opened 2022-06-13
Add the concept of semantic similarity to the type system

### LINKED ISSUE #1103: Generic impl selfs with semantic type constraints have undefined behavior
For example, the compiler is not equipped to handle this case:

```rust
script;

struct DoubleIdentity<T, F> {
  first: T,
  second: F,
}

impl<T, F> DoubleIdentity<T, F> {
 ..
}

impl DoubleIdentity<u8, u8> {
 ..
}

// this case ↓↓↓↓↓
impl<T> DoubleIdentity<T, T> {
  fn new_from_single(x: T) -> DoubleIdentity<T, T> {
    DoubleIdentity {
      first: x,
      second: x,
    }
  }
}


fn main() -> u32 {
  ..
}

```

### LINKED ISSUE #1747: Unable to find associated methods for types in some cases
Potentially related: a recurring issue where `unwrap()` can't be found when trying to call `result.unwrap()`. Turns out this can be worked around by explicitly annotating the type:

```rs
let sender: Result<Sender, AuthError> = msg_sender(); // if you remove `: Result<Sender, AuthError>` this doesn't work!
sender.unwrap();
```

_Originally posted by @adlerjohn in https://github.com/FuelLabs/sway/issues/1734#issuecomment-1139883131_

emilyaherbert:
# The previous way:

When type checking a method call, the compiler does a look-up to find the definition for the `TypedFunctionDefinition` given the `TypeInfo` of the type upon which the method is being called. So, given:

```rust
struct DoubleIdentity<T, F> {
    first: T,
    second: F,
}

impl<T, F> DoubleIdentity<T, F> {
    fn get_second(self) -> F {
        self.second
    }
}

fn main() {
    let a = DoubleIdentity {
        first: 0u8,
        second: 1u8
    };
    let b = a.get_second();
}
```

The call to `a.get_second()` will do a lookup for `get_second` under the `DoubleIdentity<u8, u8>` type, inside of the `TraitMap` inside of the namespace, retrieving all methods that are inserted under the `DoubleIdentity<u8, u8>` type. `DoubleIdentity<u8, u8>` is added to the `TraitMap` during monomorphization (the process by which the `DoubleIdentity<T, F>` type is resolved to the `DoubleIdentity<T', F'>` type, where `T'` and `F'` are then in the next step resolved to `u8` and `u8`.) During monomorphization, the methods defined for `DoubleIdentity<T, F>` are copied, `T` and `F` are replaced with `T'` and `F'`, and the new copies are inserted in the `TraitMap` under `DoubleIdentity<T', F'>`. When `T'` and `F'` are resolved to `u8` and `u8`, the `T'` and `F'` inside of the methods for `DoubleIdentity<T', F'>` are also resolved behind-the-scenes by virtue of the type engine.

However, this breaks down when you encounter semantically-bound type implementations like:

```rust
impl<T> DoubleIdentity<T, T> {
    fn get_first(self) -> T {
        self.first
    }
}
```

because the type system does not have a way of knowing that `DoubleIdentity<u8, u8>` should take both the methods defined under `DoubleIdentity<T, F>` and the methods defined under `DoubleIdentity<T, T>`.

# The new way:

Similar to the old way, when type checking a method call, the compiler does a look-up to find the definition for the `TypedFunctionDefinition` given the `TypeInfo` of the type upon which the method is being called. So, given:

```rust
struct DoubleIdentity<T, F> {
    first: T,
    second: F,
}

impl<T> DoubleIdentity<T, T> {
    fn get_first(self) -> T {
        self.first
    }
}

impl<T, F> DoubleIdentity<T, F> {
    fn get_second(self) -> F {
        self.second
    }
}

impl DoubleIdentity<u8, u8> {
    fn add(self) -> u8 {
        self.first + self.second
    }
}

fn main() {
    let a = DoubleIdentity {
        first: 0u8,
        second: 1u8
    };
    let b = DoubleIdentity {
        first: true,
        second: false,
    };
    let c = DoubleIdentity {
        first: 0u64,
        second: "hi"
    };

    let d = a.get_first();
    let e = a.get_second();
    let f = a.add();

    let g = b.get_first();
    let h = b.get_second();
    // should fail
    let i = b.add();

    // should fail
    let j = c.get_first();
    let k = c.get_second();
    // should fail
    let l = c.add();
}
```

The call to `a.get_second()` still does a lookup for `get_second` under the `DoubleIdentity<u8, u8>` type, inside of the `TraitMap` inside of the namespace. _Except now it retrieves all methods inserted under all types for which `DoubleIdentity<u8, u8>` is a subset type_. The concept of type subsets are defined here:
https://github.com/FuelLabs/sway/blob/ff97c79c0e0abd652db3a7b69e5dfd6ba7dc5f5f/sway-core/src/type_engine/type_info.rs#L930

At a high level, type `T` is a subset of type `T'` if `T` maintains the same semantic constraints as `T'` and is either equally as "generic" or less "generic". So `DoubleIdentity<u8, u8>` is a subset of both the `DoubleIdentity<T, F>` type and the `DoubleIdentity<T, T>` type, `DoubleIdentity<T, T>` is a subset of `DoubleIdentity<T, F>`, but `DoubleIdentity<T, F>` is _not_ a subset of `DoubleIdentity<T, T>`.

Now, types are only inserted into the `TraitMap` during type checking of `impl` blocks. So, the `TraitMap` should expect to see an entry for `DoubleIdentity<u8, u8>` after the `impl DoubleIdentity<u8, u8>` block, but should never see an entry for `DoubleIdentity<u64, str[2]>`.

# The bug from #1747

The compiler was having an issue where, in certain cases, it was unable to find the associated methods for some types. Given:

```rust
let sender: Result<Sender, AuthError> = msg_sender(); // if you remove `: Result<Sender, AuthError>` this doesn't work!
sender.unwrap();
```

the compiler would fail to find the associated method `unwrap()` for the type `Result<Sender, AuthError>`. Here is the definition of `msg_sender`: https://github.com/FuelLabs/sway/blob/3a2e7199245cc14e1ec335f513d439237b0e886d/sway-lib-std/src/chain/auth.sw#L38

`msg_sender` returns a `Result<Sender, AuthError>`. In the old way of doing things, the compiler would fail to find `unwrap` for a list of compounding reasons. 1) Only certain "locations" of type definitions would trigger monomorphization and the function return type was not one of them. 2) Any iterations of monomorphization that occur while type checking the function definition are caught by the internal namespace of the function definition, meaning that even if a `Result<Sender, AuthError>` was monomorphized and its associated methods copied to the new type in the `TraitMap`, then those associated methods would be strictly contained within the namespace of the function definition, and 3) because the mechanism for associated method lookup does not account for semantically similar types, then `unwrap` could not be found.

To be clear, of the elements that were contributing to the bug, (1) and (2) are not bad things---they are desired behavior. (3) is really where to problem needed to be fixed. Because this PR introduces the concept of semantic similarity in types, during method lookup, looking up the type `Result<Sender, AuthError>` does not require that `Result<Sender, AuthError>` be in the `TraitMap` of the current namespace, it instead requires that `Result<T, E>` be in the `TraitMap` of the current namespace. And that definition is imported from the `impl` block in the call to `use std::result::*;`!

---

This allows the compiler to:
1. reduce overhead for monomorphization
2. reduce complexity in the type engine and namespace engine
3. anticipate the addition of the declaration engine and eliminating function inlining
4. support type's and `impl` blocks with semantic trait constraints

- Blocked by #1942
- Blocked by #1963
- Closes #1103
- Closes #1747


otrho:
Looks good to me, nothing stood out as potential problems, lots of tests... :+1:  Don't let Syd see the `|x| ...` closures though. :wink:

mohammadfawaz:
@Braqzen is having an issue with something like:
```
```
where `unwrap()` is not found if though `result::*` is imported. Does this PR fix that as well? Of so, it's worth adding that pattern as a test case.

emilyaherbert:
> @Braqzen is having an issue with something like:
> let sender = msg_sender().unwrap();
> where unwrap() is not found even though result::* is imported. Does this PR fix that as well? If so, it's worth adding that pattern as a test case.

Yes it does---i will add that test case.

otrho (approved):

PR #1960, opened 2022-06-13
Revert "Implement Add for b256"

nfurfaro:
Reverts FuelLabs/sway#1707

>We will need to revert this PR (https://github.com/FuelLabs/sway/pull/1707). This PR introduces syntax that imports an external Sway library into a nested submodule in a different Sway project, for which Sway does not currently have support (https://github.com/FuelLabs/sway/blob/b7b6bc9c5b3730402c792c984629a9848794fb7c/sway-lib-core/src/ops.sw#L3). The only reason that this passed CI is because 1) the error is being thrown away here ([https://github.com/FuelLabs/sway/blob/b7b6bc9c5b3730402c792c984629a9848794fb7c/swa[…]y-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs](https://github.com/FuelLabs/sway/blob/b7b6bc9c5b3730402c792c984629a9848794fb7c/sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs#L417)) and 2) because there is a bug in the import+type system which allows declarations to leak into the global namespace (this PR is a partial fix to that https://github.com/FuelLabs/sway/pull/1941, but it won't solve it entirely)


nfurfaro:
Going to try reverting manually vs by using the GitHub UI.

PR #1962, opened 2022-06-13
Enum formating for sway-fmt-v2

### LINKED ISSUE #1873: `swayfmt` v2 should be able to format enums
This issue is for tracking the progress of `sway-fmt-v2` implementation for `enums`.

related & part of #1516.

kayagokalp:
- ~This PR is WIP~. I believe this is ready to give us a good starting point/something to iterate on with feedback.

## About this PR

closes #1873.

~Blocked by: #1968~

This PR implements the basic `enum` formatting for `sway-fmt-v2`.

I tested with the following inputs.

```
contract;

enum Color {
    Blue: (),
    Green: (),
    Red: (),
    Silver: (),
    Grey: (),
}
```

To test enum variant alignment you can create a `swayfmt.toml` at the root of your test project directory and use the following config:

```
[structures]
enum_variant_align_threshold = 20
```

Without alignment applied (default) output will be like:

```
enum Color {
 Blue : (),
 Green : (),
 Red : (),
 Silver : (),
 Grey : (),
}
```

With alignment:

```
enum Color {
 Blue   : (),
 Green  : (),
 Red    : (),
 Silver : (),
 Grey   : (),
}
``` 

Note:  since the `contract` part in the example is not handled yet by a part of the `sway-fmt-v2` it is deleted.  It will be corrected once we handle that part too (possibly in a separate PR : #1995 ).

If you would like to try something a little bit crazier you can try something like:

```
contract;

enum Color {
    Blue: (), Green: (),
    		Red: (),
  Silver: (),
    				Grey: (), }
```


sway-fmt-v2/src/utils/helpers.rs
```diff
@@ -0,0 +1,37 @@
+use crate::Formatter;
```
eureka-cpu, line 1: We should avoid using ambiguous file names like `helpers.rs` and instead give descriptive names as to what utility the file provides like `bracket.rs` might be better here since it should only be handling brackets. If you intend on adding more punctuation handling in this file you could call it `punctuation.rs`.

sway-fmt-v2/src/utils/helpers.rs
```diff
@@ -0,0 +1,37 @@
...
+use sway_parse::Ty;
+use sway_types::Spanned;
+
+/// Handling of a ty, currently this does not apply any formatting.
+/// It simply pushes the string version of the ty.
+pub fn handle_ty(ty: &Ty, push_to: &mut String) {
```
eureka-cpu, line 7: since this function has more to do with changing the type to `String`, you may want to move it to `sway_parse::Ty` as a method on `Ty` and give it a more clear name like `to_string()`

sway-fmt-v2/src/utils/helpers.rs
```diff
@@ -0,0 +1,37 @@
...
+    push_to.push_str(ty.span().as_str());
+}
+
+/// Handles bracket open scenerio. Checks the config for the placement of the bracket.
+/// Modifies the current shape of the formatter.
+pub fn handle_bracket_open(push_to: &mut String, formatter: &mut Formatter) {
```
eureka-cpu, line 13: small nit but maybe make this `handle_open_bracket` & `handle_close_bracket` instead

sway-fmt-v2/src/utils/helpers.rs
```diff
@@ -0,0 +1,37 @@
...
+    let bracket_on_new_line = formatter.config.items.item_brace_style;
+    let mut shape = formatter.shape;
+    match bracket_on_new_line {
+        crate::config::items::ItemBraceStyle::AlwaysNextLine => {
```
eureka-cpu, line 17: we should bring the imports this to the top, and just match on `ItemBraceStyle::AlwaysNextLine` etc.

eureka-cpu (changes requested):
Looking good so far, just some placement stuff and naming conventions

sway-fmt-v2/src/utils/helpers.rs
kayagokalp: We can continue to use `to_string` of` ty` and add this kind of routine later when we need it.

sway-fmt-v2/src/utils/punctuation.rs
```diff
@@ -0,0 +1,30 @@
+use crate::config::items::ItemBraceStyle;
```
eureka-cpu, line 1: We can merge these two imports here

sway-fmt-v2/src/items/item_enum.rs
```diff
@@ -1,8 +1,78 @@
...
+            formatted_code.push_str(" : ");
+
+            // Currently does not apply custom formatting for ty,
+            let current_variant_length = variant_length[index];
+
+            if current_variant_length < max_valid_variant_length {
+                // We need to add allignment between : and ty
+                // max_valid_variant_length: the length of the variant that we are taking as a reference to allign
+                // current_variant_length: the length of the current variant that we are trying to format
+                let required_allignment = max_valid_variant_length - current_variant_length;
```
eureka-cpu, line 63: small nit: `required_alignment`

sway-fmt-v2/src/items/item_enum.rs
```diff
@@ -1,8 +1,78 @@
...
+                // TODO: Improve handling this
+                formatted_code.push_str(&(0..required_allignment).map(|_| ' ').collect::<String>());
```
eureka-cpu, line 65: here if we are starting from the first element, I'm pretty sure we can just do `..required_alignment` instead of `0..`

eureka-cpu (commented):
a few more small things

sway-fmt-v2/src/utils/punctuation.rs
```diff
@@ -0,0 +1,30 @@
...
+use crate::Formatter;
+
+/// Handles bracket open scenerio. Checks the config for the placement of the bracket.
+/// Modifies the current shape of the formatter.
+pub fn handle_open_bracket(push_to: &mut String, formatter: &mut Formatter) {
```
eureka-cpu, line 6: One thing we need to be careful of, is that if we are using this to maintain brackets for all items that we add a match statement so that it will target each specific case. For now you could add the match statement in and have everything else as `todo!()`

eureka-cpu: An instance where this might occur is in `use` statements where we don't want a space before `{` like `use a:: {b, c, d};` which would break the code.

eureka-cpu: If it's not too much trouble, could you push this to a separate PR as well?

sway-fmt-v2/src/items/item_enum.rs
kayagokalp: Unfortunately providing just the upper limit like `..required_alignment` creates a `RangeTo` rather than a `Range` and `RangeTo` cannot be used as an iterator according to the [doc](https://doc.rust-lang.org/std/ops/struct.RangeTo.html). I couldn't find a way to use it like that but if you have any suggestions would be happy to apply

sway-fmt-v2/src/utils/punctuation.rs
kayagokalp: 👍 opened-up #1968.

eureka-cpu (dismissed):
Looks good! Awesome job!

eureka-cpu:
The program type formatting will need to be done in a separate PR, this is okay for now I think

sway-fmt-v2/src/items/item_enum.rs
```diff
@@ -4,10 +4,77 @@ use crate::{
...
+        // Find the maximum length in the variant_length vector that is still smaller than enum_variant_align_threshold.
+        let mut max_valid_variant_length = 0;
+
+        variant_length.iter().for_each({
+            |length| {
+                if *length > max_valid_variant_length && *length < enum_variant_align_threshold {
+                    max_valid_variant_length = *length;
+                }
+            }
+        });
```
mitchmindtree, line 46: I don't think the extra nested braces are necessary here, e.g. should be able to do:

```rust
variant_length.iter().for_each(|length| {
    if *length > max_valid_variant_length && *length < enum_variant_align_threshold {
        max_valid_variant_length = *length;
    }
});
```

mitchmindtree (commented):
Looking good!

Is it possible to add small unit tests for each of these item implementations as they're added?

JoshuaBatty (approved):
Looks good.

PR #1966, opened 2022-06-14
Refactor `optimize.rs` into an `ir_generation` module.

otrho:
This is reorganising IRgen, particularly after #1948 introduced the `const_eval.rs` which needed to go somewhere sensible.

Everything has gone into `ir_generation` and it exposes only `compile_program()` and the `PurityChecker` to the crate, all else is private.


mohammadfawaz (approved):
LGTM! Much cleaner.

vaivaswatha (approved):

PR #1968, opened 2022-06-14
`sway-fmt-v2` brace handler and implementation for enum added

### LINKED ISSUE #1967: Add brace handling to `sway-fmt-v2`
We need to have a baseline implementation of handling different punctuations possibly starting from `{` and `}`.

kayagokalp:
## About this PR

closes #1967.

This PR adds the base implementation for `brace` handling and implements the case of `enum` so that it can be an example for other item types as well (`use`, `struct`, ...)


eureka-cpu:
After looking over this I think maybe having another match expression for each type of punctuation may not be the best direction. I think maybe a better approach might be to just implement `Delimiter` as a trait and have functions for each type of delimiter like so: 
```
pub trait CurlyDelimiter { // there may be a better name for this, but just be descriptive
    fn handle_open_bracket(push_to: &mut String, formatter: &mut Formatter);

    fn handle_closed_bracket(push_to: &mut String, formatter: &mut Formatter);
}
```

And this would look like:

```
impl Delimiter for ItemEnum {...}
```

Which you could then use in the `format` function, this would take out the need to match for the `VirtualItemKind`.

eureka-cpu:
just realized, looking at `sway-parse` what we've been calling `bracket` is actually called `brace`

kayagokalp:
Oh, I see bracket is a more general name (i.e, `<` is a type of bracket) while `{` is a brace.

eureka-cpu:
> Oh, I see bracket is a more general name (i.e, `<` is a type of bracket) while `{` is a brace.

We could also call the file `bracket.rs` or `delimiter.rs` I think, since that is all that file should handle

sway-fmt-v2/src/items/item_enum.rs
```diff
@@ -1,8 +1,39 @@
...
 
 impl Format for ItemEnum {
     fn format(&self, _formatter: &mut Formatter) -> FormattedCode {
         todo!()
     }
 }
+
+impl CurlyDelimiter for ItemEnum {
+    fn handle_open_brace(push_to: &mut String, formatter: &mut Formatter) {
+        let brace_on_new_line = formatter.config.items.item_brace_style;
```
eureka-cpu, line 16: this could just be called `brace_style` I think

eureka-cpu (approved):

mitchmindtree (approved):

PR #1969, opened 2022-06-14
Improvements around `forc plugins` command

tritao:
Hey, so I was getting started with Sway, and tried calling `forc plugins` command while I was exploring the functionality.

This lead to a crash:

```
~/dev/sway/target/debug$ forc plugins
Installed Plugins:
thread 'main' panicked at 'Could not get plugin description.: Os { code:
2, kind: NotFound, message: "No such file or directory" }',
forc/src/cli/commands/plugins.rs:43:10
stack backtrace:
   0: rust_begin_unwind
at
/rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
at
/rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:143:14
   2: core::result::unwrap_failed
at
/rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/result.rs:1785:5
   3: core::result::Result<T,E>::expect
at
/rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/result.rs:1035:23
   4: forc::cli::commands::plugins::parse_description_for_plugin
at
/home/joao/dev/sway/forc/src/cli/commands/plugins.rs:40:16
   5: forc::cli::commands::plugins::format_print_description
at
/home/joao/dev/sway/forc/src/cli/commands/plugins.rs:81:23
   6: forc::cli::commands::plugins::print_plugin
at
/home/joao/dev/sway/forc/src/cli/commands/plugins.rs:97:5
   7: forc::cli::commands::plugins::exec
at
/home/joao/dev/sway/forc/src/cli/commands/plugins.rs:28:21
```

The first commit deals with this, which happens due to not using the full path while exec'ing the plugin, so it just changes the function to receive the full path. The docs say `unwrap` should not generally be used because it can lead to a panic, but here we know we have a valid path, so I thought it was okay. If that is not the case, let me know.

After that, we get: 

```
Installed Plugins:
forc-cd4662b67fa2a5c9
forc-7550923f7d96c76d
forc-7eada75316346bfc
forc-fac57f493a01e50b
forc-lsp
forc-fmt-v2
forc-explore
forc-fmt
forc-gm
```

The second commit just skips those executables from being considered plugins. While I generally don't like hardcoding the `deps` folder in the check, it seemed like a sensible solution for this not too frequent edge case, since it will probably only happens for developers. What do you think?


kayagokalp (commented):
Congrats on your first PR 🎉 (sorry if it is not 👀). Thank you for the fix and the fix seems good to me. But I think you may need to run `cargo fmt` since CI is resulting in an error.

kayagokalp (dismissed):
LGTM will defer the final approval to @ra0x3 or @FuelLabs/tooling

ra0x3:
@tritao Does this PR have an associated issue?

tritao:
> @tritao Does this PR have an associated issue? I see the PR fixes something (from your PR description), but trying to see what the originating issue is/was

No, found the issue today and sent the PR directly. Should I create one?

ra0x3:
@tritao For housekeeping purposes, yes that would be great if you could create one :)

forc/src/cli/commands/plugins.rs
```diff
@@ -78,7 +78,7 @@ fn format_print_description(
             .to_string()
     };
 
-    let description = parse_description_for_plugin(&display);
+    let description = parse_description_for_plugin(path.to_str().unwrap());
```
ra0x3, line 81: Tested and works fine. I think we can leave this as is though.

forc/src/cli/plugin.rs
```diff
@@ -63,6 +63,11 @@ fn is_executable(path: &Path) -> bool {
 
 /// Whether or not the given path points to a valid forc plugin.
 fn is_plugin(path: &Path) -> bool {
+    // This is to check that we do not match executables like `deps/forc-cd4662b67fa2a5c9`,
```
ra0x3, line 66: +1 for the helpful comment

ra0x3 (changes requested):
Looks good. One minor/small requested change.

forc/src/cli/commands/plugins.rs
ra0x3: @tritao 
- I agree :) 
- I'll defer to @mitchmindtree on this as I'm not sure why there are arbitrary folders in `~/.cargo/bin` or how many there can/might be. But other than that, LGTM

mitchmindtree: @tritao instead of passing in a `&str` here, it looks like we might be better off changing the `parse_description_for_plugin` function to take a `&Path` instead.

We can still call `Command::new(plugin)` directly inside `parse_description_for_plugin` (without any str conversion) as `Path` implements `AsRef<OsStr>`.

We should also update its comment to reflect that it expects the canonical path to the plugin, not just the plugin name.

forc/src/cli/plugin.rs
mitchmindtree: Out of curiosity, where these `deps` coming from? I'm curious how they end up in your `PATH`?

I don't think it's feasible for us to filter out every possible unwanted directory of executables that a user might have careless placed in their `PATH`, however if this is particularly common for some reason I'm open to it.

I think the proper fix here is: Rather than `panic!`ing when failing to retrieve a description for a plugin, we should instead use an empty description, or perhaps some default string along the lines of "Unable to retrieve a description for this plugin."

mitchmindtree (changes requested):
Congrats on the first PR @tritao and thanks for catching this!

I've left some comments inline.

forc/src/cli/plugin.rs
mitchmindtree: Thinking on this further, if we fail to retrieve a description, perhaps we can just skip the plugin altogether? We could instead update our definition of `forc` plugins to *require* a description in order to be enumerated under the `forc plugins` command?

tritao: > Out of curiosity, where these `deps` coming from? I'm curious how they end up in your `PATH`?

They exist after a `cargo build` inside `target/debug/deps` folder. If `PATH` contains `sway/target/debug`, then it triggers this issue, since files are recursively enumerated when searching for plugins.

> I don't think it's feasible for us to filter out every possible unwanted directory of executables that a user might have carelessly/accidentally placed in their `PATH`, however if this is particularly common for some reason I'm open to it.

Agreed, but in this case it happens in a default build environment, so I thought it was worth dealing with it directly.

Might be worth noting these executables respond to `-h`:

```
~/dev/sway/target/debug$ deps/forc-cd4662b67fa2a5c9 -h
Usage: -h [OPTIONS] [FILTERS...]

Options:
        --include-ignored 
                        Run ignored and not ignored tests
        --ignored       Run only ignored tests
...
```

And that is why they are being considered as plugins. 

I think the most reliable fix here would be to make sure that forc plugins provide some kind of well-specified manifest (say with a `--manifest` flag which could return a JSON description). That would make sure there are no conflicts with other executables providing an `-h` flag.

For my first PR, however, I decided to take the easy route and just fix the panic, and skip them directly since I wasn't sure it was worth spending too much time on this.

If you think it's worth the time fixing this in a more solid way let me know.

forc/src/cli/commands/plugins.rs
tritao: Thanks, that is cleaner for sure, will update the PR

forc/src/cli/plugin.rs
mitchmindtree: Thanks for the explanation @tritao!

> I think the most reliable fix here would be to make sure that forc plugins provide some kind of well-specified manifest (say with a --manifest flag which could return a JSON description).

That sounds like a great idea! But probably best for a future PR like you mention.

> They exist after a cargo build inside target/debug/deps folder. If PATH contains sway/target/debug, then it triggers this issue, since files are recursively enumerated when searching for plugins.

I'm still a little unsure of how `sway/target/debug` ends up in your `PATH` in this case? Do you normally manually add the target directory of projects you work on to your `PATH`?

I think the more conventional approach for having Rust bins available on `PATH` is to have `~/.cargo/bin` on the `PATH` - that way when `cargo install <crate>` is ran, the `<crate>` bin is available via `PATH`.

If you want the debug build to be available on your `PATH` while you're testing etc, another option is to use `cargo install --debug --path /path/to/crate`.

When I'm testing in Sway locally, I find it more useful to make a habit of running `forc` with the `cargo run` command to ensure I'm always using the local state, and to avoid the need to have two steps. E.g.

```sh
cargo run --bin forc -- build --path /path/to/forc/project
```

mitchmindtree:
My review was requested again recently, however I still have some outstanding Qs [here](https://github.com/FuelLabs/sway/pull/1969#discussion_r898592808). To clarify, the premise for some of these changes is that it may be common to have the `deps` of a target directory on a user's `PATH`, however I'm not so sure that this is the common case and I suspect these errors might result from misuse of `PATH`, hence the Qs in the linked comment.

forc/src/cli/plugin.rs
tritao: I added the target dir to the `PATH` because I was just getting started and didn't know any better regarding the workflow.

Thanks for explaining your workflow with `cargo`, I can just use that going forward.

tritao:
> My review was requested again recently, however I still have some outstanding Qs [here](https://github.com/FuelLabs/sway/pull/1969#discussion_r898592808). To clarify, the premise for some of these changes is that it may be common to have the `deps` of a target directory on a user's `PATH`, however I'm not so sure that this is the common case and I suspect these errors might result from misuse of `PATH`, hence the Qs in the linked comment.

I've updated the PR with just the fix to use the full path to the plugin, which seems like the correct thing to do either way.

mitchmindtree (approved):
Perfect, thanks so much @tritao!

adlerjohn:
Poking @ra0x3 for re-review

kayagokalp (commented):
lgtm, deferring to @ra0x3 for final approval

ra0x3 (approved):

Issue #1943, opened 2022-06-11
Using `+=` on a `storage` variable results in an error stating that the storage is not scope


Braqzen:
# Reproduce

If you comment out `test3()` in the `abi` and trait `impl` then the code will build otherwise the error listed below is presented.

```rust
contract;

abi MyContract {
    fn test1() -> bool;
    fn test2() -> bool;
    fn test3() -> bool;
}

storage {
    value: u64,
}

impl MyContract for Contract {
    fn test1() -> bool {
        let mut a = 1;
        a += 1;
        true
    }

    fn test2() -> bool {
        storage.value = storage.value + 1;
        true
    }

    fn test3() -> bool {
        storage.value += 1;
        true
    }
}
```

# Screenshots
![1](https://user-images.githubusercontent.com/103777923/173186543-71539b33-7475-4619-bc58-c34bf0a43ee9.png)

cc @FuelLabs/sway-compiler

mohammadfawaz:
Nice catch. Yeah I guess `+=` does not support storage accesses yet. Should be an easy fix in the parser.

PR #1970, opened 2022-06-14
Support `+=` and other similar operators for storage accesses

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/1943

The code I added is pretty similar to https://github.com/FuelLabs/sway/blob/95518c2e42f535e0d5c55181e164ee2143a722f2/sway-core/src/convert_parse_tree.rs#L1595
The idea is to special case "storage" like we do in other places.


sway-core/src/convert_parse_tree.rs
```diff
@@ -2863,11 +2863,36 @@ fn assignable_to_expression(
...
+            let mut idents = vec![&name];
+            let mut base = &*target;
+            let storage_access_field_names_opt = loop {
+                match base {
+                    Assignable::FieldProjection { target, name, .. } => {
+                        idents.push(name);
+                        base = target;
+                    }
+                    Assignable::Var(name) => {
+                        if name.as_str() == "storage" {
```
otrho, line 2876: Man, I was looking at this code trying to work out how it was related to storage stuff, and here it is!  Feels a bit hacky matching against a string like this -- perhaps the parser should become more storage-aware?

otrho (approved):

sway-core/src/convert_parse_tree.rs
mohammadfawaz: Absolutely. Let me create an issue. 

I think Andrew probably wanted to commonize storage accesses and subfield accesses as much as possible.

mohammadfawaz: https://github.com/FuelLabs/sway/issues/1986

PR #1974, opened 2022-06-14
Prevent failure while creating BuildPlan from lock file in the case of a dependency removal

### LINKED ISSUE #1778: `BuildPlan::from_lock()` should be able to create a `BuildPlan` even there is a package removal in the manifest file
Currently `BuildPlan::from_lock()` fails when there is a package removal in the manifest file. We should be able to get some sort of an "unsafe" `BuildPlan` rather than a complete failure. Because #1686 needs to be able to generate a `diff` on the `BuildPlan` generated from the `lock` file and the updated manifest file. 

## Proposed Solution

We can do that by introducing an "unsafe" version of the `graph_to_path_map()`. Current `graph_to_path_map()` returns an error ("dependency required for path reconstruction has been removed from the manifest") when the dependency coming from the graph is not included in its parent manifest file (which is the case for the removed packages). Instead, we can point them to a default removed path (something like `$HOME/.forc/removed`). This way we can create an unsafe version of the BuildPlan which can be used to produce the diff. Once we have the diff `apply_new_manifest()` from #1686 will use that diff to create an actual BuildPlan representing the updated manifest file.

kayagokalp:
## About this PR
~Still in WIP, opening up for visibility. I will be adding testing steps~

- [x]  Merge #1836 first and add required patching checks to this one as well

closes #1778.

### Motivation
Before this PR, `BuildPlan` generation from the lock file was failing if after the lock file is generated, a dependency is removed from the manifest file (specifically if the removed dependency is a path dependency).

This was resulting in the `BuildPlan` being generated from scratch. To generate the `BuildPlan` from scratch we generally need connectivity since `std` is implicitly added as a git dependency (if we do not explicitly prevent it with `implicit-std = false`). Ideally, we should be able to remove any dependency in offline mode. 

With this PR, we should be able to create the `BuildPlan` from the lock file even if a path dependency is removed. This *should* mean `apply_pkg_diff` will be able to remove any dependency in `offline` mode.

### To Test

1. run `forc new parent`
2. run `forc new --library child`
3. add child as a dependency to parent
4. run `forc build` in parent
5. remove the child from parent's dependencies
6. run `forc build` in parent

This should output something like

```
Creating a new `Forc.lock` file. (Cause: lock file did not match manifest `diff`)
```

Rather than 

```
Creating a new `Forc.lock` file. (Cause: dependency required for path reconstruction has been removed from the manifest)
```


kayagokalp:
I found out a bug about this one (as it can be seen from CI failure) so leaving this as a draft and will come back to here after I am done with some formatter stuff.

forc-pkg/src/pkg.rs
```diff
@@ -163,6 +163,13 @@ pub struct BuildPlan {
     graph: Graph,
     path_map: PathMap,
     compilation_order: Vec<NodeIx>,
+    /// The differences between graph described by lock_file and the graph that actually constructed.
+    ///
+    /// If the BuildPlan is constructed from a manifest file, this will be empty.
+    ///
+    /// If the BuildPlan is constructed from a lock file there might be some differences between the graph described by the lock file and the one that actually created.
+    /// The differences between them are the path dependencies that are removed from the manifest file after the lock file is created.
+    removed_deps: Vec<DependencyName>,
```
kayagokalp, line 172: I am not happy to put a `removed_deps` field in the BuildPlan but couldn't find a cleaner way to pass the `removed_deps` vec from `graph_to_path_map` to validate. So any input would be highly appreciated 🙌

mitchmindtree: Perhaps rather than storing this in the `BuildPlan` itself, we can return this in a tuple alongside `BuildPlan::from_lock`?

forc-pkg/src/pkg.rs
```diff
@@ -271,18 +279,23 @@ impl BuildPlan {
...
         })
     }
 
     /// Attempt to load the build plan from the `Lock`.
     pub fn from_lock(proj_path: &Path, lock: &Lock, sway_git_tag: &str) -> Result<Self> {
-        let graph = lock.to_graph()?;
+        let mut graph = lock.to_graph()?;
+        let tmp_compilation_order = compilation_order(&graph)?;
+        let (path_map, removed_deps) =
+            graph_to_path_map(proj_path, &mut graph, &tmp_compilation_order, sway_git_tag)?;
```
mitchmindtree, line 291: It would be nice if we could keep `graph_to_path_map` as a pure function, particularly as its name doesn't really imply any modification of the `graph` itself.

Perhaps rather than mutating the `graph` *during* this function, `graph_to_path_map` can return the set of removed dependencies as something like a `HashSet<NodeIx>`, and then this can be used to update the `graph` in a following step?

mitchmindtree (commented):
Thanks for diving into this @kayagokalp, it's certainly a bit of an awkward one to address.

I'm starting to wonder if we should be doing the `validate` step as a part of the `BuildPlan` construction... Ideally it shouldn't be possible to construct a `BuildPlan` that is invalid. I think originally the idea behind having the `validate` method be in a following step was to allow for the comparison of the build plans to work out the package diff to print to the user, but perhaps this can just be returned alongside the `BuildPlan` constructor?

kayagokalp:
I am applying the changes requested rn and while doing so realized that introduction of #1836 will have a tiny effect on this one. So I would like to merge that one first if possible so that I can add the required changes in this PR as well

kayagokalp:
> I'm starting to wonder if we should be doing the validate step as a part of the BuildPlan construction... Ideally it shouldn't be possible to construct a BuildPlan that is invalid. I think originally the idea behind having the validate method be in a following step was to allow for the comparison of the build plans to work out the package diff to print to the user, but perhaps this can just be returned alongside the BuildPlan constructor?

That sounds better to me as I feel like it would allow us to do some extra cleaning as well. Should I file an issue and address it with a separate PR or do you think we should have it with this one?

forc-pkg/src/pkg.rs
```diff
@@ -279,37 +271,54 @@ impl BuildPlan {
...
         })
     }
 
     /// Attempt to load the build plan from the `Lock`.
-    pub fn from_lock(proj_path: &Path, lock: &Lock, sway_git_tag: &str) -> Result<Self> {
+    pub fn from_lock(
+        proj_path: &Path,
+        lock: &Lock,
+        sway_git_tag: &str,
+    ) -> Result<(Self, Vec<String>)> {
```
mitchmindtree, line 282: `Vec<String>` is a very ambiguous return type and it's not at all clear from reading the function signature what it represents. Either we should update the docs for this function, or use a type to represent the deps removed due to invalid paths, or both.

forc-pkg/src/pkg.rs
```diff
@@ -279,37 +271,54 @@ impl BuildPlan {
...
+                path_map,
+                compilation_order,
+            },
+            removed_deps_vec,
+        ))
     }
 
     /// Attempt to load the build plan from the `Forc.lock` file.
-    pub fn from_lock_file(lock_path: &Path, sway_git_tag: &str) -> Result<Self> {
+    pub fn from_lock_file(lock_path: &Path, sway_git_tag: &str) -> Result<(Self, Vec<String>)> {
```
mitchmindtree, line 307: Same as above w.r.t. `Vec<String>` in return type.

forc-pkg/src/pkg.rs
```diff
@@ -275,28 +275,50 @@ impl BuildPlan {
...
     }
 
     /// Ensure that the build plan is valid for the given manifest.
-    pub fn validate(&self, manifest: &Manifest, sway_git_tag: &str) -> Result<PkgDiff> {
+    pub fn validate(
+        &self,
+        removed: Vec<String>,
+        manifest: &Manifest,
+        sway_git_tag: &str,
+    ) -> Result<PkgDiff> {
```
mitchmindtree, line 319: Might be useful for this method's docs to mention that the `removed` packages represent the set of packages already removed from the graph due to invalid paths.

forc-pkg/src/pkg.rs
```diff
@@ -406,7 +429,7 @@ fn remove_deps(
             .is_none()
             || to_remove
                 .iter()
-                .any(|removed_dep| removed_dep.1 == graph[node].unpinned(path_map))
+                .any(|removed_dep| *removed_dep == graph[node].unpinned(path_map).name)
```
mitchmindtree, line 432: Are you sure it's enough to only check the `name` here? Keep in mind that a dependency graph can contain two instances of the same package at different versions or branches. Perhaps we should re-add the `Pkg` field to the `to_remove` list here?

forc-pkg/src/pkg.rs
```diff
@@ -750,20 +773,79 @@ pub fn compilation_order(graph: &Graph) -> Result<Vec<NodeIx>> {
...
+                    graph.add_edge(parent_node, child, graph[child].name.clone());
+                }
+            }
+            println_red(&format!("  Removing {}", dep_name));
+            // Remove the deleted node
+            graph.remove_node(dep);
+        }
+    }
+    Ok(())
+}
```
mitchmindtree, line 833: Is it possible to re-use `remove_deps` in this implementation?

mitchmindtree (commented):
> Should I file an issue and address it with a separate PR or do you think we should have it with this one?

Yeah I think it would be great if we could address it in this PR!

forc-pkg/src/pkg.rs
kayagokalp: Hmm, that might be a little problematic to implement. Since the dependency is removed from the manifest file, we do not have any way of converting `Pinned` (which we can get from the `graph` that is created from the lock file) to `Pkg` so for removed packages I couldn't find a way to obtain their packages.

forc-pkg/src/pkg.rs
kayagokalp: This got me thinking and realized that this approach was wrong for handling removed packages after the lock file is generated. 

What I was doing was while visiting in the parent-to-child direction I was looking for the current node's parent's manifest file for the same reason if I cannot find the dependency,  I was deleting the current node and connecting the current node's child nodes to the current node's parent. 

**An example scenario where this seems wrong to me:**

If `a` -> (depends on) `b` -> (depends on) `c`, and If I remove `b`, `c` was getting connected to `a` and this is not necessarily correct. Imagine a package depending on `std` and we are removing `std` from the manifest file (in the case of an explicit declaration) after the lock file is generated, this approach would result in `core` to be getting connected to the project_node.

We should be visiting in the child-to-parent direction and checking if the parent of that node's manifest file does not contain a dependency declaration for the current node, if that is the case current node is a removed dependency that should be removed from the dependency graph generated from the lock file.

I think it seems not easily possible to re-use `remove_deps` since we should be visiting nodes in the child-to-parent manner, which is the opposite direction of the `removed_deps` and also, different from the `removed_deps` we need to be checking parent's manifest file. So the operation seems different enough to create the need for a new implementation but I am open to any suggestion as it would be looking better if we can re-use it somehow.

forc-pkg/src/pkg.rs
mitchmindtree: Whether or not we use `Pkg`, the problem remains the same: by only checking packages by name, we will potentially remove unrelated, valid packages.

Can we perhaps pass through the node index or pinned ID instead? If not, we need to rethink how to approach this removal.

forc-pkg/src/pkg.rs
```diff
@@ -276,28 +278,82 @@ impl BuildPlan {
     }
 
     /// Attempt to load the build plan from the `Lock`.
-    pub fn from_lock(proj_path: &Path, lock: &Lock, sway_git_tag: &str) -> Result<Self> {
-        let graph = lock.to_graph()?;
+    /// Returns the best effor BuildPlan and the packages removed
```
mitchmindtree, line 281:
```suggestion
    /// Returns the best effort BuildPlan and the packages removed  
```

forc-pkg/src/pkg.rs
```diff
@@ -779,6 +892,33 @@ pub fn graph_to_path_map(
...
+            if parent_manifest.dep(&dep.name).is_none()
+                && !parent_manifest
+                    .deps()
+                    .filter_map(|dep| dep.1.package())
+                    .any(|package| package == dep.name)
+            {
+                removed_deps.insert(dep_node);
+                continue;
+            }
+        }
```
mitchmindtree, line 921: This section repeats a lot of code that already exists in the `Path` branch within the following `match &dep_source` expression that immediately follows this block. Could we move this check to that branch, and re-use the existing retrieved `parent_manifest` etc?

forc-pkg/src/pkg.rs
mitchmindtree: > we need to be checking parent's manifest file

Why do we need to check the parent's manifest file here? It seems to me that we have already done this check when we collected this set of `deps_to_remove`?

forc-pkg/src/pkg.rs
```diff
@@ -751,20 +813,71 @@ pub fn compilation_order(graph: &Graph) -> Result<Vec<NodeIx>> {
...
+            println_red(&format!("  Removing {}", dep_name));
+            // Remove the deleted node
+            graph.remove_node(*dep);
+        }
+    }
+    Ok(())
+}
+
 /// Given graph of pinned dependencies and the directory for the root node, produce a path map
 /// containing the path to the local source for every node in the graph.
```
mitchmindtree, line 868: This function comment should explain what the returne `HashSet<NodeIx>` represents.

mitchmindtree (changes requested):

forc-pkg/src/pkg.rs
kayagokalp: The problem that comes with moving this check to there is that, If this is indeed a removed dependency I don't want to continue to check its parent's manifest file (since we cannot find it) and just continue with the next dep. That's why I am checking this at the start of each iteration, if it is removed path dependency we continue with the next dependency.

I couldn't find a clean way of doing that while using that part of the code but open to any suggestion 🙌

forc-pkg/src/pkg.rs
kayagokalp: Yep... I confused myself unnecessarily since we already collected the required set of deps_to_remove we should not check it again.

kayagokalp:
I can use a re-review at this point, especially about https://github.com/FuelLabs/sway/pull/1974#discussion_r904808968.

Also, I will apply https://github.com/FuelLabs/sway/pull/1974#discussion_r902207832 too other than that I believe It started to look a lot better. Thanks for the extensive reviews so far 🙌

forc-pkg/src/pkg.rs
mitchmindtree: > If this is indeed a removed dependency I don't want to continue to check its parent's manifest file (since we cannot find it) and just continue with the next dep

I think we should still be able to do this in the following `match` expression with `continue`, just as you've done here? I don't believe there is anything else happening between this section and the `Path` branch below, so it should be fine?

mitchmindtree (commented):
Perhaps now that we have a `load_from_manifest` constructor, we can make the `from_lock` and `from_lock_file` constructors private to avoid exposing an invalid `BuildPlan` to downstream users? We can probably also make the `apply_pkg_diff` and `generate_diff` methods private now too.

forc-pkg/src/pkg.rs
kayagokalp: Yeah yeah, somehow I had the idea I couldn't do that but it was my mistake. This should be good to go now. Thanks  a lot 👍

kayagokalp:
I think this should be good to go for a final review but I couldn't make the `from_lock_file` private since `sway-lsp` uses it.

forc-pkg/src/pkg.rs
```diff
@@ -333,26 +340,58 @@ impl BuildPlan {
...
         let lock = Lock::from_path(lock_path)?;
         Self::from_lock(proj_path, &lock, sway_git_tag)
     }
 
-    /// Ensure that the build plan is valid for the given manifest.
-    pub fn validate(&self, manifest: &Manifest, sway_git_tag: &str) -> Result<PkgDiff> {
+    /// Attempt to find the difference between the graph constructed from the lock file
+    /// and the graph that is described by the manifest file. The difference is used to
+    /// update BuildPlan.
+    pub fn generate_diff(&self, manifest: &Manifest) -> Result<PkgDiff> {
```
mitchmindtree, line 394: I think we can make this private too

sway-lsp/src/core/document.rs
```diff
@@ -162,7 +162,7 @@ impl TextDocument {
             pkg::ManifestFile::from_dir(&manifest_dir, forc::utils::SWAY_GIT_TAG).unwrap();
         let lock_path = forc_util::lock_path(manifest.dir());
         let plan = pkg::BuildPlan::from_lock_file(&lock_path, forc::utils::SWAY_GIT_TAG).unwrap();
```
mitchmindtree, line 164: Ahh I wonder if this `BuildPlan` should be constructed via `load_from_manifest` rather than `from_lock_file` - otherwise LSP won't pick up changes to the manifest until the user manually triggers a build and updates the lock file again. @JoshuaBatty does this sound right?

If we can change this over to `load_from_manifest`, we should be able to make `from_lock_file` private.

kayagokalp: Depending on the input from @JoshuaBatty maybe we can do something like 

```rust
let plan =
  pkg::BuildPlan::load_from_manifest(&manifest, false, false, forc::utils::SWAY_GIT_TAG)
      .unwrap_or_else(|_| {
          pkg::BuildPlan::from_lock_file(&lock_path, forc::utils::SWAY_GIT_TAG)
              .map(|plan| plan.0)
              .unwrap()
      });
```

mitchmindtree: I think rather than falling back to `from_lock_file`, we might be better off producing an error from LSP to the user so that they're aware of whatever the failure is. It might be a bit unexpected if LSP keeps using an old `BuildPlan` after they've updated the manifest.

JoshuaBatty: If I add a create to the `Cargo.toml` file and save it in a rust project then I am able to start using types from that crate and even jump to the definition of items in that crate without needing to build the project again. Not sure if it's rust or rust-analyzer that is downloading the added crate once the manifest file has been saved. It would be nice if we could replicate this behavior somehow.

mitchmindtree: In that case it sounds like we do want to change the use of `from_lock_file` to `load_from_manifest here - @kayagokalp are you happy to do so in this PR so we can make `from_lock_file` private too?

kayagokalp: Should be good to go but didn't add an explicit error throwing for the failure as we didn't have one for the current version too:

https://github.com/FuelLabs/sway/blob/92ec0a655bd11d19b5595fcaa940e233feee883a/sway-lsp/src/core/document.rs#L164

Should we change the current behavior and explicitly return an error? or keep the current logic there (which is simply unwrapping)

mitchmindtree:
Hey @kayagokalp, just a heads up that I have my eyes on this! I'm working on a PR with some refactoring in an attempt to simplify all of this a little further and cover a few more failure cases, aiming to post tonight or tomorrow.

mitchmindtree:
Closing as these commits were all included as a part of #2234.

PR #1979, opened 2022-06-14
Add `std::alloc`

AlicanC:
This adds a new library to `sway-lib-std` called `alloc`.

(This PR is a stripped down version of #1627 with minor improvements/rebasing.)

I needed these for my work on the [Multicall script](https://github.com/FuelLabs/fuels-ts/pull/288) and just wanted to send them upstream.

- Adds `alloc::alloc`, `alloc::realloc`
- Adds `intrinsics::size_of_val`, `intrinsics::addr_of`, `intrinsics::copy`, `intrinsics::raw_eq`


AlicanC:
@emilyaherbert @mohammadfawaz also added you as reviewers since it seems like you've been working on vec. This should be useful for allocation you do there.

sway-lib-std/src/intrinsics.sw
```diff
@@ -10,3 +10,39 @@ pub fn is_reference_type<T>() -> bool {
...
+/// Returns the size of a value in bytes.
+pub fn size_of_val<T>(val: T) -> u64 {
+    __size_of_val::<T>(val)
+}
+
+/// Returns the address of the given value.
+pub fn addr_of<T>(val: T) -> u64 {
+    // TODO: Replace with intrinsic: https://github.com/FuelLabs/sway/issues/855
+    if !__is_reference_type::<T>() {
+        // std::revert not available here
```
adlerjohn, line 23: Why can't you import revert?

sway-lib-std/src/intrinsics.sw
```diff
@@ -10,3 +10,39 @@ pub fn is_reference_type<T>() -> bool {
...
+}
+
+/// Determines whether the raw bytes at two points of memory are equal.
+pub fn raw_eq(first: u64, second: u64, len: u64) -> bool {
+    // TODO: Replace with intrinsic: https://github.com/FuelLabs/sway/issues/855
+    asm(first: first, second: second, len: len, result) {
+        meq result first second len;
+        result: bool
+    }
+}
```
adlerjohn, line 48: These aren't really intrinsics, they're just low-level memory operations.

sway-lib-std/src/intrinsics.sw
AlicanC: I get an error. My understanding is that the ordering where both `intrinsics` and `revert` are imported matters and `intrinsics` comes before `revert`.

sway-lib-std/src/intrinsics.sw
AlicanC: My reasoning was that they are in `std::intrinsics` in Rust std. If I should put these somewhere else I'm open to suggestions.

AlicanC: Oh, you already edited. I can put these into `mem`, yeah. Should I keep the Rust naming? Maybe `std::mem::cpy`/`std::mem::cmp`, or `std::mem::memcpy`/`std::mem::memcmp`?

sway-lib-std/src/intrinsics.sw
adlerjohn: You can change the order of `dep`s to put `revert` first, since it doesn't depend on intrinsics.

sway-lib-std/src/intrinsics.sw
otrho: Like pretty much all ASM blocks in the std lib I think they should become intrinsics eventually, in which case they won't need to exists as lib functions at all.  But until then `std::mem` is good.

We don't anticipate adding anything more to `intrinsics.sw`, it's there just to make `__size_of` look a bit nicer.

adlerjohn: Maybe they're intrinsics in Rust (probably), but in our case they're not so it would be weird to put them under `intrinsics`.

The suggested names aren't bad, but maybe `memeq` would be clearer since unlike C's memcmp it's a Boolean equal-or-not.

otrho: I think just `eq` is an OK name -- `mem::eq()`, rather than `raw_eq`.

AlicanC: Naming?

otrho (changes requested):
Yep, so move everything new in `intrinsics.sw` to `mem.sw` and then I think it's :+1: .

sway-lib-std/src/intrinsics.sw
AlicanC: Whoops. Just seeing the comments.

AlicanC:
> Yep, so move everything new in `intrinsics.sw` to `mem.sw` and then I think it's 👍 .

I'm leaving `size_of_val` in `intrinsics`, right?

otrho:
> I'm leaving `size_of_val` in `intrinsics`, right?

Yep, that's OK.

otrho (dismissed):

AlicanC:
Added two commits:

[Reorganize](https://github.com/FuelLabs/sway/pull/1979/commits/4a0d9c9b9f98606c54b8f83b915c12cda3c8ffaa): The changes we discussed.

[Add mem::{read,write}](https://github.com/FuelLabs/sway/pull/1979/commits/3be7b380f2a1845c52ee40cd7ce1c10d5b9ef99a): Since we have `mem` now, it made sense to add these two too. With these there won't really be a need for https://github.com/FuelLabs/sway/pull/1627.

sway-lib-std/src/alloc.sw
```diff
@@ -0,0 +1,45 @@
...
+/// See: https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/opcodes.md#aloc-allocate-memory
+pub fn alloc(size: u64) -> u64 {
+    asm(size: size, ptr) {
+        aloc size;
+        // `$hp` points to unallocated space and heap grows downward so
+        // our newly allocated space will be right after it
+        addi ptr hp i1;
+        ptr: u64
+    }
+}
```
mohammadfawaz, line 34: How about something like this:
```suggestion
pub fn alloc(size: u64) -> u64 {
    asm(size: size, ptr) {
        aloc size;
    }
    // `$hp` points to unallocated space and heap grows downward so
    // our newly allocated space will be right after it
    heap_ptr() + 1
}
```
You would need to import `use ::context::registers::heap_ptr;`

sway-lib-std/src/intrinsics.sw
```diff
@@ -10,3 +10,8 @@ pub fn is_reference_type<T>() -> bool {
 pub fn size_of<T>() -> u64 {
     __size_of::<T>()
 }
+
+/// Returns the size of a value in bytes.
```
mohammadfawaz, line 14:
```suggestion
/// Returns the size of the type of a value in bytes.
```

sway-lib-std/src/mem.sw
```diff
@@ -0,0 +1,55 @@
...
+pub fn addr_of<T>(val: T) -> u64 {
+    if !__is_reference_type::<T>() {
+        revert(0);
+    }
+    asm(ptr: val) {
+        ptr: u64
+    }
+}
+
+/// Copies bytes from src to dst.
```
mohammadfawaz, line 17:
```suggestion
/// Copies `size` bytes from `src` to `dst`.
```

sway-lib-std/src/mem.sw
```diff
@@ -0,0 +1,55 @@
...
+pub fn copy(dst: u64, src: u64, size: u64) {
+    asm(dst: dst, src: src, size: size) {
+        mcp dst src size;
+    };
+}
+
+/// Determines whether the raw bytes at two points of memory are equal.
```
mohammadfawaz, line 24:
```suggestion
/// Compares `len` raw bytes in memory at addresses `first` and `second`.
```

mohammadfawaz (approved):
Looks good! Some comments but not blocking.

sway-lib-std/src/alloc.sw
mohammadfawaz: That's sad 😞

mohammadfawaz (approved):

PR #1984, opened 2022-06-14
Add `std::tx::get_predicate_data`

### LINKED ISSUE #1981: Add predicate data helper into std lib
Add helper to collect predicate data, see also:
https://github.com/FuelLabs/fuels-rs/issues/341
https://github.com/FuelLabs/fuels-ts/issues/304

camsjams:
Closes #1981


sway-lib-std/src/tx.sw
```diff
@@ -205,6 +207,27 @@ pub fn tx_input_coin_owner(input_ptr: u32) -> Address {
...
+pub fn tx_predicate_data_start_offset() -> u64 {
+    let is = instrs_start();
+    let predicate_length_ptr = is - 16;
+    let predicate_code_length = asm(r1, r2: predicate_length_ptr) {
+        lw r1 r2 i0;
+        r1: u64
+    };
+
+    let predicate_data_ptr = is + predicate_code_length;
+    predicate_data_ptr + predicate_data_ptr % 8
```
otrho, line 223: Is this line attempting to round the pointer up to a multiple of 8?  If so, it... isn't.  🙂

`predicate_data_ptr + (8 - (predicate_data_ptr % 8))` is slightly more accurate, except it would add 8 when the ptr is already aligned.

`(predicate_data_ptr + 7) - ((predicate_data_ptr + 7) % 8)` is a better way to do it.

Or if this is not what you were trying to do, adding the value mod 8 to itself doesn't really make sense.

OTOH, if your tests are passing then maybe the predicate data pointer is always aligned anyway.  But I doubt it.

otrho (changes requested):

sway-lib-std/src/tx.sw
adlerjohn: Not rounding, but padding. `predicate_data_ptr % 8` is guaranteed to be either `0` (if `predicate_data_ptr` is word-aligned already, i.e. there are an even number of instructions), or `4` (if there are an odd number of instructions).

sway-lib-std/src/tx.sw
```diff
@@ -5,6 +5,8 @@ library tx;
 use ::address::Address;
 use ::contract_id::ContractId;
 use ::intrinsics::is_reference_type;
+use ::context::registers::instrs_start;
```
adlerjohn, line 8: Alphabetize imports.

sway-lib-std/src/tx.sw
```diff
@@ -205,6 +207,27 @@ pub fn tx_input_coin_owner(input_ptr: u32) -> Address {
...
+
+pub fn tx_predicate_data_start_offset() -> u64 {
+    let is = instrs_start();
+    let predicate_length_ptr = is - 16;
+    let predicate_code_length = asm(r1, r2: predicate_length_ptr) {
+        lw r1 r2 i0;
+        r1: u64
+    };
+
+    let predicate_data_ptr = is + predicate_code_length;
```
adlerjohn, line 222: This should be 
```suggestion
    let predicate_data_ptr = is + predicate_code_length * 4;
```

The fact that it's not means there's a bug in the SDKs where they're not dividing the lengths by 4 to get the number of instructions instead of the number of bytes, as described in the specs.

adlerjohn: Only (1). Code length is in instructions (each instruction is 4 bytes), data length is in bytes.

otrho (dismissed):

sway-lib-std/src/tx.sw
otrho: So does this still need to be resolved somewhere?  Is there an open issue to fix the code length thing?

otrho (approved):

adlerjohn (approved):

PR #1987, opened 2022-06-15
preprocessor strictness change

### LINKED ISSUE #1918: Rework strictness or logic of `mdbook-forc-documenter`
`mdbook-forc-documenter` was built with failing in the CI in mind should there be any new or removed commands/plugins that we have changed but update in the book. The introduction of `forc-fmt-v2` in the `forc-plugins` directory has caused devs to not be able to build the book locally as a result, since it was designed to fail on any errors.

Possible solutions: 

1) We could rework the logic of the preprocessor. Right now it's looking for plugins within the `forc-plugins` directory and running `forc <plugin> --help` - this is causing `forc-fmt-v2` to be picked up locally and ran if your machine has the v2 formatter installed. We could change the method of detection or perhaps introduce some way to exclude plugins or commands that are still WIP.

2) We could also make the `mdbook build` not fail if it's not ran in the CI eg. introduce some sort of flag(?)

cc: @mohammadfawaz

binggh:
Closes #1918 

mdbook preprocessors allow you to use [environment variables](https://rust-lang.github.io/mdBook/format/configuration/environment-variables.html) to do certain things within the preprocessor. Introduced a new `strict` env var here that causes `mdbook build` to behave like before (failing at any errors), or pass the build regardless if it's not enabled. 

## Motivation: 

Local builds were failing erroneously since the addition of `forc-fmt-v2`, since it wasn't included in the book as a chapter but was getting detected by the preprocessor.

This env var is **false** by default - the command that the CI runs will run `mdbook build` on strict mode which fails the build:

```sh
MDBOOK_preprocessor__FORC_documenter__STRICT="true" mdbook build docs
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,7 +139,7 @@ jobs:
           mdbook-version: "0.4.15"
       - name: Emit logs to tmp.txt, fail if build logs contain 'ERROR'
         run: |
-          mdbook build docs &> tmp.txt
+          MDBOOK_preprocessor__FORC_documenter__STRICT="true" mdbook build docs &> tmp.txt
           if cat tmp.txt | grep 'ERROR'
           then
             rm tmp.txt && exit 1
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -39,7 +39,7 @@ jobs:
         with:
           mdbook-version: '0.4.15'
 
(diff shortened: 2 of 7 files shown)


mitchmindtree (approved):
Lgtm - nice to see the user will still get a warning at least :+1:

kayagokalp (approved):
LGTM 🎉

PR #1988, opened 2022-06-15
Translate assignments to `_` as variable declaration

### LINKED ISSUE #1867: Tuple destructuring results in warnings about variables being ignored when `_` is used
# Reproduce

```rust
library tuples;

fn tuple() {
    let tuple = (5, true, ("Sway", 8));

    let(_, truthness, _) = tuple;
}
```

The variables are intentionally ignored. There should not be any complaints (warnings) in this case.
Tangentially, line 6 (in screenshot) clearly "uses" the variable.

# Screenshots

vaivaswatha:
Closes #1867.

A new no-span variable named `_` is created and the assignment to `_` is now compiled as if it was being assigned to an actual
variable. This stops the warning about the RHS expression being ignored.

### DIFF
--- a/sway-core/src/convert_parse_tree.rs
+++ b/sway-core/src/convert_parse_tree.rs
@@ -2552,14 +2552,12 @@ fn statement_let_to_ast_nodes(
         span: Span,
     ) -> Result<Vec<AstNode>, ErrorEmitted> {
         let ast_nodes = match pattern {
-            Pattern::Wildcard { .. } => {
-                let ast_node = AstNode {
-                    content: AstNodeContent::Expression(expression),
-                    span,
+            Pattern::Wildcard { .. } | Pattern::Var { .. } => {
+                let (mutable, name) = match pattern {
+                    Pattern::Var { mutable, name } => (mutable, name),
+                    Pattern::Wildcard { .. } => (None, Ident::new_no_span("_")),
+                    _ => panic!("Internal error"),
                 };
-                vec![ast_node]
-            }
-            Pattern::Var { mutable, name } => {
                 let (type_ascription, type_ascription_span) = match ty_opt {
                     Some(ty) => {
                         let type_ascription_span = ty.span();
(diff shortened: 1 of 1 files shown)


vaivaswatha:
What's the best way to add a test for this? (i.e., how do we check expected and unexpected warnings?)

sway-core/src/convert_parse_tree.rs
```diff
@@ -2552,14 +2552,12 @@ fn statement_let_to_ast_nodes(
         span: Span,
     ) -> Result<Vec<AstNode>, ErrorEmitted> {
         let ast_nodes = match pattern {
-            Pattern::Wildcard { .. } => {
-                let ast_node = AstNode {
-                    content: AstNodeContent::Expression(expression),
-                    span,
+            Pattern::Wildcard { .. } | Pattern::Var { .. } => {
+                let (mutable, name) = match pattern {
```
otrho, line 2556: I feel like there must be a better way to do this than to match on the same pattern twice, but nothing obvious springs to mind in this case.  Perhaps we can rethink how we manage wildcard patterns, and instead of having a separate variant for `Wildcard` just add an `is_wildcard` field to `Var`.

Dunno.  This is OK for now. :slightly_smiling_face:

otrho (approved):

mohammadfawaz:
> What's the best way to add a test for this? (i.e., how do we check expected and unexpected warnings?)

Not yet possible. Toby wants to introduce FileCheck to the integration tests so that will help!

sway-core/src/convert_parse_tree.rs
```diff
@@ -2552,14 +2552,12 @@ fn statement_let_to_ast_nodes(
...
+                    Pattern::Var { mutable, name } => (mutable, name),
+                    Pattern::Wildcard { .. } => (None, Ident::new_no_span("_")),
+                    _ => panic!("Internal error"),
```
mohammadfawaz, line 2559: `unreachable!()` here is maybe better? Not blocking though.

mohammadfawaz (dismissed):

mohammadfawaz (approved):

Issue #1985, opened 2022-06-15
Predicate hash should be predicate bytecode root


simonr0204:
Issue #1913 added script and predicate bytecode hash to the output on compilation. For predicates, this should actually be the [root ](https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/identifiers.md#contract-id) of the bytecode.   (i.e. `Contract::root_from_code(&bytecode)`)
 
 This is my bad - I thought it was the hash we needed as we've been referring to 'predicate hashes', but it turns out this is a legacy phrase from bitcoin script nomenclature (P2SH) . In Fuel, it's the root of the bytecode that we need.
 
 Note : for scripts, the correct value is the hash (so no need to change for scripts)

PR #1992, opened 2022-06-15
Don't emit hashes for predicates

mohammadfawaz:
Before fixing https://github.com/FuelLabs/sway/issues/1985 properly, I'm stopping `forc build` from emitting the bytecode hash for predicates.


kayagokalp (approved):
Tested locally, it works & LGTM

PR #1997, opened 2022-06-15
sway-fmt-v2 adds program type to the output

### LINKED ISSUE #1995: sway-fmt-v2 should include program type to the output
Currently, `sway-fmt-v2` consumes the program type part of the source code it is formatting. It should handle the program type otherwise it is going to be breaking the build process.

For details: see https://github.com/FuelLabs/sway/pull/1962#issue-1269910143

kayagokalp:
closes #1995.

simply tested with a main.sw file that only contains:

```
contract;
```


sway-fmt-v2/src/fmt.rs
```diff
@@ -1,4 +1,4 @@
-use crate::utils::indent_style::Shape;
+use crate::utils::{indent_style::Shape, program_type::insert_program_type};
```
eureka-cpu, line 1: Would you mind merging the two `crate`s here?

sway-fmt-v2/src/fmt.rs
```diff
@@ -37,8 +42,20 @@ impl Formatter {
...
-        let formatted_raw_newline = items
+        let module = sway_parse::parse_file(src, path)?;
+        // Get parsed items
+        let items = module.items;
+        // Get the program type (script, predicate, contract or library)
+        let program_type = module.kind;
+
+        // Formatted code will be pushed here with raw newline stlye.
+        // Which means newlines are not converted into system-specific versions by apply_newline_style
+        let mut formatted_raw_newline = String::new();
```
eureka-cpu, line 53: would you mind changing this variable name to `raw_formatted_code`?

eureka-cpu: Also looks like merging in master added in a duplicate import

mitchmindtree (commented):
> simply tested with a main.sw file that only contains: `contract;`

Perhaps this could act nicely as a small unit test?

kayagokalp:
I added a small test to test this but there is this strange thing happening so I am not sure which way to continue:

I realized that the program type got deleted once there is a match to an item so just supplying 

```
contract;
```
for the test actually does not test this completely. I cannot add anything that would match an item in the source code since we do not have any of the logic for them implemented. (Before #1962 is merged.) So I think we should merge that first and then supply a test like

```
contract;

enum Color {
    Blue: (),
    Green: (),
    Red: (),
    Silver: (),
    Grey: (),
}
```

eureka-cpu:
@mitchmindtree one problem we are trying find a solution for is that the initial idea of the formatter was to leave anything it didn't recognize as an `item` as is, but this PR is actually a side effect of it _removing_ everything that isn't an `item`.

kayagokalp:
Should we close this one and open an issue to track not-touching non-item source code, or should we continue with this one until we have a fix for that

eureka-cpu:
> Should we close this one and open an issue to track not-touching non-item source code, or should we continue with this one until we have a fix for that?

Let's leave this open and keep it as a short term solution, since we do want to format the program type anyway and also open an issue to track the aforementioned problem (:

kayagokalp:
Opened #2008 for tracking that 👍

sway-fmt-v2/src/fmt.rs
```diff
@@ -93,8 +106,9 @@ enum Color {
                     Grey: (), }
         "#;
 
-        // Until #1995 is addressed we will not have contract; in the output
-        let correct_sway_code = r#"enum Color {
+        let correct_sway_code = r#"contract;
```
kayagokalp, line 109: I removed the separate test which was not testing things correctly (even though I remove the changes from this PR, it wouldn't fail as the failure depends on the presence of other items in the source code that we are trying to format). Although we do not have any separate test for program type, it is checked in each test.

mitchmindtree (approved):
Lgtm! Will leave the final review to @eureka-cpu.

eureka-cpu (approved):

PR #2000, opened 2022-06-15
Updates `apply_newline_style` function to use determined setting

eureka-cpu:
Currently we have this set to just use `Auto` but we are determining which setting to use from the `Config` before the `format` function is called so we can change this to be dynamic.


kayagokalp (approved):
Ohh, that's my bad sorry. Thanks for the fix and lgtm 👍

adlerjohn (approved):

PR #2003, opened 2022-06-16
Config driven E2E testing.

otrho:
OK, I've added hundreds of `test.toml` files to all the E2E tests and now they're gathered up and run from `mod.rs`.  Hopefully this means no more annoying merge conflicts at the end of the test list.

This is PHASE ONE.  There's no new actual test functionality here, just a different discovery model.

 PHASE TWO will include the ability to specify expected output from the compilation process.  This will be especially useful for the `should_fail` tests.

Another feature we could add is for the harness to complain about test projects which don't have a `test.toml`.  At the moment there are a few of these in there, which I've explicitly disabled.  Some are waiting for new features, but some seem to just have been forgotten.  And I think I even rediscovered some tests in here which weren't being run by the old `mod.rs`, probably the victim of some merge faults.


emilyaherbert:
@otrho , would you be able to include a small fix addressing this comment in your PR? https://github.com/FuelLabs/sway/pull/2009#discussion_r899400618

otrho:
I've also noticed that for the `should_pass` tests which verify their ABI (most of them) it's quite inefficient -- it's compiling everything twice, which as compiling `libstd` gets slower is compounding the problem.  I reckon if we could do both in a single pass we'd cut 30% off the total run time.

test/src/e2e_vm_tests/test_programs/README.md
```diff
@@ -0,0 +1,82 @@
...
+## `test.toml`
+
+To add a new test to the E2E suite place a `test.toml` file at the root of the test Forc package,
+i.e., next to the `Forc.toml` file.  This file may contain a few basic fields.
+
+## `category`
+
+The `category` field is mandatory and must be one of the following strings:
+
+* `"run"` - The test is compiled an run in a VM.
```
nfurfaro, line 15: typo: "The test is compiled _**and**_ run in a VM."

test/src/e2e_vm_tests/test_programs/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+The `action` field describe what sort of result to expect:
+
+* `"return"` - An integer value returned by success in the VM.
+* `"return_data"` - An array of bytes returned by the VM.
+* `"result"` - An integer word returned by the Fuel Core node.
+* `"revert"` - An integer value returned by failure in the VM.
+
+The `value` field is the actual expected value.  For `"return"`, `"result"` and `"revert"` actions
+is must be an integer.
```
nfurfaro, line 34: typo: "**_it_** must be an integer"

test/src/e2e_vm_tests/test_programs/README.md
```diff
@@ -0,0 +1,82 @@
...
+deployed to the node prior to deploying and running the test code.  These are specified with the
+`contracts` field.
+
+It must be an array of strings each containing only the path to the Forc project for the contract to
+be compiled and deployed.  It is important that these paths remain relative to the
+`test/src/e2e_vm_tests/test_programs` directory.
+
+## `validate_abi`
+
+Some tests also require their ABI is verified.  To indicate this the `validate_abi` field may be
```
Braqzen, line 50:
```suggestion
Some tests also require their ABI to be verified.  To indicate this the `validate_abi` field may be
```

test/src/e2e_vm_tests/test_programs/should_pass/language/enum_padding/test.toml
```diff
@@ -0,0 +1,3 @@
+category = "run"
+expected_result = { action = "return_data", value = [ 0xce, 0x55, 0xff, 0x05, 0x11, 0x3a, 0x24, 0x2e, 0xc7, 0x9a, 0x23, 0x75, 0x0c, 0x7e, 0x2b, 0xab, 0xaf, 0x98, 0xa8, 0xdc, 0x41, 0x66, 0x90, 0xc8, 0x57, 0xdd, 0x31, 0x72, 0x0c, 0x74, 0x82, 0xb6 ] }
```
Braqzen, line 2: The value here is in different encoding than in the README.md. Should the README.md be updated to reflect this?

test/src/e2e_vm_tests/mod.rs
```diff
@@ -1,730 +1,349 @@
...
-    // Deploy and then test.
-    number_of_tests_run += projects.len();
-    let mut contract_ids = Vec::<fuel_tx::ContractId>::with_capacity(contracts.len());
-    for name in contracts {
-        let contract_id = harness::deploy_contract(name, locked);
-        contract_ids.push(contract_id);
-    }
+            TestCategory::Compiles => {
+                assert!(crate::e2e_vm_tests::harness::compile_to_bytes(&name, locked).is_ok());
+                number_of_tests_run += 1;
```
Braqzen, line 90: Should we be counting tests that don't run as ones that do run? Compilation, skipping etc. should not be counted as being run, no? Am I misunderstanding?

otrho: We currently count all tests that are 'tested'.  So maybe that variable should just be `number_of_tests_executed`.  Not that important, could change it one day.

mohammadfawaz (approved):

test/src/e2e_vm_tests/mod.rs
otrho: I've already had to do a rebase to take in @vaivaswatha's PR changes, so I renamed this var while I was at it.

PR #2004, opened 2022-06-16
Introduce a type check `Context`. Replaces `TypeCheckArguments`.

### LINKED ISSUE #1810: Provide a "context" to the type-check phase with the aim of improving ergonomics
Currently, the set of `TypeCheckArguments` we pass throughout type-checking is rather ad-hoc and requires frequent destructuring and re-structuring as it is passed throughout the AST.

With the incoming `FunctionEngine`  and `DeclarationEngine` (#1692) types to be introduced with the new collections phase (#1693), and plans to de-globalize the type engine, it's time to consider refactoring the `TypeCheckArguments` into something a little more ergonomic.

_TODO: Update this with a pseudo context type (or open a draft)._

mitchmindtree:
Closes #1810.

~This is very WIP, but just posting as a draft so that folks can see what direction it's going in in the meantime! Will elaborate here on the changes when this is closer to ready-for-review.~

OK, this should be good to go!

The main change is the removal of the `TypeCheckArguments` in favour of a new type check `Context` type throughout `semantic_analysis`.

The `Context` carries through many of the original fields from `TypeCheckArguments`, but attempts to expose a more ergonomic API for the type-checking process.

A key difference is that the generic `checkee` field has been removed in favour of using dedicated function arguments. This generic `checkee` resulted in the need to constantly destructure and restructure the `TypeCheckArguments` throughout semantic analysis due to each node having a different checkee type, and in turn a different filled `TypeCheckArguments` type. This destructuring added quite a bit of noise and often made it tricky to keep track of which parts of the type-check context were actually getting updated, and which were simply being plumbed through from the parent node.

This should be much clearer now with the new `Context`, which provides builder types to only update the fields that are necessary. There are some locations where I'm unsure if it's truly necessary to reset the `help_text` to `""`, or the `type_annotation` to `insert_type(TypeInfo::Unknown)`, however I added them anyway to ensure we match the original semantics exactly. If it turns out some of these are unnecessary, perhaps we can remove them in future PRs.


mitchmindtree:
Woops, just requested reviews individually forgetting there's a `FuelLabs/sway-compiler` team now. Will remember to use the team in the future!

sway-core/src/ir_generation/function.rs
```diff
@@ -10,7 +10,7 @@ use crate::{
 use super::{compile::compile_function, convert::*, lexical_map::LexicalMap, types::*};
 
 use fuel_crypto::Hasher;
-use sway_ir::*;
+use sway_ir::{Context, *};
```
otrho, line 13: What is the thinking here?  Trying to emphasise `Context` somehow?  Does it clash with the new `Context`?

otrho (approved):
More red than green is always a good thing. :)

sway-core/src/ir_generation/function.rs
mitchmindtree: Yeah there's a glob import from both `sway_ir` and `semantic_analysis` here :smiling_face_with_tear: the explicit import takes precedence in Rust so this resolves the conflict.

otrho: I didn't know you could do that.  Cool. :slightly_smiling_face:

JoshuaBatty (dismissed):
This is great, the code is much cleaner!

adlerjohn (commented):
Merge conflicts

mitchmindtree:
Okydoke, rebased!

Would love a review from @emilyaherbert or @sezna in particular as you both spend the most time in `semantic_analysis.rs` :mage_woman:  :mage:

sway-core/src/semantic_analysis/context.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    // The following set of fields are intentionally private. When a `Context` is passed into a new
+    // node during type checking, these fields should be updated using the `with_*` methods which
+    // provides a new `Context`, ensuring we don't leak our changes into the parent nodes.
+    self_type: TypeId,
+    /// While type-checking an expression, this indicates the expected type.
+    ///
+    /// Assists type inference.
+    type_annotation: TypeId,
+    mode: Mode,
```
emilyaherbert, line 25: Add description here.

sway-core/src/semantic_analysis/context.rs
```diff
@@ -0,0 +1,186 @@
...
+        Namespace,
+    },
+    type_engine::{insert_type, unify_with_self, TypeId, TypeInfo},
+    CompileResult, CompileWarning, TypeArgument, TypeError,
+};
+use sway_types::{span::Span, Spanned};
+
+/// Contextual state tracked and accumulated throughout type-checking.
+pub struct Context<'ns> {
+    pub(crate) namespace: &'ns mut Namespace,
```
emilyaherbert, line 15: Add description here.

sway-core/src/semantic_analysis/context.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    /// Create a new context that mutably borrows the inner `namespace` with a lifetime bound by
+    /// `self`.
+    ///
+    /// This is particularly useful when type-checking a node that has more than one child node
+    /// (very often the case). By taking the context with the namespace lifetime bound to `self`
+    /// rather than the original namespace reference, we instead restrict the returned context to
+    /// the local scope and avoid consuming the original context when providing context to the
+    /// first visited child node.
+    pub fn by_ref(&mut self) -> Context {
```
emilyaherbert, line 62: Would you consider renaming this to `by_ref_mut`? To stay consistent with the pseudo-equivalent `ref mut` Rust pattern.

sway-core/src/semantic_analysis/context.rs
```diff
@@ -0,0 +1,186 @@
...
+    type_annotation: TypeId,
+    mode: Mode,
+    // TODO: Should this be removed to `CompileError` and `CompileWarning`?
+    help_text: &'static str,
+    /// Tracks the purity of the context, e.g. whether or not we should be allowed to write to
+    /// storage.
+    purity: Purity,
+}
+
+impl<'ns> Context<'ns> {
```
emilyaherbert, line 33: Would you be able to rename this to `TypeSystemContext` or something like that? With the new collection context and stuff going in we will need to have differentiating names.

emilyaherbert (changes requested):
This looks great! Just a few nits.

I'm thinking that we should remove the globally-accessible type engine and instead place it inside of this new context. However I think that is too much for this PR. This PR should go in and then it should be relatively straightforward to place the type engine inside of this context later.

sway-core/src/semantic_analysis/context.rs
mitchmindtree: In this case, I was aiming to follow the precedent set by [`Iterator::by_ref`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.by_ref) which does something kiiiind of similar. There can be no immutable case here, so I'm hoping there should be no ambiguity anyways.

emilyaherbert: Ah gotcha. Feel free to ignore this then.

sway-core/src/semantic_analysis/context.rs
mitchmindtree: I was hoping we could keep the brevity of `Context` seeing as it's used in so many locations, but if there will also be a `CollectionContext` that sounds fair! Perhaps `TypeCheckContext`?

mitchmindtree:
Thanks for the review @emilyaherbert! Will address these today.

> I'm thinking that we should remove the globally-accessible type engine and instead place it inside of this new context. However I think that is too much for this PR. This PR should go in and then it should be relatively straightforward to place the type engine inside of this context later.

Yeah agreed, I'd be happy to have a go at this but need to prioritise formatter V2 stuff this week. I'll hold off on assigning myself this in case someone else gets a chance to have a go in the meantime.

emilyaherbert:
> Yeah agreed, I'd be happy to have a go at this but need to prioritise formatter V2 stuff this week. I'll hold off on assigning myself this in case someone else gets a chance to have a go in the meantime.

Will you make an issue for this? (if there is not one already .. 😄 )

mitchmindtree:
@emilyaherbert OK, I think I've addressed everything!

I've also moved the construction of the `TypeCheckContext` from the module-level to the beginning of `TypedProgram::type_check`. This should hopefully make it a little clearer on how to provide the declaration, type and function engines to the `TypeCheckContext` and pass them through to each of the submodules.

otrho (dismissed):

emilyaherbert:
> Would we eventually want to implement these functions as methods on context, or do you think it is better architecture this way? I personally think it could go either way, not sure if there's a tiebreaker

What functions are you talking about here @sezna ? I think maybe the context got lost on GitHub.

Maybe this is what you meant, but I feel that we should move everything that has to do with type engine stuff and type resolving/lookup/monomorphization to inside of the type engine `Engine` type (which would live inside of this new context type).

emilyaherbert (dismissed):

mitchmindtree:
Just merged master and resolved some conflicts with the PR that changes resolution to use `TypeId`s rather than `TypeInfo`s.

JoshuaBatty (approved):

mohammadfawaz (approved):

PR #2007, opened 2022-06-16
remove `fuels-abigen-macro` dependency

Salka1988:
Forc init template include no longer the abigen macro dependency directly.


adlerjohn:
This is blocked by https://github.com/FuelLabs/fuels-rs/pull/393, no?

mohammadfawaz (approved):
Seems reasonable to me. Thanks.

adlerjohn (changes requested):

forc/src/cli/commands/test.rs
```diff
@@ -70,7 +70,6 @@ pub(crate) fn exec(command: Command) -> Result<()> {
 
     // Pass through cargo test args.
     if !command.cargo_test_args.is_empty() {
-        cmd.arg("--");
```
mitchmindtree, line 73: Could you elaborate on why this was removed?

The `--` separator is necessary in order to distinguish between the cargo test options and cargo test arguments. Perhaps you removed it when you were trying to pass some cargo test options to `--cargo-test-args`, when you perhaps instead wanted to use `--cargo-test-opts`?

mitchmindtree (changes requested):

test/src/sdk-harness/test_projects/tx_fields/mod.rs
```diff
@@ -158,7 +159,9 @@ async fn can_get_witnesses_count() {
 #[tokio::test]
 async fn can_get_receipts_root() {
     let (contract_instance, _, _) = get_contracts().await;
-    let zero_receipts_root = Bytes32::default();
+    let zero_receipts_root =
+        Bytes32::from_str("4be973feb50f1dabb9b2e451229135add52f9c0973c11e556fe5bce4a19df470")
```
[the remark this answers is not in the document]
Salka1988: Previously, this function returned only zeros, but this has been fixed. Now this function always returns this value that I set as the harcoded value. I saw earlier that certain tests were written like this, so I wrote like this.

forc/src/utils/defaults.rs
```diff
@@ -30,10 +30,12 @@ edition = "2021"
 license = "Apache-2.0"
 
 [dependencies]
-fuels = "0.15"
-fuels-abigen-macro = "0.15"
+fuels = {{ version = "0.16", features = ["fuel-core-lib"] }}
 tokio = {{ version = "1.12", features = ["rt", "macros"] }}
 
+[features]
+fuel-core-lib = ["fuels/fuel-core-lib"]
```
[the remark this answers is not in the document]
Salka1988: I don't think the user has to write "forc test - fuel-core-lib features" but "forc test" is enough. In this way, I wanted to enable the "fuel-core-lib" flag that the user would use as needed. Correct me if I'm wrong.
Should I implement `` forc init --feature fuel-core-lib`` and in that way enable the use of `` fuel-core-lib`` flag, and by default generate Cargo.toml without `` fuel-core- lib`` flag?

forc/src/cli/commands/test.rs
Salka1988: Jup! That is right. When I wanted to use `` --features``, I ran into problems until I deleted `` --``.

ControlCplusControlV (commented):
Testing locally branch is working fine for me, will defer to sway people for code style changes as they will be primary maintainers

test/src/sdk-harness/Cargo.toml
```diff
@@ -24,3 +23,6 @@ name = "integration_tests"
 path = "test_projects/harness.rs"
 
 [workspace]
+
+[features]
+fuel-core-lib = ["fuels/fuel-core-lib"]
```
mitchmindtree, line 28: Could you elaborate a little on why the `sdk-harness` crate needs to expose a `fuel-core-lib` feature here? Currently it appears to be unused by the `sdk-harness` crate.

If we need the `fuels` dependency's `fuel-core-lib` feature in order to be able to successfully build the tests, I think it's fine to just specify this feature directly under the `fuels` dependency (as you've already done). We may not to expose the feature from the crate itself.

Salka1988: I think we need to expose it if we want to use "fuel-core-lib" in tests.

.github/workflows/ci.yml
```diff
@@ -179,7 +179,7 @@ jobs:
       - name: Build test project
         run: forc build --path test-proj
       - name: Run test project's test suite
-        run: cd test-proj && forc test
+        run: cd test-proj && forc test --cargo-test-opts="--features fuel-core-lib"
```
adlerjohn, line 182: Please revert changes to this file, as they're orthogonal to the what this PR targets: "remove fuels-abigen-macro dependency."

Transitively revert any changes that require this file to be changed.

adlerjohn (changes requested):

.github/workflows/ci.yml
adlerjohn: In the future, avoid bundling multiple independent issues into a single PR, as the combinatorial explosion in interactions between features and fixes can make it impossible to review quickly.

Salka1988: done

forc/src/utils/defaults.rs
adlerjohn, line 37: Is the reason we don't have to specify `fuel-core-lib` as a Rust feature to `cargo test` a bug, or a feature? https://github.com/FuelLabs/fuels-rs/pull/419 is supposed to enable fuel-core bin as default instead of lib IIUC. If it's a bug, then these two lines should be removed.

test/src/sdk-harness/Cargo.toml
mitchmindtree: > I think we need to expose it if we want to use "fuel-core-lib" in tests.

I believe we only need to enable the `fuels` dependency's `fuel-core-lib` (which you've already done). I don't believe there's any need to expose the feature from this crate as we require the feature from `fuels` whether or not the user passes a `--features "fuel-core-lib"` flag.

Could you try removing this `[features]` table?

adlerjohn: Ditto

Salka1988: You're right! I changed this too.

adlerjohn (approved):

adlerjohn:
poking @mitchmindtree

mitchmindtree (approved):

PR #2011, opened 2022-06-16
Add `attribute` to `utils`

eureka-cpu:
This adds the `format_attributes` function to `utils::attributes`.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -37,16 +37,16 @@ jobs:
       - name: Setup mdBook
         uses: peaceiris/actions-mdbook@v1
         with:
-          mdbook-version: '0.4.15'
+          mdbook-version: "0.4.15"
 
       - run: MDBOOK_preprocessor__FORC_documenter__STRICT="true" mdbook build docs
 
-      - name: Deploy latest
+      - name: Deploy master
         uses: peaceiris/actions-gh-pages@v3
         with:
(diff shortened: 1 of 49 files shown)


kayagokalp (commented):
looks awesome, but I have a little question, should we add a `\n` after putting the attribute? I tested by putting something static like to `item_enum.format()`: 

```rust
       r#"enum Color {
        Red:(),
       }"#.to_string() 
 }
```

Tested with the following input file:
```
contract;

#[storage(read)]
enum Color {
    Blue: (),
    Green: (),
    Red: (),
    Silver: (),
    Grey: (),
}

#[storage(read)]
enum Color {
    Blue: (),
    Green: (),
    Red: (),
    Silver: (),
    Grey: (),
}

```

Output becomes:
```
#[storage(read)]enum Color {
        Red:(),
       }
#[storage(read)]enum Color {
        Red:(),
       }
```

P.S: ~I think attributes are not applicable to enums but~ I just wanted to quickly test the interaction with an item formating.

I guess we either need to push a `\n` after the attribute or before each of item formattings.

mitchmindtree (approved):
> this just adds the outline, not the complete functionality. Going to complete that in a separate PR.

Happy to see this land and the proper `Format` implementation in a follow-up PR :+1: Will just add a small reminder to add a test or two (like Alex's snippet) when the time comes :)

JoshuaBatty (approved):

PR #2022, opened 2022-06-16
Add the `ResolveTypes` trait and remove the use of `TypeMapping` from inside of monomorphization

emilyaherbert:


sezna:
closing due to inactivity

Issue #606, opened 2022-01-09
Unnecessary warning when implicitly casting small-enough literal integers to narrower types


Dentosal:
This shouldn't produce a warning, as the compiler should know that the literal is small enough.

### Input

```rust
script;
fn main() {
  let a: u32 = 1;
  let b: u32 = a + 1;
}
```

### Expected output

No warnings about casts

### Actual output

```
warning
 --> fuel_project/src/main.sw:0:16
  |
...
2 |   let a: u32 = 1;
  |                - This cast, from integer type of width sixty four to integer type of width thirty two, will lose precision.
  |
warning
 --> fuel_project/src/main.sw:0:20
  |
...
3 |   let b: u32 = a + 1;
  |                    - This cast, from integer type of width sixty four to integer type of width thirty two, will lose precision.
  |
  Compiled script "Fuel example project" with 2 warnings.
```

mohammadfawaz:
This should be fixed with this: https://github.com/FuelLabs/sway/pull/711

adlerjohn:
This isn't fixed with arrays. See #2024

anton-trunov:
Fixed with #4829 I believe

PR #2024, opened 2022-06-17
Minor fix to test warning.

adlerjohn:
Tests throw a warning of loss of precision. Use same type to avoid warning.


mohammadfawaz (approved):

otrho:
Are these the loss of precision warnings?  If so they shouldn't really be a thing and this change shouldn't be necessary.   Yeah, found #606.

otrho (approved):

PR #2026, opened 2022-06-17
Add `forc check` command

JoshuaBatty:
This PR adds a `forc check` command that compiles the project to a typed AST but doesn't produce a binary. Similar to how ` cargo check` works. 

The language server has been updated to call into this instead of `compile_to_ast` directly. This now provides the infrastructure for sway-lsp to parse whole projects instead of individual files, as was previously the case.


adlerjohn (approved):
utACK

mohammadfawaz (commented):
LGTM! Will defer the second tick to someone on the tooling team.

forc-pkg/src/pkg.rs
```diff
@@ -1444,6 +1459,54 @@ pub fn build(
...
+        }
+        source_map.insert_dependency(path.clone());
+
+        // We only need to return the final CompileAstResult
+        let ast_res = compile_ast(&manifest, config, dep_namespace)?;
+        if i == build_plan.compilation_order.len() - 1 {
+            return Ok(ast_res);
+        }
+    }
+    Err(anyhow!("unable to check sway program from build plan"))
```
mitchmindtree, line 1507:
```suggestion
    bail!("unable to check sway program from build plan")
```

mitchmindtree (changes requested):
For the most part what's here is looking great!

That said, there are still a couple of important steps missing:

- We need to check the validity of the lock file against the current state of the manifest.
- We need to update the dependency graph in the case that `[dependencies]` has changed, and potentially fetch dependencies in the case that new dependencies have been added. 

Without the above, running `forc check` currently assumes that the lock file is *always* valid which is certainly not always the case.

All of this is already solved in `forc::ops::forc_build`, which you could use that as a reference. However, rather than just copying chunks out of the build op, we should probably try to abstract out the logic that can be shared between `check` and `build` into the `forc-pkg` crate.

JoshuaBatty:
Thanks @mitchmindtree , I've now abstracted logic shared by build and check into a `load_from_manifest` constructor on `BuildPlan`. Think this should be ready now.

forc-pkg/src/pkg.rs
```diff
@@ -235,6 +235,61 @@ impl BuildPlan {
         })
     }
 
+    /// Create a new build plan and syncronize manifest and lock file
```
mitchmindtree, line 238: Might be useful to elaborate on this a little:

> Create a new build plan taking into account the state of both the `Manifest` and the existing lock file if there is one.
> 
> This will first attempt to load a build plan from the lock file and validate the resulting graph using the current state of the `Manifest`. This includes checking if the `[dependencies]` or `[patch]` tables have changed and checking the validity of the local `path` dependencies. If any changes are detected, the graph is updated and any new packages that require fetching are fetched.
> 
> The resulting build plan should always be in a valid state that is ready for `build`ing or `check`ing.

forc-pkg/src/pkg.rs
```diff
@@ -1444,6 +1514,43 @@ pub fn build(
...
+    let mut namespace_map = Default::default();
+    let mut source_map = SourceMap::new();
+    for (i, &node) in plan.compilation_order.iter().enumerate() {
+        let dep_namespace =
+            dependency_namespace(&namespace_map, &plan.graph, &plan.compilation_order, node);
+        let pkg = &plan.graph[node];
+        let path = &plan.path_map[&pkg.id()];
+        let manifest = ManifestFile::from_dir(path, sway_git_tag)?;
+        let (_, maybe_namespace) =
+            compile(pkg, &manifest, conf, dep_namespace.clone(), &mut source_map)?;
```
mitchmindtree, line 1539: Should this be `compile_to_ast` rather than `compile`? `compile` does the full compilation of the package (including asm gen) - I think the point of `check` should be to avoid this.

sway-lsp/Cargo.toml
```diff
@@ -10,6 +10,8 @@ description = "LSP server for Sway."
 
 [dependencies]
 dashmap = "5.3.4"
+forc = { version = "0.16.1", path = "../forc" }
```
mitchmindtree, line 13: Looks like you might not need this anymore now that `check` is in `forc-pkg`?

JoshuaBatty: I still need to access `forc::utils::SWAY_GIT_TAG` inside of `sway-lsp` so I can call into the `check` function in `forc-pkg`

forc-pkg/src/pkg.rs
JoshuaBatty: Good catch, thanks this description is very clear. I'll add it in.

mitchmindtree:
cc @kayagokalp so you're aware of these changes :) I think the new `BuildPlan::load_from_manifest` here might help to clean up #1974 a little.

forc-pkg/src/pkg.rs
JoshuaBatty: omg how did I miss that, I was doing this to get the optional namespace back. I've added that functionality to `compile_ast` now which now fixes this.

sway-lsp/src/core/document.rs
```diff
@@ -94,7 +94,8 @@ impl TextDocument {
         self.clear_tokens();
         self.clear_hash_maps();
 
-        //self.test_typed_parse();
+        self.test_typed_parse();
+        eprintln!("{:#?}", self.token_map);
```
mitchmindtree, line 98: Local debug stuff?

forc-pkg/src/pkg.rs
```diff
@@ -1373,18 +1380,31 @@ pub fn dependency_namespace(
...
+        CompileAstResult::Success { typed_program, .. } => match typed_program.kind.tree_type() {
+            TreeType::Library { .. } => {
+                let lib_namespace = typed_program.root.namespace.clone();
+                Some(lib_namespace.into())
+            }
+            TreeType::Contract | TreeType::Predicate | TreeType::Script => None,
+        },
+    };
+
+    Ok((ast_res, namespace))
```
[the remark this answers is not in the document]
JoshuaBatty: Sure that makes sense.

forc-pkg/src/pkg.rs
```diff
@@ -1444,6 +1518,43 @@ pub fn build(
...
+            }
+        }
+        source_map.insert_dependency(path.clone());
+
+        // We only need to return the final CompileAstResult
+        if i == plan.compilation_order.len() - 1 {
+            return Ok(ast_res);
+        }
+    }
+    bail!("unable to check sway program from build plan")
```
mitchmindtree, line 1555:
```suggestion
    bail!("unable to check sway program: build plan contains no packages")
```

mitchmindtree (commented):
Last nit otherwise looks good :+1:

mitchmindtree (approved):
:shipit:

eureka-cpu (approved):

PR #2028, opened 2022-06-17
Added Eq for identity

### LINKED ISSUE #2027: Implement the `Eq` trait for `Identity`
# Summary

Being able to compare identities is important, I find myself doing that a bit.

E.g.

```rust
let sender: <Identity, AuthError> = msg_sender();
let sender = sender.unwrap();

if sender == owner {
    // Owl
}
```

Braqzen:
# Summary

Changes made

- Added Eq trait for Identity allowing
    - `Identity1 == Identity2`
    - `Identity1 != Identity2`
- Run the formatter on `std` which "fixed" the formatting via the broken formatter
    - It actually passed CI because the formatter is not being applied there

# Related

Closes #2027 
Bug #2031

### DIFF
--- a/sway-lib-std/src/identity.sw
+++ b/sway-lib-std/src/identity.sw
@@ -12,9 +12,9 @@ pub enum Identity {
 impl core::ops::Eq for Identity {
     fn eq(self, other: Self) -> bool {
         match (self, other) {
-            (Identity::Address(address1), Identity::Address(address2)) => address1.value == address2.value,
-            (Identity::ContractId(asset1), Identity::ContractId(asset2)) => asset1.value == asset2.value,
-            _ => false,
+            (Identity::Address(address1), Identity::Address(address2)) => { address1.value == address2.value },
+            (Identity::ContractId(asset1), Identity::ContractId(asset2)) => { asset1.value == asset2.value },
+            _ => { false },
         }
     }
 }
--- a/sway-lib-std/src/identity.sw
+++ b/sway-lib-std/src/identity.sw
@@ -12,8 +12,8 @@ pub enum Identity {
 impl core::ops::Eq for Identity {
     fn eq(self, other: Self) -> bool {
         match (self, other) {
-            (Identity::Address(address1), Identity::Address(address2)) => { address1.value == address2.value },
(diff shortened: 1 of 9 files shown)


sway-lib-std/src/option.sw
```diff
@@ -30,16 +30,14 @@ impl<T> Option<T> {
     /// Returns `true` if the result is [`Some`].
     fn is_some(self) -> bool {
         match self {
-            Option::Some(_) => true,
-            _ => false, 
+            Option::Some(_) => true, _ => false,
```
adlerjohn, line 33: There are some formatting changes here that aren't directly related to this PR, can you remove them? Do them all in one go when resolving #2031

sway-lib-std/src/identity.sw
```diff
@@ -3,8 +3,18 @@ library identity;
...
 
 pub enum Identity {
     Address: Address,
     ContractId: ContractId,
 }
+
+impl core::ops::Eq for Identity {
+    fn eq(self, other: Self) -> bool {
+        match(self, other) {
+            (Identity::Address(address1), Identity::Address(address2)) => eq(addr_of(address1), addr_of(address2), size_of_val(self)), (Identity::ContractId(asset1), Identity::ContractId(asset2)) => eq(addr_of(asset1), addr_of(asset2), size_of_val(self)), _ => false,
```
adlerjohn, line 17: To avoid poor formatting until fmt-v2 lands, can you enclose the body of each arm in `{  }`? E.g. `{ eq(addr_of(address1), addr_of(address2), size_of_val(self)) }`. That will also least format things semi-nicely.

adlerjohn, line 17: Instead of using `mem::eq` here, why not just use `address1 == address2`? Address implements `Eq`: https://github.com/FuelLabs/sway/blob/7fb883c029fa008887b413dc95c2763a4cc73887/sway-lib-std/src/address.sw#L13

You might have to import `::address::*`.

adlerjohn (approved):

mohammadfawaz (approved):
Nice. Very reasonable.

PR #2029, opened 2022-06-17
Fix couple of bugs in handling returns in if blocks

### LINKED ISSUE #1663: Compiler doesn't fully handle `return` statements in control flow
Found this issue while compiling a predicate:
This compiles:
```
if let Result::Ok(address) = result_address {
        true
} else {
       false
}
```
image

However, this doesn't:
```
if let Result::Ok(address) = result_address {
        return true
} else {
        return false
}
```
image

Likewise, this doesn't compile either:
```
if let Result::Ok(address) = result_address {
        return true;
} else {
        return false;
};
```
image

vaivaswatha:
Closes #1663

### DIFF
--- a/sway-core/src/ir_generation/function.rs
+++ b/sway-core/src/ir_generation/function.rs
@@ -718,14 +718,15 @@ impl FnCompiler {
         if codeblock.contents.is_empty() {
             Some(insert_type(TypeInfo::Tuple(Vec::new())))
         } else {
-            codeblock
-                .contents
-                .iter()
-                .find_map(|node| match &node.content {
-                    TypedAstNodeContent::ReturnStatement(trs) => Some(trs.expr.return_type),
-                    TypedAstNodeContent::ImplicitReturnExpression(te) => Some(te.return_type),
-                    _otherwise => None,
-                })
+            codeblock.contents.iter().find_map(|node| {
+                match node.gather_return_statements().first() {
+                    Some(TypedReturnStatement { expr }) => Some(expr.return_type),
+                    None => match &node.content {
+                        TypedAstNodeContent::ImplicitReturnExpression(te) => Some(te.return_type),
+                        _otherwise => None,
+                    },
+                }
+            })
         }
     }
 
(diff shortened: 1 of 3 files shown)


test/src/e2e_vm_tests/test_programs/should_pass/language/impure_ifs/src/main.sw
```diff
@@ -0,0 +1,46 @@
+script;
```
mohammadfawaz, line 1: Nice change! Thanks.

Do you mind also adding a test case with `if let`?

vaivaswatha: > Nice change! Thanks.
> 
> Do you mind also adding a test case with `if let` like the one in the linked issue?

Done. But there's a failure in one of the vec tests. Let me try and debug that.

otrho (commented):
The changes to the type checking are good, I'm a bit iffy on the IR as mentioned below.  But will happily approve once it's passing CI.

sway-ir/src/constant.rs
```diff
@@ -171,4 +171,21 @@ impl Constant {
         ));
         Value::new_constant(context, value, span_md_idx)
     }
+
+    pub fn eq(&self, context: &Context, other: &Constant) -> bool {
```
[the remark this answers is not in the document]
vaivaswatha: > Where is this used? I did a quick search and couldn't find anything.

Ah sorry. One more my earlier attempts used this function. I'll remove it.

sway-ir/src/block.rs
```diff
@@ -149,6 +149,11 @@ impl Block {
         })
     }
 
+    pub fn is_terminated_by_ret(&self, context: &Context) -> bool {
+        self.get_term_inst(context)
+            .map_or_else(|| false, |i| matches!(i, Instruction::Ret { .. }))
```
otrho, line 154:
```suggestion
            .map_or(false, |i| matches!(i, Instruction::Ret { .. }))
```

otrho (approved):
Yay.

otrho (approved):

mohammadfawaz (approved):
Looks good! Thanks.

Issue #1568, opened 2022-05-17
Implement "range pattern" (i.e. `..`)


emilyaherbert:
It would be cool to have the range pattern available for ignoring cases in a pattern that aren't important. Like:

```
struct Point {
  x: u64,
  y: u64
}

let p = Point {
  x: 1u64,
  y: 2u64,
};

match p {
  Point { x, .. } => { /* something with x */ },
  _ => { /* something else */ },
}
```

PR #2032, opened 2022-06-17
Improve struct patterns with new warnings and rest pattern support.

tritao:
This commit improves a couple aspects of handling of struct patterns.

First of all, it adds semantic checking (with a new error) when
patterns are missing usage of all fields, which previously compiled fine with no errors:

```
error
  --> src/main.sw:15:9
   |
13 | 
14 |     let z = match p {
15 |         Point { x } => { x },
   |         ^^^^^^^^^^^ Pattern does not mention field: y
16 |     };
   |
____
```

Then it adds support for rest pattern, `..`, which can be used as the
last token in a pattern to not have to specify all the struct fields, like so:

```rust
fn rest_pattern() {
    let p = Point {
        x: 1u64,
        y: 2u64,
    };

    match p {
        Point { x, .. } => { x },
    };
}
```

The semantic AST model was updated to support modeling this pattern,
and further type checking was added to the code.

There is also a new error for when the rest pattern doesn't appear as
the last location, or when it appears multiple times:

```
error
  --> src/main.sw:15:20
   |
13 | 
14 |     match p {
15 |         Point { x, .., .. } => { x },
   |                    ^^ Unexpected rest token, must be at the end of pattern.
16 |     };
17 | }
   |
____

```

And lastly, tests were added to cover the changes and new functionality.

Closes https://github.com/FuelLabs/sway/issues/1568.

### DIFF
--- a/sway-parse/src/pattern.rs
+++ b/sway-parse/src/pattern.rs
@@ -75,6 +75,19 @@ impl Parse for Pattern {
             return Ok(Pattern::Constructor { path, args });
         }
         if let Some(fields) = Braces::try_parse(parser)? {
+            let _fields: &Punctuated<PatternStructField, CommaToken> = fields.get();
+            let rest_pattern = _fields
+                .value_separator_pairs
+                .iter()
+                .find(|(p, _)| matches!(p, PatternStructField::Rest { token: _ }));
+
+            if let Some((rest_pattern, _)) = rest_pattern {
+                return Err(parser.emit_error_with_span(
+                    ParseErrorKind::UnexpectedRestPattern,
+                    rest_pattern.span(),
+                ));
+            }
+
             return Ok(Pattern::Struct { path, fields });
         }
         match path.try_into_ident() {
@@ -88,22 +101,39 @@ impl Parse for Pattern {
 }
 
 #[derive(Clone, Debug)]
-pub struct PatternStructField {
-    pub field_name: Ident,
-    pub pattern_opt: Option<(ColonToken, Box<Pattern>)>,
+pub enum PatternStructField {
+    Rest {
+        token: DoubleDotToken,
+    },
+    Field {
+        field_name: Ident,
+        pattern_opt: Option<(ColonToken, Box<Pattern>)>,
+    },
 }
 
 impl Spanned for PatternStructField {
     fn span(&self) -> Span {
-        match &self.pattern_opt {
-            Some((_colon_token, pattern)) => Span::join(self.field_name.span(), pattern.span()),
-            None => self.field_name.span(),
+        use PatternStructField::*;
+        match &self {
+            Rest { token } => token.span(),
+            Field {
+                field_name,
+                pattern_opt,
+            } => match pattern_opt {
+                Some((_colon_token, pattern)) => Span::join(field_name.span(), pattern.span()),
+                None => field_name.span(),
+            },
         }
     }
 }
 
 impl Parse for PatternStructField {
     fn parse(parser: &mut Parser) -> ParseResult<PatternStructField> {
+        if parser.peek::<DoubleDotToken>().is_some() {
+            let token = parser.parse()?;
+            return Ok(PatternStructField::Rest { token });
+        }
+
         let field_name = parser.parse()?;
         let pattern_opt = match parser.take() {
             Some(colon_token) => {
@@ -112,7 +142,7 @@ impl Parse for PatternStructField {
             }
             None => None,
         };
-        Ok(PatternStructField {
+        Ok(PatternStructField::Field {
             field_name,
             pattern_opt,
         })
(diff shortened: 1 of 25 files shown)


sway-parse/src/pattern.rs
```diff
@@ -75,6 +75,19 @@ impl Parse for Pattern {
             return Ok(Pattern::Constructor { path, args });
         }
         if let Some(fields) = Braces::try_parse(parser)? {
+            let _fields: &Punctuated<PatternStructField, CommaToken> = fields.get();
+            let rest_pattern = _fields
```
otrho, line 79:
```suggestion
            let fields: &Punctuated<PatternStructField, CommaToken> = fields.get();
            let rest_pattern = fields
```
Typically you only use `_name` to ignore it explicitly.  Is there a reason to not ust use `fields.get().value_separator_pairs...` though?

otrho (commented):
Looks good!  Could you expand the `should_pass` test to match against some more complex patterns?  Nested structs, enums and arrays maybe.  Having `rest` match with a bunch of other fields rather than just one and matching more than one missing field.

sway-parse/src/pattern.rs
tritao: I originally tried to do it like that, but Rust gives me this:

```
error[E0282]: type annotations needed
  --> sway-parse/src/pattern.rs:79:18
   |
78 |               let rest_pattern = fields
   |  ________________________________-
79 | |                 .get()
   | |__________________^^^_- this method call resolves to `&T`
   |                    |
   |                    cannot infer type for type parameter `T`
   |
   = note: type must be known at this point

error[E0609]: no field `value_separator_pairs` on type `&_`
  --> sway-parse/src/pattern.rs:80:18
   |
80 |                 .value_separator_pairs
   |                  ^^^^^^^^^^^^^^^^^^^^^

Some errors have detailed explanations: E0282, E0609.
```

I am still new to Rust so not really sure if there is a better way to fix it.

If I try your suggestion then I also get:

```
error[E0308]: mismatched types
  --> sway-parse/src/pattern.rs:91:47
   |
91 |             return Ok(Pattern::Struct { path, fields });
   |                                               ^^^^^^ expected struct `brackets::Braces`, found reference
   |
   = note: expected struct `brackets::Braces<punctuated::Punctuated<PatternStructField, keywords::CommaToken>>`
           found reference `&punctuated::Punctuated<PatternStructField, keywords::CommaToken>`
```

Should I just rename to something else without an underscore, maybe `inner_fields` or something?

tritao:
> Looks good! Could you expand the `should_pass` test to match against some more complex patterns? Nested structs, enums and arrays maybe. Having `rest` match with a bunch of other fields rather than just one and matching more than one missing field.

Thanks for the review. Sure thing, I will add some more testing.

sway-parse/src/pattern.rs
mohammadfawaz: So this is one way that works:
```rust
            let rest_pattern = Braces::<Punctuated<PatternStructField, CommaToken>>::get(&fields)
                .value_separator_pairs
                .iter()
                .find(|(p, _)| matches!(p, PatternStructField::Rest { token: _ }));
```
Note that we're using `get()` here as an associated function instead of calling `fields.get()` which seems to need additional type information. `get()` itself is not generic but the impl block where it's defined is. That's why this gets a bit annoying. Maybe @otrho has better ideas.

otrho: Ahh, right, I didn't notice that you need the original fields later on, so you can't shadow them as I suggested.  And as Mo pointed out the type attribution is needed since `get` is from `Braces<T>`.  I think `Braces::<Punctuated<_, _>>` would still work here, but still a type is needed.

So, really you've done it all correctly, and my only nit was that identifiers beginning with `_` are explicitly ignored by the compiler when reporting unused values.  `rustc` will suggest to prefix unused arguments or patterns with `_` when it finds them 'if this is intentional'.

So yeah, originally I saw this and thought 'why is it unused', but it obviously isn't.  Perhaps renaming `_fields` to `inner_fields` or something similar would be better.

mohammadfawaz (approved):
LGTM! @emilyaherbert should probably also take a look.

otrho:
> @otrho Updated the PR with some more tests and with the previous suggestion to use `_` in the explicit typed field. I tried adding some tests with arrays, but AFAICT struct destructuring inside arrays is not supposed to work.

Sure, if we don't have destructuring in arrays yet then that's fine.  I was just trying to think of more test examples.
 
> Though maybe you meant rest pattern support in regular array, types, and enums (without structs). If that is the case, then I would rather just handle struct patterns in this PR, and I will take a look at those kinds of patterns in a following PR.

No, I meant the former. 🙂 That's certainly something for a later PR.

tritao:
Rebased to latest master with config-driven testing.

mohammadfawaz (approved):
Will you create an issue to document this new behavior in the Sway Book? Looks good otherwise!

tritao:
> Will you create an issue to document this new behavior in the Sway Book? Looks good otherwise!

Sure, opened https://github.com/FuelLabs/sway/issues/2067.

PR #2038, opened 2022-06-17
Small nits in the standard library

mohammadfawaz:


adlerjohn (approved):

sway-lib-std/src/storage.sw
```diff
@@ -4,8 +4,7 @@ use ::hash::sha256;
 use ::context::registers::stack_ptr;
 
 /// Store a stack variable in storage.
-#[storage(write)]
-pub fn store<T>(key: b256, value: T) {
+#[storage(write)]pub fn store<T>(key: b256, value: T) {
```
nfurfaro, line 7: Do we have to live with these attributes formatted to be on the same line for now?

nfurfaro (approved):
One minor nit that shows up in a couple spots, but otherwise good changes. I love seing `+=`!

sway-lib-std/src/storage.sw
mohammadfawaz: Fmt v2 fixes this!

PR #2041, opened 2022-06-18
Add `Identity` type to Sway Book

### LINKED ISSUE #2040: Add Identity section to the Sway Book
The `Identity` type definition is not included in the Sway Book. There should be a section that includes its definition and how to cast to and from the `Identity` type.

bitzoic:
- Added a section under Blockchain Types that defines the `Identity` type. 
- Added an example of casting from `Address` type to `Identity` type
- Added an example of returning to `ContractId` type from `Identity` type
- Added an example of using the `Identity` type but needed different execution based on the underlying type

Closes #2040


docs/src/basics/blockchain_types.md
```diff
@@ -43,3 +43,46 @@ let my_number: b256 = 0x00000000000000000000000000000000000000000000000000000000
...
+
+The `Identity` type is an enum that allows for the handling of both `Address` and `ContractId` types. This is useful in cases where either type is needed.
+
+An `Identity` is implemented as follows.
+
+```sway
+pub enum Identity {
+    Address: Address,
+    ContractId: ContractId,
+}
```
adlerjohn, line 57: These should all be in actual Sway files, tested in CI, and included using `{{#include }}`: https://rust-lang.github.io/mdBook/format/mdbook.html#including-files. Otherwise it's too easy for them to drift and we end up with non-working code in The Book.

nfurfaro:
Maybe add an example of access control using `msg_sender()` and a stored `owner: Identity` ?
Something like:
```
require(msg_sender().unwrap() == storage.owner, MyError::UnauthorizedUser);
```
I'm not sure if this exact syntax works yet, but should be close if not.

docs/src/basics/blockchain_types.md
```diff
@@ -43,3 +43,46 @@ let my_number: b256 = 0x00000000000000000000000000000000000000000000000000000000
 let my_contract_id: ContractId = ~ContractId::from(my_number);
 let forty_two: b256 = my_contract_id.into();
 ```
+
+## `Identity` Type
+
+The `Identity` type is an enum that allows for the handling of both `Address` and `ContractId` types. This is useful in cases where either type is needed.
```
adlerjohn, line 49:
```suggestion
The `Identity` type is an enum that allows for the handling of both `Address` and `ContractId` types. This is useful in cases where either type is accepted, e.g. receiving funds from an identified sender, but not caring if the sender is an address or a contract.
```

docs/src/basics/blockchain_types.md
```diff
@@ -43,3 +43,46 @@ let my_number: b256 = 0x00000000000000000000000000000000000000000000000000000000
...
+    Address: Address,
+    ContractId: ContractId,
+}
+```
+
+Casting to an `Identity` must be done explicity:
+
+```sway
+let my_number: b256 = 0x000000000000000000000000000000000000000000000000000000000000002A;
+let my_address: Address = ~Address::from(my_number);
```
adlerjohn, line 64: I would simply this to just `~Address::from(0x000000000000000000000000000000000000000000000000000000000000002A)` for brevity.

adlerjohn:
> The `==` should work now

`Eq` has been implemented for `Identity` as of #2028

docs/src/basics/blockchain_types.md
mohammadfawaz: I think we need a special directory for all these little snippets like https://github.com/rust-lang/book/tree/main/listings and use anchors with them. These wouldn't be full examples like the examples directory that we have today, but small test cases that are meant only for the book but are also kept in sync with the compiler if anything changes.

mohammadfawaz: Not suggesting this to be done in this PR.. for now, maybe include a single full example with all the different cases you're showing in this section?

docs/src/basics/blockchain_types.md
```diff
@@ -43,3 +43,35 @@ let my_number: b256 = 0x00000000000000000000000000000000000000000000000000000000
...
 ```
+
+## `Identity` Type
+
+The `Identity` type is an enum that allows for the handling of both `Address` and `ContractId` types. This is useful in cases where either type is accepted, e.g. receiving funds from an identified sender, but not caring if the sender is an address or a contract.
+
+An `Identity` is implemented as follows.
+
+```sway
+{{#include ../../../sway-lib-std/src/identity.sw:7:10}}
```
mohammadfawaz, line 54: While I like the idea here, I worry these line number can get out of date for any number of reasons. We have the following options:
1. Keep it as is
2. Inline the code directly here
3. Use anchors in the standard library itself. 

If we're going to start documenting the standard library directly in the Sway Book (due to the concerns that @adlerjohn raised with docstrings in Sway code), we may have to do 3 eventually in other places. John, what do you think?

mohammadfawaz: Or maybe just have a link to the `identity.sw` file instead of having code in here.

examples/identity/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[project]
```
mohammadfawaz, line 1: We don't need this file right?

bitzoic: Yeah I think you're right

docs/src/basics/blockchain_types.md
adlerjohn: There's no issue, but only add them as needed.

examples/identity/.gitignore
```diff
@@ -0,0 +1,4 @@
+out
+target
+
+Forc.lock
```
adlerjohn, line 4: Commit `Forc.lock` to git, see other examples.

adlerjohn (commented):
CI failing

examples/identity/Forc.toml
```diff
@@ -0,0 +1,7 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "identity"
+
+[dependencies]
```
adlerjohn, line 7: This needs to be a path to the stdlib in this repo, so that that's tested instead of the latest release. Note that the committed `Forc.lock` is using `v0.16.1` instead of this branch's stdlib. See how other examples do this.

bitzoic: Should pass now

bitzoic: Nevermind, looks like there is a formatting error from an earlier commit suggestion

adlerjohn (approved):

PR #2044, opened 2022-06-18
Add `abi` handling to `sway-fmt-v2`

### LINKED ISSUE #2043: Add `abi` handling to `sway-fmt-v2`

eureka-cpu:
WIP
- [x] #2061 merged first
- [x] Add indentation handling
- [x] Unit test pass

Closes #2043


mitchmindtree:
Review was requested but looks like this is in draft mode? Don't forget to hit "Ready for review" when ready!

eureka-cpu:
> Review was requested but looks like this is in draft mode? Don't forget to hit "Ready for review" when ready!

It's still a work in progress, my bad!

eureka-cpu:
@kayagokalp @mitchmindtree having some trouble figuring out why the indentations aren't sticking for `abi` but seem to work for `enum`. Could use a review, thanks!

sway-fmt-v2/src/items/item_abi.rs
```diff
@@ -74,12 +74,12 @@ impl CurlyBrace for ItemAbi {
             ItemBraceStyle::AlwaysNextLine => {
                 // Add openning brace to the next line.
                 line.push_str(&format!("\n{}\n", open_brace));
-                shape = shape.block_indent(1);
+                shape = shape.block_indent(4);
```
mitchmindtree, line 77: All of these `4` literals should be behind a const or come from config

mitchmindtree (commented):
@eureka-cpu what are your thoughts on first tackling `fn` item handling? Even empty ones first, i.e. ignoring statements and expressions within the body to begin.

Then you perhaps you could re-use the fn logic in the `abi` and `trait` items for handling the methods?

sway-fmt-v2/src/items/item_abi.rs
eureka-cpu: Agreed, I was thinking that as I placed it there haha.

As for the `fn` handling, I also agree. It's sort of a big task since the rest of the formatter will rely on it. For the time being I think this could still go through, and I can work on `fn` next then implement it here. Either way is fine by me since they're both necessary (:

sway-fmt-v2/src/items/item_abi.rs
```diff
@@ -1,8 +1,97 @@
...
+
+        // Add item fields
+        // abi_items
+        formatted_code += self
+            .abi_items
+            .clone()
+            .into_inner()
+            .iter()
+            .map(|item| -> FormattedCode {
+                let mut buf = String::new();
```
kayagokalp, line 29: For indentation to be added, you can add something like

`buf.push_str(&formatter.shape.indent.to_string(formatter)); ` 

so that the current level of indentation is pushed before the formatted code pieces.

I am following a similar approach at [structs](https://github.com/FuelLabs/sway/blob/60da558a85563ddb5ffb4cffd46563c13b0d5935/sway-fmt-v2/src/items/item_struct.rs#L128) and [enums](https://github.com/FuelLabs/sway/blob/bdfb7449cac1e80ee4258abf1882c3caec99111c/sway-fmt-v2/src/items/item_enum.rs#L51) as well.

kayagokalp (dismissed):
LGTM. But let's note that formatting the following example does not change anything. 
```
contract;

abi MyContract {
        fn test_function()    -> bool;

        fn      test_function() -> bool;

        fn   test_function() -> bool;
}
```
Since we are not applying formatting to function signatures any trailing spaces are also kept in the formatted output. 

We can trim the function signatures so that we are not pushing the extra whitespaces of fn signatures to the formatted `abi` output but we could also deal with it while we are implementing the actual formatting logic for `fn`. Leaving this decision up to you and approving 👍

eureka-cpu:
@kayagokalp I think for now it's okay, since the formatter is still a WIP. I will tackle `fn` next!

JoshuaBatty (approved):
Killing it. LGTM

sway-fmt-v2/src/items/item_enum.rs
```diff
@@ -79,26 +79,30 @@ impl Format for ItemEnum {
...
         let brace_style = formatter.config.items.item_brace_style;
+        let extra_width = formatter.config.whitespace.tab_spaces;
         let mut shape = formatter.shape;
+        let open_brace = Delimiter::Brace.as_open_char();
         match brace_style {
             ItemBraceStyle::AlwaysNextLine => {
                 // Add openning brace to the next line.
-                line.push_str("\n{\n");
-                shape = shape.block_indent(1);
+                line.push_str(&format!("\n{}\n", open_brace));
```
mitchmindtree, line 88: A nicer alternative to using `format!` when appending to an existing `String` (or any buffer, including `stdout`/`stderr`!) is to use `write!` or `writeln!`, e.g.

```suggestion
                writeln!(line, "\n{}", open_brace);
```

sway-fmt-v2/src/utils/attribute.rs
```diff
@@ -21,7 +21,7 @@ impl<T: Parse + Format> Format for Annotated<T> {
     }
 }
 
-trait FormatDecl {
+pub trait FormatDecl {
```
mitchmindtree, line 24:
```suggestion
pub(crate) trait FormatDecl {
```

We should prefer `pub(crate)` in general unless we are prepared to support downstream use of the API and handle breaking changes to it in semver accordingly.

mitchmindtree (approved):
I've left a couple of optional things to address, but neither are worth blocking this.

PR #2058, opened 2022-06-20
Add struct formatting to sway-fmt-v2

### LINKED ISSUE #2053: `sway-fmt-v2` should be able to format `structs`
This issue is for tracking the progress of `sway-fmt-v2` implementation for structs.

related & part of https://github.com/FuelLabs/sway/issues/1516.

kayagokalp:
closes #2053.

~[WIP] I am still testing this, opening up for visibility.~

- [x] Implement alignment
- [x] Add test for alignment


mitchmindtree (commented):
@kayagokalp @eureka-cpu you guys are killing it!

sway-fmt-v2/src/fmt.rs
```diff
@@ -102,7 +102,87 @@ pub const TEST:u16=10;"#;
...
+        let sway_code_to_format = r#"contract;
+pub struct Foo {
+   barbazfoo: u64,
+   baz  : bool,
+}
+"#;
+        let correct_sway_code = r#"contract;
+
+pub struct Foo {
+ barbazfoo: u64,
```
eureka-cpu, line 142: This will need to be 4 spaces, since the default is 4

sway-fmt-v2/src/items/item_struct.rs
```diff
@@ -1,8 +1,213 @@
...
+    fn get_formatted_len(&self) -> usize {
+        // TODO while determininig the length we may want to format to some degree and take length.
+        let str_item = &self.span().as_str().len();
+        *str_item as usize
+    }
+}
+
+impl CurlyBrace for ItemStruct {
+    fn open_curly_brace(line: &mut String, formatter: &mut Formatter) {
+        let brace_style = formatter.config.items.item_brace_style;
```
eureka-cpu, line 180: Add a variable, ` let extra_width = formatter.config.whitespace.tab_spaces`

sway-fmt-v2/src/items/item_struct.rs
```diff
@@ -1,8 +1,213 @@
...
+        let mut shape = formatter.shape;
+        match brace_style {
+            ItemBraceStyle::AlwaysNextLine => {
+                // Add openning brace to the next line.
+                line.push_str("\n{");
+                shape = shape.block_indent(1);
```
eureka-cpu, line 186: And use it here

sway-fmt-v2/src/items/item_struct.rs
```diff
@@ -1,8 +1,213 @@
...
+            }
+            _ => {
+                // Add opening brace to the same line
+                line.push_str(" {");
+                shape = shape.block_indent(1);
```
eureka-cpu, line 191: here

sway-fmt-v2/src/items/item_struct.rs
```diff
@@ -1,8 +1,213 @@
...
+            }
+        }
+
+        formatter.shape = shape;
+    }
+
+    fn close_curly_brace(line: &mut String, formatter: &mut Formatter) {
+        line.push('}');
+        // If shape is becoming left-most alligned or - indent just have the defualt shape
+        formatter.shape = formatter.shape.shrink_left(1).unwrap_or_default();
```
eureka-cpu, line 201: same here, except no need to store in a variable just call it directly

sway-fmt-v2/src/fmt.rs
```diff
@@ -102,7 +102,87 @@ pub const TEST:u16=10;"#;
         let correct_sway_code = r#"contract;
 
 pub const TEST: u16 = 10;"#;
+        let mut formatter = get_formatter(Config::default(), Shape::default());
```
eureka-cpu, line 105: instead of using `get_formatter()` you can do `Formatter::default()`

eureka-cpu (changes requested):

sway-fmt-v2/src/items/item_struct.rs
```diff
@@ -1,8 +1,210 @@
...
+/// Format the struct if the multiline is passed as false struct will be formatted into a single line.
+///
+/// Example (multiline : false):
+/// struct Foo { bar: u64,  baz: bool }
+///
+/// Example (multiline : true):
+/// struct Foo {
+///  bar: u64,
+///  baz: bool,
+/// }
```
mitchmindtree, line 57: When writing doc comments, try to be considerate of the markdown formatting, even if it's just for private functions. I believe they can show up in an LSP-enabled editor with markdown formatting when hovered, and they can also be generated using `cargo doc --document-private-items` which I personally find handy when familiarising with a new code-base.

Here's how the current formatting looks:

Screenshot from 2022-06-23 11-22-53

Here's how it can look with some slight reformatting:

Screenshot from 2022-06-23 11-27-27

```rust
/// Format the struct if the multiline is passed as false struct will be formatted into a single line.
///
/// ## Examples
///
/// (multiline : false):
///
/// ```rust,ignore
/// struct Foo { bar: u64,  baz: bool }
/// ```
///
/// (multiline : true):
///
/// ```rust,ignore
/// struct Foo {
///     bar: u64,
///     baz: bool,
/// }
/// ```
```

mitchmindtree (approved):
Looking good. Thanks for the extensive tests, they make it much easier to review!

I've left a comment around good doc comment practise. It shouldn't block these changes, and I realise there are already a lot of places throughout the repo where formatting is a bit broken, but it would be sweet if we could try to be more aware of this going forward.

sway-fmt-v2/src/items/item_struct.rs
kayagokalp: Oh, that is really handy thanks. I will be using that to ensure that docs seem as I intended 👍

eureka-cpu: @kayagokalp did you want to add this, or should I go ahead and approve as is?

kayagokalp: @eureka-cpu After merging this I will be completing https://github.com/FuelLabs/sway/pull/2097. So I was planning to fix the doc there so that mitch won't need to approve again but if you prefer the other way we can go that way too.

eureka-cpu (approved):

PR #2017, opened 2022-06-16
Add conceptual distinction between replacing `TypeInfo::Self` and monomorphization

emilyaherbert:
- Blocked by https://github.com/FuelLabs/sway/pull/1982

PR #2059, opened 2022-06-20
Preventing the first use of a generic method from fixing its generic type

### LINKED ISSUE #1628: First use of a generic method fixes its generic type

emilyaherbert:
- Blocked by #2017
- Closes #1628


emilyaherbert:
Closed in favor of #2195

PR #2061, opened 2022-06-20
Adds `attribute` handling to `sway-fmt-v2`

### LINKED ISSUE #2012: Add `attribute` handling to `sway-fmt-v2`

eureka-cpu:
WIP, won't be able to fully test until `abi` is done

Closes #2012

### DIFF
--- a/sway-fmt-v2/src/items/item_abi.rs
+++ b/sway-fmt-v2/src/items/item_abi.rs
@@ -1,8 +1,9 @@
 use crate::fmt::{Format, FormattedCode, Formatter};
 use sway_parse::ItemAbi;
+use sway_types::Spanned;
 
 impl Format for ItemAbi {
     fn format(&self, _formatter: &mut Formatter) -> FormattedCode {
-        todo!()
+        self.span().as_str().to_string()
     }
 }
--- a/sway-fmt-v2/src/items/item_use.rs
+++ b/sway-fmt-v2/src/items/item_use.rs
@@ -1,8 +1,9 @@
 use crate::fmt::{Format, FormattedCode, Formatter};
 use sway_parse::ItemUse;
+use sway_types::Spanned;
 
 impl Format for ItemUse {
     fn format(&self, _formatter: &mut Formatter) -> FormattedCode {
-        todo!()
+        self.span().as_str().to_string()
     }
 }
--- a/sway-fmt-v2/src/utils/attributes.rs
+++ b/sway-fmt-v2/src/utils/attributes.rs
@@ -1,11 +1,38 @@
 use crate::Formatter;
-use sway_parse::AttributeDecl;
+use sway_parse::attribute::AttributeDecl;
 use sway_types::Spanned;
-pub fn format_attributes(attributes: Vec<AttributeDecl>, _formatter: &mut Formatter) -> String {
-    // TODO format attributes
-    attributes
-        .into_iter()
-        .map(|x| x.span().as_str().to_string())
-        .collect::<Vec<String>>()
-        .join("\n")
+
+use super::bracket::SquareBracket;
+
+trait Format {
+    fn format(&self, line: &mut String, formatter: &mut Formatter);
+}
+
+pub fn format_attributes(attributes: Vec<AttributeDecl>, formatter: &mut Formatter) -> String {
+    let mut formatted_code = String::new();
+    for attr in &attributes {
+        AttributeDecl::format(attr, &mut formatted_code, formatter);
+    }
+
+    formatted_code
+}
+
+impl Format for AttributeDecl {
+    fn format(&self, line: &mut String, _formatter: &mut Formatter) {
+        line.push_str(self.hash_token.span().as_str());
+        Self::open_square_bracket(line, _formatter);
+        // TODO: attributes are joined with their args during `span()`
+        // but we may need to do formatting there eventually
+        line.push_str(self.attribute.span().as_str());
+        Self::close_square_bracket(line, _formatter);
+    }
+}
+
+impl SquareBracket for AttributeDecl {
+    fn open_square_bracket(line: &mut String, _formatter: &mut Formatter) {
+        line.push('[');
+    }
+    fn close_square_bracket(line: &mut String, _formatter: &mut Formatter) {
+        line.push_str("test]\n");
+    }
 }
--- a/sway-fmt-v2/src/utils/attributes.rs
+++ b/sway-fmt-v2/src/utils/attributes.rs
@@ -2,7 +2,7 @@ use crate::Formatter;
 use sway_parse::attribute::AttributeDecl;
 use sway_types::Spanned;
 
-use super::bracket::SquareBracket;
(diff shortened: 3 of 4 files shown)


sway-fmt-v2/src/items/item_abi.rs
```diff
@@ -1,8 +1,9 @@
 use crate::fmt::{Format, FormattedCode, Formatter};
 use sway_parse::ItemAbi;
+use sway_types::Spanned;
 
 impl Format for ItemAbi {
     fn format(&self, _formatter: &mut Formatter) -> FormattedCode {
-        todo!()
+        self.span().as_str().to_string()
```
kayagokalp, line 7: Is this change relevant to this PR, (if not) maybe we can leave them as `todo!()`.

sway-fmt-v2/src/items/item_use.rs
```diff
@@ -1,8 +1,9 @@
 use crate::fmt::{Format, FormattedCode, Formatter};
 use sway_parse::ItemUse;
+use sway_types::Spanned;
 
 impl Format for ItemUse {
     fn format(&self, _formatter: &mut Formatter) -> FormattedCode {
-        todo!()
+        self.span().as_str().to_string()
```
kayagokalp, line 7: Same with the `item_abi.rs`

sway-fmt-v2/src/utils/attributes.rs
```diff
@@ -1,11 +1,69 @@
...
+                .into_inner()
+                .value_separator_pairs
+                .iter()
+                .map(|arg| format!("{}{}", arg.0.as_str(), arg.1.span().as_str()))
+                .collect::<Vec<String>>()
+                .join(" ");
+            args.pop(); // pop the ending space
+            args.pop(); // pop the ending comma
+            line.push_str(&args);
+        }
```
kayagokalp, line 47: Should we have something like `Self::close_parenthesis(line, formatter);` here?

kayagokalp (commented):
just left some small nits, otherwise looks like a solid ground 👍

sway-fmt-v2/src/items/item_abi.rs
eureka-cpu: Ah, these are so that the test included will pass. I can remove them and the test for the time being since the test won't pass the way we want to until both `abi` and this PR land

sway-fmt-v2/src/utils/attributes.rs
eureka-cpu: It is a bit needless, but I figured I'd use the traits for consistency

eureka-cpu:
@mitchmindtree @kayagokalp Just added something I've been thinking about for a while. Instead of having exposed characters in the formatter, maybe we can use the `tokens` directly from `sway_parse` leaving the only thing the formatter actually exposes is whitespaces (plus I think it looks nicer). Let me know what you think!

sway-fmt-v2/src/utils/attributes.rs
```diff
@@ -1,11 +1,82 @@
...
+impl<T: Parse> Format for Annotated<T> {
+    fn format(&self, formatter: &mut Formatter) -> FormattedCode {
+        let attributes = &self.attribute_list;
+        let mut formatted_code = String::new();
+
+        for attr in attributes {
+            AttributeDecl::format(attr, &mut formatted_code, formatter);
+        }
+
+        formatted_code
```
mitchmindtree, line 20: Should we also format `T` here? E.g. do we want to change `T: Parse` to `T: Parse + Format` and do something like:

```suggestion
        formatted_code + self.value.format(formatter)
```

Or is this being handled separately?

sway-fmt-v2/src/utils/attributes.rs
```diff
@@ -1,11 +1,82 @@
...
+        // ')'
+        Self::close_parenthesis(line, formatter);
+        // `]\n`
+        Self::close_square_bracket(line, formatter);
+    }
+}
+
+impl SquareBracket for AttributeDecl {
+    fn open_square_bracket(line: &mut String, _formatter: &mut Formatter) {
+        line.push(Delimiter::as_open_char(Delimiter::Bracket));
```
mitchmindtree, line 65:
```suggestion
        line.push(Delimiter::Bracket.as_open_char());
```

sway-fmt-v2/src/utils/attributes.rs
```diff
@@ -1,11 +1,82 @@
...
+    }
+    fn close_square_bracket(line: &mut String, _formatter: &mut Formatter) {
+        line.push_str(&format!(
+            "{}\n",
+            Delimiter::as_close_char(Delimiter::Bracket)
```
mitchmindtree, line 70:
```suggestion
            Delimiter::Bracket.as_close_char()
```

sway-fmt-v2/src/utils/attributes.rs
```diff
@@ -1,11 +1,82 @@
...
+        ));
+    }
+}
+
+impl Parenthesis for AttributeDecl {
+    fn open_parenthesis(line: &mut String, _formatter: &mut Formatter) {
+        line.push(Delimiter::as_open_char(Delimiter::Parenthesis))
```
mitchmindtree, line 77:
```suggestion
        line.push(Delimiter::Parenthesis.as_open_char())
```

sway-fmt-v2/src/utils/attributes.rs
```diff
@@ -1,11 +1,82 @@
...
+    }
+    fn close_parenthesis(line: &mut String, _formatter: &mut Formatter) {
+        line.push(Delimiter::as_close_char(Delimiter::Parenthesis))
```
mitchmindtree, line 80:
```suggestion
        line.push(Delimiter::Parenthesis.as_close_char())
```

sway-fmt-v2/src/utils/attributes.rs
eureka-cpu: I wasn't sure if we should include item formatting in the scope of a function intended to format annotations. Lmk if we should do this and I'll move that logic here, it's still in fmt.rs

eureka-cpu: Also would like to tackle that in a separate PR if we intend to do it

eureka-cpu: Went ahead and did it, I think it does feel cleaner after all

kayagokalp (approved):
lgtm, I think this will make things a lot cleaner 👍

JoshuaBatty (approved):
LGTM

sway-fmt-v2/src/utils/attributes.rs
mitchmindtree: > I wasn't sure if we should include item formatting in the scope of a function intended to format annotations.

I guess it's worth keeping in mind that when we say `impl` `Format` for `Annotated<T>`, we are actually saying that this implementation applies to all annotated `T`s, not just the annotations themselves. I think this turned out nice :ok_hand:

PR #2064, opened 2022-06-21
Fixed point numbers library

rostyslavtyshko:
Adding fixed point numbers library akin to https://github.com/abdk-consulting/abdk-libraries-solidity

Adding basic math functionality to other connected types  i.e. u8/16/32/64/128


test/src/e2e_vm_tests/test_programs/should_pass/stdlib/fixed_point_test/fixed_point_test/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[project]
+name = "fixed_point_test"
+version = "0.1.0"
+authors = ["Rostyslav Tyshko"]
+edition = "2021"
+license = "Apache-2.0"
+
+[dependencies]
+fuels = "0.14"
+fuels-abigen-macro = "0.14"
```
Braqzen, line 10:
```suggestion
fuels = "0.15"
fuels-abigen-macro = "0.15"
```

Unless the rest of the projects require an older version try to use the latest release if possible. [Version 16](https://github.com/FuelLabs/fuels-rs) is out within the last 24h. I have not used it / tested it so I suggest you try at least version 15

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/logarythmic_test/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["Rostyslav Tyshko"]
```
Braqzen, line 2:
```suggestion
authors = ["Fuel Labs <contact@fuel.sh>"]
```

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/logarythmic_test/Forc.toml
```diff
@@ -0,0 +1,8 @@
...
+entry = "main.sw"
+license = "Apache-2.0"
+name = "logarythmic_test"
```
Braqzen, line 5:
```suggestion
name = "logarithmic_test"
```

You may need to rename your directory etc. if you change this name

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u128_pow_test/Forc.toml
```diff
@@ -0,0 +1,7 @@
+[project]
+authors = ["Rostyslav Tyshko"]
```
Braqzen, line 2:
```suggestion
authors = ["Fuel Labs <contact@fuel.sh>"]
```

sway-lib-std/src/fixed_point.sw
```diff
@@ -0,0 +1,130 @@
+library fixed_point;
+//! A wrapper around U128 type for a library for Sway for mathematical functions operating with signed 64.64-bit fixed point numbers. 
+
+use ::u128::U128;
+use ::assert::assert;
+
+pub struct UFP64 {
```
Braqzen, line 7: Idk about this name. It doesn't really roll off the tongue that well although I have no suggestions for how to make it better

sway-lib-std/src/fixed_point.sw
```diff
@@ -0,0 +1,130 @@
...
+        self.value > other.value
+    }
+
+    pub fn lt(self, other: Self) -> bool {
+        self.value < other.value
+    }
+}
+
+impl core::ops::Add for UFP64 {
+    /// Add a UFP64 to a UFP64. Panics on overflow.
```
Braqzen, line 28: Is reverting on an overflow desirable behaviour? cc @nfurfaro @adlerjohn

sway-lib-std/src/fixed_point.sw
```diff
@@ -0,0 +1,130 @@
...
+    pub fn add(self, other: Self) -> Self {
+        UFP64 {
+            value: self.value + other.value
+        }
+    }
+}
+
+impl core::ops::Subtract for UFP64 {
+    /// Subtract a UFP64 from a UFP64. Panics of overflow.
```
Braqzen, line 37: Same here, isn't this an "underflow" instead of an "overflow"?

sway-lib-std/src/fixed_point.sw
```diff
@@ -0,0 +1,130 @@
...
+    pub fn subtract(self, other: Self) -> Self {
+        // If trying to subtract a larger number, panic.
+        assert(!(self.value < other.value));
```
Braqzen, line 40: Should we be using assertions here or `require` instead?

sway-lib-std/src/fixed_point.sw
```diff
@@ -0,0 +1,130 @@
...
+
+        let base = ~U128::from(1,0);
+
+        let self_up = ~U128::from(0, self.value.upper);
+        let self_lo = ~U128::from(0, self.value.lower);
+
+        let other_up = ~U128::from(0, other.value.upper);
+        let other_lo = ~U128::from(0, other.value.lower);
+
+        let mut up_up = self_up * other_up;
```
Braqzen, line 60: I'd appreciate non-abbreviated names. Ones that are descriptive to any reader who does not have insider knowledge on what is going on

sway-lib-std/src/fixed_point.sw
```diff
@@ -0,0 +1,130 @@
...
+            value: ~U128::max(),
+        }
+    }
+
+    /// The size of this type in bits.
+    pub fn bits() -> u32 {
+        128
+    }
+
+    pub fn inv(number: UFP64) -> Self {
```
Braqzen, line 110: What is `inv`? Can we use proper names?

sway-lib-std/src/fixed_point.sw
nfurfaro: Generally, yes. When this is not the desired behavior something like `overflowing_add()` must be used explicitly (or the vm flags must be modified to allow overflowing behavior).

sway-lib-std/src/fixed_point.sw
nfurfaro: I usually go with assertions for things that should never fail, vs requires to prevent bad user input.
A failing assertion would then indicate a bug somewhere that needs to be fixed.

emilyaherbert:
@rostyslavtyshko  can you clarify what was happening that was causing the related bug on this PR?

Braqzen:
> @rostyslavtyshko can you clarify what was happening that was causing the related bug on this PR?

Compiling Sway and the library works as expected however when you attempt to import the U128 type and use the Expo trait with the pow() function it would complain that pow() does not exist no matter how it was imported and where it was located.

E.g. `some_u128_var.pow()` with the error that `pow()` is not found on the type.

Sorry, got ahead of myself and edited your comment by accident

sway-lib-std/src/fixed_point.sw
adlerjohn: ☝️ By default all arithmetic operations panic (or rather, revert) on overflow. One could add some logic to not revert if the panic-on-overflow flag is disabled, but that can be left for future work since it's not nearly as useful with floating point numbers.

sway-lib-std/src/fixed_point.sw
adlerjohn: This is actually also an overflow. Underflow is only when the floating-point result has too many decimals, not when it's too negative.

sway-lib-std/src/fixed_point.sw
```diff
@@ -0,0 +1,130 @@
...
+        let one = ~U128::from(0,1);
+        
+        UFP64 {
+            value: one / number.value
+        }
+    }
+
+    pub fn avg(left: UFP64, right: UFP64) -> Self {
```
adlerjohn, line 118: We don't want to overload the base implementation in the standard library with all the convenience methods. As a reference, see the Rust implementation of `f64`: https://doc.rust-lang.org/std/primitive.f64.html

If a method isn't there, then it shouldn't be implemented in this PR. It can be filed as an issue where we can discuss its addition.

sway-lib-std/src/fixed_point.sw
```diff
@@ -0,0 +1,130 @@
...
+    /// The largest value that can be represented by this type,
+    pub fn max() -> UFP64 {
+        UFP64 {
+            value: ~U128::max(),
+        }
+    }
+
+    /// The size of this type in bits.
+    pub fn bits() -> u32 {
+        128
```
adlerjohn, line 107: HMMM. Why is a 64-bit floating point number 128 bits?

rostyslavtyshko: Because it is 64x64, not 64

sway-lib-std/src/fixed_point.sw
adlerjohn: Ah actually these are fixed point, not floating point, so `FP` might make sense.

sway-lib-std/src/fixed_point.sw
adlerjohn: Hmmm. I think it should be `UFP128` then, no?

adlerjohn (commented):
Would it be possible to make this closer to https://docs.rs/fixed/latest/fixed/ in terms of interface and naming? (Most of the changes to this PR would just be naming, since everything other than the core ops should be deferred to a future PR.)

Specifically something like https://docs.rs/fixed/latest/fixed/struct.FixedU64.html, with a hard-coded 32:32.

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/logarythmic_test/Forc.toml
rostyslavtyshko: Fixed

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u128_pow_test/Forc.toml
rostyslavtyshko: Fixed

sway-lib-std/src/fixed_point.sw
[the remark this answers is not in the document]
rostyslavtyshko: Deleted

sway-lib-std/src/fixed_point.sw
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    pub fn avg(left: UFP64, right: UFP64) -> Self {        
+        UFP64 {
+            value: (left.value + right.value) / 2,
+        }
+    }
+
+    pub fn gavg(left: UFP64, right: UFP64) -> Self {
```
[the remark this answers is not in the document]
rostyslavtyshko: Deleted

sezna (approved):

Braqzen (commented):
Seems like this PR is dependent upon #2340 so I would wait for that to go in before resolving the commented out code.

All of the logging should be removed btw. It's not too bad when testing via scripts but it should not be in the `std` (if there are any). The scripts should just contain the setup and assertions since there is no use in spamming logs during an automated workflow.

In the PR linked above I've mentioned that `Self` should be used consistently where possible as this allows the rest of the code to be more flexible. This should also be done here.

sway-lib-std/src/ufp128.sw
```diff
@@ -0,0 +1,227 @@
...
+        ~Self::from(nominator.upper, nominator.lower)
+    }
+}
+
+impl Exponentiate for UFP128 {
+    fn pow(self, exponent: Self) -> Self {
+        let nominator_pow = self.value.pow(exponent.value);
+        let one_u128 = ~U128::from(0, 1);
+        let two_u128 = ~U128::from(0, 2);
+        let u128_64 = ~U128::from(0, 64);
```
SwayStar123, line 193: Should probably have consistent names
```
let u128_1 = ~U128::from(0, 1);
let u128_2 = ~U128::from(0, 2);
let u128_64 = ~U128::from(0, 64);
```

rostyslavtyshko: Fixed

simonr0204 (approved):

sezna (approved):

simonr0204 (approved):

Braqzen (commented):
1. There are various sections which are commented out. Should all of them remain commented out or are they able to be completed? Is there an issue ready to track this? We would not want this to get buried.
2. Some of the tests do not look like tests since they simply log and do not assert anything. Logs should be removed from tests and only assertions should remain.
3. There are various imports that use the glob operator and I'm unsure if they are needed. There's at least one which I think can be explicit, maybe more.
4. Sway now has a [prelude](https://fuellabs.github.io/sway/v0.24.5/introduction/standard_library.html#standard-library-prelude) which means that there are certain types that do not have to be imported. I've seen some of those in this PR. Can these imports be removed (since they are unnecessary) or must they still be imported because this is part of the standard library and there's something that I'm unaware of?
5. `impl core::ops::Divide for UFP128` has a `let mut s = self;`. in `divide()`. Can this not be changed to something like `ref mut self` in the params instead?

mohammadfawaz:
As discussed previously, it's better to move this PR to the [`sway-libs` repository](https://github.com/FuelLabs/sway-libs). Once https://github.com/FuelLabs/sway/issues/2873 is resolved, we can bring back all the math libraries into `sway-lib-std`.

Changing this to a draft in the meantime.

mohammadfawaz:
I'm going to close this in favor of https://github.com/FuelLabs/sway-libs/pull/32 and https://github.com/FuelLabs/sway-libs/pull/31.

Issue #1996, opened 2022-06-15
The compiler should protect against using keywords as identifiers


mohammadfawaz:
https://github.com/FuelLabs/sway/issues/1946 is one example where the `storage` keyword was used as a type. Another example is
```rust
let mut mut = 0;
```
which compiles without any errors and that seems wrong. I suggest a general solution (in the parser probably?) that protects against the use of any keywords for `Ident`.

PR #2066, opened 2022-06-21
Improve reserved keywords checking and add support for raw identifiers.

tritao:
This PR deals with the usage and checking of reserved keywords as identifiers, for code like:

```
fn main() {
    let mut mut = 0;
}

It introduces a new error that checks if an identifier is a reserved keyword:

```
error
 --> /main.sw:4:13
  |
2 |
3 | fn main() {
4 |     let mut mut = 0;
  |             ^^^ Identifiers cannot be a reserved keyword.
5 | }
  |
____
```

There was an existing issue in the standard library, which has a library/module named `storage`.

Instead of working around this by renaming it to something else, an alternative solution with raw identifiers is implemented.

This raw identifier feature is implemented at the lexer level, and allows you to use keywords as identifiers in places that generally wouldn't be allowed.

Rust and a bunch of other modern languages also provide this escape hatch, and it seemed the simplest solution for me to handle the issue.

It activates by declaring an identifier prefixed with `r#`, just like Rust.

The complexity on the codebase to support this feature is pretty minimal, but if there any objections to this, I can easily remove it, but some other solution to the issue above will need to be figured out.

Closes https://github.com/FuelLabs/sway/issues/1996.


otrho:
The rule we tend to follow for Sway is to ask 'what would Rust do?' and AFAICT this is spot on. :slightly_smiling_face:

tritao:
> Thank you so much for this. Two things:
> 
> * We have a check here: https://github.com/FuelLabs/sway/blob/5318a09edf064d7c750a15c78343df83d5774af2/sway-core/src/semantic_analysis/ast_node/mod.rs#L268
>    that can now be removed. The function `check_if_name_is_invalid` can also be deleted along with the constant array `INVALID_NAMES` that it uses.

:+1:  

> * With the storage library now called `r#storage`, would users still be able to import items from it using `use std::storage::...`? I'm assuming yes given that the storage tests are still passing. That being said, i think we should still rename `storage` to something else (while we still can) because we want to avoid highlighting `storage` in `std::storage::` like the `storage` keyword in other spots in the user's code. But please keep the code you have now that allows for this escape hatch.

Yes, importing still works fine. I don't really know how much code "in the wild" is relying on the storage name, that so I didn't want to break anything. I think if the only issue with using storage as a name is highlighting, then that can be probably be fixed by using semantic-based highlighting. Either way I think that should probably be handled as a follow up PR.

mohammadfawaz (dismissed):
Great stuff! Thanks.

otrho (approved):

mohammadfawaz (approved):

sezna (approved):

mohammadfawaz (approved):

Issue #1185, opened 2022-04-07
Handle storage initializers


mohammadfawaz:
Blocker for supporting initialization with literals (primitive types):
* JSON ABI spec for how storage initializers should show up: https://github.com/FuelLabs/fuel-specs/issues/302

Additional blocker for supporting arbitrary initializers:
* Compile-time constant evaluation: https://github.com/FuelLabs/sway/issues/407

mohammadfawaz:
What is the right approach here? Does adding a new array to the JSON file like this work?
```json
[
   {
      "type":"storage_initializer",
      "location":"0x0101010101010101010101010101010101010101010101010101010101010101",
      "initial_value":"0x1234123412341234123412341234123412341234123412341234123412341234",
   },
   {
      "type":"storage_initializer",
      "location":"0x0101010101010101010101010101010101010101010101010101010101010102",
      "initial_value":"0x1234123412341234123412341234123412341234123412341234123412340000",
   }
]
[
    
   {
      "type":"function",
      "inputs":[
         {
            "name":"arg",
            "type":"u64"
         }
      ],
      "name":"entry_one",
      "outputs":[

      ]
   }
]
```

adlerjohn:
Note that constant evaluation isn't needed for allow initial storage for primitive types.

mohammadfawaz:
Yup, clarified the description.

mohammadfawaz:
The spec for that is now ready: https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/storage_initialization.md. This can now be implemented in the compiler.

PR #2078, opened 2022-06-21
Enable storage initializers and emit a storage initialization JSON

mohammadfawaz:
As per https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/storage_initialization.md
* Support storage initializers in the parser
* Type check storage initializers
* Serialize storage initializers using the corresponding state index into appropriate storage slots (key-value pairs) in such a way to match how storage accesses are transformed into `state_read_*/state_write_*` instructions. The initializers that are supported are ones that can be converted to constants by the const expr evaluator.
* In `lib.rs`, call the function that gets the storage slots and do all the error checking there as part of `TypeProgram`. `Vec<StorageSlot>` is a part of a `TypedProgram`.
* Changes in `forc-pkg` to simply read the storage slots from `TypedPorgram`.
* In `forc build`, generate a JSON file with the storage slots.
* In `forc deploy`, _sort_ the storage slots and use them to build the contract transaction. The VM expects sorted slots.
* Update the e2e tests to use the json-abi output directly from `compile_to_bytes` (without relying on `forc json-abi`) and do the same for storage_slots.

> **Note**: `forc json-abi` should be removed in a subsequent PR.

> **Note**: Initializers won't work in the SDK yet. Pending https://github.com/FuelLabs/fuels-rs/issues/87

Closes https://github.com/FuelLabs/sway/issues/1185


sway-core/src/ir_generation/const_eval.rs
```diff
@@ -288,3 +310,180 @@ fn const_eval_typed_ast_node(
...
     }
 }
+
+pub fn serialize_to_storage_slots(
+    constant: &Constant,
+    context: &Context,
+    ix: &StateIndex,
+    ty: &Type,
+    indices: &[usize],
+) -> JsonStorageInitializers {
```
otrho, line 320: This seems out of place in this file to me.  It's very storage/JSON specific, rather than 'constant evaluation' specific.

mohammadfawaz: Yeah I agree, I had a similar feeling. Generally that JSON stuff goes into the impl of the different types so it makes sense for this to go in Constant.rs but I didn't want to bloat sway-it with this stuff. So maybe a separate file in sway-core somewhere.

otrho (dismissed):

mohammadfawaz:
I'm missing some error handling. Turning into a draft until I finish that.

adlerjohn (commented):
Should the e2e test configuration also have an initial storage slots oracle, similar to the JSON ABI oracle?

mohammadfawaz:
> Should the e2e test configuration also have an initial storage slots oracle, similar to the JSON ABI oracle?

Done!

docs/src/SUMMARY.md
```diff
@@ -62,6 +62,7 @@
     - [forc deploy](./forc/commands/forc_deploy.md)
     - [forc init](./forc/commands/forc_init.md)
     - [forc json-abi](./forc/commands/forc_json-abi.md)
+    - [forc json-storage-slots](./forc/commands/forc_json-storage-slots.md)
```
adlerjohn, line 65: Is this command actually needed? It's an automatic output from the `build` process, same with `abi`. I would vote to remove `forc json-abi` instead of introducing more redundant commands.

mohammadfawaz: Alright, so I started by stopping the e2e test from relying on `forc json-abi` and removed the new `forc storage-slots-json` command. I agree that they seem redundant except maybe for specifying an output file. We could however add a flag `-json-abi-output-file` to `forc build` (and others) to specify where the JSON file should go, but that's not too important.

I will remove `forc json-abi` from `forc` in a subsequent PR.

adlerjohn: >  We could however add a flag -json-abi-output-file to forc build (and others) to specify where the JSON file should go

Yep, that looks like the better alternative to having entirely new commands.

forc-pkg/Cargo.toml
```diff
@@ -11,6 +11,7 @@ description = "Building, locking, fetching and updating Sway projects as Forc pa
 [dependencies]
 anyhow = "1"
 forc-util = { version = "0.16.2", path = "../forc-util" }
+fuel-tx = "0.12"
```
adlerjohn, line 14: `StorageSlot` should be under `fuel-types` (not `fuels-types`) I think. Can you file an issue to follow-up on moving that? We probably don't want to introduce a dependency on `fuel-tx` just for a type.

forc/src/ops/forc_deploy.rs
```diff
@@ -99,11 +103,14 @@ fn create_contract_tx(
 
     let salt = Salt::new([0; 32]);
     let static_contracts = vec![];
-    let storage_slots = vec![];
 
     let contract = Contract::from(compiled_contract);
     let root = contract.root();
-    let state_root = Contract::default_state_root();
+
+    // The VM requires that storage slots are sorted.
```
adlerjohn, line 110: Technically, there's no requirement that the JSON ABI is sorted. Would be nice to add that this step isn't strictly necessary and not guaranteed, and downstream tooling should sort it themselves.

test/src/e2e_vm_tests/test_programs/should_fail/nonconst_storage_init/json_abi_oracle.json
```diff
@@ -0,0 +1 @@
+[]
```
adlerjohn, line 1: Do you need this for a failing test?

forc-pkg/Cargo.toml
mohammadfawaz: Done 
https://github.com/FuelLabs/sway/issues/2115
https://github.com/FuelLabs/fuel-tx/issues/152

forc/src/ops/forc_deploy.rs
mohammadfawaz: I created this issue: https://github.com/FuelLabs/fuel-tx/issues/153. 

If I don't currently sort the slots, I get a VM error about the slots not being sorted 🤷

test/src/e2e_vm_tests/test_programs/should_fail/nonconst_storage_init/json_abi_oracle.json
mohammadfawaz: Nops, removed.

adlerjohn (approved):

mitchmindtree (commented):
Lgtm! Will let @otrho give the final tick in case he wants a final look at the IR stuff.

otrho (approved):

PR #2082, opened 2022-06-22
Update all the E2E should_fail tests to verify their output.

### LINKED ISSUE #723: Refactor the E2E test driver.

### LINKED ISSUE #1580: Update E2E test harness to verify specific failures.

otrho:
Using the `FileCheck` crate it can now pattern match against the compiler output to be sure the errors and/or warnings are exactly what we expect.

Closes #723.
Closes #1580.

### DIFF
--- a/test/src/e2e_vm_tests/test_programs/should_fail/excess_fn_arguments/test.toml
+++ b/test/src/e2e_vm_tests/test_programs/should_fail/excess_fn_arguments/test.toml
@@ -1,1 +1,5 @@
-category = "fail"
+# This test should be a "fail" but it's currently passing.
+# See https://github.com/FuelLabs/sway/issues/2081
+category = "compile"
+
+# check: ... something about too many args ...
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1072,6 +1072,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "filedescriptor"
+version = "0.8.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7199d965852c3bac31f779ef99cbb4537f80e952e2d6aa0ffeb30cce00f4f46e"
+dependencies = [
+ "libc",
+ "thiserror",
+ "winapi",
+]
+
+[[package]]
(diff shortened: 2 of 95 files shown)


kayagokalp:
I am in the middle of merging some stuff and will look into this as soon as it finishes 👍

mitchmindtree:
Okydoke @otrho, should be solved via #2090.

test/src/e2e_vm_tests/test_programs/should_fail/excess_fn_arguments/test.toml
```diff
@@ -1 +1,5 @@
-category = "fail"
+# This test should be a "fail" but it's currently passing.
```
mitchmindtree, line 1: I wonder we should have a flag or something that indicates that the test is known not to pass right now but _should_ in the future? Then we could potentially print a list of known not-yet-passing tests at the end of the test output as a reminder? That said I guess having an issue for these is probably enough. Either way, not a concern of this PR!

mitchmindtree: Ah I just noticed we have a `disabled` category - guess we could use that?

mitchmindtree (approved):
Wicked, TIL about `filecheck`.

Thanks loads for tackling these test issues @otrho, looking so much better!

test/src/e2e_vm_tests/test_programs/should_fail/excess_fn_arguments/test.toml
otrho: Yar, it prints the number of disabled tests at the end, I guess we could keep an eye on that.

Having an actual GitHub issue for things that should change in the future is a bit more robust.  I reckon having a flag in the suite would just get forgotten too easily.

tritao (approved):
Lgtm

Issue #1241, opened 2022-04-13
Convert remaining `Ident`s into `CallPath`s


mohammadfawaz:
We still parse things like variables and function declarations using `Ident`. `CallPath` is more powerful and it would be nice to switch everything over to use it. Probably worth waiting for the new parser before doing that though.

PR #2083, opened 2022-06-22
Use function arg types as a key when indexing functions in CFA

### LINKED ISSUE #758: Incorrect warning when the same method type is used for two different traits
The following code results in an incorrect warning:
```rust
script;

trait Numeric {
    fn get_number_of_bytes(self) -> u64;
}

trait Boolean {
    fn get_number_of_bytes(self) -> u64;
} 

struct U64 {
    n: u64,
}
struct Bool {
    b: bool,
}

impl Numeric for U64 {
    fn get_number_of_bytes(self) -> u64 {
        8
    }
}

impl Boolean for Bool {
    fn get_number_of_bytes(self) -> u64 {
        1
    }
}

fn main() {
    let n = U64 { n : 0 };
    let b = Bool { b : true };
    let bytes1 = n.get_number_of_bytes();
    let bytes2 = b.get_number_of_bytes();
}
```
The warning is as follows:
```
warning
  --> test/src/main.sw:19:5
   |
17 |   
18 |   impl Numeric for U64 {
19 |       fn get_number_of_bytes(self) -> u64 {
   |  _____-
20 | |         8
21 | |     }
   | |_____- This method is never called.
22 |   }
   |
warning
  --> test/src/main.sw:20:9
   |
18 | 
19 |     fn get_number_of_bytes(self) -> u64 {
20 |         8
   |         - This code is unreachable.
21 |     }
22 | }
```
I tracked this down a bit and it seems to be an issue in `control_flow_anlysis` where the `ControlFlowNamespace` relies on function names to insert and get the function node indices corresponding to a function application.

vaivaswatha:
This enables us to distinguish b/w same named different implementations of (possibly) different traits.

Another way to achive this would've been to use CallPaths, but those aren't available here (after type checking). We *could*
retain them during type checking but may be not worth doing all that.

Closes #758


mohammadfawaz:
This is great! Thanks!

What about situations where the arguments also have the same type (or no arguments at all)? Wouldn't we still get the warning?

E.g.:
```rust
trait Numeric {
    fn get_number_of_bytes() -> u64;
}

trait Boolean {
    fn get_number_of_bytes() -> u64;
} 

struct U64 {
    n: u64,
}
struct Bool {
    b: bool,
}

impl Numeric for U64 {
    fn get_number_of_bytes() -> u64 {
        8
    }
}

impl Boolean for Bool {
    fn get_number_of_bytes() -> u64 {
        1
    }
}

fn main() {
    let bytes1 = U64::get_number_of_bytes();
    let bytes2 = Bool::get_number_of_bytes();
}
```

vaivaswatha:
> 

You're right. That was my overlook. Let me use `CallPath`s and fix this rights.

vaivaswatha:
> > 
> 
> You're right. That was my overlook. Let me use `CallPath`s and fix this right.

@mohammadfawaz While at applications I can get the full call path, I'm unable to get that path for declarations. Without that I can't see how to unique out different declarations. 

I pinged @otrho for help and he pointed me out to the "declaration engine" work that's ongoing and I'll be able to get full paths to declarations after that. @mohammadfawaz how do you suggest I progress?

vaivaswatha:
@emilyaherbert 

@mohammadfawaz pointed me to https://github.com/FuelLabs/sway/issues/1241. Implementing this could work for me, as far as I can tell.
and earlier, @otrho had pointed me to https://github.com/FuelLabs/sway/issues?q=is%3Aissue+is%3Aopen+declaration+engine, which I gather you're planning around.

So your suggestion on what's the best step to take would be helpful.

otrho:
So @mohammadfawaz do you think this is urgent enough to fix now, or can it wait until after the declaration engine?  If #1241 will fix it _and_ it's fairly easy to do then perhaps @vaivaswatha should pivot there and then come back?

mohammadfawaz:
> So @mohammadfawaz do you think this is urgent enough to fix now, or can it wait until after the declaration engine? If #1241 will fix it _and_ it's fairly easy to do then perhaps @vaivaswatha should pivot there and then come back?

Yeah I agree. This is certainly not urgent.

vaivaswatha:
> > So @mohammadfawaz do you think this is urgent enough to fix now, or can it wait until after the declaration engine? If #1241 will fix it _and_ it's fairly easy to do then perhaps @vaivaswatha should pivot there and then come back?
> 
> Yeah I agree. This is certainly not urgent.

So shall I abandon this PR ?

vaivaswatha:
Abandoning this in favour of the declaration engine, as discussed above.

PR #2085, opened 2022-06-22
Fix: Failing BuildPlan generation with --locked does not return the cause of the error

### LINKED ISSUE #2084: BuildPlan generation is attempted even if there is a dependency cycle
discovered at https://github.com/FuelLabs/sway/pull/2082#issuecomment-1162796119.

If there is a dependency cycle, forc should not attempt to create the `BuildPlan`. Currently, after failing to create the `BuildPlan` from the lock file it also tries to create it from scratch.

kayagokalp:
closes #2084.

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -263,10 +263,10 @@ impl BuildPlan {
         // If there are errors we will need to create the BuildPlan from scratch, i.e fetch & pin everything
         let mut new_lock_cause = None;
         let mut plan = plan_result.or_else(|e| -> Result<BuildPlan> {
-            if e.to_string().contains("dependency cycle detected") {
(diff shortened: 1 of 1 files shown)


kayagokalp:
This is not enough to solve this as when there is a dependency cycle a lock file cannot be generated. The build process checks for the lock file and if it is not there tries to create the BuildPlan from the manifest file. That is where we find out the cycle actually so, the check for locked is happening before we found out about the cycle. I will try to work-out something.

kayagokalp:
This way it should be ready to review but I needed to move the `BuildPlan` generation up so that it is done before the check of locked. That way any problem coming from generation will be emitted. This solves the problem but has the downside of having to create the `BuildPlan` in every circumstance even though we provided `--locked` (before, if the lock file needed to change and `--locked` provided we were not trying to create the `BuildPlan`). cc @mitchmindtree

PR #2091, opened 2022-06-23
Adding `--time-phases` to `forc build`

### LINKED ISSUE #1953: Add a `--time-phases` flag that shows the duration of each compiler pass
In #1952 I added some basic time-stamping to the compiler in a temporary branch in order to check what has been causing the compiler to take so long while building `std`.

It would be nice to have something like this (but formatted much more nicely) that could be enabled by a `--time-phases` flag that could be passed to `forc build` and forwarded on to the compiler via the `BuildConfig`.

We could also enable it in our CI in order to better detect regressions in performance while working on the compiler.

#318 is related, this could act as a simple step in that direction in lieu of something more comprehensive.

seem-less:
Closes #1953


otrho (dismissed):
It'd be nice if the IR generation got one too, but it's not worth futzing with the pipeline to make it work...

mitchmindtree (commented):
Another great contribution from @seem-less :)

I wonder if we can clean this up a little by using a little `macro_rules!` at the beginning of `compile` that acts a bit like a closure over `build_config.time_phases`? I'm thinking something like:
```rust
// Time the given expression and print the result if `build_config.time_phases` is true.
macro_rules! time_expr {
    ($description:expr, $expression:expr) => {{
        if build_config.time_phases {
            let expr_start = std::time::Instant::now();
            let output = { $expression };
            println!("  Time elapsed to {}: {:?}", $description, expr_start.elapsed());
            output
        } else {
            $expression
        }
    }};
}
```
Then we could use it without needing the temporary `get_time_now()` bindings:
```rust
let ast_res = time_expr!("compile to AST", {
    compile_ast(manifest, build_config, namespace)?
});
```
The reason I suggest a macro and not a closure here is that the macro allows more flexibility around control flow with the given expression (e.g. returning early with `?`).

There must be a crate out there that does something like this already... I thought `tracing` had something like this, but after a quick look it appears no :thinking:

mitchmindtree (approved):
:ok_hand: Noice

otrho (approved):

PR #2092, opened 2022-06-23
Rename formatting `trait`s to specify what they format

eureka-cpu:
As a side effect of the formatter growing some code is becoming duplicated, this is an effort to reduce that by naming the formatting traits more appropriately to what they are intended to format.

### DIFF
--- a/sway-fmt-v2/src/fmt.rs
+++ b/sway-fmt-v2/src/fmt.rs
@@ -10,16 +10,24 @@ pub use crate::{
     error::{ConfigError, FormatterError},
 };
 
+/// Finalized Sway formatting configurations, and code shape.
 #[derive(Debug, Default)]
 pub struct Formatter {
     pub shape: Shape,
     pub config: Config,
 }
 
+/// Formatted Sway code.
 pub type FormattedCode = String;
 
-pub trait Format {
+/// Formatting for `ItemKind`.
+pub trait FormatItem {
     fn format(&self, formatter: &mut Formatter) -> FormattedCode;
+}
+
+/// Formatting for complex `ItemKind` fields e.g. `AttributeDecl` or `FnSignature`.
+pub trait FormatInner {
+    fn format(&self, line: &mut String, formatter: &mut Formatter);
 }
 
 impl Formatter {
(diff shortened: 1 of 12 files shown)


mitchmindtree (commented):
Hmmm I can see where you're coming from, but I'm not sure we should split `Format` into `FormatItem` and `FormatInner` like this

Generally, `trait`s are useful to allow functions to generalise over their inputs. E.g. allowing functions to generalise over `T` where `T: Format`. In our case, `Format` is quite useful because there are cases where we generalise over some `Format`-able `T` while traversing the tree (e.g. `Annotated<T>`).

I think adding traits solely because some function signature's match in a few different places can seem really attractive (e.g. "we're detecting some pattern, yay!"), but can also be a bit of a trap at times. If they're not assisting polymorphism, `traits` can add needless indirection, particularly if their role is quite ambiguous, and require the user to import the trait everywhere they use the method.

They can also make refactoring down the road a little trickier if it turns out one of the cases needs to be handled uniquely, and you can end up with a bunch of arguments in the trait methods that only some implementations use but most ignore.

Rather than adding a `FormatInner`, perhaps we can continue to use inherent methods and functions as we need them for now?

eureka-cpu:
> Hmmm I can see where you're coming from, but I'm not sure we should split `Format` into `FormatItem` and `FormatInner` like this
> 
> 
> 
> Generally, `trait`s are useful to allow functions to generalise over their inputs. E.g. allowing functions to generalise over `T` where `T: Format`. In our case, `Format` is quite useful because there are cases where we generalise over some `Format`-able `T` while traversing the tree (e.g. `Annotated<T>`).
> 
> 
> 
> I think adding traits solely because some function signature's match in a few different places can seem really attractive (e.g. "we're detecting some pattern, yay!"), but can also be a bit of a trap at times. If they're not assisting polymorphism, `traits` can add needless indirection, particularly if their role is quite ambiguous, and require the user to import the trait everywhere they use the method.
> 
> 
> 
> They can also make refactoring down the road a little trickier if it turns out one of the cases needs to be handled uniquely, and you can end up with a bunch of arguments in the trait methods that only some implementations use but most ignore.
> 
> 
> 
> Rather than adding a `FormatInner`, perhaps we can continue to use inherent methods and functions as we need them for now?

Gotcha! Thanks for the review (:

PR #2094, opened 2022-06-23
Refactor `forc_pkg::BuildConfig` -> `BuildProfile`, fix CLI arg handling

mitchmindtree:
Addresses some of the followup left after #1800.

Previously, if any of the `print` args were set, the rest of the
selected build profile was ignored. This changes the behaviour so that
the command line arguments only override their associated build profile
fields.

Also renames `BuildConfig` to `BuildProfile` and moves it from
`forc_pkg::pkg` to `forc_pkg::manifest` along with the rest of the
serializable manifest types.

### DIFF
--- a/forc-pkg/src/lib.rs
+++ b/forc-pkg/src/lib.rs
@@ -9,6 +9,6 @@ pub mod manifest;
 mod pkg;
 
 pub use lock::Lock;
-pub use manifest::{Manifest, ManifestFile};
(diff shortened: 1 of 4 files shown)


JoshuaBatty (approved):
nice, looks good.

binggh (approved):
Looks good to me - not entirely familiar with this part of the codebase yet but it seems like it's mostly just renaming and a last bit of functionality change at the end there with the print checks. I'll chip in an approval as well

PR #2095, opened 2022-06-23
forc: Improve the `print_*_asm` CLI option docs

mitchmindtree:
Previously, these implied no bytecode was generated if the flag was not set, however this is not the case.

cc @otrho do these ASM states in the new docs kinda make sense? Let me know if you think there's a better way to put this!

### DIFF
--- a/forc/src/cli/commands/build.rs
+++ b/forc/src/cli/commands/build.rs
@@ -21,13 +21,18 @@ pub struct Command {
     /// Path to the project, if not specified, current working directory will be used.
     #[clap(short, long)]
     pub path: Option<String>,
-    /// Whether to compile to bytecode (false) or to print out the generated ASM (true).
+    /// Print the finalized ASM.
+    ///
+    /// This is the state of the ASM with registers allocated and optimisations applied.
     #[clap(long)]
     pub print_finalized_asm: bool,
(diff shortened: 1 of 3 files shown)


otrho (approved):

otrho:
There isn't much difference between intermediate and finalised, mostly the removal of labels and redundant jumps, and the reg allocation.  But  there might be more in the future.  We introduced intermediate initially just for the labels because finalised asm just jumps to an offset which is inscrutable.

forc/src/cli/commands/build.rs
```diff
@@ -21,13 +21,18 @@ pub struct Command {
     /// Path to the project, if not specified, current working directory will be used.
     #[clap(short, long)]
     pub path: Option<String>,
-    /// Whether to compile to bytecode (false) or to print out the generated ASM (true).
+    /// Print the finalized ASM.
+    ///
+    /// This is the state of the ASM with registers allocated and optimisations applied.
```
binggh, line 26:
```suggestion
    /// This is the state of the ASM with registers allocated and optimizations applied.
```

American English perhaps? Seems good otherwise

binggh (approved):

kayagokalp (approved):

Issue #2081, opened 2022-06-22
The test `e2e_vm_tests/test_programs/should_fail/excess_fn_arguments` is erroneously passing.


otrho:
While implementing a fix for #1580 this test was discovered to be failing because it didn't import `core` properly, and once that was fixed the compilation succeeds when it definitely shouldn't.

```rust
script;

fn bar(a: bool, b: u64) -> u64 {
  if a {
    b * 2
  } else {
    0
  }
}

fn main() -> u64 {
  bar(true, 0, true) // <--- an extra arg.
}
```

After fixing this issue the test must be reverted from currently passing back to failing.  See its `test.toml`.

tritao:
This is happening because of the way typed argument checking is done in, which drops the extra arguments:

```rust
    let typed_call_arguments = arguments
        .into_iter()
        .zip(function_decl.parameters.iter())
```

Working on a fix.

PR #2096, opened 2022-06-23
Fix checking for too many function call arguments.

tritao:
This PR adds explicit checking for the case of too many function call arguments in function application instantiation.

This needs to be done before type checking because after type checking we drop the extra function call arguments due to having no type to check against.

During this PR development, it was found that Sway was allowing associated functions to be called with method call syntax,
which should not be allowed (Rust also throws an error in this case), and previously was just working by chance and a side
effect of the zip calls dropping the extra arguments.

So this adds a new error and check for this and updates some tests that were relying on this non-Rust-like behaviour.

There was another issue found, where the IR (at least the IR parser) currently cannot deal with empty structs, which was triggered by an existing IR test after being updated to work with the above changes.

Also re-enables the `should_fail/excess_fn_arguments` test.

Closes https://github.com/FuelLabs/sway/issues/2081.


vaivaswatha:
>This needs to be done before type checking because after type checking we drop the extra function call arguments due to having no type to check against.

Does this mean that, with your changes, `instantiate_function_application_inner` can never see `TooManyArgumentsForFunction`? That means we could also check for `TooFewArgumentsForFunction` in the outer functions and remove the checks entirely from `instantiate_function_application_inner` ?

sway-core/tests/sway_to_ir/get_storage_key.sw
```diff
@@ -1,16 +1,18 @@
 contract;
 
-struct Empty {}
+struct Foo {
+    foo: u32
```
mohammadfawaz, line 4: As discussed, please create an issue to investigate the need for this change and the issues with the IR parser.

mohammadfawaz (changes requested):

vaivaswatha (dismissed):

tritao:
Updated the PR with fixes to the IR parsing issue instead of changing the test.

Meanwhile I've also found a couple more issues with associated methods while reworking the checks I had earlier on this PR.

The new checking code is more robust and found a couple more issues related to traits, as such I've decided to just split it from this one, and will be sending a different PR later to deal with that with more tests once I handle those extra edge cases.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression/function_application.rs
```diff
@@ -107,59 +120,71 @@ pub(crate) fn instantiate_function_application_simple(
...
+    selector: Option<ContractCallMetadata>,
+    is_constant: IsConstant,
+    self_state_idx: Option<StateIndex>,
+    span: Span,
+) -> TypedExpression {
+    assert_eq!(
+        arguments.len(),
+        function_decl.parameters.len(),
+        "different number of function parameters and arguments"
+    );
```
otrho, line 173: We tend not to have assertions in the compiler as permanent code (sometimes there'll be a `todo!()` or `unimplemented!()` temporarily) but instead prefer to return a `CompileError::Internal` to encourage users to file an issue if something like this occurs.

sway-ir/src/parser.rs
```diff
@@ -389,8 +389,8 @@ mod ir_builder {
                 }
 
             rule struct_const() -> IrAstConstValue
-                = "{" _ flds:(field_or_element_const() ++ comma()) "}" _ {
-                    IrAstConstValue::Struct(flds)
+                = "{" _ flds:(field_or_element_const() ++ comma())? "}" _ {
+                    IrAstConstValue::Struct(flds.unwrap_or_default())
```
otrho, line 393: A better way to do this is with the `**` operator.  It's a bit more explicit and doesn't need the `unwrap()`.

otrho (commented):
Looking good, just a couple of changes needed.

sway-core/tests/sway_to_ir/get_storage_key.sw
otrho: Yep, I don't think this has been done yet?

sway-core/src/semantic_analysis/ast_node/expression/typed_expression/method_application.rs
```diff
@@ -242,6 +245,14 @@ pub(crate) fn type_check_method_application(
                 None
             };
 
+            if method.parameters.is_empty() && !method.is_contract_call {
```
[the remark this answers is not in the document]
mohammadfawaz: I guess this is no longer relevant.

mohammadfawaz (approved):
LGTM! Thanks.

sway-core/tests/sway_to_ir/get_storage_key.sw
mohammadfawaz: I guess this particular change is no longer there but an issue about empty structs in the IR and how they work if they're passed around etc. would still be useful.

otrho (approved):

PR #2099, opened 2022-06-23
update fuelup related instructions

### LINKED ISSUE #2098: Update `fuelup` install instructions
https://github.com/FuelLabs/sway/blob/91b8eee1208fc47efea607e8820ed08ca0028975/docs/src/introduction/installation.md?plain=1#L27

binggh:
Closes #2098 

Instruction for fuelup is a bit outdated in a few ways:

- Remove args from the command since we no longer pass args, and instead pass `toolchain install latest` by default
- Short explanation on `--no-modify-path` and provide the command for it
- Explain that `fuelup toolchain install latest` is ran right after the script is ran.
- Also remove the note for 404 failure since an error message was added here: https://github.com/FuelLabs/fuelup/pull/48/


adlerjohn (approved):

ra0x3 (approved):
🚢

Issue #855, opened 2022-03-01
Introduce intrinsics for VM operations.


otrho:
See FuelLabs/sway#1074

ASM blocks are very hard to optimise for and are generally error prone, as they are untyped and the author often needs to make assumptions about the compiler code generation.  In an effort to minimise the need for ASM blocks in the standard library we can make compiler intrinsics which wrap useful VM opcodes such as `MCP` & `MEQ` or `BAL` & `BURN`, etc.

mohammadfawaz:
Relevant links:
https://gcc.gnu.org/onlinedocs/gcc-4.9.2/gcc/Other-Builtins.html#Other-Builtins
https://en.wikipedia.org/wiki/Intrinsic_function

mohammadfawaz:
We can basically have them all start with `__builtin` like C/C++ to minimize collision and make it clear that they are an advanced feature.

mohammadfawaz:
@vaivaswatha now that https://github.com/FuelLabs/sway/pull/2100 is close to be merged and we have a prototype of how these things can look like, maybe it's a good time to step back a bit and write up an RFC for the full feature and its implementation. This serves multiple purposes:
1. Thinking through the big picture and the scope of this project: what intrinsics do we want to support, What IR instructions do we need to add, Interactions with the different data types, Testing plan, etc.
2. Document all the intrinsics in one place (addresses the last point in https://github.com/FuelLabs/sway/pull/2100#discussion_r908096738)
3. Uncover any overlooked issues in the existing implementation (hopefully there are none!)
4. Lay out the benefits of this project in terms of type checking and future optimizations.

otrho:
We have intrinsics and an RFC now.

PR #2100, opened 2022-06-24
Introduce `__eq` intrinsic

vaivaswatha:
This PR introduces an intrinsic `__eq<T>` (currently T is constrained to integers and bool) that can replace (as shown by the changes in `ops.sw`) assembly definitions of `eq` for these types.

edit: 
1. The intrinsic is lowered to `Instruction::Cmp` in the IR.
2. The existing intrinsic framework is refactored to normalise how we represent all intrinsics.

#855

### DIFF
--- a/sway-lib-std/src/u128.sw
+++ b/sway-lib-std/src/u128.sw
@@ -12,6 +12,10 @@ pub struct U128 {
     lower: u64,
 }
 
+pub enum U128Error {
+    LossOfPrecision: (),
+}
+
 pub trait From {
     /// Function for creating U128 from its u64 components.
     pub fn from(upper: u64, lower: u64) -> Self;
@@ -24,7 +28,7 @@ pub trait From {
 impl From for U128 {
     pub fn from(upper: u64, lower: u64) -> U128 {
         U128 {
-            upper, lower, 
+            upper, lower,
         }
     }
 }
@@ -102,14 +106,15 @@ impl U128 {
         }
     }
 
-    /// Downcast to `u64`. Err if precision would be lost, Ok otherwise.
-    pub fn to_u64(self) -> Result<u64, ()> {
+    /// Safely downcast to `u64` without loss of precision.
+    /// Returns Err if the number > ~u64::max()
+    pub fn as_u64(self) -> Result<u64, U128Error> {
         match self.upper {
             0 => {
                 Result::Ok(self.lower)
             },
             _ => {
-                Result::Err(())
+                Result::Err(U128Error::LossOfPrecision)
             },
         }
     }
@@ -245,7 +250,7 @@ impl core::ops::Subtract for U128 {
         }
 
         U128 {
-            upper, lower, 
+            upper, lower,
         }
     }
 }
--- a/test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u128_test/src/main.sw
+++ b/test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u128_test/src/main.sw
@@ -78,5 +78,30 @@ fn main() -> bool {
     assert(three_left_shift_one.upper == 0);
     assert(three_left_shift_one.lower == 6);
 
+    // test as_u64()
+    let eleven = ~U128::from(0, 11);
+    let unwrapped = eleven.as_u64().unwrap();
+    assert(unwrapped == 11);
+
+    let err_1 = ~U128::from(42, 11).as_u64();
+    assert(match err_1 {
+        Result::Err(U128Error::LossOfPrecision) => {
+            true
+        },
+        _ => {
+            false
+        },
+    });
+
+    let err_1 = ~U128::from(42, 0).as_u64();
+    assert(match err_1 {
+        Result::Err(U128Error::LossOfPrecision) => {
+            true
+        },
+        _ => {
+            false
+        },
+    });
+
     true
 }


Braqzen:
I've not looked at the code but I have noticed that there is an issue linked in the description. 
If this PR is meant to close the issue then you can use one of a few special keywords that github provides in order to automatically close that issue when this PR is merged in (so that you do not have to close it yourself). You can find more info [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

otrho:
The IR already as a `cmp eq` instruction which I'd hope could be used.

https://github.com/FuelLabs/sway/blob/master/sway-ir/src/instruction.rs#L34-L35
https://github.com/FuelLabs/sway/blob/master/sway-ir/src/instruction.rs#L110-L114

If it's not quite functional enough then it should be made so.

Before we get to IR the compiler needs to confirm that the types of each argument are only the builtin types and not structs or other custom types.  And then for reference types like string or b256 they should use different IR, or potentially use a `MEQ`/`memcmp` intrinsic instead.

Creating ASM blocks in IRgen isn't really where we want to go, sorry.

vaivaswatha:
> The IR already as a `cmp eq` instruction which I'd hope could be used.
> If it's not quite functional enough then it should be made so.

Makes sense. I'll look at using this and get back.
 
> Before we get to IR the compiler needs to confirm that the types of each argument 
I'm already doing this, more or less.

>And then for reference types like string or b256 they should use different IR, or potentially use a `MEQ`/`memcmp` intrinsic instead.
Okay. I'm not handling this right now right now though.
 
> Creating ASM blocks in IRgen isn't really where we want to go, sorry.
I understand this. That's why I pointed out the two options we have.

vaivaswatha:
> I've not looked at the code but I have noticed that there is an issue linked in the description. If this PR is meant to close the issue then you can use one of a few special keywords that github provides in order to automatically close that issue when this PR is merged in (so that you do not have to close it yourself). You can find more info [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

Thank you. This PR isn't meant to close the issue, it only implements it partially.

otrho:
> > Creating ASM blocks in IRgen isn't really where we want to go, sorry.
> 
> I understand this. That's why I pointed out the two options we have.

OK, cool.  Hopefully `cmp` will work well enough.  The IR needs a bunch of work still -- it doesn't do pointers properly IMO (or to my satisfaction, I guess) which makes by-reference semantics a bit ambiguous, like comparing larger types.  So if it's not yet manageable it wouldn't be the end of the work to allow `cmp eq` to compare by-ref types for now (and do the correct magic in ASMgen) and we can fix it for pointers later.

sway-core/src/ir_generation/function.rs
```diff
@@ -318,41 +319,65 @@ impl FnCompiler {
...
                 let span_md_idx = MetadataIndex::from_span(context, &span);
                 Ok(self
                     .current_block
                     .ins(context)
                     .get_storage_key(span_md_idx, None))
             }
+            Intrinsic::Eq => {
+                // We index the array with confidence that the type-checker did its job.
+                let lhs = arguments[0].clone();
+                let rhs = arguments[1].clone();
```
sezna, line 371: One thing we _could_ do is codify the arguments into the enum itself, so we don't have a panic risk here.

e.g.

```rust
enum Intrinsic {
  Eq { lhs: TypedExpression, rhs: TypedExpression },
  SizeOf { type_argument: TypeInfo } 
}
```

etc...

I'm just suggesting this, but I wouldn't block this PR on it, as it could be a more major refactor. It would give us type safety here, though.

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
```diff
@@ -1,172 +1,244 @@
...
+        let mut result = vec![];
+        self.type_arguments.iter().fold(&mut result, |accum, targ| {
+            accum.append(&mut targ.type_id.check_for_unresolved_types());
+            accum
+        });
+        self.arguments.iter().fold(&mut result, |accum, texpr| {
+            accum.append(&mut texpr.check_for_unresolved_types());
+            accum
+        });
+        result
```
sezna, line 66: maybe a flat map would be a bit cleaner? No biggie either way

```rust
        self.type_arguments.iter().flat_map(|targ| {
             targ.type_id.check_for_unresolved_types())
         })
        .chain(self.arguments.iter()
          .flat_map(
            UnresolvedTypeCheck::check_for_unresolved_types
          )
         );
```

sezna: (didn't test that code, just writing in github)

sway-lib-core/src/ops.sw
```diff
@@ -210,46 +210,31 @@ pub trait Eq {
...
-        asm(r1: self, r2: other, r3) {
-            eq r3 r1 r2;
-            r3: bool
-        }
+        __eq(self, other)
     }
 }
 
 impl Eq for u8 {
     fn eq(self, other: Self) -> bool {
```
sezna, line 236: nice, looks clean

sezna (commented):
Looks good, just leaving some comments. W.R.T. my statement about using enums for extra type safety on the intrinsic arguments -- I realize that'd make iterating over the arguments harder, so _if_ you want to do that, you may have to add an `.arguments()` iterator method to abstract over the varying types in the enum variants.
[here's what I mean.](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8df6b6ae8eb07d217977c1e022188c07)

Do you think that the extra type safety would be worth it?

sezna:
Ah I see this commit: 
> @vaivaswatha
[Refactor intrinsics to all have arg and type arg vectors](https://github.com/FuelLabs/sway/pull/2100/commits/69173ac5154d5cab66af70c521fa47209f98e13c)

I see you've already had the same thoughts. Do you find it more ergonomic?

vaivaswatha:
> Ah I see this commit:
> 
> > @vaivaswatha
> > [Refactor intrinsics to all have arg and type arg vectors](https://github.com/FuelLabs/sway/pull/2100/commits/69173ac5154d5cab66af70c521fa47209f98e13c)
> 
> I see you've already had the same thoughts. Do you find it more ergonomic?

@otrho suggested that we represent intrinsics in a uniform way, for example during parsing, and later do the checks (number and type of arguments and type arguments) later on. So that motivated me to have this current representation. The idea is that when we have more intrinsics, we can put such information (number of args, type args etc) in a table and validate it with just one piece of code for all intrinsics (I'm sure we'll need exceptions, but in general it should work).

sway-core/src/convert_parse_tree.rs
```diff
@@ -1445,134 +1436,37 @@ fn expr_to_expression(ec: &mut ErrorContext, expr: Expr) -> Result<Expression, E
...
-                        let exp = match <[_; 1]>::try_from(arguments) {
-                            Ok([exp]) => Box::new(exp),
-                            Err(..) => {
-                                let error = ConvertParseTreeError::SizeOfValOneArg { span };
-                                return Err(ec.error(error));
-                            }
-                        };
                         Expression::IntrinsicFunction {
-                            kind: IntrinsicFunctionKind::SizeOfVal { exp },
+                            #[allow(clippy::unnecessary_unwrap)]
```
otrho: This is weird, why is clippy complaining?  `intrinsic` is an `Option` yeah?

test/src/e2e_vm_tests/test_programs/should_fail/intrinsics/test.toml
```diff
@@ -0,0 +1,4 @@
+category = "fail"
+
+# check: $()Unsupported argument type to eq.
```
otrho, line 3: The error should probably say `__eq` to avoid confusion.

test/src/e2e_vm_tests/test_programs/should_pass/language/intrinsics/src/main.sw
```diff
@@ -0,0 +1,14 @@
...
+fn main() -> u64 {
+
+  assert(true == true);
+  assert(true != false);
+
+  assert(__eq(1, 22) == (1 == 22));
+  assert(__eq(1, 1) == (1 == 1));
+
+  2
+}
```
otrho, line 14: Can you be a bit more thorough here please?  We need more testing in general I think:
- This is only testing `__eq` so the test name should probably be `should_pass/language/eq_intrinsic`.
- Confirm that we can pass all the built-in types to `__eq` without type error.
- Confirm (in a `should_fail`) that you can't pass custom types (structs, enums) or all by-ref types (tuples, strings, B256, etc).
- Now that intrinsics are generalised, does the parser always parse `__xxx` calls as intrinsic calls?  What happens if you try a `___xxx()` call for an unknown intrinsic?  I think we already check for this so there might be an existing E2E test, but please check that we get a coherent error message.
- We should have an IR test (under `sway-core/tests`), just a small one confirming `__eq` generates `cmp eq` and not a `call`.  

We should also document these somewhere as we add them.  I'm not sure if they belong in the book... maybe not as they're only meant to be used by library devs.  @sezna @mohammadfawaz Ideas?

otrho (changes requested):
I really like the generalisation of intrinsics here, good stuff.  We might need to refine the data driven approach to type checking, control flow/dead code analysis for them, I think there are some (OK) assumptions being made now which may not always hold in the future.

But it needs a bit more testing please.

test/src/e2e_vm_tests/test_programs/should_pass/language/intrinsics/src/main.sw
vaivaswatha: I've addressed all these comments, except for

> What happens if you try a ___xxx() call for an unknown intrinsic? 

There already are tests that address this, such as `double_underscore_fn`.

>We should also document these somewhere as we add them.

I'll wait for @sezna @mohammadfawaz to respond.

otrho (dismissed):

otrho (approved):

test/src/e2e_vm_tests/test_programs/should_pass/language/intrinsics/src/main.sw
mohammadfawaz: Ref: https://github.com/FuelLabs/sway/issues/855#issuecomment-1170054431 as a starting point.

More generally, I've suggested in the past the idea of introducing a "Sway Reference" book that is similar to the ["Rust Reference"](https://doc.rust-lang.org/reference/) which would include the language spec (we don't have `sway.pest` anymore so we don't have that in any form) as well as all the language features that are not necessarily useful for end users but only for library authors. This is a big undertaking but probably worth investing the time into long term. It may even include some compiler internals such as how monomorphization works for example.

sezna (approved):
If it is good with @otrho, it is good with me. I think a good next item would be a more wholistic intrinsic design doc: https://github.com/FuelLabs/sway/issues/855#issuecomment-1170054431

Issue #1798, opened 2022-06-01
Add math & bitwise ops to `U256` type


nfurfaro:
Probably not critical at this point, but it may be something we want to have eventually. 
Possible uses:
- handle _**very**_ large integers (ie: overflowing `U128` maths)
- a convenient way to pack 4 `u64`'s together as a single logical entity. Currently, users are trying to pack `u64`s into a `b256` which is not really what a `b256` is for semantically. This may in part be due to a misunderstanding of storage slots and word size in the FuelVM; Devs familiar with Solidity are often familiar with packing as much data as possible into a `uint256` (a single storage slot). 

Note: The basic `U256` is implemented in #2103.
TODO:
- [ ] add math operations to `U256`
- [ ] add bitwise operations to `U256`

mohammadfawaz:
Seems that this is now implemented in https://github.com/FuelLabs/sway/pull/2445

PR #2103, opened 2022-06-24
Add the U256 type

nfurfaro:
This adds the start of a `U256` type to the stdlib. 
The current implementation _should_ be sufficient to unblock the first use-case ( the L1 <=> L2 bridge messaging system).
- modelled on the `U128` type (composed of inner `u64`s
- implements `new()`, `from()`, `into()`, `Eq`, `min()`, `max()`, `bits()` and `as_u64()`

Math operations and bit-shifting is left to be implemented in subsequent PRs as needed.

ref #1798


otrho (approved):
If #1798 is going to be closed do you have replacement issues for finishing it?

nfurfaro:
@otrho good catch. 
I will NOT close the issue, only ref it and will add notes to the issue for the remaining work on this type.

sway-lib-std/src/u256.sw
```diff
@@ -0,0 +1,85 @@
+library u256;
+
+use core::num::*;
+use ::result::*;
+
+/// The 256-bit unsigned integer type.
+/// Represented as four u64-bit components: `(a, b, c, d)`, where `value = (a << 192) + (b << 128) + (c << 64) + d`.
```
adlerjohn, line 7:
```suggestion
/// Represented as four 64-bit components: `(a, b, c, d)`, where `value = (a << 192) + (b << 128) + (c << 64) + d`.
```

adlerjohn (approved):

adlerjohn (approved):

mohammadfawaz (approved):

sway-lib-std/src/u256.sw
```diff
@@ -0,0 +1,85 @@
+library u256;
+
+use core::num::*;
```
Braqzen, line 3: Imports should always be explicit even if the compiler only brings into scope the code that is used. This is for the sake of code quality for the developers.

sway-lib-std/src/u256.sw
```diff
@@ -0,0 +1,85 @@
...
+        U256 {
+            a: 0,
+            b: 0,
+            c: 0,
+            d: 0,
+        }
+    }
+
+    /// Downcast to `u64`. Err if precision would be lost, Ok otherwise.
+    pub fn to_u64(self) -> Result<u64, ()> {
```
Braqzen, line 52: I dislike seeing the unit type being returned as an error. It's not descriptive in what has occured. This should really return some enum or something that has a descriptive name because this allows the dev to immediately see the problem via the name rather than having to go and read the source code / look up the docs.

Braqzen, line 52: Do we want to also have `to_uXX` methods e.g. `to_u128`?

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u256_test/src/main.sw
```diff
@@ -0,0 +1,63 @@
...
+
+use core::num::*;
+use std::{
+    assert::assert,
+    result::Result,
+    u256::*
+};
+
+fn main() -> bool {
+    let new = ~U256::new();
```
Braqzen, line 11: No test to see if `new` returns as expected

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u256_test/src/main.sw
```diff
@@ -0,0 +1,63 @@
...
+    assert(0 == 0);
+
+    let a = 11;
+    let b = 42;
+    let c = 101;
+    let d = 69;
+    let x = ~U256::from(a, b, c, d);
+    let y = ~U256::from(a, b, c, d);
+
+    assert(x.a == a);
```
Braqzen, line 25: This only checks that the two cases are equal. It does not check the opposite. It could be that no matter what you add into the associated call it always returns the same values.

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u256_test/src/main.sw
```diff
@@ -0,0 +1,63 @@
...
+
+    let(e, f, g, h) = x.into();
+    assert(e == a);
+    assert(f == b);
+    assert(g == c);
+    assert(h == d);
+
+    assert(x == y);
+
+    let max = ~U256::max();
```
Braqzen, line 38: This max tests only 1 condition rather than checking each field namely if `a` is too large then if `a` is 0 but `b` is too large etc. Each case should be tested individually to prove that all cases pass

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u256_test/src/main.sw
```diff
@@ -0,0 +1,63 @@
...
+    });
+
+    let(one, two, three, four) = max.into();
+    assert(one == ~u64::max());
+    assert(two == ~u64::max());
+    assert(three == ~u64::max());
+    assert(four == ~u64::max());
+
+    let eleven = ~U256::from(0, 0, 0, 11);
+    let ok: Result<u64, ()> = eleven.to_u64();
```
Braqzen, line 56: Using the latest release you should not need to add a type to unwrap nor use 2 lines. You should be able to chain the calls

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u256_test/src/main.sw
```diff
@@ -0,0 +1,63 @@
...
+    let unwrapped = ok.unwrap();
+    assert(unwrapped == 11);
+
+    assert(~U256::bits() == 256u32);
+
+    true
```
Braqzen, line 62: I don't see a test for `min()`

sway-lib-std/src/u256.sw
```diff
@@ -0,0 +1,85 @@
...
+        U256 {
+            a: ~u64::max(),
+            b: ~u64::max(),
+            c: ~u64::max(),
+            d: ~u64::max(),
+        }
+    }
+
+    /// The size of this type in bits.
+    pub fn bits() -> u32 {
```
Braqzen, line 82: Why does this return a `u32` instead of a `u8`?

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u256_test/test.toml
```diff
@@ -0,0 +1,3 @@
+category = "run"
+expected_result = { action = "return", value = 1 }
```
Braqzen, line 2: I may be misremembering but why is a bool (true has the value of 1) being treated as a 1 here? Should this be a bool or is this fine? I guess it is fine

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u256_test/src/main.sw
```diff
@@ -0,0 +1,63 @@
+script;
+
+use core::num::*;
+use std::{
+    assert::assert,
+    result::Result,
+    u256::*
```
Braqzen, line 7: Ree @ `*`

Braqzen (changes requested):
1. Updating to the latest release allows for cleaner code
2. Ree @ `*` imports
3. main file looks fine (generally), minor nit and a couple questions
4. testing does not meet my standards because it misses 1 function entirely and misses coverage across all other boundary conditions

sway-lib-std/src/u256.sw
nfurfaro: @Braqzen any thoughts on the syntax for this import: `use core::num::*;`
where I only need the functions defined on the `u64` type.

sway-lib-std/src/u256.sw
nfurfaro: If needed, yes.

sway-lib-std/src/u256.sw
nfurfaro: It's a rust thing. Following their example as there's no good reason to deviate.

adlerjohn: In addition to following Rust, 256 also doesn't fit in a `u8`

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u256_test/test.toml
adlerjohn: The output here is untyped. True is 1 and false is 0 in the VM

sway-lib-std/src/u256.sw
Braqzen: Unfortunately I do not know of a way to directly import that content without using `*` therefore that should stay as it is unless there is another way to be explicit

Braqzen (commented):
I'm almost satisfied with the rigour of the tests.
The `to_u64()` tests contain 2 out of 4 test cases.

1. We return a non-error (a, b & c are all 0)
2. We return an error when `a != 0`

We should also test individual cases where

1. `a == 0` & `b != 0`
2. `a == 0` & `b == 0` & `c != 0` (all errors)

For completeness.

I won't block this going through but for the reason stated above I will not approve until those 2 tests are added in. Aside from that it looks good to me!

sway-lib-std/src/u256.sw
```diff
@@ -0,0 +1,89 @@
...
+        U256 {
+            a: 0,
+            b: 0,
+            c: 0,
+            d: 0,
+        }
+    }
+
+    /// Downcast to `u64`. Err if precision would be lost, Ok otherwise.
+    pub fn to_u64(self) -> Result<u64, U256Error> {
```
adlerjohn, line 56: Do we want this to be like https://docs.rs/primitive-types/0.6.1/primitive_types/struct.U256.html#method.as_u64? Same as: https://docs.rs/bigint/4.2.0/bigint/uint/struct.U256.html#method.as_u64

Braqzen (approved):

Braqzen:
Note that https://github.com/FuelLabs/sway/blob/e59cac351ce242fa1933261404b42983083a3eb8/test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u256_test/src/main.sw#L64

and

https://github.com/FuelLabs/sway/blob/e59cac351ce242fa1933261404b42983083a3eb8/test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u256_test/src/main.sw#L74

are the same test.
You can remove one if you want.

sway-lib-std/src/u256.sw
nfurfaro: Yes, it makes sense to rename `to_u64()` => `as_u64()` here.
This sets us up for implementing more methods in line with what rust does and following the same naming convention.

sway-lib-std/src/lib.sw
```diff
@@ -12,7 +12,7 @@ dep constants;
 dep contract_id;
 dep context;
 dep hash;
-dep r#storage;
+dep storage;
```
Braqzen, line 15: Should the storage be changed in this issue? I think if that is a problem then it should be its own issue which is merged into master and then master is merged into this. Why was it changed in the first place?

nfurfaro: nope, I actually just reverted this change as it's not up to this PR to touch it.
Not sure why it was changed, I certainly didn't do it intentionally, so maybe fmt did it?

Braqzen (approved):

sway-lib-std/src/lib.sw
adlerjohn: CI fails because storage is a keyword and forbidden as a library name. Revert this change to fix.

sway-lib-std/src/storage.sw
```diff
@@ -1,4 +1,4 @@
-library r#storage;
+library storage;
```
adlerjohn: Revert

PR #2110, opened 2022-06-25
Adds `Generics` handling to `sway-fmt-v2`

### LINKED ISSUE #2073: Update `Punctuated` to handle the `final_value_opt`
Currently, if there is no trailing `,` after the last enum variant, the variant is completely removed from the formatted output.

Example:
```
contract;

enum Color {
 Red : (),
 Blue : ()
}
```

is formatted to 

```
contract;

enum Color {
 Red : (),
}
```

### LINKED ISSUE #2108: Add formatting for `Punctuated` pairs
Currently we're neglecting the `Option<Box<T>>` field on struct `Punctuated`, it may be better to have specific formatting for `Punctuated` itself instead of only formatting the `value_separator_pairs` like we currently do in the `ItemKind` formatting.

This is the current state of `Punctuated` formatting, inside of each `ItemKind` formatting function:
```Rust
        let params = self.parameters.clone().into_inner().value_separator_pairs;
        let mut buf = params
            .iter()
            .map(|param| format!("{}{}", param.0.as_str(), param.1.span().as_str()))
            .collect::<Vec<String>>()
            .join(" ");
```

### LINKED ISSUE #2109: Add `Generics` handling to `sway-fmt-v2`

eureka-cpu:
Handles formatting of generic arguments in-line, as well as `Punctuated` pairs. If need be I can also split this into two PR's, they are just dependent on one another so I have them here during the draft for simplicity's sake.

- [x] #2108 (interdependent)
- [x] if #2111 gets in first this PR will need some refactoring
- [x] handle `final_value_opt`

Closes #2108 
Closes #2109 
Closes #2073


mitchmindtree (commented):
Looking good! Do we have any tests for generics yet?

eureka-cpu:
> Looking good! Do we have any tests for generics yet?

Yep! Just did some editing to fix other spots where we could've used `write`, and then updated one of the `struct` tests with some generics. Again, there is a known problem where if the user doesn't add a trailing comma it will act strangly. @kayagokalp and I are tracking that issue here: #2073

eureka-cpu:
#2134 is a child of this PR

mitchmindtree (approved):

JoshuaBatty (approved):

PR #2111, opened 2022-06-25
internal: Reduce amount of String::new() in sway-fmt-v2

### LINKED ISSUE #2107: Reduce amount of `String::new()` in `sway-fmt-v2`
I think our general approach to creating new allocations can be changed slightly.

The biggest change I have in mind would be to change the `Format` trait itself to instead take in the string originally created in `Formatter::format()`. This way we aren't creating new strings for every `Format` function, and instead just amend the original.
The trait would now look like:
```Rust
pub trait Format {
    fn format(&self, line: &mut String, formatter: &mut Formatter) -> Result<(), std::fmt::Error>;
}
``` 

The second and smaller change, would be to organize how each `line` is formatted with `write!()` and `writeln!()`. So instead of something like:
```Rust
line.push_str("...");
line.push(' ');
// or
line.push_str(&format!("{} ", ...span().as_str());
```
we would just have:
```Rust
write!(line, "{} ", ...span().as_str());
```
This will save us several new string allocations and return possible errors along the way.

XFFXFF:
close #2107   

changed `format` method   
from 
``` rust
fn format(&self, formatter: &mut Formatter) -> FormattedCode;
```
to
``` rust
fn format(&self, formatted_code: &mut FormattedCode, formatter: &mut Formatter,) -> Result<(), FormatterError>;
```
This may reduce the number of memory allocations.  

~~This pr does not fully resolve the above issue, still using `push_str` instead of `write!`~~

### DIFF
--- a/sway-fmt-v2/src/fmt.rs
+++ b/sway-fmt-v2/src/fmt.rs
@@ -41,6 +41,7 @@ impl Formatter {
         build_config: Option<&BuildConfig>,
     ) -> Result<FormattedCode, FormatterError> {
         let path = build_config.map(|build_config| build_config.canonical_root_module());
+        let src_len = src.len();
         let module = sway_parse::parse_file(src, path)?;
         // Get parsed items
         let items = module.items;
@@ -49,7 +50,9 @@ impl Formatter {
 
         // Formatted code will be pushed here with raw newline stlye.
         // Which means newlines are not converted into system-specific versions by apply_newline_style
-        let mut raw_formatted_code = String::new();
+        // Use the length of src as a hint of the memory size needed for raw_formatted_code, 
+        // which will reduce the number of reallocations
+        let mut raw_formatted_code = String::with_capacity(src_len);
 
         // Insert program type to the formatted code.
         insert_program_type(&mut raw_formatted_code, program_type);
--- a/sway-fmt-v2/src/fmt.rs
+++ b/sway-fmt-v2/src/fmt.rs
@@ -50,14 +50,14 @@ impl Formatter {
 
         // Formatted code will be pushed here with raw newline stlye.
         // Which means newlines are not converted into system-specific versions by apply_newline_style
-        // Use the length of src as a hint of the memory size needed for raw_formatted_code, 
+        // Use the length of src as a hint of the memory size needed for raw_formatted_code,
         // which will reduce the number of reallocations
         let mut raw_formatted_code = String::with_capacity(src_len);
 
         // Insert program type to the formatted code.
         insert_program_type(&mut raw_formatted_code, program_type);
-        // Insert parsed & formatted items into the formatted code.
 
+        // Insert parsed & formatted items into the formatted code.
         let mut iter = items.iter().peekable();
         while let Some(item) = iter.next() {
             // format Annotated<ItemKind>
--- a/sway-fmt-v2/src/fmt.rs
+++ b/sway-fmt-v2/src/fmt.rs
@@ -3,7 +3,6 @@ use crate::utils::{
 };
 use std::{path::Path, sync::Arc};
 use sway_core::BuildConfig;
-use sway_parse::attribute::Annotated;
 
 pub use crate::{
     config::manifest::Config,
@@ -58,7 +57,8 @@ impl Formatter {
 
         let mut iter = items.iter().peekable();
         while let Some(item) = iter.next() {
-            Annotated::format(item, &mut raw_formatted_code, self)?;
+            // format Annotated<ItemKind>
+            item.format(&mut raw_formatted_code, self)?;
             if iter.peek().is_some() {
                 raw_formatted_code.push('\n');
             }
--- a/sway-fmt-v2/src/fmt.rs
+++ b/sway-fmt-v2/src/fmt.rs
@@ -19,7 +19,11 @@ pub struct Formatter {
 pub type FormattedCode = String;
 
 pub trait Format {
-    fn format(&self, formatter: &mut Formatter) -> FormattedCode;
+    fn format(
+        &self,
+        formatted_code: &mut FormattedCode,
+        formatter: &mut Formatter,
+    ) -> Result<(), FormatterError>;
 }
 
 impl Formatter {
@@ -51,14 +55,15 @@ impl Formatter {
         // Insert program type to the formatted code.
         insert_program_type(&mut raw_formatted_code, program_type);
         // Insert parsed & formatted items into the formatted code.
-        raw_formatted_code += &items
-            .into_iter()
-            .map(|item| -> Result<String, FormatterError> {
-                // format Annotated<ItemKind>
-                Ok(Annotated::format(&item, self))
-            })
-            .collect::<Result<Vec<String>, _>>()?
-            .join("\n");
+
+        let mut iter = items.iter().peekable();
+        while let Some(item) = iter.next() {
+            Annotated::format(item, &mut raw_formatted_code, self)?;
+            if iter.peek().is_some() {
+                raw_formatted_code.push('\n');
+            }
+        }
+
         let mut formatted_code = String::from(&raw_formatted_code);
         apply_newline_style(
             // The user's setting for `NewlineStyle`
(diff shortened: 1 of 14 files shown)


sway-fmt-v2/src/fmt.rs
```diff
@@ -51,14 +55,15 @@ impl Formatter {
...
-            .map(|item| -> Result<String, FormatterError> {
-                // format Annotated<ItemKind>
-                Ok(Annotated::format(&item, self))
-            })
-            .collect::<Result<Vec<String>, _>>()?
-            .join("\n");
+
+        let mut iter = items.iter().peekable();
+        while let Some(item) = iter.next() {
+            Annotated::format(item, &mut raw_formatted_code, self)?;
```
eureka-cpu, line 61: Would you mind retaining the comment here?

sway-fmt-v2/src/items/item_abi.rs
```diff
@@ -2,80 +2,73 @@ use crate::{
...
-                        AttributeDecl::format(&attr, &mut buf, formatter)
-                    }
+        let mut abi_items_iter = self.abi_items.get().iter().peekable();
+        while let Some(item) = abi_items_iter.next() {
+            let attribute_list = item.0.attribute_list.clone();
+            // add indent + format attribute if it exists
+            if !attribute_list.is_empty() {
+                formatted_code.push_str(&formatter.shape.indent.to_string(formatter));
+                for attr in attribute_list {
+                    AttributeDecl::format(&attr, formatted_code, formatter)
```
eureka-cpu, line 33: Here as well, it can be `attr.format(formatted_code, formatter)`

sway-fmt-v2/src/items/item_abi.rs
```diff
@@ -2,80 +2,73 @@ use crate::{
...
-                            AttributeDecl::format(&attr, &mut buf, formatter)
-                        }
+            let mut iter = abi_defs.get().iter().peekable();
+            while let Some(item) = iter.next() {
+                let attribute_list = item.attribute_list.clone();
+                // add indent + format attribute if it exists
+                if !attribute_list.is_empty() {
+                    formatted_code.push_str(&formatter.shape.indent.to_string(formatter));
+                    for attr in attribute_list {
+                        AttributeDecl::format(&attr, formatted_code, formatter)
```
eureka-cpu, line 58: here as well

sway-fmt-v2/src/items/item_enum.rs
```diff
@@ -24,7 +28,7 @@ impl Format for ItemEnum {
 
         // Add name of the enum.
         formatted_code.push_str(self.name.as_str());
-        ItemEnum::open_curly_brace(&mut formatted_code, formatter);
+        ItemEnum::open_curly_brace(formatted_code, formatter);
```
eureka-cpu, line 31: This can be `Self`

sway-fmt-v2/src/items/item_enum.rs
```diff
@@ -71,8 +75,8 @@ impl Format for ItemEnum {
             // from the config we may understand next enum variant should be in the same line instead.
             formatted_code.push('\n');
         }
-        ItemEnum::close_curly_brace(&mut formatted_code, formatter);
-        formatted_code
+        ItemEnum::close_curly_brace(formatted_code, formatter);
```
eureka-cpu, line 78: Here as well

sway-fmt-v2/src/utils/attribute.rs
```diff
@@ -9,15 +12,18 @@ use sway_types::Spanned;
...
+        &self,
+        formatted_code: &mut FormattedCode,
+        formatter: &mut Formatter,
+    ) -> Result<(), FormatterError> {
         let attributes = &self.attribute_list;
-        let mut formatted_code = String::new();
 
         for attr in attributes {
-            AttributeDecl::format(attr, &mut formatted_code, formatter);
+            AttributeDecl::format(attr, formatted_code, formatter);
```
eureka-cpu, line 23: this can also be as previously mentioned

eureka-cpu (changes requested):
As a general by-product of adding `write!()` in the future, you may have to add `std::fmt::Error` to `FormatterError`

This is starting to look better, thanks so much for opening this ticket!

XFFXFF:
Thanks for the quick reply! I made some changes based on your comments, please review again.

eureka-cpu:
> Thanks for the quick reply! I made some changes based on your comments, please review again.

Looking better! Just needs `write!()` and update `FormatterError` with `std::fmt::Error` and I'll happily approve. Also, we will need one other reviewer to have PR's merged on this repo. I'll go ahead and request from @FuelLabs/tooling

eureka-cpu (approved):
Approving to dismiss my previously requested changes

sway-fmt-v2/src/items/item_enum.rs
```diff
@@ -2,15 +2,19 @@ use crate::{
...
 use sway_types::Spanned;
 
 impl Format for ItemEnum {
-    fn format(&self, formatter: &mut Formatter) -> FormattedCode {
+    fn format(
+        &self,
+        formatted_code: &mut FormattedCode,
+        formatter: &mut Formatter,
+    ) -> Result<(), FormatterError> {
         // TODO: creating this formatted_code with String::new() will likely cause lots of
```
kayagokalp, line 16:
```suggestion

```

sway-fmt-v2/src/items/item_enum.rs
```diff
@@ -2,15 +2,19 @@ use crate::{
...
         // reallocations maybe we can explore how we can do this, starting with with_capacity may help.
```
kayagokalp, line 17:
```suggestion

```

kayagokalp (changes requested):
Thanks for the contribution! This looks great, just a single nit. Maybe we can get rid of the outdated comment so that someone else does not get confused by that 👍

sway-fmt-v2/src/fmt.rs
```diff
@@ -46,19 +50,23 @@ impl Formatter {
 
         // Formatted code will be pushed here with raw newline stlye.
         // Which means newlines are not converted into system-specific versions by apply_newline_style
-        let mut raw_formatted_code = String::new();
+        // Use the length of src as a hint of the memory size needed for raw_formatted_code, 
+        // which will reduce the number of reallocations
+        let mut raw_formatted_code = String::with_capacity(src_len);
```
XFFXFF, line 55: Use the length of src as a hint of the memory size needed for `raw_formatted_code`, which will reduce the number of reallocations.

XFFXFF:
I just introduced `write!` to avoid using something like
```rust
line.push_str(&format!("{} ", xxx);
```
Now #2107  can be closed, if this pr lands. @eureka-cpu @kayagokalp Please review again

sway-fmt-v2/src/utils/attribute.rs
```diff
@@ -58,26 +65,43 @@ impl FormatDecl for AttributeDecl {
...
 }
 
 impl Parenthesis for AttributeDecl {
-    fn open_parenthesis(line: &mut String, _formatter: &mut Formatter) {
-        line.push(Delimiter::Parenthesis.as_open_char())
+    fn open_parenthesis(
+        line: &mut String,
+        _formatter: &mut Formatter,
+    ) -> Result<(), FormatterError> {
+        line.push(Delimiter::Parenthesis.as_open_char());
```
eureka-cpu, line 97: Would you mind making `Parenthesis` use `writeln!()` as well?

sway-fmt-v2/src/utils/attribute.rs
```diff
@@ -58,26 +65,43 @@ impl FormatDecl for AttributeDecl {
...
+        Ok(())
     }
-    fn close_parenthesis(line: &mut String, _formatter: &mut Formatter) {
-        line.push(Delimiter::Parenthesis.as_close_char())
+    fn close_parenthesis(
+        line: &mut String,
+        _formatter: &mut Formatter,
+    ) -> Result<(), FormatterError> {
+        line.push(Delimiter::Parenthesis.as_close_char());
```
eureka-cpu, line 104: and here

sway-fmt-v2/src/fmt.rs
```diff
@@ -46,19 +50,23 @@ impl Formatter {
 
         // Formatted code will be pushed here with raw newline stlye.
         // Which means newlines are not converted into system-specific versions by apply_newline_style
-        let mut raw_formatted_code = String::new();
+        // Use the length of src as a hint of the memory size needed for raw_formatted_code,
+        // which will reduce the number of reallocations
+        let mut raw_formatted_code = String::with_capacity(src_len);
```
eureka-cpu, line 55: I don't think this will be necessary since our formatting will do more than just simple whitespace augmentation. We may end up reformatting things like long `generics` with a `where` clause which could cause collisions in the future if we give `raw_formatted_code` an initial capacity.

eureka-cpu (changes requested):
Okay looking great just a few last things and I'll defer final approval to @FuelLabs/tooling

sway-fmt-v2/src/fmt.rs
XFFXFF: Hmm, I don't understand what the "collision" means here, would you mind explain it more?

eureka-cpu: My mistake, I was really tired yesterday and somehow thought `with_capacity` limited the capacity which isn't true at all 🤦🏻

eureka-cpu (approved):
Thanks again for doing this!

kayagokalp (approved):
LGTM, great work! thanks for the contribution 🎉

PR #2112, opened 2022-06-25
`break` and `continue`

mohammadfawaz:
Adds `break` and `continue` to loops:
* Introduce `break` and `continue` to the parser.
* Treat them as `declarations`.
* In IR gen, when compiling a loop, remember what blocks we need to jump to in case of a `break` or a `continue`. Also make sure not to break the nested loops case which I'm testing in the E2E test I added.

* Few notes on control flow analysis:
  * DCA is simple... just don't hand over any leaves and the code after `break` and `continue` will show up as dead/unreachable.
  * Analyze return paths is a bit more complicated. We now have to dig into while loops: keep track of what nodes to jump to in case of a `break` or a `continue` and use those to handle `break` and `continue` nodes later. Also, error out if those nodes are not set (that means that the `break` or the `continue` is used outside a while loop which is illegal). Finally, we also have to keep track of the return nodes that are inside the while loop so "bubble" those up and connect them to the exit node.

* Not handling `break` and `continue` in `sway-lsp` Created a separate issue for that here: https://github.com/FuelLabs/sway/issues/2117
* Not handling `break` and `continue` in `sway-fmt-v2` Created a separate issue for that here: https://github.com/FuelLabs/sway/issues/2118

Closes https://github.com/FuelLabs/sway/issues/587
Closes https://github.com/FuelLabs/sway/issues/2116


sway-core/tests/sway_to_ir/break.sw
```diff
@@ -0,0 +1,13 @@
+script;
+
+fn main() {
```
Braqzen, line 3: Should this return a value or is exiting at all acceptable? I'm wondering about this potentially being stuck in an infinite loop and however the test is run it might not detect it except for when a dev notices that this has been running for forever.

That being said, should this have counters to exit intentionally with a failing bool if the limit is reached? This could be another indicator that there is a bug.

Similar comments for continue.sw

sway-fmt-v2/src/utils/item.rs
```diff
@@ -14,6 +14,8 @@ impl Format for Item {
             Abi(item_abi) => item_abi.format(formatter),
             Const(item_const) => item_const.format(formatter),
             Storage(item_storage) => item_storage.format(formatter),
+            Break(_item_break) => unimplemented!(),
```
Braqzen, line 17: I don't recall if this macro panics. Iirc then there is a newer (shorter) `todo!()` which will panic when reached. I think it was meant to replace this more verbose macro.

Anyway, does this mean that is someone uses these keywords and attempts to format their code then it will panic? That is not ideal if true. I'd rather get a message  informing me of this rather than crashing - unless this only produces a warning which is fine

sway-ir/src/block.rs
```diff
@@ -154,6 +154,15 @@ impl Block {
             .map_or(false, |i| matches!(i, Instruction::Ret { .. }))
     }
 
+    pub fn is_terminated_by_a_branch(&self, context: &Context) -> bool {
+        self.get_term_inst(context).map_or(false, |i| {
+            matches!(
+                i,
```
Braqzen, line 160: What is `i`?

sway-core/src/error.rs
```diff
@@ -995,6 +995,10 @@ pub enum CompileError {
     NonConstantDeclValue { span: Span },
     #[error("Declaring storage in a {program_kind} is not allowed.")]
     StorageDeclarationInNonContract { program_kind: String, span: Span },
+    #[error("\"break\" used outside of a loop")]
```
Braqzen, line 998: I've made a comment below in the formatter regarding the `unimplemented!()` macro. Is this a warning / error that would be displayed in that section or is this error used somewhere else?

test/src/e2e_vm_tests/test_programs/should_pass/language/break_and_continue/src/main.sw
```diff
@@ -0,0 +1,70 @@
+script;
+
+const J = 10;
+const K = 20;
+const I = 30;
+const N = 100;
+
+fn break_test() -> u64 {
+    let m = 5;
```
Braqzen, line 9: I don't understand the point of all of these variables and calculations. Could this be made simpler? Why is the output so large?

sway-parse/src/item/item_control_flow.rs
```diff
@@ -0,0 +1,51 @@
+use crate::priv_prelude::*;
```
Braqzen, line 1: Importing explicitly is a good practice. Is there an issue which prevents this explicit import or is this just lazy coding?

Braqzen (commented):
Just some questions. Trying to look around other parts of the code base

sway-core/tests/sway_to_ir/break.sw
adlerjohn: Test use a metered FuelVM, so it will error with out of gas effectively instantly in real time.

mohammadfawaz: This is simply a unit test that doesn't actually run. It's just used to verify the IR generated, so the simpler I can make it the better, hence the weird structure of loops here.

sway-fmt-v2/src/utils/item.rs
mohammadfawaz: I'm not sure which macro is preferred honestly, but I opened https://github.com/FuelLabs/sway/issues/2118 and I'm hoping we don't release `sway-fmt-v2` without this fixed. cc @eureka-cpu.

sway-ir/src/block.rs
mohammadfawaz: `i` is for instruction.. that's pretty common in IR land. Either way, simply copied the function above (`is_terminated_by_ret`) here and kept it the same for consistency.

sway-core/src/error.rs
mohammadfawaz: This is a regular compiler error. This is the desirable behavior when `break` is used outside a loop. This is exactly what Rust tells you when you do that.

sway-parse/src/item/item_control_flow.rs
mohammadfawaz: Another instance of consistency between all the files in this directory. I didn't really think about this much. Might be worth cleaning up at some point.

test/src/e2e_vm_tests/test_programs/should_pass/language/break_and_continue/src/main.sw
mohammadfawaz: Good question. I really tried to go wild here to stress test `break` and `continue` in a complex setting with many nested loops. I can probably simplify the math a bit but I was worried about losing cover my mistake by choosing the wrong numbers. Crafting a better test is certainly possible here though. I can do that tomorrow.

sway-parse/src/item/item_control_flow.rs
otrho: The point of `priv_prelude` is you just import it all at once.  Pretty much everywhere.  It's a private prelude.

sway-fmt-v2/src/utils/item.rs
otrho: It doesn't make that much difference, they both panic.  `todo!()` is meant to convey the meaning that this will definitely be done soon, but `unimplemented!()` can mean we'll never do this...

sway-core/tests/sway_to_ir/break.ir
```diff
@@ -0,0 +1,43 @@
...
+        while:
+        v0 = const bool true, !2
+        cbr v0, while_body, end_while
+
+        while_body:
+        br while0
+
+        end_while:
+        v1 = const unit ()
+        ret () v1
```
otrho, line 15: This is out of order and a bit confusing.  I've tried fairly hard to make most jumps go downwards and for `ret` to be at the end of the function.  I still need to review the IRgen so maybe it's a bit hard to predict ahead of time what blocks are needed when there could be many breaks or continues pending... maybe we should reorder the blocks after finalising a function and/or create a mechanism for inserting new blocks at a cursor rather than always appending.

test/src/e2e_vm_tests/test_programs/should_pass/language/break_and_continue/src/main.sw
otrho: I like the complexity but I don't want to be the one who has to debug it if it starts failing. :smile:

sway-core/src/ir_generation/function.rs
```diff
@@ -390,29 +435,34 @@ impl FnCompiler {
...
-            ),
-            LazyOp::Or => cond_builder.conditional_branch(
-                lhs_val,
-                final_block,
-                rhs_block,
-                Some(lhs_val),
-                span_md_idx,
-            ),
-        };
+        if !self.current_block.is_terminated_by_a_branch(context) {
```
otrho, line 438: This needs to be done to avoid a double branch due to a break/continue, yeah?  And the verifier would complain in that case.  It _does_ feel like this sort of thing would be better handled by DCE... I'm starting to wonder if the block verifier is too strict... maybe we only need to call `verify` once, after we've finished opts and are ready to hand over to ASMgen... thoughts?

otrho: I feel like IRgen shouldn't have to stress out over introducing dead code.

sway-core/src/ir_generation/function.rs
```diff
@@ -869,35 +923,59 @@ impl FnCompiler {
...
         // Create the final block after we're finished with the body.
         let final_block = self
             .function
             .create_block(context, Some("end_while".into()));
 
+        // Keep track of the previous blocks we have to jump to in case of a break or a continue.
+        // This should be `None` if we're not in a loop already or the previous break or continue
+        // destinations for the outer loop that contains the current loop.
+        let prev_block_to_break_to = self.block_to_break_to;
+        let prev_block_to_continue_to = self.block_to_continue_to;
```
otrho, line 947: This feels a bit clumsy (and I balked a bit when I saw `FnCompiler::block_to_break_to` initially) but I can't think of an obviously better technique.  If it wasn't for the `lexical_map` I'd say create a copy of `self` for the recursion, but that doesn't help that much anyway.  Maybe some preprocessing which tags the break or continue with their destinations before compilation, but that's adding unneeded complexity too.

So, nevermind, this is good. :slightly_smiling_face:

otrho (dismissed):
I'll approve though I feel like it has some rough edges we should smooth over in time. :)

sezna:
Thanks for creating all the corresponding issues!

sway-core/src/control_flow_analysis/dead_code_analysis.rs
```diff
@@ -414,7 +414,9 @@ fn connect_declaration(
             connect_storage_declaration(storage, graph, entry_node, tree_type);
             Ok(leaves.to_vec())
         }
-        ErrorRecovery | GenericTypeForFunctionScope { .. } => Ok(leaves.to_vec()),
+        ErrorRecovery | GenericTypeForFunctionScope { .. } | Break | Continue => {
```
sezna, line 417: Here, we should return `vec![]` I think, since break and continue both abort the flow.

sezna (changes requested):
I think we need to update the path analysis stuff -- see comments.

sway-core/src/ir_generation/function.rs
mohammadfawaz: My reasoning here is that a basic block (at least in LLVM) must have a single terminator. A block with multiple `br` instructions in it is malformed by definition. We can choose to change that if we like.

On the topic of only running the verifier at the end... I actually think we should go in the opposite direction and run the verifier as many times as possible to make sure that no optimization pass (nor IRgen itself) is generating illegal IR.

sway-core/tests/sway_to_ir/break.ir
mohammadfawaz: Yeah those two options you mentioned are possible! I had to do this unfortunate reordering because I need to know the final block before processing the body. That being said, some future loop optimization passes may decide to reorder those blocks anyways (things like loop rotation) so readability of the IR may be compromised in the future for other reasons. But everything functions well with reordered blocks so that's a good sign!

test/src/e2e_vm_tests/test_programs/should_pass/language/break_and_continue/src/main.sw
mohammadfawaz: I added two smaller tests and reduced the complexity of the existing test a bit. I do want to keep the nested structure and sibling loops and combine break and continue in one test to see what happens.

sway-core/tests/sway_to_ir/break.ir
otrho: True, readability isn't really number-1 priority.

sway-core/src/ir_generation/function.rs
otrho: OK, fair call. 🙂 I initially thought the verifier should be run during optimisation by the pass manager, when we get one.   

OTOH seeing friction in IRgen is annoying.  We _could_ eventually allow crap from IRgen, not verify prior to opts, but mandate DCE at the top of the pass manager list and then verify like crazy.  That seems a bit convoluted, but if it means we can have dead terminators and empty blocks, etc. come out of IRgen without breaking anything it might be easier overall.

sway-core/src/control_flow_analysis/dead_code_analysis.rs
mohammadfawaz: Makes sense. Updated!

sway-core/src/control_flow_analysis/analyze_return_paths.rs
```diff
@@ -173,7 +173,9 @@ fn connect_declaration(
         | StructDeclaration(_)
         | EnumDeclaration(_)
         | StorageDeclaration(_)
-        | GenericTypeForFunctionScope { .. } => leaves.to_vec(),
+        | GenericTypeForFunctionScope { .. }
+        | Break
+        | Continue => leaves.to_vec(),
```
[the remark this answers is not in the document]
mohammadfawaz: Alright, I am now handling `break` and `continue` in this pass.

sezna (approved):
Approving to clear my requested changes regarding the graph algos

otrho (approved):

PR #2113, opened 2022-06-25
Added storage field alignment threshold to formatter config

### LINKED ISSUE #2102: Add storage field alignment threshold to formatter config
I think we would like to offer the option of aligning fields (just like we are doing for `enums` and `structs`) inside the storage like the following example:

```
storage {
    long_var_name: Type1,
    var2: Type2,
    ...
}
```
to
```
storage {
    long_var_name: Type1,
    var2         : Type2,
    ...
}
```

kayagokalp:
closes #2102.

### DIFF
--- a/sway-fmt-v2/src/config/user_def.rs
+++ b/sway-fmt-v2/src/config/user_def.rs
@@ -1,7 +1,8 @@
 //! Configuration options related to formatting user-defined structures.
 
 use crate::constants::{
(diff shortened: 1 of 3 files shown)


eureka-cpu:
Looks good! I realized last night that I didn't add the option to opt in or out of using the alignment, so I'll add an issue to address that.

eureka-cpu:
> Looks good! I realized last night that I didn't add the option to opt in or out of using the alignment, so I'll add an issue to address that.

#2114

eureka-cpu (approved):
I think for now we can get this PR in so it can unblock you for `storage`, and later we can address #2114 in a separate PR

mitchmindtree (approved):

PR #2121, opened 2022-06-27
Collect tokens into a `HashMap` instead of a `Vec` in the language server.

JoshuaBatty:
This introduces the following type for collecting tokens during parsing. 

```
pub type TokenMap = HashMap<(Ident, Span), TokenType>;

#[derive(Debug, Clone)]
pub struct TokenType {
    pub parsed: AstToken,
    pub typed: Option<TypedAstToken>,
}
```

First, we traverse the untyped parse tree and collect the `AstToken` into `parsed`.  Then we traverse the typed_parse_tree and optionally collect a `TypedAstToken` if one was found. After both traversal stages, we end up with a `HashMap` that contains all of the tokens found in the sway file. 

This PR replaces the current `Token` module with the above data structure. As such, quite a lot of logic around the current capabilites needs to be re-written to accomodate for the new `TokenMap ` type. Marking this as a draft for now but opening a PR for visiblity.


adlerjohn (commented):
Off the top of my head, we've had issues with using hash maps before due to their non-determinism. See #459

mitchmindtree:
> Off the top of my head, we've had issues with using hash maps before due to their non-determinism

True, worth keeping in mind @JoshuaBatty that `BTreeMap` is a safer default in this regard. The non-determinism of `HashMap` comes into play when iterating over it, whereas `BTreeMap`'s keys are always ordered. If we don't iterate over the `HashMap` at all no need to worry, but worth keeping in mind.

JoshuaBatty:
Ok good to know, I did originally set out to use a `BTreeMap` but i'm unable to use `Span` as the key.

`the trait bound `sway_types::Span: Ord` is not satisfied`

sway-lsp/src/capabilities/document_symbol.rs
```diff
@@ -21,30 +32,113 @@ pub fn to_symbol_information(tokens: &[Token], url: Url) -> Vec<SymbolInformatio
...
+                TypedDeclaration::VariableDeclaration(_) => SymbolKind::VARIABLE,
+                TypedDeclaration::ConstantDeclaration(_) => SymbolKind::CONSTANT,
+                TypedDeclaration::FunctionDeclaration(_) => SymbolKind::FUNCTION,
+                TypedDeclaration::TraitDeclaration(_) => SymbolKind::INTERFACE,
+                TypedDeclaration::StructDeclaration(_) => SymbolKind::STRUCT,
+                TypedDeclaration::EnumDeclaration(_) => SymbolKind::ENUM,
+                TypedDeclaration::ImplTrait { .. } => SymbolKind::INTERFACE,
+                TypedDeclaration::AbiDeclaration(_) => SymbolKind::INTERFACE,
+                TypedDeclaration::GenericTypeForFunctionScope { .. } => SymbolKind::TYPE_PARAMETER,
+                // currently we return `variable` type as default
```
mitchmindtree, line 110: Weird, is there no keyword symbol kind or anything long those lines? Or should we possibly be returning `None` or similar? Not a blocker as I can see this was the original behaviour, just curious.

sway-lsp/src/capabilities/document_symbol.rs
```diff
@@ -21,30 +32,113 @@ pub fn to_symbol_information(tokens: &[Token], url: Url) -> Vec<SymbolInformatio
 
 #[allow(warnings)]
 // TODO: the "deprecated: None" field is deprecated according to this library
-fn create_symbol_info(token: &Token, url: Url) -> SymbolInformation {
+fn create_symbol_info(ident: &Ident, token: &TokenType, url: Url) -> SymbolInformation {
```
mitchmindtree, line 35:
```suggestion
fn symbol_info(ident: &Ident, token: &TokenType, url: Url) -> SymbolInformation {
```

sway-lsp/src/capabilities/hover.rs
```diff
@@ -1,98 +1,108 @@
...
 
-fn get_hover_format(token: &Token, documents: &Documents) -> Hover {
-    let value = match &token.token_type {
-        TokenType::VariableDeclaration(var_details) => {
-            let var_type = match &var_details.var_body {
-                VarBody::FunctionCall(fn_name) => get_var_type_from_fn(fn_name, documents),
-                VarBody::Type(var_type) => var_type.clone(),
-                _ => "".into(),
-            };
+fn get_hover_format(token: &TokenType, ident: &Ident) -> Hover {
```
mitchmindtree, line 41:
```suggestion
fn hover_format(token: &TokenType, ident: &Ident) -> Hover {
```

sway-lsp/src/capabilities/hover.rs
```diff
@@ -1,98 +1,108 @@
...
+        format!(
+            "{}{} {}",
+            extract_visibility(&visibility),
+            decl_name,
+            token_name
+        )
+    };
+
+    let format_variable_hover = |is_mutable: bool, type_name: String| -> String {
+        format!("let{} {}: {}", is_mutable, token_name, type_name,)
```
mitchmindtree, line 55: It looks like this would format as something like:

```rust
letfalse token_name: type_name
```

Is that the intent?

sway-lsp/src/capabilities/semantic_tokens.rs
```diff
@@ -28,43 +34,51 @@ pub fn get_semantic_tokens_full(
...
-        let semantic_token = create_semantic_token(&tokens[i], Some(&tokens[i - 1]));
         semantic_tokens.push(semantic_token);
+        prev_token_span = Some(span);
     }
 
     semantic_tokens
 }
 
-fn create_semantic_token(next_token: &Token, prev_token: Option<&Token>) -> SemanticToken {
+fn create_semantic_token(
```
mitchmindtree, line 52:
```suggestion
fn semantic_token(
```

sway-lsp/src/capabilities/semantic_tokens.rs
```diff
@@ -28,43 +34,51 @@ pub fn get_semantic_tokens_full(
...
+    prev_token_span: Option<&Span>,
+) -> SemanticToken {
+    let next_token_range = get_range_from_span(next_token_span);
+    let next_token_line_start = next_token_range.start.line;
+
     // TODO - improve with modifiers
     let token_modifiers_bitset = 0;
-    let token_type = get_type(&next_token.token_type);
-    let length = next_token.length;
+    let length = next_token_range.end.character - next_token_range.start.character;
```
mitchmindtree, line 62:
```suggestion
    let length = next_token_range.len_chars();
```
^ Maybe this range type should have a `len_chars()` method or something along these lines?

sway-lsp/src/capabilities/semantic_tokens.rs
```diff
@@ -75,27 +89,45 @@ fn create_semantic_token(next_token: &Token, prev_token: Option<&Token>) -> Sema
...
-fn get_type(token_type: &TokenType) -> u32 {
-    match token_type {
-        TokenType::FunctionDeclaration(_)
-        | TokenType::FunctionApplication
-        | TokenType::TraitFunction => TokenTypeIndex::Function as u32,
-        TokenType::Library => TokenTypeIndex::Namespace as u32,
-        TokenType::FunctionParameter => TokenTypeIndex::Parameter as u32,
-        TokenType::VariableDeclaration(_) | TokenType::VariableExpression => {
-            TokenTypeIndex::Variable as u32
+fn get_type_idx(ast_token: &AstToken) -> u32 {
```
mitchmindtree, line 100:
```suggestion
fn type_idx(ast_token: &AstToken) -> u32 {
```

sway-lsp/src/core/document.rs
```diff
@@ -38,84 +34,93 @@ impl TextDocument {
...
-        let line = position.line;
-
-        if let Some(indices) = self.lines.get(&line) {
-            for index in indices {
-                let token = &self.tokens[*index];
-                if token.is_within_character_range(position.character) {
-                    return Some(token);
-                }
+    /// Check if the code editor's cursor is currently over an of our collected tokens
+    pub fn get_token_at_position(&self, position: Position) -> Option<(Ident, &TokenType)> {
```
mitchmindtree, line 44:
```suggestion
    pub fn token_at_position(&self, position: Position) -> Option<(Ident, &TokenType)> {
```

sway-lsp/src/core/document.rs
```diff
@@ -38,84 +34,93 @@ impl TextDocument {
...
     }
 
-    pub fn get_all_tokens_by_single_name(&self, name: &str) -> Option<Vec<&Token>> {
-        if let Some(indices) = self.values.get(name) {
-            let tokens = indices.iter().map(|index| &self.tokens[*index]).collect();
-            Some(tokens)
-        } else {
-            None
-        }
+    pub fn get_all_references(&self, token: &TokenType) -> Vec<(&Ident, &TokenType)> {
```
mitchmindtree, line 56:
```suggestion
    pub fn all_references(&self, token: &TokenType) -> Vec<(&Ident, &TokenType)> {
```

sway-lsp/src/core/document.rs
```diff
@@ -38,84 +34,93 @@ impl TextDocument {
...
+            .collect()
     }
 
-    pub fn get_declared_token(&self, name: &str) -> Option<&Token> {
-        if let Some(indices) = self.values.get(name) {
-            for index in indices {
-                let token = &self.tokens[*index];
-                if token.is_initial_declaration() {
-                    return Some(token);
+    pub fn get_declared_token_ident(&self, token: &TokenType) -> Option<Ident> {
```
mitchmindtree, line 72:
```suggestion
    pub fn declared_token_ident(&self, token: &TokenType) -> Option<Ident> {
```

sway-lsp/src/core/document.rs
```diff
@@ -38,84 +34,93 @@ impl TextDocument {
...
+                    _ => None,
                 }
             }
+            None => None,
         }
-        None
     }
 
-    pub fn _get_token_map(&self) -> &TokenMap {
+    pub fn get_token_map(&self) -> &TokenMap {
```
mitchmindtree, line 94:
```suggestion
    pub fn token_map(&self) -> &TokenMap {
```

sway-lsp/src/core/session.rs
```diff
@@ -74,15 +76,14 @@ impl Session {
     // Token
     pub fn get_token_ranges(&self, url: &Url, position: Position) -> Option<Vec<Range>> {
```
mitchmindtree, line 77:
```suggestion
    pub fn token_ranges(&self, url: &Url, position: Position) -> Option<Vec<Range>> {
```

sway-lsp/src/core/traverse_parse_tree.rs
```diff
@@ -0,0 +1,420 @@
...
+            to_ident_key(&parameter.name),
+            TokenType::from_parsed(AstToken::FunctionParameter(parameter.clone())),
+        );
+    }
+
+    handle_custom_type(&func.return_type, tokens);
+}
+
+fn handle_custom_type(_type_info: &TypeInfo, _tokens: &mut TokenMap) {
+    // TODO: Not obvious how to handle this now with the new types
```
mitchmindtree, line 51:
```suggestion
    todo!("Not obvious how to handle this now with the new types");
```

If we don't want to panic here, we should maybe open an issue for this and mention it under this `TODO` and move the commented out code to the issue?

sway-lsp/src/utils/token.rs
```diff
@@ -0,0 +1,64 @@
...
+    false
+}
+
+// We need to do this work around as the custom PartialEq for Ident impl
+// only checks for the string, not the span.
+pub(crate) fn to_ident_key(ident: &Ident) -> (Ident, Span) {
+    (ident.clone(), ident.span())
+}
+
+pub fn get_type_id(token_type: &TokenType) -> Option<TypeId> {
```
mitchmindtree, line 40:
```suggestion
pub fn type_id(token_type: &TokenType) -> Option<TypeId> {
```

mitchmindtree (commented):
Looking good for the most part! Just a few small comments and clarifications.

I also added a bunch of name change recommendations (mostly removing extraneous `get_` prefixes) but feel free to dismiss these comments in favour of a future PR if you'd prefer.

sway-lsp/src/capabilities/semantic_tokens.rs
JoshuaBatty: Yeah that would be cool. The `next_token_range` here is actually a `Range` type in the `lsp_types` crate though.

sway-lsp/src/capabilities/hover.rs
JoshuaBatty: Nope that's the intent :) good spot!

sway-lsp/src/core/traverse_parse_tree.rs
JoshuaBatty: Good idea, I've opened #2274 and moved the code over in there for the time being.

sway-lsp/src/capabilities/document_symbol.rs
JoshuaBatty: Nice good call, i've gone ahead and applied this change and the subsequent suggestions of removing the get_ and create_ prefexies.

sway-lsp/src/capabilities/document_symbol.rs
JoshuaBatty: Na there isn't a keyword variant. I agree though that we should scrutinize exactly the role and mappings of nodes to SymbolKind's in a future PR. I was just keeping the original behavior in place for now.

JoshuaBatty:
> Looking good for the most part! Just a few small comments and clarifications.
> 
> I also added a bunch of name change recommendations (mostly removing extraneous `get_` prefixes) but feel free to dismiss these comments in favour of a future PR if you'd prefer.

Thanks for the review @mitchmindtree! Just addressed all your feedback :)

sezna:
+1 on hashmap nondeterminism, but in stuff outside of core compiler functionality, the nondeterminism should be less catastrophic

mitchmindtree (approved):
:shipit:

Issue #2114, opened 2022-06-25
Update `Config` to have a single field alignment option


eureka-cpu:
This could look like:

```Rust
pub struct Structures {
    /// Align fields of user defined structures to a threshold.
    pub field_alignment: FieldAlignment,
    /// Put small struct literals on a single line.
    pub struct_lit_single_line: bool,
}

pub enum FieldAlignment {
    AlignFields(usize),
    Off,
}
```

This would reduce the clutter in `Config`, remove the need for extra constants and take away the need for the user to input the same number for each type of alignment.

kayagokalp:
Hmm, that looks good but in doing so wouldn't we block people from selecting which type to align (with different thresholds)? For example, would there be a case where a user wants to align enums and structs but not storage fields?

eureka-cpu:
One thing mentioned before is that we don't want to make `swayfmt` "fully customizable" like `rustfmt`. I don't really see a user in the future that would want one aligned but not the other, but if those users do arise it's an easy fix. In the mean time it makes the use case much clearer and cleaner in the code itself.

PR #2124, opened 2022-06-27
Add storage formatting to sway-fmt-v2

### LINKED ISSUE #2069: Add `storage` formatting to `sway-fmt-v2`

kayagokalp:
closes #2069.


sway-fmt-v2/src/items/item_storage.rs
```diff
@@ -1,15 +1,139 @@
...
+        formatter.shape = shape;
+        Ok(())
+    }
+    fn close_curly_brace(
+        line: &mut String,
+        formatter: &mut Formatter,
     ) -> Result<(), FormatterError> {
-        todo!()
+        line.push(Delimiter::Brace.as_close_char());
+        // If shape is becoming left-most alligned or - indent just have the defualt shape
```
sezna, line 132: I don't quite understand this comment here

sway-fmt-v2/src/items/item_storage.rs
```diff
@@ -1,15 +1,139 @@
...
+        line: &mut String,
+        formatter: &mut Formatter,
+    ) -> Result<(), FormatterError> {
+        let brace_style = formatter.config.items.item_brace_style;
+        let extra_width = formatter.config.whitespace.tab_spaces;
+        let mut shape = formatter.shape;
+        let open_brace = Delimiter::Brace.as_open_char();
+        match brace_style {
+            ItemBraceStyle::AlwaysNextLine => {
+                // Add openning brace to the next line.
```
sezna, line 113:
```suggestion
                // Add opening brace to the next line.
```

sway-fmt-v2/src/items/item_storage.rs
```diff
@@ -1,15 +1,139 @@
...
+            formatted_code.push_str(&formatter.shape.indent.to_string(formatter));
+
+            // Push the storage field name
+            formatted_code.push_str(item.name.as_str());
+
+            let current_variant_length = variant_length[item_index];
+
+            if current_variant_length < max_valid_variant_length {
+                // We need to add alignment between : and ty
+                // max_valid_variant_length: the length of the variant that we are taking as a reference to allign
```
sezna, line 60:
```suggestion
                // max_valid_variant_length: the length of the variant that we are taking as a reference to align
```

sway-fmt-v2/src/fmt.rs
```diff
@@ -273,4 +273,50 @@ pub const TEST1: u16 = 10;"#;
...
+    fn test_storage_with_alignment() {
+        let sway_code_to_format = r#"contract;
+
+storage {
+ long_var_name: Type1,
+      var2: Type2,
+}
+"#;
+        let correct_sway_code = r#"contract;
+
```
sezna, line 309: Could we get some tests that are on one line? e.g.

```rust
storage { foo: Test, bar: Test, baz: u64 } 
```

and maybe something more gross?
```rust
storage{foo:Test,bar
: 
    Test
, baz: u64 } 
```

sezna (changes requested):
Just a couple more tests and I think we are good!

sway-fmt-v2/src/items/item_storage.rs
kayagokalp: I refactored the comment here 👍 Basically `shrink_left` will return an error and if the current indentation level is becoming invalid, (i.e, current indentation level is becoming `< 0`)  in that case we should use the `Shape::default` which is basically `indentation level = 0`.

kayagokalp:
@eureka-cpu I had to add `storage_width`, and `storage_single_line` for deciding when to format to multiple lines, just like `structs`. This might be introducing an "extra" layer of configurability that we do not want. While addressing  #2114 we might want to touch these places as well. I think the issue is general enough to cover this as well so we should not need a separate issue but would love to hear your thought on this. I can also link this comment to there for more visibility.

sway-fmt-v2/src/fmt.rs
```diff
@@ -319,4 +317,24 @@ storage {
...
+storage { long_var_name: Type1, var2: Type2, }"#;
+        let mut config = Config::default();
+        config.structures.struct_lit_single_line = true;
+        config.whitespace.max_width = 300;
+        let mut formatter = get_formatter(config, Shape::default());
+        let formatted_sway_code =
+            Formatter::format(&mut formatter, Arc::from(sway_code_to_format), None).unwrap();
+        assert!(correct_sway_code == formatted_sway_code)
```
mitchmindtree, line 339: Ahh I've only just noticed we've been using `assert!` in these tests when checking equality.

It's totally fine to do so for catching errors, but it's better to use `assert_eq!(a, b)` when asserting equality when possible. This is because the `assert_eq!` macro will actually display the left and right values in the failed test output, giving a better indication of why a test might be failing. On the other hand, `assert!` only shows the expression itself which is less useful.

We can probably change all of these at once in a follow-up PR though.

mitchmindtree (approved):

sway-fmt-v2/src/fmt.rs
eureka-cpu: I've been wrapping them in `dbg!()` when testing then removing them when finished

kayagokalp:
Once again I am in need of your reviews 🙌  😄 @mitchmindtree @sezna @eureka-cpu

sway-fmt-v2/src/constants.rs
```diff
@@ -14,9 +14,9 @@ pub const DEFAULT_TAB_SPACES: usize = 4;
 pub const DEFAULT_FN_CALL_WIDTH: usize = 60;
 /// Default max width of the args of a function-like attributes before falling back to vertical formatting.
 pub const DEFAULT_ATTR_FN_LIKE_WIDTH: usize = 70;
-/// Default max width in the body of a struct literal before falling back to vertical formatting.
+/// Default max width in the body of a structure literal before falling back to vertical formatting.
```
eureka-cpu, line 17: Would you mind if we are explicit here and say `user-defined structure`? This is how I worded it in `Config` just forgot to update this

sway-fmt-v2/src/constants.rs
```diff
@@ -14,9 +14,9 @@ pub const DEFAULT_TAB_SPACES: usize = 4;
...
 pub const DEFAULT_STRUCT_LIT_WIDTH: usize = 18;
-/// Default max width in the body of a struct variant before falling back to vertical formatting.
+/// Default max width in the body of a structure variant before falling back to vertical formatting.
```
eureka-cpu, line 19: Same here except we should call them `fields`, `variants` only apply to `enum` so we can be more broad since it will be used for all structure types

sway-fmt-v2/src/constants.rs
eureka-cpu: Also the name of this `const` could be `DEFAULT_STRUCTURE_LIT_WIDTH`

sway-fmt-v2/src/constants.rs
eureka-cpu: Also here this can be `DEFAULT_STRUCTURE_FIELD_WIDTH`

sway-fmt-v2/src/fmt.rs
```diff
@@ -271,4 +271,69 @@ pub const TEST1: u16 = 10;"#;
...
+
+storage {
+ long_var_name: Type1,
+      var2: Type2,
+}
+"#;
+        let correct_sway_code = r#"contract;
+
+storage {
+    long_var_name: Type1,
```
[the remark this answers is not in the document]
eureka-cpu: e.g. the end result is: 
```Rust
storage {
    long_var_name : Type1,
    var2          : Type2,
}
```

eureka-cpu (commented):
Just a few small things and should be good to go!

JoshuaBatty (approved):

adlerjohn:
Poking @sezna for review

mitchmindtree (approved):

sezna (approved):

PR #2132, opened 2022-06-27
Demonstrate that `T` for `Vec` can now be inferred by arguments

### LINKED ISSUE #2049: `new()` for `Vec<T>` requires type information
Currently, we need to write: 
```rust
let vector = ~Vec::new::<u8>();
```
The type annotation `u8` has to be supplied to `new()`. The following, however, doesn't work:
```rust
let vector: Vec<u8> = ~Vec::new();
```
The second syntax is a bit more intuitive to users and should probably work.

### LINKED ISSUE #2050: Infer the `T` in `Vec<T>` from the elements being pushed into it
The following should work:
```rust
let vector = ~Vec::new();
vector.push(42);
```
Just like Rust, we should be able to infer that `T` resolves to `u64`.

emilyaherbert:
- Closes #2050 
- Closes #2049


test/src/e2e_vm_tests/test_programs/should_fail/vec/src/main.sw
```diff
@@ -0,0 +1,716 @@
...
+    assert(vector.len() == 0);
+    assert(vector.capacity() == 0);
+    assert(vector.is_empty());
+
+    vector.push(number0);
+    vector.push(number1);
+    vector.push(number2);
+    vector.push(number3);
+    vector.push(number4);
+    vector.push(false);
```
mohammadfawaz, line 51: Is it possible to reduce the size of this failing test to the bare minimum by removing all or most of the things that don't fail?

emilyaherbert: Done.

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
```diff
@@ -27,7 +27,7 @@ fn main() -> bool {
 }
 
 fn test_vector_new_u8() {
-    let mut vector: Vec<u8> = ~Vec::new::<u8>();
+    let mut vector = ~Vec::new();
```
mohammadfawaz, line 30: Very nice!

mohammadfawaz (approved):

otrho (approved):

PR #2137, opened 2022-06-28
#2020 - changed BASE_ASSET_ID type to ContractId

### LINKED ISSUE #2020: Make `BASE_ASSET_ID` a constant `ContractId`
Now that #1948 is in, https://github.com/FuelLabs/sway/blob/3ce5945cbe8d169f5ec17a2b731521c0284de553/sway-lib-std/src/constants.sw#L3 should be

```rust
const BASE_ASSET_ID = ~ContractId::from(0x0000000000000000000000000000000000000000000000000000000000000000);
```

r-sitko:
Closes #2020


test/src/e2e_vm_tests/test_programs/should_pass/stdlib/contract_id_type/src/main.sw
```diff
@@ -9,8 +9,8 @@ fn main() -> bool {
     let id_2 = ~ContractId::from(0x0000000000000000000000000000000000000000000000000000000000000000);
 
     assert(id_1 == id_2);
-    assert(id_1 == ~ContractId::from(BASE_ASSET_ID));
-    assert(id_1.into() == BASE_ASSET_ID);
+    assert(id_1 == BASE_ASSET_ID);
+    assert(id_1.into() == BASE_ASSET_ID.into());
```
Braqzen, line 13: I think you can remove this line. Previously the `BASE_ASSET_ID` was a constant instead of the `ContractId` type so `.into()` was needed. This is no longer the case as the test above covers it.

Braqzen (changes requested):
Good work.
There's 1 thing that I'd like you to change though

examples/identity/src/main.sw
```diff
@@ -25,7 +25,7 @@ storage {
 impl IdentityExample for Contract {
     fn cast_to_identity() {
         // ANCHOR: cast_to_identity
-        let my_address: Address = ~Address::from(BASE_ASSET_ID);
+        let my_address: Address = ~Address::from(BASE_ASSET_ID.into());
```
adlerjohn, line 28: Hmmm. This doesn't really make sense now since you don't want to be using an asset ID as an address. Can you change to something like
```suggestion
        let raw_address: b256 = 0xddec0e7e6a9a4a4e3e57d08d080d71a299c628a46bc609aab4627695679421ca;
        let my_address: Address = ~Address::from(raw_address);
```

adlerjohn (changes requested):

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/contract_id_type/src/main.sw
r-sitko: Yeah, you are right. What do you think about changing this line to below line to make testing into()?
assert(0x0000000000000000000000000000000000000000000000000000000000000000 == BASE_ASSET_ID.into());

examples/identity/src/main.sw
Braqzen: I did think that the `.into()` was a bit magical here in the conversion but thought that the next issue where a `const b256` would just update this after

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/contract_id_type/src/main.sw
r-sitko: Done

examples/identity/src/main.sw
r-sitko: Done

Braqzen (approved):

PR #2153, opened 2022-06-28
Updates `user_def` with `FieldAlignment` & fix corresponding use cases

### LINKED ISSUE #2114: Update `Config` to have a single field alignment option

### LINKED ISSUE #2139: Add support for single line enum formatting
Currently, enums are always formatted into multiple lines but I guess we should do the same thing we are doing for `structs` and `storage` (once #2124 is merged) which is deciding whether to format into single or multiple lines based on the length. 

So once we address this issue we should be able to format:

```
enum Color {
    Blue: (),
    Green: (),
}
``` 
to something like:
```
enum Color { Blue: (), Green: () }
```

eureka-cpu:
This PR adds a unifying field for aligning user defined structures if given a threshold. 

This previously would require the user to give 3 separate thresholds (`structs`, `enums`, `storage`). This reduces the clutter down to a single field: `field_alignment` with a value of `FieldAlignment` (`AlignFields(usize)`, `Off`). It also updates the other field on `user_def::Structures`, `struct_lit_single_line`, to be named more appropriately since it will be used to reference all user defined structures as well. 

Tests have also been updated to reflect these changes, and the logic for these structures is much clearer as well. `ItemEnum` now looks more like `ItemStruct` in terms of code clarity, and I've added the `single_line` functionality from `ItemStruct` to it as well.

During my workflow I also added formatting for `TypeField` in `punctuated.rs` for ease of use. See #2154 for more.

Closes #2114 
Closes #2139


sway-fmt-v2/src/items/item_enum.rs
```diff
@@ -17,74 +17,167 @@ impl Format for ItemEnum {
...
+/// Format the enum if the multiline is passed as false enum will be formatted into a single line.
+///
+/// Example (multiline : false):
+/// enum Foo { bar: u64,  baz: bool }
+///
+/// Example (multiline : true):
+/// enum Foo {
+///     bar: u64,
+///     baz: bool,
+/// }
```
mitchmindtree, line 49: We should try to be weary of doc comment formatting - see my comment [here](https://github.com/FuelLabs/sway/pull/2058#discussion_r904440534).

mitchmindtree (commented):
Just having a quick look before stepping out for a bit, will look closer soon.

I just wanted to add that I don't think we should worry about these non-default configurations until we've landed the MVP V1 replacement. I think we're better off focusing on landing the minimal work to replace V1 before getting fancy with features like this.

I think field/variant type alignment in particular is probably not something we want to encourage. While it looks very pretty, it's very unfriendly to diffs. Specifically, when adding or removing a field or variant that happens to be longer than any of the existing set, every other field/variant will need to be updated by the auto-formatter. This is the kind of thing that can lead to unnecessary/irrelevant conflicts in downstream Sway code e.g.

Apologies if we have an open issue for this that I missed, I likely should have left a comment like this there first.

sway-fmt-v2/src/items/item_enum.rs
eureka-cpu: Ah yes, I did see that comment just wasn't on my mind. I'll fix that. 

I just did this one since I started to see the code get a little disorganized, shouldn't really impact the time frame too much. I'll stay focused on the MVP tho, just trying to stay organized and on top of things (:

eureka-cpu: Also I agree with the above sentiment about alignment, it's something that was deliberated a bit before starting the new formatter

kayagokalp (dismissed):
LGTM, just left a tiny comment but it is not a blocker as we can always address it later with a different PR.

sway-fmt-v2/src/items/item_enum.rs
```diff
@@ -17,74 +17,167 @@ impl Format for ItemEnum {
...
+
+    // Handle closing brace
+    ItemEnum::close_curly_brace(formatted_code, formatter)?;
+    Ok(())
+}
+
+impl ItemLen for ItemEnum {
+    fn get_formatted_len(&self) -> usize {
+        // TODO while determininig the length we may want to format to some degree and take length.
+        let str_item = &self.span().as_str().len();
```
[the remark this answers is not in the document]
eureka-cpu: Let's get that in a separate PR

mitchmindtree (approved):

kayagokalp (approved):

PR #2159, opened 2022-06-29
Speedup `find_dead_code` pass in control flow analysis

### LINKED ISSUE #1952: Dead Code Analysis suspiciously slow (up to 28x slower than `type_check`) particularly for `std`
I've noticed the E2E tests and stdlib tests have been progressively getting slower and slower as we add new APIs to `std`.

This is particularly problematic for `sway-lsp`. @JoshuaBatty is in the process of updating `sway-lsp` to use `compile_to_ast`, however the current state of things (taking multiple seconds just for `std`) is unfeasible for the kind of feedback a user would normally expect in an IDE environment.

To work out what's going on, I added some basic time-stamping between all of the major compiler phases to see where most time was being spent. You can see the exact approach in [this commit](https://github.com/FuelLabs/sway/commit/1576f1dfdcb0c45c96b7b505cba1102d398f4034).

Here is the output of `forc build --path examples/counter` with `forc` built in `debug`:

```sh
$ cargo run --bin forc -- build --path examples/counter
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/forc build --path examples/counter`
Parsing: 1.06626ms
Type Check: 1.344043ms
Dead Code Analysis: 294.733µs
Return Path Analysis: 146.552µs
  Compiled library "core_utils".
Parsing: 12.502972ms
Type Check: 75.687842ms
Dead Code Analysis: 113.27732ms
Return Path Analysis: 4.030499ms
  Compiled library "core".
Parsing: 26.412109ms
Type Check: 159.818236ms
Dead Code Analysis: 4.463190539s
Return Path Analysis: 9.589824ms
  Compiled library "std".
Parsing: 566.406µs
Type Check: 861.018µs
Dead Code Analysis: 209.432µs
Return Path Analysis: 89.74µs
Finalize Types: 51.21µs
Compile IR: 1.146012ms
Purity Check: 16.96µs
Inline Fn Calls: 54.66µs
Combine Constants: 3.84µs
Compile ASM: 2.341203ms
  Compiled contract "counter".
  Bytecode size is 256 bytes.
```

Curiously, the Dead Code Analysis phase is taking vastly longer than everything else when compiling `std`: `Dead Code Analysis: 4.463190539s`.

Compiling `forc` in `--release` does run significantly faster:

```sh
$ forc build --path examples/counter
Parsing: 202.012µs
Type Check: 311.453µs
Dead Code Analysis: 32.1µs
Return Path Analysis: 20.86µs
  Compiled library "core_utils".
Parsing: 1.419324ms
Type Check: 8.038199ms
Dead Code Analysis: 10.101929ms
Return Path Analysis: 585.396µs
  Compiled library "core".
Parsing: 2.477274ms
Type Check: 20.874826ms
Dead Code Analysis: 215.495149ms
Return Path Analysis: 1.455874ms
  Compiled library "std".
Parsing: 79.69µs
Type Check: 139.272µs
Dead Code Analysis: 19.1µs
Return Path Analysis: 7.08µs
Finalize Types: 6.3µs
Compile IR: 123.721µs
Purity Check: 1.94µs
Inline Fn Calls: 5.301µs
Combine Constants: 410ns
Compile ASM: 220.282µs
  Compiled contract "counter".
  Bytecode size is 256 bytes.
```

However Dead Code Analysis still takes ~10x the type-check stage for `std`. It appears there might be an exponential blow-up the bigger the program/library.

It would be worth investigating this soon in order to speed up the debug build, cut down our CI times and improve responsiveness in the upcoming LSP update. It also looks like sway could get close to instantaneous / real-time build times in release mode which would be a nice impression to leave :)

cc @sezna I think I remember saying you had an idea of what might be causing this?

mitchmindtree:
While waiting for the tests to pass on a PR I thought I'd have a quick
look to see if I could find any quick wins for dead code analysis #1952.

I noticed that we're using `has_path_connecting` for every
combination of node and entry point. This means we were re-checking the
same nodes many, many times, searching from scratch each time and not
re-using any of the knowledge of already visited nodes in each
consecutive traversal.

This commit refactors the approach to first collect all known live nodes
into a set by traversing from the entry points. We re-use the same `Dfs`
when searching from each entry in order to re-use its inner set of
visited nodes and avoid re-searching sections of the graph that we've
already visited.

The dead nodes are those not contained in the live set after traversal.

This reduces the time taken within the `find_dead_code` call when
building the `std` library in debug from ~7.9 seconds down to ~3.3
milliseconds. 1000x+ speedup in DCA :)

Hopefully this speeds up our CI a bit!

Edit: locally, this gets my E2E tests finishing in under ~18 secs with a release build.

Closes #1952.

### DIFF
--- a/sway-core/src/control_flow_analysis/dead_code_analysis.rs
+++ b/sway-core/src/control_flow_analysis/dead_code_analysis.rs
@@ -16,29 +16,33 @@ use crate::{
     type_engine::{resolve_type, TypeInfo},
     CompileError, CompileWarning, Ident, TreeType, Warning,
 };
+use std::collections::BTreeSet;
 use sway_types::{span::Span, Spanned};
 
 use crate::semantic_analysis::TypedStorageDeclaration;
 
-use petgraph::algo::has_path_connecting;
 use petgraph::prelude::NodeIndex;
+use petgraph::visit::Dfs;
 
 impl ControlFlowGraph {
     pub(crate) fn find_dead_code(&self) -> Vec<CompileWarning> {
-        // dead code is code that has no path to the entry point
-        let mut dead_nodes = vec![];
-        for destination in self.graph.node_indices() {
-            let mut is_connected = false;
-            for entry in &self.entry_points {
-                if has_path_connecting(&self.graph, *entry, destination, None) {
-                    is_connected = true;
-                    break;
-                }
-            }
-            if !is_connected {
-                dead_nodes.push(destination);
+        // Dead code is code that has no path to the entry point.
+        // Collect all connected nodes by traversing from the entries.
+        // The dead nodes are those we did not collect.
+        let mut connected = BTreeSet::new();
+        let mut dfs = Dfs::empty(&self.graph);
+        for &entry in &self.entry_points {
+            dfs.move_to(entry);
+            while let Some(node) = dfs.next(&self.graph) {
+                connected.insert(node);
             }
         }
+        let dead_nodes: Vec<_> = self
+            .graph
+            .node_indices()
+            .filter(|n| !connected.contains(n))
+            .collect();
+
         let dead_enum_variant_warnings = dead_nodes
             .iter()
             .filter_map(|x| match &self.graph[*x] {


JoshuaBatty:
~3.3 milliseconds sounds pretty damn blazingly fast to me! So good, nice work man!

otrho:
> This reduces the time taken within the `find_dead_code` call when building the `std` library in debug from ~7.9 seconds down to ~3.3 milliseconds. 1000x+ speedup in DCA :)

SHUT UP.

otrho (approved):

adlerjohn (approved):
:shipit:

Issue #2151, opened 2022-06-28
Const evaluator fails when initializing a `const` with another `const`


mohammadfawaz:
Example:
```rust
script;

const ZERO_B256 = 0x0000000000000000000000000000000000000000000000000000000000000000;
const KEY = ZERO_B256;

fn main() { }
```
Error:
```console
3 | const ZERO_B256 = 0x0000000000000000000000000000000000000000000000000000000000000000;
4 | const KEY = ZERO_B256;
  |             ^^^^^^^^^ Could not evaluate initializer to a const declaration.
```

cc @vaivaswatha

vaivaswatha:
This seems to be related to the order in which `const` declarations are processed. Renaming `ZERO_B256` to `AERO_B256` gets it compiled fine. `get_all_declared_symbols()`, on whose results `compile_constant_expression` is called, stores an ordered (by name) list of symbols and returns that, not based on actual dependences.

otrho:
Are these not ordered properly by the `order_ast_nodes_by_dependency()`?  Sounds like a bug!

vaivaswatha:
> Are these not ordered properly by the `order_ast_nodes_by_dependency()`? Sounds like a bug!

That's done when type-checking. This is afterwards, when compiling to the IR.

otrho:
Hrmm, so why is this specifically a new problem?

We have a typed AST which we're trying to compile to IR.  The current method will start with 'main' or contract entry points and recursively compile nodes as they're needed.

Is it that until now, function declarations and traits, etc. have been embedded/inlined into the AST and so finding them has been straight forward, and type declarations like structs and enums are just recompiled each time they're used..?  But now with top-level `const` declarations depending on each other there isn't a direct link between them in the AST already, and they aren't found unless they've already been compiled?

vaivaswatha:
> Hrmm, so why is this specifically a new problem?

It's a new problem only because until now we didn't allow a global (top level) `const` declaration to depend on anything else (the RHS had to be a literal).

>But now with top-level const declarations depending on each other there isn't a direct link between them in the AST already, and they aren't found unless they've already been compiled?

That's right. `const_eval`, when it tries to evaluate an initializer looks up `get_global_constant`, and in this case doesn't find it because of the order not being right.

PR #2162, opened 2022-06-29
create `ZERO_B256` const

### LINKED ISSUE #2131: Repurpose `BASE_ASSET_ID` `b256` value as the zero value
# Summary

There is an issue which wants to turn the `BASE_ASSET_ID` into a `ContractId` type. Ref: #2020 
This makes sense however that "zero address" is used beyond the contract type.

We should still have the following

```rust
const ZERO_B256 = 0x0000000000000000000000000000000000000000000000000000000000000000;
```

r-sitko:
closes #2131


Braqzen (commented):
Seems pretty reasonable

sway-lib-std/src/constants.sw
```diff
@@ -3,3 +3,4 @@ library constants;
 use ::contract_id::ContractId;
 
 const BASE_ASSET_ID = ~ContractId::from(0x0000000000000000000000000000000000000000000000000000000000000000);
+const ZERO_B256 = 0x0000000000000000000000000000000000000000000000000000000000000000;
```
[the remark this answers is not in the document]
r-sitko: Unfortunately not. I tried to do that, but it doesn't work and it is related to #2151 issue.

examples/storage_example/src/main.sw
```diff
@@ -1,21 +1,19 @@
...
     #[storage(write)]fn store_something(amount: u64);
     #[storage(read)]fn get_something() -> u64;
 }
 
-const STORAGE_KEY: b256 = 0x0000000000000000000000000000000000000000000000000000000000000000;
-
 impl StorageExample for Contract {
     #[storage(write)]fn store_something(amount: u64) {
-        store(STORAGE_KEY, amount);
+        store(ZERO_B256, amount);
```
adlerjohn, line 12: This shouldn't be used here, since keys are means to be 0, 1, 2, 3, etc. There's nothing special about the first one. `ZERO_B256` can be used e.g. for comparisons.

test/src/e2e_vm_tests/test_programs/should_pass/language/const_inits/src/main.sw
```diff
@@ -33,12 +32,15 @@ const EN1a = En1::Int(101);
...
+
 const TUP1_idx2 = TUP1.2;
 
 fn main() -> u64 {
+    let ETH_ID0 = ~ContractId::from(ZERO_B256);
+    let ETH_ID0_VALUE = ETH_ID0.value;
+
     // initialization through function applications.
-    let eth_id0 = ~ContractId::from(0x0000000000000000000000000000000000000000000000000000000000000000);
+    let eth_id0 = ~ContractId::from(ZERO_B256);
```
adlerjohn, line 43: Ditto. Semantically, `eth_id0` and `eth_id1` use the values zero and one, but there's nothing special about zero specifically so I would avoid using `ZERO_B256`.

test/src/e2e_vm_tests/test_programs/should_pass/language/many_stack_variables/src/main.sw
```diff
@@ -17,133 +19,133 @@ fn main() -> u64 {
...
-    let b = 0x0000000000000000000000000000000000000000000000000000000000000000;
-    let b = 0x0000000000000000000000000000000000000000000000000000000000000000;
-    let b = 0x0000000000000000000000000000000000000000000000000000000000000000;
-    let b = 0x0000000000000000000000000000000000000000000000000000000000000000;
-    let b = 0x0000000000000000000000000000000000000000000000000000000000000000;
-    let b = 0x0000000000000000000000000000000000000000000000000000000000000000;
-    let b = 0x0000000000000000000000000000000000000000000000000000000000000000;
-    let b = 0x0000000000000000000000000000000000000000000000000000000000000000;
-    let b = 0x0000000000000000000000000000000000000000000000000000000000000000;
+    let b = ZERO_B256;
```
adlerjohn, line 22: This is an okay use of `ZERO_B256` because we just need _some_ value for `b`.

adlerjohn (changes requested):
Revert any changes of the literal `0x0000000000000000000000000000000000000000000000000000000000000000` to `ZERO_B256` where the literal was meant as the first of a sequence (e.g. storage keys).

Put another way, if you have two lines with `ZERO_B256` and `0x0000000000000000000000000000000000000000000000000000000000000001`, that's indicative you should be using the literal `0x0000000000000000000000000000000000000000000000000000000000000000`.

examples/storage_example/src/main.sw
nfurfaro: IIRC, storage keys don't increment yet, but use a hash onion approach.
(I had to revert the PR which implemented `Add` for `b256`.)
Still, I think the suggestion here is a good one.

sway-lib-std/src/constants.sw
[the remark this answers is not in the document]
nfurfaro: @r-sitko could you add a `// Todo: ` here with a link to the issue tracking this?

test/src/e2e_vm_tests/test_programs/should_pass/language/many_stack_variables/src/main.sw
r-sitko: @adlerjohn 
I tried to follow your guidlines. Please check if everything is correct.

test/src/e2e_vm_tests/test_programs/should_pass/language/empty_method_initializer/src/main.sw
```diff
@@ -1,10 +1,10 @@
 script;
 
-use std::b512::B512;
+use std::{b512::B512, constants::ZERO_B256};
 
 fn main() -> bool {
     let hi_bits: b256 = 0x7777777777777777777777777777777777777777777777777777777777777777;
-    let lo_bits: b256 = 0x0000000000000000000000000000000000000000000000000000000000000000;
+    let lo_bits: b256 = ZERO_B256;
```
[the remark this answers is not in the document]
r-sitko: Done

adlerjohn (approved):

Issue #2167, opened 2022-06-29
Handle `Tuple` case for `Ty` formatting


kayagokalp:
part of #2101.

There are 5 different scenarios we should consider while we are formatting a `Ty`. These are:

1. `Array`
2. `Infer`
3. `Path`
4. `Str`
5. `Tuple`

This issue is for tracking the `Tuple` scenario.

PR #2168, opened 2022-06-29
Add 'Ty' formatting to `sway-fmt-v2`

### LINKED ISSUE #2101: Add `ty` formatting to `sway-fmt-v2`
Currently, if we come across a `ty` during the formatting, we simply push its string version without any formatting. This can be seen in https://github.com/FuelLabs/sway/blob/a43aec0d3fb61b7ff7875ecd0259f055cf2a32e3/sway-fmt-v2/src/items/item_enum.rs#L66-L68

I think if we have a re-usable method for formatting `ty` that would come in handy since there are a couple of more places we will need the same logic (i.e, `storage` formatting).

### Possible motivations to format a `ty`:

I am not sure how we should be formatting `ty` but if we can create a list of the motivations that might be helpful while implementing this.

1. It might not fit the total line width.
2. ?

any thoughts @eureka-cpu?

--------
## TODO

- [ ] Address #2163 
- [ ] Address #2164 
- [ ] Address #2165 
- [ ] Address #2166 
- [ ] Address #2167

### LINKED ISSUE #2163: Handle `Infer` case for `Ty` formatting
part of #2101.

There are 5 different scenarios we should consider while we are formatting a `Ty`. These are:

1. `Array`
2. `Infer`
3. `Path`
4. `Str`
5. `Tuple`

This issue is for tracking the `Infer` scenario.

kayagokalp:
This should be good to go for a review. One thing to note is that this PR shares the `path` formatting with #2173 (#2173 has a couple more things implemented for `path` so this `path` formatting is a subset of that one). While merging there might be some conflicts that should be easy to deal with.

## TODO
- [x] Address #2163 
- [x] Address #2164 
- [x] Address #2165 
- [x] Address #2166 
- [x] Address #2167 

closes #2101, #2163, #2164, #2165, #2166, #2167.


sway-fmt-v2/src/items/item_enum.rs
```diff
@@ -70,14 +71,13 @@ impl Format for ItemEnum {
...
             write!(
                 formatted_code,
-                " {} {}{}",
+                " {} ",
                 type_field.colon_token.span().as_str(),
-                type_field.ty.span().as_str(),
-                PunctKind::Comma.as_char(),
             )?;
+            Ty::format(&type_field.ty, formatted_code, formatter)?;
+            write!(formatted_code, "{}", PunctKind::Comma.as_char())?;
```
eureka-cpu, line 80: this can just be `formatted_code.push(PunctKind::Comma.as_char());`

sway-fmt-v2/src/utils/ty.rs
```diff
@@ -36,34 +35,38 @@ fn format_infer(
...
+/// Currently does not apply formatting, just pushes the str version of span
 fn format_array(
-    _formatted_code: &mut FormattedCode,
-    _formatter: &mut Formatter,
-    _underscore_token: SquareBrackets<TyArrayDescriptor>,
+    formatted_code: &mut FormattedCode,
+    array_descriptor: SquareBrackets<TyArrayDescriptor>,
 ) -> Result<(), FormatterError> {
-    todo!()
+    write!(formatted_code, "{}", array_descriptor.span().as_str())?;
```
eureka-cpu, line 43: Same here and a few other places, we don't need the `write` macro if we aren't joining anything I think

eureka-cpu (commented):
Just a few things to fix optionally, we may also want to add something to the tests for this

sway-fmt-v2/src/utils/ty.rs
```diff
@@ -0,0 +1,71 @@
...
+            Self::Infer { underscore_token } => format_infer(formatted_code, underscore_token),
+            Self::Path(path_ty) => format_path(formatted_code, path_ty),
+            Self::Str { str_token, length } => {
+                format_str(formatted_code, str_token.clone(), length.clone())
+            }
+            Self::Tuple(tup_descriptor) => format_tuple(formatted_code, tup_descriptor.clone()),
+        }
+    }
+}
+
```
eureka-cpu, line 27: `Path`, `Tuple` and `Array` can all implement `Format` I think

eureka-cpu: But I think it could be okay either way, I'll defer to your preference

eureka-cpu: If you can, I think it would be cleaner

eureka-cpu (approved):

sway-fmt-v2/src/utils/ty.rs
eureka-cpu: For sure! Don't want to block you, just a suggestion to keep in mind that certain parts may benefit from using the trait instead of just duplicating the function for each one

kayagokalp:
Decided to handle all of the `Ty` with this pr as it seems like I will not be blocked by anything. So converting this to draft until that is done.

eureka-cpu:
Looking good so far!

sway-fmt-v2/src/utils/ty.rs
mitchmindtree: One benefit of using `write!` is that it gives us the flexibility of easily refactoring the kind of buffer that we write into as `write!` supports writing into any type `T` where `T: std::fmt::Write`. I think it's fine to use `write!` generally, perhaps even preferred.

sway-fmt-v2/src/fmt.rs
```diff
@@ -271,4 +271,39 @@ pub const TEST1: u16 = 10;"#;
             Formatter::format(&mut formatter, Arc::from(sway_code_to_format), None).unwrap();
         assert_eq!(correct_sway_code, formatted_sway_code)
     }
+    #[test]
+    fn test_ty_formatting() {
+        let sway_code_to_format = r#"contract;
+
+enum test_ty {
```
mitchmindtree, line 278: I realise this is just a test, but the convention for Rust and Sway types is to use capital camel case for type names, e.g. `TestTy` rather than `test_ty`. Folks sometimes refer to tests and examples upstream as demonstrative of conventions, so we should try to follow them wherever we can.

sway-fmt-v2/src/utils/path.rs
```diff
@@ -0,0 +1,71 @@
...
+use sway_types::Spanned;
+
+impl Format for QualifiedPathRoot {
+    fn format(
+        &self,
+        formatted_code: &mut FormattedCode,
+        formatter: &mut Formatter,
+    ) -> Result<(), FormatterError> {
+        // Commenting this out since this will err until `Ty` formatting gets in
+        // self.ty.format(formatted_code, formatter)?;
```
mitchmindtree, line 13: Should this be commented out? The comment mentions we can uncomment when `Ty` formatting lands, but it seems like that's happening as a part of this PR? Does this get tested yet?

mitchmindtree (changes requested):
Mostly looking good! Just left a couple nits

sway-fmt-v2/src/fmt.rs
kayagokalp: Ohh yeah, of course. I must have missed this one and now it triggers me pretty badly 😄

sway-fmt-v2/src/utils/path.rs
kayagokalp: I am uncommenting it but for actually testing this I think we need something like the following to the existing test which requires us to have the baseline for the `fn` formatting (which is #2173):

```rust
fn test_path_ty() { 
    match TestTy {
        TestTy::String(str) = > {},
         _ => {}
    }
}
```
I couldn't come up with another way to test that part and if we cannot find another way a test for this can be added with #2173 like above.

mitchmindtree (approved):
:ship:

eureka-cpu (approved):

Issue #2014, opened 2022-06-16
Expand `Vec<T>` with additional utility methods


mohammadfawaz:
An example of a good method to add would be `contains()`. There are probably plenty of other methods that we can add from https://doc.rust-lang.org/stable/std/vec/struct.Vec.html

Braqzen:
1. [Indexing](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#indexing) and being able to assign to a specific index
2. [Inserting](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.insert) at specified index
3. [Removing](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.remove) at specified index
4. [Pop](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.pop) from the end
5. [Append](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.append) vectors together
6. [Split into two](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.split_off)
7. [Split at](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.split_at)
8. [First](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.first) element of vec
9. [Last](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.last) element of vec
10. [Swap](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.swap) two elements
11. [Reverse](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.reverse) the vector
12. [Fill](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.fill)
13. [Sorting](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.sort_by) by various keys, algorithms etc

SwayStar123:
mutating an already existing value at given index is also a necessary function

mohammadfawaz:
> mutating an already existing value at given index is also a necessary function

See https://github.com/FuelLabs/sway/issues/2836

jtriley-eth:
Writing some notes since this was partially completed prior to assignment.

## Features Requested

> Note: (✅ if complete)

1. ✅ [Indexing](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#indexing) and being able to assign to a specific index
[...]

Issue #2155, opened 2022-06-29
memory overflow occurs when using `copy`


XFFXFF:
I'm trying to implement `remove` of `Vec` (ref #2014), but found that memory overflow occurs when I use `copy`, and then I think `copy` seems to have a bug.  

to reproduce:
``` rust
script;

use std::alloc::alloc;
use std::mem::copy;

fn main() {
    let ptr = alloc(32);
    copy(ptr, ptr + 8, 24);  // memory overflow error
}
```
Interestingly, sway is happy with the following example
``` rust
script;

use std::alloc::alloc;
use std::mem::copy;

fn main() {
    let ptr = alloc(32);
    let new_ptr = alloc(24);
    copy(new_ptr, ptr + 8, 24);
}
```
So I guess the memory overflow error occurs when the memory of `src` and the memory of `dst` overlap.

adlerjohn:
Oh, that's expected behavior. https://github.com/FuelLabs/fuel-specs/blob/bd6ec935e3d1797a192f731dadced3f121744d54/specs/vm/instruction_set.md#mcp-memory-copy

> The memory ranges `MEM[$rA, $rC]` and `MEM[$rB, $rC]` overlap

This issue can be re-purposed to improve the documentation on the stdlib functions though

PR #2170, opened 2022-06-29
add `remove`, `insert` and `pop` to `Vec`

XFFXFF:
ref #2014 

add three methods to `Vec`:
* remove
* insert
* pop

`contains`  is blocked by trait constraints, which requires element `T` to implement `core::ops::Eq`    
Some other methods like `first` or `last` will be easy to implement once we can call other methods defined in the same `impl` block.   

I also want to write a test to catch `out of index`, but don't know how, I noticed there is a `should_fail` folder, but still don't know how to write `test.toml`


sway-lib-std/src/vec.sw
```diff
@@ -141,4 +142,71 @@ impl<T> Vec<T> {
...
+
+        // Read from `ptr`
+        let ret = read(ptr);
+
+        // Shift everything down to fill in that spot.
+        let end = buf_start + val_size * self.len;
+        while ptr < end {
+            copy(ptr, ptr + val_size, val_size);
+            ptr += val_size;
+        }
```
XFFXFF, line 164: The loop is needed, can't use `copy(ptr, ptr + val_size, val_size * (self.len - index - 1)` , https://github.com/FuelLabs/sway/issues/2155 discusses this

mohammadfawaz (changes requested):
This is really good! Thank you so much for your contribution. 

I don't really have any comments on the library code itself. It looks correct, clean, and readable. 

As for the test, one thing that I'd like to also see is tests for the new functions called with data types other than `u8`. I suggest that for each of the existing test functions `test_vector_new_*()`, append some calls to `remove`, `insert`, and `pop` at the end just to double check that they work with complex data types.

sway-lib-std/src/vec.sw
```diff
@@ -141,4 +142,71 @@ impl<T> Vec<T> {
...
+        // Increment length.
+        self.len += 1;
+    }
+
+    /// Removes the last element from a vector and returns it, or [`None`] if it
+    /// is empty.
+    fn pop(mut self) -> Option<T> {
+        if self.len == 0 {
+            return Option::None;        
+        } else {
```
Braqzen, line 207: You do not need to use the `else` here unless the compiler hasn't reached the point of development where it understands that is okay. You should be able to continue to control flow as usual without `else`.
If there are no complaints from the compiler then can you remove the "else" so that there is less spaghetti?

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
```diff
@@ -705,3 +708,127 @@ fn test_vector_with_capacity_u64() {
...
+    match vector.pop() {
+        Option::Some(val) => assert(val == number1), Option::None => revert(0), 
+    }
+
+    match vector.pop() {
+        Option::Some(val) => assert(val == number0), Option::None => revert(0), 
+    }
+    
+    match vector.pop() {
+        Option::Some(val) => revert(0), Option::None => {},
```
Braqzen, line 779: I'd feel a little better if we could assert that the last value returned is none rather than not reverting. if that's not possible in sway then perhaps it should be an issue?

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
```diff
@@ -705,3 +708,127 @@ fn test_vector_with_capacity_u64() {
...
+    assert(vector.capacity() == 1);
+    match vector.get(0) {
+        Option::Some(val) => assert(val == number2), Option::None => revert(0), 
+    }
+
+    // insert at first
+    vector.insert(0, number0);
+    assert(vector.len() == 2);
+    assert(vector.capacity() == 2);
+    match vector.get(0) {
```
Braqzen, line 803: this needs 1 more `.get()` to make sure that the value that you have previously inserted is at the next index

Braqzen (changes requested):
Looks good to me. I would add 1 more line in a test and make a minor change if the compiler allows it. After that I will approve

Braqzen:
> This is really good! Thank you so much for your contribution.
> 
> I don't really have any comments on the library code itself. It looks correct, clean, and readable.
> 
> As for the test, one thing that I'd like to also see is tests for the new functions called with data types other than `u8`. I suggest that for each of the existing test functions `test_vector_new_*()`, append some calls to `remove`, `insert`, and `pop` at the end just to double check that they work with complex data types.

Oh right, this too. I thought about this and then I got too stuck into the code so I forgot.

sway-lib-std/src/vec.sw
Braqzen: My preference is to remove as many indents as possible. I've seen the horrors that it could lead to so I'm generally put off when I see them and cannot get around them easily. 
Following rust is fine but if they write spaghetti then it shouldn't be followed imo

XFFXFF:
Thanks to both of you @mohammadfawaz @Braqzen , I read all your comments and they all seem reasonable to me. I'll change it tonight if I have time

XFFXFF:
```rust
let s0 = "fuel";
let s1 = "john";
let s2 = "nick";

vector.push(s0);
vector.push(s1);
vector.push(s2);

let val = vector.remove(0);
assert(sha256(val) == sha256(s0));
```
Strange, test for `str[4]` can not pass, any ideas? @mohammadfawaz @Braqzen

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
XFFXFF: Not possible now, this requires to implement `Eq` for `Option<T>`, which is blocked by trait constraints

mohammadfawaz:
> > I tested this out and used `log`. It looks like `val` is equal to `s1` for some reason.
> 
> image
> 
> `log(ret)` twice and it seems that `ret` changed during the loop of `copy`
> 
> image
> 
> https://github.com/FuelLabs/sway/blob/d0060d285f38bb027434b0e21bbe362ab0bd3b4e/sway-lib-std/src/mem.sw#L32-L44
> 
> Could it be because [str;4] is a reference type? Then why does `test_vector_new_array` pass?

That is indeed very weird. `b256` is also a reference type and works fine. The only difference with strings is where the padding bits go (right instead of left). I'm investigating.

mohammadfawaz:
So I think `let ret = read(ptr)` is simply giving you back a reference to the value you're reading and then when you do this shifting, that value is being overwritten which is why you get the new value which is `s1`. I'm thinking through the best way to fix this.

mohammadfawaz:
Ok I **_think_** I know what's going on. For strings that fit in a word (i.e. 8 characters or smaller), `let ret = read()` simply does a move instead of a copy. If you change the string type to `str[9]` or bigger, then the problem goes away. This needs a compiler fix which I can make in a separate PR. For the time being, please include an `str[9]` example and add a TODO with a comment saying that `str[8]` and smaller don't really work yet. I'll create a separate issue asap.

Thank you for catching this!

mohammadfawaz:
Alright, fix in review: https://github.com/FuelLabs/sway/pull/2189

XFFXFF:
> Ok I **_think_** I know what's going on. For strings that fit in a word (i.e. 8 characters or smaller), `let ret = read()` simply does a move instead of a copy. If you change the string type to `str[9]` or bigger, then the problem goes away. This needs a compiler fix which I can make in a separate PR. For the time being, please include an `str[9]` example and add a TODO with a comment saying that `str[8]` and smaller don't really work yet. I'll create a separate issue asap.
> 
> Thank you for catching this!

Thanks a lot @mohammadfawaz , I think I just need to update `test_vector_new_string` after #2189 lands

XFFXFF:
I think everything is ready, take a look @mohammadfawaz @Braqzen . It took me a lot of time to write tests =)

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
```diff
@@ -409,24 +817,233 @@ fn test_vector_new_enum() {
         },
         Option::None => revert(0), 
     }
-}
 
-fn test_vector_new_tuple() {
-    let mut vector = ~Vec::new();
+    vector.push(SimpleEnum::Y(b1));
+    vector.push(SimpleEnum::Y(b2));
```
Braqzen, line 822: You've pushed 2 elements so you should assert the before condition again to ensure that the removal works correctly here too

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
```diff
@@ -409,24 +817,233 @@ fn test_vector_new_enum() {
...
+
+    match vector.pop() {
+        Option::Some(val) => {
+            match val {
+                SimpleEnum::Y(b) => assert(b == b0),
+                _ => revert(0),
+            }
+        }, 
+        Option::None => revert(0), 
+    }
```
Braqzen, line 1012: Assert that your pop worked here too

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
```diff
@@ -409,24 +817,233 @@ fn test_vector_new_enum() {
...
+
+    match vector.pop() {
+        Option::Some(val) => {
+            match val {
+                SimpleEnum::X => {},
+                _ => revert(0),
+            }
+        }, 
+        Option::None => revert(0), 
+    }
```
Braqzen, line 1022: Assert that your pop worked here too

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
```diff
@@ -575,6 +1364,120 @@ fn test_vector_new_string() {
         },
         Option::None => revert(0), 
     }
+
+    vector.push(s3);
+    vector.push(s4);
+
```
Braqzen, line 1370: You've just pushed, assert the before conditions

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
```diff
@@ -623,6 +1528,196 @@ fn test_vector_new_array() {
         },
         Option::None => revert(0), 
     }
+
+    vector.push(a3);
+    vector.push(a4);
```
Braqzen, line 1533: Assert the before conditions here

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
```diff
@@ -704,4 +1799,4 @@ fn test_vector_with_capacity_u64() {
     match vector.get(5) {
         Option::Some(val) => revert(0), Option::None => (), 
     }
-}
+}
```
Braqzen, line 1802:
```suggestion
}

```

new line

Braqzen (changes requested):
Few more assertions to make sure it's all solid

emilyaherbert (commented):
I agree with @Braqzen suggested. Once those are in I will be ready to approve.

emilyaherbert:
This is an excellent addition @XFFXFF !

sway-lib-std/src/vec.sw
```diff
@@ -141,4 +142,70 @@ impl<T> Vec<T> {
...
+
+        // Increment length.
+        self.len += 1;
+    }
+
+    /// Removes the last element from a vector and returns it, or [`None`] if it
+    /// is empty.
+    fn pop(mut self) -> Option<T> {
+        if self.len == 0 {
+            return Option::None;
```
adlerjohn, line 206: There are some formatting issues, please format this file.
```suggestion
            return Option::None;
```

mohammadfawaz (approved):
Thank you again for working on this. This is great work!

Braqzen (approved):

PR #2173, opened 2022-06-29
Adds `ItemFn` formatting to `sway-fmt-v2`

### LINKED ISSUE #2077: Add `fn` formatting to `sway-fmt-v2`
Necessary for `abi`, `impl`, `trait` and `storage` formatting

### LINKED ISSUE #2245: Add `WhereClause` formatting to `sway-fmt-v2`

### LINKED ISSUE #2246: Add `Statement` formatting to `sway-fmt-v2`

### LINKED ISSUE #2247: Add `Pattern` formatting to `sway-fmt-v2`

### LINKED ISSUE #2248: Add `Path` formatting to `sway-fmt-v2`

### LINKED ISSUE #2256: Add `SameLineWhere` handling to `sway-fmt-v2`
I think we may benefit from having a field on `Shape` that stores a boolean value for the `SameLineWhere` case.

somehting like: 
```Rust
pub struct Shape {
// snip...
    /// Used in determining `SameLineWhere` formatting.
    /// Default is false.
    pub active_where_clause: bool,
}
```

Once this is added we can then do something like:
```Rust
match formatter.config.newline_style {
    SameLineWhere => {
        if formatter.shape.where_clause == true {
            // where formatting...
        } else {
            // normal formatting...
        }
    // snip...
}
```

eureka-cpu:
This PR doesn't take into account every scenario and is the bare minimum for handling `ItemFn`.

- [x] #2168 `ty`
- [x] Closes #2245 & Closes #2256
- [x] `*trait` <-- only spans
- [x] Closes #2246 
- [x] `*expr` <-- only spans
- [x] Closes #2247 
- [x] Closes #2248 
- [x] `*literal` <-- only spans
- [x] tests

Closes #2077


sway-fmt-v2/src/utils/where_clause.rs
```diff
@@ -0,0 +1,61 @@
...
+    fn format(
+        &self,
+        formatted_code: &mut FormattedCode,
+        formatter: &mut Formatter,
+    ) -> Result<(), FormatterError> {
+        writeln!(
+            formatted_code,
+            "\n{}{}",
+            &formatter.shape.indent.to_string(formatter),
+            self.where_token.span().as_str(),
```
kayagokalp, line 16: Here we may want to check `ItemBraceStyle::SameLineWhere` here and place the `where` according to that.

sway-fmt-v2/src/items/item_fn.rs
```diff
@@ -1,15 +1,165 @@
...
+    }
+}
+
+// For now this just pushes the char
+impl CurlyBrace for ItemFn {
+    fn open_curly_brace(
+        line: &mut FormattedCode,
+        _formatter: &mut Formatter,
+    ) -> Result<(), FormatterError> {
+        line.push(Delimiter::Brace.as_open_char());
```
kayagokalp, line 33: I will need to double-check but shouldn't we also alter the `shape`?

sway-fmt-v2/src/items/item_fn.rs
```diff
@@ -1,15 +1,165 @@
...
+        Ok(())
+    }
+    fn close_curly_brace(
+        line: &mut FormattedCode,
+        _formatter: &mut Formatter,
+    ) -> Result<(), FormatterError> {
+        line.push(Delimiter::Brace.as_close_char());
```
kayagokalp, line 40: same with the `open_curly_brace` case.

kayagokalp (commented):
I am going to be doing one more pass over this PR for a more detailed review but left some initial thoughts for discussion. It is coming around very nicely but I think we will need the tests (as you stated in the todo) for the parts that we are adding with this PR as it covers a lot of stuff.

sway-fmt-v2/src/utils/where_clause.rs
eureka-cpu: Right, this PR's scope is to get everything necessary for basic `fn` in. I think we can do this in a separate PR.

sway-fmt-v2/src/items/item_fn.rs
eureka-cpu: It's possible, I haven't added tests yet

sway-fmt-v2/src/utils/where_clause.rs
eureka-cpu: Also, `SameLineWhere` is described as: 
```rust
/// Prefer the same line except where there is a where-clause, in which
/// case force the brace to be put on the next line.
```

so this shouldn't be handled inside `where_clause`

eureka-cpu: It should actually be handled inside of `CurlyBrace::format()` I think

eureka-cpu: I think we may benefit from having a function on `Shape` that returns a boolean value for a `WhereClause` case. I'll give it a test and update it after this PR since it's already grown to this size.

sway-fmt-v2/src/items/item_fn.rs
eureka-cpu: This has to do with the above in matching the `ItemBraceStyle`

eureka-cpu:
@kayagokalp going to start on a few other pieces to this like `trait` and `expr` since I'm still waiting for `Ty` and as well

sway-fmt-v2/src/items/item_fn.rs
```diff
@@ -1,15 +1,206 @@
...
+            .clone()
+            .into_inner()
+            .format(formatted_code, formatter)?;
+        Self::close_curly_brace(formatted_code, formatter)?;
+
+        Ok(())
+    }
+}
+
+// For now this just pushes the char
```
kayagokalp, line 28:
```suggestion
// For now this just pushes the char
```
Is this comment still relevant, I feel like you have implemented logic for placing the where as well? Maybe we can get rid of this if it is not.

kayagokalp (approved):
Looks good to me 🚀 🚀. Just left a small nit regarding a comment which is not a blocker.

sway-fmt-v2/src/utils/pattern.rs
```diff
@@ -0,0 +1,116 @@
...
+                    .into_inner()
+                    .format(formatted_code, formatter)?;
+                Self::close_parenthesis(formatted_code, formatter)?;
+            }
+        }
+        Ok(())
+    }
+}
+
+// Currently these just push their respective chars, we may need to change this
```
mitchmindtree, line 59: W.r.t these `Parenthesis` and `CurlyBrace` traits, I agree it might be worth rethinking whether or not we want these as traits in general, particularly as we don't seem to be abstracting over their implementors in a generic way anywhere.

One way that we might be able to add some use to them could be to add a default method to the trait along the lines of:

```rust
trait Parenthesis {
    // open and close methods...

    /// Applies the given `format` function between a pair of open and close parentheses.
    fn parenthesised<F>(
        formatted_code: &mut FormattedCode,
        formatter: &mut Formatter,
        format: F,
    ) -> Result<(), FormatterError>
    where
        F: FnOnce(&mut FormattedCode, &mut Formatter) -> Result<(), FormatterError>,
    {
        self.open_parenthesis(formatted_code, formatter)?;
        format(formatted_code, formatter)?;
        self.close_parenthesis(formatted_code, formatter)
    }
}
```

This would allow us to do stuff like:

```rust
block.parenthesised(formatted_code, formatter, |formatted_code, formatter| {
    // format stuff inside block
})?;
```

where the open/close parens are automatically placed.

That said, it is still a bit awkward piping through the `formatted_code` and `formatter` and I'm not sure a fancy method like this is worth it?

Either way, these are considerations for a future PR - I think it's fine as is :)

mitchmindtree (approved):
Looks like after this we're getting very close to replacing V1!

sway-fmt-v2/src/utils/pattern.rs
eureka-cpu: That's kinda cool actually, I wouldn't mind looking further into that.

sway-fmt-v2/src/items/item_fn.rs
eureka-cpu: I'll get that out in my next PR (:

PR #2127, opened 2022-06-27
Warning about using zero address as ownership in SwayBook

bitzoic:
- Added warning of zero address to Access Control in Sway Docs
- Created new Ownership example
- Minor fix in Identity example

Closes #2126

PR #2174, opened 2022-06-29
Bump to v0.17.0

mohammadfawaz:
* Bump to v0.17.0
* Run `cargo update`

Waiting on:
- [x] https://github.com/FuelLabs/sway/pull/2189
- [x] https://github.com/FuelLabs/sway/pull/2112
- [ ] https://github.com/FuelLabs/sway/pull/2127
- [x] https://github.com/FuelLabs/sway/pull/2096
- [x] https://github.com/FuelLabs/sway/pull/2193


mitchmindtree:
I wonder if we should set a convention of trying to land a `cargo update` along with each of these releases, to make sure we're shipping with the latest security patches etc? If CI fails after `cargo update`, we can always go without.

mohammadfawaz:
> I wonder if we should set a convention of trying to land a `cargo update` along with each of these releases, to make sure we're shipping with the latest security patches etc? If CI fails after `cargo update`, we can always go without.

I think that's a very good idea. Should this be done with every release (including non-breaking ones) or just minor releases? Cc @adlerjohn @sezna @JoshuaBatty.

sezna:
My inclination is to say every minor release, to avoid the risk of transitive dependencies perhaps not upholding semver

sezna (approved):

mohammadfawaz:
> My inclination is to say every minor release, to avoid the risk of transitive dependencies perhaps not upholding semver

Makes sense. I had a similar feeling which is why I asked :)

mitchmindtree (approved):
> My inclination is to say every minor release, to avoid the risk of transitive dependencies perhaps not upholding semver

Yeah SGTM

JoshuaBatty:
> My inclination is to say every minor release, to avoid the risk of transitive dependencies perhaps not upholding semver

Also agree this is the correct approach.

PR #2178, opened 2022-06-29
Add the `SMO` opcode to `sway-core`

### LINKED ISSUE #2176: Add the `SMO` opcode to `sway-core`
## Summary

Related issue: #1868 
[Specification](https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/instruction_set.md#smo-send-message-to-output)

cc @nfurfaro @sezna @mohammadfawaz

Braqzen:
## Changes Made

1. sway-core
    1. allocated_ops.rs
        - Added to enum `AllocatedOpcode`
        - Added to Display trait for `AllocatedOpcode`
        - Added to `impl AllocatedOp` function `to_fuel_asm`
    2. mod.rs
        - Added to `impl Op` function `parse_opcode`
        - Added to Display trait for `Op`
    3. virtual_ops.rs
        -  Added to enum `VirtualOp`
        - Added to `impl VirtualOp` function 
            - `registers`
            - `use_registers`
            - `def_registers`
            - `update_register`
            - `allocate_registers`
2. sway-parse
    1. Defined opcode in `define_op_codes!()` from `op_code.rs`
3. smo_opcode (test)
    - Given that the client is not ready the test is only made to compile similar to the `TRO` opcode

## Related

Closes #2176


Braqzen:
Where should this be tested?

I swear that I did not simply copy and paste things without knowing what they are. 
I assure you dear reader that I am competent.

Braqzen:
12 failing checks huh. I have been duped. I will fight you tomorrow.

mohammadfawaz:
cargo fmt
cargo clippy
cargo run (in sway/test)
And then the stdlib tests which I think you know how to run.

Braqzen:
Looks like sway-core needs to be bumped to 0.6 to get the SMO opcode but that turns out to remove a bunch of other opcodes... That issue needs to come first...

Should I try that one or leave it to the compiler team?
@mohammadfawaz

mohammadfawaz:
> Looks like sway-core needs to be bumped to 0.6 (fuel-asm) to get the SMO opcode but that turns out to remove a bunch of other opcodes... That issue needs to come first...
> 
> 
> 
> Should I try that one or leave it to the compiler team?
> 
> @mohammadfawaz 

Do you want to just remove those opcode then? I don't think they are used in any useful way anywhere in the compiler or the tests. They have to be removed anyways. 

Cc @nfurfaro who wants to add `gtf` too.

nfurfaro:
The discussion we had in the app-dev sync was around adding these opcodes (`smo` and `gtf`) pre-emptively, knowing that they won't work until the upstream changes trickle down to the compiler. You could try bumping deps, I'm not sure if there are any blockers there or not. I'm ok to wait till the compiler is ready for these new instructions and then merge them.

Braqzen:
Note that the SMO is *UNTESTED* because I don't know how to test this.
I have disabled it for now, the TRO opcode is also disabled for some reason so despite disagreeing with this being disabled I thought that it might get picked up later.

emilyaherbert (commented):
> Skeleton test that is disabled (and panics when enabled because it is incorrect)

I might be missing some context, but could you explain why this is the case?

Braqzen:
> > Skeleton test that is disabled (and panics when enabled because it is incorrect)
> 
> I might be missing some context, but could you explain why this is the case?

Comment [here](https://github.com/FuelLabs/sway/pull/2178#issuecomment-1172619158)

emilyaherbert:
> Note that the SMO is UNTESTED because I don't know how to test this.

@Braqzen  do you not know how to test this because you are not familiar with the Sway compiler, or because it is not currently possible to write an accurate test? I'm trying to figure out if I would be able to think of test suggestions to offer 😄

Braqzen:
> > Note that the SMO is UNTESTED because I don't know how to test this.
> 
> @Braqzen do you not know how to test this because you are not familiar with the Sway compiler, or because it is not currently possible to write an accurate test? I'm trying to figure out if I would be able to think of test suggestions to offer smile

I don't understand the [spec](https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/instruction_set.md#smo-send-message-to-output) and how to implement the test for it

mohammadfawaz:
> > > Note that the SMO is UNTESTED because I don't know how to test this.
> 
> > 
> 
> > @Braqzen do you not know how to test this because you are not familiar with the Sway compiler, or because it is not currently possible to write an accurate test? I'm trying to figure out if I would be able to think of test suggestions to offer smile
> 
> 
> 
> I don't understand the [spec](https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/instruction_set.md#smo-send-message-to-output) and how to implement the test for it

Do we have all the changes from fuel-core and the VM for this to be functional? I suspect not. I wonder if we also have to bump their version up when they're ready.

Braqzen:
> > > > Note that the SMO is UNTESTED because I don't know how to test this.
> > 
> > 
> > > 
> > 
> > 
> > > @Braqzen do you not know how to test this because you are not familiar with the Sway compiler, or because it is not currently possible to write an accurate test? I'm trying to figure out if I would be able to think of test suggestions to offer smile
> > 
> > 
> > I don't understand the [spec](https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/instruction_set.md#smo-send-message-to-output) and how to implement the test for it
> 
> Do we have all the changes from fuel-core and the VM for this to be functional?

According to @Voxelot the client has some pending code for review for the SMO opcode.

Braqzen:
According to @adlerjohn I just need to get it to compile.

otrho:
It's not a great idea to add the empty test -- do you have a simple example of how it should be used, @nfurfaro?

adlerjohn (approved):
Can you file an issue to add a test for correct functionality once a new version of fuel-core is released? Until then, this will have to suffice since there's no way to actually test anything beyond "can the compiler emit this instruction."

otrho (approved):

nfurfaro:
>do you have a simple example of how it should be used

There's really no simple way to test this. The most straightforward solution is to add SDK tests, which makes more sense for testing a stdlib wrapper function around this opcode.

mohammadfawaz (approved):

PR #2182, opened 2022-06-30
#2177 - Use assert_eq instead of assert in sway-fmt-v2 tests

### LINKED ISSUE #2177: Use `assert_eq!` rather than `assert!` in `sway-fmt-v2` tests
> It's totally fine to do so for catching errors, but it's better to use assert_eq!(a, b) when asserting equality when possible. This is because the assert_eq! macro will actually display the left and right values in the failed test output, giving a better indication of why a test might be failing. On the other hand, assert! only shows the expression itself which is less useful.

posted by @mitchmindtree at https://github.com/FuelLabs/sway/pull/2124#discussion_r909090620

r-sitko:
closes #2177


kayagokalp (approved):
lgtm 🎉, will defer the final approval to @FuelLabs/tooling

PR #2185, opened 2022-06-30
Check for unknown file check directives and fix a broken test.

tritao:
This commit makes sure we emit an error when tests contain an unknown FileCheck directive, to prevent introducing broken tests to the tree.

Previously we just ignored unknown directives.

Instead of passing the entire test text to FileCheck, we now parse the test for directives and add it to the checker one-by-one, making sure to deal with errors.

For tests with unknown directives, this now emits:

```
thread 'main' panicked at 'Discovering tests
should_fail/match_expressions_rest/test.toml: Failed to parse: Unknown
FileCheck directive: nextnl', test/src/e2e_vm_tests/mod.rs:45:9
```

Also fixes a non-working test originally introduced in https://github.com/FuelLabs/sway/commit/2c5564b56eb1b15d1834fdb7ca164ebe2373d6b2, due to a simple typo.


otrho:
@vaivaswatha has used LLVM FileCheck and was using `check-next:` instead of `nextln:` inadvertently so this would've helped there too.

mohammadfawaz (dismissed):
Nice!

test/src/e2e_vm_tests/mod.rs
```diff
@@ -218,12 +219,24 @@ fn discover_test_configs() -> Result<Vec<TestDescription>, String> {
...
+                if let Ok(false) = checker.directive(&cap[0]) {
+                    return Err(format!("Unknown FileCheck directive: {}", &cap[1]));
+                }
+            }
+
+            // Then parse the file into TOML.
+            checker
                 .text(&toml_content_str)
                 .map_err(|e| e.to_string())
                 .and_then(|checker| {
```
otrho, line 242: GitHub won't let me highlight bits that aren't part of the diff, but you don't need to do this any more if you're adding the directives individually.

You only need to keep this:
```rust
                    toml_content_str
                        .parse::<toml::Value>()
                        .map_err(|e| e.to_string())
                        .map(|toml_content| (toml_content, checker.finish()))
```

In fact it could probably be simpilified some more.  It was originally a pipeline, hence all the `and_then`s etc, but now that `checker` is `mut` and built manually it can probably be done separately to parsing from TOML.

test/src/e2e_vm_tests/mod.rs
```diff
@@ -218,12 +219,24 @@ fn discover_test_configs() -> Result<Vec<TestDescription>, String> {
     Ok(configs)
 }
 
+const DIRECTIVE_RX: &str = r"\b#\s*(\w+):\s+(.*)";
```
otrho, line 222: This _could_ be stricter by insisting the comment is the only thing on the line, with `^\s*#...` but I guess in a TOML file it'd be pretty hard to actually be a mismatch.

otrho (changes requested):

tritao:
Addressed the feedback.

emilyaherbert (approved):

otrho (approved):

PR #2189, opened 2022-06-30
Rely on `is_copy_type` when compiling `load` and `store` into `asm`

mohammadfawaz:
Ref: https://github.com/FuelLabs/sway/pull/2170#issuecomment-1171327753

The impact of this is situations where loading a small string (or a small struct etc.) does not actually make a copy.. it instead just moves the address into a new variable. So, if we overwrite the original memory location (needed for `remove()` in `Vec`), the variable read is also overwritten.

The current behavior is also a bit inconsistent since small strings and large strings are treated differently for loads and stores even though we agreed that strings are always non-copy types regardless of their size. So this seems more sensible I think.


sezna (approved):
LGTM! :shipit:

otrho (approved):
Wasn't the small string copy string vs large ref string all about the ABI?  This shouldn't break that though?

mohammadfawaz:
> Wasn't the small string copy string vs large ref string all about the ABI?  This shouldn't break that though?

Yeah small strings and large strings are both reference types now so this aligns with what we agreed on for the ABI.

Issue #1673, opened 2022-05-25
Document how monomorphization works


mohammadfawaz:
I'm, of course, talking about some internal documentation which could be in the form of extensive comments in the codebase or a separate README file. 

I do find most things in the compiler to be self-explanatory for the most part except for monomorphization (and a few other things). Could be just me! But I find the implementation of monomorphization and its interaction with the type engine to be quite complex and hard to understand from just the code, even for someone who has read about monomorphization.

The recent changes by @emilyaherbert made the code cleaner and easier to read so that's a great step in the right direction. But concrete documentation would probably be extremely helpful for the current and future contributors.

emilyaherbert:
Relevant: https://github.com/FuelLabs/sway/pull/2191#discussion_r912583256

PR #2191, opened 2022-06-30
Turn `CallPath` into a wrapper type

### LINKED ISSUE #2018: Allow the use of type `Self` for struct expressions
This does work in some cases, but something like:
```rust
impl<T> RawVec<T> {
    fn new() -> Self { // Returning `Self` here is fine
        Self { // <- Problem is here.. Need to replace `Self` with `RawVec` for this to work.
            ptr: alloc(0),
            cap: 0,
        }
    }
}
```
does not work and gives `Could not find symbol "Self" in this scope`

emilyaherbert:
This PR introduces `CallPath<T>` where `suffix: T`. `CallPath<T>` is used to represent `CallPath<Ident>` and `CallPath<(TypeInfo, Span)>`, representing two different cases in which the suffix of the call path is either an ident pointing towards a declaration, or the name of a type.

- Blocked by #2201
- Closes #2018


sezna:
I know you're already kinda into this, but what about a `Spanned<T>` type to be consistent with the parser? E.g. `Spanned<TypeInfo>` is a `TypeInfo` with a span.

emilyaherbert:
> I know you're already kinda into this, but what about a Spanned<T> type to be consistent with the parser? E.g. Spanned<TypeInfo> is a TypeInfo with a span.

Wait @sezna, I'm not seeing how this would work... `Spanned` is a trait with the method `span`, it would not add any `span` field to `TypeInfo`.

sway-core/src/parse_tree/call_path.rs
```diff
@@ -5,16 +5,17 @@ use crate::Ident;
 use sway_types::{span::Span, Spanned};
 
 /// in the expression `a::b::c()`, `a` and `b` are the prefixes and `c` is the suffix.
+/// `c` can be any type `T`, but in practice `c` is either an `Ident` or a `TypeInfo`.
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
-pub struct CallPath {
+pub struct CallPath<T> {
```
mitchmindtree, line 10: I wonder if we could specify `Ident` as a default type here in the case that one isn't specified? E.g.

```suggestion
pub struct CallPath<T = Ident> {
```

This might help to easy the common case and cut down the diff a little?

emilyaherbert: Good idea!

sway-core/src/parse_tree/call_path.rs
```diff
@@ -24,19 +25,25 @@ impl std::convert::From<Ident> for CallPath {
...
+    T: fmt::Display,
+{
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let mut buf = String::new();
         for prefix in self.prefixes.iter() {
             buf.push_str(prefix.as_str());
             buf.push_str("::");
         }
-        buf.push_str(self.suffix.as_str());
+        buf.push_str(&self.suffix.to_string());
```
mitchmindtree, line 38: **_Hot tip_**: Not super important, but if `suffix` is a `T` where `T: Display`, you can do:

```rust
write!(&mut buf, "{}", self.suffix);
```
which avoids the extra `String` alloc from `to_string()`.

Alternatively, this whole `impl` should probably write directly into the `Formatter` with `write!(f, ...)`, but a change like that could probably be left for a separate PR.

sway-core/src/semantic_analysis/namespace/root.rs
```diff
@@ -64,34 +64,58 @@ impl Root {
...
+        self.resolve_type(
+            type_id,
+            span,
+            enforce_type_arguments,
+            type_info_prefix,
+            mod_path,
+        )
+    }
+
     pub(crate) fn resolve_type(
```
mitchmindtree, line 86: Perhaps this could be a good time to add some docs for `resolve_type` and `resolve_type_with_self`, particularly as we're adding a new `type_info_prefix` arg?

otrho:
Looks like there's some conflicts with the new IR tests being moved.  I found trying to resolve these could be a pain, I can help out if you like.

sway-core/src/semantic_analysis/namespace/root.rs
emilyaherbert: I added this to this issue: https://github.com/FuelLabs/sway/issues/1673

emilyaherbert:
@otrho the IR tests seem to pass right out of the box---which makes sense as there should be no behavior change other than introducing a new feature.

sway-core/tests/sway_to_ir/mutable_struct.ir
```diff
@@ -0,0 +1,31 @@
...
+    fn main() -> u64, !1 {
+        local mut ptr { u64, u64 } record
+
+        entry:
+        v0 = const { u64, u64 } { u64 undef, u64 undef }, !2
+        v1 = const u64 40, !3
+        v2 = insert_value v0, { u64, u64 }, v1, 0, !2
+        v3 = const u64 2, !4
+        v4 = insert_value v2, { u64, u64 }, v3, 1, !2
+        v5 = get_ptr mut ptr { u64, u64 } record, ptr { u64, u64 }, 0, !5
```
otrho, line 11: It looks like you're re-instating these files when they should be deleted?  Or am I reading this wrongly?  The `sway_to_ir` directory shouldn't exist on `master` any more.

sway-core/tests/sway_to_ir/mutable_struct.ir
```diff
@@ -0,0 +1,31 @@
+script {
```
mohammadfawaz, line 1: Toby moved all IR tests to https://github.com/FuelLabs/sway/tree/master/test/src/ir_generation and changed them to use `FileCheck`.

sway-lib-std/src/vec.sw
```diff
@@ -14,7 +14,7 @@ struct RawVec<T> {
 impl<T> RawVec<T> {
     /// Create a new `RawVec` with zero capacity.
     fn new() -> Self {
-        RawVec {
+        Self {
```
mohammadfawaz, line 17: Nice 😄

mohammadfawaz (commented):
Looks great! Just the comment about the IR tests.

sway-core/tests/sway_to_ir/mutable_struct.ir
emilyaherbert: Good eye---I didn't realize that the merge added these files back in.

sway-core/tests/sway_to_ir/mutable_struct.ir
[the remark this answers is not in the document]
emilyaherbert: Yes, this was a snafu with the merge.

mohammadfawaz (approved):

sway-core/src/semantic_analysis/node_dependencies.rs
```diff
@@ -574,7 +574,10 @@ impl Dependencies {
                 type_arguments,
             } => {
                 self.deps.insert(DependentSymbol::Symbol(name.clone()));
-                self.gather_from_type_arguments(type_arguments)
+                match type_arguments {
+                    Some(type_arguments) => self.gather_from_type_arguments(type_arguments),
+                    None => self,
+                }
```
otrho, line 580:
```suggestion
                type_arguments
                    .map(|type_arguments| self.gather_from_type_arguments(type_arguments))
                    .unwrap_or(self)
```

otrho (approved):

PR #2194, opened 2022-06-30
chore: use `concurrency` to cancel previous CI runs

AlicanC:
(This PR is a port of https://github.com/FuelLabs/fuels-ts/pull/361.)

From this SO page, it seems like the way we cancel previous runs is outdated and GitHub's native `concurrency` setting is the current practice: https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,21 +8,16 @@ on:
   release:
     types: [published]
 
+concurrency:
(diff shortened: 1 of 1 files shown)


mitchmindtree (approved):
Oooo nice, much cleaner, good spot!

sezna (approved):
nice change!

PR #2195, opened 2022-06-30
Introduce the `TypeBinding` type

### LINKED ISSUE #1628: First use of a generic method fixes its generic type

emilyaherbert:
This PR introduces the `TypeBinding` type, which represents the application of turbofish upon some LHS declaration or type, which will bind the type arguments of the turbofish to the type parameters in the LHS. This allows us to do things like:

```rust
struct Data<T> {
  value: T
}

impl<T> Data<T> {
  fn noop<F>(other: F) -> F {
    other
  }
}

let data = ~Data::<bool>::noop::<u64>(1u64);
```

previously, we were unable to have different generics between the data type and the internal methods.

- Closes #1628
- Blocked by #2191


otrho (approved):
I will admit I've been avoiding this review since it's big and I don't fully understand exactly what is going on, but it still looks good to me.

emilyaherbert:
> I will admit I've been avoiding this review since it's big and I don't fully understand exactly what is going on, but it still looks good to me.

I'd say about 50% of the green additions are from introducing a new test, hopefully that helps calm your feelings 😄

mohammadfawaz (approved):
The new syntax looks more correct and reasonable. The rest of the PR looks good! And the testing seems quite extensive.

PR #2196, opened 2022-07-01
const_eval can try (recursively) compile unresolved decls

### LINKED ISSUE #2151: Const evaluator fails when initializing a `const` with another `const`

vaivaswatha:
When the constant evaluator is executing, and it encounters a declaration that hasn't been added to "global_constants" yet, it can try and recursively compile that declaration.

Note that constant declarations are processed (const evaluated) in an arbitrary order and hence dependences aren't resolved. This change just dynamically resolves unmet dependences (when possible).

Closes #2151


sway-core/src/ir_generation/const_eval.rs
```diff
@@ -20,14 +22,66 @@ use sway_types::{ident::Ident, span::Spanned};
...
+    pub module: Module,
+    pub module_ns: Option<&'a namespace::Module>,
+    pub public_only: bool,
+    pub lookup: fn(&mut LookupEnv, &Ident) -> Result<Option<Value>, CompileError>,
+}
+
+pub fn compile_const_decl(
+    env: &mut LookupEnv,
+    name: &Ident,
+) -> Result<Option<Value>, CompileError> {
```
otrho, line 36: What does returning `Ok(None)` signify?  It succeeded but there's no value?

otrho (commented):
Looks good.  I'm leaning towards suggesting `LookupEnv` be a const evel context and all those functions which take a `lookup` be methods (which take a `&mut self`... so what's the difference?)

Is this still meant to be a draft?

sway-core/src/ir_generation/const_eval.rs
vaivaswatha: This function does two things
1. Provides a lookup.
2. Recursive entry to the actual const evaluator. 

In the case of a lookup not happening, it isn't the duty of this function to return an error, but just say `None`. In case the call to the actual evaluator fails, it passes on that error.

otrho (approved):

r-sitko:
@vaivaswatha 
By the way, if this PR resolves #2151 can it also make change in sway-lib-std/src/constants.sw to use ZERO_B256 const in BASE_ASSET_ID initialization? There is a TODO comment left to do this when #2151 is resolved.

r-sitko (commented):
By the way, if this PR resolves https://github.com/FuelLabs/sway/issues/2151 maybe it can also make change in sway-lib-std/src/constants.sw to use ZERO_B256 const in BASE_ASSET_ID initialization? There is a TODO comment left to do this when https://github.com/FuelLabs/sway/issues/2151 is resolved.

sezna (approved):

PR #2198, opened 2022-07-01
Check for calling of associated functions as method calls.

tritao:
This adds a new error and checking for calls of associated functions using method call syntax.

This means for code like the following:

```rust
struct Bar {}

impl Bar {
  fn associated() {}
}

fn main() -> u64 {
  let bar = Bar {};
  bar.associated();
  0
}
```

we now emit an error:

```rust
error
  --> main.sw:11:3
   |
 9 |
10 |   let bar = Bar {};
11 |   bar.associated();
|   ^^^^^^^^^^^^^^^^ Cannot call associated function "associated" as
method call. Use associated function syntax instead.
12 |   0
13 | }
   |
```

This brings us closer to Rust behaviour, which to my understanding also checks for this and only allows method call syntax for functions taking self.

It adds a few more tests for this and also updates some standard library trait methods to take self so they work with the new check.


sway-core/src/error.rs
```diff
@@ -514,6 +514,11 @@ pub enum CompileError {
         variable_name: Ident,
         span: Span,
     },
+    #[error(
+        "Cannot call associated function \"{fn_name}\" as method call. Use associated function \
```
sezna, line 518:
```suggestion
        "Cannot call associated function \"{fn_name}\" as a method. Use associated function \
```

test/src/e2e_vm_tests/test_programs/should_fail/associated_fn_as_method_call/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+license = "Apache-2.0"
+name = "associated_fn_as_method_call"
+entry = "main.sw"
+
+[dependencies]
+core = { path = "../../../../../../sway-lib-core" }
```
sezna, line 8: i think we can omit core here, since we don't use operators in this test. Will help with compile speed.

```suggestion
name = "associated_fn_as_method_call"
entry = "main.sw"
implicit-std = false
```

test/src/e2e_vm_tests/test_programs/should_fail/associated_fn_params_as_method_call/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+license = "Apache-2.0"
+name = "associated_fn_params_as_method_call"
+entry = "main.sw"
+
+[dependencies]
+core = { path = "../../../../../../sway-lib-core" }
```
sezna, line 8: Same thing here, for faster test compilation we can omit the dependenecy.
```suggestion
entry = "main.sw"
implicit-std = false
```

sezna (changes requested):
This is an awesome change and great for compiler UX. Thanks!

emilyaherbert (commented):
I agree with what Alex said, but other than that LGTM. When those changes go through I will approve.

mohammadfawaz (approved):

mohammadfawaz (approved):

sezna (approved):

PR #2201, opened 2022-07-01
Remove duplicate check to length of function arguments

emilyaherbert:
This PR:
- removes the duplicate check of arguments length inside of type checking for function call
- improves code quality of the type checking for method calls

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/expression/typed_expression/function_application.rs
+++ b/sway-core/src/semantic_analysis/ast_node/expression/typed_expression/function_application.rs
@@ -37,6 +37,7 @@ pub(crate) fn instantiate_function_application(
         });
     }
 
+    // check that the number of parameters and the number of the arguments is the same
     check!(
         check_function_arguments_arity(arguments.len(), &function_decl, &call_path),
         return err(warnings, errors),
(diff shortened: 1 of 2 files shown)


sway-core/src/semantic_analysis/ast_node/expression/typed_expression/function_application.rs
```diff
@@ -102,6 +98,7 @@ pub(crate) fn instantiate_function_application_simple(
     let mut warnings = vec![];
     let mut errors = vec![];
 
+    // check that the number of parameters and the number of the arguments is the same
     check!(
         check_function_arguments_arity(arguments.len(), &function_decl, &call_path),
```
sezna, line 103: could we just do this inside of `instantiate_function_application_inner()` instead of calling it beforehand in both callsites?

emilyaherbert: It depends on what we want to show users. Putting the check here aborts type checking before the arguments are type checked. If we wanted to make this check after type checking the arguments, then we could do it inside of `instantiate_function_application_inner`

sezna: Instead of `check!()` with `return err(...)`, could we aggregate the compile results and join them?

emilyaherbert: I'm not sure what you mean @sezna? In my head there are two scenarios when there are a different number of arguments and parameters:
1. We display one error message reporting the differing numbers
2. We display 1+ error messages---one reporting the different numbers and possibly many more coming from type checking the arguments.

I prefer option (1) because option (2) creates a situation in which we would only type check a subset of the arguments (or a subset of the parameters!).

emilyaherbert: If we want to go with option (1) but also move the check to inside of `instantiate_function_application_inner`, that would require us to keep track of multiple error vecs and we would still need to separately aggregate the error vecs in every place that `instantiate_function_application_inner` is called. IMO that would add more code complexity than it saves.

otrho (approved):
I like option 1.

mohammadfawaz (approved):

PR #2209, opened 2022-07-04
forc-pkg: Improve error handling around removed path deps and patches

mitchmindtree:
Previously, if a path dependency's directory was removed from its
expected `path` location, the produced error would refer to a "patch"
even if there was no patch being applied:

```
Error: Cant apply patch from ../child, cause: No such file or directory (os error 2)
```

This improves the error message for this case:

```
Error: Failed to source dependency: Failed to canonicalize dependency path "parent/../child": No such file or directory (os error 2)
```

Also improves the error handling around patches more generally. In
particular, this fixes a case where an error could be produced in the
case that a dependency's URL was patched, but its name wasn't found
within the patch table. Instead, we assume no patch is desired. This is
necessary to allow users to source multiple packages from a single
repository, and to allow for patching them individually.

*Spotted while testing #1974, though we should land that first.*

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -1068,7 +1068,8 @@ fn fetch_deps(
         .collect();
     for (dep_name, dep) in deps {
         let name = dep.package().unwrap_or(&dep_name).to_string();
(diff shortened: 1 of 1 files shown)


kayagokalp (approved):
lgtm

JoshuaBatty (approved):

PR #2210, opened 2022-07-04
Add `parsed_to_ast` fn in `sway_core`

JoshuaBatty:
This PR refactors `compile_to_ast` and adds a `parsed_to_ast` function. This avoids reparsing the whole program again in the language server when doing the type-checking phase as we can call `parsed_to_ast` with our already obtained `CompileResult<ParseProgram>`.

### DIFF
--- a/sway-core/src/lib.rs
+++ b/sway-core/src/lib.rs
@@ -206,28 +206,11 @@ pub enum BytecodeCompilationResult {
 }
 
 pub fn parsed_to_ast(
-    parsed: CompileResult<ParseProgram>,
+    parse_program: ParseProgram,
     initial_namespace: namespace::Module,
 ) -> CompileAstResult {
     let mut warnings = Vec::new();
     let mut errors = Vec::new();
-
-    let CompileResult {
-        value: parse_program_opt,
-        warnings: new_warnings,
-        errors: new_errors,
-    } = parsed;
-
-    warnings.extend(new_warnings);
-    errors.extend(new_errors);
-    let parse_program = match parse_program_opt {
-        Some(parse_program) => parse_program,
-        None => {
-            errors = dedup_unsorted(errors);
-            warnings = dedup_unsorted(warnings);
-            return CompileAstResult::Failure { errors, warnings };
-        }
-    };
 
     let CompileResult {
         value: typed_program_result,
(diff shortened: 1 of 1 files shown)


otrho (approved):
We have a couple of these little functions now, like `parsed_to_ast`, `compile_to_ast` and `compile_to_asm`.  We should look at cleaning up the public interface -- it seems every little step in the pipeline is useful to either tooling or testing, so they all might as well be publicly exposed.

sway-core/src/lib.rs
```diff
@@ -205,10 +205,9 @@ pub enum BytecodeCompilationResult {
     },
 }
 
-pub fn compile_to_ast(
-    input: Arc<str>,
+pub fn parsed_to_ast(
+    parsed: CompileResult<ParseProgram>,
```
mitchmindtree, line 209: This should probably be:

```suggestion
    parsed: ParseProgram,
```

To make this change, we'll also need to change the `compile_to_ast` body to:

```rust
    parse(input, build_config).flat_map(|parsed| {
        parsed_to_ast(parsed, initial_namespace)
    })
```

---

The `CompileResult`'s `flat_map` method should probably be called `and_then` to better align with the `std` `Option` and `Result` methods, but that's a future PR.

JoshuaBatty: But don't we still want to carry any errors and warnings across from the parsing stage?

mitchmindtree: The result returned from `sway_core::parse` should have already been handled by the user as they likely require access to the `ParseProgram` within the result (otherwise they would have just called `compile_to_ast` in the first place).

JoshuaBatty: Thanks @mitchmindtree I've changed `parsed_to_ast` to take a `ParseProgram` instead of a `CompileResult<ParseProgram>`.

I've also tried to retain the original behaviour of carrying forward any warnings and errors found during the parsing stage when calling `compile_to_ast`.

mitchmindtree (approved):
Agh I forgot `CompileAstResult` isn't just an alias for `CompileResult<TypedProgram>`, so you couldn't have used the `CompileResult::flat_map` method anyway.

:ship:

otrho:
Yeah, it gets a little complicated in `lib.rs` as we have 2 or 3? different 'result' types and error types.  Again, might be worth revisiting in a future refactoring.

otrho (approved):
You might be able to use `check!()` in `compile_to_ast` rather than destructuring the result from `parse()` but that doesn't let you dedup the errors... but that might happen later anyway?  I can't be bothered checking.  This is fine. :slightly_smiling_face:

PR #2211, opened 2022-07-04
Move all the `sway_to_ir` tests from `sway_core` to the integration tests.

otrho:
This puts all the IR tests from `sway-core` under `sway/test`.

It's switching to `FileCheck` for verification.  I've added the ability to have checks for IR and also for ASM, in the same file using markers.  But I haven't moved the `ir_to_asm` tests over yet, since this has taken me ages and I'm a bit sick of it.  Need to get this merged and take a break before tackling those.

It also provides an instance of `std::core` to the tests, which is especially useful for pattern matching enums which needs `core` for comparisons.


otrho:
I've had to make a bunch of things in `sway-core` `pub` instead of `pub(crate)`.  I was looking at how `forc` does it, with a `test` feature, and turning it on just for these tests, but it wasn't that easy.  You can't export modules twice and you can't give them arbitrary namespace paths to avoid collisions because they have to match the filesystem.

I actually didn't try _that_ hard, it may be possible, but it would probably require moving stuff around a bit.  Might be worth putting in an issue, but then again it might not matter.  Exporting most things from `sway-core` probably isn't the end of the world, because it's 'core' by definition.

test/src/main.rs
```diff
@@ -21,5 +22,6 @@ fn main() {
         filter_regex = Some(regex);
     }
 
-    e2e_vm_tests::run(locked, filter_regex);
+    e2e_vm_tests::run(locked, filter_regex.as_ref());
+    ir_generation::run(filter_regex.as_ref());
```
mohammadfawaz, line 26: Ah so these can be run with `cargo run` in `sway/test`. Nice.

mohammadfawaz (approved):
This is great. Thank you so much for spending the time on this. Much cleaner!

tritao:
Nice. :+1: 

One thing that comes to mind is if we should define the most common patterns up front in the runner (things like `// regex: VAL=v\d+`) instead of having them repeated in each test.

I can see advantages to both approaches, having them on the tests is more explicit/obvious and discoverable, so I am on the fence about it.

Either way looks good.

tritao (approved):

otrho:
> One thing that comes to mind is if we should define the most common patterns up front

Yeah, I was thinking the same thing -- every time I copied and pasted them from older tests!  If they're added automatically and documented in a README then it would be OK.  There isn't a single test that doesn't used the `VAL` one...

PR #2212, opened 2022-07-04
improvement(formatter): Enums and Structs are formatted to a single line first before getting their length to determine multiline formatting

### LINKED ISSUE #2138: Struct & Enum formatting in `sway-fmt-v2` should take formatted length into account while deciding single line formatting
Currently, to decide whether to format it into a single line or not, the formatter looks at the length of the unformatted enum/struct block. But it should actually get the length of the formatted (to a single line) version of that block and use it for deciding whether the single line construct would fit into the constraints we have for that line. Just like #2124 does it.

kayagokalp:
## About this PR
closes #2138.

Before this PR, while deciding on formatting type (`multi-line` or `single-line`), `sway-fmt-v2` was checking the raw char length of the structure for line width comparison. This would mean any extra whitespace in the input would cause the formatter to format things to multiline unnecessarily. 

For example:

```rust
pub enum Color{
    Blue:                                                                                               ()
}
```

Since the raw length of the input above would be 122 formatter may format that to the following depending on the line-width limit.

```rust
pub enum Color {
    Blue: ()
}
```

But normally it should be possible to format to a single line like the following.

```rust
pub enum Color { Blue: () }
```

This logic was already in-place for `storage` formatting at #2124 but this PR moves the `ItemLen` trait to `items.rs` as requested in https://github.com/FuelLabs/sway/issues/2138#issuecomment-1170199277

## TO DO
Before marking this ready to review, it would be great to land #2124 first.

- [x] Merge #2124


kayagokalp:
Since #2124  is merged now, we can have this small pr for improving the current logic of multi/single line formatting.

sway-fmt-v2/src/items/item_struct.rs
```diff
@@ -187,9 +187,11 @@ fn format_struct(
 
 impl ItemLen for ItemStruct {
     fn get_formatted_len(&self) -> Result<usize, FormatterError> {
```
eureka-cpu, line 189: Since this function is the same for each of its use cases, would you mind making it simply a `pub(crate)` function?

sway-fmt-v2/src/items/item_enum.rs
```diff
@@ -182,9 +182,11 @@ fn format_enum(
...
 impl ItemLen for ItemEnum {
     fn get_formatted_len(&self) -> Result<usize, FormatterError> {
-        // TODO while determininig the length we may want to format to some degree and take length.
-        let str_item = &self.span().as_str().len();
-        Ok(*str_item as usize)
+        // Format to single line and return the length
+        let mut str_item = String::new();
+        let mut formatter = Formatter::default();
+        format_enum(self, &mut str_item, &mut formatter, false)?;
+        Ok(str_item.len() as usize)
```
mitchmindtree, line 189: We likely want to use `chars().count()` rather than `len` here. The `len` method gives us the length in _bytes_, but in UTF-8 (which is what the `str` and `String` types represent) its possible for a single character to occupy more than one byte. The [str::len docs](https://doc.rust-lang.org/std/primitive.str.html#method.len) briefly mention this.

Using `str_item.chars().count()` will give the length in UTF-8 characters which is closer to what we're talking about when we're checking things like "line width".

We might want to consider changing the `ItemLen` trait to `ItemLenChars` and change its method to `len_chars` to make this a bit clearer. We should review the rest of the formatter code-base for uses of `str::len` too where this might apply.

Apologies for not spotting this sooner on the other implementations!

eureka-cpu: Ah, I also encountered this my last run in with the old formatter. I should've noticed as well.

kayagokalp: Ohh, I see. I am changing it rn and will look into the existing code base for other uses of `str::len` 👍

sway-fmt-v2/src/items/item_struct.rs
kayagokalp: I did follow the trait path because of two reasons:

1. To handle all of them with a single function I need to take `ItemKind` as a function parameter and then match for struct, enum, and storage cases to call the respective format functions. This function can falsely imply that it will also return the lengths of other kinds of `ItemKind`. 
2. We may want to use other implementations later on for other `ItemKinds`.

That being said, I can also use a single function for this purpose and document it (regarding the 1st point), if you think that should be the way to go.

mitchmindtree (approved):

PR #2214, opened 2022-07-04
Add a `tx_id()` function to std::tx

### LINKED ISSUE #1629: Create a `tx_id()` std library function
The transaction ID is simply at address zero: https://github.com/FuelLabs/fuel-specs/blob/0d56640183c3fc1eee055ab380f796565955a973/specs/vm/main.md#vm-initialization

nfurfaro:
Adds a simple wrapper function to get the tx id.
Also adds a test for this.

closes #1629


sway-lib-std/src/tx.sw
```diff
@@ -302,3 +303,10 @@ pub fn tx_output_amount(index: u64) -> u64 {
         r1: u64
     }
 }
+
+/// Get the id of the current transaction.
+pub fn tx_id() -> b256 {
+    asm() {
```
mohammadfawaz, line 309: Can this `asm` block be replaced with `read(VM_ADDRESS_ZERO)`?

sway-lib-std/src/tx.sw
```diff
@@ -43,6 +43,7 @@ const TX_OUTPUTS_COUNT_OFFSET = 10304;
 const TX_WITNESSES_COUNT_OFFSET = 10312;
 const TX_RECEIPTS_ROOT_OFFSET = 10320;
 const TX_SCRIPT_START_OFFSET = 10352;
+const VM_ADDRESS_ZERO = 0;
```
mohammadfawaz, line 46: Maybe change this to `TX_ID_OFFSET = 0` following the format of the other consts? It is possible that the transaction ID may be stored at a different location in the future (though unlikely I suspect).

nfurfaro: Sure. 
A big change coming soon (TM) will be the change to using the `GTF` opcode rather than hardcoded offsets, but I will modify this for the time being.

mohammadfawaz: Oh nice.. I didn't know if `GTF` handles tx id or not but I guess it does!

nfurfaro: Now I'm second_guessing myself 😆 
But I will make this change regardless.

nfurfaro: image

sway-lib-std/src/tx.sw
nfurfaro: yep!

adlerjohn (approved):

sway-lib-std/src/tx.sw
adlerjohn: No that's not tx ID, that's the transaction ID part of the pointer for an input. `GTF` only handles tx fields, not the transaction ID. It could be added to `GM` though!

mohammadfawaz (approved):
Looks good! Thanks.

Issue #1816, opened 2022-06-02
Support matching on constants in pattern matching


emilyaherbert:
We should support matching on constants directly in pattern matching.

For example:

```rust
contract;

use std::{address::Address, chain::auth::{AuthError, Sender, msg_sender}, result::*};

abi MyContract {
    fn test_function() -> bool;
}

const ORACLE_1: b256 = 0x6b63804cfbf9856e68e5b6e7aef238dc8311ec55bec04df774003a2c96e0418e;
const ORACLE_2: b256 = 0x54944e5b8189827e470e5a8bacfc6c3667397dc4e1eef7ef3519d16d6d6c6610;
const ORACLE_3: b256 = 0xe10f526b192593793b7a1559a391445faba82a1d669e3eb2dcd17f9c121b24b1;

impl MyContract for Contract {
    fn test_function() -> bool {
        let sender: Result<Sender, AuthError> = msg_sender();
        let is_one_of_signers = match sender.unwrap() {
            Sender::Address(address) => {
                match address.value {
                    ORACLE_1 => true, 
		    ORACLE_2 => true, 
                    ORACLE_3 => true, 
                    _ => false, 
                }
            },
            _ => {
                false
            },
        };
        is_one_of_signers
    }
}
```

inspired by: https://github.com/FuelLabs/sway/issues/1765

emilyaherbert:
Some notes for the future:

At parse time, constants on the LHS of a match arm (i.e. `ORACLE_1 => true`) could be parsed to a `Scrutinee::Variable { name: "ORACLE_1" }`. Each constant will need to be converted into one of the literal `Pattern`'s:
https://github.com/FuelLabs/sway/blob/057e83aee87d17c3812f223f5fc7884cf6d3468a/sway-core/src/semantic_analysis/ast_node/expression/match_expression/analysis/pattern.rs#L101

using this method:
https://github.com/FuelLabs/sway/blob/057e83aee87d17c3812f223f5fc7884cf6d3468a/sway-core/src/semantic_analysis/ast_node/expression/match_expression/analysis/pattern.rs#L125

with a lookup into the namespace to grab the `value` of the constant:
https://github.com/FuelLabs/sway/blob/e57de1b4b90f6626f69b7f972e9cef28fcceba9a/sway-core/src/semantic_analysis/ast_node/declaration.rs#L439

[...]

PR #2215, opened 2022-07-04
Add `std::tx::tx_input_owner`

### LINKED ISSUE #2056: Fix bug in std::auth::get_coins_owner()
The changes needed are:
1. a pre-check on the value of inputs_count:  i.e: `assert(inputs_count < 1);`
2. return a Result<AuthError::NoCoinInputs> from the code path where `inputs_count` is set to 1, but there are no `InputCoins`.
This leads to the while loop exiting, but `candidate` still being `Option::None`. Then, when trying to return a `Result::Ok` value the function will revert when unwrapping `candidate`.

Discovered by @emilyaherbert 
cc @adlerjohn

nfurfaro:
This adds a new function `get_input_owner()`.
- it replaces the former `get_input_coin_owner`, while also handling `InputMessage`s.
- rather than "undefined behavior" in some cases, `get_input_owner()` now returns an `Option<Address>`.
- refactored to accept an index rather than a pointer, building on @simonr0204 's recent work.
- refactors `std::auth::get_coins_owner()` to use the new `get_input_owner()` function.
- adds a new check to `get_coins_owner as discussed in #2056 to ensure we don't return an Option::None and inadvertantly cause a revert. This last item may not be needed as per @adlerjohn 's comments on 2056. (it's also tangential to this PR so I may remove it either way, but will leave it in for now to gather some feedback).

closes #2056


nfurfaro:
Oops, need to update the test contract.

adlerjohn:
converting to draft until ready

sway-lib-std/src/chain/auth.sw
```diff
@@ -61,28 +62,33 @@ fn get_coins_owner() -> Result<Identity, AuthError> {
...
             };
-        };
+        }
     }
 
+    // we want to return an error here rather than reverting
+    if candidate.is_none() {
+        return Result::Err(AuthError::NoCoinOrMessageInputs);
+    };
+
```
nfurfaro, line 91: I might be able to remove this, as _I think_ the only way `candidate` can be `None` here is on an invalid tx (which shouldn't happen. (i.e: a tx with only a single input which is **not** an `InputCoin`).

sway-lib-std/src/chain/auth.sw
```diff
@@ -49,10 +50,10 @@ pub fn msg_sender() -> Result<Identity, AuthError> {
 fn get_coins_owner() -> Result<Identity, AuthError> {
```
simonr0204, line 50: Orthogonal to the actual change, but should we drop the `get_` here as we did for the other function in this lib?

sway-lib-std/src/chain/auth.sw
simonr0204: A transaction must have at least one inputCoin for fees, so I agree that you can safely remove this.

test/src/sdk-harness/test_artifacts/tx_contract/src/main.sw
```diff
@@ -18,7 +19,8 @@ abi TxContractTest {
     fn get_tx_script_start_offset() -> u64;
 
     fn get_tx_input_pointer(index: u64) -> u32;
-    fn get_tx_input_type(ptr: u32) -> u8;
+    fn get_tx_input_type_from_ptr(ptr: u32) -> u8;
```
simonr0204, line 22: I saw a PR from @AlicanC earlier today where all pointers were changed to `u64`. Should that be the same here ?

sway-lib-std/src/chain/auth.sw
nfurfaro: yes!

test/src/sdk-harness/test_artifacts/tx_contract/src/main.sw
nfurfaro: I have no objections, consistency is good. Any obvious benefits to switching?

adlerjohn: +1

sway-lib-std/src/chain/auth.sw
adlerjohn: Please file an issue!

sway-lib-std/src/chain/auth.sw
adlerjohn: +1. If you remove this though then there's no need for `AuthError::NoCoinOrMessageInputs`, right?

nfurfaro: correct, I'll be removing that error variant.

adlerjohn (commented):
merge conflicts

sway-lib-std/src/chain/auth.sw
nfurfaro: done: 
https://github.com/FuelLabs/sway/issues/2219

sway-lib-std/src/chain/auth.sw
```diff
@@ -61,28 +61,28 @@ fn coins_owner() -> Result<Identity, AuthError> {
...
-                // `candidate` and `input_owner` must be `Option::Some` at this point,
-                // so can unwrap safely.
+                // `candidate` and `input_owner` must be `Option::Some`
+                // at this point, so we can unwrap safely.
                 if input_owner.unwrap() == candidate.unwrap() {
                     // Owners are a match, continue looping.
                     i += 1;
                 } else {
                     // Owners don't match. Return Err.
+                    // TODO: Use break keyword when possible
```
mohammadfawaz, line 78: We can use `break` now 😄

sway-lib-std/src/tx.sw
```diff
@@ -6,6 +6,7 @@ use ::address::Address;
 use ::context::registers::instrs_start;
 use ::contract_id::ContractId;
 use ::intrinsics::is_reference_type;
+use ::option::Option;
```
mohammadfawaz, line 9: Maybe move this under `::mem` to keep in alphabetical order?

sway-lib-std/src/tx.sw
```diff
@@ -195,6 +196,32 @@ pub fn tx_input_type_from_pointer(ptr: u64) -> u8 {
     }
 }
 
+/// If the input's type is `InputCoin` or `InputMessage`,
+/// return the owner as an Option::Some(owner).
+/// Otherwise, returns Option::None.
+pub fn tx_input_owner(index: u64) -> Option<Address> {
+    let type = tx_input_type(index);
+    let owner_offset = match type {
+        0u8 => {
```
mohammadfawaz, line 205: Are there constants representing these types that we could use here? Hardcoded integers are a bit harder to to follow.

test/src/sdk-harness/test_artifacts/tx_contract/src/main.sw
```diff
@@ -2,6 +2,7 @@ contract;
 
 use std::address::Address;
 use std::tx::*;
+use std::option::Option;
```
mohammadfawaz, line 5: Alphabetize.

sway-lib-std/src/chain/auth.sw
```diff
@@ -61,28 +61,28 @@ fn coins_owner() -> Result<Identity, AuthError> {
             i += 1;
         } else {
             // type == InputCoin
-            let input_owner = Option::Some(tx_input_coin_owner(i));
+            let input_owner = tx_input_owner(i);
             if candidate.is_none() {
```
mohammadfawaz, line 65: Not to be done in this PR necessarily, but I feel a lot of this if/else nesting can be cleaned up a bit using `match`.

sway-lib-std/src/tx.sw
nfurfaro: can't wait for forc-fmt to handle this!

sway-lib-std/src/tx.sw
mohammadfawaz: Yeah you're right... Looks like https://github.com/FuelLabs/sway/issues/1816 is still open.

sway-lib-std/src/tx.sw
mohammadfawaz: https://github.com/FuelLabs/sway/issues/2226

sway-lib-std/src/chain/auth.sw
nfurfaro: I'll open an issue.

sway-lib-std/src/tx.sw
simonr0204: Can we match on `const` variables though?  https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/tx.sw#L49

sway-lib-std/src/chain/auth.sw
nfurfaro: #2237

PR #2216, opened 2022-07-04
Swap order of operands for `std::mem::copy`

### LINKED ISSUE #2203: Swap order of operands for `std::mem::copy`
https://github.com/FuelLabs/sway/blob/850e0de4434d412f458692760fd0de019ed03687/sway-lib-std/src/mem.sw#L18

Rust's `std::ptr::copy` uses source then dest, instead of the more C-link dest then source. https://doc.rust-lang.org/std/ptr/fn.copy.html

AlicanC:
Swapped `src` and `dst` args in `std::mem::copy` so `src` comes first:
```rs
pub fn copy(src: u64, dst: u64, size: u64) {
    // ...
}
```

Resolves #2203.

### DIFF
--- a/sway-lib-std/src/alloc.sw
+++ b/sway-lib-std/src/alloc.sw
@@ -38,7 +38,7 @@ pub fn realloc(ptr: u64, size: u64, new_size: u64) -> u64 {
     if new_size > size {
         let new_ptr = alloc(new_size);
         if size > 0 {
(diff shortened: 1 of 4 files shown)


mohammadfawaz (approved):
:shipit:

adlerjohn (approved):

Issue #1993, opened 2022-06-15
Emit type arguments for generic structs/enums in the JSON ABI file and update the function selector to include those type arguments


mohammadfawaz:
This is needed to support `Vec<T>` in the ABI. The struct `Vec` does not have data fields that actually use the type `T`. Therefore, the ABI won't currently have that type information that is needed by the SDK. This needs to change. Spec changes to be made as well once we're done spec-ing this out.

Note that unresolved generic types in the ABI are not allowed. However, something like this is allowed:
```rust
abi Foo {
    fn foo(v: Vec<u64>); // This `u64` should show up in the JSON ABI file
}
```

We may also need to update the function selector to include the type information above. This is probably needed to be consistent with how we handle non-generic types. `Vec<u64>` is a totally different type from `Vec<bool>`.

mohammadfawaz:
Here's how this might look like in the general case:
```rust
contract;

struct Foo<T, U> {
}

struct S {
    x: u64,
    y: u64,
}

abi MyContract {
    fn test_function(f: Foo<str[3], S>) -> bool;
}

impl MyContract for Contract {
    fn test_function(f: Foo<str[3], S>) -> bool {
        true
    }
}
```

JSON ABI:
```json
[
  {
    "type": "function",
    "inputs": [
      {
        "name": "f",
        "type": "struct Foo",
        "components": [],
        "type_arguments": [
          {
            "name": "T",
            "type": "str[3]",
            "components": null,
            "type_arguments": null
          },
          {
            "name": "U",
            "type": "struct S",
            "components": [
              {
                "name": "x",    
                "type": "u64",
                "components": null
                "type_arguments": null
              },
              {
                "name": "y",    
                "type": "b256",
                "components": null
                "type_arguments": null
              }
            ],
            "type_arguments": [] 
          }
        ]
      }
    ],
    "name": "test_function",
    "outputs": [
      {
        "name": "",
        "type": "bool",
        "components": null
        "type_arguments": null
      }
    ]
  }
]
```
Note the new field `"type_arguments"` which will be 
* `[]` for structs and enums that don't have type parameters.
* `null` for types that are not structs or enums.
* Non-empty array otherwise.

Thoughts? @digorithm @adlerjohn

digorithm:
@mohammadfawaz That looks reasonable to me, though implementation problems/challenges may or may not arise once we're actually implementing it on the SDK side. But at first glance, it looks simple enough!

PR #2218, opened 2022-07-04
Type arguments in JSON ABI and the function selector

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/1993

Spec change in https://github.com/FuelLabs/fuel-specs/pull/382

Mostly similar to how `components` work. 

I had to switch to using copies of `Property` and `Function` from `fuels-types` to local versions in `sway-types` because `type_arguments` still need to be added in `fuels-types`. This is temporary until `fuels-types` is ready.


otrho (approved):

adlerjohn:
Is there an issue in `fuels-rs` to track adding that to `fuels-types`?

otrho (approved):

tritao (approved):

sway-core/src/type_engine/type_argument.rs
```diff
@@ -59,6 +58,10 @@ impl ToJsonAbi for TypeArgument {
             name: "__tuple_element".to_string(),
             type_field: self.type_id.json_abi_str(),
             components: self.type_id.generate_json_abi(),
+            type_arguments: self
+                .type_id
+                .get_type_parameters()
+                .map(|v| v.iter().map(|param| param.generate_json_abi()).collect()),
```
sezna, line 64: i'm running out of creativity but the same code golf could apply here

PR #2220, opened 2022-07-05
Return a `CompileResult` & a `CompileAstResult` from `forc_pkg::check`

JoshuaBatty:
This PR allows `forc_pkg::check` to return a tuple of `CompileResult` and `CompileAstResult`. 

In the language server, we need access to both types, this change improves the ergonomics of needing access to both types and eliminates the need to call `sway_core::parse` in the language server which then would need to pass the returned `ParseProgram`  into `forc_pkg::check`.

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -1721,39 +1721,48 @@ pub fn build(plan: &BuildPlan, profile: &BuildProfile) -> anyhow::Result<(Compil
 }
 
 /// Compile the entire forc package and return a CompileAstResult.
-pub fn check(plan: &BuildPlan, silent_mode: bool) -> anyhow::Result<(CompileResult<ParseProgram>, CompileAstResult)> {
-    let profile = BuildProfile {
-        silent: silent_mode,
-        ..BuildProfile::debug()
-    };
+pub fn check(
+    plan: &BuildPlan,
+    silent_mode: bool,
+) -> anyhow::Result<(CompileResult<ParseProgram>, CompileAstResult)> {
     let mut namespace_map = Default::default();
     let mut source_map = SourceMap::new();
     for (i, &node) in plan.compilation_order.iter().enumerate() {
         let dep_namespace = dependency_namespace(&namespace_map, &plan.graph, node);
         let pkg = &plan.graph[node];
         let manifest = &plan.manifest_map()[&pkg.id()];
+        let parsed_result = parse(manifest, silent_mode)?;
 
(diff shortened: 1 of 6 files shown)


mohammadfawaz (approved):
Compiler changes look good! I'll defer to @FuelLabs/tooling for the other files.

forc-pkg/src/pkg.rs
```diff
@@ -1661,17 +1661,25 @@ pub fn check(
...
         let manifest = ManifestFile::from_dir(path, sway_git_tag)?;
-        let ast_res = compile_ast(&manifest, &profile, dep_namespace)?;
-        if let CompileAstResult::Success { typed_program, .. } = &ast_res {
-            if let TreeType::Library { .. } = typed_program.kind.tree_type() {
-                namespace_map.insert(node, typed_program.root.namespace.clone());
+
+        let source = manifest.entry_string()?;
+        let sway_build_config =
+            sway_build_config(manifest.dir(), &manifest.entry_path(), &profile)?;
+        let parsed_result = sway_core::parse(source, Some(&sway_build_config));
```
mitchmindtree, line 1668: Maybe we should refactor these 4 lines into a `parse` function that we can publicly export from `forc_pkg`?

Something like:

```rust
pub fn parse(manifest: &ManifestFile, silent: bool) -> CompileResult<ParseProgram> { ... }
```
^ I can imagine this might be a useful short-hand for the formatter too.

forc-pkg/src/pkg.rs
```diff
@@ -1661,17 +1661,25 @@ pub fn check(
...
+
+        if let Some(parse_program) = &parsed_result.value {
```
mitchmindtree, line 1670: If parsing has failed and this package is *not* the project package but is one of its dependencies, we should return an error. We cannot continue compilation if we cannot parse the dependency.

Currently, it looks like if a dependency fails to parse this will `panic!` on a following iteration when we attempt to retrieve the namespace for the package (which won't exist as we could not type check it).

mitchmindtree (changes requested):

forc-pkg/src/pkg.rs
JoshuaBatty: Good call, i'll make this change.

JoshuaBatty:
Thanks @mitchmindtree i've made those changes now, should be good to go.

mitchmindtree (approved):
Perfect, nice one bruv!

sezna (approved):

PR #2221, opened 2022-07-05
Switch the check if an IR block is terminated by a branch to just be a check if terminated at all.

### LINKED ISSUE #2150: ICE for a while loops containing a `return` statement
Example:
```rust
script;

fn main() -> bool {
    while true {
        return false;
    }

    true
}
```
Error:
```console
error: Internal compiler error: Block while_body has a misplaced terminator.
```
The code above should compile with no issues.

otrho:
I noticed low hanging fruit in #2150 and realised that everywhere we were checking if a block was terminated by a branch would actually care if the block was already terminated at all.  We're now essentially checking if a branch is already terminated before adding a new terminator.

Closes #2150.


mohammadfawaz (approved):
Very cool! Thanks.

mitchmindtree (approved):

Issue #2217, opened 2022-07-04
EvmAddress::from leads to memory overflow error


nfurfaro:
I originally raised this issue in the SDK channel, and even opened an issue for it (since closed).
It turns out this occurs even when not using the SDK, as discovered by @segfault-magnet .
He created a minimal repro without the SDK here: https://github.com/segfault-magnet/memory-overflow-issue/tree/42ebe730d219da50ade0f12c5994cea463ec72a3

mohammadfawaz:
So this works fine:
```
impl EvmTest for Contract {
    fn test_function() -> EvmAddress {
        let x = 0x0000000000000000000000000000000000000000000000000000000000000000;
        ~EvmAddress::from(x)
    }
}
```
which tells me that this is the same problem as https://github.com/FuelLabs/sway/issues/1455 but a bit worse given that the above is a bad workaround.

mohammadfawaz:
Because this is in the standard library and is critical to fix, I propose the following fix to `from()` which seems to work fine:
```rust
/// Functions for casting between the b256 and Address types.
impl From for EvmAddress {
    fn from(bits: b256) -> EvmAddress {
        // An EVM address is only 20 bytes, so the first 12 are set to zero
        let mut local_bits = bits;
        asm(r1: local_bits) {
            mcli r1 i12;
        };

        EvmAddress {
            value: local_bits,
        }
    }
}
```
In fact, the fact that we're changing `bits` in the original code is slightly incorrect because `bits` is not passed as `mut`. Changing that (once `mut` is supported for all function arguments) won't solve the problem though. The root cause of the problem is that we're trying to change a constant using `mcli`.

PR #2227, opened 2022-07-05
Fixing `from` in `evm_address` to make a `mut` local copy of the `bits` argument before calling `mcli`

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/2217

The test I added is not meant to be comprehensive for the `evm` library at all.


test/src/sdk-harness/test_projects/evm/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+    )
+    .await
+    .unwrap();
+    let instance = EvmTestContract::new(id.to_string(), wallet);
+
+    (instance, id)
+}
+
+#[tokio::test]
+async fn can_call_from() {
```
Braqzen, line 23: This should be split into 2 tests. How do you know which one will fail if this test fails? They are two separate functions. This makes it harder to debug

Braqzen (changes requested):

test/src/sdk-harness/test_projects/evm/mod.rs
mohammadfawaz: Done.

Braqzen (approved):

nfurfaro (approved):

Issue #2224, opened 2022-07-05
Collect positional data to comment map for sway-fmt-v2: comment formatting


kayagokalp:
This issue is for tracking a part of #2122.

kayagokalp:
closed by #2229

Issue #2122, opened 2022-06-27
Handle comments in `sway-fmt-v2`


mitchmindtree:
This is a dedicated issue for tracking the handling of comments in `sway-fmt-v2`, part of #1516. I'll add thoughts/ideas here as I progress.

---

Currently, `sway-fmt-v2` works by using `sway_parse` to parse the file into a syntax tree. This is useful as it allows us to take advantage of the same parsing that the compiler uses and trivially apply formatting based on the item types.

Unfortunately, comments are omitted during the parsing stage, as they are irrelevant to the compilation of the program. In `sway-fmt-v2`'s case this means we end up ignoring user comments and the current implementation deletes them entirely.

## Retrieving spans

This problem is discussed in #1517. We can at least start by retrieving the Spans during lexing, however we likely still do not want to include them within the syntax tree.

## Interleaving comments

Without having comments in the syntax tree itself, we'll need a general way of interleaving the comments into the syntax tree traversal so that we know if items we are formatting precede/follow/intersect comments. Comments can appear *anywhere*, including in the middle of a leaf node's span (e.g. between a function argument and the next comma punctuation).

One way of approaching this might be to add a `Span::intersects` method that allows for testing whether a span overlaps another. One way to achieve this could be to use a `BTreeMap<Span, Comment>`, and lookup this map with the item's span as we traverse the AST to check whether or not we intersect any comments.

This requires `Span` being `Ord`, which currently is not the case. One `Ord` implementation for `Span`s that could make sense is to order them by their AST visitation order. In other words, enclosing spans should always precede sub-spans. Handling file paths in the `Ord` implementation is less clear. Potentially, this should be split into a separate `FileSpan` type to separate this concern from the `Span` itself.

## Preserving whitespace

The whitespace between a comment's span and the following/preceding item can also be intentional and should likely be preserved to some extent. E.g. if a comment follows an item on the same line of text, we don't want to push the comment below by inserting a newline after the item.

## MVP requirements

As a very initial implementation, we should not attempt to do any reformatting of the comments themselves, but instead simply preserve the comments as they are. For comments that exist *between* AST nodes, we can start by adding them them verbatim with the same surrounding whitespace as what they had previously. We can think about fancier comment handling / rearranging of comments after crossing the MVP line and replacing V1.

---

## TODO

[...]

PR #2229, opened 2022-07-05
Add means of comment formatting to `sway-fmt-v2`

kayagokalp:
~Opening up for visibility. This is highly WIP.~

If we decide to have the actual formatting in a separate PR, this is ready to review. 

Related to #2224 and #2122.

## To Do

- [x] Handle comments inside a code block
- [x] Improve tests


sway-fmt-v2/src/utils/comments.rs
```diff
@@ -0,0 +1,78 @@
...
+pub struct CommentSpan {
+    // The byte position in the string of the start of the span.
+    start: usize,
+    // The byte position in the string of the end of the span.
+    end: usize,
+}
+
+impl Ord for CommentSpan {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.start.cmp(&other.start)
```
mitchmindtree, line 17: When it comes to ordering spans, I think the correct approach might be to reflect not just the order in which the `start` occurs, but also the order in which the spans would be visited during traversal (e.g. larger enclosing spans preceding inner spans.)

E.g. when comparing the spans `2..6` and `2..4`, the `2..6` span should come first as it encloses the `2..4` span. I think this would look something like:
```rust
match self.start.cmp(&other.start) {
    Ordering::Equal => other.end.cmp(&self.end),
    ord => ord,
}
```

Anyway, this is more relevant to the ordering of AST spans generally, and shouldn't be relevant to our case as I suppose comments cannot overlap :joy:

kayagokalp: I was actually thinking about this while writing that part but later on couldn't think of an example where comments would be overlapping as you pointed out 😄. But I will fix the ordering like this. Although it seems like overlapping won't be happening, if somehow it happens I feel like this will be really hard to debug.

kayagokalp:
This needs couple more touches but at the current status it can format the following code piece:

```rust
contract;

// This is a comment.
struct Foo {
    barasdd: u64,
    baz: bool,
}









        // This is the second comment.
struct Foo2 {
    barasdd: u64,
    baz: bool,
}

```

to 

```rust
contract;

// This is a comment.
struct Foo {
    barasdd: u64,
    baz: bool,
}
// This is the second comment.
struct Foo2 {
    barasdd: u64,
    baz: bool,
}
```

Currently, it cannot handle the inside code-block comment correctly (although they are correctly placed in the comment map). I should have that shortly and after some final clean-ups, this will be ready to go.

sway-fmt-v2/src/utils/comments.rs
```diff
@@ -14,7 +14,12 @@ pub struct CommentSpan {
 
 impl Ord for CommentSpan {
     fn cmp(&self, other: &Self) -> Ordering {
-        self.start.cmp(&other.start)
+        // If the starting position is the same encapsulatig span (i.e, wider one) should come
+        // first
+        match self.start.cmp(&other.start) {
+            Ordering::Equal => self.end.cmp(&other.end),
```
mitchmindtree, line 20:
```suggestion
            Ordering::Equal => other.end.cmp(&self.end),
```
^ I *think* it should be this

mitchmindtree:
For now, we might want to preserve the whitespace "gap" between comments and their preceding/following items too. We might be better off thinking about how to format comments and the whitespace gaps between comments and surrounding items in future PRs, as there is probably a lot to consider when trying to re-arrange or format comments.

People do all kinds of stuff in comments like diagrams, post-fix on the same line, prefix to the line above, etc.

Could be good to keep some gross examples in mind, or even some as tests to check that we don't accidentally remove them or break them. I'm imagining stuff like:

```rust
// The following items are all related to the foo bar baz. Blah blah blah.

// This is a Foobar.
struct Foobar {
    a: A, // A is a A.
    b: B, // TODO: We should change this to a C.
    /// D is important and pub.
    /// blah blah.
    /// -----
    /// | D |
    /// -----
    pub d: D,
}

// This is a foo function.
fn foo(
    bar: Bar, // this is a bar
    baz: Baz, // this is a baz
    // Qux is a little more complicated.
    // So I'm writing multiline on top to describe why.
    // Blah blah blah
    qux: Qux,
) -> /* TODO: Result<Foobar> */ Result<Barfoo> {
}
```

kayagokalp:
As I started to explore how I can insert the comments to relevant places without formatting them, I realized that it is going to be somewhat tricky. I am thinking to have a separate PR for the actual comment insert operation and keep this one for only providing the means of comment formatting which will complete these 2 todos in #2122:

>  - [ ] Implement Ord for Span as mentioned above. If unsuitable for sway-types due to paths, use our own Span type with conversions implemented. Edit: after taking another look at Span I think we don't want to use it directly. Instead we should use our own span (i.e. a simple byte range) and handle one file at a time.
> - [ ] Use CommentedTokenStream to collect a map of spans -> comments in sway-fmt-v2.

Would that make sense @mitchmindtree or should I keep going on this one until we have the complete formatting implemented?

Also once you are available another brainstorming session about actually inserting the comments would be great for me because as I started to dive deeper into actually inserting them, I started to have some concerns regarding the comments inside some code-block. (I feel like I need some sort of mapping between the spans of items before formatting to after formatting, I can do that by re-parsing the formatted code and adding comments in a separate pass after I constructed while traversing the AST (that contains formatted code spans) but I feel like there could be a better approach for this)

sway-fmt-v2/src/utils/comments.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    for comment in tts {
+        get_comment_from_token_stream(comment, &mut comment_map);
+    }
+    Ok(comment_map)
+}
+
+/// Get `Comment` from the token stream and insert it with its span to the `CommentMap`.
+/// Handles both the standalone and in-block comments.
+fn get_comment_from_token_stream(
```
mitchmindtree, line 51: In Rust, `get` is generally used to refer to the retrieval of a `&` to an element within a datastructure (e.g. `Vec::get`, `HashMap::get`, `BTreeMap::get`, etc). Perhaps a better name for this would `collect_comments_from_token_stream`?

sway-fmt-v2/src/utils/comments.rs
```diff
@@ -0,0 +1,182 @@
...
+    fn partial_cmp(&self, other: &CommentSpan) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+pub type CommentMap = BTreeMap<CommentSpan, Comment>;
+
+/// Get the CommentedTokenStream and collect the spans -> Comment mapping for the input source
+/// code.
+pub fn construct_comment_map(input: Arc<str>) -> Result<CommentMap, FormatterError> {
```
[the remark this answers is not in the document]
mitchmindtree: I think just `comment_map` should be pretty unambiguous here.

Otherwise, if you do prefer to clarify that it is a constructor, the general convention for constructors in Rust is to use `new` or `from`, e.g. `new_comment_map` or `comment_map_from_src`.

mitchmindtree (commented):
Good idea, I agree it makes sense to land the `CommentMap` and `CommentSpan` stuff first.

It's looking really good! I've just added a couple small suggestions.

Also, more than happy to have a chat again early next wk about how to go forward from here :relaxed:

eureka-cpu (dismissed):
Nice one!

mitchmindtree (approved):

sway-fmt-v2/src/utils/comments.rs
```diff
@@ -0,0 +1,169 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::{comment_map_from_src, CommentSpan};
+    use std::{ops::Bound::Included, sync::Arc};
+
+    #[test]
+    fn test_comment_span_ordering() {
+        let first_span = CommentSpan { start: 2, end: 4 };
+        let second_span = CommentSpan { start: 2, end: 6 };
```
mitchmindtree, line 80: Oh, I just saw your response to my Ordering suggestion. It looks like this test is incorrect - I think the `start: 2, end: 6` span should occur *before* the `start: 2, end: 4` span, as I think we want the "visit" order where the enclosing spans precede the enclosed spans. Does that make sense?

mitchmindtree (changes requested):

mitchmindtree (approved):

eureka-cpu (approved):

JoshuaBatty (approved):

Issue #2143, opened 2022-06-28
Hidden dead code warnings due to a small issue in the collection of warnings in DCA


mohammadfawaz:
I think the condition here: https://github.com/FuelLabs/sway/blob/2264b2d898a512d899e0ae23461802a7b4259598/sway-core/src/control_flow_analysis/dead_code_analysis.rs#L99 should be inverted i.e. `!all_warnings.iter().any(`. When I do this, I get some new warnings from `core` and `std`:
```console
╰─ forc build
warning
   --> /Users/mohammad/Desktop/fuel/sway/sway-lib-core/src/ops.sw:467:7
    |
465 |
466 |
467 | trait OrdEq: Ord + Eq {
    |       ----- This trait is never implemented.
468 | } {
469 |     fn ge(self, other: Self) -> bool {
    |
____

  Compiled library "core" with 1 warning.
warning
 --> /Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/constants.sw:3:1
  |
1 |
2 |
3 | const BASE_ASSET_ID = 0x0000000000000000000000000000000000000000000000000000000000000000;
  | ----------------------------------------------------------------------------------------- This declaration is never used.
  |
____

warning
  --> /Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/tx.sw:32:1
   |
30 |
31 |
32 | const TX_TYPE_OFFSET = 10240;
   | --------------------------------- This declaration is never used.
33 | const TX_GAS_PRICE_OFFSET = 10248;
34 | const TX_GAS_LIMIT_OFFSET = 10256;
   |
____

warning
  --> /Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/tx.sw:33:1
   |
31 |
32 | const TX_TYPE_OFFSET = 10240;
33 | const TX_GAS_PRICE_OFFSET = 10248;
   | ---------------------------------- This declaration is never used.
34 | const TX_GAS_LIMIT_OFFSET = 10256;
35 | const TX_BYTE_PRICE_OFFSET = 10264;
   |
____

warning
  --> /Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/tx.sw:34:1
   |
32 |
33 | const TX_GAS_PRICE_OFFSET = 10248;
34 | const TX_GAS_LIMIT_OFFSET = 10256;
   | ---------------------------------- This declaration is never used.
35 | const TX_BYTE_PRICE_OFFSET = 10264;
[...]

Issue #306, opened 2021-10-11
`pub const` for exporting constants from libraries


sezna:
This issue has two parts:

1. Only make `pub const` values exported in libraries
2. Fix dead code analysis so `pub const` values are not warned as dead code but private `const` ones are.

PR #2230, opened 2022-07-05
Correctly display DCA warnings and fix some incorrect warnings that showed up

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/2143
Closes https://github.com/FuelLabs/sway/issues/306

The missing `!` in DCA was hiding some warnings. When added, a bunch of warnings showed up from `lib-core` and `lib-std`:
- One in `sway-lib-core` due to a duplicate declaration of `OrdEq` (likely a git merge issue?)
- A few others related to unused constant declarations:
  - Some of those need to be `pub` so I enabled that keyword for constants and fixed them.
  - Some don't need to be `pub` but still emit a warning. Turns out `asm` blocks needed some additional handling in DCA.
    - Those that don't need `pub` still caused an error in IR when the library code that uses them is imported. The fix is to remove the visibility checking in the const expr evaluator because the actual check happens earlier in type checking anyways and the existing check is causing the const not to be available to the library code itself.

Another instance of incorrect warnings showed up with `storage` declarations which I also fixed.


sway-core/src/control_flow_analysis/dead_code_analysis.rs
```diff
@@ -148,53 +148,70 @@ impl ControlFlowGraph {
...
+                            // Don't emit warnings for constant declarations in libraries. We
+                            // cannot yet figure out whether they're actually used outside the
+                            // library or not.
+                            ControlFlowGraphNode::ProgramNode(TypedAstNode {
+                                content:
+                                    TypedAstNodeContent::Declaration(
+                                        TypedDeclaration::ConstantDeclaration(_),
+                                    ),
+                                ..
+                            }) => true,
```
mohammadfawaz, line 209: This is the only actual addition to this file. Everything else is due to the formatter.

sway-lib-core/src/ops.sw
```diff
@@ -459,16 +459,6 @@ trait OrdEq: Ord + Eq {
     }
 }
 
-trait OrdEq: Ord + Eq {
```
mohammadfawaz, line 462: Duplicate trait declaration.

sway-core/src/control_flow_analysis/dead_code_analysis.rs
```diff
@@ -100,7 +95,7 @@ impl ControlFlowGraph {
             .filter(|CompileWarning { span, .. }| {
                 // if any other warnings contain a span which completely covers this one, filter
                 // out this one.
-                all_warnings.iter().any(
+                !all_warnings.iter().any(
```
mohammadfawaz, line 98: This is the main fix.

sway-lib-std/src/assert.sw
```diff
@@ -3,7 +3,7 @@ library assert;
 use ::logging::log;
 use ::revert::revert;
 
-const FAILED_REQUIRE_SIGNAL = 42;
+pub const FAILED_REQUIRE_SIGNAL = 42;
```
mohammadfawaz, line 6: This shouldn't really need the `pub` keyword but without it I get an IR error. Will create an issue.

sway-lib-core/src/ops.sw
mohammadfawaz, line 462: Duplicate trait declaration.

sway-core/src/ir_generation/const_eval.rs
```diff
@@ -47,15 +46,8 @@ pub fn compile_const_decl(
                 TypedDeclaration::ConstantDeclaration(TypedConstantDeclaration {
                     name,
                     value,
-                    visibility,
-                }) => {
-                    // XXX Do we really only add public constants?
```
otrho, line 52: This was a comment from when I copied the older AST->ASM code which only compiled pubs, and it has languished here for 6 months!  Thanks. 🙂

otrho (approved):

mitchmindtree (approved):
Lgtm!

PR #2231, opened 2022-07-05
Add swap method to Vec<T>

r-sitko:
ref #2014


sway-lib-std/src/vec.sw
```diff
@@ -208,4 +208,22 @@ impl<T> Vec<T> {
...
+        assert(a < self.len);
+        assert(b < self.len);
+
+        let val_size = size_of::<T>();
+
+        let a_ptr = self.buf.ptr() + a * val_size;
+        let b_ptr = self.buf.ptr() + b * val_size;
+
+        let a_val = read(a_ptr);
+        copy(b_ptr, a_ptr, val_size);
```
mohammadfawaz, line 226: I suspect this call to `copy` will panic if `a` and `b` are the same. We should probably just return early if `a == b` and not do anything.

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
```diff
@@ -1810,3 +1811,27 @@ fn test_vector_with_capacity_u64() {
         Option::Some(val) => revert(0), Option::None => (), 
     }
 }
+
+fn test_vector_swap() {
```
mohammadfawaz, line 1815: I suggest that we expand the `swap` tests to cover all possible data types. For inspiration, take a look at this PR that did that recently. So, for each function `test_vector_new_*`, append some lines that test `swap` for the given data type.

mohammadfawaz (changes requested):
This looks great! Two comments that require some attention but otherwise on-point. Thank you for your contribution 😄

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
Braqzen: The referenced PR is #2170

mohammadfawaz: Lol thank you @Braqzen!

sway-lib-std/src/vec.sw
Braqzen: I don't see why it would panic however I agree that if `a == b` then that is a noop. Then again how often is someone going to swap the same elements? Is it worth adding an additional check and expanding the bytecode a little bit more just to do a noop that otherwise is a mistake from the user (that I do not immediately see from looking at the implementation of copy why it would panic)? In terms of efficiency and bytecode size I would argue that there is no reason to include it unless it actually panics and if it panics then perhaps that is a compiler issue that needs to be solved instead of a high level library bandaid.

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
```diff
@@ -1810,3 +1811,27 @@ fn test_vector_with_capacity_u64() {
...
+    let number0 = 0u8;
+    let number1 = 1u8;
+    let number2 = 2u8;
+    let number3 = 3u8;
+
+    vector.push(number0);
+    vector.push(number1);
+    vector.push(number2);
+    vector.push(number3);
+
```
Braqzen, line 1827: add an assertion to prove that your vec is the size that you expect it to be and also capacity. after that you must check that your element 1 and element 3 are what you expect them to be. this is a "before condition" so before you use your method you must assert that your initial state is what you expect it to be otherwise your test might pass falsely because the implementation could be incorrect or there might be an underlying bug

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
```diff
@@ -1810,3 +1811,27 @@ fn test_vector_with_capacity_u64() {
...
+    vector.swap(1, 3);
+
+    match vector.get(1) {
+        Option::Some(val) => assert(val == number3), Option::None => revert(0),
+    }
+
+    match vector.get(3) {
+        Option::Some(val) => assert(val == number1), Option::None => revert(0),
+    }
```
Braqzen, line 1836: There should also be two tests for when `a` and `b` are out of bounds i.e. greater than or equal to the length. you have added in the assertions as preconditions to your method so those conditions must be checked as well

Braqzen (changes requested):

sway-lib-std/src/vec.sw
mohammadfawaz: The reason I expect a panic is because `copy` calls `mcp` under the hood which does not allow overlapping ranges (ref: https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/instruction_set.md#mcp-memory-copy). I do agree that an additional check will cause extra bytecode that is not going to be needed in most cases, but we have to handle that case in a public API unfortunately. If we don't want to do this here, then we have to change the spec (and implementation) of `mcp`.

adlerjohn: Ref: https://github.com/FuelLabs/fuel-vm/issues/147

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
mohammadfawaz: I agree with @Braqzen. How about we start by creating a new test called `vec_swap` which is similar to `vec` but tests on swaps (for all data types)? We can then further split the existing `vec` test as we see fit.

sway-lib-std/src/vec.sw
```diff
@@ -208,4 +208,22 @@ impl<T> Vec<T> {
         self.len -= 1;
         Option::Some(read(self.buf.ptr() + self.len * size_of::<T>()))
     }
+
+    /// Swaps two elements
+    /// a - The index of the first element
+    /// b - The index of the second element
+    /// Panics if `a >= self.len || b >= self.len` .
```
adlerjohn, line 215: Taking from https://doc.rust-lang.org/src/core/slice/mod.rs.html#593-604

```suggestion
    /// Swaps two elements.
    ///
    /// # Arguments
    ///
    /// * a - The index of the first element
    /// * b - The index of the second element
    ///
    /// # Reverts
    ///
    /// Reverts if `a` or `b` are out of bounds.
```

Braqzen: Guess I'll start bringing this up outside of sway-apps too

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
r-sitko: > I agree with @Braqzen. How about we start by creating a new test called `vec_swap` which is similar to `vec` but tests on swaps (for all data types)? We can then further split the existing `vec` test as we see fit.

Do you mean creating new project `vec_swap` in e2e tests something like test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec?

sway-lib-std/src/vec.sw
```diff
@@ -208,4 +208,22 @@ impl<T> Vec<T> {
...
+    pub fn swap(mut self, a: u64, b: u64) {
```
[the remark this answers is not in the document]
r-sitko: > Can you rename `a` and `b` to be a little more meaningful e.g. `element1` and `element2`? I know that rust uses `a` and `b` but that is bad practice and those devs need to get a talking to.

Maybe better will be to rename to `element1_index` and `element2_index`, because `a` and `b` are not elements to swap, but their indexes.
Btw, C++ also uses `a` and `b` naming convention.

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
mohammadfawaz: Yup! Just a separate test called should_pass/stdlib/vec_swap for example.

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
r-sitko: > add an assertion to prove that your vec is the size that you expect it to be and also capacity. after that you must check that your element 1 and element 3 are what you expect them to be. this is a "before condition" so before you use your method you must assert that your initial state is what you expect it to be otherwise your test might pass falsely because the implementation could be incorrect or there might be an underlying bug

Do we really need to assert vec size, capacity, element 1 and element 3 in "before condition" that they have proper initial values? I think it would be unnecessary as other tests in vec test suite should catch implementation errors of `Vec` initialization and push. What do you think? Wouldn't it be double checking same functionality?

Braqzen: Yes you do. You must assert your initial condition. It doesn't matter how you got there but it matters that you can prove your initial condition. You cannot rely on other tests to help test your current test. That's coupling code.

sway-lib-std/src/vec.sw
```diff
@@ -208,4 +208,32 @@ impl<T> Vec<T> {
...
+    /// # Arguments
+    ///
+    /// * element1_index - The index of the first element
+    /// * element2_index - The index of the second element
+    ///
+    /// # Reverts
+    ///
+    /// Reverts if `element1_index` or `element2_index` is greater than or equal to the length of vector.
+    pub fn swap(mut self, element1_index: u64, element2_index: u64) {
+        if element1_index == element2_index {
```
mohammadfawaz, line 223: I would actually move the `if` after the `asserts`. We do want the `swap` function to panic if the two indices are out-of-bounds even if they're equal.

mohammadfawaz (changes requested):

mohammadfawaz (approved):
Looks great from my point of view! Thanks again 🙇‍♂️

adlerjohn (approved):

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec_swap/src/main.sw
```diff
@@ -0,0 +1,787 @@
...
+        Option::Some(val) => {
+            assert(val == number0)
+        },
+        Option::None => {
+            revert(0)
+        },
+    }
+}
+
+fn test_vector_swap_u128() {
```
Braqzen, line 104: I wouldn't bother testing custom types like this. Lots will be added over time and this is just a repeat of a struct test

Braqzen (commented):
Seems fine. I won't block but I won't approve until the `u128` is test is removed since that's a duplicate test for a struct

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec_swap/src/main.sw
r-sitko: I removed test_vector_swap_u128 now.

Braqzen (approved):

mohammadfawaz (approved):

Issue #1787, opened 2022-06-01
When `--offline` flag specified for `forc build`, attempt to use local git checkouts for git dependencies before failing


mitchmindtree:
Supporting `forc build --offline` is trivial for the case where the `[dependencies]` table in the `Forc.toml` manifest remains untouched, as `forc` can recreate the dependency graph entirely from the lock file. However, if the `[dependencies]` table is edited whatsoever, forc will try to recreate the whole graph by re-fetching dependencies.

This behaviour will be greatly improved by #1686 which makes more concise modifications to the package graph based on the difference between the dependencies read from the lock file and those read from the manifest. These means we can now at least support dependency removal and editing/adding `path`-sourced dependencies without failing in `--offline` mode. 

We should also do our best to support editing/adding `git` dependencies in `--offline` mode.

## Potential Approach

In the case that `--offline` was specified, we should attempt to find an existing git checkout for the specified dependency within `~/.forc/git/checkouts` that adheres to the described git source.

For example, In the case that the following dependency was added:

```toml
foo = { git = "https://domain.com/owner/repo", branch = "bar" }
```

We should attempt to find all existing checkouts of the `foo` package whose commit is known to be of the `bar` branch from the `https://domain.com/owner/repo` repo and select the most recent one. Finding local checkouts for git sources that specify a `tag` or `rev` should be easier, as they should map to a specific commit.

We could possibly consider maintaining an index that maps from `GitSource` to the path of the latest known checkout for that git source. However, we should be weary that if we maintain an index it will need to handle the case where multiple processes or threads are attempting to read-from, add-to or edit it simultaneously. Ideally we shouldn't have to introduce a whole db crate just for this case.

Alternatively, when checking out local git repositories into `~/.forc/git/checkouts/`, we can copy over the `.git` directory too (currently we omit it). Doing so would allow us to use the `git2` crate to check whether the existing checkout is valid for the specified git source.

It's worth noting that searching and checking `~/.forc/git/checkouts` without and index should not be a major performance hit, as we only require checking out directories that start with the same package name, and the search only needs to be performed once as consecutive calls to `forc build` will retrieve the dependency directly from the lock file.

kayagokalp:
I was planning to start this but moved formatter stuff to the front in my current work queue. Removing my assignment for now so that if someone wants to tackle it until we are done with formatter stuff. Otherwise, I can take a look after formatter as I am pretty excited about the offline support 😄

kayagokalp:
[...]

Issue #895, opened 2022-03-07
Forc doesn't automatically try to build in `--offline` mode if there is no active internet connection


sezna:
I was banging out some proper LOC on the airplane, as one does, and I ran the sway test suite. I didn't pass `--offline`, although, on the airplane, I indeed had no network.

```
Creating a new `Forc.lock` file
    Cause: Manifest dependencies do not match
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: failed to clone package 'core' from 'http://github.com/FuelLabs/sway-lib-core': failed to resolve address for github.com: nodename nor servname provided, or not known; class=Net (12)', test/src/e2e_vm_tests/harness.rs:69:46
```

I think we should either:
1. Automatically try `--offline` if no network is present
2. Provide a message instead of a panic, e.g. _Unable to connect to Github. Is your computer connected to the internet?_

mitchmindtree:
Good call @sezna - I've opened #900 which I think we might want to solve first. I might remove "good first issue" as it could be a little more involved than a newcomer might expect.

mitchmindtree:
#900's almost addressed now in #1686.

Next blocker is #1787.

kayagokalp:
Before #1778 has a solution we cannot handle removed dependencies as we wanted with #1686. So I guess that is also a blocker.

kayagokalp:
As https://github.com/FuelLabs/sway/pull/2366 is about to finalize, I am considering tackling this to offer a complete offline support. For detecting the internet connectivity the most straightforward way seems like checking if we can ping a specific website (it can be [github](https://github.com) or something else) but this also means we are limited by the uptime of the selected target(s).

PR #2234, opened 2022-07-06
`forc-pkg`: Refactor `BuildPlan` construction and validation to minimize fetching, improve offline support, reduce I/O, fix bugs.

### LINKED ISSUE #1778: `BuildPlan::from_lock()` should be able to create a `BuildPlan` even there is a package removal in the manifest file

### LINKED ISSUE #1903: `forc` can create an invalid package dependency graph if the `[dependencies]` table of a `path` dependency manifest has been modified
Caught by @nfurfaro and @otrho.

## Problem

Right now we only check the immediate dependencies specified in the current project's manifest when validating the graph loaded from the lock file. This means if the `[dependencies]` table of one of the current project's `path` dependencies is modified, we don't correctly catch or account for this change when loading the graph from the lock file.

## Solution

When loading the graph from the `Forc.lock` file, we should not only validate it against the top-level project manifest, but also against the manifest of each of the path dependencies as it's possible that any of them have been modified since the `Forc.lock` file was created.

Note that this is not important for git dependencies where we know they haven't changed due to the hash.

## Temporary Workaround

Delete the `Forc.lock` and re-run `forc build` or `forc update` to allow `forc` to rebuild the graph from scratch.

mitchmindtree:
This continues @kayagokalp's great work on #1974 with a wider overhaul of the package graph `BuildPlan` validation and construction.

## `BuildPlan` refactor

During kaya's work on #1974 it became clear we needed to reconsider the approach to package graph validation in order to be able to more generally account for all of the different ways a node might have been invalidated since the lock file was last written, and to be able to handle invalidated nodes in a more consistent manner.

This refactors the build plan construction process into the following distinct steps:

1. Load the project manifest.
2. Load the project lock.
3. Convert the lock to a package graph.
4. Validate the package graph, collecting all invalid nodes. A node is invalid if:
    - We are unable to construct its local path.
    - We are unable to load its manifest.
    - It is no longer present within its parent's manifest.
    - Its `Source` differs from the source or patch described in the parent's manifest.
    - Its name differs from the name within its manifest.
5. Remove all invalid nodes and any remaining nodes with no valid parents. Remaining nodes are known to be valid.
6. Fetch the missing components of the graph until it is complete. If in `offline`, only fail if we are missing a `git` node.
7. Determine the compilation order.

Closes #1903.
Closes #1778.
Makes progress on #895, #1787.

## ~`PathMap`~ `ManifestMap`

This PR also reduces the number of times that we need to read each dependency's manifest from file in order to reduce the amount of I/O performed. Each dependency's manifest should now be loaded once during either `graph_to_manifest_map` or `fetch_graph`.

This required implementing `Clone` for `ManifestFile` to allow for constructing the build plan in a manner that does not require consuming the manifest file. `ManifestFile::clone` is only ever called for the project manifest, and never for the dependencies.

As a side-effect, the `build` and `check` commands no longer require loading each package's manifest as they are already provided by the build plan. As a result both commands no longer require the sway git tag to be passed through.


kayagokalp (approved):
Looks good to me 🙌 , just left a small nit (which is not a blocker, just a simple typo).

 Also a side note I realized the build process is considerably faster (while adding a new dep) with this PR. A small benchmark on my m1 MBA:

**Adding dep before this PR:** 3.18 seconds
**Adding dep with this PR:** 0.79 seconds

kayagokalp (approved):

JoshuaBatty:
I think it's fine to merge this into master and close #1974.

JoshuaBatty (approved):
Nice work. Looks good.

mitchmindtree:
> Should this be merged into https://github.com/FuelLabs/sway/pull/1974 instead of master?

Ah I should have clarified this in my OP a bit further! I decided to base directly against `master` as the diff was just about as large compared to #1974 as it was to `master`, so I thought I'd try saving on the intermediary review. But you're right and I agree, in most cases it's better to base against the existing PR when possible.

Issue #1167, opened 2022-04-05
Struct destructuring


emilyaherbert:
This Sway code should work:

```
script;

struct Data<T> {
    value: T
}

impl<T> Data<T> {
    fn new(value: T) -> Data<T> {
        Data {
            value: value
        }
    }
}

fn main() {
    let data = Data::new(7);
    let Data { value } = data;
    // use value to do something
}
```

The equivalent Rust code: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a745f0543c5c918f865e130de85dd6da

matt-user:
How difficult would this be to implement for someone with limited compiler experience?  I'm interested in working on this issue, and learning more about sway and its compiler, but I don't want my inexperience to be a development blocker.  Would this feature be implemented primarily in `parse_tree` and `semantic_analysis`?  If someone could point me in the right direction it would be appreciated.

emilyaherbert:
Hi @matt-user , this is a good first issue to start with because there is a pretty good example of how to get started here: https://github.com/FuelLabs/sway/blob/19586172e455a499d09d03b0aaca3d44bb47665c/sway-core/src/convert_parse_tree.rs#L2606

In Sway, tuple patterns on the LHS like `let (a, b) = (0, 1);` are desugared as syntactic sugar. So the LOC:

```rust
let (a, b) = (0, 1);
```

gets desugared into:

```rust
let <new name> = (0, 1);
let a = <new name>.0;
let b = <new name>.1;
```

So, I recommend approaching structs similarly:

```rust
let Data { value } = data;
```

would get desugared to:

```rust
let <new name> = data;
let value = <new name>.value;
```

matt-user:
Thank you so much for the information!  I'll be sure to ask questions as I go along :)

PR #2243, opened 2022-07-06
feat: struct destructuring

matt-user:
closes issue #1167 
There is definitely some repeated code between the Struct and Tuple Pattern branches in the statement_let_to_ast_nodes function.   Let me know if I should add some helper functions.


sway-core/src/convert_parse_tree.rs
```diff
@@ -2571,19 +2660,31 @@ fn statement_let_to_ast_nodes(
...
                     span: span.clone(),
                 };
+
+                // from the possible type annotation, if the annotation was a tuple annotation,
+                // extract the internal types of the annotation
                 let tuple_tys_opt = match ty_opt {
                     Some(Ty::Tuple(tys)) => Some(tys.into_inner().to_tys()),
                     _ => None,
                 };
+
```
sezna, line 2676: thanks for these comments, great addition

test/src/e2e_vm_tests/test_programs/should_pass/language/struct_destructuring/test.toml
```diff
@@ -0,0 +1,3 @@
+category = "run"
+expected_result = { action = "return", value = 42 }
+validate_abi = true
```
sezna, line 3: as this isn't a contract, do we need to validate the abi?

sezna (changes requested):
Love the changes -- clean, in line with @emilyaherbert's awesome tuple destructuring impl, and well commented.

Two things -- 
1. Could you document this in the `docs/` folder as a part of this PR?
2. Does this support nested destructuring? If so, we should test it. If not, we should document that limitation for both structs and tuples.

test/src/e2e_vm_tests/test_programs/should_pass/language/struct_destructuring/test.toml
matt-user: Good catch!  I'm not sure, there are other test cases in the same folder such as tuple_access that are scripts and validate the abi.

sezna: Hm, I don't think it is necessary. Cc @otrho for consensus on that.

adlerjohn: Time for an RFC!

test/src/e2e_vm_tests/test_programs/should_pass/language/nested_struct_destructuring/src/main.sw
```diff
@@ -0,0 +1,21 @@
+script;
+
+// Tests nested struct destructuring
+
+fn main() -> u64 {
+    let point1 = Point { x: 0, y: 0 };
+    let point2 = Point { x: 1, y: 1 };
+    let line = Line { p1: point1, p2: point2 };
+    let Line { p1: Point { x: x0, y: y0 }, p2: Point { x: x1, y: y1} } = line;
```
sezna, line 9: alright, don't hate me, but what about a nested tuple in a struct and vice versa?

test/src/e2e_vm_tests/test_programs/should_pass/language/struct_destructuring/test.toml
mohammadfawaz: We've bee validating the ABI for all passing tests including scripts because scripts emit a JSON ABI now. Not sure how important that is though.

test/src/e2e_vm_tests/test_programs/should_pass/language/nested_struct_destructuring/src/main.sw
matt-user: Yeah I can add a test for that.  Should I also include an example for the docs?

sezna: Docs example could be cool, to show people just how deep destructuring can go!

sezna (dismissed):
💪🏻

sezna (approved):

test/src/e2e_vm_tests/test_programs/should_pass/language/struct_destructuring/test.toml
otrho: BTW, all I did was refactor the test harness.  This doesn't make me the authority on how or what to test. :wink:

otrho (approved):

Issue #2207, opened 2022-07-03
`forc init` has an additional subcommand which is added by CLI automatically


Braqzen:
## Summary

When runnning `forc init --help` you see the options one of which is `--verbose`.
When appending that to your command it turns out that nothing happens and @sezna said that it's added automatically by a CLI tool.

The `--verbose` should probably be removed / prevent the CLI from adding it.

### Help
![1](https://user-images.githubusercontent.com/103777923/177048487-b00dd597-709b-49b6-91d0-189efb563ba0.png)

### Verbose
![1 (1)](https://user-images.githubusercontent.com/103777923/177048380-8c303a18-94d7-4be6-b3e3-eeefafb01cc7.png)

### Non-verbose
![1 (2)](https://user-images.githubusercontent.com/103777923/177048385-41baa16b-f6b5-4b3d-a81f-5256e36d7330.png)

PR #2244, opened 2022-07-06
Change verbosity from debug -> info in forc init

ra0x3:
### Description
- Looks like `forc init` was using `debug` level tracing for verbosity instead of `info`
- This PR swaps `debug` for `info` and hides the verbosity behind the `command.verbose` option


adlerjohn (approved):

adlerjohn:
Is this related to #2207?

ra0x3:
> Is this related to #2207?

@adlerjohn 
- Yes, the description in #2207 is a bit of a misnomer (suggesting to remove the `verbose` flag). 
- The issue was the use of `debug!` as opposed to `info!` looks like

adlerjohn:
:shipit:

mitchmindtree:
I wonder if instead we should keep the usage of `debug!`, and have `verbose` be an alias for setting the logger to the `debug` level? That way we can still allow for configuring everything via the more flexible environment logging approach too, e.g. `RUST_LOG=debug forc <cmd>`.

ra0x3:
@mitchmindtree 
- 🤦 You're totally right, I forgot we were passing `RUST_LOG=` to `forc` commands now. I'll push a change

mitchmindtree:
> - But I'm thinking -- doesn't me specifying verbose supersede RUST_LOG?
>     - Meaning, even if I have no RUST_LOG shouldn't I see verbose output if verbose is specified?
>     - Or was that change to use tracing for verbosity (I forget which PR it was) a permanent change to RUST_LOG for verbosity as opposed to verbose?
>     - I guess it's fine if I have to pass RUST_LOG as a sort of verbose alias, but feels slightly unintuitive?

Yeah, I think it's fine for `--verbose` to override the log level provided by the `RUST_LOG=<level>` environment variable.

My point is more that, rather than going down a road where we have `if command.verbose { ... }` everywhere, we could potentially take advantage of the integrated `tracing` logging and have `verbose` be an alias for a more verbose level of logging (e.g. `debug` or `trace` level logging).

But I agree, the CLI argument should probably take precedence over the `RUST_LOG` environment variable.

mohammadfawaz:
Is this ready to be merged?

ra0x3:
@mohammadfawaz I'm OK to merge. @mitchmindtree If you prefer, can follow this up with a quick PR to change the `info!` back to `debug!`

mitchmindtree:
Hey @ra0x3 I've opened https://github.com/FuelLabs/sway/issues/2305 to track this. I think it would be good to address this eventually, but it's certainly not particularly high priority so no stress :ok_hand:

Issue #2240, opened 2022-07-06
Build hangs on large crypto code


adlerjohn:
https://github.com/hashcloak/fuel-crypto/tree/da305181a77f3c6898571c7632b4cae3c2345c4a/edwards25519

`forc build` hangs with `v0.17.0`.

mohammadfawaz:
Looks like this is another instance of a bug in codegen that is very similar to https://github.com/FuelLabs/sway/pull/1541. Will have a PR ready asap.

mohammadfawaz:
Fixed in https://github.com/FuelLabs/sway/pull/2252

PR #2251, opened 2022-07-06
Fix how warnings/errors are displayed to avoid a crash

mohammadfawaz:
The code linked to in https://github.com/FuelLabs/sway/issues/2240 crashes with the following panic:
```
thread 'main' panicked at 'SourceAnnotation range `(92, 1272)` is bigger than source length `1268`', /Users/mohammad/.cargo/registry/src/github.com-1ecc6299db9ec823/annotate-snippets-0.9.1/src/display_list/from_snippet.rs:286:9
```
from https://github.com/rust-lang/annotate-snippets-rs/blob/448b80421b5a9c39c3ed1a708636014aec82eac1/src/display_list/from_snippet.rs#L287

This wasn't happening in `0.17.0` because the problematic warning was dropped in DCA but now show up due to https://github.com/FuelLabs/sway/pull/2230. 

The requirement is here is that the length of `input` is at least as big as `*end_ix`.

With this change, the problematic warning shows up as intended:
image

Honestly, I'm not 100% sure that this is the best way to fix this. I don't fully understand how `construct_window()` works 🤔 . Soliciting help if someone has a better idea.

### DIFF
--- a/forc-util/src/lib.rs
+++ b/forc-util/src/lib.rs
@@ -407,7 +407,10 @@ fn construct_window<'a>(
     *start_ix -= std::cmp::min(calculated_start_ix, *start_ix);
     *end_ix -= std::cmp::min(calculated_start_ix, *end_ix);
     start.line = lines_to_start_of_snippet;
-    &input[calculated_start_ix..calculated_end_ix]
+
+    // The length of input must be at least as big as *end_ix. This is required by the
+    // annotate_snippets library
+    &input[calculated_start_ix..std::cmp::max(calculated_start_ix + *end_ix, calculated_end_ix)]
 }
 
 const LOG_FILTER: &str = "RUST_LOG";


sezna:
Hm, so in this case, after DCA, we essentially remove any warnings that have spans that  got eliminated? Or, shorten any spans that have regions which got eliminated?

I'm not sure this is the best way forward. Ideally, we can perform DCA maybe after rendering warnings, so they're not lost based on the result of optimizations. Compile errors and warnings are effectively outputs of the compiler, and optimizations should not cause them to get malformed or omitted.

mohammadfawaz:
> Hm, so in this case, after DCA, we essentially remove any warnings that have spans that  got eliminated? Or, shorten any spans that have regions which got eliminated?
> 
> 
> 
> I'm not sure this is the best way forward. Ideally, we can perform DCA maybe after rendering warnings, so they're not lost based on the result of optimizations. Compile errors and warnings are effectively outputs of the compiler, and optimizations should not cause them to get malformed or omitted.

I agree that something fishy is going on. I don't know that the warning and its span have been modified... it's just that in this file, we construct a window into the code (input) based on the span and display that but this is a case where the calculations were a bit off. This particular warning is from DCA before any optimizations happen. Do you think there is something wrong with the formation of the warning itself? Or do you think the problem is in construct_window? My fix to construct_windows is a bit of a bandaid fix.

forc-util/src/lib.rs
```diff
@@ -407,7 +407,10 @@ fn construct_window<'a>(
     *start_ix -= std::cmp::min(calculated_start_ix, *start_ix);
     *end_ix -= std::cmp::min(calculated_start_ix, *end_ix);
     start.line = lines_to_start_of_snippet;
-    &input[calculated_start_ix..calculated_end_ix]
+
+    // The length of input must be at least as big as *end_ix. This is required by the
+    // annotate_snippets library
+    &input[calculated_start_ix..std::cmp::max(calculated_start_ix + *end_ix, calculated_end_ix)]
```
otrho, line 413: This whole function is hard to read.  Is `end_ix` an index or a length?  Adding two indices together doesn't make sense does it?  I can't tell what's going on.

mohammadfawaz: You know what? I'm going to study that whole function tomorrow and rewrite it. This change is a bandaid fix and will probably lead to more issues in the future.

PR #2252, opened 2022-07-06
Fix some small issues in compiling load and store IR instructions when the word offset is > 4096

mohammadfawaz:


test/src/e2e_vm_tests/test_programs/should_pass/language/many_stack_variables/src/main.sw
```diff
@@ -147,5 +147,186 @@ fn main() -> u64 {
...
+    let c = 0;
+    let c = 0;
+    let c = 0;
+    let c = 0;
+    let c = 0;
+    let c = 0;
+    let c = 0;
+    let c = 0;
+    let c = 0;  
+    let c = 0; // problematic
```
mohammadfawaz, line 201: This test still fails with this change. The failure used to be a memory overflow but now we just get the wrong result. There's something else going on that I'm still trying to figure out.

otrho (approved):
Thanks for this. 👍

test/src/e2e_vm_tests/test_programs/should_pass/language/many_stack_variables/src/main.sw
mohammadfawaz: Resolved.

sezna (approved):

Issue #2261, opened 2022-07-07
Support the LSP rename capability


JoshuaBatty:
Renaming a symbol should ask the language server to find all references to that token in the file that share the same type information. The server needs to know about the scope and symbol table that the token belongs to in order to correctly rename the token throughout the project. 

This was temporarily removed in #1431

Add the following line when constructing a ServerCapabilities in server.rs
```
rename_provider: Some(OneOf::Right(RenameOptions {
    prepare_provider: Some(true),
    work_done_progress_options: WorkDoneProgressOptions {
        work_done_progress: Some(true),
    },
})),
execute_command_provider: Some(ExecuteCommandOptions {
    commands: vec![],
    ..Default::default()
}),
```

Issue #2260, opened 2022-07-07
Support the LSP hover capability


JoshuaBatty:
Hovering over a token should look up the definition and show `type` information, and it's visibility i.e `pub` or not. 

For functions then the function signature should be also shown.   

Once doc comments have made it into the AST Nodes then we can also display the comments as formatted markdown in the hover pop-up box. 

Add the following line when constructing a ServerCapabilities in server.rs
`hover_provider: Some(HoverProviderCapability::Simple(true))`

JoshuaBatty:
closed with:
#3084
#3258

Issue #2259, opened 2022-07-07
Support the LSP go_to definition capability


JoshuaBatty:
Add go to definition support when right-clicking on a token. If the current token passed the type checking phase, get the `Ident` of the token definition and use the file name and position to jump the editor to that location.  

Add the following line when constructing a ServerCapabilities in server.rs
`definition_provider: Some(OneOf::Left(true)),`

JoshuaBatty:
This is mostly done now apart from tuples and arrays. There is an issue tracking that here. https://github.com/FuelLabs/sway/issues/2486

JoshuaBatty:
addressed with #2620

Issue #2258, opened 2022-07-07
Support the LSP highlighting capability


JoshuaBatty:
When a token is double-clicked in the client, we only want to highlight token names that are related to the current type. i.e we can't just highlight all words that share the same name, we need to take the symbol table into account.

Add the following line when constructing a `ServerCapabilities` in server.rs
`document_highlight_provider: Some(OneOf::Left(true)),`

PR #2262, opened 2022-07-07
Remove unsupported LSP Capabilities

JoshuaBatty:
The language server is going through a bit of a re-write so we can handle typed tokens. The foundation of this work is being done in #2121. Before that lands, we should disable LSP capabilities that we can't reasonably support without knowledge of the type information of the tokens. 

I've outlined my thoughts on why we can't support the definition, rename, highlighting, and hover, capabilites without knowledge about the tokens type in #2257.

I've created the following separate issues about properly adding back in support for these capabilities one by one once we have the typed tokens in the language server. 

#2258 
#2259 
#2260 
#2261


mitchmindtree (approved):
Makes sense to me!

sezna (approved):

PR #2269, opened 2022-07-07
Remove `Spanned` and add `Format` trait to `Punctuated`

eureka-cpu:
This PR removes the `Spanned` trait in favor of the `Format` trait to allow dynamic formatting for type `T` and `P` on `Punctuated`.

### DIFF
--- a/sway-fmt-v2/src/utils/punctuated.rs
+++ b/sway-fmt-v2/src/utils/punctuated.rs
@@ -3,18 +3,18 @@ use crate::{
     FormatterError,
 };
 use std::fmt::Write;
-use sway_parse::{punctuated::Punctuated, StorageField, TypeField};
-use sway_types::Spanned;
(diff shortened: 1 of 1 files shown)


eureka-cpu:
> would it be possible to add some more tests specific to the behavior of punctuated lists?

There's a few other examples that I need to add yes, but they require a good bit of refactoring so I am adding them on a separate PR.

eureka-cpu:
> > would it be possible to add some more tests specific to the behavior of punctuated lists?
> 
> There's a few other examples that I need to add yes, but they require a good bit of refactoring so I am adding them on a separate PR.

To add clarity to this, the refactoring is in regards to `enum`, `struct` and `storage` that also use `Punctuated` but at present they are handled inside of their respective `format()` functions instead of in `Punctuated::format()`.

mitchmindtree (approved):

kayagokalp (approved):
lgtm

PR #2273, opened 2022-07-08
Bump to fuels 0.17.0

mohammadfawaz:
- [x] Bump SDK tests to fuels 0.17.0
  - one test case is failing that I'll look into. Otherwise, everything looks good. 
- [x] Bump `forc new` to use 0.17.0


forc/src/utils/defaults.rs
```diff
@@ -100,13 +100,22 @@ abigen!(MyContract, "out/debug/"#,
...
+        &wallet,
+        TxParameters::default(),
+        StorageConfiguration::with_storage_path(Some(
+            "./out/debug/"#,
+        project_name,
+        r#"-storage_slots.json".to_string(),
+        )),
+    )
+    .await
+    .unwrap();
```
mohammadfawaz, line 118: I tried this locally with a reasonable project name and ran `cargo fmt` and it did nothing which means that the output of this is already well formatted.

test/src/sdk-harness/test_projects/tx_fields/mod.rs
```diff
@@ -251,7 +254,7 @@ async fn can_get_tx_output_type() {
     // Contract output
     let output_type = 1;
     let result_ptr = contract_instance
-        .get_tx_output_pointer(1)
+        .get_tx_output_pointer(0)
```
mohammadfawaz, line 257: @digorithm is this change expected? Are the outputs added in a different order now?

digorithm: We did touch on some parts related to that in recent changes, but I'm not sure about the context here -- what's this test doing? What's the expected behavior?

mohammadfawaz: This methods basically calls the following standard library function: https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/tx.sw#L277-L283.
It just looks like the tx output at index 0 is now of "type 1" and the tx output at index 1 is of "type 3". Previously, they were in the opposite order, that's all.

sezna (approved):
LGTM but will defer to @digorithm for the above comment

digorithm (approved):
After discussing this with @segfault-magnet, yes, the change related to `Output`s was introduced in the `fuels-rs`'s Multicall PR. Here's what @segfault-magnet said:

```
originally the first output we'd insert was the asset change one:
https://github.com/FuelLabs/fuels-rs/pull/399/files#diff-436dedff71aee8cbbb04812b7dd33fd33ee52ea2fccb5d2d83ea8bba68d2e34eL77

then we would do the outputs for the external contracts:
https://github.com/FuelLabs/fuels-rs/pull/399/files#diff-436dedff71aee8cbbb04812b7dd33fd33ee52ea2fccb5d2d83ea8bba68d2e34eL135

now we first do the outputs for the external contracts:
https://github.com/FuelLabs/fuels-rs/pull/399/files#diff-436dedff71aee8cbbb04812b7dd33fd33ee52ea2fccb5d2d83ea8bba68d2e34eR218

and then the asset change
https://github.com/FuelLabs/fuels-rs/pull/399/files#diff-436dedff71aee8cbbb04812b7dd33fd33ee52ea2fccb5d2d83ea8bba68d2e34eR241

The external contracts ids are now stored in a hashset, while they were initially in a vec, so if there ever was an ordering with the vec, now there definitely isn't one with HashSet. 
```


So, yes, the order of asset change and outputs for external contracts was flipped.

Issue #2275, opened 2022-07-08
Unnecessary visibility qualifiers when defining trait functions.


tritao:
The following code will currently compile fine with Sway:

```rust
trait A {} {
    pub fn f(self) -> bool {
        false
    }
}
```

The `pub` here should be implied, and Rust indeed will issue an error for this:

```
error[E0449]: unnecessary visibility qualifier
```

PR #2276, opened 2022-07-08
Check for unnecessary visibility qualifier in trait items.

tritao:
This introduces a new check for unnecessary visibility qualifiers when defining trait items.

For code like:

```rust
trait A {} {
    pub fn f(self) -> bool {
        false
    }
}
```

We now emit:

```rust
error
 --> main.sw:4:5
  |
2 |
3 | trait A {} {
4 |     pub fn f(self) -> bool {
  |     ^^^ Unnecessary visibility qualifier.
5 |         false
6 |     }
  |
____
```

Also adds some tests for both trait interface and method blocks.

Closes https://github.com/FuelLabs/sway/issues/2275.


sezna:
It looks like Rust emits two error strings in this case ([reference playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6bcf4d7182739d0354222420e266dfe6)). Could we add the second string to our error message, since it provides more clarity?


> `pub` not permitted here because it's implied

tritao:
> It looks like Rust emits two error strings in this case ([reference playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6bcf4d7182739d0354222420e266dfe6)). Could we add the second string to our error message, since it provides more clarity?
> 
> > `pub` not permitted here because it's implied

Done.

mohammadfawaz (approved):

otrho (approved):
I feel like there might be other places where `pub` isn't wanted, in which case doing this check in the parser might be premature (as opposed to semantic analysis), but I can't think of where, so... all good. :)

Issue #2238, opened 2022-07-06
Forbid the use of storage variables without initializers


mohammadfawaz:
Refer to the discussion in https://github.com/FuelLabs/sway/pull/2222

Currently, all storage slots are initialized to zero in the VM but zero is not always the _desired default_ initializer for a given type. Zero initializers may be desired for integers and booleans but are certainly not desired for enums. `Option<T>` is a good example where the default value should be `None` but `Option::None` is not necessarily stored as zeros (it is in fact _not_ stored as zeros because the tag for `None` is 1). 

We don't currently have a way to specify default values for particular types like Rust:
```rust
impl<T> const Default for Option<T> {
    /// Returns [`None`][Option::None].
    ///
    /// # Examples
    ///
    /// ```
    /// let opt: Option<u32> = Option::default();
    /// assert!(opt.is_none());
    /// ```
    #[inline]
    fn default() -> Option<T> {
        None
    }
}
```
In the absence of the above, we can start by forcing all storage variables to have initializers. In fact, this would be consistent with the behavior of local variables today where the following does not compile because it expects an `=` after `x: u64`.
```rust
fn main() {
    let mut x: u64;
}
```
Once we have the ability to initialize variables (local and in storage) implicitly using `default()`, we can then allow such variables to be declared without explicit initializers.

Braqzen:
StorageMap will be fun

mohammadfawaz:
So that's indeed a drawback. This will work:
```
storage {
    map: StorageMap<u64, u64> = StorageMap { },
}
```

PR #2278, opened 2022-07-08
Make storage initializers required, just like local variables

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/2238

- Make storage initializers required
- Update all the tests and examples
- Improve the docs for storage variables
image


docs/src/blockchain-development/storage.md
```diff
@@ -8,33 +8,23 @@ Some basic use cases of storage include declaring an owner address for a contrac
...
 ```sway
-storage.var1 = v;
+{{#include ../../../examples/storage_variables/src/main.sw:storage_write}}
 ```
 
 To read a storage variable, you also need to use the `storage` keyword as follows:
 
 ```sway
-let v = storage.var1;
+{{#include ../../../examples/storage_variables/src/main.sw:storage_read}}
```
sezna, line 25: good catch

examples/storage_map/src/main.sw
```diff
@@ -9,8 +9,11 @@ struct Data {
 
 storage {
     map1: StorageMap<u64,
-    u64>, map2: StorageMap<(b256,
-    bool), Data>, 
+    u64> = StorageMap {
```
sezna, line 12: @eureka-cpu wen formatter

sezna (approved):
Looks good to me. To summarize, if I understand correctly:
1. Update parser to remove optionality around initializer 
2. remove any optionality in the type checker and constant evaluation of initializers
3. add initializers to all examples which were missing them

No new logic seems to be added, so this seems safe -- just removing the option stuff. 

Nice PR, very thorough with all the examples and sway fmt stuff. 💪🏻

mohammadfawaz:
> Looks good to me. To summarize, if I understand correctly:
> 
> 1. Update parser to remove optionality around initializer
> 2. remove any optionality in the type checker and constant evaluation of initializers
> 3. add initializers to all examples which were missing them
> 
> No new logic seems to be added, so this seems safe -- just removing the option stuff.
> 
> Nice PR, very thorough with all the examples and sway fmt stuff. 💪🏻

That's a good summary. As for the formatter stuff, `sway-fmt-v2` seems to be dropping initializers actually so I filed in an issue here: https://github.com/FuelLabs/sway/issues/2279

examples/storage_map/src/main.sw
eureka-cpu: soon 😇

adlerjohn (approved):

PR #2283, opened 2022-07-09
 Add math & bitwise ops to U256 type 

### LINKED ISSUE #1798: Add math & bitwise ops to `U256` type

rostyslavtyshko:
Closes #1798


Braqzen (commented):
- `u256.sw`
    - Has a commented out import for functions which have been copy and pasted into this module. Could you remove the duplicated functions and re-import them as before?
    - The bitwise traits have the first param as `val: self`. Is there a reason why `self` cannot be used and the `val` part must be specified? It seems to work in the `Shiftable` trait implementation.
    - The `Shiftable` trait has an `else` which contains the `()` type. Afaik this doesn't do anything here. Can it be removed?
    - The `Multiply` trait implementation has a comment about a missing `break` keyword. This is now a feature in sway so you can add that into the `multiply` method (same for `Divide`)
- The tests use the `*` to import. Could you check where that is absolutely needed and change it to explicitly import where it is not?
- Could you also run the formatter using `forc fmt` to format the code?

sway-lib-std/src/u256.sw
```diff
@@ -88,4 +107,268 @@ impl U256 {
...
+        let word_2 = value_word_2 ^ other_word_2;
+        let word_3 = value_word_3 ^ other_word_3;
+        let word_4 = value_word_4 ^ other_word_4;
+        let rebuilt = ~U256::from(word_1, word_2, word_3, word_4);
+        rebuilt
+    }
+}
+
+impl core::ops::Shiftable for U256 {
+    fn lsh(self, shift_amount: u64) -> Self {
```
nfurfaro, line 167: I didn't realize that we would be able to use operators like `<<` if a trait was implemented in `std`. I thought it had to be implemented in `core` for that to work, but I guess because the actual traits are defined in core this works.

nfurfaro (commented):
A few things I commented on in addition to @Braqzen, but this is looking like it's almost there. good work!

sway-lib-std/src/u256.sw
```diff
@@ -88,4 +107,268 @@ impl U256 {
     pub fn bits() -> u32 {
         256
     }
+
+    /// Get 4 64 bit words from a single U256 value.
+    fn decompose(self) -> (u64, u64, u64, u64) {
```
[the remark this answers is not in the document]
adlerjohn: Hmm. So ideally this would use from and into, but that requires additional compiler work IIRC. I wouldn't be against just keeping it as-is until then, since it's going away in the not too distant future.

sway-lib-std/src/u256.sw
```diff
@@ -88,4 +106,263 @@ impl U256 {
...
+            w2 = shifted_3 + carry_4;
+            w3 = shifted_4;
+        } else if w == 2 {
+            let(shifted_4, carry_4) = lsh_with_carry(word_4, b);
+            w1 = (word_3 << b) + carry_4;
+            w2 = shifted_4;
+        } else if w == 3 {
+            w1 = word_4 << b;
+        } else {
+            ();
```
otrho, line 197:
```suggestion
```
Is the final else required by the compiler?

sway-lib-std/src/u256.sw
```diff
@@ -88,4 +106,263 @@ impl U256 {
...
+        local_res = ~U128::from(0, word_2) + ~U128::from(0, other_word_2) + ~U128::from(0, overflow);
+        let result_b = local_res.lower;
+        overflow = local_res.upper;
+
+        local_res = ~U128::from(0, word_1) + ~U128::from(0, other_word_1) + ~U128::from(0, overflow);
+        let result_a = local_res.lower;
+        if local_res.upper != 0 {
+            // panic on overflow
+            revert(0);
+        };
```
otrho, line 263: This could probably be an `assert` like below, although we're moving to using `require` instead of `assert` in library code I think.

sway-lib-std/src/u256.sw
```diff
@@ -88,4 +106,263 @@ impl U256 {
...
+impl core::ops::Multiply for U256 {
+    /// Multiply a U256 with a U256. Panics on overflow.
+    fn multiply(self, other: Self) -> Self {
+        let zero = ~U256::from(0, 0, 0, 0);
+        let one = ~U256::from(0, 0, 0, 1);
+
+        let mut total = zero;
+        // The algorithm loops <from number of bits - 1> to <zero>.
+        // Need to add 1 here to invalidate the while loop once i == 0 since we
+        // don't have a break keyword.
```
otrho, line 316: We have `break` and `continue` now.

sway-lib-std/src/vec.sw
```diff
@@ -14,7 +14,7 @@ struct RawVec<T> {
 impl<T> RawVec<T> {
     /// Create a new `RawVec` with zero capacity.
     fn new() -> Self {
-        Self {
+        RawVec {
```
otrho, line 17: Are these changes required?  I'm surprised if it used to work but no longer does.  Or is it more about verbosity and self-documentation?

otrho (changes requested):
Getting there, some small changes requested and a couple of questions.

I'd also like to see the tests fleshed out some more.  The existing tests could maybe try some bigger numbers, and also we could test for overflow (although I'm not sure how the `should_fail` tests deal with asserts at the moment, they might need to be fleshed out in a different issue/PR).  And tests for `add`, `sub` and shifting should be added.

sway-lib-std/src/vec.sw
mohammadfawaz: It used to be `RawVec` and was changed to `Self` recently by Emily because she introduced a change that allows the use of `Self` in this context. I suggest we keep it as `Self`.

sway-lib-std/src/u256.sw
```diff
@@ -88,4 +106,263 @@ impl U256 {
...
+
+    /// Get 4 64 bit words from a single U256 value.
+    fn decompose(self) -> (u64, u64, u64, u64) {
+        (self.a, self.b, self.c, self.d)
+    }
+}
+
+impl core::ops::Ord for U256 {
+    pub fn gt(self, other: Self) -> bool {
+        self.a > other.a || self.a == other.a && self.b > other.b || self.a == other.a && self.b == other.b && self.c > other.c || self.a == other.a && self.b == other.b && self.c == other.c && self.d > other.d
```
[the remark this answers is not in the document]
rostyslavtyshko: It doesn't, and currently formatter has some nasty bugs like changing -= to unworkable - =

sway-lib-std/src/vec.sw
rostyslavtyshko: Reverted back, as it works now

sway-lib-std/src/u256.sw
```diff
@@ -1,7 +1,26 @@
 library u256;
 
 use core::num::*;
+// use core::ops::{lsh_with_carry, rsh_with_carry};
```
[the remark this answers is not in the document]
rostyslavtyshko: After last changes it works fine, so I deleted the copied code and import existing functions

rostyslavtyshko:
> Getting there, some small changes requested and a couple of questions.
> 
> I'd also like to see the tests fleshed out some more. The existing tests could maybe try some bigger numbers, and also we could test for overflow (although I'm not sure how the `should_fail` tests deal with asserts at the moment, they might need to be fleshed out in a different issue/PR). And tests for `add`, `sub` and shifting should be added.

Added bigger numbers to mul_test and added a `should_fail` test for overflow. Test cases for `add`, `sub` and shifting  are in the u256_ops_test.

sway-lib-std/src/u256.sw
```diff
@@ -311,19 +303,19 @@ impl core::ops::Multiply for U256 {
...
-        // don't have a break keyword.
-        let mut i = 256 - 1 + 1;
-
-        while i > 0 {
-            // Workaround for not having break keyword
-            let shift = i - 1;
+
+        let mut i = 256 - 1;
+
+        while i >= 0 {
```
otrho, line 325: `i` is unsigned, so this will always be true.  The `break` below is always ending the loop so this might as well be `while true` to avoid ambiguity.  Personally I'd prefer a var name like `bit_idx` rather than `i` too.
```suggestion
        while true {
```

sway-lib-std/src/u256.sw
```diff
@@ -341,25 +333,25 @@ impl core::ops::Divide for U256 {
...
```
otrho, line 355: Ditto.
```suggestion
        while true {
```

sway-lib-std/src/u256.sw
```diff
@@ -88,4 +91,286 @@ impl U256 {
...
+                if self.c > other.c {
+                    return true
+                } else if self.c == other.c {
+                    if self.d > other.d {
+                        return true
+                    }
+                }
+            }  
+        }
+        false
```
otrho, line 118: OK, I know I compained about the other long line and suggested nested ifs and return statements, but this isn't great either.  My fault.  But I think this is the same and might be the simplest yet still readable version?
```suggestion
        self.a > other.a
            || (self.a == other.a && self.b > other.b
                || (self.b == other.b && self.c > other.c
                    || (self.c == other.c && self.d > other.d)))
```

otrho (commented):
Yep, other than simplifying `gt` and `lt` again I think this is OK.

otrho (approved):
OK, thanks, all good.  Hopefully the CI problem can be fixed, or was maybe an anomaly.

JoshuaBatty:
I'm not sure why CI would be failing on this test. It just simulates the json messages when opening a file. 

Perhaps rebasing this PR onto master might help, looks like it hasn't been rebased in 20 days.

JoshuaBatty:
hmm still not passing CI after rebasing. It's strange that this test is passing CI for all other PR's except this one.

JoshuaBatty:
Looking closer at the diff. Has this actually been rebased onto master properly? Looking at just the changes under the `sway-lsp` folder and there are heaps of changes that are unrelated to this PR at all....

otrho:
> Looking closer at the diff. Has this actually been rebased onto master properly?

Considering the age of the original PR (you said it was 20 days old?) I imagine a `git merge` would be better suited than a rebase.  But yeah, it doesn't quite look right.

@rostyslavtyshko It might be worth going back to where you were for the original PR and do a `git merge origin/master` to try and get up to date.

JoshuaBatty:
Closing this as the work was already merged in #2445

PR #2285, opened 2022-07-09
Move `std::assert::require` to `std::revert::require`

AlicanC:
This PR creates a new library called `std::require` and moves `std::assert::require` to it.

I believe this separation will help prevent assertion and regular error handling from being confused.

Assertions are useful in that they signal the developer that a condition will always be true and for that reason they can also be stripped out from compiled code and have no runtime cost.

If we agree on this meaning and usage of `assert()`s, then we could also follow this up with an issue for an optimization option to strip `assert()`s.

---

Quoting from: https://en.wikipedia.org/wiki/Assertion_(software_development)
> An assertion is a [predicate](https://en.wikipedia.org/wiki/Predicate_(mathematical_logic)) connected to a point in the program, that always should evaluate to true at that point in code execution. Assertions can help a programmer read the code, help a [compiler](https://en.wikipedia.org/wiki/Compiler) compile it, or help the program detect its own defects.

Following that definiton, we should *not* be using `std::assert`, but instead `std::require` for checking, for example:
- If the caller has forwarded an enough amount of assets.
- If a deadline has passed
- If a signature is valid

We should be using `std::assert` for tests or for cases like in this TypeScript example:
```ts
  async call(options?: CallOptions) {
    // We do regular error handling here because the user can create a Contract
    // without a wallet and call `call`.
    // In production, we can't remove this check. We need to properly notify the user
    // about what was wrong.
    if (!this.contract.wallet) {
      throw new Error('Wallet is required!');
    }
   
    // ... snip ...
    
    return this.call_internal(options);
  }

  async call_internal(options?: CallOptions) {
    // We do an assertion here because even though TypeScript thinks `this.contract.wallet`
    // can be undefined, it is impossible for this to happen.
    // We call this method from only one place and we make this check already.
    // In production, we can safely remove this check.
    assert(this.contract.wallet, 'Expected a wallet.');

    // ... snip ...
  }
  ```

### CODE REVIEW

**Review by adlerjohn — COMMENTED** [review]
I'm not opposed to moving requiring out as a distinct thing from assert. However, I am curious if `assert` is then even useful in this model without a way of removing assertion when doing a release build 🤔. What are your thoughts there?

**Comment by nfurfaro** [comment]
Directly related to #1759

**Comment by otrho** [comment]
> I don't fully understand the idea of stripping out assertions from compiled code. 

This is how it works in C, asserts only trigger in debug code but aren't there for release for perf reasons.  So your tests are supposed to find failed assertions and release code will just be either broken or crash.

So the difference between `assert` and `require` is that `require` logs a code and reverts with a special code other than zero.  I'm also not sure we need both.  And to me it is probably better to pass the revert code to `require` and not necessarily log it, rather than use `42`.  A bit like `errno` in C.

**Comment by nfurfaro** [comment]
> So your tests are supposed to find failed assertions and release code will just be either broken or crash.
This seems very bad in the context of a blockchain where release code is immutable and permanent!?

**Comment by adlerjohn** [comment]
+1 that stripping out checks in release code is dangerous. In this case, it could result in loss of user funds if the contract enters an undefined state. From today's sync call, I'm in favor of removing `assert` completely and just having `revert`. Users can use `()` as the revert error if they're lazy, but really they should have an enum/struct for every possible error.

**Comment by adlerjohn** [comment]
`require` is definitely useful to have, yes.

**Comment by nfurfaro** [comment]
Ok, so it sounds like we could:
- rename the current module to either `std::revert` or `std::require`, 
- remove `assert` completely, keeping `require()` and the raw `revert()` grouped together.

I prefer `std::revert` as this is the base functionality, and the `require` function is just a special-case wrapper around `revert`.

**Comment by AlicanC** [comment]
> @AlicanC I just want to check with you on this, but is it worth removing `assert` as it means replacing it with `requires` in literally every single test script in the e2e test suite?

Right? I was going to ask the same.

**Comment by nfurfaro** [comment]
Yeah, might make sense to just keep `assert` unless there's a burning need to remove it.

**Comment by AlicanC** [comment]
@adlerjohn ?

**Comment by adlerjohn** [comment]
We could keep assert for now then slowly replace it.

**Comment by AlicanC** [comment]
So now `assert` is back and this is almost like the initial PR, but instead of moving `require` to its own `std::require`, we move it to `std::revert` (which I believe is an improvement).

**Review by otrho — APPROVED** [review]

File: sway-lib-std/src/assert.sw [hunk 1]
```diff
@@ -1,25 +1,14 @@
...
-use ::logging::log;
 use ::revert::revert;
 
-const FAILED_REQUIRE_SIGNAL = 42;
-
-/// Assert that a value is true
-pub fn assert(a: bool) {
-    if !a {
+/// Asserts that the given `condition` will always be `true` during runtime.
+/// To check for conditions that may not be `true`, use `std::revert` instead.
```
nfurfaro, line 6: @AlicanC Was this meant to read:
 "/// To check for conditions that may not be `true`, use `std::revert::require` instead." ?

test/src/sdk-harness/test_artifacts/methods_contract/src/main.sw
```diff
@@ -6,7 +6,7 @@ use std::result::*;
 use std::identity::*;
 use std::chain::auth::*;
 use std::option::*;
-use std::assert::*;
+use std::revert::*;
```
nfurfaro, line 9: Maybe see how many of these `*` imports can be cleaned up. Some of them were required at one point and no longer are.

nfurfaro (commented):
a couple of small comments but looking good overall.

sway-lib-std/src/assert.sw
AlicanC: Yup, fixed.

test/src/sdk-harness/test_artifacts/methods_contract/src/main.sw
AlicanC: `std::result::*` wasn't being used, removed.

mohammadfawaz (approved):

adlerjohn (commented):
lgtm, but would like @otrho to confirm

otrho (approved):

PR #2289, opened 2022-07-11
feat: wrapper for low-level call opcode

### LINKED ISSUE #2144: Add a function to the std-lib to wrap the CALL opcode for low-level calls
It would be useful to have a function in the std that wraps the low-level CALL opcode specified [here](https://github.com/FuelLabs/fuel-specs/blob/1be31f70c757d8390f74b9e1b3beb096620553eb/specs/vm/instruction_set.md#call-call-contract)

The function would be used to call external contracts not known at compile time.  For example, a proposal in a DAO voting contract.

The signature for such a function might look like `make_call(call_data: CallData, amount: u64, asset: ContractId, gas: u64)`

matt-user:
closes #2144
[related opcode](https://github.com/FuelLabs/fuel-specs/blob/1be31f70c757d8390f74b9e1b3beb096620553eb/specs/vm/instruction_set.md#call-call-contract)


sezna:
Hey Matt, great work! Before continuing on this, though, do see my comment here: https://github.com/FuelLabs/sway/issues/2144#issuecomment-1186920584

Issue #2179, opened 2022-06-30
Add the `GTF` opcode to the compiler


nfurfaro:
ref: https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/instruction_set.md#gtf-get-transaction-fields

mohammadfawaz:
Can now be closed with https://github.com/FuelLabs/sway/pull/2315

Issue #810, opened 2022-02-17
Allow use of `consts` in place of immediates in asm blocks


sezna:
## Motivation
https://github.com/FuelLabs/sway-lib-std/pull/46 
We should be able to directly use the `const` in the asm block if it is small enough to fit in that immediate size.

## Summary

Currently,  immediates are size-checked in the compiler. If an opcode takes a 12-bit immediate, the size of that immediate is checked at compile time. Since we also know the value of `const`s at compile time, we should be able to directly allow the use of `const`s inside of `asm` blocks instead of passing them in. This will require a modification to the grammar and then size checking the const.

otrho:
Could be done along with #801.

sezna:
Also requires a compile time check that the constant is within the immediate size for a particular opcode.

nfurfaro:
This would be a huge help for exposing all of the `gtf` opcodes functionality, as I could condense ~70 asm blocks with hardcoded immediates down to a single asm block in a function called ~70 times with different const values!

mohammadfawaz:
The alternative for `gtf` proposed yesterday is https://github.com/FuelLabs/sway/issues/2310. I suspect this particular PR is not trivial to implement (not hard but not trivial). https://github.com/FuelLabs/sway/issues/2310 on the other hand is trivial and allows us to proceed quickly I think.

PR #2292, opened 2022-07-11
Refactor `std::tx` to use `gtf` opcode

### LINKED ISSUE #2146: Update the std library to use `gtf` in place of hardcoded offsets
blocked by: implement `gtf` opcode in compiler.

nfurfaro:
Lots of changes in this one, but it won't compile yet as the `gtf` opcode hasn't been fully implemented yet.
- I've tried to avoid renaming _for the most part_ to minimize breaking changes.
- Some of the fields we need to read are exposed under more than one hex immediate, for example:
`GTF_INPUT_COIN_TX_ID` & `GTF_INPUT_CONTRACT_TX_ID` both read the same field, and this is a bit of a pattern.
The approach I've taken is to expose a single wrapper function which matches on the type of Input/Output, and returns an `Option::Some` if possible, else an `Option::None`. 
- I haven't exposed the individual getters directly (i.e: this has a `tx_input_owner()` function, but not `tx_input_coin_owner()` or `tx_input_message_owner()` functions.) but it wouldn't be hard to do so.
- Some follow-up PRs should probably split out a few different modules from `std::tx`while adding the remaining functions, as this thing is going to get pretty big with the addition of all the `gtf` variants (~64 wrapper functions in total).

blocked by: #2179 
closes #2146


simonr0204:
What's the `index` required everywhere as a parameter? For example `tx_type(index: u64)`

sway-lib-std/src/tx.sw
```diff
@@ -10,41 +10,95 @@ use ::mem::read;
...
+const GTF_SCRIPT_GAS_LIMIT = 0x003;
+const GTF_SCRIPT_MATURITY = 0x004;
+const GTF_SCRIPT_SCRIPT_LENGTH = 0x005;
+const GTF_SCRIPT_SCRIPT_DATA_LENGTH = 0x006;
+const GTF_SCRIPT_INPUTS_COUNT = 0x007;
+const GTF_SCRIPT_OUTPUTS_COUNT = 0x008;
+const GTF_SCRIPT_WITNESSES_COUNT = 0x009;
+const GTF_SCRIPT_RECEIPTS_ROOT = 0x00A;
+const GTF_SCRIPT_SCRIPT = 0x00B;
+const GTF_SCRIPT_SCRIPT_DATA = 0x00C;
```
nfurfaro, line 28: Now that I added these, I don't think I can even use them.

simonr0204:
> > What's the index required everywhere as a parameter? For example tx_type(index: u64)
> 
> The `gtf` opcode requires an Imput/Output index as a param for all operation.

But since it's the same transaction type (and gas price, limit, script bytecode, etc) for all inputs and outputs, shouldn't we just hard-code 0 as the index for those fields that apply to the whole transaction?

nfurfaro:
>But since it's the same transaction type (and gas price, limit, script bytecode, etc) for all inputs and outputs, shouldn't we just hard-code 0 as the index for those fields that apply to the whole transaction?

Possibly? I'm not 100% sure if input//output ordering is fixed (or even matters), but I don't know how a user is expected to know the index of the Input/Output they're interested in otherwise.
Hardcoding the index to `0` would certainly make things much simpler.

simonr0204:
> > But since it's the same transaction type (and gas price, limit, script bytecode, etc) for all inputs and outputs, shouldn't we just hard-code 0 as the index for those fields that apply to the whole transaction?
> 
> Possibly? I'm not 100% sure if Input/Output ordering is fixed (or even matters), but I don't know how a user is expected to know the index of the Input/Output they're interested in otherwise. Hardcoding the index to `0` would certainly make things much simpler.

Input/output ordering is one thing, but my comment is more that there are field that are common across the whole _transaction_. E.g. the gas price doesn't have anything to do with a given input or output. It's a property of the transaction.

sway-lib-std/src/tx.sw
```diff
@@ -60,120 +114,116 @@ pub const OUTPUT_VARIABLE = 4u8;
 pub const OUTPUT_CONTRACT_CREATED = 5u8;
 
 /// Get the transaction type.
-pub fn tx_type() -> u8 {
-    asm(r1, r2: TX_TYPE_OFFSET) {
-        lw r1 r2 i0;
-        r1: u8
+pub fn tx_type(index: u64) -> u8 {
```
mohammadfawaz, line 117: `index` here is not really required right? From the spec, `index` is only useful and required for some of the tx fields like `GTF_SCRIPT_INPUT_AT_INDEX`

sway-lib-std/src/tx.sw
mohammadfawaz: I assume you're blocked by https://github.com/FuelLabs/sway/issues/810 for this right?

sway-lib-std/src/tx.sw
nfurfaro: yup, exactly.
>looks like I need to give this another pass to find all such fields and hardcode the index to 0 for them (an index is mandatory for gtf usage).

mohammadfawaz:
@nfurfaro https://github.com/FuelLabs/sway/pull/2315 is now in and you should be able to use the `__gtf` intrinsic instead of `asm` blocks. Some documentation for the intrinsic can be found in the description but please let me know if you have any questions!

PR #2294, opened 2022-07-11
Adds `ItemTrait` to `sway-fmt-v2`

### LINKED ISSUE #2070: Add `trait` formatting to `sway-fmt-v2`

eureka-cpu:
Closes #2070


sway-fmt-v2/src/fmt.rs
```diff
@@ -391,6 +401,51 @@ where
...
+    fn university(self) -> String;
+}
+
+trait Programmer {
+    fn fav_language(self) -> String;
+}
+
+trait CompSciStudent: Programmer + Student {
+    fn git_username(self) -> String;
+}"#;
```
mitchmindtree, line 430: It looks like there's no difference between the pre-formatted and expected output in this test - should we mess up the `sway_code_to_format` a bit here?

sway-fmt-v2/src/items/item_trait.rs
```diff
@@ -1,18 +1,116 @@
...
+        // `trait name`
+        write!(
+            formatted_code,
+            "{} {} ",
+            self.trait_token.span().as_str(),
+            self.name.span().as_str()
+        )?;
+        // `: super_trait + super_trait`
+        if let Some(super_traits) = &self.super_traits {
+            formatted_code.pop(); // pop the ending space if there is a `super_trait`
```
mitchmindtree, line 25: Rather than `pop`ping from `formatted_code` in cases like this, we might want to consider omitting the extra space from the previous formatting and re-adding it later on when necessary or writing it in an `else` block. This might make future refactoring easier if we decide to change `formatted_code` from `&mut String` to some other type of buffer or `T: fmt::Write` that does not provide a `pop` method.

sway-fmt-v2/src/items/item_trait.rs
```diff
@@ -1,18 +1,116 @@
...
+            write!(
+                formatted_code,
+                "{}",
+                formatter.shape.indent.to_string(formatter)
+            )?;
+            // format `Annotated<FnSignature>`
+            trait_items.0.format(formatted_code, formatter)?;
+            writeln!(formatted_code, "{}\n", trait_items.1.span().as_str())?;
+        }
+        formatted_code.pop(); // pop last ending newline
```
mitchmindtree, line 41: Same here

sway-fmt-v2/src/fmt.rs
```diff
@@ -74,6 +74,10 @@ impl Formatter {
             &mut formatted_code,
             &raw_formatted_code,
         )?;
+        if formatted_code.ends_with('\n') {
+            formatted_code.pop();
```
mitchmindtree, line 78: and here

mitchmindtree (changes requested):
Looking good! Just left some comments for a couple minor things

Also, it looks like a `Cargo.lock` might have been downgraded accidentally as a part of this PR - perhaps a merge conflict resolution error?

sway-fmt-v2/src/fmt.rs
eureka-cpu: Ah yes, I forgot to do that thanks 😅

sway-fmt-v2/src/fmt.rs
eureka-cpu: Was having some kind of issue with a `None` value returning the block for `Some` when I refactored this as a peekable iterator. May take me a bit to figure out what's going on, this was my short term solution.

sway-fmt-v2/src/items/item_trait.rs
eureka-cpu: This one was an easy fix, see my comment below for more details on current status

mitchmindtree (approved):
:ship:

kayagokalp (approved):
🚀

Issue #2226, opened 2022-07-05
The formatter should alphabetize `use` statements if they're not separated by an empty line


mohammadfawaz:
We already alphabetize items within a single `use` statements. It would be nice to also alphabetize the `use` statements themselves if they're not separate by an empty line. This replicates the behavior of the Rust formatter.

eureka-cpu:
> This will be included in the `ItemUse::format()` functionality !

Since the deadline is approaching I've unassigned for the time being since I will be working on finishing out the basic features. If someone picks it up by all means! if not I will continue once the MVP is finished.

eureka-cpu:
To be clear, it currently does sort alphabetically just in the scope of each use statement and not as a group of use statements.

eureka-cpu:
@cr-fuel to do this you'll need to add some extra logic to separate the `use` statements into individual blocks to be formatted with what is already there. You'll need to collect the use statements first, by looking ahead for the next use statement. If it finds there are two consecutive newlines, only then you should format the previous block.

PR #2295, opened 2022-07-11
Adds `ItemUse` to `sway-fmt-v2`

### LINKED ISSUE #1808: Formatter incorrectly reformatting use declarations that exceed line length limit
before reformatting 
``` rust
use std::{address::Address, assert::assert, chain::auth::*, context::*, context::call_frames::*, token::*};
```

after reformatting: 
``` rust
stduse:: {
    address::Address, assert::assert, chain::auth:: * , context:: * , context::call_frames:: * , token:: * 
};
```

This then fails to build. 

setup:
``` sh
forc 0.14.5
fuel-core v0.8.0
```

Editor: VSCode

### LINKED ISSUE #2298: Add `ItemUse` to `sway-fmt-v2`

eureka-cpu:
Opening this up for visibility.

- [x] #2226 <-- I've added sorting but only in the scope of each import, not the collective of imports
- [x] Closes #1808
- [x] #2299 #2390

Closes #2298


sway-fmt-v2/src/items/item_use/tests.rs
```diff
@@ -0,0 +1,103 @@
...
+                            }
+                        }
+                    }
+                }
+                assert_eq!(&formatted_code, $desired_output)
+            }
+        }
+    )+
+}
+}
```
mitchmindtree, line 80: It could be nice to have a brief comment on these macros explaining what's happening, just to give future readers/maintainers a head-start.

sway-fmt-v2/src/items/item_use/tests.rs
```diff
@@ -0,0 +1,103 @@
...
+                        DiffOp::Replace(old, new) => {
+                            for o in old {
+                                println_red(&format!("-{}", o));
+                            }
+                            for n in new {
+                                println_green(&format!("+{}", n));
+                            }
+                        }
+                    }
+                }
```
mitchmindtree, line 74: Do we do this diff printing with different colours in other places? If so, maybe we can move this to a `print_diff` function or similar? If not, we can probably wait until we do this in multiple places before we refactor it out.

mitchmindtree (approved):

kayagokalp:
Merging master (as #2518 merged) here to resolve CI failure

sway-fmt-v2/src/items/item_use/tests.rs
eureka-cpu: Since it's a copypasta I'll go ahead an update both in a separate PR

sway-fmt-v2/src/items/item_use/tests.rs
eureka-cpu: It is in one other place, `expr`, but I can do that in another PR as well. Perhaps the same PR for adding comments describing the macros

kayagokalp (approved):
Tested locally, works & looks perfect!

Maybe something that we can address with the later PR you mentioned in @mitchmindtree's review: After this PR merges we only have `item_use` folder (not counting `expr` as it resides in utils folder) which holds the tests for formatting. For consistency maybe we can either have this kind of structure for all of the items and removing the corresponding tests from `fmt.rs`, or move these tests to `fmt.rs` (possibly expending/reusing your cool macro for other tests)

PR #2206, opened 2022-07-03
#2205 - Use ZERO_B256 const to initialize BASE_ASSET_ID const

r-sitko:

PR #2296, opened 2022-07-11
refactor: refactor call_frames to use registers read function 

### LINKED ISSUE #2284: no sig mathod for msg function
sway doesn't have msg_sign function which return The first four bytes of the calldata for a function that specifies the function to be called.

matt-user:
closes #2284


sway-lib-std/src/context/call_frames.sw
```diff
@@ -43,19 +45,15 @@ pub fn code_size() -> u64 {
...
+    read(frame_ptr() + 584)
 }
 
 /// Get the second parameter from the current call frame.
-pub fn second_param() -> u64 {
-    asm(size, ptr, offset: 592) {
-        add size fp offset;
-        size: u64
-    }
+/// Where 592 is the current offset in bytes from the start of the call frame.
```
nfurfaro, line 54: same comment as above.

nfurfaro (commented):
Looking good.Just a few comments.

sway-lib-std/src/context.sw
```diff
@@ -1,11 +1,11 @@
 //! Functionality for accessing context-specific information about the current contract or message.
 library context;
-dep context/call_frames;
 dep context/registers;
+dep context/call_frames;
```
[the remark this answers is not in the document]
matt-user: Yes, if the deps are ordered alphabetically then I get an error in call_frames saying it can not resolve the module registers.

nfurfaro (dismissed):
LGTM!

sway-lib-std/src/context/call_frames.sw
```diff
@@ -43,19 +49,13 @@ pub fn code_size() -> u64 {
 }
 
 /// Get the first parameter from the current call frame.
-pub fn first_param() -> u64 {
-    asm(size, ptr, offset: 584) {
-        add size fp offset;
-        size: u64
-    }
+pub fn first_param<T>() -> T {
```
adlerjohn, line 52: The first parameter is always a `u64` (technically `u32`), so don't think this needs to be generic.

sway-lib-std/src/context/call_frames.sw
```diff
@@ -12,6 +14,10 @@ use ::contract_id::ContractId;
 
 const SAVED_REGISTERS_OFFSET: u64 = 64;
 const PREV_FRAME_POINTER_OFFSET: u64 = 48;
+/// Where 584 is the current offset in bytes from the start of the call frame.
+const FIRST_PARAMETER_OFFSET: u64 = 584;
+/// Where 592 is the current offset in bytes from the start of the call frame.
+const SECOND_PARAMETER_OFFSET: u64 = 592;
```
Braqzen, line 20: I wonder if this should be `const SECOND_PARAMETER_OFFSET: u64 = FIRST_PARAMETER_OFFSET + 8;`

test/src/sdk-harness/test_projects/call_frames/mod.rs
```diff
@@ -41,14 +41,30 @@ async fn can_get_code_size() {
 async fn can_get_first_param() {
     let (instance, _id) = get_call_frames_instance().await;
     let result = instance.get_first_param().call().await.unwrap();
-    assert!(is_within_range(result.value));
+    // The hash of "get_first_param" is 0x35fd8826
```
Braqzen, line 44: Can this be fetched and then converted to a value automatically instead of hardcoding a magic number with a comment?

sway-lib-std/src/context/call_frames.sw
Braqzen: Strange, I thought this would work given #2206

sway-lib-std/src/context/call_frames.sw
```diff
@@ -44,18 +50,12 @@ pub fn code_size() -> u64 {
...
+    read(frame_ptr() + FIRST_PARAMETER_OFFSET)
 }
 
 /// Get the second parameter from the current call frame.
-pub fn second_param() -> u64 {
-    asm(size, ptr, offset: 592) {
-        add size fp offset;
-        size: u64
-    }
+pub fn second_param<T>() -> T {
```
mohammadfawaz, line 57: I also think the second parameter doesn't need to be a generic. The second parameter itself is a u64 and in your tests T is actually a u64 that you then treat as a pointer and read from. I think if you change T to u64, everything should continue to work as expected.

test/src/sdk-harness/test_projects/call_frames/src/main.sw
```diff
@@ -21,7 +22,16 @@ impl CallFramesTest for Contract {
         first_param()
     }
 
-    fn get_second_param() -> u64 {
+    fn get_second_param_u64(arg0: u64) -> u64 {
         second_param()
     }
+
+    fn get_second_param_bool(arg0: bool) -> bool {
+        read(second_param())
```
mohammadfawaz, line 30: Is this read needed here? Bool and u64 should exhibit the same behaviour. I believe a read is only needed for multiple arguments or single arguments of reference types. You could test this out by covering the possible cases:
- Single copy type argument (u64 or bool). You already have tests for that but the bool probably needs an small update.
- Single reference type argument (b256 or struct for example). This is missing a test case.
- Multiple copy type arguments. You have a test for that but would be nice to throw in some reference types in there.

mohammadfawaz (commented):
One more comment on the tests and I'm done, I promise 😃

Sorry, should've made all my comments in one go.

test/src/sdk-harness/test_projects/call_frames/src/main.sw
matt-user: since I changed second_param to return a u64 instead of T I think I'm unable to just return second_param() like I had before

matt-user: I'll do some experimenting

mohammadfawaz: Ohhh you're right. This is annoying because second_param is meant to give you the value itself if you only have a single copy type argument and so calling read on the result is not really correct I think. So what you want is a cast from u64 to bool. 

The fact that param2 sometimes contains a value and sometimes contains a pointer is what's causing the confusion here. So maybe we should go back to keeping second_param a generic and calling read inside it with <T> and see if this covers all the cases without having to call read anywhere else in the tests?

I'm sorry for the back and forth on this. This is what I get for trying to be helpful from my phone on a day off 😞

matt-user: All good I appreciate the help you've given me!  Enjoy your day off :)

matt-user: so should second_param() ever return a pointer or should it handle pointers itself?  That is should second_param() handle reading  such that the test functions don't have to.

mohammadfawaz: The cleanest implementation seems to be to have second_param return T and just call read inside it with <T> as a type argument. Then just calling second_param from the test with the same type arguments should give you the right result in all cases I believe.

mohammadfawaz: But we should document the meaning of T in this case and how to set it properly based on the arguments of the contract method.

matt-user: hmm I tried
```
pub fn second_param<T>() -> T {
    read::<T>(frame_ptr() + SECOND_PARAMETER_OFFSET)
}
```
and
```
fn get_second_param_multiple_params(arg0: bool, arg1: u64) -> (bool, u64) {
        second_param::<(bool, u64)>()
}
```
But it's not working as intended.  I'll take another look at it later.

mohammadfawaz: So I looked into this a bit and it seems that one solution is to implement `second_param` as follows:
```rust
/// Get the second parameter from the current call frame.
pub fn second_param<T>() -> T {
    if !is_reference_type::<T>() {
        read::<T>(frame_ptr() + SECOND_PARAMETER_OFFSET)
    }
    else {
        read::<T>(read::<u64>(frame_ptr() + SECOND_PARAMETER_OFFSET))
    }
}
```
Basically, if `T` is a reference type, then we need to read `param2` itself as a `u64` and then read from the pointer stored there. Otherwise, `param2` is just the value itself (this would be the case of single argument that is an integer or bool).

Another alternative is to implement 4 functions as follows:

- `fn first_param()`: gives back the raw first param as `u64` using a single `read()`.
- `fn second_param(0`: gives back the raw second param as `u64` using a single `read()`.
- `fn selector()`: returns the selector which happens to be the first parameter, so we just call `first_param()` in the body of `fn selector()`.
- `fn call_data()`: gives back the call data implemented as above.

What do you think? 
cc @nfurfaro @adlerjohn

mohammadfawaz: With this, we never need to call `read()` from the test.

mohammadfawaz (approved):
LGTM! Thanks!

test/src/sdk-harness/test_projects/call_frames/mod.rs
```diff
@@ -41,14 +44,62 @@ async fn can_get_code_size() {
...
+    // Grab the first 4 bytes of the hash per https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#function-selector-encoding
+    let function_name_hash = &function_name_hash[0..4];
+    // Convert the bytes to decimal value
+    let selector = function_name_hash[3] as u64 + 256 * (function_name_hash[2] as u64 + 256 * (function_name_hash[1] as u64 + 256 * function_name_hash[0] as u64));
+    assert_eq!(result.value, selector);
 }
 
 #[tokio::test]
-async fn can_get_second_param() {
+async fn can_get_second_param_u64() {
```
Braqzen, line 60: It's good to see that there is a mixture of params unfortunately this is not an exhaustive list. For these tests to thoroughly prove that the second param works as expected we would need to perform at least a check on each native type and it only gets more complex considering the permutations.

That being said having 1 test for each native type and then another few for a mixture of types would generally suffice as a bug being discovered in the permutations is an unlikely edge case if each native type works and a few tests which mix also pass.

Could you add some tests to make sure that each native type is checked?

Braqzen (changes requested):

test/src/sdk-harness/test_projects/call_frames/mod.rs
matt-user: @mohammadfawaz what do you think?

Braqzen: hm, I see. As long as `read()` is thoroughly tested then this does not need to be

Braqzen (approved):

test/src/sdk-harness/test_projects/call_frames/mod.rs
matt-user: I could add a test for a mixture of reference and copy types as [this](https://github.com/FuelLabs/fuels-rs/issues/470) was just merged.

Braqzen: This looks like an SDK issue rather than a Sway issue so I don't believe that to be necessary

test/src/sdk-harness/test_projects/call_frames/src/main.sw
nfurfaro: I definitely would prefer to not have to call `read()` from the test.
It sounds like the second alternative may unblock @matt-user for his work on a wrapper function for `call`?

PR #2302, opened 2022-07-12
Call `sway_core::clear_lazy_statics` at the beginning of build and check

JoshuaBatty:
PR #2290 introduced a temporary solution for clearing the type engine. 

I've moved this call to the top of `compile` and also added it to the beginning of the `check` function. As we call `check` on every didChange event in the language server we can now reset the type engine to an empty state on each call. 

This call should be removed once the type engine isn't residing in global memory.

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -1587,6 +1587,9 @@ pub fn compile(
     namespace: namespace::Module,
     source_map: &mut SourceMap,
 ) -> Result<(Compiled, Option<namespace::Root>)> {
+    //TODO remove once type engine isn't global anymore.
+    sway_core::clear_lazy_statics();
+
     // Time the given expression and print the result if `build_config.time_phases` is true.
     macro_rules! time_expr {
         ($description:expr, $expression:expr) => {{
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Looks like CI is showing lots of these

>  thread 'main' panicked at 'index out of bounds: the len is 342 but the index is 625', sway-core/src/concurrent_slab.rs:22:9

There must be some interaction with the type engine that we depend on that happens *prior* to `check` and `compile` that's getting cleared here :thinking:

The joys of global state :woozy_face:

forc-pkg/src/pkg.rs
```diff
@@ -1587,6 +1587,9 @@ pub fn compile(
     namespace: namespace::Module,
     source_map: &mut SourceMap,
 ) -> Result<(Compiled, Option<namespace::Root>)> {
+    //TODO remove once type engine isn't global anymore.
+    sway_core::clear_lazy_statics();
```
mitchmindtree, line 1591: Ahhh this should be at the beginning of `build`, not `compile` - otherwise we clear all the type state of dependency packages.

mohammadfawaz (approved):
Thanks!

mitchmindtree (approved):

emilyaherbert (approved):

Issue #2305, opened 2022-07-12
Change `--verbose` flag behaviour to increase the level of verbosity within the logger


mitchmindtree:
Currently as of #2244 we handle the `--verbose` flag by manually checking the `command.verbose` condition before outputting certain content via `info!`.

Instead, we should switch back to using `debug!`, but have the `--verbose` flag increase the log level. E.g. if the default is `info!`, using `--verbose` would increase it to `debug!`.

This will allows us to avoid checking a `command.verbose` condition everywhere, and allow us to take advantage of the existing logging support in supporting `--verbose`.

mitchmindtree:
For future reference, I've elaborated a little further on what the implementation for this might involve in [this comment](https://github.com/FuelLabs/sway/pull/2308#pullrequestreview-1036769377).

sdankel:
> but have the --verbose flag increase the log level. E.g. if the default is info!, using --verbose would increase it to debug!

I wonder if it would be better to have `--verbose`/ mean the same thing across all systems (DEBUG level) and have a `-vv` option for the most verbosity (TRACE level). It *looks* like that's how cargo's `--verbose` and `-vv` work. If users want to target a level other than DEBUG/TRACE they can set `RUST_LOG`.

mitchmindtree:
Yeah I think that sounds good, I suggested something similar in [the linked comment](https://github.com/FuelLabs/sway/pull/2308#pullrequestreview-1036769377), but clarifying that `-v` could be DEBUG and `-vv` could be TRACE sounds good!

PR #2308, opened 2022-07-12
Upgrade verbosity if --verbose passed

ra0x3:


mitchmindtree (commented):
Ahh it looks like the aim for this PR is to address #2305, however there seems to be a misinterpretation of the issue (or lack of clarity on my part!).

My aim with #2305 was to suggest that we use the `--verbose` option to check whether we should upgrade the level of verbosity within the tracing subscriber itself, not at every call-site of potentially verbose output.

## Context

Currently, we call `init_tracing_subscriber()` at the beginning of `main` which instantiates the global subscriber. By default, this retrieves the log-level or "filter" from the `RUST_LOG` environment variable, or falls back to `info`. My thinking is that if we want to support something like `--verbose` we should do so by modifying the filter level here to a more verbose level (either `debug` or `trace`).

It might be easier to achieve this by having the `--verbose` flag as an option on the top-level `forc` command, rather than having it as an option on multiple subcommands. Otherwise we would need to delay initialisation of the subscriber until we parse each subcommand's options.

## Remove `--verbose`?

Alternatively, we could remove the `--verbose` flag altogether in favour of requiring that users specify the desired log-level via the `RUST_LOG=<level>` env var, seeing as this was part of the motivation behind using `tracing` in the first place?

If we do want to allow specifying the log-level via CLI args, we might be better off doing so via a `--log-filter=<level>` flag. That way, if we decide we want to include `--verbose`, it could act as an alias for `--log-filter=trace`.

PR #2311, opened 2022-07-12
Inserting Comments

### LINKED ISSUE #2122: Handle comments in `sway-fmt-v2`

kayagokalp:
## About this PR
closes #2122.

Comments are stripped in the parsing step which causes them to disappear in the formatted output because the formatter reconstructs the code from the parsed source code. With this PR `sway-fmt-v2`will be able to conserve the comments as they were inserted by the user. To accomplish that, the following concepts are introduced:

1. A mapping between `Span` -> `Comment`
2. A visitor that can visit and collect possible positions inside a parsed code piece.
3. While finding comments, collecting some context so that user-given indentations, etc are intact (we are not applying any formatting logic to the `Comment`s yet)

#### Mapping between `Span` -> `Comment`

This is used for fast retrieval of comments with a given range of spans. This is essential since to insert comments we should first find them with respect to items that are also present in the formatted output. To do that we will be searching for comments in possible spots

#### Visitor that can visit and collect possible positions inside a parsed code piece.

> To do that we will be searching for comments in possible spots for possible positions. 

To be able to accomplish this goal of ours we would either need to implement searching logic for all possible spanned pieces in the code like the following:

```rust
impl AddComments for ItemUse {
    // ...
}
impl AddComments for ItemFn {
    // ...
}
.
.
.
```

or we should have a generic logic for searching & insertion of comments with given possible spans. This requires us to be able to visit each possible parsed code piece and collect possible spans from that code piece. Then a generic search & insert would be able to insert all the comments for all possible places that a comment can be seen. Since this seemed a cleaner and shorter way, I followed this path and introduced a `CommentVisitor` trait. If `CommentVisitor` is implemented for a type we can get all of the possible `Span`s by simply calling `collect_spans()`.

**A little side note here:** Since for BTreeMap to keep an ordering of `Span` -> `Comment`, I needed `Span` to implement `Ord`. Since `Span` is heavily used in the other parts of the code as well and for comment insertion, I do only require the `start` and `end` fields of it, I created a stripped version that also implements `Ord` (`CommentSpan`).

#### Collecting some context while finding comments from unformatted code

Since we are finding the comment by simply searching in the mapping for the given range of spans, we are only getting the comment itself. Inserting just the comment itself deletes the user's original alignment whitespaces. We are collecting some extra information (context) while searching for comments to conserve the user's alignment.

### General structure of `Comment` insertion

1. We get the unformatted version of the code, and parse it to get the unformatted code's `module`.
2. We format the unformatted version of the code and parse it again to get the formatted code's `module`.
3. We collect spans for these two modules and traverse them together. (This way we have a way of mapping a `span` range from unformatted code to a `span` range in formatted code. This is required since we do not have another way of finding the correct spot for the comment in the formatted code because when the code is formatted places of `item`s etc can heavily change in various scenarios. For example, input with a large number of new lines between items can be considered. When such input is formatted, the number of newlines between `item`s will reduce dramatically. Although we will be able to find the `comment`s easily in the unformatted code placing them in the correct places w.t.r to the item they were originally commenting on would be an open problem if we didn't traverse both formatted and unformatted code together.). 
4. In each iteration of the traverse we search comments between the last `CommentSpan` and the current one. If we can find some we are inserting them with their collected context to the corresponding place in the formatted output.

### Current status of `Comment`s

This is a little tricky to write about since while working on comment stuff I had a couple of "a-ha moments" about `comment` insertion. I think this can be enough to reach MVP formatter but this part of the code will require a little bit more love. Some points to keep in mind around `Comment` stuff:

1. If a structure was multiline in unformatted code and formatter transformed that stuff to a single line, comment formatting needs to be aware of that because single-line comments that are valid in the form of a multiline structure might not be valid in a single-line case. (In that case, we can either transform the comments to something like `/*...*/` from `//...` or we can prevent single-line formatting while we have comments.)
2. We might want to collect more meaningful context for formatting the actual comment.

A final note about the current status: until `expr` formatting lands we will have duplicate comments inside `expr`'s since the formatter also adds them (because it is directly pushing raw `str` until it is implemented. @eureka-cpu  brewing some cool stuff for that over #2338)

## Initial explanation of this PR (written while it was still a draft leaving this here for visibility purposes) 

> This is very WIP, opening up for visibility and discussion.
> 
> ~Please mind that I created this branch from #2229 since it is not merged yet, GitHub diff shows that changes as a part of this one too.~
> 
> Comments are not included in the AST which requires us to construct a mapping from `Span` -> `Comment` and search for them in possible places. Ideally, we should be able to do it as generic as possible and while doing so we should be preserving the whitespace around user comments as they might be intentional (at least for now, we will be exploring how to format them in a later PR).
> 
> 
> This implementation tries to be as generic as possible this is trivial once we are in the `Item` level like the following example:
> 
> 
> ```rust
> // This is a comment
> struct Foo {
>     baz: u64,
> }
> // This is another comment
> struct Baz {
>     foo: u64,
> }
> ```
> 
> Finding and inserting these types of comments are easier since they can be searched between items that all have `Span`s so we can have a generic implementation for finding and adding them easily. Currently, this is not implemented. 
> 
> The problem is visiting all possible places in the `Item` which might contain a comment in between. For example:
> 
> ```rust
> struct Foo {
>     baz: u64, // This is another comment
>     fooo: u64,
> }
> ```
> 
> Finding out (and adding) these type of comments require us to visit each possible place in a struct (like in between fields or after `{`, etc.). For searching for the comments we require the spans of the things we are going to be searching in between (like from the `baz` field to the `fooo` field we should be searching for comments.)
> 
> To be able to implement this as generic as possible I created this draft PR to further our previous discussion yesterday @mitchmindtree @eureka-cpu.
> 
> With this type of implementation, we will need to implement the following portion of the code for items that we are going to be visiting (like `enum`s, `struct`s, `fn`s ...):
> 
> ```rust
> impl CommentVisitor for ItemStruct {
>     fn collect_spans(&self) -> Vec<CommentSpan> {
>         let mut collected_spans = Vec::new();
>         // Add visibility token's span if it exists
>         if let Some(visibility) = &self.visibility {
>             collected_spans.push(CommentSpan::from_span(visibility.span()));
>         }
>         // Add struct_token's span
>         collected_spans.push(CommentSpan::from_span(self.struct_token.span()));
>         // Add name's span
>         collected_spans.push(CommentSpan::from_span(self.name.span()));
>         // Add generics' span if it exist
>         if let Some(generics) = &self.generics {
>             collected_spans.push(CommentSpan::from_span(generics.parameters.span()))
>         }
>         // Collect fields' spans.
>         collected_spans.append(&mut self.fields.collect_spans());
>         collected_spans
>     }
> }
> ```
> 
> This implementation currently formats the following input:
> 
> ```rust
> contract;
> 
> struct Foo { // Here I am 4
>     // Here I am
>     barasdd: u64, // Here i am 2 
>     baz: bool,
> }
> struct Foo2 {
>     barasdd: u64,
>     baz: bool,
> }
> ```
> 
> to
> 
> ```rust
> contract;
> 
> struct Foo {// Here I am 4
> // Here I am
>     barasdd: u64,// Here i am 2 
>     baz: bool,
> }
> struct Foo2 {
>     barasdd: u64,
>     baz: bool,
> }
> ```
> ## To Do
> 
> - [x] Collect more context regarding the user's whitespace around their comments
> - [x] Implement handling of comments in-between `Item`s
> - [x] Implement `CommentVisitor` for other item type's as well
> 
> ~I feel like this type of implementation (depending on collecting more context part's implementation) would be under 1000 lines which will be an improvement over my first approach 😄~
> 
> Would love to hear your thoughts about this one and also we can still explore modifying the AST if this still seems crazy. But I am not sure if that is going to be easier than this one.
> 
> To the future reviewers of this PR: May the force be with you. 😄


kayagokalp:
I solved the whitespace changes in the formatted output, and will be editing this message with more details on my progress.

mitchmindtree:
Would you mind merging in or rebasing onto `master` so that the diff no longer includes the `CommentMap` changes? It might make it a little easier to review the new comment insertion work.

kayagokalp:
Currently, I am adding other visitor implementations as well. I am nearly done with functions and once it is finished it will also take care of traits etc. Couple more visitor implementations and hopefully. we will be able to insert comments correctly. After that, I will deal with a more robust way of handling offsets (between comments and items associated with them, so that users' alignments are not altered while inserting the comments.)

kayagokalp:
Left some couple of remarks and information about StorageInitializer formatting at #2339. (Also related to https://github.com/FuelLabs/sway/pull/2311/commits/f2a3ab10b5210979926c3acc3b1789fa3ed727ba)

sway-fmt-v2/src/items/item_const.rs
```diff
@@ -44,3 +45,22 @@ impl Format for ItemConst {
...
     }
 }
+
+impl CommentVisitor for ItemConst {
+    fn collect_spans(&self) -> Vec<CommentSpan> {
+        let mut collected_spans = Vec::new();
+        if let Some(visibility) = &self.visibility {
+            collected_spans.push(CommentSpan::from_span(visibility.span()));
+        }
+        collected_spans.push(CommentSpan::from_span(self.const_token.span()));
```
eureka-cpu, line 55: should we use `push` instead of `write` here?

eureka-cpu: Ah, it's a vector, my mistake!

JoshuaBatty (approved):
Epic work @kayagokalp ! Looks good to me but I'll defer to Austin or Mitch for final approval

sway-fmt-v2/src/utils/comments.rs
```diff
@@ -69,6 +100,308 @@ fn collect_comments_from_token_stream(
...
+    formatted_code.clear();
+    formatted_code.push_str(&src_rope.to_string());
+    Ok(comment_str.len())
+}
+
+/// Applies formatting to the comment.
+/// Currently just checks if it is a multiline comment, if that is the case it adds a trailing `/` to the end.
+fn format_comment(comment: &Comment) -> String {
+    if comment.span().str().starts_with("/*") {
+        format!("{}/", comment.span().str())
```
mitchmindtree, line 394: Why do we add the trailing '/' here? Is this missing from from the original comment span? If so, we should likely fix the span in the `lex_comments` function.

sway-fmt-v2/src/utils/comments.rs
```diff
@@ -12,6 +29,20 @@ pub struct CommentSpan {
     pub end: usize,
 }
 
+impl CommentSpan {
+    /// Takes `start` and `end` from `sway::types::Span` and constructs a `CommentSpan`
+    pub fn from_span(span: Span) -> CommentSpan {
```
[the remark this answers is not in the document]
mitchmindtree: Also, this `from_span` method could be removed in favour of a `From<Span>` implementation.

sway-fmt-v2/src/utils/path.rs
```diff
@@ -132,3 +135,21 @@ impl Format for PathTypeSegment {
         Ok(())
     }
 }
+
+impl CommentVisitor for PathExpr {
+    fn collect_spans(&self) -> Vec<CommentSpan> {
+        // TODO: Should we look for a comment inside the path expression? if so we will need to implement CommentVisitor for PathExpr
+        // For now we are assuming there will be no comments inside the PathExpr so the following comment will be omitted
+        // root::parent/* i am a comment*/::child
+        // I am not sure if this is something the language will allow.
```
mitchmindtree, line 144: Wow I have never seen or would have considered a comment within a path expr :smiling_face_with_tear: Does that even parse? What does Rust do?

I'd be totally fine with removing this comment and keeping the assumption that paths must be contiguous (without space or comments between separators).

sway-fmt-v2/src/utils/path.rs
```diff
@@ -132,3 +135,21 @@ impl Format for PathTypeSegment {
...
+        vec![CommentSpan::from_span(self.span())]
+    }
+}
+
+impl CommentVisitor for PathType {
+    fn collect_spans(&self) -> Vec<CommentSpan> {
+        // TODO: Should we look for a comment inside the path type? if so we will need to implement CommentVisitor for PathType
+        // For now we are assuming there will be no comments inside the PathType so we are visiting it as a single unit.
```
mitchmindtree, line 152: ^ Same as above

sway-fmt-v2/src/utils/comments.rs
```diff
@@ -69,6 +100,308 @@ fn collect_comments_from_token_stream(
...
+    comments_to_insert: Vec<CommentWithContext>,
+    offset: usize,
+    formatted_code: &mut FormattedCode,
+) -> Result<usize, FormatterError> {
+    let mut src_rope = Rope::from_str(formatted_code);
+    // prepare the comment str
+    let mut comment_str = format!(
+        "{}{}",
+        comments_to_insert[0].1,
+        format_comment(&comments_to_insert[0].0)
```
mitchmindtree, line 379: This `insert_after_span` function could be made a little safer if we were to use `if let` to remove the assumption that the `comments_to_insert` is not empty, e.g.

```rust
let mut iter = comments_to_insert.iter();
if let Some(first) = iter.next() {
    let mut comment_str = format!("{}{}", first.1, format_comment(&first.0));
    for comment in iter {
        // ...
    }
    return Ok(comment_str.len());
}
Ok(0)

File: sway-fmt-v2/src/utils/comments.rs [hunk 7]
```diff
@@ -69,6 +100,308 @@ fn collect_comments_from_token_stream(
...
+/// Returns a list of comments between given spans. For each comment returns the Context
+/// Context of a comment is basically the offset (the characters between the last item/comment) to the current comment
+fn get_comments_between_spans(
+    from: &CommentSpan,
+    to: &CommentSpan,
+    comment_map: &CommentMap,
+    unformatted_code: &Arc<str>,
+) -> Vec<CommentWithContext> {
+    let mut comments_with_context = Vec::new();
+    for (index, comment_tuple) in comment_map
```
mitchmindtree, line 339: We should never need to write `comment_tuple`, as `tuple` and tuple indexing (e.g. `.0`, `.1`, etc) is unnecessarily ambiguous for the reader in the case they don't happen to know what the tuple contains. Instead, we could match this pattern further, e.g.

```rust
for (index, (comment_span, comment)) in comment_map
```

I think it's good to keep in mind that while useful in some niche cases, tuple indexing is generally a bit of a code smell, particularly when a more descriptive binding name could be given to the elements we care about. We should always try to prioritize the reader of the code ahead of the writer.

sway-fmt-v2/src/utils/comments.rs
kayagokalp: Yes, it is missing in the original comment span. I didn't bring it up because it had a test without the trailing `/` so I thought it was intentional 😄  I just created #2355 for tracking this issue. Would you prefer that handled separately or should we pass it with this one?

sway-fmt-v2/src/utils/path.rs
kayagokalp: :smile: just tested and it does not parse, so I am deleting this comment and leaving this assumption in-place.

kayagokalp:
Thanks for the nice comments @JoshuaBatty @mitchmindtree :pray: . I applied the changes requested and wanted to comment about the current approach. As @mitchmindtree touched in the review above this comment, although this works this can be done more efficiently. But we need some ninja level refactoring to reach there and this PR is already big and I think we need the formatter v2 to replace the current one as soon as possible. So as @mitchmindtree suggested in the review, I am creating a separate issue for improving the efficiency and handle that later on with a clean look.

sway-fmt-v2/src/utils/comments.rs
```diff
@@ -69,6 +100,308 @@ fn collect_comments_from_token_stream(
...
+        .range((Included(from), Excluded(to)))
+        .enumerate()
```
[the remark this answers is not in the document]
kayagokalp: I ended up simplifying there differently but couldn't utilize `Iterator::peekable` with the current one either, do you think we should still have it with the current implementation as well? If that is the case I can try to utilize it with the current version as well.

sway-fmt-v2/src/utils/comments.rs
```diff
@@ -69,6 +100,308 @@ fn collect_comments_from_token_stream(
...
+fn format_comment(comment: &Comment) -> String {
+    if comment.span().str().starts_with("/*") {
+        format!("{}/", comment.span().str())
+    } else {
+        comment.span().str()
+    }
+}
+/// While searching for a comment we need the possible places a comment can be placed in a structure
+/// `collect_spans` collects all field's spans so that we can check in between them.
+pub trait CommentVisitor {
```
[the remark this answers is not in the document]
mitchmindtree: @kayagokalp looks like you might have missed this one?

sway-fmt-v2/src/utils/comments.rs
mitchmindtree: Agh I just took a look and that `assert!` is definitely wrong! I must have just copy pasted it from the output without thinking :facepalm:

I think we should address it before landing more code that works around it - maybe in a separate independent PR?

sway-fmt-v2/src/utils/comments.rs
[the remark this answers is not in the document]
kayagokalp: Ahh, yeah my bad. This really makes sense, I remember struggling to find a name for this one :smile: Just pushed the name change and merged master which includes your multi-line comment lexer fix :tada:

mitchmindtree (approved):
:ship:

sway-fmt-v2/src/error.rs
```diff
@@ -9,6 +9,8 @@ pub enum FormatterError {
     FormatError(#[from] std::fmt::Error),
     #[error("Error while lexing file: {0}")]
     LexError(#[from] sway_parse::LexError),
+    #[error("Error while adding comments")]
```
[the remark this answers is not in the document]
eureka-cpu: like maybe include some extra pointer to information as the above errors do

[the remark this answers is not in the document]
kayagokalp: Hmm, not so sure if we have some extra information to print here _for now,_ because this error is produced only if there are no spans collected for the given unformatted/formatted code. Which can only happen if the input is empty. But in that case, this error is still not possible because this check is done _while_ traversing the given input. So if there is something that we can traverse and do this check this means there are some `span`s present in the input. So basically I introduced this because I do not like `.unwrap()` stuff around 😄 also I believe this error will need to be re-used in the future when we are refactoring this part of the code for a better efficient approach. Once we have some other cases where we are producing this error, we will have some other information to print just like the above ones.

So this was more like a preferred style thingy over `.unwrap()` (they are interchangeable in this context as this will not fail 😄) but if you prefer having an explicit `anyhow!("..")` kinda approach other than this approach I can also go that way too.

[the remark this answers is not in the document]
eureka-cpu: Okay great! Just wanted to see where your head was at (: good work!

eureka-cpu (approved):

Issue #2310, opened 2022-07-12
Create a `__gtf` intrinsic


mohammadfawaz:
A simple intrinsic that would help with two things:
1. The lack of const support in place of immediates in `asm` blocks: https://github.com/FuelLabs/sway/issues/810
2. Remove the need for `asm` blocks altogether in https://github.com/FuelLabs/sway/pull/2292

PR #2315, opened 2022-07-12
Add a new `__gtf` intrinsic

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/2310

- Added `gtf` to codegen
- Added a new `__gtf` intrinsic that is generic: `__gtf<T>(<index>, <field_id>)` 
  - It's up to user (library author) to decided how to read the result of a `__gtf` (i.e. as `u64`, or as `b256`, etc.)
  - `<index>` can be any integer variable. This is a compile error otherwise.
  - `<field_id>` has to be const-evaluatable integer. This is a compile error otherwise.
- Added a new `gtf` IR instruction that will map exactly to the `gtf` opcode.
  - Need an `int_to_ptr` instruction after `gtf` if the generic type specified is not a copy type.


sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
```diff
@@ -240,6 +240,97 @@ impl TypedIntrinsicFunctionKind {
...
+                    resolve_type(index.return_type, &index.span).unwrap(),
+                    TypeInfo::UnsignedInteger(_)
+                ) {
+                    errors.push(CompileError::IntrinsicUnsupportedArgType {
+                        name: kind.to_string(),
+                        span: index.span.clone(),
+                    });
+                }
+
+                // Make sure that the tx field ID is an unsigned integer
```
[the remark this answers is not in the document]
nfurfaro: So this (being able to pass consts to `__gtf`) is dependent on #810 then?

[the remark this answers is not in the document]
mohammadfawaz: Because we're not using `asm` block directly, the intrinsic will work with anything that is const-evaluatable like `const` declarations. So this will work:
```rust
    let u64_field = __gtf::<u64>(1, SOME_TX_FIELD);
    let b256_field = __gtf::<b256>(2, SOME_OTHER_TX_FIELD);
```
I added a small test to show this. The test compiles and the assembly has the right opcodes, but the test reverts as expected because the `fuel-core` changes are not done yet.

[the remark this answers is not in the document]
nfurfaro: A few things: 
- I think I'll close my `gtf` opcode PR in favor of this one because my test doesn't even compile yet.
>intrinsic will work with anything that is const-evaluatable like const declarations. 

great!

>it is possible that the gtf instruction gives us back a pointer to a b256

Yeah, I had been using `read` for the fields that return a pointer to a `b256` so this makes sense.

test/src/e2e_vm_tests/test_programs/should_pass/language/gtf_intrinsic/src/main.sw
```diff
@@ -0,0 +1,11 @@
+script;
+
+const SOME_TX_FIELD = 0x42;
+const SOME_OTHER_TX_FIELD = 0x77;
+
+fn main() -> u64 {
+    // Test expected to compile but revert because `fuell-core` does not support `gtf` yet.
+    let u64_field = __gtf::<u64>(1, SOME_TX_FIELD);
+    let b256_field = __gtf::<b256>(2, SOME_OTHER_TX_FIELD);
```
vaivaswatha, line 9: I don't know what the `__gtf` op is supposed to do, but if the values of `u64_field` and `b256_field` are deterministic, it might be useful to add an assert to validate the values.

test/src/e2e_vm_tests/test_programs/should_pass/language/gtf_intrinsic/src/main.sw
```diff
@@ -0,0 +1,11 @@
+script;
+
+const SOME_TX_FIELD = 0x42;
+const SOME_OTHER_TX_FIELD = 0x77;
+
+fn main() -> u64 {
+    // Test expected to compile but revert because `fuell-core` does not support `gtf` yet.
```
vaivaswatha, line 7:
```suggestion
    // Test expected to compile but revert because `fuel-core` does not support `gtf` yet.
```

vaivaswatha (dismissed):
Looks fine to me.

sway-core/src/asm_generation/from_ir.rs
```diff
@@ -1380,6 +1400,11 @@ impl<'ir> AsmBuilder<'ir> {
         self.reg_map.insert(*instr_val, base_reg);
     }
 
+    fn compile_int_to_ptr(&mut self, instr_val: &Value, int_to_ptr_val: &Value) {
+        let val_reg = self.value_to_register(int_to_ptr_val);
+        self.reg_map.insert(*instr_val, val_reg);
+    }
```
otrho, line 1406: Yep, this will be a conflict with `master` now.

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
```diff
@@ -244,6 +244,97 @@ impl TypedIntrinsicFunctionKind {
...
+                    TypedExpression::type_check(ctx.by_ref(), arguments[1].clone()),
+                    return err(warnings, errors),
+                    warnings,
+                    errors
+                );
+
+                // Make sure that the index argument is an unsigned integer
+                if !matches!(
+                    resolve_type(index.return_type, &index.span).unwrap(),
+                    TypeInfo::UnsignedInteger(_)
```
otrho, line 291: I think this should be a fixed size int, probably `u64`.  Same for the field ID although that should probably be `u16` since it's a 12 bit val in the ASM.  Or maybe just make them both `u64` for simplicity.

Having them as wishy-washy any-size-integer makes documentation hard -- ideally we want to say this concretely is `__gtf<T>(index: u64, field_id: u64) -> T`.

Is that right?  `index` first, which is `RB` in the `GTF` instruction?  Should the index be `Option<u64>` to be a little more correct?  And while we're at it we could make an enum for all the field types -- or do we expect the stdlib to do that instead?

otrho: I honestly would rather see these functions be type checked as regular functions.  We would add their signatures to the type checker before we start and then just assume they're regular functions until we get back to codegen.

Since we're enforcing the `__name` only for intrinsics they can be easily identified at that point and get special case codegen.  This might be worth looking at as a part of the global declaration refactor.

sway-ir/src/parser.rs
```diff
@@ -225,6 +227,11 @@ mod ir_builder {
                     IrAstOperation::GetPtr(name, ty, offset)
                 }
 
+            rule op_gtf() -> IrAstOperation
+                = "gtf" _ index:id() comma() tx_field_id:(decimal())  {
```
otrho, line 231:
```suggestion
                = "gtf" _ index:id() comma() tx_field_id:decimal()  {
```

sway-ir/tests/inline/int_to_ptr.ir
```diff
@@ -0,0 +1,25 @@
...
+
+// check: fn main
+    fn main() -> b256 {
+        entry:
+
+// check: $(arg0=$VAR) = const u64 11
+        v0 = const u64 11
+// check: $(arg1=$VAR) = int_to_ptr $arg0 to b256
+// not: call
+        v1 = call a(v0)
```
otrho, line 19: I know this is from the other PR but looking at it here I think these tests can be reorganised a bit.  The `inline` tests should really only be testing the inlining pass (and regressions), and we don't need to add a new test for every new opcode that we add.

The Sway `ir_generation` tests can verify that the right instruction is generated from the Sway source (e.g., an `int_to_ptr` pops out and not a `call`), and those tests also do the print -> parse -> print -> compare.  But we should be adding _something_ here for new opcodes.

I just took a peek at the `sway-ir/tests` dir and there _is_ already a `serialize` directory there..!  I forgot.  So that's where we should put tests for coverage in general.  It might need a bit of TLC, I'm not sure it's parsing it's own output like the `ir_generation` tests are.

sway-core/src/asm_generation/from_ir.rs
mohammadfawaz: All good now :)

test/src/e2e_vm_tests/test_programs/should_pass/language/gtf_intrinsic/src/main.sw
mohammadfawaz: Sadly, we can't do this just yet because `fuel-core` is not ready yet. So, the test will always revert at the `gtf` instruction before anything gets assigned.

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
otrho: I asked @vaivaswatha for his opinion here and he pointed out that while this makes sense, it's not really possible while we still have AST function call inlining.  Definitely worth revisiting after declaration engine, but not yet.

sway-ir/tests/inline/int_to_ptr.ir
mohammadfawaz: Yeah the test would have to change so I created a separate issue here: https://github.com/FuelLabs/sway/issues/2348

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
mohammadfawaz: Ah interesting. Alright, so kept it as is for now but change it to only expect `u64`s.

> Should the index be Option<u64> to be a little more correct?

Hmm potentially, though we'd have to do some magic here to make sure to expect a Sway `Option`.

> And while we're at it we could make an enum for all the field types

I was hoping not to bake this into the compiler as these can change. But also yeah the standard library will handle most of this.

Really the idea with this intrinsic is to keep it low level as it's only to be used by library authors. It's mean to replace an `asm` block effectively and allow the use of const-evaluatable tx field IDs (second argument) which `asm` don't allow yet.

vaivaswatha (approved):

PR #2317, opened 2022-07-12
Refactor coins_owner

### LINKED ISSUE #2148: Implement bitwise negation for `b256`
This will use the `not` opcode.

### LINKED ISSUE #2235: Improve msg_sender() to look at Message Inputs.
As both pubkey hashes && predicate roots are valid Addresses, we should really be looking at Message Inputs as well when trying to determine an owner (not just Coin Inputs).

nfurfaro:
This addresses most of the concerns in the 2 different linked issues. It:
- removes manual invalidation of the loop condition.  I decided `break` isn't needed here as we already have an explicit `Return` statement in that arm of the conditional statement.
- imports and uses 2 consts from `std::tx`, and removes the `*` import
- adds the ability to get the owner from `InputMessage`s as well as coins, extending the usefulness of `msg_sender()`
- renames the function to `inputs_owner()` to encompass the new functionality

I didn't end up refactoring if/else statements to use `match`, as it wasn't a good fit here. We can't match on const values or dynamic values, so my attempts to do this all ended at unreachable `match` arm errors.

closes #2148 
closes #2235


sway-lib-std/src/chain/auth.sw
```diff
@@ -8,7 +8,7 @@ use ::contract_id::ContractId;
 use ::identity::Identity;
 use ::option::*;
```
mohammadfawaz, line 9: Unrelated, but do you want to change this to `use ::option::Option;`? 😄

mohammadfawaz (dismissed):

otrho (dismissed):

Braqzen (approved):

Braqzen (approved):

mohammadfawaz (approved):

PR #2328, opened 2022-07-14
Switch the order of the fields for `Option` from `Some` followed by `None` to `None` followed by `Some`

### LINKED ISSUE #2326: Switch the order of the fields for `Option` from `Some` followed by `None` to `None` followed by `Some`
## Summary

When using `Option` my assumption is that the default value is `None` until it is set to something. 
This may or may not be a correct assumption nevertheless this becomes an issue when initializing a `StorageMap` with an `Option<CustomType>`.
The reason this is an issue is that the default value for this `Option` is now `CustomType` instead of `None` which means that checking for `var.is_some()` is now insufficient to ensure that the variable is / is not set. Now `is_some()` must be used and the fields of `CustomType` must be validated to not be some 0th value but that's also an issue because those fields might all have valid fields when the value of 0 is used. This means that there is no way to distinguish whether `CustomType` has been set or if it has a default value and it's unreasonable to assume that the contract will somehow either set every value in a mapping to `None` upon initialization or that additional variables are juggled to ensure that each `CustomType` is actually valid.

The simple bandaid is to switch the order of the fields in `Option` so that the default value is `None` and now a variable must be explicitly set in order to pass through the `var.is_some()` check.

P.S. This was only discovered in some extensive testing and I did not realize this until I hit a test that was disfunctional. This could have created a vulnerability in the contract and adding documentation with the hope that each dev is aware of this current flaw is unacceptable UX. Moreover, this might slip past testing if it's done incorrectly.

Braqzen:
## Changes made

- Changed Option field order from (1) to (2)

(1)

```rust
pub enum Option<T> {
    /// Contains the value
    Some: T,

    /// Signifies the absence of a value
    None: (),
}
```

(2)

```rust
pub enum Option<T> {
    /// Signifies the absence of a value
    None: (),
    
    /// Contains the value
    Some: T,
}
```

## Related

Closes #2326

### DIFF
--- a/sway-lib-std/src/option.sw
+++ b/sway-lib-std/src/option.sw
@@ -11,11 +11,11 @@ use ::revert::revert;
 /// `Option` is a type that represents either the existence of a value ([`Some`]) or a value's absence
 /// ([`None`]).
 pub enum Option<T> {
-    /// Contains the value
-    Some: T,
-
     /// Signifies the absence of a value
     None: (),
+    
+    /// Contains the value
+    Some: T,
 }
 
 /////////////////////////////////////////////////////////////////////////////


bitzoic (approved):

otrho:
@Braqzen I would really have liked to see a new test added which tests/enforces this behaviour.  This is assuming that the tag values always start at zero for the first variant.  In the future we may change the codegen to do it differently (unlikely but definitely possible) so a specific test that would break in that case would be valuable.

PR #2330, opened 2022-07-14
Simplify function IR generation

tritao:
So while working on https://github.com/FuelLabs/sway/issues/1188 I've been making some IR gen changes which led me to try some simplifications to the code.

All the tests still seem to work fine, but maybe there are good reasons for the current way the generation code is working that I am not aware of.

Would you take a look and let me know if maybe I am missing something this might break?


mohammadfawaz:
@otrho should know! The existing code is a bit hacky and would be nice to remove it 😃

sway-core/src/ir_generation/function.rs
```diff
@@ -734,48 +728,8 @@ impl FnCompiler {
...
-                span: callee_span,
-                return_type,
-                type_parameters: Vec::new(),
-                return_type_span: crate::span::Span::new(" ".into(), 0, 0, None).unwrap(),
-                visibility: Visibility::Private,
-                is_contract_call: false,
-                purity: callee_purity,
-            };
+            let mut callee_fn_decl = callee;
+            callee_fn_decl.type_parameters.clear();
```
otrho, line 732: I am extremely surprised this isn't breaking any tests.  I'm almost positive that until we have the declaration engine this function recreation is necessary to avoid symbols clashes and/or mismatches.  i.e., a function called `foo` in two different libraries are both going to just be inserted as a local function `foo` in this module and when inlining later on it will probably just choose the first one it finds.

At least that's what I thought was going on but maybe I'm wrong.  I'm going to do a couple of tests on this branch to confirm before I approve...

tritao: > I am extremely surprised this isn't breaking any tests. I'm almost positive that until we have the declaration engine this function recreation is necessary to avoid symbols clashes and/or mismatches. i.e., a function called `foo` in two different libraries are both going to just be inserted as a local function `foo` in this module and when inlining later on it will probably just choose the first one it finds.
> 
> At least that's what I thought was going on but maybe I'm wrong. I'm going to do a couple of tests on this branch to confirm before I approve...

Makes sense, if that is indeed an issue, perhaps we can rename functions before generation to add a unique suffix.

otrho: So, the problem is a problem and not a problem.   Here's an example:

```rust
script;

struct A {
    a: u64,
}

impl A {
    fn foo(self, val: u64) -> u64 {
        asm(a: val, res) {
            add res a a;
            res: u64
        }
    }
}

struct M {
    m: u64,
}

impl M {
    fn foo(self, val: u64) -> u64 {
        asm(a: val, res) {
            mul res a a;
            res: u64
        }
    }
}

fn main() -> u64 {
    let a = (A { a: 0 }).foo(10);        // 10 + 10 = 20
    let m = (M { m: 0 }).foo(4);         // 4 * 4 = 16
    asm (l: a, r: m, res) {
        sub res l r;                    // 20 - 16 = 4
        res: u64
    }
}
```

I used this to try and repro the problem.  Two methods which have the exact same signatures but are obviously different functions.  This is the output from compilation _before_ inlining using your new branch:

```
script {
    fn main() -> u64, !1 {                       
        local ptr u64 a                 
        local ptr u64 m                                                                            
                                                 
        entry:                                   
        v0 = const { u64 } { u64 undef }, !2                                                       
        v1 = const u64 0, !3                     
        v2 = insert_value v0, { u64 }, v1, 0, !2                                                   
        v3 = const u64 10, !4                                                                      
        v4 = call foo(v2, v3), !5
        v5 = get_ptr ptr u64 a, ptr u64, 0, !6
        store v4, ptr v5, !6                     
        v6 = const { u64 } { u64 undef }, !7     
        v7 = const u64 0, !8                                                                       
        v8 = insert_value v6, { u64 }, v7, 0, !7
        v9 = const u64 4, !9                   
        v10 = call foo(v8, v9), !10
        v11 = get_ptr ptr u64 m, ptr u64, 0, !11
        store v10, ptr v11, !11           
        v12 = get_ptr ptr u64 a, ptr u64, 0, !12
        v13 = load ptr v12, !12                                                                    
        v14 = get_ptr ptr u64 m, ptr u64, 0, !13                                                                                                                                                      
        v15 = load ptr v14, !13                                                                    
        v16 = asm(l: v13, r: v15, res) -> u64 res, !14 {
            sub    res l r, !15                  
        }                               
        ret u64 v16                                                                                
    }               
                                                                                                   
    fn foo(self !16: { u64 }, val !17: u64) -> u64, !18 {                                          
        entry:                                   
        v0 = asm(a: val, res) -> u64 res, !19 {                                                    
            add    res a a, !20                                                                    
        }            
        ret u64 v0   
    }                                                                                                                                                                                                 
                                                                                                                                                                                                      
    fn foo(self !21: { u64 }, val !22: u64) -> u64, !23 {                                          
        entry:                                                                                     
        v0 = asm(a: val, res) -> u64 res, !24 {
            mul    res a a, !25                                                                                                                                                                       
        }                                        
        ret u64 v0                                                                                                                                                                                    
    }                              
}
```
Notice the two functions just at the global scope called `foo`.  In `master` the `foo`s would be `anon_0` and `anon_1`.  In the IR itself, as in not this serialised version, it _will_ inline correctly because the `call` instructions refer to the correct function in its args.

But if we were to deserialise this IR it will do as I expected and inline the first `foo` in both calls (the `add` one).

So, while it's not exactly an issue with Sway -> IR -> ASM now, it's definitely broken.  And it frustrates me, but once we get the declaration engine going all symbols will have proper scope/callpaths and we'll be able to add proper calling and choose to inline discriminately, and it will be all rainbows and sunshine, and I'll be very happy.

So, I'd much rather we just leave the function calling code alone as it is, if you don't mind.

tritao: Thanks for looking into this and coming up with this test case, I agree its definitely an issue as is in this PR and not something we should merge. I will try to fix it.

Bwt the reason I need to rework this atm is because we lose some mutability information when creating these fake parameters, and making sure the fake parameters have all the correct information is more trouble than is worth it.

mohammadfawaz: So one thing regarding mutability is that it is only relevant at type checking time. I don't think the IR should or need to care about that. All mutability checking should be done in type checking, so I wouldn't worry to much about that here.

otrho: I don't quite follow.  All arguments are pointer types aren't they?  And I think they're all mutable by default, we don't really do  const pointers.

tritao: Maybe I misunderstood the IR semantics?

Let's take this code (which is not allowed today, but I which I am trying to add support for):

```rust
fn mut_prim(mut b: u32) {
    b = 20;
}

fn main()  {
    let mut b = 0u32;
    mut_prim(b);
}
```

Right now we generate (inlining disabled):

```rust
fn main() -> u64, !1 {
    local mut ptr u64 b
    ...
    v2 = get_ptr mut ptr u64 b, ptr u64, 0, !4
    v3 = load ptr v2, !4
    v4 = call mut_prim_0(v3), !5
}

fn mut_prim_0(b !8: u64) -> (), !9 {
    entry:
    v0 = const u64 20, !5
    store v0, ptr b, !6
    v1 = const unit ()
    ret () v1
}
```

Given `v3 = load ptr v2, !4`, and non ptr type in `mut_prim_0`, I assumed `v3` is a non-reference type.
Did I get this wrong?

With my changes we now generate:

```rust
fn main() -> u64, !1 {
    ...
    v2 = get_ptr mut ptr u64 b, ptr u64, 0, !4
    v3 = call mut_prim_0(v2), !5
}

fn mut_prim_0(b !7: ptr u64) -> (), !8 {
    ...
}   
```

Does this make sense or am I in the wrong path here?

sway-ir/src/context.rs
```diff
@@ -55,7 +55,12 @@ impl Context {
     ///
     /// The name will be in the form `"anon_N"`, where `N` is an incrementing decimal.
     pub fn get_unique_name(&mut self) -> String {
-        let sym = format!("anon_{}", self.next_unique_sym_tag);
+        format!("anon_{}", self.get_unique_id())
+    }
+
+    /// Get a globally unique symbol id.
+    pub fn get_unique_id(&mut self) -> u64 {
```
otrho, line 62: Is this now used somewhere else?  If not, there's no real need to split a function up because we might want it one day, and to make it `pub`.  But I'm looking at only the most recent changes, so maybe you're using `get_unique_id()` elsewhere...

otrho (approved):
OK, thanks Joao.  Overall this simplifies a lot.

It also reminds me of how the IR needs a tonne of work -- the library we have now is pretty much the version 0.0.1 which was really the bare MVP needed to switch to using a middle end.

There's a bunch of stuff which needs refinement or all out replacement, especially regarding pointers vs immediates, managing loads and stores and the stack, mutability and even function calls.  So, yay. :slightly_smiling_face:

sway-core/src/ir_generation/function.rs
otrho: Yeah, OK.  I think the typing regarding pointers/references is just not that well defined and needs to be nailed down.

mohammadfawaz (approved):

PR #2264, opened 2022-07-07
Remove Rust integration testing behaviour from `forc test` in anticipation of unit testing support

mitchmindtree:
This PR is a WIP that aims to address the first 3 steps in #1833.

This is in anticipation of using `forc test` to support unit testing (rather than Rust integration testing) #1832.

The `forc test` command remains for now, but outputs a message explaining that the command is now reserved for unit testing and links to the issues above.

## TODO

- [x] Create a new `sway-test-rs` repo or similar that can be used as a `cargo generate` template.
- [x] Update Rust integration testing docs in the Sway book to describe how to use the `cargo generate` command to easily add Sway integration testing to an existing Rust project.

## Follow-up

- Create a `forc-test-rs` crate that re-exports and extends `fuels` with useful `forc` functionality for integration testing (e.g. re-exporting `forc_pkg::build` to ensure sway code is built and available under `out/` at the start of testing).

PR #2332, opened 2022-07-15
Switch `forc new` to create a test that uses a wallet with some coins in it

mohammadfawaz:


digorithm (approved):
Nice!

mitchmindtree (approved):
This will be moved out of `forc` into a `cargo generate` template as a part of https://github.com/FuelLabs/sway/pull/2264, but happy to see it land in the meantime.

PR #2334, opened 2022-07-15
Add a test to verify the order of `Option` variants

### LINKED ISSUE #2333: Add a test to verify the order of `Option` variants
Ref: https://github.com/FuelLabs/sway/pull/2328#issuecomment-1185070560

We should add a test to verify that the variant `None` is stored as zeros (in memory or storage)

Braqzen:
Closes #2333


test/src/sdk-harness/test_projects/option_field_order/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+
+abigen!(
+    MyContract,
+    "test_projects/option_field_order/out/debug/option_field_order-abi.json"
+);
+
+#[tokio::test]
+async fn default_is_none() {
+    let instance = setup().await;
+    assert_eq!(true, instance.is_none().call().await.unwrap().value);
```
otrho, line 11: `assert_eq!(true, ...)` is a bit redundant, could just be `assert!(...)`.

otrho (approved):

nfurfaro (approved):

mohammadfawaz (approved):

Issue #2324, opened 2022-07-14
Create an `int_to_ptr` IR instruction


mohammadfawaz:
Roughly similar to https://llvm.org/docs/LangRef.html#inttoptr-to-instruction. The goal is to be able to interpret a `u64` value in IR as a pointer. That allows us to do the equivalent of the following cast:
```rust
asm(x: u64) {
    x: b256
}
```
in IR.

otrho:
Off the top of my head, wouldn't it be pretty much like `get_ptr`? It's an instruction which still returns a `Value`, but it's a pointer.  I guess there's other places where we assume a pointer value is a `get_ptr` but now it could be an `int_to_ptr` too?

PR #2335, opened 2022-07-15
Add `int_to_ptr` IR instruction

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/2324

Simple instruction that allows us to interpret an integer as a pointer to a reference type. Exhibits a similar behaviour to:
```rust
asm(x: 0) {
    x: T
}
```


sway-ir/src/instruction.rs
```diff
@@ -163,6 +165,7 @@ impl Instruction {
             Instruction::GetStorageKey => Some(Type::B256),
             Instruction::InsertElement { array, .. } => array.get_type(context),
             Instruction::InsertValue { aggregate, .. } => aggregate.get_type(context),
+            Instruction::IntToPtr(_, ty) => Some(*ty),
```
otrho, line 168: If I was going to be pedantic I might ask for this to be moved below next to `GetPointer` since they're semantically similar.  But I'm not. ;)

sway-ir/src/verify.rs
```diff
@@ -468,6 +469,26 @@ impl<'a> InstructionVerifier<'a> {
         }
     }
 
+    fn verify_int_to_ptr(&self, value: &Value, ty: &Type) -> Result<(), IrError> {
+        // We want the source vaule to be an integer and the destination type to be a reference
```
otrho, line 473:
```suggestion
        // We want the source value to be an integer and the destination type to be a reference
```

sway-ir/src/verify.rs
```diff
@@ -468,6 +469,26 @@ impl<'a> InstructionVerifier<'a> {
...
+        // type.
+        let val_ty = value
+            .get_type(self.context)
+            .ok_or(IrError::VerifyIntToPtrUnknownSourceType)?;
+        if !matches!(val_ty, Type::Uint(64)) {
+            return Err(IrError::VerifyIntToPtrFromNonIntegerType(
+                val_ty.as_string(self.context),
+            ));
+        }
+        if ty.is_copy_type() {
```
otrho, line 483: Interesting.  We never want to convert a `u64` to `ptr u64`?  I guess not yet, until we really have proper reference support.

otrho (approved):

sway-ir/src/instruction.rs
mohammadfawaz: Oh I thought I was following alphabetical order here but I just noticed that this list is grouped into sublists :)

mohammadfawaz: Moved it to right after `GetPointer` :)

sway-ir/src/verify.rs
mohammadfawaz: Yeah I didn't want think too hard about this just yet 😄

otrho (approved):

mitchmindtree (approved):

otrho:
Merging as @mohammadfawaz is probably busy with Paris and I have an upcoming PR depending on this.

PR #2338, opened 2022-07-15
Add `expr` formatting to `sway-fmt-v2`

### LINKED ISSUE #992: Formatting contract call arguments
The formatter turns contract calls into something that looks like this:
```rust
caller.method {
    gas: 10000
}
();
```
I think it's better to have the `()` next to the `}`, so:
```rust
caller.method {
    gas: 10000
} ();
```

### LINKED ISSUE #2300: Add `expr` formatting to `sway-fmt-v2`
Currently we only have the `expr` spans, I've gone ahead and added the setup for the logic we just need to fill it out.

### LINKED ISSUE #2467: Only allow single line formatting for structure literals in `sfv2`
Currently the line width of an item to use the `Inline` formatting (single line) is only `18usize`, and this almost never is a realistic number for inline formatting of small structures. Most of our tests are about `200-220usize` to allow formatting of structs or enums to single line. Thoughts? @kayagokalp @mitchmindtree @sezna @mohammadfawaz @emilyaherbert 

See below comments for resolved context

eureka-cpu:
Closes #2300 
Closes #992 
Closes #2467 

Currently not all expressions have vertical handling, but they do have horizontal. There could also be _more_ tests, but it may be a good idea to start getting reviews now since this has grown beyond the scope of what it was originally. This also fixes an issue of accidentally adding single line formatting for non-literal structures. There are also some changes to `Shape` that made multiline formatting easier. I've gone ahead and removed/added features that we weren't/are using.

- [x] abi
- [x] asm
- [x] assignable
- [x] code_block
- [x] collections
- [x] conditionals
- [x] structs
- [x] expr
- [x] #2467 
- [ ] vertical formatting for each expression
  - [x] structs/method calls
  - [x] tuples
- [ ] tests
  - [x] method calls
  - [x] structs
  - [x] tuples
  - [x] function applications
  - [x] literals
  - [x] paths
  - [x] field projection
  - [x] abi cast


kayagokalp (commented):
Hey @eureka-cpu, I am taking a quick look and wanted to have a little comment. First of all, this seems really great 🔥, for the tests of comments at #2311, I couldn't add comments with `expr` test because we didn't have `expr` formatting back then and the way comment stuff works were duplicating comments next to `expr`s.  With this PR, we should be able to add `expr` comments as well, so it might be a good idea to add a `test_expr_comments` with this introduction of `expr` formatting. 

I don't know if were planning to add such a test but wanted to give a heads-up on that 👍

eureka-cpu:
> Hey @eureka-cpu, I am taking a quick look and wanted to have a little comment. First of all, this seems really great 🔥, for the tests of comments at #2311, I couldn't add comments with `expr` test because we didn't have `expr` formatting back then and the way comment stuff works were duplicating comments next to `expr`s. With this PR, we should be able to add `expr` comments as well, so it might be a good idea to add a `test_expr_comments` with this introduction of `expr` formatting.
> 
> I don't know if were planning to add such a test but wanted to give a heads-up on that 👍

I have some better tests written and I'll update asap, there's just quite a lot of things to do!

sway-fmt-v2/src/utils/expr/mod.rs
```diff
@@ -0,0 +1,953 @@
...
+            Self::If(if_expr) => if_expr.format(formatted_code, formatter)?,
+            Self::Match {
+                match_token,
+                value,
+                branches,
+            } => {
+                write!(formatted_code, "{} ", match_token.span().as_str())?;
+                value.format(formatted_code, formatter)?;
+                MatchBranch::open_curly_brace(formatted_code, formatter)?;
+                let branches = branches.clone().into_inner();
```
mitchmindtree, line 148: I've noticed there are quite a lot of places where we `clone()` in order to call `into_inner()`, even though we only need the inner type by reference.

In this case, the `Braces` type has [a `get` method](https://docs.rs/sway-ast/0.19.2/sway_ast/brackets/struct.Braces.html#method.get) which allows us to use the existing instance directly rather than cloning the whole structure.

It might be worth checking all our uses of `.clone().into_inner()` for similar cases to this as we can likely avoid quite a bit of unnecessary allocation.

sway-fmt-v2/src/utils/expr/mod.rs
```diff
@@ -0,0 +1,953 @@
...
+}
+
+impl LeafSpans for Expr {
+    fn leaf_spans(&self) -> Vec<ByteSpan> {
+        visit_expr(self)
+    }
+}
+
+/// Collects various expr field's ByteSpans.
+fn visit_expr(expr: &Expr) -> Vec<ByteSpan> {
```
mitchmindtree, line 645: This function might be better named `expr_leaf_spans` as we are just matching on the expr to collect leaf spans, not really doing visitor traversal or anything like that.

sway-fmt-v2/src/utils/expr/tests.rs
```diff
@@ -0,0 +1,117 @@
...
+                assert_eq!(&formatted_code, $desired_output);
+            }
+        }
+    )+
+}
+}
+
+// TODO:
+// path, literal, abicast, struct, tuple, parens, block, array, asm, return, if, match, while, func app, index, field projection, tuple field projection, ref, deref, operators, bitwise stuff, comp operators
+// logical operators, reassignment
```
mitchmindtree, line 83: Could we open an issue for these and link it here?

sway-fmt-v2/src/utils/punctuated.rs
```diff
@@ -4,9 +4,13 @@ use crate::{
     FormatterError,
 };
 use std::fmt::Write;
-use sway_ast::{keywords::CommaToken, punctuated::Punctuated, StorageField, TypeField};
+use sway_ast::{
+    keywords::CommaToken, punctuated::Punctuated, token::PunctKind, StorageField, TypeField,
+};
 use sway_types::{Ident, Spanned};
```
mitchmindtree, line 11:
```suggestion
```

In general I think it's good to keep `use` statements together as it allows rustfmt to automatically order them, but not that important.

mitchmindtree (approved):
Looking super clean and readable @eureka-cpu, nice work!

I've left a few nits that would be good to address, but no showstoppers. I'll leave it up to you whether we want to address them in this PR or if we want to handle them in smaller follow-up PRs.

---

### `Shape` changes

I have a slight hesitation about exposing the formatter's `Shape` state as done in this PR as I can imagine it might lead to cases where we could easily forget to reset the line style or width heuristics at the end of an item's formatting, leading to following items being mysteriously incorrect only on occasion.

We could potentially avoid this with methods on `Formatter` that scope these temporary shaping changes, e.g.

```rust
formatter.with_line_style(new_style, |formatter| {
    // Do some formatting with `new_style`.
});
// Automatically revert to original style after scope exited.
```

With an approach along these lines, a `Format` implementation could never accidentally leave the `Formatter`'s inner state in some unexpected state.

All that said, this might require some more significant redesign, and we can investigate diving into this in future issues/PRs. Your existing approach appears to be working so happy to land it!

kayagokalp (commented):
Looks great 🚀. Left couple of nits regarding some possible rebasing issue, and expanded @mitchmindtree's great review around the usage of `while let`. 

If you would like to merge this and address changes in a separate PR, just ping me and I can happily approve as they are not blocking issues. I am not approving rn just to give you a chance to look to the reviews 🙌

sway-fmt-v2/src/fmt.rs
```diff
@@ -475,6 +474,88 @@ trait CompSciStudent: Programmer + Student {
...
+    }
+    (3, fuelcoin_id, balance_test_id);
+
+    fuelcoin_balance = balance_of(fuelcoin_id, fuelcoin_id);
+    let balance_test_contract_balance = balance_of(fuelcoin_id, balance_test_id);
+    assert(fuelcoin_balance == 1);
+    assert(balance_test_contract_balance == 3);
+
+    true
+}"#;
```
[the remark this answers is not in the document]
eureka-cpu: This test is actually specifically what @mohammadfawaz gave me for formatting method calls, hence why it was used. But I can see that smaller tests would be better, since we may only end up breaking one specific thing like inline method calls.

sway-fmt-v2/src/items/item_enum.rs
```diff
@@ -23,182 +18,93 @@ impl Format for ItemEnum {
...
-                        )?;
-                        // TypeField
-                        type_field.format(formatted_code, formatter)?;
-
-                        if value_pairs_iter.peek().is_some() {
-                            writeln!(formatted_code, "{}", comma_token.span().as_str())?;
+                });
+
+                let mut value_pairs_iter = value_pairs.iter().enumerate().peekable();
+                for (var_index, (type_field, comma_token)) in value_pairs_iter.clone() {
```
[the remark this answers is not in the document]
eureka-cpu: I was curious of that, I believe this has been there for some time

sway-fmt-v2/src/utils/expr/tests.rs
eureka-cpu: sure! I have taken care of a few of them. I believe we would need all of them as tests before this PR can be approved.

eureka-cpu: *at least the horizontal ones

kayagokalp (approved):
🚀

eureka-cpu:
> Looking super clean and readable @eureka-cpu, nice work!
> 
> I've left a few nits that would be good to address, but no showstoppers. I'll leave it up to you whether we want to address them in this PR or if we want to handle them in smaller follow-up PRs.
> ### `Shape` changes
> 
> I have a slight hesitation about exposing the formatter's `Shape` state as done in this PR as I can imagine it might lead to cases where a `Format` implementation could easily forget to reset the line style or width heuristics at the end of an item's formatting, leading to following items being mysteriously incorrect only on occasion.
> 
> We could potentially avoid this with methods on `Formatter` that scope these temporary shaping changes, e.g.
> 
> ```rust
> formatter.with_line_style(new_style, |formatter| {
>     // Do some formatting with `new_style`.
> });
> // Automatically revert to original style after scope exited.
> ```
> 
> With an API along these lines, a `Format` implementation could never accidentally leave the `Formatter`'s inner state in some unexpected state.
> 
> All that said, this might require some more significant redesign, and we can investigate diving into this in future issues/PRs. Your existing approach appears to be working so happy to land it!

I'm curious if this way will solve the problem of having nested formats, for instance in a `let` statement where the inner value is an `expr` that has another `expr` within it. We would need a way to continually update the state of the shape to fit its formatting needs. I took this approach with this in mind, but if the way you mentioned also takes this into account I'd rather that!

eureka-cpu:
I will create some issues around these comments and address them in separate PRs, thanks so much!

eureka-cpu:
> I will create some issues around these comments and address them in separate PRs, thanks so much!

#2493 #2494 #2495 #2496 #2497 #2498 #2499 #2500 #2501

PR #2340, opened 2022-07-17
Signed integers

### LINKED ISSUE #2422: Signed integers in std lib
The implementation of signed integers using corresponding unsigned integers as underlying values. Actual value is underlying value minus 2 ^ (number of bits - 1), in other words the values are indented by 2 ^ (number of bits - 1).

As for addition and subtraction we  need to keep track of this indentation, so that we don't indent twice.

For multiplication and division we need to keep track of the sign, to correctly take the indentation as well (i.e. either subtract the indent, or subtract from indent).

rostyslavtyshko:
i8, i16, i32, i64

Basic arithmetic operations and tests

Closes #2422


otrho:
Does this have a related issue?

sway-lib-std/src/i16.sw
```diff
@@ -0,0 +1,140 @@
...
+
+pub trait From {
+    /// Function for creating i16 from its u16 and bool components.
+    fn from(value: u16) -> Self;
+}
+
+impl From for i16 {
+    fn from(value: u16) -> i16 {
+        i16 {
+            underlying: value,
```
sezna, line 20: Awesome implementation! A couple of notes here -- since we follow Rust generally when it comes to these kinds of things,  we should probably follow Rust's conventions on integer width conversions: https://doc.rust-lang.org/std/primitive.i32.html#impl-From%3Ci16%3E

Note how `i32` can only be constructed from `u16`, `u8`, `i16`, or `i8` because they have no risk of overflow. 

Additionally, check out this example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=849026d21ceb5f4632f8b179665f1e7d

I believe with your `from()` here, `i16::from(5u32)` would be `i16::min() + 5`, which is not consistent.

sezna:
BTW, is there an issue for this? I didn't find one -- could be good to make one for reference so we can document motivating factors.

rostyslavtyshko:
> BTW, is there an issue for this? I didn't find one -- could be good to make one for reference so we can document motivating factors.

Added one

sway-lib-std/src/i16.sw
rostyslavtyshko: The From trait in Sway is not the same, as in Rust. Basically we don't have one From trait (e.g. the for U128 differs significantly). From normally is for direct conctructing and should be used with caution.
About overflows: Sway seems to have an undesirable behavior to ignore them. E.g. you could use u32 instead of u16, and you wouldn't get any panics. I'd like it to be different, but I guess that's out of the scope of this PR

sway-lib-std/src/i8.sw
```diff
@@ -0,0 +1,141 @@
...
+use ::assert::assert;
+
+/// The 8-bit signed integer type.
+/// Represented as an underlying u8 value.
+pub struct i8 {
+    underlying: u8,
+}
+
+pub trait From {
+    /// Function for creating i8 from its u8 and bool components.
```
adlerjohn, line 13: Wait, there's no bool component (presumably there was one in the past)!

sway-lib-std/src/i8.sw
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /// The largest value that can be represented by this type,
+    pub fn max() -> i8 {
+        i8 {
+            underlying: ~u8::max(),
+        }
+    }
+
+    /// The size of this type in bits.
```
adlerjohn, line 69: Similar to the `U256` docs, can you include the actual max, min, and number of bits in the docstrings?

sway-lib-std/src/i8.sw
```diff
@@ -0,0 +1,141 @@
...
+        i8 {
+            underlying
+        }
+    }
+}
+
+impl core::ops::Add for i8 {
+    /// Add a i8 to a i8. Panics on overflow.
+    pub fn add(self, other: Self) -> Self {
+        ~i8::from(self.underlying - ~i8::indent() + other.underlying) // subtract 1 << 7 to avoid double move
```
adlerjohn, line 91: These comments are hard to read on smaller screens. Can you place them on top of instead of beside lines?

sway-lib-std/src/vec.sw
```diff
@@ -14,7 +14,7 @@ struct RawVec<T> {
 impl<T> RawVec<T> {
     /// Create a new `RawVec` with zero capacity.
     fn new() -> Self {
-        Self {
+        RawVec {
```
adlerjohn, line 17: These changes look orthogonal to PR, please revert.

sway-lib-std/src/i128.sw
```diff
@@ -0,0 +1,145 @@
+library i128;
+
+use core::num::*;
+use ::assert::assert;
+use ::u128::U128;
+
+/// The 128-bit signed integer type.
+/// Represented as an underlying U128 value.
+pub struct i128 {
```
adlerjohn, line 9: One philosophical thing: should these instead be `I8`, `I16`, `I32`, `I64`, and `I128` (i.e. uppercase `I`)? They're structs after all, not primitives.

rostyslavtyshko: Good point. Applied it.

mohammadfawaz:
> > BTW, is there an issue for this? I didn't find one -- could be good to make one for reference so we can document motivating factors.
> 
> Added one

Do you mind further explaining how this library works and how to use it with additional comments (preferably docstrings) and in-depth description in this PR's description or in the corresponding issue? In particular, I'm interested in providing proper documentation for end users as well as some design notes for future maintainers.

adlerjohn (dismissed):
utACK as a first pass. I guess we'll find out soon enough if there are any correctness issues as people use it!

Can you file a follow-up issue to make the internal representation twos-complement, and also add Rust tests? Ultimately if this type exists, it'll be used at the ABI level, so it should `Just Work` with Rust.

Braqzen (commented):
1. I recommend chaning the syntax in the `From` trait for each type in addition to any `impl` that uses the name of the type rather than `Self` e.g.

```rust
impl From for I128 {
    /// Helper function to get a signed number from with an underlying
    fn from(value: U128) -> I128 {
        I128 {
            underlying: value,
        }
    }
}
```

would become

```rust
impl From for I128 {
    /// Helper function to get a signed number from with an underlying
    fn from(value: U128) -> Self {
        Self {
            value  // This requires the struct field to be updated. Alternatively, change the param to be "underlying"
        }
    }
}
```

2. I find the control flow in the `Multiply` & `Divide` implementations to be a little difficult to read. I presume the compiler can optimize away the number of `~Type::indent()` calls but I'm wondering if assigning to a variable, similar to `res`, and then using that variable in the flow is easier to read (perhaps also without an overhead).
3. In the `Sway` tests there are imports using `*`. Are these still needed to import the `impl` blocks?
4. The indentation for the struct declarations is messed up in the following tests
    - `i128`
    - `i256`
5. I would personally restructure the tests. I see two approaches.
    - Currently, each "test" (assertion) has its data defined directly above it and they flow from 1 test to another. You can definitely see where one test starts and one test ends if you look for it but it's not as quick and simple (imo) as it would be in a unit test function format.
    - The alternative is to define everything in one place (usually at the top) and then keep the tests (assertions) bundled together at the bottom. This does not allow you to shadow variable names so it might be tedious to name things but I think this enforces a higher quality structure and it's what I would prefer to see here. Otherwise, unit tests should be used but that cannot be done in a script atm

rostyslavtyshko:
> 1. I recommend chaning the syntax in the `From` trait for each type in addition to any `impl` that uses the name of the type rather than `Self` e.g.
> 
> 
> ```rust
> impl From for I128 {
>     /// Helper function to get a signed number from with an underlying
>     fn from(value: U128) -> I128 {
>         I128 {
>             underlying: value,
>         }
>     }
> }
> ```
> 
> would become
> 
> ```rust
> impl From for I128 {
>     /// Helper function to get a signed number from with an underlying
>     fn from(value: U128) -> Self {
>         Self {
>             value  // This requires the struct field to be updated. Alternatively, change the param to be "underlying"
>         }
>     }
> }
> ```
> 
>     2. I find the control flow in the `Multiply` & `Divide` implementations to be a little difficult to read. I presume the compiler can optimize away the number of `~Type::indent()` calls but I'm wondering if assigning to a variable, similar to `res`, and then using that variable in the flow is easier to read (perhaps also without an overhead).
> 
>     3. In the `Sway` tests there are imports using `*`. Are these still needed to import the `impl` blocks?
> 
>     4. The indentation for the struct declarations is messed up in the following tests
>        
>        * `i128`
>        * `i256`
> 
>     5. I would personally restructure the tests. I see two approaches.
>        
>        * Currently, each "test" (assertion) has its data defined directly above it and they flow from 1 test to another. You can definitely see where one test starts and one test ends if you look for it but it's not as quick and simple (imo) as it would be in a unit test function format.
>        * The alternative is to define everything in one place (usually at the top) and then keep the tests (assertions) bundled together at the bottom. This does not allow you to shadow variable names so it might be tedious to name things but I think this enforces a higher quality structure and it's what I would prefer to see here. Otherwise, unit tests should be used but that cannot be done in a script atm

1. Fixed
2. In earlier PR for this issue several approaches were checked and we came up to this solution.
3. Fixed
4. Fixed
5. I'm using the common pattern used in say u256, b512 or exponentiation test. Though I agree that unit tests should be do instead, as soon as they become avaliable.

simonr0204:
Is this ready for final review?

mohammadfawaz:
All the E2E tests need a `test.toml` and `json_abi_oracle.json` (unless you want to set `validate_abi` to `false` in `test.toml`. See https://github.com/FuelLabs/sway/tree/master/test/src/e2e_vm_tests/test_programs/should_pass/stdlib/assert_test for an example.

sway-lib-std/src/i128.sw
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+impl I128 {
+    /// Initializes a new, zeroed I128.
+    pub fn new() -> I128 {
+        I128 {
+            underlying: ~I128::indent(),
+        }
+    }
```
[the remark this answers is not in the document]
simonr0204: Resolved by [980e1c6](https://github.com/FuelLabs/sway/pull/2340/commits/980e1c65c97bef812c09b194aed1fc3f2cdcd12c), @tyshkor ?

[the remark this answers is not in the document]
rostyslavtyshko: Yes

Braqzen (commented):
1. There's a lot of sections that use `~Type::method()` rather than `~Self::method()`. Does the latter not work?
2. The `I256` and`I32` have a few areas where `Self` can be used. I suggest using a search and replace tool to highlight all of them for you so that it's easy to see where this change can be made

simonr0204:
@tyshkor don't forget to ping for final review If you think this is ready.

sway-lib-std/src/i128.sw
```diff
@@ -0,0 +1,155 @@
...
+            res = ~Self::from(~Self::indent() - (other.underlying - self.underlying));
+        }
+        res
+    }
+}
+
+impl core::ops::Multiply for I128 {
+    /// Multiply a I128 with a I128. Panics of overflow.
+    fn multiply(self, other: Self) -> Self {
+        let mut res = ~I128::new();
```
Braqzen, line 125: Can `Self` be used here?

sway-lib-std/src/i16.sw
```diff
@@ -0,0 +1,151 @@
...
+    fn add(self, other: Self) -> Self {
+        // subtract 1 << 15 to avoid double move
+        ~Self::from(self.underlying - ~Self::indent() + other.underlying)
+    }
+}
+
+impl core::ops::Subtract for I16 {
+    /// Subtract a I16 from a I16. Panics of overflow.
+    fn subtract(self, other: Self) -> Self {
+        let mut res = ~I16::new();
```
Braqzen, line 106: Can `Self` be used here?

sway-lib-std/src/i256.sw
```diff
@@ -0,0 +1,157 @@
...
+    fn add(self, other: Self) -> Self {
+        // subtract 1 << 63 to avoid double move
+        ~Self::from(self.underlying - ~Self::indent() + other.underlying)
+    }
+}
+
+impl core::ops::Subtract for I256 {
+    /// Subtract a I256 from a I256. Panics of overflow.
+    fn subtract(self, other: Self) -> Self {
+        let mut res = ~I256::new();
```
Braqzen, line 112: Can `Self` be used here?

sway-lib-std/src/i32.sw
```diff
@@ -0,0 +1,151 @@
...
+    fn add(self, other: Self) -> Self {
+        // subtract 1 << 31 to avoid double move
+        ~Self::from(self.underlying - ~Self::indent() + other.underlying)
+    }
+}
+
+impl core::ops::Subtract for I32 {
+    /// Subtract a I32 from a I32. Panics of overflow.
+    fn subtract(self, other: Self) -> Self {
+        let mut res = ~I32::new();
```
Braqzen, line 106: Can `Self` be used here?

sway-lib-std/src/i64.sw
```diff
@@ -0,0 +1,151 @@
...
+    pub fn indent() -> u64 {
+        9223372036854775808u64
+    }
+}
+
+impl I64 {
+    /// Initializes a new, zeroed I64.
+    pub fn new() -> Self {
+        Self {
+            underlying: ~I64::indent(),
```
Braqzen, line 55: Can `Self` be used here?

sway-lib-std/src/i64.sw
```diff
@@ -0,0 +1,151 @@
...
+    fn add(self, other: Self) -> Self {
+        // subtract 1 << 63 to avoid double move
+        ~Self::from(self.underlying - ~Self::indent() + other.underlying)
+    }
+}
+
+impl core::ops::Subtract for I64 {
+    /// Subtract a I64 from a I64. Panics of overflow.
+    fn subtract(self, other: Self) -> Self {
+        let mut res = ~I64::new();
```
Braqzen, line 106: Can `Self` be used here?

sway-lib-std/src/i128.sw
simonr0204: @rostyslavtyshko

rostyslavtyshko: Fixed

sway-lib-std/src/i256.sw
rostyslavtyshko: Fixed

sway-lib-std/src/i32.sw
rostyslavtyshko: Fixed

sway-lib-std/src/i64.sw
rostyslavtyshko: Fixed

sway-lib-std/src/i64.sw
rostyslavtyshko: Fixed

simonr0204 (approved):

sezna (approved):

Braqzen (approved):

PR #2343, opened 2022-07-18
Fix utf8 related bug in computing window for errors and warnings.

### LINKED ISSUE #2297: Panic while displaying a compile error related to unavailable struct field
The following fails to compile with a panic:
```rust
script;

struct Data<T> {
    value: T
}

impl<T> Data<T> {
  fn new(value: T) -> Data<T> {
    // this ↓↓↓↓↓
    Data {
        val // val is not a data member so this should be a compile error
    }
  }
}

fn main() -> bool {
    false
}
```
Panic:
```
thread 'main' panicked at 'SourceAnnotation range `(33, 57)` is bigger than source length `53`', /Users/mohammad/.cargo/registry/src/github.com-1ecc6299db9ec823/annotate-snippets-0.9.1/src/display_list/from_snippet.rs:286:9
```
We're likely generating an error but not actually printing it correctly.
Note that replacing `val` with `value` does work and compiles as expected.

vaivaswatha:
Closes #2297

### DIFF
--- a/forc-util/src/lib.rs
+++ b/forc-util/src/lib.rs
@@ -406,8 +406,16 @@ fn construct_window<'a>(
     let calculated_start_ix = calculated_start_ix.unwrap_or(0);
     let calculated_end_ix = calculated_end_ix.unwrap_or(input.len());
 
-    *start_ix -= std::cmp::min(calculated_start_ix, *start_ix);
-    *end_ix -= std::cmp::min(calculated_start_ix, *end_ix);
+    let start_ix_bytes = *start_ix - std::cmp::min(calculated_start_ix, *start_ix);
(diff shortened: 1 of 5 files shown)


otrho (approved):
This function... :shakes-fist:

mohammadfawaz:
> This function... :shakes-fist:
> 
> I mean, I reckon we've had 4 or 5 separate issues with `construct_window()` over the months. So hard to get right, and it's still fairly unreadable.

Yeah I was planning on rewriting it my last PR about utf8 characters but ended up being lazy and only fixing the immediate problem.  I'll open a `code quality` issue to clean this up.

mohammadfawaz (approved):

Issue #1849, opened 2022-06-04
Enable the use of `[]` for `StorageMap`.


mohammadfawaz:
It would be nice to be able to do this:

```rust
storage {
    map: StorageMap<u64, u64>,
}

...

storage.map[key] += value; // destructures to storage.map.insert(key, storage.map.get(key) + value);
```

adlerjohn:
Does this not require enshrining `StorageMap` into the compiler? Or will this work for anything that implements a storage insertion with a key?

mohammadfawaz:
Yeah we may need an Index trait like Rust has think 🤔

otrho:
#429

Issue #429, opened 2021-11-30
Index and IndexMut trait.


otrho:
#424 introduced static arrays and indexing them though indexing other types we need the `Index` and possibly `IndexMut` traits.  These would mimic the Rust equivalents.

The existing code tries to call `self.index(arg)` by in fact just calling `std::ops::index(self, arg)` which may itself be a bit wrong... I'm not sure how it should work mechanically.

PR #2345, opened 2022-07-18
StorageVec fixing issues and adding features

### LINKED ISSUE #2325: StorageVec issues and features
Please reply with any features you would also like in StorageVec

Known issues:

- Insert will panic on an underflow if length (and index) is 0

Needed features:

- a mutate or replace method, as there cannot be any references for elements here (I think)
- abstractions on the calculations of keys for store()

SwayStar123:
closes #2325


sway-lib-std/src/storage.sw
```diff
@@ -236,6 +236,31 @@ impl<V> StorageVec<V> {
...
+    /// # Arguments
+    /// 
+    /// * `index` - The index of the vec to set the value at
+    /// * `value` - The value to be set
+    ///
+    /// # Reverts
+    /// 
+    /// Reverts if index is larger than the length of the vec
+    #[storage(read, write)]
+    pub fn set(self, index: u64, value: V) {
```
Braqzen, line 250: @mohammadfawaz could the compiler somehow desguar into this from `storage.storage_vec[1] = 42;`?

mohammadfawaz: Not yet unfortunately. This requires the `Index` trait: https://github.com/FuelLabs/sway/issues/429

Also related: https://github.com/FuelLabs/sway/issues/1849

sway-lib-std/src/storage.sw
```diff
@@ -236,6 +236,31 @@ impl<V> StorageVec<V> {
...
+        let len = get::<u64>(__get_storage_key());  
+        // if the index is higher than len, cannot set an element.
+        assert(index <= len);
```
mohammadfawaz, line 253: Isn't it better to only allow setting indices that are within range? I don't expect this function to change the length of the vector  (unlike `push` and `insert`).

Also, soliciting feedback from @sezna @otrho @mitchmindtree regarding adding a function `set()` that I don't believe Rust has (but Rust has mutable references which we don't have yet). If we had `set()` here we also have to add `set()` in `Vec`.

SwayStar123: Well you could make the same argument for insert. If there nothing to insert between why do we allow it? But we still do. Itd be inconsistent to allow it for insert and not for set IMO

mohammadfawaz: Hmm yeah I see where you're coming from. My argument is that `insert()` always increases the length by 1 but `set()` does not. A function called `insert()` is always expected to grow the size because it's inserting something somewhere, and the end of a vector is a possible place for where the new element could go. A function called `set()` however is expected to set (aka change, update, etc.) an existing value. In the case of `Vec()` for example, I do not expect `set()` to ever incur allocation cost. Also, `set()` is meant to replace updating a mutable reference, and you can't really get a valid mutable reference to index `len()`.

SwayStar123: hmm yeah i see what you mean, ill change that

Braqzen (approved):
- I have noticed that two files have been formatted in this PR when they should be changed in a separate PR. For future reference, could you undo the formatting there and make sure that only the work related to your ticket is being committed?
    - test_projects/call_frames/mod.rs
    - test_projects/option_field_order/mod.rs

otrho (approved):

Issue #324, opened 2021-10-24
improve LSP dealing with imports


leviathanbeak:
current LSP implementation is oblivious to dependencies/imports from other files,
that is, it does not know from which (other) file was a value imported - so consequently we cannot make changes across the whole workspace, only within a single file.

@sezna informed me that Span contains path to a File, but for me that value is always None,
will update here, after discussing it some more with Alex, and then either fix it on my side or report a bug on compiler side

JoshuaBatty:
This should be mostly resolved now.

PR #2346, opened 2022-07-19
Parse tokens from all dependencies instead of single files in the language server

### LINKED ISSUE #2318: Reorganize language server to handle full projects instead of single files
Will involve reorganizing how the `document.rs` and `session.rs` files are currently structured. We will also want to write some nice helper methods around the `TokenMap` to get all tokens for a specific file etc...

This should mostly solve #271 once implemented properly

### LINKED ISSUE #2319: Start using the typed AST tokens in the language server
At the moment we are only using non-typed tokens collected during traversal of the `ParseProgram` from `sway-core`. Most of the work for collecting typed tokens has already been done in #2121. We just need to properly handle working with full projects instead of single files first which is being tracked in #2318. 

After this is done we can start adding back LSP capabilities that require knowledge of the tokens type information.

JoshuaBatty:
This PR now traverses the submodules in a project and collects those tokens into the `TokenMap`. 

I've moved a lot of the logic that was previously in `document.rs` into `session.rs`. Also ended moving / simplifying a bit of code so it's easier to read.

Finally, I've changed our `TokenMap` type to use a `DashMap` instead of a `HashMap`.

closes #2318 

partially addresses #324


sway-lsp/src/core/session.rs
```diff
@@ -19,16 +29,98 @@ pub type Documents = DashMap<String, TextDocument>;
...
 impl Session {
     pub fn new() -> Self {
         Session {
             documents: DashMap::new(),
             config: RwLock::new(SwayConfig::default()),
+            token_map: DashMap::new(),
+        }
+    }
+
+    /// Check if the code editor's cursor is currently over an of our collected tokens
```
mitchmindtree, line 44:
```suggestion
    /// Check if the code editor's cursor is currently over one of our collected tokens.
```

sway-lsp/src/core/session.rs
```diff
@@ -19,16 +29,98 @@ pub type Documents = DashMap<String, TextDocument>;
...
+                (key.clone(), token.clone())
+            })
+            .collect()
+    }
+
+    pub fn declared_token_ident(&self, token: &TokenType) -> Option<Ident> {
+        // Look up the tokens TypeId
+        match utils::token::type_id(token) {
+            Some(type_id) => {
+                tracing::info!("type_id = {:#?}", type_id);
```
mitchmindtree, line 102: Is this left over from debugging?

sway-lsp/src/core/session.rs
```diff
@@ -19,16 +29,98 @@ pub type Documents = DashMap<String, TextDocument>;
...
+
+                // Use the TypeId to look up the actual type
+                let type_info = sway_core::type_engine::look_up_type_id(type_id);
+                tracing::info!("type_info = {:#?}", type_info);
```
mitchmindtree, line 106: Same here

sway-lsp/src/server.rs
```diff
@@ -70,24 +72,27 @@ fn capabilities() -> ServerCapabilities {
...
+        diagnostics: Vec<Diagnostic>,
+        token_map: &TokenMap,
+    ) {
         // If parsed_tokens_as_warnings is true, take over the normal error and warning display behavior
         // and instead show the parsed tokens as warnings.
         // This is useful for debugging the lsp parser.
         if self.config.parsed_tokens_as_warnings {
-            if let Some(document) = self.session.documents.get(uri.path()) {
-                let diagnostics = debug::generate_warnings_for_parsed_tokens(document.token_map());
+            //let diagnostics = debug::generate_warnings_for_parsed_tokens(token_map);
```
mitchmindtree, line 85: Left over from debugging?

mitchmindtree (commented):
Nice one, looking much cleaner! Just a couple nits, otherwise happy to approve :heavy_check_mark:

sway-lsp/src/core/session.rs
JoshuaBatty: Yep good spots

sway-lsp/src/server.rs
JoshuaBatty: I actually had the wrong line commented out, thanks for this.

JoshuaBatty:
> Nice one, looking much cleaner! Just a couple nits, otherwise happy to approve ✔️

Thanks, @mitchmindtree just addressed those nits, should be good to go.

mitchmindtree (approved):
:shipit:

otrho (approved):

PR #2359, opened 2022-07-22
Fix lexing of multiline comments to include the trailing '/'

### LINKED ISSUE #2355: Comment lexing misses trailing `/` for multi-line comments
Comment lexing works perfectly while the comments are single-line comments like the following:

```
// This is a comment
```

but if the comments are in the form of multiline we are missing the trailing `/` in the comment's span. So for this example:

```
/*This is a comment*/
```

`.span().as_str()` returns  `/*This is a comment*`

I feel like this shouldn't be hard to tackle so added the "good first issue" label but I did not touch lexer enough to be sure about that so please remove it if that is not the case 🙌 . 

Depending on the discussion at https://github.com/FuelLabs/sway/pull/2311#discussion_r926167383 I can either fix it in #2311 or take a look at this after that is merged.

Related to https://github.com/FuelLabs/sway/pull/2311#discussion_r926167383

mitchmindtree:
Previously, both the implementation and the test missed the trailing
slash during lexing. This fixes both, closes #2355 and unblocks #2311.

### DIFF
--- a/sway-fmt-v2/src/utils/comments.rs
+++ b/sway-fmt-v2/src/utils/comments.rs
@@ -163,7 +163,7 @@ mod tests {
             .unwrap();
         assert_eq!(
             found_comment.1.span.as_str(),
-            "/* multi-\n             * line-\n             * comment *"
+            "/* multi-\n             * line-\n             * comment */"
         );
     }
 }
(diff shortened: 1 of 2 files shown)


JoshuaBatty (approved):

mitchmindtree:
Hrmmrmmm seemingly unrelated `cargo-unused-deps-check` error.

Seems like potentially some issue in nightly-Rust causing a new compile error to appear in `mdbook` `0.4.19` :thinking:

JoshuaBatty (approved):

otrho (approved):

Issue #154, opened 2021-08-18
Reference Types


sezna:
The `ref` and `deref` keywords should be used to construct reference types and dereference reference types. This issue involves the parsing, type checking, and code generation of reference types.

PR #2363, opened 2022-07-22
Add pointers to IR and implement mutable parameters.

tritao:
So this PR does a few different things:

- Adds first-class pointer type in the IR
- Changes the type of `get_ptr` IR instruction to return a pointer
- Updates the IR verifier to deal with this
- Allows Sway to parse mutable function parameters
- Updates IR generation to generate a pointer for mutable parameters
- Adds support for a ref keyword when declaring parameters
- Adds new errors to check for unsupported cases with the ref keyword

These changes support code like mutable primitive parameters:

```rust
fn mut_prim(mut b: u32) {
    b = 20;
}

fn main() -> u32 {
    let mut b = 0u32;
    mut_prim(b);
    b
}
```

`fn mut_prim(mut b: u32)` now generates the following IR:

```rust
fn mut_prim_0(b !8: mut ptr u64) -> (), !9
```

And calling it generates:

```rust
v2 = get_ptr mut ptr u64 b, ptr u64, 0, !4
v3 = call mut_prim_0(v2), !5
```

Update: There is also support for general mutable reference parameters now. 

I am looking for feedback on this, particularly on the IR changes and the verifier. Adding pointers like this both simplifies some semantics but also complicates the verifier a bit, what do you think?

Another step towards a full fix for https://github.com/FuelLabs/sway/issues/1188.


sway-core/src/asm_generation/from_ir.rs
```diff
@@ -381,6 +381,9 @@ impl<'ir> AsmBuilder<'ir> {
                             &ptr_content.ty
                         ));
                     }
+                    Type::Pointer(_) => {
+                        unreachable!("pointers not expected here")
```
sezna, line 385: could we explain more about why this is an invariant in the panic message?

otrho (commented):
I haven't looked thoroughly at all of this change yet because it's actually quite significant and I'd like to discuss the design a bit before committing. 

Until now the IR hasn't really had proper pointers, even though it has the `Pointer` struct.  They've been used exclusively for local variables, which are _always_ pointers.  Hence there has been no need to add a pointer variant to `Type`.  Also, by adding the existing `Pointer` designed for locals directly as a `Type` variant it doesn't make sense for `Pointer` to have an initialiser.  `Pointer` really should be called `LocalVariable` or something.

Aggregates are types which pretty much need them, as they're 'reference types' in Sway, but they currently aren't explicitly pointers in IR.  They should be though, and part of the aggregates refactor will be to make them so and to stop using `load` and `store` on them in such a high level.  i.e., right now you can create an elaborate const struct and just `store` it like you would an integer.  An explicit `memcpy` should be called instead.

Recently-ish LLVM switched to using 'opaque' pointers for all sorts of reasons which seem to make sense, and I was thinking we should probably do the same thing, but I'm not 100% sure and would like to discuss it with the team.  It makes sense to decouple the pointee type and the pointer value as it reduces friction and clutter around casting and lower level memory operations.  But I don't know for sure. :slightly_smiling_face: 

I wonder though, for this change to pass mutable args to Sway and IR, it might be simpler to just have a richer `Argument` which can tag that it's mutable and/or a reference.   Then the bigger question of how to properly to pointers can come later.

sway-ir/src/context.rs
```diff
@@ -43,6 +43,10 @@ pub struct Context {
     pub(crate) metadata_storage_indices: HashMap<StorageOperation, MetadataIndex>,
 
     next_unique_sym_tag: u64,
+    pub fn_param_is_mutable_primitive: bool,
+
+    // Enables extra debug information while printing IR
+    pub debug_ir: bool,
```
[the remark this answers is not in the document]
tritao: I agree with you, I don't like how this was done in the PR, it doesn't feel clean, but at the same time the "correct" fix would mean adding a ton of changes which would make this PR harder to review, so I thought we could clean it up in a following PR as we already got some stuff in there anyway. 

Maybe we can leave `Context` as-is, but extract all the relevant fields into a new type, say `CompilationResult` or something, which we return instead.

sway-ir/src/instruction.rs
```diff
@@ -165,6 +165,167 @@ pub enum Register {
 }
 
 impl Instruction {
+    /// Return a string representation of type, used for debug printing.
+    pub fn as_debug_string(&self, context: &Context) -> String {
+        match self {
```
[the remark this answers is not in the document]
tritao: I thought about that, but the format of this `as_debug_string` is in the style of `fmt::Debug`, the idea was to hopefully keep a consistent style of debugging output, whereas the printer outputs the information in a different style.

Anyway this debug stuff is not really necessary for this PR, so I can take it out and can send a separate PR adding the debugging-specific code.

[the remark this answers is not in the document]
otrho: I still disagree here.  The serialised IR should contain everything about the IR, particularly since we parse it back into IR from this format.  In that regard, it should be fine for debug output and duplicating all the effort is wasteful and harder to maintain.

It'd be nice to have `Debug` (and `dbg!()`) or `Display` but it isn't possible due to needing the `Context` for nice output.

sway-core/src/ir_generation/function.rs
```diff
@@ -98,6 +100,7 @@ impl FnCompiler {
         // Filter out all ast nodes *after* a `break` statement. Those nodes are essentially dead.
         let value = ast_block
             .contents
+            .clone()
             .into_iter()
```
otrho, line 104: Usually rather than cloning and then moving into the iterator it's probably better to just iterate over the references.  This is only because `ast_block` hasn't been moved here though, which I'm not sure why you've changed.

sway-core/src/semantic_analysis/ast_node/declaration/function/function_parameter.rs
```diff
@@ -67,17 +72,13 @@ impl TypedFunctionParameter {
                     is_constant: IsConstant::No,
                     span: parameter.name.span(),
                 },
-                is_mutable: if parameter.is_mutable {
-                    VariableMutability::Mutable
-                } else {
-                    VariableMutability::Immutable
-                },
+                is_mutable: mutability.clone(),
```
otrho, line 75: `VariableMutability` is just an enum, should probably be `Copy` so we don't need to explicitly `.clone()` it, but this is fine for now.

sway-ir/src/irtype.rs
```diff
@@ -81,9 +86,32 @@ impl Type {
...
         }
     }
+
+    pub fn strip_ptr_type(&self, context: &Context) -> Option<Type> {
+        if let Type::Pointer(ptr) = self {
+            Some(*ptr.get_type(context))
+        } else {
+            Some(*self)
+        }
+    }
```
otrho, line 101: This can just return `Type`.

otrho (commented):
OK, just a few questions below.

I'm going to make a new issue addressing a pointers refactor for IR in general, including formalising how args and locals should work and especially to fix the copy-type / reference-type ambiguity which still remains.

But until then this is good. 🙂

mohammadfawaz:
Do you mind updating the description of the PR with the latest version of this? For example, "Changes the type of get_ptr IR instruction to return a pointer" seems to be outdated?

mohammadfawaz:
> 

Oh misread that part in the code, sorry! Thanks for keeping up-to-date :)

sway-core/src/ir_generation/compile.rs
```diff
@@ -155,16 +156,30 @@ pub(super) fn compile_function(
...
+        Ok((
+            param.name.as_str().into(),
+            if param.is_mutable && look_up_type_id(param.type_id).is_copy_type() {
+                Type::Pointer(Pointer::new(context, ty, param.is_mutable, None))
+            } else {
+                ty
+            },
+            param.name.span(),
+        ))
+    })?
```
otrho, line 180: I think you can remove that `Ok` and the `?` on line 180?

`convert_resolved_typeid()` returns a `Result`, and so will the call to `.map()`, so there's no need to introduce another `Result`.

otrho (commented):
Just the one little nit.

otrho (approved):

otrho:
No, I thought the whole point was to get an MVP going now to unblock whatever needs mutable copy type args (`Vec`?).

We had the `ref` keyword for a while, then we took it out.  Unlike Rust we've adopted an implicit 'copy-type' or 'ref-type' model rather than explicit references.  We probably should add explicit references, just like #2449 will for IR.  But I was under the impression this issue is a bit of a blocker for the stdlib and we needed something now.

If it's not blocking/urgent then yeah, we might as well just do #2449 and add explicit `ref` to Sway, but that's far from a trivial job.  The changes to the type checker alone will be very tricky, I imagine.  We've been thinking about this for a while -- see #154 from a year ago, which is still `priority: low`.  #1188 OTOH is `priority: critical`.

otrho:
What do you think @sezna ?

mohammadfawaz:
> What do you think @sezna ?

Yes being able to pass `mut Vec` as an argument is very important and can be easily explained by saying reference types are passed by reference by default so it's fine as is until we have the `ref` keyword. Being able to pass `mut` primitive types is not as important. So if we're concerned with the lack of `ref`, we can skip the logic for `mut` primitive types or keep it but disallow it in the parser.

otrho:
I was thinking about this last night.  I don't think we should ever introduce explicit references.  Sway is designed to be as safe as possible, since it's for writing contracts!  Adding references but keeping it safe will be a lot of work -- we'd probably need to implement a borrow checker, which IMO would not be worth the effort.

We can keep the conventions, and document them heavily, so we obviously _do_ have references, just implicitly.  If we were to allow references to be first class types, especially references to references without the guarantees of a whole lot of static analysis then we're asking for trouble.

sezna:
> I was thinking about this last night. I don't think we should ever introduce explicit references. Sway is designed to be as safe as possible, since it's for writing contracts! Adding references but keeping it safe will be a lot of work -- we'd probably need to implement a borrow checker, which IMO would not be worth the effort, or have a bunch of hardcoded exceptions as to where they're allowed to be used. Not in storage, or across the FFI for example.

Yeah, the main reason we don't allow explicit references right now is precisely that: we'd need to implement a borrow checker. 
In Rust, since it is a "systems" language, the ability to very specifically control memory and memory usage is important. However, in Sway, we shouldn't expose that much power (and vulnerability) to the user unless there's a very strong use case. Additionally, Sway is VM-agnostic and the more complex we make the memory  model on the front end, the more work must be done to ensure consistent behavior across back ends.

If we were to introduce reference types, I'd be more keen to introduce something like `Rc<T>` which would be a safe, reference counted pointer. That could be accomplished mostly (entirely?) in library code and would not require a borrow checker.

Another thing to keep in mind is that Sway doesn't free or drop memory like Rust does. Since freeing memory costs gas, and contracts typically won't run out of memory, we omit freeing memory. This actually removes the need to check for use-after-free. We could perhaps leverage that fact to make a simpler borrow checking model.


> We can keep the conventions, and document them heavily, so we obviously do have references, just implicitly. If we were to allow references to be first class types, especially references to references without the guarantees of a whole lot of static analysis then we're asking for trouble.


If I'm not mistaken, this PR specifically unblocks/implements mutable primitives ight? As we already have the ability to support mutable structs and enums since they're implicitly reference types. So really this boils down to the priority of mutable primitive types, and a design for them.

From the frontend perspective, what @tritao has written here where `mut` implies a reference, but we don't introduce a specific ref keyword, is actually pretty intuitive to me. If a var is immutable, then you definitely don't care about whether it was copied or not. If it is mutable, of course it wouldn't be. 

> Being able to pass mut primitive types is not as important

This is true that there are higher priority things, but we should certainly be able to pass mutable primitive types by v1

sezna:
I did some more thinking and we certainly won't ever need a ref counted type since we don't ever free memory. And if structs and enums are always reference types, we can create a library type for references like this:
```rust
struct Ref<T> {
  inner: T
}

fn foo(mut inner: Ref<u32>) {
// omitted
```

test/src/e2e_vm_tests/test_programs/should_pass/language/ref_mutable_fn_args_bool/src/main.sw
```diff
@@ -0,0 +1,11 @@
+script;
+
+fn mut_arg(ref mut b: bool) {
+    b = true;
+}
+
+fn main() -> bool {
+    let ref mut b = false;
```
mohammadfawaz, line 8: Hmm I thought we didn't want the `ref` keyword to be allowed anywhere other than function/method parameters. This example would basically be:
```rust
fn mut_arg(ref mut b: bool) {
    b = true;
}

fn main() -> bool {
    let mut b = false;
    mut_arg(b);
    b
}
```
Having to define the variable as `ref` has a lot of implications and I'm not sure we want to go down that rabbit hole 🤔

test/src/e2e_vm_tests/test_programs/should_fail/ref_mutable_fn_args_struct/test.toml
```diff
@@ -0,0 +1,3 @@
+category = "fail"
+
+# check: $()This parameter was declared as a reference, which is implied for non-primitive types.
```
mohammadfawaz, line 3: While this is true, I think we still want the `ref` keyword for reference types. Basically, my understanding is that we want the `ref` keyword to be added anywhere we had `mut` only before. So, our options are:
```rust
fn mut_arg_non_primitive(ref mut s: S) {
    s = S{}
}
fn mut_arg_primitive(ref mut b: bool) {
    b = true;
}
fn mut_arg_non_primitive(s: S) {
    // Can't modify `s` here
}
fn mut_arg_primitive(b: bool) {
    // Can't modify `b` here
}
```

otrho: Oh dear, so tricky.  Syntactically I thought the only changes we're making were mut args now _must_ have the `ref` qualifier, but this is _only_ for args, not local variables, and yeah, for all mut args regardless of their type.

So for mut args, it's always `ref mut` and for mut locals it's always just `mut`.

sway-fmt-v2/src/items/item_fn.rs
```diff
@@ -105,9 +105,14 @@ impl Format for FnSignature {
             }
             FnArgs::NonStatic {
                 self_token,
+                ref_self,
                 mutable_self,
                 args_opt,
             } => {
+                // `mut `
```
mohammadfawaz, line 112:
```suggestion
                // `ref `
```

sway-core/src/semantic_analysis/ast_node/mod.rs
```diff
@@ -795,7 +797,7 @@ fn type_check_trait_methods(
                             span: name.span(),
                         },
                         // TODO allow mutable function params?
```
mohammadfawaz, line 799: Is this TODO relevant now?

sway-core/src/ir_generation/function.rs
```diff
@@ -593,7 +595,7 @@ impl FnCompiler {
 
                 // We're still undecided as to whether this should be decided by type or size.
                 // Going with type for now.
```
mohammadfawaz, line 597: This comment can be removed. We did decide to go with the type :)

mohammadfawaz (commented):
This all makes sense to me and the testing is great! Seems like the best use model given what we would like to support. Thank you doing for this!

Just 3 tiny comments and I'll approve.

mohammadfawaz (approved):
Excellent!

sway-core/src/error.rs
```diff
@@ -538,6 +538,10 @@ pub enum CompileError {
         variable_name: Ident,
         span: Span,
     },
+    #[error(
+        "This parameter was declared as mutable, which is not supported yet, did you mean to use ref mut?"
+    )]
```
otrho, line 543: I would argue we'll never support mutable parameters, the `ref` bit is pretty much documentation.  But that's not a deal breaker.

otrho (approved):

Issue #2365, opened 2022-07-22
Search locally for git sources before fetching them in online mode


kayagokalp:
While working on #1787, I realized that the implementation for it would also open up the way for some easy speed-ups for online mode as well. Currently we are directly trying to fetch git repos from remote which in the case of `std` takes some time as it fetches the `sway` repo. What we can do (as a TODO comment in the `fetch_git()` suggests) is basically looking for the repo locally first. If we cannot find it we should fallback to fetching the remote repo. 

> I implemented the logic for git sources other than `branch` as it seemed to be easier to start with and I am providing some benchmarks so that we can confirm the searching for a local version is manageable. For comparison i created an empty project which (implicitly) depends on `std`. 
> 
> Building the empty project without searching for a local repo (and doing the fetch & pin stuff takes) **11.5 seconds** on my system 
> Building the empty project with searching for a local repo takes **1.35 seconds** on my system which feels like butter :butter: 
> 
> I think it is worth searching for a local copy even if we are in `online` mode and if we cannot find one then we can fetch & pin (as a `TODO` suggests at the relevant part of the code).

_Originally posted by @kayagokalp in https://github.com/FuelLabs/sway/issues/1787#issuecomment-1192562698_

Issue #919, opened 2022-03-10
Add a `Filesystem` abstraction similar to cargo's for managing access to the `forc` user directory


mitchmindtree:
As discussed [here](https://github.com/FuelLabs/sway/pull/917#issuecomment-1063618316), it would be nice to have a clearer approach to managing exclusive access to shared directories/files within the user's `~/.forc` directory.

We should consider taking inspiration [from cargo](https://github.com/rust-lang/cargo/blob/master/src/cargo/util/flock.rs#L102-L111), and perhaps investigate whether or not there's a dedicated crate to managing concurrent access to a directory tree between multiple instances of the same program.

bingcicle:
Unassigning myself and setting this as low priority for now since after working on this we realized that there isn't much motivation nor need for a `Filesystem` abstraction at the moment ([PR](https://github.com/FuelLabs/sway/pull/3484), see comment [here](https://github.com/FuelLabs/sway/pull/3484#pullrequestreview-1201893997)).

PR #2366, opened 2022-07-22
When --offline flag specified for forc build, attempt to use local git checkouts for git dependencies before failing

### LINKED ISSUE #1787: When `--offline` flag specified for `forc build`, attempt to use local git checkouts for git dependencies before failing

kayagokalp:
closes #1787.

unblocks #2365.


forc-pkg/src/pkg.rs
```diff
@@ -590,7 +590,7 @@ impl GitReference {
     pub fn resolve(&self, repo: &git2::Repository) -> Result<git2::Oid> {
         // Find the commit associated with this tag.
         fn resolve_tag(repo: &git2::Repository, tag: &str) -> Result<git2::Oid> {
-            let refname = format!("refs/remotes/origin/tags/{}", tag);
```
sezna, line 593: was this wrong before? is there documentation for this and should we make it a constant or something?

kayagokalp: The thing is we were fetching the package with an anonymous remote, which [cannot](https://docs.rs/git2/latest/git2/struct.Repository.html#method.remote_anonymous) be converted to a persisted remote. Since I needed the fetched package to be a valid git repo that we can interact with, later on, I changed that. After this change once `resolve` was failing and once I checked the fetched repo with `git show-ref`, I realized `refnames` changed. I think it is because of the change from `remote_anonymous` to `remote`.

I couldn't find why this is happening in the documentation of `git2`.

forc-pkg/src/pkg.rs
```diff
@@ -1364,17 +1374,138 @@ pub fn fetch_git(fetch_id: u64, name: &str, pinned: &SourceGitPinned) -> Result<
...
-
+        let repo_path = repo.path();
         // Checkout HEAD to the target directory.
         let mut checkout = git2::build::CheckoutBuilder::new();
         checkout.force().target_dir(&path);
         repo.checkout_head(Some(&mut checkout))?;
+
+        // Copy current_dir to target
+        let copy_options = fs_extra::dir::CopyOptions::new();
+        fs_extra::copy_items(&[repo_path], &path, &copy_options)?;
```
mitchmindtree, line 1385: A few lines above, we checkout only the commit that we care about to the target `path`.

It looks like this new code copies *everything* within the repo into the same `path`, rather than only the desired commit.

It looks like the aim here is to provide access to the associated git repository as a sub-directory, so that this repo can be queried (via `git`) during future searches of the checkouts directory. Does this sound correct?

If that's the case, ideally we should not have both the checkout and the whole repo as a subdirectory. Repos can get large, e.g. the `std` checkout is only ~6 MB for the Sway repo, but some repos can contain other large assets like media, etc. We should do our best to keep these checkouts as small as possible, and especially avoid doubling the repo contents.

As an alternative to this copy, I wonder if we can include the `.git` directory along with the checkout somehow? Or alternatively, strip down this repo to the commit we need, and copy it *rather* than doing the checkout?

Another alternative might be, rather than cloning the whole repo, we could only store some tiny file that just contained the `SourceGit` that was used to produce the commit. I guess this would only work for git sources specified via tag or commit hash, and wouldn't work for branches as we couldn't know which is the latest commit :thinking:

forc-pkg/src/pkg.rs
```diff
@@ -1215,7 +1215,7 @@ where
     if tags {
         fetch_opts.download_tags(git2::AutotagOption::All);
     }
-    repo.remote_anonymous(source.repo.as_str())?
+    repo.remote("origin", source.repo.as_str())?
```
mitchmindtree, line 1218: Now that this is no longer an anonymous remote, perhaps we should make this a constant and re-use it wherever "origin" appears?

mitchmindtree (commented):
Nice one @kayagokalp! Just had a quick first review as I've been pushing back reviewing this for too long :)

forc-pkg/src/pkg.rs
kayagokalp: Hmm, that is correct. I just checked the difference between the old method and this approach, for fetching std there seems to be around 8 MB of difference and the size of the `.git` folder is about also 8MB. Thanks for the detailed possible workarounds but I think I am a little confused.

By adding the copy, aren't we just adding the `.git` directory (on top of the already present checkout, because before executing the checkout I believe there is only the `.git` directory present in the `tmp` folder)? If that is the case can we really avoid it? Because in offline mode if the user wanted to add a git package by providing a branch, we should be able to find the most recent one we have locally. If we got the `.git` folders intact in the _all_ checkouts, we can find the most recent one in a checkout that we did for a non-branch version of the same repo. 

For example, imagine we are adding an `std` dependency that points to this branch, and after some while, we remove that and point to the latest release implicitly. In the meantime, this branch is also updated. In offline mode, if we try to add a dependency that points to this branch, (i believe) the most recent commit that we could find from this branch would be coming from the second checkout (the one that we did for the implicit dep declaration). So to be able to find those, I thought we need the full `.git` dir. 

On the other hand, this whole `.git` and size stuff got me thinking and I realized if we got the `.git` dir we don't need to store anything else and expand to the full repo from there once we need it. 
> if that's the case, ideally we should not have both the checkout and the whole repo as a subdirectory

I am not sure if that is what you meant here 😄 

Testing this approach yields  a difference from 20MB(with this PR) to 8MB. Before this PR same checkout seems to take about 11mb on disk. I will look into this in more in detail after our wallet stuff but wanted to clarify a little as I am a little confused rn 😄

kayagokalp:
Converting this into draft until we are done with `fuel-client` stuff as it has a higher priority.

kayagokalp:
Quick update on my status :) @mitchmindtree, so as we discussed briefly yesterday, just storing `.git` in the checkouts could mean that different `forc` processes might get into a racing condition. Then I switched over to another approach, just storing the HEAD commit should significantly make the checkout smaller. Unfortunately only the CLI , `libgit` does not support `--depth`. So `git2-rs` also cannot offer this. I am thinking of doing something like:

1. Check if the user have `git` CLI installed. If they have we can fetch it with `--depth 1` and this will reduce the size and increase the speed a lot.
2. If they do not have `git` CLI installed and if the git repo is from github, we can get the exact commit hash with a tmp repo and fetch exactly that similar to https://github.com/rust-lang/cargo/issues/10078
3. If we cant go with step 1 or 2, unfortunately we need to get the repo and the `.git` folder so that it can be queried. Since we couldn't do step 1 or 2, we will have the complete `.git` folder intact

Does this sound reasonable? 👀

kayagokalp:
After a quick discussion over slack with @mitchmindtree, decided to go different path. I guess this is ready for a review 👍 **Quick summary of what is happening here:**

1. In online mode, when the `forc` fetches a dependency, it also creates a small indexing file `.forc_index` which holds the information about that repo (time of the HEAD commit, HEAD commit hash and type of the reference: branch, rev, tag). We are keeping this because we are omitting `.git` folder to save disk space and we need a way to import some information about the repo once we are in offline mode.
2. Later on, when the user want's to build in `offline` mode, we search locally available checkouts. We do that by first filtering by the name of the package we are searching for and then we read the index file from each available checkout. If the git reference of the desired package is `rev` or `tag`. We search for the exact match. If the git reference of the desired package is `branch` we look for the best match. Best match here means, the most recent commit we have for the desired branch. 

Since `.forc_index` is generated deterministically (for the same repo/reference we are generating the exact same `.forc_index` file), multiple instances of `forc` can work together doing the same checkout. Since the file content in the end will be intact, there won't be a problem later on when we got into `offline` mode. Similarly in the `offline` mode, multiple instances of `forc` can work with the same checkout as they will only read the indexing file.

Only possible problem might be with a `forc` process fetching a repo that already exists in online mode , while another `forc` process trying to find it in the offline mode. Since before the checkout is completed we check if the target dir exists. If that is the case we are deleting it. So the time between the deletion of the target folder and the checkout, the searching `forc` process won't be able to find the repo it is looking for. I believe this can also happen before this PR too, because the checkout is done after everything is removed and there is a small time-window where we don't have the repo.

I tried to run 16 concurrent `build` operation 10 times and couldn't catch something like this. I am not sure if I am missing something but, if this seems possible and significant enough we can introduce something like a `.forc_ready` file. If a `.forc_read` file is not present, searching could wait until it is there (with a time limit). I believe i saw something similar in `cargo` somewhere but couldn't find it now 🤔 

### My Testing Procedure:

1. Create a new project
2. Build the project with implicit std (default) in online mode
3. Add std as an explicit dependency and point to `master` branch
4. Build the same project in online mode.

After these steps you will have both master and latest tag checkouts in your local.

1. Remove the explicit std dependency declaration from the toml file
2. Build the project with `--offline` flag
3. Add back the std explicit dependency which points to `master` branch
4. Build the project with `--offline` flag

forc-pkg/src/pkg.rs
```diff
@@ -1369,12 +1412,174 @@ pub fn fetch_git(fetch_id: u64, name: &str, pinned: &SourceGitPinned) -> Result<
...
+fn find_repo_with_rev(
+    rev: &str,
+    package_name: &str,
+    checkouts_dir: PathBuf,
+) -> Result<Option<(PathBuf, String)>> {
+    let mut found_local_repo = None;
+    with_search_checkouts(
+        checkouts_dir,
+        package_name,
+        &mut |repo_index, repo_dir_path| {
```
mitchmindtree, line 1536:
```suggestion
        |repo_index, repo_dir_path| {
```

forc-pkg/src/pkg.rs
```diff
@@ -1369,12 +1412,174 @@ pub fn fetch_git(fetch_id: u64, name: &str, pinned: &SourceGitPinned) -> Result<
...
+            for repo_dir in fs::read_dir(entry.path())? {
+                // Iterate over all dirs inside the `name-***` directory and try to open repo from
+                // each dirs inside this one
+                let repo_dir = repo_dir
+                    .map_err(|e| anyhow!("Cannot find local repo at checkouts dir {}", e))?;
+                if repo_dir.file_type()?.is_dir() {
+                    // Get the path of the current repo
+                    let repo_dir_path = repo_dir.path();
+                    // Get the index file from the found path
+                    let index_file = fs::read_to_string(repo_dir_path.join(".forc_index"))?;
```
mitchmindtree, line 1573: Rather than returning on `Err` here, lets just skip this repo if it doesn't have an index file. This should allow us to land this without users having to delete their old `.forc` directories, and should avoid returning an `Err` in the case that a repo has been checked out but the `.forc_index` file has not yet been written.

mitchmindtree (commented):
Awesome stuff @kayagokalp, this is looking great! Just a few nits then happy to approve :+1: 

> I believe this can also happen before this PR too, because the checkout is done after everything is removed and there is a small time-window where we don't have the repo.

Yeah, that's correct! See #2603 which managed to trigger this with a very specific setup on CI. So far it's remained rare enough that this hasn't been a big priority just yet.

> I am not sure if I am missing something but, if this seems possible and significant enough we can introduce something like a .forc_ready file. If a .forc_read file is not present, searching could wait until it is there (with a time limit). I believe i saw something similar in cargo somewhere but couldn't find it now :thinking:

I think the long-term solution here might be something along the lines of #919. *Edit: to clarify, we can leave that for a future PR!*

forc-pkg/src/pkg.rs
```diff
@@ -1369,12 +1412,174 @@ pub fn fetch_git(fetch_id: u64, name: &str, pinned: &SourceGitPinned) -> Result<
...
+            }
+            Ok(())
+        },
+    )?;
+    Ok(found_local_repo)
+}
+
+/// Search local checkouts directory and apply the given function. This is used for iterating over
+/// possible options of a given package.
+fn with_search_checkouts<F>(checkouts_dir: PathBuf, package_name: &str, f: &mut F) -> Result<()>
```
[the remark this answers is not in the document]
kayagokalp: Hmm it seems like if I do this I am getting a compiler error at `f(index, repo_dir_path)` line complaining about not being able to capture it mutably. Once I remove the unnecessary `&mut` statements in front of the closures I pass to this function and change function signature to `fn with_search_checkouts<F>(checkouts_dir: PathBuf, package_name: &str, mut f: F) -> Result<()>` it works. I might be missing something but to get it compiling I had to go like the [doc example](https://doc.rust-lang.org/std/ops/trait.FnMut.html#using-a-fnmut-parameter)

JoshuaBatty (approved):

PR #2371, opened 2022-07-25
Provide intrinsic `__addr_of`.

### LINKED ISSUE #2281: Calling `std::mem::addr_of` on a copy type should not be a runtime panic
Example:
```rust
let number0 = 1u8;
let xyz = std::mem::addr_of(number0);
```
currently panics. It's better to emit a compile-time error instead given that we know the type of `number0` at compile-time.

vaivaswatha:
Closes #2281 (see the discussion there on design choices).


sway-core/src/type_engine/type_info.rs
```diff
@@ -619,6 +619,10 @@ impl TypeInfo {
         }
     }
 
+    pub fn is_copy_type(&self) -> bool {
+        matches!(self, TypeInfo::Boolean | TypeInfo::UnsignedInteger(_))
```
otrho, line 623: The version in `ResolvedType` also has `Byte` and `Unit`.  Not sure if you should call `is_unit()` here... probably, as AFAIK we don't actually use tuple/structs to represent it.

Are you actually calling this method though?  I see in other places you're using `ResolvedType::is_copy_type()`.  When it's declared `pub` like this the compiler won't tell you if it's never called.

sway-ir/src/instruction.rs
```diff
@@ -429,6 +433,15 @@ impl<'a> InstructionInserter<'a> {
         asm_val
     }
 
+    pub fn addrof(self, value: Value, span_md_idx: Option<MetadataIndex>) -> Value {
```
otrho, line 436: I think it reads a bit weirdly.  `addr_of` is better.
```suggestion
    pub fn addr_of(self, value: Value, span_md_idx: Option<MetadataIndex>) -> Value {
```

sway-ir/src/verify.rs
```diff
@@ -175,6 +176,16 @@ impl<'a> InstructionVerifier<'a> {
         Ok(())
     }
 
+    fn verify_addrof(&self, value: &Value) -> Result<(), IrError> {
```
otrho, line 179:
```suggestion
    fn verify_addr_of(&self, value: &Value) -> Result<(), IrError> {
```

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
```diff
@@ -333,6 +333,42 @@ impl TypedIntrinsicFunctionKind {
...
+                    .with_type_annotation(insert_type(TypeInfo::Unknown));
+                let exp = check!(
+                    TypedExpression::type_check(ctx, arguments[0].clone()),
+                    return err(warnings, errors),
+                    warnings,
+                    errors
+                );
+                let copy_ty = resolve_type(exp.return_type, &span).unwrap().is_copy_type();
+                if copy_ty {
+                    errors.push(CompileError::IntrinsicUnsupportedArgType {
```
otrho, line 356: I can see down in the test below that this just says 'Unsupported argument type to intrinsic "addr_of"' which would be better if it actually said 'Intrinsic "addr_of" cannot be passed a non-reference type."' or similar.

OTOH, we only expect this to be called by the library and library devs so verbosity isn't the highest priority.  Is there an easy way to add a hint to `IntrinsicUnsupportedArgType`, like an `Option<str>` arg or something?

test/src/e2e_vm_tests/test_programs/should_pass/language/addrof_intrinsic/src/main.sw
```diff
@@ -0,0 +1,21 @@
...
+        revert(0);
+    }
+    asm(ptr: val) {
+        ptr: u64
+    }
+}
+
+fn main() {
+    let sender = Identity::Address(B1);
+    assert (__addr_of(sender) == addr_of(sender));
```
otrho, line 20: This is great.  Maybe throw in some simple string, array, enum, tuple, etc. tests too?

otrho (commented):
Thanks Vaivas.  Just a few comments...

sway-core/src/type_engine/type_info.rs
vaivaswatha: > The version in `ResolvedType` also has `Byte` and `Unit`. Not sure if you should call `is_unit()` here... probably, as AFAIK we don't actually use tuple/structs to represent it.
> 
> Are you actually calling this method though? I see in other places you're using `ResolvedType::is_copy_type()`. When it's declared `pub` like this the compiler won't tell you if it's never called.

I'm using this in `intrinsic_function.rs`, i.e., during type checking, where I can't use `ResolvedType::is_copy_type()`. 

Let me check if I should call `is_unit()`.

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
vaivaswatha: Formatting the `Option` here could be tricky. Let me try doing it.

vaivaswatha:
Once I modified the existing library function `addr_of` in `mem.sw` to use the new intrinsic, I realized (due to a test failing) that we have the following function in `mem.sw`.

```
pub fn write<T>(ptr: u64, val: T) {
    if is_reference_type::<T>() {
        copy(addr_of(val), ptr, size_of_val(val));
    } else {
        asm(ptr: ptr, val: val) {
            sw ptr val i0;
        };
    }
}
```

and this can be called with both a reference type and a non-reference type. When called with a non-reference type, we would still (statically, not at runtime) have a call to `addr_of` with a non-reference argument, and that results in an error[1].

We have two options to go forward with
1. Drop the idea of having this intrinsic and live with a runtime check.
2. Have the intrinsic (because it's still better than having `ASM` blocks), but relax the static check, and instead, add a runtime check.

[1] The error, while it should've been in the newly added checks is not, because at the time of type checking, the `T` is instantiated. So the failure happens later in the verifier.

otrho:
This is unfortunate!  I vote for (2) with a relaxed verifier just to avoid the ASM blocks.

vaivaswatha:
Since we will, at some point, have trait bounds, and which means that we can have specialized implementations of `write` (see my previous comment), the implementation of `__addr_of` that we have today is right. It's just that, until we have those specialized implementations, we cannot use `__addr_of` to implement `std::addr_of`.

I suggest that we merge this PR as-is (except for the `std::addr_of` change which I'll revert now) - subject to review feedback, of course - and when we have trait bounds, we can re-implement `write` and *then* modify `std::addr_of` to use `__addr_of`.

@mohammadfawaz what do think? I've discussed this with @otrho offline too.

mohammadfawaz:
> Since we will, at some point, have trait bounds, and which means that we can have specialized implementations of `write` (see my previous comment), the implementation of `__addr_of` that we have today is right. It's just that, until we have those specialized implementations, we cannot use `__addr_of` to implement `std::addr_of`.
> 
> I suggest that we merge this PR as-is (except for the `std::addr_of` change which I'll revert now) - subject to review feedback, of course - and when we have trait bounds, we can re-implement `write` and _then_ modify `std::addr_of` to use `__addr_of`.
> 
> @mohammadfawaz what do think? I've discussed this with @otrho offline too.

I see so the plan to introduce the intrinsic but not use it just yet? Yeah I suppose that's reasonable.

vaivaswatha:
> > Since we will, at some point, have trait bounds, and which means that we can have specialized implementations of `write` (see my previous comment), the implementation of `__addr_of` that we have today is right. It's just that, until we have those specialized implementations, we cannot use `__addr_of` to implement `std::addr_of`.
> > I suggest that we merge this PR as-is (except for the `std::addr_of` change which I'll revert now) - subject to review feedback, of course - and when we have trait bounds, we can re-implement `write` and _then_ modify `std::addr_of` to use `__addr_of`.
> > @mohammadfawaz what do think? I've discussed this with @otrho offline too.
> 
> I see so the plan to introduce the intrinsic but not use it just yet? Yeah I suppose that's reasonable.

Yep.

sway-core/src/asm_generation/from_ir.rs
```diff
@@ -470,6 +470,7 @@ impl<'ir> AsmBuilder<'ir> {
         let mut errors = Vec::new();
         if let ValueDatum::Instruction(instruction) = &self.context.values[instr_val.0].value {
             match instruction {
+                Instruction::AddrOf(arg) => self.compile_addrof(instr_val, arg),
```
mohammadfawaz, line 473:
```suggestion
                Instruction::AddrOf(arg) => self.compile_addr_of(instr_val, arg),
```

sway-core/src/asm_generation/from_ir.rs
```diff
@@ -745,6 +746,11 @@ impl<'ir> AsmBuilder<'ir> {
         ok((), warnings, errors)
     }
 
+    fn compile_addrof(&mut self, instr_val: &Value, arg: &Value) {
```
mohammadfawaz, line 749:
```suggestion
    fn compile_addr_of(&mut self, instr_val: &Value, arg: &Value) {
```

sway-ir/src/error.rs
```diff
@@ -118,6 +120,14 @@ impl fmt::Display for IrError {
                 f,
                 "Verification failed: Argument specifier for function '{callee}' is not an argument value."
             ),
+            IrError::VerifyAddrOfUnknownSourceType => write!(
+                f,
+                "Verification failed: AddrOf unable to determine source type."
```
mohammadfawaz, line 125:
```suggestion
                "Verification failed: addr_of unable to determine source type."
```

sway-ir/src/error.rs
```diff
@@ -118,6 +120,14 @@ impl fmt::Display for IrError {
...
+            ),
+            IrError::VerifyAddrOfCopyType => write!(
+                f,
+                "Verification failed: AddrOf argument must be non-copy (memory) type."
```
mohammadfawaz, line 129:
```suggestion
                "Verification failed: addr_of argument must be non-copy (memory) type."
```

mohammadfawaz (commented):
Looks good! Some nits. Also, changes to `sway-ir/src/parser.rs` are required + maybe a short IR test in https://github.com/FuelLabs/sway/tree/master/sway-ir/tests/serialize

vaivaswatha:
All review comments addressed.

sway-core/src/ir_generation/function.rs
```diff
@@ -482,6 +482,11 @@ impl FnCompiler {
                         .add_metadatum(context, span_md_idx))
                 }
             }
+            Intrinsic::AddrOf => {
+                let exp = arguments[0].clone();
+                let value = self.compile_expression(context, md_mgr, exp)?;
+                Ok(self.current_block.ins(context).addr_of(value))
```
otrho, line 488:
```suggestion
                Ok(self.current_block.ins(context).addr_of(value).add_metadatum(context, span_md_idx))
```
Need to add the span when creating instructions.

sway-core/src/error.rs
```diff
@@ -187,6 +191,26 @@ where
...
+}
+
+impl Hint {
+    pub fn empty() -> Hint {
+        Hint { msg: None }
+    }
+    pub fn new(msg: String) -> Hint {
+        Hint { msg: Some(msg) }
+    }
+}
```
otrho, line 206: This doesn't seem very useful, but I assume you've made it just to implement `Display` so the `#[error ...]` attribute works below?  Feels overengineered but I can't think of a better solution, other than to get `Option` support working automatically in `#[error]` somehow.

sway-ir/tests/serialize/intrinsic_addr_of.ir
```diff
@@ -0,0 +1,26 @@
...
+        store v1, ptr v0, !2
+        // check: $(arg0=$VAR) = get_ptr
+        v2 = get_ptr ptr { u64, ( { b256 } | { b256 } ) } sender, ptr { u64, ( { b256 } | { b256 } ) }, 0, !5
+        // check: addr_of $arg0
+        v3 = addr_of v2
+        ret u64 v3
+    }
+}
+
+!0 = "/home/vaivaswatha/c/sway/test/src/e2e_vm_tests/test_programs/should_pass/language/addrof_intrinsic/src/main.sw"
```
otrho, line 21: You're leaking some personal directory stuff here.  Probably doesn't matter but it's a little weird to keep in the repo.

Since you're not actually testing metadata I'd just strip all the metadata out.  You could strip a lot of it out -- you're just testing that `addr_of` is parsed and re-printed correctly, so it only really needs a `get_ptr` and the `addr_of` instructions, none of the rest...

sway-core/src/error.rs
vaivaswatha: >I assume you've made it just to implement Display so the #[error ...] attribute works below?

Yes.

sway-ir/tests/serialize/intrinsic_addr_of.ir
vaivaswatha: > You're leaking some personal directory stuff here. Probably doesn't matter but it's a little weird to keep in the repo.
> 
> Since you're not actually testing metadata I'd just strip all the metadata out. You could strip a lot of it out -- you're just testing that `addr_of` is parsed and re-printed correctly, so it only really needs a `get_ptr` and the `addr_of` instructions, none of the rest...

Ah I overlooked this. Fixing it now.

sway-core/src/ir_generation/function.rs
```diff
@@ -485,7 +485,12 @@ impl FnCompiler {
             Intrinsic::AddrOf => {
                 let exp = arguments[0].clone();
                 let value = self.compile_expression(context, md_mgr, exp)?;
-                Ok(self.current_block.ins(context).addr_of(value))
+                let span_md_idx = md_mgr.span_to_md(context, &span);
```
otrho, line 488: Thanks.  Looks like this could be called once at the top and shared between variants.  But that can happen next time.

otrho (approved):

mohammadfawaz (approved):

Issue #2341, opened 2022-07-18
RWLock panic with small contract program (repro inside)


sezna:
the following `main.sw` file causes this panic:
```
thread 'main' panicked at 'rwlock read lock would result in deadlock', /rustc/ddcbba036aee08f0709f98a92a342a278eae5c05/library/std/src/sys/unix/locks/pthread_rwlock.rs:72:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

```rust
contract;
use std::contract::ContractId;

abi MyContract {
    fn test_function() -> bool;
}

impl MyContract for Contract {
    fn test_function() -> bool {
        true
    }
}

fn caller(address: ContractId) -> ContractCaller<_> {
  abi(MyContract, address.0)
}

```

sezna:
It is related to the return type. If the return type is a `ContractCaller` of any kind, then we get this panic.

PR #2378, opened 2022-07-25
Fix concurrency bug in ConcurrentSlab.replace.

tritao:
We get the following panic during `ConcurrentSlab.replace` execution:

```
thread 'main' panicked at 'rwlock read lock would result in deadlock',
/rustc/ddcbba036aee08f0709f98a92a342a278eae5c05/library/std/src/sys/unix/locks/pthread_rwlock.rs:72:13
```

This happens due to `operator !=`, we call `PartialEq` and eventually end up calling `ConcurrentSlab.get`, which fails due to existing read/write lock already on the stack.

The fix is to make sure to get a read lock only, do the comparison, and then get a read/write lock just for the write.

Closes https://github.com/FuelLabs/sway/issues/2341.

### DIFF
--- a/sway-core/src/concurrent_slab.rs
+++ b/sway-core/src/concurrent_slab.rs
@@ -23,11 +23,20 @@ where
     }
 
     pub fn replace(&self, index: TypeId, prev_value: &T, new_value: T) -> Option<T> {
-        let mut inner = self.inner.write().unwrap();
-        let actual_prev_value = &inner[*index];
-        if actual_prev_value != prev_value {
-            return Some(actual_prev_value.clone());
+        // The comparison below ends up calling functions in the slab, which
+        // can lead to deadlocks if we used a single read/write lock.
+        // So we split the operation: we do the read only operations with
+        // a single scoped read lock below, and only after the scope do
+        // we get a write lock for writing into the slab.
+        {
+            let inner = self.inner.read().unwrap();
+            let actual_prev_value = &inner[*index];
+            if actual_prev_value != prev_value {
+                return Some(actual_prev_value.clone());
+            }
         }
+
+        let mut inner = self.inner.write().unwrap();
         inner[*index] = new_value;
         None
     }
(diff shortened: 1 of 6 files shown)


sway-core/src/concurrent_slab.rs
```diff
@@ -23,11 +23,15 @@ where
...
-        let actual_prev_value = &inner[*index];
-        if actual_prev_value != prev_value {
-            return Some(actual_prev_value.clone());
+        {
+            let read_inner = self.inner.read().unwrap();
+            let actual_prev_value = &read_inner[*index];
+            if actual_prev_value != prev_value {
+                return Some(actual_prev_value.clone());
+            }
         }
```
mohammadfawaz, line 32: Can we just do this instead?
```suggestion
        let read_inner = self.inner.read().unwrap();
        let actual_prev_value = &read_inner[*index];
        if actual_prev_value != prev_value {
            return Some(actual_prev_value.clone());
        }
```

tritao: Doesn't look like it, just gave it a try and the test hangs.

otrho: Because it's waiting for a lock which never comes?  If so this is very fragile and either needs a different approach or lots and lots of comments.

sezna: The read lock needs to be dropped before a write lock can be given out. The extra scope specifies that the lifetime of the locks allocated here are dropped right after the if statement.  Having a `.read()` and a `.write()` in the same scope sequentially will always deadlock. Scoping the variable down like what @tritao is what I'm used to doing to fix that, because IMO it is prettier than using `drop(lock)`. But yeah some comments wouldn't hurt!

tritao: Yep exactly, the extra scope is used to limit the lifetime of the lock. I've added some comments to make it a bit clearer.

otrho (approved):

PR #2380, opened 2022-07-26
Return the local struct Span as a new field on `TypedExpressionVariant::StructExpression`

### LINKED ISSUE #2372: StructExpression Ident's differ between AstNode and TypedAstNode
Given the following small example: 

```
struct X{}

fn main() {
  let var = X{};
}
```

If we inspect the `Ident` of our `let` expression as an `AstNode` we get.

```
name: Ident {
    name_override_opt: None,
    span: Span {
        src (ptr): 0x000000012a64ea90,
        path: Some(
            "/Users/joshuabatty/Documents/rust/fuel/sway/sway-lsp/test_programs/array_basics/src/main.sw",
        ),
        start: 45,
        end: 46,
        as_str(): "X",
    },
    is_raw_ident: false,
},
```

And if we inspect this line as a `TypedAstNode` we can see the `Ident` is actually pointing to the struct declaration and the `Span` information of the token has been discarded. 

```
struct_name: Ident {
    name_override_opt: None,
    span: Span {
        src (ptr): 0x0000000121ed1f70,
        path: Some(
            "/Users/joshuabatty/Documents/rust/fuel/sway/sway-lsp/test_programs/array_basics/src/main.sw",
        ),
        start: 16,
        end: 17,
        as_str(): "X",
    },
    is_raw_ident: false,
},
```

We need to carry forward the original `Span` information during type checking as this is required for the language server.

JoshuaBatty:
This PR uses the local span of the call_path_binding to create a `Span` for the `TypedExpressionVariant::StructExpression`. Previously it was only passing along the `Ident` of the declaration which would prohibit the token from being collected in the language server. 

closes #2372

### DIFF
--- a/sway-core/src/control_flow_analysis/dead_code_analysis.rs
+++ b/sway-core/src/control_flow_analysis/dead_code_analysis.rs
@@ -821,6 +821,7 @@ fn connect_expression(
         StructExpression {
             struct_name,
             fields,
+            ..
         } => {
             let decl = match graph.namespace.find_struct_decl(struct_name.as_str()) {
                 Some(ix) => *ix,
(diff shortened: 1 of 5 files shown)


mohammadfawaz (approved):
Seems reasonable. Thanks!

otrho (approved):

PR #2382, opened 2022-07-26
Deprecate forc gm

### LINKED ISSUE #1472: Publish `forc-gm` once the crate name is available

### LINKED ISSUE #2381: Deprecate `forc-gm`
Since we seemed to have reached no resolution on the `forc-gm` crate squatting issue (and the act also isn't technically against the crates.io public policy), I think it's best to deprecate and remove `forc-gm` from the repository, since there is no need to maintain it anymore. We are also no longer using it as an example within the docs.

cc @SilentCicero @adlerjohn @mitchmindtree

bingcicle:
closes #2381 
closes #1472


kayagokalp (approved):
LGTM :)

adlerjohn (approved):

PR #2389, opened 2022-07-27
Set the CallPath prefix of `FunctionApplication` to call_path_binding suffix

JoshuaBatty:
Previously, the `prefixes` field of `CallPath` was always set to `call_path_binding.inner.prefixes`. However, this always is an empty Vec. 

I confirmed this by putting in an `asert!(call_path_binding.inner.prefixes.len==0)` and running all the e2e tests. 

The motivation behind this change is to collect the `Ident` of the Struct name in the language server. Consider the following line:
```
let p = ~MyStruct::new();
```

the `prefixes` field now = `MyStruct`
and the `suffix` field = `new`


sezna (approved):
good catch!

vaivaswatha (approved):

PR #2390, opened 2022-07-27
Adds `LineStyle` to `Shape`

### LINKED ISSUE #2299: Add `multiline` field to `Shape`
This will store whether formatting should be in-line or multi-line

eureka-cpu:
This PR cleans up `Shape` and `Indent` in general, but mainly it introduces a field `line_style` to `Shape` which is an `enum` `LineStyle` and resolves to either `Inline` or `Multiline`. I've added functions to make navigating this easier, tests to ensure they work as intended and updated their corresponding use cases. Overall this will boost readability and be useful for keeping track of the formatting style of structures/statements that may get formatted to either single or multiline, in the future. E.g. a function with lots of parameters, or a `let` statement with a long expression.

Closes #2299


kayagokalp (approved):
lgtm

sway-fmt-v2/src/fmt.rs
```diff
@@ -359,27 +359,6 @@ storage {
         assert_eq!(correct_sway_code, formatted_sway_code)
     }
     #[test]
-    fn test_storage_single_line() {
```
sezna, line 362: why was this test removed?

eureka-cpu: I was under the impression we didn't want single line formatting for `storage`, but if that's not the case I can absolutely replace it

sezna (approved):

kayagokalp (approved):

PR #2394, opened 2022-07-27
Generalise IR metadata.

### LINKED ISSUE #1906: Refactor to generalise IR metadata.
### Problem

At the moment we have only a couple of types of metadata in IR, spans and storage indices.  I'm adding storage attributes now too.  The spans are actually two different types: a path type and a segment type.

So each of these types of metadatum are hard coded in the IR library.  If we want to add more then we need to add new data structures and types to the IR library.  This is not really general enough, especially since the IR usually doesn't care what is in the metadata.

### Fix

I propose that we refactor the metadata to use a general structured type which can reasonably represent anything in metadata but without needing specifics in the IR library.  A bit like JSON or s-exprs.

Each metadatum could have an identifying tag or key, and then tuples or nested tuples of strings and ints.  A special metadata index would also be a usable type, to reference other metadata.

e.g., a Span would be tagged with `span`, and have a `(idx, int, int)` tuple, for path index, start and end.  When serialised it would look similar to what we have now: `!11 = span (!0 10 20)`.

otrho:
Closes #1906.


mohammadfawaz (approved):

tritao (approved):
LGTM, and the new code looks cleaner than before :+1: 

The only exception to this is the new `md_mgr` being passed around everywhere.

Would it make sense to put it inside `Context`? 

 You commented in my other PR how you wanted to get rid of some of the existing fields in `Context` , but then we will have to pass even more stuff around as parameters. A potential solution to that would be to introduce another type to represent the result of the final IR context, and either move the fields in `Context` in the end of the compilation or just have `Context` itself provide a field of that other type.

Anyway, approving this, and we can always think about that as a later cleanup.

otrho:
@tritao Yep, the main point of `md_mgr` was to separate it from the IR library and keep it in core -- it has all the core specific stuff, like knowing what a span is -- and the IR just knows generic opaque data.  So putting it in `Context` would break that abstraction.

But an alternative would be to put one of (or both) `context` and `md_mgr` into another 'builder' type struct and pass them around in `self`.  This is how `FnCompiler` does it when generating IR.  But I've run into problems with the borrow checker when putting mutable references into those builders, as it often produces errors about multuple mutable borrows.  So I didn't actually bother trying in this PR, but I think the borrow checker has improved lately when dealing with `self` and putting `context` and/or `md_mgr` into `self` where they're used might now be an option.

Issue #2392, opened 2022-07-27
Repeated function or method argument name does not cause a compile error


mohammadfawaz:
Example:
```rust
contract;

fn foo(x: u64, x: u64) {

}

abi MyContract {
    fn test_function(x: u64, x: u64) -> bool;
}

impl MyContract for Contract {
    fn test_function(x: u64, x: u64) -> bool {
        true
    }
}
```
I expect both `foo` and `test_function` not to compile but they do.

otrho:
Curious!  Have you tested what actually happens?  Is the second one shadowing the first one?

mohammadfawaz:
The behavior is quite weird actually. For example, the following returns 1:
```
script;

fn foo(x: u64, x: u64) -> u64 {
    x
}

fn main() -> u64 {
    foo(1, 2)
}
```
but when I try to change the type of the second `x` to `bool`, I get type checking errors. Same if I switch the types around. 

The code below actually gave me an ICE from the IR verifier:
```rust
script;

fn foo(x: bool, x: u64) -> u64 {
    x
}

fn main() -> u64 {
    foo(true, 2)
}
```

PR #2395, opened 2022-07-27
Check for function parameters with an already used identifier.

tritao:
This adds a new semantic check to make sure function parameter identifiers are unique.

If an identifier is re-used, then it emits a new error:

```
3 | fn foo(x: u32, x:u32) {
  |                ^ identifier "x" bound more than once in this parameter list
```

Closes https://github.com/FuelLabs/sway/issues/2392.


test/src/e2e_vm_tests/test_programs/should_pass/test_contracts/abi_with_generic_types/src/main.sw
```diff
@@ -23,7 +23,7 @@ struct MyArrayStruct<V, W> {
 abi MyContract {
     fn complex_function(arg1: MyStruct<[b256;
     3], u8>, arg2: [MyStruct<u64, bool>;
-    4], arg3: (str[5], bool), arg3: MyOtherStruct, ) -> str[6];
+    4], arg3: (str[5], bool), arg4: MyOtherStruct, ) -> str[6];
```
mohammadfawaz, line 26: Nice 😄

mohammadfawaz (approved):
Great! Thanks for the quick fix!

otrho (approved):

mohammadfawaz:
Will merge this so that it's in `0.19.0`

PR #2405, opened 2022-07-28
Add minimum Forc version to manifest

### LINKED ISSUE #2188: Add minimum Forc version field to manifest
# Proposal

Add a field, e.g. `project.forc_version` that defines the minimum version of Forc supported. Example:

```toml
[project]
# ...
forc_version = "0.16.2"
```

Cargo ref for similar field: https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

If the current version of `Forc` is too low, it should error out and tell the user how to update by linking to the Sway Book's page on installing and updating.

# Motivation

1. A common source of user issues is simply an outdated `forc`. This will eliminate tech support and provide actual information to the end user.
2. Will also avoid using a library that assumes certain compiler features (e.g. safety checks) from being used in an older codebase.

# Misc

cc @mitchmindtree and @kayagokalp for thoughts. Should this also apply transitively?

kayagokalp:
closes #2188.

## TODO

- [x] Documentation


forc-pkg/src/pkg.rs
```diff
@@ -379,6 +386,31 @@ fn find_proj_node(graph: &Graph, proj_name: &str) -> Result<NodeIx> {
...
 
+/// Check minimum forc version given in the manifest file and the current toolchain version
+///
+/// If required minimum forc version is higher than current forc version return an error with
+/// upgrade instructions
+fn validate_version(proj_manifest: &ManifestFile) -> Result<()> {
+    match &proj_manifest.project.forc_version {
+        Some(min_forc_version) => {
+            // Get the current version of the toolchain
+            let crate_version = env!("CARGO_PKG_VERSION");
```
sezna, line 397: This only works when you're building via `cargo` -- when you install a binary and use it elsewhere, this env var will no longer be available unfortunately.

In the past I have had to get the version at compile time. What I did then was I parsed the `Cargo.toml` file using `toml-rs` inside of `build.rs` and output just the version as a text file. In my code, I then did `include_str!()` to get that file, which was just a string representation of the version, at compile time. I think it's the cleanest solution. I'm happy to help you write it out, as I've done it before.

sezna (changes requested):

forc-pkg/src/pkg.rs
kayagokalp: > This only works when you're building via cargo -- when you install a binary and use it elsewhere, this env var will no longer be available unfortunately.

Oh, that is a bummer, thanks for the review. I will try to get it like your suggestion and can ping you if I get stuck in the process.

forc-pkg/src/pkg.rs
```diff
@@ -379,6 +386,31 @@ fn find_proj_node(graph: &Graph, proj_name: &str) -> Result<NodeIx> {
...
 
+/// Check minimum forc version given in the manifest file and the current toolchain version
+///
+/// If required minimum forc version is higher than current forc version return an error with
+/// upgrade instructions
+fn validate_version(proj_manifest: &ManifestFile) -> Result<()> {
+    match &proj_manifest.project.forc_version {
+        Some(min_forc_version) => {
+            // Get the current version of the toolchain
+            let crate_version = include_str!("../.version");
```
sezna, line 397: looks like you got it! this is exactly what I meant. 😄

sezna (dismissed):
approving to dismiss my requested change

forc-pkg/src/pkg.rs
mitchmindtree: I think we probably do want to use this original approach.

@sezna keep in mind that [the `env!` macro](https://doc.rust-lang.org/std/macro.env.html) is evaluated at *compile time*, so it won't matter whether or not the env var exists on the system after installation etc.

Using this env var is also the same approach that `clap` takes with [its `crate_version!` macro](https://docs.rs/clap/latest/clap/macro.crate_version.html).

mitchmindtree (changes requested):
https://github.com/FuelLabs/sway/pull/2405#discussion_r935035875

forc-pkg/src/pkg.rs
sezna: Ohhh. I thought that while `env!` is expanded at compile time, all it is doing is getting the runtime environment variable. I stand corrected, and I'm sorry for the churn that I have caused. 😓

JoshuaBatty (approved):

sezna (dismissed):

forc-pkg/src/pkg.rs
```diff
@@ -454,6 +485,20 @@ fn validate_dep_manifest(dep: &Pinned, dep_manifest: &ManifestFile) -> Result<()
...
+        if current_forc_version < *dep_forc_version {
+            bail!(
+                "{:?} requires forc version {} but current forc version is {}\nUpdate the toolchain by following: https://fuellabs.github.io/sway/v{}/introduction/installation.html",
+                dep.name,
+                dep_forc_version,
+                current_forc_version,
+                crate_version
+            );
+        }
+    }
```
mitchmindtree, line 501: It looks like this whole `if let` block could be replaced with your new function? I.e.
```rust
    validate_version(dep_manifest)?;
```

mitchmindtree (changes requested):

forc-pkg/src/pkg.rs
kayagokalp: Ahh yeah my bad. Thanks for pointing it!

sezna (approved):

mitchmindtree (approved):
:ok_hand: nice!

PR #2406, opened 2022-07-28
Forc init appends to .gitignore rather than overriding it

### LINKED ISSUE #2402: Running `forc init` in a git checkout clobbers gitignore
Steps:

1. Clone a git repo, or init a new git repo, with a `.gitignore`
2. Run `forc init`
3. ???
4. Clobbered!

`forc init` should instead append the Forc ignores to an existing `.gitignore`.

kayagokalp:
closes #2402.

This looked like an easy fix so I wanted to push a quick fix. Before this PR the `.gitignore` was written with [`fs::write`](https://doc.rust-lang.org/std/fs/fn.write.html) which, according to the doc, overrides the content if the file already exists. We should be appending our stuff to the existing `.gitignore`


kayagokalp:
Had a little bit of a problem with git 😅 but should be good to go now!

ra0x3 (approved):
utAck

eureka-cpu (approved):

PR #2409, opened 2022-07-29
Collect more tokens in the language server and store definitions

### LINKED ISSUE #2274: Handle collecting `TypeInfo::Custom` tokens during AST traversal
Before PR #2121 we were collecting tokens that were of type `TypeInfo::Custom` in the language server using the function below. It's not entirely clear to me at this stage how to handle collecting these tokens using the new `TokenMap`. Leaving this here for future reference when we decide how to handle this properly. 

```
fn handle_custom_type(_type_info: &TypeInfo, _tokens: &mut TokenMap) {
    if let TypeInfo::Custom { name, .. } = type_info {
        //Iterate through the tokens and find the first token that has the same name as the custom type.
        //Extract the token type of the found token, this should help determine if the custom type
        //is a struct or an enum.
        let found_token = tokens.iter().find(|token| token.name == name.as_str());
        if let Some(token_type) = found_token.map(|token| &token.token_type) {
            if let TokenType::StructDeclaration(_) = token_type {
                let token = Token::from_ident(name, TokenType::Struct);
                tokens.push(token);
            } else if let TokenType::EnumDeclaration(_) = token_type {
                let token = Token::from_ident(name, TokenType::EnumApplication);
                tokens.push(token);
            }
        }
    }
}
```

JoshuaBatty:
Most of the changes in the PR result from renaming `TokenType` to `Token`. I did this as we are now storing the type definition of the token during the collection phase. This can either be a `TypeId` or the `Ident` of the declaration token. 

The rest is just collecting more tokens that I noticed were missing when working through the e2e programs. There are still more tokens that need collecting but happy to address them in separate PR's.

closes #2274


kayagokalp (approved):
LGTM 🚀

sezna (approved):

PR #2416, opened 2022-07-29
Refactor `sway-fmt-v2`: remove and implement old todos & add where comment handling & code clean-up

### LINKED ISSUE #2404: Change dot notation to variable notation when destructuring in `sfv2`
Currently a lot of the code looks like: 
```rust
if let Some(foo: &(Bar, Baz)) = &self.foo_opt {
    write!(formatted_code, "{}{}", foo.0.span().as_str(), foo.1.span().as_str())?;
}
```
when it should look like:
```rust
if let Some((bar, baz)) = &self.foo_opt {
    write!(formatted_code, "{}{}", foo.span().as_str(), bar.span().as_str())?;
}
```

### LINKED ISSUE #2407: Refactor and use already implemented format functions for unblocked todos for `sway-fmt-v2`
There are some `todo`s in various places in `sway-fmt-v2` to be implemented later (because they were blocked at that time) like https://github.com/FuelLabs/sway/blob/22859f34f7a10c1ae766e1adc3ba8d9555631cfe/sway-fmt-v2/src/items/item_abi.rs#L54. Since we have most of the formatter implemented we can fix those. This issue is for tracking this kind of old todos.

### LINKED ISSUE #2415: Handle comments for where clauses in `sway-fmt-v2`
The initial implementation for the comment handler omitted `where` clauses. We need to support that too since it is possible to have comments next to/inside of a `where` clause.

kayagokalp:
## About this PR

closes #2404.
closes #2407.
closes #2415.

This PR implements comment handling for `where clauses` & changes dot notation to tuple matching & implements unblocked todo's around the codebase.

### DIFF
--- a/sway-fmt-v2/src/fmt.rs
+++ b/sway-fmt-v2/src/fmt.rs
@@ -606,7 +606,8 @@ fn hello_world(baz: /* this is a comment */ u64) { // This is a comment inside t
 abi StorageMapExample {
     // insert_into_map is blah blah
     #[storage(write)] // this is some other comment
-    fn insert_into_map(key: u64, value: u64); // this is the last comment inside the StorageMapExample
+    fn insert_into_map(key: u64, value: u64);
+    // this is the last comment inside the StorageMapExample
 }"#;
         let correct_sway_code = r#"contract;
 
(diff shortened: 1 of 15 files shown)


eureka-cpu (approved):
lgtm!

sezna (approved):
TBH I think we can be a lot meaner with the comment tests. But there's no end to the complexity you can have in comments, so this is probably a good start.

E.g.

```rust

struct /* a comment*/ Foo<T: /* another comment! */ Eq +
 /* another comment */ Ord {
  field_1: Result<
       /* a comment, believe it or not */ i32,
       /*wow*/ Error /* and another! */ >
}

PR #2417, opened 2022-07-29
Removed angle bracket trait from sway-fmt-v2

### LINKED ISSUE #2253: Remove `AngleBracket` trait
So far all instances of `AngleBracket` formatting have been the same. If there are no other instances by the time all `ItemKind` formatting is done, I think we can remove it and either add a public function to replace it or just add the brackets in-line.

gr00vytvniks:
This PR removes the angle bracket trait from the formatter due to the implementations of the trait being the same in each use case.

Closes #2253


eureka-cpu:
Looks like CI is failing because you need to run `cargo fmt`

sezna (approved):
Looks reasonable. Could also be accomplished with a [trait default implementation](https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations) if you'd like to keep the trait.

kayagokalp (approved):
lgtm 👍

Issue #2119, opened 2022-06-25
Use `break` in the `U128` library as per the comment


mohammadfawaz:
https://github.com/FuelLabs/sway/pull/2112 introduces the `break` keyword. It should be possible after that to use `break` in the `U128` library where needed: https://github.com/FuelLabs/sway/blob/8ae2fcafa93ab57c66a04a065535900ffe769c50/sway-lib-std/src/u128.sw#L262

@nfurfaro I will let you decide how you want to rewrite and clean that up.

mohammadfawaz:
Blocked by https://github.com/FuelLabs/sway/pull/2112

PR #2431, opened 2022-07-31
General code quality enhancements to the standard library

mohammadfawaz:
* Some formatting issues
* Some TODOs
* Some general code refactoring
* Removal of some `*` imports
etc.

Also closes https://github.com/FuelLabs/sway/issues/2119

### DIFF
--- a/sway-lib-std/src/storage.sw
+++ b/sway-lib-std/src/storage.sw
@@ -6,7 +6,6 @@ use ::hash::sha256;
 use ::option::Option;
 use ::result::Result;
 
-
 /// Store a stack variable in storage.
 #[storage(write)]pub fn store<T>(key: b256, value: T) {
     if !__is_reference_type::<T>() {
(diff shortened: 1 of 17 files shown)


vaivaswatha (approved):

otrho (approved):

sway-lib-std/src/storage.sw
```diff
@@ -134,14 +133,13 @@ impl<V> StorageVec<V> {
     }
 
     /// Removes the last element of the vector and returns it, None if empty
-    #[storage(read, write)]
-    pub fn pop(self) -> Option<V> {
+    #[storage(read, write)]pub fn pop(self) -> Option<V> {
```
tritao, line 136: Most of the languages I have used in the past prefer to keep attributes as a separate line, which at least to me looks a bit more readable. I guess this is fine too, just wondering if this was automatically or manually formatted and if this is way we prefer to format them so I can be consistent in the future.

mohammadfawaz: Yup, this is very ugly but that's what the current formatter does. Luckily, the new formatter will not have this issue. So yeah, this happened because I ran the formatter on all of the standard library which generally should be done with every change and should probably be checked in CI (I think there's an issue about that somewhere)

Issue #2397, opened 2022-07-27
Some usability issues around `abi` and `Contract`


mohammadfawaz:
## Tracking sub-issues mentioned below

- [x] Self `impl` instead of `impl` ABI for `Contract` (fixed in #3021)
- [x] Mismatch in return types (fixed in #3022)
- [x] Repeated ABI declarations (#3661)
- [x] Multiple definitions of ABI methods (fixed in #3031)
- [x] `self` inside ABI function signature (#3662)

## Detailed description

Below are some examples that should not compile but do. I bundled those together in the same issue because they are somewhat related and there may be others. Whoever takes this on should also think about other corner cases that are not covered below.

### Self `impl` instead of `impl` ABI for `Contract`
```rust
contract;

abi MyContract {
    fn foo();
}

impl Contract { // self impl for a `Contract`?
    fn foo() {
    }
}
```

### Mismatch in return types
Interestingly this is only a problem for integer types. I still don't think this should compile (Similar playground [link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3ad4f1d1c04611ef4920b9a2c364a178))
```rust
contract;

abi MyContract {
    fn foo() -> u32;
}

impl MyContract for Contract {
    fn foo() -> u64 {
        0 
    }
}
```
Traits in Sway have the same problem btw. The following should not compile but does:
```rust
trait T {
    fn foo() -> u64;
}

struct S {
    x: u64
}

impl T for S {
    fn foo() -> u32 {
        0
    }
}
```
This also seems to be a problem in this situation which compiles but shouldn't:
```rust
contract;

abi MyContract {
    fn foo()-> u64;
}

impl Contract {
    fn foo() { // No return type here
    }
}
```

### Repeated `abi` declarations:
The following compiles but I don't expect it to:
```rust
contract;

abi MyContract {
    fn foo() -> u64;
}

abi MyContract {
    fn bar() -> u64;
}
```

### Multiple definitions of `abi` methods
Example:
```rust
contract;

abi MyContract {
    fn foo() -> u64;
}

impl MyContract for Contract {
    fn foo() -> u64 {
        0
    }
}

impl MyContract for Contract {
    fn foo() -> u64 {
        0
    }
}
```
Multiple `impl` blocks for the same type or trait should be allowed but they should not allow redefinition of methods. This applies for ABIs, regular traits, and types.

### `self` inside ABI function signature
ABI function should not accept `self` as the first argument.
```rust
contract;

[...]

PR #2432, opened 2022-07-31
Additional checks for unnecessary visibility qualifier (ABI definitions, ABI impls and trait impls)

### LINKED ISSUE #2419: Additional checks for unnecessary visibility qualifier
Similarly to https://github.com/FuelLabs/sway/pull/2276, we should
1. Check for unnecessary visibility qualifier in ABIs:
```rust
abi MyContract {
    pub fn test_function() -> bool;
}
```
2. Check for unnecessary visibility qualifier in ABI `impl` blocks:
```rust
    pub fn test_function() -> bool {
        true
    }
}
```
3. Check for unnecessary visibility qualifier in trait `impl` blocks:
```rust
trait A {
    fn foo();
}

struct S {}

impl A for S {
     pub fn foo() {}
}
```

kayagokalp:
closes #2419.

With this PR, for the following cases, we are now emitting an error, similar to #2276.
```rust
abi MyContract {
    pub fn test_function() -> bool;
}
```
```rust
    pub fn test_function() -> bool {
        true
    }
}
```
```rust
trait A {
    fn foo();
}

struct S {}

impl A for S {
     pub fn foo() {}
}
```

### DIFF
--- a/test/src/e2e_vm_tests/test_programs/should_fail/abi_pub_fn_interface/src/main.sw
+++ b/test/src/e2e_vm_tests/test_programs/should_fail/abi_pub_fn_interface/src/main.sw
@@ -1,0 +1,5 @@
+contract;
+
+abi MyContract {
+    pub fn f(self) -> bool;
+}
--- a/test/src/e2e_vm_tests/test_programs/should_fail/abi_pub_fn_interface/src/main.sw
+++ b/test/src/e2e_vm_tests/test_programs/should_fail/abi_pub_fn_interface/src/main.sw
@@ -1,5 +1,5 @@
 contract;
 
 abi MyContract {
-    pub fn f(self) -> bool;
+    pub fn f() -> bool;
 }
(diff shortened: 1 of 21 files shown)


tritao (dismissed):

test/src/e2e_vm_tests/test_programs/should_fail/abi_pub_fn_interface/src/main.sw
```diff
@@ -0,0 +1,5 @@
+contract;
+
+abi MyContract {
+    pub fn f(self) -> bool;
```
mohammadfawaz, line 4: Actually I would avoid `self` here and in the next example. Don't think this should be allowed. Do you mind adding to this case to this issue: https://github.com/FuelLabs/sway/issues/2397? Feel free to modify the description :)

kayagokalp: Oh, that makes sense, thanks for pointing it 👍

tritao (approved):

mohammadfawaz (approved):

PR #2434, opened 2022-08-01
Enhancements to the `Contracts` section in the Sway Book

### LINKED ISSUE #1805: Revamp the "Contracts" section in the Sway Book
Link: https://fuellabs.github.io/sway/latest/sway-program-types/smart_contracts.html

The section is mostly fine but the examples have some mistakes and they can probably be replaced with actual compilable examples.

mohammadfawaz:
closes #1805

### DIFF
--- a/docs/src/sway-program-types/smart_contracts.md
+++ b/docs/src/sway-program-types/smart_contracts.md
@@ -7,12 +7,7 @@ A smart contract is no different than a script or predicate in that it is a piec
 As with any Sway program, the program starts with a declaration of what [program type](./index.md) it is. A contract must also either define or import an [ABI declaration](#the-abi-declaration) and implement it. It is considered good practice to define your ABI in a separate library and import it into your contract. This allows callers of your contract to simply import the ABI directly and use it in their scripts to call your contract. Let's take a look at an ABI declaration in a library:
 
 ```sway
-library wallet_abi;
-
-abi Wallet {
-    fn receive_funds();
-    fn send_funds(amount_to_send: u64, recipient_address: b256);
-}
+{{#include ../../../examples/wallet_abi/src/main.sw:abi_library}}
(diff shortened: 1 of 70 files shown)


docs/src/sway-program-types/smart_contracts.md
```diff
@@ -81,29 +65,22 @@ Note that the above implementation of the ABI follows the [Checks, Effects, Inte
...
 ```
 
 The main new concept is the _abi cast_: `abi(AbiName, contract_address)`. This returns a `ContractCaller` type which can be used to call contracts. The methods of the ABI become the methods available on this contract caller: `send_funds` and `receive_funds`. We then directly call the contract ABI method as if it was just a regular method. You also have the option of specifying the following special parameters inside curly braces right before the main list of parameters:
 
-1. `gas`: represents the gas being forwarded to the contract when it is called.
-2. `coins`: represents how many coins are being forwarded with this call.
-3. `asset_id`: represents the ID of the _asset type_ of the coins being forwarded.
+1. `gas`: a `u64` that represents the gas being forwarded to the contract when it is called.
+2. `coins`: a `u64` that represents how many coins are being forwarded with this call.
+3. `asset_id`: a `b256` that represents the ID of the _asset type_ of the coins being forwarded.
```
otrho, line 80: I always feel that 'which' is better than 'that' in cases like this, but I'm not sure if one or the other is more correct or whether it matters, really.
```suggestion
1. `gas`: a `u64` which represents the gas being forwarded to the contract when it is called.
2. `coins`: a `u64` which represents how many coins are being forwarded with this call.
3. `asset_id`: a `b256` which represents the ID of the _asset type_ of the coins being forwarded.
```

otrho (approved):

sezna (approved):

PR #2436, opened 2022-08-01
Add a new 'simplify CFG` optimisation pass.

### LINKED ISSUE #2398: Write initial 'simplify CFG' pass.
Simplify the control flow graph.

* Remove unused blocks.
* Merge blocks with only a single predecessor.
* Remove PHI instructions with a single entry.

These sorts of redundant blocks tend to be created by inlining and make further analysis like const folding more difficult.

otrho:
- Remove dead blocks.
- Merge twin blocks which only jump from one to the other.

Also updated the `opt` binary to use a more structured pass manager.

Closes #2398.

> Note: this pass isn't called by the compiler yet as it will take hours to complete on very large files like the `stdlib/vec` E2E test.

### DIFF
--- a/sway-ir/src/block.rs
+++ b/sway-ir/src/block.rs
@@ -138,7 +138,7 @@ impl Block {
     /// Get a reference to the block terminator.
     ///
     /// Returns `None` if block is empty.
-    pub fn get_term_inst<'a>(&self, context: &'a Context) -> Option<&'a Instruction> {
+    pub fn get_terminator<'a>(&self, context: &'a Context) -> Option<&'a Instruction> {
         context.blocks[self.0].instructions.last().and_then(|val| {
             // It's guaranteed to be an instruction value.
             if let ValueDatum::Instruction(term_inst) = &context.values[val.0].value {
@@ -160,7 +160,7 @@ impl Block {
 
     /// Return whether this block is already terminated specifically by a Ret instruction.
     pub fn is_terminated_by_ret(&self, context: &Context) -> bool {
-        self.get_term_inst(context)
+        self.get_terminator(context)
             .map_or(false, |i| matches!(i, Instruction::Ret { .. }))
     }
 
@@ -244,7 +244,7 @@ impl Block {
             //
             // Copying the candidate blocks and putting them in a vector to avoid borrowing context
             // as immutable and then mutable in the loop body.
-            for to_block in match new_block.get_term_inst(context) {
+            for to_block in match new_block.get_terminator(context) {
                 Some(Instruction::Branch(to_block)) => {
                     vec![*to_block]
                 }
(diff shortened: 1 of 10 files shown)


mitchmindtree (dismissed):
Not sure I'm 100% grokking the new pass, but the code's looking lurvely! Left a few nits but none are critical.

sway-ir/src/block.rs
```diff
@@ -273,26 +273,53 @@ impl Block {
...
+    ) -> impl Iterator<Item = Block> + 'a {
+        self.function.block_iter(context).flat_map(|block| {
+            block
+                .get_terminator(context)
+                .map(|term_inst| match term_inst {
+                    Instruction::ConditionalBranch {
+                        true_block,
+                        false_block,
+                        ..
+                    } => vec![*true_block, *false_block].into_iter(),
```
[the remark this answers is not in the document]
otrho: Ahh, thanks.  I wasn't happy with this code, and not as familiar with `flat_map` as I should be.  I was trying to use `Empty` and stuff, didn't think of `default`.

sway-ir/src/optimize/simplify_cfg.rs
```diff
@@ -0,0 +1,156 @@
...
+    let twin_blocks = function
+        .block_iter(context)
+        .filter_map(|from_block| {
+            // Filter all blocks with a Branch terminator.
+            block_terms_with_br(from_block)
+        })
+        .find(|pair_candidate| {
+            // Find branching blocks where they are singly paired.
+            are_uniquely_paired(pair_candidate)
+        });
```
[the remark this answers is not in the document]
otrho: Yep, nice.  Clippy often suggests these, maybe it didn't because of the comments... 🤔

sway-ir/tests/tests.rs
```diff
@@ -72,8 +78,23 @@ fn inline() {
...
+}
+
+// -------------------------------------------------------------------------------------------------
+
+#[allow(clippy::needless_collect)]
+#[test]
+fn simplify_cfg() {
+    run_tests("simplify_cfg", |ir: &mut sway_ir::Context| {
+        let fn_idcs: Vec<_> = ir.functions.iter().map(|func| func.0).collect();
+        fn_idcs.into_iter().fold(false, |acc, fn_idx| {
```
sezna, line 95: Do we need to collect here? cant we keep the iterator lazy?

sway-ir/src/optimize/simplify_cfg.rs
```diff
@@ -0,0 +1,150 @@
+use crate::{
+    block::Block, context::Context, error::IrError, function::Function, instruction::Instruction,
+    value::ValueDatum,
+};
+
```
sezna, line 5: could we get a high level description of how simplify cfg works at the top of this module?

sezna (approved):
approved but left nits

sezna:
oh that was a quick auto merge 😓 guess the nits can happen in a follow up haha

sway-ir/tests/tests.rs
otrho: Borrow checker says No.

PR #2440, opened 2022-08-01
Split client interaction commands into a forc-client plugin

### LINKED ISSUE #2412: Split client interaction commands into a `forc-client` plugin
The aim here is the separation of two major kinds of functionality currently provided by `forc`:

1. **Managing Sway code projects / hosting plugins** (`init`, `clean`, `check`, `build`, `update`, `plugins`). This is the role of the standalone `forc` binary.
2. **Interacting with a Fuel node as a client** (`deploy`, `run`). This would be the role of a `forc-client` plugin.

One of our primary upcoming priorities is integrating wallet support into the deployment process. Before we add this integration to `deploy` and `run` and further bloat the standalone `forc` tool, it makes sense to separate this client functionality into a dedicated crate.

By keeping the client interaction logic in a dedicated crate, we'll likely also greatly improve the build times and binary size of the standalone `forc` tool.

## To-do

- [ ] **Create a `forc-client` plugin crate** under the sway repo's `forc-plugins` directory.
- [ ] **Move the `forc deploy` and `forc run` commands** to the `forc-client` crate.
    - **Note:** this will be the first forc plugin that provides more than one new forc command. To achieve providing multiple commands under a single crate, the best approach might be to declare multiple `[[bin]]` targets under the `forc-client` manifest with a separate `main` for each command. Before diving into this, we should consider checking how the `cargo-edit` crate (which provides the `cargo add`, `cargo rm` and `cargo upgrade` commands via the installation of a single `cargo-edit` crate) achieves this.
- [ ] **Update E2E tests that use the deploy, run commands.**
- [ ] **Update Sway documentation.**
    - The forc chapter documentation should for the most part be an automatic update, but we may need to add a file stem for the new plugin. We should also review the rest of the Sway docs.
- [ ] **Ensure `forc-client` is published and shipped with `fuelup`**
    - This will include:
        1. Ensuring the new plugin is published under the release assets along with forc and the rest of the plugins and
        2. Ensuring `fuelup` is aware of the plugin and distributes it in each of the toolchains. I *believe* this should be automated for the most part, but it may be the case that we need to update a list of plugin names or similar somewhere in `fuelup` to make it aware.

---

cc @adlerjohn I know we briefly discussed splitting out `forc deploy` into a `forc-deploy` plugin in our roadmap chat, but it occurred to me that it might be useful to treat this as a general `forc-client` plugin that acts as the home of any commands that interact with a fuel node generally, I.e. any command that initalises a fuel client with `FuelClient::new(node_url)`. Does this sound OK?

kayagokalp:
closes #2412 

~blocked by #2478~


scripts/mdbook-forc-documenter/src/plugins.rs
```diff
@@ -28,10 +28,38 @@ pub fn plugin_commands() -> Vec<String> {
...
         }
     }
     plugins
 }
+/// Collects child plugins for a given plugin's Cargo.toml path
+fn collect_child_plugins(manifest_path: &PathBuf) -> Result<Vec<String>> {
+    let mut child_plugins = Vec::new();
+    let forc_toml: toml::Value = toml::de::from_str(&fs::read_to_string(manifest_path)?)?;
+    if let Some(table) = forc_toml.as_table() {
+        if let Some(values) = table.get("bin").and_then(|bin| bin.as_array()) {
```
sezna, line 52: code golf
```suggestion
    forc_toml.as_table().map(|table| {
        if let Some(values) = table.get("bin").and_then(|bin| bin.as_array()) {
```

forc-client-ops/src/deploy/cmd.rs
```diff
@@ -1,11 +1,9 @@
-use crate::ops::forc_deploy;
-use anyhow::{bail, Result};
 use clap::Parser;
 
-/// Deploy contract project.
-/// Crafts a contract deployment transaction then sends it to a running node.
 #[derive(Debug, Default, Parser)]
-pub struct Command {
+#[clap(version)]
+#[clap(bin_name = "forc deploy")]
```
sezna, line 5: does this mean that we maintain the command line interface of `forc run` and `forc deploy`? And we don't have to type `forc client deploy`, right?

sezna:
Is this still WIP? If so, could you convert it to a draft?

forc-client-ops/src/deploy/cmd.rs
kayagokalp: Yes, `forc deploy` is sufficient 👍🏻

kayagokalp:
Oh I forgot to change title, this is ready for review.

Cargo.toml
```diff
@@ -3,11 +3,13 @@ resolver = "2"
 members = [
     "docstrings",
     "forc",
+    "forc-client-ops",
```
mitchmindtree, line 6: Instead of adding an extra `forc-client-ops` crate, perhaps we can add a lib.rs and `[[lib]]` target to the `forc-client` crate itself? Then we can add an `ops` submodule if necessary.

scripts/mdbook-forc-documenter/src/plugins.rs
```diff
@@ -28,10 +28,38 @@ pub fn plugin_commands() -> Vec<String> {
...
+                // Check for parent plugin (like `forc-client`)
+                let child_plugins = collect_child_plugins(&path.join("Cargo.toml")).unwrap();
+                if child_plugins.is_empty() {
+                    let plugin = name.split_once('-').unwrap().1;
+                    plugins.push(plugin.to_string());
+                } else {
+                    for child_plugin in child_plugins {
+                        let plugin = child_plugin.split_once('-').unwrap().1;
+                        plugins.push(plugin.to_string());
+                    }
```
mitchmindtree, line 40: We've already established that `child_plugins` is empty before this branch. It looks like rather than:
 
```rust
                    for child_plugin in child_plugins {
                        let plugin = child_plugin.split_once('-').unwrap().1;
                        plugins.push(plugin.to_string());
                    }
```
this section should be
```rust
                    let plugin = name.split_once('-').unwrap().1;
                    plugins.push(plugin.to_string());
```

mitchmindtree (changes requested):
Looking good so far!

cc @bingcicle would be great to get your eyes on the mdbook changes in case we're missing anything else. Also, just a heads up that this adds two external commands (`forc deploy`, `forc run`) via the installation of a single plugin `forc-client` - just want to highlight this in case it might be relevant to any assumptions made in mdbook or fuelup related to a single plugin being associated with a single command?

forc-client-ops/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "forc-client-ops"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1"
+clap = { version = "3", features = ["derive"] }
+forc = {path = "../forc"}
```
mitchmindtree, line 9: Ahh I missed this. Can you share why we need the `forc` crate itself? Is it just for the git tag?

In retrospect, that constant might be better off living in the `forc-pkg` crate. This is because the git tag constant is solely used with `forc-pkg` functionality (i.e. the inclusion of implicit std). This change might be better off made in a dedicated PR though. This would also help both `sway-lsp` and the recent SDK/Sway testing as they would no longer require depending on `forc` itself, and would only require `forc-pkg` (and as a result, not need the entire `clap` dependency tree).

If this is the case, would you mind if we open a dedicated PR for this first?

forc-client-ops/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+forc-pkg = {path = "../forc-pkg"}
+forc-util = {path = "../forc-util"}
```
mitchmindtree, line 11: Could you make the formatting of these dependencies consistent with the rest of the file? e.g.

```rust
forc-util = { path = "../forc-util" }
```

Also, for dependencies within the same workspace, we need to specify the version of these dependencies in order for `cargo publish` to work, e.g.

```suggestion
forc-util = { version = "0.19.1", path = "../forc-util" }
```

mitchmindtree (changes requested):

forc-client-ops/Cargo.toml
mitchmindtree: Just a heads up, I'm going to have a go at moving this constant now as I think it's the right move either way!

kayagokalp: Hmm, we need `forc` because before running `deploy` or `run` we need to build. We are currently doing it by creating a `BuildCommand` (which lives in `forc`) from the parameters we pass to `deploy` or `run`. But I can probably refactor that place to only use `forc-pkg`. I will have a look at this shortly 👍🏻

scripts/mdbook-forc-documenter/src/plugins.rs
kayagokalp: I am not sure if I understood this change. If `child plugins` is empty we need to add `name.split_once('-').unwrap().1;` part to the plugins same like we did before this PR. If it is not empty, it means that we have a parent plugin that hosts multiple bins just like `forc-client`. So we need to add all child plugins to the list of plugins. I added the check for child plugins so that we can document `forc run` and `forc deploy`. Maybe @bingcicle might have some input here 👀

forc-client-ops/Cargo.toml
kayagokalp: I took a second look, and it seems like we can remove `forc` dependency but there is some logic around `build` in forc regarding `build_profile` and some other stuff. To remove that we may need to duplicate that logic on the `forc-client` side. Would that be acceptable?

scripts/mdbook-forc-documenter/src/plugins.rs
bingcicle: Semantically, I see a plugin as one installable component of the toolchain, so `forc-client` would be the only plugin here instead, with `run` and `deploy` as the available commands to the plugin - it just so happened that for the plugins so far, there was only 1 command available to them. I think I would document them as such since it might potentially be confusing if we documented them as separate plugins, but installed together in `forc-client`.

Perhaps documenting both `forc deploy` and `forc run` under `forc client` would make it clearer that `forc-client` is the installable component here:

```md
...
- 11.4. Plugins
  - 11.4.1 forc client
    - 11.4.1.1 forc deploy
    - 11.4.1.2 forc run
  - 11.4.2 forc explore
  - 11.4.3 forc fmt
  - 11.4.4 forc lsp
```

You might have to update the logic within the forc-documenter script to support sub-chapters though, since it currently naively assumes the commands to run are all unnested.

bingcicle (changes requested):
https://github.com/FuelLabs/sway/blob/2816c13698a35752136d4843dbfe5e1b95c26e10/.github/workflows/gh-pages.yml#L27-L30

The new plugin should also be installed within the gh-pages workflow, so that the CI knows to run the commands.

https://github.com/FuelLabs/sway/blob/2816c13698a35752136d4843dbfe5e1b95c26e10/.github/workflows/ci.yml#L587-L624

This should also be handled so that we package the new `forc-client` for release.

scripts/mdbook-forc-documenter/src/plugins.rs
kayagokalp: I made the required changes for supporting sub-chapters under plugins to have the documentation structure like the following.

```md
...
- 11.4. Plugins
  - 11.4.1 forc client
    - 11.4.1.1 forc deploy
    - 11.4.1.2 forc run
  - 11.4.2 forc explore
  - 11.4.3 forc fmt
  - 11.4.4 forc lsp
```
First of all, I want to discuss a little bit about the status of `forc-client`. `forc-client` is a plugin that does not produce any executable (unlike `forc-fmt` or any other `forc-*`) once installed. It packs 2 distinct bins, through `[[bins]]` table of its `Cargo.toml`. So we cannot collect any documentation from that, as the document collection phase is basically trying to do `forc-client --help`. So I think we should have the concept of `Virtual Plugins` in the script. If we do not have this kind of separate logic, adding `forc-client` to list of plugins will always result in warning stating that `forc client` has been removed from the toolchain, because there is no `forc-client`. And not adding it to the list of plugins, but having it in the `SUMMARY.md` will also fail. I strongly agree that the structural completeness of the Sway Book requires `forc client` as a chapter under `Plugins` which holds `forc deploy` and `forc run`. So I implemented the logic for supporting these kind of sub-chapters, currently, it supports only a single level of depth, meaning that we cannot have sub-sub-sub chapters 😄. I think if we want to support an arbitrary amount of nesting, we should introduce a graph-like structure to hold this information and introduce a larger refactor to the documenter. Depending on the input from @mitchmindtree and @bingcicle, I can open a separate issue for that as with the current state of the documenter we can create the structure above.

bingcicle (changes requested):
https://github.com/FuelLabs/sway/blob/2816c13698a35752136d4843dbfe5e1b95c26e10/.github/workflows/ci.yml#L611-L624

You also have to handle the job where we zip up the binaries for release

scripts/mdbook-forc-documenter/src/plugins.rs
bingcicle: Ahh right, makes sense. I think it'd be fair to handle it in a separate issue!

test/src/e2e_vm_tests/harness.rs
```diff
@@ -16,7 +20,7 @@ pub(crate) fn deploy_contract(file_name: &str, locked: bool) -> ContractId {
 
     tokio::runtime::Runtime::new()
         .unwrap()
-        .block_on(forc_deploy::deploy(DeployCommand {
+        .block_on(deploy(Deploy {
```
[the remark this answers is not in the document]
kayagokalp: Yep, my bad. I must have done this change subconsciously

bingcicle (changes requested):
Looking really good!

One last nit with regards to the docs - `forc deploy` and `forc run` pages are documented like before as intended, but the `forc client` page is empty. Perhaps it would look nicer to insert a short description of what it is like we have for the other plugins? eg. for `forc-explore`, the description is:

```
Forc plugin for running the Fuel Block Explorer.
```

Perhaps a short description like:

```
Forc plugin for interacting with a Fuel node.
```

would be nice.

I think manually inserting this description text might be fine, since new plugins wouldn't be introduced too frequently.

bingcicle (dismissed):
LGTM!

forc-client-ops/Cargo.toml
mitchmindtree: If we can use `forc-pkg` directly I think that would be best. If there is duplicated code, perhaps there's a chance to refactor some of it into `forc-pkg`?

Ideally we shouldn't need `forc` directly, as I think `forc` itself should just be a thin CLI wrapper on top of `forc-pkg` + plugin support.

kayagokalp: That makes perfect sense. I opened up #2478 for that refactoring and marking this PR as draft as this one is blocked until we land that one 👍

forc-plugins/forc-client/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "forc-client"
+version = "0.19.2"
+edition = "2021"
+
+[dependencies]
+anyhow = "1"
+clap = { version = "3", features = ["derive", "env"] }
+forc-pkg = { path = "../../forc-pkg" }
```
mitchmindtree, line 9: This one's still missing a version, e.g.
```suggestion
forc-pkg = { version = "0.19.2", path = "../../forc-pkg" }
```

forc-plugins/forc-client/src/ops/run/cmd.rs
```diff
@@ -1,11 +1,10 @@
-use crate::ops::forc_run;
-use anyhow::{bail, Result};
 use clap::Parser;
 
 /// Run script project.
 /// Crafts a script transaction then sends it to a running node.
 #[derive(Debug, Default, Parser)]
-pub struct Command {
+#[clap(bin_name = "forc run")]
```
mitchmindtree, line 6: Perhaps we want `version` here too like we have for the deploy command?
```suggestion
#[clap(version)]
#[clap(bin_name = "forc run")]
```
Possibly can also be written as:
```rust
#[clap(bin_name = "forc run", version)]
```
though I'm not sure

scripts/mdbook-forc-documenter/src/plugins.rs
mitchmindtree: > I think if we want to support an arbitrary amount of nesting, we should introduce a graph-like structure to hold this information and introduce a larger refactor to the documenter.

I don't think this should be necessary, I'm happy for us to draw the line at one level of nesting. I think all we're after is a way for a single plugin crate to support more than one command.

@kayagokalp I wonder if we could use something like a `PluginCommand` enum here to make it a little clearer on what's going on here for future maintainers? E.g.

```rust
enum PluginCommand {
    /// The plugin has a single command. Its name matches the plugin.
    Single,
    /// The plugin installs a group of commands with the given names.
    Group(Vec<String>),
}

/// The list of plugins alongside their associated commands.
pub fn plugin_commands() -> BTreeSet<String, PluginCommand> { ... }
```

mitchmindtree (changes requested):
Just a few more small things then we should be good to go!

scripts/mdbook-forc-documenter/src/plugins.rs
kayagokalp: I ended up using a `BTreeMap<String, PluginCommand>` as the return type of the `plugin_commands()` but other than that I created the `PluginCommand` enum and with it I think it is a lot more clearer now, thanks for the review 🙌.
 
Also returning a  `BTreeMap<String, PluginCommand>`, reduced the complexity of the doc addition phase from O(n^2) to O(n). I don't think this is significant as number of plugins is very limited but I enjoyed the fact that it became more readable and more time-efficient at the same time 😄

mitchmindtree:
I've opened a PR against this branch that removes some unused deps and attempts to improve the readability throughout mdbook-forc-documenter a little here https://github.com/FuelLabs/sway/pull/2520. If we can land that, I think this should be good to go!

JoshuaBatty (approved):
lgtm

bingcicle (approved):

mitchmindtree (approved):

Issue #1674, opened 2022-05-25
`sway-core` API needs to return a type-checked malformed AST


JoshuaBatty:
Currently in the language server, we call `compile_to_ast()` in order to get back the `TypedParseTree` for token collection. This works fine if the program was able to be successfully compiled, but returns any encountered errors otherwise. 

So for example, if the user starts typing a new line of code, then we are unable to collect any `TypedAstNode`s from the file, meaning all `TypeId`s and definitions are lost until the user line of code is properly formed leading to a successful compilation. 

In the compiler discussion this morning, @sezna mentioned that it should be possible for the `sway-core` API to return a malformed AST that contained not only any errors or warnings that the compiler detected but also any successfully type-checked `TypedAstNode`s. Going to mark this as critical as without this, the language server isn't able to perform how most users would expect it to during the editing process.

PR #2441, opened 2022-08-02
Implement a runnables custom method and collect typed tokens

JoshuaBatty:
This PR enables the language server to collect and use typed tokens. As a result, go to definition works for tokens that have an associated type definition (roughly 50% of tokens).

This also allows us to get the main_function for script and predicate files, allowing us to send the code position on the `main` fn back to the client. Code lens is implemented on the client side for now which allows us to have a play button show up above the `main` fn to execute `forc run`. 

Currently, we are only able to use the typed tokens, and send information about the `main` fn location if the program is able to successfully compile.  This should be fixed once #1674 is addressed.

### DIFF
--- a/sway-lsp/src/lib.rs
+++ b/sway-lsp/src/lib.rs
@@ -13,7 +13,7 @@ pub async fn start(config: DebugFlags) {
     let stdout = tokio::io::stdout();
 
     let (service, socket) = LspService::build(|client| Backend::new(client, config))
-        .custom_method("sway/runnable", Backend::runnable)
+        .custom_method("sway/runnables", Backend::runnables)
         .finish();
     Server::new(stdin, stdout, socket).serve(service).await;
 }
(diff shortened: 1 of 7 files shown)


sway-lsp/src/lib.rs
```diff
@@ -12,6 +12,8 @@ pub async fn start(config: DebugFlags) {
     let stdin = tokio::io::stdin();
     let stdout = tokio::io::stdout();
 
-    let (service, socket) = LspService::new(|client| Backend::new(client, config));
+    let (service, socket) = LspService::build(|client| Backend::new(client, config))
+        .custom_method("sway/runnables", Backend::runnables)
```
sdankel, line 16: In the future we should move these method name strings, param and response interfaces to a single file that has a corresponding file in the typescript client.

sdankel (approved):
🚢

sezna (approved):

JoshuaBatty:
closes #2319

PR #2444, opened 2022-08-02
Add an option to print the typed program AST.

tritao:
This adds a new `--print-ast` option which prints the typed program AST.

I've now needed this a few times for helping with debugging, so might as well add it.

You can use it via: `cargo run --bin forc -- build --print-ast --path test > test.ast`

And it will print something like:

```
TypedProgram {
    kind: Script {
        main_function: TypedFunctionDeclaration {
            name: main,
            body: TypedCodeBlock {
                contents: [
                    TypedAstNode {
                        content: Declaration(
                            ConstantDeclaration(
                                TypedConstantDeclaration {
                                    name: C2,
                                    value: TypedExpression {
                                        expression: VariableExpression {
                                            name: C1,
                                        },
                                        return_type: u64,
                                        is_constant: Yes,
                                        span: Span {
                                            as_str(): "C1",
                                        },
                                    },
                                    visibility: Private,
                                },
                            ),
                        ),
                        ... etc
```

Spans and Idents are a bit verbose by default, which can be improved by commenting a few lines in their Debug trait implementation. Not sure if we should change those defaults, so left it out of the PR.

### DIFF
--- a/docs/src/forc/manifest_reference.md
+++ b/docs/src/forc/manifest_reference.md
@@ -65,6 +65,7 @@ The `[build-profiles]` tables provide a way to customize compiler settings such 
 
 The following fields needs to be provided for a build-profile:
 
+* `print-ast` - Whether to print out the generated AST (true) or not (false).
 * `print-finalized-asm` - Whether to compile to bytecode (false) or to print out the generated ASM (true).
 * `print-intermediate-asm` - Whether to compile to bytecode (false) or to print out the generated ASM (true).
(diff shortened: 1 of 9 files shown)


sezna (approved):
Looks good enough to me, although it may conflict with #2440 

Eventually we should probably have an AST pretty-printer, since the default Debug print is not particularly readable.

ra0x3 (approved):
utACK :shipit:

PR #2445, opened 2022-08-02
Add math & bitwise ops to U256 type 2

rostyslavtyshko:
No new changes compared [to existing PR about math & bitwise ops to U256](https://github.com/FuelLabs/sway/pull/2283) created only due to rebaseing issues and already solved compiler issue


JoshuaBatty:
Hi @rostyslavtyshko I've spent today looking into why this test isn't passing. It's actually happening in the sway compiler and crashing with a stack over flow message. 

With the `u256` lib in the std-lib we are *just* sneaking in under the maximum stack size when compiling, but when the compiler is invoked from a test (such as in LSP tests) the stack is getting tipped over when some of the stack is also being used for the test. 

I think perhaps the parser needs to be optimized slightly before we can land this PR unfortunately.

adlerjohn (commented):
Is this PR meant to supersede #2283?

sway-lib-core/src/ops.sw
```diff
@@ -618,15 +618,15 @@ impl Shiftable for b256 {
 /////////////////////////////////////////////////
 
 /// Left shift a u64 and preserve the overflow amount if any
-fn lsh_with_carry(word: u64, shift_amount: u64) -> (u64, u64) {
+pub fn lsh_with_carry(word: u64, shift_amount: u64) -> (u64, u64) {
```
adlerjohn, line 621: `core` is meant to just contain ops, necessary so `std` can use operators. Everything else in `core` should be private. If you need access to these two helper functions, then copy them into `std`.

The only other thing so far that's acceptable is `core::num`, to imitate Rust's https://doc.rust-lang.org/src/core/num/mod.rs.html#849-853 (and even then we should consider moving that to `std` as per https://doc.rust-lang.org/std/primitive.u64.html#associatedconstant.MAX).

There was a suggestion to add a second library in between `core` and `std`, but I can't find it. cc @nfurfaro

sway-lib-std/src/u256.sw
```diff
@@ -89,4 +91,512 @@ impl U256 {
...
+            i -= 1;
+        }
+
+        quotient
+    }
+}
+
+impl core::ops::Ord for U256 {
+    fn gt(self, other: Self) -> bool {
+        self.a > other.a || self.a == other.a && self.b > other.b ||  self.a == other.a && self.b == other.b  && self.c > other.c ||  self.a == other.a && self.b == other.b  && self.c == other.c && self.d > other.d
```
adlerjohn, line 356: Can you put brackets on this one so it's formatted the same as `lt` and `eq`?

sway-lib-std/src/u256.sw
```diff
@@ -89,4 +91,512 @@ impl U256 {
...
+            w3 = shifted_1;
+        } else if w == 3 {
+            w4 = word_1 >> b;
+        };
+
+        ~U256::from(w1, w2, w3, w4)
+    }
+}
+
+impl core::ops::Add for U256 {
```
adlerjohn, line 475: Wait...why is there a duplicate here?

Braqzen (commented):
Spacing should be consistent. 
If the formatter doesn't spot this then could you raise an issue so that `let(a, b, c)` gets formatted to `let (a, b, c)`?

adlerjohn (approved):

Braqzen (commented):
I think the tests should also test `a` & `b` to prove that their values are correct.

sezna (approved):

PR #2446, opened 2022-08-02
Add an overall description to the `simplify-cfg` IR pass.

otrho:
Addressing nits in #2436 from Alex.

### DIFF
--- a/sway-ir/src/optimize/simplify_cfg.rs
+++ b/sway-ir/src/optimize/simplify_cfg.rs
@@ -1,3 +1,13 @@
+//! ## Simplify Control Flow Graph
+//!
+//! The optimizations here aim to reduce the complexity in control flow by removing basic blocks.
+//! This may be done by removing 'dead' blocks which are no longer called (or in other words, have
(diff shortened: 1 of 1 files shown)


sway-ir/src/optimize/simplify_cfg.rs
```diff
@@ -130,20 +140,9 @@ fn merge_blocks(context: &mut Context, function: &Function) -> Result<bool, IrEr
     // `from_block`.
     let succs = context.blocks[final_to_block.0]
         .successors(context)
-        .collect::<Vec<crate::block::Block>>();
+        .collect::<Vec<_>>();
```
otrho, line 143: There are quite a few places throughout `sway-ir` where this is going on and it looks like the `Vec` and/or `.collect()` could be omitted.  But we're often passing the `Context` around as a mutable reference and so the borrow checker isn't happy with e.g., iterating over `.successors()` and `.update_phi_source_block()`  in the single expression.

mitchmindtree (approved):

sezna (approved):

Issue #2423, opened 2022-07-30
Incorrect DCA warning for `const` initialized with another `const`


mohammadfawaz:
Example:
```rust
script;

const C1 = 66;

fn main() -> u64 {
    const C2 = C1;
    C2 
}
```
Diagnostic emitted:
```console
1 |
2 |
3 | const C1 = 66;
  | -------------- This declaration is never used.
4 |
5 | fn main() -> u64 {
  |
```

PR #2450, opened 2022-08-03
Refactor E2E test runner to run FileCheck on compile category tests.

tritao:
This refactors the existing code in the test runner and generalizes it so FileCheck-based checks are also done for tests in the compile category.

This will be used in an upcoming PR for https://github.com/FuelLabs/sway/issues/2423 to make sure that DCA warnings are not present in tests that should pass.

### DIFF
--- a/test/src/e2e_vm_tests/harness.rs
+++ b/test/src/e2e_vm_tests/harness.rs
@@ -95,52 +95,52 @@ pub(crate) fn runs_in_vm(file_name: &str, locked: bool) -> (ProgramState, Compil
     (*i.transact(tx_to_test).unwrap().state(), script)
 }
 
-/// Returns Err(()) if code _does_ compile, used for test cases where the source
-/// code should have been rejected by the compiler.  When it fails to compile the
-/// captured stdout is returned.
-pub(crate) fn does_not_compile(file_name: &str, locked: bool) -> Result<String, ()> {
-    use std::io::Read;
-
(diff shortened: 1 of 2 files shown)


test/src/e2e_vm_tests/harness.rs
```diff
@@ -95,60 +96,70 @@ pub(crate) fn runs_in_vm(file_name: &str, locked: bool) -> (ProgramState, Compil
...
+    // Capture the result of the compilation (i.e., any errors Forc produces) and append to
+    // the stdout from the compiler.
+    if let Err(ref e) = result {
+        write!(output, "\n{}", e).expect("error writing output");
     }
+
+    (result, output)
 }
 
 /// Returns `true` if a file compiled without any errors or warnings,
```
sezna, line 123: could we update this comment to also describe what the `String` return type contains?

test/src/e2e_vm_tests/mod.rs
```diff
@@ -233,6 +233,18 @@ fn build_file_checker(content: &str) -> Result<filecheck::Checker, String> {
     Ok(checker.finish())
 }
 
+fn check_file_checker(checker: filecheck::Checker, name: &String, output: &str) {
```
sezna, line 236: could we document this function, since its name is not particularly descriptive?

sezna (changes requested):

sezna (approved):

mohammadfawaz (approved):

PR #2453, opened 2022-08-03
Identify types suited only for storage and find incorrect decls

### LINKED ISSUE #1796: Improve error messaging for `StorageMap` when used illegally (or allow those cases?)
Here are some cases that we need to think about and either support or simply disallow with clear error messages:

### Case 1:

```rust
storage {
    map1: StorageMap<u64, u64>,
}

...

let local_map1 = storage.map1;
local_map1.insert(..);

```
This currently emits an internal compiler error that should be improved. 

### Case 2
```rust
let local_map1: StorageMap<u64, u64> = StorageMap {};
local_map1.insert(..);
```
This currently errors out with a type inference issue.

### Case 3
```rust
Struct Wrapper {
    map1: StorageMap<u64, u64>,
    map2: StorageMap<u64, u64>,
}

storage {
    w: Wrapper
}

...

storage.w.map1.insert(..);
```
This does compile but I _think_ it will cause collision between `map1` and `map2` because they have the same `storage` index (that of `w`), so we need additional information which is available but not yet propagated. But then again, we should think about whether we should even allow this.

### Case 4
Storage maps that contain other storage maps. Not sure what the semantics of this would be really. And this is not going to work with the way we do things (i.e. using an intrinsic that relies on the state index in the `storage` block). We need some sort of `storage` reference here.

### LINKED ISSUE #1947: Internal compiler error: State index for get_storage_key is not available as a metadata
# Reproduce

Passing in a mapping into a function, not just in a library, produces an ICE.

Related: #1809 

```rust
contract;

use std::storage::StorageMap;

storage {
    map: StorageMap<u64, u64>,
}

fn insert(mapping: StorageMap<u64, u64>) {
    mapping.insert(1, 1);
}

abi MyContract {
    fn test_function() -> bool;
}

    fn test_function() -> bool {
        insert(storage.map);
        true
    }
}
```

# Screenshot

cc @FuelLabs/sway-compiler

vaivaswatha:
This analysis works in two phases:

1. Identify types which are suited *only* to be declared as fields in the `storage` construct. A type is `storage-only` if any method in its `impl` uses `get_storage_index`. This analysis is performed in `impl_trait.rs`.
2. Traverse the AST and inspect all declarations. If a declaration involves a type that is marked `storage-only` (with the obvious exception that the declaration is directly in a `storage` clause), then it is flagged as an error.

While both the issues (that this PR is supposed to close) mention the case of passing storage decls to functions, it isn't clear at this point on how that should work. So those are flagged as errors too at the moment. We can revisit this analysis later once we implement the mechanics of making that work.

Closes #1796
Closes #1947


mohammadfawaz (approved):
Wow this turned out a much bigger change than I thought it was going to be but it all makes sense I think. I don't see a different way of doing this. This seems to cover the cases where a `StorageVec`/`StorageMap` is used as a struct/enum field, as a type for a local variable, or as a type parameter, which is fantastic. Thanks!

With the collection context work, I wonder if we will be able to do this in the collection phase and tag types there earlier, which would make the type checking phase much easier.

sway-core/src/semantic_analysis/storage_only_types.rs
```diff
@@ -0,0 +1,256 @@
...
+        | TypedDeclaration::Continue { .. } => {}
+    }
+    ok((), warnings, errors)
+}
+
+pub fn validate_decls_for_storage_only_types_in_ast(
+    ast_n: &TypedAstNodeContent,
+) -> CompileResult<()> {
+    ast_node_validate(ast_n)
+}
```
otrho, line 245: This is a bit redundant isn't it?  `ast_node_validate()` isn't `pub` so there's a slight differentiator but they might as well be one function?

otrho (approved):

sway-core/src/semantic_analysis/storage_only_types.rs
vaivaswatha: Yes, kind of. I just wanted to make the entry point explicit, since the private functions are mutually recursive.

PR #2459, opened 2022-08-04
Send the program kind along with the range to the client code_lens request

JoshuaBatty:
Currently, the LSP client has no information about the current file that is open. This changes the `runnables` callback from sending a `vec![lsp_types::Range]` to `vec![(lsp_types::Range, String)]`

The `String` can be either `"predicate"` or `"script"`

This should allow the client to match on the program kind string and call `forc run` for scripts and `forc build` for predicate files when the code_lens run button is clicked.


kayagokalp (approved):
lgtm 👍

bingcicle (approved):

PR #2460, opened 2022-08-04
Switch between viewing typed or parsed tokens as warnings in the LSP client.

JoshuaBatty:
Change `forc-lsp` to take an optional string for specifying which tokens should be shown as warnings when debugging the language server.

This makes it much more convenient to switch back and forth between inspecting the parsed and typed tokens that have been collected by the language server.

I've opened up a PR in the VSCode plugin which allows the user to specify this using a drop-down menu under the sway plugin options tab. https://github.com/FuelLabs/sway-vscode-plugin/pull/83


kayagokalp (approved):
utACK

mitchmindtree (approved):
:shipit:

PR #2461, opened 2022-08-04
Fix DCA for constant declaration value expressions.

tritao:
This connects the value expression of a constant declaration when performing dead code analysis.

Also adds a new test to make sure this does not regress.

Closes https://github.com/FuelLabs/sway/issues/2423.


vaivaswatha (approved):

mohammadfawaz (approved):
Looks reasonable! Thanks for the quick fix!

PR #2466, opened 2022-08-04
bump sdk version to v0.20 in stdlib tests and forc

nfurfaro:
All in the title.
I had to go through the tests and change instances of `StorageConfiguration` from `StorageConfiguration::with_storage_path` to `StorageConfiguration::default()` for tests using contracts _**with no storage**_.


Braqzen (approved):
I don't know what the difference is between the default and the previous version but this seems reasonable so I'll approve

mohammadfawaz:
Are we not able to bump to fuels 0.19? Sorry if this was asked before!

nfurfaro:
@mohammadfawaz Based on my attempt to bump to 19 yesterday, it looks like we need to wait for a new client release before bumping to 0.19. 
I will double-check that now before merging this though.

mohammadfawaz:
Oh excellent. I see you're able to go to 0.19 now. Two more things :)
1. Do you mind updating the title of the PR?
2. Do you mind also bumping the SDK version in https://github.com/FuelLabs/sway/blob/master/forc/src/utils/defaults.rs?

Thanks!

Braqzen (approved):
Could we also get an update to `forc` so when a new project is created it uses the latest version? :)

forc/src/utils/defaults.rs
```diff
@@ -100,12 +100,7 @@ abigen!(MyContract, "out/debug/"#,
 
 async fn get_contract_instance() -> (MyContract, ContractId) {
     // Launch a local network and deploy the contract
-    let mut wallets = launch_custom_provider_and_get_wallets(
-        WalletsConfig::new_single(Some(1), Some(1000000)),
-        None,
-    )
-    .await;
-    let wallet = wallets.pop().unwrap();
+    let wallet = launch_provider_and_get_wallet().await;
```
mohammadfawaz, line 103: Do we not need to pass in coins anymore? @digorithm recommended a few weeks ago to do that which is why this was changed to specify a custom provider.

nfurfaro: I see. There are 2 different ways to do this from what I understand. I believe this is the simpler of the 2, but has UTXO validation disable IIRC.
I can switch to using the other method if that's what was recently recommended.

mohammadfawaz:
> I had to go through the tests and change instances of `StorageConfiguration` from `StorageConfiguration::with_storage_path` to `StorageConfiguration::default()` for tests using contracts _**with no storage**_.

What is the reason for this change? Nothing should break even if storage is not used. The JSON file for storage slots should exist regardless of whether storage is used or not. Now whether the file has the correct path in those tests or not is different issue 😄 - I remember @SwayStar123 saying that the test for `StorageMap` has the wrong path.

nfurfaro:
> What is the reason for this change?

While trying to bump to SDK v0.18, I found that for the tests which **did not** actually use storage, they were failing when using `StorageConfiguration::with_storage_path`. Switching just these tests to use `StorageConfiguration::default()` had the tests passing again.
I'll admit I did not try changing the paths, but could test that now...

nfurfaro:
@mohammadfawaz I've reverted some changes to this and got all the tests green again.
1.) I went back to using a custom provider config for the forc test_proj setup as recommended.
2.) I fixed all of the broken storage paths so that they work as expected.

Braqzen (approved):

sezna (approved):

Issue #2425, opened 2022-07-30
Incorrect DCA warning for unused struct when it's not used in an expression


mohammadfawaz:
Example:
```rust
contract;

struct S {
    x: u64,
}

abi MyContract {
    fn get(s: S) -> S;
}

impl MyContract for Contract {
    fn get(s: S) -> S {
        s
    }
}
```
`S` is not used here in an expression but the type is technically used. The above is valid code and should not emit an warning. We currently get the following:
```console
2 |
3 | / struct S {
4 | |     x: u64,
5 | | }
  | |_- This struct is never instantiated.
6 |
7 |   abi MyContract {
  |
```

otrho:
And the return type too?  Do we link type declarations or only expressions?  I can't remember.

PR #2468, opened 2022-08-05
Fix DCA for used struct/enum function parameters.

tritao:
This fixes DCA of struct/enum types that are used as function parameters, but otherwise not instantiated in the module.

This is done by introducing a new graph edge between struct/enum parameters and their respective struct/enum declarations.

Closes https://github.com/FuelLabs/sway/issues/2425.

### DIFF
--- a/sway-parse/src/attribute.rs
+++ b/sway-parse/src/attribute.rs
@@ -7,15 +7,15 @@ use sway_types::Ident;
 
 impl<T: Parse> Parse for Annotated<T> {
     fn parse(parser: &mut Parser) -> ParseResult<Self> {
+        // Parse the attribute list.
         let mut attribute_list = Vec::new();
-        loop {
-            if parser.peek::<HashToken>().is_some() {
-                attribute_list.push(parser.parse()?);
-            } else {
-                break;
-            }
+        while let Some(attr) = parser.guarded_parse::<HashToken, _>()? {
+            attribute_list.push(attr);
         }
+
+        // Parse the `T` value.
         let value = parser.parse()?;
+
         Ok(Annotated {
             attribute_list,
             value,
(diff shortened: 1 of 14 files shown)


vaivaswatha (dismissed):

sway-core/src/control_flow_analysis/dead_code_analysis.rs
```diff
@@ -514,6 +514,35 @@ fn connect_impl_trait(
             tree_type,
         )?;
         methods_and_indexes.push((fn_decl.name.clone(), fn_decl_entry_node));
+        // if we are dealing with a contract abi impl, then we want to connect reference
+        // parameters to their respective declaration, since they can be passed externally
+        // and so do not need to be necessarily instantiated in the module.
+        if tree_type == &TreeType::Contract {
```
mohammadfawaz, line 520: Two questions:
1. Do we need to limit this to only contract methods? If we have other types of methods or even free functions that take a struct/enum or return a struct/enum, wouldn't we also want to make the graph connections here? If the function is not used in the current module (and it's private), then we will emit a warning for the whole function being unused anyways. If the function is `pub`, then we don't have visibility into whether it's used outside the module or not so we can't emit a warning.
2. If we really want to limit this to ABI `impl`, then we probably need a better way such as checking that the trait we're implementing is actually an `abi` not a regular trait. Checking that we're in a contract is not enough right? Because we could have other non-abi traits and their impls in the contract.

Also, if the struct is `pub` we can never emit a warning. I assume this is working already?

mohammadfawaz: Thanks! One last question: what happens if we have a free `pub` function that takes a struct and returns it but is not used in the library? Do we still emit a warning? Same question goes for trait methods implemented for a type in a library (recall that trait methods are `pub` by default if the type is `pub`).

mohammadfawaz: So `pub` in a `contract` is not really useful because we can't import a contract. But you can imagine the same code but in a `library`. So something like:
```rust
library foo;

struct S {}

pub fn free_fn(s: S) -> S {
    s
}
```
Or something like:
```rust
library;

struct S {}

pub struct F {}

impl F {
    fn free_fn(s: S) -> S {
        s
    }
}
```
or even 
```rust
library foo;

struct S {}

trait T {
    fn bar(t: T) -> T; // pub by default
}

impl T for S {
    fn bar(t: T) -> T {
        t
    }
}
```
In all the cases above, because the functions are effectively `pub`, then the functions themselves can't be dead. At the same time, we shouldn't emit warnings for unused `S`. Is that correct?

sway-core/src/control_flow_analysis/dead_code_analysis.rs
```diff
@@ -630,6 +634,36 @@ fn connect_typed_fn_decl(
     graph
         .namespace
         .insert_function(fn_decl.name.clone(), namespace_entry);
+
+    connect_fn_params_struct_enums(fn_decl, graph, entry_node)?;
+    Ok(())
+}
+
+fn connect_fn_params_struct_enums(
```
sezna, line 642: could we get a descriptive documentation comment here?

sezna (commented):
nice thorough tests.

sezna (approved):

mohammadfawaz (approved):
Great! LGTM!

PR #2470, opened 2022-08-07
Move `test` project to `clap`

AlicanC:
Our `test` project parsed its CLI args manually. This PR moves it to `clap` (which we already use in `forc` and `fuel-core`).


sezna (approved):
Worth noting that this changes the interface from `cargo run -- <filter_regex>` to `cargo run -- --filter-regex <filter_regex>`

otrho:
> Worth noting that this changes the interface from `cargo run -- <filter_regex>` to `cargo run -- --filter-regex <filter_regex>`

Umm, I'm not sure I'm a fan of that.  Most (all?) test harnesses I've used in the past, including `cargo test` will just take a filter as a string on the command line.

sezna:
Oh. I am wrong. These docs [suggest otherwise](https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html#positionals). I just checked with this PR and I'll eat my words. I thought `value_parser` did something else.

otrho (approved):
Ah, yep, cool.  I haven't used the `derive` method in `clap` before.  `value_parser` is for non-option args...

Issue #2421, opened 2022-07-30
`main()` should not accept any arguments just yet


mohammadfawaz:
This currently compiles without any diagnostics:
```rust
script;

fn main(x: u64) -> u64 {
    x
}
```
and the assembly generated doesn't make much sense.

Until https://github.com/FuelLabs/sway/issues/845 is implemented, we should probably disallow passing arguments to `main()`.

ameer-khan-ashraf:
I would like to work on this.

mohammadfawaz:
Thank you! Go for it and please reach out for any questions 😃

ameer-khan-ashraf:
I have some questions. How can I contact you @mohammadfawaz ?

mohammadfawaz:
@ameer-khan-ashraf discord is the best place to discuss this! As a starting point, one possible plate to introduce the check is right before type-checking function parameters for a function declaration: https://github.com/FuelLabs/sway/blob/376010e7b8f84334ffacd0f99e64f185c971d213/sway-core/src/semantic_analysis/ast_node/declaration/function.rs#L152
You'd have to introduce a new compile error, add the check (there is a helper call `is_main_function` somewhere I think), and emit the error if `is_main_function()` AND the number of function parameters is non-zero. You will then have to add a negative test in https://github.com/FuelLabs/sway/tree/master/test/src/e2e_vm_tests/test_programs/should_fail

denizsurmeli:
Hey, can I try this one if it's still open ?

denizsurmeli:
How can we use `is_main_function()` in the `type_check`, and should we ? Rather than wrapping the `FunctionDeclaration` into `TypedAstNode` and let the `is_main_function()` unwrap it, shouldn't we just check the `name == crate::constants::DEFAULT_ENTRY_POINT_NAME`. It's effectively the same thing, but we take the longer path if we try to use `is_main_function()` I believe. Also is there a way of finding out the `TreeType` in scope of `type_check` ? I believe I should have find my way in the `ctx`.

otrho:
I just took a quick look and I think @mohammadfawaz 's suggestion is OK but might have a couple of shortcomings (I'm sure he just looked for the most obvious starting point).  As you've pointed out the `TypedAstNode` wrapper is gone, but also it's a little deep into the process to be checking for edge cases.  If you were to put the check there we'd be checking every single function and method in the program just to see if they're `main()`.

[...]

Issue #845, opened 2022-02-25
Allow access of script data via arguments in the main function


sezna:
Script data is currently inaccessible without using `asm` to extract it from the call frame. Ideally, the compiler should magically handle that as a language feature and expose it to the user. 

This should probably take the form of arguments to the main function, e.g.:
```rust
fn main(a: u64, b: bool, c: etc...) -> bool { ... }
```

Where `a`, `b`, and `c` are script data. 

The compiler should probably revert if these values are not provided in the script data as expected.

adlerjohn:
> The compiler should probably revert if these values are not provided in the script data as expected.
For stack-only types, this can maybe be done with a single check against the `scriptDataLength` field of the tx.

Types that aren't full-width might also have to be checked to be `0` on their high bits.

nfurfaro:
For reference, we have this in the stdlib that could be removed when this issue is closed:
https://github.com/FuelLabs/sway-lib-std/blob/2c4f9ff384917442d4c2bf5cce530953673de59f/src/chain.sw#L54

simonr0204:
This seems to "just work" with predicates:

https://github.com/FuelLabs/sway-applications/blob/simon-otc-swap-predicate/otc-swap-predicate/src/main.sw#L17

Was this implemented separately to scripts or is my code not doing what I think it's doing ?

mohammadfawaz:
@AlicanC would be nice to generalize this to predicates as well. See https://github.com/FuelLabs/sway/issues/2581

I expect the work to be basically identical.

simonr0204:
> For reference, we have this in the stdlib that could be removed when this issue is closed: https://github.com/FuelLabs/sway-lib-std/blob/2c4f9ff384917442d4c2bf5cce530953673de59f/src/chain.sw#L54

This probably _shouldn't_ be removed, as other contexts might want to inspect the script data (e.g. a predicate spent by a given script might want to check the script data and bytecode). These wouldn't have access to the `main()` args directly as it's a different VM instance.

PR #2472, opened 2022-08-08
2421/no main args

denizsurmeli:
#2421


sway-core/src/lib.rs
```diff
@@ -384,6 +384,14 @@ pub(crate) fn compile_ast_to_ir_to_asm(
         errors
     );
 
+    // Check if we have `script` or `predicate` and their main function have arguments
```
otrho, line 387:
```suggestion
    // Check if we have `script` or `predicate` and their main functions have no arguments.
```

sway-core/src/semantic_analysis/ast_node/declaration/function.rs
```diff
@@ -226,7 +226,7 @@ impl TypedFunctionDeclaration {
             is_contract_call: ctx.mode() == Mode::ImplAbiFn,
             purity,
         };
-
+
```
otrho, line 229:
```suggestion

```

sway-core/src/semantic_analysis/program.rs
```diff
@@ -196,6 +196,29 @@ impl TypedProgram {
...
+                }
+            },
+            _ => (),
+        }
+
+        CompileResult{
+            value:None,
+            warnings,
+            errors,
+        }
```
otrho, line 219: Rather than create a `CompileResult` directly you should use `ok()` or `err()`, particularly since this isn't quite valid.  A `CompileResult` with a `None` value usually indicates an error.  I'll make an issue for this.
```suggestion
        ok((), warnings, errors)
```

test/src/e2e_vm_tests/test_programs/should_fail/main_should_not_have_args/test.toml
```diff
@@ -0,0 +1,4 @@
+category = "fail"
+
+# check: main(42)
+# nextln: $()Function "main" can't have any arguments.
```
otrho, line 4: Isn't the error message `arguments to "main()" are not yet supported.`?  That's what's in `error.rs` above...

otrho (changes requested):
I'm happy with this approach. :slightly_smiling_face: A few small changes, and as of right now the CI hasn't run... not sure why, but I imagine it might come up with some stuff.

Thanks for your contribution!

otrho:
> as of right now the CI hasn't run... not sure why, but I imagine it might come up with some stuff.

Aha, first time contribs need to have the CI approved by a contributor.  So hopefully it's running now.

sway-core/src/semantic_analysis/ast_node/declaration/function.rs
```diff
@@ -226,7 +226,6 @@ impl TypedFunctionDeclaration {
             is_contract_call: ctx.mode() == Mode::ImplAbiFn,
             purity,
         };
-
```
otrho, line 229: Whoops, did I do this with my suggestion?  It's unrelated to your PR so we should probably leave this file alone and keep the empty line.

denizsurmeli: I might have touched it accidentally, will fix it in a minute

sway-core/src/semantic_analysis/program.rs
```diff
@@ -196,6 +196,25 @@ impl TypedProgram {
...
+                if !main_function.parameters.is_empty() {
+                    errors.push(CompileError::MainArgsNotYetSupported {
+                        span: main_function.span.clone(),
+                    })
+                }
+            }
+            _ => (),
+        }
+
+        ok((), warnings, errors)
```
kayagokalp, line 215: Looks like event if we detect an error we are still returning `ok(...)`. By looking at the `finalize_types()` we may need to do something like this: 
```suggestion
        if errors.is_empty() {
            ok((), warnings, errors)
        } else {
            err(warnings, errors)
        }
```

kayagokalp (commented):
Thanks for the contribution 🚀 I am not from @FuelLabs/sway-compiler so it might make sense wait a comment from someone from the compiler team 😄  but I realized CI was failing and it looks like a simple fix. My local test shows that suggested change fixes the problem.

sway-core/src/semantic_analysis/program.rs
denizsurmeli: Oh that makes sense 😅  Thanks a lot !

denizsurmeli:
I couldn't set up the negative test probably, is there a guide or should I look the tests in the repo and find my way around ?

denizsurmeli (commented):
CI seems ok, should I do anything else ?

sezna:
We do perform other validations on `main()` functions [here](https://github.com/FuelLabs/sway/blob/master/sway-core/src/semantic_analysis/program.rs#L148), do you think it would be better to call your check there?

For posterity and documentation, we will eventually support this: #845. Perhaps you could link to the issue in the error message?

denizsurmeli:
Oh, thanks for the detailed explanation, I will add this as soon as I can

otrho (commented):
It looks like this still needs some work -- it's turning out to be a little bit more involved than we originally thought... :slightly_smiling_face: 

I'll mention here a couple more things -- we don't want to leave commented code in the source, it's just noise.  Also, I'm not sure whether this should return `ok` or `err` when `main()` has args.  You can see (in `program.rs`) on lines 158 and 181 that it's returning `err` when `main()` is missing, but also at lines 162 and 185 that it pushes an error but still returns `ok` at the bottom.  These were done by different authors... not sure which is better.

@sezna? Is it OK to return non-empty `errors` with `ok`?

denizsurmeli:
Yeah kinda wrestling with it, but I believe if we perform the check on `validate_root` we should wrap it in `ok` otherwise the tests start to break up. Following some traces but kinda lost 😄 Also sorry about the comments, fixing those too.

denizsurmeli:
I'm trying to run some tests on my local machine but getting
 ```
 thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Response errors; TransactionCreateBytecodeWitnessIndex', test/src/e2e_vm_tests/harness.rs:28:10
 ``` 
 when I try to run tests under the `e2e_vm_tests/test_programs/should_pass/test_contracts`.  Any suggestions ?

denizsurmeli:
> > I'm trying to run some tests on my local machine but getting
> > ```
> > thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Response errors; TransactionCreateBytecodeWitnessIndex', test/src/e2e_vm_tests/harness.rs:28:10
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > when I try to run tests under the `e2e_vm_tests/test_programs/should_pass/test_contracts`. Any suggestions ?
> 
> Hmm, never seen that error before. Something to keep in mind though, you have to restart the Fuel node between test runs.
> 
> I guess you are already running a node to get that error, but if you aren't, you can launch it by cloning the fuel-core repo and doing: `cargo run --release --bin fuel-core -- --db-type in-memory`

same error remains, any other suggestion ? thanks for your help

sezna:
Hm, you could try deleting your `~/.fuel` directory. 

`ok()` represents either an okay state, or at least a state that recovered (`Some` value + errors means it recovered). In the case of recovery, the errors will be reported but the compiler can continue on.

So, yes, it is okay to return non-empty `errs` with `ok()`, this is actually how we do recovery after a compile error. If an error is non-recoverable, then we use `err()` which has no value at all. 

Therefore, we should be using `ok()` here and reporting the errors back to the user, but continuing the compilation process.

otrho:
> Therefore, we should be using `ok()` here and reporting the errors back to the user, but continuing the compilation process

Yep, that's what I thought, except I didn't think we were catching them in this case.  But double checking I see that we do capture the errors when calling `validate_root()`, so `ok()` is correct.  Thanks.

denizsurmeli:
Thanks for detailed description first of all, the problem remains but I believe it's not due to my changes because the `master` branch fails with the same error also. The negative test passes for my part, do you want me to push the changes ? I'm open to any other suggestion also

sezna:
The test runner in CI is failing for two reasons. One, [an outdated lockfile](https://github.com/FuelLabs/sway/runs/7740476755?check_suite_focus=true#step:6:1169). Secondarily, the filecheck string you provided in `test.toml` (`}`)  is not being found.

See [this filecheck string](https://github.com/FuelLabs/sway/blob/76ae557ca143fbc66648186264da1f0e90cd6d6c/test/src/e2e_vm_tests/test_programs/should_fail/bad_generic_annotation/test.toml#L3) as an example. You'll want to put a string that matches the location of the error.

sezna:
As for the transaction error, that's not really relevant to this change and could be caused by a version mismatch. Feel free to filter out that test for now by running `cargo run -- <FILTER_STRING>` in `test`. E.g. `cargo run -- main_should_not_have_args`

denizsurmeli:
I know I have been a headache for the last couple of days but I really don't want to spend any more of your CI minutes 😅 . I have merged the master to my branch and on `v0.19.2` , but `Forc.lock` still has some [GitHub](https://github.com/fuellabs/sway?tag=v0.19.2#6808861389966f99887f71476918a562a9edd90e)  link, with version `0.19.2`. I have run the test explicitly on my local with `cargo run --locked --bin test -- main_should_not_have_args --locked` on my local machine and it passes. Is it going to break CI again, if so what are the steps I should take ? Thanks for your time again, sorry for your CI minutes.

sezna:
Not a headache at all, we love any and all contributions :) Also github CI minutes [are free for public respositories](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#about-billing-for-github-actions). 

If you just delete the `Forc.lock` file in your test, then run `forc build` again, it should regenerate the lockfile. Then, commit the regenerated file.

otrho (approved):

sezna (approved):

mohammadfawaz (approved):

PR #2478, opened 2022-08-08
refactor: move build logic to `forc-pkg` from `forc`

### LINKED ISSUE #2479: Move build logic to `forc-pkg`
Ideally we should have the building logic inside `forc-pkg` so that any other places that requires to perform a build would not need to depend on `forc`. 

This is originated from https://github.com/FuelLabs/sway/pull/2440#discussion_r940064544.

Quoting @mitchmindtree:

> as I think forc itself should just be a thin CLI wrapper on top of forc-pkg + plugin support

kayagokalp:
closes #2479


kayagokalp:
Adding the "critical" tag as this is blocking a issue tagged with "critical"

JoshuaBatty:
Hi @kayagokalp i'm happy to approve this, just need to bump the versions to `0.19.2` first.

JoshuaBatty (approved):

eureka-cpu (approved):
LGTM

PR #2480, opened 2022-08-08
Misc cleanups in the parser

Centril:
This refactors `sway-parse` with various de-duplications, simplifications, as well as adding docs in some places.


sezna:
We will actually be moving to sway fmt v2 very soon, which is a complete rewrite. That being said, one of the [old formatter tests](https://github.com/FuelLabs/sway/runs/7727128114?check_suite_focus=true#step:5:944) is failing.

mitchmindtree (approved):
Very nice :ok_hand: 

Maybe not the most critical stuff, but I can empathize with refactoring being a great way to familiarize with a code-base.

I've left one note, but it shouldn't be a blocker - will happily re-approve if you wish to address it.

I'll cc @canndrew too as `sway-parse` is his baby and he might want to stay in the loop.

sway-parse/src/item/mod.rs
```diff
@@ -20,67 +19,66 @@ mod item_use;
 
 impl Parse for ItemKind {
     fn parse(parser: &mut Parser) -> ParseResult<ItemKind> {
-        if parser.peek::<UseToken>().is_some() || parser.peek2::<PubToken, UseToken>().is_some() {
+        // FIXME(Centril): Visibility should be moved out of `ItemKind` variants,
+        // introducing a struct `Item` that holds the visibility and the kind,
+        // and then validate in an "AST validation" step which kinds that should have `pub`s.
```
mitchmindtree, line 24: In general, we try to keep a tracking issue for each TODO/FIXME and link it inline. Maybe we should add one and link it here? Edit: Woops, forgot to add this with previous review.

canndrew:
Nice. I really like the `guarded_parse` method, that does simplify the code in a lot of places.

LGTM :+1:

sezna (approved):

sway-parse/src/parser.rs
```diff
@@ -21,9 +23,9 @@ impl<'a, 'e> Parser<'a, 'e> {
     }
 
     pub fn emit_error(&mut self, kind: ParseErrorKind) -> ErrorEmitted {
-        let span = match self.token_trees.first() {
-            Some(token_tree) => token_tree.span(),
-            None => {
+        let span = match self.token_trees {
+            [token_tree, ..] => token_tree.span(),
```
canndrew, line 27: Using `first`/`Some`/`None` is slightly more robust, eg. because of [this](https://github.com/FuelLabs/sway/pull/2480/commits/52d3bad98d85b3b98338e7645556e0be6fdc23da) (wildcard patterns scare me in general). It's not a blocker on merging this though.

PR #2482, opened 2022-08-08
Refactor std::tx to use gtf intrinsic

nfurfaro:
### Main changes
- made small changes to both `token.sw` and `auth.sw` to handle the return types of some tx functions switching to enums
- Adds 2 new modules `Inputs` & `Outputs` and moves related stuff to correct module.
- follow-up pr's will add in all the remaining getters to expose all of the `gtf` fields.
- the list of `cont` values near the top of the `tx` module currently has all the needed consts listed, but the ones not yet used are commented out (I've been un-commenting as I implement them to help track my work).
- I've added some new `pub` enums to clean up some return types of functions:
```

pub enum Transaction {
    Script: (),
    Create: (),
}

pub enum Input {
    Coin: (),
    Contract: (),
    Message: (),
}

pub enum Output {
    Coin: (),
    Contract: (),
    Message: (),
    Change: (),
    Variable: (),
}
```


sway-lib-std/src/inputs.sw
```diff
@@ -0,0 +1,220 @@
...
+// const GTF_INPUT_CONTRACT_STATE_ROOT = 0x111;
+// const GTF_INPUT_CONTRACT_TX_POINTER = 0x112;
+// const GTF_INPUT_CONTRACT_CONTRACT_ID = 0x113;
+
+// const GTF_INPUT_MESSAGE_MESSAGE_ID = 0x114;
+// const GTF_INPUT_MESSAGE_SENDER = 0x115;
+// const GTF_INPUT_MESSAGE_RECIPIENT = 0x116;
+const GTF_INPUT_MESSAGE_AMOUNT = 0x117;
+// const GTF_INPUT_MESSAGE_NONCE = 0x118;
+const GTF_INPUT_MESSAGE_OWNER = 0x119;
```
pixelcircuits, line 49: The latest [spec changes](https://github.com/FuelLabs/fuel-specs/commit/5dcf71f162b11fb9ce998315a7078630d78536b7#diff-0ef30bdacf4ca9f9f500f9aaf4a50326591005bb730a703d5b8abc049bb475e7) removes the `owner` field from `InputMessages` and some of these constants have also updated to remove the gap. If [this issue](https://github.com/FuelLabs/fuel-asm/issues/92) is already closed, then those updates will need to be made here as well. Otherwise we will need to create a separate issue for the `owner` field on `InputMessage`

nfurfaro: Ok, good to know @pixelcircuits.
I will create an issue to track this and try to keep moving forward with this as is for now until a new fuel-core is released with the changes removing the owner field. Otherwise I'm aiming for a moving target with this PR and it will never be merged.

sway-lib-std/src/outputs.sw
```diff
@@ -118,11 +118,11 @@ pub fn output_amount(index: u64) -> u64 {
...
         // ues GTF_OUTPUT_MESSAGE_AMOUNT as there's no simlar const for OutputChange
         Output::Change => {
-            __gtf::<u64>(index, GTF_OUTPUT_MESSAGE_AMOUNT)
+            __gtf::<u64>(index, GTF_OUTPUT_COIN_AMOUNT)
         },
         // use GTF_OUTPUT_MESSAGE_AMOUNT as there's no simlar const for OutputVariable
         Output::Variable => {
-            __gtf::<u64>(index, GTF_OUTPUT_MESSAGE_AMOUNT)
+            0
         },
```
otrho, line 124: The comments here are now out of date...

sway-lib-std/src/outputs.sw
```diff
@@ -0,0 +1,126 @@
...
+        },
+        Output::Message => {
+            __gtf::<u64>(index, GTF_OUTPUT_MESSAGE_AMOUNT)
+        },
+        // Output changes are always guaranteed to have an amount of zero since
+        // they're only set after execution terminates
+        Output::Change => {
+            0
+        },
+        // use `__gtf` when GTF_OUTPUT_VARIABLE_AMOUNT is available
```
otrho, line 116: This is tracked in an issue somewhere?

otrho (approved):
A couple of questions, but I know this has been a long time coming and is semi-urgent.

pixelcircuits:
Looks good to me! The only note I have is that the hardcoded GTF values are set to the current spec which removes `owner` from `InputMessages`, but this has not been implemented yet in fuel-asm, fuel-tx, or fuel-core yet ([see here](https://github.com/FuelLabs/fuel-specs/issues/399)). There could be some minor headaches until all repos are caught up

pixelcircuits (dismissed):

sway-lib-std/src/chain/auth.sw
```diff
@@ -47,34 +47,42 @@ pub fn msg_sender() -> Result<Identity, AuthError> {
...
+            },
+            Input::Message => {
+                ();
+            },
+            _ => {
+                // type != InputCoin or InputMessage, continue looping.
+                i += 1u8;
+                continue;
+            }
         }
```
Braqzen, line 69: Does the `_` check for multiple types? Would it be cleaner to use an `if` expression instead? In `inputs.sw` it looks like it looks for a single input, a Contract. I think this should use an `if` expression since that would halve the number of lines of code

sway-lib-std/src/token.sw
```diff
@@ -95,13 +97,23 @@ pub fn transfer_to_output(amount: u64, asset_id: ContractId, to: Address) {
...
-        if tx_output_type(index) == OUTPUT_VARIABLE && tx_output_amount(index) == 0 {
-            output_index = index;
-            output_found = true;
-            break; // break early and use the output we found
-        };
+    let outputs = output_count();
+    //log(outputs);
+    while index < outputs {
+        let type_of_output = output_type(index);
+        log(type_of_output);
```
Braqzen, line 104: Does this function, from the `std`, need to log?

sway-lib-std/src/token.sw
```diff
@@ -95,13 +97,23 @@ pub fn transfer_to_output(amount: u64, asset_id: ContractId, to: Address) {
...
+        match type_of_output {
```
Braqzen, line 105: Similar comment regarding `if` expressions here. If it's only checking for a single type then it makes more sense to use an `if` imo

test/src/sdk-harness/test_artifacts/tx_contract/src/main.sw
```diff
@@ -3,12 +3,24 @@ contract;
 use std::address::Address;
```
Braqzen, line 3: I would join all imports from the same place so that the formatter can handle it, including alphabetical ordering (in all files), that aside, does `tx` need to import using `*`?

test/src/sdk-harness/test_artifacts/tx_contract/src/main.sw
```diff
@@ -66,26 +74,23 @@ impl TxContractTest for Contract {
...
-    fn get_tx_input_type_from_ptr(ptr: u64) -> u8 {
-        tx_input_type_from_pointer(ptr)
+        input_pointer(index)
     }
-    fn get_tx_input_type(index: u64) -> u8 {
-        tx_input_type(index)
+    fn get_input_type(index: u64) -> Input {
+        input_type(index)
     }
     // TODO: Add test for getting InputMessage owner when we have InputMessages
```
Braqzen, line 82: Is this tracked in an issue?

test/src/sdk-harness/test_projects/tx_fields/mod.rs
```diff
@@ -34,18 +32,18 @@ async fn can_get_tx_type() {
...
     let result = contract_instance.get_tx_type().call().await.unwrap();
     // Script transactions are of type = 0
-    assert_eq!(result.value, 0);
+    assert_eq!(result.value, Transaction::Script());
 }
 
 #[tokio::test]
 async fn can_get_gas_price() {
     let (contract_instance, _, _) = get_contracts().await;
     // TODO set this to a non-zero value once SDK supports spending coins.
```
Braqzen, line 41:
```suggestion
```

TODO no longer needed?

Braqzen (commented):
Some suggestions

sway-lib-std/src/chain/auth.sw
otrho: The code is skipping inputs which aren't `Coin` or `Message` by continuing the loop, as the comment says.  The whole thing is a bit fragile, but the best alternative is iterators.

sway-lib-std/src/token.sw
otrho: Yeah, in this case it could be an `if let ...`.

sway-lib-std/src/token.sw
mohammadfawaz: Oh shoot no. I'll remove. Good catch.

nfurfaro: fixed.

sway-lib-std/src/outputs.sw
mohammadfawaz: Done.

sway-lib-std/src/chain/auth.sw
mohammadfawaz: As per Toby's comment, leaving this as is for now.

mohammadfawaz:
> Looks good to me! The only note I have is that the hardcoded GTF values are set to the current spec which removes `owner` from `InputMessages`, but this has not been implemented yet in fuel-asm, fuel-tx, or fuel-core yet ([see here](https://github.com/FuelLabs/fuel-specs/issues/399)). There could be some minor headaches until all repos are caught up

Reverted to the old constants (most of them were commented out anyways) and added a comment explaining why they don't match the spec.

mohammadfawaz (approved):

test/src/sdk-harness/test_artifacts/tx_contract/src/main.sw
mohammadfawaz: Done

test/src/sdk-harness/test_artifacts/tx_contract/src/main.sw
```diff
@@ -1,21 +1,23 @@
...
+        input_type,
+    },
+    outputs::{
+        Output,
+        output_count,
+        output_pointer,
+        output_type,
+    },
+    tx::*,
+    };
```
Braqzen, line 20:
```suggestion
};
```

:)

Braqzen (approved):

sway-lib-std/src/inputs.sw
```diff
@@ -0,0 +1,160 @@
...
+/// (transaction-script or transaction-create).
+pub fn input_count() -> u8 {
+    let type = tx_type();
+    match type {
+        Transaction::Script => {
+            log(17);
+            __gtf::<u8>(0, GTF_SCRIPT_INPUTS_COUNT)
+        },
+        Transaction::Create => {
+            log(19);
```
pixelcircuits, line 156: What are these logs for?

otrho: Ooh, definitely debug logs!  Whoops.

PR #2483, opened 2022-08-08
Remove the global `TypeEngine` in favor of adding it to the `TypeCheckContext`

### LINKED ISSUE #2063: Remove the global `TypeEngine` and `DeclarationEngine` in favour of moving them into the `TypeCheckContext`
As mentioned [here](https://github.com/FuelLabs/sway/pull/2004#pullrequestreview-1012828683), this is an important follow-up to #2004.

It should make it much easier to enable type-checking the same program multiple times within the same process. Currently this might be possible, but doing so will endlessly bloat the global arena as it's never cleaned up.

Accumulating the type state from the beginning of `TypedProgram::type_check` and returning it within the `TypedProgram` will allow us to take advantage of Rust's ownership semantics and leave its cleanup to RAII.

emilyaherbert:
- Closes #2063


emilyaherbert:
Closing this branch because it is stale.

Issue #2486, opened 2022-08-08
Support individual element spans for tuples and arrays in `TypedStructFields` and `TypedEnumVariants`


JoshuaBatty:
I'm currently adding in support for collecting the fields from Struct and Storage Declarations in the language server. I have everything working apart from arrays and tuples. 

The image below demonstrates the fields that we can and can't currently collect. I'm drawing squiggly error lines under the tokens which demonstrates the span information accessible from the compiler.

<img width="633" alt="Screen Shot 2022-08-09 at 9 49 28 am" src="https://user-images.githubusercontent.com/1289413/183533544-e23db43f-f1a1-4401-bd78-bf099c9cf26c.png">

Currently, the `Span` for an array is `"[u64, 3]"` whereas we need to have 2 unique spans. One for `"u64"` and one for `"3"`. 

Similarly with tuples. we need unique spans for the individual elements. At the moment the compiler just has spans for each element in the outer tuple. We also require unique span information for inner tuples and arrays within a tuple.

PR #2488, opened 2022-08-09
Collect `TypedStructField` tokens in the language server.

JoshuaBatty:
This PR allows us to collect both the name and the type as tokens for fields in Struct and Storage Declarations. This is working for most types apart from arrays and tuples as outlined in #2486. 

Go to definition support is now also working for fields in Struct expressions.


sway-lsp/src/server.rs
```diff
@@ -291,25 +291,25 @@ mod tests {
...
             .unwrap()
             .parent()
             .unwrap()
             .join("test/src/e2e_vm_tests/test_programs/should_pass/language")
-            .join("enum_if_let")
+            .join("struct_field_access")
     }
 
-    fn sway_example_dir() -> PathBuf {
+    fn _sway_example_dir() -> PathBuf {
```
sezna, line 303: what's this for?

sezna (dismissed):

sway-lsp/src/server.rs
JoshuaBatty: I have 2 helper methods for testing against the sway examples folder and another for the e2e language folder. Clippy complains if there are any unused functions so I underscore the one I'm not currently using for development. 

This is actually one of the main motivations for me wanting to inspect the syntax tree in the editor for the currently open file in VScode.

sezna: Ah ok, fair. what about just doing `#[allow(dead_code)]` on both?

sezna (approved):

sway-lsp/src/server.rs
JoshuaBatty: Yeah good idea. I'll add that to my next PR. Cheers.

mohammadfawaz (approved):
Compiler changes seem reasonable to me!

PR #2502, opened 2022-08-09
Organize imports in `sfv2`

### LINKED ISSUE #2501: Remove extra new lines from imports in `sfv2`

eureka-cpu:
This removes newlines from imports so they can be properly organized.
Closes #2501


JoshuaBatty (dismissed):

sway-fmt-v2/src/config/comments.rs
```diff
@@ -1,7 +1,6 @@
 //! Configuration options related to formatting comments.
-use crate::constants::DEFAULT_MAX_COMMENT_WIDTH;
-
 use super::user_opts::CommentsOptions;
+use crate::constants::DEFAULT_MAX_COMMENT_WIDTH;
```
kayagokalp, line 3: Do we want to keep this kind of separate imports rooted from `super::`? If we also convert those imports to `crates::` type (and add it to the others) they would also be ordered by `rustfmt`. But I would also understand if you prefer having these separated as it is a shorter relative path. 

```suggestion
use crate::{config::user_opts::CommentsOptions, constants::DEFAULT_MAX_COMMENT_WIDTH};
```

sway-fmt-v2/src/config/heuristics.rs
```diff
@@ -1,13 +1,12 @@
 //! Configuration options related to heuristics.
+use super::user_opts::HeuristicsOptions;
```
kayagokalp, line 2: ^ Same

sway-fmt-v2/src/config/items.rs
```diff
@@ -1,8 +1,7 @@
 //! Configuration options related to item formatting.
-use serde::{Deserialize, Serialize};
-
 use super::{lists::ListTactic, user_opts::ItemsOptions};
 use crate::constants::{DEFAULT_BLANK_LINES_LOWER_BOUND, DEFAULT_BLANK_LINES_UPPER_BOUND};
```
kayagokalp, line 3: ^ Same

sway-fmt-v2/src/config/whitespace.rs
```diff
@@ -1,9 +1,7 @@
 //! Standard system and editor whitespace configuration options. Advanced whitespace options will be deferred to their corresponding sub-classes.
-use serde::{Deserialize, Serialize};
-
-use crate::constants::{CARRIAGE_RETURN, DEFAULT_MAX_LINE_WIDTH, DEFAULT_TAB_SPACES, LINE_FEED};
-
 use super::user_opts::WhitespaceOptions;
+use crate::constants::{CARRIAGE_RETURN, DEFAULT_MAX_LINE_WIDTH, DEFAULT_TAB_SPACES, LINE_FEED};
```
kayagokalp, line 3: ^ Same

sway-fmt-v2/src/utils/attribute.rs
```diff
@@ -1,3 +1,7 @@
+use super::{
+    bracket::{Parenthesis, SquareBracket},
+    comments::{ByteSpan, LeafSpans},
+};
```
kayagokalp, line 4: Same ^

sway-fmt-v2/src/utils/expr/mod.rs
```diff
@@ -1,3 +1,7 @@
+use super::{
+    bracket::{CurlyBrace, Parenthesis, SquareBracket},
+    shape::{ExprKind, LineStyle},
+};
```
kayagokalp, line 4: ^

sway-fmt-v2/src/utils/generics.rs
```diff
@@ -1,9 +1,8 @@
 use super::bracket::{close_angle_bracket, open_angle_bracket};
+use super::shape::LineStyle;
```
kayagokalp, line 2: ^

sway-fmt-v2/src/utils/path.rs
```diff
@@ -1,3 +1,4 @@
+use super::bracket::{close_angle_bracket, open_angle_bracket};
```
kayagokalp, line 1: ^

sway-fmt-v2/src/utils/pattern.rs
```diff
@@ -1,3 +1,4 @@
+use super::bracket::{CurlyBrace, Parenthesis};
```
kayagokalp, line 1: ^

sway-fmt-v2/src/utils/program_type.rs
```diff
@@ -1,11 +1,9 @@
+use super::comments::{ByteSpan, LeafSpans};
+use crate::FormatterError;
```
kayagokalp, line 2: ^

sway-fmt-v2/src/utils/punctuated.rs
```diff
@@ -1,3 +1,4 @@
+use super::shape::LineStyle;
```
kayagokalp, line 1: ^

sway-fmt-v2/src/utils/ty.rs
```diff
@@ -1,3 +1,4 @@
+use super::shape::LineStyle;
```
kayagokalp, line 1: ^

sway-fmt-v2/src/utils/where_clause.rs
```diff
@@ -1,10 +1,9 @@
+use super::comments::{ByteSpan, LeafSpans};
```
kayagokalp, line 1: ^

kayagokalp (commented):
Looking good! I just left a single nit (on multiple places) about mixing `super::` and `crate::` type paths. I think if we also merge those automatic `rustfmt` ordering that we are trying to ensure with this PR will benefit a little more but I understand that addressing the same piece by `crate::` rather than `super::` will be longer, so I guess this is more like a design decision and thus I am leaving this up to you. I just wanted to provide an option to consider 👍 Since automatic merge is enabled for this PR, I am not approving right now to provide a chance for you to look at the review but if you decide to go like this, you can ping me and I will happily approve.

sway-fmt-v2/src/config/comments.rs
eureka-cpu: Personally I like `super` (: if @mitchmindtree or @JoshuaBatty prefer `crate` only then I'd gladly change it. But I have no qualms as is!

mitchmindtree: I think `crate` is perhaps a little more conventional, though `super` can be useful when you're in a module that you know will always be nested and have the same relative path, e.g.

```rust
// Inside module `foo.rs`

fn foo() { ... }

mod tests {
    use super::foo;
    // ...
}
```

^ Even if `foo.rs` gets moved or nested within another module, `tests` will always be nested within `foo.rs` so it's pretty safe to use `super` from within `tests`.

For most of the cases in this lib I'd personally go with `crate`.

JoshuaBatty: Yeah I prefer `crate`, makes it a bit more flexible for the reasons mitchie stated above.

eureka-cpu: gotcha, I'll add those changes in the morning

sway-fmt-v2/src/items/item_use.rs
```diff
@@ -7,6 +7,7 @@ use crate::{
     },
 };
 use std::fmt::Write;
+use std::vec;
```
kayagokalp, line 10: Maybe we can also merge them like:

```suggestion
use std::{fmt::Write, vec};
```

kayagokalp (commented):
Just left a single nit that might be slipped through otherwise I think it is good to go!

sway-fmt-v2/src/items/item_use.rs
kayagokalp: Also I think there are no problems if we do not have this `use` for `vec`, if you prefer we can also drop it.

kayagokalp (approved):

JoshuaBatty (approved):

PR #2504, opened 2022-08-09
Changes `.clone().into_inner()` to `.get()` in `sfv2`

### LINKED ISSUE #2499: Use `.get()` instead of `.clone().into_inner()` when obtaining inner value from braces in `sfv2`

eureka-cpu:
~Running into a problem that I'm assuming came from this, which is that clippy is complaining about IR using an unstable feature. cc @otrho any idea what I can do to fix it?~

JK, somehow clippy is passing here but not on my machine...

Rust update fixed it

Closes #2499


otrho:
It'll be about `bool::then_some()` which is new.  You just need to update your Rust.

eureka-cpu:
> It'll be about `bool::then_some()` which is new. You just need to update your Rust.

I appreciate the swift reply <3

mitchmindtree (approved):
Nailed it!

kayagokalp (approved):
looks perfect and a lot cleaner!

PR #2506, opened 2022-08-10
Use rust 1.62 in CI docker image

sezna:
This is what is breaking CI. We probably want to remove this docker image eventually anyway, now that we have `fuelup`. The image gets zero downloads.﻿


Voxelot (approved):

otrho (approved):
Aha, there it is.  Was scratching my head over this all morning.  Thanks.

mohammadfawaz (approved):

PR #2507, opened 2022-08-10
Add separate expression kind types

### LINKED ISSUE #2485: Compiler is hitting a stack overflow error in debug mode
The `u256` PR #2445 from @rostyslavtyshko is blocked from a stack overflow in the compiler. With the `u256`
lib in the `std-lib` we are just sneaking in under the maximum stack size when compiling, but when the compiler is invoked from a test (such as in LSP tests) the stack is getting tipped over when some of the stack is also being used for the test.

I will note that running the test runs fine when running with the `--release` flag set. 

It seems that some optimizations are needed to reduce the stack size, perhaps in the parser or in other parts of the compiler.

canndrew:
* Refactor the `Expression` enum into a struct with separate fields for the span and the expression kind.
* Add an `ExpressionKind` enum, which is just `Expression` but with all the span fields removed.
* Add separate types for the different kinds of expressions, and turn all of `ExpressionKind`'s variants into single-field tuple variants.
* Box the larger `ExpressionKind` variants to reduce its size.

This fixes #2485

### DIFF
--- a/sway-ast/src/lib.rs
+++ b/sway-ast/src/lib.rs
@@ -21,7 +21,7 @@ pub mod where_clause;
 pub use crate::{
     assignable::Assignable,
     attribute::AttributeDecl,
-    brackets::{AngleBrackets, Braces},
+    brackets::{AngleBrackets, Braces, Parens},
     dependency::Dependency,
     expr::{
         asm::{AsmBlock, AsmRegisterDeclaration},
@@ -44,11 +44,12 @@ pub use crate::{
         item_use::{ItemUse, UseTree},
         FnArg, FnArgs, FnSignature, Item, ItemKind, TypeField,
     },
-    keywords::{DoubleColonToken, PubToken},
+    keywords::{CommaToken, DoubleColonToken, PubToken},
     literal::{LitInt, LitIntType, Literal},
     module::{Module, ModuleKind},
     path::{PathExpr, PathExprSegment, PathType, PathTypeSegment, QualifiedPathRoot},
     pattern::{Pattern, PatternStructField},
+    punctuated::Punctuated,
     statement::{Statement, StatementLet},
     ty::Ty,
     where_clause::{WhereBound, WhereClause},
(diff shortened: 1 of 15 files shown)


otrho (commented):
From the 4 dot points above, I'm surprised any difference is made with the first 3 -- an enum's size should be approximately the size of the largest variant, so delegating it down to `ExpressionKind` wouldn't really make much difference, would it?

The 4th point about boxing larger types makes sense though.

Have you rebased #2445 onto here experimentally to confirm it fixes that problem?

canndrew:
> From the 4 dot points above, I'm surprised any difference is made with the first 3 -- an enum's size should be approximately the size of the largest variant, so delegating it down to ExpressionKind wouldn't really make much difference, would it?

Boxing some of the variants means making them they're own types. And since I was going to add types for some variants I figured I'd do it for all variants (or at least the ones that weren't already a single field). And since every variant had a `span` field I figured it made sense to factor it out.

> Have you rebased #2445 onto here experimentally to confirm it fixes that problem?

Yep, and it does.

canndrew:
> From the 4 dot points above, I'm surprised any difference is made with the first 3 -- an enum's size should be approximately the size of the largest variant, so delegating it down to ExpressionKind wouldn't really make much difference, would it?

Boxing some of the variants means making them they're own types. And since I was going to add types for some variants I figured I'd do it for all variants (or at least the ones that weren't already a single field). And since every variant had a `span` field I figured it made sense to factor it out.

> Have you rebased #2445 onto here experimentally to confirm it fixes that problem?

Yep, and it does.

canndrew:
I've rebased this off of master and added another commit which factors some functions out of `convert_parse_tree::expr_to_expression`. This reduces the (debug mode) stack usage of that function by a further 25%. Boxing the enum variants reduced the stack usage by 40%.

sezna (approved):
> I'm surprised any difference is made with the first 3 -- an enum's size should be approximately the size of the largest variant,

While this PR doesn't do this a lot, one place this saves space is in the situation where you were previously returning, or taking as an argument, an `Expression` that was always a single variant type. Now we can specify arguments and return types as the variant type itself. e.g. what used to be `fn parse_literal(a: &str) -> Expression` can become `fn parse_literal(a: &str) -> Literal`

mohammadfawaz (approved):

JoshuaBatty (approved):
Looks good, thanks for taking a look at this!

Issue #2505, opened 2022-08-10
Refactor the `sway-core/src/asm_generation` directory.


otrho:
When switching from AST -> ASM to AST -> IR -> ASM all the new IR -> ASM code went in `from_ir.rs` to keep it simple and contained.

But the old AST -> ASM code has been removed now and `from_ir.rs` is ~2500 lines and should be broken up and refactored.  All the tests called from there and hosted in `sway-core/tests/ir_to_asm` still need to be moved over to `sway/test/src/ir_generation` or somewhere similar, and switched to use `FileCheck`.

otrho:
I took a birds eye view of what's in `asm_generation` and the pipeline from IR to ASM and took some notes.   Might be useful in a refactor.  Of note is the duplication of opcodes between `VirtualOp` and `AllocatedOp` which is nice for type safety but does involve a lot of duplication.

```
* ir -> asm:
  * from_ir.rs:
    * compile_ir_to_asm():
      * compile_module_to_asm()         sway_ir::Context -> AbstractInstructionSet
      * remove_unnecessary_jumps()      AbstractInstructionSet -> JumpOptimizedAsmSet
      * allocate_registers()            JumpOptimizedAsmSet -> RegisterAllocatedAsmSet
        * realize_labels()              AbstractInstructionSet -> RealizedAbstractInstructionSet
        * allocate_registers()          RealizedAbstractInstructionSet -> InstructionSet via register allocator
      * optimize()                      InstructionSet -> FinalizedAsm

* JumpOptimizedAsmSet                   - wrapper around AbstractInstructionSet
* RegisterAllocatedAsmSet               - wrapper around InstructionSet
* FinalizedAsm                          - wrapper around InstructionSet

* AbstractInstructionSet                - collection of Op
* RealizedAbstractInstructionSet        - collection of RealizedOp
* InstructionSet                        - collection of AllocatedOp

* Op                                    - Either<VirtualOp, OrganizationalOp>
  * VirtualOp                           - giant list of opcodes
  * OrganizationalOp                    - tiny list of opcodes (label, jumps, data section placeholder)
* AllocatedOp                           - giant list of opcodes
* RealizedOp                            - VirtualOp

* VirtualOp:
  * Opcodes with VirtualRegisters and VirtualImmediates
* AllocatedOp:
  * Opcodes with AllocatedRegisters and AllocatedImmediates
```

otrho:
This has changed a bit with #2843 but is still desperately needed.

[...]

PR #2508, opened 2022-08-10
Intrinsics for the VM's storage opcodes.

### LINKED ISSUE #2514: Implement intrinsics for storage operations
With the [RFC](https://github.com/FuelLabs/sway-rfcs/blob/master/rfcs/0003-intrinsics.md) on intrinsics accepted, we should plan to slowly replace assembly usage in our libraries with intrinsics. As a step towards this, this Issue is to track the implementation of intrinsics for storage ops, namely: load/store word/quad.

vaivaswatha:
This PR provides the following intrinsics:

1. `__state_load_word (key: b256) -> u64;`
2. `__state_store_word (key: b256, value: u64);`
3. `__state_load_quad (key: b256, addr: u64);`
4. `__state_store_quad (key: b256, addr: u64);`

All of them compile to equivalent IR instructions. The last two (quad versions) required some changes to asm generation since they use `int_to_ptr` as inputs to the IR instructions for quad storage.

While adding the tests, a flaw I discovered was that: If the intrinsics are called from a script, passing an address that belongs to the script, we hit a runtime error (because the address is now referenced in the contract, not the calling script). This indicates that we'll need some static analysis around passing addresses - flagging such issues. Another issue would be if we take the address of a `const`, if at all that is allowed - I haven't checked. These are issues with the `__addr_of` intrinsic rather than this one, so I suppose I can file a separate Issue on that?

Closes #2514


mohammadfawaz:
> a flaw I discovered was that: If the intrinsics are called from a script

We should certainly error out at compile-time if storage intrinsics are used in a script!

vaivaswatha:
> > a flaw I discovered was that: If the intrinsics are called from a script
> 
> We should certainly error out at compile-time if storage intrinsics are used in a script!

I think that already happens. The problem is if you take a local variable's address in a script, then call a (wrapper) function in the contract that in-turn calls one of these intrinsics using that address, then we have a runtime error.

mohammadfawaz:
Three small requests:
- Do you mind creating a short issue that explains the motivation behind this change and link the PR to it? It's probably good to mention that this is one task towards implementing the intrinsics RFC that you proposed.
- Will you also describe the source of the runtime error in another issue so that we don't forget about it?
- Finally, can we now use the intrinsic in the standard library to replace `asm` blocks? Is it a good idea to this in this PR?

vaivaswatha:
> This is great! Thanks Vaivas :)
> 
> Three small requests:
> 
>     * Do you mind creating a short issue that explains the motivation behind this change and link the PR to it? It's probably good to mention that this is one task towards implementing the intrinsics RFC that you proposed.
> 
>     * Will you also describe the source of the runtime error in another issue so that we don't forget about it?
> 
>     * Finally, can we now use the intrinsic in the standard library to replace `asm` blocks? Is it a good idea to this in this PR?

I meant to do the 3rd one in this PR and forgot all about it 🤦‍♂️ . I'll do all three of these and update back here. Thanks Mo.

vaivaswatha:
> This is great! Thanks Vaivas :)
> 
> Three small requests:
> 
> * Do you mind creating a short issue that explains the motivation behind this change and link the PR to it? It's probably good to mention that this is one task towards implementing the intrinsics RFC that you proposed.
> * Will you also describe the source of the runtime error in another issue so that we don't forget about it?
> * Finally, can we now use the intrinsic in the standard library to replace `asm` blocks? Is it a good idea to this in this PR?

New issues: #2515 and #2514

sezna:
One thing I am curious about is the portability of these intrinsics. Given that the EVM's word size is different, the quad word intrinsic would have to load `256 * 4 == 1024` bits in order for the name to make sense. Is there a naming scheme that could give us more freedom to implement these optimally? Perhaps `state_load_256` or something?

mohammadfawaz:
> One thing I am curious about is the portability of these intrinsics. Given that the EVM's word size is different, the quad word intrinsic would have to load `256 * 4 == 1024` bits in order for the name to make sense. Is there a naming scheme that could give us more freedom to implement these optimally? Perhaps `state_load_256` or something?

+ 1 for this.

otrho:
Aren't we going to have different intrinsics for different targets?  So maybe since these are Fuel specific they should be `__fuelvm_state_load_word()` and we can have `__evm_state_load_word()` eventually too.  And, then I guess different instructions in the IR... :thinking: 

I was :thinking: about this the other day actually -- platform specific stuff in the IR should probably be generalised with wrappers.  We're going to have to do this with ASM blocks which are currently FuelVM specific in the IR, and so any platform specific stuff will need the same.  Storage counts here -- it's very, very similar between FuelVM and EVM but not the same.

sezna:
> Aren't we going to have different intrinsics for different targets?

Hmm, my mental model has been that we will use the same intrinsics and accomplish portability by implementing the codegen differently. Otherwise we aren't increasing portability with intrinsics relative to asm blocks, since you'd have to rewrite both of them. Platform specific stuff can live in asm perhaps? 

How do portable intrinsics work in e.g. rust? I guess they're platform specific too, thinking about things like the AVX intrinsics....maybe there's a middle ground where some are disallowed for certain backends, like the ones related to the fuel tx frame -- those would need to be disallowed for EVM..

Just having a stream of consciousness here...

otrho:
I was thinking yeah, a certain subset of intrinsics will just not compile, with an error, if your target isn't right for them.  Some intrinsics simply won't exist for certain targets, especially if we wrap basic ASM ops (in the ongoing effort to remove ASM blocks from the libs in the name of type safety).

ASM blocks themselves are going to have to change drastically, though this isn't really the place for that discussion.  But they're going to have to be more opaque than they are now, and maybe intrinsics should be the same.

We could actually have just one 'intrinsic' which is parameterised with an enum.  So `__size_of<T>()` becomes `__builtin<T>(SizeOf)`, or we have e.g., `__builtin<T>(Eq, lhs: T, rhs: T)` and `__builtin` is very much special cased and parameterised based on target.   It would have to magically allow any number of type parameters and args.

I'm not sure if this is less work and/or easier to implement though.

sezna:
We also need to standardize how we define what intrinsics are allowable for what backends, and it should be thoroughly considered..

> We could actually have just one 'intrinsic' which is parameterised with an enum. So __size_of<T>() becomes __builtin<T>(SizeOf), or we have e.g., __builtin<T>(Eq, lhs: T, rhs: T) and __builtin is very much special cased and parameterised based on target. It would have to magically allow any number of type parameters and args.

What would be the benefit of restructuring in this way?

otrho:
There is one intrinsic whose behaviour is entirely data driven based on the target.

vaivaswatha:
I've now updated `storage.sw` to use the new intrinsics. To workaround type limitations on the arguments and return type of the intrinsics, I'm using `asm` to do type casts (this isn't a problem at runtime because of the `is_reference_type` guard). Ideally, this should be solved with type constraints and not using `is_reference_type` as detailed [here](https://github.com/FuelLabs/sway/pull/2371#issuecomment-1196669323).

So that leaves only the question of names for these intrinsics (the discussion going on above related to different targets). I'll update the names once that is finalized, but the rest of the code can be reviewed now I think.

sway-core/src/asm_generation/from_ir.rs
```diff
@@ -1602,39 +1603,50 @@ impl<'ir> AsmBuilder<'ir> {
...
-        if val_ptr.value.is_none() {
-            return val_ptr.map(|_| ());
-        }
-        let (val_ptr, ptr_ty, offset) = val_ptr.value.unwrap();
-
-        // Expect the ptr_ty for val to also be B256
-        assert!(ptr_ty.eq(self.context, &Type::B256));
+        let val_reg = if matches!(
+            &self.context.values[val.0].value,
+            ValueDatum::Instruction(Instruction::IntToPtr(..))
```
otrho, line 1608: Rather than special case `IntToPtr` here would it work instead to update `resolve_ptr()` to handle both `GetPointer` and `IntToPtr`?

sway-core/tests/ir_to_asm/storage_store_load_intrinsics.ir
```diff
@@ -0,0 +1,486 @@
+contract {
+    fn main<ea1a0f91>() -> u64, !3 {
```
otrho, line 2: As per #2505 we really need to move these tests away from here and use `FileCheck`.  This is a bit ridiculous. :smiley:

sway-lib-std/src/storage.sw
```diff
@@ -40,21 +39,18 @@ use ::result::Result;
...
-        // If copy type, then it's a single word and can be read with a single
-        // SRW.
-        asm(k: key, v) {
-            srw v k;
-            v: T
+        // If copy type, then it's a single word
+        let loaded_word = __state_load_word(key);
+        asm (l: loaded_word) {
+            l: T
         }
```
otrho, line 54: Can we not turbofish this one here?  Was that also part of the discussion around this issue?

sway-lib-std/src/storage.sw
```diff
@@ -66,12 +62,12 @@ use ::result::Result;
...
 
         while size_left > 32 {
-            // Read 4 words (32 bytes) at a time using `srwq`
+            // Read 4 words (32 bytes) at a time
             let current_pointer = stack_ptr();
-            asm(k: local_key, v: current_pointer) {
+            asm() {
                 cfei i32;
-                srwq v k;
             };
```
otrho, line 69: Next is an `__alloca()` intrinsic to wrap `CFEI` since it's a bit weird to grab the stack pointer with  a call but then allocate some space with an ASM block.  I guess it could go in a library function for now...

otrho (commented):
Comments, below, all good otherwise.

sway-core/src/asm_generation/from_ir.rs
vaivaswatha: > Rather than special case `IntToPtr` here would it work instead to update `resolve_ptr()` to handle both `GetPointer` and `IntToPtr`?

I considered that, but there's some computation that's done later which isn't needed for `IntToPtr` (but needed if resolved as a stack pointer in `resolve_ptr`. So it wouldn't serve any purpose since I'll need to handle them differently here anyway.

sway-lib-std/src/storage.sw
vaivaswatha: > Can we not turbofish this one here? Was that also part of the discussion around this issue?

You mean using `std::mem::read<T>`? That wouldn't work as it ends up actually doing a load, assuming the value to be an address (in case of non-reference types).

otrho: The asm block is just casting the loaded word to `T` -- if it's a copy type then we should be able to use `__state_load_word::<T>(key)` to get it in the right type from the start, no?  I didn't check, is `__state_load_word()` just returning a `u64` right now?  Maybe it should return `T`.

otrho: Or is this just the problem we can't solve until we have trait constraints?

vaivaswatha: It really only returns `u64`, so having its type to return `T` would be incorrect. But if it's returning only `u64`, then when used with `is_reference_type<T>` in a generic function, we end up with this problem we have. If we could restrict it with trait contraints, then we wouldn't have this call in a non `u64` context at all.

otrho: Right, OK.  I kinda feel like we should only have two intrinsics -- read and write, and whether they need to use `SRW`/`SWW` or `SRWQ`/`SWWQ` can be decided by the back end.  Then the read functions _could_ just return `T`.

But I'm sure that won't work either for some reason.  OK, this'll do for now but I can feel the technical debt growing weekly. :smile:

sway-lib-std/src/alloc.sw
```diff
@@ -45,3 +46,12 @@ pub fn realloc(ptr: u64, size: u64, new_size: u64) -> u64 {
         ptr
     }
 }
+
+// Allocate a type on the stack.
+pub fn alloca<T>() -> u64 {
+    let current_pointer = stack_ptr();
+    asm() {
+        cfei i32;
```
otrho, line 54: This is allocating 32 bytes, where it needs to allocate '__size_of<T>()`.

vaivaswatha: > This is allocating 32 bytes, where it needs to allocate '__size_of()`.

Oops, yes. Let me fix this.

vaivaswatha: Dropping (and reverting related commits) this as it needs to be an intrinsic. The argument to `cfei` must be an immediate :-(

otrho (dismissed):
Alright.  Approving and then I'm out of here.  Have a good weekend!

vaivaswatha:
> Alright. Approving and then I'm out of here. Have a good weekend!

Thank you and you too !

otrho (approved):

sezna (approved):

PR #2510, opened 2022-08-10
Remove send address check

joerlop:
Given that send() can be called by any address, we shouldn't check whether it's being called by the contract creator.


examples/subcurrency/src/main.sw
```diff
@@ -90,15 +90,6 @@ impl Token for Contract {
         // Note: The return type of `msg_sender()` can be inferred by the
         // compiler. It is shown here for explicitness.
         let sender: Result<Identity, AuthError> = msg_sender();
```
mohammadfawaz, line 92: We still need to extract the `Address` from `msg_sender()` because the `balances` expects an `Address` as its key, right? We probably still need the `match` statement but without the `assert`. Something like:
```
        let sender: Address = match sender.unwrap() {
            Identity::Address(addr) => {
                addr
            },
            _ => {
                revert(0);
            },
        };
```

mohammadfawaz (changes requested):
Thank you for your contribution!

examples/subcurrency/src/main.sw
joerlop: Thanks! Just added the requested changes.

mohammadfawaz (approved):
Looks good now! Thank you :)

otrho (approved):

sezna (approved):

PR #2516, opened 2022-08-11
Add Outer DocComment Parsing

AlicanC:
This PR extends the parser so that the following:
```rs
/// This is a doc comment.
/// This is another doc comment.
fn f() -> bool {
    false
}
```
is basically parsed like:
```rs
#[doc( This is a doc comment.)]
#[doc( This is another doc comment.)]
fn f() -> bool {
    false
}
```

Notes:
- Corners were cut to unblock the rest of the work.
- Inner DocComment (`//!`) support was left for later. The lexer has commented out code for supporting these in the future.
- Enum and Struct fields (`TypeField`s) aren't annotated ATM so they can't have DocComments. Since we do have code that uses DocComments for these fields, we started getting errors. As a temporary solution I made the parser ignore these DocComments. When we start supporting annotations for `TypeField`s, we can remove the temp code and everything should work fine.
- There's some ugliness in the `DocComment` to `AttributeDecl` conversion code. Adding `Literal` support for `Attribute` args or having two types of attributes, `Normal` and `DocComment`, [like Rust](https://github.com/rust-lang/rust/blob/aeb5067967ef58e4a324b19dd0dba2f385d5959f/compiler/rustc_ast/src/ast.rs#L2551-L2560) could help.

### DIFF
--- a/sway-ast/src/token.rs
+++ b/sway-ast/src/token.rs
@@ -119,6 +119,24 @@ impl Spanned for Comment {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Hash)]
+pub enum DocStyle {
+    Outer,
+    Inner,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Hash)]
+pub struct DocComment {
+    pub span: Span,
+    pub doc_style: DocStyle,
+}
+
+impl Spanned for DocComment {
+    fn span(&self) -> Span {
+        self.span.clone()
+    }
+}
+
 /// Allows for generalizing over commented and uncommented token streams.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Hash)]
 pub enum GenericTokenTree<T> {
@@ -126,6 +144,7 @@ pub enum GenericTokenTree<T> {
     Ident(Ident),
     Group(GenericGroup<T>),
     Literal(Literal),
+    DocComment(DocComment),
 }
 
 pub type TokenTree = GenericTokenTree<TokenStream>;
@@ -154,6 +173,7 @@ impl<T> Spanned for GenericTokenTree<T> {
             Self::Ident(ident) => ident.span(),
             Self::Group(group) => group.span(),
             Self::Literal(literal) => literal.span(),
+            Self::DocComment(doc_comment) => doc_comment.span(),
         }
     }
 }
@@ -188,6 +208,12 @@ impl<T> From<GenericGroup<T>> for GenericTokenTree<T> {
 impl<T> From<Literal> for GenericTokenTree<T> {
     fn from(lit: Literal) -> Self {
         Self::Literal(lit)
+    }
+}
+
+impl<T> From<DocComment> for GenericTokenTree<T> {
+    fn from(doc_comment: DocComment) -> Self {
+        Self::DocComment(doc_comment)
     }
 }
 
@@ -307,6 +333,7 @@ impl CommentedTokenTree {
             CommentedTree::Ident(ident) => ident.into(),
             CommentedTree::Group(group) => group.strip_comments().into(),
             CommentedTree::Literal(lit) => lit.into(),
+            CommentedTree::DocComment(doc_comment) => doc_comment.into(),
         };
         Some(tt)
     }
--- a/sway-ast/src/token.rs
+++ b/sway-ast/src/token.rs
@@ -128,6 +128,7 @@ pub enum DocStyle {
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Hash)]
 pub struct DocComment {
     pub span: Span,
+    pub content_span: Span,
     pub doc_style: DocStyle,
 }
 
--- a/sway-parse/src/attribute.rs
+++ b/sway-parse/src/attribute.rs
@@ -30,7 +30,7 @@ impl<T: Parse> Parse for Annotated<T> {
         while let Some(DocComment {
             doc_style: DocStyle::Outer,
             ..
-        }) = parser.peek::<DocComment>()
+        }) = parser.peek()
         {
             let doc_comment = parser.parse::<DocComment>()?;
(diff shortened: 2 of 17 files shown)


sway-ast/src/token.rs
```diff
@@ -119,13 +119,32 @@ impl Spanned for Comment {
...
+}
+
 /// Allows for generalizing over commented and uncommented token streams.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Hash)]
 pub enum GenericTokenTree<T> {
     Punct(Punct),
     Ident(Ident),
     Group(GenericGroup<T>),
     Literal(Literal),
+    DocComment(DocComment), // <- dis new
```
AlicanC, line 147: Inspired from Rust: https://github.com/rust-lang/rust/blob/792bc5a0102d0973d42183a2b267850bb905236f/compiler/rustc_ast/src/token.rs#L254

AlicanC, line 147: Inspired from Rust: https://github.com/rust-lang/rust/blob/792bc5a0102d0973d42183a2b267850bb905236f/compiler/rustc_ast/src/token.rs#L254

sway-parse/src/attribute.rs
```diff
@@ -1,13 +1,78 @@
...
+use sway_ast::token::{DocComment, DocStyle};
+use sway_types::{Ident, Span};
+
+impl Peek for DocComment {
+    fn peek(peeker: Peeker<'_>) -> Option<DocComment> {
+        peeker
+            .peek_doc_comment()
+            .ok()
+            .map(|doc_comment| doc_comment.clone())
+        // .map(DocComment::clone)
```
AlicanC, line 17: `.map(DocComment::clone)` did not work, but Clippy still wants me to do it. Need to have a look.

sway-parse/src/attribute.rs
```diff
@@ -1,13 +1,78 @@
...
+        }
+    }
+}
 
 impl<T: Parse> Parse for Annotated<T> {
     fn parse(parser: &mut Parser) -> ParseResult<Self> {
         let mut attribute_list = Vec::new();
+
+        // Parse outer doc comment lines
+        let mut doc_lines = Vec::new();
```
AlicanC, line 35: Maybe instead of collecting lines here, we should just build the final string?

sway-parse/src/attribute.rs
```diff
@@ -1,13 +1,78 @@
...
+                parser.peek::<DocComment>(),
+                Some(DocComment {
+                    doc_style: DocStyle::Outer,
+                    ..
+                })
+            ) {
+                let doc_comment = parser.parse::<DocComment>()?;
+                let span = doc_comment.span.clone().as_str().to_string();
+                let line = span.trim_start_matches("///").to_string();
+                let line = line.trim().to_string();
```
AlicanC, line 47: Ended up with this ugliness trying to make the borrow checker happy. This needs cleaning up.

sway-parse/src/attribute.rs
```diff
@@ -1,13 +1,78 @@
...
+                doc_lines.push(line);
+            } else {
+                break;
+            }
+        }
+
+        // Join doc comments and add as an AttributeDecl
```
AlicanC, line 54: Should we even merge these? Maybe every doc comment should be a separate attribute?

sway-parse/src/attribute.rs
```diff
@@ -1,13 +1,78 @@
...
+        if !doc_lines.is_empty() {
+            let doc = doc_lines.join(" ");
+            let doc: &'static str = Box::leak(doc.into_boxed_str());
```
AlicanC, line 57: This doesn't look like the right thing to do.

sway-parse/src/attribute.rs
```diff
@@ -1,13 +1,78 @@
...
+                                value_separator_pairs: vec![],
+                                final_value_opt: Some(Box::new(Ident::new_no_span(doc))),
+                            },
+                            Span::dummy(),
+                        )),
+                    },
+                    Span::dummy(),
+                ),
+            });
+        }
```
AlicanC, line 74: This is looks messy. Maybe we should have two attribute kinds like Rust: `Normal` and `DocComment`

https://github.com/rust-lang/rust/blob/aeb5067967ef58e4a324b19dd0dba2f385d5959f/compiler/rustc_ast/src/ast.rs#L2551-L2560

sway-parse/src/item/mod.rs
```diff
@@ -201,6 +201,37 @@ mod tests {
         }
     }
 
+    #[test]
+    fn parse_doc_string() {
```
AlicanC, line 205: Rust calls them "doc comments" so rename to `parse_doc_comment`.

sway-parse/src/attribute.rs
```diff
@@ -1,13 +1,78 @@
...
+            let doc: &'static str = Box::leak(doc.into_boxed_str());
+            attribute_list.push(AttributeDecl {
+                hash_token: HashToken::new(Span::dummy()),
+                attribute: SquareBrackets::new(
+                    Attribute {
+                        name: Ident::new_no_span("doc"),
+                        args: Some(Parens::new(
+                            Punctuated {
+                                value_separator_pairs: vec![],
+                                final_value_opt: Some(Box::new(Ident::new_no_span(doc))),
```
AlicanC, line 66: Turning the literal to `Ident` here. Hacky.

sway-parse/src/token.rs
```diff
@@ -891,4 +914,53 @@ mod tests {
         }
         assert!(tts.next().is_none());
     }
+
+    #[test]
+    fn lex_doc_strings() {
```
AlicanC, line 919: `lex_doc_comments`

sway-parse/src/attribute.rs
AlicanC: Seems like Rust doesn't so not merging.

sezna (approved):

Issue #2455, opened 2022-08-03
`while` loop "not allowed outside of blocks"


Braqzen:
Is this a bug?
What is a "block" here?

```rust
contract;

abi MyContract {
    fn test_function() -> bool;
}

impl MyContract for Contract {
    fn test_function() -> bool {
        let mut counter = 0;
        let truth = false;
        let result = while counter < 10 {
            counter += 1;
            true
        };
        result
    }
}
```

![1](https://user-images.githubusercontent.com/103777923/182673464-06293591-ec90-40e6-84e7-5a95d2289e77.png)

mohammadfawaz:
Yeah don't get that either. Marking this as critical. Either this should work or it should emit a much better error message. And I don't see why it shouldn't work...

otrho:
Are `while` loops expressions?  I was under the impression their bodies couldn't return a type.  :thinking:  If they turn out to be statements then only allowing them as direct children of code blocks makes sense, but that shouldn't be the error message.

otrho:
Personally I don't like the idea of having statements as a special case, but having an expression which must return `()` is the same, but better because it can type-check.  So I think what Rust does is right; the body must have the type `()` and so does the `while` loop.

PR #2517, opened 2022-08-11
Convert while loops to be an expression instead of a statement

tritao:
This commit changes while loops to be handled as an expression instead of as a statement.

This means while loops now return the unit type `()` as their return type.

And means the following code now compiles:

```rust
fn main() {
    let result = while true { break; };
}
```

Which is also allowed in Rust.

Closes https://github.com/FuelLabs/sway/issues/2455.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -119,6 +119,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "eab1c04a571841102f5345a8fc0f6bb3d31c315dec879b5c6e42e40ce7ffa34e"
 
 [[package]]
+name = "assert_matches"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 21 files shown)


sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -1582,6 +1605,55 @@ impl TypedExpression {
...
+            ),
+            warnings,
+            errors
+        );
+        let exp = TypedExpression {
+            expression: TypedExpressionVariant::WhileLoop {
+                condition: Box::new(typed_condition),
+                body: typed_body,
+            },
+            return_type: block_implicit_return,
```
mohammadfawaz, line 1650: Shouldn't this return type be just `()`?

mohammadfawaz: Oh so `block_implicit_return` is always `TypeInfo::Tuple(Vec::new())`?

mohammadfawaz (approved):
Excellent. Looks good to me! Thanks.

sezna (approved):

PR #2518, opened 2022-08-11
Fix new clippy lints (with rust version 1.63)

kayagokalp:
With rust version 1.63 released, we have some new clippy warnings in our CI.


sezna (approved):
thanks for spotting this!

mohammadfawaz (approved):
Great! Thanks.

PR #2519, opened 2022-08-11
Temporarily disable the 3 LSP tests that open a file

JoshuaBatty:
CI seems to be hanging on LSP tests that open a file. 

Disabling them for now in order to unblock the sway repo. Will look at re-adding them back in and figuring out what is causing CI to hang when I'm back on Monday.


sezna (approved):

mitchmindtree (approved):
So weird. Maybe also open an issue that links to this so we can remember to re-add?

PR #2520, opened 2022-08-12
Small cleanup / refactoring for `forc-client` plugin PR

mitchmindtree:
Based on #2440.

This removes some unused deps from `forc` and refactors `mdbook-forc-documenter` a little for clearer "plugin" and "command" terminology. Also aims to improve the readability a little of the content injection section within lib.rs, and reorders the "Plugins" and "Commands" sections to match the order of the book.

**Edit**: OK, it looks like my refactoring is causing some issues with mdbook generation! I've removed the mdbook refactor commit in favour of opening a dedicated PR in the future. For now it's best we just land these necessities to unblock #2440.


bingcicle (approved):
did a brief look through to confirm deps are indeed unused in `forc`.

kayagokalp (approved):

PR #2523, opened 2022-08-12
Fix: `Impl` formatting without generics generates unparsable output

### LINKED ISSUE #2522: `Impl` formatting without generics generates unparsable output
Trying to format the following code piece:

```rust
script;

struct Foo {
    bar: u64,
    baz: bool,
}

trait Qux {
    fn is_baz_true(self) -> bool;
}

impl Qux  for  Foo {
fn is_baz_true(self) -> bool {
        self.baz
    }
}
```

Fails in comment addition phase because it tries to parse the formatted code generated by the formatter. Further investigation showed that the formatter currently formats the given code piece to the following form:

```rust
script;

struct Foo {
    bar: u64,
    baz: bool,
}
trait Qux {
    fn is_baz_true(self) -> bool;
}

implQux for Foo {
    fn is_baz_true(self) -> bool {
        self.baz
    }
}
```

Current tests only cover with generics case in which this problem does not happen.

kayagokalp:
closes #2522.


eureka-cpu (approved):
Gah, Idk how I missed this, thanks!

JoshuaBatty (approved):

PR #2524, opened 2022-08-12
New JSON ABI format supporting generics

### LINKED ISSUE #2386: Generalize the handling of generics in JSON ABI
https://github.com/FuelLabs/sway/pull/2218 introduces the `typeArgument` field to the JSON ABI which got us one step closer to supporting (resolved) generics in the ABI. However, as @iqdecay  pointed out, this doesn't solve the full problem because even though we know what type parameters a given struct/enum has and what they resolve to, we do not know what generic parameter each component actually uses. For example:

```rust
struct MyStruct1<T, U> {
    x: T,
    y: U
}

struct MyStruct2<T, U> {
    x: U,
    y: T
}

fn some_abi_funct(s1: MyStruct1<u64, u64>, s2: MyStruct2<u64, u64>);
```

Both structs above would be represented in the same way in the JSON ABI (except for the name of course) and that's ambiguous.

The `typeArguments` field gave us half the solution. The other half is that we should modify the `type` field to contain the type parameter applied. For example, the first input would be represented as follows:
```json
      {
        "name": "s1",
        "type": "struct MyStruct1<T, U>",
        "components": [
          {
            "name": "x",
            "type": "T",
            "components": null,
            "typeArguments": null
          },
          {
            "name": "y",
            "type": "U",
            "components": null,
            "typeArguments": null
          }
        ],
        "typeArguments": [
          {
            "name": "T",
            "type": "u64",
            "components": null,
            "typeArguments": null
          },
          {
            "name": "U",
            "type": "u64",
            "components": null,
            "typeArguments": null
          }
        ]
      },
```
The above has all the information we need to recreate the generic struct declaration and also recreate its specialization (i.e. `MyStruct1<u64, u64>`).

This gets a bit more complicated when the field of a struct is a complex type that _uses_ one of the generics. For example:

```rust
struct MyStruct2<T, U> {
    x: U,
    y: T
}

struct MyStruct3<V, W> {
    x:  MyStruct2<V, W>,
}

fn some_other_abi_funct(s3: MyStruct3<bool, u64>);
```
but the same concepts still applies:

```json
      {
        "name": "s1",
        "type": "struct MyStruct3<V, W>",
        "components": [
          {
            "name": "x",
            "type": "struct MyStruct2<V, W>",
            "components": [
              {
                "name": "x",
                "type": "U",
                "components": null,
                "typeArguments": null
              },
              {
                "name": "y",
                "type": "T",
                "components": null,
                "typeArguments": null
              }
            ],
            "typeArguments": [
              {
                "name": "T",
                "type": "u64",
                "components": null,
                "typeArguments": null
              },
              {
                "name": "U",
                "type": "u64",
                "components": null,
                "typeArguments": null
              }
            ]
          }
        ],
        "typeArguments": [
          {
            "name": "V",
            "type": "u64",
            "components": null,
            "typeArguments": null
          },
          {
            "name": "W",
            "type": "u64",
            "components": null,
            "typeArguments": null
          }
        ]
      }
```

This introduces no changes to non-generic types so the current working flows continue to work as expected.

mohammadfawaz:
Closes #2386 

The main changes are as follows:
- Introduce `initial_type_id` or `initial_return_type` in various places in the type system. This is needed because the original type information (potentially generic) is often lost and cannot be inferred at the end of the type checking phase. Keeping that information around helps generate the JSON ABI information needed in a much cleaner way.
- New data structures for the new JSON ABI format in `sway-types`
- Generate the new JSON ABI format... Most of the work in the `type_system` directory.
- Currently generate both the old and the new JSON ABI files. We will quickly deprecate the old one as soon as the SDKs are ready to make the switch (hopefully within a week or two)
- Added testing for contract tests for now. The tests show how the new JSON looks like. Again, most of the old stuff will be cleaned up and replaced with the new stuff.
- In `forc-pkg`, I added code to standardize the output of the JSON by 
  - Removing duplicate types
  - Ordering the types in alphabetical order
  - Assigning new IDs to the types according to the alphabetical order above.

The standardization of the output is not really required but helps make the JSON more compact and more robust for testing in case of future changes in the type system, given that the type IDs that the type engine hands out are arbitrary (deterministic but arbitrary)


mohammadfawaz:
This is nearly done but needs some tests. For now, the SDK team can experiment with the `JSON` file generated. The old one is still there... the new one is called `<proj_name>-flat-abi.json`.

mohammadfawaz:
Need to fix something related to the stable ordering of the output. Turning into a draft temporarily.

luizstacio:
@mohammadfawaz some notes from testing this PR.  On the examples we created, `typeArguments` was an object which looked cleaner by just having the `typeArguments: [4, 5]`. 

What are the motivations for having the object instead?

mohammadfawaz:
> @mohammadfawaz On the examples, we created, `typeArguments` was an object which looked cleaner by just having the `typeArguments: [4, 5]`.
> 
> What are the motivations for having the object instead?

Good question! It turns out that this is not enough. A type argument can be complex such as `MyStruct<MyEnum<str[4]>>` for example. Here, `MyEnum<str[4]>` has to be a `JsonTypeApplication` and not a `JsonTypeDeclaration`, and it's not enough to pass the ID of `MyEnum<T>` for example.

forc-pkg/src/pkg.rs
```diff
@@ -1907,22 +1934,149 @@ pub fn build(plan: &BuildPlan, profile: &BuildProfile) -> anyhow::Result<(Compil
...
+        json_abi_program,
         storage_slots,
         tree_type,
     };
     Ok((compiled, source_map))
 }
 
+/// Standardize the JSON ABI data structure by eliminating duplicate types. This is an iterative
+/// process because every time two types are merged, new opportunities for more merging arise.
+fn standardize_json_abi_types(json_abi_program: &mut JsonABIProgram) {
```
emilyaherbert, line 1965: Good catch, I didn't think about this possibility... it makes me want to implement this in the `TypeEngine` 😄

emilyaherbert (approved):
so beautiful... so flat! 🙌 

In a future PR, let's make sure to write a blurb for the book documenting the new design of the ABI.

mohammadfawaz:
> In a future PR, let's make sure to write a blurb for the book documenting the new design of the ABI.

100%. I will start by adding the new format to the spec (https://github.com/FuelLabs/fuel-specs/issues/401) and then go from there.

digorithm (approved):
Approving on behalf of the SDK team (and regarding SDK/ABI-related stuff only); All tests with the new experimental `fuels-rs` using this new `flat-abi.json` outputted from a local copy of this branch (including `sway`'s SDK harness tests) are passing!

sezna (approved):

PR #2525, opened 2022-08-12
Bump deps in sway repo to switch to fuel-core v0.10.1

nfurfaro:
Bumps a bunch of deps and get us ready for fuel-core v0.10 !


Braqzen (commented):
Some points

1. I assume all of those OPCODES are now obsolete. I think it would be helpful to add some issues which removed them in their respective repos so that anyone that reviews this PR can have an easier time in figuring out why bumping deps includes the removal of some OPCODES.
2. I don't really understand the difference in the harness test for the transaction but if it works then it works?
3. I'm wondering if this requires the default forc generated tests etc. to be updated. If the CI passes then I guess not however I have recently been having some issues manually bumping outdated SDK tests in sway-apps to v0.20 of the SDK which involved some config changes in e.g. the Tx repo and perhaps others.

mohammadfawaz:
Thanks @nfurfaro! Looks great. Two things:
1. We should also remove the dead opcodes from https://github.com/FuelLabs/sway/blob/master/sway-ast/src/expr/op_code.rs
2. If we do the above, I suspect the standard library won't compile anymore because https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/tx.sw uses `xis` and `xos` for example and potentially the other opcodes are used somewhere else. In fact, does the standard library compiles with your change as is?

In light of the above, I'm worried we cannot actually bump to `fuel-core 0.10.1` without supporting `gtf` first and replacing all uses of `xis`, `xos`, etc. with uses of `gtf`. What do you think?

AlicanC:
This is a soft-blocker for #845 since the changes in this PR adds the necessary support for GTF. The issue can be solved without GTF but GTF is a lot cleaner.

nfurfaro:
Closing as I had to include these changes with https://github.com/FuelLabs/sway/pull/2482.

PR #2528, opened 2022-08-13
Annotate TypeFields

AlicanC:
This PR adds support for attaching Attributes to fields of enums and structs.

The primary motivation was to remove this hack introduced in https://github.com/FuelLabs/sway/pull/2516:
https://github.com/FuelLabs/sway/blob/afb59ca69615e9a1e65ea54a4ae3c6f3e357ba64/sway-parse/src/item/mod.rs#L78-L96

The second commit is for fixing the issues in fmt-v2. The tests pass, but correct formatting is left for another PR as we did in #2516.


kayagokalp (dismissed):
This looks really good to me. Can't wait `forc doc` with these doc comment support you are brewing lately🧪

 About the formatter changes I left some suggestions but none of them are blockers as we will _probably_ address this kind of changes in a future refactor (I did actually thought about this today). So all good to me like this!

The thing is currently the places we are adding the indentations is a little unorganized  throughout `sway-fmt-v2`. I think it would be better for us to have the addition of indentations inside the `format()` for each applicable `Format` implementation rather than adding them before calling `format()` on that item throughout the codebase. I suggested some changes to this PR, since doing it like that would be shorter and keep this one more focused. On top of the suggested changes, for tests to pass correctly we should add the indentation addition at the beginning of `Format` implementation for `FnSignature`

```rust
write!(
  formatted_code,
  "{}",
  formatter.shape.indent.to_string(&formatter.config)?,
)?;
```
to:

https://github.com/FuelLabs/sway/blob/fddb2067dd1b382b507bae76d0258948dca1331a/sway-fmt-v2/src/items/item_fn.rs#L77-L83

kayagokalp (approved):

sezna (approved):

PR #2529, opened 2022-08-13
fix: Setting the contract ID in an abi cast to a storage access fails to compile

### LINKED ISSUE #2490: Setting the contract ID in an `abi` cast to a storage access fails to compile
Example:
```rust
contract;

abi MyContract {
    fn test_function();
}

abi MyContract2 {
    #[storage(read)]
    fn test_function2();
}

storage {
    b: b256 = 0x0000000000000000000000000000000000000000000000000000000000000000,
}

impl MyContract2 for Contract {
    #[storage(read)]
    fn test_function2() {
        let caller = abi(MyContract, storage.b);
        caller.test_function();
    } 
}
```
Fails with:
```console
   |                ^^^^^^^^^^^^^ No method named "test_function" found for type "contract caller MyContract".
```
However, storing `storage.b` in a local variable first works fine:
```rust
// ...
   let local_b = storage.b;
   let caller = abi(MyContract, local_b);
// ...
```

kayagokalp:
closes #2490

After this PR

```rust
contract;

abi MyContract {
    fn test_function();
}

abi MyContract2 {
    #[storage(read)]
    fn test_function2();
}

storage {
    b: b256 = 0x0000000000000000000000000000000000000000000000000000000000000000,
}

impl MyContract2 for Contract {
    #[storage(read)]
    fn test_function2() {
        let caller = abi(MyContract, storage.b);
        caller.test_function();
    } 
}
```

compiles successfully. The problem was in the implantation of `PartialEq` for `TypedExpressionVariant` which did not have specific match case for `StorageAccess` thus it was always returning false.

I left a review regarding the validity of my approach 👀. Adding critical as the associated issue has the critical tag.


sway-core/src/semantic_analysis/ast_node/expression/typed_expression_variant.rs
```diff
@@ -559,7 +559,7 @@ impl TypeCheckedStorageAccess {
 }
 
 /// Describes a single subfield access in the sequence when accessing a subfield within storage.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq)]
```
kayagokalp, line 562: I am not sure if this is the right approach. Should I use `look_up_type_id` and implement `PartialEq` myself?

mohammadfawaz: I think this perfectly fine!

mohammadfawaz (approved):
Thanks! Great catch 😄

PR #2532, opened 2022-08-15
Adds `control_flow` items to `sfv2`

### LINKED ISSUE #2118: Handle `break` and `continue` in `sway-fmt-v2`
https://github.com/FuelLabs/sway/pull/2112/files introduces `break` and `continue` but doesn't do anything with them in `sway-fmt-v2`.

eureka-cpu:
- [x] Add tests

Closes #2118


sezna:
They should have a semicolon, although it is technically possible to omit the semicolon if `break` or `continue` are the last item in a function which returns the unit type, since you'd be implicitly returning `()` which is therefore valid.

tl;dr semicolon required unless in implicit return position

Braqzen (approved):

sezna (approved):

PR #2534, opened 2022-08-15
Create markdown-lint.yml

adlerjohn:
Add markdownlint check to CI.


otrho (approved):

docs/src/common-collections/storage_map.md
```diff
@@ -34,9 +34,12 @@ To insert key-value pairs into a storage map, we can use the `insert` method, as
 
 Note two details here. First, in order to use `insert`, we need to first access the storage map using the `storage` keyword. Second, because `insert` requires _writing_ into storage, a `#[storage(write)]` annotation is required on the ABI function that calls `insert`.
 
-> **Note**: The storage annotation is also required for any private function defined in the contract that tries to insert into the map.
+> **Note**
```
mohammadfawaz, line 37: Does the linter not like the colon after `**Note**`? Without it, the note looks a bit weird:
image
Looking at the Rust book, notes that start with `Note` do have a colon.

adlerjohn: Note strictly, no. It's to make it look like this on GitHub

note

and the above screenshot on Mdbook.

mohammadfawaz: But shouldn't we care more about how it renders on Mdbook? It's a shame that book does not render notes the same was as Github 😞

adlerjohn: The book still renders them as a box, instead of just quote text, with the changed format.

mohammadfawaz: Yeah, I am just more concerned with the lack of the colon (`:`) after `Note` which makes it look weird IMO, but maybe the fact that `Note` is in bold make that acceptable.

mohammadfawaz (approved):

PR #2535, opened 2022-08-15
Reassignment expressions

canndrew:
Make reassignments an expression of type `()` rather than a declaration.


canndrew:
This code has a bunch of FIXMEs in places where I haven't changed the current behaviour but the current behaviour looks wrong. I'm going to point them out individually so I can get some feedback on them.

sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs
```diff
@@ -257,6 +257,11 @@ impl TypedImplTrait {
                     kind,
                     ..
                 }) => matches!(kind, sway_ast::intrinsics::Intrinsic::GetStorageKey),
+                TypedExpressionVariant::Reassignment(reassignment) => {
+                    expr_contains_get_storage_index(&reassignment.rhs)
+                }
+                // FIXME: shouldn't this check the rhs?
```
canndrew, line 263: This is in `fn expr_contains_get_storage_index`. Do we need to check the rhs of a storage assignment for `GetStorageKey`?

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -282,6 +297,14 @@ impl DeterministicallyAborts for TypedExpression {
             AbiName(_) => false,
             EnumTag { exp } => exp.deterministically_aborts(),
             UnsafeDowncast { exp, .. } => exp.deterministically_aborts(),
+            Reassignment(_reassignment) => {
+                // FIXME: This seems very wrong. We should check rhs, no?
+                false
+            }
+            StorageReassignment(_storage_reassignment) => {
+                // FIXME: This seems very wrong. We should check rhs, no?
+                false
```
canndrew, line 306: This is in `fn deterministically_aborts`. We always return `false` for assignments, but if the rhs of an assignment deterministically aborts then the whole expression does. So we should check rhs here, yeah?

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -367,8 +390,18 @@ impl TypedExpression {
                 }
                 buf
             }
+            TypedExpressionVariant::Reassignment(reassignment) => {
+                reassignment.rhs.gather_return_statements()
+            }
+            TypedExpressionVariant::StorageReassignment(_storage_reassignment) => {
+                // FIXME: we should check rhs here too, no?
+                vec![]
```
canndrew, line 398: This is in `fn gather_return_statements`. We don't check the rhs of storage assignments for return statements.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -367,8 +390,18 @@ impl TypedExpression {
...
+            }
             // if it is impossible for an expression to contain a return _statement_ (not an
             // implicit return!), put it in the pattern below.
+
+            // FIXME: lots of these expression kinds can contain arbitrary sub-expressions. But
+            // we're not checking those for returns?
```
canndrew, line 404: We also don't check any of the expression kinds below for return statements, even though many of them could contain return statements in sub-expressions.

sway-core/src/semantic_analysis/storage_only_types.rs
```diff
@@ -109,6 +109,21 @@ fn expr_validate(expr: &TypedExpression) -> CompileResult<()> {
...
+            check!(
+                check_type(rhs.return_type, lhs_base_name.span(), false),
+                (),
+                warnings,
+                errors,
+            );
+            check!(expr_validate(rhs), (), warnings, errors)
+        }
+        TypedExpressionVariant::StorageReassignment(_storage_reassignment) => {
+            // FIXME: there's nothing to check here?
```
canndrew, line 125: This is in `storage_only_types::expr_validate`. I'm not sure what this function is doing, but given we recurse into non-storage assignments for validation it seems like we should have to recurse into storage assignments to, no?

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -1582,6 +1618,131 @@ impl TypedExpression {
...
+                            TypedReassignment {
+                                lhs_base_name: base_name,
+                                lhs_type: final_return_type,
+                                lhs_indices: names_vec,
+                                rhs,
+                            },
+                        )),
+                        return_type: crate::type_system::insert_type(TypeInfo::Tuple(Vec::new())),
+                        // TODO: if the rhs is constant then this should be constant, no?
+                        is_constant: IsConstant::No,
```
canndrew, line 1713: Reassignment expressions are always treated as being non-const. But I'm pretty sure we could just take the const-ness of the rhs (?)

otrho:
This generally LGTM, but I'm not really sure on those issues you've pointed out.  I think there's a restriction on what the RHS of a storage reassignment can be, but off the top of my head I don't know for sure.  But it might explain why we're not so rigorous with them.  @mohammadfawaz might know more?

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
sezna: yeah, technically the RHS could be a code block with a return in it, so we should check that.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
sezna: Ditto on my above comment:
> yeah, technically the RHS could be a code block with a return in it, so we should check that.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
sezna: yeah, but I think const evaluation actually doesn't rely on this after all. cc @vaivaswatha -- this could perhaps be removed entirely.

sezna (dismissed):
I'm approving for now, but would you mind making an issue that summarizes all of the FIXMEs you've pointed out? Or just handling them in this PR? Either works, but the former will probably result in less conflicts.

sway-core/src/semantic_analysis/storage_only_types.rs
vaivaswatha: This function is checking that there are no declarations of "storage-only" types. I don't know if syntactically we can have declarations inside here (but semantically, as per the AST definition, we can). So it wouldn't harm to recurse, but I'm not sure if it's necessary.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
vaivaswatha: > yeah, but I think const evaluation actually doesn't rely on this after all. cc @vaivaswatha -- this could perhaps be removed entirely.

That's right. Constant evaluation doesn't worry about `is_constant`.

vaivaswatha: And yes, from what I can tell, `is_constant` shouldn't be needed anymore at all.

sway-core/src/semantic_analysis/storage_only_types.rs
vaivaswatha: > This function is checking that there are no declarations of "storage-only" types. I don't know if syntactically we can have declarations inside here (but semantically, as per the AST definition, we can). So it wouldn't harm to recurse, but I'm not sure if it's necessary.

Based on https://github.com/FuelLabs/sway/pull/2535#discussion_r949053844, I change my stance from "it wouldn't harm to recurse" to "we must recurse".

sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs
vaivaswatha: Yes, since it could contain any expression (as per the AST definition) including a `get_storage_index`.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
sezna: Ok, created #2572 to address this. Therefore we can resolve this conversation as `IsConstant` is dead code.

mohammadfawaz (dismissed):
Thanks! Just re-emphasizing Alex's point to create issues for the FIXMEs.

canndrew:
I've gone ahead and fixed the FIXMEs here, since it wasn't much work. I've rebased off of master too, so this should be good to go.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
canndrew: I've modified this to also recurse into all the sub-expression kinds that weren't being recursed into.

sezna (approved):

otrho (approved):

sezna (approved):

otrho (approved):

PR #2537, opened 2022-08-15
feat: `sway-fmt-v2` preserves newline with given threshold

### LINKED ISSUE #2379: `sway-fmt-v2` should be able to preserve whitespace
Currently all newlines aren't handled, so something like:
```rust
let foo = "foo";

let bar = "bar";

let baz = "baz";
```

will format to: 
```rust
let foo = "foo";
let bar = "bar";
let baz = "baz";
```

I think we could add some kind of threshold with the default being 2, that keeps track of newlines and inserts them similar to how we handle inserting comments.

kayagokalp:
~Opening up for visibility.~
closes #2379.

## TODO:

- [x] Move threshold to config
- [x] Add comma


eureka-cpu:
Huge 🙌🏻 Very excited to get this in

sway-fmt-v2/src/fmt.rs
```diff
@@ -855,4 +863,37 @@ impl Qux for Foo {
...
+    let number3: u64 = 30;
+
+
+
+}"#;
+
+        let correct_sway_code = r#"script;
+
+fn main() {
+    let number: u64 = 10;
```
eureka-cpu, line 887: Wouldn't we want to retain the extra space after this line? and if there are more lines than the threshold those extra lines would be removed?

eureka-cpu (approved):
Fantastic work, thanks so much @kayagokalp !

JoshuaBatty (approved):
:shipit:

Issue #2492, opened 2022-08-09
`self` should not be allowed as a parameter of a free function


mohammadfawaz:
Example:
```rust
script;

fn main() {}
fn foo(self) {}
```
Rust errors out with: `^^^^^ not semantically valid as function parameter` but we do not error out.

PR #2538, opened 2022-08-15
Check for self parameters in free functions.

tritao:
This checks and emits a new error in the case of self parameters in free functions.

The following code:

```rust
script;

fn free_fn(self) {}
fn main() {}
```

Will now emit the following error:

```
error
 --> main.sw:3:12
  |
1 |
2 |
3 | fn free_fn(self) {}
  |            ^^^^ not semantically valid as function parameter
4 |
5 | fn main() {}
  |
____

```

Closes https://github.com/FuelLabs/sway/issues/2492.


vaivaswatha (approved):
The change itself looks ok to me, but I'm not sure of the error message (and the enum name for it). Sounds vague to me. Maybe someone else has a better suggestion. (I'm okay merging this as-is too, hence the approval).

mohammadfawaz:
> The change itself looks ok to me, but I'm not sure of the error message (and the enum name for it). Sounds vague to me. Maybe someone else has a better suggestion. (I'm okay merging this as-is too, hence the approval).

The error message was proposed because it's what `rustc` emits. We could improve if we want though!

mohammadfawaz (dismissed):

mohammadfawaz (approved):

otrho (approved):

PR #2545, opened 2022-08-16
Update 'Sway toolchain' references to 'Fuel toolchain' where appropriate

bingcicle:
This PR updates several instances of references to the 'Sway toolchain' to the 'Fuel toolchain'.

For some context: created this PR as a result of working on `fuelup` and realizing that I started calling it the _Fuel toolchain_ there (to refer to `sway` + `fuel-core`), saw that in the Sway book it was the _Sway toolchain_ but this seems to be referring to `sway` alone.

Note that several instances were left untouched - places where the content was mostly talking about  `sway` only, I left the _Sway toolchain_ references alone; places where the content was mostly described in the context of working with both `sway` + `fuel-core`, I updated them to be _Fuel toolchain_ instead.

Very minor difference but I think it's worth the differentiation between the two, unless there's a better way to call `sway` instead of _the Sway toolchain_.

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -5,7 +5,7 @@
 - [Introduction](./introduction/index.md)
   - [Installation](./introduction/installation.md)
   - [Getting Started](./introduction/overview.md)
-  - [The Sway Toolchain](./introduction/sway-toolchain.md)
+  - [The Fuel Toolchain](./introduction/fuel-toolchain.md)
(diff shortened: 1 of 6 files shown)


sezna (approved):
While someday we may want to differentiate sway from fuel in the docs, as we will support multiple backends, today is not that day. Since we only support a single back end, I think it's good to unify our terminology here. Thanks for catching that!

kayagokalp (approved):
lgtm 👍

PR #2546, opened 2022-08-16
Break/continue expressions

canndrew:
Make `break` and `continue` expressions rather than declarations in the compiler's internal representation of the AST.

### DIFF
--- a/sway-core/src/control_flow_analysis/analyze_return_paths.rs
+++ b/sway-core/src/control_flow_analysis/analyze_return_paths.rs
@@ -15,7 +15,7 @@ impl ControlFlowGraph {
         // do a depth first traversal and cover individual inner ast nodes
         let mut leaves = vec![];
         for ast_entrypoint in module_nodes {
-            let l_leaves = connect_node(ast_entrypoint, &mut graph, &leaves, None, None)?.0;
+            let l_leaves = connect_node(ast_entrypoint, &mut graph, &leaves)?.0;
             if let NodeConnection::NextStep(nodes) = l_leaves {
                 leaves = nodes;
             }
@@ -121,8 +121,6 @@ fn connect_node(
     node: &TypedAstNode,
     graph: &mut ControlFlowGraph,
     leaves: &[NodeIndex],
-    break_to_node: Option<NodeIndex>,
-    continue_to_node: Option<NodeIndex>,
 ) -> Result<(NodeConnection, ReturnStatementNodes), CompileError> {
     let span = node.span.clone();
     match &node.content {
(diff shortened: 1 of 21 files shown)


sezna:
The code looks reasonable, quick question: is there a motivation for this, or some sort of documentation as to why this is the better way?

sezna (approved):
Ah I see the other PR now. #2536

otrho (approved):

sway-core/src/control_flow_analysis/dead_code_analysis.rs
```diff
@@ -1101,6 +1100,9 @@ fn connect_expression(
             }
             Ok(vec![while_loop_exit])
         }
+        // TODO: do we need to create a node here and connect the leaves to it?
```
otrho, line 1103: Dunno.

sway-core/src/control_flow_analysis/analyze_return_paths.rs
```diff
@@ -262,32 +244,6 @@ fn connect_declaration(
...
-            let entry_node = graph.add_node(node.into());
-            for leaf in leaves {
-                graph.add_edge(*leaf, entry_node, "".into());
-            }
-            match break_to_node {
-                Some(break_to_node) => {
-                    graph.add_edge(entry_node, break_to_node, "".into());
-                    Ok(vec![break_to_node])
-                }
-                None => Err(CompileError::BreakOutsideLoop { span }),
```
mohammadfawaz, line 275: I suppose this error will be emitted later in `ir_generation` so that's okay. However, are the graph connections above no longer needed? The idea was to transition the control flow to `break_to_node` or `continue_to_node` because everything after a break or continue in a given scope is essentially dead.

canndrew: This PR *shouldn't* change the current behaviour at all since the way `break` and `continue` expressions are handled is (if I made no mistakes) exactly the same as if they were enclosed in a single-declaration `CodeBlock`. I think the problem here is that `break_to_node` and `break_to_continue` weren't getting plumbed through to where they were needed anyway, so those variables ended up becoming unused and so I removed them.

sway-core/src/control_flow_analysis/dead_code_analysis.rs
sezna, line 1103: if there's no node, we can't ever report break or continue as dead code. That's _alright_, but the proper thing to do here is to connect the current leaves to the break and continue, and then return an empty vec for leaves as a result. That means that the points leading up to the break/continue are connected, and then the path breaks since everything directly after break/continue is actually dead code.

mohammadfawaz (approved):
Seems reasonable, thanks!

sway-core/src/control_flow_analysis/dead_code_analysis.rs
canndrew: @sezna: You mean like this?

https://github.com/FuelLabs/sway/commit/609f7077890718faa77e4124d05cf69fe8f0accb#diff-8877b4bcb40460d28f3808f38d7efae417662ed07fe5462b224207d08c910130R1104

Note that some of the other cases here which I looked at for guidance seem a bit off. `WhileLoop` for instance only connects `leaves[0]` and ignores any others. `Array` connects it's sub-expressions as if they're all going to be executed in parallel rather than sequentially. I didn't look at any others.

sezna: > You mean like this? 

yup LGTM

sezna (approved):

otrho (approved):

PR #2548, opened 2022-08-16
[simplify-cfg] Separate `remove_dead_blocks` from `merge_blocks`

vaivaswatha:
Currently dead block removal and block merging are both run in a loop. However, block merging cannot expose any new dead blocks (assuming dead blocks removal was run prior). So it doesn't make sense to try dead block removal again, interleaved with merge blocks.

Some performance numbers:
On a debug build, `forc build` of `test/src/e2e_vm_tests/test_programs/should_pass/test_contracts/basic_storage` takes about 14s without this change, and about 3.5s with this change (and the latter number is close to the compile time without having simplify-cfg).

On a release build, `forc build` of `test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec` takes about 25m with this change. Earlier, it was [reported](https://github.com/FuelLabs/sway/pull/2436#issue-1323901821) that this takes ~hours. @otrho  can you verify?

### DIFF
--- a/sway-ir/src/optimize/simplify_cfg.rs
+++ b/sway-ir/src/optimize/simplify_cfg.rs
@@ -20,13 +20,13 @@ pub fn simplify_cfg(context: &mut Context, function: &Function) -> Result<bool, 
             modified = true;
             continue;
         }
-
+        break;
+    }
+    loop {
         if merge_blocks(context, function)? {
             modified = true;
             continue;
         }
-
-        // Other passes here... always continue to the top if pass returns true.
         break;
     }
     Ok(modified)


otrho (approved):

sezna (approved):

otrho:
Yep, good catch.  I kinda copy pasted that loop from other passes, as it follows the iterate-until-there-are-no-more-changes idiom.  But checking for dead blocks after every merge is very wasteful and separating the loops makes sense.

It might be worth extending the optimisation further (hard to tell without experimentation) to not iterate at all for dead blocks, but to instead do a single pass to find all the blocks with no predecessors and remove them all at once afterwards.  Should be O(n).

A method could be added to `Function` to gather the predecessor counts for every block found within, returned in a map:
```
fn get_block_predecessor_counts(&self, context: &Context) -> HashMap<Block, usize> { ... }
```
Right now it iterates for each block and asks for its predecessor count which in turn iterates for each block searching for branches to that block.  O(n^2).

vaivaswatha:
> Yep, good catch. I kinda copy pasted that loop from other passes, as it follows the iterate-until-there-are-no-more-changes idiom. But checking for dead blocks after every merge is very wasteful and separating the loops makes sense.
> 
> It might be worth extending the optimisation further (hard to tell without experimentation) to not iterate at all for dead blocks, but to instead do a single pass to find all the blocks with no predecessors and remove them all at once afterwards. Should be O(n).
> 
> A method could be added to `Function` to gather the predecessor counts for every block found within, returned in a map:
> 
> ```
> fn get_block_predecessor_counts(&self, context: &Context) -> HashMap<Block, usize> { ... }
> ```
> 
> Right now it iterates for each block and asks for its predecessor count which in turn iterates for each block searching for branches to that block. O(n^2).

Yep, i realised the predecessor inefficiency and am already on a fix for it. Hope it'll help.

otrho:
I remember LLVM having a `has_zero_predecessors()` call which would be more efficient than `count_predecessors() == 0` since it could abort the search early, as soon as a predecessor is found.

But again, my experience tells me that the best way to optimise code is to profile, as often you think of things like this, spend an hour or two implementing it and then find a 0.1% improvement.

PR #2549, opened 2022-08-16
Introduce configuration-time constants in `Forc.toml`

### LINKED ISSUE #1498: Add support for configuration-time constants
We currently have `const` values which must be hardcoded.
For values which can't be hardcoded, or are subject to change, we have `storage`.

Configuration-time constants (in Solidity these are `immutable` variables) give us a third option which fits nicely in between the existing 2.

These are variables which are like `const` values in that they can't be changed once set, but their value is set at configuration time. In Solidity this is done via the constructor.
We don't have a concept of `constructor` in Sway, so this might be done in an `init` function (just a normal abi function which must be made to be callable only once) or perhaps via some other mechanism.

sezna:
This PR adds the ability to specify configuration-time constants. [Example](https://github.com/FuelLabs/sway/blob/7c7b085ec41ed7f3f83267a9e66d9b10d4fe316a/test/src/e2e_vm_tests/test_programs/should_pass/language/config_time_constants/Forc.toml#L9). These are constants that are known and put into the `Forc.toml` manifest, and then available to the Sway program as constants. This is accomplished by injecting them as `ConstantDeclaration`s when generating the initial namespace.

I had to add a little mini-compilation pass for each constant. While this may appear messy, I directly reused our normal compilation pipeline. What this means is that anything that is constant-evaluatable (valid inside of a `const` expression in Sway) is valid as a config-time constant. As we continue to expand constant evaluation, config-time constants get those benefits for free.

While the span data isn't _perfect_ (always line 0), we do get rendered errors for config constants: 
image

Warnings are omitted for now. 

closes #1498


canndrew:
What's the reason for adding this to Forc.toml rather than in the code (eg. in a `constants.sw` file imported into the root module).

I've wanted something similar to this for rust, as an alternative to rust's (kinda broken) features system. But the point of it would be that you could specify the constant values when building the project, eg. on the command line, or in a dependent package's manifest. Are the values here meant to only be default values? Or are they intended to be overrideable?

otrho:
> What's the reason for adding this to Forc.toml rather than in the code (eg. in a `constants.sw` file imported into the root module).

I think it's probably easier to inject constants automatically into a TOML file with a build system or other mechanic than into Rust.  It might be better in a different `constants.toml` file though..?

sezna:
We discussed verbally adding config-time constants as a flag to `forc`. I looked into this and it isn't trivial as clap doesn't currently support providing multiple values for non-flag arguments. I've decided to punt that work and created #2575 for further bikeshedding. I think this is sufficient as a standalone PR, with the understanding that we may introduce more ways to add these constants in the future.

mohammadfawaz (approved):
:shipit:

PR #2560, opened 2022-08-17
Speed improvements and bug fixes for `simplify-cfg`.

vaivaswatha:
#### Algorithmic changes
- New worklist based algorithm for `remove_dead_blocks`
- Cached predecessor counts for `merge_blocks`

#### Bugfix:
- `BlockContent::successors()` was incorrect. It was returning successors for all blocks in the function.
- Allow empty PHIs in `to_block` when merging blocks.

#### Performance
With this change, we now can compile the `stdlib/vec` test in about 35s (on my i7-7th-gen/32GB RAM laptop).
It was [previously taking](https://github.com/FuelLabs/sway/pull/2548#issue-1340432847) about 25m.

#### Testing
 The `stdlib/vec` test's bytecode with this change is identical to the one without this change. The existing regression tests for `simplify-cfg` pass.

I've also enabled `simplify-cfg` optimization pass by default now.

### DIFF
--- a/sway-ir/src/optimize/simplify_cfg.rs
+++ b/sway-ir/src/optimize/simplify_cfg.rs
@@ -13,17 +13,33 @@ use crate::{
     value::ValueDatum,
 };
 
+use std::collections::HashMap;
+
+type PredCounts = HashMap<Block, usize>;
+
+fn count_predecessors(context: &Context, function: &Function) -> PredCounts {
+    let mut pred_counts = PredCounts::new();
+
+    for block in function.block_iter(context) {
+        for succ in block.successors(context) {
+            match pred_counts.get_mut(&succ) {
+                Some(count) => *count += 1,
+                None => {
+                    pred_counts.insert(succ, 1);
+                }
+            }
+        }
+    }
+    pred_counts
+}
+
 pub fn simplify_cfg(context: &mut Context, function: &Function) -> Result<bool, IrError> {
     let mut modified = false;
+    modified |= remove_dead_blocks(context, function)?;
+
+    let mut pred_counts = count_predecessors(context, function);
     loop {
-        if remove_dead_blocks(context, function)? {
-            modified = true;
-            continue;
-        }
-        break;
-    }
-    loop {
-        if merge_blocks(context, function)? {
+        if merge_blocks(context, &mut pred_counts, function)? {
             modified = true;
             continue;
         }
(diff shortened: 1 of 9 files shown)


sway-ir/src/optimize/simplify_cfg.rs
```diff
@@ -13,17 +13,33 @@ use crate::{
...
-    let mut modified = false;
-    loop {
-        if remove_dead_blocks(context, function)? {
-            modified = true;
-            continue;
+use std::collections::HashMap;
+
+type PredCounts = HashMap<Block, usize>;
+
+fn count_predecessors(context: &Context, function: &Function) -> PredCounts {
```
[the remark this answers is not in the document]
otrho: We might want to adjust `Block::num_predecessors()` too, if this algorithms is better.

sway-ir/src/optimize/simplify_cfg.rs
```diff
@@ -33,43 +49,62 @@ pub fn simplify_cfg(context: &mut Context, function: &Function) -> Result<bool,
...
-
-fn merge_blocks(context: &mut Context, function: &Function) -> Result<bool, IrError> {
-    // Get the block a block branches to iff its terminator is `br`.
-    let block_terms_with_br = |from_block: Block| -> Option<(Block, Block)> {
-        from_block.get_terminator(context).and_then(|term| {
-            if let Instruction::Branch(to_block) = term {
-                Some((from_block, *to_block))
-            } else {
-                None
+    let mut worklist = std::collections::VecDeque::<Block>::new();
```
otrho, line 52: I'm curious -- it might be slightly more efficient to just use a `Vec` here and `pop_back()`.  The order of visiting the blocks doesn't matter and while `VecDeque` is efficient it will need to do more allocations than `Vec` if you're going to use it as a queue.

otrho: I was going to suggest you recurse instead of using a work list at all, but then the CFG for a large function could be quite deep so you'd need to guarantee tail calls to be sure not to blow up the stack.

sway-ir/src/optimize/simplify_cfg.rs
```diff
@@ -134,14 +165,13 @@ fn merge_blocks(context: &mut Context, function: &Function) -> Result<bool, IrEr
 
         // Remove `to_block`.
         function.remove_block(context, &to_block)?;
+        // We probably don't need to update pred_counts, but anyway ...
+        pred_counts.remove(&to_block);
```
otrho, line 169: No, I wouldn't, and then `pred_counts` doesn't have to be `mut`.

sway-ir/src/optimize/simplify_cfg.rs
```diff
@@ -80,25 +115,20 @@ fn merge_blocks(context: &mut Context, function: &Function) -> Result<bool, IrEr
...
-                    break;
-                }
+                block_chain.push(next_pair.1);
             }
         }
     }
 
     // Keep a copy of the final block in the chain so we can adjust the successors below.
     let final_to_block = block_chain.last().copied().unwrap();
+    let final_to_block_succs = final_to_block.successors(context);
```
otrho, line 131: Does this need to be up here, or can we declare it down below where it's used?

otrho (commented):
Nice speed ups!  Some preliminary comments below.

sway-ir/src/optimize/simplify_cfg.rs
vaivaswatha: > Does this need to be up here, or can we declare it down below where it's used?

This needs to be here because, after merging, the terminator is moved to a different block.

sway-ir/src/optimize/simplify_cfg.rs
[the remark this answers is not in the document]
vaivaswatha: > I have a feeling this would be better as a method on `Function` as it will probably be useful in other passes. 

Okay, I'll move it there.

> And there's probably no need to iterate through `successors()` to just count them, you could use `len()`.

I'm not counting the `successors()` here but for each of them, updating their predecessor count. So I'll need to iterate through them right? Or I misunderstood your comment.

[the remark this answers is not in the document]
vaivaswatha: > We might want to adjust `Block::num_predecessors()` too, if this algorithms is better.

I've address all review comments except this one. I think we should do it separately if needed. At this point I don't know what is better for here.

sway-ir/src/optimize/simplify_cfg.rs
```diff
@@ -33,43 +31,62 @@ pub fn simplify_cfg(context: &mut Context, function: &Function) -> Result<bool,
...
-
-fn merge_blocks(context: &mut Context, function: &Function) -> Result<bool, IrError> {
-    // Get the block a block branches to iff its terminator is `br`.
-    let block_terms_with_br = |from_block: Block| -> Option<(Block, Block)> {
-        from_block.get_terminator(context).and_then(|term| {
-            if let Instruction::Branch(to_block) = term {
-                Some((from_block, *to_block))
-            } else {
-                None
+    let mut worklist = std::vec::Vec::<Block>::new();
```
otrho, line 34: `Vec` is in the prelude, you shouldn't need to fully qualify it here.  I'm surprised `clippy` doesn't mention it.

sway-ir/src/optimize/simplify_cfg.rs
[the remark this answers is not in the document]
otrho: Yep, and yeah, I misread the 'counting successors' bit, sorry.  Probably shouldn't be reviewing PRs at night. :)  But here I am anyway...

otrho (approved):
Other than the `Vec` thing I think it's all OK now.  Thanks.

sway-ir/src/optimize/simplify_cfg.rs
vaivaswatha: > `Vec` is in the prelude, you shouldn't need to fully qualify it here. I'm surprised `clippy` doesn't mention it.

I just fixed this. Please re-approve.

vaivaswatha:
Thanks for the quick review @otrho.

mohammadfawaz (approved):
Fantastic, thanks!

I'm not surprised the bytecode is more or less the same for most tests. As Toby mentioned before on Slack, codegen already tries to optimize the bytecode by:
- Removing unnecessary jumps
- Coalescing registers with redundant moves between them

This pass's main benefit is reducing the IR size and make the IR more canonical so that other future passes can do better. I also imagine we'll have to run this pass often in the future as we start introducing more passes that alter the CFG (such as loop optimization passes and such, if we feel like this is necessary).

PR #2561, opened 2022-08-17
Make return into an expression

canndrew:
This PR makes `return` into an expression, rather than just a kind of `AstNode`. This change was made because:
 (a) It enables using `return` is positions other than where statements are expected.
 (b) It *is* an expression (to the extent that we're following rust and not earlier less-sane languages), and sorting out this stuff,  eg. organising the compiler's internal code representations and putting things into the right types, will help with later refactorings such as the declaration engine.

Note that this PR shouldn't change any existing behaviour, but I've added a couple of TODOs to the code where the existing behaviour seems questionable. So having clarification on those would be useful.


sezna:
Looks like this one still needs a `fmt` check and some investigation into why `asm_expr_basic` in the test suite is failing.

mohammadfawaz:
Changing to draft until this is ready.

canndrew:
The problem is that, on master, the way we handle returns is broken. For instance this won't compile on master:

```
fn foo() -> u32 {
    {
        return 123;
    }
}
```

This PR is exposing the bug since it treats a return expression the same way that a code block with a single return ast node would get treated on master.

canndrew:
This is now rebased off of the [diverging expressions PR](https://github.com/FuelLabs/sway/pull/2627), so should be good once that gets merged.

mohammadfawaz:
This should be ready now. I've resolved the conflicts and fixed on small issue with it in the const evaluator.

mohammadfawaz (approved):

Issue #2465, opened 2022-08-04
Support `StorageMap` and `StorageVec` in complex *storage* types


mohammadfawaz:
For example, it would be nice to be able to write:
```rust
struct Wrapper {
    map1: StorageMap<u64, u64> = StorageMap {},
    map2: StorageMap<u64, u64> = StorageMap {},
}

storage {
    wrapper: Wrapper = Wrapper {
        map1: StorageMap {},
        map2: StorageMap {},        
    },
    tuple_wrapper: (StorageVec<u64>, StorageVec<bool>) = (StorageVec {}, StorageVec {}),
    // ... enum example
    // ... array example (once arrays are supported in storage)
}
```
This requires the `__get_storage_key()` intrinsic to pass in more information such as the field index in a struct and so on.

This is a follow up on https://github.com/FuelLabs/sway/issues/1796 which is being closed with https://github.com/FuelLabs/sway/pull/2453. https://github.com/FuelLabs/sway/pull/2453 disallow the above until we implement it.

PR #2562, opened 2022-08-17
Introduce the Sway reference, a more technical Sway source of documentation.

### LINKED ISSUE #2443: Begin work on a brand new `Sway Book` to replace the current version
## Summary

Note: The original issue has been hijacked and completely reworked to contain significantly more work than the original issue.

This issue is now the first issue tracking a brand new sway book which will eventually take place of the current book. It's purpose is to make the book intuitive, organized and concise.

The book is meant to be self contained and thus the source code and documentation is in a brand new directory `sway-book`. Once the book is completed the current `./docs` and `./examples` directories are planned to be deleted since the new book will contain all of the information and much more.

Another future enhancement is to add further anchor tags into the standard library so that any section that is re-written in the source code of the new book can reference the literal standard library without having to re-write that component in its own `src` directory.

The structure is split into two sections.

- `src`
- `documentation`

Those two structures follow each other closely so that it's easy to navigate from once section to the other while keeping the separation of source code and documentation.

Braqzen:
## Summary

More information can be found in #2443 .

Notes for PR:

- The review is meant to take place using `mdbook` so check out the main `README.md` to see how to run it in the browser
- This book is incomplete however there are a lot of pages so the idea is to PR this into the repo soon so that subsequent reviews are smaller
- No CI added until the book is ready to replace the current book
- Ideally reviews are split between multiple people because there is a lot of content

Closes #2443 

## Pages Reviewed
@matt-user : Program Types -> Functions, Misc


sway-book/README.md
```diff
@@ -0,0 +1,28 @@
...
+2. Install [mdbook](https://rust-lang.github.io/mdBook/).
+
+   ```bash
+   cargo install mdbook
+   ```
+
+3. To [build](https://rust-lang.github.io/mdBook/cli/build.html) the book make sure that you are in `/sway/sway-book` and run
+
+   ```bash
+   mdbook build
```
anton-trunov, line 17: I have Rust and `mdbook` installed but I get this error message at this step:
```
$ mdbook build
2022-11-01 14:33:09 [INFO] (mdbook::book): Book building has started
2022-11-01 14:33:09 [WARN] (mdbook::preprocess::cmd): The command wasn't found, is the "forc-documenter" preprocessor installed?
2022-11-01 14:33:09 [WARN] (mdbook::preprocess::cmd): 	Command: mdbook-forc-documenter
2022-11-01 14:33:09 [ERROR] (mdbook::preprocess::links): Error updating "{{#include ../../../../examples/msg_sender/src/main.sw}}", Could not read file for link {{#include ../../../../examples/msg_sender/src/main.sw}} (/Users/anton/fuel.network/sway/sway-book/src/documentation/operations/call-data/../../../../examples/msg_sender/src/main.sw)
2022-11-01 14:33:09 [WARN] (mdbook::preprocess::links): Caused By: No such file or directory (os error 2)
2022-11-01 14:33:09 [ERROR] (mdbook::preprocess::links): Error updating "{{#include ../../../examples/ownership/src/main.sw:set_owner_example}}", Could not read file for link {{#include ../../../examples/ownership/src/main.sw:set_owner_example}} (/Users/anton/fuel.network/sway/sway-book/src/documentation/examples/../../../examples/ownership/src/main.sw)
2022-11-01 14:33:09 [WARN] (mdbook::preprocess::links): Caused By: No such file or directory (os error 2)
2022-11-01 14:33:09 [ERROR] (mdbook::preprocess::links): Error updating "{{#include ../../../examples/ownership/src/main.sw:revoke_owner_example}}", Could not read file for link {{#include ../../../examples/ownership/src/main.sw:revoke_owner_example}} (/Users/anton/fuel.network/sway/sway-book/src/documentation/examples/../../../examples/ownership/src/main.sw)
2022-11-01 14:33:09 [WARN] (mdbook::preprocess::links): Caused By: No such file or directory (os error 2)
2022-11-01 14:33:09 [INFO] (mdbook::book): Running the html backend
```

Braqzen: Does it prevent the book from building? I think this should allow you to open the book in the browser.

Those links are pointing at the current examples in the live sway book. I have been moving things around and commenting things out so I have not updated all of the links / re-created new examples so if there is a link that does not point at a file then it will show a warning, or an error. I usually ignore them because I can use the book and over time get to the places where I need to update the sections.

If this prevents the book from building then let me know otherwise this can be ignored.

sway-book/src/code/operations/storage/storage_init/src/main.sw
```diff
@@ -0,0 +1,33 @@
...
+}
+// ANCHOR_END: data_structures
+// ANCHOR: initialization
+storage {
+    current_owners: u64 = 0,
+    explicit_declaration: Owner = Owner {
+        maximum_owners: 10,
+        role: Role::FullAccess,
+    },
+    encapsulated_declaration: Owner = ~Owner::default(),
```
supiket, line 31: Is this `~` an exception to "no more `~`"?

supiket (commented):
This book structure is so refreshing. I especially like that the examples are left towards the end.
1. Will testing have its own section?
2. I know from experience that `StorageMap`s cannot be returned from contract functions. Does it make sense to add a general explanation around returning collections that can grow / shrink in size from functions? 
3. There is no reference to `lifetime`s in the previous book or the Rust SDK book. Is it because Sway lifetimes are exactly the same as Rust lifetimes? Do you think it is worth mentioning anywhere in this book, perhaps in the "Language Comparison - Rust" section?
4. For people who are previously unfamiliar with Rust, the concept of "borrowing" can be unfamiliar and is most likely going to cause multiple failing `forc build`s for first-timers. Does it make sense to add a pointer to an explanation from The Rust Language Book, perhaps?
5. I would consider adding some code chunks (if that does not work, visuals) of the Sway compiler warnings / errors in action to show how powerful it is.

sway-book/src/code/operations/storage/storage_init/src/main.sw
Braqzen: Not an exception, an oversight on my part. That should be removed.

sway-book/src/documentation/language/variables/const.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Declaration
+
+To define a constant the `const` keyword is used followed by a name and an assignment of a value.
+
+```sway
+{{#include ../../../code/language/variables/src/lib.sw:constants}}
+```
+
+The example above hardcodes the value of `5` however function calls may also be used alongside [built-in types](../built-ins/index.md).
```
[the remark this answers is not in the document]
Braqzen: That is correct. You may use a function with a return value to assign that returned value to the constant. I am under the impression that this will only work if the value can be figured out at compile time i.e. static dispatch.

Braqzen:
> This book structure is so refreshing. I especially like that the examples are left towards the end.
> 
> 1. Will testing have its own section?
> 2. I know from experience that `StorageMap`s cannot be returned from contract functions. Does it make sense to add a general explanation around returning collections that can grow / shrink in size from functions?
> 3. There is no reference to `lifetime`s in the previous book or the Rust SDK book. Is it because Sway lifetimes are exactly the same as Rust lifetimes? Do you think it is worth mentioning anywhere in this book, perhaps in the "Language Comparison - Rust" section?
> 4. For people who are previously unfamiliar with Rust, the concept of "borrowing" can be unfamiliar and is most likely going to cause multiple failing `forc build`s for first-timers. Does it make sense to add a pointer to an explanation from The Rust Language Book, perhaps?
> 5. I would consider adding some code chunks (if that does not work, visuals) of the Sway compiler warnings / errors in action to show how powerful it is.

1. My intention is to not have any testing in this book except when in-language testing is available i.e. you can test sway with sway rather than Rust or TS. The SDKs will be referenced and any testing pertinent to those repositories should be contained and maintained in those repositories. Having that info be duplicated in this book and maintained as well doubles the work.
2. That is a user error i.e. the user does not know what type they are dealing with and how it works. I don't believe that it should be mentioned even more so that it's an implementation in the standard library. Once the documentation tool is out perhaps as part of its documentation it may mention it there. I'm not entirely against mentioning it somewhere in the book, e.g. known issues, however this isn't an issue with the code and instead with the understanding of the user. You may point out that the rest of the book explains content, specifically the built-in types, so why would this particular type be excluded. The reason is that 1) standard library docs and 2) iirc nothing explains the implementation and instead presents the usage and some misconceptions / possibly actual "issues" in the "known issues" section. This means that now the book has to start teaching about the implementation and fundamental concepts of data structures rather than simple usage. The compiler should also produce an appropriate warning to let them know that they cannot return ... the entire memory of a contract.
3. Sway aims to make lifetimes disappear so that it's simpler for the user. I'm not opposed to adding such a comment however the implication is that since it's not mentioned anywhere and you can use variables like in any other language that does not force the handling of some form of a lifetime then it's like you've never been told about the concept and should not worry about it. I will keep it in mind as this is a point I will likely add to the Rust language section.
4. Ref point 3, there is no borrowing in sway afaik.
5. The compiler is being updated all the time and there is no easy way to maintain this. In the future we'd like to allow a user to run a sway code block directly in the book and have it be annotated as to whether it compiles or errors. This way the user can directly run and edit the snippets plus see the outcome of the compiler. I am unsure of what type of work is needed for this however this works by default for Rust. We need to implement some preprocessor (which we already have for some things that are unknown to me) which would accomplish this. This will likely be done by the tooling team and I'm looking forward to it.

sway-book/src/documentation/language/program-types/script.md
```diff
@@ -0,0 +1,19 @@
...
+
+Some properties of a script include:
+
+- It cannot be called by a contract
+- It is stateless but can interact with storage through a contract
+- Can call multiple contracts
+
+## Example
+
+The following example demonstrates a sciprt which takes one argument and returns the boolean value of true.
```
nfurfaro, line 15: typo: script

nfurfaro (commented):
Just a couple nits to start with.

sway-book/src/documentation/language/style-guide/index.md
```diff
@@ -0,0 +1,7 @@
+# Style Guide
+
+Programming languages have different ways of styling code i.e. how variables, functions, structures etc. are written.
+
+The following snippets present the style for writing `Sway`.
```
LuizAsFight, line 5:
```suggestion
The following snippets present the style guide for writing `Sway`.
```

sway-book/src/documentation/language/style-guide/annotations.md
```diff
@@ -0,0 +1,9 @@
+# Type Annotations
+
+When declaring a variable it is possible to annotate it with a type however the compiler can usually infer that information.
+
+The general approach is to omit a type if the compiler does not throw an error however if it is deemed clearer by the developer to indicate the type then that is also encouraged.
```
LuizAsFight, line 5:
```suggestion
The general approach is to omit a type if the compiler does not throw an error.
However if it is deemed clearer by the developer to indicate the type, then that is also encouraged.
```

sway-book/src/documentation/language/style-guide/comments.md
```diff
@@ -0,0 +1,10 @@
+# Comments
+
+In [regular comments](../comments/index.md#regular-comments) we outline two forms:
+
+- `// comment`
+- `/* comment */`
+
+The first form is generally encouraged however there may be instances where a comment needs to be placed in the middle of some code in which case the second form is encouraged.
+
+For example, in [function decleration](../functions/index.md#function-decleration) the second form is used to indicate additional parameters.
```
LuizAsFight, line 10:
```suggestion
For example, in [function declaration](../functions/index.md#function-decleration) the second form is used to indicate additional parameters.
```

LuizAsFight (commented):
Just reviewed the `Style Guide` section.
Looks really good, concise and coherent. Just few minor comments

eureka-cpu:
This may help with organizing and maintaining links for the future should they break: Add link references at the bottom of markdown pages instead of inline with the thing they are referencing.

sway-book/src/documentation/language/variables/index.md
```diff
@@ -0,0 +1,14 @@
+# Variables
+
+A variable is a way to reference some information by a specific name and it can take the form of a variety of [data structures](../built-ins/index.md).
+
+In Sway there are two states that a variable can take:
+
+- immutable
+  - Can be read but cannot be changed after it has been declared
+- mutable
+  - Can be read and have its value changed but only if the type is the same
```
bitzoic, line 10:
```suggestion
  - Can be read and can have its value changed if and only if the new value is of the same type
```
?

sway-book/src/documentation/language/program-types/index.md
```diff
@@ -0,0 +1,41 @@
+# Sway Program Types
+
+A Sway program is a file ending with the extension `.sw`, e.g. `main.sw`, and the first line of the file is a decleration of the _type_ of program that it is.
```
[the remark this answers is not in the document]
matt-user: Also "that it is." seems a bit redundant

sway-book/src/documentation/language/variables/const.md
```diff
@@ -0,0 +1,17 @@
...
+
+- The constant is evaluated at compile-time
+- A constant can be declared in any scope i.e. inside of a [function](../functions/index.md) and outside
+- The `mut` keyword cannot be used
+
+## Declaration
+
+To define a constant the `const` keyword is used followed by a name and an assignment of a value.
+
+```sway
```
bitzoic, line 13: This example should probably use `FOO` and state that constants are usually defined in all upper case.

sway-book/src/documentation/language/functions/function.md
```diff
@@ -0,0 +1,19 @@
+# Functions
+
+In this section we will define a function that takes two numerical inputs and returns a [boolean](../built-ins/boolean.md) value indicating whether they are equal. We will also take a look at how to use the function.
+
+## Declaration
+
+The following function is called `equals` and it takes two parameters of type `u64` (64-bit unsigned integers). It performs a comparison and [implicitly](./return.md) returns the result of that comparison.
```
bitzoic, line 7:
```suggestion
The following function is named `equals` and it takes two parameters of type `u64` (64-bit unsigned integers). It performs a comparison and [implicitly](./return.md) returns the result of that comparison.
```

sway-book/src/documentation/language/functions/method.md
```diff
@@ -0,0 +1,27 @@
+# Methods
+
+Methods are defined within the context of a [struct](../built-ins/structs.md) (or [enum](../built-ins/enums.md)) and either refer to the type or mutate it.
+
+The first parameter of a method is always `self`, which represents the instance of the type the method is being called on.
+
+## Decleration
```
bitzoic, line 7:
```suggestion
## Declaration
```

sway-book/src/documentation/language/functions/associated-function.md
```diff
@@ -0,0 +1,27 @@
+# Associated Functions
+
+Associated functions are similar to methods in that they are also defined in the context of a [struct](../built-ins/structs.md) or [enum](../built-ins/enums.md), but they do not use any of the data in the struct and as a result do not take `self` as a parameter.
+
+Associated functions could be standalone functions, but they are included in a specific type for organizational or semantic reasons.
+
+## Decleration
```
bitzoic, line 7:
```suggestion
## Declaration
```

sway-book/src/documentation/language/program-types/libraries/external.md
```diff
@@ -0,0 +1,52 @@
...
+
+#### Import
+
+Use the `use` keyword to selectively import our code from `my_other_library`
+
+```sway
+{{#include ../../../../code/language/program-types/libraries/external/my_library/src/lib.sw}}
+```
+
+#### Add to Dependencies
```
matt-user, line 46: I would move "Add to Dependencies" above "Import" because a dev can't really import until they add it to their deps.  Also the comment in the code example says "// `quix` from `my_other_library` is now available throughout the file", but that's only correct if you add the dep to the toml first.

eureka-cpu:
I think the introduction could use a section in regards to the target audience how Rust does [here](https://doc.rust-lang.org/book/ch00-00-introduction.html#who-this-book-is-for).

sway-book/src/documentation/language/control-flow/index.md
```diff
@@ -0,0 +1,7 @@
+# Control Flow
+
+A control flow in a program is the order in which instructions are executed.
+
+For example, a function may take an input `u64` and if the value is greater than `5` then it calls one function otherwise it calls a different function.
+
+Controlling the order of instructions can be done through the use of conditional expressions such as [if](if-expressions.md) and [match](match/complex/index.md) and through [looping](looping/index.md).
```
bitzoic, line 7:
```suggestion
Controlling the order of instructions can be done through the use of conditional expressions such as [if](if-expressions.md) and [match](match/index.md) and through [looping](looping/index.md).
```

Wrong link

eureka-cpu:
There are some sections toward the end of the book that assume certain things like `catch_all` which don't have links to other sections of the book. I'll try to point to as many as I can find, but keep in mind that the book may also be just a reference for some and those links will be helpful!

sway-book/src/documentation/language/control-flow/if-expressions.md
```diff
@@ -0,0 +1,49 @@
...
+
+In the following example we have a hardcoded variable `number` set to the value of `5` which is put through some conditional checks.
+
+```sway
+{{#include ../../../code/language/control_flow/src/lib.sw:conditional}}
+```
+
+The conditional checks are performed in the order that they are defined therefore the first check is to see if the `number` is divisible by `3`.
+
+If the condition evaluates to the [boolean](../built-ins/boolean.md) value of `true` then we call `function 1` and we move on to the end where the comment `more code here` is written. We do not evaluate the remaining conditions.
```
bitzoic, line 15: I would bullet point these cases rather than a wall of text

sway-book/src/documentation/language/control-flow/if-expressions.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Using if & let together
+
+In [Conditional Branching](#conditional-branching) we have opted to call some functions depending on which condition is met however that is not the only thing that we can do. Since `if`'s are expressions in Sway we can use them to match on a pattern.
+
+### if let
+
+In the following example we combine `if` and `let` into `if let` followed by some comparison which must evaluate to a [boolean](../built-ins/boolean.md).
+
+```sway
```
bitzoic, line 29: This example does not use `if let`

bitzoic: Looks like this is referring to the examples below. Maybe change "example" to "examples" and move the enum declaration into the next block of code rather than it's own

sway-book/src/documentation/language/control-flow/match/index.md
```diff
@@ -0,0 +1,13 @@
+# match
+
+[If expressions](if-expressions.md) can be used to check a large number of conditions however there is an alternative syntax which allows us to perform advanced pattern matching.
```
bitzoic, line 3: Link does not work

sway-book/src/documentation/examples/contract-ownership.md
```diff
@@ -0,0 +1,34 @@
+# Contract Ownership
+
+THe following example implements access control to restrict functionality to a priviledged user.
```
eureka-cpu, line 3: *The

sway-book/src/documentation/language/control-flow/match/single-line.md
```diff
@@ -0,0 +1,13 @@
+# Single Line Arm
+
+The following example demonstrates how a type can be matched on and its output is assigned to a variable. The assignment to a variable is optional.
+
+```sway
```
bitzoic, line 5: The following example uses "catch_all" over "_". I believe the convention is to use the underscore and if you'd like a comment to mark that it is a catch all that would be better.

sway-book/src/documentation/language/control-flow/match/single-line.md
```diff
@@ -0,0 +1,13 @@
...
+{{#include ../../../../code/language/control_flow/src/lib.sw:simple_match}}
+```
+
+The left side of the arrow `=>` is the pattern that we are matching on and the right side of the arrow `=>` is the logic that we want to perform, in this case we are returning a different multiple of `10` depending on which arm is matched.
+
+We check each arm starting from `0` and make our way down until we either find a match on our pattern or we reach the `catch_all` case.
+
+The `catch_all` case is equivalent to an `else` in [if expressions](if-expressions.md) and it does not have to be called `catch_all`. Any pattern declared after a `catch_all` case will not be matched because once the compiler sees the first `catch_all` it stop performing further checks.
```
bitzoic, line 13: Link does not work

sway-book/src/documentation/language/program-types/libraries/index.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Accessibility
+
+Code defined inside a library, but more generally anywhere inside a Sway project, is considered to be `private` which means that it is inaccessible to other files unless explicitly exposed.
+
+Code can be exposed through a two step process:
+
+- Add a `pub` keyword at the start of some code
+- Specify the [library](external.md) in the `Forc.toml` file as a dependency and then import the `pub` declaration
+
```
ControlCplusControlV, line 21: Should note that functions within an `impl` block are implicitly public

sway-book/src/documentation/language/program-types/predicate.md
```diff
@@ -0,0 +1,17 @@
+# Predicates
+
+A predicate is an executable that represents a UTXO spending condition, such as a multisig predicate, which has restrictions on the VM instructions that can be used (e.g. no jumps).
```
ControlCplusControlV, line 3: It would be accurate to say that the restrictions are that a predicate cannot be stateful, rather than using jumps as an example

sway-book/src/documentation/language/program-types/script.md
```diff
@@ -0,0 +1,19 @@
+# Scripts
+
+A script is an executable that does not need to be deployed because it only exists during a transaction.
+
+It can be used to replicate the functionality of contracts, such as routers, without the cost of deployment or increase of the blockchain size.
+
+Some properties of a script include:
+
+- It cannot be called by a contract
+- It is stateless but can interact with storage through a contract
```
ControlCplusControlV, line 10: Scripts can impact the state but the scripts themselves aren't stored on the chain, perhaps a better way to say this than `stateless`?

sway-book/src/documentation/language/control-flow/looping/index.md
```diff
@@ -0,0 +1,11 @@
+# looping
```
bitzoic, line 1:
```suggestion
# Looping
```

sway-book/src/documentation/language/control-flow/looping/index.md
```diff
@@ -0,0 +1,11 @@
...
+
+A loop is a type of operation which allows us to perform computation a certain number of times. For example, given a collection of items we could call a method on the first item and iterate until the method has been called on each item.
+
+Usually, a loop has a condition which prevents it from continuing indefinitely however it is possible to create a loop that never stops i.e. an infinite loop.
+
+Programming languages have various forms of syntax for declaring a loop which may slightly alter how the iteration takes place.
+
+Sway has the following loops:
```
bitzoic, line 9: There should be a **Note** that Sway does not support for loops

sway-book/src/documentation/misc/workarounds/pattern-matching.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Nested Match Expressions
+
+In [nested match expressions](../../language/control-flow/match/complex/nested-expression.md) we nest a `match` expression by embedding it inside the `{}` brackets on the right side of the arrow `=>`.
+
+Match expressions cannot be used as a pattern, the left side of the arrow `=>`.
+
+## Constants
+
+When matching on [constants](../../language/control-flow/match/complex/constant.md) we specify that a constant must be used in order to match on a variable. Dynamic values, such as an argument to a function, cannot be matched upon because it will be treated as a `catch_all` case and thus any subsequent patterns will not be checked.
```
eureka-cpu, line 11: Here's the `catch_all` I referenced before, I think things like this where an example isn't present could have a link to somewhere else in the book with more context

sway-book/src/documentation/language/built-ins/index.md
```diff
@@ -0,0 +1,38 @@
...
+   5. `hexadecimal`, `binary` & `base-10` syntax
+2. [Boolean](boolean.md)
+   1. `bool` (true or false)
+3. [Strings](string.md)
+   1. `str[n]` (fixed-length string of size n)
+4. [Bytes](b256.md)
+   1. `b256` (256 bits / 32 bytes, i.e. a hash)
+
+<!-- TODO: The following sentence does not belong here. We need to convey the default size, including word size, somewhere however not on this page -->
+The default numeric type is `u64`. The FuelVM's word size is 64 bits, and the cases where using a smaller numeric type saves space are minimal.
```
matt-user, line 25:
```suggestion
The default numeric type is `u64`. The FuelVM's word size is 64 bits, and the cases where using a smaller numeric type to save space are minimal.
```

sway-book/src/documentation/misc/workarounds/general.md
```diff
@@ -0,0 +1,7 @@
+# General
+
+> TODO: need help filling this in, might remove this page and move content into individual sections
+
+- [Issue: #870](https://github.com/FuelLabs/sway/issues/870)
+  - All impl blocks need to be defined before any of the functions they define can be called. This includes sibling functions in the same impl declaration, i.e., functions in an impl can't call each other yet.
```
eureka-cpu, line 6: Here's another example where maybe `impl` could be wrapped and have an internal link to where there's more context on them

sway-book/src/documentation/operations/storage/libraries/store-get.md
```diff
@@ -0,0 +1,31 @@
...
+
+To use `store()` & `get()` we must import them however we are not required to declare a `storage` block.
+
+```sway
+{{#include ../../../../code/operations/storage/store_get/src/main.sw:import}}
+```
+
+## Store
+
+To store a generic value `T` we must provide a key of type `b256`.
```
eureka-cpu, line 15: Another link could be given to `generic` here to reference generics in Sway

sway-book/src/documentation/language/built-ins/numeric.md
```diff
@@ -0,0 +1,42 @@
...
+
+In order to achieve this one _bit_ must be kept for tracking the sign (+ or -) of the value and thus the range of available values is smaller than an unsigned integer.
+
+For those inclined, the range for an n-bit signed integer is -2<sup>n-1</sup> to 2<sup>n-1</sup>-1.
+
+Sway does not natively support signed integers however there is nothing stopping a library from using primitives to create types that act like signed types.
+
+## Unsigned Integers
+
+An unsigned integer is a whole number which can take the value of zero and any positive number, but cannot be negative. This allows for one more _bit_ of values to be used for the positive numbers and thus the positive range is significantly larger than for signed integers.
```
matt-user, line 25:
```suggestion
An unsigned integer is a whole number which can take the value of zero or any positive number, but cannot be negative. This allows for one more _bit_ of values to be used for the positive numbers and thus the positive range is significantly larger than for signed integers.
```

sway-book/src/documentation/language/annotations/attributes/index.md
```diff
@@ -0,0 +1,3 @@
+# Attributes
+
+An attribute is a metadatum which provides some additional functionality.
```
bitzoic, line 3: There should be another attribute for the test attribute recently added.

sway-book/src/documentation/operations/storage/libraries/store-get.md
eureka-cpu: Scratch that, I see there's a link above this 😅

eureka-cpu:
In the language comparison sections which is empty, it may be good to show explicit examples of the differences between how Rust and Sway handle references e.g. `&` and `ref` in Rust vs. just `ref` in Sway

bitzoic (commented):
First iteration, laptop is dying

sway-book/src/code/language/built-ins/structs/src/lib.sw
```diff
@@ -0,0 +1,53 @@
...
+// ANCHOR_END: instantiation
+// ANCHOR: destructuring
+fn destructuring() {
+    let foo = Foo {
+        bar: 42,
+        baz: false,
+    };
+
+    // bar and baz are now accessible as variables
+    let Foo{bar, baz} = foo;
```
matt-user, line 43: I'm pretty sure you don't need the `Foo` here.  You should just be able to write it as `let { bar, baz } = foo;`

sway-book/src/documentation/language/annotations/attributes/index.md
eureka-cpu: cc @mitchmindtree for that absolutely incredible PR

sway-book/src/code/language/built-ins/structs/src/lib.sw
eureka-cpu: This was fixed recently in the formatter, I think it just wasn't applied to `std`

eureka-cpu: And the `Foo` does still need to be there since it is destructuring `foo`

eureka-cpu: formatted it should look like ` let Foo { bar, baz } = foo;`

sway-book/src/documentation/language/control-flow/looping/index.md
eureka-cpu: *yet 😉

deekerno:
Reviewed the `Common Operations` section for now. Looks good. 👍🏽

eureka-cpu:
I think the book could use an [Appendix](https://doc.rust-lang.org/book/appendix-00.html) like the Rust book

sway-book/src/documentation/misc/advanced-concepts/structs.md
```diff
@@ -0,0 +1,5 @@
+# Struct Memory Layout
```
matt-user, line 1: Structs and enums are mentioned here and in "built in types". we probably only need one section

sway-book/src/documentation/misc/workarounds/missing-features.md
```diff
@@ -0,0 +1,14 @@
+# Missing Features
+
+> TODO: also copied from the current book (some old version). Needs work
+
+- [Issue: #1182](https://github.com/FuelLabs/sway/issues/1182)
+  - Arrays in a storage block are not yet supported. See the [Manual Storage Management](https://fuellabs.github.io/sway/v0.24.5/blockchain-development/storage.html#manual-storage-management) section for details on how to use `store` and `get` from the standard library to manage storage slots directly. Note, however, that `StorageMap<K, V>` does support arbitrary types for `K` and `V` without any limitations.
+- [Issue: #428](https://github.com/FuelLabs/sway/issues/428)
+  - Arrays are currently immutable which means that changing elements of an array once initialized is not yet possible.
+- [Issue: #1796](https://github.com/FuelLabs/sway/issues/1796)
```
matt-user, line 9: I don't think this is the correct issue.  I think we want this [one](https://github.com/FuelLabs/sway/issues/2465)

sway-book/src/documentation/misc/workarounds/importing.md
```diff
@@ -0,0 +1,16 @@
+# Importing
+
+In [external libraries](../../language/program-types/libraries/external.md) we have looked at how a library can be imported into a project so that code can be reused.
+
+When it comes to importing only external libraries can be imported and any other type of program will result in an error.
+
+This means that the following projects cannot be imported:
+
+- [contracts](../../language/program-types/contract.md)
+- [internal libraries](../../language/program-types/libraries/internal.md)
```
matt-user, line 10: Idk how I feel about us saying you can't import internal libraries.  IMO it's a bit confusing because don't we import them using the `dep` keyword.

sway-book/src/code/misc/known-issues/strings/src/lib.sw
```diff
@@ -0,0 +1,17 @@
...
+    // ANCHOR: single_quotes
+    // Will error if uncommented
+    // let fuel = 'fuel';
+    // ANCHOR_END: single_quotes
+}
+
+
+fn indexing() {
+    // ANCHOR: indexing
+    // Will error if uncommented
```
matt-user, line 13: I would move this below the next line

sway-book/src/documentation/language/functions/index.md
```diff
@@ -0,0 +1,24 @@
...
+The distinction between a function, method and associated function is as follows:
+
+- A [function](function.md) is a grouping of code that is independent of any object
+- A [method](method.md) is a function that is associated with an object and it uses `self` as the first parameter
+- An [associated function](associated-function.md) is a method but without the `self` parameter
+
+## Function Decleration
+
+A function decleration consists of a few components
+
```
matt-user, line 14: Should we include visibility specifiers here ie "pub" etc

sway-book/src/documentation/language/built-ins/numeric.md
```diff
@@ -0,0 +1,42 @@
...
+
+- -42
+- 0
+- 42
+
+In order to achieve this one _bit_ must be kept for tracking the sign (+ or -) of the value and thus the range of available values is smaller than an unsigned integer.
+
+For those inclined, the range for an n-bit signed integer is -2<sup>n-1</sup> to 2<sup>n-1</sup>-1.
+
+Sway does not natively support signed integers however there is nothing stopping a library from using primitives to create types that act like signed types.
```
simonr0204, line 21: Should we link to the existing library in sway-libs? On the one hand, if someone is looking for signed int we should point them in the right direction. On the other hand, we probably don't want the book to depend on types that are maintained in a different repo.

sway-book/src/documentation/language/functions/return.md
```diff
@@ -0,0 +1,39 @@
+# Returning from functions
+
+In the previous sections we have seen how functions return values without much information. In this section we will take a closer look at how we can return data from a function.
```
simonr0204, line 3: unclear what "without much information" refers to here. I _think_ you mean that we've seen it happen but haven't discussed it in much detail. But it could also be taken to mean that the return values didn't contain much information.

sway-book/src/documentation/language/functions/return.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Implicit Return
+
+To return from a function implicitly we do not use the `return` keyword and we omit the semi-colon at the end of the line.
+
+```sway
+{{#include ../../../code/language/functions/src/implicit.sw:main}}
+```
+
+Similarly to the explicit usage of a `return` this will typically be used at the end of a function but it can also be used anywhere.
```
simonr0204, line 35: A return expression will only return from the function if it is itself _inside an expression_ . If you're inside a statement (something ending with a `;`, more or less) then you need to explicitly `return;`

In my mental model, an "implicit return" is just a special case of the way expressions work in Sway (and Rust), rather than a thing in its own right.

sway-book/src/documentation/language/program-types/predicate.md
```diff
@@ -0,0 +1,17 @@
...
+
+It does not need to be deployed to a blockchain because it only exists during a transaction. That being said, the predicate hash (specifically the root) is on-chain in the UTXO set.
+
```
simonr0204, line 6: maybe "is on chain as the owner of one or more UTXOs"  is more precise. Also, it is the predicate _root_ rather than the hash.

sway-book/src/documentation/language/program-types/predicate.md
```diff
@@ -0,0 +1,17 @@
+# Predicates
```
simonr0204, line 1: In case it's useful, my attempt at explaining predicates on Fuel can be consulted [here](https://github.com/FuelLabs/sway-applications/tree/master/otc-swap-predicate#predicates-in-fuel)

sway-book/src/documentation/examples/contract-ownership.md
simonr0204: Also *privileged

sway-book/src/documentation/examples/subcurrency.md
```diff
@@ -0,0 +1 @@
+# Subcurrency
```
nfurfaro, line 1: Please, can we do away with the subcurrency example, unless we supplement it with a fuel native token example?
For new users, the subcurrency example makes it seem like ledger-based tokens are the recommended way to implement a token contract on forc.

sway-book/src/documentation/language/style-guide/enums.md
```diff
@@ -0,0 +1,25 @@
...
+
+```sway
+{{#include ../../../code/language/style-guide/enums/src/lib.sw:avoid}}
+```
+
+Key points to note:
+
+- You must import all of the enums i.e. `Error`, `StateError` & `UserError`
+- The lines may get unnecessarily long (depending on the names)
+- The syntax unergonomic
```
mohammadfawaz, line 25:
```suggestion
- The syntax is unergonomic
```

sway-book/src/documentation/language/variables/const.md
```diff
@@ -0,0 +1,17 @@
+# const
+
+A `const` is similar to an [immutable let](./let.md#immutable) however there are a few differences.
```
mohammadfawaz, line 3:
```suggestion
Constants are similar to [immutable let](./let.md#immutable) variables. However there are a few differences:
```

sway-book/src/documentation/language/variables/const.md
```diff
@@ -0,0 +1,17 @@
...
+
+- The constant is evaluated at compile-time
```
mohammadfawaz, line 5:
```suggestion
- Constants are always evaluated at compile-time
```

sway-book/src/documentation/language/variables/const.md
```diff
@@ -0,0 +1,17 @@
...
+- A constant can be declared in any scope i.e. inside of a [function](../functions/index.md) and outside
+- The `mut` keyword cannot be used
```
mohammadfawaz, line 7:
```suggestion
- The `mut` keyword cannot be used with constants.
```

sway-book/src/documentation/language/variables/const.md
```diff
@@ -0,0 +1,17 @@
+# const
+
+A `const` is similar to an [immutable let](./let.md#immutable) however there are a few differences.
+
+- The constant is evaluated at compile-time
+- A constant can be declared in any scope i.e. inside of a [function](../functions/index.md) and outside
```
mohammadfawaz, line 6:
```suggestion
- Constants can be declared both inside of a [function](../functions/index.md) and at global scope.
```

sway-book/src/documentation/language/variables/const.md
```diff
@@ -0,0 +1,17 @@
...
+- The `mut` keyword cannot be used
+
+## Declaration
+
+To define a constant the `const` keyword is used followed by a name and an assignment of a value.
```
mohammadfawaz, line 11:
```suggestion
To define a constant, the `const` keyword is used followed by a name and an assignment of a value. Providing the type of a `const` is optional.
```

sway-book/src/documentation/language/variables/const.md
```diff
@@ -0,0 +1,17 @@
+# const
```
mohammadfawaz, line 1: We should probably call this section "Constants".

sway-book/src/documentation/language/variables/index.md
```diff
@@ -0,0 +1,14 @@
+# Variables
+
+A variable is a way to reference some information by a specific name and it can take the form of a variety of [data structures](../built-ins/index.md).
+
+In Sway there are two states that a variable can take:
+
+- immutable
```
mohammadfawaz, line 7:
```suggestion
- Immutable
```

sway-book/src/documentation/language/variables/index.md
```diff
@@ -0,0 +1,14 @@
...
+  - Can be read but cannot be changed after it has been declared
```
mohammadfawaz, line 8: I would move this to the previous line:
" - Immutable: can be read but cannot be changed after it has been declared"

sway-book/src/documentation/language/variables/index.md
```diff
@@ -0,0 +1,14 @@
...
+- mutable
```
mohammadfawaz, line 9:
```suggestion
- Mutable
```

sway-book/src/documentation/language/variables/index.md
mohammadfawaz, line 10: Same.

sway-book/src/documentation/language/variables/index.md
```diff
@@ -0,0 +1,14 @@
...
+  - Can be read and have its value changed but only if the type is the same
+
+By default all variables in Sway are immutable unless declared as mutable through the use of the `mut` keyword. This is one of the ways how Sway encourages safe programming, and many modern languages have the same default.
```
mohammadfawaz, line 12:
```suggestion
By default all variables in Sway are immutable unless declared as mutable through the use of the `mut` keyword. This is one of the ways in which Sway encourages safe programming, and many modern languages have the same default.
```

sway-book/src/documentation/language/variables/index.md
```diff
@@ -0,0 +1,14 @@
...
+
+In the proceeding sections we'll take a look at two keywords that are used to instantiate information ([let](let.md) & [const](const.md)) and a way to temporarily reuse a variable name without affecting the original instantation through [variable shadowing](variable-shadowing.md).
```
mohammadfawaz, line 14:
```suggestion
In the following sections, we'll take a look at two keywords that are used to instantiate information ([let](let.md) & [const](const.md)) and a way to temporarily reuse a variable name without affecting the original instantiation through [variable shadowing](variable-shadowing.md).
```

sway-book/src/documentation/language/variables/let.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Immutable
+
+We can declare a variable that cannot have its value changed in the following way.
+
+```sway
+{{#include ../../../code/language/variables/src/lib.sw:immutable}}
+```
+
+By default `foo` is an immutable [`u64`](../built-ins/index.md#primitive-types) with the value of `5`. This means that we can pass `foo` around and its value can be read however it cannot have its value changed from `5` to any other number.
```
mohammadfawaz, line 13:
```suggestion
By default `foo` is an immutable [`u64`](../built-ins/index.md#primitive-types) with the value of `5`. This means that we can pass `foo` around and its value can be read, but it cannot have its value changed from `5` to any other value.
```

sway-book/src/documentation/language/variables/variable-shadowing.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Variable Shadowing
+
+If we do not want to alter the original variable but we'd like to temporarily reuse the variable name then we can use block scope to constrain the variable.
+
+```sway
+{{#include ../../../code/language/variables/src/lib.sw:shadowing}}
+```
+
+`foo` & `bar` defined inside the curly brackets `{}` only exist inside the `{}` scope therefore the original `foo` & `bar` with the value of `5` maintain their value.
```
mohammadfawaz, line 21:
```suggestion
`foo` and `bar` defined inside the curly brackets `{ }` only exist inside the `{ .. }` scope. Therefore, the original `foo` and `bar` variables with the value of `5` maintain their value.
```

sway-book/src/documentation/misc/advanced-concepts/enums.md
```diff
@@ -0,0 +1,7 @@
+# Enum Memory Layout
+
+Enums have some memory overhead. To know which variant is being represented, Sway stores a one-word (8-byte) tag for the enum variant.
+
+The space reserved after the tag is equivalent to the size of the _largest_ enum variant.
+
+To calculate the size of an enum in memory, add 8 bytes to the size of the largest variant. For example, in the [`Color`](../../language/built-ins/enums.md) example where the variants are all `()`, the size would be 8 bytes since the size of the largest variant is 0 bytes.
```
mohammadfawaz, line 7: It would be nice to provide a more complex example here such as an enum with variants of type `u64` and `b256`.

sway-book/src/documentation/misc/advanced-concepts/structs.md
```diff
@@ -0,0 +1,5 @@
...
+
+Structs have zero memory overhead meaning that each field is laid out sequentially in memory. No metadata regarding the struct's name or other properties is preserved at runtime.
```
mohammadfawaz, line 3:
```suggestion
Structs have zero memory overhead, meaning that each field is laid out sequentially in memory. No metadata regarding the struct's name or other properties is preserved at runtime.
```

sway-book/src/documentation/misc/workarounds/general.md
mohammadfawaz, line 6:
```suggestion
  - All `impl` blocks need to be defined before any of the functions they define can be called. This includes sibling functions in the same `impl` declaration, i.e., functions in an `impl` can't call each other yet.
```

sway-book/src/documentation/misc/workarounds/general.md
```diff
@@ -0,0 +1,7 @@
...
+- No compiler optimization passes have been implemented yet, therefore bytecode will be more expensive and larger than it would be in production. Note that eventually the optimizer will support zero-cost abstractions, avoiding the need for developers to go down to inline assembly to produce optimal code.
```
mohammadfawaz, line 7: Let's remove this as it is not correct anymore.

sway-book/src/documentation/misc/workarounds/importing.md
```diff
@@ -0,0 +1,16 @@
...
+- [scripts](../../language/program-types/script.md)
+- [predicates](../../language/program-types/predicate.md)
+
+Contracts can be imported however the workaround is to move the `ABI` into an [external library](../../language/program-types/libraries/external.md) and import that library.
```
mohammadfawaz, line 14:
```suggestion
While contracts cannot be imported, a workaround is to move the contract's `abi` declaration into an [external library](../../language/program-types/libraries/external.md) and import that library anywhere the ABI it is needed.
```

sway-book/src/documentation/operations/assertions/assert.md
```diff
@@ -0,0 +1,13 @@
+# assert
+
+The `assert` function is automatically imported into every program from the [prelude](../../misc/prelude.md) and it takes an expression which must evaluate to a [boolean](../../language/built-ins/boolean.md). If the boolean is `true` then nothing will happen and the code will continue to run otherwise the virtual machine will revert.
```
mohammadfawaz, line 3:
```suggestion
The `assert` function is automatically imported into every program from the [prelude](../../misc/prelude.md) and it takes an expression which must evaluate to a [Boolean](../../language/built-ins/boolean.md). If the Boolean is `true` then nothing will happen and the code will continue to run. Otherwise, the virtual machine will revert.
```

sway-book/src/documentation/operations/assertions/assert.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Example
+
+Here we have a function which takes two `u64` arguments and subtracts them. A `u64` cannot be negative therefore the assertion enforces `b` must be less than or equal to `a`.
+
+If the condition is not met then the virtual machine will revert.
```
mohammadfawaz, line 9:
```suggestion
If the condition is not met, then the virtual machine will revert.
```

sway-book/src/documentation/operations/assertions/assert.md
```diff
@@ -0,0 +1,13 @@
+# assert
+
+The `assert` function is automatically imported into every program from the [prelude](../../misc/prelude.md) and it takes an expression which must evaluate to a [boolean](../../language/built-ins/boolean.md). If the boolean is `true` then nothing will happen and the code will continue to run otherwise the virtual machine will revert.
+
+## Example
+
+Here we have a function which takes two `u64` arguments and subtracts them. A `u64` cannot be negative therefore the assertion enforces `b` must be less than or equal to `a`.
```
mohammadfawaz, line 7:
```suggestion
Here we have a function which takes two `u64` arguments and subtracts them. A `u64` cannot be negative therefore the assertion enforces that `b` must be less than or equal to `a`.
```

sway-book/src/documentation/operations/assertions/index.md
```diff
@@ -0,0 +1,14 @@
+# Assertions
+
+An assertion is a condition which must evaluate to the [boolean](../../language/built-ins/boolean.md) value of `true` and its purpose is to prevent undesirable computation when the condition is evaluated to `false`.
```
mohammadfawaz, line 3:
```suggestion
An assertion is a condition which must evaluate to the [Boolean](../../language/built-ins/boolean.md) value of `true` and its purpose is to prevent undesirable computation when the condition is evaluated to `false`.
```

sway-book/src/documentation/operations/assertions/index.md
```diff
@@ -0,0 +1,14 @@
...
+
+For example, a function may only work if the condition `argument < 5` is `true`. We can use an assertion to enforce this condition by:
+
+- Crashing the program when `5 <= argument`
```
mohammadfawaz, line 7:
```suggestion
- Forcing a revert in the program when `5 <= argument`
```

simonr0204:
As a general remark, I think the content is good and certainly much more complete than the original Sway book. It covers all the features with no stone left unturned, which is absolutely an improvement. The book structure is good and the code examples are clean.

I do think I prefer the "voice" (for want of a better term) of the original book. It mirrors the Rust book's "friendly" disposition which I think is also important. This is a harder thing to nail down as it's totally subjective, but I'd be interested in hearing what other reviewers think about this. If there's a consensus we can work out some concrete examples in order to guide the rest of the prose. Comments, anyone?

sway-book/src/documentation/language/program-types/predicate.md
Braqzen: There is a sentence below that mentions the state. Does that address this concern?

sway-book/src/documentation/language/functions/index.md
```diff
@@ -0,0 +1,24 @@
...
+Functions, and by extension methods and associated functions, are a way to group functionality together in a way that allows for code reuse without having to re-write the code in each place that it is used.
+
+The distinction between a function, method and associated function is as follows:
+
+- A [function](function.md) is a grouping of code that is independent of any object
+- A [method](method.md) is a function that is associated with an object and it uses `self` as the first parameter
+- An [associated function](associated-function.md) is a method but without the `self` parameter
+
+## Function Decleration
```
mitch-fuel, line 11:
```suggestion
## Function Declaration
```

sway-book/src/documentation/language/functions/index.md
```diff
@@ -0,0 +1,24 @@
...
+
+A function decleration consists of a few components
```
mitch-fuel, line 13:
```suggestion
A function declaration consists of a few components
```

sway-book/src/code/language/built-ins/tuples/src/lib.sw
```diff
@@ -0,0 +1,37 @@
...
+    let first = balances.0;
+    let second = balances.1;
+    // ANCHOR_END: index
+
+    // ANCHOR: internal_mutability
+    // 12 has the same type as 42 (u64) therefore this is valid
+    balances.0 = 12;
+
+    // true is a boolean and the tuple expects a u64 therefore this is invalid
+    // balances.0 = true;
```
bingcicle, line 20: is this supposed to be uncommented?

sway-book/src/documentation/language/functions/index.md
```diff
@@ -0,0 +1,24 @@
...
+
+- The `fn` keyword
+- A unqiue name for the function
```
bingcicle, line 16:
```suggestion
- A unique name for the function
```

sway-book/src/documentation/language/functions/associated-function.md
```diff
@@ -0,0 +1,27 @@
...
+
+In this example we will take a look at a struct however an enum will work in the same way.
```
bingcicle, line 9:
```suggestion
In this example, we will take a look at a struct. Note that an enum will work in the same way.
```

sway-book/src/documentation/language/control-flow/match/index.md
bingcicle, line 3:
```suggestion
[If expressions](if-expressions.md) can be used to check a large number of conditions. However, there is an alternative syntax which allows us to perform advanced pattern matching.
```

sway-book/src/documentation/language/control-flow/match/index.md
```diff
@@ -0,0 +1,13 @@
...
+
+A `match` expression matches on a variable and checks each case, also known as an `arm`, to see which branch of logic should be performed.
+
+The cases are checked sequentially in the order they are declared, i.e. from top to bottom, and the last arm must ensure that all cases in the pattern are covered otherwise the compiler will not know how to handle an unspecified pattern and it will error.
```
bingcicle, line 7:
```suggestion
The cases are checked sequentially in the order they are declared, i.e. from top to bottom, and the last arm must ensure that all cases in the pattern are covered otherwise the compiler will not know how to handle an unspecified pattern and will throw an error.
```

sway-book/src/documentation/operations/log.md
```diff
@@ -0,0 +1,24 @@
+# Logging
+
+Logging is a way to record data as the program runs.
+
+The [standard library](https://github.com/FuelLabs/sway/tree/master/sway-lib-std) provides a [`logging`](https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/logging.sw) module which contains a [generic](../language/generics/index.md) `log` function that is used to log a variable of any type.
+
+Each call to `log` appends 1 of 2 types of a [`receipt`](https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#receipt) to the list of receipts
```
bingcicle, line 7: Link doesn't work

sway-book/src/documentation/operations/log.md
```diff
@@ -0,0 +1,24 @@
...
+
+- [`Log`](https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#log-receipt)
```
bingcicle, line 9: Link doesn't work

sway-book/src/documentation/operations/log.md
```diff
@@ -0,0 +1,24 @@
...
+  - Generated for _non-reference_ types: `bool`, `u8`, `u16`, `u32`, and `u64`
+- [`LogData`](https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#logdata-receipt)
```
bingcicle, line 11: Link doesn't work

sway-book/src/documentation/operations/reentrancy.md
```diff
@@ -0,0 +1,32 @@
+# Re-entrancy
+
+Re-entrency occurs when a contract makes a call back into the contract that called it, e.g. `Contract A` calls `Contract B` but then `Contract B` makes a call back into `Contract A`.
```
bingcicle, line 3:
```suggestion
Re-entrancy 
```

bingcicle, line 3:
```suggestion
Re-entrancy occurs when a contract makes a call back into the contract that called it, e.g. `Contract A` calls `Contract B` but then `Contract B` makes a call back into `Contract A`.
```

sway-book/src/documentation/operations/reentrancy.md
```diff
@@ -0,0 +1,32 @@
...
+
+To mitigate security concerns there are two approaches that are commonly used:
+
+- [Implement a guard](#re-entrency-guard): detect when a re-entrency occurs
+- [Defensive programming](#checks-effects-interactions-pattern): perform calls after all state changes have been made
+
+## Re-entrency Guard
```
bingcicle, line 10:
```suggestion
## Re-entrancy Guard
```

sway-book/src/documentation/operations/asset/index.md
```diff
@@ -0,0 +1,22 @@
+# Asset Operations
+
+A common application of a smart contract is the creation of an asset / token i.e. a crytocurrency.
```
bingcicle, line 3:
```suggestion
A common application of a smart contract is the creation of an asset / token i.e. a cryptocurrency.
```

sway-book/book.toml
```diff
@@ -0,0 +1,18 @@
...
+authors = ["Fuel Labs <contact@fuel.sh>"]
+language = "en"
+multilingual = false
+src = "src"
+title = "The Sway Book"
+
+[output.html]
+git-repository-url = "https://github.com/FuelLabs/sway"
+
+[preprocessor.forc-documenter]
```
bingcicle, line 11:
```suggestion
```

We can remove the `forc-documenter` processor here since we probably are not documenting `forc` commands here(?)

sway-book/src/code/language/program-types/scripts/simple/src/main.sw
```diff
@@ -0,0 +1,6 @@
+script;
+
+// All scripts require a main function. The return type is optional.
+fn main(amount: u64) -> bool {
+    true
+}
```
bingcicle, line 6: An optional suggestion: what do you think about including a 'realistic' short script/predicate within the examples? As a new Sway developer this could give some context into what would be a useful script/predicate to write, instead of having an fn main that just returns true.

bingcicle:
As a general suggestion, how about a section at the beginning of the book introducing how the book should be read/used? I think it does seem more like a reference than a book as others have mentioned, perhaps a short section indicating how the book is intended to be consumed might be helpful?

sway-book/src/documentation/language/functions/associated-function.md
Braqzen: This breaks the flow of how the sentence is read. It becomes more of like a bullet point by bullet point explanation rather than a sentence.

sway-book/src/documentation/language/functions/function.md
Braqzen: Pretty sure names belong to people while things are `called` something

sway-book/src/documentation/language/program-types/predicate.md
```diff
@@ -0,0 +1,17 @@
...
+
+A predicate is an executable that represents a UTXO spending condition, such as a multisig predicate, which has restrictions on the VM instructions that can be used (e.g. no jumps).
+
+It does not need to be deployed to a blockchain because it only exists during a transaction. That being said, the predicate hash (specifically the root) is on-chain in the UTXO set.
+
+Similar to a [script](script.md), a predicate consists of a single `main()` function which can take any number of arguments but it must return a [boolean](../built-ins/boolean.md) and in order to be valid the returned boolean value must be `true`.
+
+Unlike scripts, predicates are stateless functions which means they can neither read from nor write to any contract state.
```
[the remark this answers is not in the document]
Braqzen: Scripts do not hold state but they can access it via contracts while predicates cannot even touch state

Braqzen (commented):
Addressing some comments.

sway-book/src/documentation/language/functions/associated-function.md
simonr0204: technically speaking, if you want this to be one sentence, it must be :

> In this example, we will take a look at a struct; however, an enum will work in the same way.

That said, I do not think any of the versions discussed will make or break the readability of the sentence. 🤷

sway-book/src/documentation/language/control-flow/match/index.md
simonr0204: As above, technically the only correct punctuation of "however" as a coordinating adverb is  

> ... ; however, ...

sway-book/src/documentation/language/variables/const.md
Braqzen: What's the purpose of mentioning the `optional` constant?

sway-book/src/documentation/language/variables/const.md
Braqzen: I've updated the casing however the latter part is left for the style guide.

sway-book/src/documentation/language/variables/index.md
Braqzen: I don't understand

sway-book/src/documentation/operations/hashing/index.md
```diff
@@ -0,0 +1,11 @@
...
+
+The [`hashing module`](https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/hash.sw) contains the following functions:
+
+<!-- no toc -->
+- [`sha256`](sha256.md)
+- [`keccak256`](keccak256.md)
+
+They take one [`generic`](../../language/generics/index.md) argument `T` and return a [`b256`](../../language/built-ins/b256.md) (hash of `T`). 
+
+To hash multiple values the values must be wrapped into one type such as a [`tuple`](../../language/built-ins/tuples.md), [`array`](../../language/built-ins/arrays.md), [`struct`](../../language/built-ins/structs.md) & [`enum`](../../language/built-ins/enums.md).
```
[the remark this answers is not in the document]
Braqzen: The index page of the built-ins mentions the default word size. Perhaps it needs to be moved and emphasised elsewhere. I'll leave this as it is for now since lots of sections are missing.

otrho (dismissed):
:shipit:

bingcicle:
I added a redirect file that redirects links to `/master` to default to `/master/book` + fixed the `publish_dir`s inputs for the workflow. 

Working example w/ workflow available: https://bingcicle.github.io/mdbook-ci-test/master - redirects to https://bingcicle.github.io/mdbook-ci-test/master/book ([repo](https://github.com/bingcicle/mdbook-ci-test))

otrho (approved):

mohammadfawaz (approved):

simonr0204 (approved):

Issue #2543, opened 2022-08-15
Some issues with `ref mut` parameters


mohammadfawaz:
### Issue 1
This does not compile due to an ICE:
```rust
script;

fn foo(ref mut x: u64) {
    x += 1;
}

fn main() {
    let mut x = 1;
    foo(x);
}
```
ICE:
```console
error: Internal compiler error: Verification failed: Type mismatch found for call to 'add_1'.
```

### Issue 2
The compiles but shouldn't because `x` is not `mut` (same problem if `x` is a reference type).
```rust
script;

fn foo(ref mut x: u64) {
    x = 1;
}

fn main() {
    let x = 1;
    foo(x);
}
```

### Issue 3
This fails with an ICE but I expect a clear compile error (`0` is not really mutable)
```rust
script;

fn foo(ref mut x: u64) {
    x = 1;
}

fn main() {
    foo(0);
}
```

### Issue 4
The following compiles but shouldn't. The `x` parameter of `foo()` should be `ref mut`.
```rust
script;

fn foo(x: u64) {
    bar(x)
}

fn bar(ref mut x: u64) {
    x = 1;
}

fn main() {
    let mut x = 0;
    bar(x);
}
```

mohammadfawaz:
@tritao assigning this issue to you as the person who's most familiar with this feature :)

otrho:
That ICE about `add_1` will be about the function recreation in IR gen `compile_fn_call()`.  I'm wondering whether it's a better idea to revert the change where I tried to avoid recreating the same function over and over again -- the hard bit is determining whether two functions are the same:
- The spans must be same.
- The monomorphised parameter and return types must be the same.
- The monomorphised type argument types must be the same.
- ??? More?  The above ICE may indicate further discrimination is needed.

otrho:
Issue 3 is contentious too.  Could we consider it to be equivalent to this (valid) Rust?
```rust
fn foo(x: &mut u64) {
    *x = 1;
}

fn main() {
    foo(&mut 0);
}
```

otrho:
> I gues we could make it work for Sway, but it doesn't seem particularly useful so I am just disallowing it.

It's useful because it means the caller doesn't have to create a temporary themselves if they don't need the mutated value after the call.  You'd be forcing them to do:
```rust
fn main() {
    let mut tmp = 0;
    foo(tmp);    // Don't care about tmp after here.
}
```
Another example is allowing the following:
```rust
fn a() -> u64 {
    22
}

[...]

PR #2563, opened 2022-08-17
Fix some issues with `ref mut` parameters

tritao:
This PR fixes the following issues originally identified by @mohammadfawaz.

### Issue 1
This does not compile due to an ICE:
```rust
script;

fn foo(ref mut x: u64) {
    x += 1;
}

fn main() {
    let mut x = 1;
    foo(x);
}
```
ICE:
```console
error: Internal compiler error: Verification failed: Type mismatch found for call to 'add_1'.
```

### Issue 2
The compiles but shouldn't because `x` is not `mut` (same problem if `x` is a reference type but `ref mut self` behaves as expected).
```rust
script;

fn foo(ref mut y: u64) {
    y = 1;
}

fn main() {
    let x = 1;
    foo(x);
}
```

### Issue 3
This fails with an ICE but I expect a clear compile error (`0` is not really mutable)
```rust
script;

fn foo(ref mut x: u64) {
    x = 1;
}

fn main() {
    foo(0);
}
```

### Issue 4
The following compiles but shouldn't. The `x` parameter of `foo()` should be `ref mut`.
```rust
script;

fn foo(x: u64) {
    bar(x)
}

fn bar(ref mut x: u64) {
    x = 1;
}

fn main() {
    let mut x = 0;
    bar(x);
}
```

### Issue 5
`ref mut` contract method parameters should not be allowed:
```rust
contract;

abi MyAbi {
    fn foo(ref mut x: u64);
}

impl MyAbi for Contract {
    fn foo(ref mut x: u64) {

    }
}
```

Closes https://github.com/FuelLabs/sway/issues/2543.


sway-core/src/semantic_analysis/ast_node/declaration/abi.rs
```diff
@@ -67,6 +67,16 @@ impl TypedAbiDeclaration {
             warnings,
             errors
         );
+        for typed_fn in &interface_surface {
+            for param in &typed_fn.parameters {
+                if param.is_reference {
+                    errors.push(CompileError::RefMutableNotAllowedInContractAbi {
+                        param_name: param.name.clone(),
+                    })
+                }
```
mohammadfawaz, line 76: How about 
```suggestion
                if param.is_reference && param.is_mutable {
                    errors.push(CompileError::RefMutableNotAllowedInContractAbi {
                        param_name: param.name.clone(),
                    })
                }
```
just for readability and if we decide to support only `mut` params or only `ref` params in the future?

sway-core/src/semantic_analysis/ast_node/declaration/abi.rs
```diff
@@ -75,6 +85,16 @@ impl TypedAbiDeclaration {
             warnings,
             errors
         );
+        for typed_fn in &methods {
+            for param in &typed_fn.parameters {
+                if param.is_reference {
```
mohammadfawaz, line 90: Same as above.

Basically, I think we should avoid the assumption that `ref` => `mut` or vice versa, even though that's the only thing we support today.

mohammadfawaz (approved):
Great! Thanks.

vaivaswatha (approved):

PR #2539, opened 2022-08-15
Adds better formatting to `match_branch_kind`

eureka-cpu:
This makes some changes to `CodeBlockContents` handling as well as `ItemFn` regarding new line handling, as well as indentation level for `MatchBranchKind`.

Issue #2493, opened 2022-08-09
[Tracking] Finish vertical formatting in `sfv2`


eureka-cpu:
Currently only `structs` and `tuples` are handled for multiline formatting. That still leaves: 
- [x] arrays (mostly copy & paste from tuple) this will also close #726 #2593 
- [x] control flow (`match`) #604 #1689
- [x] conditionals ( #2564 )
- [x] & chained operators e.g. #2597 
```rust
if really_long_var_name > other_really_long_var || really_long_var_name <= 0 && other_really_long_var != 0 { ... }
```
should be: 
```rust
if really_long_var_name > other_really_long_var
    || really_long_var_name <= 0 
    && other_really_long_var != 0
{
    ...
}
```
- [x] fn params e.g. #2566 
```rust
fn foo(really_long_var: String, other_really_long_var: String, hi_its_another_long_var: String) -> Result<(), Error> { ... }
```
should be:
```rust
fn foo(
really_long_var: String,
other_really_long_var: String,
hi_its_another_long_var: String,
) -> Result<(), Error> {
    ...
}
```

Braqzen:
Additional comment for **params** in version `0.24.4`.

The following code

```rust
fn claim(
    amount: u64,
    key: u64,
    num_leaves: u64,
    proof: [b256; 2],
    to: Identity,
);
```

turns into

```rust
fn claim(amount: u64, key: u64, num_leaves: u64, proof: [b256; 2], to: Identity, );
```

instead of the preferred

```rust
fn claim(amount: u64, key: u64, num_leaves: u64, proof: [b256; 2], to: Identity);
```

PR #2564, opened 2022-08-17
Adds formatting for `if_else` control flow blocks

eureka-cpu:
Ref #2493 

This will have to change slightly if #2539 gets in first and vice versa since it adds changes to `CodeBlockContents` which they both rely on.


Braqzen:
I may be misunderstanding something but it seems to me like the desired output of the test and the intermediate whitespace are duplicates. Shouldn't they be different so that a messy form gets formatted into the desired form?

eureka-cpu:
> I may be misunderstanding something but it seems to me like the desired output of the test and the intermediate whitespace are duplicates. Shouldn't they be different so that a messy form gets formatted into the desired form?

Right you are, I was expecting to have to resolve some conflicts so I got lazy. Thanks for the catch 😄

bingcicle:
When I test this with a sample `forc init` project and add the test code snippet into `main.sw`, the closing braces get pushed to the start of the line:

```
contract;

abi MyContract {
    fn test_function() -> bool;
}
    fn test_function() -> bool {
        if foo{
        foo();
}else if bar{
        bar();
}else {
        baz();
}
    }
}⏎       
```

Is this intended?

bingcicle (approved):
LGTM - tested some simple variations on a `forc init` project as well using the test code snippet too.

Braqzen (approved):

PR #2566, opened 2022-08-17
Adds multiline formatting for `fn_signature` in `sfv2`

eureka-cpu:
Ref #2493


Braqzen (commented):
Some comments

1. There seems to be some duplication of code e.g. `format_fn_args` in the `match`. I think the duplication should be refactored
2. The test seems to have been written and added across more than 1 PR so that multiple PRs could be submitted for review at once. Now that `match_branch` has gone in perhaps the duplicated code should be removed for the sake of keeping the PR cleaner? It doesn't really matter tbh but it makes reviewing the PRs a little easier if someone were to look back at the code. This could be ignored though
3. Something that caught my attention is that the function ends the formatting with a trailing `,` (comma) after the last argument. Since there are no more arguments I think it should really remove the last comma so that it ends with the arg followed by no spaces and the `)` e.g. `..., bar: Bar,     )` -> `..., bar: Bar)`

eureka-cpu:
> Some comments
> 
>     1. There seems to be some duplication of code e.g. `format_fn_args` in the `match`. I think the duplication should be refactored
> 
>     2. The test seems to have been written and added across more than 1 PR so that multiple PRs could be submitted for review at once. Now that `match_branch` has gone in perhaps the duplicated code should be removed for the sake of keeping the PR cleaner? It doesn't really matter tbh but it makes reviewing the PRs a little easier if someone were to look back at the code. This could be ignored though
> 
>     3. Something that caught my attention is that the function ends the formatting with a trailing `,` (comma) after the last argument. Since there are no more arguments I think it should really remove the last comma so that it ends with the arg followed by no spaces and the `)` e.g. `..., bar: Bar,     )` -> `..., bar: Bar)`

The traililng comma is there since it is a multiline format! And about the code dup, I can make a function called `format_self` and have that called in both places of that match statement.

bingcicle:
I'm trying it with:

```rs
contract;

abi MyContract {
    fn test_function() -> bool;
}
    fn test_function(
        helloitsverylong: String,
        whatisgoingonthisistoolong: String,
        yetanotherlongboy: String,
    ) -> bool {
        match foo {
            Foo::foo => true,
            _ => false,
        }
    }
}
```

After running `forc-fmt-v2` it seems to insert a newline after the opening parenthesis, even if I already have a newline:

```rs
contract;

abi MyContract {
    fn test_function() -> bool;
}
    fn test_function(

        helloitsverylong: String,
        whatisgoingonthisistoolong: String,
        yetanotherlongboy: String,
    ) -> bool {
        match foo {
            Foo::foo => true,
            _ => false,
        }
    }
}
```

Also, this might be unrelated but `forc-fmt-v2` doesn't insert a `,` at the end of the final match arm as part of formatting, but `rustfmt` does so. Is this intentional behaviour?

sway-fmt-v2/src/items/item_fn.rs
```diff
@@ -70,103 +78,201 @@ impl CurlyBrace for ItemFn {
...
         line: &mut FormattedCode,
-        _formatter: &mut Formatter,
+        formatter: &mut Formatter,
     ) -> Result<(), FormatterError> {
-        write!(line, "{}", Delimiter::Parenthesis.as_open_char())?;
+        let open_paren = Delimiter::Parenthesis.as_open_char();
+        match formatter.shape.code_line.line_style {
+            LineStyle::Multiline => {
+                formatter.shape.block_indent(&formatter.config);
+                writeln!(line, "{open_paren}")?;
```
bingcicle, line 248: Perhaps this line is the culprit of my findings, should this be `write!` instead of `writeln!`?

eureka-cpu: could've swore I got rid of that one, since the tests passed..... weird

eureka-cpu: Ah, I see, my tests all have `self` as the first arg.

eureka-cpu:
> Also, this might be unrelated but `forc-fmt-v2` doesn't insert a `,` at the end of the final match arm as part of formatting, but `rustfmt` does so. Is this intentional behaviour?

It seems to be doing it here in this code example. Do you mean if it's wrapped in curly braces? rustfmt does have an option to add trailing commas after curly braced match statements but I didn't implement that by default

bingcicle (approved):
> Also, this might be unrelated but forc-fmt-v2 doesn't insert a , at the end of the final match arm as part of formatting, but rustfmt does so. Is this intentional behaviour?

~I'm still curious about this^~ You frontran my comment! Nope, I just mean that it seems like trailing commas are added to the final match arm like so:

Before:

```rs
match foo {
  Foo::foo => true,
  _ => false
}
```

After:

```rs
match foo {
  Foo::foo => true,
  _ => false, // added here
}
```

But `forc-fmt-v2` fails to parse entirely given a `.sw` file in the __Before__ state, unless you append the `,`.

With that said, the above seems unrelated to this PR, so I'm giving this the tick here 👍

eureka-cpu:
> > Also, this might be unrelated but forc-fmt-v2 doesn't insert a , at the end of the final match arm as part of formatting, but rustfmt does so. Is this intentional behaviour?
> 
> I'm still curious about this^
> 
> With that said, the above seems unrelated to this PR, so I'm giving this the tick here 👍

Looks like we may have some time since I have some conflicts to resolve, would you mind adding an example?

bingcicle:
> Looks like we may have some time since I have some conflicts to resolve, would you mind adding an example?

Edited my original comment above

eureka-cpu:
> > Also, this might be unrelated but forc-fmt-v2 doesn't insert a , at the end of the final match arm as part of formatting, but rustfmt does so. Is this intentional behaviour?
> 
> ~I'm still curious about this^~ You frontran my comment! Nope, I just mean that it seems like trailing commas are added to the final match arm like so:
> 
> Before:
> 
> ```rust
> match foo {
>   Foo::foo => true,
>   _ => false
> }
> ```
> 
> After:
> 
> ```rust
> match foo {
>   Foo::foo => true,
>   _ => false, // added here
> }
> ```
> 
> But `forc-fmt-v2` fails to parse entirely given a `.sw` file in the **Before** state, unless you append the `,`.
> 
> With that said, the above seems unrelated to this PR, so I'm giving this the tick here 👍

This should keep trailing commas

bingcicle:
> This should keep trailing commas

I meant that the trailing comma isn't appended by default but the formatter runs into 

```
Error parsing file: Unable to parse: Expected `,`.
```

Given something in the __Before__ state.

eureka-cpu:
> > This should keep trailing commas
> 
> I meant that the trailing comma isn't appended by default but the formatter runs into
> 
> ```
> Error parsing file: Unable to parse: Expected `,`.
> ```
> 
> given something in the **Before** state.

Ah, I see what you're saying now. Unfortunately, at the moment it's up to the user to add that as part of the language syntax and not the formatter. I do think this could be an issue you could open for the language team to tackle, and allow non-trailing commas on the ending match block. In the future, we would want it to be able to add commas there if they were missing but presently we can only format if the parsing succeeds.

JoshuaBatty (approved):

PR #2568, opened 2022-08-17
Add LSP tests back in.

JoshuaBatty:
These were temporarily removed in #2519 as some tests were getting stuck in CI. 

Also adds tests for the following LSP capabilities. 

- documentHighlight
- hover
- formatting
- definition
- documentSymbol
- semanticTokens

### DIFF
--- a/sway-lsp/src/server.rs
+++ b/sway-lsp/src/server.rs
@@ -90,6 +90,7 @@ fn capabilities() -> ServerCapabilities {
         }),
         document_formatting_provider: Some(OneOf::Left(true)),
         definition_provider: Some(OneOf::Left(true)),
+        hover_provider: Some(HoverProviderCapability::Simple(true)),
         ..ServerCapabilities::default()
     }
 }
@@ -376,8 +377,10 @@ mod tests {
 
     use super::*;
     use futures::stream::StreamExt;
-    use tower_lsp::jsonrpc::{self, Request, Response};
-    use tower_lsp::LspService;
+    use tower_lsp::{
+        jsonrpc::{self, Request, Response},
+        {ClientSocket, LspService},
+    };
 
     #[allow(dead_code)]
     fn e2e_test_dir() -> PathBuf {
@@ -398,8 +401,17 @@ mod tests {
             .join("examples/storage_variables")
     }
 
-    fn load_sway_example() -> (Url, String) {
-        let manifest_dir = e2e_test_dir();
+    #[allow(dead_code)]
+    fn doc_comments() -> PathBuf {
+        env::current_dir()
+            .unwrap()
+            .parent()
+            .unwrap()
+            .join("test/src/e2e_vm_tests/test_programs/should_pass/language")
+            .join("doc_comments")
+    }
+
+    fn load_sway_example(manifest_dir: PathBuf) -> (Url, String) {
         let src_path = manifest_dir.join("src/main.sw");
         let mut file = fs::File::open(&src_path).unwrap();
         let mut sway_program = String::new();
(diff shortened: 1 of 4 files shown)


sway-lsp/src/server.rs
```diff
@@ -463,6 +474,23 @@ mod tests {
...
+    ) -> Request {
+        let did_change = Request::build("textDocument/didChange")
+            .params(params)
+            .finish();
+        let response = service
+            .ready()
+            .await
+            .unwrap()
+            .call(did_change.clone())
+            .await;
```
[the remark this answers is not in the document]
JoshuaBatty: Nice! :)

sway-lsp/src/server.rs
```diff
@@ -375,31 +375,39 @@ mod tests {
...
+
+    fn sway_workspace_dir() -> PathBuf {
+        env::current_dir().unwrap().parent().unwrap().to_path_buf()
+    }
+
+    fn e2e_language_dir() -> PathBuf {
+        PathBuf::from("test/src/e2e_vm_tests/test_programs/should_pass/language")
+    }
 
     #[allow(dead_code)]
```
sdankel, line 392: Why is this needed?

sway-lsp/src/server.rs
```diff
@@ -476,247 +504,418 @@ mod tests {
...
+                "line": 44,
+                "character": 24
+            }
+        });
+        let definition = build_request_with_id("textDocument/definition", params, 1);
+        let response = call_request(service, definition.clone()).await;
+        let ok = Response::from_ok(
+            1.into(),
+            json!({
+                "range": {
```
sdankel, line 551: nit: it'd be easier to read the tests if these long-ish json blobs were imported from another file

sway-lsp/src/server.rs
```diff
@@ -476,21 +504,177 @@ mod tests {
...
+        let (uri, sway_program) = load_sway_example(manifest_dir);
+
+        // send "textDocument/didOpen" notification for `uri`
+        did_open_notification(service, &uri, &sway_program).await;
+
+        uri
+    }
+
+    async fn shutdown_and_exit(service: &mut LspService<Backend>) {
+        // send "shutdown" request
```
sdankel, line 669: nit: remove these comments as they aren't adding much

sway-lsp/src/server.rs
```diff
@@ -486,237 +514,448 @@ mod tests {
...
+                let (mut service, _) = LspService::new(|client| Backend::new(client, config()));
+                let uri = init_and_open(&mut service, $example_dir).await;
+                // Call the specific LSP capability function that was passed in.
+                let _ = $capability(&mut service, &uri).await;
+
+                shutdown_and_exit(&mut service).await;
+            });
+            rt.shutdown_background();
+        }};
+    }
```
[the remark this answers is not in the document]
mitchmindtree: We had a good go at this @Centril , but it would require being able to pass an `async` function as an argument to the higher order `test_lsp_capability` function, which is a bit tricky at the moment as currently Rust only supports `impl Trait` in function argument and return type position (i.e. not in trait position, which would be required for the function argument). But feel free to have a go at it!

As a result, I suggested going with a small well-documented `macro_rules!` instead seeing as it's not publicly exposed and only used by the tests.

sdankel (approved):
Nice cleanup of the tests! Excited to see these running in CI.

sway-lsp/src/server.rs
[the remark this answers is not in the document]
JoshuaBatty: As @mitchmindtree stated above, this is as far as we got. Would love to know if you could find a solution though. 

```rust 
pub async fn test_lsp_cap<F, Fut>(example_dir: PathBuf, capability: F)
    where
        F: Fn(&mut LspService<Backend>, &Url) -> Fut,
        Fut: std::future::Future<Output = Request> + 'static + Send,
{
    let (mut service, _) = LspService::new(|client| Backend::new(client, config()));
    let uri = init_and_open(&mut service, example_dir).await;
    // Call the specific LSP capability function that was passed in.
    let _ = capability(&mut service, &uri).await;
    shutdown_and_exit(&mut service).await;
}

#[tokio::test]
async fn semantic_tokens() {
    test_lsp_cap(doc_comments_dir(), semantic_tokens_request);
}
```

trying to compile this I get the following error,
```
error[E0308]: mismatched types
   --> sway-lsp/src/server.rs:858:9
    |
858 |         test_lsp_cap(doc_comments_dir(), semantic_tokens_request);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ one type is more general than the other
    |
    = note: expected trait `for<'r, 's> <for<'r, 's> fn(&'r mut LspService<server::Backend>, &'s tower_lsp::lsp_types::Url) -> impl Future<Output = tower_lsp::jsonrpc::Request> {semantic_tokens_request} as FnOnce<(&'r mut LspService<server::Backend>, &'s tower_lsp::lsp_types::Url)>>`
               found trait `for<'r, 's> <for<'r, 's> fn(&'r mut LspService<server::Backend>, &'s tower_lsp::lsp_types::Url) -> impl Future<Output = tower_lsp::jsonrpc::Request> {semantic_tokens_request} as FnOnce<(&'r mut LspService<server::Backend>, &'s tower_lsp::lsp_types::Url)>>`
note: the lifetime requirement is introduced here
   --> sway-lsp/src/server.rs:843:50
    |
843 |         F: Fn(&mut LspService<Backend>, &Url) -> Fut,
    |                                                  ^^^

For more information about this error, try `rustc --explain E0308`.
```

sway-lsp/src/server.rs
JoshuaBatty: just a convenience function as this path is called in 2 other places.

sway-lsp/src/server.rs
JoshuaBatty: good call, just removed them.

sway-lsp/src/server.rs
JoshuaBatty: That's certainly a good possible option. Perhaps we could refactor the tests to live in their own folder within LSP and we could have a related json file for the tests under that test folder as well. Let's chat more about this and do this in a separate PR if we decide to move ahead.

mohammadfawaz (approved):

PR #2569, opened 2022-08-18
Collect SymbolKind info and generate correct semantic tokens

JoshuaBatty:
This PR introduces a new SymbolKind enum which is stored in the `Token` type. We are populating the kind of each token during the parse-tree traversal stage. This makes it a lot cleaner when going from our Token type to the various `lsp_types` enums such as `CompletionItemKind`, `SemanticTokenType` etc..

We are now providing accurate semantic token information to the client which is resulting in much nicer syntax highlighting for sway files.

### DIFF
--- a/sway-lsp/src/capabilities/semantic_tokens.rs
+++ b/sway-lsp/src/capabilities/semantic_tokens.rs
@@ -15,7 +15,7 @@ pub fn semantic_tokens_full(session: &Session, url: &Url) -> Option<SemanticToke
     let tokens = session.tokens_for_file(url);
 
     // The tokens need sorting by thier span so each token is sequential
-    // If this is done then the bit offsets used for the lsp_types::SemanticToken are incorrect.
+    // If this step isn't done, then the bit offsets used for the lsp_types::SemanticToken are incorrect.
     let mut tokens_sorted: Vec<_> = tokens
         .iter()
         .map(|item| {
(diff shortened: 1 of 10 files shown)


sway-lsp/src/capabilities/semantic_tokens.rs
```diff
@@ -1,174 +1,174 @@
...
-#[repr(u32)]
-enum TokenTypeIndex {
-    Function = 1,
-    Parameter = 5,
-    String = 6,
-    Variable = 9,
-    Enum = 10,
-    Struct = 11,
-    Interface = 12,
+static TOKEN_RESULT_COUNTER: AtomicU32 = AtomicU32::new(1);
```
mitchmindtree, line 96: Rather than using a global, is it possible to track this in the Session or something along those lines?

That said this looks pretty harmless, and appears to only ever increment upon finishing each token collection. If it's easier to keep it, maybe we could move this static into the `semantic_tokens` function to make it clear that that's the only place it is used?

mitchmindtree (approved):
Is there any chance that we are able to infer the `SymbolKind` from the `AstToken` rather than storing both side-by-side for each `Token`? Then we could have a method on `AstToken` like `fn symbol_kind(&self) -> SymbolKind` or something along these lines?

That said, if it's not always possible to infer the `SymbolKind` purely from the `AstToken` and you need the extra context during token collection, then the current approach looks fine!

I've left another comment about the new static variable, but will leave that up to you if you think it's worth addressing.

JoshuaBatty:
Yeah, my original idea was to infer the SymbolKind from the AstToken but it's just so much easier to infer the correct symbol kind during the token collection phase that it makes more sense to use the current approach.

sway-lsp/src/capabilities/semantic_tokens.rs
JoshuaBatty: Yeah cool, i'll move it into the `semantic_tokens` function.

mitchmindtree (approved):

sezna (approved):

Issue #2487, opened 2022-08-09
Pretty print the AST from sway-core


JoshuaBatty:
It would be nice if we could implement a pretty printer for the AST produced from sway-core. 

Currently, it's pretty hard to read when using a `println!` on the return AST in rust. Alex mentioned he had used the [ptree](https://docs.rs/ptree/latest/ptree/) crate before. Perhaps it might be worth also looking at leveraging this crate for pretty printing.

PR #2570, opened 2022-08-18
Provide LSP extension for printing the parsed or typed AST in VS Code

JoshuaBatty:
This provides a custom LSP extension method that the client can call when either a "show parsed ast" or a "show typed ast" command is executed from the command palette. 

For example, in VS Code if you executed a "show typed ast" command while working on a sway file, a second window will open up containing the formatted `TypedProgram` of the current file. 

Once #2487 happens we can update the VS Code plugin with options for pretty printing and verbosity levels etc..


kayagokalp (approved):
utACK

sezna (approved):

PR #2571, opened 2022-08-18
Fix bug in serializing strings for storage

### LINKED ISSUE #2458: Assertion failure on a simple storage declaration with a string variable
Example
```rust
contract;

storage {
    name: str[4] = "test",
}
```
Error:
```rust
thread 'main' panicked at 'assertion failed: s.len() % 8 == 0', sway-core/src/ir_generation/storage.rs:191:13
```
Interrestingly:
* Sizes 1, 2, 3, 4 fail
* Sizes 5, 6, 7, 8 pass
* Sizes 9, 10, 11, 12 fail,
... and so on

Note: the following test fails because of the above: https://github.com/FuelLabs/sway/tree/master/test/src/e2e_vm_tests/test_programs/should_pass/language/storage_declaration but it's disabled. It should be re-enabled (but without `fn main()` as it is a a contract and not a script)

vaivaswatha:
Also remove the `storage_declaration` test as it was disabled, and now, with this PR, covered by the `basic_storage` test.

Closes #2458


sezna (commented):
guess we don't have to pad to the word for storage, although does this work if you reassign? e.g. if you reassign slot 3 to `b`, does it become `b` or `baa`?

vaivaswatha:
> guess we don't have to pad to the word for storage, 
We need to pad. That's done in two stages. (1) convert the byte sequence to word sequence. (2) Convert the word sequence to quad-word sequence. The bug was that (1) and (2) were both doing (2). So now I changed (1) to do what it must.

>although does this work if you reassign? e.g. if you reassign slot 3 to `b`, does it become `b` or `baa`?
The type of the slot is fixed right? (and `str[1] != str[3]`).

sezna:
Ah yeah all of that makes sense. Type safety does defend us there.

sezna (approved):

PR #2573, opened 2022-08-18
Fixing the defaults for `forc new`

### LINKED ISSUE #2567: `forc new` generates code that asks for too many coins
[This bumping of the sdk version to 0.20](https://github.com/FuelLabs/sway/commit/3b9413f4ba21c7d12d06f371adf25046e389fa7b#diff-b3cc1ff7749f6a5157a3649a13c1f7eb820555bab130abcb267445e29674e659R104) changed the template for the `harness.rs` to ask for one wallet with a million coins each having a million amount. 

This causes the node initialization to hang due to the work needed to initialize the coins.

Originally there was only one coin generated with a million amount.

Investigate why this change was made and, if deemed needless/over the top, revert it to a more sensible value.

mohammadfawaz:
Closes #2567

- Changing the number of coins to 1 so that the test doesn't hang while trying to create 1_000_000 coins.
- Adding comments to make it clear what `WalletsConfig` is doing.

### DIFF
--- a/forc/src/utils/defaults.rs
+++ b/forc/src/utils/defaults.rs
@@ -101,7 +101,11 @@ abigen!(MyContract, "out/debug/"#,
 async fn get_contract_instance() -> (MyContract, ContractId) {
     // Launch a local network and deploy the contract
     let mut wallets = launch_custom_provider_and_get_wallets(
-        WalletsConfig::new(Some(1), Some(1000000), Some(1_000_000)),
+        WalletsConfig::new(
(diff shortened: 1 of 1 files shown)


hal3e (approved):

kayagokalp (approved):

adlerjohn:
It looks like this isn't in a release yet. Could a new patch release of `forc` be released to include this @mohammadfawaz, since it's a pretty big blocker to actually making new projects usable?

PR #2582, opened 2022-08-19
Fix binary release; add verify tag

bingcicle:
CI failed packaging binaries - https://github.com/FuelLabs/sway/runs/7908826524?check_suite_focus=true#step:12:20

- Fix the build procedure for forc binaries (`forc-deploy` and `forc-run` are the actual names of bins installed, not `forc-client`)
- Add `verify-tag` step for `forc-client`


mohammadfawaz (approved):
Thanks!

adlerjohn (approved):
Can you file an issue to maybe do a dry-run of as much as this as possible if the PR has something like a `release` tag?

PR #2583, opened 2022-08-19
Replace old JSON ABI format with the new one

mohammadfawaz:
Removing everything related to the old JSON ABI format and update all the tests.

Remaining tasks:
- [ ] Bump the Rust SDK tests to use the newer version of the SDK that will include support for the new JSON ABI format
- [ ] Bump the `fuels` version in `forc new`


mohammadfawaz:
This PR cannot be merged until we have a new version of `fuels` that supports the new JSON ABI format.

PR #2588, opened 2022-08-19
Annotate ItemStorage fields

AlicanC:
This PR adds support for attaching Attributes to storage fields, similar to #2528 which adds this support for enum and struct fields.


sezna:
I notice both this PR and #2528 say `TODO -- handle annotations` -- do we just parse them and discard them right now? If so, could you file an issue to handle adding them to the AST?

mohammadfawaz (approved):
Seems reasonable. Thanks!

AlicanC:
> I notice both this PR and #2528 say `TODO -- handle annotations` -- do we just parse them and discard them right now? If so, could you file an issue to handle adding them to the AST?

Yup, that's the case. Created the issue: https://github.com/FuelLabs/sway/issues/2589

sezna (approved):
Thanks!

PR #2592, opened 2022-08-19
Allow script main fns to have args

AlicanC:
This PR adds support for main args:
```rs
script;

struct TestStruct {
    val: u64,
}

fn main(baba: TestStruct, keke: u64) -> u64 {
    baba.val + keke
}
```

TODO:
- [x] This PR needs GTF so it might get blocked by #2482 if other items are finished before it lands on master. When it lands, the incomplete GTF support commit in this PR should be removed.
- [x] Predicate support.
- [x] Checking if the length of script data is correct.
- [x] ~~Implement copying, test mutating the args.~~ Disallow `ref mut` args in main fns.
- [ ] Docs
- [x] Test w/ enums, strings, arrays


mohammadfawaz:
Do you mind adding a small section in the Sway book about this? Potentially [here](https://fuellabs.github.io/sway/v0.22.0/sway-program-types/scripts.html?highlight=main()#scripts-and-the-sdks). We basically should mention that `main()` does accept arguments, that those arguments cannot be mutable, and where those arguments live (in script data, maybe a link to the specs?). Also, is there a way to exercise this feature today using `forc`? Or nothing can be done without support from the SDK?

test/src/e2e_vm_tests/test_programs/should_pass/language/main_args/main_args_mutation/test.toml
```diff
@@ -0,0 +1,4 @@
+category = "fail"
```
mohammadfawaz, line 1: Should test be moved to `should_fail`?

sway-core/src/asm_generation/from_ir.rs
```diff
@@ -346,6 +367,77 @@ impl<'ir> AsmBuilder<'ir> {
...
+        let data_len_reg = self.reg_seqr.next();
+        self.bytecode.push(Op {
+            opcode: either::Either::Left(VirtualOp::GTF(
+                data_len_reg.clone(),
+                VirtualRegister::Constant(ConstantRegister::Zero),
+                VirtualImmediate12 {
+                    value: GTFArgs::ScriptDataLength as u16,
+                },
+            )),
+            comment: "Get length of script data".into(),
```
mohammadfawaz, line 382: Are main args the only script data we have or are the SDKs allowed to have more things in there (such as the content of a `Vec` in a contract's ABI?)

sway-core/src/error.rs
```diff
@@ -1058,12 +1058,12 @@ pub enum CompileError {
     BreakOutsideLoop { span: Span },
     #[error("\"continue\" used outside of a loop")]
     ContinueOutsideLoop { span: Span },
-    #[error("arguments to \"main()\" are not yet supported. See the issue here: github.com/FuelLabs/sway/issues/845")]
-    MainArgsNotYetSupported { span: Span },
     #[error("Configuration-time constant value is not a constant item.")]
     ConfigTimeConstantNotAConstDecl { span: Span },
     #[error("Configuration-time constant value is not a literal.")]
     ConfigTimeConstantNotALiteral { span: Span },
+    #[error("ref mut parameters are not allowed for script or predicate main functions.")]
```
mohammadfawaz, line 1065:
```suggestion
    #[error("ref mut parameter not allowed for main()")]
```

sway-core/src/asm_generation/from_ir.rs
```diff
@@ -346,6 +367,77 @@ impl<'ir> AsmBuilder<'ir> {
...
+                cmp_result_reg.clone(),
+                data_len_reg,
+                expected_data_len_reg,
+            )),
+            comment: "Compare lengths".into(),
+            owning_span: None,
+        });
+        let label = self.reg_seqr.get_label();
+        self.bytecode.push(Op {
+            opcode: Either::Right(OrganizationalOp::JumpIfNotEq(
```
mohammadfawaz, line 411: It's unfortunate that we have to pay for this in the bytecode itself. What happens if we don't do these checks here?

mohammadfawaz (commented):
Other than my comments about script data and other things it may contain, this looks great! I like the reuse of code. 

Regarding testing, I'd feel a bit more confident with the change if you can add examples that use other complex data types such as enums, strings, and arrays.

sway-core/src/asm_generation/from_ir.rs
AlicanC: Yep. That's the reason the check isn't an EQ but a GTE. In fuels-ts's Multicall we use this for sending some variable length data:

https://github.com/FuelLabs/fuels-ts/blob/5d4d6ce7fa1a23deae3f41be94c9fe2ee9851772/packages/contract/src/contracts/multicall/src/main.sw#L53-L56

sway-core/src/asm_generation/from_ir.rs
AlicanC: Nothing really. I was just following what's described in the issue. This whole length-checking feels like a crime against the gas golfing community :D

test/src/e2e_vm_tests/test_programs/should_pass/language/main_args/main_args_mutation/test.toml
AlicanC: But it's so good that all these tests related to this feature are in one place. Do we even need the `should_fail`/`should_pass` separation? Separating by feature seems more reasonable.

AlicanC: But it's so good that all these tests related to this feature are in one place. Do we even need the `should_fail`/`should_pass` separation? Separating by feature seems more reasonable.

AlicanC:
TODO:

- [ ] Docs
- [ ] Test w/ enums, strings, arrays

AlicanC:
TODO:

- [ ] Docs
- [ ] Test w/ enums, strings, arrays

sway-core/src/asm_generation/from_ir.rs
mohammadfawaz: This seems a little fragile anyways because script data could contain other things and this check may pass even in cases where the user has made a mistake. In other words, this check is necessary but not sufficient and paying gas for it, as you said, feels like a crime against the gas golfing community 😛 .

For users interacting with the script from the SDK, this shouldn't be a problem right?

AlicanC: At this time, IDK, but we do generate ABIs and SDKs can check and warn if the user is not passing the right data. (So yes, we can easily handle this at the SDK level instead.)

ATM, we will need an escape hatch for variable length data though, which won't be a problem when we support `Vec` coding in SDKs.

AlicanC: At this time, IDK, but we do generate ABIs and SDKs can check and warn if the user is not passing the right data. (So yes, we can easily handle this at the SDK level instead.)

ATM, we will need an escape hatch for variable length data though, which won't be a problem when we support `Vec` coding in SDKs.

test/src/e2e_vm_tests/test_programs/should_pass/language/main_args/main_args_mutation/test.toml
sezna: I'm okay with unifying the folders, `should_fail` and `should_pass` are remnants of a pre-test.toml world. But yeah this PR should follow current practices and we can do a sweeping update afterwards.

AlicanC: Created this issue for discussion: https://github.com/FuelLabs/sway/issues/2690

AlicanC:
> Do you mind adding a small section in the Sway book about this? Potentially [here](https://fuellabs.github.io/sway/v0.22.0/sway-program-types/scripts.html?highlight=main()#scripts-and-the-sdks). We basically should mention that `main()` does accept arguments, that those arguments cannot be mutable, and where those arguments live (in script data, maybe a link to the specs?). Also, is there a way to exercise this feature today using `forc`? Or nothing can be done without support from the SDK?

Both SDKs have a way of providing script data, but (at least on the TS side) the user is responsible of encoding both the args and the return value. I could doc this mentioning the `script_data` field but it will be confusing when SDKs eventually start managing this for the user. Maybe we should wait for SDK support to doc this. (Or I could just add some args to the example `main()` to signal support without mentioning how to get the data there.)

sezna (approved):
lgtm, although we should make an issue regarding avoiding the len check in bytecode when possible

AlicanC:
So I've:
- Added a new test that involves strings, enums, tuples, structs and arrays: https://github.com/FuelLabs/sway/pull/2592/files#diff-4b244577e2be7c4b79eed1f45ece8b31fa082219e46ea034e3e8ac2a169dd026
- Moved the failing test to `should_fail` directory.

What should I do with:
- Docs? Leave it for another PR when SDKs support script args?
- Length checking? Remove and let the SDKs handle it?

sezna:
Yeah we can postpone docs for now. Will circle back on the runtime check...

mohammadfawaz (approved):

sezna (approved):

PR #2593, opened 2022-08-19
Adds multiline formatting for `array`

### LINKED ISSUE #726: `forc fmt` doesn't format arrays very well.
Here's what it did to a recent PR of mine:
```diff
diff --git a/src/b512.sw b/src/b512.sw
index 2bcf573..ff5f664 100644
--- a/src/b512.sw
+++ b/src/b512.sw
@@ -4,7 +4,8 @@ library b512;
 /// Stores two b256s in contiguous memory.
 /// Guaranteed to be contiguous for use with ec-recover: std::ecr::ec_recover().
 pub struct B512 {
-    bytes: [b256; 2],
+    bytes: [b256;
+    2],
 }

 // @todo use generic form when possible
@@ -19,7 +20,8 @@ pub trait From {
 impl From for B512 {
     fn from(h: b256, l: b256) -> B512 {
         B512 {
-            bytes: [h, l],
+            bytes: [h,
+            l],
         }
     }
 }
@@ -30,7 +32,7 @@ impl B512 {
     fn new() -> B512 {
         B512 {
             bytes: [0x0000000000000000000000000000000000000000000000000000000000000000,
-                    0x0000000000000000000000000000000000000000000000000000000000000000],
+            0x0000000000000000000000000000000000000000000000000000000000000000],
         }
     }
 }
```

eureka-cpu:
Closes #726 
Ref #2493


Braqzen (commented):
1. How come a tuple is being tested in an array PR?
2. Should there be another test for 

```rust
[ some, args, here, ...,
and, more,
here ]
```

and a few variations of that?

eureka-cpu:
> 1. How come a tuple is being tested in an array PR?
> 
>     2. Should there be another test for
> 
> 
> ```rust
> [ some, args, here, ...,
> and, more,
> here ]
> ```
> 
> and a few variations of that?

I can add it in another PR, they're both collections and use generally the same logic so I added it while testing to be sure they both functioned

eureka-cpu:
@Braqzen I'm not sure I understand the second part, if you'd like I can add an explicit example though! Currently we have either normal formatting
```rust
[some, thing, here, ...]
```
and once it reaches a threshold it gets vertical formatting
```rust
[
    some,
    thing,
    here,
    ...,
]
```

Braqzen:
> @Braqzen I'm not sure I understand the second part, if you'd like I can add an explicit example though! Currently we have either normal formatting
> 
> ```rust
> [some, thing, here, ...]
> ```
> 
> and once it reaches a threshold it gets vertical formatting
> 
> ```rust
> [
>     some,
>     thing,
>     here,
>     ...,
> ]
> ```

What will the formatter do if it is presented with the input that I have added in the code block?
What would it do if instead of having the `]` on the same line, as in my example, it would be on a new line?

eureka-cpu:
> > @Braqzen I'm not sure I understand the second part, if you'd like I can add an explicit example though! Currently we have either normal formatting
> > ```rust
> > [some, thing, here, ...]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and once it reaches a threshold it gets vertical formatting
> > ```rust
> > [
> >     some,
> >     thing,
> >     here,
> >     ...,
> > ]
> > ```
> 
> What will the formatter do if it is presented with the input that I have added in the code block? What would it do if instead of having the `]` on the same line, as in my example, it would be on a new line?

So, since this is an AST based formatter, it takes the actual items into account but not the white space. You can think of it like this, your example:
```rust
[ some, args, here, ...,
and, more,
here ]
```
gets collected, which you could represent like:
(keep in mind, I'm not sure how this actually looks, this is a rough representation for the purpose of explaining)
```rust
[Collection(Var("some"), Var("args"), Var("here"), Var("..."), Var("and"), Var("more"), Var("here"))]
```
it does not take into account any whitespace, the formatter strips that by design and we inject whitespace and newlines situationally
So the output would look like:
```rust
[some, args, here, ..., and, more, here]
```

Braqzen:
> > > @Braqzen I'm not sure I understand the second part, if you'd like I can add an explicit example though! Currently we have either normal formatting
> > > ```rust
> > > [some, thing, here, ...]
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > and once it reaches a threshold it gets vertical formatting
> > > ```rust
> > > [
> > >     some,
> > >     thing,
> > >     here,
> > >     ...,
> > > ]
> > > ```
> > 
> > 
> > What will the formatter do if it is presented with the input that I have added in the code block? What would it do if instead of having the `]` on the same line, as in my example, it would be on a new line?
> 
> So, since this is an AST based formatter, it takes the actual items into account but not the white space. You can think of it like this, your example:
> 
> ```rust
> [ some, args, here, ...,
> and, more,
> here ]
> ```
> 
> gets collected, which you could represent like: (keep in mind, I'm not sure how this actually looks, this is a rough representation for the purpose of explaining)
> 
> ```rust
> [Collection(Var("some"), Var("args"), Var("here"), Var("..."), Var("and"), Var("more"), Var("here"))]
> ```
> 
> it does not take into account any whitespace, the formatter strips that by design and we inject whitespace and newlines situationally So the output would look like:
> 
> ```rust
> [some, args, here, ..., and, more, here]
> ```

That sounds really cool. Thanks for letting me know!

Braqzen (approved):

mitchmindtree (approved):

PR #2596, opened 2022-08-19
feat: check for dependency program type in dependency manifest validation

### LINKED ISSUE #1681: forc-pkg panics when building a parent package that depends on a non-library package
I believe depending on something other than a library is not something we want to allow but this should not be panicking but logging an error like "depending on a contract is not permitted".

Steps to reproduce:

1. Run `forc init child`
2. Run `forc init parent`
3. Inside parent's `Forc.toml` add the "child" as a dependency (something like `child = {path = "path/to/child"}`)
4. Run `forc build` in parent's directory.

Which will output something like the following:

```
Compiled library "core".
Compiled library "std".
Compiled contract "child".
thread 'main' panicked at 'no entry found for key', forc-pkg/src/pkg.rs:974:30
```

kayagokalp:
closes #1681.

This has happened couple of times and I think we should be providing a descriptive error. I believe packages should only depend on libraries but please correct me if I am wrong cc @mitchmindtree.

~Leaving this as a draft because I want to tweak something a little.~

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -479,15 +479,12 @@ fn validate_dep(
 }
 /// Part of dependency validation, any checks related to the depenency's manifest content.
 fn validate_dep_manifest(dep: &Pinned, dep_manifest: &ManifestFile) -> Result<()> {
-    // Check that the dependency is a library.
-    match dep_manifest.program_type()? {
-        TreeType::Library { .. } => {}
-        _ => {
-            bail!(
-                "\"{}\" is not a library! Depending on a non-library package is not supported.",
-                dep.name
-            );
-        }
+    // Ensure that the dependency is a library.
+    if !matches!(dep_manifest.program_type()?, TreeType::Library { .. }) {
+        bail!(
+            "\"{}\" is not a library! Depending on a non-library packages is not supported.",
+            dep.name
+        );
     }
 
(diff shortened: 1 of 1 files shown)


forc-pkg/src/pkg.rs
```diff
@@ -479,6 +479,17 @@ fn validate_dep(
...
+    // Check that the dependency is a library.
+    match dep_manifest.program_type()? {
+        TreeType::Library { .. } => {}
+        _ => {
+            bail!(
+                "\"{}\" is not a library! Depending on a non-library package is not supported.",
+                dep.name
+            );
+        }
+    }
```
mitchmindtree, line 491: Could also be written as something like:

```rust
    if !matches!(dep_manifest.program_type()?, TreeType::Library { .. }) {
        bail!(
            "\"{}\" is not a library! Depending on a non-library package is not supported.",
            dep.name
        );
    }
```

mitchmindtree (commented):
Nice one @kayagokalp, thanks for addressing this!

forc-pkg/src/pkg.rs
kayagokalp: TIL `matches!`, I was actually looking for this! thanks a lot!

mitchmindtree (approved):

JoshuaBatty (approved):

PR #2597, opened 2022-08-20
Adds formatting for long conditional statements

eureka-cpu:
As an added bonus, it also formats small ones too!

Ref #2493


eureka-cpu:
I've noticed an issue with this so I've converted to draft

eureka-cpu:
Just an update that the extra functionality I've added here will not work properly until we add better handling for indentation tracking

eureka-cpu:
Found another bug, converting to draft while I figure it out

mitchmindtree (approved):
:shipit:

kayagokalp (commented):
🚀

kayagokalp (approved):

PR #2599, opened 2022-08-20
Evaluate const decls before storage decls

### LINKED ISSUE #2424: Initializing a storage slot with a `const` doesn't seem to work just yet
I know that the const evaluator is still WIP, but this is a simple case that should probably be easy to support:
```rust
contract;

const C1 = 42;

storage {
    x: u64 = C1,
}
```
We currently get the following error:
```console
4 |
5 | storage {
6 |     x: u64 = C1,
  |              ^^ Could not evaluate initializer to a const declaration.
7 | }
  |
```

Also, the error is not great because we're not actually trying to initialize a `const` declaration here. The error should be tailored for storage initializers.

vaivaswatha:
This allows storage decls to use const decls for initialization.

I'm not very happy with this change because it introduces sway-ir stuff earlier on. But looks like that was already happening when we tried to const-evaluate storage initializers. This change only adds evaluating the module const decls as well early on, so that they can be used when evaluating the storage initializers.

Closes #2424


sway-core/src/metadata.rs
```diff
@@ -14,7 +14,7 @@ use std::{collections::HashMap, path::PathBuf, sync::Arc};
 ///! indices repeatedly.
 
 #[derive(Default)]
-pub(crate) struct MetadataManager {
+pub struct MetadataManager {
```
otrho, line 17: Why does this need to leak out of the crate?

test/src/e2e_vm_tests/test_programs/should_pass/test_contracts/basic_storage/json_storage_slots_oracle.json
```diff
@@ -1,42 +1,46 @@
 [
   {
     "key": "02dac99c283f16bc91b74f6942db7f012699a2ad51272b15207b9cc14a70dbae",
-    "value": "6161610000000000000000000000000000000000000000000000000000000000"
+    "value": "6161000000000000000000000000000000000000000000000000000000000000"
```
otrho, line 4: Are these changes all expected?

otrho (commented):
I'm still not really a fan of using IR to evaluate stuff in the compiler, I feel like it's a hack, but it was the easiest solution for getting something quickly.  So this change doesn't really change much, but ideally we have a better compiler specific evaluator one day.

sway-core/src/metadata.rs
vaivaswatha: Without this, I get `can't leak crate-private type` in `compile.rs`.

test/src/e2e_vm_tests/test_programs/should_pass/test_contracts/basic_storage/json_storage_slots_oracle.json
vaivaswatha: > Are these changes all expected?

Yes. Because I added another storage field, the storage indices changed, and hence the keys.

sway-core/src/ir_generation.rs
```diff
@@ -1,4 +1,4 @@
-mod compile;
+pub mod compile;
```
otrho, line 1: So maybe this should be `pub(crate)`?

otrho (approved):

PR #2612, opened 2022-08-23
Reorganize module tree and modules

eureka-cpu:
This PR adds the `Format` trait to `Module` and adds formatting for `Dependency` and `DependencyPath` which we previously neglected. It also re-organizes the modules in the new formatter to be more clear as to what their roles are, and make it easier to navigate on the whole.

I have a few very small PRs I'd like to get in before this one:
- [x] #2611 
- [x] #2610


mitchmindtree (approved):
:+1:

sezna (approved):

PR #2617, opened 2022-08-24
Make `ConcurrentSlab` more general.

emilyaherbert:
PR to make `ConcurrentSlab` more general so that it can be used in other situations.

### DIFF
--- a/sway-core/src/type_system/type_id.rs
+++ b/sway-core/src/type_system/type_id.rs
@@ -6,7 +6,7 @@ use crate::types::*;
 use super::*;
 
 /// A identifier to uniquely refer to our type terms
-#[derive(PartialEq, Eq, Hash, Clone, Copy)]
+#[derive(PartialEq, Eq, Hash, Clone, Copy, Default)]
 pub struct TypeId(usize);
 
 impl std::ops::Deref for TypeId {
--- a/sway-core/src/concurrent_slab.rs
+++ b/sway-core/src/concurrent_slab.rs
@@ -1,28 +1,32 @@
-use crate::type_system::TypeId;
-use std::sync::RwLock;
+use std::{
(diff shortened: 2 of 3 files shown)


sway-core/src/type_system/type_id.rs
```diff
@@ -6,7 +6,7 @@ use crate::types::*;
 use super::*;
 
 /// A identifier to uniquely refer to our type terms
-#[derive(PartialEq, Eq, Hash, Clone, Copy)]
+#[derive(PartialEq, Eq, Hash, Clone, Copy, Default)]
```
sezna, line 9: Won't this default impl just give the type id of 0, which is undefined behavior? Instead, could we remove the constraint on `TypeEngine` and use `::new()` instead of default there?

sezna (changes requested):

sway-core/src/type_system/type_id.rs
emilyaherbert: Good catch!

Side note. I fixed it by doing this:

```rust
#[derive(Debug)]
pub(crate) struct ConcurrentSlab<I, T> {
    indexer: PhantomData<I>,
    inner: RwLock<Vec<T>>,
}

impl<I, T> Default for ConcurrentSlab<I, T> where T: Default {
    fn default() -> Self {
        Self { indexer: PhantomData, inner: Default::default() }
    }
}
```

and one would think that Rust would use `PhantomData` for `#[derive(Default)]` on `PhantomData<T>` but apparently not. It is a known issue: https://github.com/rust-lang/rust/issues/69037.

otrho (approved):

emilyaherbert:
Ping @sezna for rereview plz 😜

sezna (approved):

PR #2619, opened 2022-08-25
Formatter bug fixes

eureka-cpu:
This PR gets in lots of small fixes for the upcoming swap, it takes into account all of the test programs we have in the test directory.


eureka-cpu:
I got most of the kinks out, there are a few things that are still affecting the overall performance but this should be a nice bump closer !

sway-fmt-v2/src/items/item_enum/mod.rs
```diff
@@ -28,12 +28,8 @@ impl Format for ItemEnum {
             write!(formatted_code, "{} ", visibility.span().as_str())?;
         }
         // Add enum token and name
-        write!(
-            formatted_code,
-            "{} {}",
-            self.enum_token.span().as_str(),
-            self.name.as_str()
-        )?;
+        write!(formatted_code, "{} ", self.enum_token.span().as_str(),)?;
```
JoshuaBatty, line 31: shouldn't this be `write!(formatted_code, "{} ", self.enum_token.span().as_str())?;`

sway-fmt-v2/src/items/item_fn/mod.rs
```diff
@@ -130,12 +130,8 @@ fn format_fn_sig(
         write!(formatted_code, "{} ", visibility_token.span().as_str())?;
     }
     // `fn ` + name
-    write!(
-        formatted_code,
-        "{} {}",
-        fn_sig.fn_token.span().as_str(),
-        fn_sig.name.as_str()
-    )?;
+    write!(formatted_code, "{} ", fn_sig.fn_token.span().as_str(),)?;
```
JoshuaBatty, line 133: dangling `,`

sway-fmt-v2/src/items/item_struct/mod.rs
```diff
@@ -28,12 +28,8 @@ impl Format for ItemStruct {
             write!(formatted_code, "{} ", visibility.span().as_str())?;
         }
         // Add struct token and name
-        write!(
-            formatted_code,
-            "{} {}",
-            self.struct_token.span().as_str(),
-            self.name.as_str(),
-        )?;
+        write!(formatted_code, "{} ", self.struct_token.span().as_str(),)?;
```
JoshuaBatty, line 31: and here

sway-fmt-v2/src/utils/language/expr/code_block.rs
```diff
@@ -15,18 +15,26 @@ impl Format for CodeBlockContents {
...
-            statement.format(formatted_code, formatter)?;
-        }
-        if let Some(final_expr) = &self.final_expr_opt {
-            write!(
-                formatted_code,
-                "{}",
-                formatter.shape.indent.to_string(&formatter.config)?
-            )?;
-            final_expr.format(formatted_code, formatter)?;
+        if self.statements.is_empty() && self.final_expr_opt.is_none() {
```
JoshuaBatty, line 18: could be written to just be an if statement without the else

JoshuaBatty (changes requested):
Seems reasonable, only a few nits other wise looks good.

sway-fmt-v2/src/items/item_enum/mod.rs
mitchmindtree: I'm guessing rustfmt struggles a little more with `macro_rules!`, as they can basically have arbitrary syntax! I imagine they'll eventually at least add some special cases for these std macros like `println!`, `write!`, etc.

sway-fmt-v2/src/utils/language/expr/code_block.rs
mitchmindtree: I.e.
```rust
if !self.statements.is_empty() || self.final_expr_opt.is_some() {
```

mitchmindtree (approved):
Looking good!

It looks like some of the E2E tests might have been accidentally reformatted as a part of your testing and those changes made it into this PR. Is it possible to remove these changes for now? Or are they necessary in preparation for the switch?

I'm happy to approve either way as the logic appears to remain the same.

sway-fmt-v2/src/items/item_enum/mod.rs
eureka-cpu: Ah, thanks I missed this

sway-fmt-v2/src/utils/language/expr/code_block.rs
eureka-cpu: Thanks (:

eureka-cpu:
> Looking good!
> 
> It looks like some of the E2E tests might have been accidentally reformatted as a part of your testing and those changes made it into this PR. Is it possible to remove these changes for now? Or are they necessary in preparation for the switch?
> 
> I'm happy to approve either way as the logic appears to remain the same.

So the only ones that changed were ones that couldn't be formatted, so I had to change something about them. Once I saved they auto-formatted with the old formatter which isn't necessary but also will get formatted with the new formatter when we switch over so I didn't think to revert them.

test/src/e2e_vm_tests/test_programs/should_pass/language/generic_type_inference/Forc.toml
```diff
@@ -2,7 +2,7 @@
 authors = ["Fuel Labs <contact@fuel.sh>"]
 entry = "main.sw"
 license = "Apache-2.0"
-name = "generic-type-inference"
+name = "generic_type_inference"
```
eureka-cpu, line 5: I changed the name of this test to be consistent with the rest of the tests, but I didn't change the actual contents. Do you know why it would be failing? @mitchmindtree

eureka-cpu: Ah, must've been the lock file it's passing now!

JoshuaBatty (approved):

sway-fmt-v2/src/items/item_fn/tests.rs
```diff
@@ -110,3 +110,37 @@ fmt_test!(  non_self_fn
...
+"fn returns_msg_sender(expected_id: ContractId) -> bool {
+    let result: Result<Identity, AuthError> = msg_sender();
+    let mut ret = false;
+    if result.is_err() {
+        ret = false;
+    }
+    let unwrapped = result.unwrap();
+    match unwrapped {
+        Identity::ContractId(v) => {
+            ret = true
```
Braqzen, line 124: Since we are not assigning to a variable from the match (instead we assign to `ret` from above) shouldn't the line end with `;`?

sway-fmt-v2/src/items/item_impl/tests.rs
```diff
@@ -0,0 +1,153 @@
...
+);
+
+fmt_test!(  normal_with_generics
+"impl<T> Option<T> {
+    fn some(value: T) -> Self {
+        Option::Some::<T>(value)
+    }
+    fn none() -> Self {
+        Option::None::<T>()
+    }
```
Braqzen, line 127: I think each function should be separated by a new line

sway-fmt-v2/src/utils/language/expr/asm_block.rs
```diff
@@ -1,8 +1,8 @@
 use crate::{
-    formatter::*,
+    formatter::{shape::LineStyle, *},
```
Braqzen, line 2: `*` :triggered:
oof, it's done in other places too

test/src/e2e_vm_tests/test_programs/should_pass/language/array_basics/src/main.sw
```diff
@@ -1,32 +1,42 @@
...
+    foo: u64,
+    bar: u64,
 }
 
 fn main() -> bool {
-    let a: [bool; 5] = [true, true, true, false, true];
-    let b: [u32; 10] = [3; 10];
+    let a: [bool;
+    5] = [true, true, true, false, true];
+    let b: [u32;
```
Braqzen, line 11: What's going on with the arrays here?

sway-fmt-v2/src/items/item_fn/tests.rs
eureka-cpu: I think in this instance the semicolon is optional, otherwise if it wasn't the test wouldn't pass

sway-fmt-v2/src/items/item_impl/tests.rs
eureka-cpu: In this instance since it's purpose is to test the generics themselves, the formatting of the inner functions is irrelevant. I could remove them if that makes it more clear though!

sway-fmt-v2/src/utils/language/expr/asm_block.rs
eureka-cpu: What's wrong here?

test/src/e2e_vm_tests/test_programs/should_pass/language/array_basics/src/main.sw
eureka-cpu: This is the result of the old formatter, meaning the code wasn't formatted according to it previously. This will change when we switch over, and it will look much better (:

Braqzen (approved):

PR #2625, opened 2022-08-25
fix: ForwardSlash span collection

### LINKED ISSUE #2624: `ForwardSlashToken` span is incorrect
Somehow the span of `/` is the span of the source code up until `/`, here's the dbg output of the catch:
```rust
forward_slash_token.span().as_str() = "script;\n        fn hello() {\n\n        let sum = 1 + 2;\n        let quo = 2 /"
```
I caught this while testing out the formatter, and it gets its information from the AST

kayagokalp:
closes #2624.

### DIFF
--- a/sway-parse/src/token.rs
+++ b/sway-parse/src/token.rs
@@ -274,7 +274,7 @@ pub fn lex_commented(
                     } else {
                         Spacing::Alone
                     };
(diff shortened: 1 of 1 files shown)


eureka-cpu:
Tested and it works! 🥇

eureka-cpu (approved):

emilyaherbert (approved):

PR #2627, opened 2022-08-26
Correctly handle diverging expressions in blocks

canndrew:
~This is still a draft. I still need to fix the tests, removing some debugging stuff, separate it into sane commits and then rebase it.~

This PR fixes type-checking and codegen of blocks that contain diverging sub-expressions (such as `break`, `continue` or `result`). This will allow (for example) this code to compile:

```
fn foo() {
    let x: u64 = {
        return;
    };
}
```


otrho (commented):
A couple of things:
- With all those changes to the type checker and IR I feel like there should be more tests added than just one new one.
- I've felt that adding debug printing for the IR is duplicated effort when there's already the pretty printer in `sway-ir/src/printer.rs`.  I see a bunch of macros added and the new `DebugWithContext` but I'm not sure it was used anywhere?  Does `DebugWithContext::fmt()` need to be implemented for everything or does the macro do that?  And then how is it used?  Is there a new `print!()` macro which I missed, or does it somehow work with `std::fmt`?

canndrew:
> I've felt that adding debug printing for the IR is duplicated effort when there's already the pretty printer in `sway-ir/src/printer.rs`. I see a bunch of macros added and the new `DebugWithContext` but I'm not sure it was used anywhere? Does `DebugWithContext::fmt()` need to be implemented for everything or does the macro do that? And then how is it used? Is there a new `print!()` macro which I missed, or does it somehow work with `std::fmt`?

The debugging stuff isn't used anywhere because it was only useful to me when I was debugging my changes.

The derive macro implements `DebugWithContext` for any type you stick it on and there's manual impls for some standard library types in `pretty.rs`. You use it like this:

```rust
// In (for instance) sway-core
fn compile_something(context: &Context) {
    let val = make_some_value();
    println!("val is {:?}", val.with_context(context));
}
```

That is, you can call `.with_context(context)` on any type `T: DebugWithContext`. That returns a `WithContext<'_, '_, T>` which implements `Debug` by calling `T::fmt_with_context` and passing through the context. The derived `DebugWithContext` impls then work like ordinary derived `Debug` impls except that they also pass through the context by calling `.with_context` on their fields. Any reference type `&T` where `T: Debug` also implements `DebugWithContext` by just forwarding to the `Debug` impl and autoref will cause this impl to get used whenever `T: !DebugWithContext`. That means that you don't have to implement `DebugWithContext` for everything - anything that doesn't implement it will just forward to the normal `Debug` impl. The reason `DebugWithContext` is implemented for containers like `Vec<T>` is so that the values in the container will get formatted via `DebugWithContext` rather than just `Debug`.

The derive macro has a helper attribute `#[in_context(field_name)]` which is used like this:

```rust
#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, DebugWithContext)]
pub struct Value(#[in_context(values)] pub generational_arena::Index);
```

This causes the field marked with `#[in_context(field_name)]` (which must have type `Index`) to get formatted by looking up `context.field_name[index]` and formatting that instead. This attribute is pretty much the whole point of `DebugWithContext`, since it allowed me to do printf debugging while being able to automatically/recursively/transparently see through all these `Index` fields. There didn't seem to be a way to do that with the stuff in `printer.rs` and I'm not sure it would have made sense to add that kind of functionality.

sway-core/src/semantic_analysis/ast_node/code_block.rs
```diff
@@ -47,33 +47,40 @@ impl TypedCodeBlock {
...
-                        ref return_type,
-                        ..
-                    }),
-                ..
-            } if !x.deterministically_aborts() => Some(*return_type),
-            _ => None,
-        });
+        let block_type =
+            {
+                // find the implicit return, if any, and use it as the code block's return type.
```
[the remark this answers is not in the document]
canndrew: Because I write code the way I think looks sensible then I run rustfmt before pushing in order to "fix" it and sometimes it outputs stuff like that. I've fixed it now.

otrho:
OK, for new data structures we can derive `DebugWithContext` along with `Debug` and it'll just work.  Fair enough, that's useful without being intrusive or redundant.  :grinning:

canndrew:
> With all those changes to the type checker and IR I feel like there should be more tests added than just one new one.

I've added another test which tests that the code type-checks, compiles and runs properly with a whole bunch of different diverging expressions.

sway-core/src/type_system/type_info.rs
```diff
@@ -609,6 +609,15 @@ impl TypeInfo {
         }
     }
 
+    pub fn can_safely_ignore(&self) -> bool {
+        match self {
+            TypeInfo::Tuple(fields) => fields.is_empty(),
```
emilyaherbert, line 614: What about an `TypeInfo::Array` of `TypeInfo::Tuple`, could that be ignored? (does Sway even support that? 😅 )

sway-ir/src/pretty.rs
```diff
@@ -0,0 +1,109 @@
+pub use sway_ir_macros::*;
+use {crate::Context, std::fmt};
+
+pub struct WithContext<'a, 'c, T: ?Sized> {
```
emilyaherbert, line 4: So sick 🙏

sway-ir/sway-ir-macros/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "sway-ir-macros"
```
emilyaherbert, line 2: What is the reasoning for why this needs to be in a separate crate?

emilyaherbert (changes requested):

sway-core/src/type_system/type_info.rs
otrho: An empty tuple is actually Unit, which is why it's here with the others.  I don't think ignoring non-empty data structures is desirable?

sway-ir/sway-ir-macros/Cargo.toml
canndrew: Rust needs to build procedural macros before they can be run. This (currently) requires that procedural macros are defined in a separate crate to the crate which uses them so that the macro crate can be built first.

sway-core/src/type_system/type_info.rs
canndrew: I've updated the method in [this commit](https://github.com/FuelLabs/sway/pull/2627/commits/8463fcc6c4e16a2e9795a4ef7c8d4710859dd197) to make the check more complete.

sway-ir/sway-ir-macros/Cargo.toml
emilyaherbert: Gotcha.

emilyaherbert (dismissed):

mohammadfawaz (approved):

otrho (approved):

vaivaswatha:
> > I've felt that adding debug printing for the IR is duplicated effort when there's already the pretty printer in `sway-ir/src/printer.rs`. I see a bunch of macros added and the new `DebugWithContext` but I'm not sure it was used anywhere? Does `DebugWithContext::fmt()` need to be implemented for everything or does the macro do that? And then how is it used? Is there a new `print!()` macro which I missed, or does it somehow work with `std::fmt`?
> 
> The debugging stuff isn't used anywhere because it was only useful to me when I was debugging my changes.
> 
> The derive macro implements `DebugWithContext` for any type you stick it on and there's manual impls for some standard library types in `pretty.rs`. You use it like this:
> 
> ```rust
> // In (for instance) sway-core
> fn compile_something(context: &Context) {
>     let val = make_some_value();
>     println!("val is {:?}", val.with_context(context));
> }
> ```
> 
> That is, you can call `.with_context(context)` on any type `T: DebugWithContext`. That returns a `WithContext<'_, '_, T>` which implements `Debug` by calling `T::fmt_with_context` and passing through the context. The derived `DebugWithContext` impls then work like ordinary derived `Debug` impls except that they also pass through the context by calling `.with_context` on their fields. Any reference type `&T` where `T: Debug` also implements `DebugWithContext` by just forwarding to the `Debug` impl and autoref will cause this impl to get used whenever `T: !DebugWithContext`. That means that you don't have to implement `DebugWithContext` for everything - anything that doesn't implement it will just forward to the normal `Debug` impl. The reason `DebugWithContext` is implemented for containers like `Vec<T>` is so that the values in the container will get formatted via `DebugWithContext` rather than just `Debug`.
> 
> The derive macro has a helper attribute `#[in_context(field_name)]` which is used like this:
> 
> ```rust
> #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, DebugWithContext)]
> pub struct Value(#[in_context(values)] pub generational_arena::Index);
> ```
> 
> This causes the field marked with `#[in_context(field_name)]` (which must have type `Index`) to get formatted by looking up `context.field_name[index]` and formatting that instead. This attribute is pretty much the whole point of `DebugWithContext`, since it allowed me to do printf debugging while being able to automatically/recursively/transparently see through all these `Index` fields. There didn't seem to be a way to do that with the stuff in `printer.rs` and I'm not sure it would have made sense to add that kind of functionality.

I just came across this and haven't tried it yet. I think it's going to be very useful.

Issue #2622, opened 2022-08-25
Migration to `fuel-core` v0.10+ is blocked


nfurfaro:
While trying to make the switch to`fuel-core` v0.10` I'm encountering a number of issues.

### Context
I'm working on this PR here: https://github.com/FuelLabs/sway/pull/2482
The changeset is fairly large; multiple dependencies are being bumped, opcodes have been removed and added, and a big chunk of the stdlib has been refactored. These have to happen together in 1 PR to make it all work.

### Current Issues

- Multiple failing stdlib tests using the SDK

 The current released SDK version doesn't currently support `fuel-core` v0.10` so I'm using a patched version of the SDK for the stdlib tests with this patch:
```
[patch.crates-io]
fuels = { git = "https://github.com/FuelLabs/fuels-rs", branch = "hal3e/bump-fuel-core"}
```
Among other things, the failing tests indicate that the number of inputs returned by the `input_amount()` getter == `0`.
More discussion here: https://fuellabs.slack.com/archives/C02CX4JVCSJ/p1661344259170089

- Apparent bug in the `__gtf` intrinsic
If I log out the value returned by the intrinsic for a specific field, and then get that same field using raw `asm` + the `gtf` opcode I get different values.  For example:
```
    let type_1 = __gtf::<u8>(0, GTF_TYPE);
    let type_2 = asm(r1, r2: 0) {
        gtf r1 r2 i1;
        r1: u8
    };
    log(type_1); // 10376
    log(type_2); // 0 (the expected value
```
I've not tested extensively so I don't know if this is the case for more than 1 field.

- While trying to produce a minimal repro of the above (WIP can be found on the `furnic/gtf-intrinsic-bug` branch), I started hitting this new error:
```
Error: Response errors; TransactionCreateBytecodeWitnessIndex
```
More discussion here: https://fuellabs.slack.com/archives/C01KBNGSWGJ/p1661440389568419

- and finally, some strange behavior around logging values in sway apparently affecting control flow?
More discussion here: https://fuellabs.slack.com/archives/C01KBNGSWGJ/p1661430694033159

mohammadfawaz:
@nfurfaro now that we're done with the migration to `fuel-core 0.10.1`, most of the items above are resolved. Do you mind creating separate descriptive issues of the remaining problems?

nfurfaro:
All sub-issue in this have been resolved, closing.

PR #2629, opened 2022-08-26
feat: sign forc-deploy transactions

### LINKED ISSUE #2623: Refactor `forc-deploy` so that it returns an transaction ID and asks for a signature
As we are moving forward with #2413, we need `forc-deploy` to report the transaction id (deployment transaction) back to the user and then wait for the signature. A signature will be generated by the `forc-wallet` as we sign the transaction with that. After the signature is generated we need to give it back to `forc-deploy`. 

Also, it is important that (maybe using a flag) we should be able to keep the current behavior as testing suite depends on that

## Blocked by:

- [x] https://github.com/FuelLabs/fuels-rs/pull/540 (need a release which includes this)
- [x] https://github.com/FuelLabs/forc-wallet/pull/24

kayagokalp:
## About this PR
closes #2623.

~Blocked until https://github.com/FuelLabs/fuels-rs/pull/540 is merged and released. While testing I pointed to a local instance of the SDK,~ most of the diff is actually coming from the lock file.

### TODO

- [x] Accept transaction parameters 
- [x] Point to released fuels-rs (Blocked until the next release of fuels-rs)


mitchmindtree:
Just adding a note (previously discussed on slack with @kayagokalp):

I think it might be best if we first wait to update `fuel-core`, `fuels-rs` and their dependencies that we use throughout the Sway repo before this lands. We want to avoid the massive lock file additions this PR would otherwise cause, which would require anyone building the repo to build both entire version sets and likely slow down CI quite a bit.

https://github.com/FuelLabs/sway/issues/2622 tracks the `fuel-core` update blockers in particular.

mitchmindtree:
Ahh it looks like https://github.com/FuelLabs/sway/pull/2482 is doing the heavy-work of updating to the latest `fuel-core` and `fuels-rs`.

forc-plugins/forc-client/src/ops/deploy/cmd.rs
```diff
@@ -69,4 +69,13 @@ pub struct DeployCommand {
     /// Output the time elapsed over each part of the compilation process.
     #[clap(long)]
     pub time_phases: bool,
+    /// Do not sign the transaction
+    #[clap(long)]
+    pub not_sign: bool,
```
mitchmindtree, line 74: Was just taking a quick look at #2656 and noticed this flag - perhaps a better name might be `unsigned`?

mitchmindtree (dismissed):
:rocket: 

Looking great! Just a couple conflicts to resolve.

sezna (dismissed):

sezna (approved):

mitchmindtree (approved):

Issue #2636, opened 2022-08-26
Introduce a monomorphization cache


emilyaherbert:
Instead of creating one monomorphized copy per expression, we should create a monomorphization cache to keep track of monomorphized copies that we have created in the past, so that we can eliminate duplicate computation

emilyaherbert:
See this closed PR description for additional context and insight into a possible design: https://github.com/FuelLabs/sway/issues/2723#issue-1361870315

emilyaherbert:
Closing this issue based on chatting with @IGI-111 as it will be included in the implementation for #1267.

Issue #2635, opened 2022-08-26
Add functions to the declaration engine


emilyaherbert:
This can be modeled after: https://github.com/emilyaherbert/declaration-engine-and-collection-context-demo/blob/3a7b3d9e0096956a60cb0e36af93bcf7f2ecbd48/de/src/semantic_analysis/inference/declaration.rs#L51

PR #2637, opened 2022-08-29
Add a monomorphization cache to the declaration engine.

tritao:
This introduces a cache to the declaration engine that keeps track of functions/structs that have been instantiated from a specific set of type parameters.

The approach here is relatively simple and relies on a linear search. I originally thought about having more maps to speed it up by keying on the type arguments, but since it adds a bit of complexity I opted for this for now. If it proves to be a bottleneck in the future we can always speed it up.

This was prototyped in https://github.com/tritao/declaration-engine-and-collection-context-demo/commit/c3aa5e718211ee1a88d7ebc51798280e8321e387, where these APIs where wired to the function application and struct expressions instantiation code.

However given the declaration engine is not hooked up yet here, this PR is just adding the APIs, and more proper testing will be added as we wire things up.

Closes https://github.com/FuelLabs/sway/issues/2636.

### DIFF
--- a/sway-core/src/declaration_engine/declaration_engine.rs
+++ b/sway-core/src/declaration_engine/declaration_engine.rs
@@ -4,11 +4,15 @@ use sway_types::{Span, Spanned};
 
 use crate::{
     concurrent_slab::ConcurrentSlab,
+    error::{err, ok},
+    namespace::{Path, Root},
     semantic_analysis::{
         TypedImplTrait, TypedStorageDeclaration, TypedStructDeclaration, TypedTraitDeclaration,
         TypedTraitFn,
     },
-    CompileError, TypedFunctionDeclaration,
+    type_system::TypeArgument,
+    type_system::{type_engine::monomorphize, EnforceTypeArguments},
+    CompileError, CompileResult, MonomorphizeHelper, TypedFunctionDeclaration,
 };
 
 use super::{declaration_id::DeclarationId, declaration_wrapper::DeclarationWrapper};
@@ -24,6 +28,7 @@ pub(crate) struct DeclarationEngine {
     // *declaration_id -> vec of monomorphized copies
     // where the declaration_id is the original declaration
     monomorphized_copies: RwLock<HashMap<usize, Vec<DeclarationId>>>,
+    is_monomorph_cache_enabled: bool,
 }
 
 impl DeclarationEngine {
@@ -39,14 +44,10 @@ impl DeclarationEngine {
 
     fn de_add_monomorphized_copy(&self, original_id: DeclarationId, new_id: DeclarationId) {
         let mut monomorphized_copies = self.monomorphized_copies.write().unwrap();
-        match monomorphized_copies.get_mut(&*original_id) {
-            Some(prev) => {
-                prev.push(new_id);
-            }
-            None => {
-                monomorphized_copies.insert(*original_id, vec![new_id]);
-            }
-        }
+        monomorphized_copies
+            .entry(*original_id)
+            .and_modify(|f| f.push(new_id.clone()))
+            .or_insert_with(|| vec![new_id.clone()]);
     }
 
     fn de_get_monomorphized_copies(&self, original_id: DeclarationId) -> Vec<DeclarationWrapper> {
@@ -74,30 +75,6 @@ impl DeclarationEngine {
         span: &Span,
     ) -> Result<TypedFunctionDeclaration, CompileError> {
         self.slab.get(*index).expect_function(span)
-    }
-
-    fn de_add_monomorphized_function_copy(
-        &self,
-        original_id: DeclarationId,
-        new_copy: TypedFunctionDeclaration,
-    ) {
-        let span = new_copy.span();
-        let new_id = DeclarationId::new(
-            self.slab.insert(DeclarationWrapper::Function(new_copy)),
-            span,
-        );
-        self.de_add_monomorphized_copy(original_id, new_id)
-    }
-
-    fn de_get_monomorphized_function_copies(
-        &self,
-        original_id: DeclarationId,
-        span: &Span,
-    ) -> Result<Vec<TypedFunctionDeclaration>, CompileError> {
-        self.de_get_monomorphized_copies(original_id)
-            .into_iter()
-            .map(|x| x.expect_function(span))
-            .collect::<Result<_, _>>()
     }
 
     fn de_insert_trait(&self, r#trait: TypedTraitDeclaration) -> DeclarationId {
@@ -129,7 +106,7 @@ impl DeclarationEngine {
         self.slab.get(*index).expect_trait_fn(span)
     }
 
-    fn insert_trait_impl(&self, trait_impl: TypedImplTrait) -> DeclarationId {
+    fn de_insert_trait_impl(&self, trait_impl: TypedImplTrait) -> DeclarationId {
         let span = trait_impl.span.clone();
         DeclarationId::new(
             self.slab.insert(DeclarationWrapper::TraitImpl(trait_impl)),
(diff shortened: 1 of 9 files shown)


sway-core/src/declaration_engine/declaration_engine.rs
```diff
@@ -64,13 +61,64 @@ impl DeclarationEngine {
...
         original_id: DeclarationId,
-        new_copy: TypedFunctionDeclaration,
-    ) {
-        let new_id = self.slab.insert(DeclarationWrapper::Function(new_copy));
-        self.add_monomorphized_copy(original_id, new_id)
+        type_arguments: &Vec<TypeArgument>,
+    ) -> Option<TypedFunctionDeclaration> {
+        for copy in self.get_monomorphized_copies(original_id).iter() {
+            let monomorphized_fn_decl = match copy {
+                DeclarationWrapper::Function(fn_decl) => fn_decl,
```
vaivaswatha, line 71: There is an `expect_function` method that can be used here I think.

sway-core/src/declaration_engine/declaration_engine.rs
```diff
@@ -118,13 +166,64 @@ impl DeclarationEngine {
...
         original_id: DeclarationId,
-        new_copy: TypedStructDeclaration,
-    ) {
-        let new_id = self.slab.insert(DeclarationWrapper::Struct(new_copy));
-        self.add_monomorphized_copy(original_id, new_id)
+        type_arguments: &Vec<TypeArgument>,
+    ) -> Option<TypedStructDeclaration> {
+        for copy in self.get_monomorphized_copies(original_id).iter() {
+            let monomorphized_struct_decl = match copy {
+                DeclarationWrapper::Struct(struct_decl) => struct_decl,
```
vaivaswatha, line 176: Same as before, `expect_struct`.

vaivaswatha (approved):

vaivaswatha (dismissed):

emilyaherbert:
> so I've disabled the cache by default and we can enable it by default at a later stage

Good plan :+1:

emilyaherbert (approved):

sway-core/src/declaration_engine/declaration_id.rs
```diff
@@ -1,7 +1,7 @@
 use std::fmt;
 
 /// An ID used to refer to an item in the [DeclarationEngine]
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
```
emilyaherbert, line 4: Can you remove `Hash, PartialEq, Eq`? Those will need to be special cased I think.

emilyaherbert (changes requested):

sway-core/src/declaration_engine/declaration_id.rs
emilyaherbert: @tritao you should be able to pull in  #2675 and solve this now :smile:

sway-core/src/declaration_engine/declaration_engine.rs
```diff
@@ -76,17 +77,69 @@ impl DeclarationEngine {
...
-            self.slab.insert(DeclarationWrapper::Function(new_copy)),
-            span,
+        type_arguments: &Vec<TypeArgument>,
+    ) -> Option<TypedFunctionDeclaration> {
+        for copy in self
+            .de_get_monomorphized_copies(original_id)
+            .iter()
+            .cloned()
+        {
+            let monomorphized_fn_decl = copy.expect_function(&Span::dummy()).unwrap();
```
emilyaherbert, line 90: Modify the `de_get_monomorphized_function` function to return either a `Result<(), CompileError>` or a `CompileResult<()>` and remove the `.unwrap()`

sway-core/src/declaration_engine/declaration_engine.rs
```diff
@@ -76,17 +77,69 @@ impl DeclarationEngine {
...
+            if &monomorphized_fn_decl.type_parameters == type_arguments {
```
emilyaherbert, line 91: I think that we may need to think more deeply about the logic of this line. In general, most generic functions are called without type arguments, so most of the time this check will fail, missing valid cache opportunities.

sway-core/src/declaration_engine/declaration_engine.rs
```diff
@@ -158,26 +211,69 @@ impl DeclarationEngine {
...
+        type_arguments: &Vec<TypeArgument>,
+        span: &Span,
+    ) -> Option<TypedStructDeclaration> {
+        for copy in self
+            .de_get_monomorphized_copies(original_id)
+            .iter()
+            .cloned()
+        {
+            let monomorphized_struct_decl = copy.expect_struct(span).unwrap();
+            if &monomorphized_struct_decl.type_parameters == type_arguments {
```
emilyaherbert, line 226: Same as above---most generic structs are called without type arguments, so most of the time this check will fail, missing valid cache opportunities.

sway-core/src/declaration_engine/declaration_engine.rs
```diff
@@ -158,26 +211,69 @@ impl DeclarationEngine {
         self.slab.get(*index).expect_struct(span)
     }
 
-    fn de_add_monomorphized_struct_copy(
+    pub(crate) fn de_get_monomorphized_struct(
```
emilyaherbert, line 214: I recommend making the `de_get_monomorphized_struct` and `de_get_monomorphized_function`, etc, functions all one function `de_get_monomorphized_version` that takes `original_id: DeclarationId,` and returns `Option<T> where T: MonomorphizeHelper`

sway-core/src/declaration_engine/declaration_engine.rs
```diff
@@ -76,17 +77,69 @@ impl DeclarationEngine {
...
+                self.de_get_monomorphized_function(decl_id.clone(), type_arguments);
+            if let Some(cached_fn_decl) = cached_fn_decl {
+                return ok(cached_fn_decl, warnings, errors);
+            }
+        }
+
+        // monomorphize the function declaration into a new copy
+        let mut typed_function_declaration = self
+            .de_get_function(decl_id.clone(), call_site_span)
+            .unwrap();
```
emilyaherbert, line 121: Remove this `.unwrap()`

sway-core/src/declaration_engine/declaration_engine.rs
```diff
@@ -158,26 +211,69 @@ impl DeclarationEngine {
...
-        self.de_add_monomorphized_copy(original_id, new_id)
+        type_arguments: &Vec<TypeArgument>,
+        span: &Span,
+    ) -> Option<TypedStructDeclaration> {
+        for copy in self
+            .de_get_monomorphized_copies(original_id)
+            .iter()
+            .cloned()
+        {
+            let monomorphized_struct_decl = copy.expect_struct(span).unwrap();
```
emilyaherbert, line 225: Modify this function to return either a `Result<(), CompileError>` or a `CompileResult<()>` and remove the `.unwrap()`

sway-core/src/declaration_engine/declaration_engine.rs
```diff
@@ -158,26 +211,69 @@ impl DeclarationEngine {
...
+            let cached_struct_decl =
+                self.de_get_monomorphized_struct(decl_id.clone(), type_arguments, call_site_span);
+            if let Some(cached_struct_decl) = cached_struct_decl {
+                return ok(cached_struct_decl, warnings, errors);
+            }
+        }
+
+        // monomorphize the struct declaration into a new copy
+        let mut typed_struct_declaration =
+            self.de_get_struct(decl_id.clone(), call_site_span).unwrap();
```
emilyaherbert, line 255: Remove the `.unwrap()`

emilyaherbert (changes requested):
We need to think a little bit longer about how to achieve optimal results with the cache I think, could you iterate on the design another time?

sway-core/src/type_system/mod.rs
```diff
@@ -7,7 +7,7 @@ mod resolved_type;
 mod trait_constraint;
 mod type_argument;
 mod type_binding;
-mod type_engine;
+pub(crate) mod type_engine;
```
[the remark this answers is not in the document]
emilyaherbert: Instead of changing the visibility of the module, change the import statement to match this example: https://github.com/FuelLabs/sway/blob/9076eb232cb24d5ed9136d7e8d7a77d4f7460467/sway-core/src/semantic_analysis/type_check_context.rs#L6

sway-core/src/declaration_engine/declaration_engine.rs
```diff
@@ -158,26 +135,78 @@ impl DeclarationEngine {
         self.slab.get(*index).expect_struct(span)
     }
 
-    fn de_add_monomorphized_struct_copy(
+    pub(crate) fn de_get_monomorphized_decl(
```
emilyaherbert, line 138: Remove the `pub(crate)` because all of these methods are instead accessed via the global methods that access `lazy_static`.

sway-core/src/declaration_engine/declaration_engine.rs
```diff
@@ -158,26 +135,78 @@ impl DeclarationEngine {
...
-        self.de_add_monomorphized_copy(original_id, new_id)
+        type_arguments: &Vec<TypeArgument>,
+        span: &Span,
+    ) -> Result<DeclarationWrapper, CompileError> {
+        for monomorphized_decl in self
+            .de_get_monomorphized_copies(original_id)
+            .iter()
+            .cloned()
+        {
+            if monomorphized_decl.type_parameters() == type_arguments {
```
emilyaherbert, line 149: This line will cause the cache to fail in some cases (e.g. nested generic function calls).

For example this:

```rust
fn foo<T, F>(a: T, b: F) -> F {
    b
}

fn bar<T, F>(a: T, b: F) -> F {
    foo::<T, F>(a, b)
}
```

would generate a false cache hit due to the implementation of `PartialEq` on `TypeInfo::UnknownGeneric`: https://github.com/FuelLabs/sway/blob/a989e6ca0caf6ace4630cd72a2ff8c875c141a43/sway-core/src/type_system/type_info.rs#L205

sway-core/src/declaration_engine/declaration_engine.rs
```diff
@@ -158,26 +135,78 @@ impl DeclarationEngine {
...
+            }
+        }
+        Err(CompileError::Internal(
+            "could not find monomorphized decl",
+            span.clone(),
+        ))
     }
 
-    fn de_get_monomorphized_struct_copies(
+    pub(crate) fn de_get_or_create_monomorphized_decl<T>(
```
emilyaherbert, line 159: Remove the `pub(crate)`

sway-core/src/declaration_engine/declaration_wrapper.rs
```diff
@@ -17,6 +17,7 @@ use crate::{
 pub(crate) enum DeclarationWrapper {
     // no-op variant to fulfill the default trait
     Unknown,
+    Enum(TypedEnumDeclaration),
```
emilyaherbert, line 20: There is already a PR for adding enum support: https://github.com/FuelLabs/sway/pull/2713. Please remove these additions on this PR.

sway-core/src/declaration_engine/declaration_wrapper.rs
```diff
@@ -69,11 +71,47 @@ impl CopyTypes for DeclarationWrapper {
...
 
+impl MonomorphizeHelper for DeclarationWrapper {
+    fn type_parameters(&self) -> &[TypeParameter] {
+        match self {
+            DeclarationWrapper::Trait(_)
+            | DeclarationWrapper::TraitFn(_)
+            | DeclarationWrapper::TraitImpl(_)
+            | DeclarationWrapper::Storage(_)
+            | DeclarationWrapper::Unknown => {
+                panic!("declaration type does not support type parameters")
```
emilyaherbert, line 82: In general, we should not use `panic!` (`unwrap!`, etc), unless absolutely necessary. The compiler is designed so that if there is an error when type checking one AST nodes, the other AST nodes are relatively unaffected and can continue to be type checked, providing the best end-user experience. `panic!` generates an early exit, preventing the compiler from offering this feature to users. This is why we have the heafty `CompileResult` struct. In this particular case, remove this `panic` and I recommend that you return an empty vec instead.

sway-core/src/declaration_engine/declaration_wrapper.rs
```diff
@@ -69,11 +71,47 @@ impl CopyTypes for DeclarationWrapper {
...
+            DeclarationWrapper::Function(decl) => decl.name(),
+            DeclarationWrapper::Trait(decl) => &decl.name,
+            DeclarationWrapper::TraitFn(decl) => &decl.name,
+            DeclarationWrapper::TraitImpl(decl) => &decl.trait_name.suffix,
+            DeclarationWrapper::Struct(decl) => decl.name(),
+        }
+    }
+
+    fn to_wrapper(&self) -> DeclarationWrapper {
+        panic!("not expected to be called for DeclarationWrapper")
```
emilyaherbert, line 105: Remove the `panic!`

emilyaherbert (changes requested):
Adding the monomorphization cache is not a blocker for adding the declaration engine, and the cache-aware API is similar enough to the non-cache-aware API that switching API functions would only be a small change. Considering we are not able to test the code in the cache until the declaration engine goes in, I recommend that we merge #2631 and  #2635 in before we add the cache. Would you close this PR? I recommend keeping this branch alive though so that we can bring it back when we are ready to add the cache again.

PR #2645, opened 2022-08-29
Fix in how the `__gtf` intrinsic is lowered to asm

mohammadfawaz:
Fixes one problem in https://github.com/FuelLabs/sway/pull/2642


otrho (approved):

nfurfaro (approved):
looks good!

PR #2646, opened 2022-08-29
Introduce the `CompileWrapper` type so that we can feign `PartialEq` and `Hash` while still using local copies of the `DeclarationEngine` and `TypeEngine`

emilyaherbert:
Unfortunately, doing this without making the `DeclarationEngine` a `lazy_static` is a HUGE pain


emilyaherbert:
Closing in favor of #2675

PR #2651, opened 2022-08-30
Bump to `v0.21.0`

mohammadfawaz:
Also ran `cargo update` as usual since this is a minor release.


otrho (approved):

mitchmindtree (approved):
Did we want to wait for the `fuel-core` and `fuels-rs` version update?

I guess we can always just cut another if need be!

PR #2653, opened 2022-08-30
fix: `swayfmt-v2` deletes empty tuples

### LINKED ISSUE #2652: `swayfmt-v2` deletes empty tuples
`sway-fmt-v2` removes empty tuples but instead it should be placing `()`.

### LINKED ISSUE #2654: Newline map is placing new lines incorrectly
Another example of a conflict is in https://github.com/FuelLabs/sway/blob/master/test/src/e2e_vm_tests/test_programs/should_pass/language/generic_impl_self/src/main.sw when it is formatted with the corrent v2, newlines are pushed in the wrong places

kayagokalp:
closes #2652.
closes #2654.


eureka-cpu (approved):
LGTM

sway-fmt-v2/src/utils/language/expr/collections.rs
```diff
@@ -16,7 +16,10 @@ impl Format for ExprTupleDescriptor {
         formatter: &mut Formatter,
     ) -> Result<(), FormatterError> {
         match self {
-            Self::Nil => {}
+            Self::Nil => {
+                Self::open_parenthesis(formatted_code, formatter)?;
```
eureka-cpu, line 20: Just realized we can actually open and close the parenthesis before and after the match statement instead of calling the functions twice, one per match arm

mohammadfawaz (approved):

mitchmindtree (approved):

PR #2660, opened 2022-08-30
Replace old JSON ABI format with the new one

mohammadfawaz:
The json file emitted now has `-flat` in its name, until the SDK switches to only supporting the new format.


emilyaherbert (commented):
> The json file emitted now has -flat in its name, until the SDK switches to only supporting the new format.

I'm not sure I know what you mean by this---the diff for this PR doesn't show any files with -flat in the name in the "changed files" list. Is that intended?

mohammadfawaz:
> 

Right so the JSON files emitted in the test suite are named differently than the ones generated by default when you run `forc build` for an independent project. When running `forc build`, the JSON file we get in `out/debug` is now called `<project_name>-flat-abi.json` which is the same as before except for the `-flat`. This is needed temporarily because the SDK relies that added `-flat` to unlock the new behavior. This will quickly change when the SDK team switches to using the new format for all JSON files (not just the ones that have `-flat`).

emilyaherbert:
> Right so the JSON files emitted in the test suite are named differently than the ones generated by default when you run forc build for an independent project

Makes sense, thanks for the clarification!

emilyaherbert (approved):

PR #2663, opened 2022-08-30
Fixes consecutive `fn_sig` formatting

### LINKED ISSUE #2662: Consecutive `fn_sig` will format incorrectly
Found while testing on Sway applications, consecutive fn signatures will take into account the width of the previous function. In order to stop this we replace the add and sub width methods with an update method instead.

eureka-cpu:
Found while testing on Sway applications, consecutive fn signatures will take into account the width of the previous function. In order to stop this we replace the add and sub width methods with an update method instead.

Closes #2662


kayagokalp (approved):
utACK

PR #2665, opened 2022-08-30
fix: sway-fmt-v2 newline/comment handler makes duplicate searches

### LINKED ISSUE #2664: sway-fmt-v2 newline/comment handler makes duplicate searches
Comment/newline handler searches in between point a and b and then point b and a. This is unnecessary and causes weird problems regarding doc comment formatting.

kayagokalp:
closes #2664.

This is blocking #2668


eureka-cpu (approved):
utACK

mitchmindtree (approved):

PR #2669, opened 2022-08-31
Replace old formatter with new formatter

### LINKED ISSUE #1516: [Tracking] `swayfmt-v2` MVP

eureka-cpu:
Closes #1516 

- [x] Rename and replace old fmt with new fmt
- [x] Fix any conflicts that arise from this
- [ ] #2691


mitchmindtree:
This is legendary @eureka-cpu! :zap: :rocket: :100:

examples/hashing/src/main.sw
```diff
@@ -73,36 +92,57 @@ fn main() {
     log(sha_hashed_enum);
     log(sha_hashed_struct);
 
+
```
mohammadfawaz, line 95: What is the reason for the extract newlines here and in other places?

mohammadfawaz: and does this mean that when I rerun the formatter another new line will be added?

mitchmindtree: @mohammadfawaz I just noticed this issue https://github.com/FuelLabs/sway/issues/2691 - @eureka-cpu possibly related?

emilyaherbert:
Omg it's happening

examples/hashing/src/main.sw
eureka-cpu: @mohammadfawaz & @mitchmindtree yes, there's an issue currently where newlines will continue to get added if there is a comment with a newline in its span. Doesn't seem to happen for comments that don't have an extra newline or in places without comments.

sway-lsp/src/sway_config.rs
```diff
@@ -34,11 +42,24 @@ impl SwayConfig {
 // note `FormattingOptions` and `SwayConfig` may be similar at this moment,
 // but they are not the same thing, `SwayConfig` contains all the data related to the LanguageServer
 // while `FormattingOptions` is only the part that is necessary for 'formating'
-impl From<SwayConfig> for FormattingOptions {
+impl From<SwayConfig> for Formatter {
```
mitchmindtree, line 45: Perhaps a closer mapping to the old behaviour would be:
```suggestion
impl From<SwayConfig> for Config {
```
where `Config` is the formatter's `Config` type?

Eventually we might want to simply remove `SwayConfig` in favour of using the formatter's `Config` directly, but we can discuss this with @JoshuaBatty once he's back in the office. For now, I think you're on the right track with simply using the defaults where necessary :+1:

mitchmindtree:
@eureka-cpu I've just merged in #2698 (which should have @kayagokalp's fixes for the newline issues) into your branch here!

I was expecting this to resolve these `sway-fmt-v2` conflicts as there were no conflicts that arose locally when merging in `master`, but it seems they remain on this PR :thinking: Maybe GitHub has some approach to merging that differs from a simple local `git merge`?

If you have trouble resolving these conflicts within this PR, feel free to open up a new one as it's more important we land this sooner than later :)

eureka-cpu:
> @eureka-cpu I've just merged in #2698 (which should have @kayagokalp's fixes for the newline issues) into your branch here!
> 
> I was expecting this to resolve these `sway-fmt-v2` conflicts as there were no conflicts that arose locally when merging in `master`, but it seems they remain on this PR thinking Maybe GitHub has some approach to merging that differs from a simple local `git merge`?
> 
> If you have trouble resolving these conflicts within this PR, feel free to open up a new one as it's more important we land this sooner than later :)

When I pull the branch it doesn't show any conflicts 🤔

eureka-cpu:
Merging master did the trick oddly enough

sway-lsp/src/sway_config.rs
eureka-cpu: Checked in on this and while converting to `Config` may be closer to the previous code, it doesn't seem that we're using it anywhere else other than for the end result of constructing a `Formatter` and calling `.format()`. In other words we would be constructing a `Config` and then construct a `Formatter` and then replace the formatter's config with said config.

examples/hashing/src/main.sw
```diff
@@ -35,30 +34,50 @@ fn main() {
     let sha_hashed_u32 = sha256(~u32::max());
     let sha_hashed_u64 = sha256(~u64::max());
 
+
```
Braqzen, line 37: Additional new lines? This happens in multiple files

examples/match_statements/src/main.sw
```diff
@@ -1,12 +1,11 @@
...
+fn foo() {}
     // do something
-}
-fn bar() {
-    // do something
-}
+fn bar() {}
 
+
+    // do something
```
Braqzen, line 8: Comments are broken here

examples/structs/src/main.sw
```diff
@@ -40,73 +45,52 @@ fn shorthand_instantiation() -> Foo {
...
-    let Point {
-        x, y
-    }
-    = point1;
-
-    let point2 = Point {
-        x: 1,
-        y: 1,
-    };
+    let Point{x, y} = point1;
```
Braqzen, line 64: This should be `let Point {x, y}` or imo `let Point { x, y }` is better

sway-lsp/src/server.rs
```diff
@@ -79,7 +79,7 @@ fn capabilities() -> ServerCapabilities {
             trigger_characters: None,
             ..Default::default()
         }),
-        document_formatting_provider: Some(OneOf::Left(true)),
+        // document_formatting_provider: Some(OneOf::Left(true)),
```
Braqzen, line 82: Commented out line?

examples/hashing/src/main.sw
eureka-cpu: Ah this is from the previous formatting I did, I'll go in and fix that rq

examples/match_statements/src/main.sw
eureka-cpu: I'll open an issue to address it, this happens when placing comments inside an empty block

examples/hashing/src/main.sw
eureka-cpu: As it turns out this is actually a side effect of the previous bug fix for preceding newlines on comments. I'll reopen the issue.

eureka-cpu: #2691

eureka-cpu: I added a bandaid fix for now, and will continue looking into it further

Braqzen (approved):

mohammadfawaz (approved):

PR #2677, opened 2022-08-31
Add new macro to use with the `DeclarationEngine`

emilyaherbert:
Adds a simple new macro to use with the DE


otrho:
This is already implemented via `CompileResult::from()` -- is there a reason it needs to be a macro?

emilyaherbert:
> This is already implemented via `CompileResult::from()` -- is there a reason it needs to be a macro?

I didn't realize that! We can close this PR then.

PR #2689, opened 2022-09-01
Collect logged types and include them in the JSON ABI + `__log` intrinsic

### LINKED ISSUE #2130: Generate type and type IDs for event logging in a JSON file
In a separate JSON file (called `*-types.json`, we can emit all the types used by a contract and assign them IDs. When logging events, the `log<T>()` call can be modified to include an intrinsic that would help point to the right type that's being logged so that the SDK can pick that up from the JSON file and interpret the data being logged and display it properly.

mohammadfawaz:
Closes #2130 
## Main changes:
- Introduce a new intrinsic `__log` which takes a value of any type. Type check the intrinsic as usual.
- Change the pass that checks for unresolved types to a pass that collects information about that types. This collects logged type as well as unresolved types (and maybe more things in the future?). Unresolved types lead to errors as before while logged types are collected and store in `TypedProgram`. So, the method `check_for_unresolved_types` is now called `collect_types_metadata` and simply collects type information instead of returning errors, and that information is interpreted later in `lib.rs`.
- Update the JSON ABI generation to also include a `loggedTypes` array. This is an array of `JsonLoggedType` which contains a unique ID per `log` as well as a `JsonTypeApplication`. Types that are logged are added to the types array which is shared with the contract functions.
- Introduce an IR instruction `log` which takes a value to log, its type, and an ID (has to be `u64`). This ID is basically the ID of the log that is emitted in the JSON file.
- Codgen transforms the `log` instruction into either `log` or `logd` depending on the type of the value being logged
  - In the case of a `log`, `$rA` contains the value being logged and `$rB` contains the ID of the log
  - In the case of a `logd`, `$rB` contains the ID of the log while `$rC` and `$rD` both describe the value being logged (base pointer and length).
  - The above matches the behaviour of the standard library today if the log ID field was omitted.
- Update the standard library to use the new `__log` intrinsic


sway-ir/src/parser.rs
```diff
@@ -251,6 +252,11 @@ mod ir_builder {
                     IrAstOperation::Load(src)
                 }
 
+            rule op_log() -> IrAstOperation
+                = "log" _ log_ty:ast_ty() log_val:id() comma() _ log_id:id() {
```
otrho, line 256:
```suggestion
                = "log" _ log_ty:ast_ty() log_val:id() comma() log_id:id() {
```
The convention is whitespace is handled at the lowest rule, any call to another rule doesn't need `_` explicitly.

sway-ir/src/printer.rs
```diff
@@ -549,6 +549,21 @@ fn instruction_to_doc<'a>(
...
             ),
+            Instruction::Log {
+                log_val,
+                log_ty,
+                log_id,
+            } => maybe_constant_to_doc(context, md_namer, namer, log_val)
+                .append(maybe_constant_to_doc(context, md_namer, namer, log_id))
+                .append(Doc::line(
+                    Doc::text(format!(
+                        "log {} {}, {}",
```
otrho, line 560: No comma after the type?  Feels inconsistent -- but then I took a quick look elsewhere in the file and there is no consistency already. 🙁 We did consider removing commas but they're needed for multi-word args like types.  This still needs to be looked at one day.

otrho (dismissed):

sway-core/src/type_system/type_id.rs
```diff
@@ -31,18 +31,18 @@ impl From<usize> for TypeId {
...
         let span_override = if let TypeInfo::Ref(_, span) = look_up_type_id_raw(*self) {
             Some(span)
         } else {
             None
         };
         match look_up_type_id(*self) {
-            UnknownGeneric { name } => vec![CompileError::UnableToInferGeneric {
-                ty: name.as_str().to_string(),
-                span: span_override.unwrap_or_else(|| name.span()),
+            UnknownGeneric { name } => vec![TypeMetadata::UnresolvedType {
```
sezna, line 43: Since this is after type checking, there will be no further type unification happening -- should we also generate `TypeMetadata::UnresolvedType` if the type is `TypeInfo::Unknown`?

sway-ir/src/printer.rs
mohammadfawaz: I was honestly just following `ret`. `ret` and `log` are pretty much the same except for the extra ID parameter and the fact that `log` does not terminate the program.

sway-core/src/type_system/type_id.rs
mohammadfawaz: Done!

mohammadfawaz: Alright, so CI fails because apparently some `Unknown` types linger after type checking 🤔 . I reverted the change for now but I'll create an issue to investigate that weird behavior.

sezna: Hmmmmmm ok. Thanks for looking into it.

sezna (approved):

otrho (approved):

PR #2692, opened 2022-09-01
fix: Unformatted comment spans add extra newline

### LINKED ISSUE #2691: Unformatted comment spans add extra newline
```rust
script;

// use statements
use std::*;
```
formats into:
```rust
script;


// use statements
use std::*;
```
Each additional format call adds more extra newlines
2nd format:
```rust
script;



// use statements
use std::*;
```

kayagokalp:
closes #2691.

## TODO

- [x] Add a test

### DIFF
--- a/sway-fmt-v2/src/utils/map/newline.rs
+++ b/sway-fmt-v2/src/utils/map/newline.rs
@@ -172,6 +172,23 @@ fn format_newline_sequnce(newline_sequence: &NewlineSequence, threshold: usize) 
     }
 }
 
+/// Checks for newlines that are already in the source code.
+fn find_already_present_extra_newlines(from: usize, src: String) -> usize {
+    let mut number_of_newlines_present = 0;
+    for char in src.chars().skip(from) {
+        if char == '\n' {
+            number_of_newlines_present += 1;
+        } else {
+            break;
+        }
+    }
+    if number_of_newlines_present == 0 {
+        0
+    } else {
+        number_of_newlines_present - 1
+    }
+}
+
 /// Inserts after given span and returns the offset.
 fn insert_after_span(
     from: &ByteSpan,
@@ -182,12 +199,18 @@ fn insert_after_span(
 ) -> Result<usize, FormatterError> {
     let iter = newline_sequences_to_insert.iter();
     let mut sequence_string = String::new();
+    let mut newlines_to_skip =
+        find_already_present_extra_newlines(from.end, formatted_code.to_string());
     for newline_sequence in iter {
-        write!(
-            sequence_string,
-            "{}",
-            format_newline_sequnce(newline_sequence, threshold)
-        )?;
+        if newlines_to_skip != 0 {
+            newlines_to_skip -= 1;
+        } else {
+            write!(
+                sequence_string,
+                "{}",
+                format_newline_sequnce(newline_sequence, threshold)
+            )?;
+        }
     }
     let mut src_rope = Rope::from_str(formatted_code);
     src_rope.insert(from.end + offset, &sequence_string);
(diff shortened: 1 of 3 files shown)


sway-fmt-v2/src/utils/map/newline.rs
```diff
@@ -182,12 +199,18 @@ fn insert_after_span(
...
+        if newlines_to_skip != 0 {
+            newlines_to_skip -= 1;
+        } else {
+            write!(
+                sequence_string,
+                "{}",
+                format_newline_sequnce(newline_sequence, threshold)
+            )?;
+        }
     }
```
mitchmindtree, line 214: It looks like this could also be written as:

```rust
for newline_sequence in iter.skip(newlines_to_skip) {
    write!(
        sequence_string,
        "{}",
         format_newline_sequnce(newline_sequence, threshold),
    )?;
}
```

mitchmindtree (commented):
Thanks for digging into this!

Could we add a couple tests in with this PR? Even if we just modify some existing tests to reformat the `expected` output that could be a good start?

sway-fmt-v2/src/utils/map/newline.rs
kayagokalp: Actually, I first wrote it like this but realized that a newline sequence can have multiple newlines in it.

For example

```rust
script;

fn main() {
    let number = 10;
    let number5 = 10;
}
```

would produce the following newline sequences:

1. newline_sequence { start: .., end: .. length: 2}
2. newline_sequence { start: .., end: .. length: 1}
3. newline_sequence { start: .., end: .. length: 1}

(I did not took those from a debug output so exact field name etc can change but the logic is like this)

kayagokalp:
I had to change how function signatures inside traits are formatted since it was failing in the stability check but I think the new one looks more reasonable and it is now matching `rustfmt`
```rust
trait test {
    fn test();
    fn non_test();
}
```

was getting formatted into

```rust
trait test {
    fn test();

    fn non_test();
}
```

it is now matching `rustfmt` and becomes

```rust
trait test {
    fn test();
    fn non_test();
}
```

If user inserts a newline in between them like: 
```rust
trait test {
    fn test();

    fn non_test();
}
```
it will be formatted into:
```rust
trait test {
    fn test();

    fn non_test();
}
```

sway-fmt-v2/src/utils/map/newline.rs
mitchmindtree: Hmm I'm not quite sure I'm following here - I think your implementation in this PR and my suggestion look to be semantically equivalent?

```rust
    for newline_sequence in iter {
        if newlines_to_skip != 0 {
            newlines_to_skip -= 1;
        } else {
            write!(
                sequence_string,
                "{}",
                format_newline_sequnce(newline_sequence, threshold)
            )?;
        }
    }
```
^ Here, for each value of `newlines_to_skip` that is greater than `0`, we skip writing that `newline_sequence` element. It looks to me like this is the same as skipping the first `newlines_to_skip` number of elements. A more concise way of writing this is using the `skip` adapter like so:

```rust
for newline_sequence in iter.skip(newlines_to_skip) {
    write!(
        sequence_string,
        "{}",
         format_newline_sequnce(newline_sequence, threshold),
    )?;
}
```

Let me know if I'm still missing something here! Maybe the intended logic is a little different?

kayagokalp: Ahh, yeah my bad. While trying to understand the problem that popped up during testing (the one that I mentioned through slack) I thought this was the problem than realized it was not the case 🤦

mitchmindtree (approved):
Nice one!

ra0x3 (approved):

Issue #2551, opened 2022-08-16
Introduce a new `raw_ptr` type a disallow it to be returned from scripts


mohammadfawaz:
For example:
```rust
struct RawVec<T> {
    #[pointer]
    ptr: u64,
    cap: u64,
}
```
The immediate benefit of this is preventing [`RawVec<T>`](https://github.com/FuelLabs/sway/blob/88820bed80e9b7e68ec0e1a78c64f6b70d7e1dd4/sway-lib-std/src/vec.sw#L9) (and by extension `Vec<T>`) from being returned from a script because when we cross the script boundary, the `ptr` value is not meaningful.

This annotation may have future uses as well.

An alliterative is to introduce pointer/reference types (other than `mut ref` function parameters) but that's a rabbit hole we probably don't want to explore.

Edit: The above proposal has been voted against and the preferred way of doing this seems to be to introduce a `raw_ptr` type as @canndrew as described in https://github.com/FuelLabs/sway/pull/2699

otrho:
This feels a bit hacky, as does a lot of our avoidance of pointers.  But I really don't want to add references.

Would an alternative be an 'address' type, which is very much locked down?  I.e. you can't create them from anywhere other than a heap or stack allocation.  It'd be different to a reference or pointer, where you would take the address of a value.  And then we limit it specifically to only do the operations we need in `Vec` and similar data structures.  :thinking:

PR #2699, opened 2022-09-02
Add raw untyped pointer type

canndrew:
This PR adds a `raw_ptr` type to sway. These are implemented in sway-core as `TypeInfo::RawUntypedPtr`, but are (currently) lowered to `Uint(64)` in sway-ir. Raw pointers can only be created and manipulated using `asm` blocks or by using the memory-manipulating functions in the standard library which have been updated to use this type rather than `u64`.

It's still possible to return a `raw_ptr` across a script boundary, so #2551 isn't fixed yet. But this PR puts the groundwork in place so that the necessary checks can be added in a follow-up PR.


otrho:
CI failure @canndrew

canndrew:
@otrho: This should be good to go now.

sway-lib-std/src/vec.sw
```diff
@@ -3,11 +3,11 @@ library vec;
 use ::alloc::{alloc, realloc};
 use ::assert::assert;
 use ::intrinsics::size_of;
-use ::mem::{copy, read, write};
+use ::mem::{copy, ptr_offset, read, write};
 use ::option::Option;
 
 struct RawVec<T> {
-    ptr: u64,
+    ptr: raw_ptr,
```
mohammadfawaz, line 10: @digorithm what are the implications of this on the work the SDK team is doing to support `Vec`? A few things:
- `raw_ptr` is basically just a `u64` in the bytecode
- `raw_ptr` shows up as `raw untyped ptr` in the JSON ABI.
- It is not possible to initialize a `raw_ptr` without an explicit `asm` block.

test/src/sdk-harness/test_projects/registers/src/main.sw
```diff
@@ -3,47 +3,51 @@ contract;
...
 }
 
 impl Registers for Contract {
-    fn get_overflow() -> u64 {
-        overflow()
-    }
-
-    fn get_program_counter() -> u64 {
+    /*
+    fn get_program_counter() -> raw_ptr {
```
mohammadfawaz, line 26: How come you've had to disable these tests?

canndrew: It's not currently possible to return raw pointers back the rust through the SDK. We need to discuss how this is supposed to work. For instance if someone returns a `Vec` then they probably want to allocate on the rust side and copy the data across rather than returning a (wrapped) `u64` representing the pointer. So for now the parts of the SDK tests that return raw pointers have been disabled.

sway-lib-std/src/vec.sw
canndrew: > `raw_ptr` is basically just a u64 in the bytecode.

Correct. I'm not sure if we'd want something different in the IR or bytecode since the point of this (as I understand it) is to prevent misuse of pointers earlier during compilation, ie. during type-checking.

> `raw_ptr` shows up as `raw untyped ptr` in the JSON ABI.

Is that a bad thing? I could make it show up as `raw_ptr` easily enough. Either way the SDK will have to know how to handle this type.

> It is not possible to initialize a `raw_ptr` without an explicit `asm` block.

Currently yes, though that's sort of intentional. Raw pointers are intended to be highly locked-down. You can obtain them via standard library functions like `alloc` and `stack_ptr` or by modifying an existing pointer using `ptr_offset`. Any other way that someone might want to create a pointer is likely to be highly unsafe.

sway-core/src/type_system/type_info.rs
```diff
@@ -86,6 +86,8 @@ pub enum TypeInfo {
     Storage {
         fields: Vec<TypedStructField>,
     },
+    /// Raw untyped pointers.
```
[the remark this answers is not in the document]
canndrew: I've expanded this doc comment. Does that look good?

sway-core/src/convert_parse_tree.rs
```diff
@@ -1039,6 +1039,7 @@ fn type_name_to_type_info_opt(name: &Ident) -> Option<TypeInfo> {
         "unit" => Some(TypeInfo::Tuple(Vec::new())),
         "byte" => Some(TypeInfo::Byte),
         "b256" => Some(TypeInfo::B256),
+        "raw_ptr" => Some(TypeInfo::RawUntypedPtr),
```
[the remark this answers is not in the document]
canndrew: Not that I'm aware of. Currently intrinsic functions and operators are implemented rather hackily, so there's definitely a need for it.

sezna (approved):

sezna (approved):

sezna:
Marking as a draft pending discussions w/ SDK team and an RFC.

mohammadfawaz:
This is certainly the right approach here and I think this is a good change! Thanks @canndrew.

Thant being said, this feels a bit rushed and I'm not comfortable with disabling tests until further notice. I'd like to sync up with the SDK team on this first and make sure we have path forward before merging it and breaking a bunch of things.

I will turn this into a draft in the mean time.

mohammadfawaz:
Closing this in favor of https://github.com/FuelLabs/sway/pull/2828.

PR #2705, opened 2022-09-02
Extend support for `Input::Message` type fields in std::inputs

nfurfaro:
Also adds tests for existing untested functions in the stdlib !


nfurfaro:
Hmm. I need to review the git history of this branch, there should not be so many commits.

PR #2707, opened 2022-09-02
Extend support for Input::Message type fields in std::inputs

nfurfaro:
Also adds tests for untested functions currently in the stdlib.


pixelcircuits:
Can we add support for getting the hash of the script bytecode? I've been using this in the current predicate for the bridge:
```

const GTF_SCRIPT_SCRIPT_LENGTH = 0x005;
const GTF_SCRIPT_SCRIPT = 0x00B;

/// Get the hash of the script bytecode
pub fn tx_script_bytecode_hash() -> b256 {
    // Get the script memory details
    let mut result_buffer: b256 = ZERO_B256;
    let script_length = __gtf::<u64>(0, GTF_SCRIPT_SCRIPT_LENGTH);
    let script_ptr = __gtf::<u64>(0, GTF_SCRIPT_SCRIPT);
    
    // Run the hash opcode for the script in memory
    asm(hash: result_buffer, ptr: script_ptr, len: script_length) {
        s256 hash ptr len;
        hash: b256
    }
}
```

I feel like it belongs in `tx.sw` since it requires the constant values for certain GTF fields.

nfurfaro:
>Can we add support for getting the hash of the script bytecode?

@pixelcircuits I'm not opposed to this. Would you mind opening a separate issue to track it?

pixelcircuits:
> > Can we add support for getting the hash of the script bytecode?
> 
> @pixelcircuits I'm not opposed to this. Would you mind opening a separate issue to track it?

Sure! #2721

sway-lib-std/src/inputs.sw
```diff
@@ -156,3 +175,57 @@ pub fn input_count() -> u8 {
...
+    __gtf::<u64>(index, GTF_INPUT_MESSAGE_PREDICATE_LENGTH)
+}
+
+/// Get the predicate data length of the input message at `index`.
+pub fn input_message_predicate_data_length(index: u64) -> u64 {
+    __gtf::<u64>(index, GTF_INPUT_MESSAGE_PREDICATE_DATA_LENGTH)
+}
+
+/// Get the data of the input message at `index`.
+pub fn input_message_data<T>(index: u64) -> T {
```
pixelcircuits, line 224: This should include an `offset: u64` field so individual reads in a larger data set can be performed

nfurfaro: @pixelcircuits Would the user need to pass in a `start_offset` and a `length` for the data they actually care about then?

pixelcircuits: No, I've been using my own version of this function like this: 
```
pub fn input_message_data<T>(index: u64, offset: u64) -> T {
    read::<T>(__gtf::<u64>(index, GTF_INPUT_MESSAGE_DATA) + offset)
}
```

It let's me parse data from a message like this:
```
// Parse the message data
let contract_id: b256 = input_message_data(msg_idx, 0);
storage.data1 = ~ContractId::from(contract_id);

let num: u64 = input_message_data(msg_idx, 32);
storage.data2 = num;

let big_num: b256 = input_message_data(msg_idx, 32 + 8);
storage.data3 = big_num;

let address: b256 = input_message_data(msg_idx, 32 + 8 + 32);
storage.data4 = ~Address::from(address);
```

It seems to work just fine.

sway-lib-std/src/inputs.sw
```diff
@@ -156,3 +175,57 @@ pub fn input_count() -> u8 {
...
     }
 }
+
+////////////////////////////////////////
+// Message Inputs
+////////////////////////////////////////
+
+/// Get the message id of the input message at `index`.
+pub fn input_message_msg_id(index: u64) -> b256 {
+    __gtf::<b256>(index, GTF_INPUT_MESSAGE_MESSAGE_ID)
```
pixelcircuits, line 185: Should the functions be checking that the input type is actually a message? I'm not sure if we should return `Option::None` (like `input_predicate_data_pointer`) or just `revert(0)` (like `input_predicate_data`). I'm kinda leaning towards a revert

sway-lib-std/src/inputs.sw
```diff
@@ -156,3 +175,57 @@ pub fn input_count() -> u8 {
...
+pub fn input_message_recipient(index: u64) -> Address {
+    ~Address::from(__gtf::<b256>(index, GTF_INPUT_MESSAGE_RECIPIENT))
+}
+
+/// Get the nonce of input message at `index`.
+pub fn input_message_nonce(index: u64) -> u64 {
+    __gtf::<u64>(index, GTF_INPUT_MESSAGE_NONCE)
+}
+
+/// Get the witness index of the input messasge at `index`.
```
pixelcircuits, line 203: Typo for "messasge"

sway-lib-std/src/inputs.sw
nfurfaro: Yeah, it could probably go either way here...
There are a bunch of these functions using `__gtf` that only make sense for specific types of inputs/outputs, and only some check the type IIRC.
I'll need to review for consistency and probably refactor a bit.

pixelcircuits:
This PR should wait until #2810 is finished since there is a lot of overlap

sway-lib-std/src/inputs.sw
```diff
@@ -115,12 +160,142 @@ pub fn input_predicate_data<T>(index: u64) -> T {
...
+    let type = input_type(index);
+    match type {
+        Input::Coin => Option::Some(__gtf::<u16>(index, GTF_INPUT_COIN_PREDICATE_DATA_LENGTH)),
+        Input::Message => Option::Some(__gtf::<u16>(index, GTF_INPUT_MESSAGE_PREDICATE_DATA_LENGTH)),
+        Input::Contract => Option::None,
+    }
+}
+
+////////////////////////////////////////
+// Message Inputs
```
pixelcircuits, line 239: did you mean `Coin Inputs`?

pixelcircuits:
This PR is looking good. I just wanted to point to the [tests on my now defunct branch](https://github.com/FuelLabs/sway/blob/38317cac5c0a58f2a9a7174db7dbac2586efc06a/test/src/sdk-harness/test_projects/tx_fields/mod.rs#L340) for testing things like predicate bytecode and data when you get to that.

sway-lib-std/src/inputs.sw
nfurfaro: I've made a change here by implementing `Eq` for `Input`, and then adding a function to validate an Input type which can be reused:

```
impl Eq for Input {
        match (self, other) {
            (Input::Coin, Input::Coin) => true,
            (Input::Contract, Input::Contract) => true,
            (Input::Message, Input::Message) => true,
            _ => false,
        }
    }
}

fn valid_input_type(index: u64, expected_type: Input) -> bool {
    let input_type: Input = input_type(index);
    input_type == expected_type
}

/// Get the message id of the input message at `index`.
pub fn input_message_msg_id(index: u64) -> b256 {
    assert(valid_input_type(index, Input::Message));
    __gtf::<b256>(index, GTF_INPUT_MESSAGE_MESSAGE_ID)
}
```

pixelcircuits:
Can we go ahead and use this issue to make all GTF_ const definitions public (`pub`) so that they can be referenced for more lower level solutions like the current bridge message-contract-predicate?

nfurfaro:
> Can we go ahead and use this issue to make all GTF_ const definitions public (`pub`) so that they can be referenced for more lower level solutions like the current bridge message-contract-predicate?

I think so. So far I've only been exposing them as needed, but no reason to keep any of them private.

test/src/sdk-harness/test_projects/tx_fields/mod.rs
```diff
@@ -519,10 +515,9 @@ mod inputs {
                     .await
                     .unwrap();
                 let messages = wallet.get_messages().await?;
-                let inputs = wallet.get_inputs_for_messages(0).await?;
-                // let contract_bytes: [u8; 32] = result.value.into();
-                // let message_id: [u8; 32] = *inputs[0].message_id().unwrap().into();
-                // assert_eq!(result.value, Bits256(message_id));
+                let message_id: [u8; 32] = *messages[0].message_id.0 .0;
```
pixelcircuits, line 518: What exactly is this doing? The space between the two `.0` is throwing me off...

nfurfaro:
hmm. @mohammadfawaz any ideas on why CI would fail for this unrelated test?
image

sway-lib-std/src/inputs.sw
```diff
@@ -3,8 +3,12 @@
 library inputs;
 
 use ::address::Address;
+use ::assert::assert;
+use ::bytes::Bytes;
 use ::option::Option;
 use ::revert::revert;
+use ::contract_id::ContractId;
+use ::constants::BASE_ASSET_ID;
```
mohammadfawaz, line 11: Alphabetize :)

sway-lib-std/src/inputs.sw
```diff
@@ -73,7 +91,16 @@ pub fn input_type(index: u64) -> Input {
...
+/// Get the transaction inputs count
+pub fn input_count() -> u8 {
+    let type = tx_type();
+    match type {
+        Transaction::Script => __gtf::<u8>(0, GTF_SCRIPT_INPUTS_COUNT),
+        Transaction::Create => __gtf::<u8>(0, GTF_CREATE_INPUTS_COUNT),
+    }
+}
+
+// Get the pointer of the input at `index`.
```
mohammadfawaz, line 103:
```suggestion
/// Get the pointer of the input at `index`.
```

sway-lib-std/src/inputs.sw
```diff
@@ -82,7 +109,21 @@ pub fn input_pointer(index: u64) -> u64 {
...
+    let type = input_type(index);
+    match type {
+        Input::Coin => Option::Some(__gtf::<u64>(index, GTF_INPUT_COIN_AMOUNT)),
+        Input::Message => Option::Some(__gtf::<u64>(index, GTF_INPUT_MESSAGE_AMOUNT)),
+        Input::Contract => return Option::None,
+    }
+}
+
 /// If the input's type is `InputCoin` the owner as an Option::Some(owner).
+/// If the input's type is `InputMessage` the owner as an Option::Some(recipient).
```
mohammadfawaz, line 126:
```suggestion
/// If the input's type is `InputCoin`, return the owner as an `Option::Some(owner)`.
/// If the input's type is `InputMessage`, return the recipient as an `Option::Some(recipient)`.
```

sway-lib-std/src/inputs.sw
```diff
@@ -82,7 +109,21 @@ pub fn input_pointer(index: u64) -> u64 {
     }
 }
 
+/// Get amount field from input at `index`.
+/// If the input's type is `InputCoin` or `InputMessage`,
+/// return the amount as an Option::Some(u64).
+/// Otherwise, returns Option::None.
```
mohammadfawaz, line 115:
```suggestion
/// return the amount as an `Option::Some(u64)`.
/// Otherwise, returns `Option::None`.
```

test/src/sdk-harness/test_artifacts/tx_contract/src/main.sw
```diff
@@ -1,20 +1,53 @@
...
+        tx_script_bytecode_hash,
+        tx_script_data_length,
+        tx_script_data_start_pointer,
+        tx_script_length,
+        tx_script_start_pointer,
+        tx_type,
+        tx_witness_data,
+        tx_witness_data_length,
+        tx_witness_pointer,
+        tx_witnesses_count,
```
mohammadfawaz, line 49: At this point, it must be better to just use `*` import for `inputs::` and `tx::`.

test/src/sdk-harness/Cargo.toml
```diff
@@ -25,3 +25,6 @@ name = "integration_tests"
 path = "test_projects/harness.rs"
 
 [workspace]
+
+# [patch.crates-io]
+# fuels = { path = "../../../../fuels-rs/packages/fuels" }
```
mohammadfawaz, line 30: Is this needed?

test/src/sdk-harness/test_artifacts/tx_contract/src/main.sw
```diff
@@ -1,20 +1,53 @@
...
         input_owner,
         input_pointer,
+        input_predicate,
+        input_predicate_data_length,
+        input_predicate_data_pointer,
+        input_predicate_length,
         input_type,
+        input_witness_index,
     },
+    logging::log,
```
mohammadfawaz, line 25: This is probably not needed?

sway-lib-std/src/inputs.sw
mohammadfawaz: Would be nice to standardize this across the whole file by adding the backticks like this.

sway-lib-std/src/inputs.sw
```diff
@@ -113,12 +157,149 @@ pub fn input_predicate_data<T>(index: u64) -> T {
...
+        Input::Contract => Option::None,
+    }
+}
+
+/// Get the predicate from the input at `index`.
+/// If the input's type is `InputCoin` or `InputMessage`,
+/// return the data, otherwise reverts.
+pub fn input_predicate(index: u64) -> Bytes {
+    let data = input_predicate_pointer(index);
+    let length = input_predicate_length(index).unwrap();
```
mohammadfawaz, line 216: Can this `unwrap` fail?

sway-lib-std/src/inputs.sw
```diff
@@ -113,12 +157,149 @@ pub fn input_predicate_data<T>(index: u64) -> T {
...
+/// Otherwise, returns Option::None.
+pub fn input_maturity(index: u64) -> Option<u32> {
+    let type = input_type(index);
+    match type {
+        Input::Coin => {
+            Option::Some(__gtf::<u32>(index, GTF_INPUT_COIN_MATURITY))
+        },
+        _ => {
+            Option::None
+        }
```
mohammadfawaz, line 257: The curly braces are not needed here.

sway-lib-std/src/outputs.sw
```diff
@@ -61,7 +61,7 @@ pub fn output_pointer(index: u64) -> u64 {
 
 /// Get the transaction outputs count for either tx type
 /// (transaction-script or transaction-create).
-pub fn output_count() -> u64 {
+pub fn output_count() -> u8 {
```
mohammadfawaz, line 64: This means that this is breaking change right?

sway-lib-std/src/inputs.sw
```diff
@@ -113,12 +157,149 @@ pub fn input_predicate_data<T>(index: u64) -> T {
...
+    }
+}
+
+/// Get the predicate data length from the input at `index`.
+/// If the input's type is `InputCoin` or `InputMessage`,
+/// return the data length as an Option::Some(u16).
+/// Otherwise, returns Option::None.
+pub fn input_predicate_data_length(index: u64) -> Option<u16> {
+    let type = input_type(index);
+    match type {
```
mohammadfawaz, line 235: Remind me, does writing `match input_type(index)` directly not work here?

mohammadfawaz (commented):
Great! A few minor comments.

sway-lib-std/src/outputs.sw
nfurfaro: I reverted this; not sure why I changed it.

sway-lib-std/src/inputs.sw
nfurfaro: It does indeed. fixed.

sway-lib-std/src/inputs.sw
nfurfaro: yes. It would just revert which is fine, but I've made that behaviour explicit now.

mohammadfawaz (approved):
All seem reasonable to me. Someone from @FuelLabs/application-dev team should also review though.

sway-lib-std/src/inputs.sw
```diff
@@ -13,112 +17,273 @@ use ::tx::{
...
-/// If the input's type is `InputCoin` the owner as an Option::Some(owner).
+/// Get amount field from input at `index`.
+/// If the input's type is `InputCoin` or `InputMessage`,
+/// return the amount as an `Option::Some(u64)`.
+/// Otherwise, returns `Option::None`.
+pub fn input_amount(index: u64) -> Option<u64> {
+    match input_type(index) {
+        Input::Coin => Option::Some(__gtf::<u64>(index, GTF_INPUT_COIN_AMOUNT)),
+        Input::Message => Option::Some(__gtf::<u64>(index, GTF_INPUT_MESSAGE_AMOUNT)),
+        Input::Contract => return Option::None,
```
Braqzen, line 117: How come some of these have a `return` and other `match` expressions do not? Is it needed in these cases?

sway-lib-std/src/inputs.sw
```diff
@@ -13,112 +17,273 @@ use ::tx::{
...
-// const GTF_INPUT_CONTRACT_STATE_ROOT = 0x111;
-// const GTF_INPUT_CONTRACT_TX_POINTER = 0x112;
-// const GTF_INPUT_CONTRACT_CONTRACT_ID = 0x113;
-// const GTF_INPUT_MESSAGE_MESSAGE_ID = 0x114;
-// const GTF_INPUT_MESSAGE_SENDER = 0x115;
-const GTF_INPUT_MESSAGE_RECIPIENT = 0x116;
-
-// const GTF_INPUT_MESSAGE_AMOUNT = 0x117;
-// const GTF_INPUT_MESSAGE_NONCE = 0x118;
+// pub const GTF_INPUT_COIN_TX_ID = 0x102;
```
Braqzen, line 27: Should these be commented out?

sway-lib-std/src/inputs.sw
```diff
@@ -13,112 +17,273 @@ use ::tx::{
...
+// Message Inputs
+////////////////////////////////////////
+/// Get the message id of the input message at `index`.
+pub fn input_message_msg_id(index: u64) -> b256 {
+    assert(valid_input_type(index, Input::Message));
+    __gtf::<b256>(index, GTF_INPUT_MESSAGE_MESSAGE_ID)
+}
+
+/// Get the sender of the input message at `index`.
+pub fn input_message_sender(index: u64) -> Address {
```
Braqzen, line 256: Should we have support for `EVMAddress`?

test/src/sdk-harness/test_projects/tx_fields/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "tx_predicate"
+
+[dependencies]
+std = { path = "../../../../../sway-lib-std" }
```
Braqzen, line 8:
```suggestion
std = { path = "../../../../../sway-lib-std" }

```

test/src/sdk-harness/test_projects/tx_fields/mod.rs
```diff
@@ -1,298 +1,747 @@
...
+const MESSAGE_DATA: [u8; 3] = [1u8, 2u8, 3u8];
+
 abigen!(
     TxContractTest,
     "test_artifacts/tx_contract/out/debug/tx_contract-abi.json",
 );
 
-async fn get_contracts() -> (TxContractTest, ContractId, WalletUnlocked) {
-    let wallet = launch_provider_and_get_wallet().await;
+async fn get_contracts() -> (TxContractTest, ContractId, WalletUnlocked, WalletUnlocked) {
```
Braqzen, line 19: This file contains modules inside modules. It's quite a lot of scrolling to look for things. I would consider splitting up the content into files similar to how the apps repo does it

sway-lib-std/src/inputs.sw
nfurfaro: For now, yes. They are being enabled by consecutive PRs and help to highlight what is still needed.

test/src/sdk-harness/test_projects/tx_fields/Forc.toml
nfurfaro: weird, locally this file already has a newline at the end.

test/src/sdk-harness/test_projects/tx_fields/mod.rs
nfurfaro: I reeealy want to merge this before the next big set of breaking changes from the SDK lands as I've had to update these files too many times already. I'll be back in here adding more tests soon enough after the next dependency version bump, and could perform a general refactoring at that time.

test/src/sdk-harness/Cargo.toml
nfurfaro: nope.

Braqzen (approved):

pixelcircuits (approved):

PR #2708, opened 2022-09-02
Fix some anchors that were removed by the new formatter

mohammadfawaz:


mohammadfawaz:
I guess this won't work because of CI failure.

PR #2724, opened 2022-09-05
Recover on type checked AST with errors

Centril:
Works towards https://github.com/FuelLabs/sway/issues/1674.

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -2116,7 +2116,7 @@ fn update_json_type_declaration(
 pub fn check(
     plan: &BuildPlan,
     silent_mode: bool,
-) -> anyhow::Result<(CompileResult<ParseProgram>, CompileResult<TypedProgram>)> {
+) -> anyhow::Result<CompileResult<(ParseProgram, Option<TypedProgram>)>> {
     //TODO remove once type engine isn't global anymore.
     sway_core::clear_lazy_statics();
     let mut namespace_map = Default::default();
@@ -2127,17 +2127,26 @@ pub fn check(
         let constants = manifest.config_time_constants();
         let dep_namespace =
             dependency_namespace(&namespace_map, &plan.graph, node, constants).expect("TODO");
-        let parsed_result = parse(manifest, silent_mode)?;
+        let CompileResult {
+            value,
+            mut warnings,
+            mut errors,
+        } = parse(manifest, silent_mode)?;
 
-        let parse_program = match &parsed_result.value {
-            None => bail!("unable to parse"),
+        let parse_program = match value {
+            None => return Ok(CompileResult::new(None, warnings, errors)),
             Some(program) => program,
         };
 
-        let ast_result = sway_core::parsed_to_ast(parse_program, dep_namespace, false);
+        let ast_result = sway_core::parsed_to_ast(&parse_program, dep_namespace, false);
+        warnings.extend(ast_result.warnings);
+        errors.extend(ast_result.errors);
 
-        let typed_program = match &ast_result.value {
-            None => bail!("unable to type check"),
+        let typed_program = match ast_result.value {
+            None => {
+                let value = Some((parse_program, None));
+                return Ok(CompileResult::new(value, warnings, errors));
+            }
             Some(typed_program) => typed_program,
         };
 
@@ -2149,7 +2158,8 @@ pub fn check(
 
         // We only need to return the final `TypedProgram`.
         if i == plan.compilation_order.len() - 1 {
-            return Ok((parsed_result, ast_result));
+            let value = Some((parse_program, Some(typed_program)));
+            return Ok(CompileResult::new(value, warnings, errors));
         }
     }
     bail!("unable to check sway program: build plan contains no packages")
--- a/sway-lsp/src/core/session.rs
+++ b/sway-lsp/src/core/session.rs
@@ -162,10 +162,23 @@ impl Session {
         if let Ok(manifest) = pkg::ManifestFile::from_dir(&manifest_dir) {
             if let Ok(plan) = pkg::BuildPlan::from_lock_and_manifest(&manifest, locked, offline) {
                 //we can then use them directly to convert them to a Vec<Diagnostic>
-                if let Ok((parsed_res, ast_res)) = pkg::check(&plan, silent_mode) {
-                    // First, populate our token_map with un-typed ast nodes
+                if let Ok(CompileResult {
+                    value,
+                    warnings,
+                    errors,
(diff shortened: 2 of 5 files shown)


mohammadfawaz:
The error expected in the `should_fail` failing test in CI is collected [after type checking](https://github.com/FuelLabs/sway/blob/d79bb51bb1c031f9c0d6af4f73ba9775610383fd/sway-core/src/semantic_analysis/program.rs#L135).. I wonder if those error are somewhat missed now?

Either way, I'm converting this to a draft until CI passes :)

emilyaherbert (approved):

JoshuaBatty:
Hi @Centril, thanks for tackling this issue. I would love to review this PR properly before it's merged if possible. I'm currently away in holidays without a laptop but will be back at the start of next week.

sway-core/src/lib.rs
```diff
@@ -265,191 +228,141 @@ pub fn parsed_to_ast(
...
+        errors.push(e);
     }
 
     // Check that all storage initializers can be evaluated at compile time.
-    let CompileResult {
-        value: typed_program_with_storage_slots_result,
-        warnings: new_warnings,
-        errors: new_errors,
-    } = typed_program.get_typed_program_with_initialized_storage_slots(
+    let typed_wiss_res = typed_program.get_typed_program_with_initialized_storage_slots(
```
sezna, line 262: took me a second to figure out that `wiss` is `with_initialized_storage_slots` 😆

sezna (dismissed):
cc @JoshuaBatty

mohammadfawaz:
Merge conflicts :)

JoshuaBatty (changes requested):
Hi @Centril back from holiday and have had some time to look over this. 

So LSP actually calls `forc::check()` internally on every didChange event in order to get the parsed and typed AST's back. There are currently 2 issues that need resolving. Considering the following example: 

```
script;

fn main() {
  let y: u64 = 0;
  let x = 
}
```

Running `forc check` on this file returns with an `Error: unable to parse`. Even the though the `let x = ` line is not completed yet, the compiler should still return the parsed and typed AST's so we can traverse them in the language server and extract info such as semantics, definitions, types etc.. while the user is coding. 

Similarly in the following example: 

```
script;

fn main() {
  let y: u64 = 0;
  let x: str[3] = y;
}
```
Running `forc check` on this file returns with an `Error: unable to type check`. Obviously here `x` is of type `str[3]` and is being assigned a `u64` which is incorrect. In this situation instead of bailing and only returning the errors and warnings from the compiler, we also need the parsed and typed program for the rest of the file. Otherwise we are left in the same situation where we can only traverse the AST's in the language server and derive the information we need when fully formed code able to be compiled. 

Hope this makes sense? Let me know if I can help clarify anything further.

JoshuaBatty:
> @JoshuaBatty This is a matter of adding more _parser recovery_ for various cases, of which there are an infinite number. To handle your example, you'd need to handle the specific case of `let x = ` being incomplete. In other words, your first example never reaches the type checker to begin with. There was never a concrete syntax tree or untyped AST to begin with.

Ok that's fair. Is there a plan to enable this in future PR's as this is what #1674 was requesting which is what this PR apparently solves.... ?

> My understanding is the compiler will assign `x` the type `ErrorRecovery`, allowing type checking for other bits to continue, at least that's what the compiler sources suggest. Is that not so?

If I print the `typed_res` that is assigned at line 210 in the `parsed_to_ast` function then I get the following. I would assume that along with the warnings and errors, that there would be `Some(TypedAst)` assigned to `value` as well but in this case, containing a `ErrorRecovery` type as one of the nodes?

```
CompileResult {
    value: None,
    warnings: [],
    errors: [
        TypeError(
            MismatchedType {
                expected: str[3],
                received: u64,
                help_text: "Variable declaration's type annotation does not match up with the assigned expression's type.",
                span: Span {
                    src (ptr): 0x000055bcdd85f180,
                    path: Some(
                        "/home/josh/rust/fuel/sway/sway-lsp/test_programs/scratch_pad/src/main.sw",
                    ),
                    start: 61,
                    end: 62,
                    as_str(): "y",
                },
            },
        ),
    ],
}
```

forc-pkg/src/pkg.rs
```diff
@@ -2110,9 +2108,9 @@ pub fn check(
 
         let ast_result = sway_core::parsed_to_ast(parse_program, dep_namespace);
 
-        let typed_program = match &ast_result {
-            CompileAstResult::Failure { .. } => bail!("unable to type check"),
-            CompileAstResult::Success { typed_program, .. } => typed_program,
+        let typed_program = match &ast_result.value {
+            None => bail!("unable to type check"),
```
JoshuaBatty, line 2112: We need to return the warnings and errors here so was can access them in LSP. 

Same above where we match on `parsed_result.value`. Errors and warnings need to be returned.

JoshuaBatty:
Thanks @Centril. I've also opened #2817 so we can have the compiler return a typed AST even if parts of the code fail type checking. I noticed that we now at least get back the `ParseProgram` in this instance, but we then lose type information for the rest of the program which needs to be recovered still.

sway-lsp/src/core/session.rs
```diff
@@ -233,26 +242,29 @@ impl Session {
...
-                }
+                let root_nodes = typed_program.root.all_nodes.iter();
+                let sub_nodes = typed_program
+                    .root
+                    .submodules
+                    .iter()
+                    .flat_map(|(_, submodule)| &submodule.module.all_nodes);
+                root_nodes
+                    .chain(sub_nodes)
+                    .for_each(|node| traverse_typed_tree::traverse_node(node, &self.token_map));
```
JoshuaBatty, line 253: This is much nicer :)

sway-lsp/src/core/session.rs
```diff
@@ -162,10 +162,23 @@ impl Session {
...
                 //we can then use them directly to convert them to a Vec<Diagnostic>
-                if let Ok((parsed_res, ast_res)) = pkg::check(&plan, silent_mode) {
-                    // First, populate our token_map with un-typed ast nodes
+                if let Ok(CompileResult {
+                    value,
+                    warnings,
+                    errors,
+                }) = pkg::check(&plan, silent_mode)
+                {
+                    // FIXME(Centril): Refactor parse_ast_to_tokens + parse_ast_to_typed_tokens
```
JoshuaBatty, line 171: Is the plan to address this FIXME in this PR or in a follow up PR?

JoshuaBatty (approved):
Just left a comment about a fixme you left for yourself. Happy to land this as is for now if you want to address it in a follow up PR.

Issue #2495, opened 2022-08-09
Making `Shape` state changes in `Format` API more robust in `sfv2`


eureka-cpu:
@mitchmindtree : 
>I have a slight hesitation about exposing the formatter's Shape state as done in this PR as I can imagine it might lead to cases where a Format implementation could easily forget to reset the line style or width heuristics at the end of an item's formatting, leading to following items being mysteriously incorrect only on occasion.
>
>We could potentially avoid this with methods on Formatter that scope these temporary shaping changes, e.g.
>```rust
>formatter.with_line_style(new_style, |formatter| {
>    // Do some formatting with `new_style`.
>});
>// Automatically revert to original style after scope exited.
>```
>With an API along these lines, a Format implementation could never accidentally leave the Formatter's inner state in some unexpected state.
>
>All that said, this might require some more significant redesign, and we can investigate diving into this in future issues/PRs.

PR #2727, opened 2022-09-05
Fixes `Shape::width` bleed from file to file

eureka-cpu:
This is an artifact of shape state changes and this PR is a quick fix for the issue #2495 which I am currently tackling. This at least will solve the problem of files being incorrectly formatted when formatted as a group for the time being before a PR for the previously mentioned issue is ready.


Braqzen (approved):

kayagokalp (approved):
utACK

PR #2729, opened 2022-09-05
doc: added a section about interaction with the testnet

### LINKED ISSUE #2728: Add a small section to `forc-client`doc about deploying to test net
We should mention default gas price for `forc-deploy` is 0 and also include the end-point.

kayagokalp:
closes #2728.

### DIFF
--- a/docs/src/forc_client.md
+++ b/docs/src/forc_client.md
@@ -47,3 +47,19 @@ forc wallet account <account_index>
 > ```sh
 > forc-run --unsigned
 > ```
+
+## Interacting with the testnet
+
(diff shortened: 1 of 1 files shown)


mohammadfawaz (approved):

Braqzen (approved):
Looks good to me.

Does this mean that I can create a wallet using the section above, locally, and then use that address in the faucet to get the coins sent to the address on-chain?
After this when I try to deploy it checks the amount that I have on-chain and based on that it will figure out whether I have enough?

I think the steps by themselves are fine but I'd like more info on how this actually works so that there is a seamless step-by-step from creation to deployment.

PR #2732, opened 2022-09-06
Make `Shape` state changes in `Format` API more robust

### LINKED ISSUE #2495: Making `Shape` state changes in `Format` API more robust in `sfv2`

eureka-cpu:
Closes #2495

This PR adds automatic handling of `Shape` state changes, which will stop file bleed and clean up the API


examples/structs/src/main.sw
```diff
@@ -84,5 +81,5 @@ fn struct_destructuring() {
         Point { x: 2, y: 4 },
         Point { x: 3, y: 6 },
     );
-    let (Point{x:x0, y:y0, }, Point{x:x1, y:y1, }) = struct_in_tuple;
+    let (Point{x:x0, y:y0}, Point{x:x1, y:y1}) = struct_in_tuple;
```
Braqzen, line 84: How come the `Point` on lines 81 & 82 is formatted as `Point<space>{` but on line 84 there is no space?

swayfmt/src/items/item_fn/mod.rs
```diff
@@ -26,15 +26,26 @@ impl Format for ItemFn {
...
+                &formatter.shape,
+                Some(0), // In some cases we will want to update parts of Shape
+                Some(CodeLine::new(LineStyle::Normal, ExprKind::Function)),
+            ),
+            |formatter| -> Result<(), FormatterError> {
+                self.fn_signature.format(formatted_code, formatter)?;
+                let body = self.body.get();
+                if !body.statements.is_empty() || body.final_expr_opt.is_some() {
+                    Self::open_curly_brace(formatted_code, formatter)?;
+                    self.body.get().format(formatted_code, formatter)?;
```
Braqzen, line 40: Could this be `body.format(formatted_code, formatter)?;`?

eureka-cpu: ah yep

examples/structs/src/main.sw
eureka-cpu: this has to do with how inner expressions are formatted within the tuple expr type, I've meant to file an issue about this

swayfmt/src/items/item_enum/mod.rs
```diff
@@ -19,94 +22,101 @@ impl Format for ItemEnum {
...
-        if let Some(generics) = &self.generics {
-            generics.format(formatted_code, formatter)?;
-        }
+        formatter.with_shape(
+            Shape::from(
+                &formatter.shape,
+                Some(0), // In some cases we will want to update parts of Shape
+                Some(CodeLine::new(LineStyle::Multiline, ExprKind::Undetermined)),
+            ),
+            |formatter| -> Result<(), FormatterError> {
```
mitchmindtree, line 31: One thing that might help to cut down on indentation a little is to first assign the new shape to a binding, e.g.

```rust
let new_shape = Shape::from(
    &formatter.shape,
    Some(0), // In some cases we will want to update parts of Shape
    Some(CodeLine::new(LineStyle::Multiline, ExprKind::Undetermined)),
);
formatter.with_shape(new_shape, |formatter| -> Result<(), FormatterError> {
```

Not important though, purely a style thing!

swayfmt/src/items/item_fn/mod.rs
```diff
@@ -26,15 +26,26 @@ impl Format for ItemFn {
...
-        } else {
-            write!(formatted_code, " {{}}")?;
-        }
+        formatter.with_shape(
+            Shape::from(
+                &formatter.shape,
+                Some(0), // In some cases we will want to update parts of Shape
+                Some(CodeLine::new(LineStyle::Normal, ExprKind::Function)),
+            ),
+            |formatter| -> Result<(), FormatterError> {
```
mitchmindtree, line 35: Same here ^

swayfmt/src/formatter/shape.rs
```diff
@@ -188,6 +194,18 @@ impl Shape {
     // |<---->|          indent
     //        |<--->|    width
     //
+    /// Create a copy of
+    pub(crate) fn from(shape: &Shape, width: Option<usize>, code_line: Option<CodeLine>) -> Self {
```
mitchmindtree, line 198: One alternative to using multiple `Option`s in a constructor's set of arguments (which can be a little tricky to read on the calling side if you don't know what the arguments happen to refer to) is to use the "builder method" pattern. E.g.

```rust
impl Shape {
    // ...

    pub(crate) fn width(mut self, w: usize) -> Self {
        self.update_width(w);
        self
    }

    pub(crate) fn line(mut self, code_line: CodeLinde) -> Self {
        self.update_line_settings(code_line);
        self
    }

    // ...
}
```

This way, rather than constructing a new Shape like so:

```rust
let new_shape = Shape::from(&formatter.shape, Some(new_width) Some(new_line));
```

user's could construct new `Shape` instances like this:

```rust
let new_shape = formatter.shape.clone().width(new_width).line(new_line);
```
This approach is particularly useful in the case that you don't want to update *all* of the options at once, e.g. here we only update the width:
```rust
let new_shape = formatter.shape.clone().width(new_width);
```
whereas the current approach would require specifying a redundant `None`, e.g.
```rust
let new_shape = Shape::new(&formatter.shape, Some(new_width), None);
```

You start to notice the benefits of the builder pattern more when you have many more optional parameters (e.g. imagine if we had 5 optional parameters but only wanted to specify `width`).

### Sidenotes

Some folks prefer the `with_` prefix for builder methods, e.g. `shape.with_width(w).with_line(l)`, but I personally don't mind either approach, as long as they're used consistently and don't cause ambiguity with any of the other methods. 

There are also other variations on the builder method where a dedicated `ShapeBuilder` type would be used with a `.build()` call chained on at the end to construct the actual `Shape`, however this approach is more useful when building much larger sized types that store a lot of data.

You can read more about the builder pattern [here](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html).

mitchmindtree (dismissed):
Nice work!

Just a few codestyle suggestions, but I'll leave it up to you if you want to address them. Some of the indentation is getting pretty extreme in some places which might be a good indicator that we could refactor some of this a little (e.g. moving some code out into sub functions), but functionally everything's looking good!

swayfmt/src/items/item_enum/mod.rs
eureka-cpu: One thing I plan to do after this PR is refactor `CodeLine` to have the `width` and `has_where_clause` fields since I think it actually belongs there instead of just on `Shape` itself. I think this will reduce the clutter and also remove the need to have the extra `Some(0)` 😄

swayfmt/src/formatter/shape.rs
eureka-cpu: Ah much better 😄 Thank you! I'll add these changes in!

eureka-cpu: I think I found a decent compromise to this, let me know if it needs any other changes!

swayfmt/src/formatter/shape.rs
```diff
@@ -103,26 +103,51 @@ impl Sub for Indent {
...
-        self.expr_kind = expr_kind
+    pub(crate) fn with_expr_kind(&mut self, expr_kind: ExprKind) {
+        self.expr_kind = expr_kind;
     }
-    pub(crate) fn new(line_style: LineStyle, expr_kind: ExprKind) -> Self {
-        Self {
-            line_style,
-            expr_kind,
+    /// Update the value of `has_where_clause`.
+    pub(crate) fn update_where_clause(&mut self) {
```
mitchmindtree, line 147: Maybe `toggle_where_clause` could be a clearer name for this?

That said, it might instead be worth passing the value explicitly here (like `update_width`) so that it's clear to the caller what the resulting state will be, as the implementer of `Format` doesn't necessarily know what state the value will be before calling `update_where_clause`.

swayfmt/src/formatter/shape.rs
```diff
@@ -103,26 +103,51 @@ impl Sub for Indent {
...
+    pub(crate) fn add_width(&mut self, extra_width: usize) {
+        self.width += extra_width;
+    }
+    pub(crate) fn sub_width(&mut self, extra_width: usize) {
+        self.width -= extra_width;
+    }
     /// Update `CodeLine::line_style` with a given LineStyle.
-    pub(crate) fn update_line_style(&mut self, line_style: LineStyle) {
-        self.line_style = line_style
+    pub(crate) fn with_line_style(&mut self, line_style: LineStyle) {
```
mitchmindtree, line 139: Perhaps these should remain `update_line_style` and `update_expr_kind` to remain consistent with `update_width` which also just assigns the given value?

In Rust, the `with` prefix is more commonly used with builder style methods (where you return a new `Self` **with** a new value) or for functions that will call a given function **with** some argument (like `Formatter::with_shape`).

swayfmt/src/formatter/shape.rs
```diff
@@ -103,26 +103,51 @@ impl Sub for Indent {
...
+        self.line_style = line_style;
     }
     /// Update `CodeLine::expr_kind` with a given ExprKind.
-    pub(crate) fn update_expr_kind(&mut self, expr_kind: ExprKind) {
-        self.expr_kind = expr_kind
+    pub(crate) fn with_expr_kind(&mut self, expr_kind: ExprKind) {
```
mitchmindtree, line 143: as above^

swayfmt/src/formatter/shape.rs
```diff
@@ -194,22 +214,9 @@ impl Shape {
...
-        }
-
-        new_shape
-    }
     /// A wrapper for `to_width_heuristics()` that also checks if the settings are default.
     pub(crate) fn apply_width_heuristics(&mut self, width_heuristics: WidthHeuristics) {
-        if width_heuristics == WidthHeuristics::default() {
-        } else {
+        if width_heuristics != WidthHeuristics::default() {
             self.width_heuristics = width_heuristics
```
mitchmindtree, line 220: Just noticed this - are we sure we want to deny users of this method from resetting this value back to the default state? Perhaps we can just remove the condition altogether and let the function body be `self.width_heuristics`? Let me know if I'm missing the point or use-case here!

Also, the method mentions it's a wrapper for `to_width_heuristics()`, but it appears there's no call to `to_width_heuristics()`.

swayfmt/src/items/item_enum/mod.rs
mitchmindtree: Nicce, just noticed you made these changes already - looking much better!

I think we can still reduce the indentation a little by first assigning the new shape to a variable like in my example above. E.g. with your new changes, this might look like:

```rust
        let new_shape = formatter
            .shape
            .with_code_line_from(LineStyle::Multiline, ExprKind::default());
        formatter.with_shape(new_shape, |formatter| -> Result<(), FormatterError> {
            // Code starts here.
```
This removes ~4 spaces of indentation over what we currently have e.g.
```rust
        formatter.with_shape(
            formatter
                .shape
                .with_code_line_from(LineStyle::Multiline, ExprKind::default()),
            |formatter| -> Result<(), FormatterError> {
                // Code starts here.
```
That said, this is a very minor nit - feel free to ignore this one if you prefer the current way!

swayfmt/src/formatter/shape.rs
```diff
@@ -252,68 +245,67 @@ impl Shape {
...
-    pub(crate) fn update_line_settings(&mut self, line_settings: CodeLine) {
-        self.code_line = line_settings
-    }
-    /// Reset `Shape::code_line` to default.
-    pub(crate) fn reset_line_settings(&mut self) {
-        self.update_line_settings(CodeLine::default())
+    /// Create a new `Shape` with a new `CodeLine` from a given `LineStyle` and `ExprKind`.
+    pub(crate) fn with_code_line_from(self, line_style: LineStyle, expr_kind: ExprKind) -> Self {
+        Self {
+            code_line: CodeLine::from(line_style, expr_kind),
```
mitchmindtree, line 300: Note that the returned `Shape`'s `code_line` here will not have the original `self.code_line`'s `width` or `has_where_clause` (both are set to `default()` in `CodeLine::from`). Is that definitely what we want here? Or should the returned `Shape` have the original `Shape`'s `code_line.width` and `code_line.has_where_clause`?

mitchmindtree (commented):
Nice, I'm liking the refactor of the `width` and `has_where_clause` fields to the `CodeLine` type. Also liking the new `Shape` builder methods - looks much nicer!

I've left just a few more small suggestions, otherwise looking great overall.

swayfmt/src/formatter/shape.rs
eureka-cpu: I was thinking that as well, but left it as is since it was working. I'll go ahead and add that change!

swayfmt/src/formatter/shape.rs
eureka-cpu: So the heuristic settings are what we gather from the `swayfmt.toml`, I don't believe we would want this to change during formatting. This function just takes in the settings that are generated from `to_width_heuristics`. If I recall correctly, I attempted to make it a direct application but this is the conclusion I came to at the time.

swayfmt/src/formatter/shape.rs
eureka-cpu: Hm, I added an option for them previously while I was building the first iteration of this new API, and it seemed that they were always `width: 0` and `has_where_clause: None`, and I think this is the right move. Each time we are addressing a new `CodeLine` we want to have a fresh start although sometimes we already know some information like `ExprKind` or in the case of say `if/else` we already know it's a `Conditional` so we give that information at the start. We never know the `width` before analyzing the `CodeLine` nor if it has a `where` clause so I think this is the safest route, at least for now.

eureka-cpu: I'm open ears for a better approach!

swayfmt/src/formatter/shape.rs
eureka-cpu: To be more clear, at the beginning of each `Formatter::format` call, we are reassigning the `heuristics` settings from a `config`. This is just the method that applies those settings to `Shape`.

swayfmt/src/items/item_enum/mod.rs
eureka-cpu: It's just my personal preference to avoid adding variables where we won't re-use or update them, but this isn't based on any logic, just how I like to see it. Is it better to have it behind a variable for any reason other than reducing indentation?

Braqzen (commented):
1. There are some imports that use `*`. This should be changed to be explicit.
2. Are these "tests" in the `examples` the only things that are expected to be fixed or are there additional changes that should have been fixed but might not be showing up in our projects?

swayfmt/src/items/item_enum/mod.rs
Braqzen: People have different monitor sizes and may set up multiple windows / tabs side by side. If things are heavily indented then they have to keep scrolling from side to side to see the content. If indentation can be cleaned up by assignment to variables and parsing around those vars then that's generally preferred.

eureka-cpu:
> 1. There are some imports that use `*`. This should be changed to be explicit.
> 
>     2. Are these "tests" in the `examples` the only things that are expected to be fixed or are there additional changes that should have been fixed but might not be showing up in our projects?

Are we not allowed to use the glob operator? 🤔 I personally think it looks cleaner
And the tests here changed to be what they were intended to be, we have unit tests that reflect the formatting we now see, we just weren't seeing it previously because of the width bleed. What you saw was bleed from file to file, but this was also happening on a smaller scale too.

Braqzen:
> > 1. There are some imports that use `*`. This should be changed to be explicit.
> >    
> >    2. Are these "tests" in the `examples` the only things that are expected to be fixed or are there additional changes that should have been fixed but might not be showing up in our projects?
> 
> Are we not allowed to use the glob operator? thinking I personally think it looks cleaner And the tests here changed to be what they were intended to be, we have unit tests that reflect the formatting we now see, we just weren't seeing it previously because of the width bleed. What you saw was bleed from file to file, but this was also happening on a smaller scale too.

Might as well obscure every import and use `*` for everything so that no one knows where anything is coming from. The glob operator should be removed from programming languages. If you import things explicitly then it's clear where things are coming from. `*` is tech debt.

eureka-cpu:
> > > 1. There are some imports that use `*`. This should be changed to be explicit.
> > >    
> > >    2. Are these "tests" in the `examples` the only things that are expected to be fixed or are there additional changes that should have been fixed but might not be showing up in our projects?
> > 
> > 
> > Are we not allowed to use the glob operator? thinking I personally think it looks cleaner And the tests here changed to be what they were intended to be, we have unit tests that reflect the formatting we now see, we just weren't seeing it previously because of the width bleed. What you saw was bleed from file to file, but this was also happening on a smaller scale too.
> 
> Might as well obscure every import and use `*` for everything so that no one knows where anything is coming from. The glob operator should be removed from programming languages. If you import things explicitly then it's clear where things are coming from. `*` is tech debt.

I understand the frustration, I only ever use the glob op when importing all of a crate's parts which I think is its actual intended use. I don't mind opening an issue to exchange all glob ops in the formatter for their explicit pieces though if this is an agreed upon best practice!

Braqzen (approved):

JoshuaBatty (approved):

PR #2743, opened 2022-09-07
Extend `simplify-cfg` to remove empty blocks

vaivaswatha:
Code size impact:

sway-application | master | this PR
-- | -- | --
dao-voting | 22024 | 21104
escrow | 73068 | 69132
fundraiser | 27616 | 26760
multisig-wallet | 8084 | 7860
NFT | 22520 | 21760

edit: I've updated this table after merging changes from `master`.


vaivaswatha:
>Then we should generate an ICE if this is violated or maybe I'm misunderstanding the comment.

I just wrote a mutable version of the function that's right above. 

>Any reason this isn't a HashSet<_>?

The set of predecessors is most often small. I could make it a `HashSet`, but my intuition (because we don't have data) is that it's better a vector.

Have made all the other changes you've suggested.

sway-ir/src/optimize/simplify_cfg.rs
```diff
@@ -42,7 +42,7 @@ fn unlink_empty_blocks(context: &mut Context, function: &Function) -> Result<boo
             match block.get_terminator(context) {
                 // Except for a PHI and a branch, we don't want anything else.
                 // num_instructions doesn't count PHI though.
-                Some(Instruction::Branch(to_block)) if block.num_instructions(context) > 1 => {
+                Some(Instruction::Branch(to_block)) if block.num_instructions(context) <= 1 => {
```
otrho, line 45: I was literally looking at this line thinking 'hang on..?' when you committed this. :slightly_smiling_face:

sway-ir/src/block.rs
```diff
@@ -176,6 +190,49 @@ impl Block {
         }
     }
 
+    /// Replace successor `old_succ` with `new_succ`
+    pub(super) fn replace_successors(
```
otrho, line 194: I think this should be `replace_successor()`.  The plural is a little confusing.

sway-ir/src/optimize/simplify_cfg.rs
```diff
@@ -27,6 +27,69 @@ pub fn simplify_cfg(context: &mut Context, function: &Function) -> Result<bool,
...
+                    // If the value coming to `to_phi` is `block_phi`, we replace it
+                    // with all the incoming values to `block_phi` itself.
+                    to_phi_pairs.append(&mut cur_phi_pairs);
+                } else {
+                    // Otherwise, it gets `v` from every `pred`.
+                    let v_pred_pairs = preds.iter().map(|b| (*b, v));
+                    to_phi_pairs.extend(v_pred_pairs);
+                }
+                modified = true;
+            } // We don't need to bother if there is no value coming in from block.
```
otrho, line 86: I think this is over complicating the process by trying to update the pairs in place.  Most of the time a block will have at most a handful of predecessors so trying to be overly efficient isn't worth it.

Instead a new PHI could be created based on the old ones.  It should be OK to just delete any reference the the empty block PHI from the successor PHIs since we'll inject the predecessor pairs into the PHI anyway, which is all the empty block PHI would've contained.  So it's two steps - remove the empty block PHI value and then merge the predecessor PHIs with the successor PHIs.  They're essentially sets of pairs.

All the PHI manipulation should be managed in `block.rs` too, rather than poking around directly in the optimisation passes, if possible.  How PHIs working internally shouldn't be exposed.  So a `Block::merge_into_phi(other_phi: Value)` method or something.

It might be worth actually making them `BTreeSet`s instead of `Vec`.

otrho:
> > Any reason this isn't a HashSet<_>?
> 
> The set of predecessors is most often small. I could make it a `HashSet`, but my intuition (because we don't have data) is that it's better a vector.

As I've mentioned above I actually think we should make these `BTreeSet`s now that we need to treat them as such -- adding, removing elements and merging.  I'd avoid `HashSet` purely because we iterate over them which is non-deterministic, though 99% of the time would be OK.  But that 1% might bite us.

sway-ir/src/optimize/simplify_cfg.rs
vaivaswatha: >merge the predecessor PHIs with the successor PHIs. 

This is not what I'm doing in both the cases. I'm separating the following two scenarios:

```
block0:
v1 = const 1
br block1

block1:
x = phi(block0: v1)
```
and

```
block0:
v1 = phi(...)
br block1

block1:
x = phi(block0: v1, ...)
```

and also note that in the latter case, I actually remove the pair `(block0, v1)` from `block1`'s PHI before doing the merge.

So I can't see how to best do this in a single function call to a new method for `Block`.

vaivaswatha:
Regarding the data structure, considering that we're dealing with a small set, isn't vector the best one to have?

sway-ir/src/optimize/simplify_cfg.rs
otrho: No, it's two steps.  Remove `v1` altogether from the successor PHIs because we're merging those in in the second step.  Two methods -- `remove_pred_from_phi()` and `merge_phi()`.  But that's only if the value coming from the empty block is its PHI.

That first example is tricky though.  It'll be empty because `const` isn't an instruction.  OK, I didn't think of that.   So you create a new PHI which is all the pred PHIs except with the values replaced with the `const` and then merge that into the succ PHIs.

Hrmm, let's see:
1. Remove `block0` and `v1` from successor PHIs.
2. Let `new_phi` = `v1` if `v1` is a PHI, else `new_phi` is a new PHI with all the pred blocks but `v1` as the values.
3. Merge `new_phi` with successor PHIs.

Does that work?  No editing in place and all of the gnarly stuff in `block.rs`.

otrho:
> Regarding the data structure, considering that we're dealing with a small set, isn't vector the best one to have?

`HashSet` and `BTreeSet` have set-style methods for adding without duplicating, and removing and finding the union, which is what we want here.

sway-ir/src/optimize/simplify_cfg.rs
otrho: When I say merge, I mean something like [BTreeSet::append()](https://doc.rust-lang.org/std/collections/struct.BTreeSet.html#method.append) which will handle duplicates.  If the check for conflicts remains then it should work.

Or maybe a new set with `BTreeSet::from_iter(a.union(b))`.  Whichever works.

otrho: Or even... don't merge, just call `Block::add_phi()` to the successors with the old pairs, I guess.  If we've guaranteed no conflicts then just adding the new ones like you've done with `to_phi_pairs.extend(v_pred_pairs)` above shouldn't create dupes.

I'd just mostly like to see the direct pair manipulation kept in `block.rs`.

vaivaswatha:
C

> > Regarding the data structure, considering that we're dealing with a small set, isn't vector the best one to have?
> 
> `HashSet` and `BTreeSet` have set-style methods for adding without duplicating, and removing and finding the union, which is what we want here.

So should we do this change as part of this PR?

vaivaswatha:
> C
> 
> > > Regarding the data structure, considering that we're dealing with a small set, isn't vector the best one to have?
> > 
> > 
> > `HashSet` and `BTreeSet` have set-style methods for adding without duplicating, and removing and finding the union, which is what we want here.
> 
> So should we do this change as part of this PR?

If I understood it right, you mean to change the definition of PHI itself to hold a BTreeSet rather than a Vec right? (not just here in my function).

otrho:
> If I understood it right, you mean to change the definition of PHI itself to hold a BTreeSet rather than a Vec right? (not just here in my function).

Yeah.  Maybe.   :slightly_smiling_face: Hrrm, it fits quite well except for the invariant where we don't want to duplicate pred blocks in the pairs.  In which case set-merging isn't what we want.  Umm, how about we put that on the Maybe Pile for now?

vaivaswatha:
5

> > If I understood it right, you mean to change the definition of PHI itself to hold a BTreeSet rather than a Vec right? (not just here in my function).
> 
> Yeah. Maybe. slightly_smiling_face Hrrm, it fits quite well except for the invariant where we don't want to duplicate pred blocks in the pairs. In which case set-merging isn't what we want. Umm, how about we put that on the Maybe Later Pile for now?

I can create an Issue for that (if we should). So where does that leave this PR?

otrho:
> I can create an Issue for that (if we should). So where does that leave this PR?

OK.  So simplify the pass to just gather the candidates and then use `Block` methods to update the PHIs rather than poking around in there directly.

otrho (dismissed):
Thanks @vaivaswatha, this is much more readable IMO.

mohammadfawaz:
This is great! Thanks @vaivaswatha 

My only concerns with this is the lack of unit tests. While reviewing this by inspection is valuable, having unit tests showing exactly what's happening and preventing regressions is even more valuable. What do you think?

vaivaswatha:
> This is great! Thanks @vaivaswatha
> 
> My only concerns with this is the lack of unit tests. While reviewing this by inspection is valuable, having unit tests showing exactly what's happening and preventing regressions is even more valuable. What do you think?

I fully agree. I'll add them and update the PR.

vaivaswatha:
> This is great! Thanks @vaivaswatha
> 
> My only concerns with this is the lack of unit tests. While reviewing this by inspection is valuable, having unit tests showing exactly what's happening and preventing regressions is even more valuable. What do you think?

I've added this now. Please re-review.

mohammadfawaz (approved):

sway-ir/src/optimize/simplify_cfg.rs
```diff
@@ -27,6 +27,63 @@ pub fn simplify_cfg(context: &mut Context, function: &Function) -> Result<bool,
...
+            match block.get_terminator(context) {
+                // Except for a PHI and a branch, we don't want anything else.
+                // num_instructions doesn't count PHI though.
+                Some(Instruction::Branch(to_block)) if block.num_instructions(context) <= 1 => {
+                    Some((block, *to_block))
+                }
+                _ => None,
+            }
+        })
+        .collect();
```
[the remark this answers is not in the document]
otrho: There is very likely a borrowing conflict on `context` if the collect is removed...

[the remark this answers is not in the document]
vaivaswatha: Yep.

otrho (approved):

PR #2744, opened 2022-09-07
Disable the generation of logged types in the JSON ABI by default

mohammadfawaz:
- Disable the generation of logged types in the JSON ABI by default
- The flag `--generate-logged-types` can be used to generate those types until the full support is done in the SDK.

https://github.com/FuelLabs/sway/issues/2745 tracks reverting this change once the SDK team is finished with implementing generics and even logging.


Braqzen (commented):
The current version has code which tests the `loggedTypes` (at least it includes it in the `.json`) but now a flag is being added which removes tests. Shouldn't there be 1 set of tests when the flag is used and another when the flag is not used? This way it would default to not being used but it is still proven to work when the flag is provided.

mohammadfawaz:
> The current version has code which tests the `loggedTypes` (at least it includes it in the `.json`) but now a flag is being added which removes tests. Shouldn't there be 1 set of tests when the flag is used and another when the flag is not used? This way it would default to not being used but it is still proven to work when the flag is provided.

Absolutely. I don't like that I had to update the tests the way I did. To fix this, I need to update the infrastructure of the E2E tests to allow passing in `forc` flags through `test.toml` which I skipped here due to urgency. I will create an issue do that and work on it as soon as this is done and released so that nothing breaks without us knowing about it.

Braqzen (approved):
Tentatively approving. A compiler dev should really double check.

emilyaherbert (approved):

Issue #2733, opened 2022-09-06
Add misc parser recovery in appropriate places


Centril:
For example, we might want to recover `class` as a typo for `struct`.

Adding more details here or in the PR once they become clearer.

PR #2751, opened 2022-09-07
Recover `class Foo { .. }` as `struct`

Centril:
Recover `class Foo { ... }` as if `struct Foo { ... }` had been written.

Work towards https://github.com/FuelLabs/sway/issues/2733.


sway-parse/src/error.rs
```diff
@@ -74,6 +74,8 @@ pub enum ParseErrorKind {
     UnnecessaryVisibilityQualifier { visibility: Ident },
     #[error("Expected a doc comment.")]
     ExpectedDocComment,
+    #[error("Use `struct`s to define records, instead of `class`.")]
```
sezna, line 77: Hmm, what do you think about 

```suggestion
    #[error("Use the `struct` keyword to define records, instead of `class`.")]
```
?

otrho (approved):
Is this sort of thing a priority?  Are people complaining about not knowing how to make structs?

mohammadfawaz (approved):

Issue #2731, opened 2022-09-06
Add `TIME` instruction support


adlerjohn:
https://github.com/FuelLabs/fuel-asm/pull/89

mohammadfawaz:
I'm assuming this simply involves: 
- Adding `time` to opcode to the parser
- Adding `time` in codegen.
- Adding a standard library function in [`block.sw`](https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/block.sw) called `timestamp()` that wraps the `time` instruction. Propsal:
  - `fn timestamp()`: returns the timestamp for the current block, i.e., it calls `height()` under the hood.
  - `fn timestamp_of_block(height: u64)`: returns the timestamp for block at height `height`.

PR #2752, opened 2022-09-08
Add support for the `TIME` instruction

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/2731

### DIFF
--- a/test/src/sdk-harness/test_projects/block/mod.rs
+++ b/test/src/sdk-harness/test_projects/block/mod.rs
@@ -1,0 +1,109 @@
+use fuels::prelude::*;
+use std::time::{SystemTime, UNIX_EPOCH};
+use tokio::time::{sleep, Duration};
+
+abigen!(
+    BlockTestContract,
+    "test_projects/block/out/debug/block-abi.json"
+);
+
+async fn get_block_instance() -> (BlockTestContract, ContractId) {
+    let wallet = launch_provider_and_get_wallet().await;
+    let id = Contract::deploy(
+        "test_projects/block/out/debug/block.bin",
+        &wallet,
+        TxParameters::default(),
+        StorageConfiguration::with_storage_path(Some(
+            "test_projects/block/out/debug/block-storage_slots.json".to_string(),
+        )),
+    )
+    .await
+    .unwrap();
+    let instance = BlockTestContractBuilder::new(id.to_string(), wallet).build();
+
+    (instance, id.into())
+}
+
+#[tokio::test]
+async fn can_get_block_height() {
+    let (instance, _id) = get_block_instance().await;
+    let block_0 = instance.get_block_height().call().await.unwrap();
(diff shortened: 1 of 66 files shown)


test/src/sdk-harness/test_projects/block/src/main.sw
```diff
@@ -0,0 +1,22 @@
+contract;
+
+use std::block::*;
+use block_test_abi::*;
```
Braqzen, line 4: ew

Braqzen (dismissed):

test/src/sdk-harness/test_projects/block/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+
+#[tokio::test]
+async fn can_get_block_height() {
+    let (instance, _id) = get_block_instance().await;
+    let block_0 = instance.get_block_height().call().await.unwrap();
+    let block_1 = instance.get_block_height().call().await.unwrap();
+    let block_2 = instance.get_block_height().call().await.unwrap();
+
+    // Consecutive blocks
+    assert_eq!(block_1.value, block_0.value + 1);
```
nfurfaro, line 35: This seems brittle; it will break when there is more than 1 tx per block right?
That said, it's probably good enough for now and can be updated when it does break.

test/src/sdk-harness/test_projects/block/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    // The difference should be 1 second in most cases, but be conservative to guarantee stability
+    // of the test
+    assert!(block_1_time.value - block_0_time.value <= 2);
+
+    // Wait 2 seconds and request another block
+    sleep(Duration::from_secs(2)).await;
+    let block_2_time = instance.get_timestamp().call().await.unwrap();
+
+    // The difference should be 2 seconds in most cases, but be conservative to guarantee stability
```
nfurfaro, line 63: For these conservative tests, would it make sense to specify a range instead? i.e: 
`    assert!(2 <= block_2_time.value - block_1_time.value <= 3);`

nfurfaro (commented):
A couple of minor comments, but looks good!

test/src/sdk-harness/test_projects/block/mod.rs
mohammadfawaz: Good idea. Updated the test.

test/src/sdk-harness/test_projects/block/mod.rs
mohammadfawaz: Yes you're right. I changed the checks to be a bit more flexible and more future proof.

Braqzen (approved):
Note that testing a range is a bad idea because if it fails then the assertion would need to be split up to be debugged. If the range is for a single value in order to not use 2 assertions then that is fine. Debugging boundary conditions is simpler when they are split into multiple assertions.

nfurfaro (approved):

Issue #2719, opened 2022-09-03
Internal compiler error: Store must be to a pointer


Braqzen:
Two different signatures, one with `ref mut`, causes an ICE.

```rust
contract;

abi MyContract {
    fn test_function(variable: u64) -> bool;
}

impl MyContract for Contract {
    fn test_function(ref mut variable: u64) -> bool {
        variable += 1;
        true
    }
}
```

![1](https://user-images.githubusercontent.com/103777923/188289229-62103589-ab8e-473c-ab89-f44b1ecb853e.png)

otrho:
Note: the fix for this issue is to return a better non-ICE error, rather than to actually support this behaviour.

PR #2754, opened 2022-09-08
Improve parameter signature checking for trait functions.

tritao:
This commit adds two new parameter signature checks when type checking trait functions.

1) parameter mutability mismatches between trait declaration and its implementation
2) ref parameters inside contract abi functions

Given the following code:

```rust
contract;

abi MyContract {
    fn test_function(variable: u64) -> bool;
}

    fn test_function(ref mut variable: u64) -> bool {
        true
    }
}

```

We now emit:

```rust
error
  --> main.sw:8:22
   |
 6 |
 7 | impl MyContract for Contract {
 8 |     fn test_function(ref mut variable: u64) -> bool {
|                      ^^^^^^^ Parameter mutability mismatch between
the trait function declaration and its implementation.
 9 |         true
10 |     }
   |
____

error
  --> main.sw:8:22
   |
 6 |
 7 | impl MyContract for Contract {
 8 |     fn test_function(ref mut variable: u64) -> bool {
|                      ^^^^^^^ Ref mutable parameter is not supported
for contract ABI function.
 9 |         true
10 |     }
   |
____
```

Closes https://github.com/FuelLabs/sway/issues/2719.


mohammadfawaz (approved):
Nice. Thanks!

otrho (approved):

Issue #2759, opened 2022-09-10
Incorrect JSON ABI generated for a struct containing a tuple


mohammadfawaz:
Example:

```rust
contract;

struct AGenericStruct<K, T> {
	a: (K, T)
}

abi MyContract {
    fn test_function(arg: AGenericStruct<u32, u32>);
}
```

Generated JSON ABI
```json
    {
      "typeId": 1,
      "type": "(_, _)",
      "components": [
        {
          "name": "__tuple_element",
          "type": 2,
          "typeArguments": null
        },
        {
          "name": "__tuple_element",
          "type": 5,
          "typeArguments": null
        }
      ],
      "typeParameters": null
    },
    {
      "typeId": 2,
      "type": "generic K",
      "components": null,
      "typeParameters": null
    },
    {
      "typeId": 3,
      "type": "generic T",
      "components": null,
      "typeParameters": null
    },
    {
      "typeId": 4,
      "type": "struct AGenericStruct",
      "components": [
        {
          "name": "a",
          "type": 1,
          "typeArguments": null
        }
      ],
      "typeParameters": [
        2,
        3
      ]
    },
    {
      "typeId": 5,
      "type": "u32",
      "components": null,
      "typeParameters": null
    }
  ]
```
The type of the second element of the tuple should not be `u32` - it should just be `T`. Somehow the first tuple element behaves as expected but the second does not 🤔

PR #2760, opened 2022-09-10
Fix JSON ABI for generic tuples in structs

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/2759

Simple change that was missed from the original change that introduced then new JSON ABI format.


otrho (approved):
utACK

Issue #841, opened 2022-02-25
Create an implicit Sway `prelude`


emilyaherbert:
Per the discussion on #830, we will eventually want to have a Sway prelude akin to the Rust prelude (https://doc.rust-lang.org/std/prelude/index.html).

adlerjohn:
Related: #330

mitchmindtree:
Adding some quick thoughts on this as it was raised again recently in #2433:

The simplest approach might be to add prelude names while constructing the initial `Namespace`. The `Namespace` contains an `init` module that is `clone`d when initialising each module within a Sway project. It contains the names that should be implicitly available from every module. Perhaps we can add the prelude names to this `init` module upon constructing the `Namespace` with the condition that the `std` library must be a dependency.

sezna:
Hm, I think a lot of the above can remain scoped. Address, ContractId, and Identity, Vec, assert, require, and revert should probably be included. But `msg_sender`, `transfer`, `msg_amount`, `msg_asset_id`, and `StorageMap` can remain scoped IMO. Following Rust's prelude design, there aren't many global functions and all data structures except `Vec` are in `std::collections`

SwayStar123:
> Hm, I think a lot of the above can remain scoped. Address, ContractId, and Identity, Vec, assert, require, and revert should probably be included. But `msg_sender`, `transfer`, `msg_amount`, `msg_asset_id`, and `StorageMap` can remain scoped IMO. Following Rust's prelude design, there aren't many global functions and all data structures except `Vec` are in `std::collections`

In this regard it doesnt make sense to follow Rust, but rather Solidity's design, right?

sezna:
That's a good question. This is really subjective. Personally, I dislike magical global functions in general, although data structures are less offensive to me. That's purely subjective though, so let's go ask in the community discord what people think.

mohammadfawaz:
> Personally, I dislike magical global functions in general

Agree 100%. If we implement the following proposal https://github.com/FuelLabs/sway/issues/2463, then a lot of these functions become methods and they will be automatically available by only importing `Contract` (which itself might be added to the prelude already). We could even consider adding more similar types such as `Block` which would expose `timestamp()` and `height()` methods (instead of the free functions in https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/block.sw).

xyx202:
Very good

PR #2769, opened 2022-09-12
Add standard library prelude

Centril:
Closes https://github.com/FuelLabs/sway/issues/841.

This adds the standard library prelude with some initial suggested items as per https://github.com/FuelLabs/sway/issues/841#issuecomment-1242594164.

The approach is to have the module `std::prelude` which is then "glob re-exported" by the compiler into `init` in every module. Extending the prelude is then done in Sway and not in Rust.

### DIFF
--- a/sway-core/src/semantic_analysis/namespace/module.rs
+++ b/sway-core/src/semantic_analysis/namespace/module.rs
@@ -213,6 +213,7 @@ impl Module {
             warnings,
             errors
         );
+
         let implemented_traits = src_ns.implemented_traits.clone();
         let mut symbols = vec![];
         for (symbol, decl) in src_ns.symbols.iter() {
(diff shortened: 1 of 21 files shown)


sway-core/src/semantic_analysis/namespace/module.rs
```diff
@@ -230,13 +231,65 @@ impl Module {
...
+
+        for symbol in symbols {
+            try_add(symbol, src.to_vec());
+        }
+        for (symbol, (mod_path, _)) in use_synonyms {
+            // N.B. We had a path like `::bar::baz`, which makes the module `bar` "crate-relative".
+            // Given that `bar`'s "crate" is `foo`, we'll need `foo::bar::baz` outside of it.
+            //
+            // FIXME(Centril): Seems like the compiler has no way of
+            // distinguishing between external and crate-relative paths?
```
mitchmindtree, line 287: Perhaps worth opening an issue and linking it at this FIXME?

test/src/e2e_vm_tests/test_programs/should_fail/shadow_import/src/main.sw
```diff
@@ -8,7 +8,7 @@ use ::bar::{Bar1, Bar1 as Foo};
 // This is not okay because symbol `Foo` is already reserved
 use ::bar::Bar2 as Foo;
 
-// This is not okay 
+// This is not okay
 use ::bar::{Bar2, Bar2};
 
 // This not okay now because Bar1 and Bar2 have already been imported
```
mitchmindtree, line 14: I see you've already removed the checks from the `test.toml`, but perhaps we should remove this line (and the one under it) too? Otherwise it could be a bit confusing to come across for future maintainers.

Great to see we can now shadow glob imports :partying_face:

mitchmindtree (commented):
Nice one, looks good! Just a couple tiny nits

test/src/e2e_vm_tests/test_programs/should_pass/language/prelude_access/src/main.sw
```diff
@@ -0,0 +1,11 @@
+script;
+
+struct A {
+    addr: Address,
+}
+
+// Ensure shadowing is allowed.
+use std::address::Address;
```
mohammadfawaz, line 8: Do you mind adding another test to check that every symbol in the prelude is actually available without the need to import it?

mohammadfawaz (approved):

otrho (approved):

mohammadfawaz:
I'm merging this because I'm very excited about it 😄

PR #2771, opened 2022-09-12
Add intrinsics for `add`, `sub`, `mul` and `div`

### LINKED ISSUE #2758: Introduce intrinsics (and consequently IR instructions) for basic arithmetic operations
Following up on #2755 and #855, introduce `__add`, `__sub`, `__mul` and `__div` opcodes, represent them via a new IR opcode `BinaryOp` (with an internal enum on what the operation specifically is) and generate code for these new opcodes.

As a followup, modify the library functions for these operations to use these intrinsics.

vaivaswatha:
Consequently, modify `stdlib` to use the new arithmetic intrinsics.

Closes #2758

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
+++ b/sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
@@ -550,7 +550,7 @@ impl TypedIntrinsicFunctionKind {
                     });
                     return err(warnings, errors);
                 }
-                if type_arguments.len() != 0 {
+                if !type_arguments.is_empty() {
                     errors.push(CompileError::IntrinsicIncorrectNumTArgs {
                         name: kind.to_string(),
                         expected: 0,
--- a/sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
+++ b/sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
@@ -550,6 +550,15 @@ impl TypedIntrinsicFunctionKind {
                     });
                     return err(warnings, errors);
                 }
+                if type_arguments.len() != 0 {
+                    errors.push(CompileError::IntrinsicIncorrectNumTArgs {
+                        name: kind.to_string(),
+                        expected: 0,
+                        span,
+                    });
+                    return err(warnings, errors);
+                }
+
                 let mut ctx = ctx
                     .by_ref()
                     .with_type_annotation(insert_type(TypeInfo::Unknown));
--- a/sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
+++ b/sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
@@ -541,6 +541,60 @@ impl TypedIntrinsicFunctionKind {
                 let return_type = insert_type(TypeInfo::Tuple(vec![]));
                 (intrinsic_function, return_type)
             }
+            Intrinsic::Add | Intrinsic::Sub | Intrinsic::Mul | Intrinsic::Div => {
+                if arguments.len() != 2 {
+                    errors.push(CompileError::IntrinsicIncorrectNumArgs {
+                        name: kind.to_string(),
+                        expected: 2,
+                        span,
+                    });
+                    return err(warnings, errors);
+                }
+                let mut ctx = ctx
+                    .by_ref()
+                    .with_type_annotation(insert_type(TypeInfo::Unknown));
+
+                let lhs = arguments[0].clone();
+                let lhs = check!(
+                    TypedExpression::type_check(ctx.by_ref(), lhs),
+                    return err(warnings, errors),
+                    warnings,
+                    errors
+                );
+
+                // Check for supported argument types
+                let arg_ty = resolve_type(lhs.return_type, &lhs.span).unwrap();
+                let is_valid_arg_ty = matches!(arg_ty, TypeInfo::UnsignedInteger(_));
+                if !is_valid_arg_ty {
+                    errors.push(CompileError::IntrinsicUnsupportedArgType {
+                        name: kind.to_string(),
+                        span: lhs.span,
+                        hint: Hint::empty(),
+                    });
+                    return err(warnings, errors);
+                }
+
+                let rhs = arguments[1].clone();
+                let ctx = ctx
+                    .by_ref()
+                    .with_help_text("Incorrect argument type")
+                    .with_type_annotation(lhs.return_type);
+                let rhs = check!(
+                    TypedExpression::type_check(ctx, rhs),
+                    return err(warnings, errors),
+                    warnings,
+                    errors
+                );
+                (
+                    TypedIntrinsicFunctionKind {
+                        kind,
+                        arguments: vec![lhs, rhs],
+                        type_arguments: vec![],
+                        span,
+                    },
+                    insert_type(arg_ty),
+                )
+            }
         };
         ok((intrinsic_function, return_type), warnings, errors)
     }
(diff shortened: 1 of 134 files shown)


sway-ir/src/instruction.rs
```diff
@@ -181,6 +195,7 @@ impl Instruction {
         match self {
             Instruction::AddrOf(_) => Some(Type::Uint(64)),
             Instruction::AsmBlock(asm_block, _) => Some(asm_block.get_type(context)),
+            Instruction::BinaryOp { arg1, .. } => arg1.get_type(context),
```
otrho, line 198: We might want to add a return type to the `BinaryOp` struct for the future, as the assumption that both args have the same return type as the expression won't necessarily be true always.  But not in this PR...

sway-ir/src/parser.rs
```diff
@@ -174,6 +178,26 @@ mod ir_builder {
...
+            rule op_binary_mul() -> IrAstOperation
+                = "mul" _ arg1:id() comma() arg2:id() {
+                    IrAstOperation::BinaryOp(BinaryOpKind::Mul, arg1, arg2)
+                }
+
+            rule op_binary_div() -> IrAstOperation
+                = "div" _ arg1:id() comma() arg2:id() {
+                    IrAstOperation::BinaryOp(BinaryOpKind::Div, arg1, arg2)
+                }
+
```
otrho, line 200: This is a bit redundant.  Instead have a rule for `op_binary_kind` returning `BinaryOpKind` which matches the `"add"` / `"sub"` / etc. and then a single `op_binary` rule which calls it.

sway-lib-core/src/ops.sw
```diff
@@ -6,37 +6,25 @@ pub trait Add {
 
 impl Add for u64 {
     fn add(self, other: Self) -> Self {
-        asm(r1: self, r2: other, r3) {
-            add r3 r2 r1;
-            r3: u64
-        }
+        __add(self, other)
```
otrho, line 9: Ahh, so nice. 😀

otrho (commented):
Can you add a quick `should_fail` test with mismatched integer types, and maybe some ref-types and stuff?

vaivaswatha:
>Can you add a quick should_fail test with mismatched integer types, and maybe some ref-types and stuff?

Right now, `__add(a: u64, b: u32)` works automatically and `__add(a: u32, b: u64)` works with a warning that `b` might lose precision. So it looks like our default behaviour is to convert between integer types implicitly. 

@emilyaherbert @mohammadfawaz @sezna I spoke with @otrho about this offline, and he suggested to ask you on the best course of action here. Should we retain the current behaviour or should I modify the PR to be stricter and disallow different integer type arguments?

@otrho I've done all the other changes you suggested.

sway-ir/src/parser.rs
```diff
@@ -121,15 +121,18 @@ mod ir_builder {
                     mdi
                 }
 
+            rule binary_op_kind() -> BinaryOpKind 
+                = "add" { BinaryOpKind::Add }
+                / "sub" { BinaryOpKind::Sub }
+                / "mul" { BinaryOpKind::Mul }
+                / "div" { BinaryOpKind::Div }
```
otrho, line 128: Thanks, but another small nit -- by convention throughout this grammar the `_` is always placed right next to the literal strings.  I know this breaks my other complaint about reducing redundancy, but to keep it consistent could you put the `_` after each of these strings and take it out of `op_binary()` please?

otrho:
> Should we retain the current behaviour or should I modify the PR to be stricter and disallow different integer type arguments?

And my (and I think @vaivaswatha's) feeling is we _should_ be stricter and only allow exact matches for the args and return type.  But I'm interested in further consensus.

otrho (approved):
Approving in principle, even though I think we need to tighten the type checks. Thanks!

mohammadfawaz:
> @emilyaherbert @mohammadfawaz @sezna I spoke with @otrho about this offline, and he suggested to ask you on the best course of action here. Should we retain the current behaviour or should I modify the PR to be stricter and disallow different integer type arguments?

I personally prefer stricter types with an escape hatch (i.e. explicit conversion), just like Rust, across our whole type system. I'm not really a big fan of the implicit conversion that we do for two reasons:
1. Implicit conversions allow the user to write less expressive code which goes against the philosophy of Sway.
2. Implicit conversions can generate a warning if there's a possibility of precision loss but there is now way to get rid of this warning today (please correct me if I'm wrong!) even if the user knows that precision loss is never going to happen in their particular use case (in Rust they would've used `as`).

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
```diff
@@ -541,6 +541,60 @@ impl TypedIntrinsicFunctionKind {
                 let return_type = insert_type(TypeInfo::Tuple(vec![]));
                 (intrinsic_function, return_type)
             }
+            Intrinsic::Add | Intrinsic::Sub | Intrinsic::Mul | Intrinsic::Div => {
```
mohammadfawaz, line 544: I noticed this in other places in this file as well but it is possible to pass in type arguments to these intrinsics and the compiler won't complain. Is that correct? Maybe we should add a check here that the number of type arguments is exactly zero.

vaivaswatha:
> > @emilyaherbert @mohammadfawaz @sezna I spoke with @otrho about this offline, and he suggested to ask you on the best course of action here. Should we retain the current behaviour or should I modify the PR to be stricter and disallow different integer type arguments?
> 
> I personally prefer stricter types with an escape hatch (i.e. explicit conversion), just like Rust, across our whole type system. I'm not really a big fan of the implicit conversion that we do for two reasons:
> 
> 1. Implicit conversions allow the user to write less expressive code which goes against the philosophy of Sway.
> 2. Implicit conversions can generate a warning if there's a possibility of precision loss but there is now way to get rid of this warning today (please correct me if I'm wrong!) even if the user knows that precision loss is never going to happen in their particular use case (in Rust they would've used `as`).

So my question now is, should this PR make it strict (and differ from the existing behaviour otherwise) or should we do it at one go everywhere?

mohammadfawaz:
> 

I say we move forward with this to make it non-breaking and document the discussion above in a new issue which we can address separately. There may be other opinions here as to why we deviated from Rust in the first place (is it the lack of the `as` keyword?)

mohammadfawaz (approved):
:shipit:

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
vaivaswatha: > I noticed this in other places in this file as well but it is possible to pass in type arguments to these intrinsics and the compiler won't complain. Is that correct? Maybe we should add a check here that the number of type arguments is exactly zero.

I've added this now.

mohammadfawaz (dismissed):

otrho (approved):

mohammadfawaz (approved):

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
vaivaswatha: > Long term, I think we should let the type checker proper handle this by exposing the signature of this intrinsic in the `core` library itself, e.g.,
> 
> ```rust
> #[intrinsic]
> fn my_special_function(...);
> ```
> 
> Note the `;` here -- the intrinsic is assigned no body. The type checker will see that this is `#[intrinsic]` and therefore allow the body-less function. However, it will collect the signature of the function and any calls to it will consult that signature. The only special code in the compiler is then recognizing to the standard library item and assigning it dynamic semantics (fairly deep in the compiler, in IR generation, with type substitutions etc. already known and with the sound assumption that there are no inference variables left).

I like this suggestion. Can you also post this in the RFC for intrinsics? I (and @otrho) had suggested a table based approach but this could be better.

PR #2772, opened 2022-09-13
feat: display long versions for forc crates

bingcicle:
Related #2773 

Includes changes to show long versions for `forc` crates in the format `<name> <version> (<git_commit_hash> <date>)`

Certain commands of `fuelup` (eg. `show`, `check`) to support nightly toolchains require comparing dates. 

Currently a hacky way of doing this is implemented within the CI for publishing channels where we manually format the strings ([example of what is published](https://github.com/FuelLabs/fuelup/blob/gh-pages/channel-fuel-nightly.toml)). Downloading a toolchain will download the channel as well, which `fuelup` uses to compare versions. However, the actual version printed by the binaries do not change and makes it difficult to support the above commands, especially if we want to support a specific nightly version of a component in a custom toolchain.

~For this to work, we have to build with the options `CI_COMMIT_HASH` and `CI_COMMIT_DATE` in the release steps in the CI. Otherwise, the current format of versioning will be used.~

Instead of the above, we will be using [`vergen`](https://github.com/rustyhorde/vergen) to get the hash and date at build-time with `build.rs`.

Example:

Screenshot 2022-09-13 at 11 25 32 AM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1419,12 +1419,14 @@ dependencies = [
  "fs_extra",
  "fuel-asm",
  "hex",
+ "once_cell",
  "serde",
  "serde_json",
  "sway-core",
  "sway-types",
  "sway-utils",
  "term-table",
+ "time 0.3.14",
  "tokio",
  "toml",
  "toml_edit",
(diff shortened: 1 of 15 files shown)


bingcicle:
after a short discussion with @mitchmindtree we decided to only include this information within the nightlies by setting the date/hash in Cargo.toml within the CI, using cargo set-version command.

PR #2774, opened 2022-09-13
`sway-lsp`: Collect `storage` token in `sway_ast::Attribute` 

JoshuaBatty:
This PR collects the `storage` attribute into the `TokenMap` type used in the language server. In order to do so, we are now looping through the items in `sway_parse` so we can access tokens in `sway_ast::Attribute`. The storage token in an attribute is being assigned a kind of `SymbolKind::BuiltinAttr` which matches the semantic definition of what rust-analyzer is doing. 

Also removes an old unused `test_typed_parse` function. 

@SilentCicero this (should) solve the highlighting issue with the storage token.


sway-lsp/src/core/session.rs
```diff
@@ -158,17 +158,29 @@ impl Session {
...
         if let Ok(manifest) = pkg::ManifestFile::from_dir(&manifest_dir) {
             if let Ok(plan) = pkg::BuildPlan::from_lock_and_manifest(&manifest, locked, offline) {
+                // First, parse items to collect keyword tokens
+                if let Ok(source) = manifest.entry_string() {
+                    if let Ok(sway_build_config) =
+                        pkg::sway_build_config(manifest.dir(), &manifest.entry_path(), &profile)
+                    {
+                        parse_items::parse(source, Some(&sway_build_config), &self.token_map);
+                    }
+                }
```
mitchmindtree, line 176: Is there no way we can collect these `storage` tokens during our regular traversal of items, rather than needing a whole distinct parse step? We should avoid parsing or type-checking more than we need to, especially considering how frequently LSP will do both.

JoshuaBatty: I agree we should limit the amount of parsing we need to do. One option is to also return the `sway_ast::Module` along with the `ParseProgram` and `TypedProgram` when calling `forc::check`. Would probably need to rework a bit of API in `sway-core` to make this ergonomic which might not be worth the trouble. 

The other option is to pass through the attributes into the `ParseProgram` and access the `storage` token from there. I was planning on doing this so we could access the `doc` attribute. Maybe it's worth just closing this PR and adding this functionality back in once that stage has been completed.

sway-lsp/src/core/parse_items.rs
```diff
@@ -0,0 +1,36 @@
...
+    core::token::{AstToken, SymbolKind, Token, TokenMap},
+    utils::token::to_ident_key,
+};
+use std::sync::Arc;
+use sway_ast::{AttributeDecl, ItemKind};
+use sway_core::BuildConfig;
+
+pub(crate) fn parse(src: Arc<str>, build_config: Option<&BuildConfig>, tokens: &TokenMap) {
+    let path = build_config.map(|build_config| build_config.canonical_root_module());
+    if let Ok(module) = sway_parse::parse_file_standalone(src, path) {
```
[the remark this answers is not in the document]
JoshuaBatty: We definitely should log to the output console if an error was returned. We have an issue open to go through all the sway-lsp code and add this functionality. At the moment we just ignore it as the server can never panic.

JoshuaBatty:
Thanks for the review @Centril. I'm going to close this in favor of accessing the attribute tokens in the `ParseProgram` once #2792 lands.

Issue #2463, opened 2022-08-04
Change the Trait `impl for Contract` to contain free functions from `std` like `contract_id()`


Braqzen:
## Summary

Currently the `Contract` does not behave like an object that takes `self` as a parameter meaning it does not allow the user to do something like `self.contract_id()` where `contract_id()` is a function defined in our standard library that returns a pointer to the Tx ID in an external context and the contract id in an internal context.

There are many functions in the standard library that arguably should be moved into the context of the `Contract` rather than have them exist in the standard library where the user must explicitly import them. Perhaps `Contract` may be a library in and of itself.

There's a trade-off between unwanted bloat and additional out-of-the-box functionality that should be evaluated but it's unlikely to be a huge issue unless we strictly follow the standard from Rust where the user does not pay for what they do not use.

That being said, here is a list of possible functions that could be added into `Contract`:

- [msg_sender()](https://github.com/FuelLabs/sway/blob/a3ffd94c413ca62a8fc130a55b296ee61d6ab1e5/sway-lib-std/src/chain/auth.sw#L36-L45)
- [msg_asset_id()](https://github.com/FuelLabs/sway/blob/a3ffd94c413ca62a8fc130a55b296ee61d6ab1e5/sway-lib-std/src/context/call_frames.sw#L36-L42)
- [msg_amount()](https://github.com/FuelLabs/sway/blob/a3ffd94c413ca62a8fc130a55b296ee61d6ab1e5/sway-lib-std/src/context.sw#L30-L33)
- [balance()](https://github.com/FuelLabs/sway/blob/a3ffd94c413ca62a8fc130a55b296ee61d6ab1e5/sway-lib-std/src/context/registers.sw#L68-L73) or [this_balance()](https://github.com/FuelLabs/sway/blob/a3ffd94c413ca62a8fc130a55b296ee61d6ab1e5/sway-lib-std/src/context.sw#L10-L13)
- [contract_id()](https://github.com/FuelLabs/sway/blob/a3ffd94c413ca62a8fc130a55b296ee61d6ab1e5/sway-lib-std/src/context/call_frames.sw#L27-L34)
- [balance_of()](https://github.com/FuelLabs/sway/blob/a3ffd94c413ca62a8fc130a55b296ee61d6ab1e5/sway-lib-std/src/context.sw#L15-L21)
[...]

PR #2778, opened 2022-09-13
fix contract caller types

### LINKED ISSUE #1261: Revamp the `ContractCaller` type
When instantiating a `ContractCaller` "manually" as per : 

https://github.com/FuelLabs/sway/blob/master/test/src/e2e_vm_tests/test_programs/should_pass/language/contract_caller_as_type/src/main.sw#L26

and attempting to call a function implemented for that abi on a deployed contract (i.e. `caller.foo()`), the function is not recognized.

Minimal example : https://github.com/FuelLabs/sway/blob/simon-abi_wrapper/sway-lib-std/tests/test_projects/abi/src/main.sw

sezna:
This fixes some incorrect type unification and closes #1261﻿. Really this was just the result of some behavior being untested, so it regressed over time. The type engine code for unifying contract callers had some typos.

I've disabled the `contract_caller_as_type` test, as this breaks that functionality, but that functionality is very much a special case and will be enabled by #2463 and a rework of the `Contract` type -- this should at least unblock the DAO contract and the berlin workshop.


otrho (approved):

mohammadfawaz:
Thanks Alex! Does this also address https://github.com/FuelLabs/sway/issues/1261#issuecomment-1129175700? If so, then do you mind adding some tests for that use case?

sezna:
Unfortunately, you still can't return a caller as a type from a function. That functionality is currently broken. I could fix that in a follow-on, but I think it will get fixed by introducing the `Contract` type eventually anyway, so i wasn't sure if it was necessary. Additionally, that functionality isn't strictly necessary as you can do dynamic-address calls without functions.

mohammadfawaz (approved):

PR #2781, opened 2022-09-14
Extend support for getting the transaction script hash in std::tx

### LINKED ISSUE #2721: Extend support for getting the transaction script hash in std::tx
Since the operation involves knowing the constant values associated with the GTF opcode, it makes sense to move the following used function for the bridge predicate into the standard library itself:

```

const GTF_SCRIPT_SCRIPT_LENGTH = 0x005;
const GTF_SCRIPT_SCRIPT = 0x00B;

/// Get the hash of the script bytecode
pub fn tx_script_bytecode_hash() -> b256 {
    // Get the script memory details
    let mut result_buffer: b256 = ZERO_B256;
    let script_length = __gtf::<u64>(0, GTF_SCRIPT_SCRIPT_LENGTH);
    let script_ptr = __gtf::<u64>(0, GTF_SCRIPT_SCRIPT);
    
    // Run the hash opcode for the script in memory
    asm(hash: result_buffer, ptr: script_ptr, len: script_length) {
        s256 hash ptr len;
        hash: b256
    }
}
```

pixelcircuits:
Adds new function for efficiently getting the hash of the script bytecode.
Resolves #2721


sway-lib-std/src/tx.sw
```diff
@@ -5,6 +5,7 @@ use ::address::Address;
 use ::mem::read;
 use ::option::Option;
 use ::revert::revert;
+use ::constants::ZERO_B256;
```
mohammadfawaz, line 8: Do you mind keeping these imports in alphabetical order?

sway-lib-std/src/tx.sw
```diff
@@ -232,6 +233,29 @@ pub fn tx_script_bytecode<T>() -> T {
...
 
+/// Get the hash of the script bytecode.
+/// Reverts if not a transaction-script
+pub fn tx_script_bytecode_hash() -> b256 {
+    let type = tx_type();
+    match type {
+        Transaction::Script => {
+            // Get the script memory details
+            let mut result_buffer: b256 = ZERO_B256;
+            let script_length = __gtf::<u64>(0, GTF_SCRIPT_SCRIPT_LENGTH);
```
mohammadfawaz, line 244: Can we use `tx_script_length()` here?

sway-lib-std/src/tx.sw
```diff
@@ -232,6 +233,29 @@ pub fn tx_script_bytecode<T>() -> T {
...
+            let script_ptr = __gtf::<u64>(0, GTF_SCRIPT_SCRIPT);
```
mohammadfawaz, line 245: Can we use `tx_script_start_pointer()` here?

sway-lib-std/src/tx.sw
```diff
@@ -232,6 +233,29 @@ pub fn tx_script_bytecode<T>() -> T {
...
+            
+            // Run the hash opcode for the script in memory
+            asm(hash: result_buffer, ptr: script_ptr, len: script_length) {
+                s256 hash ptr len;
+                hash: b256
+            }
+        },
+        _ => {
+            revert(0);
+        }
```
mohammadfawaz, line 255:
```suggestion
        _ => revert(0),
```

sway-lib-std/src/tx.sw
```diff
@@ -232,6 +233,29 @@ pub fn tx_script_bytecode<T>() -> T {
...
+    match type {
+        Transaction::Script => {
+            // Get the script memory details
+            let mut result_buffer: b256 = ZERO_B256;
+            let script_length = __gtf::<u64>(0, GTF_SCRIPT_SCRIPT_LENGTH);
+            let script_ptr = __gtf::<u64>(0, GTF_SCRIPT_SCRIPT);
+            
+            // Run the hash opcode for the script in memory
+            asm(hash: result_buffer, ptr: script_ptr, len: script_length) {
+                s256 hash ptr len;
```
mohammadfawaz, line 249: I guess there is no way to use the `sha256` library here :/

test/src/sdk-harness/test_artifacts/tx_contract/src/main.sw
```diff
@@ -2,22 +2,22 @@ contract;
 
 use std::{
     address::Address,
```
mohammadfawaz, line 4: `Address` no longer needed here with the prelude :)

test/src/sdk-harness/test_projects/tx_fields/mod.rs
```diff
@@ -170,6 +172,29 @@ async fn can_get_script_start_offset() {
     assert_eq!(result.value, script_start_offset as u64);
 }
 
+#[tokio::test]
+async fn can_get_script_bytecode_hash() {
```
Braqzen, line 176: Should there be another test which somehow checks the negative i.e. comparing one bytecode to another should produce unequal / different results? This proves that the call doesn't return the same output each time

sway-lib-std/src/tx.sw
```diff
@@ -232,6 +233,29 @@ pub fn tx_script_bytecode<T>() -> T {
     read::<T>(tx_script_start_pointer())
 }
 
+/// Get the hash of the script bytecode.
+/// Reverts if not a transaction-script
+pub fn tx_script_bytecode_hash() -> b256 {
+    let type = tx_type();
+    match type {
```
Braqzen, line 240:
```suggestion
    match tx_type() {
```

Since the variable `type` is unused and doesn't cause indentation it is cleaner to include it in the `match` directly

sway-lib-std/src/tx.sw
Braqzen: Why is that?

sway-lib-std/src/tx.sw
```diff
@@ -232,6 +233,29 @@ pub fn tx_script_bytecode<T>() -> T {
...
+        Transaction::Script => {
+            // Get the script memory details
+            let mut result_buffer: b256 = ZERO_B256;
```
Braqzen, line 243:
```suggestion
            let mut result_buffer = ZERO_B256;
```

Compiler can infer on its own

sway-lib-std/src/tx.sw
mohammadfawaz: I have a feeling this doesn't work but I could be wrong - I hope I'm wrong. If not, please file an issue @pixelcircuits

sway-lib-std/src/tx.sw
mohammadfawaz: Mainly because all we have is a pointer and a length and not a concrete value that we can just pass to `sha256`. If we somehow can be build an array of bytes (`u8`s) out of this, then we can just use `sha256` directly, but I'm worried that's not possible just yet (and `Vec` doesn't help here because we'd only be hashing the struct itself and not the heap data).

pixelcircuits: @mohammadfawaz is correct. It's not currently possible to get an array of `u8`s through a simple compiler supported casting or anything. You could get an array of 64 bit "bytes", zero-padded to `u64` length, or the whole data rounded to the nearest `u64` with padded zeros at the end, but then this messes up the computed hash (and also is memory intensive). Since this is being used for bridging in a predicate that is likely to be used often, I think it makes sense to have an ultra optimized assembly defined function rather than try and fix current gaps.

sway-lib-std/src/tx.sw
pixelcircuits: I don't think it works. Every function in the library does it this way (which I copy-pasted from). I'll test to verify and then file an issue if I can confirm it

test/src/sdk-harness/test_artifacts/tx_contract/src/main.sw
pixelcircuits: I get a compile error if I don't include this line, even after updating to the latest forc version

nfurfaro: Not sure if the forc version with the prelude has been released yet... might be only on master?

test/src/sdk-harness/test_projects/tx_fields/mod.rs
pixelcircuits: So it seems like all simple contract calls share the same script bytecode, so it's hard to test against different scripts. I do think it's a good idea that we check that the script is more than a byte long, which ensures that the script has some complexity to it with a unique hash we should expect

sway-lib-std/src/tx.sw
pixelcircuits: So it works! Whatever issue there was before, it seems to have been fixed now

test/src/sdk-harness/test_artifacts/tx_contract/src/main.sw
mohammadfawaz: Wait, this test _is_ being added on `master`... why do we do need a release `forc`?

mohammadfawaz (approved):

sway-lib-std/src/tx.sw
mohammadfawaz: Great! Glad to see these cleaned up.

pixelcircuits:
@Braqzen Does this PR look good now?

Braqzen (approved):
I don't remember if I've said anything about `tx::*` as an import but I always complain about glob imports. Approving anyway

PR #2783, opened 2022-09-14
Prelude docs and updating some examples

mohammadfawaz:
- New section in the book about the prelude.
- Updating the examples to not add unnecessary imports.


docs/src/blockchain-development/calling_contracts.md
```diff
@@ -33,8 +33,6 @@ fn return_45() -> u64 {
 // ./contract_b.sw
 contract;
 
-use contract_a::ContractA;
```
Braqzen, line 36: `make_call()` uses `ContractA` in the `abi` cast however it has been removed as an import. The code should really be extracted into a project so that it can be included and have CI run on it

mohammadfawaz: On my tasks list is to actually rework this whole chapter. So many things wrong there.

Braqzen (dismissed):

Issue #2649, opened 2022-08-29
Problem with newline not being inserted after comments when a token follows it


eureka-cpu:
something like:
```rust
if self == PrimaryColor::Blue {
            Rgb {
                red: 0,
                blue: 255,
                green: 0,
            }
        }
        // TODO remove this else when exhaustive ifs are checked for
      else {
            Rgb {
                red: 0,
                green: 0,
                blue: 0,
            }
        }
```
turns into:
```rust
if self == PrimaryColor::Blue {
            Rgb {
                red: 0,
                blue: 255,
                green: 0,
            }
        }
        // TODO remove this else when exhaustive ifs are checked for else { <-- part of the comment now, should be on newline
            Rgb {
                red: 0,
                green: 0,
                blue: 0,
            }
        }
```

- [x] #2654

eureka-cpu:
This is also happening when multiple comments are before the `module_kind` eg 
```rust
//! A reentrancy check for use in Sway contracts.
//! Note that this only works in internal contexts.
//! to prevent reentrancy: `assert(!is_reentrant);library reentrancy; <-- Here
```

bingcicle:
I tried playing with this for a bit on `forc-fmt v0.33.0`, writing down what I found here:

It seems like indentation matters when it comes to this issue. For example, in the code sample below i intentionally mis-indented the `else` statement here:

```rust
impl MyContract for Contract {
    fn test_function() -> bool {
        if true {
            Rgb {
                red: 0,
                blue: 255,
                green: 0,
            }
        }
        // TODO remove this else when exhaustive ifs are checked for
    else {
            Rgb {
                red: 0,
                blue: 0,
                green: 0,
            }
        }
    }
}
```

The formatter ran fine and left it as is - `else` seems to respect the same column that `fn` is in, instead of `if`. The same thing happens if you moved `else` to the right, the formatter respects the spacing left there and leaves it as is.

Then I tried this:

```rust
impl MyContract for Contract {
    fn test_function() -> bool {
        if true {
            Rgb {
                red: 0,
                blue: 255,
                green: 0,
            }
        }
        // TODO remove this else when exhaustive ifs are checked for
else {
            Rgb {
                red: 0,
                blue: 0,
                green: 0,
            }
        }
    }
}
```

This errors:

```console
$ forc fmt --path qwe
This file: "/.../fuellabs/sway/qwe/src/main.sw" is not part of the build
Error parsing file: Unable to parse: unexpected close delimiter
```

[...]

PR #2787, opened 2022-09-15
Fix comments preceding `module_kind`

### LINKED ISSUE #2739: Fix (and then run) `forc fmt` for `sway-lib-core` and `sway-lib-std`
1. Running `forc fmt` on `sway-lib-core` leads to the following error:
```console
❯ forc fmt

This file: "/Users/mohammad/Desktop/fuel/sway/sway-lib-core/src/ops.sw" is not part of the build
Error parsing file: Unable to parse: Expected `fn`.
```

2. Running `forc fmt` on `sway-lib-std` leads to the following error:
```console
❯ forc fmt

This file: "/Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/reentrancy.sw" is not part of the build
Error parsing file: Unable to parse: Expected a module kind (script, contract, predicate or library).


This file: "/Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/option.sw" is not part of the build
Error parsing file: Unable to parse: Expected a module kind (script, contract, predicate or library).


This file: "/Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/inputs.sw" is not part of the build
Error parsing file: Unable to parse: Expected a module kind (script, contract, predicate or library).


This file: "/Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/hash.sw" is not part of the build
Error parsing file: Unable to parse: Unexpected token after asm return type.


This file: "/Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/result.sw" is not part of the build
Error parsing file: Unable to parse: Expected a module kind (script, contract, predicate or library).


This file: "/Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/alloc.sw" is not part of the build
Error parsing file: Unable to parse: Expected a module kind (script, contract, predicate or library).


This file: "/Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/outputs.sw" is not part of the build
Error parsing file: Unable to parse: Expected a module kind (script, contract, predicate or library).


This file: "/Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/context/registers.sw" is not part of the build
Error parsing file: Unable to parse: Expected a module kind (script, contract, predicate or library).


This file: "/Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/context/call_frames.sw" is not part of the build
Error parsing file: Unable to parse: Expected a module kind (script, contract, predicate or library).
```
Some of these errors go away if I remove the first few lines of comments at the start of the library file (such as `option.sw).

eureka-cpu:
Closes #2739

also ref #2649 though I don't believe this PR will close it, they are similar issues

### DIFF
--- a/swayfmt/src/utils/map/comments.rs
+++ b/swayfmt/src/utils/map/comments.rs
@@ -33,11 +33,11 @@ impl CommentRange for CommentMap {
         // since both beginning of the file and first byte span have their start = 0. If we are looking from STARTING_BYTE_SPAN to `to`, we need to collect all until `to` byte span.
         if from == &byte_span::STARTING_BYTE_SPAN {
             self.range(..to)
-                .map(|items| (items.0.clone(), items.1.clone()))
+                .map(|(byte_span, comment)| (byte_span.clone(), comment.clone()))
                 .collect()
         } else {
             self.range((Included(from), Excluded(to)))
-                .map(|items| (items.0.clone(), items.1.clone()))
+                .map(|(byte_span, comment)| (byte_span.clone(), comment.clone()))
                 .collect()
         }
     }
--- a/swayfmt/src/utils/map/comments.rs
+++ b/swayfmt/src/utils/map/comments.rs
@@ -232,14 +232,21 @@ fn insert_after_span(
     let iter = comments_to_insert.iter();
     let mut offset = offset;
     let mut comment_str = String::new();
+    let mut pre_module_comment = false;
     for comment_with_context in iter {
         let (comment_value, comment_context) = comment_with_context;
+        if comment_value.span.start() == 0 {
+            pre_module_comment = true;
+        }
         write!(
             comment_str,
             "{}{}",
             format_context(comment_context, 2),
             &format_comment(comment_value)
         )?;
+    }
+    if pre_module_comment {
+        writeln!(comment_str)?;
     }
     let mut src_rope = Rope::from_str(formatted_code);
     // If the position we are going to be inserting from + 1 is a \n we are moving that \n after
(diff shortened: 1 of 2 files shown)


swayfmt/src/utils/map/comments.rs
```diff
@@ -232,15 +232,22 @@ fn insert_after_span(
     let iter = comments_to_insert.iter();
     let mut offset = offset;
     let mut comment_str = String::new();
+    let mut pre_module_comment = false;
     for comment_with_context in iter {
         let (comment_value, comment_context) = comment_with_context;
+        if comment_value.span.start() == 0 {
```
mitchmindtree, line 238: I wonder if this condition should be:

```suggestion
        if comment_value.span.start() < from.start() {
```

Though I'm not sure exactly what the `from` span represents here - is it the span of the current item being checked for comments?

---

The issue that comes to mind with using `0` is that this might not work in the case that there's some empty whitespace at the top of the file, e.g.

```rust
 
// Blah blah
library foo;
```
Arguably, there shouldn't be any unnecessary leading whitespace at the top of the file, though we can't be sure the user won't run the formatter with an extra newline at the top.

One potentially hacky solution might be to always `.trim()` the unformatted code `str` before we begin any formatting or parsing to make sure there's no empty whitespace at the beginning or end of the `str`. I think in that case, the comparison to `== 0` here should be OK as if there is a comment preceding the module kind, it would always start at `0`. We might want to add a test to check the case of having an extra newline before the first comment just in case.

mitchmindtree (approved):
I've left a comment at one potentially problematic spot, however I'm happy to approve either way if this improves the majority of troublesome cases in the meantime.

swayfmt/src/utils/map/comments.rs
eureka-cpu: This is actually something I was thinking as well, though in a slightly different context. The comments I think should never have a preceding new line, all new lines should be handled in `newline_handling` including ones that are on comments.

eureka-cpu: #2790

JoshuaBatty (approved):

Issue #1702, opened 2022-05-26
Switch back to tracing for test-suite and sway-core


kayagokalp:
Tracing subscribers was not initialized for `test-suite` and `sway-core`, so we experienced some outputs getting missing after #884. 

To fix this we had to switch back to `println!` for those places but initializing the `tracing subscriber` in the correct place for those parts will enable us to use `tracing` there as well.

- [x] test-suite
- [x] sway-core

Related to:

#1719 
 #1624 
https://github.com/FuelLabs/sway/pull/1688#discussion_r882920929

Issue #1661, opened 2022-05-25
Coloring the output while using tracing


kayagokalp:
Currently, we are using tracing throughout the sway repo after https://github.com/FuelLabs/sway/issues/884, but we do not have a way of coloring outputs while using tracing so currently we are switching back to print once we need colors. We should be able to have colors with tracing as well.

> Currently, the init message is printed with `tracing::info!` but we don't have a solid way of coloring this output with tracing. Maybe, for now, we can switch back to good old `print`. We likely need coloring with `tracing`, so I am opening a new issue to track that. Once that is closed, we can also output with colors with `tracing`.
> 
> _Originally posted by @kayagokalp in https://github.com/FuelLabs/sway/issues/1659#issuecomment-1137125145_

PR #2788, opened 2022-09-15
Replace println with tracing [sway-core]

sdankel:
Closes https://github.com/FuelLabs/sway/issues/1702

Related https://github.com/FuelLabs/sway/issues/1661

### DIFF
--- a/sway-core/src/asm_generation/from_ir.rs
+++ b/sway-core/src/asm_generation/from_ir.rs
@@ -2639,7 +2639,7 @@ mod tests {
 
         let asm_script = format!("{}", asm);
         if asm_script != expected {
-            print!("{}", prettydiff::diff_lines(&expected, &asm_script));
+            tracing::info!("{}", prettydiff::diff_lines(&expected, &asm_script));
             panic!();
         }
     }


sway-core/src/asm_generation/from_ir.rs
```diff
@@ -2639,7 +2639,7 @@ mod tests {
 
         let asm_script = format!("{}", asm);
         if asm_script != expected {
-            print!("{}", prettydiff::diff_lines(&expected, &asm_script));
+            tracing::info!("{}", prettydiff::diff_lines(&expected, &asm_script));
```
mitchmindtree, line 2642: This might not be *quite* the same - it looks like `print!` was used previously which (unlike `println`) doesn't insert a newline at the end, whereas `info!` does. That said, I don't think it matters as this is debug output? cc @otrho

sdankel: True! My guess was that it didn't matter since this looks like it's printing a large blob, and the newline would only be applied at the end. 

A few cases I've seen where it's important to preserve `print` without a newline are where we're expecting user input on the same line, such as `print!("Please provide the signature for this transaction:")`

Do you think should should be using `tracing::debug` instead? (That is, only visible with `--verbose`)

mitchmindtree: I think it's fine to use `info!` for now, as I believe printing this is already gated behind another compiler flag - it might be too confusing to have to enable two different flags.

mitchmindtree: Ahh on a closer look, it seems it's only printed in the test anyway! It's probably fine if we close this and let the test just use `print!`? Using `tracing` is more about giving the user a consistent form of control over the output of forc and related tools, and these tests aren't exposed publicly to the user anyway.

PR #2798, opened 2022-09-17
Make verbose flag a global option in forc

sdankel:
Closes https://github.com/FuelLabs/sway/issues/2305

Adds the `--verbose` flag to all forc commands and makes it so `tracing::debug` logs shows when `-v` or `--verbose` is used, and `tracing::trace` logs show when `-vv` is used.

If both `RUST_LOG` and the verbose flag are set, the verbose flag takes precedence.

Tested with permutations of:
```
cargo run -p forc -- clean --verbose
cargo run -p forc -- init -vv
```

### DIFF
--- a/forc/src/cli/mod.rs
+++ b/forc/src/cli/mod.rs
@@ -6,9 +6,10 @@ use addr2line::Command as Addr2LineCommand;
 use anyhow::{anyhow, Result};
 pub use build::Command as BuildCommand;
 pub use check::Command as CheckCommand;
-use clap::Parser;
+use clap::{Parser, Subcommand};
 pub use clean::Command as CleanCommand;
 pub use completions::Command as CompletionsCommand;
+use forc_util::init_tracing_subscriber;
 pub use init::Command as InitCommand;
 pub use new::Command as NewCommand;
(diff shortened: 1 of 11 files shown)


forc/src/cli/mod.rs
```diff
@@ -26,9 +27,12 @@ struct Opt {
     /// the command to run
     #[clap(subcommand)]
     command: Forc,
+    /// Use verbose output.
+    #[clap(short, long, parse(from_occurrences), global = true)]
```
sdankel, line 31: This will have to change when we upgrade to clap v3.2:
https://github.com/clap-rs/clap/blob/v3.2.22/tests/derive/nested_subcommands.rs#L21

mitchmindtree (approved):
Looks good!

JoshuaBatty (approved):

kayagokalp (approved):

kayagokalp:
Hey @sdankel, just a heads-up looks like CI failed here too (a PR of mine is in the same situation) due to clippy error because rust v1.64 released. I opened-up https://github.com/FuelLabs/sway/pull/2830 to resolve this, once that is in, this should again go green :)

PR #2801, opened 2022-09-17
Re-enable formatting in the `lsp`

### LINKED ISSUE #2802: Re-enable the formatting capabilities of the `lsp`
This includes the removal of `SwayConfig` being replaced by the formatter's `Config` and converting the `serde_json::Value` to its corresponding `ConfigOptions`.

eureka-cpu:
Opening for visibility

Closes #2802


sway-lsp/src/core/session.rs
```diff
@@ -127,7 +129,10 @@ impl Session {
     // update sway config
     pub fn update_config(&self, options: Value) {
```
JoshuaBatty, line 130: We can remove this function entirely as we aren't allowing users to update the config from the plugin anymore.

sway-lsp/src/core/session.rs
```diff
@@ -350,8 +355,8 @@ impl Session {
         if let Some(document) = self.documents.get(url.path()) {
             match self.config.read() {
                 std::sync::LockResult::Ok(config) => {
-                    let config: SwayConfig = *config;
-                    let mut formatter = Formatter::from(config);
+                    let config: Config = *config;
```
JoshuaBatty, line 358: let's just call `swayfmt::Config::default()` here directly.

sway-lsp/src/core/session.rs
```diff
@@ -38,7 +40,7 @@ pub struct CompiledProgram {
 #[derive(Debug)]
 pub struct Session {
     pub documents: Documents,
-    pub config: RwLock<SwayConfig>,
+    pub config: RwLock<Config>,
```
JoshuaBatty, line 43: Let's remove config from the Session struct and just initialize swayfmt where it's used.

kayagokalp (approved):
utACK

JoshuaBatty (approved):
Nice one, thanks!

PR #2804, opened 2022-09-18
`trim_start` src code before creating `newline_map`

### LINKED ISSUE #2791: `.trim()` source code before formatting
As stated in #2787 , if we trim the extra space around the source code we can guarantee that in the instance comments are found before `module_kind` they will be on the first line.

eureka-cpu:
Closes #2791


kayagokalp (approved):
LGTM

mitchmindtree (approved):

PR #2805, opened 2022-09-18
Fix relationship between `handle_comments` & `handle_newlines`

### LINKED ISSUE #2790: Fix relationship between `handle_comments` & `handle_newlines`
Comments should not have preceding newlines, nor extra ending newlines. All extra newlines should be handled by the newline handler instead. Going about it this way, we ensure that there are no conflicts between `comment` and `newline` maps and as a result we also get the benefit of securing the logic from #2787 so that comments preceding `module_kind` are guaranteed to start at the beginning of a file.

eureka-cpu:
Opening for visibility

Closes #2790


eureka-cpu:
Closing for now since we will likely come up with a new strategy for handling newlines and comments cc @mitchmindtree

PR #2807, opened 2022-09-19
Remove unwraps in the language server when calling into the declaration engine. 

JoshuaBatty:
Removes the calls to `unwrap()` when getting types from the declaration engine and only continues if the result is `Ok`.

### DIFF
--- a/sway-lsp/src/capabilities/hover.rs
+++ b/sway-lsp/src/capabilities/hover.rs
@@ -63,21 +63,28 @@ fn hover_format(token: &Token, ident: &Ident) -> Hover {
                 }
                 TypedDeclaration::FunctionDeclaration(func) => extract_fn_signature(&func.span()),
                 TypedDeclaration::StructDeclaration(decl_id) => {
-                    // TODO: do not use unwrap
-                    let struct_decl =
-                        declaration_engine::de_get_struct(decl_id.clone(), &decl.span()).unwrap();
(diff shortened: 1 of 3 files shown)


sway-lsp/src/capabilities/hover.rs
```diff
@@ -63,21 +63,28 @@ fn hover_format(token: &Token, ident: &Ident) -> Hover {
...
-                    // TODO: do not use unwrap
-                    let struct_decl =
-                        declaration_engine::de_get_struct(decl_id.clone(), &decl.span()).unwrap();
-                    format_visibility_hover(struct_decl.visibility, decl.friendly_name())
+                    match declaration_engine::de_get_struct(decl_id.clone(), &decl.span()) {
+                        Ok(struct_decl) => {
+                            format_visibility_hover(struct_decl.visibility, decl.friendly_name())
+                        }
+                        Err(_) => token_name,
+                    }
```
otrho, line 71: These could probably be `.map(|struct_decl| format_visibility_hover(...)).unwrap_or(token_name)` which is better than a `match`?  I don't know, it's subjective I think.

otrho (dismissed):

sway-lsp/src/capabilities/hover.rs
JoshuaBatty: That's a good suggestion, thanks! I'll make this change now.

otrho (approved):

mitchmindtree (approved):

PR #2808, opened 2022-09-19
Refactor `should_pass/stdlib/vec` tests

sezna:
Noticed a lot of repeated code in this test at the hackathon, this just simplifies the copy/pasted tests into a generic version. I'll take a look at `should_fail` vec tests next.


sezna:
Hitting gas limits...will revisit when we have proper calls.

PR #2810, opened 2022-09-20
Extend support for Input::Coin type fields in std::inputs

pixelcircuits:
Waiting on SDK message integration to finish tests


Braqzen (commented):
There are some merge conflicts, that aside:

1. I would match on the function return rather than create a local variable and then `match var` if that var is only used in the `match`
2. I might not bother with cleaning up existing tests, it depends, however I would add new tests into some hierarchical structure where it's clear which concept is being tested by looking at the `mod <name>`

pixelcircuits:
This issue is now covered by #2707 since there was so much overlap

PR #2824, opened 2022-09-22
feat: handle implicit_std_dep for nightly builds

### LINKED ISSUE #2821: handle checking out repo for nightly builds
Currently `forc check` (and also `build`) on a `nightly` build fails due to the version formatting:

forccheck

We need to handle the case where the tag is appended with the build metadata of the date/commit hash for forc-pkg to work as intended for nightlies.

bingcicle:
closes #2821 

Update `implicit_std_dep()` to handle the case where forc-pkg is used within a nightly build e.g. `0.24.4+nightly.20220922.abcdefg`. 

Previously, this would parse the entire version + build metadata as a single tag. Here, we parse the build metadata and prefer the commit hash if available.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,7 +120,7 @@ jobs:
           cargo install --debug --path ./forc-plugins/forc-fmt
           cargo install --debug --path ./forc-plugins/forc-lsp
           cargo install --debug --path ./forc-plugins/forc-explore
-          cargo install --debug --path ./forc-plugins/forc-client
+          cargo install --debug forc-client
       - name: Install mdbook-forc-documenter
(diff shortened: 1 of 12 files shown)


mitchmindtree (approved):
Noice

kayagokalp (dismissed):
👍

mitchmindtree (approved):

bingcicle:
@mitchmindtree @kayagokalp Requesting for a re-review because I made some unnecessary changes :see_no_evil: wanted to handle empty strings but realized that empty strings is technically impossible because of semver. Reverted the change to make it less complicated.

kayagokalp (approved):

PR #2827, opened 2022-09-22
Reduce redundant `TypeId` creation

emilyaherbert:
Nowadays `TypeId::update_type` is called from `CopyTypes` and is used in monomorphization to create new copies of the types present in the given `TypeMapping`. When the type system first went into the compiler way back when, this functionality was overloaded a bit for other uses, so it was necessary to _always_ replace `self` with a new `TypeId`, even if there was no match in the given `TypeMapping`. With the developments on the type system, this is no longer necessary, so we can remove it. Removing this bit of code cuts down drastically on the number of `TypeId`'s created during type checking.

- There are also some unrelated clippy changes in this PR because clippy released a new version and it was causing CI to fail.

This table shows the number of `TypeId`'s created during type checking before this PR and after this PR. The number of `TypeId`'s created for tests cases that import `core` or `std` will always include the number of `TypeId`'s created for `core` or `std`. For example, `std` imports `core`, so 3,078 of the 170,349 are just `TypeId`'s from `core`.

|                                                                                 | before:   | after:  |
|---------------------------------------------------------------------------------|-----------|---------|
| core                                                                            | 57,570    | 3,078   |
| std (w/ core)                                                                   | 1,019,669 | 170,349 |
| should_pass/language/generic_type_inference (w/ core + std)                     | 1,019,741 | 170,421 |
| should_pass/language/funcs_with_generic_types (w/ core)                         | 57,969    | 3,429   |
| should_pass/language/fixing_generic_type (w/ core + std)                        | 1,042,578 | 179,518 |
| should_pass/language/generic_impl_self (w/ core)                                | 61,815    | 6,019   |
| should_pass/language/match_expressions_inside_generic_functions (w/ core + std) | 1,026,071 | 170,823 |
| should_pass/language/generic_structs (w/ core)                                  | 57,746    | 3,254   |
| should_pass/language/generic_enum (w/ core)                                     | 57,606    | 3,110   |
| should_pass/language/generic_inside_generic (w/ core)                           | 58,272    | 3,220   |
| should_pass/language/array_generics (w/ core)                                   | 57,728    | 3,232   |
| should_pass/language/generic_functions (w/ core)                                | 57,696    | 3,204   |
| should_pass/language/generic_struct (w/ core)                                   | 57,634    | 3,142   |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,7 +120,7 @@ jobs:
           cargo install --debug --path ./forc-plugins/forc-fmt
           cargo install --debug --path ./forc-plugins/forc-lsp
           cargo install --debug --path ./forc-plugins/forc-explore
-          cargo install --debug --path ./forc-plugins/forc-client
+          cargo install --debug forc-client
       - name: Install mdbook-forc-documenter
         uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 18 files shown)


mohammadfawaz (approved):
Very cool!

Issue #2449, opened 2022-08-03
Refactor IR memory management and pointers.


otrho:
As of #2363 we have explicit reference semantics for mutable args including copy-type args.

## Problem

This is a bit of a stop-gap measure though as memory management in the IR is inconsistent and sometimes ambiguous.

In particular:
- There are no explicit memory manipulation instructions like `mem_cpy` or `mem_set`.
- Pointers are only for locals and now some args.
- Mutability is neither explicit nor enforced.
- There's a lack of distinction between (Sway's) copy types and reference types.
- Arrays and structs are both aggregates but are actually handled orthogonally.

There's probably other things I've forgotten right now.

## Proposal

I think to address some of these issues we should change the following:
- Use `store` and `load` _only_ for integer 64-bit types and with explicit pointer src/dst.
- Introduce memory instructions for copying larger values.
- No such thing as Sway's implicit copy or ref types, instead pointers must be explicitly used when necessary.
- All pointers are mutable.  It's not up to the IR to worry about mutability, the front-end compiler can enforce those restrictions.  No `mut` keyword needed.
- Split arrays and structs.  Abandon 'aggregates' as they're implemented now.
- Use `get_ptr` as LLVM's `GEP`, update where needed in the land of explicit pointers.  Use `int_to_ptr` and `ptr_to_int` instead of casting in `get_ptr`.
- **Introduce (or at least consider) 'opaque' pointers which are untyped.  This is the approach LLVM has switched to.**

I guess not all of these should be done in a single PR, but much of it will need to be done together to remain workable without intermediate hacks.  This is another of those not-urgent-but-should-be-done-sooner-rather-than-later-to-avoid-pain.

mohammadfawaz:
> Introduce (or at least consider) 'opaque' pointers which are untyped. This is the approach LLVM has switched to.

When I saw the keyword `ptr` in IR when I first started contributing in January, I thought you designed pointers to be opaque just like LLVM's future direction. The challenge with opaque pointers is losing some information related to size of the type so that it's harder to compute offsets (e.g. opaque pointer arguments etc.). LLVM was trying to fix that with some metadata and some other ways of carrying type information IIRC, but I haven't been keeping track of its recent progress. On the flip side, analyzing IR with opaque pointers is much easier and enables more optimization opportunities. I think we should seriously consider this option and it's clearly the better option long term (WWLLVMD?)

I also agree with the rest of the proposal. Thanks!

otrho:
[...]

PR #2828, opened 2022-09-22
Add `raw_ptr` type

AlicanC:
This PR:
- Introduces the `raw_ptr` type.
- Changes RawVec's `ptr` field from `u64` to `raw_ptr`.
- Disallows returning `raw_ptr`s (and aggregate types that contain it) from being returned from `main()` fns.

image

---

Note to @sezna:

On our meeting we discussed `checks.rs`, which led me to this implementation: https://github.com/FuelLabs/sway/pull/2828/commits/090a4d1a4d133cbaa9d4c2499a83f3dfb066bddb

That didn't work because `raw_ptr`s are converted to `sway_ir::irtype::Type::Uint(64)`s so there wasn't a way to differentiate them. I've reverted that and went with this instead: https://github.com/FuelLabs/sway/pull/2828/commits/fc6bf34084a47502cb281dc2533714a4b666b4e4


sway-core/src/lib.rs
```diff
@@ -376,6 +376,34 @@ pub(crate) fn compile_ast_to_ir_to_asm(
     let mut warnings = Vec::new();
     let mut errors = Vec::new();
 
+    // If the program is a script or a predicate,
+    // check if its main() returns a type containing a raw_ptr.
+    match &program.kind {
```
mohammadfawaz, line 381: Can we do this check in https://github.com/FuelLabs/sway/blob/c096c4a640319ac020e4f20baa898e081f6825fb/sway-core/src/semantic_analysis/program.rs#L61? We have a bunch of checks there that seems quite similar.

mohammadfawaz: Also, predicates need to return a `bool` anyways so hopefully that's checked somewhere and doesn't need to be checked again.

sway-core/src/type_system/type_info.rs
```diff
@@ -87,6 +87,13 @@ pub enum TypeInfo {
     Storage {
         fields: Vec<TyStructField>,
     },
+    /// Raw untyped pointers.
+    /// These are represented in memory as u64 but are a different type since pointers only make
+    /// sense in the context they were created in. Users can obtain pointers via standard library
+    /// functions such `alloc` or `stack_ptr`. These functions are implemented using asm blocks
```
mohammadfawaz, line 93: Maybe only mention the `Vec` example here? `alloc` and `stack_ptr` did not really change in this PR. I know Andrew changed them in his previous PR but not sure if that's something we really want to do just yet.

sway-core/src/type_system/type_info.rs
```diff
@@ -178,6 +185,9 @@ impl Hash for TypeInfo {
                 look_up_type_id(*elem_ty).hash(state);
                 count.hash(state);
             }
+            TypeInfo::RawUntypedPtr => {
+                state.write_u8(20);
```
mohammadfawaz, line 189: Should this just be `19`?

sway-error/src/error.rs
```diff
@@ -653,6 +653,8 @@ pub enum CompileError {
     ConfigTimeConstantNotALiteral { span: Span },
     #[error("ref mut parameter not allowed for main()")]
     RefMutableNotAllowedInMain { param_name: Ident },
+    #[error("returning raw_ptrs are not allowed for main()")]
```
mohammadfawaz, line 656:
```suggestion
    #[error("returning a `raw_ptr` from `main()` is not allowed")]
```

test/src/e2e_vm_tests/test_programs/should_fail/raw_ptr/vec_ret/src/main.sw
```diff
@@ -0,0 +1,8 @@
+script;
+
+use std::vec::Vec;
```
mohammadfawaz, line 3: No need to import `Vec` anymore.

sway-lib-std/src/vec.sw
```diff
@@ -6,16 +6,28 @@ use ::intrinsics::size_of;
...
+        r1: raw_ptr
+    }
+}
+
+fn raw_ptr_into_u64(ptr: raw_ptr) -> u64 {
+    asm(r1: ptr) {
+        r1: u64
+    }
+}
+
```
mohammadfawaz, line 20: I dislike these 😢 but I understand the need for them. Maybe we should, in fact, have `alloc` return a `raw_ptr` and [follow what Andrew did](https://github.com/FuelLabs/sway/pull/2699/files) (the whole thing with `fn offset_ptr()`. I know that in that case, some standard library tests will fail because of our check, but that's okay because we can modify them to return a `u64` via an `asm` block that converts that `raw_ptr` into a `u64` (in the tests themselves). What do you think?

sway-lib-std/src/vec.sw
```diff
@@ -6,16 +6,28 @@ use ::intrinsics::size_of;
...
 struct RawVec<T> {
-    ptr: u64,
+    ptr: raw_ptr,
```
mohammadfawaz, line 22: This is going to break the input `Vec` support in the SDks. We should communicate that asap with new issues. We clearly don't have any SDKs tests that use `Vec` as an input yet which is why CI is passing here :) I opened [this issue ](https://github.com/FuelLabs/sway/issues/2922)recently and I hope to get to it soon.

test/src/e2e_vm_tests/test_programs/should_fail/raw_ptr/raw_ptr_ret/src/main.sw
```diff
@@ -0,0 +1,6 @@
+script;
+
+fn main() -> raw_ptr {
```
mohammadfawaz, line 3: I would be nice to also have a positive test here.. even something simple, like a unit test, where you convert a `raw_ptr` to a `u64`, do some math, and then back, just to show that the type is functional.

sway-lib-std/src/vec.sw
mohammadfawaz: Actually I've also opened https://github.com/FuelLabs/fuels-rs/issues/600 which might be enough.

sway-core/src/type_system/type_info.rs
AlicanC: Why not `18`?

mohammadfawaz: Good question. I think I forgot what comes after 17.

sway-lib-core/src/raw_ptr.sw
```diff
@@ -0,0 +1,67 @@
+library raw_ptr;
+
+impl raw_ptr {
```
mohammadfawaz, line 3: Ouu these are interesting 👀

sway-lib-core/src/raw_ptr.sw
```diff
@@ -0,0 +1,67 @@
...
+    /// Returns `true` if the pointer is null.
+    pub fn is_null(self) -> bool {
+        let addr = asm(ptr: self) { ptr: u64 };
+        __eq(addr, 0)
+    }
+
+    /// Gets the address of the pointer.
+    pub fn addr(self) -> u64 {
+        asm(ptr: self) { ptr: u64 }
+    }
```
mohammadfawaz, line 13: How about we be more explicit here and call this `as_u64` or something like that? `addr()` seems like something you'd apply on a variable to get its address, and not on a pointer.

sway-lib-core/src/raw_ptr.sw
```diff
@@ -0,0 +1,67 @@
...
+    
+    /// Copies `size` bytes from `self` to `dst`.
+    pub fn copy_to(self, dst: raw_ptr, count: u64) {
+        asm(dst: dst, src: self, count: count) {
+            mcp dst src count;
+        };
+    }
+    
+    /// Copies `size` bytes from `src` to `self`.
+    pub fn copy_from(self, src: raw_ptr, count: u64) {
```
mohammadfawaz, line 49: While all the methods in this file are tested (directly or indirectly), this one is not tested at all. It would be nice to have some tests for it.

sway-lib-core/src/raw_ptr.sw
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /// Gets the address of the pointer.
+    pub fn addr(self) -> u64 {
+        asm(ptr: self) { ptr: u64 }
+    }
+
+    /// Calculates the offset from the pointer.
+    pub fn add(self, count: u64) -> raw_ptr {
+        let addr = asm(ptr: self) { ptr: u64 };
```
mohammadfawaz, line 17: Ugh too bad we can't use the methods of an `impl` block in other methods in the same `impl` block :(

This will get fixed soon though!

sway-lib-core/src/ops.sw
```diff
@@ -302,6 +308,21 @@ impl Ord for b256 {
...
+            gt r3 r1 r2;
+            r3: bool
+        }
+    }
+    fn lt(self, other: Self) -> bool {
+        asm(r1: self, r2: other, r3) {
+            lt r3 r1 r2;
+            r3: bool
+        }
+    }
```
mohammadfawaz, line 323: These aren't tested anywhere as well.

mohammadfawaz (commented):
Generally very happy with the direction of this. Thank you 😄

sway-lib-core/src/raw_ptr.sw
AlicanC: Pointers can have more than an address so it made sense to me. Are we going to keep `raw_ptr`, or will it eventually turn into a typed pointer where it has a length too?

I was thinking:
- Turn `raw_ptr` into `ptr<T>` (or `ptr T` or `&T`).
- Add slices (`ptr<[T]>`?) so pointers can have a length in runtime which can be used for serialization.
- Extend the spec so our serialization format supports dynamic lengths.
- Implement `Serialize/Deserialize` traits for primitives and slices.
- Implement `Into<ptr<[T]>> for Vec<T>`.
- Implicitly call `Serialize::serialize` and `Deserialize::deserialize` when reading `script_data` or returning from a script. (Only allow `Into<Serialize>` to be returned, `Into<Deserialize>` to be inputted.)

Notes:
- Might need https://github.com/FuelLabs/sway/issues/3018
- Related: https://github.com/FuelLabs/sway/issues/2449

mohammadfawaz: I'm not sure what the future of `raw_ptr` is going to be and your suggestions could certainly be possible! (Worth documenting in an issue for bikeshedding).

> Pointers can have more than an address so it made sense to me.

What does that mean specifically? At any given point in execution, a `raw_ptr` will have a single `u64` value. When I read `addr(ptr)`, in my head this is "address of `ptr`" while in fact all we're doing is an explicit conversion of a pointer to a `u64`, hence the name I suggested.

AlicanC: ATM yes but they could have metadata: https://doc.rust-lang.org/std/ptr/trait.Pointee.html#pointer-metadata

otrho: > * Turn `raw_ptr` into `ptr<T>` (or `ptr T` or `*T`).

We need to be careful and ensure users can't _easily_ get the address of variables and/or create hacks for accessing data they shouldn't really be allowed to.  (Hard to ensure while we have ASM blocks, but still...)  It'd be good to try and keep devs from thinking in terms of pointers as much as possible.

Sway isn't a 100% recreation of Rust for the blockchain, even though it looks that way -- we need to keep it as safe as possible for writing robust contracts without needing a borrow checker.

mohammadfawaz: Yeah I'd imagine only library authors should ever think about pointers.

mohammadfawaz:
Log IDs don't need to be changed anymore. Revering those changes to `master` should be fine.

sezna (approved):

sezna:
JSON ABI mismatch -- likely from a `raw_ptr`

mohammadfawaz:
All is reasonable! I resolved all the comments that have been addressed and left a few related to testing but I'm okay with those being addressed later. Once the conflicts are resolved, I will approve ✅

sway-lib-core/src/raw_ptr.sw
AlicanC: We actually don't need `addr()`. I've removed it.

AlicanC:
I've addressed all the issues, but it looks like some tests are failing. (Merge issues?) Trying to fix those now.

mohammadfawaz:
I think the contract IDs in scripts need to be updated?

AlicanC:
Actually added `addr()` back to get this in quick. Some tests needed refactoring without it.

mohammadfawaz (approved):

mohammadfawaz:
@JC - in [this commit](https://github.com/FuelLabs/sway/pull/2828/commits/0101dba0967a1fd5e9105c84a228b2f2e684b956), I:
* Updated the `token_ops` test to not pass in data from the SDK using a `Vec` because that's what's causing the issue, as expected (`abigen!` cannot handle `raw_ptr` yet).
* With the above, both disabled tests should pass. However, the changes you had to `send_message` seemed a bit buggy. `send_message` is only tested in `token_ops` currently so no surprise that Ci was passing. Regardless, I reverted the implementation of `send_message` to how it was, for the most part. We can iterate on that in the future.

This should be good to go. Once the SDK supports `raw_ptr`, we should add some extensive tests for `Vec` in the inputs and probably revert `token_ops` to how it was before.

mohammadfawaz (approved):

otrho (approved):
This is great, thanks!

PR #2833, opened 2022-09-22
Improve `TypeMapping`

emilyaherbert:
This PR cuts down on some of the unnecessary bloat of the `matches_type_parameter` function and replaces it with an API on an improved `TypeMapping` type.


sway-core/src/type_system/type_mapping.rs
```diff
@@ -1,130 +1,253 @@
...
+                    insert_type(TypeInfo::UnknownGeneric {
+                        name: x.name_ident.clone(),
+                    }),
+                )
+            })
+            .collect();
+        TypeMapping { mapping }
+    }
+
+    pub(crate) fn from_superset_and_subset(
```
mohammadfawaz, line 26: Do you mind adding a comment before this method explaining what it does? Specifically, I'm looking for what `superset` and `subset` exactly refer to.

mohammadfawaz (commented):
All looks good to me. It would be nice to see detailed documentation for all the methods in `TypeMapping` 🙌

sway-core/src/type_system/type_mapping.rs
emilyaherbert: Done!

mohammadfawaz (approved):
Great! Thanks for all the additional comments :)

otrho (approved):

PR #2835, opened 2022-09-22
Move the `resolve_types` function to the `TypeEngine`

emilyaherbert:
Resolving types is the work of the type engine so this PR moves it from `Root` to `TypeEngine`.


otrho (dismissed):
`check_type_is_not_unknown` is a little bit awkward, but I can't really think of a much better name.  Feels like it could be `confirm_type` or something, but whatevs.

mohammadfawaz:
How about something simple like `get_typeinfo` or `get_resolved_typeinfo`? The main things this method does is convert a `TypeId` into a `TypeInfo`... the fact that it returns a `Result` tells us that it can fail.

otrho:
Oh, true, I thought it was returning the same thing and just rejecting unkowns, I didn't notice the conversion.

emilyaherbert:
I renamed it to `to_typeinfo`... let me know if that feels better!

mohammadfawaz (approved):
LGTM! thanks!

otrho (approved):

mohammadfawaz (approved):

Issue #2836, opened 2022-09-23
Add set method to `Vec<T>`


mauricewbr:
ref #2014

Add functionality that enables to assign a value to a specific index

PR #2837, opened 2022-09-23
Add `set` function to `Vec<T>`

mauricewbr:
#2836 and #2014 

Adding a set function to Vec<T> which updates an element at position `index` with a new element `value`.


sway-lib-std/src/vec.sw
```diff
@@ -232,4 +232,22 @@ impl<T> Vec<T> {
...
+    /// * index - The index of the element to be set
+    /// * value - The value of the element to be set
+    ///
+    /// # Reverts
+    ///
+    /// Reverts if `index` is greater than or equal to the length of vector.
+    pub fn set(ref mut self, index: u64, value: T) {
+        assert(index < self.len);
+
+        let end = self.buf.ptr() + index * size_of::<T>();
```
mohammadfawaz, line 249:
```suggestion
        let index_ptr = self.buf.ptr() + index * size_of::<T>();
```

sway-lib-std/src/vec.sw
```diff
@@ -232,4 +232,22 @@ impl<T> Vec<T> {
...
+
+        write(end, value);
```
mohammadfawaz, line 251:
```suggestion
        write(index_ptr, value);
```

test/src/e2e_vm_tests/test_programs/should_fail/vec_set_index_out_of_bounds/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[project]
```
mohammadfawaz, line 1: This file is not needed when adding tests under `e2e_vm_tests` because we rely on `forc` directly for these tests (and not the SDK)

test/src/e2e_vm_tests/test_programs/should_fail/vec_set_index_out_of_bounds/tests/harness.rs
```diff
@@ -0,0 +1,40 @@
+use fuels::{prelude::*, tx::ContractId};
```
mohammadfawaz, line 1: Same comment as before. This file can be removed.

test/src/e2e_vm_tests/test_programs/should_fail/vec_set_index_out_of_bounds/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
```
mohammadfawaz, line 1: A `test.toml` is missing here. You can look at other `should_fail` `Vec` tests to see how to write one :)

test/src/e2e_vm_tests/test_programs/should_fail/vec_set_index_out_of_bounds/.gitignore
```diff
@@ -0,0 +1,2 @@
+out
```
mohammadfawaz, line 1: Probably not needed either.

mohammadfawaz (changes requested):
Seems reasonable to me! Thank you for writing this :)

Just a few comments.

test/src/e2e_vm_tests/test_programs/should_fail/vec_set_index_out_of_bounds/src/main.sw
```diff
@@ -0,0 +1,19 @@
...
+
+    vector.push(number0);
+    vector.push(number1);
+
+    assert(vector.len() == 2);
+    assert(vector.capacity() == 2);
+    assert(vector.is_empty() == false);
+
+    vector.set(3, 5);
+}
```
Braqzen, line 19:
```suggestion
}

```

test/src/e2e_vm_tests/test_programs/should_fail/vec_set_index_out_of_bounds/src/main.sw
```diff
@@ -0,0 +1,19 @@
+script;
+
+use std::{assert::assert, vec::Vec};
```
Braqzen, line 3:
```suggestion
```

[Prelude](https://fuellabs.github.io/sway/v0.24.4/introduction/standard_library.html#standard-library-prelude)  automatically includes these imports.

test/src/e2e_vm_tests/test_programs/should_fail/vec_set_index_out_of_bounds/Forc.toml
```diff
@@ -0,0 +1,8 @@
...
+authors = ["Maurice Weber"]
```
Braqzen, line 2:
```suggestion
authors = ["Fuel Labs <contact@fuel.sh>"]
```

test/src/e2e_vm_tests/test_programs/should_fail/vec_set_index_out_of_bounds/Forc.toml
```diff
@@ -0,0 +1,8 @@
...
+entry = "main.sw"
+license = "Apache-2.0"
+name = "vec_set_index_out_of_bounds"
+
+[dependencies]
+std = { path = "../../../../../../sway-lib-std" }
```
Braqzen, line 8:
```suggestion
std = { path = "../../../../../../sway-lib-std" }

```

test/src/e2e_vm_tests/test_programs/should_fail/vec_set_index_out_of_bounds/Forc.toml
Braqzen: [Ref](https://github.com/FuelLabs/sway/blob/master/test/src/e2e_vm_tests/test_programs/should_fail/vec_swap_param1_out_of_bounds/test.toml)

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
```diff
@@ -1809,4 +2406,4 @@ fn test_vector_with_capacity_u64() {
     match vector.get(5) {
         Option::Some(val) => revert(0), Option::None => (), 
     }
-}
+}
```
Braqzen, line 2409:
```suggestion
}

```

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
```diff
@@ -225,6 +225,64 @@ fn test_vector_new_u8() {
...
+        Option::Some(val) => assert(val == number3), Option::None => revert(0), 
+    }
+
+    match vector.get(1) {
+        Option::Some(val) => assert(val == number1), Option::None => revert(0), 
+    }
+
+    match vector.get(2) {
+        Option::Some(val) => assert(val == number2), Option::None => revert(0), 
+    }
```
Braqzen, line 253: The first step to performing a test is proving the initial conditions are what they are expected to be.
This means that the workflow is as follows:

1. Set up the conditions
2. Prove that the subject of the test has been correctly set in the set up
3. Perform computation
4. Prove that the computation has changed the originally set subject and that it has changed to the expected result

The code in this section, and subsequent sections, sets up the conditions and performs computation without proving the original conditions. 
This could lead to unexpected results depending on the coverage of tests.
For example, if you imagine that the rest of the test suite does not exist and `push` has been changed then it's possible that `push` ignores whatever the first argument is passed in and sets the value of 3 (as does the computation in the new function) and then it works as expected.

If the initial conditions are proven then such a change to `push` would be caught and the test would not falsely pass.

Braqzen (changes requested):

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec/src/main.sw
mauricewbr: Alright, so i'll incorporate additional `assert` checks after using `push` for the set up

Braqzen (approved):

mohammadfawaz (approved):
Awesome, thank you so much for this and congrats on your first contribution 😃

Issue #2838, opened 2022-09-23
Self type inference/unification not working for trait method application


tritao:
This issue comes up on top of https://github.com/FuelLabs/sway/pull/2839 with the following test case:

```rust
script;

trait Store {
    fn store(self, key: b256);
}

impl Store for b256 {
    fn store(self, key: b256) {
        asm(r1: key, r2: self) {
            swwq r1 r2;
        };
    }
}

fn foo<T>(key: b256, value: T) where T: Store {
    value.store(key)
}

fn main() {
    let x: b256 = 0x000000000000000000000000000000000000000000000000FFFFFFFFFFFFFFFF;
    foo(x, x);
}
```

The issue is that in the final AST after type checking, the `store` declaration attached to the `value.store(key)` call still contains references to the `Self` type.

Which later causes us to fail with:

```
3 | trait Store {
4 |     fn store(self, key: b256);
  |              ^^^^ Internal compiler error: Self type cannot be resolved in IR.
Please file an issue on the repository and include the code that triggered this error.
5 | }
```

emilyaherbert:
Just took a look at this, and I think the error that your seeing is due to an issue with the implementation of #2839---closing this for now.

PR #2839, opened 2022-09-23
Add initial support for trait constraints

tritao:
So this is the initial work for trait contraints and is still not ready for review.

The final PR will actually be split up since this one also contains a fix for https://github.com/FuelLabs/sway/issues/2786.

Sending this as a draft so @emilyaherbert can take a look at an issue that came up: https://github.com/FuelLabs/sway/issues/2838

### DIFF
--- a/sway-core/src/declaration_engine/declaration_engine.rs
+++ b/sway-core/src/declaration_engine/declaration_engine.rs
@@ -111,6 +111,17 @@ impl DeclarationEngine {
         self.slab.get(*index).expect_trait(span)
     }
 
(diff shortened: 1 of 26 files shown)


mohammadfawaz:
> The final PR will actually be split up since this one also contains a fix for https://github.com/FuelLabs/sway/issues/2786.

So a fix for #2786 will be made in a separate PR? That's good! We like incremental PRs :)

> So this is the initial work for trait constraints and is still not ready for review.

Are we ready for this or are we planning on closing out all the declaration engine work before finishing trait constraints?

PR #2840, opened 2022-09-23
Bump to `v0.24.5`

mohammadfawaz:


kayagokalp (approved):

bingcicle (approved):
adding a note here that `forc-client` will no longer be built/released from this version onwards - this will be done from its own repo

PR #2842, opened 2022-09-23
Remove `asm` block via function reuse

nfurfaro:


mohammadfawaz (approved):
Looool how did this pass code review 😂 

I don't know why I wrote that code that way originally. Brain freeze moment.

Thanks for fixing!

otrho (approved):

Issue #2819, opened 2022-09-21
Aggregates and memory management in the IR is generally broken.


otrho:
Although the distinction between copy and reference types is more explicit within the compiler now, it wasn't always the case and this is apparent in the IR design.

E.g.,
* Operations like `store` and `load` are used on entire aggregates as a blanket 'write' or 'read'.
* `insert_value` and `extract_value` were attempting to simplify using structs but just ambiguate them.
* Arrays and structs are treated collectively as 'aggregates' but they actually don't share any utility.
* Constant 'aggregates' are created out of thin air where necessary and just permanently stored (and leaked) on the stack.
* The use of pointers is confused and inconsistent.  `get_ptr` was initially only for getting the address of locals but now it's for arguments too and should support globals.

I propose at least the following fixes:
* The removal of `insert_value`, `extract_value`, `insert_element` and `extract_element` and instead use `get_ptr`,  `load` and `store`.
* `load` and `store` are only for writing copy type values to a pointer.
* Pointers are opaque, a la LLVM.
* `get_ptr` is improved to be more like `GEP` from LLVM.
* A global data section is introduced and mutability be strictly specified and adhered to.   Reference typed constants may then be constructed there at compile time.

otrho:
Also currently `get_ptr` is compiled to an address and stored in a register and is then ignored.  And the address calculations in ASM gen are extremely complicated and still have the potential to just fall over for large values that might not fit in an immediate and/or addressing is often quite redundant, calc'ing the same offsets repeatedly.

So, the ASM gen around memory needs a lot of work.

otrho:
I'm also finding now that managing enums, which are tagged unions in IR and ASM, to be extremely painful in the current ASM gen.  In particular, ensuring the padding around the union to make sure the memory accesses are aligned and correct in a union of types of different sizes is very clumsy and should be made explicit in IR with decent memory management.

The workarounds in `sway_core/src/asm_generation/data_section.rs` need to be removed.

otrho:
Also thinking that in IR there's no point in making distinctions between mutable pointers and immutable.  That sort of correctness can be guaranteed by the compiler and so all pointers might as well be mutable.  If we create code which tries to write to the data section then that's a compiler bug, not an issue with IR.

vaivaswatha:
Summary of discussion on slack:

* We'll introduce Alloca instructions. Loads and stores directly use Alloca as their pointer operand. (i.e., the Alloca Value has type "Pointer").
[...]

Issue #1823, opened 2022-06-02
Restructure how functions are handled in the IR and codegen to eliminate function inlining


emilyaherbert:
Once #1824 goes in, we will have the option to support recursive functions. But, in order to do so, we will need to make changes to the IR and codegen.

Blocked by:
- [x] #1824

---

The IR would like to organise a program into a tree.  Instructions gathered into blocks, gathered into functions, gathered into modules.  So having distinct functions and all their meta available to IR would be beneficial structurally and analytically.

To the best of my knowledge, the debugger is mostly interested in attributing spans to instructions, but being able to resolve full paths of symbols to instructions is important to set breakpoints (or inspect memory values).  Without the absolute naming of functions this becomes impossible.  @Dentosal knows more.

_Originally posted by @otrho in https://github.com/FuelLabs/sway/issues/1557#issuecomment-1127188087_

otrho:
The hardest part of removing inlining (and supporting recursion) is supporting actual function calls.  I can't find an issue for this though at the moment.

Function calls are typically done with a `call`/`ret` mechanism, where `ret` is essentially a 'pop an arbitrary value off the stack and jump to it'.

The Fuel VM doesn't support arbitrary jumps so to get around this the current proposal is to use a 'jump table'.  Each 'call' would include a token describing where the function should 'return' to which would be looked up in a static local table.

This may end up less efficient as popular functions would have a very large return table.  But it could also prove more efficient in that [CPS](https://en.wikipedia.org/wiki/Continuation-passing_style) call chains could be generated and tail call optimisations are pretty much built in.

Another downside to using a return table is that library code can never be precompiled and included verbatim (i.e., linked) with user code as the return table for a library function depends on where it is called from, which varies per program of course.

mohammadfawaz:
Just jotting down some things here to make sure I understand the full picture correctly:

Inlining has been used in our conversations to mean two different things:
1. Inlining the function body in the FunctionApplication AST node itself. This is not what the term "inlining" is usually used for I think.
2. Inlining in IR (inliner pass) where the function body is actually moved and replaces the `call` instruction in IR. (`call` here is not a contract call... the contract call instruction in IR is `contract_call` and, of course, no inlining happens there).

We want to avoid (1) by creating a declaration engine that avoids making copies of the function body everywhere it is called. This would be a cleaner and lighter design and enables a much easier solution for the trait constraints problem.

[...]

PR #2843, opened 2022-09-25
Add function calls to the back end.

otrho:
Probably closes some issues, not sure.  Maybe #1823?

Partially blocked by #2885, but the test which depends has been disabled until it's merged.

**This is huge, I've tried to split each commit up into different areas, so it might be easier to review per commit.**

### Done

Here's most of what we need to have non-inlined function calls in ASM gen.  It introduces a calling convention which resembles any typical ABI, using call frames on the stack, reserved registers for args and dynamic jumps to return, much like `BL` on ARM architectures.

A lot of the work has been in tidying up memory use on the stack, to make sure nothing is leaked and doesn't corrupt the call frames.

The data section has been rewritten to no longer use `Literal`s from the parser and instead has its own very basic values, which are almost untyped.  Just words, byte arrays, or collections of those values are supported.

The IR generation has been tightened up a bit to prevent inconsistent state.

The ASM generation has changed order.  There are 3 stages:
- 'virtual' ops which are abstract in the sense that they use an infinite number of registers, use symbolic labels for control flow and can load values by id from the data section.
- 'realised' ops which still use infinite registers but have the labels converted to concrete addresses for control flow, and the data section has explicit storage and addressing.
- 'allocated' ops which have been through register allocation, using only those registers actually available in the VM.

Prior to this change it went 'virtual' -> 'realised' -> 'allocated'.

Register allocation needs to be per function though and after address realisation it's no longer possible to add or remove instructions.  A part of setting up a call frame is to store any registers used in the function to the stack and to restore them before returning, to avoid corrupting values in the caller.  Which registers are used and therefore need to be preserved is unknown until after register allocation, and the act of preserving those registers involves inserting new `STORE` and `LOAD` instructions which must be done before address realisation.

So, we now do 'virtual' -> 'allocated' -> 'realised'.  This meant changing `asm_lang` and `asm_generation` quite a bit.

### To Do

- We still can't make calls to functions with more than 6 args, as they will need to be passed on the stack.
- We still can't make calls to functions which return a reference-type value, as it will also need to be passed (back) on the stack.

For both these cases we still force inlining of those calls.  This was done only due to the desire to get this change in sooner rather than later, and can be addressed in follow-up PRs.

- `asm_lang` is a bit of a mess.  It has been hacked upon to make the changes needed for the new pipeline order, but it needs to be redesigned and to have much of the redundancy removed.
- `asm_generation` is a bit of a mess.  It still needs to be refactored, as what used to be in `from_ir.rs` is now still mostly in `asm_builder.rs`.
- Memory management and handling of aggregates is still a warty part of this code, but I've been complaining about this for ages and aim to clean it up pronto.
- `sway_ir` has had a very open interface and this needs to be locked down some more.  It has grown very organically as needed by the compiler, but it's starting to show some cracks and needs work to straighten it out with a nice consistent API.
- The `inline` pass in `sway-ir` could be tuned, its heuristic is currently just using some parameters I thought would work OK.
- `--print-intermediate-asm` will print the ASM twice at the moment, before and after allocation.  I think we need a 'preliminary' asm output flag so we can see it at each stage.

- https://github.com/FuelLabs/sway/issues/2904
- https://github.com/FuelLabs/sway/issues/2905
- https://github.com/FuelLabs/sway/issues/2505
- https://github.com/FuelLabs/sway/issues/2906
- https://github.com/FuelLabs/sway/issues/2819
- https://github.com/FuelLabs/sway/issues/2907


mohammadfawaz:
>  we now do 'virtual' -> 'allocated' -> 'realised'. 

I do like that, assuming there are no performance implications. It's nice not to have to worry about jumps labels like I was previously doing in the register allocator. I see you've removed that now which is great.

mohammadfawaz (approved):
Can't say I understand every line of code in here, but the overall approach and the changes to the assembly all look good to me. Approving given that tests seem to pass. Did you get a chance to test this PR in the Sway applications repo?

otrho:
> Approving given that tests seem to pass. Did you get a chance to test this PR in the Sway applications repo?

Ahh, I did not, yet.  Will take a look over the weekend.

mohammadfawaz:
We still can't make calls to functions which return a reference-type value

> Sigh, not extraordinary, though these apps are pretty small.

I bet a lot of the functions being called are either too small or return reference types. It'll be interesting to get a count of how many functions are actually *not* inlined per-application.

otrho:
True, let's not get ahead of ourselves.  Need to re-assess after #2905 goes in.

vaivaswatha:
Echoing @mohammadfawaz 's  words, I haven't looked at every detail, but overall this seems ok.

vaivaswatha (dismissed):

vaivaswatha (approved):

mohammadfawaz (approved):

Issue #2854, opened 2022-09-26
Parser: Recover on missing method or field after dot (`.`)


Centril:
(Desired for LSP)

Suppose that we have `foo.`. The user might be writing `foo.field` or `foo.method(...)`. Either way, we should recover in this case, allowing parsing to continue.

PR #2858, opened 2022-09-26
Parser: recovery for auto-completion

Centril:
Recovers on `expr.`

Fixes https://github.com/FuelLabs/sway/issues/2854.

### DIFF
--- a/swayfmt/src/formatter/mod.rs
+++ b/swayfmt/src/formatter/mod.rs
@@ -52,6 +52,7 @@ impl Formatter {
                 .heuristics_pref
                 .to_width_heuristics(self.config.whitespace.max_width),
         );
+        let src = src.trim();
 
         let path = build_config.map(|build_config| build_config.canonical_root_module());
         // Formatted code will be pushed here with raw newline stlye.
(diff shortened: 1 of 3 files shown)


sway-parse/src/expr/mod.rs
```diff
@@ -555,7 +559,11 @@ fn parse_projection(parser: &mut Parser, ctx: ParseExprCtx) -> ParseResult<Expr>
                 };
                 continue;
             }
-            return Err(parser.emit_error(ParseErrorKind::ExpectedFieldName));
+
+            // Nothing expected followed. Now we have parsed `expr .`.
+            // Try to recover as an unknown sort of expression.
+            parser.emit_error(ParseErrorKind::ExpectedFieldName);
+            return Ok(Expr::Error(Span::join(target.span(), dot_token.span())));
```
JoshuaBatty, line 566: I wonder if instead of joining the span's here, if we could instead return a vector of them? This would make accessing the unique parts of the string more ergonomic on the lsp side.

JoshuaBatty (changes requested):
Thanks @Centril just pulled down and tested locally. Definitely heading in the right direction. We just need the `TypedProgram` AST to also be returned so we don't lose type information for the rest of the file while the user is typing.

mohammadfawaz (approved):

sway-core/src/parse_tree/expression/mod.rs
```diff
@@ -117,6 +117,10 @@ pub struct ReassignmentExpression {
 
 #[derive(Debug, Clone)]
 pub enum ExpressionKind {
+    /// A malformed expression.
+    ///
+    /// Used for parser recovery when we cannot form a more specific node.
+    Error,
```
JoshuaBatty, line 123: Is it possible to access the `Box<[Span]>` here? It looks like you have added them to `sway-ast` but not carried them forward into the `ParseProgram`. The `ParseProgram` is the lowest data we access in LSP so anything that isn't carried forward isn't accessible.

JoshuaBatty (changes requested):

sway-core/src/parse_tree/expression/mod.rs
JoshuaBatty: Excellent thanks!

JoshuaBatty (approved):
LGTM

mohammadfawaz (approved):

PR #2861, opened 2022-09-26
Remove extra call to `resolve_type`

emilyaherbert:
This removes a no-op call to `resolve_type`.


test/src/e2e_vm_tests/test_programs/should_pass/language/generic_impl_self/src/main.sw
```diff
@@ -1,5 +1,7 @@
 script;
 
+use std::result::Result;
+
```
mohammadfawaz, line 4: No need for this anymore.

mohammadfawaz (commented):
Is there something blocking this from going in?

Issue #2375, opened 2022-07-25
Add a `--log-level=<level>` CLI flag to `forc` as an alternative to using `RUST_LOG` env var


mitchmindtree:
Currently, the only way to filter certain kinds of logging output is to specify a filter level via `RUST_LOG`, e.g. `RUST_LOG=trace`, `RUST_LOG=off`, and so on.

It could be useful to provide a CLI flag alternative for this, e.g. `--log-level "warn"`. This would have the added benefit of appearing in our auto-generated `forc` documentation, and is arguably a bit cleaner than requiring uses pass a `RUST_LOG` environment variable.

Any `--verbose` options could act as an alias for `--log-level "debug"` or `--log-level "trace"`.

We would need to decide on what option should take precedence in the case that both `RUST_LOG` and `--log-level` were specified. Perhaps we should select the highest-level of the two.

Originally raised here https://github.com/FuelLabs/sway/pull/2308#pullrequestreview-1036769377.

Issue #2362, opened 2022-07-22
Forc stderr verbosity control


segfault-magnet:
If you run `forc build --silent` you would get see some output on your stderr, for example:
```
❯ forc build --silent
  Compiled library "core".
  Compiled library "std".
  Compiled contract "transaction_block_height".
  Bytecode size is 84 bytes.
```

There is [a case](https://unix.stackexchange.com/questions/191254/why-do-many-commands-provide-a-quiet-option) of why `--quiet` or `--silent` flags exist.

> quiet doesn't necessarily mean silent. For some commands, it means reduce verbosity (the opposite of -v|--verbose). For instance, mplayer has a --quiet and a --really-quiet. With some commands, you can use -qqq to decrease verbosity 3 times.

See the [following](https://ell.stackexchange.com/questions/11454/when-to-use-quiet-and-silent#:~:text=Silence%20is%20the%20absence%20of,something%20is%20not%20very%20loud) for the meaning of `silent` vs `quiet`.

I'm having difficulties using `forc` as a library since its design seems to be geared towards it being used only as a command-line app.

Ideally, I could call 
```rust
  forc_build::build(BuildCommand {
                path: Some(path),
                locked: false,
                silent_mode: true,
                ..Default::default()
            })
```
in my process without it spamming my `stderr`. Currently, my workaround is to run this in a separate process. 

Although there are [ways](https://docs.rs/gag/latest/gag/) of controlling when can somebody write to stderr, they are difficult to get right in an async environment.

Regardless of my use case, `forc` should allow for differing levels of verbosity. Ultimately to have a super-quiet aka silent mode whose only means of communicating is via the exit code.

mitchmindtree:
`forc` currently uses [`tracing`](https://docs.rs/tracing/latest/tracing/index.html) for its logging implementation. The [tracing subscriber](https://github.com/FuelLabs/sway/blob/master/forc-util/src/lib.rs#L434-L442) we use for handling logged messages applies an environment filter which [defaults to the "info" log-level](https://github.com/FuelLabs/sway/blob/master/forc-util/src/lib.rs#L431).

The environment filter level can be set via the `RUST_LOG` env var, which you can read a little more about [here](https://docs.rs/env_logger/latest/env_logger/#enabling-logging). You *should* be able to do something like `RUST_LOG=off forc build` to disable all of `forc`'s logging output altogether.

[...]

PR #2870, opened 2022-09-27
Add --log-level and --silent to forc

sdankel:
Closes https://github.com/FuelLabs/sway/issues/2375 https://github.com/FuelLabs/sway/issues/2362

Related https://github.com/FuelLabs/sway/pull/2798

image

1. adds a global `--log-level` option.

2. changes "forc completions --shell" short form from "-s" to "-S" because "-s" is now globally the short form of "--silent"

image

3. similarly, changes "forc addr2line --search-dir" short from from  "-s" to "-S"

image

4. adds a global `--silent` option to all forc commands and renames the old silent mode to "terse mode". 

What used to be called silent mode was not really silent; it would still output 1 line if there were warnings or errors during the build (for example, during CI runs). 
 
There is still a need for a non-global option for terse mode because the test harness runs the build commands directly (without the CLI), so there's no way to pass the global options into the command (for example, `forc::test::forc_check::check` which takes a `forc::cli::CheckCommand` argument).

image

I can't simply set the log level to "ERROR" for the test harness (effectively like a terse mode) because it would be applied to all tests running in that thread, and some of the tests depend on the INFO output to stdout to determine if the test passed. To have separate log levels for `ir_generation` and `e2e_vm_tests`, we'd have to separate those into separate build steps.


kayagokalp (approved):
Looks cool to me! One minor discussion point:

 I wonder if changes to `forc-client` should be done to it's own dedicated repo as `forc-client` from sway repo is about to be depreciated 🤔  Maybe @bingcicle can have a better input for that.

JoshuaBatty (approved):
Nice work @sdankel ! Also cc'ing @bingcicle for his thoughts on migrating the forc-client changes over to the new repo.

bingcicle:
^ sorry I missed this but yeah, the changes to `forc-client` should definitely be in the [new repo](https://github.com/FuelLabs/forc-client) instead

JoshuaBatty:
@sdankel would you mind porting these `forc-client` changes to the new repo in a separate PR? Thanks

PR #2879, opened 2022-09-27
Handle leading newlines and whitespaces before pre-module comments 

### LINKED ISSUE #2868: Some formatter issues in libraries related to new line before comments
### Problem 1
Note the empty line before the first comment (actually you can't see it in the rendering, but it's there if you just copy the code or look at the raw text)
```rust

//! A reentrancy check for use in Sway contracts.
//! Note that this only works in internal contexts.
//! to prevent reentrancy: `assert(!is_reentrant);
library reentrancy;

use std::assert::assert;
use std::context::{call_frames::*, registers::frame_ptr};

/// Returns `true` if the reentrancy pattern is detected, and `false` otherwise.
pub fn is_reentrant() -> bool {
    // Get our current contract ID
    let this_id = contract_id();

    // Reentrancy cannot occur in an external context. If not detected by the time we get to the
    // bottom of the call_frame stack, then no reentrancy has occured.
    let mut call_frame_pointer = if frame_ptr() == 0 {
        0
    } else {
        get_previous_frame_pointer(frame_ptr())
    };
    // The current contract ID wasn't found in any contract calls prior to here.
    false
}
```
`forc fmt` emits:
```console
Error parsing file: unexpected close delimiter
```

### Problem 2
Again, note the empty line before the first comment (same as before)
```rust

//! A reentrancy check for use in Sway contracts.
//! Note that this only works in internal contexts.
//! to prevent reentrancy: `assert(!is_reentrant);
library reentrancy;

use std::assert::assert;
use std::context::{call_frames::*, registers::frame_ptr};

/// Returns `true` if the reentrancy pattern is detected, and `false` otherwise.
pub fn is_reentrant() -> bool {
    // Get our current contract ID
    let this_id = contract_id();

    // The current contract ID wasn't found in any contract calls prior to here.
    false
}
```
Formats without errors but the output is (not empty line before the comment here):
```rust
//! A reentrancy check for use in Sway contracts.
//! Note that this only works in internal contexts.
//! to prevent reentrancy: `assert(!is_reentrant);
library reentrancy;

use std::assert::assert;
use std::context::{call_frames::*, registers::frame_ptr};

/// Returns `true` if the reentrancy pattern is detected, and `false` otherwise.
pub fn is_reentrant() -> bool {

    // Get our current contract ID    let this_id = contract_id();

    // The current contract ID wasn't found in any contract calls prior to here.    false
}
```
which is incorrect. The `let` statement is now in the preceding comment.

I don't know if these are known issues or not, but filing anyways just in case :)

### Problem 3
Running the formatter on the output of Problem 2 leads to an error:
```console
Error parsing file: unclosed delimiter
```

### LINKED ISSUE #2875: Formatter does not trim correctly the input code when there is a whitespace other than '\n' at the beginning
realized working on #2868. I am planning to fix them together.

Example (it is not rendered but can be seen in raw text): 

```
	
//! A reentrancy check for use in Sway contracts.
//! Note that this only works in internal contexts.
//! to prevent reentrancy: `assert(!is_reentrant);
library reentrancy;

use std::assert::assert;
use std::context::{call_frames::*, registers::frame_ptr};

/// Returns `true` if the reentrancy pattern is detected, and `false` otherwise.
pub fn is_reentrant() -> bool {
    // Get our current contract ID
    let this_id = contract_id()
;
    // Reentrancy cannot occur in an external context. If not detected by the time we get to the
    // bottom of the call_frame stack, then no reentrancy has occured.
    let mut call_frame_pointer = if frame_ptr() == 0 {
        0
    } else {
        get_previous_frame_pointer(frame_ptr())
    };
    // The current contract ID wasn't found in any contract calls prior to here.
    false
}

```

kayagokalp:
closes #2875.
closes #2868.

### DIFF
--- a/swayfmt/src/utils/map/comments.rs
+++ b/swayfmt/src/utils/map/comments.rs
@@ -50,10 +50,14 @@ pub fn comment_map_from_src(input: Arc<str>) -> Result<(CommentMap, usize), Form
 
     // pass the input through lexer
     let mut start = 0;
-    if input.starts_with('\n') {
-        let mut input_chars = input.chars().peekable();
-        while let Some('\n') = input_chars.next() {
-            start += 1
+    for (char_index, curr_char) in input.chars().enumerate() {
+        if char_index == 0 && !curr_char.is_whitespace() {
+            break;
+        }
+        if curr_char.is_whitespace() {
+            start += 1;
+        } else {
+            break;
         }
     }
     let commented_token_stream = lex_commented(&input, start, input.len(), None)?;
--- a/swayfmt/src/utils/map/comments.rs
+++ b/swayfmt/src/utils/map/comments.rs
@@ -45,29 +45,18 @@ impl CommentRange for CommentMap {
 
 /// Get the CommentedTokenStream and collect the spans -> Comment mapping for the input source
 /// code.
-pub fn comment_map_from_src(input: Arc<str>) -> Result<(CommentMap, usize), FormatterError> {
+pub fn comment_map_from_src(input: Arc<str>) -> Result<CommentMap, FormatterError> {
     let mut comment_map = BTreeMap::new();
 
     // pass the input through lexer
-    let mut start = 0;
-    for (char_index, curr_char) in input.chars().enumerate() {
-        if char_index == 0 && !curr_char.is_whitespace() {
-            break;
-        }
-        if curr_char.is_whitespace() {
-            start += 1;
-        } else {
-            break;
-        }
-    }
-    let commented_token_stream = lex_commented(&input, start, input.len(), None)?;
+    let commented_token_stream = lex_commented(&input, 0, input.len(), None)?;
     let tts = commented_token_stream.token_trees().iter();
 
     for comment in tts {
         collect_comments_from_token_stream(comment, &mut comment_map);
     }
 
-    Ok((comment_map, start))
+    Ok(comment_map)
 }
 
 /// Collects `Comment`s from the token stream and insert it with its span to the `CommentMap`.
@@ -128,7 +117,7 @@ pub fn handle_comments(
 /// traversal. When `add_comments` is called we have already parsed the unformatted_code so there is no need
 /// to parse it again.
 fn add_comments(
-    comment_map: (CommentMap, usize),
+    comment_map: CommentMap,
     unformatted_module: &Module,
     formatted_module: &Module,
     formatted_code: &mut FormattedCode,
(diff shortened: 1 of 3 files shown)


eureka-cpu:
Thanks for handling this @kayagokalp should I go ahead and close my PR that's trying to address this #2805 ?

swayfmt/src/utils/map/comments.rs
```diff
@@ -50,10 +50,14 @@ pub fn comment_map_from_src(input: Arc<str>) -> Result<(CommentMap, usize), Form
 
     // pass the input through lexer
     let mut start = 0;
-    if input.starts_with('\n') {
-        let mut input_chars = input.chars().peekable();
-        while let Some('\n') = input_chars.next() {
-            start += 1
+    for (char_index, curr_char) in input.chars().enumerate() {
```
eureka-cpu, line 53: If you are trimming the `src` I'm not sure there's a need for this anymore

eureka-cpu: Also can remove the `.trim_start()` from `newline.rs` and the logic from my previous PR #2844

eureka-cpu: My attempt was to try and keep it as in tact as possible but your approach seems better

kayagokalp: Thanks for pointing out!

kayagokalp:
@eureka-cpu, hmm not sure about that but if you think this solves everything that PR is aiming maybe we can close and if it turns out there are some more cases needed to be handled we can re-open 🤔

eureka-cpu (approved):

bingcicle:
Trying to understand this solution a bit more before I give the tick, but was the issue with [this block](https://github.com/FuelLabs/sway/pull/2879/files#diff-23a28d973354ebae49e506d66fe99a95d62481d03232f85c80003800a0e9578bL52-L59) that unintentionally formatted/hid newline characters before the pre-module comments?

kayagokalp:
So https://github.com/FuelLabs/sway/pull/2844 introduced that code block to take into account preceding newlines before module statement. Since the solution from #2844 does not solve it for all cases and it is a little bit complicated to iterate on, this PR replaces it with a different approach.

The desired behavior of the formatter would be removing those extra newlines at the beginning of the file before placing any pre-module comments. So rather than trying to count the preceding newlines and doing complicated shifting operations necessary for removing preceding newlines before any pre-module comments and correctly placing other comments (for solving #2868 it took me several hours to come-up with that shifting logic and I suspect that there were still some edge cases that does not working with that approach). I handled them before doing anything else and removed them directly from the source code as they would be removed at the end so that we do not require to calculate offset and shift other comments accordingly etc. 

cc @bingcicle

bingcicle:
I see, makes sense - thanks for the explanation!

Issue #2876, opened 2022-09-27
`revert` in an `if` branch causes an IR verification error


mohammadfawaz:
Example:
```rust
script;

pub fn revert(code: u64) {
    asm(r1: code) {
        rvrt r1;
    }
}

fn main() -> u64 {
    if true {
        revert(0)
    } else {
        42
    }
}
```
Error:
```console
error: Internal compiler error: Verification failed: Function main return type must match its RET instructions.
```

Basically, the type of the `asm` block resolves to `()` because `revert()` returns a `()`. But because `revert(0)` deterministically aborts, we don't emit a compiler error about mismatched types (i.e. type of `42` is different from the type of `revert(0)`). So this ends up being caught by the IR verifier. 

Note that switching the bodies of `if` and `else` compiles fine. I expect the above to compiler with or without a semicolon after `revert(0)` AND with or without a semicolon after the `asm` block. See https://github.com/FuelLabs/sway/issues/2822 for another issue related to that.

mohammadfawaz:
Now that `rvrt` is not allowed in `asm` blocks, the reproducer looks like:
```
script;

fn main() -> u64 {
    if true {
        std::revert::revert(0) 
    } else {
        42
    }
}
```

otrho:
I think this is supposed to be handled by the `is_diverging()` stuff which Andrew added.  It's most likely just a missed case in the `if` compilation.

mohammadfawaz:
The issue is simply that the type of `revert(0)` is `()` because `fn revert` is defined as 
```rust
pub fn revert(code: u64) {
    __revert(code)
}
```
so that's a conflict of types between the `if` branch and the `else` branch.

esdrubal:
This issue appears to no longer occur in master.


> Now that `rvrt` is not allowed in `asm` blocks, the reproducer looks like:
> 
> ```rust
> script;
> 
> fn main() -> u64 {
>     if true {
>         std::revert::revert(0) 
>     } else {
>         42
>     }
> }
> ```

Added the above test case in #3285

PR #2880, opened 2022-09-27
Small fix to the IR verifier

mohammadfawaz:
The result of the a `reduce` needs to be unwrapped first before checking its content. The `VerifyPhiInconsistentTypes` was never actually reached previously. I was seeing some inconsistent behavior while looking at https://github.com/FuelLabs/sway/issues/2876 and found this issue.

### DIFF
--- a/sway-ir/src/verify.rs
+++ b/sway-ir/src/verify.rs
@@ -622,6 +622,7 @@ impl<'a> InstructionVerifier<'a> {
             .iter()
             .map(|(_, v)| v.get_type(self.context))
             .reduce(|a, b| if self.opt_ty_not_eq(&a, &b) { None } else { b })
+            .unwrap() // Safe to unwrap because `pairs` is *not* empty as this was checked earlier
             .is_none()
         {
             Err(IrError::VerifyPhiInconsistentTypes)


mohammadfawaz:
> Some tests and examples are failing

Yeah :( converted to a draft and taking a look.

otrho:
@vaivaswatha is in the process of removing PHIs from the IR altogether, so unless this is super urgent I'd say it's not worth fixing.

mohammadfawaz:
> @vaivaswatha is in the process of removing PHIs from the IR altogether, so unless this is super urgent I'd say it's not worth fixing.

Yeah that's fair. I can close. But I'm worried about the type mismatch issues that showed up when I made this change. @vaivaswatha are you hitting these issues (with `unwrap` for example) in your current task?

vaivaswatha:
> > @vaivaswatha is in the process of removing PHIs from the IR altogether, so unless this is super urgent I'd say it's not worth fixing.
> 
> Yeah that's fair. I can close. But I'm worried about the type mismatch issues that showed up when I made this change. @vaivaswatha are you hitting these issues (with `unwrap` for example or #2876) in your current task?

I agree with @otrho. This function is going to go.

Regarding the fix, it looks correct to me, so I don't think you should be hitting the type mismatch. I'd say it's worth looking into why that's happening.

If #2881 is the fix for #2876, i.e., we're going to have a `Never` type for one branch of this `if`, I'd say the solution should be, at the IR generation level, to not branch from that arm of the `if` to the merge block, thus avoiding that particular PHI argument (it isn't a predecessor at all, so it won't have a corresponding PHI argument). Does that sound reasonable?

mohammadfawaz:
> Regarding the fix, it looks correct to me, so I don't think you should be hitting the type mismatch. I'd say it's worth looking into why that's happening.

That's happening specifically because the implementation of `unwrap` has a diverging path (a revert in a match) and so when I made the fix here, the type mismatch showed up - the current IR is indeed problematic as is. Now when lowering to assembly, the problem doesn't have any symptoms at runtime, but the IR for unwrap is currently malformed.

vaivaswatha:
> > Regarding the fix, it looks correct to me, so I don't think you should be hitting the type mismatch. I'd say it's worth looking into why that's happening.
> 
> That's happening specifically because the implementation of `unwrap` has a diverging path (a revert in a match) and so when I made the fix here, the type mismatch showed up - the current IR is indeed problematic as is. Now when lowering to assembly, the problem doesn't have any symptoms at runtime, but the IR for unwrap is currently malformed.

At the type checker level, for the two branches of an `if`:
  - Types have to match OR at least one of them must be `Never`
  - The `Never` arm of the `if` will not have a branch out (it's an "exit block"). So no value flows from this block to the PHI of the merge block.
At the IR level,
  - Types have to strictly match. No concept of "Never". (hence the fix in *this* PR is correct).

vaivaswatha:
> > > Regarding the fix, it looks correct to me, so I don't think you should be hitting the type mismatch. I'd say it's worth looking into why that's happening.
> > 
> > 
> > That's happening specifically because the implementation of `unwrap` has a diverging path (a revert in a match) and so when I made the fix here, the type mismatch showed up - the current IR is indeed problematic as is. Now when lowering to assembly, the problem doesn't have any symptoms at runtime, but the IR for unwrap is currently malformed.
> 
> At the type checker level, for the two branches of an `if`:
> 
> * Types have to match OR at least one of them must be `Never`
> * The `Never` arm of the `if` will not have a branch out (it's an "exit block"). So no value flows from this block to the PHI of the merge block.
> 
> At the IR level,
> 
> * Types have to strictly match. No concept of "Never". (hence the fix in _this_ PR is correct).

And to be able to detect this, we'll need a `__revert` intrinsic (we can't deep dive into the ASM to check that).

mohammadfawaz:
Alright I think we all agree on a path forward then. Closing this for now despite the malformed IR (with no side effects luckily). Thanks!

PR #2882, opened 2022-09-27
Add `append` macro

emilyaherbert:
This macro takes an instance of `(Vec<T>, Vec<F>)` `&mut Vec<T>`, and `&mut Vec<F>`. It appends the `Vec<T>` to the `&mut Vec<T>`, and the `Vec<F>` to the `&mut Vec<F>`.

### DIFF
--- a/sway-core/src/error.rs
+++ b/sway-core/src/error.rs
@@ -29,6 +29,14 @@ macro_rules! check {
     }};
 }
 
+macro_rules! append {
+    ($fn_expr: expr, $warnings: ident, $errors: ident $(,)?) => {{
+        let (mut l, mut r) = $fn_expr;
+        $warnings.append(&mut l);
+        $errors.append(&mut r);
+    }};
+}
+
 macro_rules! assert_or_warn {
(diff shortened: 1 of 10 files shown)


sway-core/src/error.rs
```diff
@@ -29,6 +29,14 @@ macro_rules! check {
     }};
 }
 
+macro_rules! append {
```
mohammadfawaz, line 32: How come we're diverging from the usual `CompileResult` for methods like `unify_with_self`? Can we not achieve the same behavior using `CompileResult<()>` and `check!`?

emilyaherbert: No significant reason, `unify_with_self` returns `(Vec<CompileWarning>, Vec<TypeError>)` with no value, while `CompileResult` needs a `Vec<CompileError>` and does return a value. It would be simple to make `unify_with_self` return a `CompileResult<()>` though.

otrho (approved):

sway-core/src/error.rs
mohammadfawaz: Yeah that's fine I suppose. Looks good as is!

mohammadfawaz (approved):

PR #2885, opened 2022-09-28
Change `storage::get()` in libstd to use the heap.

otrho:
Until now it read non-copy types to a leaked buffer on the stack.  This will interfere with function call frames which require stack use symmetry within functions.

It now uses the heap, although it still leaks.

OK, see my note below though.  I don't know why it's failing, maybe there's an off-by-one thing happening or something.  @AlicanC in particular, you might see what's going on here.  @vaivaswatha or @mohammadfawaz you might have some insights too.  Am I missing something obvious?

### DIFF
--- a/sway-lib-std/src/storage.sw
+++ b/sway-lib-std/src/storage.sw
@@ -2,7 +2,6 @@ library r#storage;
 
 use ::alloc::alloc;
 use ::assert::assert;
-use ::context::registers::stack_ptr;
 use ::hash::sha256;
 use ::option::Option;
 use ::result::Result;
@@ -42,7 +41,10 @@ pub fn store<T>(key: b256, value: T) {
     };
 }
 
-/// Load a stack variable from storage.
+/// Load a variable from storage.
+///
+/// If the value size is larger than 8 bytes it is read to a heap buffer which is leaked for the
+/// duration of the program.
 #[storage(read)]
 pub fn get<T>(key: b256) -> T {
     if !__is_reference_type::<T>() {
--- a/sway-lib-std/src/storage.sw
+++ b/sway-lib-std/src/storage.sw
@@ -1,5 +1,6 @@
 library r#storage;
 
+use ::alloc::alloc;
 use ::assert::assert;
 use ::context::registers::stack_ptr;
 use ::hash::sha256;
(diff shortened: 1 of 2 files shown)


sway-lib-std/src/storage.sw
```diff
@@ -50,34 +51,33 @@ pub fn get<T>(key: b256) -> T {
...
 
-        // Keep track of the base pointer for the final result
-        let result_ptr = stack_ptr();
+        // Allocate a buffer for the result.  It needs to be a multiple of 32 bytes so we can make
+        // 'quad' storage reads without overflowing.
+        //
+        // XXX This fails for tests that exercise the loop below unless we allocate an extra byte!
+        // This must be a bug?!
+        //
+        let result_ptr = alloc(((size_left + 31) & 0xffffffe0) + 1);
```
otrho, line 64: If we don't allocate that extra byte then this will fail with a `MemoryOverflow` from the VM.

mohammadfawaz: I could not figure it out. I played around with it a bit and here are some observations:
* Even the case with `b256` (so the loop is not really exercised) is failing, but not all reference types are failing!
* The issue is when _reading_ the final value, so all the steps work fine until we get to the `asm` block at the end.
* Reading from `result_ptr - 1` at the end instead of `result_ptr` does not overflow, but shows that the data is there correctly (except for the first byte which is always 0 because we're reading `result_ptr - 1` and not `result_ptr`. So the data is there but we can't read it for some reason? I also tried replacing the `asm` block with `mem::read` and got the same behavior (expectedly).

mohammadfawaz: I even logged the size allocated and it's correct (` = 32` in the case of `b256`).

mohammadfawaz: And for whatever reason, this works:
```rust
        let result_ptr = alloc(1);
        let result_ptr = realloc(result_ptr, 1, ((size_left + 31) & 0xffffffe0) + 0);
```

otrho: There definitely is a minor weirdness with `ALOC` and `$hp` in that `$hp` always points to the last _unallocated_ byte, though `std::alloc::alloc()` accounts for that.  https://github.com/FuelLabs/fuel-vm/issues/137

The heap grows _down_, so the above snippet doesn't surprise me, it's pretty much the same as `size + 1`.

I was thinking it was only reads that used the loop, but it must be reads which are multiples of 32 bytes and need the entire buffer.  I think most of those tests are actually only 16 bytes.

Reading from `result_ptr - 1` should _not_ work, this is the biggest surprise to me.  It's almost as if the change in the issue I proposed above has been made.  :thinking:  If that were true `Vec` should be broken, as the primary user of `alloc()`.

otrho:
I've done some more digging and I think it's actually a bug in `RETD`.  Since this is a fairly minor change it could just wait as it won't particularly go stale.  It isn't quite blocking my other calls PR yet, as any ref-type return functions will be inlined, and so the old stack based version still works.

I hope to have an issue and maybe PR in `fuel-vm` later on, if I'm right, and so I'll make all the links between here and there when I do.

otrho:
I'm pretty sure https://github.com/FuelLabs/fuel-vm/pull/229 is the culprit here, though I can't _easily_ confirm until we pull it across to our Sway test suite.

mohammadfawaz:
@otrho changing this into a draft until we're ready to merge it.

otrho:
Is fixed in #3055.

PR #2889, opened 2022-09-28
Update constants.sw

SwayStar123:
added more constants


sway-lib-std/src/constants.sw
```diff
@@ -1,6 +1,12 @@
...
+use ::{
+  address::Address,
+  contract_id::ContractId,
+  identity::Identity,
+};
 
 pub const BASE_ASSET_ID = ~ContractId::from(ZERO_B256);
+pub const ZERO_ADDRESS = ~Address::from(ZERO_B256);
 pub const ZERO_B256 = 0x0000000000000000000000000000000000000000000000000000000000000000;
+pub const ZERO_IDENTITY = ~Identity::Address(ZERO_ADDRESS);
```
mohammadfawaz, line 12:
```suggestion
pub const ZERO_IDENTITY = Identity::Address(ZERO_ADDRESS);
```

mohammadfawaz (commented):
What is the reasoning for adding address zero as a special constant? I worry this will encourage bad user code where the `ZERO_ADDRESS` will be used as a default value for addresses.

sway-lib-std/src/constants.sw
SwayStar123: We have a zero contractid already (the base asset) so thought a zero address would also be useful.

adlerjohn: The zero contract ID exists because people might want to refer to the base asset, but TBH it should be renamed to `BASE_ASSET_ID`, where its value should not be relied on to be zero. The zero address serves no purpose.

mohammadfawaz: It is called `BASE_ASSET_ID` already :) We also have `ZERO_B256` but that's just a raw `b256` with no meaning.

And +1 to what John said.

SwayStar123: Okay so should I remove the zero address and instead use the base_asset_id in the identity? The problem I see is that the zero contract id actually refers to something, that being ETH, whereas the 0 address doesnt mean anything, making it more suitable for a "zero identity"

mohammadfawaz: Would `Option` not solve your problem in this case?

Braqzen: Identity is a wrapper of 2 variants. Selecting one of the variants is a bad idea.

mohammadfawaz: Yeah relying on default values is a source of vulnerabilities. `Option` is the way to go here I think even if it's a bit more expensive. Do you agree? And if so, can this PR be closed?

Issue #2864, opened 2022-09-26
Lexer: recover 'abc' as "abc"


Centril:
(Opportunity found as part of refactoring the lexer.)

In the lexer, we can recover the literal `'abc'` as `"abc"` such that the error `ExpectedCloseQuote` is non-aborting.

Issue #2815, opened 2022-09-20
Add parser recovery for incomplete `let x = ` statement


Centril:
As per https://github.com/FuelLabs/sway/pull/2724#pullrequestreview-1104953282, we need to recover if the user wrote e.g., `let x = ` but did not enter a following expression, or failed to terminate the statement with `;`.

This probably needs some lexer tweaks, so that we can even do an `enter_delimited`. Otherwise, we'll not even attempt parsing the block.
However, the lexer probably also lacks necessary info to do this intelligently, if we just attempt to recover for any unclosed delimiters. So we should let the parser know where recovery was made. See https://github.com/rust-lang/rust/blob/87788097b776f8e3662f76627944230684b671bd/compiler/rustc_parse/src/parser/diagnostics.rs#L2239-L2312 for some relevant code.

Issue #2734, opened 2022-09-06
Use `Handler` infra instead of `CompileResult`


Centril:
The idea is to stop using the `CompileResult<T>` monad in `sway_core`. The type is a beautiful concept, and would be the go-to solution in Haskell. However, there is lacking proper language support for early returns (i.e. `do` notation in Haskell). Meanwhile, the `?` operator is inappropriate for logging (writer monad), which is half of what `CompileResult` is (roughly `(Maybe a, [Warning], [Error])`).

Currently, use of `CompileResult<T>` results in a lot of noise dealing with error handling. For example, there's a bunch of code dealing with extending errors and warnings.

Instead of `CompileResult`, we can use the `Handler` concept in the parser, which is also the way [rustc deals with emitting errors](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/struct.Handler.html).

mohammadfawaz:
https://github.com/FuelLabs/sway/pull/3057 can be used for inspiration to resolve this issue.

PR #2891, opened 2022-09-28
Lexer + Parser recovery & Gathering errors in sway-error

Centril:
This PR achieves a couple of related things.

1. All compiler errors are moved into the common `sway-error` crate. The `Handler` type is now also moved there, and now uses `CompileError` instead of `ParseError`. That's also the main goal, to generalize `Handler` to all sorts of errors, which will help solve https://github.com/FuelLabs/sway/issues/2734 soon.

2. Using 1, we can now also emit `LexError`s into the `Handler`, so this enables lexer recovery.

3. Lexer recovery is added for e.g., `42y8`, which fixes https://github.com/FuelLabs/sway/issues/2865.

4. Lexer recovery is added for `'abc'` char literals, which fixes https://github.com/FuelLabs/sway/issues/2864. The AST supports them, but the rest of the compiler does not, so this will be useful later on, but was good for recovery framework testing.

5. Lexer recovery is added for mismatched open/close delimiters, e.g., `fn foo() { )`.

6. Lexer recovery is added for unexpected closing delimiters, e.g., `fn foo() }`.

7. Lexer recovery is added for unclosed delimiters.

8. Parser recovery is added for `let x =`, which fixes https://github.com/FuelLabs/sway/issues/2815.

9. Misc lexer refactoring and simplifications are made.

### DIFF
--- a/sway-error/Cargo.toml
+++ b/sway-error/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sway-error"
-version = "0.25.1"
+version = "0.25.2"
 authors = ["Fuel Labs <contact@fuel.sh>"]
 edition = "2021"
 homepage = "https://fuel.network/"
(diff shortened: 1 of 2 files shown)


sway-error/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "sway-error"
+version = "0.24.5"
```
mohammadfawaz, line 3: This should now be `0.25.0`

emilyaherbert (approved):

JoshuaBatty (approved):
This is great, thanks! :shipit:

JoshuaBatty (approved):

mitchmindtree (approved):
Great stuff!

PR #2892, opened 2022-09-29
Revert signed integers, ufp64

rostyslavtyshko:


mohammadfawaz (approved):

mohammadfawaz:
For context, these two libraries slow down the compilation of the standard library. Despite this being a solvable problem in the compiler and we will get to it eventually, it's better to keep *very large* libraries out of the standard library for now and keep them in `sway-libs` instead.

simonr0204 (approved):

Braqzen (approved):
Perhaps these should have been disabled rather than deleted, no?

mohammadfawaz:
> Perhaps these should have been disabled rather than deleted, no?

Potentially, but it's probably better to have them and maintain them in `sway-libs` (they could change in the coming weeks!) and then port back here when the compiler is ready.

PR #2894, opened 2022-09-29
Refactor manifest handling around `forc-pkg` to accommodate workspace manifest files

### LINKED ISSUE #2893: Introduce the concept of workspace manifest file for holding [workspace] information
related to #1777.

I think first step towards to implement `[workspace]` logic should be being able to differentiate (and hold) a package manifest file and a workspace manifest file similiar to `cargo`. So this issue is for tracking required refactor around or `manifest` logic in `forc-pkg`.

kayagokalp:
closes #2893.

~Note: I will be double checking the changes before marking it ready to review but other than that I feel like it is ready.~

With this PR, we reach a point where we can have a workspace manifest that lists members of the workspace and `forc-client` (more particularly `forc-deploy`) is able to deploy workspace members in the order of their declaration. I think this is a good starting point to iterate on to add fully fledged workspace support.


kayagokalp:
Would love some input whether this is a good approach especially from @mitchmindtree, we could also separate `Manifest` and `VirtualManifest` completely rather than merging them in a single enum as different variants. 🤔

mitchmindtree:
Thanks for diving into this @kaya!

It looks like the purpose of the "Virtual Manifest" is to act as a workspace manifest - is that right?

> we could also separate Manifest and VirtualManifest completely rather than merging them in a single enum as different variants. thinking

After a quick first look I think this sounds a lot better. Currently in `forc` our `Manifest` type is very specific to an individual package. I think we should avoid conflating the package manifest with the workspace manifest and try to separate the concerns best we can.

Maybe we can also:

1. Rename the existing `Manifest` type to `PackageManifest` as it's specific to a single package.
2. Following suit, rename `ManifestFile` to `PackageManifestFile`.
3. Rather than using `VirtualManifest`, use `WorkspaceManifest` to refer to the workspace manifest.

Perhaps we can have a design chat before going any further? I'll message you on slack!

kayagokalp:
Hey @mitchmindtree, I couldn't decide whether to strip all `forc-deploy` changes (which is basically utilizing the `WorkspaceManifest` introduced in this PR to detect and deploy all of the workspace members in the order of their declaration) from this PR since this might still be an improvement for application developers while we are brewing the full workspace support. So leaving this decision up to you and marking this as ready to go!

forc-plugins/forc-client/src/ops/deploy/op.rs
```diff
@@ -14,8 +14,10 @@ use tracing::info;
 
 use crate::ops::{deploy::cmd::DeployCommand, parameters::TxParameters};
 
-/// Checks if we have a WorkspaceManifest and if that is the case we will be deploying the packages
-/// listed in the members field of the WorkspaceManifest in the order of their declaration
+/// If the provided path (`DeployCommand.path`, or if it is not provided the current directory)
+/// contains a workspace manifest file, `deploy` will be deploying the workspace members in the
+/// order that their listing in the manifest file. If the provided path contains only a package
```
mitchmindtree: > `deploy` will be deploying the workspace members in the order that their listing in the manifest file

I think we'll need to make sure we actually deploy these in order of dependency, which we can get by doing a toposort of the subgraph of all `contract-dependencies` filtered down to those that exist in our workspace members set. Otherwise if we try to deploy a contract before one of its dependencies is deployed we'll run into problems :) We could require the user to specify these in order, but it would be nice if this weren't necessary.

I think this graph of contract dependencies is something that could be returned from the result of the `forc_pkgs::build` functions, so it might make sense to first address #2931 and get contract dependencies into the pkg graph before addressing this.

mitchmindtree (changes requested):
> I couldn't decide whether to strip all forc-deploy changes

Yeah I think stripping out the `forc-client` changes into a follow-up PR is a good idea - I think in general, to handle workspace members in the correct order it will be beneficial to have contract dependencies in the graph first.

forc-pkg/src/manifest.rs
```diff
@@ -14,19 +14,19 @@ use sway_utils::constants;
...
 type PatchMap = BTreeMap<String, Dependency>;
 
-/// A [Manifest] that was deserialized from a file at a particular path.
+/// A [PackageManifest] that was deserialized from a file at a particular path.
 #[derive(Clone, Debug)]
-pub struct ManifestFile {
+pub struct PackageManifestFile {
     /// The deserialized `Forc.toml`.
-    manifest: Manifest,
+    manifest: PackageManifest,
```
bingcicle, line 21: Will there be different files for declaring packages vs declaring a workspace? If they are still within `Forc.toml`, I think I'd find the `PackageManifestFile` naming to be a bit confusing :thinking:

kayagokalp: Just to clarify the current approach, I can summarize them like the following: 

With this approach package manifests (manifest files that are residing in packages) and workspace manifests are going to be two different files both named `Forc.toml` in different locations in the project structure as they are both manifest files for `forc`. Just like a `Cargo.toml` can be a virtual manifest file (which holds workspace information) at the root of the project structure and package manifest file that holds information about the package itself and resides inside the package's folder.

Since we have two different structures with different functionalities we have `PackageManifest` and `WorkspaceManifest` as different structures. These structures are direct mappings of these two types of manifests. A `PackageManifestFile` and `WorkspaceManifestFile` is a packing of these mappings with their corresponding paths, as throughout the flow of the `forc-pkg` operations we sometimes need to access both the manifest information and the path of that manifest information coming from. 

Since we changed name of the default `Manifest` structure to `PackageManifest` as we now have one more type, I also renamed the structure that holds both the `PackageManifest` and the path for that specific `Forc.toml` file to `PackageManifestFile`. The same logic goes for `WorksapceManifestFile` too. We either need to do it like this or merge them under a single enum and check & produce errors based on their types which was the approach at the beginning but it resulted in a fairly complex code so we decided to separate them as much as possible.

The logic behind the `PackageManifest` wording was the fact that these manifests are going to be inside a package and we are using the "package" name for referencing individual forc projects throughout `forc-pkg`.

These being said I am open for any suggestions too 🙌

bingcicle: Ahh okay, got it. Thank you for the detailed explanation!

mitchmindtree (approved):

Issue #2895, opened 2022-09-29
Update `fuel-vm` dependency


vlopes11:
Currently we can't deploy hotfixes to sway because it is using the version `0.15` of the fuel-vm, and the latest version is `0.18`

https://github.com/FuelLabs/sway/blob/15b9d4322f35d7fb33f885bffd8d816371b3e9ca/sway-core/Cargo.toml#L20

We need to update the version of the fuel-vm to be aligned with the same used in `fuel-core-interfaces`

https://github.com/FuelLabs/fuel-core/blob/ee8cd4269fcd3c5c3d20792b699ace9451863f63/fuel-core-interfaces/Cargo.toml#L18

This way, we will keep the stack consistent with the latest changes, improvements and fixes

Issue #2659, opened 2022-08-30
Using `fuel-tx` directly in `forc-pkg` causes some issues with `cargo install`


mohammadfawaz:
This is referring to https://github.com/FuelLabs/sway/pull/2482...

With this branch, the problem is reproducible as follows:
- Use `fuel-tx` directly in `forc-pkg` instead of through `fuel-gql-client`
- Run `cargo install --debug --path ./forc`. You get the following error:
![image](https://user-images.githubusercontent.com/59666792/187533941-d5bee8ec-6a49-4b59-a196-4e1ff3e9a84e.png)

Note that `cargo build` compiles without any issues. It's only `cargo install` that exposes the problem.

kayagokalp:
I am not sure but this might be happening because of a missing `serde` feature in the `fuel-tx` entry of `forc-pkg` manifest file. I had a similar issue while I was playing around at #2629 https://github.com/FuelLabs/sway/blob/184f6e4ec4a9a7b6a776531df7b04a68a5919a93/forc-pkg/Cargo.toml#L15

mohammadfawaz:
Oh interesting. I'll play around with that, thanks!

PR #2896, opened 2022-09-29
Bump a few dependencies

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/2895
Closes https://github.com/FuelLabs/sway/issues/2659

Not bumping everything here just yet. Bumping `sway-core` separately does not have side effects. Bumping versions in `forc-pkg` and `forc-plugins` is a bit more complicated because a new version of `fuel-core` is needed I believe.

Also bumping `fuels` in places outside the sdk tests


forc-plugins/forc-client/src/ops/run/op.rs
```diff
@@ -75,7 +75,7 @@ pub async fn run(command: RunCommand) -> Result<Vec<fuel_tx::Receipt>> {
         let inputs_to_add = locked_wallet
             .get_asset_inputs_for_amount(
                 AssetId::default(),
-                DEFAULT_SPENDABLE_COIN_AMOUNT,
+                1_000_000,
```
mohammadfawaz, line 78: I'm not sure about this. I've seen the whole thing removed in other places (mainly tests) but I don't know if that's safe to do here. cc @nfurfaro and @digorithm.

digorithm: Yeah, we've removed that `DEFAULT_SPENDABLE_COIN_AMOUNT` here https://github.com/FuelLabs/fuels-rs/pull/556.

nfurfaro (dismissed):
LGTM.

forc-pkg/src/pkg.rs
```diff
@@ -10,7 +10,7 @@ use forc_util::{
 };
 // Using `fuel_tx` directly instead of `fuel_gql_client` transitively is causing some weird issue.
 // See https://github.com/FuelLabs/sway/issues/2659
```
kayagokalp, line 12:
```suggestion
// Using `fuel_tx` directly instead of `fuel_gql_client` transitively is causing some weird issue.
// See https://github.com/FuelLabs/sway/issues/2659
```

Maybe we can also remove this as this is solving the mentioned issue.

kayagokalp (commented):
Just a single nit, I can approve and fix that in a later PR as I will have some refactoring related PR for `forc-pkg` 👍

forc-pkg/src/pkg.rs
mohammadfawaz: Good point! Removed.

kayagokalp (approved):
🚀

otrho (commented):
Do we bump `fuel-vm` for `sway/test` here too?
```
: rg fuel-vm
sway/test/Cargo.toml
18:fuel-vm = { version = "0.15", features = ["random"] }

sway/test/src/sdk-harness/Cargo.toml
13:fuel-vm = "0.15"
```

mohammadfawaz:
> Do we bump `fuel-vm` for `sway/test` here too?
> 
> ```
> : rg fuel-vm
> sway/test/Cargo.toml
> 18:fuel-vm = { version = "0.15", features = ["random"] }
> 
> sway/test/src/sdk-harness/Cargo.toml
> 13:fuel-vm = "0.15"
> ```

We may have to wait for a new version of `fuel-core` to do this because the tests directory relies on `forc-run` which relies on `fuels` and `fuels-core`. I tried various combinations but somehow always ended up needing a new version of `fuel-core`. Anyways, I think that's coming soon.

otrho (approved):
OK, I'm going to eventually need `fuel-vm` 0.18 with the `RETD` fix for #2885 to not need that hack... and then for the disabled test in `sdk-harness` in my calls PR... argh, painful.

Issue #2897, opened 2022-09-29
Revert with a special error code when `transfer_to_output` fails due to missing variable outputs


mohammadfawaz:
Basically this line https://github.com/FuelLabs/sway/blob/15b9d4322f35d7fb33f885bffd8d816371b3e9ca/sway-lib-std/src/token.sw#L108 should be `revert(SOME_SPECIAL_NUMBER)`

PR #2898, opened 2022-09-29
Add a special revert code for `transfer_to_output`

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/2897

Eventually, we should consolidate all the error codes into a separate file


digorithm (approved):
Nice :shipit:

sway-lib-std/src/token.sw
```diff
@@ -8,6 +8,8 @@ use ::identity::Identity;
 use ::revert::revert;
 use ::outputs::{Output, output_amount, output_count, output_type};
 
+const FAILED_TRANSFER_TO_OUTPUT_SIGNAL = 0xffff_ffff_ffff_0001;
```
nfurfaro, line 11: Looks reasonable, but what's the significance of the chosen hex value? Will this actually make it easier for devs to debug, or is this more for the SDK to decode and work with?

mohammadfawaz: Honestly 100% arbitrary. `0xffff_ffff_ffff_0000` just seemed like a reasonable _high_ offset for errors.

nfurfaro: This makes me think that the `Revert(42)` in `std::revert::require` would be a good candidate for a similarly-offset error value.

nfurfaro (approved):
Looks good.

sway-lib-std/src/token.sw
mohammadfawaz: True. Didn't want to touch it just yet because I don't know if it's been relied on anywhere right now. If not, we can certainly change it. I'll create an issue to clean up the errors and move them into a separate file or something.

PR #2909, opened 2022-09-30
Bitwise not for all unsigned integers

rostyslavtyshko:


mohammadfawaz:
I know this is a draft but some early comments :)
1. We probably shouldn't be using the name `binary_not` since this is not a "binary" operation. It's a unary operation. A better name would probably just be `bitwise_not`.
2. Because this is fairly similar to `not()` for `bool`, maybe it's worth creating a `Not` trait that bools AND integers use? @sezna what do you think?
3. If we go with (2), then `!` will automatically work for integers (I think) which means that the compiler will desugar `!5` into `not(5)` which is consistent with Rust. If we decide to keep a separate trait and separate method name (like `bitwise_not()` for integers instead, then we have to modify the compiler to also handle that case separately which gets annoying when we don't know whether something is a bool or an integer, which is why I recommend we go with (2).

simonr0204:
Hey @rostyslavtyshko , did you get round to looking at @mohammadfawaz 's suggestions above?

sezna:
> Because this is fairly similar to not() for bool, maybe it's worth creating a Not trait that bools AND integers use? @sezna what do you think?

Agree -- we should use this trait here.

mohammadfawaz:
Changing this to a draft while the comments are being address :)

mohammadfawaz:
@rostyslavtyshko 
* I updated the parser to now convert `!` to a the method application `not()` instead of the old function call `not()`. \
* Also merge with `master` and resolved conflicts. 
* Finally, I added two missing `.not()` in `core` and replaced hardcoded max values with `max()` from `core::num`.

test/src/e2e_vm_tests/test_programs/should_pass/language/bitwise_not/src/main.sw
```diff
@@ -0,0 +1,14 @@
+script;
+
+use core::ops::*;
+use std::assert::assert;
+use std::logging::log;
```
mohammadfawaz, line 5: I think none of these are needed. `assert` is also in the prelude.

rostyslavtyshko: Removed

mohammadfawaz (dismissed):
Looks good! One thing is maybe the test should be added to `stdlib/` instead of `language`, but that's not a big issue.

mohammadfawaz (dismissed):

simonr0204 (approved):

mohammadfawaz:

mohammadfawaz (approved):

simonr0204 (approved):

PR #2915, opened 2022-10-02
Remove unused `peekable` operations from `swayfmt`

### LINKED ISSUE #2496: Use `while let` in place of `for` in `peekable()` iterations in `sfv2`
```rust
let mut value_pairs_iter = value_pairs.iter().enumerate().peekable();
for (var_index, (type_field, comma_token)) in value_pairs_iter.clone() {}
```
should be:
```rust
let mut value_pairs_iter = value_pairs.iter().enumerate().peekable();
while let Some((var_index, (type_field, comma_token))) = value_pairs_iter.next() {}
```
There are many cases of this in `sway-fmt-v2`, more examples in #2338 comments

eureka-cpu:
Closes #2496


eureka-cpu:
@mitchmindtree hmm, I know this change was at your request initially but `clippy` seems to think we _should_ use a `for` loop 🤔

mitchmindtree:
> @mitchmindtree hmm, I know this change was at your request initially but clippy seems to think we should use a for loop thinking

Ahh, clippy is likely complaining as we're no longer using `peak`. Are you sure this PR keeps the semantics we're after?

Previously in these loops we were:

1. Cloning the iterator.
2. Using the cloned iterator to iterate over the elements.
3. Calling `peak` on the **old** iterator, which meant that *only the first element was ever peaked*.

The reason I suggested checking if we should be using `while let` is that step 3 looked suspicious to me. I.e. if we're using `peak`, the intention was likely to always peak the *next* element and not repeatedly `peak` the *first* element. Does that sound correct?

eureka-cpu:
> > @mitchmindtree hmm, I know this change was at your request initially but clippy seems to think we should use a for loop thinking
> 
> Ahh, clippy is likely complaining as we're no longer using `peak`. Are you sure this PR keeps the semantics we're after?
> 
> Previously in these loops we were:
> 
>     1. Cloning the iterator.
> 
>     2. Using the cloned iterator to iterate over the elements.
> 
>     3. Calling `peak` on the **old** iterator, which meant that _only the first element was ever peaked_.
> 
> 
> The reason I suggested checking if we should be using `while let` is that step 3 looked suspicious to me. I.e. if we're using `peak`, the intention was likely to always peak the _next_ element and not repeatedly `peak` the _first_ element. Does that sound correct?

I think I see now, I'll take another look tomorrow morning

eureka-cpu:
> @mitchmindtree is there a reason we wouldn't want to consume the iterator here? Looking again, I removed the `.peek()` because it removes the last newline which we actually want to keep. Adding the `.peek()` back in feels sort of redundant because when there isn't another field, we still want to add another newline.

I went ahead and removed `.peekable()` from `enum`, maybe this is the right direction?

mitchmindtree:
Ahhhh in that case, if there's no need for peek, we can switch back to consuming the iterator with `for` 👍

bingcicle (commented):
LGTM once the warning is ignored

kayagokalp (approved):
:shipit:

JoshuaBatty (approved):

PR #2913, opened 2022-10-01
Fix `doc` comments and apply to `lib std`

eureka-cpu:
Closes #2845

PR #2916, opened 2022-10-02
Missing changes from 2913

eureka-cpu:
Somehow the actual changes to the formatter didn't make it into my commit... Super weird. See #2913 which for some reason only applied the changes to `sway-lib-std`.


bingcicle (approved):
utACK

Braqzen (approved):

PR #2919, opened 2022-10-03
LSP synchronize workspace

### LINKED ISSUE #2847: Language server falls out of sync with users workspace while editing
@Braqzen noticed some strange syntax highlighting behavior which was noted in https://github.com/FuelLabs/sway-vscode-plugin/issues/92 

I was able to replicate this locally and have tracked down the cause of the problem. In the language server, we are calling `forc::check` on each didChange event. This fn takes the path of the manifest and then loads the project from disk to compile and return the parsed and typed AST's.

The problem is, if the user has not _saved_ the file in between each edit, then the `Span`s returned in the ASTs do not reflect the current edits that the user has made. This is then causing the capabilities such as, semantic tokens, warnings, go_to etc to have mismatched locations. 

I'm currently working on a solution that clones the workspace into a temp directory that LSP can use for processing. On every didChange event, we can write the file to disk in order to get an up-to-date directory in which to produce ASTs that reflect the local environment.

JoshuaBatty:
Previously, the language server was falling out of sync with the users current workspace edits. For example, incorrect syntax highlighting can be seen in the below image if the user creates a new line above the function. The syntax then takes on the previous highlighting and all other attributes like definition locations, warnings/error locations etc..

Screen Shot 2022-10-12 at 12 39 47 pm

This was happening because we load the project from disk and then get ASTs to work with from the compiler from these files. As the users workspace does not match the files saved on disk, then incorrect `Span`s are in the AST. To get around this, we are cloning the users workspace into a temporary directory on initialization of the server. On each keystroke or edit, we update the temp files and save to disk so that we can get an accurate AST back from the compiler for the language server to work with. The temp directory is synchronized with the users workspace on each did_save event. 

Finally, we spawn a tokio thread and use the `notify` crate to look for changes to the projects `Forc.toml` file. We then convert any dependencies that have a relative path to an absolute path.

closes #2847

~~There are a few places where we should avoid using `unwrap`s and return better errors. I've left `TODO` comments in the code where this should happen. This will be a lot easier once #2968 lands.~~
I've added error handling now as #2968 was merged before this one was approved.


otrho (dismissed):

Braqzen (commented):
Could tests be written to make sure the issue is fixed?

JoshuaBatty:
> Could tests be written to make sure the issue is fixed?

Sure, I can modify the current `did_change` test so that it can confirm the fix.

otrho (approved):

sway-lsp/src/test_utils.rs
```diff
@@ -1,15 +1,35 @@
+#![allow(unused)]
```
Braqzen, line 1: Shouldn't this be removed / are these used anywhere to make sure things work?

JoshuaBatty: The idea is for this file to have convenience functions for accessing different examples when developing the server. These used to be placed on top of each function but after moving these over to this file I figured it was less noisy to just have this statement at the top of the file. 

Not having this in the past had led to unnecessary changes on git when testing various things and this just seemed like a simpler approach to facilitate rapid testing.

Braqzen (approved):

otrho (approved):

Issue #2286, opened 2022-07-11
Use the `thiserror` crate and report any internal language server errors to the client


JoshuaBatty:
The language server could benefit from a more structured approach to handling and reporting errors. 

We should integrate the `thiserror` crate for defining the various types of errors that can occur in the language server. If an error does occur for any reason, we should report the error to the client and have it printed in the output window. 

Something along the lines of. `self.client.log_message(MessageType::ERROR, error_message).await;`

Care should be taken to make sure that errors are gracefully handled to avoid any panics so the server never shutdown.

JoshuaBatty:
closed with #2925 | #2968 | #3094

PR #2925, opened 2022-10-03
[LSP] Replace logging with tracing

sdankel:
Related https://github.com/FuelLabs/sway/issues/2286

Replacing the tower_lsp logging with calls to the `tracing` lib.

Added an option to `init_tracing_subscriber` to allow all tracing logs to be piped through stderr instead of split between stderr/stdio. Because the main thread in the LSP uses stdio to communicate with the LSP client, tracing logs must go through stderr.

image


JoshuaBatty:
Nice, this is going to be super handy to have! Looks like there are just a few clippy warnings that need addressing.

sway-lsp/src/lib.rs
```diff
@@ -9,8 +11,16 @@ mod server;
 pub mod test_utils;
 pub mod utils;
 use server::Backend;
+use tracing::metadata::LevelFilter;
 
 pub async fn start() {
+    let tracing_options = TracingSubscriberOptions {
+        log_level: Some(LevelFilter::DEBUG), // TODO: Set this based on IDE config
```
JoshuaBatty, line 18: sounds good, we can have a chat next week and I can talk you though how to go about setting something like this up.

sway-lsp/src/server.rs
```diff
@@ -186,10 +178,8 @@ impl Backend {
 #[tower_lsp::async_trait]
 impl LanguageServer for Backend {
     async fn initialize(&self, params: InitializeParams) -> jsonrpc::Result<InitializeResult> {
-        eprintln!("InitializeParams = {:#?}", params);
-        self.client
-            .log_message(MessageType::INFO, "Initializing the Sway Language Server")
-            .await;
+        tracing::error!("InitializeParams = {:#?}", params);
```
JoshuaBatty, line 181: The existing `eprintln!("InitializeParams = {:#?}", params);` was accidentally left over from a previous PR. Can we remove this line?

JoshuaBatty (changes requested):
LGTM, thanks for this! Just one thing to be removed and I'll be happy to approve.

JoshuaBatty (dismissed):

mitchmindtree (approved):

otrho (approved):

JoshuaBatty (approved):

PR #2936, opened 2022-10-04
Remove the `TypeInfo::Ref` variant

emilyaherbert:
This PR removes the `TypeInfo::Ref` variant, instead refactoring instances of where this variant was used with direct references to the `TypeId`'s themselves. This change is introduced (1) because `TypeInfo::Ref` has mostly become obsolete in the type system and was adding indirection, (2) this is one PR in a series of PR's in an effort to fix in a bug with the type system (TODO create the GH issue and link it), and (3) this is slowly crawling towards a type system where we can draw more direct comparisons between types (i.e. an eventual goal might be to use `&TypeInfo` directly, allocated on an arena)


mohammadfawaz (approved):
Seems reasonable, thanks!

PR #2829, opened 2022-09-22
Remove redundant call to `insert_type`

emilyaherbert:
This PR eliminates a redundant call to `insert_type` leading to extra `TypeId`'s being created.

This table shows the number of TypeId's created during type checking before this PR and after this PR. The number of TypeId's created for tests cases that import core or std will always include the number of TypeId's created for core or std. For example, std imports core, so 1,481 of the 164,779 are just TypeId's from core.

|                                                                                 | before #2827:   | after #2827:  | after this PR:  |
|---------------------------------------------------------------------------------|-----------|---------|---------|
| core                                                                            | 57,570    | 3,078   | 1,481   |
| std (w/ core)                                                                   | 1,019,669 | 170,349 | 164,779 |
| should_pass/language/generic_type_inference (w/ core + std)                     | 1,019,741 | 170,421 | 164,839 |
| should_pass/language/funcs_with_generic_types (w/ core)                         | 57,969    | 3,429   | 1,756   |
| should_pass/language/fixing_generic_type (w/ core + std)                        | 1,042,578 | 179,518 | 173,653 |
| should_pass/language/generic_impl_self (w/ core)                                | 61,815    | 6,019   | 4,228   |
| should_pass/language/match_expressions_inside_generic_functions (w/ core + std) | 1,026,071 | 170,823 | 165,090 |
| should_pass/language/generic_structs (w/ core)                                  | 57,746    | 3,254   | 1,633   |
| should_pass/language/generic_enum (w/ core)                                     | 57,606    | 3,110   | 1,505   |
| should_pass/language/generic_inside_generic (w/ core)                           | 58,272    | 3,220   | 1,599   |
| should_pass/language/array_generics (w/ core)                                   | 57,728    | 3,232   | 1,598   |
| should_pass/language/generic_functions (w/ core)                                | 57,696    | 3,204   | 1,580   |
| should_pass/language/generic_struct (w/ core)                                   | 57,634    | 3,142   | 1,534   |

PR #2938, opened 2022-10-04
Introduce `unify_right`

emilyaherbert:
This PR fixes a bug found by both @tritao and myself related to the separate implementations of #2867 and #2786. The bug involved wackiness in the type system when calling nested generics, resulting in some generic arguments resolving preemptively. For example, this Sway code:

```rust
fn third_if<D>(value: D) -> u8 {
  if true {
    5u8
  } else {
    revert(0);
  }
}

fn second_if<E>(value: E) -> bool {
  let third = third_if(value);
  if third == 1u8 || third == 2u8 || third == 3u8 || third == 4u8 {
    false
  } else if third == 5u8 {
    true
  } else {
    false
  }
}

fn first_if<F>(value: F) -> u64 {
  let second = second_if(value);
  if second == false {
    2u64
  } else {
    3u64
  }
}

fn main() -> u64 {
  let c = first_if(true);
  assert(c == 3);

  let d = first_if(1u8); // error created here because it expected first_if to take a boolean
  assert(d == 3);

  1
}
```

`first_if` would resolve preemptively to `first_if<bool>`, causing an error at the function application `first_if(1u8)`.

Previously, function application would use `unify` to unify the types of the arguments and the parameters. At a high level, `unify` takes two arguments, a `received` type and an `expected` type, and tries to unify `received` and `expected`. `unify` was working fine during function application in most cases, except for the fact that `unify` is slightly left-biased, meaning it prefers to unify upon `received` before unifying upon `expected`. This created situations where, given a `received` type and a different `expected` `TypeInfo::UnknownGeneric { .. }` type, `unify` would replace `received` with `expected`, in certain cases, causing the bug.

This PR introduces a new function `unify_right` that acts similarly to `unify`, except it can only unify upon the RHS argument, i.e. the `expected` argument. I've included a test case in `should_fail` that demonstrates the effect of this new functionality---this test cases was passing before this PR and now correctly fails. For reference here is the same code in Rust: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=41e6df297e37b72d7a8b7c32e43706d0


mohammadfawaz:
Do you mind adding a descriptive title? :)

otrho:
Bah, _all_ PRs should be 'Fix bug' from now on, I say!

otrho (commented):
It's hard to tell from the diff -- is it not enough to just swap the order of the args to `unify()` rather than using `unify_right()`?  There's a lot of duplication going on there.

emilyaherbert:
> Do you mind adding a descriptive title? :)

Oops sorry haha, I meant to replace that.

emilyaherbert:
> It's hard to tell from the diff -- is it not enough to just swap the order of the args to `unify()` rather than using `unify_right()`?

No, unfortunately not. I'd be happy to discuss other solutions regarding architecture, but `unfiy_right` needs to _exclude_ this case, which `unify` _includes_, so swapping the arguments wouldn't be sufficient: https://github.com/FuelLabs/sway/blob/37422d04df4b02babf9710236756ab878eeb5be7/sway-core/src/type_system/type_engine.rs#L467.

That would also cause issues downstream where simple mismatched function application argument errors would appear to be caused from the function parameters.

otrho (dismissed):

sezna (dismissed):
Could this be related to the removal of lots of `insert_type`s? I used to use `insert_type` when I wanted to unify types `a` and `b` without actually modifying `a`. So you could do `unify(insert_type(a), b)` or something to prevent the original copy from being modified. 

Either way this is a better way to handle it w/ less type engine work, so 🎊

emilyaherbert:
> Could this be related to the removal of lots of insert_types? I used to use insert_type when I wanted to unify types a and b without actually modifying a. So you could do unify(insert_type(a), b) or something to prevent the original copy from being modified.

Yep! This bug popped up when #2829 and #2827 went in.

mohammadfawaz (dismissed):

mohammadfawaz (approved):

sezna (approved):

PR #2941, opened 2022-10-05
Rename `transfer_to_output` to `transfer_to_address`

### LINKED ISSUE #2912: Consider renaming `transfer_to_output` to `transfer_to_address`
should `transfer_to_output` be renamed to `transfer_to_address`? That way we would have
1. `transfer_to_address`
2. `force_transfer_to_contract` (don’t like force here but I understand why it’s needed)
3. `transfer`
We also have `mint_to_address` so that would somewhat be parallel to `transfer_to_address`

mohammadfawaz:
Closes #2912

Even though `transfer_to_output` is technically more correct, users really understand this as transferring to an address. Besides, the name `transfer_to_address` aligns better with the names of its parameters:

```rust
(amount: u64, asset_id: ContractId, to: Address)
```

And also with other functions in `token.rs` (such as `mint_to_address`).

I also did some minor refactoring of the function.

### DIFF
--- a/examples/identity/src/main.sw
+++ b/examples/identity/src/main.sw
@@ -17,7 +17,7 @@ use std::{
     },
     token::{
         force_transfer_to_contract,
-        transfer_to_output,
+        transfer_to_address,
     },
 };
 
@@ -48,7 +48,7 @@ impl IdentityExample for Contract {
 
         // ANCHOR: different_executions
         match my_identity {
-            Identity::Address(identity) => transfer_to_output(amount, token_id, identity),
+            Identity::Address(identity) => transfer_to_address(amount, token_id, identity),
             Identity::ContractId(identity) => force_transfer_to_contract(amount, token_id, identity),
         };
         // ANCHOR_END: different_executions
--- a/examples/identity/src/main.sw
+++ b/examples/identity/src/main.sw
@@ -48,8 +48,8 @@ impl IdentityExample for Contract {
 
         // ANCHOR: different_executions
         match my_identity {
-            Identity::Address(identity) => transfer_to_address(amount, token_id, identity),
-            Identity::ContractId(identity) => force_transfer_to_contract(amount, token_id, identity),
+            Identity::Address(address) => transfer_to_address(amount, token_id, address),
+            Identity::ContractId(contract_id) => force_transfer_to_contract(amount, token_id, contract_id),
         };
         // ANCHOR_END: different_executions
     }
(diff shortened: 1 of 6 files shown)


Braqzen (approved):
Just noting that this is a breaking change so a new major release is required

mohammadfawaz:
> Just noting that this is a breaking change so a new major release is required

Yup! Draft PR for that up: https://github.com/FuelLabs/sway/pull/2942

examples/identity/src/main.sw
```diff
@@ -48,7 +48,7 @@ impl IdentityExample for Contract {
 
         // ANCHOR: different_executions
         match my_identity {
-            Identity::Address(identity) => transfer_to_output(amount, token_id, identity),
+            Identity::Address(identity) => transfer_to_address(amount, token_id, identity),
```
nfurfaro, line 51: I find this example slightly misleading. the inner value name below would be better as just `a` or `addr` or similar. Having the value named `identity` could be confused with the `Identity` type.
```
Identity::Address(identity) => transfer_to_address(amount, token_id, identity),
```

sway-lib-std/src/token.sw
```diff
@@ -84,33 +84,24 @@ pub fn force_transfer_to_contract(amount: u64, asset_id: ContractId, to: Contrac
...
-    let mut output_found = false;
 
     // If an output of type `OutputVariable` is found, check if its `amount` is
     // zero. As one cannot transfer zero coins to an output without a panic, a
     // variable output with a value of zero is by definition unused.
-    let outputs = output_count();
-    while index < outputs {
-        let type_of_output = output_type(index);
-        if let Output::Variable = type_of_output {
+    while index < output_count() {
```
nfurfaro, line 94: Does `output_count()` get called for each iteration of the loop here?
Or does the compiler somehow cache this value without the need to explicitly cache it in a temp var like `outputs` ?

nfurfaro (commented):
A couple of comments, but looks good, and improves the token module usability.

sway-lib-std/src/token.sw
Braqzen: I recall comments about the compiler being smart enough to see if something is `read-only` or something. The implication was that if something does not change in some order of operations then repeating the value / call can have some pruning so that it's not called over and over. Not my area though

mohammadfawaz: Hmmm that's actually a very good question. I don't know. What I know is that we most definitely should *not* be doing that, at least in the future, once all of our loop optimizations are implemented. For that reason, I will revert this to how it was to be on the same side. Thanks!

nfurfaro (approved):
LGTM!

Braqzen (approved):
Use the force, @mohammadfawaz . Force push.

PR #2950, opened 2022-10-05
Error messages distinguish abi and trait methods

### LINKED ISSUE #2420: Enhance error message for when a contract method is implemented but is not part of the ABI
Example:
```rust
contract;

abi MyContract {
}

    fn test_function() -> bool {
        true
    }
}
```
We currently get the following error:
```console
6 | impl MyContract for Contract {
7 |     fn test_function() -> bool {
  |        ^^^^^^^^^^^^^ Function "test_function" is not a part of trait "MyContract"'s interface surface.
8 |         true
9 |     }
  |
```
which doesn't make much sense because `MyContract` is not a trait. ABIs and traits are largely similar in the compiler but we should have more tailored diagnostic messages for ABIs.

anton-trunov:
close #2420

### DIFF
--- a/sway-core/src/error.rs
+++ b/sway-core/src/error.rs
@@ -457,6 +457,21 @@ impl fmt::Display for Warning {
     }
 }
 
+#[derive(Error, Debug, Clone, PartialEq, Eq, Hash)]
+pub enum InterfaceName {
+    Abi(Ident),
+    Trait(Ident),
+}
+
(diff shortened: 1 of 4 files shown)


mohammadfawaz (approved):
Everything seem reasonable here. Thank you!

vaivaswatha (approved):

PR #2954, opened 2022-10-05
Rename the `Typed` nodes to `Ty` nodes

### LINKED ISSUE #2911: Rename the `Typed` nodes to `Ty` nodes
When a node is a `Typed` node, it indicates that it is fully typed. But when a node is a `Ty` node, it instead indicates that a node is _typeable_. This distinction is important for disentangling AST transformation from type inference and type checking.

emilyaherbert:
This PR renames all of the nodes that start with `Typed` to `Ty`, for instance `TypedExpression` has been changed to `TyExpression`.

closes #2911


mitchmindtree (commented):
If we're to go this route, perhaps we could also consider `Expression` -> `Expr`, `Declaration` -> `Decl`, `Context` -> `Ctx`, `Function` -> `Fn` etc too?

Currently we have quite verbose naming as a general trend in `sway-core`. I'd be happy to switch to a more terse approach generally (as long as the names remain unambiguous) but we might want to be consistent on which direction we go down? I.e. either clear shorthand (like Rust's std lib, and the direction of this PR), or verbose and descriptive (like Java BeanFactories :smiling_face_with_tear:)?

mitchmindtree (commented):
Ahhhh disregard my last comment! I just read through #2911 properly and realised the motivations behind this are totally different. This is more about distinguishing between type-able and typed nodes.

mohammadfawaz:
So are we later going to introduce some `Typed` nodes once we disentangle type inference and type checking?

emilyaherbert:
> Currently we have quite verbose naming as a general trend in sway-core. I'd be happy to switch to a more terse approach generally (as long as the names remain unambiguous) but we might want to be consistent on which direction we go down? I.e. either clear shorthand (like Rust's std lib, and the direction of this PR), or verbose and descriptive (like Java BeanFactories 🥲)?

@mitchmindtree I'm definitely a fan of shorter, more uniform naming. I'd support an RFC for this 😄

emilyaherbert:
> So are we later going to introduce some Typed nodes when we disentangle type inference and type checking?

We certainly could, although I wasn't really planning to include it in the immediate roadmap. I think doing an ast transformation from `Ty` to a new `Typed` language before ir gen has advantages (if the ast transformation is successful, then you can be assured that any functions that take a `Typed` node are taking a _fully typed node_, i.e. once you have a `Typed` node you can eliminate all checks for well-typed-ness in ir gen). But, just changing the old `Typed` definitions to `Ty` doesn't affect the fact that ir gen still does some lifting to check for well-typed-ness, regardless of if the old `Typed` definitions or the new `Ty` definitions are used.

In other words, we could say that the old `Typed` definitions were a bit of a misnomer, as we couldn't actually be sure that a `Typed` definition was well-typed (i.e. why we needed to check for lingering generics). So, even in the shorterm without introducing a new `Typed` definition, changing to `Ty` is a more reflective name.

mohammadfawaz (approved):

PR #2955, opened 2022-10-05
Create a `language` module

emilyaherbert:
This PR:
- creates the `language` module
- moves the `parse_tree` module into `language` and renames it `parsed`
- moves the `literal`, `call_path`, `lazy_op`, `purity`, `asm`, and `visibility` modules out of the `parsed` module and into the `language` module because they are used by more than just the `parsed` module
- *the next PR will create a `ty` module inside of `language` that contains the `TyProgram` nodes and stuff but I'm splitting it up into multiple PR's for readability*


otrho:
I don't think `asm_lang` belongs in there.  The rest is all front-end where `asm_lang` is back-end.  If anything we should put `asm_lang` and `asm_generation` in their own module or even a new crate.

emilyaherbert:
> I don't think asm_lang belongs in there. The rest is all front-end where asm_lang is back-end. If anything we should put asm_lang and asm_generation in their own module or even a new crate.

Makes sense! I'll move it out.

mohammadfawaz:
I wonder if, long term, a lot of these things should be pulled out of `sway-core`. For example, there is no reason why `forc-pkg` should depend on `sway-core`. Data crates are a better fit for data structures that are required by `sway-core` *and* other crates. This would be similar to what @Centril has done recently with the `sway-ast` crate.

mohammadfawaz (approved):

otrho (approved):
> I wonder if, long term, a lot of these things should be pulled out of sway-core... This would be similar to what @Centril has done recently with the sway-ast crate.

Eventually there'll be no `sway-core` left. :slightly_smiling_face:

PR #2956, opened 2022-10-06
Update finalized asm example in Sway Book

mohammadfawaz:
The finalized assembly looks a little different now. Also switching to using `cargo test` instead of `forc test` as I think `forc test` will be repurposed for in-Sway testing.


otrho (approved):
I forgot I changed the mismatched selector revert code to 123.  It was annoying me as I tested things that `revert(0)` was used so much -- I'd have to dig into the bytecode to work out which `RVRT` was actually triggering.

docs/src/introduction/forc_project.md
```diff
@@ -55,36 +55,37 @@ We now compile our project with `forc build`, passing the flag `--print-finalize
...
-data_1 .u32 0x2151bd4b
+data_0 .word 559005003
 
   Compiled contract "my-fuel-project".
-  Bytecode size is 68 bytes.
+  Bytecode size is 60 bytes.
 ```
 
-To test this contract, use `forc test`:
+To test this contract, use `cargo test`:
```
mitchmindtree, line 80: I've removed from this line onwards in https://github.com/FuelLabs/sway/pull/2264 - would you mind reverting the unrelated lines that touch the test parts to avoid conflicts?

mohammadfawaz:
> I forgot I changed the mismatched selector revert code to 123. It was annoying me as I tested things that `revert(0)` was used so much -- I'd have to dig into the bytecode to work out which `RVRT` was actually triggering.

That's probably a good change. Not sure what value we should select though 😆  - We've started using these code in the standard library: `0xffff_ffff_ffff_0001`, `0xffff_ffff_ffff_0002`, and so on, so maybe `0xffff_ffff_ffff_0000` is a good fit for the selector revert.

otrho (approved):

emilyaherbert (approved):

PR #2957, opened 2022-10-06
Replace PHIs with block arguments

### LINKED ISSUE #2775: Remove PHIs and introduce block args
The Sway-IR currently limits basic blocks to have only a single PHI node. That wouldn't work, for example, in the following case:
```
let mut x;
let mut y;
if (...) {
    x = 10;
    y = 11;
} else {
    x = 11;
    y = 10;
}
merge_block:
     = x;
     = y;
```
At the merge block we'll need two PHIs, one for `x` and another for `y`.

We have two choices:
1. Remodel the IR to allow multiple PHIs and teach the various IR transformations / analyses about it.
2. Go for the "more modern" [block argument](https://mlir.llvm.org/docs/Rationale/Rationale/#block-arguments-vs-phi-nodes) approach.

Working with multiple PHIs is a bit clumsy, and in our case, since there is an assumption throughout the IR that there is only a single PHI, making changes to it to accept multiple PHIs has the risk of breaking this assumption somewhere. It's safer to just remove PHIs altogether and move to block arguments.

vaivaswatha:
Closes #2775.


sway-ir/src/instruction.rs
```diff
@@ -36,16 +35,16 @@ pub enum Instruction {
     /// Cast the type of a value without changing its actual content.
     BitCast(Value, Type),
     /// An unconditional jump.
-    Branch(Block),
+    Branch((Block, Vec<Value>)),
```
otrho, line 38: I'm looking at where we use these branch variants and seeing `block.1.iter()...` or `block_to_label(&block.0)` and I think it's a little confusing.  Instead of a pair I think it should be a struct, so we can use `branch.block` and `branch.args`.

sway-ir/src/function.rs
```diff
@@ -59,18 +60,11 @@ impl Function {
...
-                    name,
-                    Value::new_argument(context, ty).add_metadatum(context, arg_metadata),
-                )
-            })
-            .collect();
         let content = FunctionContent {
             name,
-            arguments,
+            // Arguments to a function are the arguments to its entry block.
+            // We set it up after creating the entry block below.
```
otrho, line 66: Ahh, cool. :slightly_smiling_face:

sway-ir/src/verify.rs
```diff
@@ -292,13 +296,13 @@ impl<'a> InstructionVerifier<'a> {
         Ok(())
     }
 
-    fn verify_br(&self, dest_block: &Block) -> Result<(), IrError> {
-        if !self.cur_function.blocks.contains(dest_block) {
+    fn verify_br(&self, dest_block: &(Block, Vec<Value>)) -> Result<(), IrError> {
+        if !self.cur_function.blocks.contains(&dest_block.0) {
             Err(IrError::VerifyBranchToMissingBlock(
-                self.context.blocks[dest_block.0].label.clone(),
+                self.context.blocks[dest_block.0 .0].label.clone(),
```
otrho, line 302: Here's an example too.  `self.context.blocks[dest_branch.block.0].label.clone()` would be more readable.

sway-ir/tests/dce/dce1.cfg
```diff
@@ -2,14 +2,14 @@
 
 script {
     fn main() -> bool {
-        entry:
+        entry():
```
otrho, line 5: Why is this file `dce1.cfg`, instead of `dce1.ir`?

otrho (changes requested):
This is awesome.

sway-ir/src/instruction.rs
vaivaswatha: > I'm looking at where we use these branch variants and seeing `block.1.iter()...` or `block_to_label(&block.0)` and I think it's a little confusing. Instead of a pair I think it should be a struct, so we can use `branch.block` and `branch.args`.

Fixed this now.

sway-ir/tests/dce/dce1.cfg
vaivaswatha: > Why is this file `dce1.cfg`, instead of `dce1.ir`?

Fixed this now.

sway-ir/src/block.rs
```diff
@@ -10,13 +10,16 @@
 //!
 //! Every [`Function`] has at least one block, the first of which is usually labeled `entry`.
 
+use rustc_hash::FxHashSet;
```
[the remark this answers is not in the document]
vaivaswatha: > As a pre-emptive optimisation? I assume we could replace all uses of `HashMap` and `HashSet` throughout the crate, if it makes a noticeable difference.

Let's do this as a separate PR. As far as I can tell, for simpler key types, `FxHashSet` is supposed to be faster, but not for others. So we'll need to check this manually and replace.

otrho (approved):

otrho (dismissed):

sway-ir/src/block.rs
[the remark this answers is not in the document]
vaivaswatha: Another reason that I chose to use it is that it isn't randomised, and hence deterministic.

vaivaswatha:
Thank you for the quick review @otrho

sway-ir/src/block.rs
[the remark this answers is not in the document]
otrho: Oh, that's helpful.  Fast and deterministic. :+1:

otrho (approved):

sway-core/src/asm_generation/asm_builder.rs
```diff
@@ -492,33 +491,44 @@ impl<'ir> AsmBuilder<'ir> {
...
+        ok((), vec![], vec![])
     }
 
-    fn compile_branch_to_phi_value(&mut self, from_block: &Block, to_block: &Block) {
-        if let Some(local_val) = to_block.get_phi_val_coming_from(self.context, from_block) {
-            // We only need a MOVE here if get_phi_val_coming_from() is actually assigned to a
-            // register
-            if let Some(local_reg) = self.opt_value_to_register(&local_val) {
-                let phi_reg = self.value_to_register(&to_block.get_phi(self.context));
+    fn compile_branch_to_phi_value(&mut self, to_block: &BranchToWithArgs) {
```
mohammadfawaz, line 522: Does it still make sense to call this function `compile_branch_to_phi_value`? Same goes for `phi_reg` below.

mohammadfawaz (approved):
Seems very reasonable. I've never worked with block arguments before so I'm looking forward to dipping my toes 😄

sway-core/src/asm_generation/asm_builder.rs
vaivaswatha: I considered renaming it, but couldn't think of a better name, and conceptually, it's still a PHI. So I let it be.

Issue #2929, opened 2022-10-04
Refactor `forc-client` to be aware of workspace manifest files


kayagokalp:
After #2893 is addressed we are going to have a workspace manifest file that holds various information for that workspace. As a first step going towards full workspace support we may want to do some refactoring around `forc-client` so that it can be aware of the workspace manifest file. Until we add the deployment dependencies to the workspace manifest file, we can deploy the members of the workspace as they are declared.

mitchmindtree:
> Until we add the deployment dependencies to the workspace manifest file

Just a note to add that we'll want to do this in the package manifest (not the workspace manifest) as mentioned here https://github.com/FuelLabs/sway/issues/2931#issuecomment-1267775595.

@kayagokalp we might be better off tackling #2931 first and leave any necessary refactoring of `forc-client` until after this. If we end up outputting the contract ID during `forc build` instead as mentioned [here](https://github.com/FuelLabs/sway/issues/2930#issuecomment-1267800047), then we shouldn't need to do much in `forc-client` at all other than:

1. Refactor `deploy` and `run` to use the contract ID from the `out` directory (or directly from the `forc_pkg` build function call used within the `deploy` function).
2. Add support for `deploy`ing all contract packages within a workspace at once.

kayagokalp:
Ahh just realized this exists after I created #3156. Since we longer have 

> Refactor deploy and run to use the contract ID from the out directory (or directly from the forc_pkg build function call used within the deploy function).

section of your summary as we handled that with introducing `contract-dependencies` at https://github.com/FuelLabs/sway/pull/3016. And second part of this issue is tracked by newer smaller issue #3156. I am closing this one but if you feel like tracking with this makes more sense I can reverse

PR #2958, opened 2022-10-06
feat: Output contract-id to .out folder with forc build

### LINKED ISSUE #2930: Introduce a way for `forc-build` to output deployment information
Currently the information related to a specific deployment is only `print`ed and we do not have a non-volatile way of storing this information. We may want to output something like a deployment file which is output in a predetermined structure so that later on workspace members can be aware of some deployment information (like the contract id) of other members so that we can send that information in between them.

kayagokalp:
closes #2930.


mohammadfawaz (approved):
This all seems reasonable but I wonder if we can reuse this code in https://github.com/FuelLabs/sway/blob/4740fb2f9b9c4c4d6e8ba2110831604b40ccc645/forc-plugins/forc-client/src/ops/deploy/op.rs#L148 and wherever else we compute the contract ID in that file.

mitchmindtree:
@mohammadfawaz good spot! The plan is to eventually have `forc-client` use this :+1: https://github.com/FuelLabs/sway/issues/2929#issuecomment-1267824974

forc-pkg/src/pkg.rs
```diff
@@ -2118,12 +2118,25 @@ pub fn build_with_options(build_options: BuildOptions) -> Result<Compiled> {
...
             let res = if minify_json_storage_slots {
-                serde_json::to_writer(&file, &compiled.storage_slots)
+                serde_json::to_writer(&storage_slots_file, &compiled.storage_slots)
             } else {
-                serde_json::to_writer_pretty(&file, &compiled.storage_slots)
+                serde_json::to_writer_pretty(&storage_slots_file, &compiled.storage_slots)
             };
+            // Construct the contract ID
+            let contract = Contract::from(compiled.bytecode.clone());
+            let salt = fuel_tx::Salt::new([0; 32]);
```
mitchmindtree, line 2129: Maybe we can open an issue for specifying the salt via flag?

mitchmindtree (approved):
:ok_hand:

PR #2968, opened 2022-10-07
Move DocumentError to use thiserror

sdankel:
Related https://github.com/FuelLabs/sway/issues/2286

- adds `thiserror` to sway LSP and moves the existing `DocumentError` types over
- adds `path` or `dir` to the document errors
- logs the parsing errors using the LSP logging interface
- adds 2 new error types, distinguishing between BuildPlan, Parse, and Compile errors
- adds `ManifestFileNotFound` error
- refactors functions to use `map_err` pattern instead of `match`
- adds tests for the DocumentErrors


sway-lsp/src/error.rs
```diff
@@ -0,0 +1,25 @@
...
+#[derive(Debug, Error)]
+pub enum LanguageServerError {
+    #[error(transparent)]
+    DocumentError(#[from] DocumentError),
+
+    #[error("Failed to create build plan. {0}")]
+    BuildPlanFailed(anyhow::Error),
+    #[error("Failed to compile. {0}")]
+    FailedToCompile(anyhow::Error),
+    #[error("Failed to parse document. {:?}", diagnostics)]
```
JoshuaBatty, line 13: I think just `Failed to parse document` is fine here instead of also printing the diagnostics. The user can already see the warnings and errors reported from the diagnostics in the editor, so printing them to the console as well might make it overly noisy.

sway-lsp/src/core/session.rs
```diff
@@ -328,3 +330,59 @@ impl Session {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::env;
+
+    use super::*;
+
+    fn sway_workspace_dir() -> PathBuf {
```
JoshuaBatty, line 340: Perhaps it's worth making a `test_utils.rs` file under the new tests folder that contains useful functions like this and  `get_absolute_path` below. Then we don't need to duplicate them in other tests such as in the `document.rs` file above.

sway-lsp/src/core/session.rs
```diff
@@ -122,132 +123,133 @@ impl Session {
...
+                Err(DocumentError::DocumentAlreadyStored { path: uri })
+            })
     }
 
     pub fn remove_document(&self, url: &Url) -> Result<TextDocument, DocumentError> {
-        match self.documents.remove(url.path()) {
-            Some((_, text_document)) => Ok(text_document),
-            None => Err(DocumentError::DocumentNotFound),
-        }
+        self.documents
```
JoshuaBatty, line 135: nice :)

JoshuaBatty (changes requested):
Nice work Sophie! This is already so much better. Just left a few little nits. Happy to see this merged once those are addressed.

sway-lsp/src/core/session.rs
sdankel: I was thinking this too, will do.

JoshuaBatty (approved):

kayagokalp (approved):
lgtm

Issue #2948, opened 2022-10-05
Type arguments on method calls are not able to parse


emilyaherbert:
This Sway code:

```rust
script;

struct Data {
    value: u64
}

impl Data {
    fn inner_generic<T>(self) -> u64 {
        __size_of::<T>()
    }
}

fn main() -> u64 {
    let data = Data { 
        value: 42u64
    };
    data.inner_generic::<bool>()
}
```

gives this error:

```bash
   |
15 | 
16 |     };
17 |     data.inner_generic::<bool>()
   |                       ^ Unexpected token in statement.
18 | }
   |
____

  Aborting due to 1 error.
```

PR #2974, opened 2022-10-07
Support type arguments on methods

Centril:
Fixes https://github.com/FuelLabs/sway/issues/2948.

Now it's possible to write `foo.bar::<Baz>()`.
The parser will also do recovery for `foo.field::<Args>`.


emilyaherbert (approved):
Looks great!

otrho (dismissed):

otrho:
Problem with the E2E test lock file.  Need to make it work with `--locked`...

otrho (approved):

mitchmindtree (approved):

PR #2976, opened 2022-10-07
Remove `byte` type and associated literal

### LINKED ISSUE #1936: Consider removing the type `byte`
It is not really documented in the Sway Book and it's pretty similar to a `u8`. One added benefit to `byte` is that it doesn't cast to other integer types like `u8`.

Edit: actually we do mention `byte` in passing here: https://fuellabs.github.io/sway/latest/basics/built_in_types.html#numeric-types
> All of the unsigned integer types are numeric types, and the _**byte**_ type can also be viewed as an 8-bit unsigned integer.

anton-trunov:
close #1936

### DIFF
--- a/docs/src/basics/built_in_types.md
+++ b/docs/src/basics/built_in_types.md
@@ -20,7 +20,7 @@ All other types in Sway are built up of these primitive types, or references to 
 
 ## Numeric Types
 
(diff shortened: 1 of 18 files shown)


mohammadfawaz (approved):
Great! Thank you :)

emilyaherbert (approved):

PR #2979, opened 2022-10-10
Forward attributes from `ParseProgram` to `TyProgram`

mitchmindtree:
This is useful for both upcoming `forc doc` (`#[doc]`) and `forc test` (`#[test]`) behaviour.

`forc test` requires this for in-language unit tests as it needs access to the attributes of the function declaration during dead code analysis. This is so that we can add the test functions as entry points and avoid producing dead function warnings for all `#[test]` functions.

@eureka-cpu also mentioned that this would be useful for `#[doc]` support, though I don't quite remember the details.

Following parsing, each item's set of `Attribute`s remains unmodified and used in a read-only manner throughout the rest of compilation. This PR also moves the `HashMap` used to store the attributes behind an `Arc` to cut down on memory usage slightly.


sezna:
This simplifies forc doc very much, sweet

Issue #1832, opened 2022-06-02
`#[test_script]` for unit testing


adlerjohn:
Add a `#[test_script]` that defines a script with no input parameters and an optional output value. Only valid for `contract`s. The compiler must:

1. compile the contract entirely, then
1. compile each script test independently, allowing implicit use of any ABI declarations the contract implements, then
1.  output each script test's compiled output to a different `.bin` file.

This will be sufficient for downstream tools such as `forc test` to pick them up and automatically run them, as a first step to in-language testing.

---

*Following added by @mitchmindtree.*

## TODO

- [x] #2979.
- [x] #2980.
    - [x] Support the `#[test]` attribute on Sway functions.
    - [x] Add `#[test]` to set of known attributes.
    - [x] Update original `compile` function (i.e. `forc build`) to ignore all `#[test]` functions. Likely best done during `convert_parse_tree`.
- [ ] Update `sway-core` to provide an alternative to `compile` called `compile_tests` that lazily generates bytecode for each test function, treating each as an entry point.
    1. Type check full program (including `#[test]` functions).
    2. Add `#[test]` functions as entry points for dead code analysis.
    3. Generate bytecode for each `#[test]` function entrypoint.
- [ ] Add a `forc test` command that executes the bytecode.
- [ ] Add a `forc check --tests` flag for use by `sway-lsp` to type-check all tests (without codegen or running them).

## Follow-up

- Refactor E2E tests to take advantage of new `#[test]` feature where possible.
- Consider moving some std lib unit tests back to std library and use `#[test]`, rather than in E2E tests?
- Introduce `[dev-dependencies]` table to `Forc.toml`.

mitchmindtree:
@adlerjohn are you happy to go with `#[test]` for this over `#[test_script]`? I don't recall if there was a particular motivation for `#[test_script]` instead :thinking:

adlerjohn:
The motivation is that there needs to be a way to write tests that involve running a script transaction and calling one or more ABI methods, and a way to run a unit test that `just happens` without having to go through a transaction.

mitchmindtree:
Ahhh yes, perhaps we could go with something like this?

- `#[test]` - base test declaration, "just happens".
- `#[test(script)]` - indicates a script transaction is required in order to call one or more ABI methods.

[...]

PR #2980, opened 2022-10-10
Add `#[test]` to list of known attributes in `sway-core`

mitchmindtree:
First steps of #1832, this adds basic awareness to the compiler for the new `#[test]` attribute.

For now, we omit test functions when constructing the `ParseTree` during regular compilation (i.e. `forc build` or `forc check`) so that:

1. Programs may continue to compile even if tests do not typecheck (like Rust).
2. We don't do extra work type-checking test functions when we don't have to.
3. We don't get a bunch of dead code warnings for `#[test]` fns.

Later on, I'll add an option to enable these during `forc test`, `forc check --tests`, etc.

This PR doesn't enable compilation or execution of test functions just yet. This will be addressed in a following PR (see #1832).

Adds a basic test that makes sure we don't provide any undesired warnings about unknown attributes or dead code.

### DIFF
--- a/sway-core/src/convert_parse_tree.rs
+++ b/sway-core/src/convert_parse_tree.rs
@@ -86,7 +86,10 @@ impl ErrorContext {
     }
 }
 
-pub fn convert_parse_tree(module: Module) -> CompileResult<(TreeType, ParseTree)> {
+pub fn convert_parse_tree(
+    module: Module,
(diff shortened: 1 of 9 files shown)


sezna (approved):

JoshuaBatty (approved):
:shipit:

PR #2981, opened 2022-10-10
CompileWarning -> sway_error

Centril:
Moves `CompileWarning` into `sway_error`. This is work towards #2734. After this PR, we can integrate warnings into `Handler` so that it is functionally on par with `CompileResult`, and then remains to use `Handler` in various places, which can probably be done incrementally over a series of PRs.

This also moves `style` and `IntegerBits` into `sway-types`. The former is also made more "pure".


mitchmindtree (approved):
:ship:

mohammadfawaz (approved):

vaivaswatha (approved):

PR #2985, opened 2022-10-11
Sway unit test (i.e. `#[test]` fn) compilation

### LINKED ISSUE #1832: `#[test_script]` for unit testing

mitchmindtree:
OK, this should be good to go!

## Highlights

- Adds support for multiple entry points to IR. Adds a tests for serializing entry functions to/from IR.
- Enables ASM generation for libraries (to support test function entry points).
- Track entry points through ASM generation so we can return entry point metadata as a part of the compilation result. This doesn't affect ASM generation, but allows tools using `sway-core` as a library to work with different entry points.
- Updated E2E test harness with a new test category "UnitTestsPass".
- Added E2E tests with multiple unit tests for each type of Sway program (library, script, predicate, contract).
- Gets `forc test` working with pretty output!

Here's the output from the new `lib_multi_test`:

Screenshot from 2022-11-03 19-00-57

## Test running implementation

Currently, it seems like there's no publicly accessible approach to execute a script directly from a custom entry point. As a result, for each test, we patch the bytecode with a `JI` instruction that jumps from after the data section setup to the test's entry point. This is a bit hairy, but works for now!

As a follow-up, we may want to consider adding support upstream in `fuel-vm` for executing scripts directly from a given entry point. Even if this was only exposed from the interpreter API, this would make the `forc-test` implementation quite a bit cleaner and avoid the need to patch the bytecode.

Alternatively, when building projects we could return more metadata along with the compiled output (whether in memory or bytecode header) to indicate how to work with different entry points in a more reliable manner (rather than the magic const offset currently used in `forc-test`).

# TODO

- [x] Add `include_tests` flag to `BuildConfig`, allowing `forc` to trigger compilation of test functions.
- [x] Include `#[test]` fns as entry points within dead code analysis.
- [x] IR and ASM generation for test entry points.
- [x] Add `forc build --tests`.
- [x] Add `forc test`.
- [x] Always include test fns in `TyProgram` (for DCA), but omit from IR if not building tests.
- [x] Work out how to iterate over different entry points during `forc test`.
- [x] Only generate tests for top-level "members" (not all dependencies).

### Follow-up:

- #3260
- #3261
- #3262
- #3263
- #3264
- #3265
- #3266
- #3267
- #3268
- #3278

Closes #1832.


nfurfaro:
First, I can't wait to start using this!
Will give a more thorough look over, but one thing I noticed right away is:
>Updated E2E test harness with a new test category "UnitTestsPass".

We currently have subdirectories "should_pass"/"should_fail", but these are legacy artifacts from a time before @otrho introduced the test.toml file, which is now the canonical way to declare whether a test should pass or fail. We should remove these subdirectories as they're confusing and just add noise.
However, that could be a small dedicated PR to avoid crowding this one with unrelated changes.

sway-core/src/semantic_analysis/program.rs
```diff
@@ -70,21 +72,17 @@ impl ty::TyProgram {
                         storage_slots.sort();
                         ok(
                             Self {
-                                kind: self.kind.clone(),
-                                root: self.root.clone(),
                                 storage_slots,
-                                logged_types: self.logged_types.clone(),
+                                ..self
```
mohammadfawaz, line 76: Nice.

sway-ir/src/parser.rs
```diff
@@ -66,31 +67,23 @@ mod ir_builder {
...
-                        metadata,
-                        locals,
-                        blocks,
-                        selector
+                        selector,
+                        is_entry
                     }
                 }
 
+            rule is_public() -> bool
```
mohammadfawaz, line 79: This is probably going away eventually. `pub` is really not relevant in the IR - it doesn't serve any purpose outside of the frontend.

mohammadfawaz (approved):
Good stuff. This is very clean change and easy to understand. Thank you! No actionable feedback from my side, but have two questions:
* Are we planning on disallowing certain opcodes in the tests? For example, we should presumably disallow storage operations, logging (?), contract operations in general, etc.
* Would you say this in a state that we can advertise it already? I assume not before documentation is written including limitations etc.
* Are we meant to update `sway/Cargo.toml` to add the new crate?

mitchmindtree:
@mohammadfawaz thanks for the review!

> Are we planning on disallowing certain opcodes in the tests? For example, we should presumably disallow storage operations, logging (?), contract operations in general, etc.

Ahh good point! Yes denying storage or contract specific operations makes a lot of sense, I've opened another follow-up issue https://github.com/FuelLabs/sway/issues/3278. Maybe logging is something we want to work out how to support for providing feedback, e.g. similar to printing to stdio and using `cargo test`'s `--nocapture` flag?

> Would you say this in a state that we can advertise it already? I assume not before documentation is written including limitations etc.

Yeah perhaps it's worth holding off until we get some docs up?

> Are we meant to update sway/Cargo.toml to add the new crate?

Good spot! I wonder why I'm not getting warnings about this :thinking:  it looks like it has still been included within the `Cargo.lock` file - perhaps a recent cargo update now implicitly includes all nested crates as members or something? I'll add this either way!

mohammadfawaz (approved):

JoshuaBatty (approved):
Epic, so cool that have this in sway!

PR #2986, opened 2022-10-11
Handler warnings + misc related simplifications

Centril:
After https://github.com/FuelLabs/sway/pull/2981, which this PR is based on, we can now add `fn emit_warn` to the `Handler`, which is what this PR does.

This PR also does some related refactoring, in particular in `swayfmt`.
This is work towards https://github.com/FuelLabs/sway/issues/2734.

PR #2988, opened 2022-10-11
Use `Handler` for convert_parse_tree + sway-core parsing logic

Centril:
Based on https://github.com/FuelLabs/sway/pull/2986, this PR, which works towards https://github.com/FuelLabs/sway/issues/2734, uses `Handler` in the convert-to-parse-tree logic, and then uses this to move away from `CompileResult` in `sway-core` for parsing-related stuff, up to the boundary, but not including, of `pub fn parse(input: Arc<str>, config: Option<&BuildConfig>) -> CompileResult<parsed::ParseProgram>`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4468,6 +4468,7 @@ dependencies = [
  "futures",
  "notify",
  "notify-debouncer-mini",
+ "parking_lot 0.12.1",
  "ropey",
  "serde",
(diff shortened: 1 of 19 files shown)


mohammadfawaz:
Conflicts :(

sezna (approved):
solid refactor!

Issue #2951, opened 2022-10-05
Refactor the front end of the compiler such that ast transformation is seperate from semantic analysis


emilyaherbert:

emilyaherbert:
This was closed by mistake.

PR #3006, opened 2022-10-11
Move struct definition to `ty` module.

emilyaherbert:
This PR is one of several in a process to address https://github.com/FuelLabs/sway/issues/2951. This PR moves the struct definition to the ty module. Currently, things are a little awkward organizationally, as the definitions and impl blocks are in different modules, but this is done in (quick) iterations to reduce merge conflicts overall.


emilyaherbert:
Closing in favor of #3024.

Issue #2970, opened 2022-10-07
Add a way to specify salt with `forc-build`


kayagokalp:
Currently we do not have a CLI option to specify `Salt` while deploying contracts. https://github.com/FuelLabs/sway/pull/2958#discussion_r989633174

adlerjohn:
#2958 seems like an overloading of `build`. Would it be better to instead add a new subcommand e.g. `forc contract-id`that does a `build` then outputs the contract ID, accepting a `--salt` parameter?

mitchmindtree:
> https://github.com/FuelLabs/sway/pull/2958 seems like an overloading of build.

True, I guess we don't *need* the contract ID in `out/`, as the ID is more relevant to other scripts or contracts that will call into this contract :thinking:

---

Perhaps a better alternative might be to specify the `salt` for each entry under `[contract-dependencies]` instead, as the ID of a contract is generally only needed by dependents, E.g.

```toml
[contract-dependencies]
foo = { path = "../foo", salt = "blah" }
bar = { git = "https://host.com/user/repo", tag = "v1.0.0" }
```

In the above, `foo` would be built, have its contract ID generated using the salt "blah", and that ID would be provided via the const `foo::CONTRACT_ID` to our package. `bar` would be built similarly, but it's contract ID would be generated without salt (the default case with no `salt` specified) before being provided to our package as the const `bar::CONTRACT_ID`. This approach would also enable specifying a unique salt per contract dependency, which would likely be necessary for some folk.

Thoughts @kayagokalp @adlerjohn ?

adlerjohn:
The salt is exactly and always 32 bytes, so you can't have "no salt." You could have 32 bytes of zeroes. Proposal seems fine I guess otherwise.

kayagokalp:
I think specifying them with the `[contract-dependencies]` makes sense but if two different package depends on the same `contract` they should be providing the same `salt` to get the same contract id. Could this be confusing for the users? Maybe we could start like this and add a `salt` declaration to the members of the `WorkspaceManifest` later on. In that case we could check if the `[contract-deployment]` table entry consists the `salt` information. if it does we can use that. Otherwise check the `WorkspaceManifest`. In the absence of both we could fallback to 32 bytes of zeros.

mitchmindtree:
> I think specifying them with the `[contract-dependencies]` makes sense but if two different package depends on the same contract (for deployment) they should be providing the same salt to get the same contract id. Could this be confusing for the users?

[...]

PR #3016, opened 2022-10-12
feat: Introduce the concept of contract dependencies to the package manifest file

### LINKED ISSUE #2931: Introduce the concept of deployment dependencies to the package manifest file
After #2893 we will have the basics of the workspace manifest file which will hold just the members of the workspace. We will need a way to declare deployment dependencies for members of the workspace so that while deploying we can build and deploy them in the correct order.

kayagokalp:
closes #2931.

### Summary
`[contract-dependencies]` introduced to `Forc.toml`. Listed dependencies are added to dependency graph so that they are built and their contract ID is collected before root package is built and inserted to its namespace by injecting them as config time constants.

### My approach

I first thought of creating a new graph with just the root nodes of the declared dependencies under `[contract-dependencies]` so that I can do builds in the correct order and that would be basically building all of them exclusively in the correct order. But doing so creates the problem of double builds. For example if `contract a` and `contract b` both depends on `std` and `contract a` declares `contract-b` as a `[contract-dependency]` we will be building `std` twice. To overcome this issue, I added contract dependencies to the original dependency graph while doing so I made sure that root nodes of `[contract-dependencies]` are immediate childs of the root node of the original package. Doing like that ensures that `std` built just once. After dependencies are fetched and the dependency graph is generated, while building I collected the contract IDs of the ones that are originally declared under `[contract-dependencies]` for the original package. After collection of the contract IDs, I injected the collected ids to the namespace of the root package by adding them to the configuration time constants for that package.

This approach currently works but it is a little rough around the edges.

I am currently working on improving the doc and changing the injected contract ID variable names. ~Currently the injected contract ids are named like `{dependency_name}_CONTRACT_ID`~. I wanted to make it like `{dependency_name}::CONTRACT_ID` but I guess since they are inserted as a configuration time constants to the root namespace we can't have that since `{dependency_name}::CONTRACT_ID` is not a valid variable name. I also tried to inject these contract ids to the submodule part of the namespace (by creating a default namespace with config time constants for the contract dependency) but I guess by default config time constants are non-public so we cannot use it from the root module. I an currently looking for other options for inserting this constants publicly to the submodules so that they can be used like:

```rust
let x = contract_b::CONTRACT_ID;
```

## Updated Note

With slight changes to the `ConfigTimeConstants` I added an option to make it public this enabled me to crate submodules for `[contract-dependencies]` and insert their contract ids to these newly created submodules. So currently they can be used like the following (same `example of contract_a` depending on `contract_b` for deployment):

```rust
script;
use contract_b:::CONTRACT_ID as b_contract_id;

fn main() {
 let x = b_contract_id;
}
```

Unfortunately the following seems to be failing which seems unrelated to this PR:

```rust
script;

fn main() {
 let x = contract_b:::CONTRACT_ID;
}
```

## TODO

- [x] Documentation
- [x] Change the injected contract ID (config-time-constant) name
- [x] Re-enable invalid dependency error (it was disabled to test depending on a contract)
- [x] Update `forc-check` so that it can also type check contract ids coming from contract dependencies
- [x] Fix the bug introduced with this PR which can be seen from the [CI](https://github.com/FuelLabs/sway/actions/runs/3233996550/jobs/5296563594)
- [x] Add tests


forc-pkg/src/pkg.rs
```diff
@@ -450,19 +461,35 @@ fn validate_deps(
     graph: &Graph,
     node: NodeIx,
     node_manifest: &PackageManifestFile,
+    contract_dependencies: &mut HashSet<String>,
```
mitchmindtree, line 464: Rather than passing in a `HashSet` of `contract_dependencies` here to determine whether or not dependencies are contract dependencies throughout the dependency validation process, we might want to describe the dependency *kind* in the `Edge` type.

Currently, the graph's `Edge` type is just the dependency name, however we could update it to also contain the dependency kind like so:

```rust
struct Edge {
    /// The name of the dependency as declared under `[dependencies]` or `[contract-dependencies]`.
    /// This may differ from the package name as declared under the dependency package's manifest.
    name: String,
    kind: DepKind,
}

enum DepKind {
    /// The dependency is a library and declared under `[dependencies]`.
    Library,
    /// The dependency is a contract and declared under `[contract-dependencies]`.
    Contract,
}
```

This should help to simplify a lot of the implementation involved here, as rather than having to repeatedly check if a name exists within a `contract_dependencies` set, we can instead simply check the edge's dependency kind when we need to distinguish behaviour.

mitchmindtree (changes requested):
Nice start @kayagokalp!

I've made a start in reviewing this, but noticed one detail that I think could help to simplify the implementation here quite a bit. If you could have a go at addressing the `Edge` and `DepKind` comment, and move toward using those where feasible rather than the `contract_dependencies` set, I'll take another look!

docs/src/forc/manifest_reference.md
```diff
@@ -147,3 +149,35 @@ foo = { git = "https://github.com/foo/foo", branch = "test" }
...
+Example usage:
+
+```rust
+script;
+
+use foo::CONTRACT_ID as foo_contract_id;
+
+fn main() {
+  let x = foo_contract_id;
+}
```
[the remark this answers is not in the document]
kayagokalp: I wanted to give the example like this but I believe this does not work currently.

I tested out with a script depending on a library with a pub const.

lib:

```rust
library test_lib;

pub const NUMBER: u64 = 10;
```

script that depends on the lib:

```rust
script;

fn main() {
  let x = test_lib::NUMBER;
}
```

and trying to build the script above, fails with:

``` ^^^^^^ Could not find symbol "NUMBER" in this scope. ```

This seems like an unrelated issue with the compiler and I think I saw this issue but couldn't find it right now. I will look into and raise an issue if it does not already exists 👍

File: forc-pkg/src/pkg.rs (hunk 1 above)
Comment (kayagokalp) [reply forc-pkg/src/pkg.rs]: Thanks for the suggestion I think it is much more clearer right now!

**Comment by kayagokalp** [comment]
@mitchmindtree Thanks for pointing out a better direction. I think with the introducing `Edge` struct we are in a better place now. Sorry for the diff produced by introducing `dep-kind` to the lock file (I needed it while reconstructing the graph from lock file) which lead to updating lock files for tests. I think this is ready for an another review now!

File: sway-types/src/lib.rs [hunk 3]
```diff
@@ -102,6 +102,8 @@ where
 pub struct ConfigTimeConstant {
     pub r#type: String,
     pub value: String,
+    #[serde(default)]
+    pub public: bool,
```
mitchmindtree, line 106: While the `ContractId`s *currently* use the same method of being added to the namespace, they don't have anything to do with the concept of "configuration time" which will likely be removed in the future.

The `ConfigTimeConstant` is also implemented in a bit of a strange way which requires them to be parsed from a string. Instead, we likely want to construct the AST directly instead as mentioned [here](https://github.com/FuelLabs/sway/blob/8ca5234d85358a24e23434cbf4ef93c7674980c9/sway-core/src/semantic_analysis/namespace/module.rs#L84).

Could you open an issue to track this? Something along the lines of

> Contract ID constants should not use `ConfigTimeConstant` and should construct AST directly

Then we can leave this as-is to get this working for now, then come back and refactor this in a follow-up PR and I can help with the AST construction stuff.

test/src/e2e_vm_tests/test_programs/should_pass/forc/contract_dependencies/contract_a/src/main.sw
```diff
@@ -0,0 +1,14 @@
+contract;
+
+use contract_b::CONTRACT_ID as contract_b_id;
```
mitchmindtree, line 3:
```suggestion
use contract_b::CONTRACT_ID as CONTRACT_B_ID;
```

test/src/e2e_vm_tests/test_programs/should_pass/forc/contract_dependencies/contract_a/src/main.sw
```diff
@@ -0,0 +1,14 @@
...
+
+abi MyContract {
+      fn test_function();
+}
+  
+  impl MyContract for Contract {
+      fn test_function() {
+    	  let contract_b_id = contract_b_id;
```
mitchmindtree, line 11:
```suggestion
    	  let CONTRACT_B_ID = CONTRACT_B_ID;
```

mitchmindtree (changes requested):
Nice work! The new `Edge` type looks perfect, and the changes are already looking much nicer as a result.

I've done a more thorough review this time, just a few more things to address but we're much closer :rocket: 

## `build` and `check`

There should be no need to special case the `root_node` within the `build` and `check` functions, particularly as each of the `[contract-dependencies]` might also have their own set of `[contract-dependencies]` which need to be handled as well.

Also worth noting is that a contract dependency's ID may differ between packages once [the `salt` field is added to the `contract-dependencies` table](https://github.com/FuelLabs/sway/issues/2970#issuecomment-1273927559) in follow-up work, so rather than keeping a `contract_dependency_ids` map, we likely want to recalculate each contract dependency's ID for each package.

Keeping this in mind, rather than adding the `node == *root_node` checks in each of `build` and `check`, I think we likely want to address adding the `ContractId`s in the `dependency_namespace` function instead.

In `fn dependency_namespace`, during the `// Add direct dependencies` loop, we should check the `edge`'s kind and:

1. If `edge.kind` is `Library`, retrieve the namespace from the `namespace_map` (the existing behaviour). Otherwise,
2. If `edge.kind` is `Contract`, do not use the `namespace_map`. Instead:
    1. Retrieve the dependency's `ContractId`.
    2. Create a new, empty `namespace::Module` for the contract dependency.
    3. Add the `ContractId` to the contract dependency module.
    4. Insert the contract dependency module into the package's namespace.

We also might wish to rename the `namespace_map` to `lib_namespace_map` to clarify that it only contains the namespace modules for dependencies that are `DepKind::Library`?

## Tests

New test is looking good! Could you possibly also add a `contract_c` and add it as a dependency to both `a` and `b` so that we can make sure that contract dependencies can also have their own contract dependencies?

docs/src/forc/manifest_reference.md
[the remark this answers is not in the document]
mitchmindtree: #3063 is tracking this issue for anyone reading this :)

forc-pkg/src/pkg.rs
```diff
@@ -2187,21 +2271,35 @@ pub fn build(plan: &BuildPlan, profile: &BuildProfile) -> anyhow::Result<(Compil
...
+            };
         let res = compile(pkg, manifest, profile, dep_namespace, &mut source_map)?;
         let (compiled, maybe_namespace) = res;
+        // If the current node is a contract dependency, collect the contract_id
+        if root_manifest
+            .contract_deps()
+            .any(|dep| *dep.0 == manifest.project.name)
+        {
+            compiled_map.insert(node, compiled.clone());
+        }
```
mitchmindtree, line 2302: This still doesn't account for the case where the package is only a contract dependency of another contract dependency.

We can remove all the `root_`-specific logic and instead check if the current node is depended on as a contract dependency by any other node.

We can do so by checking if any of the incoming edges of this node are `DepKind::Contract`, e.g. something like:

```rust
if plan.graph().edges_directed(node, Incoming).any(|e| e.weight().kind == DepKind::Contract) {
    compiled_map.insert(node, compiled.clone());
}
```

forc-pkg/src/pkg.rs
```diff
@@ -2363,12 +2461,14 @@ pub fn check(
     sway_core::clear_lazy_statics();
     let mut namespace_map = Default::default();
     let mut source_map = SourceMap::new();
+    let compiled_map = HashMap::new();
```
mitchmindtree, line 2464: For readers of this code, seeing a collection that is forever empty might be confusing, so a comment might be useful:
```suggestion
    // During `check`, we don't compile so this stays empty.
    // In turn, default contract IDs are used when constructing the dep namespace.
    let compiled_contract_deps = HashMap::new();
```

forc-pkg/src/pkg.rs
```diff
@@ -1720,19 +1779,45 @@ pub fn sway_build_config(
 /// then the std prelude will also be added.
 pub fn dependency_namespace(
     namespace_map: &HashMap<NodeIx, namespace::Module>,
+    compiled_map: &HashMap<NodeIx, Compiled>,
```
mitchmindtree, line 1782: Let's clarify this should only contain compiled contract deps:
```suggestion
    compiled_contract_deps: &HashMap<NodeIx, Compiled>,
```

forc-pkg/src/pkg.rs
```diff
@@ -2187,21 +2271,35 @@ pub fn build(plan: &BuildPlan, profile: &BuildProfile) -> anyhow::Result<(Compil
     let mut storage_slots = vec![];
     let mut bytecode = vec![];
     let mut tree_type = None;
+    let root_node = plan
+        .compilation_order()
+        .last()
+        .expect("Compilation order is empty!");
+    let root_node_pinned = &plan.graph()[*root_node];
+    let root_manifest = &plan.manifest_map()[&root_node_pinned.id()];
+    let mut compiled_map = HashMap::new();
```
mitchmindtree, line 2280:
```suggestion
    let mut compiled_contract_deps = HashMap::new();
```

mitchmindtree (changes requested):
Just a few more small tweaks and suggestions then I think this should be ready!

forc-pkg/src/pkg.rs
kayagokalp: Ahh, yeah I missed this part thanks for pointing out! I Maybe I should add this case to the test too

forc-pkg/src/pkg.rs
```diff
@@ -1720,19 +1779,45 @@ pub fn sway_build_config(
...
-        namespace.insert_submodule(dep_name, dep_namespace.clone());
+        let dep_name = kebab_to_snake_case(&edge.weight().name);
+        let dep_edge = edge.weight();
+        let dep_namespace = match dep_edge.kind {
+            DepKind::Library => namespace_map
+                .get(&dep_node)
+                .map(Clone::clone)
+                .expect("no namespace module for {dep_name}"),
+            DepKind::Contract => {
+                let mut constants = constants.clone();
```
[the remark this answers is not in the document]
kayagokalp: Yeah it seems like it is not doing what i intended at the beginning. So if a contract dep declares config time constants that are `public` they are not transitively moved to root node. Since the infrastructure that we are using is not going to be used for config time constants, but if I understood correctly, I guess it does not need to work as this infra is going to be just used our `CONTRACT_ID` logic and config time constants are going to be separate.

sway-types/src/lib.rs
kayagokalp: Created #3077 to track that 👍

mitchmindtree (approved):
Wicked stuff @kayagokalp! Looking great to me :rocket: :ship:

mitchmindtree (changes requested):
Actually @kayagokalp, I just realised we can avoid a lot of downstream breakage in lock files (including the local lock files) if we use separate lists in the lock file entries for `dependencies` and `contract-dependencies`. This would mean existing lock files no longer need to change at all, which would be great for avoiding causing a lot of conflicts. Would you mind having a go at this?

forc-pkg/src/lock.rs
```diff
@@ -151,15 +183,34 @@ impl Lock {
...
+            // If `pkg.dependencies` is None, we will be collecting an empty list of
+            // lib_deps so that we will omit them during edge adding phase
+            let lib_deps: Vec<(String, DepKind)> = pkg
+                .dependencies
+                .clone()
+                .unwrap_or_default()
+                .iter()
+                .map(|lib_dep| (lib_dep.clone(), DepKind::Library))
+                .collect();
+            for (dep_line, dep_kind) in lib_deps.iter().chain(contract_deps.iter()) {
```
mitchmindtree, line 204: Another way of writing this section might be:

```rust
            let contract_deps: Vec<(String, DepKind)> = pkg
                .contract_dependencies
                .as_ref()
                .into_iter()
                .flat_map(|vec| vec)
                .map(|contract_dep| (contract_dep, DepKind::Contract));
            let lib_deps: Vec<(String, DepKind)> = pkg
                .dependencies
                .as_ref()
                .into_iter()
                .flat_map(|vec| vec)
                .map(|lib_dep| (lib_dep, DepKind::Library));
    for (dep_line, dep_kind) in lib_deps.chain(contract_deps) {
```
^ This should achieve the same behaviour avoiding both the `Vec` clones and the `String` clones.

mitchmindtree (approved):
Left one minor optional suggestion, otherwise looking great!

forc-pkg/src/lock.rs
kayagokalp: Thanks! I can address this in the next PR to not to dismiss your approval 👍

forc-pkg/src/lock.rs
```diff
@@ -34,7 +34,8 @@ pub struct PkgLock {
     version: Option<semver::Version>,
     // Short-hand string describing where this package is sourced from.
     source: String,
-    dependencies: Vec<PkgDepLine>,
+    dependencies: Option<Vec<PkgDepLine>>,
+    contract_dependencies: Option<Vec<PkgDepLine>>,
```
bingcicle, line 38: out of curiosity, why are dependencies turned into an `Option` here?

mitchmindtree: This was done so that, in the case that the package has no dependencies or contract dependencies, the `dependencies` or `contract_dependencies` field is omitted from the lock file entirely rather than shown with an empty list.

test/src/e2e_vm_tests/test_programs/should_pass/forc/contract_dependencies/contract_a/Forc.lock
```diff
@@ -0,0 +1,16 @@
+[[package]]
+name = 'contract_a'
+source = 'root'
+contract_dependencies = [
```
mitchmindtree, line 4: One more minor request I just noticed after addressing Bing's Q: Could we change this to `contract-dependencies` (use slash rather than underscore) to match the `Forc.toml` name? I think there's a `serde` attribute that allows for renaming a field to a kebab case name.

mitchmindtree (changes requested):

forc-pkg/src/lock.rs
bingcicle: Hmm so the change to `Option` isn't entirely related introducing contract dependencies? Was wondering if not having to specify dependencies was an intentional part of the change, or if it was just convenient to update this along the way.

mitchmindtree: Good question! The reason why I gave the pass for this change in this PR is that a decision needed to be made for the new `contract-dependencies` field. Without the `Option`, every test lock file in the Sway repo would require a new empty `contract-dependencies` entry, so Kaya added the `Option`. Then we figured the behaviour should be consistent for both of the dependencies fields, so the decision was made to have `dependencies` use an `Option` too.

bingcicle: Ah, got it, that makes sense, thanks :+1:

bingcicle (dismissed):

mitchmindtree (approved):

bingcicle (approved):

PR #3023, opened 2022-10-12
Bump to `v0.26.0`

mohammadfawaz:
Also run `cargo update`


JoshuaBatty (dismissed):

mitchmindtree:
It looks like this includes the removal of integration testing from [`forc test`](https://github.com/FuelLabs/sway/pull/2264) which is quite breaking for those who were previously using `forc test` - should we consider a bump to `0.26`?

mohammadfawaz:
> It looks like this includes the removal of integration testing from [`forc test`](https://github.com/FuelLabs/sway/pull/2264) which is quite breaking for those who were previously using `forc test` - should we consider a bump to `0.26`?

True! Thanks for pointing that out. Let me update the PR.

kayagokalp (approved):

sezna (approved):

PR #3030, opened 2022-10-13
Port TypedTraitFn and TyImplTrait methods to the declaration engine.

tritao:
Closes https://github.com/FuelLabs/sway/issues/2784.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -890,6 +890,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dcb25d077389e53838a8158c8e99174c5a9d902dee4904320db714f3c653ffba"
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c2dd04ddaf88237dc3b8d8f9a3c1004b506b54b3313403944054d23c0870c521"
+dependencies = [
+ "cfg-if 1.0.0",
+ "crossbeam-utils",
+]
+
+[[package]]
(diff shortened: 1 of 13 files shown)


mohammadfawaz (approved):

JoshuaBatty (approved):

sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs
```diff
@@ -482,13 +494,19 @@ fn type_check_trait_implementation(
     let mut functions_buf: Vec<ty::TyFunctionDeclaration> = vec![];
     let mut processed_fns = std::collections::HashSet::<Ident>::new();
 
+    let mut trait_fns = vec![];
+    for decl_id in trait_interface_surface.iter() {
+        match de_get_trait_fn(decl_id.clone(), block_span) {
+            Ok(decl) => trait_fns.push(decl),
+            Err(err) => errors.push(err),
+        }
+    }
```
otrho, line 503: Seeing this pattern repeated a lot in this change... feels like it could be generalised.  Below with the `methods` and `de_get_function` too.  But I'm not sure what the best way would be since they're just slices.  Something like a `scan` or maybe just a fold, called `gather_decls` or something... but meh, it's not going to improve readability that much.  :shrug: 

If you want to `for` loop over a collection to get references then you can use `for elem in &elems {}` rather than calling `elems.iter()` usually, but in this case it's a slice so I think you can just remove the `.iter()`.

otrho (approved):
Small nits.  Up to you.

otrho (approved):

JoshuaBatty (approved):

PR #3041, opened 2022-10-14
fix: add missing leaf span collection for final_expr_opt of asm blocks

### LINKED ISSUE #2716: Comments after a final expression in `asm` creates unparsable code
```rust
script;

use std::assert::assert;

fn adder(a: u64, b: u64, c: u64) -> (u64, u64) {
    let empty_tuple = (0u64, 0u64);
    asm(output: empty_tuple, r1: a, r2: b, r3: c, r4, r5) {
        add r4 r1 r2; // add a & b and put the result in r4
        add r5 r2 r3; // add b & c and put the result in r5
        sw output r4 i0; // store the word in r4 in output + 0 words
        sw output r5 i1; // store the word in r5 in output + 1 word
        output: (u64, u64) // return both values <-- when this comment is removed the code formats as intended
    }
}

fn main() -> bool {
    let(first, second) = adder(1, 2, 3);
    assert(first == 3);
    assert(second == 5);
    true
}

```

kayagokalp:
closes #2716.

### DIFF
--- a/examples/asm_return_tuple_pointer/src/main.sw
+++ b/examples/asm_return_tuple_pointer/src/main.sw
@@ -12,7 +12,7 @@ fn adder(a: u64, b: u64, c: u64) -> (u64, u64) {
 }
 
 fn main() -> bool {
-    let(first, second) = adder(1, 2, 3);
+    let (first, second) = adder(1, 2, 3);
     assert(first == 3);
     assert(second == 5);
     true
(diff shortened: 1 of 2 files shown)


mohammadfawaz (approved):

eureka-cpu:
Would you mind running the formatter on `lib-std/core`? There may be some formatting changes there, I think I saw this issue arise in one of them for multiple programs

kayagokalp:
> Would you mind running the formatter on `lib-std/core`? There may be some formatting changes there, I think I saw this issue arise in one of them for multiple programs

Seems like there are already formatter changes happening even if I run the `master` version of the formatter and same changes happen if I run with this branch too. Only difference is `sway-lib-std/src/hash.sw` does not fail to format with this branch but fails with master. (the formatter formats successfully with this branch but there is no diff generated so nothing to commit for that file, as it is already formatted)

cc @eureka-cpu

eureka-cpu (approved):

PR #3048, opened 2022-10-15
Move any general methods to live within the same file as the definition for their parent.

emilyaherbert:
Soon the `language` module will live in it's own crate, so this PR moves any general methods to live within the `impl` block defined for that definition. In other words, this PR creates several instances where there are multiple `impl` blocks for the same definitions---one in the `ty` module and another in the `semantic_analysis` module. The `impl` blocks in the `semantic_analysis` module will be refactored to be more type-checking+type-inference focused and less AST focused, and thus removed.

When reviewing this PR, the most helpful thing would be to look out for any methods that you feel are not general that I did move, or methods that are general that I missed.


mohammadfawaz (dismissed):

otrho (approved):

otrho:
Bah, @emilyaherbert conflicts.

mohammadfawaz (approved):

otrho (approved):

mohammadfawaz (approved):
This makes more sense. Thanks!

PR #3053, opened 2022-10-16
WIP: implement `ge` and `le` for U128

alphak3y:
`ge` and `le` functionality added for type `U128`.


ControlCplusControlV:
I suggest mirroring [the other impls of ge](https://github.com/FuelLabs/sway/blob/master/sway-lib-core/src/ops.sw#L400) then adding some tests to indicate the `=>` and `>=` work for the new functions

alphak3y:
// TODO this doesn't work?
// impl core::ops::OrdEq for U128 {
// }

Someone left this comment indicating they tried to implement this but were having trouble.

alphak3y:
Here is the missing piece:

trait OrdEq: Ord + Eq {
} {
    fn ge(self, other: Self) -> bool {
        self.gt(other) || self.eq(other)
    }
    fn le(self, other: Self) -> bool {
        self.lt(other) || self.eq(other)
    }
}

impl OrdEq for u64 {
}
impl OrdEq for u32 {
}
impl OrdEq for u16 {
}
impl OrdEq for u8 {
}
impl OrdEq for b256 {
}

mohammadfawaz:
@alphak3y I'm going to switch this into a draft until we figure out a way forward.

mohammadfawaz:
Tracking the work here https://github.com/FuelLabs/sway/issues/3221

PR #3055, opened 2022-10-17
Return reference type values from calls.

otrho:
Implement changes to call to allow returning reference-type values.

Uses an auxiliary 'out' parameter to receive the value, uses `mem_copy` to set it and then returns the pointer itself.

### DIFF
--- a/sway-ir/src/instruction.rs
+++ b/sway-ir/src/instruction.rs
@@ -368,7 +368,12 @@ impl Instruction {
                 replace(log_val);
                 replace(log_id);
             }
-            Instruction::MemCopy { .. } => (),
+            Instruction::MemCopy {
+                dst_val, src_val, ..
+            } => {
+                replace(dst_val);
+                replace(src_val);
+            }
             Instruction::Nop => (),
             Instruction::ReadRegister { .. } => (),
             Instruction::Ret(ret_val, _) => replace(ret_val),
--- a/sway-ir/src/instruction.rs
+++ b/sway-ir/src/instruction.rs
@@ -109,6 +109,12 @@ pub enum Instruction {
         log_ty: Type,
         log_id: Value,
     },
+    /// Copy a specified number of bytes between pointers.
+    MemCopy {
+        dst_val: Value,
+        src_val: Value,
+        byte_len: u64,
+    },
     /// No-op, handy as a placeholder instruction.
     Nop,
     /// Reads a special register in the VM.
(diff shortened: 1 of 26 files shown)


otrho:
I'm not 100% happy with this change as it's needed to be a bit hacky.  In particular, the ambiguity around IR pointers, which will be fixed by #2819 soon, means that we're sometimes using explicit pointers for values and sometimes they're implicit pointers for ref-types.

e.g.,
```
// Original code.
fn foo() -> b256 { ... }

// New code, after this PR.
fn foo(__ret_value: mut ptr b256) -> b256 {
    ...
    mem_copy(__ret_value, calculated_value, 32)
    ret b256 __ret_value
}
```
The return type is `b256` but it's actually returning a `mut ptr b256`.  But they work out to be the same thing.  Ugh.  Implicit pointers for ref types was used in the IR originally because that's what we do in Sway, but that needs to change with #2819.

otrho:
@mohammadfawaz Any idea why the JSON ABI log IDs changed in 2fb1b53 after changing storage reads to be from the heap in bac53fd?

mohammadfawaz:
> @mohammadfawaz Any idea why the JSON ABI log IDs changed in 2fb1b53 after changing storage reads to be from the heap in bac53fd?

Yeah because type IDs changed probably, and I'm using them as proxies for log IDs. There's an issue to clean that up - just remapping the IDs really but that's a bit annoying to do because the bytecode contains those IDs as well.

sezna (commented):
Is the source of the bytecode size increase just the extra move and memcopy ops around returning references? That bloat is more than I would have expected.

otrho:
> Is the source of the bytecode size increase just the extra move and memcopy ops around returning references? That bloat is more than I would have expected.

Yeah I'll look into it.  It could just be that, or something more nefarious and fixable.  We can always make the inlining more aggressive again if it's the former.

sway-ir/src/instruction.rs
```diff
@@ -473,6 +487,16 @@ pub struct InstructionInserter<'a> {
     block: Block,
 }
 
+macro_rules! make_instruction {
```
mohammadfawaz, line 490: Cool 😄

mohammadfawaz (approved):
Seems reasonable!

otrho:
@sezna Yep, I've had a look at the asm and the growth in binary size is pretty much the bytecode needed to set up addresses and then to perform the `mem_copy` for the returned values.

This is done before inlining, so my idea that making inlining more aggressive would help was incorrect.  The problem is that these new `mem_copy`s are mostly redundant, as we already have implicit copying of ref types anyway.  They used to be leaked onto the stack, but they're stricter now.

So the we really need to add a pass for fixing memory use, probably something like LLVMs [memcpyopt](https://llvm.org/docs/Passes.html#memcpyopt-memcpy-optimization), just to reduce the copying.  That is, on top of fixing pointers in general (#2819) so that we don't have implicit copying any more.

And it is probably possible to improve the ASMgen for these things.  Overripe for some improvements in general.

sezna (approved):

PR #3056, opened 2022-10-17
Use Handler in Purity checking

Centril:
Based on https://github.com/FuelLabs/sway/pull/2988 and working towards https://github.com/FuelLabs/sway/issues/2734, this PR uses the `Handler` in purity checking, also enabling error recovery on the sway.

N.B. You'll want to review this with whitespace changes hidden.

### DIFF
--- a/sway-core/src/lib.rs
+++ b/sway-core/src/lib.rs
@@ -414,42 +414,21 @@ pub(crate) fn compile_ast_to_ir_to_asm(
         errors
     );
 
-    // TODO: Experiment with putting combine-constants and simplify-cfg
-    // in a loop, but per function.
-    check!(
-        combine_constants(&mut ir, &all_functions),
-        return err(warnings, errors),
-        warnings,
-        errors
-    );
-    check!(
-        simplify_cfg(&mut ir, &all_functions),
-        return err(warnings, errors),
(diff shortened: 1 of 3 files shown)


sway-core/src/lib.rs
```diff
@@ -413,48 +414,34 @@ pub(crate) fn compile_ast_to_ir_to_asm(
...
+        simplify_cfg(handler, &mut ir, &all_functions)?;
+        // Simplify-CFG helps combine constants.
+        combine_constants(handler, &mut ir, &all_functions)?;
+        // And that in-turn enables more simplify-cfg.
+        simplify_cfg(handler, &mut ir, &all_functions)?;
+
+        // Remove dead definitions based on the entry points root set.
+        dce(handler, &mut ir, &entry_point_functions)?;
+        Ok(())
+    });
```
mohammadfawaz, line 430: This seems cleaner and more compact. Thanks!

otrho (approved):
Now that I see it in use I'm thinking `Handler` is too ambiguous.  `ErrorHandler` is pretty verbose but easier to understand in context.

mohammadfawaz:
Merged this so it doesn't get conflicts.

PR #3031, opened 2022-10-13
Implement generic traits.

emilyaherbert:
Closes #948

PR #3070, opened 2022-10-19
Expand the definition of `TraitMap` in preperation of trait constraints and generic traits.

emilyaherbert:
This PR expands the definitions of `TraitMap` to prepare it for use with trait constraints and generic traits (#3031). And it also makes an important change to the type system to eliminate redundant types.

For a high level explanation of the changes, take a look at this Sway code example (taken from a rustdoc addition in this PR):
```rust
    /// struct Data<T, F> {
    ///     first: T,
    ///     second: F,
    /// }
    ///
    /// impl<T, F> Data<T, F> {
    ///     fn get_first(self) -> T {
    ///         self.first
    ///     }
    ///
    ///     fn get_second(self) -> F {
    ///         self.second
    ///     }
    /// }
    ///
    /// impl<T> Data<T, T> {
    ///     fn switch(ref mut self) {
    ///         let first = self.first;
    ///         self.first = self.second;
    ///         self.second = first;
    ///     }
    /// }
    ///
    /// impl Data<u8, u8> {
    ///     fn add_u8(ref mut self, input: u8) {
    ///         self.first += input;
    ///         self.second += input;
    ///     }
    /// }
    ///
    /// impl Data<bool, bool> {
    ///     fn inner_and(self) -> bool {
    ///         self.first && self.second
    ///     }
    /// }
	///
    /// fn main() {
    ///     let mut foo = Data {
    ///         first: 1u8,
    ///         second: 2u8,
    ///     };
    ///
    ///     let a = foo.get_first();
    ///     let b = foo.get_second();
    ///     foo.switch();
    ///     let c = foo.add_u8(3u8);
    ///     let d = foo.inner_and();    // fails to compile
    ///
    ///     let mut bar = Data {
    ///         first: true,
    ///         second: false,
    ///     };
    ///
    ///     let e = bar.get_first();
    ///     let f = bar.get_second();
    ///     bar.switch();
    ///     let g = bar.add_u8(3u8);    // fails to compile
    ///     let h = bar.inner_and();
    ///
    ///     let mut baz = Data {
    ///         first: 1u8,
    ///         second: false,
    ///     };
    ///
    ///     let i = baz.get_first();
    ///     let j = baz.get_second();
    ///     baz.switch();               // fails to compile
    ///     let k = baz.add_u8(3u8);    // fails to compile
    ///     let l = baz.inner_and();    // fails to compile
    /// }
```

Sway allows us to define `impl` blocks that offer certain methods on certain types that are more or less "generic" than their counterparts---e.g.`Data<bool, bool>` has more methods defined for it than `Data<T, F>`, even though there are more definitions in the `impl` block for `Data<T, F>`. If we were to write down on paper all of the methods defined for `Data<bool, bool>`, these would be:
- `inner_and(self) -> bool`
- `switch(ref mut self)`
- `get_first(self) -> T`
- `get_second(self) -> F`

But because we know the type is `Data<bool, bool>` (and not `Data<T, F>`), we can actually translate these methods to be defined as:
- `inner_and(self) -> bool`
- `switch(ref mut self)`
- `get_first(self) -> bool`
- `get_second(self) -> bool`

Previously, the `TraitMap` was performing all of these translations on-the-fly when a method was retrieved from the `TraitMap` when type checking a method call. But this PR changes this so that now these translations are performed preemptively when types are added to the `TraitMap`. The benefits of this are that 1) the concepts are more straightforward and easier to debug, 2) it ensures that type inference can correctly resolve generic types across multiple invocations of different methods within the same `impl` block but with different types, and 3) fixes an edge case bug that was popping up in #3031 regarding `impl` blocks defined earlier in the program shadowing `impl` blocks defined later in the program.

This PR also introduces a change to eliminate redundant types and duplicate definitions in the `TraitMap` for improved performance. `TypeEngine` is extended with a `RwLock<HashMap<TypeInfo, TypeId>>` called `id_map`. When `TypeInfo`'s are added to the type engine in `insert_type`, if that `TypeInfo`'s exists in `id_map`, return that `TypeId` instead of creating a new `TypeId`. If that `TypeInfo` does *not* exist yet in `id_map` a new `TypeId` is created, then we check to see if it "can change" (super happy to discuss the naming on this!) and if it "can't change" then we add it to the `id_map`. To this end, going forward I'd like to think more about #3115, even if we decide that we don't want to go this route.

The types that can be added to the `id_map` are:
- boolean
- unsigned integers
- str
- b256
- ErrorRecovery
- structs that do not have type parameters (this could probably be expanded to including instances where the types of the type parameters are types that can be added to the `id_map`)
- enums that do not have type parameters (this could probably be expanded to including instances where the types of the type parameters are types that can be added to the `id_map`)

The types that cannot be added to the `id_map` are:
- unknown types
- generic types
- self
- contract
- numeric
- contract caller
- tuple (I actually think this one can be added)
- array (I actually think this one can be added)
- storage (I actually think this one can be added)
- custom types (I actually think these can be added if they do not have type parameters)

This PR err's on the cautious side until we can do further testing to include those additional variants.

1. Simplify the definition of the `TraitMap` struct to use a `im::HashMap` instead of a `im::Vector`, as the `Ref` type has been removed.
2. Extend `TraitMap` to account for the cases described above
3. Extend `TypeEngine` to reduce redundant types
4. Implement `ReplaceSelfType` on the `ty` definitions
5. Move the definition of `find_method_for_type` from `Root` to `Namespace`


JoshuaBatty (approved):

sway-core/src/semantic_analysis/namespace/trait_map.rs
```diff
@@ -5,92 +5,67 @@ use crate::{
...
-        }
-        self.trait_map
-            .push_back(((trait_name, incoming_type_id), methods_map));
+        let trait_methods: TraitMethods = methods
+            .into_iter()
+            .map(|method| (method.name.as_str().to_string(), method))
+            .collect();
+        let trait_impls: TraitImpls = vec![((trait_name, type_id), trait_methods)]
+            .into_iter()
+            .collect();
```
otrho, line 31: This is totally redundant?  Did it used to have a `.map()` in there or something?  I'm surprised Clippy isn't mentioning something, unless I'm wrong.

sway-core/src/semantic_analysis/namespace/trait_map.rs
```diff
@@ -5,92 +5,67 @@ use crate::{
...
-            );
+        for (key, other_trait_methods) in other.trait_impls.into_iter() {
+            match self.trait_impls.get_mut(&key) {
+                Some(trait_methods) => {
+                    trait_methods.extend(other_trait_methods.into_iter());
+                }
+                None => {
+                    self.trait_impls.insert(key, other_trait_methods);
+                }
+            }
```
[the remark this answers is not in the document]
emilyaherbert: Nice! I didn't know about this.

sway-core/src/semantic_analysis/namespace/trait_map.rs
otrho: Looking at `im::HashMap`, it also has a `singleton` method, so even more concise would be:
```rust
let trait_impls = TraitImpls::singleton((trait_name, type_id), trait_methods);
```

otrho:
There's a whole lot of commented out code in this PR, is that going to be addressed soon in another PR?

emilyaherbert:
> There's a whole lot of commented out code in this PR, is that going to be addressed soon in another PR?

Er, I'm not sure I know what commented out code you are referring to @otrho, could you point out which LOC specifically?

sway-core/src/type_system/type_info.rs
```diff
@@ -1093,6 +1233,35 @@ impl TypeInfo {
         }
     }
 
+    pub(crate) fn can_change(&self) -> bool {
+        // TODO: there might be an optimization here that if the type params hold
+        // only non-dynamic types, then it doesn't matter that there are type params
```
sezna, line 1238: by non-dynamic do you mean only resolved types? i.e. no generics?

mohammadfawaz (approved):

sway-core/src/type_system/type_info.rs
emilyaherbert: Yes! But also, I didn't know if I should call it that directly because the "can_change" method also checks for `TypeInfo::Contract` and such. What do you think?

sezna: Hm I think the term dynamic type invokes thoughts of scripting languages and dynamic typing, so I was confused by the comment initially. What about  "if the type params only hold concrete types"?

emilyaherbert: I like it!

sezna (approved):

emilyaherbert:
Agh--automerge got to it. I'll add a follow-up PR @sezna with the name change.

Issue #3069, opened 2022-10-19
Language Server needs to work with multiple projects simultaneously 


JoshuaBatty:
Currently, the language server can only work when a single project is loaded into the users editor. 

However, if the user opens up a projects folder, and then opens up a split window, with one project on the left and another project on the right, the language server will crash because of internal access to the global memory in the compiler. For example

```
thread 'main' panicked at 'index out of bounds: the len is 165726 but the index is 165778', sway-core/src/concurrent_slab.rs:38:9
```

This is because in the server, we call `forc_pkg::check()` on a project to produce the AST's. However, at the beginning of this function, all the global memory `lazy_static!`s are cleared. 

I feel like this is causing such a terrible UX experience that we should prioritize ripping out using global memory in the compiler once and for all :) There is an issue for this in #2063

mohammadfawaz:
Yeah we're prioritizing the removal of lazy statics; @Centril is going to work on this soon 😃

mohammadfawaz:
I'm going to close this in favor of https://github.com/FuelLabs/sway/issues/2063

PR #3072, opened 2022-10-19
LSP: code reorganize to support multiple projects simultaneously

JoshuaBatty:
The main change is instead of having a single `session: Arc<Session>`, we now have a `sessions: DashMap<PathBuf, Arc<Session>>,`. The `PathBuf` is the manifest directory of the sway project and the `Session` contains all the tokens, runnables etc.. for that project.

Although we are blocked currently to support multiple simultaneous projects, this PR fixes a current bug that now allows the users to jump between different projects in a workspace so long as they only have one window pane open. 

related: #3069 and #2063

### DIFF
--- a/sway-lsp/src/server.rs
+++ b/sway-lsp/src/server.rs
@@ -129,10 +129,19 @@ impl Backend {
             .parent()
             .ok_or(LanguageServerError::ManifestDirNotFound)?
             .to_path_buf();
-
+        
         let session = match self.sessions.get(&manifest_dir) {
             Some(item) => item.value().clone(),
             None => {
+                // TODO, remove this once the type engine no longer uses global memory: https://github.com/FuelLabs/sway/issues/2063
+                // Until then, we clear the current project and init with the new project.
+                // At least allows the user to switch between projects within a workspace but only if they
+                // have one pane open.
+                let _ = self.sessions.iter().map(|item| {
+                    let session = item.value();
+                    session.shutdown();
+                });
+
                 // If no session can be found, then we need to call init and inserst a new session into the map
                 self.init(uri)?;
                 self.sessions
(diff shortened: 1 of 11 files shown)


otrho (approved):
Fair enough.

sway-lsp/src/server.rs
```diff
@@ -124,15 +115,55 @@ fn capabilities() -> ServerCapabilities {
...
+                    session.shutdown();
+                });
+
+                // If no session can be found, then we need to call init and inserst a new session into the map
+                self.init(uri)?;
+                self.sessions
+                    .get(&manifest_dir) //
+                    .map(|item| item.value().clone())
+                    .expect("no session found even though it was just inserted into the map")
+            }
```
mitchmindtree, line 151: It looks like the previous session is not removed from `sessions`, which means if a user switches back the previous session will no longer be active? I guess this is fine for now as the real solution is addressing https://github.com/FuelLabs/sway/issues/2063 so you can actually have multiple simultaneous sessions.

sway-lsp/src/server.rs
```diff
@@ -155,12 +186,13 @@ impl Backend {
 #[tower_lsp::async_trait]
 impl LanguageServer for Backend {
     async fn initialize(&self, params: InitializeParams) -> jsonrpc::Result<InitializeResult> {
+        eprintln!("InitializeParams = {:#?}", params);
```
mitchmindtree, line 189: Leftover debug stuff? Or should this use `tracing::trace!`?

mitchmindtree (approved):

sway-lsp/src/server.rs
JoshuaBatty: Good spot, i'm going to just clear the map after shutting down the session in the meantime. Hopefully we can remove this block of code once #2063 lands and we can store multiple sessions.

PR #3075, opened 2022-10-19
Allow signing forc-client transactions without prompting

AlicanC:
We do not sign transactions while running the E2E tests, causing them to have similar IDs and collide. Some randomness could be inserted to prevent this, but signing the transactions already achieve the same result.

The only way `forc-client` could do signing was through stdio prompts, so support for signing with a provided key was also needed.

This PR:
- Moves tx related code used in `forc-client`'s `run` and `deploy` commands into `tx_util.rs`.
- Adds support for signing transactions with a provided `SecretKey` instead of prompting stdio.
- Signs transactions using `fuel-core`'s built-in key while testing.


forc-plugins/forc-client/src/ops/tx_util.rs
```diff
@@ -0,0 +1,181 @@
...
+    ) -> Result<Transaction>;
+}
+
+#[async_trait]
+impl TransactionBuilderExt for TransactionBuilder {
+    fn params(&mut self, params: TxParameters) -> &mut Self {
+        self.gas_limit(params.gas_limit).gas_price(params.gas_price)
+    }
+    fn add_contract(&mut self, contract_id: ContractId) -> &mut Self {
+        let input_index = self.inputs().len() as u8; // probably safe unless a user inputs > u8::MAX inputs
```
mitchmindtree, line 82:
```suggestion
        let input_index = u8::try_from(self.inputs().len()).expect("limit of 256 inputs exceeded");
```

forc-plugins/forc-client/src/ops/tx_util.rs
```diff
@@ -0,0 +1,181 @@
...
+        Ok(tx)
+    }
+}
+
+pub trait TransactionExt {
+    fn replace_witness(&mut self, witness_index: u8, witness: Witness) -> &mut Self;
+}
+
+impl TransactionExt for Transaction {
+    fn replace_witness(&mut self, index: u8, witness: Witness) -> &mut Self {
```
mitchmindtree, line 174: Perhaps `insert_witness` could be a better name for this, as it looks like this inserts a witness at the given index?

mitchmindtree (approved):
Really nice refactor @AlicanC, loving the red! :red_circle:

Just a couple suggestions but I'll leave these up to you, happy to re-approve if you want to address them!

Would be great to get @kayagokalp's eyes on this one too.

utACK

forc-plugins/forc-client/src/ops/tx_util.rs
AlicanC: Actually changed the behavior because replacement is needed here.

AlicanC:
Could have even more reds if TransactionBuilder supported interactive signing, etc., but didn't want to go that far and touch `fuel-tx` ATM. Might revisit this after I get to see how we use TB in other repos, etc.

kayagokalp (approved):
LGTM, tested to deploy a signed contract to local node with `--utxo-validation`.

PR #3079, opened 2022-10-19
Improve E2E test CLI

AlicanC:
- Added new args: `exclude`, `skip-until`, `abi-only`, `contract-only` and `first-only`:
```
❯ cargo run --bin=test -- --help
Usage: test [OPTIONS] [REGEX]

Arguments:
  [REGEX]  Only run tests matching this regex

Options:
  -e, --exclude <REGEX>     Exclude tests matching this regex
  -s, --skip-until <REGEX>  Skip all tests until a test matches this regex
      --abi-only            Only run tests with ABI JSON output validation [aliases: abi]
      --contract-only       Only run tests that deploy contracts [aliases: contract]
      --first-only          Only run the first test [aliases: first]
      --verbose             Print out warnings and errors [env: SWAY_TEST_VERBOSE=]
      --locked              Intended for use in `CI` to ensure test lock files are up to date
  -h, --help                Print help information
```
- Made `main()` async and refactored the rest accordingly so we don't have to create a new Tokio runtime for each `run` and `deploy`.
- Moved some assertions to `assert_matches`, which will print the value that wasn't matched instead of just printing that there was no match, which helps with debugging.
- Used `anyhow::Error`s instead of `String`s in more `Result::Err`s.


otrho (dismissed):
Nice!

mohammadfawaz (dismissed):
Very nice. These will certainly help a lot :)

What is the use case for the flag `--first-only`?

emilyaherbert (changes requested):
What does `--first` do exactly? Could you expand the help description? Other than that LGTM.

AlicanC:
@mohammadfawaz, @emilyaherbert, it just makes it run one test and stop. For example:
- Tests fail at `some/test/name`.
- You want to fix it and later continue running tests from that point so you start running:
  `test -s some/test/name --first`
- You work on the fix. Fail or pass, only the test you want to focus on runs because you have `--first`.
- When you are done fixing you remove `--first` and continue:
  `test -s some/test/name`
  
`test -s some/test/name --first` should be equal to just `test some/test/name` but `--first` is more convenient to add/remove compared to the `-s`.

What would be a fitting description for this? I'm happy to remove also.

emilyaherbert (commented):
Would it make sense to change `--first-only` to just `--only`? As in, it only runs the one test at the name you've given it? For example, I tend to want to run the `stdlib/vec` test on its own often, but this regex will match the `stdlib/vec_swap` first. In this case, would `--first-only` only run `stdlib/vec_swap`?

emilyaherbert:
If it doesn't make sense to change it, then I'm happy to approve the PR still!

AlicanC:
> Would it make sense to change `--first-only` to just `--only`? As in, it only runs the one test at the name you've given it? For example, I tend to want to run the `stdlib/vec` test on its own often, but this regex will match the `stdlib/vec_swap` first. In this case, would `--first-only` only run `stdlib/vec_swap`?

You should be able to achieve that with `test "stdlib/vec$"` already. (`--first-only` would only run `stdlib/vec_swap`, yes.)

emilyaherbert (dismissed):

mohammadfawaz:
Ahhh conflicts :(

mohammadfawaz (approved):

JoshuaBatty (approved):

PR #3080, opened 2022-10-19
Add const probing to delineated path type check

### LINKED ISSUE #3063: Assignment with call path does not work with constants
> I tested out with a script depending on a library with a pub const.
> 
> lib:
> 
> ```rust
> library test_lib;
> 
> pub const NUMBER: u64 = 10;
> ```
> 
> script that depends on the lib:
> 
> ```rust
> script;
> 
> fn main() {
>   let x = test_lib::NUMBER;
> }
> ```
> 
> and trying to build the script above, fails with:
> 
> ``` ^^^^^^ Could not find symbol "NUMBER" in this scope. ```
> 
> _Originally posted by @kayagokalp in https://github.com/FuelLabs/sway/pull/3016#discussion_r998063381_
>

kayagokalp:
closes #3063.

This enables usage of call paths that leads to const inside a declaration which would enable us to simplify given doc examples at #3016 . Example:

lib:

```rust
library test_lib;

pub const NUMBER: u64 = 10;
```

script that depends on the lib:

```rust
script;

fn main() {
  let x = test_lib::NUMBER;
}
```

is now possible.

I added const probing to the existing checks we have for `type_check_delineated_path` and tried to follow other checks there but my approach could be off too 😄


test/src/e2e_vm_tests/test_programs/should_pass/test_contracts/multiple_impl/json_abi_oracle.json
```diff
@@ -20,7 +20,7 @@
       }
     },
     {
-      "logId": 165711,
```
kayagokalp, line 23: Not sure the reason of this change 🤔

sezna: @mohammadfawaz is working on stabilizing it. Small changes to the code result in all log ids being updated. Not ideal but hopefully temporary.

mohammadfawaz: I knew this was gonna cause problems but I didn't know that every other PR is going to be affected 😢

mitchmindtree (commented):
Looking great! Just needs a test or two :nerd_face:

kayagokalp:
@mitchmindtree @Centril thanks for the review! Should be good to go now!

mohammadfawaz:

test/src/e2e_vm_tests/test_programs/should_pass/language/const_decl_with_call_path/src/main.sw
```diff
@@ -0,0 +1,8 @@
+script;
+
+dep test_lib;
+
+fn main() -> u64 {
+  let x = test_lib::NUMBER;
```
mohammadfawaz, line 6: Would be nice to also see an example with a longer path. You can try a constant from the standard library for example: `std::constants::ZERO_B256`. Also a constant with a non-primitive type such as `std::constants::BASE_ASSET_ID`. And then use `assert`s to check each of them if you want.

test/src/e2e_vm_tests/test_programs/should_pass/language/const_decl_with_call_path/src/main.sw
```diff
@@ -0,0 +1,8 @@
+script;
```
mohammadfawaz, line 1: The test should probably be formatted.

mohammadfawaz (commented):
Just some minor nits.

test/src/e2e_vm_tests/test_programs/should_pass/language/const_decl_with_call_path/src/main.sw
kayagokalp: I added the extra checks to the same test, would you prefer having them in a second separate test?

mohammadfawaz (approved):

test/src/e2e_vm_tests/test_programs/should_pass/language/const_decl_with_call_path/src/main.sw
mohammadfawaz: One test is perfectly fine :) Thanks!

mitchmindtree (approved):

Issue #2917, opened 2022-10-02
Add CI formatter checks for `sway-lib-std` & `sway-lib-core`


eureka-cpu:
Just like how we check the `examples` directory, we should also be checking all other Sway code as well

JoshuaBatty:
added in #5404

PR #3082, opened 2022-10-19
Expose CROO opcode through the std-lib 

### LINKED ISSUE #3081: Create std-lib wrapper around CROO opcode
The std-lib should expose the CROO opcode via a convenient wrapper. 

Tentatively putting this in a module called `external` (but open to suggestions. Especially since a contract can just as easily use this to read its _own_ bytecode, as indeed the test for this function does).

simonr0204:
Closes #3081


sway-lib-std/src/external.sw
```diff
@@ -0,0 +1,13 @@
...
+
+use ::contract_id::ContractId;
+
+/// Get the root of the bytecode of the contract at 'contract_id'.
+pub fn bytecode_root(contract_id: ContractId) -> b256 {
+    asm(buffer, target: contract_id.value) {
+        move buffer sp;
+        cfei i32;
+        croo buffer target;
+        buffer: b256
```
otrho, line 11: Now that Sway has calls and call-frames we can't leak anything onto the stack any more.  All `cfei`s must be paired with `cfsi`s, or you use the heap.

In this case though I think you could create a local `b256`, take its address and populate it with `croo` and return that.

In the ongoing effort to remove ASM blocks from the libs, rather than introduce new ones, :wink: this could be made into a compiler intrinsic, like `__gtf`.

sway-lib-std/src/b512.sw
```diff
@@ -27,10 +27,7 @@ impl From for B512 {
     }
 
     fn into(self) -> (b256, b256) {
-        (
-            (self.bytes)[0],
-            (self.bytes)[1],
-        )
+        ((self.bytes)[0], (self.bytes)[1], )
```
mohammadfawaz, line 30: If you're going to format these, do you mind removing the `,` at the end and reformat again? I think the formatter will then do the right thing and give you `((self.bytes)[0], (self.bytes)[1])`

test/src/e2e_vm_tests/test_programs/should_pass/language/logging/json_abi_oracle.json
```diff
@@ -12,63 +12,63 @@
   ],
   "loggedTypes": [
     {
-      "logId": 165743,
+      "logId": 165753,
```
mohammadfawaz, line 15: You will no longer need these after https://github.com/FuelLabs/sway/pull/3083

sway-lib-std/src/b512.sw
simonr0204: I can do that. I just hit format on the whole repo and a load of stuff wasn't formatted. I guess this is the first PR since the latest formatter release ?

sway-lib-std/src/lib.sw
```diff
@@ -10,6 +10,7 @@ dep mem;
 dep alloc;
 dep contract_id;
 dep constants;
+dep external;
```
mohammadfawaz, line 13: What is the reasoning for calling the library `external`?

test/src/e2e_vm_tests/test_programs/should_pass/language/logging/json_abi_oracle.json
simonr0204: That is very good to hear :)

sway-lib-std/src/lib.sw
simonr0204: Yeah, see https://github.com/FuelLabs/sway/issues/3081. I couldn't think of a better name. This reads the bytecode root of an "external" contract (although as mentioned in the issue, not strictly so)

I imagine this module eventually containing the wrappers for `CCP` and `CSIZ` , (and _maybe_ `LDC`) . Can bikeshed on the name in the issue.

sway-lib-std/src/b512.sw
simonr0204: But if I change them in a way that contradicts the formatter, won't CI fail ?

mohammadfawaz: Currently CI won't fail. See https://github.com/FuelLabs/sway/issues/2917. Also, what I'm suggesting is removing the `,` and then running the formatter again - the formatter won't do bad things in that case. Anyways, none of this matters that much :D Totally up to you how you want to do with this 😄

sway-lib-std/src/lib.sw
mohammadfawaz: How about `bytecode.sw` or something like that? All of these ops seem to be doing something related bytecode.

mohammadfawaz:
Merge with `master` and resolved conflicts by reverting to the JSON ABI on `master` which are now more stable.

sway-lib-std/src/b512.sw
mohammadfawaz: Alright I removed the commas as I was resolving the conflicts in the JSON files.

mohammadfawaz (approved):

otrho (approved):

Issue #2711, opened 2022-09-02
Follow up tasks on the generation of logged types in the JSON ABI


mohammadfawaz:
Referring to this PR: https://github.com/FuelLabs/sway/pull/2689

- [ ] Standardize the log IDs. Right now, they are simply the type IDs themselves from the type engine but we should standardize those like we do in other places in the JSON.
- [ ] Enhance the E2E testing engine to also check for `Log` receipts.
- [ ] Investigate a better approach for `collect_types_metadata` such as making it return a `struct` that contains two separate vectors. This requires implementing `FromIterator` for this struct and a refactor of that whole method everywhere.

mohammadfawaz:
Closing this now that https://github.com/FuelLabs/sway/pull/3083 is done. Other items don't need to be done really.

PR #3083, opened 2022-10-20
Update how log IDs are generated

mohammadfawaz:
Closes one task in https://github.com/FuelLabs/sway/issues/2711

### DIFF
--- a/sway-core/src/ir_generation/function.rs
+++ b/sway-core/src/ir_generation/function.rs
@@ -39,6 +39,7 @@ pub(super) struct FnCompiler {
     returns_by_ref: bool,
     lexical_map: LexicalMap,
     recreated_fns: HashMap<(Span, Vec<TypeId>, Vec<TypeId>), Function>,
+    logged_types: HashMap<TypeId, usize>,
 }
 
 impl FnCompiler {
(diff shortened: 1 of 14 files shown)


sway-core/src/ir_generation/function.rs
```diff
@@ -63,6 +66,7 @@ impl FnCompiler {
             returns_by_ref,
             recreated_fns: HashMap::new(),
             current_fn_param: None,
+            logged_types_map: logged_types_map.clone(),
```
otrho, line 69: Does it need to be cloned?

test/src/e2e_vm_tests/test_programs/should_pass/language/logging/json_abi_oracle.json
```diff
@@ -12,63 +12,63 @@
   ],
   "loggedTypes": [
     {
-      "logId": 167472,
+      "logId": 0,
```
otrho, line 15: A bit saner!  :slightly_smiling_face:

otrho (approved):

sway-core/src/ir_generation/function.rs
mohammadfawaz: So yeah `logged_types_map` is a reference and changing it to be otherwise sent me down a rabbit hole. There's probably a way to do it?

otrho: But it's a const reference, and at this stage it's never mutable again, so can't you just pass the ref around?

emilyaherbert (approved):
Thanks for adding this change @mohammadfawaz 👍

PR #3084, opened 2022-10-20
LSP: Show formatted documentation on hover

### LINKED ISSUE #2849: LSP Hover: Show formatted MarkDown documentation
eg. 

Screen Shot 2022-09-26 at 1 45 50 pm

JoshuaBatty:
This PR renders adds MarkDown documentation hover support. It works by using the doc attributes that can be found for the declaration of the currently hovered token in the editor.

We can also have blocks of ```sway code in the docs and it will render those sections out with correct sway syntax highlighting :) 

I've borrowed and modified the `markdown.rs` from `rust-analyzer` to get this to work. For example, consider we have the following doc comments above a type in our program.

```
/// Struct holding:
///
/// 1. A `value` of type `NumberOrString`
/// 2. An `address` of type `u64`
/// 2. A `name` of type `str[11]`
///
/// Lets look at some formatted sway code below.
///
/// ```sway
/// let value = NumberOrString::Number(20);
/// let address = 0; 
/// let name = "my_contract";
/// ```
/// 
/// Yay, we now have rich markdown formatting of doc comments!
struct Data {
    /// The `value` field in `Data`
    value: NumberOrString,
    /// The `address` field in `Data`
    address: u64,
    /// The `name` field in `Data`
    name: str[11],
}
```

If we hover over the `Data` type when it's used elsewhere in the program, we get the following hover box appearing.

Screen Shot 2022-10-20 at 3 14 38 pm

closes #2849


sway-lsp/src/capabilities/hover.rs
```diff
@@ -15,27 +18,51 @@ use sway_core::{
     language::{parsed::Declaration, ty, Visibility},
 };
 use sway_types::{Ident, Spanned};
-use tower_lsp::lsp_types::{Hover, HoverContents, MarkupContent, MarkupKind, Position, Url};
+use tower_lsp::lsp_types::{self, Position, Url};
 
-pub fn hover_data(session: Arc<Session>, url: Url, position: Position) -> Option<Hover> {
-    if let Some((_, token)) = session.token_at_position(&url, position) {
+pub fn hover_data(session: Arc<Session>, url: Url, position: Position) -> Option<lsp_types::Hover> {
```
sdankel, line 23: nit: IMO, it's cleaner with `Hover` being explicitly imported from `lsp_types`

sway-lsp/src/server.rs
```diff
@@ -794,16 +795,16 @@ mod tests {
             json!({
                 "contents": {
                     "kind": "markdown",
-                    "value": "```sway\nstruct Data\n```"
+                    "value": " Struct holding:\n\n 1. A `value` of type `NumberOrString`\n 2. An `address` of type `u64`"
```
sdankel, line 786: Is this extra space intentional?

```suggestion
                    "value": "Struct holding:\n\n 1. A `value` of type `NumberOrString`\n 2. An `address` of type `u64`"
```

sway-lsp/src/utils/markdown.rs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            in_code_block ^= true;
+
+            if in_code_block {
+                is_sway = is_sway_fence(header);
+
+                if is_sway {
+                    line = "```sway";
+                }
+            }
```
sdankel, line 26: should `is_sway` be reset to false if `in_code_block` is false?

sdankel (approved):

sway-lsp/src/capabilities/hover.rs
```diff
@@ -15,27 +18,51 @@ use sway_core::{
...
+    if let Some((ident, token)) = session.token_at_position(&url, position) {
+        let range = get_range_from_span(&ident.span());
         if let Some(decl_ident) = session.declared_token_ident(&token) {
             if let Some(decl_token) = session
                 .token_map()
                 .get(&to_ident_key(&decl_ident))
                 .map(|item| item.value().clone())
```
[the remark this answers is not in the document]
JoshuaBatty: Great spot. Thanks.

sway-lsp/src/utils/markdown.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// stripped down version of https://github.com/rust-lang/rust/blob/392ba2ba1a7d6c542d2459fb8133bebf62a4a423/src/librustdoc/html/markdown.rs#L810-L933
+pub fn is_sway_fence(s: &str) -> bool {
+    let mut seen_sway_tags = false;
+    let mut seen_other_tags = false;
+
+    let tokens = s
+        .trim()
+        .split(|c| c == ',' || c == ' ' || c == '\t')
```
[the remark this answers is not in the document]
JoshuaBatty: I keep forgetting about this macro, thanks.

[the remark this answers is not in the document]
otrho: Curious, why is this better?

sway-lsp/src/capabilities/hover.rs
JoshuaBatty: I definitely agree with this as well but have made an exception for places that are using types from the `lsp_types` crate. 1. It makes it obvious that this is returning a type that is on the boundary between the server and the client. 2. It avoids naming conflicts as the server and the types in `lsp_types` share a lot of similar names. I noticed this is also the way rust-analyzer is working, so decided to continue with that convention.

sway-lsp/src/server.rs
JoshuaBatty: That space is there because it's reading from this doc comment
`/// Struct holding:` 

If the doc comment had of been written like `///Struct holding:` then there would be no white space at the beginning here. 

I guess we could have a discussion how we want to treat this, i.e do we want to autostrip white space so it's all identical? I'll leave that for another PR though.

sway-lsp/src/utils/markdown.rs
JoshuaBatty: This function accounts for different types of code blocks that may or may not be `sway` code. I also borrowed this function from rust-analyzer. It seems to be working as intended. Good spot though.

sway-lsp/src/utils/markdown.rs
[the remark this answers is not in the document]
otrho: Fair enough.

otrho (approved):

mitchmindtree (approved):

PR #3091, opened 2022-10-20
Type engine bug fixes and improvements.

emilyaherbert:
This PR introduces a number of improvements + minor bug fixes to the type system that are helpful for #3031.

1. If the `copy_types` method receives a `type_mapping` that is empty, bail early instead of continuing recursion.
2. Improve the `TypeMapping::find_match` method such that a match is only created when absolutely necessary, instead of being created every time `TypeMapping::find_match` is called. This leads to a decrease in the number of `TypeId`'s generated during type checking by a factor of ~22x. For example, previously the standard library was creating ~220,000 `TypeId`'s 😱 , but now it creates ~10,000 👼  .
3. Modify the `monomorphize` method to create a `TypeMapping` with the `type_arguments` directly, instead of creating the `TypeMapping` first and then unifying with the `type_arguments`. This allows the `TypeMapping` to hold the original `TypeId`s of the `type_arguments`, instead of new `TypeId`'s that point to the same `TypeInfo` in the `TypeEngine`. This means that applications of `copy_types` with that `TypeMapping` will copy the same `TypeId` that is in the parameter, reducing complexity significantly.
4. Introduce the `unify_adt` method, which is a helper function for conveniently accessing `unify`. It should help with reducing developer overhead during implementation, as it is a little non-intuitive what the intended behavior is without the helper function. There is a pretty good rustdoc comment that will help explain what this function does in more detail.
5. Adding these changes exposed a bug relating to `ref mut` and function inlining, for which @otrho added a commit to bypass this bug for now [`f3d79ec` (#3091)](https://github.com/FuelLabs/sway/pull/3091/commits/f3d79ec715f3ac87d9ca2d95c867b5fbc8138ac8)

Inspired by the description of #3095, here are the relative performance improvements to the `stdlib/vec` test from this PR compared to `master`.

`master`:
```bash
$ ./target/release/forc build --time-phases --path ./test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec
  Time elapsed to produce `sway_core::BuildConfig`: 4.008µs
  Time elapsed to compile to ast: 30.141887ms
  Time elapsed to generate JSON ABI program: 154ns
  Compiled library "core".
  Time elapsed to produce `sway_core::BuildConfig`: 2.159µs
  Time elapsed to compile to ast: 461.510248ms
  Time elapsed to generate JSON ABI program: 219ns
  Compiled library "std".
  Time elapsed to produce `sway_core::BuildConfig`: 2.627µs
  Time elapsed to compile to ast: 2.722139253s
  Time elapsed to generate JSON ABI program: 17.576µs
  Time elapsed to compile ast to asm: 86.944279618s
  Time elapsed to compile asm to bytecode: 11.542455721s
  Compiled script "vec".
  Bytecode size is 278480 bytes.
```

this PR:
```bash
$ ./target/release/forc build --time-phases --path ./test/src/e2e_vm_tests/test_programs/should_pass/stdlib/vec
  Time elapsed to produce `sway_core::BuildConfig`: 4.169µs
  Time elapsed to compile to ast: 31.878877ms
  Time elapsed to generate JSON ABI program: 247ns
  Compiled library "core".
  Time elapsed to produce `sway_core::BuildConfig`: 2.247µs
  Time elapsed to compile to ast: 231.545843ms								// down 229.9ms
  Time elapsed to generate JSON ABI program: 221ns
  Compiled library "std".
  Time elapsed to produce `sway_core::BuildConfig`: 2.669µs
  Time elapsed to compile to ast: 2.491381456s
  Time elapsed to generate JSON ABI program: 16.036µs
  Time elapsed to compile ast to asm: 85.21828356s
  Time elapsed to compile asm to bytecode: 12.190644522s
  Compiled script "vec".
  Bytecode size is 278480 bytes.
```


sezna:
Could the parent trait define `fn copy_types()` and then call `copy_types_inner()` iff type mapping isnt empty? Then it only has to be written once.

otrho (dismissed):

mohammadfawaz:

mohammadfawaz (approved):

mohammadfawaz:
> This leads to a decrease in the number of TypeId's generated during type checking by a factor of ~22x

Nice... good stuff 🙏

otrho (approved):

PR #3098, opened 2022-10-21
Bug fixes for type checking trait declarations.

emilyaherbert:
This PR is a series of bug fixes for type checking trait declarations:

1. Remove the `type_check_method_parameter` function and instead call `ty::TyFunctionDeclaration::type_check` directly. These two functions had nearly the same behavior anyway, except that `type_check_method_parameter` also had a redundant check for checking to see if the types of the arguments were out of scope.
2. Previously we were hard-coding some methods to be public by default and skipping this hard-coding step for other methods, depending on certain factors, like if the method was defined on the interface surface or not. This PR standardizes this so that we _always hard-code every method to be public_. This doesn't change anything though in terms of the Sway that is already written---basically before this PR, any method that a user would write inside an `impl` block was hacked into public (even if the user had not written `pub`), but now all methods are public. This can be turned off easily in the future when we want to. https://github.com/FuelLabs/sway/blob/99df7b119f81bda7e0a982245585a047bce396d5/sway-core/src/semantic_analysis/ast_node/mod.rs#L551
3. When type checking `ty::TyTraitFn`, use a separate namespace to avoid overwriting type parameters.
4. When type checking a `ty::TyTraitDeclaration`, use just one instance of `insert_type(TypeInfo::SelfType)` instead of many instances.

TODOs for me:
- [ ] when this PR goes in, create an issue about turning on method visibility
- [x] create an issue for removing information about `pub` and `mut` from ir
	#3103

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/declaration/trait.rs
+++ b/sway-core/src/semantic_analysis/ast_node/declaration/trait.rs
@@ -47,8 +47,9 @@ impl ty::TyTraitDeclaration {
         }
 
         // A temporary namespace for checking within the trait's scope.
+        let self_type = insert_type(TypeInfo::SelfType);
         let mut trait_namespace = ctx.namespace.clone();
-        let mut ctx = ctx.scoped(&mut trait_namespace);
+        let mut ctx = ctx.scoped(&mut trait_namespace).with_self_type(self_type);
 
         // type check the interface surface
         let interface_surface = check!(
@@ -81,7 +82,7 @@ impl ty::TyTraitDeclaration {
                 suffix: name.clone(),
                 is_absolute: false,
             },
-            insert_type(TypeInfo::SelfType),
+            self_type,
             trait_fns
                 .iter()
                 .map(|x| x.to_dummy_func(Mode::NonAbi))
@@ -89,7 +90,6 @@ impl ty::TyTraitDeclaration {
         );
 
         // check the methods for errors but throw them away and use vanilla [FunctionDeclaration]s
-        let mut ctx = ctx.with_self_type(insert_type(TypeInfo::SelfType));
         let _methods = methods
             .iter()
             .map(|method| {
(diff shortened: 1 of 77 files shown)


sway-ir/src/parser.rs
```diff
@@ -52,16 +52,34 @@ mod ir_builder {
                 }
 
             rule fn_decl() -> IrAstFnDecl
-                = "fn" _ name:id() _ selector:selector_id()? _ "(" _
-                      args:(block_arg() ** comma()) ")" _ "->" _ ret_type:ast_ty()
-                          metadata:comma_metadata_idx()? "{" _
-                      locals:fn_local()*
-                      blocks:block_decl()*
-                  "}" _ {
+                = "pub fn" _ name:id() _ selector:selector_id()? _ "(" _
```
mohammadfawaz, line 55: I don't feel that the IR should care about things like `pub`. In my view, `pub` is only relevant for the frontend, similarly to `mut`. After they're consumed in the frontend, `pub` and `mut` become obsolete and they should not really carry any weight in IR. I'll let @otrho make the final call on this though.

mohammadfawaz (commented):
Just a small comment about IR.

We certainly need to organize our thoughts around when `pub` is needed for methods, following Rust of course. Once you create the follow up issue, do you mind laying out the rules? (i.e. trait methods v.s. impl methods, etc.)

sway-core/src/semantic_analysis/ast_node/declaration/trait.rs
```diff
@@ -85,20 +82,26 @@ impl ty::TyTraitDeclaration {
...
-        let _methods = check!(
-            type_check_trait_methods(ctx, methods.clone()),
-            vec![],
-            warnings,
-            errors
-        );
+        let _methods = methods
+            .iter()
+            .map(|method| {
+                check!(
```
otrho, line 96: If `_methods` isn't used again and all we're doing here is the checks, then it might as well be a for-loop and not `collect()` into a new (unused) `Vec`.

sway-ir/src/parser.rs
otrho, line 55: Yep, we want to remove `pub` and `mut` from the IR serialisation as with hindsight it's apparent that they aren't useful at this stage.  It's simpler to let the compiler worry about that stuff and to assume that all data are mutable and all symbols are public once we hit IR.

So I'd remove all these changes and instead go into `printer.rs` and fix the output of `function_to_doc` to not care about `pub`.

otrho (approved):

PR #3109, opened 2022-10-24
Add support for ASM control flow to/from addresses which don't fit in immediate offsets.

### LINKED ISSUE #2385: Use dynamic jump instructions when the offset is too large to fit in the immediate of static jumps
This is a bit more complicated than it seems because the register allocator does not handle dynamic jumps. In fact, I couldn't find anything about that in the literature. Dynamic jumps make "liveness analysis" really difficult to perform because we just don't know, at compile time, how the control flow graph looks like and therefore we can't build an interference graph. The good news is that we can introduce dynamic jumps _after_ register allocation: 
1. Modify the assembly pre-allocation to include "organizational ops" (just like we do before realizing jump labels) that look like `jnzi`, `jnei`, and `ji` but allow a `u64` jump offset.
2. Run the register allocator as usual. Some minor changes will be required because the organizational ops above won't be like regular virtual ops anymore
3. Replace jumps with either static jumps or dynamic jumps depending on how large the offsets are. We will need extra logic before dynamic jumps to actually store the offsets into registers.

The above works and will be more efficiently than any attempt to do register allocation _with_ dynamic jumps.

otrho:
- Introduce a new control flow op `LoadLabel` which reads an address from the data section.
- Convert `JI`, `JNZI`, `JNEI` and `MoveAddress` to use `LoadLabel` and `JMP` or `JNE` when required.
- Introduce a `BLOB` instruction which is for testing purposes only, and efficiently & conveniently inserts a raft of `NOOP`s into the binary.

Closes #2385.


mohammadfawaz (approved):
Nice 🚀

sezna (dismissed):
looks great :shipit:

sezna:
[Lock file issue.](https://github.com/FuelLabs/sway/actions/runs/3314761038/jobs/5474462109#step:6:204)

mohammadfawaz (approved):

sezna (approved):

PR #3117, opened 2022-10-24
[WIP] feat: building/deploying all members of the workspace with a single command

### LINKED ISSUE #1777: Add support for `[workspace]`s to `forc`
## Context

We are progressively coming across more and more cases where it would be great if `forc` could be aware of collections of packages and the way they interact, rather than just reasoning about one package at a time.

**Cargo Workspaces**

Cargo uses the concept of `[workspace]`s to achieve something similar and we may want to draw inspiration from it. However, it's worth noting that cargo's original major motivator was to have the ability for a bunch of related crates to share a compiler output artifact directory (`target/`) to cut down on the epicly long crate build times. Our major motivations are a bit different, and we should keep this in mind when considering the design

### Deploying multiple packages

It is not uncommon for a project to manage a set of contracts that should be deployed simultaneously. Currently, this is a PITA with `forc` as `forc` is only aware of a single package at a time.

It would be nice if `forc` had the ability to reason about a set of packages, and to be able to reason about not just build `[dependencies]`, but **deployment** dependencies and related ordering too. Deployment ordering is important as if contract foo requires interacting with contract bar, then contract foo will likely need contract bar's address which can't be known until it is deployed.

### Testing multiple packages

Currently `forc test` is a thin layer on top of `cargo test` as it only really supports integration tests via the Rust SDK. One common issue with this is that the cargo project root is not always the same as the forc project root, occasionally leading to strange and unexpected behaviour from `forc test`. This is because `cargo` considers the project root to be where the cargo workspace manifest lives, while `forc` considers the project root to be where the forc package manifest lives. This is discussed briefly in #1375 and #1521, and #1612 *might* be related.

If `forc` had the concept of `[workspace]`s, it would at least enable users to setup their project so that both `forc` and `cargo` are aware of the same workspace root, and might make it a bit easier to make `forc test` behaviour more intuitive.

---

cc @adlerjohn finally opening this as a rough tracking issue to get started and somewhere to discuss this. Feel free to add more motivations or details whether in this OP or in new comments!

I remember us having a chat about passing inputs between packages within a hypothetical workspace a while ago in another issue, but can't for the life of me remember where that issue is :thinking: 

Once we have a clearer outline of all the motivations/goals we can think of, I'll have a go at a rough potential design (if no-one beats me to it).

---

# TODO

- [x] #2893
- [x] #2930
- [x] #2931
- [x] #2929
- [ ] Document how workspaces work in the Sway Book.

### LINKED ISSUE #2929: Refactor `forc-client` to be aware of workspace manifest files

kayagokalp:
[WIP]

This is currently a PoC of the in-order building of the workspace members.

closes #2929.
closes #1777.

### DIFF
--- a/forc-pkg/src/manifest.rs
+++ b/forc-pkg/src/manifest.rs
@@ -549,16 +549,34 @@ impl WorkspaceManifestFile {
         Self::from_file(path)
     }
 
+    /// Return an iterator over the names of declared workspace members
     pub fn members(&self) -> impl Iterator<Item = &String> + '_ {
         self.members.iter()
     }
 
+    /// Return an iterator over the paths of declared workspace members
     pub fn member_paths(&self) -> Result<impl Iterator<Item = PathBuf> + '_> {
         let parent = self
             .path
             .parent()
             .ok_or_else(|| anyhow!("Cannot get parent dir of {:?}", self.path))?;
         Ok(self.members.iter().map(|member| parent.join(member)))
+    }
+
+    /// The path to the `Forc.toml` from which this manifest was loaded.
+    ///
+    /// This will always be a canonical path.
+    pub fn path(&self) -> &Path {
+        &self.path
+    }
+
+    /// The path to the directory containing the `Forc.toml` from which this manfiest was loaded.
+    ///
+    /// This will always be a canonical path.
+    pub fn dir(&self) -> &Path {
+        self.path()
+            .parent()
+            .expect("failed to retrieve manifest directory")
     }
 }
 
--- a/forc-pkg/src/manifest.rs
+++ b/forc-pkg/src/manifest.rs
@@ -246,12 +246,16 @@ impl PackageManifest {
     pub fn from_file(path: &Path) -> Result<Self> {
         let manifest_str = std::fs::read_to_string(path)
             .map_err(|e| anyhow!("failed to read manifest at {:?}: {}", path, e))?;
+        let mut warnings = Vec::new();
         let toml_de = &mut toml::de::Deserializer::new(&manifest_str);
         let mut manifest: Self = serde_ignored::deserialize(toml_de, |path| {
             let warning = format!("  WARNING! unused manifest key: {}", path);
-            println_yellow_err(&warning);
+            warnings.push(warning);
         })
         .map_err(|e| anyhow!("failed to parse manifest: {}.", e))?;
+        for warning in warnings {
+            println_yellow_err(&warning);
+        }
         manifest.implicitly_include_std_if_missing();
         manifest.implicitly_include_default_build_profiles_if_missing();
         manifest.validate()?;
@@ -583,14 +587,18 @@ impl WorkspaceManifestFile {
 impl WorkspaceManifest {
     /// Given a path to a `Forc.toml`, read it and construct a `WorkspaceManifest`.
     pub fn from_file(path: &Path) -> Result<Self> {
+        let mut warnings = Vec::new();
         let manifest_str = std::fs::read_to_string(path)
             .map_err(|e| anyhow!("failed to read manifest at {:?}: {}", path, e))?;
         let toml_de = &mut toml::de::Deserializer::new(&manifest_str);
         let manifest: Self = serde_ignored::deserialize(toml_de, |path| {
             let warning = format!("  WARNING! unused manifest key: {}", path);
-            println_yellow_err(&warning);
+            warnings.push(warning);
         })
         .map_err(|e| anyhow!("failed to parse manifest: {}.", e))?;
+        for warning in warnings {
+            println_yellow_err(&warning);
+        }
         Ok(manifest)
     }
 
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -2118,15 +2118,18 @@ fn build_workspace_with_options(
     manifest: &WorkspaceManifestFile,
     build_options: BuildOptions,
 ) -> Result<()> {
+    let workspace_dir = PathBuf::from(manifest.dir());
     // Find root member of the workspace and build it.
-    let workspace_root = find_root_workspace_member(manifest, build_options)?;
-    println!("workspace root is {workspace_root}");
+    let root_member = find_root_workspace_member(manifest, &build_options)?;
+    let root_member_path = workspace_dir.join(root_member);
+    let root_member_manifest = PackageManifestFile::from_dir(&root_member_path)?;
+    build_package_with_options(&root_member_manifest, build_options)?;
     Ok(())
 }
 
 fn find_root_workspace_member(
     manifest: &WorkspaceManifestFile,
-    build_options: BuildOptions,
+    build_options: &BuildOptions,
 ) -> Result<String> {
     let locked = build_options.locked;
     let offline = build_options.offline_mode;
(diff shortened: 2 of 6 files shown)


forc-pkg/src/manifest.rs
```diff
@@ -549,30 +553,52 @@ impl WorkspaceManifestFile {
...
     /// Given a path to a `Forc.toml`, read it and construct a `WorkspaceManifest`.
     pub fn from_file(path: &Path) -> Result<Self> {
+        let mut warnings = Vec::new();
         let manifest_str = std::fs::read_to_string(path)
             .map_err(|e| anyhow!("failed to read manifest at {:?}: {}", path, e))?;
         let toml_de = &mut toml::de::Deserializer::new(&manifest_str);
         let manifest: Self = serde_ignored::deserialize(toml_de, |path| {
             let warning = format!("  WARNING! unused manifest key: {}", path);
-            println_yellow_err(&warning);
+            warnings.push(warning);
```
kayagokalp, line 596: I had to hack this like that because to understand whether we have a package manifest or a workspace manifest, I am trying to create a `PackageManifestFile` from the dir I am checking. If that fails I am trying to see if current manifest is a `WorkspaceManifestFile`. If the manifest I am checking is indeed a `WorkspaceManifest`, checking to see if it was a `PackageManifestFile` prints out warnings of `unused key: members` as unused warnings are printed before checking success of parsing.

forc-pkg/src/pkg.rs
```diff
@@ -2090,12 +2090,92 @@ pub const SWAY_BIN_HASH_SUFFIX: &str = "-bin-hash";
...
+    let offline = build_options.offline_mode;
+    let members: HashSet<_> = manifest.members().cloned().collect();
+    for (member, member_path) in manifest.members().zip(manifest.member_paths()?) {
+        info!("Checking {member}");
+        let member_pkg_manifest = PackageManifestFile::from_dir(member_path.as_path())?;
+        let build_plan = BuildPlan::from_lock_and_manifest(&member_pkg_manifest, locked, offline)?;
+        let graph = build_plan.graph();
+        let discovered_deps = discover_contract_dependencies(graph, &members, member.clone())?;
+        // If we can find all members of the workspace from this member, this member is the root of the workspace.
+        // Meaning that building/deploying this member will ensure whole workspace will be handled in the
```
kayagokalp, line 2144: By stating "this member", I am referring to the graph of this member. I will improve the wordings here

forc-pkg/src/pkg.rs
```diff
@@ -2090,12 +2090,92 @@ pub const SWAY_BIN_HASH_SUFFIX: &str = "-bin-hash";
...
+        // correct order.
+        if discovered_deps.len() == members.len() {
```
kayagokalp, line 2146: This approach works as long as there are no 2 sub-graphs that does not have any connection in between them. Meaning that any disjoint two sets present in the workspace may create problems. But if there are two disjoint sets of dependencies in the workspace, I am not sure if they should be in the same workspace. Because to have a situation like that there should be two sets of workspace members such that they do not have dependencies in between these two subsets including both `lib` and `contract` dependencies. So it comes to the relevance of having two totally independent group of projects in the same workspace. In my opinion that should be something that we warn the user about and not even allow but there might be some use cases that I am not aware of 🤔

mitchmindtree (commented):
Solid start @kayagokalp!

I think your WIP here reveals we need to get a little deeper into the design first.

Perhaps we could start by splitting up some of this logic into smaller distinct functions, and maybe even separate PRs?

# 1. Find manifest type

The first step is determining whether the given path implies we're working with a package manifest or a workspace manifest. It would be great if we could expose this as a distinct public function so that it can be handled at a higher-level than the build function, as we'll also want the same logic for other commands. Perhaps we could add something like this to `manifest.rs`:

```rust
pub enum Manifest {
    Package(PackageManifest),
    Workspace(WorkspaceManifest),
}

impl Manifest {
    /// Returns a `PackageManifest` if the path is within a package directory, otherwise
    /// returns a `WorkspaceManifest` if within a workspace directory.
    pub fn from_path(path: &Path) -> Result<Self> {
        // ...
    }
}
```

# 2. Member compilation order

We should separate the logic for determining the dependency order of a workspace's `members` from the `build` process so that it can be shared between other commands like `check` or `test`, or 3rd party plugins. E.g. something like:

```rust
/// Returns all workspace members ordered from dependencies to dependents.
///
/// The order is determined by performing a toposort of a graph containing all members
/// and their dependencies.
pub fn member_compilation_order(
    manifest: &WorkspaceManifestFile,
    locked: bool,
    offline: bool,
) -> Result<Vec<MemberName>> {
    // ...
}
```

Now, w.r.t. to implementation, I think there are a couple of options:

### Option 1. Create a fake `PackageManifest` for the workspace

1. Create a fake `PackageManifest` for the workspace.
2. For each workspace `member`, add it as a `path` dependency to our fake package manifest.
3. Create a `BuildPlan` for the fake manifest.
4. Use the plan's compilation order to determine member compilation order (i.e. filter out all non-member packages and the fake root node from the end).

This is probably the simplest approach the best takes advantage of the existing logic. However, the caveat is that doing so would require we have a workspace-level `Forc.lock` file (like cargo) and would require we disallow project-specific lock files in the case they reside within a workspace.

This is probably for the best, as it would help to avoid duplicate versions of dependencies across the workspace, and to allow for easily updating all dependencies across a workspace to the same version at once in an atomic manner. This is likely the motivation behind why cargo projects use a single lock file at the workspace level.

The downside is that we would need to refactor `forc-pkg` a little to *always* check if a package resides within a workspace and, if-so, always create the `BuildPlan` from the workspace-wide lock file.

### Option 2. Manually construct workspace member graph

1. Create an empty graph.
2. Add a node representing the workspace root.
3. For each member:
    1. load its `BuildPlan`.
    2. Remove all non-member packages from its graph (i.e. all git dependencies, and all path deps that aren't members of the workspace).
    3. Merge the stripped graph into the main graph.
    4. Add an edge from the root to the member's node.
4. Do a `toposort` on the graph using `petgraph` to get the total order including the root.
5. Remove the workspace root node from the end of the sorted order to get the final result.

The implementation of this approach might be a little more tedious, but would require less of a refactor to the rest of `forc-pkg` and avoid the requirement of having a single workspace-level lock file.

This approach may also end up performing significantly slower, as we need to produce a `BuildPlan` (which can involve fetching dependencies) for *each member*, rather than only once for the whole workspace. As each member has its own lock file which can be updated separately, it's easier for two members to end up with different versions of the same dependency, which can lead to greater bytecode bloat.

Of these two options, my inclination is that the first option is the more correct approach, and that it's probably worth the extra refactor.

# 3. Update `build`, `check`, `update`, `test` and `deploy` commands

After coming up with these functions, we can then plug them into each of the commands, e.g.

- Check manifest type with `Manifest::from_path`:
    - If package, `<command>` package.
    - If workspace:
        - Determine member order with `member_dependency_order`.
        - `<command>` each package.

Of course, it might be easiest to test 1. and 2. by just implementing the `build` command first, but we should keep in mind that this same logic should be shareable with each of the above.

kayagokalp:
Closing in favor of #3127, #3130 and #3146

PR #3120, opened 2022-10-24
bump rust SDK to v0.26

nfurfaro:
I also attempted (but was unable) to bump other deps in this manifest:
- fuel-gql-client
- fuel-vm
- fuel-core
THis is because the SDK doesn't yet use the latest `fuel-gql-client` version.

### DIFF
--- a/test/src/sdk-harness/Cargo.toml
+++ b/test/src/sdk-harness/Cargo.toml
@@ -11,7 +11,7 @@ fuel-core = { version = "0.10", default-features = false }
 fuel-gql-client = { version = "0.10", default-features = false }
 fuel-types = "0.5"
 fuel-vm = "0.15"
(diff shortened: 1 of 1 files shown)


mohammadfawaz:
Seems reasonable. I had started doing that here: https://github.com/FuelLabs/sway/pull/3054 but as you said, we need a new `fuels` release. However, I did test #3054 with a `fuels` `master` and it compiles and passes the tests.

nfurfaro:
Ah, ok then.We could just wait and go with yours once we see the next fuels release @mohammadfawaz .

mohammadfawaz:
Closing this in favor of https://github.com/FuelLabs/sway/pull/3054 which is bumping `fuels` to `0.27` and `fuel-core` to `0.11.2`.

Issue #2550, opened 2022-08-16
Enhance purity checking to include specific checks for predicates


mohammadfawaz:
Codegen should have purity checks that are specific for predicates and ones that are specific for scripts. [Contract Instructions](https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/instruction_set.md#contract-instructions) are instructions that are not allowed in a predicate. Scripts allow only some of these instructions. The [checks we currently have](https://github.com/FuelLabs/sway/blob/master/sway-core/src/asm_generation/checks.rs#L32) should be replaced with two sets of checks:
- Predicate-specific checks
- Script-specific checks

mohammadfawaz:
Purity checks for predicates require that we know that a program is a predicate in IR and codegen. Today that's not the case as predicates are treated just like scripts. Their corresponding IR actually starts with `scripts;`. We should fix that first (See https://github.com/FuelLabs/sway/issues/1655).

PR #3125, opened 2022-10-25
Let IRgen compile predicate IR modules specifically.

### LINKED ISSUE #1655: Annotate predicate IR with `predicate` and run the appropriate `asm` checks
https://github.com/FuelLabs/sway/pull/1640 enables predicates to compile just like scripts which is perfectly fine from a functional point of view. However, if we want some predicate-specific `asm` checks to run in codegen, then we need to indicate, in the IR and beyond, that we're actually compiling a script not a predicate. This can be accomplished by making sure that the IR corresponding to predicates always start with `predicate`, just like we do for scripts and contracts. That is:
```rust
predicate {
    fn main() -> bool {
        ...
    }
    ...
}
```
Then, use this information to run whatever codegen checks we want to run (purity checks and whatnot)

otrho:
Previously we were just compiling scripts and predicates as the same and so predicates were compiled to `script` modules.

Closes #1655.

### DIFF
--- a/sway-core/src/asm_generation/from_ir.rs
+++ b/sway-core/src/asm_generation/from_ir.rs
@@ -116,7 +116,7 @@ fn compile_module_to_asm(
         Kind::Contract => ProgramKind::Contract,
         Kind::Script => ProgramKind::Script,
         Kind::Predicate => ProgramKind::Predicate,
-        Kind::Library => todo!("libraries and predicates coming soon!"),
+        Kind::Library => todo!("libraries coming soon!"),
     };
 
     ok(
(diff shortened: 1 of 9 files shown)


otrho:
#1655 talks about ASM checks but they're managed in #2550, which was basically blocked by this.

otrho:
Hrmm, shoulda probably run the tests myself instead of just :crossed_fingers: and :pray:.

sway-core/src/asm_generation/from_ir.rs
```diff
@@ -115,7 +115,8 @@ fn compile_module_to_asm(
     let kind = match module.get_kind(context) {
         Kind::Contract => ProgramKind::Contract,
         Kind::Script => ProgramKind::Script,
-        Kind::Library | Kind::Predicate => todo!("libraries and predicates coming soon!"),
+        Kind::Predicate => ProgramKind::Predicate,
+        Kind::Library => todo!("libraries and predicates coming soon!"),
```
mohammadfawaz, line 119:
```suggestion
        Kind::Library => todo!("libraries coming soon!"),
```

mohammadfawaz (approved):

mohammadfawaz:
> instead of just 🤞 and 🙏

I thought that's what CI is for 😆

PR #3126, opened 2022-10-25
Use release binaries to build and run the `sdk-harness` tests.

otrho:


otrho:
The `cargo-test-lib-std` tests usually take about 24-25 minutes.

otrho:
OK, building the ~50 contracts actually is still quicker for debug since the gains found using release is outweighed a fair bit by the time take to compile release Forc.

_And_ the SDK tests themselves have hardcoded paths to the binaries for `abigen` and also for deploying.  So... yeah, nah.

PR #3132, opened 2022-10-25
Remove `raw_ptr::addr()`

AlicanC:
With this change, `core` and `std` should stop providing a convenient way to access pointers as `u64`s.

While this might seem limiting, with `alloc()` returning `raw_ptr`s that have `add()` and `sub()`, all usages of `.addr()` could be refactored. (If not there's still `asm(r1: ptr) { r1: u64 }` of course.)

### DIFF
--- a/sway-lib-core/src/raw_ptr.sw
+++ b/sway-lib-core/src/raw_ptr.sw
@@ -7,11 +7,6 @@ impl raw_ptr {
         __eq(addr, 0)
     }
 
-    /// Gets the address of the pointer.
-    pub fn addr(self) -> u64 {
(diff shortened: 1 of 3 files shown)


mohammadfawaz:
This does seem reasonable to me, until we're proven otherwise.

mohammadfawaz (approved):
:shipit:

mohammadfawaz:
I disabled auto-merge momentarily until https://github.com/FuelLabs/sway/pull/3142 is merged and we get a new release. The reason being this PR is breaking and I'd like to delay jumping to `0.29.0` as much as possible.

otrho (approved):

PR #3134, opened 2022-10-25
Provide a way to obtain workspace member order

### LINKED ISSUE #3130: Provide a way to obtain member dependency order
# 2. Member compilation order

We should separate the logic for determining the dependency order of a workspace's `members` from the `build` process so that it can be shared between other commands like `check` or `test`, or 3rd party plugins. E.g. something like:

```rust
/// Returns all workspace members ordered from dependencies to dependents.
///
/// The order is determined by performing a toposort of a graph containing all members
/// and their dependencies.
pub fn member_compilation_order(
    manifest: &WorkspaceManifestFile,
    locked: bool,
    offline: bool,
) -> Result<Vec<MemberName>> {
    // ...
}
```

Now, w.r.t. to implementation, I think there are a couple of options:

### Option 1. Create a fake `PackageManifest` for the workspace

1. Create a fake `PackageManifest` for the workspace.
2. For each workspace `member`, add it as a `path` dependency to our fake package manifest.
3. Create a `BuildPlan` for the fake manifest.
4. Use the plan's compilation order to determine member compilation order (i.e. filter out all non-member packages and the fake root node from the end).

This is probably the simplest approach the best takes advantage of the existing logic. However, the caveat is that doing so would require we have a workspace-level `Forc.lock` file (like cargo) and would require we disallow project-specific lock files in the case they reside within a workspace.

This is probably for the best, as it would help to avoid duplicate versions of dependencies across the workspace, and to allow for easily updating all dependencies across a workspace to the same version at once in an atomic manner. This is likely the motivation behind why cargo projects use a single lock file at the workspace level.

The downside is that we would need to refactor `forc-pkg` a little to *always* check if a package resides within a workspace and, if-so, always create the `BuildPlan` from the workspace-wide lock file.

### Option 2. Manually construct workspace member graph

1. Create an empty graph.
2. Add a node representing the workspace root.
3. For each member:
    1. load its `BuildPlan`.
    2. Remove all non-member packages from its graph (i.e. all git dependencies, and all path deps that aren't members of the workspace).
    3. Merge the stripped graph into the main graph.
    4. Add an edge from the root to the member's node.
4. Do a `toposort` on the graph using `petgraph` to get the total order including the root.
5. Remove the workspace root node from the end of the sorted order to get the final result.

The implementation of this approach might be a little more tedious, but would require less of a refactor to the rest of `forc-pkg` and avoid the requirement of having a single workspace-level lock file.

This approach may also end up performing significantly slower, as we need to produce a `BuildPlan` (which can involve fetching dependencies) for *each member*, rather than only once for the whole workspace. As each member has its own lock file which can be updated separately, it's easier for two members to end up with different versions of the same dependency, which can lead to greater bytecode bloat.

Of these two options, my inclination is that the first option is the more correct approach, and that it's probably worth the extra refactor.

posted by @mitchmindtree [here](https://github.com/FuelLabs/sway/pull/3117#pullrequestreview-1154014118)

kayagokalp:
closes #3130.

After this PR, I will be able to use the order generated from the workspace members for build/check/deploy operations. ~Diff is going to be kind of dirty right now as I created this branch from #3129 but I will rebase this onto master after #3129 is merged~

Rebased and ready for review! This PR will allow #3146 to proceed which will introduce workspace level building. After that I will update the logic for deploy and check


mitchmindtree:
> Diff is going to be kind of dirty right now as I created this branch from https://github.com/FuelLabs/sway/pull/3129

One thing you can do to improve review-ability is point your PR to the same branch you forked from. That way the PR will only show the changes since the tip of that branch :+1:

forc-pkg/src/pkg.rs
```diff
@@ -561,11 +589,21 @@ fn validate_dep_manifest(
...
     let dep_program_type = dep_manifest.program_type()?;
     // Check if the dependency is either a library or a contract declared as a contract dependency
-    match (&dep_program_type, &dep_edge.kind) {
-        (TreeType::Contract, DepKind::Contract) | (TreeType::Library { .. }, DepKind::Library) => {}
+    //
+    //
+    // If this is a workspace build, we are allowing deps to be in any type. Because we are
+    // creating a dummy PackageManifestFile that declares each member of the workspace as a dependency.
+    // This condition will only happen while we are creating the workspace-wide dependency graph
+    // and it won't be possible for indivual members of the workspace to declare invalid typed dependencies.
```
mitchmindtree, line 601:
```suggestion
    // and it won't be possible for individual members of the workspace to declare invalid typed dependencies.
```

forc-pkg/src/pkg.rs
```diff
@@ -561,11 +589,21 @@ fn validate_dep_manifest(
     dep: &Pinned,
     dep_manifest: &PackageManifestFile,
     dep_edge: &Edge,
+    workspace_build: bool,
```
mitchmindtree, line 592: Perhaps rather than piping this `bool` all the way through validation, we can create a new `DepKind` variant that represents the workspace member connection, and use these for the edges from the workspace node to the members? E.g. `DepKind::WorkspaceMember`?

forc-pkg/src/pkg.rs
```diff
@@ -2262,11 +2313,68 @@ pub fn build_with_options(build_options: BuildOptions) -> Result<Built> {
...
+/// Returns the order of workspace members for build/deploy/check.
+pub fn workspace_member_order(
+    manifest: &WorkspaceManifestFile,
+    locked: bool,
+    offline: bool,
+) -> Result<Vec<String>> {
+    let mut member_order = vec![];
+    let workspace_members: HashSet<String> = manifest.members().cloned().collect();
+    let tmp_pkg_manifest = tmp_manifest_with_members(manifest)?;
+    let tmp_pkg_manifest_file = PackageManifestFile::new(tmp_pkg_manifest, manifest.path());
```
mitchmindtree, line 2358: Does using the original workspace `manifest.path()` actually work here? If so maybe this is OK, but I'm a little worried having the manifest refer to a totally different manifest might come back to bite us.

I wonder if we should instead create a temporary file for the virtual package manifest within a tmp dir, and determine the member path dependencies relative to that? What are your thoughts?

forc-pkg/src/pkg.rs
```diff
@@ -2262,11 +2313,68 @@ pub fn build_with_options(build_options: BuildOptions) -> Result<Built> {
...
+        offline,
+        workspace_build,
+    )?;
+    let graph = build_plan.graph();
+    for member in build_plan.compilation_order() {
+        let name = &graph[*member].name;
+        if workspace_members.contains(name) {
+            member_order.push(name.clone());
+        }
     }
```
mitchmindtree, line 2373: This loop could also be written as something like the following, avoiding the need for a `mut` `member_order` binding:

```rust
let member_order = build_plan
    .compilation_order()
    .map(|&n_ix| &graph[n_ix].name)
    .filter(|&name| workspace_members.contains(name))
    .cloned()
    .collect(); 
```

mitchmindtree (changes requested):
Nice one @kayagokalp, I haven't tested this locally yet but the changes are looking like the right direction! Just a few suggestions.

The only thing I'm mildly concerned about is if we're overloading the meaning of `PackageManifest` a bit with this new virtual workspace manifest thing, but I think for now it will do the trick and we can revisit this later if we start to encounter issues.

forc-pkg/src/pkg.rs
kayagokalp: Hmm, it works as we don't actually re-parse the file after that point and I can build in the next PR which is built on top of this branch. If you think this approach is Ok for now I can open an issue and we can get this improved after we have everything for workspace building

forc-pkg/src/pkg.rs
kayagokalp: Ohh yeah, it is much much better thanks 👌

forc-pkg/src/manifest.rs
```diff
@@ -591,17 +599,35 @@ impl WorkspaceManifestFile {
...
     }
+
+    /// The path to the `Forc.toml` from which this manifest was loaded.
+    ///
+    /// This will always be a canonical path.
+    pub fn path(&self) -> &Path {
+        &self.path
+    }
+
+    /// The path to the directory containing the `Forc.toml` from which this manfiest was loaded.
```
bingcicle, line 623:
```suggestion
    /// The path to the directory containing the `Forc.toml` from which this manifest was loaded.
```

forc-pkg/src/pkg.rs
```diff
@@ -2262,11 +2282,61 @@ pub fn build_with_options(build_options: BuildOptions) -> Result<Built> {
...
-/// Returns the ContractId of a built_package contract with specified `salt`.
+/// Returns a tmp PackageManifest with the members of the WorkspaceManifest inserted as
+/// dependencies
+fn tmp_manifest_with_members(
+    workspace_manifest: &WorkspaceManifestFile,
+) -> Result<PackageManifest> {
+    let mut package_manifest = PackageManifest::default();
+    let mut dependencies = BTreeMap::new();
+    let project = Project {
+        name: "worksapce".to_string(),
```
bingcicle, line 2299:
```suggestion
        name: "workspace".to_string(),
```

forc-pkg/src/pkg.rs
```diff
@@ -1824,6 +1840,10 @@ pub fn dependency_namespace(
                 constants.insert(contract_dep_constant_name.to_string(), contract_id_constant);
                 namespace::Module::default_with_constants(constants)?
             }
+            _ => lib_namespace_map
```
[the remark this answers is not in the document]
kayagokalp: Hmm actually we shouldn't have that case happening anyways so I added an explicit 3rd match statement with `panic`

mitchmindtree (changes requested):
Thanks @kayagokalp! The edge changes do seem a bit better, though the deeper we go down this path of using a fake package manifest, the more I worry it might end up causing too much ambiguity and hide too many potential issues down the road. E.g. the `PackageManifestFile` holding an incorrect manifest path, and the edges within the graph that aren't assigned a correct `DepKind` until `validate_deps` begins.

Another issue that I'm just noticing is that the generated lock file likely contains an entry for the workspace node as if it were a `[[package]]` which is not what we want either.

I realize these issues are largely my fault for sending you down this path! But perhaps it's worth considering another approach?

Maybe rather than creating a fake package manifest, we can take a step lower and try to construct a `BuildPlan` for the workspace directly. How about something like this:

1. Change `BuildPlan::from_manifest` and `BuildPlan::from_lock_and_manifest` to take a `&ManifestFile` (the enum) rather than a `&PackageManifestFile`.

   In the case that we're building from a `WorkspaceManifest`, we add a root node for each member, rather than only a single root node as we currently do for `PackageManifest`s.

   This will likely affect the `validate_graph`, `remove_deps`, `graph_to_manifest_map` and `fetch_graph` functions. Currently, each expect only a single root node, however it shouldn't be too tricky to refactor these to handle multiple root nodes (i.e. one for each member in the case we have a workspace manifest, or one in the case that we have a package manifest).

   This way we don't need to include the workspace as a root node itself (or have a fake pkg manifest or incorrect path), and in turn, we don't need to add a new `WorkspaceMember` `DepKind` variant.

2. Remove the `workspace_member_order` function in favour of a `BuildPlan::root_pkgs` method which produces a node index for each of the root nodes. This would similarly yield all the workspace member packages in order, or only a single node in the case that the BuildPlan was produced from a `PackageManifest`.

While this might involve a little more churn, I think the result would be a lot clearer and easier to maintain in the long run. I'd prefer we get this right now rather than rush it in and have to refactor it later. What are your thoughts?

Side-note: Once we look like we're arriving at the right approach, we should make sure we have a test to check we're actually getting the correct order :)

forc-pkg/src/pkg.rs
```diff
@@ -478,6 +493,65 @@ impl BuildPlan {
...
+                .node_indices()
+                .filter(|node| !matches!(graph[*node].source, SourcePinned::Root))
+                .find(|node| graph[*node].name == root_version.name);
+            if let Some(pinned_version) = pinned_version {
+                // Add this Root node as the corresponding node for `pinned_version`
+                pinned_to_root_map.insert(pinned_version, node);
+                // remove this root node from the ordering graph.
+                ordering_graph.remove_node(node);
+            }
+        }
```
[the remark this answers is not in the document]
mitchmindtree: After looking a little closer, I see that it's not enough to check something like `is_parentless`, as workspace members that are depended on by another workspace will have a parent edge :thinking: 

Ideally, the graph should *never* have two packages for a single package. That is, member nodes that are depended on by some other node should never have a second equivalent package node in the graph - we should only have one.

I think there are a couple things we can do:

1. Update `fetch_deps` so that, when determining a dependency package's `Source`, before we assume it is a `Source::Path` we first check if the path is equal to a workspace `member` path. If so, then we add a `Source::Root` node with the desired `Path` instead.
2. On that note, we should probably change the `Source` and `SourcePinned` `Root` variant names to `Member`, as a `Root` implies the package has no incoming edges which is not always the case for member packages.

With these changes, we should be able to change this method to something like this:

```rust
    /// Produce an iterator yielding all workspace member nodes in order of compilation.
    ///
    /// In the case that this `BuildPlan` was constructed for a single package,
    /// only that package's node will be yielded.
    pub fn member_nodes(&self) -> '_ + impl Iterator<Item = NodeIx> {
        self.compilation_order()
            .iter()
            .cloned()
            .filter(|&n| self.graph[n].source == SourcePinned::Member)
    }
```

mitchmindtree (changes requested):
As mentioned on slack, really awesome stuff @kayagokalp - feeling much better about this direction! I've left a few more suggestions for the new design.

forc-pkg/src/pkg.rs
```diff
@@ -824,7 +771,23 @@ fn dep_path(
...
+                parent_workspace_manifest
+                    .members()
+                    .zip(parent_workspace_manifest.member_paths()?)
+                    .find(|(member_name, _)| *member_name == dep_name)
+                    .map(|(_, member_path)| member_path)
+                    .ok_or_else(|| anyhow!("cannot find dependency in the workspace"))
+            } else {
+                bail!("a node cannot depend on a root node if it is not a member of a workspace")
+            }
+        }
```
mitchmindtree, line 790: Nice work! That said, it's a shame we have to load the workspace manifest file via I/O again here. I'll leave some thoughts on manifest file reading in the top comment.

forc-pkg/src/pkg.rs
```diff
@@ -1013,7 +967,7 @@ impl fmt::Display for GitReference {
 impl fmt::Display for SourcePinned {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            SourcePinned::Root => write!(f, "root"),
+            SourcePinned::Member => write!(f, "root"),
```
mitchmindtree, line 970:
```suggestion
            SourcePinned::Member => write!(f, "member"),
```

forc-pkg/src/pkg.rs
```diff
@@ -1113,7 +1067,7 @@ impl FromStr for SourcePinned {
     type Err = SourcePinnedParseError;
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let source = if s == "root" {
```
mitchmindtree, line 1069:
```suggestion
        // Also check `"root"` to support reading the legacy `Forc.lock` format and to
        // avoid breaking old projects.
        let source = if s == "member" || s == "root" {
```

forc-pkg/src/pkg.rs
```diff
@@ -1933,7 +1887,25 @@ fn dep_to_source(pkg_path: &Path, dep: &Dependency) -> Result<Source> {
...
+                        workspace_manifest.member_paths()?.collect();
+                    if member_paths.contains(&canonical_path) {
+                        // This is a workspace member so the source should be inserted as `Member`
+                        Source::Member(canonical_path)
+                    } else {
+                        Source::Path(canonical_path)
+                    }
+                } else {
+                    Source::Path(canonical_path)
+                }
```
mitchmindtree, line 1908: Maybe it's worth abstracting this logic into something like:
```rust
fn is_member_path(members: &MemberManifestFiles, path: &Path) -> bool
```

mitchmindtree (changes requested):
Great work @kayagokalp, I think we've hit the right approach here :) Almost there!

### Manifest File Reading

It's a bit unfortunate that we need to load the workspace manifests so many times. That is, each time we iterate over the member package manifests, we end up doing the I/O to read them all from file again. Besides being costly, it also means we run the risk of of having a package manifest file change between iterations, meaning our `BuildPlan` can end up in an inconsistent state.

Perhaps we might better off loading them all before constructing the `BuildPlan`? E.g. I'm thinking we can create a type like:

```rust
/// The name of a workspace member package.
pub type MemberName = String;
/// A manifest for each workspace member, or just one manifest if working with a single package.
pub type MemberManifestFiles = BTreeMap<MemberName, PackageManifestFile>;
```

Then we can have a method on `ManifestFile` along the lines of:

```rust
pub fn member_manifests(&self) -> MemberManifestFiles { ... }
```

Then perhaps our `BuildPlan` constructors can take a `&MemberManifestFiles` rather than a `&ManifestFile`? This way we shouldn't have to match on the `ManifestFile` multiple times while creating the `BuildPlan`, and we don't have to incur any more I/O from reading files.

kayagokalp:
Hey @mitchmindtree, I will have a second look (probably before you will see this message 😄) but I think I added the level of abstraction required to not depend on I/O for all member manifest fetching operation. Also addressed other suggestions. One thing to note: I included the changed lock files (root -> member) change. Not sure why the line that we added for legacy support didn't prevent this lock file change 🤔

forc-pkg/src/manifest.rs
```diff
@@ -61,6 +69,38 @@ impl ManifestFile {
             .parent()
             .expect("failed to retrieve manifest directory")
     }
```
mitchmindtree, line 71: Let's also add a `pub fn lock_path(&self) -> Result<PathBuf>` method here so that it can be easily retrieved for passing to the `BuildPlan::from_lock_and_manifests` constructor.

This is a little tricky as, even if we have a `ManifestFile::Package`, we still want to check if our package is *within* a workspace. This is because if our package is within a workspace, we always want to use the workspace-level lock file to ensure a consistent set of pinned dependencies.

I'm thinking we could do something like the following:

1. Add a `pub fn workspace(&self) -> Result<Option<WorkspaceManifestFile>>` method to `PackageManifestFile` that returns a `WorkspaceManifestFile` in the case that the package is within a workspace, otherwise `None`.
2. Add a `pub fn lock_path(&self) -> PathBuf` method to the `WorkspaceManifestFile`.
3. Add a `pub fn lock_path(&self) -> Result<PathBuf>` method to `PackageManifestFile` that first checks for the workspace, falling back to a path for the package lock if no workspace.

Then our `ManifestFile::lock_path` function could look something like:

```rust
pub fn lock_path(&self) -> Result<PathBuf> {
    match self {
        Self::Package(ref pkg) => pkg.lock_path(),
        Self::Workspace(ref ws) => Ok(ws.lock_path()),
    }
}

File: forc-pkg/src/manifest.rs [hunk 14]
```diff
@@ -61,6 +69,38 @@ impl ManifestFile {
...
+
+    /// Returns manifest file map from member name to the corresponding package manifest file
+    pub fn member_manifests(&self) -> Result<MemberManifestFiles> {
+        let mut member_manifest_files = BTreeMap::new();
+        match self {
+            ManifestFile::Package(pkg_manifest_file) => {
+                let member_name = &pkg_manifest_file.project.name;
+                member_manifest_files.insert(member_name.clone(), *pkg_manifest_file.clone());
```
mitchmindtree, line 79: Thinking on this more, even if we have a `ManifestFile::Package`, if we're within a workspace we still want to produce *all* members. We should use our new `PackageManifestFile::workspace` method (mentioned in my previous comment) to check if the package is within a workspace first, and if so, produce the manifest for all members (as long as our package is one of the members). This is to ensure the lock file (and in turn build plan) remains consistent for the whole workspace.

In the follow-up PR when it comes time to `build`/`check`/`test`/`deploy` a single package, we can skip the nodes in the `compilation_order` that aren't a part of the single package's dependencies to avoid building the entire workspace every time.

mitchmindtree (changes requested):

forc-pkg/src/pkg.rs
```diff
@@ -399,17 +400,33 @@ impl BuildPlan {
     // the manifest alongside some lock diff type that can be used to optionally write the updated
     // lock file and print the diff.
     pub fn from_lock_and_manifest(
-        manifest: &PackageManifestFile,
+        manifests: &ManifestFiles,
         locked: bool,
         offline: bool,
     ) -> Result<Self> {
```
mitchmindtree, line 406: Ahh, I meant to add a comment here as a part of the last review but looks like it didn't make it:

It looks like the `WorkspaceManifestFile` is solely used to determine the lock file path. Instead, I think it might be better if we can pass the lock path as an argument to this constructor, e.g.

```rust
pub fn from_lock_and_manifest(
    lock_path: &Path,
    manifests: &MemberManifestFiles,
    locked: bool,
    offline: bool,
) -> Result<Self> {
```

This way the `BuildPlan` construction can remain agnostic of whether or not it's for a workspace.

forc-pkg/src/manifest.rs
```diff
@@ -322,6 +326,46 @@ impl PackageManifestFile {
...
+
+    /// Returns the workspace manifest file if this `PackageManifestFile` is one of the members.
+    pub fn workspace(&self) -> Result<Option<WorkspaceManifestFile>> {
+        let parent_dir = self
+            .dir()
+            .parent()
+            .ok_or_else(|| anyhow!("Cannot get parent directory"))?;
+        // Only check for workspace manifest file parsing errors if the parent dir contains a
+        // manifest file.
+        if parent_dir.join("Forc.toml").exists() {
```
mitchmindtree, line 338: Do we have a `const` for `"Forc.toml"`?

forc-pkg/src/manifest.rs
```diff
@@ -322,6 +326,46 @@ impl PackageManifestFile {
...
+    /// returns the workspace level lock files location.
+    ///
+    /// This will always be a canonical path.
+    pub fn lock_path(&self) -> Result<PathBuf> {
+        // Check if this package is in a workspace
+        let workspace_manifest = self.workspace()?;
+        if let Some(workspace_manifest) = workspace_manifest {
+            Ok(workspace_manifest.lock_path())
+        } else {
+            Ok(self.dir().to_path_buf().join("Forc.lock"))
```
mitchmindtree, line 366: Do we have a `const` for `"Forc.lock"`?

forc-pkg/src/manifest.rs
```diff
@@ -688,9 +732,17 @@ impl WorkspaceManifestFile {
...
+    /// Check if given path corresponds to any workspace member's path
     pub fn is_member_path(&self, path: &Path) -> Result<bool> {
         Ok(self.member_paths()?.any(|member_path| member_path == path))
     }
+
+    /// Returns the location of the lock file for `WorkspaceManifestFile`.
+    ///
+    /// This will always be a canonical path.
+    pub fn lock_path(&self) -> PathBuf {
+        self.dir().to_path_buf().join("Forc.lock")
```
mitchmindtree, line 744: Same here

mitchmindtree (changes requested):
I'm just realising that our `WorkspaceManifest` should be storing `members` as relative `PathBuf`s, and that we shouldn't be treating those as names. E.g. if we take a look at the Sway repo's members:

```toml
members = [
    "forc",
    "forc-pkg",
    "forc-plugins/forc-client",
    "forc-plugins/forc-explore",
    "forc-plugins/forc-fmt",
```

we can see that these actually represent *relative paths* to the members.

Once we change the `WorkspaceManifest`'s `members` field to store those relative `PathBuf`s, we can change the `WorkspaceManifestFile`'s `member_paths` method to convert these to *canonical* paths using its own directory path.

forc-pkg/src/manifest.rs
```diff
@@ -322,6 +326,46 @@ impl PackageManifestFile {
...
+                .any(|member| *member == self.project.name)
+            {
+                Ok(Some(workspace_manifest))
+            } else {
+                Ok(None)
+            }
+        } else {
+            Ok(None)
+        }
+    }
```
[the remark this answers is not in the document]
kayagokalp: I added a simple check, like we do for the lock file cause detection to overcome this currently and will open up an issue to tackle after we have workspace build/deploy operating 👍

forc/src/ops/forc_update.rs
```diff
@@ -37,7 +37,8 @@ pub async fn update(command: UpdateCommand) -> Result<()> {
     let lock_path = lock_path(manifest.dir());
     let old_lock = Lock::from_path(&lock_path).ok().unwrap_or_default();
     let offline = false;
-    let new_plan = pkg::BuildPlan::from_manifest(&manifest.manifests()?, offline)?;
+    let member_manifests = manifest.member_manifests()?;
+    let new_plan = pkg::BuildPlan::from_manifest(&member_manifests, offline)?;
     let new_lock = Lock::from_graph(new_plan.graph());
     let diff = new_lock.diff(&old_lock);
     let name = if let ManifestFile::Package(pkg_manifest) = &manifest {
```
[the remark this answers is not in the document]
kayagokalp: Yep that makes sense, I now collect a set of names for the member packages and omit them while printing:

before this change adding std dependency was printing:

```
  Removing test_project
    Adding core
    Adding std git+https://github.com/fuellabs/sway?tag=v0.29.0#dc8e5f6b84741a3a690ba25b473f8b1e01e31103
    Adding test_project
   Created new lock file at /Users/kayagokalp/fuel/test_projects/virtual/Forc.lock
```

Now it prints:

```
    Adding core
    Adding std git+https://github.com/fuellabs/sway?tag=v0.29.0#dc8e5f6b84741a3a690ba25b473f8b1e01e31103
   Created new lock file at /Users/kayagokalp/fuel/test_projects/virtual/Forc.lock
```

kayagokalp:
I think over all the review process there seems to be 2 new issue to tackle after this PR. Once this is final and approved I will go over all the review comments make sure that we have issues for all of them

forc-pkg/src/pkg.rs
```diff
@@ -654,7 +697,8 @@ fn validate_dep_manifest(
             dep_manifest.project.name,
         );
     }
-    validate_version(dep_manifest)?;
+    let dep_manifests = BTreeMap::from([(dep_manifest.project.name.clone(), dep_manifest.clone())]);
+    validate_version(&dep_manifests)?;
```
mitchmindtree, line 701: Looks like we could use `validate_pkg_version` here:
```rust
    validate_pkg_version(dep_manifest)?;
```

forc-pkg/src/pkg.rs
```diff
@@ -1157,6 +1230,29 @@ pub fn fetch_id(path: &Path, timestamp: std::time::Instant) -> u64 {
...
+    for member_pkg_manifest in member_manifests.values() {
+        added_nodes.extend(&fetch_pkg_graph(
+            member_pkg_manifest,
+            offline,
+            graph,
+            manifest_map,
+        )?);
+    }
+    Ok(added_nodes)
+}
```
mitchmindtree, line 1254: Similarly to my comment [here](https://github.com/FuelLabs/sway/pull/3134/files#r1007934240), I think we can likely improve the efficiency quite a bit by re-using `visited` set within `fetch_pkg_graph` to avoid a lot of repeated work, but this seems to work nicely atm so let's leave it for a future PR :+1:

forc-pkg/src/pkg.rs
```diff
@@ -1758,7 +1854,24 @@ fn dep_to_source(pkg_path: &Path, dep: &Dependency) -> Result<Source> {
...
+                    // Check if that manifest declares this path as a member_path
+                    if workspace_manifest.is_member_path(&canonical_path)? {
+                        // This is a workspace member so the source should be inserted as `Member`
+                        Source::Member(canonical_path)
+                    } else {
+                        Source::Path(canonical_path)
+                    }
+                } else {
+                    Source::Path(canonical_path)
+                }
```
mitchmindtree, line 1874: This could also be written as:
```rust
match workspace_manifest {
    Some(ws) if ws.is_member_path(&canonical_path) => Source::Member(canonical_path),
    _ => Source::Path(canonical_path),
}
```

forc-pkg/src/pkg.rs
```diff
@@ -1758,7 +1854,24 @@ fn dep_to_source(pkg_path: &Path, dep: &Dependency) -> Result<Source> {
                 let canonical_path = path.canonicalize().map_err(|e| {
                     anyhow!("Failed to canonicalize dependency path {:?}: {}", path, e)
                 })?;
-                Source::Path(canonical_path)
+                // Check if path is a member of a workspace.
+                let workspace_manifest = canonical_path
+                    .parent()
+                    .and_then(|parent_dir| WorkspaceManifestFile::from_dir(parent_dir).ok());
```
mitchmindtree, line 1860: Rather than doing all the I/O work of finding and reading this workspace again here (which will be attempted every time we visit a `Path` dependency) we should instead be checking our already loaded `MemberManifestFiles` set.

We can land as is for now, but we should open a high priority issue and make sure to address this soon after.

forc/src/ops/forc_update.rs
```diff
@@ -32,14 +33,23 @@ pub async fn update(command: UpdateCommand) -> Result<()> {
...
     let diff = new_lock.diff(&old_lock);
-    lock::print_diff(&manifest.project.name, &diff);
+    let member_names = if let ManifestFile::Package(pkg_manifest) = &manifest {
+        HashSet::from([pkg_manifest.project.name.clone()])
+    } else {
+        member_manifests
+            .iter()
+            .map(|(_, manifest)| manifest.project.name.clone())
+            .collect()
+    };
```
mitchmindtree, line 51: We don't really need to check if `manifest` is `ManifestFile::Package` here (it will already exist in `member_manifests`). We can change this to:
```rust
let member_names = member_manifests
    .iter()
    .map(|(_, manifest)| manifest.project.name.clone())
    .collect();
```

mitchmindtree (dismissed):
Huge effort @kayagokalp - super stoked with how this has turned out :)

I've left a few more suggestions, but none are critical to landing this - happy to see this land and have these addressed as a part of follow-up work.

:rocket: :bow: :heavy_check_mark:

forc-pkg/src/pkg.rs
```diff
@@ -1758,7 +1854,24 @@ fn dep_to_source(pkg_path: &Path, dep: &Dependency) -> Result<Source> {
...
+
+                if let Some(workspace_manifest) = workspace_manifest {
+                    // There is a workspace manifest in the parent directory.
+                    //
+                    // Check if that manifest declares this path as a member_path
+                    if workspace_manifest.is_member_path(&canonical_path)? {
+                        // This is a workspace member so the source should be inserted as `Member`
+                        Source::Member(canonical_path)
+                    } else {
+                        Source::Path(canonical_path)
```
bingcicle, line 1871: Might be a bit of a stickler comment on this but I think the code explains itself here and perhaps the comments are unnecessary but it's up to you!

bingcicle (changes requested):

kayagokalp:
Solved conflicts and addressed @bingcicle crs. I will address @mitchmindtree's last suggestions in a following PR so that I can unblock next 3 PRs for now.

forc-pkg/src/pkg.rs
```diff
@@ -1056,11 +1179,35 @@ pub fn compilation_order(graph: &Graph) -> Result<Vec<NodeIx>> {
...
+        }
+        ManifestFile::Workspace(workspace_manifest_file) => {
+            let mut manifest_map = ManifestMap::new();
+            for member_pkg_manifest in workspace_manifest_file.member_pkg_manifests()? {
+                manifest_map.extend(pkg_graph_to_manifest_map(&member_pkg_manifest, graph)?);
+            }
+            Ok(manifest_map)
+        }
+    }
+}
```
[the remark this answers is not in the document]
kayagokalp: created #3281 to track this

forc-pkg/src/pkg.rs
kayagokalp: created #3282 to track this

forc-pkg/src/manifest.rs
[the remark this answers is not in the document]
kayagokalp: created #3283 to track addition of `thiserror`

kayagokalp:
### Follow Ups:

- [ ] #3281 
- [ ] #3282
- [ ] #3283

bingcicle (approved):

mitchmindtree (approved):

PR #3139, opened 2022-10-25
Use `raw_ptr` methods in `message.sw`

AlicanC:
Following up on this: https://github.com/FuelLabs/sway/pull/2828#issuecomment-1289794265

> However, the changes you had to send_message seemed a bit buggy. send_message is only tested in token_ops currently so no surprise that Ci was passing. Regardless, I reverted the implementation of send_message to how it was, for the most part. We can iterate on that in the future.

@mohammadfawaz, it seems like `raw_ptr::read()` didn't behave like I assumed it would. Still made use of `::add()` and `::write()` to pack `alloc()`s together and get rid of one `__addr_of()`, but the other one stayed.


mohammadfawaz (approved):
Awesome, much better. Thanks!

JoshuaBatty (approved):

PR #3140, opened 2022-10-25
Enable printing contract-id and accept Bech32 address in `forc-deploy`

### LINKED ISSUE #3137: forc-deploy does not print the contract-id
We cannot retrieve the contract-id as `forc-deploy` does not print it anymore.

kayagokalp:
closes #3137.

I re-added the contract id prompt and also introduced a small fix where input from the user is expected to be in `Bech32` format so that `forc-wallet list` output can be directly used. This is tested with forc-client v0.10.1, and it can sign a deployment transaction with that version. But it seems like there are some changes in the newer versions that breaks the signing process, I guess to address that we will have to bump versions to the latest. I think we might be short of some releases for that but I will have a deeper look into that after this one.


AlicanC (approved):

AlicanC:
Thanks! :shipit:

bingcicle (approved):

mohammadfawaz:
Excellent. We need a quick follow up release because this is critical. I'll do that today.

PR #3146, opened 2022-10-25
[WIP] Implement workspace building

### LINKED ISSUE #3135: Update building logic to take workspaces into account
As a next step in the full support of workspaces, we need to update building logic. This can include disallowing lock files in individual members of the workspace and use the one in the root folder. 

related and part of #1777

kayagokalp:
closes #3135.

Opening up for visibility.

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -2248,7 +2248,7 @@ pub fn build_with_options(build_options: BuildOptions) -> Result<()> {
         std::env::current_dir()?
     };
 
-    let manifest_file = ManifestFile::from_dir(&this_dir)?;
+    let manifest_file = ManifestFile::from_path(&this_dir)?;
     match manifest_file {
         ManifestFile::Package(package_manifest) => {
             build_package_with_options(&package_manifest, build_options)?
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -1,7 +1,8 @@
 use crate::{
     lock::Lock,
     manifest::{
-        BuildProfile, ConfigTimeConstant, Dependency, PackageManifest, PackageManifestFile,
+        BuildProfile, ConfigTimeConstant, Dependency, ManifestFile, PackageManifest,
+        PackageManifestFile,
     },
     CORE, PRELUDE, STD,
 };
@@ -2091,13 +2092,15 @@ pub const SWAY_BIN_HASH_SUFFIX: &str = "-bin-hash";
 /// when predicates are built.
 pub const SWAY_BIN_ROOT_SUFFIX: &str = "-bin-root";
 
-/// Builds a project with given BuildOptions
-pub fn build_with_options(build_options: BuildOptions) -> Result<Compiled> {
+pub fn build_package_with_options(
+    manifest: &PackageManifestFile,
+    build_options: BuildOptions,
+) -> Result<Compiled> {
     let key_debug: String = "debug".to_string();
     let key_release: String = "release".to_string();
 
     let BuildOptions {
-        path,
+        path: _,
         binary_outfile,
         debug_outfile,
         print_ast,
@@ -2134,17 +2137,6 @@ pub fn build_with_options(build_options: BuildOptions) -> Result<Compiled> {
             }
         }
     }
-
-    let this_dir = if let Some(ref path) = path {
-        PathBuf::from(path)
-    } else {
-        std::env::current_dir()?
-    };
-
-    let manifest = PackageManifestFile::from_dir(&this_dir)?;
-
-    let plan = BuildPlan::from_lock_and_manifest(&manifest, locked, offline_mode)?;
-
     // Retrieve the specified build profile
     let mut profile = manifest
         .build_profile(&selected_build_profile)
@@ -2163,6 +2155,8 @@ pub fn build_with_options(build_options: BuildOptions) -> Result<Compiled> {
     profile.print_intermediate_asm |= print_intermediate_asm;
     profile.terse |= terse_mode;
     profile.time_phases |= time_phases;
+
+    let plan = BuildPlan::from_lock_and_manifest(manifest, locked, offline_mode)?;
 
     // Build it!
     let (compiled, source_map) = build(&plan, &profile)?;
@@ -2242,6 +2236,27 @@ pub fn build_with_options(build_options: BuildOptions) -> Result<Compiled> {
     }
 
     Ok(compiled)
+}
+
+/// Builds a project with given BuildOptions
+pub fn build_with_options(build_options: BuildOptions) -> Result<()> {
+    let path = &build_options.path;
+
+    let this_dir = if let Some(ref path) = path {
+        PathBuf::from(path)
+    } else {
+        std::env::current_dir()?
+    };
+
+    let manifest_file = ManifestFile::from_dir(&this_dir)?;
+    match manifest_file {
+        ManifestFile::Package(package_manifest) => {
+            build_package_with_options(&package_manifest, build_options)?
+        }
+        ManifestFile::Workspace(_) => bail!("Workspace building is not supported"),
+    };
+
+    Ok(())
 }
 
 /// Returns the ContractId of a compiled contract with specified `salt`.
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -2045,7 +2045,7 @@ pub fn compile(
         Some(BytecodeOrLib::Bytecode(_)) | None => fail(&bc_res.warnings, &bc_res.errors),
     }
 }
-#[derive(Default)]
+#[derive(Default, Clone)]
 pub struct BuildOptions {
     /// Path to the project, if not specified, current working directory will be used.
     pub path: Option<String>,
@@ -2105,7 +2105,7 @@ pub const SWAY_BIN_ROOT_SUFFIX: &str = "-bin-root";
 
 pub fn build_package_with_options(
     manifest: &PackageManifestFile,
-    build_options: BuildOptions,
+    build_options: &BuildOptions,
 ) -> Result<Compiled> {
     let key_debug: String = "debug".to_string();
     let key_release: String = "release".to_string();
@@ -2127,7 +2127,7 @@ pub fn build_package_with_options(
         build_profile,
         release,
         time_phases,
-    } = build_options;
+    } = build_options.clone();
 
     let mut selected_build_profile = key_debug;
     match &build_profile {
@@ -2264,10 +2264,15 @@ pub fn build_with_options(build_options: BuildOptions) -> Result<()> {
     let manifest_file = ManifestFile::from_path(&this_dir)?;
     match manifest_file {
         ManifestFile::Package(package_manifest) => {
-            build_package_with_options(&package_manifest, build_options)?;
+            build_package_with_options(&package_manifest, &build_options)?;
         }
-        ManifestFile::Workspace(_) => {
-            bail!("Building a workspace is not supported.")
+        ManifestFile::Workspace(workspace_manifest) => {
+            let member_order = workspace_member_order(&workspace_manifest, build_options.locked, build_options.offline_mode)?;
+            for member in member_order {
+                let member_path = this_dir.join(member);
+                let member_pkg_manifest = PackageManifestFile::from_dir(&member_path)?;
+                build_package_with_options(&member_pkg_manifest, &build_options)?;
+            } 
         }
     };
 
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -2,7 +2,7 @@ use crate::{
     lock::Lock,
     manifest::{
         BuildProfile, ConfigTimeConstant, Dependency, DependencyDetails, ManifestFile,
-        PackageManifest, PackageManifestFile,
+        PackageManifest, PackageManifestFile, Project,
     },
     WorkspaceManifestFile, CORE, PRELUDE, STD,
 };
@@ -309,24 +309,6 @@ impl BuildPlan {
         })
     }
 
-    /// Create a new build plan for the project by fetching and pinning all dependencies.
-    ///
-    /// This is used for tmp manifest files created before handling workspaces.
-    fn from_manifest_without_validation(
-        manifest: &PackageManifestFile,
-        offline: bool,
-    ) -> Result<Self> {
-        let mut graph = Graph::default();
-        let mut manifest_map = ManifestMap::default();
-        fetch_graph(manifest, offline, false, &mut graph, &mut manifest_map)?;
-        let compilation_order = compilation_order(&graph)?;
-        Ok(Self {
-            graph,
-            manifest_map,
-            compilation_order,
-        })
-    }
-
     /// Create a new build plan taking into account the state of both the PackageManifest and the existing
     /// lock file if there is one.
     ///
@@ -347,9 +329,12 @@ impl BuildPlan {
         manifest: &PackageManifestFile,
         locked: bool,
         offline: bool,
+        validate: bool,
     ) -> Result<Self> {
-        // Check toolchain version
-        validate_version(manifest)?;
+        if validate {
+            // Check toolchain version
+            validate_version(manifest)?;
+        }
         // Keep track of the cause for the new lock file if it turns out we need one.
         let mut new_lock_cause = None;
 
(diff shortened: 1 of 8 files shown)


kayagokalp:
Leaving this as draft until #3130 is merged but this can build entire workspaces in the correct order. I can't wait to get to a point where we can share build artifacts between workspace members 😄

forc-pkg/src/pkg.rs
```diff
@@ -514,6 +514,35 @@ impl BuildPlan {
...
+        let visited_nodes: HashSet<NodeIx> = bfs.iter(&graph).collect();
+        graph.retain_nodes(|_, index| visited_nodes.contains(&index));
+        // Since some nodes are removed from the graph, compilation order needs to be reconstructed.
+        let compilation_order = compilation_order(&graph)?;
+        Ok(Self {
+            graph: graph.clone(),
+            manifest_map: self.manifest_map.clone(),
+            compilation_order,
+        })
+    }
```
mitchmindtree, line 545: Niccceee, nailed it!

I think we can make this even more useful if we split it up into a couple of methods? E.g.

```rust
    /// Produce the node index of the member with the given name.
    pub fn find_member_index(&self, member_name: &str) -> Option<NodeIx> {
        // ...
    }

    /// Produce an iterator yielding indices for the given node and its dependencies in BFS order.
    pub fn node_deps(&self, n: NodeIx) -> impl '_ + Iterator<Item = NodeIx> {
        // ...
    }

    // Produce a new build plan with only the packages required to build the given node.
    pub fn node_plan(&self, n: NodeIx) -> Result<Self> {
        // ...
    }
```

mitchmindtree (commented):
> Also I can't wait to get to a point where we can share build artifacts between workspace members

I think we're super close already!

The core part that needs to change is our `build` invocation. Currently, it just builds an entire `BuildPlan` and assumes the desired output is the last node. Instead, we could:

1. Provide a new argument along the lines of `outputs: &HashSet<NodeIx>`. This is the set of nodes we want to produce output artifacts for.
2. For each output, use the `node_deps` method (see suggestion) to collect and merge all necessary dependency `NodeIx`s into something like `let required: HashSet<NodeIx>`.
3. When iterating over the `BuildPlan`'s compilation order, skip nodes that aren't in `required`.
4. If the built node is in `outputs`, collect its `BuiltPackage` ready to be returned.

Would be cool to see that land in this PR, but happy to see it land in a follow-up if you prefer!

kayagokalp:
Ahh just saw the history, I think I should have rebased to keep the history clean, will try to do so in the following branches created from this one. 

As for sharing artifacts in between, explanation seems perfect thanks a lot! I will open up an issue and tackle it right after we can do all basic operations like building deploying etc 👍

forc-pkg/src/pkg.rs
```diff
@@ -2369,13 +2411,28 @@ pub fn build_with_options(build_options: BuildOpts) -> Result<Built> {
...
+                build_options.pkg.locked,
+                build_options.pkg.offline,
+            )?;
+            let manifest_map = build_plan.manifest_map();
+            let graph = build_plan.graph();
+            for member in build_plan.member_nodes() {
+                let pinned = &graph[member];
+                info!("Building {}", pinned.name);
+                build_package_with_options(&manifest_map[&pinned.id()], build_options.clone())?;
+            }
```
mitchmindtree, line 2434: We also won't need this loop or multiple calls to `build_package_with_options`.

forc-pkg/src/pkg.rs
```diff
@@ -2369,13 +2411,28 @@ pub fn build_with_options(build_options: BuildOpts) -> Result<Built> {
...
+            Ok(Built::Workspace)
```
mitchmindtree, line 2435: I think we should be returning all of the built packages as a part of the `Workspace` variant, perhaps within a `BuiltWorkspace` type.

mitchmindtree (changes requested):
@kayagokalp thanks, looking cleaner!

Thinking on this more, I think we should address [this](https://github.com/FuelLabs/sway/pull/3146#pullrequestreview-1167913135) as a part of this PR. Otherwise we're going to need to remove/change most of the code in follow-up PRs anyway, and I think the suggestions in that comment also apply equally to each of the other commands (check, deploy, test) too. I think we should get this workspace building right first, then we can use it as a reference when implementing the other commands.

test/src/e2e_vm_tests/mod.rs
```diff
@@ -109,7 +109,12 @@ impl TestContext {
 
                 let (result, ..) = harness::compile_to_bytes(&name, &context.run_config, false);
                 assert_matches!(result, Ok(_));
-                let compiled = result.unwrap();
+                let compiled = match result.unwrap() {
+                    forc_pkg::Built::Package(built_pkg) => *built_pkg,
+                    forc_pkg::Built::Workspace(_) => {
+                        panic!("workspaces are not supported in the test suite yet")
```
kayagokalp, line 115: I am looking into enabling workspace support in the test suite, it seems like I might need to play with the structure a little but I guess it would be cool to support them with this PR. What are your thoughts? cc @mitchmindtree

forc-pkg/src/pkg.rs
```diff
@@ -2369,14 +2399,66 @@ pub fn build_with_options(build_options: BuildOpts) -> Result<Built> {
...
+    // If this is a workspace we want to have all members in the output.
+    let outputs = match &manifest_file {
+        ManifestFile::Package(pkg_manifest) => std::iter::once(
+            build_plan
+                .find_member_index(&pkg_manifest.project.name)
+                .ok_or_else(|| anyhow!("Cannot found project node in the graph"))?,
+        )
+        .collect(),
+        ManifestFile::Workspace(_) => build_plan.member_nodes().collect(),
+    };
```
kayagokalp, line 2425: Basically this part substitutes for `node_plan` approach in the case of building just a single pkg of a workspace. Since any member of the workspace would result all of the member nodes, if we just want to build one of them I ended up checking like this. Otherwise it ends up building all of the members while building any member of the workspace individually.

forc-pkg/src/pkg.rs
```diff
@@ -2369,14 +2399,66 @@ pub fn build_with_options(build_options: BuildOpts) -> Result<Built> {
...
+    }
+
+    match manifest_file {
+        ManifestFile::Package(_) => {
+            let built_pkg = built_packages
+                .last()
+                .ok_or_else(|| anyhow!("Couldn't find any built package"))?;
+            Ok(Built::Package(Box::new(built_pkg.clone())))
         }
+        ManifestFile::Workspace(_) => Ok(Built::Workspace(built_packages)),
```
kayagokalp, line 2461: I am not happy with this part but couldn't pulled of another way of doing this 🤔

kayagokalp (commented):
I enabled artifact sharing between workspace members and left some review comments for some explaining and questions. Thanks for the great outline @mitchmindtree, I managed to get it working easily with that.

test/src/e2e_vm_tests/mod.rs
kayagokalp: Testing suite can now have "compile" tests for workspaces but not run. I am planning to add it with `forc-client` workspace support pr.

forc-pkg/src/pkg.rs
```diff
@@ -2369,14 +2402,66 @@ pub fn build_with_options(build_options: BuildOpts) -> Result<Built> {
...
+            build_plan
+                .find_member_index(&pkg_manifest.project.name)
+                .ok_or_else(|| anyhow!("Cannot found project node in the graph"))?,
+        )
+        .collect(),
+        ManifestFile::Workspace(_) => build_plan.member_nodes().collect(),
+    };
+    // Build it!
+    let mut built_workspaces = HashMap::new();
+    let built_packages_with_source_map = build(&build_plan, &build_profile, &outputs)?;
```
mitchmindtree, line 2431: Once we merge the `SourceMap` into the `BuiltPackage`, let's change this to
```suggestion
    let built_packages = build(&build_plan, &build_profile, &outputs)?;
```

forc-pkg/src/pkg.rs
```diff
@@ -2213,26 +2330,25 @@ pub const SWAY_BIN_HASH_SUFFIX: &str = "-bin-hash";
...
-) -> Result<BuiltPackage> {
-    let key_debug: String = "debug".to_string();
-    let key_release: String = "release".to_string();
+/// Selects the build profile from all available build profiles in the workspace using build_opts.
+fn build_profile_from_opts(
+    build_profiles: &HashMap<String, BuildProfile>,
+    build_options: &BuildOpts,
+) -> Result<(String, BuildProfile)> {
+    let key_debug = "debug".to_string();
+    let key_release = "release".to_string();
```
mitchmindtree, line 2339: We should avoid constructing (and dynamically allocating) these `String`s until we need them, and we should use `const`s rather than writing the literals here. I think we might already have `const`s somewhere? If not, let's add them.

forc-pkg/src/pkg.rs
```diff
@@ -2213,26 +2330,25 @@ pub const SWAY_BIN_HASH_SUFFIX: &str = "-bin-hash";
...
-    manifest: &PackageManifestFile,
-    build_options: BuildOpts,
-) -> Result<BuiltPackage> {
-    let key_debug: String = "debug".to_string();
-    let key_release: String = "release".to_string();
+/// Selects the build profile from all available build profiles in the workspace using build_opts.
+fn build_profile_from_opts(
+    build_profiles: &HashMap<String, BuildProfile>,
+    build_options: &BuildOpts,
+) -> Result<(String, BuildProfile)> {
```
mitchmindtree, line 2337: Nice job on moving this into its own function too!

forc-pkg/src/pkg.rs
```diff
@@ -2213,26 +2330,25 @@ pub const SWAY_BIN_HASH_SUFFIX: &str = "-bin-hash";
...
-        binary_outfile,
-        debug_outfile,
         build_profile,
         release,
         time_phases,
         tests,
-    } = build_options;
-
+        ..
+    } = build_options.to_owned();
```
mitchmindtree, line 2349: Rather than cloning the entire `BuildOpts` here, let's just clone the fields or elements that we actually need to when we need to (if we do need to).

forc-pkg/src/pkg.rs
mitchmindtree: Nice one, makes sense and looks good!

forc-pkg/src/pkg.rs
```diff
@@ -2369,14 +2402,66 @@ pub fn build_with_options(build_options: BuildOpts) -> Result<Built> {
...
+        ManifestFile::Workspace(_) => build_plan.member_nodes().collect(),
+    };
+    // Build it!
+    let mut built_workspaces = HashMap::new();
+    let built_packages_with_source_map = build(&build_plan, &build_profile, &outputs)?;
+    let output_dir = build_options
+        .clone()
+        .pkg
+        .output_directory
+        .map(PathBuf::from);
```
mitchmindtree, line 2436: This unnecessarily clones the entire `build_options` (performing allocations for all of its `String`s etc) when we only want an owned instance of the output directory. This could be something like:
```rust
let output_dir = build_options.pkg.output_directory.as_ref().map(PathBuf::from);
```

forc-pkg/src/pkg.rs
```diff
@@ -2369,14 +2402,66 @@ pub fn build_with_options(build_options: BuildOpts) -> Result<Built> {
...
+    for (node_ix, (built_package, source_map)) in built_packages_with_source_map
+        .iter()
+        .filter(|(node_ix, _)| outputs.contains(node_ix))
```
mitchmindtree, line 2439: We shouldn't have to filter by `outputs` again - we already do that during `build`.

forc-pkg/src/pkg.rs
```diff
@@ -2369,14 +2402,66 @@ pub fn build_with_options(build_options: BuildOpts) -> Result<Built> {
...
-        ManifestFile::Package(package_manifest) => {
-            let built_package = build_package_with_options(&package_manifest, build_options)?;
-            Ok(Built::Package(Box::new(built_package)))
-        }
-        ManifestFile::Workspace(_) => {
-            bail!("Workspace building is not supported")
+        ManifestFile::Package(pkg_manifest) => {
+            let built_pkg = built_workspaces
+                .get(&pkg_manifest.project.name)
+                .ok_or_else(|| anyhow!("Couldn't find the built package"))?;
```
mitchmindtree, line 2461: In this case it might be better to use use `expect` as not having this package within `built_workspace` would be an error in our logic and should never occur.
```suggestion
                .expect("package didn't exist in workspace");
```

forc-pkg/src/pkg.rs
```diff
@@ -2369,14 +2402,66 @@ pub fn build_with_options(build_options: BuildOpts) -> Result<Built> {
...
+    // Build it!
+    let mut built_workspaces = HashMap::new();
+    let built_packages_with_source_map = build(&build_plan, &build_profile, &outputs)?;
+    let output_dir = build_options
+        .clone()
+        .pkg
+        .output_directory
+        .map(PathBuf::from);
+    for (node_ix, (built_package, source_map)) in built_packages_with_source_map
+        .iter()
```
mitchmindtree, line 2438: We should use `into_iter` here to take the elements by value and avoid the need to clone the entirety of each package below.
```suggestion
        .into_iter()
```

forc-pkg/src/pkg.rs
```diff
@@ -2369,14 +2402,66 @@ pub fn build_with_options(build_options: BuildOpts) -> Result<Built> {
...
     match manifest_file {
-        ManifestFile::Package(package_manifest) => {
-            let built_package = build_package_with_options(&package_manifest, build_options)?;
-            Ok(Built::Package(Box::new(built_package)))
-        }
-        ManifestFile::Workspace(_) => {
-            bail!("Workspace building is not supported")
+        ManifestFile::Package(pkg_manifest) => {
+            let built_pkg = built_workspaces
+                .get(&pkg_manifest.project.name)
```
mitchmindtree, line 2460:
```suggestion
                .remove(&pkg_manifest.project.name)
```
Let's remove the package so that we have it by value, and so that we don't need to clone it below.

forc-pkg/src/pkg.rs
```diff
@@ -2399,15 +2484,25 @@ fn contract_id(built_package: &BuiltPackage) -> ContractId {
...
+    outputs: &HashSet<NodeIx>,
+) -> anyhow::Result<Vec<(NodeIx, (BuiltPackage, SourceMap))>> {
     //TODO remove once type engine isn't global anymore.
     sway_core::clear_lazy_statics();
+    let mut built_packages = Vec::new();
+
+    let required: HashSet<NodeIx> = outputs
+        .iter()
+        .flat_map(|output_node| plan.node_deps(*output_node))
+        .collect();
```
mitchmindtree, line 2496: Nice :ok_hand:

forc-pkg/src/pkg.rs
```diff
@@ -2437,15 +2532,12 @@ pub fn build(
         if let TreeType::Library { .. } = built_package.tree_type {
             lib_namespace_map.insert(node, namespace.into());
         }
-        last_pkg = Some(built_package);
         source_map.insert_dependency(manifest.dir());
+        standardize_json_abi_types(&mut built_package.json_abi_program);
+        built_packages.push((node, (built_package, source_map)));
```
mitchmindtree, line 2537: We should be careful not to collect all dependencies in memory unnecessarily too - we only want to collect the `outputs`.
```suggestion
        if outputs.contains(&node) {
            built_packages.push((node, (built_package, source_map)));
        }
```

mitchmindtree (changes requested):
Excellent stuff @kayagokalp! Just a few more small suggestions after the refactor, but overall looking great.

forc-pkg/src/pkg.rs
kayagokalp: I didn't had this kind of check at first but I needed to add this because without this filtering std and core also gets output artifacts even in the scenario where they are implicit depdendencies.

Considering some members of the workspace can depend on non workspace member pkgs (like std and core) without this those non-member dependencies' out folders also gets updated and their bytecode size etc gets printed.

I am on my phone rn so my explanation could be a little off, I will come-up with a cleaner explanation with examples about my concern here once i am back online

mitchmindtree: No worries! I think the reason why dependencies like `std` and `core` were showing up in the output of `build` is because we were missing this:

https://github.com/FuelLabs/sway/pull/3146#discussion_r1014923420

kayagokalp: Ohh, yeah that makes a lot more sense! Thanks

JoshuaBatty (approved):

PR #3147, opened 2022-10-26
Fixes e2e tests using checks with UNIX paths.

esdrubal:
On Windows the output error and warning paths use '\' as path separator.
e2e tests have checks with hardcoded UNIX paths that use '/' as path separator character.

Solution was to replace '\' by '/' on `compile_and_capture_output`.


mitchmindtree:
Thanks for the PR @esdrubal!

I'm not sure we should start adding patches like this until we actually decide to properly support Windows within Sway, Forc, etc too. https://github.com/FuelLabs/sway/issues/1526 shares a rough path ahead for this for anyone interested in having a go at native Windows support.

otrho (approved):
I agree.  A tiny change like this is helpful in the short term and can't really hurt.

PR #3151, opened 2022-10-26
Include `#[test]` functions as entry points during dead code analysis

mitchmindtree:
Refactors the collection of flow graph entry points into a function, and includes functions decorated with the `#[test]` attribute.

Short-hand methods have been added to `TyFunctionDeclaration` for checking if the function is `main`, a unit test or if it is an entry point.

This is another step towards #2985.

### DIFF
--- a/sway-core/src/control_flow_analysis/dead_code_analysis.rs
+++ b/sway-core/src/control_flow_analysis/dead_code_analysis.rs
@@ -2,13 +2,14 @@ use super::*;
 use crate::{
     declaration_engine::declaration_engine::*,
     language::{parsed::TreeType, ty, CallPath, Visibility},
+    transform::AttributeKind,
     type_system::{to_typeinfo, TypeInfo},
 };
 use petgraph::{prelude::NodeIndex, visit::Dfs};
 use std::collections::BTreeSet;
 use sway_error::error::CompileError;
 use sway_error::warning::{CompileWarning, Warning};
-use sway_types::{span::Span, Ident, Spanned};
+use sway_types::{constants::DEFAULT_ENTRY_POINT_FN_NAME, span::Span, Ident, Spanned};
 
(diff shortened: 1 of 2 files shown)


sway-core/src/control_flow_analysis/dead_code_analysis.rs
```diff
@@ -108,96 +108,98 @@ impl ControlFlowGraph {
...
+                    ControlFlowGraphNode::ProgramNode(ty::TyAstNode {
+                        span,
+                        content:
+                            ty::TyAstNodeContent::Declaration(ty::TyDeclaration::FunctionDeclaration(
+                                decl_id,
+                            )),
+                        ..
+                    }) => {
+                        let decl = de_get_function(decl_id.clone(), span)?;
+                        if !decl.is_entry() {
```
mitchmindtree, line 137: Here is the updated entry point check for predicate and script functions.

sway-core/src/control_flow_analysis/dead_code_analysis.rs
```diff
@@ -108,96 +108,98 @@ impl ControlFlowGraph {
...
+                    ControlFlowGraphNode::OrganizationalDominator(_) => continue,
+                    ControlFlowGraphNode::ProgramNode(ty::TyAstNode {
+                        content:
+                            ty::TyAstNodeContent::Declaration(ty::TyDeclaration::FunctionDeclaration(
+                                decl_id,
+                            )),
+                        ..
+                    }) => {
+                        let decl = de_get_function(decl_id.clone(), &decl_id.span())?;
+                        decl.visibility == Visibility::Public || decl.is_test()
```
mitchmindtree, line 158: And here is the added `#[test]` check for contract and library functions.

JoshuaBatty (approved):

Issue #2457, opened 2022-08-03
Support for mutable arrays


simonr0204:
Currently, when attempting 

```
let mut my_array: [u64; 3] = [1, 2, 3];
a[1] = 42;
```

The compiler errors with `This expression is not valid on the left hand side of a reassignment`

It would be good to support mutable arrays. This should also be possible in storage, so that:

```
storage {
    my_array: [u64; 3] = [1, 2, 3],
}

//...
storage.my_array[1] = 42
```
also works. (Note: storage arrays would be mutable by definition)

mohammadfawaz:
Yeah mutable arrays on the stack are not supported yet: https://github.com/FuelLabs/sway/issues/428. We have not prioritized that work yet because `Vec` is now available.
Arrays in storage are also not supported yet: https://github.com/FuelLabs/sway/issues/1182.

@simonr0204 are you blocked by this? We are aware of these deficiencies, we just have not prioritized the work yet.

simonr0204:
Not really _blocked_, just seen occasions where a statically sized array is just the right way to model something.

@sezna said

> "Would you mind filing an issue if there isn’t already one? It should not be too hard. Please label it as critical as we should really be able to do this"

So... I did 🙂

mohammadfawaz:
Makes sense! Thanks for filing. I can organize the issues soon :)

PR #3153, opened 2022-10-26
Add support for mutable arrays.

tritao:
This adds support for reassigning array elements.

We now support this by extending reassignment `ProjectionKind` with a new `ArrayIndex` variant which keeps track of the typed array index reassignnment during semantic analysis.

Afterwards we implement this new projection in IR generation by lowering to the `insert_element` IR instruction which updates the array element value.

Closes https://github.com/FuelLabs/sway/issues/2457.


mohammadfawaz:
Do you mind writing some documentation for this? We do have [a section on arrays ](https://fuellabs.github.io/sway/master/basics/built_in_types.html#arrays)in the Sway Book that we could add to.

mohammadfawaz:
I'm going to be a bit annoying and ask for more tests just to know how much we can do with this. For example:
* Multidimensional arrays (arrays inside arrays): am I able to read/write at any level? For example:
```rust
let a = [[0,1,2], [2,3,4], [4, 5, 6]];
a[0][1] += 2; // Does this work?
a[0] = [2, 1, 0]; // Does this work?
```
* Arrays containing structs or enums: can I update a whole struct at a given index or even just one field of it? Can I update a whole enum?
* Does `ref mut` for arrays work with and without inlining? We have an `#inline(always/never)` attribute that we can play around with now.
That's all I can think of now. Thanks!

emilyaherbert:
> Presumably, I _shouldn't_ be able to modify an element at a particular index if the array is `mut` or `ref mut`?

Sorry, I'm not following what you mean by this @mohammadfawaz, could you clarify? I would think that these would be the only two cases in which you should be allowed to modify an element at a particular index.

>  If so, then contract methods with arrays as inputs should always be immutable.

How about supporting `mut` and disallowing `ref mut`? I would think that local mutability for "non-`ref`" values would be fine. Unless I'm misunderstanding a special case regarding arrays.

sway-error/src/error.rs
```diff
@@ -247,6 +247,12 @@ pub enum CompileError {
         span: Span,
         actually: String,
     },
+    #[error("\"{name}\" is a {actually}, not an indexable expression.")]
```
emilyaherbert, line 250: Could you reword this to `\"{name}\" is a {actually}, which is not an indexable expression.` ?

emilyaherbert (changes requested):
+1 to more testing. In addition to @mohammadfawaz's suggestions for additional tests, would you include:
- tests which have return values other than `0`? Or that use `assert` to check for certain non-zero values during the test?
- a test swapping array elements between multiple arrays
- tests with arrays holding complex types
- a `should_fail` test to check for the "non indexable" error message that you added
- a `should_fail` test to check to trying to mutate a non-mutable array
- a `should_fail` test for trying to pass a non mutable array to a `ref mut` argument
- a `should_fail` test for assigning an element in the array of the wrong type

Would you also document in this PR the effects of this PR on interior mutability laws in Sway. We should also consider adding a section in the docs about this. Here is the Rust bit on this for reference: https://doc.rust-lang.org/reference/interior-mutability.html

mohammadfawaz:
> > Presumably, I _shouldn't_ be able to modify an element at a particular index if the array is `mut` or `ref mut`?
> 
> Sorry, I'm not following what you mean by this @mohammadfawaz, could you clarify? I would think that these would be the only two cases in which you should be allowed to modify an element at a particular index.

🤦 I, of course, meant  "if the array is not `mut` nor `ref mut`?"

mohammadfawaz:
@tritao converting to a draft until this is 100% ready for review.

mohammadfawaz (approved):
Does `a[0][0]` fail if `a` is 1-dimensional?

emilyaherbert (approved):

blackbeam:
@tritao, the following code reverts on beta-3 (sway 0.35.5). Is this behavior expected?

```rust
// We are inside a method with `ref mut self` receiver
// `self.keys.keys` is `[Address; LEN]`
// `authority` is `Address`

require(self.keys.keys[i].value == ZERO_B256, 200);
require(authority.value != ZERO_B256, 300);

log(self.keys.keys[0].value); // this logs zeroed value

// Also tried:
// * self.keys.keys[0] = authority
self.keys.keys[0].value = authority.value;

log(self.keys.keys[0].value); // this also logs zeroed value for some reason
log(authority.value); // this logs non-zeroed value as expected

require(self.keys.keys[0].value == authority.value, 100); // this line reverts
```

And here is an error:

```
thread 'basic' panicked at 'called `Result::unwrap()` on an `Err` value: RevertTransactionError {
  reason: "100",
  revert_id: 18446744073709486080,
  receipts: [Call {
    id: 0000000000000000000000000000000000000000000000000000000000000000,
    to: b1025663a4494e96a24773feebf8946ac3a470211e63f7b6e1473489dd78a540,
    amount: 0,
    asset_id: 0000000000000000000000000000000000000000000000000000000000000000,
    gas: 1000000,
    param1: 1084460444,
    param2: 10472,
    pc: 11728,
    is: 11728 },
  LogData {
    id: b1025663a4494e96a24773feebf8946ac3a470211e63f7b6e1473489dd78a540,
    ra: 0, rb: 16, ptr: 96384, len: 32,
    digest: 66687aadf862bd776c8fc18b8e9f8e20089714856ee233b3902a591d0d5f2925,
    data: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
    pc: 25692, is: 11728 },
  LogData {
    id: b1025663a4494e96a24773feebf8946ac3a470211e63f7b6e1473489dd78a540,
    ra: 0, rb: 17, ptr: 96384, len: 32,
    digest: 66687aadf862bd776c8fc18b8e9f8e20089714856ee233b3902a591d0d5f2925,
    data: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
    pc: 25724, is: 11728 },
  LogData {
    id: b1025663a4494e96a24773feebf8946ac3a470211e63f7b6e1473489dd78a540,
    ra: 0, rb: 18, ptr: 96176, len: 32,
    digest: db1e67338460cbf3f05736ec57909541836b655785e0717f1f3c19c2464deaab,
    data: [93, 153, 238, 150, 107, 66, 205, 143, 199, 189, 209, 54, 75, 56, 145, 83, 169, 231, 139, 66, 183, 212, 166, 145, 71, 6, 116, 232, 23, 136, 141, 78],
    pc: 25740, is: 11728 },
  Log {
    id: b1025663a4494e96a24773feebf8946ac3a470211e63f7b6e1473489dd78a540,
    ra: 100, rb: 19, rc: 0, rd: 0, pc: 28516, is: 11728 },
  Revert {
    id: b1025663a4494e96a24773feebf8946ac3a470211e63f7b6e1473489dd78a540,
    ra: 18446744073709486080, pc: 28524, is: 11728 },
  ScriptResult { result: Revert, gas_used: 8470 }]
}'
```

PR #3154, opened 2022-10-26
Patch to fix malformed test + non-determinism in trait lookup via iterating through a hashmap

emilyaherbert:
This is a fix to an issue raised by @JoshuaBatty regarding a failing `type_system` test. The main reason that this test was failing is because it was querying, either directly or indirectly, both the local copy of `TypeEngine` and the `lazy_static` copy.

This PR also patches a fix to `TraitMap` to eliminate non-determinism---iterating through a hashmap was causing some non-determinism.

### DIFF
--- a/sway-core/src/semantic_analysis/namespace/trait_map.rs
+++ b/sway-core/src/semantic_analysis/namespace/trait_map.rs
@@ -9,7 +9,7 @@ type TraitName = CallPath;
 /// Map of function name to [TyFunctionDeclaration](ty::TyFunctionDeclaration)
 type TraitMethods = im::HashMap<String, ty::TyFunctionDeclaration>;
 /// Map of trait name and type to [TraitMethods].
-type TraitImpls = im::HashMap<(TraitName, TypeId), TraitMethods>;
+type TraitImpls = im::Vector<((TraitName, TypeId), TraitMethods)>;
 
 /// Map holding trait implementations for types.
 ///
@@ -138,11 +138,27 @@ impl TraitMap {
     /// Given [TraitMap]s `self` and `other`, extend `self` with `other`,
     /// extending existing entries when possible.
     pub(crate) fn extend(&mut self, other: TraitMap) {
-        for (key, other_trait_methods) in other.trait_impls.into_iter() {
-            self.trait_impls
-                .entry(key)
-                .or_insert(other_trait_methods.clone())
-                .extend(other_trait_methods.into_iter());
+        let mut add_after = vec![];
+        for ((other_trait_name, other_type_id), other_trait_methods) in
+            other.trait_impls.into_iter()
+        {
+            let mut found_match = false;
+            for ((trait_name, type_id), trait_methods) in self.trait_impls.iter_mut() {
+                if trait_name.clone() == other_trait_name && *type_id == other_type_id {
+                    trait_methods.extend(other_trait_methods.clone());
+                    found_match = true;
+                    break;
+                }
+            }
+            if !found_match {
+                add_after.push((
+                    (other_trait_name, other_type_id),
+                    other_trait_methods.clone(),
+                ));
+            }
+        }
+        for elem in add_after.into_iter() {
+            self.trait_impls.push_back(elem);
         }
     }
 
--- a/sway-core/src/semantic_analysis/namespace/trait_map.rs
+++ b/sway-core/src/semantic_analysis/namespace/trait_map.rs
@@ -9,7 +9,7 @@ type TraitName = CallPath;
 /// Map of function name to [TyFunctionDeclaration](ty::TyFunctionDeclaration)
 type TraitMethods = im::HashMap<String, ty::TyFunctionDeclaration>;
 /// Map of trait name and type to [TraitMethods].
-type TraitImpls = im::Vector<((TraitName, TypeId), TraitMethods)>;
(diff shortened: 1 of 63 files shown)


JoshuaBatty (dismissed):
Nice work in finding the cause of this! testing locally and the test is passing again. Thanks!

sway-core/src/semantic_analysis/namespace/trait_map.rs
```diff
@@ -9,7 +9,7 @@ type TraitName = CallPath;
 /// Map of function name to [TyFunctionDeclaration](ty::TyFunctionDeclaration)
 type TraitMethods = im::HashMap<String, ty::TyFunctionDeclaration>;
 /// Map of trait name and type to [TraitMethods].
-type TraitImpls = im::HashMap<(TraitName, TypeId), TraitMethods>;
+type TraitImpls = im::Vector<((TraitName, TypeId), TraitMethods)>;
```
mitchmindtree, line 12: `im` also exposes an [`OrdMap` collection](https://docs.rs/im/latest/im/ordmap/struct.OrdMap.html) which might be closer to the original behaviour - would that simplify this change (e.g. the `extend` implementation at least)?

sway-core/src/type_system/type_engine.rs
```diff
@@ -44,6 +44,17 @@ impl TypeEngine {
         }
     }
 
+    /// Currently the [TypeEngine] is a lazy static object, so when we run
+    /// cargo tests, we can either choose to use a local [TypeEngine] and bypass
+    /// all of the global methods or we can use the lazy static [TypeEngine].
+    /// This method is for testing to be able to bypass the global methods for
+    /// the lazy static [TypeEngine] (contained within the call to hash in the
+    /// id_map).
+    #[allow(dead_code)]
```
mitchmindtree, line 53: If this is only for testing, maybe we can add a `#[cfg(test)]`?
```suggestion
    #[cfg(test)]
    #[allow(dead_code)]
```

mitchmindtree (commented):
Ahhh I ran into this last night too but forgot to check for an issue - good catch @emilyaherbert!

sway-core/src/semantic_analysis/namespace/trait_map.rs
emilyaherbert: Hmm, I don't think so ? The `insert` method on `OrdMap` is copy-on-write, which is unnecessary for this application. https://docs.rs/im/latest/im/ordmap/struct.OrdMap.html#method.insert

mitchmindtree: I mean that semantically, the `OrdMap` is much closer to the original `HashMap` behaviour, but is ordered and deterministic, and would mean we don't need any of the changes that were made to `extend` below.

> The insert method on OrdMap is copy-on-write, which is unnecessary for this application.

Note that a copy only occurs if the data is shared with another instance, (i.e. if the `OrdMap` was previously cloned elsewhere) and only the shared parts are copied.

mitchmindtree (approved):
Approving as I don't want to block this, but I still think `OrdMap` is maybe a nicer choice for the `TraitImpls` type alias.

emilyaherbert:
> Note that a copy only occurs if the data is shared with another instance, (i.e. if the OrdMap was previously cloned elsewhere) and only the shared parts are copied.

Ahhh I see. OrdMap it is!

vaivaswatha:
> > Note that a copy only occurs if the data is shared with another instance, (i.e. if the OrdMap was previously cloned elsewhere) and only the shared parts are copied.
> 
> Ahhh I see. OrdMap it is!

Sorry for the late input. If determinism is the only thing that matter and not the actual order, then [FxHashMap](https://docs.rs/rustc-hash/latest/rustc_hash/type.FxHashMap.html) could be a more efficient option.

vaivaswatha:
> > > Note that a copy only occurs if the data is shared with another instance, (i.e. if the OrdMap was previously cloned elsewhere) and only the shared parts are copied.
> > 
> > 
> > Ahhh I see. OrdMap it is!
> 
> Sorry for the late input. If determinism is the only thing that matter and not the actual order, then [FxHashMap](https://docs.rs/rustc-hash/latest/rustc_hash/type.FxHashMap.html) could be a more efficient option.

I'm already using this in the compiler. FxHashMap is faster too if the key type is small sized. If we want exactly `HashMap`'s performance, but with determinism, then we should just instantiate `HashMap` with a non-random hasher, similar to what FxHashMap does (https://docs.rs/rustc-hash/latest/src/rustc_hash/lib.rs.html#43).

So something like `pub type SwayHashMap<K, V> = HashMap<K, V, BuildHasherDefault<DefaultHasher>>;` would give us exactly what `HashMap` gives, but with determinism.

emilyaherbert:
Hi @vaivaswatha, I like this idea! I think we should push this patch through, but I created a new issue #3162 to track using FxHashMap.

mitchmindtree (approved):
@vaivaswatha cool to hear `FxHashMap` is deterministic!

I'll also just add that the only time we really want to use the `im` collections is when there is lots of cloning or structural sharing going on (hence why it's a great fit in the `TypeCheckContext` and in particular, the `Namespace`).

I haven't looked closely enough at the use of `TraitImpls` to see if it's really the right choice here, but just something to keep in mind w.r.t. the use of `im` collections!

JoshuaBatty (approved):

Issue #3112, opened 2022-10-24
U128 and common inefficient library types cause unreasonable bloat


sezna:
This issue generally tracks the fact that applications that make heavy use of wider integers like https://github.com/sway-libs/amm are taking a long time to compile, have significant bytecode bloat, and are gas inefficient.

We are currently investing in optimizing bytecode size via IR optimization passes. A few more strategic optimization passes will hopefully get us some major benefits, but we can't solely rely on that. A couple of other methods were proposed:

1. VM instructions for things like quad-add.
2. Bespoke intrinsics (or VM ops, but more likely intrinsics) for common mathematical operations like `(x * y) / z` -- this is extremely common in AMMs or other smart contracts that calculate ratios.

In general, we would like adding VM operations to be our last resort, as the VM should remain as minimal as possible. Additionally, it is a bad precedent to simply add opcodes when we have issues generating efficient code.

The consensus was that we can start by adding compiler intrinsics for specific math operations like `(x * y) / z` and quad add. These can utilize the fact that the result will fit in a `u64` while the components (`x * y`) might not, to generate more efficient code. If after this, plus more optimization passes, we are still unable to generate efficient code, we will revisit the possibility of more VM opcodes.

If this summary is correct, I'll make sub-issues for these intrinsics and tracking bytecode bloat, perhaps in the CI.

cc @FuelLabs/sway-compiler

PR #3157, opened 2022-10-26
Add intrinsics for operating on raw_ptrs

AlicanC:
This PR:
- Adds `__ptr_add<T>(ptr: raw_ptr, count: u64)` (which basically does `ptr + __size_of::<T>() * count`) and its sub counterpart to be used in pointer offset calculations.
- Makes `std::alloc::*` and `raw_ptr::*` have type arguments and work in increments of `size_of::<T>()` instead of bytes.
- Removes test `fixing_generic_type` which was broken instead of maintaining it. This test was created to reproduce an issue I was having a few months ago. Fixing this allowed other things to be built so there are other tests that touch this case.

---

This abstracts away more hacky stuff from the Sway side, and also significantly reduces bytecode size in some tests, especially in `Vec` stuff:

| Test                                      | Before | After  | Change  |
| ----------------------------------------- | ------ | ------ | ------- |
| should_pass/stdlib/alloc | 684 | 948 | 27.85% |
| should_fail/vec_swap_param2_out_of_bounds | 1028 | 1132 | 9.19% |
| should_fail/vec_swap_param1_out_of_bounds | 1028 | 1132 | 9.19% |
| should_pass/stdlib/raw_ptr | 1964 | 2020 | 2.77% |
| should_pass/stdlib/vec_swap | 32764 | 33460 | 2.08% |
| should_pass/stdlib/vec | 175952 | 178992 | 1.70% |

Related: https://github.com/FuelLabs/sway/issues/3112


mohammadfawaz:
That's amazing 🚀 . Do you mind checking the impact of this on the sway-applications repo?

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
```diff
@@ -587,7 +587,10 @@ impl ty::TyIntrinsicFunctionKind {
                     warnings,
                     errors
                 );
-                let is_valid_arg_ty = matches!(arg_ty, TypeInfo::UnsignedInteger(_));
+                let is_valid_arg_ty = matches!(
+                    arg_ty,
+                    TypeInfo::UnsignedInteger(_) | TypeInfo::RawUntypedPtr
```
mohammadfawaz, line 592: Does this mean that I can now do `__add(raw_ptr_1, raw_ptr_2)`? What about `__add(5, raw_ptr_2)`? I think both should be disallowed.

AlicanC: RHS is checked below and shouldn't allow that. If LHS is rq_ptr, RHS must be u64.

mohammadfawaz: Makes sense. I guess the type checker will complain if `with_type_annotation` contains the wrong type.

One more thing: the current code allows `__mul` and `__div` which I assume we don't want to allow for `raw_ptr`. Maybe a small check somewhere could be useful?

mohammadfawaz:
Ok so one more comment. Otherwise, is this ready for review? You also seem to have a merge conflict now :/

AlicanC:
> Otherwise, is this ready for review?

Well IDK really.

This only improves `Vec` and if we are to improve `Vec` it would be better to have separate intrinsics like `__ptr_add(ptr: raw_ptr, len: u64, count: u64)` where we do `raw_ptr += len * count` which is what `Vec` really needs.

If we had typed pointers, then the signature would be `__ptr_add<T>(ptr: ptr<T>, count: u64)` where we get `len` from `__size_of::<T>`.

mohammadfawaz:
> > Otherwise, is this ready for review?
> 
> Well IDK really.
> 
> This only improves `Vec` and if we are to improve `Vec` it would be better to have separate intrinsics like `__ptr_add(ptr: raw_ptr, len: u64, count: u64)` where we do `raw_ptr += len * count` which is what `Vec` really needs.
> 
> If we had typed pointers, then the signature would be `__ptr_add<T>(ptr: ptr<T>, count: u64)` where we get `len` from `__size_of::<T>`.

You may be right in the sense that it's odd to reuse the binary intrinsics `__add` and `__sub` and hacking them to allow accepting two values of different types (`raw_ptr` and `u64`). Besides, we also have to disallow `__mul` and `__div` for `raw_ptr`. @otrho what do you think?

otrho:
Yeah, you're (both) right.  `__ptr_add()` sounds like a way better solution.  Doing arithmetic on pointers should obviously be discouraged and constrained.

AlicanC:
Ok, so this version adds `__ptr_add<T>(ptr: raw_ptr, count: u64)` and its sub.

I've also made `std::alloc::*` and `raw_ptr::*` have type arguments and work in increments of `size_of::<T>()` instead of bytes. This introduces a constraint since `T` has to be multiples of 8 in Sway. (You can't alloc 1 byte, 7 bytes, etc. anymore, only multiples of 8.) I think this is a nice constraint to have since Sway is word-sized and not byte-sized.

I've also removed a test that was broken. It was a test I was familiar with, and what's being tested there should have no way of not being caught by other tests since we started making heavy usage of it after it was fixed.

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/alloc/src/main.sw
```diff
@@ -19,20 +18,25 @@ fn sw(ptr: raw_ptr, val: u64) {
...
 }
 
-fn main() -> bool {
-    let hp_start = heap_ptr();
+fn heap_ptr() -> raw_ptr {
+    asm(ptr) {
+        addi ptr hp i1;
+        ptr: raw_ptr
+    }
+}
```
mohammadfawaz, line 26: What is the reason for not using `heap_ptr()` from the standard library directly?

mohammadfawaz (commented):
This seems more reasonable and safer. Thanks!
* Just one small question in test
* Also, why are we deleting the test `fixing_generic_type`?
* Finally, will you please update the PR description to reflect the new approach?

mohammadfawaz (dismissed):

AlicanC:
@mohammadfawaz updated the PR desc!

mohammadfawaz (dismissed):

mohammadfawaz (approved):

vaivaswatha (approved):

PR #3166, opened 2022-10-27
Refactor `forc_pkg::BuildOptions` into its parts

mitchmindtree:
This is in anticipation of introducing a new `forc-test` library crate that will contain the logic for building and executing tests. It will expose a similar `TestOpts` type that shares some of these parts, but not all.

Also splits up the forc `BuildCommand` for easier sharing between sets of clap arguments (using the `clap` `flatten` attribute) that require building a project (e.g. build, test).

Also adds two missing args to the `RunCommand` (`release` and `build-profile`).

### DIFF
--- a/forc-plugins/forc-client/src/ops/deploy/op.rs
+++ b/forc-plugins/forc-client/src/ops/deploy/op.rs
@@ -1,5 +1,5 @@
 use anyhow::{bail, Result};
-use forc_pkg::{BuildOptions, PackageManifestFile};
+use forc_pkg::{self as pkg, PackageManifestFile};
 use fuel_gql_client::client::FuelClient;
 use fuel_tx::{Output, Salt, TransactionBuilder};
 use fuel_vm::prelude::*;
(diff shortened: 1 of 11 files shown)


kayagokalp (approved):
lgtm 👍

forc-plugins/forc-client/src/ops/deploy/op.rs
```diff
@@ -72,3 +54,31 @@ pub async fn deploy(command: DeployCommand) -> Result<fuel_tx::ContractId> {
...
+            json_storage_slots: cmd.minify_json_storage_slots,
+        },
+        build_profile: cmd.build_profile.clone(),
+        release: cmd.release,
+        time_phases: cmd.time_phases,
+        binary_outfile: cmd.binary_outfile.clone(),
+        debug_outfile: cmd.debug_outfile.clone(),
+        tests: false,
+    }
+}
```
bingcicle, line 84: nit: would we be able to share `build_opts_from_cmd` in the root of `forc-client` since it seems to be duplicated between both `deploy` and `run`, or is there a chance they could differ in future? Still going to give the tick regardless!

bingcicle (approved):

forc-plugins/forc-client/src/ops/deploy/op.rs
mitchmindtree: Unfortunately while the run and deploy commands both look the same, they're different types and so need different functions at the moment! That said, if we can share the common build arguments (as this PR starts doing between the build and test commands) then we could construct the BuildOpts from the shared build arguments. I almost did that for this PR, but didn't want to introduce `forc` as a dependency to `forc-client` without giving it more thought. I'm thinking we can possibly split out the CLI argument types into a `forc-cli-args` crate that can be shared, but I might leave that for future work.

PR #3168, opened 2022-10-27
Add CEI pattern static analysis

### LINKED ISSUE #10: Detect and handle re-entrancy patterns
Re-entrancy attacks have been a huge plague on Ethereum smart contract development. Consider the following contract:
```solidity
contract VulnerableContract {
    mapping(address => uint256) balances;

    function deposit() public payable {
        balances[msg.sender] += msg.value;
    }

    function withdraw(uint256 amount) public {
        require(balances[msg.sender] >= amount);
        msg.sender.call.value(amount)();
        balances[msg.sender] -= amount;
    }
}
```

If `msg.sender` is a contract, `call` calls the contract. A maliciously-crafted contract can then call `withdraw` again, and the `require` will pass since `balances` hasn't been updated. Repeating this potentially up to the block gas limit, the entire `VulnerableContract`'s balance can be drained.

This is resolved by the checks-effects-interactions pattern (more info: https://fravoll.github.io/solidity-patterns/checks_effects_interactions.html), where effects (writing to state) are applied before interactions (contract calls).

Detecting this should be quite easy in the compiler, which should have a flag (by default on and by default error) to automatically provide feedback to developers if they don't follow the checks-effects-interactions pattern. Developers that want to can disable it for parts of the code via a pragma.

anton-trunov:
CEI stands for "Checks, Effects, Interactions".  We check no storage reads/writes (effects) occur after calling external contracts (interaction). See this [blog post](https://fravoll.github.io/solidity-patterns/checks_effects_interactions.html) and [this blogpost](https://chainsecurity.com/curve-lp-oracle-manipulation-post-mortem) for more detail.

TODOs:
- [x] analyze libraries, scripts, and predicates (only contracts are analyzed at the moment)
- [x] process assembly blocks
- [x] what is the effects of the `__revert` intrinsic
- [x] Treat nested code blocks
- [x] more tests
  * [x] Tests using `storage.var = ...`
  * [x] Tests using `asm` blocks directly
  * [x] Tests using complex control flow
  * [x] Tests having multiple contract calls or multiple storage writes in various scenarios
  * [ ] Tests with nested code blocks (there is a very simple test, this needs more testing in a separate PR)
- [x] check for storage reads after interaction 
   * [x] Storage read intrinsics
   * [x] `StorageAccess` expressions
   * [x] `asm` blocks using `srw` or `srwq`
   * [x] `asm` blocks using [`bal`](https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/instruction_set.md#bal-balance-of-contract-id) because `bal` actually reads from storage
- [x] ~~CLI option or attribute to control the warnings~~ (an internal discussion showed that we cannot do this at the moment, I guess this should be handled as a separate issue)
- [ ] open an issue about unification of the currently existing storage annotation checks and effects inference in this PR

close #10


anton-trunov:
An example of a warning issued by the CEI analyzer:
```
warning
  --> /Users/anton/fuel.network/sway/reentrancy/src/main.sw:29:5
   |
27 |
28 |       // interaction
29 |       other_contract.deposit(gas, amount, color, unused);
   |  _____-
30 | |     // effect -- therefore violation of CEI where effect should go before interaction
31 | |     let storage_key = 0x3dba0a4455b598b7655a7fb430883d96c9527ef275b49739e7b0ad12f8280eae;
32 | |     store(storage_key, ());
   | |__________________________- Storage modification after external contract interaction. Consider making all storage writes before calling another contract
33 |     }
34 |   }
   |
____
```

sway-core/src/semantic_analysis/cei_pattern_analysis.rs
```diff
@@ -0,0 +1,258 @@
...
+    LookingForInteraction,
+    LookingForStorageWrite,
+}
+
+// TODO: process libraries, scripts and predicates
+pub(crate) fn program(prog: &ty::TyProgram) -> Vec<CompileWarning> {
+    match &prog.kind {
+        ty::TyProgramKind::Library { .. } => vec![],
+        ty::TyProgramKind::Script { .. } => vec![],
+        ty::TyProgramKind::Predicate { .. } => vec![],
```
mohammadfawaz, line 31: The good news is that you don't have to worry about libraries, scripts, or predicates because they can't access storage! Libraries may access storage in the future, but we'll cross that bridge when we understand how that looks like.

sway-core/src/semantic_analysis/cei_pattern_analysis.rs
```diff
@@ -0,0 +1,258 @@
...
+    StorageWrite, // storage modification
+}
+
+enum CEIAnalysisState {
+    LookingForInteraction,
+    LookingForStorageWrite,
+}
+
+// TODO: process libraries, scripts and predicates
+pub(crate) fn program(prog: &ty::TyProgram) -> Vec<CompileWarning> {
```
mohammadfawaz, line 27: Should `program` have a more descriptive name?

mohammadfawaz: I suppose the same applies to many functions in the module. I see that you're naming them based on what program entity they run on, but something a bit more descriptive would go a long way here.

sway-core/src/semantic_analysis/cei_pattern_analysis.rs
```diff
@@ -0,0 +1,258 @@
...
+            HashSet::new()
+        },
+    }
+}
+
+fn effects_of_intrinsic(intr: &sway_ast::Intrinsic) -> HashSet<Effect> {
+    use sway_ast::Intrinsic::*;
+    match intr {
+        StateStoreWord | StateStoreQuad => HashSet::from([Effect::StorageWrite]),
+        // TODO: figure out the effect of __revert
```
mohammadfawaz, line 232: I think a revert (or a return) between a contract call and a storage write should not result in a reentrancy warning. That being said, we should probably be warning on dead code in that case anyways?

sway-core/src/semantic_analysis/cei_pattern_analysis.rs
```diff
@@ -0,0 +1,258 @@
...
+            effs
+        },
+        Tuple { fields: exprs }
+        | Array { contents: exprs } => {
+            effects_of_expressions(exprs)
+        },
+        StructExpression {fields, ..} => {
+            effects_of_struct_expressions(fields)
+        },
+        CodeBlock(codeblock) => {
```
mohammadfawaz, line 158: Given this, I'm assuming that if I have something like
```rust
call_a_contract();
if condition {
    write_to_state()
}
```
then I will get a warning? This is good.

sway-core/src/semantic_analysis/cei_pattern_analysis.rs
```diff
@@ -0,0 +1,258 @@
...
+            effs
+        },
+        // TODO: process assembly blocks
+        AsmExpression {
+            registers: _,
+            body: _,
+            whole_block_span: _,
+            ..
+        } => {
+            // temporary solution, will remove after this is finished
```
mohammadfawaz, line 222: The quickest way to deal with this is to:
* Check the expressions initializing each register argument.
* Look for `sww` or `swwq` in the body of the `asm` block.

test/src/e2e_vm_tests/test_programs/should_pass/static_analysis/cei_pattern_violation/Forc.toml
```diff
@@ -0,0 +1,6 @@
+[project]
+name = "cei_pattern_violation"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+implicit-std = true
```
mohammadfawaz, line 6: When `implicit-std` is set to `true` (which is also the default btw), the standard library is pulled from Github (from the latest release of `forc`), which we don't want in our tests. We always want to be using the version from `master` for compatibility reasons. So, you have two options:
* Set `implicit-std` to `false` if you _don't_ need the standard library in your test.
* Remove this field and add a path `std` dependency to where the standard library lives.

test/src/e2e_vm_tests/test_programs/should_pass/static_analysis/cei_pattern_violation/src/main.sw
```diff
@@ -0,0 +1,34 @@
+contract;
```
mohammadfawaz, line 1: This test could use some formatting :)

test/src/e2e_vm_tests/test_programs/should_pass/static_analysis/cei_pattern_violation/src/main.sw
```diff
@@ -0,0 +1,34 @@
...
+impl TestAbi for Contract {
+  #[storage(write)]
+  fn deposit(gas: u64,  amount: u64, color: b256, unused: ()) {
+    let other_contract = abi(TestAbi, 0x3dba0a4455b598b7655a7fb430883d96c9527ef275b49739e7b0ad12f8280eae);
+
+    // interaction
+    other_contract.deposit(gas, amount, color, unused);
+    // [trait/abi method] effect -- therefore violation of CEI where effect should go before interaction
+    let storage_key = 0x3dba0a4455b598b7655a7fb430883d96c9527ef275b49739e7b0ad12f8280eae;
+    store(storage_key, ());
```
mohammadfawaz, line 32: I am assuming you're going to be adding more tests here. Things I'm curious to see:
* Tests using `storage.var = ...`
* Tests using `asm` blocks directly (when that's supported)
* Tests using complex control flow
* Tests having multiple contract calls or multiple storage writes in various scenarios. Do we emit multiple warnings in this case?

mohammadfawaz (commented):
Some initial thoughts! But this looks excellent! Thank you so much 🚀

sway-core/src/semantic_analysis/cei_pattern_analysis.rs
```diff
@@ -0,0 +1,258 @@
+// CEI stands for "Checks, Effects, Interactions".  We check that no storage writes
```
mohammadfawaz, line 1: It turns out that we also want to check for storage reads here which include:
* Storage read intrinsics.
* `StorageAccess` expressions.
* `asm` blocks using `srw` or srwq`.
* `asm` blocks using [`bal`](https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/instruction_set.md#bal-balance-of-contract-id) because `bal` actually reads from storage.

sway-core/src/semantic_analysis/cei_pattern_analysis.rs
anton-trunov: Yes, you'll get a warning:

```
warning
  --> /Users/anton/fuel/sway/test/src/e2e_vm_tests/test_programs/should_pass/static_analysis/cei_pattern_violation_in_if_statement/src/main.sw:16:5
   |
14 |
15 |       // interaction
16 |       other_contract.deposit(amount);
   |  _____-
17 | |
18 | |     if amount == 42 {
19 | |         // effect -- therefore violation of CEI where effect should go before interaction
20 | |         let storage_key = 0x3dba0a4455b598b7655a7fb430883d96c9527ef275b49739e7b0ad12f8280eae;
21 | |         store(storage_key, ());
22 | |     }
   | |_____- Storage modification after external contract interaction in function or method "deposit". Consider making all storage writes before calling another contract
23 |     }
24 |   }
   |
____
```
The warning could be more precise actually and point at the method call / storage field assignment / asm instruction inside the `if` codeblock. This could be either improved in this PR or as a separate PR.

test/src/e2e_vm_tests/test_programs/should_pass/static_analysis/cei_pattern_violation/src/main.sw
anton-trunov: We emit multiple warnings if there are multiple methods with CEI pattern violations, just like in test/src/e2e_vm_tests/test_programs/should_pass/static_analysis/cei_pattern_violation_english_auction/src/main.sw.

But for one method we report only the first violation, otherwise it can be very verbose, basically reporting each storage modifying statement after a contract call.

sway-core/src/semantic_analysis/cei_pattern_analysis.rs
anton-trunov: Great, I'll add these checks as a separate type of warning, so the programmer could, in principle, silence it independently of the modification-after-call warnings.

mohammadfawaz:
I'm very happy with the overall changes. My only nitpick is that the auction test is _too_ large and complex to belong in the E2E test repo. Basically, the test is hard to understand and it is unclear where the violation is. I suggest we go with a much smaller example that showcases the same violation.

anton-trunov:
@mohammadfawaz Thanks for the feedback. I simplified the English auction contract: renamed the test and removed a lot of code but preserved the warnings pattern

anton-trunov:
Just realized I need to treat nested code blocks. The fix is on the way

anton-trunov:
@mohammadfawaz Pushed an updated, deeper, version of the analysis, please have a look. It still needs a bit more testing, but it already reproduces all the existing tests and I added a very simple test with a nested code block which was not processed previously.

sway-core/src/semantic_analysis/cei_pattern_analysis.rs
```diff
@@ -0,0 +1,601 @@
...
+                warnings,
+            ),
+        };
+        accumulated_effects.extend(expr_effs)
+    }
+    accumulated_effects
+}
+
+// TODO: jumps are not processed at the moment, so the CEI analysis will work only
+// for simple linear assembly blocks, which arguably is the most important use case so far
```
mohammadfawaz, line 371: Jump are not and will porbably never be allowed in `asm` blocks, so you're good!
```suggestion
No need to worry about jumps because they are not allowed in `asm` blocks.
```

mohammadfawaz (dismissed):
Generally this looks very reasonable. I'd certainly like to see more testing. At the very least, it would be nice to have full code coverage for `cei_pattern_analysis.rs`, but that can be done in a separate PR.

mohammadfawaz (approved):

otrho (commented):
Would it be possible to add a test which uses effects (`ref mut` args?) to confirm the left-to-right evaluation of function args and struct initialisers?

sezna (approved):
Looks great! The only thing I wonder is if someday down the line we can merge this and other AST-traversing checks we do into one traversal. But the traversals themselves are slightly different so it may not be generalizable (thinking of purity checking mainly).

sezna:
> Would it be possible to add a test which uses effects (ref mut args?) to confirm the left-to-right evaluation of function args and struct initialisers?

I'd be okay with that as a follow-up PR since this one has the approvals needed. Your call @anton-trunov

mohammadfawaz:
> > Would it be possible to add a test which uses effects (ref mut args?) to confirm the left-to-right evaluation of function args and struct initialisers?
> 
> I'd be okay with that as a follow-up PR since this one has the approvals needed. Your call @anton-trunov


Actually @anton-trunov mentioned this to me recently while chatting. He thinks he has idea into how to merge purity checking into this check as he's already collecting storage accesses here!

mohammadfawaz:
> Looks great! The only thing I wonder is if someday down the line we can merge this and other AST-traversing checks we do into one traversal. But the traversals themselves are slightly different so it may not be generalizable (thinking of purity checking mainly).

Actually @anton-trunov mentioned this to me recently while chatting. He thinks he has idea into how to merge purity checking into this check as he's already collecting storag

mohammadfawaz:
@anton-trunov I've opened https://github.com/FuelLabs/sway/issues/3342 and assigned it to you so we can follow up with additional testing.

anton-trunov:
Awesome! Thanks everyone!

anton-trunov:
@otrho Great suggestion to test left-to-right evaluation on which the CEI analysis relies! In PR #3356, I added a compiler test for that: https://github.com/FuelLabs/sway/pull/3356/commits/332a240dcf3ad6706818778fe89c06739820e2e2

PR #3178, opened 2022-10-27
Initialize tracing based on user setting

sdankel:
- initializes trace _after_ the server has started and the user's settings from the IDE have been written to `config`
- uses the log level from the config to initialize tracing

Tested this change manually and confirmed that the user's setting is passed to the server and that the correct logs are shown in the output console.

Related https://github.com/FuelLabs/sway/pull/2925 
plugin PR: https://github.com/FuelLabs/sway-vscode-plugin/pull/100

### DIFF
--- a/sway-lsp/src/core/config.rs
+++ b/sway-lsp/src/core/config.rs
@@ -39,6 +39,7 @@ impl Default for LoggingConfig {
 #[derive(Deserialize, Serialize, Clone)]
 #[serde(rename_all = "lowercase")]
 #[serde(remote = "LevelFilter")]
+#[allow(clippy::upper_case_acronyms)]
 enum LevelFilterDef {
     OFF,
     ERROR,
--- a/sway-lsp/src/core/config.rs
+++ b/sway-lsp/src/core/config.rs
@@ -1,22 +1,51 @@
 use serde::{Deserialize, Serialize};
+use tracing::metadata::LevelFilter;
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Default)]
 #[serde(rename_all = "camelCase")]
 pub struct Config {
-    pub debug: Debug,
+    pub debug: DebugConfig,
+    pub logging: LoggingConfig,
     pub inlay_hints: InlayHintsConfig,
     #[serde(skip_serializing)]
-    trace: Trace,
+    trace: TraceConfig,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize, Default)]
-struct Trace {}
+struct TraceConfig {}
 
 // Options for debugging various parts of the server
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct Debug {
+pub struct DebugConfig {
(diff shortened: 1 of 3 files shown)


sway-lsp/src/core/config.rs
```diff
@@ -1,24 +1,54 @@
...
+    }
+}
+
+// This allows us to deserialize the enum that is defined in another crate.
+#[derive(Deserialize, Serialize, Clone)]
+#[serde(rename_all = "lowercase")]
+#[serde(remote = "LevelFilter")]
+#[allow(clippy::upper_case_acronyms)]
+enum LevelFilterDef {
+    OFF,
```
JoshuaBatty, line 44: Is there a reason these variants all need to be uppercase?

sway-lsp/src/core/config.rs
```diff
@@ -1,24 +1,54 @@
...
+    fn default() -> Self {
+        Self {
+            level: LevelFilter::OFF,
+        }
+    }
+}
+
+// This allows us to deserialize the enum that is defined in another crate.
+#[derive(Deserialize, Serialize, Clone)]
+#[serde(rename_all = "lowercase")]
```
JoshuaBatty, line 40: Perhaps this could be 
```suggestion
#[serde(rename_all = "PascalCase")]
```

sway-lsp/src/core/config.rs
```diff
@@ -1,24 +1,54 @@
...
+#[serde(remote = "LevelFilter")]
+#[allow(clippy::upper_case_acronyms)]
```
JoshuaBatty, line 42: We should be able to remove this line if the above change is made

sway-lsp/src/core/config.rs
JoshuaBatty: We should be able to have these `Off` `Error` `Warn` etc.. with the above change

JoshuaBatty (changes requested):
Nice, thanks! Just left some suggestions on the `LevelFilterDef ` enum.

sway-lsp/src/core/config.rs
sdankel: They're uppercase because what I'm doing here is deserializing the strings into the LevelFilter enum which is defined in another crate. The `LevelFilterDef` has to exactly match the definition of `LevelFilter` in order for `#[serde(remote = "LevelFilter")]` to work

https://serde.rs/remote-derive.html

sway-lsp/src/core/config.rs
sdankel: It shouldn't make a difference since the options are all single words

sway-lsp/src/core/config.rs
JoshuaBatty: agh I see, that makes sense. This is all good then :)

JoshuaBatty (approved):
Nice, looks good. Thanks!

sway-lsp/src/core/config.rs
mitchmindtree: TIL about `#[serde(remote = "Foo")]` - noice.

mitchmindtree (approved):
:ok_hand:

Issue #3159, opened 2022-10-26
Support the updated versions of `sww`, `swwq`, `srw`, and `srwq` in the compiler and the standard library


mohammadfawaz:
Ref:
https://github.com/FuelLabs/fuel-specs/pull/414
https://github.com/FuelLabs/fuel-specs/pull/417
https://github.com/FuelLabs/fuel-specs/pull/422

1. The first step is support the new opcodes and restore parity with the current behavior.
2. The second step is to think through how to best use the new status register that is being set. The status registers tells us whether a storage slot has been written before.

mohammadfawaz:
We can't work on this before all the required changes in the VM are made and published.

Issue #2115, opened 2022-06-25
Move the dependency on `fuel-tx` in `forc-pkg` to `fuel-types`


mohammadfawaz:
With https://github.com/FuelLabs/sway/pull/2078, `forc-pkg` now depends on `fuel-tx` just for the type `StorageSlot`. It is better to move that dependency to `fuel-types` instead, so `fuel-types` must first have `StorageSlot` as a type.

mohammadfawaz:
Blocked by https://github.com/FuelLabs/fuel-tx/issues/152

PR #3181, opened 2022-10-28
fuel-core v0.13 upgrade

Voxelot:
- $rB on state opcodes is now used as a flag for whether the storage slot was previously set
- $rD on quad word state opcodes is reserved for use as accessing a range of slots (currently hardcoded to 1)

currently blocked on fuel-rs upgrade

Closes https://github.com/FuelLabs/sway/issues/3159
Closes https://github.com/FuelLabs/sway/issues/2115 (the dependency is on `sway-core` now)

Had to disable the u128 sqrt tests for now as they are too expensive / slow, as per @mohammadfawaz suggestion.


sway-core/src/asm_generation/asm_builder.rs
```diff
@@ -1484,10 +1483,23 @@ impl<'ir> AsmBuilder<'ir> {
             _ => unreachable!("Unexpected storage locations for key and val"),
         };
 
+        // capture the status of whether the slot was set before the calling this instruction
```
mohammadfawaz, line 1486:
```suggestion
        // capture the status of whether the slot was set before calling this instruction
```

sway-core/src/asm_generation/asm_builder.rs
```diff
@@ -1514,6 +1526,10 @@ impl<'ir> AsmBuilder<'ir> {
 
         let load_reg = self.reg_seqr.next();
         let owning_span = self.md_mgr.val_to_span(self.context, *instr_val);
+
+        // capture the status of whether the slot was set before the calling this instruction
```
mohammadfawaz, line 1530:
```suggestion
        // capture the status of whether the slot was set before calling this instruction
```

sway-core/src/asm_generation/asm_builder.rs
```diff
@@ -1560,6 +1580,9 @@ impl<'ir> AsmBuilder<'ir> {
         }
         let (key_ptr, ptr_ty, offset) = key_ptr.value.unwrap();
 
+        // capture the status of whether the slot was set before the calling this instruction
```
mohammadfawaz, line 1583:
```suggestion
        // capture the status of whether the slot was set before calling this instruction
```

mohammadfawaz (commented):
This is amazing, thank you @Voxelot. Everything here seems very reasonable to me.

I suppose once we figure out how to use `$rB`, we have to expose it in IR as an output to `state_read_*` and `state_write_*` instructions.

sway-core/src/asm_lang/virtual_ops.rs
```diff
@@ -403,10 +413,10 @@ impl VirtualOp {
             LOGD(_r1, _r2, _r3, _r4) => vec![],
             MINT(_r1) => vec![],
             RVRT(_r1) => vec![],
-            SRW(r1, _r2) => vec![r1],
-            SRWQ(_r1, _r2) => vec![],
-            SWW(_r1, _r2) => vec![],
-            SWWQ(_r1, _r2) => vec![],
+            SRW(r1, r2, _r3) => vec![r1, r2],
```
Voxelot, line 416: The comment above this method states:

> /// Returns a list of all registers *written* by instruction `self`. All of our opcodes define
> /// exactly 0 or 1 register, so the size of this returned vector should always be at most 1.

Now that SRW writes to rA and rB, will this break any assumptions in the compiler?

mohammadfawaz: I think we should be fine. We iterate over the result of `def_registers()` everywhere. I don't believe there are any baked assumptions, so this comment can probably be updated.

mohammadfawaz:
As we were just talking about the `fuel-core` in CI, I remembered that https://github.com/FuelLabs/sway/blob/master/.github/workflows/ci.yml#L244 and https://github.com/FuelLabs/sway/blob/51987a96a6f0ac23f4102478290e0bc1736b2af6/.github/workflows/ci.yml#L265 should be updated.

mohammadfawaz (approved):

mohammadfawaz (dismissed):

mohammadfawaz (approved):

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u128_root_test/src/main.sw
```diff
@@ -4,26 +4,26 @@ use std::u128::*;
...
-    assert(root_of_u_128 == U128::from(0, 9));
-
-    u_128 = U128::from(0, 144);
-    root_of_u_128 = u_128.sqrt();
-    assert(root_of_u_128 == U128::from(0, 12));
-
-    u_128 = U128::from(0, 1);
-    root_of_u_128 = u_128.sqrt();
-    assert(root_of_u_128 == U128::from(0, 1));
+    //let mut u_128: U128 = U128::from(0, 49);
```
JoshuaBatty, line 7: is this meant to be commented out for this PR?

mohammadfawaz: Yeah. The operation is now too expensive with the new gas schedule and the test was running out of gas. The implementation should probably be revamped. It's a new feature anyways (like 2 days ago).

JoshuaBatty (approved):

mohammadfawaz (approved):

Issue #2944, opened 2022-10-05
Introduce an `inline` annotation


mohammadfawaz:
Expected behavior:

1. None. The compiler will decide itself if the function should be inlined. This will depend on the optimization level, the size of the function.
2. `#[inline(always)]`. This strongly suggests that the function should be inlined.
3. `#[inline(never)]`. This strongly suggests that the function should not be inlined.

I use "strongly suggests" because it is possible that inlining will fail for all kinds of reasons (it is an optimization after all) and when that happens, we shouldn't really error out.

This annotation is probably for advanced users, but I expect it to be quite useful for library authors (especially in the standard library).

mohammadfawaz:
This is done: https://github.com/FuelLabs/sway/pull/3185

PR #3185, opened 2022-10-28
Adds inline annotation to the compiler

esdrubal:
`inline_function_calls`' `inline_heuristic` now checks inline metadata and
if inline metadata is set to `Inline::Never` it returns false and the
function is not inlined.

When inline metadata is set to `Inline::Always` it will still follow the
default behavior, I left as TODO: check if inlining of function is
possible.

ref #2944


esdrubal:
I tried to write a test into test/ir_generation but it is compiling IR without running optimizations should we create an issue to add support for running optimizations as in [sway-core/src/lib.rs#L394](https://github.com/FuelLabs/sway/blob/master/sway-core/src/lib.rs#L394)?

mohammadfawaz:
> I tried to write a test into test/ir_generation but it is compiling IR without running optimizations should we create an issue to add support for running optimizations as in [sway-core/src/lib.rs#L394](https://github.com/FuelLabs/sway/blob/master/sway-core/src/lib.rs#L394)?

Thank you for this PR! You may want to look at https://github.com/FuelLabs/sway/tree/master/sway-ir/tests/inline :)

mohammadfawaz:
This generally seems reasonable to me. Thank you so much for your contribution 😄 

I'd like @otrho to review this and give his approval as he's our resident IR expert.

otrho:
Thanks @esdrubal for looking at this.  There's a couple  of issues I have with the design though...

I'm not sure that adding the `inline` member to each of those data structures, essentially pairing up with `purity`, is the best long term approach.  If we were to do this for every annotation we add in the future then those data structures (most of which don't really care about the values, they're just passing them on) will bloat out a fair bit.  I think it'd be better to generalise annotations into a map or something which can be queried when needed.  I'm assuming that would work OK, and then be general enough for further annotations.  But I'm curious what @sezna and @mitchmindtree think.

Also, yep, ideally we'd have fine grained control of which opt passes are run per test, though at the same time I think those tests are more about IR generation and not full compilation, and so they shouldn't have optimisations applied.

But it's tricky, since the unit tests under `sway-ir/tests` are better suited to testing the actual inlining except they're not supposed to really understand metadata, so confirming that the metadata works is still best suited for the compiler tests.

Hrmm, sorry, I'm not really helping here.  So I guess we need to add inlining to the `ir_generation` tests, but the `::optimisation-inline::` tag isn't right -- the other two are designed to be boundary markers for FileCheck.  So maybe, like the first line of the `sway-ir/tests/inline` files, we should have parameters describing which opt passes should be applied.

mohammadfawaz:
> I think it'd be better to generalise annotations into a map or something which can be queried when needed.

That is certainly a better long term approach. A bit like the generalized metadata map that you created.

mitchmindtree:
>  I think it'd be better to generalise annotations into a map or something which can be queried when needed. I'm assuming that would work OK, and then be general enough for further annotations. But I'm curious what @sezna and @mitchmindtree think.

Yeah, the `TyFunctionDeclaration` already has an `attributes` map. By introducing an `inline` field we end up storing the same information in two locations.

Perhaps we're better off exposing a method on `TyFunctionDeclaration` that acts as a short-hand for checking the attribute map like:
```rust
    pub fn inline(&self) -> Option<Inline>
```

This would be similar to how the `TyFunctionDeclaration::is_test` method works (it just checks the `attributes` map for a `#[test]` attribute).

By returning an `Option`, we can also remove the `Inline::Default` variant in favour of handling `None`.

esdrubal:
Thank you all for the feedback.

This new version implements an `inline` method in `TyFunctionDeclaration` instead of adding inline field to it. This method uses `TyFunctionDeclaration`'s `attributes`.

The tests no longer check for `::optimisation-inline::` instead they check if there is a comment in the first line with `optimisation-inline`.

esdrubal:
@Centril Thank you for the requested changes, looks much better now.
I applied all the changes except moving module inline into an existing module. Not sure where to put it.

PR #3186, opened 2022-10-28
Use a call-graph in the inliner

vaivaswatha:
This reduces the time spent in the inliner, in compiling the [`amm` library ](https://github.com/sway-libs/amm) from ~120s to ~5s.

Code sizes before/after on [sway-applications](https://github.com/FuelLabs/sway-applications):
-----------------------------------
benchmark | before | after
----------------|----------|--------
dao-voting | 20568 | 20496
escrow | 72740 | 72668
fundraiser | 31344 | 31440
multisig-wallet | 6592 | 7000
NFT | 23200 | 23200
oracle | 1924 | 1924

While this change doesn't touch the inlining heuristics, but only the order in which inlining of functions is undertaken, we have a heuristic that checks the size of the function being inlined. **That** depends on the order in which we inline, and hence the final set of inlining decisions aren't identical (though largely similar). Overall, I think the compile time benefits outweigh any differences in the inlining decisions.


mohammadfawaz:
from ~120s to ~5s. 😍😍😍😍😍😍

sezna (approved):
I guess the bytecode changes are just in the ordering of the function bodies in the bytecode, then?

Awesome PR, love to see the performance boost!

vaivaswatha:
> I guess the bytecode changes are just in the ordering of the function bodies in the bytecode, then?
> 
> Awesome PR, love to see the performance boost!

Suppose A calls B and B calls C. We have a limit on how big a function we can inline.
So if I inline C into B first, B may now be too big to be inlined into A. But if i inline B into A first, I can still inline C into A after that.

What this PR does is to introduce a "callee first" order of traversing the program for inlining. This ensures that you don't have to inline the same call site multiple times. But because the order of visiting call sites  has changed (compared to before this PR), there may be some differences in the final inlining decisions.

emilyaherbert:
Omg amazing work @vaivaswatha! ~120s to ~5s 🚀

PR #3193, opened 2022-10-28
Fixes parser and printer of metadata strings

esdrubal:
Metadata strings are printed with '\\' escaped as "\\\\" when we read them back
again we need to unescape "\\\\" to "\\".

Fixes the following issue while running test/src/ir_generation tests in Windows:
windows_path_error


esdrubal:
snailquote is not reliable for this use case, it single quotes strings with spaces.
We need to find an alternative to escape, unescape.

otrho:
We don't officially support Windows outside of WSL yet, so this isn't a huge priority.  We definitely want to avoid putting actual paths into the tests repo, hopefully there aren't any in there at the moment.

But eventually we'll want to make sure metadata paths for spans work on Windows, but they don't need to be portable, as in IR files created on Windows don't need to be parseable on Unix and vice-versa.  But yeah, if they don't even parse with `\` on Windows then that'll be a problem when we do officially support Windows paths.

esdrubal:
@otrho I got this error while running the tests on windows without WSL. But this error is not Windows specific the problem occurs whenever it prints metadata with '\\\\' and parse it again. For instance if a metadata string contains '\n' it will print '\\\\n' and on parse it will set the metadata string with a value that differs from the original.

The code that prints metadata strings is: [sway-ir/src/printer.rs#L983](https://github.com/FuelLabs/sway/blob/1d0b1afe4375b9e790480a508b2d9e6c68e45471/sway-ir/src/printer.rs#L983)

otrho:
OK, fair enough, if it's as simple as your new version that's great.  Thanks.  Do you mind adding a test to `sway/sway-ir/tests/serialize` to exercise it please?

otrho (approved):

PR #3203, opened 2022-10-30
Do not reuse registers for insert/extract element, they may be live

### LINKED ISSUE #3188: `mem2reg` causes a functional failure for a simple vector test
Contract and test below. I figured out that `mem2reg` is the culprit by testing the contract below with and without https://github.com/FuelLabs/sway/pull/3128.
* At commit de60a8318cd8c6a69c326e4f125037dc1f80e14c, the test passes (no revert in the list of receipts)
* At commit 74e6377c500014affc45277dd2284491d0afe5d2 (next commit after the above), the test fails (we get a revert in the list of receipts)

### Contract:
```rust
contract;

abi MyContract {
    fn array_in_vec(arg: Vec<[u64; 2]>);
}

impl core::ops::Eq for [u64; 2] {
        let mut i = 0;
        while i < 2 {
            if self[i] != other[i] {
                return false;
            }
            i += 1;
        }
        true
    }
}

impl core::ops::Eq for Vec<[u64; 2]> {
        if self.len() != other.len() {
            return false;
        }
        let mut i = 0;
        while i < self.len() {
            if self.get(i).unwrap() != other.get(i).unwrap() {
                return false;
            }
            i += 1;
        }
        true
    }
}

    fn array_in_vec(arg: Vec<[u64; 2]>) {
        let mut expected = ~Vec::new();
        expected.push([0, 1]);
        expected.push([0, 1]);

        assert(arg == expected);
    }
}
```

### Rust test (fuels 0.27):
```rust
use fuels::{prelude::*, tx::ContractId};

// Load abi from json
abigen!(MyContract, "out/debug/broken-vec-abi.json");

async fn get_contract_instance() -> (MyContract, ContractId) {
    // Launch a local network and deploy the contract
    let mut wallets = launch_custom_provider_and_get_wallets(
        WalletsConfig::new(
            Some(1),             /* Single wallet */
            Some(1),             /* Single coin (UTXO) */
            Some(1_000_000_000), /* Amount per coin */
        ),
        None,
    )
    .await;
    let wallet = wallets.pop().unwrap();

    let id = Contract::deploy(
        "./out/debug/broken-vec.bin",
        &wallet,
        TxParameters::default(),
        StorageConfiguration::with_storage_path(Some(
            "./out/debug/broken-vec-storage_slots.json".to_string(),
        )),
    )
    .await
    .unwrap();

    let instance = MyContract::new(id.clone(), wallet);

    (instance, id.into())
}

#[tokio::test]
async fn can_get_contract_id() {
    let (instance, _id) = get_contract_instance().await;

    let arg = vec![[0u64, 1u64], [0u64, 1u64]];
    dbg!(instance.methods().array_in_vec(arg.clone()).call().await);
}
```
 Compile with `cargo test -- --nocapture` to get:
```console
[tests/harness.rs:40] instance.methods().array_in_vec(arg.clone()).call().await = Err(
    RevertTransactionError(
        "Revert(0)",
        [
            Call {
                id: 0x0000000000000000000000000000000000000000000000000000000000000000,
                to: 0x4613d583fd17f43e8b63fec2c4bbcfc77f321b67e164830946b0da127a63a6f8,
                amount: 0,
                asset_id: 0x0000000000000000000000000000000000000000000000000000000000000000,
                gas: 1000000,
                param1: 3149283750,
                param2: 10472,
                pc: 11688,
                is: 11688,
            },
            Revert {
                id: 0x4613d583fd17f43e8b63fec2c4bbcfc77f321b67e164830946b0da127a63a6f8,
                ra: 0,
                pc: 12764,
                is: 11688,
            },
            ScriptResult {
                result: Revert,
                gas_used: 5521,
            },
        ],
    ),
)
tes
```

vaivaswatha:
Previously, values would be reloaded from memory, thus even if it was live and clobbered, it wouldn't matter because it would be reloaded. But now (with `mem2reg`, expectedly), a live register cannot be clobbered.

Closes #3188

### DIFF
--- a/sway-core/src/asm_generation/asm_builder.rs
+++ b/sway-core/src/asm_generation/asm_builder.rs
@@ -607,16 +607,7 @@ impl<'ir> AsmBuilder<'ir> {
 
         // Index value is the array element index, not byte nor word offset.
         let index_reg = self.value_to_register(index_val);
-        let rel_offset_reg = match index_reg {
-            VirtualRegister::Virtual(_) => {
-                // We can reuse the register.
-                index_reg.clone()
-            }
-            VirtualRegister::Constant(_) => {
-                // We have a constant register, cannot reuse it.
-                self.reg_seqr.next()
-            }
-        };
+        let rel_offset_reg = self.reg_seqr.next();
 
         // We could put the OOB check here, though I'm now thinking it would be too wasteful.
         // See compile_bounds_assertion() in expression/array.rs (or look in Git history).
@@ -915,16 +906,7 @@ impl<'ir> AsmBuilder<'ir> {
 
         // Index value is the array element index, not byte nor word offset.
         let index_reg = self.value_to_register(index_val);
-        let rel_offset_reg = match index_reg {
-            VirtualRegister::Virtual(_) => {
-                // We can reuse the register.
-                index_reg.clone()
-            }
-            VirtualRegister::Constant(_) => {
-                // We have a constant register, cannot reuse it.
-                self.reg_seqr.next()
-            }
-        };
+        let rel_offset_reg = self.reg_seqr.next();
 
         let owning_span = self.md_mgr.val_to_span(self.context, *instr_val);
 
--- a/test/src/e2e_vm_tests/test_programs/should_pass/language/insert_element_reg_reuse/.gitignore
+++ b/test/src/e2e_vm_tests/test_programs/should_pass/language/insert_element_reg_reuse/.gitignore
@@ -1,0 +1,2 @@
+out
+target
--- a/test/src/e2e_vm_tests/test_programs/should_pass/language/insert_element_reg_reuse/Forc.lock
+++ b/test/src/e2e_vm_tests/test_programs/should_pass/language/insert_element_reg_reuse/Forc.lock
@@ -1,0 +1,13 @@
+[[package]]
+name = 'core'
+source = 'path+from-root-8E2171C2F6D30B77'
+
+[[package]]
+name = 'insert_element_reg_reuse'
+source = 'root'
+dependencies = ['std']
+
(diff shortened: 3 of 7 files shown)


sway-core/src/asm_generation/asm_builder.rs
```diff
@@ -607,16 +607,7 @@ impl<'ir> AsmBuilder<'ir> {
...
-            VirtualRegister::Virtual(_) => {
-                // We can reuse the register.
-                index_reg.clone()
-            }
-            VirtualRegister::Constant(_) => {
-                // We have a constant register, cannot reuse it.
-                self.reg_seqr.next()
-            }
-        };
+        let rel_offset_reg = self.reg_seqr.next();
```
otrho, line 610: Oh, dear, yeah, it's almost surprising this problem didn't come up earlier.  I plan to refactor this so that `VirtualRegisters` can't be re-used and will behave a bit like SSA, so that we can do peephole opts without worrying about liveness.

mohammadfawaz: If we do this, then we may be able to simplify the register allocator a bit and make it faster. Liveness analysis currently **does not** assume SSA.

vaivaswatha: Just to confirm, neither of you are suggesting any changes to this PR right?

This PR is ready, except that I haven't added the failing test to our testsuite. Once I'll do that I'll mark it as ready.

mohammadfawaz: Yeah we just like to talk clearly 🤣

mohammadfawaz (approved):
Awesome, thank you for the quick fix!

otrho (approved):

PR #3212, opened 2022-10-31
LSP: Collect declaration tokens from std lib prelude and dependencies

### LINKED ISSUE #3208: LSP: Collect tokens from the Sway standard library prelude
Currently, we aren't collecting tokens from the standard library prelude. This means we can't perform hover requests for example, on items imported from the prelude. 

We should traverse and collect these tokens once on initialization.

JoshuaBatty:
We weren't collecting tokens imported by default with the new std lib prelude. This PR now collects these tokens so we can do things like show documentation for imported types on hover requests. 

partially solves #324 
closes #3208


mohammadfawaz (approved):

sway-lsp/src/core/session.rs
```diff
@@ -268,6 +266,27 @@ impl Session {
...
+            for (ident, decl) in symbols {
+                // If an ident is already in our map, skip this part as the
+                // tokens from std and core have already been collected.
+                // We only want to collect these tokens once for efficiency.
+                if self.token_map.contains_key(&to_ident_key(ident)) {
+                    break 'outer;
+                }
+                collect_symbol_map::handle_declaration(ident, decl, &self.token_map);
+            }
+        }
```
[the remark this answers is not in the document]
JoshuaBatty: nice suggestion, thanks!

mohammadfawaz (approved):

sway-lsp/src/core/dependancy.rs
```diff
@@ -0,0 +1,68 @@
+use crate::{
```
sdankel, line 1: nit: rename to dependency.rs

sway-lsp/src/test_utils.rs
```diff
@@ -13,7 +13,7 @@ pub(crate) fn e2e_language_dir() -> PathBuf {
 pub(crate) fn e2e_test_dir() -> PathBuf {
     sway_workspace_dir()
         .join(e2e_language_dir())
-        .join("struct_field_access")
+        .join("prelude_access2")
```
sdankel, line 16: Was this failing on the old e2e test dir, or is this just ensuring it's testing with preludes?

sdankel (approved):

forc/src/ops/forc_check.rs
```diff
@@ -20,5 +20,10 @@ pub fn check(command: CheckCommand) -> Result<CompileResult<ty::TyProgram>> {
     let manifest = PackageManifestFile::from_dir(&this_dir)?;
     let plan = pkg::BuildPlan::from_lock_and_manifest(&manifest, locked, offline)?;
 
-    Ok(pkg::check(&plan, terse_mode)?.flat_map(|(_, tp)| CompileResult::new(tp, vec![], vec![])))
+    let mut v = pkg::check(&plan, terse_mode)?;
+    let res = v
+        .pop()
```
sdankel, line 25: I wonder if there should be an optional argument to `pkg::check` that asserts there's at least one element in the result for you, since we check this in a few places.

sway-lsp/src/core/dependancy.rs
JoshuaBatty: good catch!

sway-lsp/src/test_utils.rs
JoshuaBatty: it was just for testing preludes locally. I'll change it back so it's less noise for the PR

forc/src/ops/forc_check.rs
JoshuaBatty: I think it's fine to keep it like this as we are already returning an error [here](https://github.com/FuelLabs/sway/pull/3212/files#diff-f398539dbba9474754d61669ea72772f8de39c155c10a613069b7f2b0e0158a4R2558) if there are no elements to return.

mohammadfawaz (approved):

sdankel (approved):

PR #3214, opened 2022-10-31
Adds error for doc comment before Dependency.

### LINKED ISSUE #3116: Docstring above `dep` keyword results in error
When adding a docstring a line above a `dep` keyword the compiler will result in an error. It should either not result in an error or inform the user that this is something that is not allowed.

For example, the following results in the error shown:

```
contract;

/// This is a doc
dep lib;
```

docstring_error

esdrubal:
Before this fix the error message for doc comments before was out of place and was "Expected an item."

This was fixed by 'pre-parsing' each `Dependency` and checking if there is a `DocComment` before. The 'pre-parsing' stores a clone of `parser.token_trees` and restores it later to avoid consuming any `DocComment` that is consumed later while parsing `Annotated<ItemKind>`.

Closes #3116

### DIFF
--- a/sway-error/src/parser_error.rs
+++ b/sway-error/src/parser_error.rs
@@ -8,6 +8,8 @@ pub enum ParseErrorKind {
     ExpectedImportNameGroupOrGlob,
     #[error("Expected an item.")]
     ExpectedAnItem,
+    #[error("Cannot doc comment a dependency.")]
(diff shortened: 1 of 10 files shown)


esdrubal:
parser.token_trees clone was only cloning a reference. And this feel too hacky. Closing this for now. If someone knows a better way of doing it please let me know.

Issue #1436, opened 2022-04-29
Eliminate the need for `~` to access associated functions


mohammadfawaz:
It seems that the following doesn't work today:
```rust
script;

trait A {
    fn get_zero() -> Self;
}

impl A for u64 {
    fn get_zero() -> Self {
        0
    }
}

fn main() {
    let zero_val:u64 = u64::get_zero();
}
```
I think we always expect trait interfaces to have methods that take `self` as the first argument. The error we get when calling `u64::get_zero()` is
```
   |                         ^^^^^^^^ Could not find symbol "get_zero" in this scope.
```
We're probably looking for a free function called `get_zero` but can't find any.

Relevant: https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#fully-qualified-syntax-for-disambiguation-calling-methods-with-the-same-name

Edit: 
The above doesn't work because a `~` is required before the type. Changing this issue to track the removal of the need for `~`.

adlerjohn:
```rust
fn main() {
    let zero_val:u64 = A::get_zero();
}
```

shouldn't that be

```rust
fn main() {
    let zero_val:u64 = u64::get_zero();
}
```

mohammadfawaz:
Interestingly, both work in Rust 🤔  but neither work in Sway.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dc62ee143500c37e3f417ad9617df1ef
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d1c8446b363a457a71a8f7ba9d16b0ec

mohammadfawaz:
The reason why this is important to implement is because the trait-based approach for storage requires associated functions to implement the `read` or `get` function:

An example starting point:
```rust
pub trait Storable {
    fn write(self, key: b256);
    fn read(key: b256) -> Self;
}

impl Storable for u64 {
    fn write(self, key: b256) {
        asm(r1: key, r2: self) {
            sww r1 r2;
        };
    }
    fn read(key: b256) -> Self {
        asm(r1: key, r2) {
            srw r2 r1;
            r2: u64
        }
    }
}
```

mohammadfawaz:
Ok so I've been completely blind here. I totally forgot that a "~" is needed for associated functions 😱  

The syntax that works is:
```
    let zero_val:u64 = ~u64::get_zero();
```

adlerjohn:
I thought the new parser removed this requirement?

[...]

Issue #1644, opened 2022-05-23
Allow `TypeInfo::Custom`s with `CallPath`s


emilyaherbert:
I want to be able to do this:

```rust
enum Foo {
  A: (),
  B: (),
}
```

```rust
use ::foo;

struct Bar {
  baz: foo::Foo
}

let x = Bar {
  baz: foo::Foo::A(())
};

match x {
  Bar { baz: foo::Foo::A(_) } => { .. },
  _ => { .. }
}
```

But right now that does not work because `TypeInfo::Custom` does not contain a `CallPath`: https://github.com/FuelLabs/sway/blob/17b9e2473001d9ab91a8f4237673a9ae8baeb492/sway-core/src/type_engine/type_info.rs#L67

esdrubal:
I tried to reproduce this error, but the following error was thrown: 
 ```
7 | struct Bar {
8 |     baz: foo::Foo
   |          ^^^^^^^^ Fully specified types are not supported in this position. Try importing the type and referring to it here.
9 | }
```
This issue no longer make sense, unless we want to support fully specified types in this position.

emilyaherbert:
There is not any bug here, this is a lack of implemented feature. We will want to support fully specified paths in this position, yes.

mohammadfawaz:
@esdrubal this seems related to what you're currently working on. I'm assigning this to you.

PR #3218, opened 2022-10-31
Eliminate need for `~` in `~Foo::bar()`

Centril:
Fixes https://github.com/FuelLabs/sway/issues/1436.

This PR first removes the need for `~` in `~Foo::bar` by first delaying the interpretation of `Foo::Bar` (associated call, free function call, or enum variant construction?) until type checking. This is achieved with `ExprKind::AmbiguousPathExpression`. During type checking, we ask whether `Foo` (and any prefixes) is a module, or an enum. If it's neither, we attempt type checking as an associated function call. Otherwise, we continue as if we had a `ExpressionKind::DelineatedPath`.

Once `~` is made redundant, it's then also removed from the language and so also from tests, examples, and the standard library.


sway-error/src/convert_parse_tree_error.rs
```diff
@@ -9,8 +9,6 @@ pub enum ConvertParseTreeError {
     FunctionArbitraryExpression { span: Span },
     #[error("generics are not supported here")]
     GenericsNotSupportedHere { span: Span },
-    #[error("fully qualified paths are not supported here")]
-    FullyQualifiedPathsNotSupportedHere { span: Span },
```
mohammadfawaz, line 13: Does this mean I can now write
```
    fn test_function(a: lib1::Foo, b: lib2::Foo) -> bool;
```
? That is, can i use call paths when referring to a type?

mohammadfawaz (commented):
🥳

sway-error/src/convert_parse_tree_error.rs
mohammadfawaz: Oh this seems related https://github.com/FuelLabs/sway/issues/1644

mohammadfawaz: Yeah nevermind, I was thinking of a different error here.

mohammadfawaz (approved):
🚀

sezna (approved):

PR #3219, opened 2022-10-31
Return undef if value being returned is not the right type

### LINKED ISSUE #3205: ICE: Verification failed: Function expiry return type must match its RET instructions
```
error: Internal compiler error: Verification failed: Function expiry return type must match its RET instructions.
```
Heres the code

```
#[storage(read)]
    fn expiry(name: str[8]) -> Result<u64, RegistrationValidityError> {
        match storage.names.get(name) {
            Option::Some(record) => {
                match timestamp() < record.expiry {
                    true => {
                        return Result::Ok(record.expiry);
                    },
                    false => {
                        return Result::Err(RegistrationValidityError::NameExpired);
                    }
                }
            },
            Option::None => {
                return Result::Err(RegistrationValidityError::NameNotRegistered);
            }
        }
    }
```

with record being of type Record
```
pub struct Record {
    /// The timestamp at which the name expires, and someone else can re-register the same name
    expiry: u64,
    /// The identity to which the name resolves to
    identity: Identity,
    /// The identity which controls the name, and can change the identity and owner
    owner: Identity,
}
```

If a semi colon is added to the end of the match statement, the error changes to 
`^ This path must return a value of type "Result<u64, RegistrationValidityError>" from function "expiry", but it does not.`

vaivaswatha:
This is an alternate fix to close #3205. 

See other PR here: https://github.com/FuelLabs/sway/pull/3210

I'm not sure if this works, so just creating a draft PR to let the CI do it's job.


vaivaswatha:
This seems to work. But this is an experiment, so closing in favour of #3210 implementing it if necessary.

otrho:
Yeah, I've been slowly taking `undef` out of the IR, as I think it's generally a code smell.  If there are times you need it then your design is probably broken.

But maybe that's a bit strict, what do you think @vaivaswatha?

vaivaswatha:
> Yeah, I've been slowly taking `undef` out of the IR, as I think it's generally a code smell. If there are times you need it then your design is probably broken.
> 
> But maybe that's a bit strict, what do you think @vaivaswatha?

I think `Undef`s are totally fine. They can be useful i many cases, like this one.

In a well typed (high level) source language, undef is bad. But this is at the IR stage after the program is type checked. So it's just a compiler convenience.

otrho:
Sorry, I mean `undef` consts in particular.  Undef can be useful for runtime dynamic stuff, but if you need an `undef` const I reckon something's broken.  Anyway...

PR #3213, opened 2022-10-31
Type check trait methods during trait declaration.

emilyaherbert:
This PR changes how trait methods are type checked---previously all "trait methods", i.e. these:

```rust
trait MyTrait {
	// interface surface
} {
	// trait methods
}
```

were type checked during type checking of the `impl` blocks that implemented those traits. Now, these trait methods are type checked during type checking of a trait declarations.

This PR implements this in a way that is trait-constraint-esque. Imagine we have this trait definition:

```rust
trait MyTrait {
	fn my_eq(self, other: Self) -> bool;
} {
	fn my_neq(self, other: Self) -> bool {
		!self.my_eq(other)
	}
}
```

We could say that from the perspective of the trait methods, the type `Self` is constrained upon the definition of the interface surface (i.e. we know that `Self` implements `my_eq`).

A similar approach is taken for trait constraints:

```rust
trait MyAdd {
	fn my_add(self, other: Self) -> Self;
}

trait MyMul {
	fn my_mul(self, other: Self) -> Self;
}

trait MyMath: MyAdd + MyMul {

} {
	fn my_double(self) -> Self {
		self.my_add(self)
	}

	fn my_exp(self) -> Self {
		self.my_mul(self);
	}
}
```

When type checking the trait methods of `MyMath`, we know that `Self` is constrained by `MyAdd`, `MySub`, and any methods inside of the interface surface (which is empty).

PR #3225, opened 2022-10-31
only run CI stages for relevant code

sezna:
Hopefully, this will make changes like #3223 faster. This change runs CI tests only when relevant files have been touched. E.g., don't re-run the entire e2e test suite if a `.md` file is changed.


sezna:
It appears that this leaves stages pending…looking into this

emilyaherbert:
> It appears that this leaves stages pending…looking into this

Hmm, this might not be just this PR. I've been having some issues with #3213 with getting the CI to trigger. I've had to push an empty commit just to trigger CI.

otrho:
Untenable?

Issue #3093, opened 2022-10-20
Incorrect compiler warning "This declaration is never used"


nfurfaro:
In the following image you can see the compiler warning about a declaration never being used, but this error enum is used in 2 different files. For now this can be reproduced by checking out the branch `furnic/bridge-token-contract` in the repo: https://github.com/FuelLabs/bridge-fungible-token. from there, `cd` into the inner `bridge-fungible-token` directory and run `forc build`.

<img width="937" alt="image" src="https://user-images.githubusercontent.com/17053076/197067029-a6c90c23-42c2-4d99-a948-3d0f4107512b.png">

mohammadfawaz:
From @nfurfaro:

This issue is a bit of a blocker for the bridge. No idea how to repro this in a more minimal way, as we import and use deps like this all the time without an issue.

pixelcircuits:
I ran into this same issue for the bridge predicates repo. Forc no longer supports testing, but they were Cargo tests anyway so all that's needed is to run the tests via Cargo vs Forc. See https://github.com/FuelLabs/bridge-message-predicates/pull/11/commits/2452de9d12fe6a7f85b65396c90edd904f9363b2

nfurfaro:
thank you @tritao !

PR #3232, opened 2022-11-01
Fix CFG node uniqueness.

tritao:
The CFG now keeps track of previously processed declarations and makes sure to re-use the previously added graph node, instead of creating duplicate nodes for the same declarations.

We do this by introducing a new IdentUnique type, which is pretty much the same as an Ident, but hashes with the full span, as opposed to an Ident which just uses the name as the hash, hence not fully ideal to use as the key for the CFG namespace, since it won't work correctly in multi-file cases.

Closes https://github.com/FuelLabs/sway/issues/3093.

<del>
This does multiple fixes:

- The declaration engine now tries to keep sure we do not create new declaration ids for previously added types.

Right now this is only implemented for enums, but should be extended in the future for other declaration types. This is kinda like the monormophization cache we previously discussed, but working for non-generic types.

- The CFG/DCA graph now also keeps track of previously processed declaration ids and makes sure to re-use the previously added graph node, instead of creating duplicate ones for the same declaration nodes.

This also needs to be extended for other declaration types.

Closes https://github.com/FuelLabs/sway/issues/3093.
</del>


emilyaherbert:
Instead of doing this, would it work to utilize the `parents` api that was just added to the declaration engine, instead? https://github.com/FuelLabs/sway/blob/3a52d50200dac7d2ad3bc1d3b7f920203aaa044d/sway-core/src/declaration_engine/declaration_engine.rs#L23

otrho:
So you're saying `IdentUnique` is the same as `Ident` except it derives `PartialEq` and `Hash` etc.?  That's what it looks like, and there's a lot of duplicated code.

Would it be less redundant to parameterise `Ident` with a hasher so that we could have `type Ident = IdentBase<OrigHasher>` and `type IdentUnique = IdentBase<FullHasher>` with a `From` from one to the other, and leave the actual `Ident` guts as is?  Then impl `Hash` for `IdentBase` in terms of `OrigHasher` and `FullHasher`.

otrho (approved):

mohammadfawaz (approved):

Issue #2923, opened 2022-10-03
Array OOB access should be a compile error


mohammadfawaz:
E.g. 
```rust
script;

fn main() -> u32 {
    let ary = [1, 2, 3, 4];
    let idx = 45;
    ary[idx]
}
```
This compiles fine but it shouldn't, at least in situations where `idx` is constant evaluatable.

otrho:
This surprises me considering https://github.com/FuelLabs/sway/tree/master/test/src/e2e_vm_tests/test_programs/should_fail/array_oob
The only difference is the indirection via the `idx` var.  Obviously it's enough.

mohammadfawaz:
Yeah `idx` is considered a dynamic index, unless we run the const-evaluator on it.

esdrubal:
I checked the code that returns the OOB access compiler error [sway-core/src/ir_generation/function.rs#L1688](https://github.com/FuelLabs/sway/blob/13ffbc3328805b8bf1abed1d34f9ecc597efa045/sway-core/src/ir_generation/function.rs#L1688).

In the case of the example above, the index_expr is compiled to an Instruction::Load with the value Instruction::GetPointer().
Maybe we can check the current block for an Instruction::Store of a const to the same pointer.

mohammadfawaz:
@esdrubal one way to do this is to leverage the const evaluator in https://github.com/FuelLabs/sway/blob/master/sway-core/src/ir_generation/const_eval.rs.
* If the const evaluator passes on the index, then we can just replace the value with the evaluated constant and do the OOB checks on it.
* If the const evaluator fails (it can only do so much, and not every expression is const-evaluatable), then we just don't do anything.

esdrubal:
@mohammadfawaz Thanks for pointing me out to [const_eval.rs](https://github.com/FuelLabs/sway/blob/master/sway-core/src/ir_generation/const_eval.rs). While using compile_constant_expression_to_constant I noticed that it is able to compile to a constant when index is a Literal. When it is a VariableExpression it doesn't know about variables in the scope and may try to compile to the value of a global const over shadowing local variables.

I am working on a solution for this and will send a PR later today.

PR #3233, opened 2022-11-01
Adds array out of bounds errors at compile time.

esdrubal:
`compile_array_index` now returns an array out of bounds error when array
 index is an oob const from a global const and when it loads a pointer that
 was previously in the same block stored with an oob const value.

Closes https://github.com/FuelLabs/sway/issues/2923

This covers the 2 new following cases:
```rust
script;

const i: u64 = 4;

fn main() -> u64 {
    // index out of bounds: the length is 3 but the index is 4
    let ary = [1, 2, 3];
    ary[i]
}
```
and:
```rust
script;

fn main() -> u64 {
    // index out of bounds: the length is 3 but the index is 4
    let ary = [1, 2, 3];
    let i = 4;
    ary[i]
}
```


mohammadfawaz:
Thank you @esdrubal! While this works in this particular case, I worry that this relies too much on matching a specific pattern, and any deviation from this pattern will cause this to fail. For example, if the index is now result of adding two literals, this will fail. In contrast, relying on the constant evaluator means this will _always_ be as good as the constant evaluator so we don't ever have to worry about it again.

esdrubal:
@mohammadfawaz Thank you for the feedback. The code now uses only `compile_constant_expression_to_constant`.

I had to pass `FnCompiler` to `compile_constant_expression_to_constant` so it can check local variables names before compiling to a global const. This will also be useful to fix #3237.

otrho:
I was never a fan of using the IR compiler for const eval and now this feels like the hackiness is leaking further out of const-eval... but I know the only alternative is probably a compile-time Sway interpreter.  How does Rust handle this stuff?

Should these errors actually be caught by more robust/user-friendly IR verification, especially now that we have span metadata pretty much working (even though they are opaque)?  Rather than pull IR into the compiler, instead push these checks back down to IR?

emilyaherbert (dismissed):

emilyaherbert (approved):

PR #3235, opened 2022-11-01
Test ec_recover using SDK + fuel-crypto

### LINKED ISSUE #3179: Rewrite `stdlib/ec_recover_test` as an `sdk-harness` test
This test relies on https://github.com/FuelLabs/sway/tree/master/test-sig-gen-util; not directly but relies on its existence I suppose. The test was written using values obtained from `test-sig-gen-util`. It would be better to write `ec_recover_test` as an SDK harness test where the values are automatically generated in the harness itself using a similar logic to what exists in `test-sig-gen-util`. 

One thing to note is that `test-sig-gen-util` uses `secp256k1_sign_compact_recoverable` which is deprecated. I think we're meant to be using `fuel-crypto` directly for similar behavior. (See https://github.com/FuelLabs/fuel-vm/pull/203)

With the new test written, we can simply remove the `test-sig-gen-util` directory.

nfurfaro:
This will allow us to deprecate & remove the `test-sig-gen-util`.
As a related follow-up, I'll add a similar new SDK test for the `ec_recover_EvmAddress` functionality which is also currently using the `test-sig-gen-util`. 
I'll also delete the old `test-sig-gen-util` in the second PR.

Closes #3179


test/src/sdk-harness/test_projects/ec_recover/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+        .call()
+        .await
+        .unwrap();
+
+    // let first = response.value.0;
+    // let second = response.value.1;
+    // let arrays: [[u8; 32]; 2] = [first.0, second.0];
+    // let joined: Vec<u8> = arrays.into_iter().flat_map(|s| s.into_iter()).collect();
+    // let joined_array: [u8; 64] = joined.try_into().unwrap();
+    // let pubkey = Bytes64::new(joined_array);
```
mohammadfawaz, line 113: Are these intentionally left here?

mohammadfawaz (commented):
Thanks @nfurfaro! Much cleaner :)

Just a nit. Also, 
* You can probably add "Closes https://github.com/FuelLabs/sway/issues/3179" in the PR description.
* Are you planning to deleting this old E2E test in a follow up PR as well?

test/src/sdk-harness/test_projects/ec_recover/mod.rs
nfurfaro: hmm, nope. Those look old, maybe I forgot to push my latest change.
I'll delete the old E2E test now. The only reason I've left test-sig-gen-util for now it that it's still kind of linked to the evm ec recover test which is the last one to be converted.

nfurfaro:
Hm. CI is failing, but it works locally. Need to investigate.

mohammadfawaz:
The test is not built? Why is it skipped?
image

mohammadfawaz (approved):

test/src/sdk-harness/test_projects/ec_recover/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+        )
+        .call()
+        .await
+        .unwrap();
+
+    let first = response.value.0;
+    let second = response.value.1;
+    let arrays: [[u8; 32]; 2] = [first.0, second.0];
+    let joined: Vec<u8> = arrays.into_iter().flat_map(|s| s.into_iter()).collect();
+    let joined_array: [u8; 64] = joined.try_into().unwrap();
```
Braqzen, line 86: Just noting that you should be able to redefine `joined` as an array `joined_array` so that you do not have to create a new name for a variable with a different type. This allows you to reuse the name as a different type.

Braqzen (approved):

PR #3236, opened 2022-11-01
Fix broken `fuel-specs` links in the Sway book and the standard library

homura:


docs/src/blockchain-development/identifiers.md
```diff
@@ -5,4 +5,4 @@ Addresses in Sway are similar to EVM addresses. The two major differences are:
 1. Sway addresses are 32 bytes long (instead of 20)
 2. Sway addresses are computed with the SHA-256 hash of the public key instead of the keccak-256 hash.
 
-Contracts, on the other hand, are uniquely identified with a contract ID rather than an address. A contract's ID is also 32 bytes long and is calculated [here](https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/identifiers.md#contract-id).
+Contracts, on the other hand, are uniquely identified with a contract ID rather than an address. A contract's ID is also 32 bytes long and is calculated [here](https://fuellabs.github.io/fuel-specs/master/protocol/identifiers#contract-id).
```
mohammadfawaz, line 8:
```suggestion
Contracts, on the other hand, are uniquely identified with a contract ID rather than an address. A contract's ID is also 32 bytes long and is calculated [here](https://fuellabs.github.io/fuel-specs/master/protocol/id/contract).
```

mohammadfawaz (changes requested):
Thank you so much for fixing all of those. Just one small link to fix.

mohammadfawaz (approved):
:shipit:

PR #3238, opened 2022-11-01
Move error signals to new module

nfurfaro:
This creates a new module `std::error_signals` and moves the 2 existing values to the module.
A follow-up PR (breaking) will be to update the value of `const FAILED_REQUIRE_SIGNAL = 42` to be `0xffff_ffff_ffff_0001`.


mohammadfawaz (commented):
Nice! Can we also move this https://github.com/FuelLabs/sway/blob/10b6c44145063fa919debc6d58ebc60dd738b033/sway-lib-std/src/token.sw#L11?

sway-lib-std/src/error_signals.sw
```diff
@@ -3,5 +3,9 @@ library error_signals;
 
 /// revert with this value for a failing call to std::revert::require.
 pub const FAILED_REQUIRE_SIGNAL = 42;
+
 /// revert with this value for a failing call to std::message::send_message.
 pub const FAILED_SEND_MESSAGE_SIGNAL = 0xffff_ffff_ffff_0002;
+
+/// revert with this value for a failing call to std::token::transfer_to_address.
+pub const FAILED_TRANSFER_TO_ADDRESS_SIGNAL = 0xffff_ffff_ffff_0001;
```
mohammadfawaz, line 11: Do you mind keeping these in numerical order? Eventually `FAILED_REQUIRE_SIGNAL` should be changed to `0xffff_ffff_ffff_0000` so we can keep it at the start.

nfurfaro: just made the change to  `FAILED_REQUIRE_SIGNAL` and ordered them numerically!

mohammadfawaz (approved):

Issue #3155, opened 2022-10-26
Temporarily disallow `asm` blocks from writing an initialized registers


mohammadfawaz:
This is due to the addition of `mem2reg` in https://github.com/FuelLabs/sway/pull/3128. Some odd patterns arise when you allow `asm` blocks to ~~read and write the same value in IR, which causes `mem2reg` not to behave well. In any case, we should disallow these patterns temporarily until https://github.com/FuelLabs/sway/issues/2819 is resolved and we have a better mental model regarding `asm` blocks in IR.~~ write an initialized register.

I think this change can be done in the parser or type checking. Example use cases:

~~Reading and writing the same register in the same instruction~~
Writing register `r1` even though `r1` is initialized to `ptr`.
```rust
    asm(r1: ptr) {
        lw r1 r1 i0;
        r1: u64
    }
```

~~Reading and writing the same register in different instructions~~
Writing registers `r1` and `r2` even though both are initialized.
```rust
    asm(r1: ptr1, r2: reg2) {
        lw r2 r1 i0;
        lw r1 r2 i0;
        r1: u64
    }
```

otrho:
Checks like these could be permanent IMO, at least in the form of warnings.

mohammadfawaz:
I spoke the @vaivaswatha and the real requirement here is that a register cannot be both _initialized_ in the args list of the `asm` block AND be assigned to in the body of the block. So my examples above are correct, but my accurate is not accurate.

Issue #3162, opened 2022-10-27
Consider using `FxHashMap` for performance


emilyaherbert:
> > > Note that a copy only occurs if the data is shared with another instance, (i.e. if the OrdMap was previously cloned elsewhere) and only the shared parts are copied.
> > 
> > 
> > Ahhh I see. OrdMap it is!
> 
> Sorry for the late input. If determinism is the only thing that matter and not the actual order, then [FxHashMap](https://docs.rs/rustc-hash/latest/rustc_hash/type.FxHashMap.html) could be a more efficient option.

I'm already using this in the compiler. FxHashMap is faster too if the key type is small sized. If we want exactly `HashMap`'s performance, but with determinism, then we should just instantiate `HashMap` with a non-random hasher, similar to what FxHashMap does (https://docs.rs/rustc-hash/latest/src/rustc_hash/lib.rs.html#43).

So something like `pub type SwayHashMap<K, V> = HashMap<K, V, BuildHasherDefault<DefaultHasher>>;` would give us exactly what `HashMap` gives, but with determinism.

_Originally posted by @vaivaswatha in https://github.com/FuelLabs/sway/issues/3154#issuecomment-1292840895_

vaivaswatha:
I suggest the following (applies to `HashSet` too equally).

1. Define `pub type SwayHashMap<K, V> = HashMap<K, V, BuildHasherDefault<DefaultHasher>>;` 
2. Use `SwayHashMap` wherever ordering doesn't matter, but just needs to be deterministic. Replace all uses of `HashMap` in the compiler with this.
3. Use `FxHashMap` whenever the key size is just an integer (or something equal in size to that or smaller).

vaivaswatha:
>Generally fxhash is faster than fnv on u32, u64, or any byte sequence with length >= 5.
This is from [FxHash](https://github.com/cbreeden/fxhash)'s GitHub page. (I somehow remembered the opposite, that for smaller keys, FxHash is better, but I think I was wrong).

This means we could just blindly use `FxHashMap` (and `FxHashSet`) everywhere as we don't typically have smaller than int size keys in the compiler.

PR #3239, opened 2022-11-01
Disallow assigning to initialized registers in `asm` blocks

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/3155

This is needed due to `mem2reg` and some unclear semantics around initialized registers being passed as ref/mut ref. Besides, it's a good idea to restrict `asm` blocks as much as possible to avoid weird unwanted behavior.

image

* I also removed the `disallow_opcodes` check because the parser takes care of that now. The parser only allows legal opcodes in `asm` blocks.
* Finally, I update the parser to accept the correct storage opcodes with `fuel-core 0.13` and updates all the tests.


sway-error/src/error.rs
```diff
@@ -670,6 +670,11 @@ pub enum CompileError {
     RefMutableNotAllowedInMain { param_name: Ident },
     #[error("returning a `raw_ptr` from `main()` is not allowed")]
     PointerReturnNotAllowedInMain { span: Span },
+    #[error(
+        "Register \"{name}\" is initialized but later assigned which is not allowed. \
+            Consider uninitializing \"{name}\"."
```
vaivaswatha, line 675:
```suggestion
        "Register \"{name}\" is initialized and later reassigned which is not allowed. \
            Consider assigning to a different register inside the ASM block."
```
Rather than advising to not initialise, it might be more indicative to advice to use a different destination register inside the block.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -31,7 +32,7 @@ use sway_error::{
 };
 use sway_types::{integer_bits::IntegerBits, Ident, Span, Spanned};
 
-use std::collections::{HashMap, VecDeque};
+use std::collections::{BTreeSet, HashMap, VecDeque};
```
vaivaswatha, line 35: From what I could understand, the order itself doesn't matter, so maybe [`FxHashSet`](https://github.com/FuelLabs/sway/issues/3162) could be faster?

sway-error/src/error.rs
mohammadfawaz: Yeah that makes sense. Thanks!

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
vaivaswatha: `FxHashMap`s also are deterministic. (though in undefined order). The issue I linked to discusses this.

mohammadfawaz: Awesome! That'll be very handy.

vaivaswatha (dismissed):

vaivaswatha (dismissed):

vaivaswatha (approved):

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -1920,24 +1929,70 @@ mod tests {
...
-    let mut errors = vec![];
 
-    match op.as_str().to_lowercase().as_str() {
-        "ji" => {
-            errors.push(CompileError::DisallowedJi { span: op.span() });
-        }
-        "jnei" => {
-            errors.push(CompileError::DisallowedJnei { span: op.span() });
-        }
-        "jnzi" => {
```
sezna, line 1933: One thing I'm concerned about is the UX of this change. Consider a user who is reading the opcode specs. They think they can type any opcode in an asm block. Currently, if they type a `ji`, the compiler will inform them that specifically that opcode is disallowed. With this change, the error will be `unrecognized opcode`. That error message is confusing, as it certainly is an opcode.

I think we should either parse disallowed opcodes and report a nicer error about them, or have extensive documentation on allowed and disallowed opcodes in the Sway Book. Preferably the former as it maintains the current UX.

mohammadfawaz: That's a good point and you're absolutely right. This list is also incomplete at the moment and that code wasn't triggering anyways because the parser won't allow you to have a `ji` for example. So, we're in no worse place than before.

What I'm going to do is address this UX issue in a separate PR where I allow all opcodes in the parser but disallow them in type checking. Let me open an issue about that and I'll make a follow up PR.

sezna: Sounds good, I'll approve this for now. You're right that this is actually a no-op currently, I missed that.

sezna (approved):

PR #3240, opened 2022-11-01
sway-test-rs template; update instruction links

K1-R1:


K1-R1:
I've just seen these instructions in the 0.29 Sway book, would it better just to link to the relevant section in there?

mitchmindtree:
Yes linking to the relevant book chapter instead would be great :+1:

templates/sway-test-rs/README.md
```diff
@@ -6,6 +6,6 @@ testing within an existing Sway project.
 This template is designed for Rust developers who wish to test integration of
 their Rust application and Sway code.
 
-See the "Testing with Rust" chapter of [the Sway
-Book](https://fuellabs.github.io/sway/v0.25.0/) for a thorough guide on how to
+See the `Testing with Rust` chapter of [the Sway
```
mohammadfawaz, line 9: I suggest we link to that chapter directly: https://fuellabs.github.io/sway/master/testing/testing-with-rust.html. This is not ideal because it links to the `master` book (we can't have `latest` in the link sadly) but maybe that's okay given that this README _is_ on `master`?

K1-R1: Yh I agree, I tried it with `latest` but forgot about `master`. I'll keep the link to `the sway book` so it's always up-to-date and add the chapter specific link as suggested

templates/sway-test-rs/README.md
```diff
@@ -6,6 +6,6 @@ testing within an existing Sway project.
 This template is designed for Rust developers who wish to test integration of
 their Rust application and Sway code.
 
-See the "Testing with Rust" chapter of [the Sway
-Book](https://fuellabs.github.io/sway/v0.25.0/) for a thorough guide on how to
+See the [Testing with Rust](https://fuellabs.github.io/sway/master/testing/testing-with-rust.html) chapter of [the Sway
+Book](https://fuellabs.github.io/sway) for a thorough guide on how to
```
mohammadfawaz, line 10: I'm suggesting this only because the two links now point to different books which is odd.
```suggestion
Book](https://fuellabs.github.io/sway/master) for a thorough guide on how to
```

bingcicle (approved):
LGTM after the suggestion has been committed

mohammadfawaz (approved):

bingcicle (approved):

PR #3241, opened 2022-11-02
Generic trait `From<T>` in the standard library

### LINKED ISSUE #1078: Remove double-definition of `From`
Ref: https://github.com/FuelLabs/sway/issues/819

mohammadfawaz:
Closes #1078 

* New trait `From<T>` in a new library called `convert.rs`
* Most changes are trivial
* Added `From` to the prelude
* Had to modify `from()` in `U128`, `U256`, and `B512` to take a tuple of components instead of multiple arguments to conform to the trait. This is a breaking change of course.
* Updating lock files for all the examples.


sway-lib-std/src/prelude.sw
```diff
@@ -17,3 +17,6 @@ use ::assert::assert;
 use ::option::Option;
 use ::result::Result;
 use ::revert::{require, revert};
+
+/* Convert */
```
nfurfaro, line 21: Is this comment needed? NOne of the other modules have comments.

nfurfaro (approved):
LGTM!

sway-lib-std/src/prelude.sw
mohammadfawaz: So blocks of imports have comments:
image
I'm hoping we'll have an `Into` trait in the future that will live under `/* Convert */` as well.

AlicanC (approved):

sezna (approved):

PR #3243, opened 2022-11-02
Trying parallel SDK tests

mohammadfawaz:


mohammadfawaz:
Closing as @Voxelot will just do that in https://github.com/FuelLabs/sway/pull/3181. This does seem to work though and cuts down the total time by around 10 min which is nice.

PR #3244, opened 2022-11-02
Fixed "Cannot infer type" error to point to call site span.

### LINKED ISSUE #3192: `Cannot infer type for parameter "T"` does not point to the location where the type was not annotated
When the compiler is unable to infer the type for `Option`, the error `Cannot infer type for type parameter "T".` is displayed in the std-lib. 

This should instead show where the type is unannotated in the user's code. This will allow users to quickly debug their Sway code and eliminate the need to comment out sections of code to find where in their code the error occurred.

For example, the shown error was generated with the following code:

```
    #[storage(write)]
    fn test_function() -> bool {
        store(VAL, Option::None());
        true
    }
```

infer_type_error

This does not inform the user where the error has occurred.

Tested with forc v0.25.2

esdrubal:
`CompileError::UnableToInferGeneric` was using as span the span of the type_id. This causes the error to point to where the type is declared instead of to where it should be annotated.

The solution was to store a call site span in `CollectTypesMetadataContext` and pass it to `TypeInfo::UnknownGeneric` so when we convert it to the actual error we can use the call site span which points to the place that is missing a type annotation.

Closes #3192

### DIFF
--- a/sway-core/src/language/ty/expression/expression.rs
+++ b/sway-core/src/language/ty/expression/expression.rs
@@ -79,7 +79,6 @@ impl CollectTypesMetadata for TyExpression {
                 call_path,
                 ..
             } => {
-                ctx.set_call_site_span(Some(call_path.span()));
                 for arg in arguments.iter() {
                     res.append(&mut check!(
                         arg.1.collect_types_metadata(ctx),
@@ -92,6 +91,12 @@ impl CollectTypesMetadata for TyExpression {
                     Ok(decl) => decl,
                     Err(e) => return err(vec![], vec![e]),
                 };
+
+                ctx.call_site_push();
+                for type_parameter in function_decl.type_parameters {
+                    ctx.call_site_insert(type_parameter.type_id, call_path.span())
+                }
+
                 for content in function_decl.body.contents.iter() {
                     res.append(&mut check!(
                         content.collect_types_metadata(ctx),
@@ -100,7 +105,7 @@ impl CollectTypesMetadata for TyExpression {
                         errors
                     ));
                 }
-                ctx.set_call_site_span(None);
+                ctx.call_site_pop();
             }
             Tuple { fields } => {
                 for field in fields.iter() {
@@ -124,7 +129,15 @@ impl CollectTypesMetadata for TyExpression {
                     }
                 }
             }
-            StructExpression { fields, .. } => {
+            StructExpression { fields, span, .. } => {
+                if let TypeInfo::Struct {
+                    type_parameters, ..
+                } = look_up_type_id(self.return_type.clone())
+                {
+                    for type_parameter in type_parameters {
+                        ctx.call_site_insert(type_parameter.type_id, span.clone());
+                    }
+                }
                 for field in fields.iter() {
                     res.append(&mut check!(
                         field.value.collect_types_metadata(ctx),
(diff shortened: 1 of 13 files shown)


sway-core/src/language/ty/expression/expression.rs
```diff
@@ -70,8 +70,10 @@ impl CollectTypesMetadata for TyExpression {
             FunctionApplication {
                 arguments,
                 function_decl_id,
+                call_path,
                 ..
             } => {
+                ctx.set_call_site_span(Some(call_path.span()));
```
mohammadfawaz, line 76: Do we need to do this for every possible expression type in this `match` statement?

emilyaherbert (changes requested):
Thank you for tackling this UX regression!

I think we should be cautious to add AST information to `TypeInfo`. One of the core problems that we are having right now with this part of the compiler is that we are in a situation where we have to try to merge external representation of types (what the user has written specifically, spans, etc) with the internal logic of types (for instance all `u64`'s should have the same "internal type" for optimal type inference) in order to achieve type inference while delivering quality UX. This can lead to an inverse relationship between adding more ast information to `TypeInfo` with degrading performance from the type engine. We should consider making these concepts more distinct, for example here is the Rust reference about their solution: https://rustc-dev-guide.rust-lang.org/type-checking.html?highlight=type%20collec#type-collection

That being said, for this PR I recommend that we take a different approach. The `DeclarationId` type exposes a distinct `Span`. https://github.com/FuelLabs/sway/blob/cfecaf4c405e46993d80ee9168605893bd18c10a/sway-core/src/declaration_engine/declaration_id.rs#L17

The `DeclarationEngine` API doesn't expose this functionality right now, but I propose that we use the `DeclarationId` span to achieve this change. When we create a new `DeclarationId`, we can (optionally?) provide a span to create that `DelcarationId` with. For example, in the case of `ty::TyExpressionVariant::FunctionApplication` this span would be the `call_path.span()`.

sway-core/src/language/ty/expression/expression.rs
esdrubal: This may be needed for other expressions, I should write test cases for the other expressions.

esdrubal:
@emilyaherbert  Thank you for the comment. I see in `ty::TyExpressionVariant::FunctionApplication` a `DeclarationId` in the field `function_decl_id` but it contains the span of `ty::TyFunctionDeclaration`. Are you proposing `ty::TyExpressionVariant::FunctionApplication` or `TyExpression` to have a new `DeclarationId` with the span of where they originate from?

emilyaherbert:
Hi @esdrubal, I was proposing that with the `DeclarationId` that is already in `ty::TyExpressionVariant::FunctionApplication`, that the span field of this `DeclarationId` be used to hold the information about the originating span from when that `DeclarationId` was generated. But I also like the current solution up on GH right now where you have added a `call_path` value to `ty::TyExpressionVariant::FunctionApplication` to accomplish the same thing.

sway-core/src/language/ty/expression/expression.rs
esdrubal: Added test case and fix for when it cannot infer struct parameter types. There may exist more cases that I didn't think of. But for now it already fixes 3 uses cases.

emilyaherbert (approved):

Issue #1759, opened 2022-05-29
Reoragnizing sway-std-lib


Braqzen:
# Is this a new feature or an improvement to something that already exists?
Select 1 checkbox
- [ ] New feature
- [ ] Improvement

# Description

## _What_ does your feature accomplish?
It makes it easier to navigate the standard library depending on the context.
Furthermore, it makes it easier to scale the size of the library.

## _How_ does your feature accomplishes your goal?
It accomplishes this by looking at the context of what each file achieves and groups them together under directories e.g. `ecr.sw` and `hash.sw` can be grouped under `hashing` or `cryptography`.

# Draft implementation

1. We have a `chain.sw` which only has a `dep` decleration inside. We also have a chain directory. This should be reorganized.
2. We have `context.sw` and also a `context` directory, this too should be fixed.
3. `assert.sw` and `require.sw` should probably be grouped together if not placed in the same file
4. `option.sw` and `result.sw` should also be together, perhaps with `logging.sw`?
5. `ecr.sw` and `hash.sw` should be together, keep in mind that we have a `vm` directory which is specifically for EVM equivalent `ecr`
6. `intrinsics.sw`, `storage.sw`, `block.sw`, `context.sw` and `chain.sw` could be together?
7. `address.sw`, `contract_id.sw`, `b512.sw` and `u128.sw` can be together?

cc @nfurfaro @adlerjohn @sezna @otrho

adlerjohn:
Things that are mirrors of the Rust standard library should be kept as they are in Rust. E.g. `Result` is in `std::result` (https://doc.rust-lang.org/std/result/) and `Option` is in `std::option` (https://doc.rust-lang.org/std/option/).

For our own types and functions, it's definitely the case that things could be reorganized. E.g. a `crypto` module for hashing and signature stuff. Note that this will require extreme care: the suggestion "`option.sw` and `result.sw` should also be together, perhaps with `logging.sw`" is exactly something we _don't_ want to do. The former two are mirrors of Rust stdlib types, and logging is something completely orthogonal and has to do with the blockchain. They have no semantic relationship.

Designing a proper API is non-trivial, and we have to take extreme care when designing the stdlib API to make it good. In the interim, avoiding grouping and having a flat structure at least doesn't impose any implicit semantics on groups of functions/types.

PR #3247, opened 2022-11-02
Flatten the stdlib

nfurfaro:
While a final design for the organization of `std` will likely involve more discussion (& consideration of the proposed `Contract` type with some of the existing free functions converted to methods on the `Contract` type), the changes I've made here remove some weirdness.
This is a breaking change, so it might be a good time to clean at least this up.
I've left the `vm` directory as currently, it seems like the only good use of a subdirectory in `std`.
If anyone thinks this needs more discussion we can mark this a draft.

ref #1759


sway-lib-std/src/context.sw
```diff
@@ -1,11 +1,11 @@
 //! Functionality for accessing context-specific information about the current contract or message.
 library context;
-dep context/registers;
-dep context/call_frames;
+dep registers;
+dep call_frames;
```
mohammadfawaz, line 4: Are those still needed?

mohammadfawaz (commented):
I'm generally in favour of this. It's currently a bit cumbersome to have to import `std::auth::chain::msg_sender()` just to access a utility that is so commonly used.

sway-lib-std/src/context.sw
nfurfaro: I thought so, but they are not. Will remove and update a few more paths.

mohammadfawaz (approved):
Awesome, :shipit:

Braqzen (approved):

PR #3248, opened 2022-11-02
feat: Implement workspace deployment/running

### LINKED ISSUE #3156: Update deploying logic to take workspaces into account
As a next step in the full support of workspaces, we need to update deploying logic. We will have workspace member ordering after #3134 and we should utilize it for deploying in the correct order too.

related and part of https://github.com/FuelLabs/sway/issues/1777

### LINKED ISSUE #3312: Update `forc run` logic to take workspaces into account
As a next step in the full support of workspaces, we need to update running logic to enable running all scripts residing in a workspace.

related and part of https://github.com/FuelLabs/sway/issues/1777

kayagokalp:
closes #3156.
closes #3312.

blocked by:

- [x] #3146


mitchmindtree (commented):
Overall looking good!

It does look like there's quite a bit of duplication between the `deploy` and `run` functions. Perhaps we can split out some of the common behaviour into a function or two?

forc-plugins/forc-client/src/ops/pkg_util.rs
```diff
@@ -0,0 +1,54 @@
...
+    match built {
+        pkg::Built::Package(built_pkg) => {
+            let pkg_name = member_manifests
+                .keys()
+                .next()
+                .expect("built package is missing");
+            Ok(std::iter::once((pkg_name.clone(), *built_pkg)).collect())
+        }
+        pkg::Built::Workspace(built_workspace) => Ok(built_workspace),
+    }
```
mitchmindtree, line 21: Perhaps this could be a method on `Built`? E.g.
```rust
impl Built {
    pub fn into_members(self) -> HashMap<String, BuiltPackage> {
        // ...
    }
}
```

forc-plugins/forc-client/src/ops/pkg_util.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use anyhow::Result;
+use forc_pkg::{self as pkg, manifest::ManifestFile, BuildOpts, BuildPlan};
+use pkg::{manifest::MemberManifestFiles, BuiltPackage, PackageManifestFile};
+
+fn built_pkgs(
+    build_opts: BuildOpts,
+    member_manifests: &MemberManifestFiles,
+) -> Result<HashMap<String, BuiltPackage>> {
+    let built = forc_pkg::build_with_options(build_opts)?;
```
mitchmindtree, line 11: Ahh, I'm just realizing that we end up building *all* members here (not just contracts). Ideally, we would do the build with only *contract* members as outputs, not all members as outputs. This is probably fine for now, but we should open an issue to address this in a future PR. We might want to expose an alternative `forc_pkg::build_contracts_with_opts` or similar, or allow for specifying the set of desired outputs.

forc-plugins/forc-client/src/ops/run/op.rs
```diff
@@ -34,25 +60,18 @@ pub async fn run(command: RunCommand) -> Result<Vec<fuel_tx::Receipt>> {
...
-
-    let build_opts = build_opts_from_cmd(&command);
-    let compiled = match forc_pkg::build_with_options(build_opts)? {
-        pkg::Built::Package(built_package) => *built_package,
-        pkg::Built::Workspace(_) => bail!("running workspaces not yet supported"),
-    };
-
     let contract_ids: Vec<ContractId> = command
         .contract
+        .clone()
```
mitchmindtree, line 65: No need to clone the `Vec<String>` here, and we don't want to swallow the error if we can't parse the `ContractId`. Perhaps this could be something like:
```rust
let contract_ids = command
    .as_ref()
    .into_iter()
    .flat_map(|contracts| contracts.iter())
    .map(|contract| ContractId::from_str(&contract))
    .collect::<Result<Vec<ContractId>>>()?;
```

test/src/e2e_vm_tests/harness.rs
```diff
@@ -64,7 +70,12 @@ pub(crate) async fn runs_on_node(
         signing_key: Some(SecretKey::from_str(SECRET_KEY).unwrap()),
         ..Default::default()
     };
-    run(command).await
+    run(command).await.map(|ran_scripts| {
+        ran_scripts
+            .first()
+            .map(|ran_script| ran_script.receipts.clone())
+            .unwrap()
```
mitchmindtree, line 77: Could avoid the clone here with:
```rust
ran_scripts
    .into_iter()
    .next()
    .map(|ran_script| ran_script.receipts)
    .unwrap()
```

mitchmindtree (dismissed):
Looking good, utACK! One minor nit, but not important at all.

test/src/e2e_vm_tests/harness.rs
kayagokalp: I will address this right away since I found a bug when trying to deploy just a single member of the workspace, so will address them together

mitchmindtree (approved):

Issue #3086, opened 2022-10-20
Add `raw_slice`


AlicanC:
Currently, there's no way to represent a DST. This causes users of Sway to use low-level escape hatches:
```rs
fn main(
  // Variable length args as `(buf_offset, len)`
  arg1: (u64, u64),
  arg2: (u64, u64),
) -> (u64, u64) {
  // Buffer is appened to script_data so it should be
  // right after `arg2`
  let buf_ptr = addr_of(arg2) + size_of_val(arg2);

  // Turn arg offsets into pointers
  arg1.0 += buf_ptr;
  arg2.0 += buf_ptr;

  // Do something with the args
  let res: (u64, u64) = concat(arg1, arg2);

  // Return the result
  asm(ptr: res.0, len: res.1) {
    retd ptr len
  }
}
```

These low level operations can be extracted from `main` as such:
```rs
struct raw_slice {
  ptr: u64,
  len: u64,
}
impl raw_slice {
  fn ret() -> raw_slice {
    asm(ptr: self.ptr, len: self.len) {
      retd ptr len
    };
    // Make compiler happy
    asm() { zero: raw_slice }
  }
}

struct ScriptDataRawSlice {
  offset: u64,
  len: u64,
}
impl Into<raw_slice> for ScriptDataRawSlice {
  fn into(self) -> raw_slice {
    let buf_ptr = /* ... */;
    raw_slice {
      ptr: buf_ptr + self.offset,
      len: self.len
    }
  }
}

fn main(
  arg1: ScriptDataRawSlice,
  arg2: ScriptDataRawSlice,
) -> raw_slice {
  concat(arg1.into(), arg2.into()).ret()
}
```

And with the compiler doing all this implicitly, `main` becomes:
```rs
fn main(
  arg1: raw_slice,
  arg2: raw_slice,
) -> raw_slice {
  concat(arg1, arg2)
}
```

---

Then we could add typed slices and use it with `Vec`:
```rs
struct slice<T> {
  ptr: u64,
  len: u64, // This isn't in bytes but in Ts
}
impl<T> slice<T> {
  fn byte_len(self) -> u64 {
    self.len * size_of::<T>()
  }
}

impl<T> Into<slice<T>> for Vec<T> {
  fn into(self) -> slice<T> {
    slice {
      ptr: self.buf.ptr,
      len: self.len,
    }
  }
}
impl<T> From<slice<T>> for Vec<T> {
  fn into(self) -> slice<T> {
    Vec {
[...]

Issue #2900, opened 2022-09-29
Special-case `Vec` when returning from a script


mohammadfawaz:
When a vector is returned from a script, instead of using `retd` to return the stuct `Vec` itself (which will be disallowed soon anyways because `Vec` will contain a `raw_ptr` - see https://github.com/FuelLabs/sway/issues/2551), use `retd` to return the actual data that lives on the heap.

mohammadfawaz:
So https://github.com/FuelLabs/sway/pull/3288 (combined with https://github.com/FuelLabs/fuels-rs/pull/743) covers the basic uses case of returning heap data from a script. There are certainly more complex cases that we would like to support such as returning a tuple with two vectors, etc. However, for now, we don't need to worry about them I think so I will close this issue.

PR #3251, opened 2022-11-02
[WIP] Allow returning and receiving `Vec`s in scripts

AlicanC:
This PR is a WIP.

This draft contains an attempt to solve the [issue](https://github.com/FuelLabs/sway/issues/2900) by generalizing the problem into "supporting dynamically sized types" and implementing a raw DST `raw_slice` as detailed on https://github.com/FuelLabs/sway/issues/3086.

ATM the code is messy but it should be good enough to ask for an initial review to see if I have a green light to proceed.

`raw_slice` aims to be just a pointer and a length that together represent a range of bytes, without a concept of unfilled capacity like `Vec`, or a hard-coded growth logic like `RawVec`. It should be roughly equivalent to `&[u8]` in Rust and it could be later evolved into `ref [T]` here in Sway.

This PR, even in this state, would unblock a [fuels-ts issue](https://github.com/FuelLabs/fuels-ts/issues/467#issuecomment-1285871841) which is blocked by the lack of `RETD`. While we could reintroduce `RETD`, it is only necessary for returning ASM-crafted DSTs so adding a native DST and support for returning it also unblocks the issue. For this reason, we could consider releasing this rather sooner and leave detailing to a follow-up PR.

Currently, this PR:
- Adds `raw_slice`, which is like `raw_ptr`, but:
  - It's a `(u64, u64)` instead of just a `u64`.
  - In IR, it doesn't use the previously available types, but has its own type `Type::Slice`.
  - If they are flat, meaning not aggregated in another type, they can be returned from scripts. (So you can't return a `Vec` which contains a `raw_slice`, but you can return `vec.as_slice()`. Implicitly calling something like `IntoSlice::as_slice` is TODO.)
  - Scripts receiving `raw_slices` in `main()` args is TODO. This is already possible and the lack shouldn't block anything though.
  - Intrinsic support for `add`, `sub`, etc. is TODO.
- Uses `raw_slice` in different parts of `std`, mainly `std::{alloc,vec}`. This allows us to remove size args from certain APIs.

If it seems like this PR is on the right direction, I'd like to proceed and:
- Check and implement what's left to unblock [fuels-ts issue](https://github.com/FuelLabs/fuels-ts/issues/467#issuecomment-1285871841).
- Clean the code up and request a merge.
- Keep following-up with PRs to support receiving slices in `main()` args and such.

### DIFF
--- a/sway-core/src/asm_generation/asm_builder.rs
+++ b/sway-core/src/asm_generation/asm_builder.rs
@@ -1331,20 +1331,42 @@ impl<'ir> AsmBuilder<'ir> {
                     comment: "".into(),
                 });
             } else {
-                // If the type not a reference type then we use RETD to return data.  First put the
-                // size into the data section, then add a LW to get it, then add a RETD which uses
-                // it.
+                // If the type is not a copy type then we use RETD to return data.
                 let size_reg = self.reg_seqr.next();
-                let size_in_bytes = ir_type_size_in_bytes(self.context, ret_type);
-                let size_data_id = self
-                    .data_section
-                    .insert_data_value(Entry::new_word(size_in_bytes, None));
+                if ret_type.eq(self.context, &Type::Slice) {
+                    // If this is a slice then return what it points to.
+                    self.cur_bytecode.push(Op {
+                        opcode: Either::Left(VirtualOp::LW(
+                            size_reg.clone(),
+                            ret_reg.clone(),
+                            VirtualImmediate12 { value: 1 },
+                        )),
+                        owning_span: owning_span.clone(),
+                        comment: "load size of returned slice".into(),
+                    });
+                    self.cur_bytecode.push(Op {
+                        opcode: Either::Left(VirtualOp::LW(
+                            ret_reg.clone(),
+                            ret_reg.clone(),
+                            VirtualImmediate12 { value: 0 },
+                        )),
+                        owning_span: owning_span.clone(),
+                        comment: "load ptr of returned slice".into(),
+                    });
+                } else {
+                    // First put the size into the data section, then add a LW to get it,
+                    // then add a RETD which uses it.
+                    let size_in_bytes = ir_type_size_in_bytes(self.context, ret_type);
+                    let size_data_id = self
+                        .data_section
+                        .insert_data_value(Entry::new_word(size_in_bytes, None));
 
-                self.cur_bytecode.push(Op {
-                    opcode: Either::Left(VirtualOp::LWDataId(size_reg.clone(), size_data_id)),
-                    owning_span: owning_span.clone(),
-                    comment: "loading size for RETD".into(),
-                });
+                    self.cur_bytecode.push(Op {
+                        opcode: Either::Left(VirtualOp::LWDataId(size_reg.clone(), size_data_id)),
+                        owning_span: owning_span.clone(),
+                        comment: "load size of returned ref".into(),
+                    });
+                }
                 self.cur_bytecode.push(Op {
                     owning_span,
                     opcode: Either::Left(VirtualOp::RETD(ret_reg, size_reg)),
--- a/sway-core/src/asm_generation/asm_builder/functions.rs
+++ b/sway-core/src/asm_generation/asm_builder/functions.rs
@@ -446,6 +446,10 @@ impl<'ir> AsmBuilder<'ir> {
                         self.ptr_map.insert(*ptr, Storage::Stack(stack_base));
                         stack_base += 1;
                     }
+                    Type::Slice => {
+                        self.ptr_map.insert(*ptr, Storage::Stack(stack_base));
+                        stack_base += 2;
+                    }
                     Type::B256 => {
                         // XXX Like strings, should we just reserve space for a pointer?
                         self.ptr_map.insert(*ptr, Storage::Stack(stack_base));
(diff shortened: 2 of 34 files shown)


otrho:
Looks nice!  I can see why you'd want to be able to special-case against a 'slice' type in the compiler, to constrain it much like `raw_ptr`.  But I don't think you need to introduce a new type to the IR.   The IR is lower level and I think `Slice` should just be a struct with either a `Pointer` and `u64` or if that doesn't work two `u64`s.

We already lower tuples and enums to structs in the IR  in the same way.

What are your thoughts @vaivaswatha?

mohammadfawaz:
Most of what I see here very reasonable. I haven't done a deep dive yet but I see how the concept of a `raw_slice` could be useful. I was hoping `Vec` would remain identical to Rust's `Vec` but I can see why it's hard to achieve that AND returning from scripts simulatenously.

> but you can return vec.as_slice(). Implicitly calling something like IntoSlice::as_slice

So for returning a vector from a contract to a script, does the current behavior change? For returning from a script, I'm assuming you imagine something like this:
```rust
fn main() -> raw_slice {
    let vector = ...
    ...
    vector.as_slice() // the type of this is `raw_slice` which works and returning a `raw_slice` is specialized in codegen
}
```

mohammadfawaz:
> 

I think the reasoning for carrying `raw_slice` all the way is because returning a `raw_slice` is special and is not similar to returning a struct. Returning a `raw_slice` becomes `retd base_reg size_reg` where `base_reg` is the first element of `raw_slice` and `size_reg` is the second element of `raw_slice`.

mohammadfawaz:
> Looks nice! I can see why you'd want to be able to special-case against a 'slice' type in the compiler, to constrain it much like `raw_ptr`. But I don't think you need to introduce a new type to the IR. The IR is lower level and I think `Slice` should just be a struct with either a `Pointer` and `u64` or if that doesn't work two `u64`s.
> 
> We already lower tuples and enums to structs in the IR in the same way.
> 
> What are your thoughts @vaivaswatha?

I think the reasoning for carrying `raw_slice` all the way is because returning a `raw_slice` is special and is not similar to returning a struct. Returning a `raw_slice` becomes `retd base_reg size_reg` where `base_reg` is the first element of `raw_slice` and `size_reg` is the second element of `raw_slice`.

cc @otrho

otrho:
This is unfortunate.  Do we already ban returning `raw_ptr` though?  I kinda feel like at that stage if we see a struct with a pointer and a `u64` then it could be specialised similarly.  Maybe that's too liberal an interpretation.

mohammadfawaz:
> This is unfortunate. Do we already ban returning `raw_ptr` though? I kinda feel like at that stage if we see a struct with a pointer and a `u64` then it could be specialised similarly. Maybe that's too liberal an interpretation.

The classic solution in this case is to use metadata indicating that a particular struct is special, but metadata can get lost so that's not great.

AlicanC:
> Edit. Oh you imagine `as_silce()` to be called automatically when ai write `return vector;`?

@mohammadfawaz, yup. Introducing a trait like Rust's [Termination](https://doc.rust-lang.org/std/process/trait.Termination.html) and implementing it on `Vec` is what I'm thinking of.

(Rust uses it for returning `Result`s from `main()`: https://rust-lang.github.io/rfcs/1937-ques-in-main.html)

> This is unfortunate. Do we already ban returning `raw_ptr` though? I kinda feel like at that stage if we see a struct with a pointer and a `u64` then it could be specialised similarly. Maybe that's too liberal an interpretation.

@otrho, this was the initial implementation I was going for but I personally found it too liberal, yeah. A similar thought was just doing checks like `if type.name == "Vec"` around the code without introducing `raw_slice` at all, but that gave the same feeling.

AlicanC:
Also, some questions:

1- While I wasn't sure, I thought we would want to have slices and operations on them recognized at the IR level to create more optimization opportunities. Is this not true? Some elaboration would be helpful here.

2- Where in the compiler we would make `Vec::<b256>::get(my_vec, 0)` compile into `asm(ptr: my_vec) { ptr: b256 }` and not this whole thing below?

```rs
asm(ptr: my_vec, item_size: 32, index: 0) {
  mul index index item_size;
  add ptr ptr index;
  ptr: b256
}
```

AlicanC:
Another Q: could `ref [T]` be a stack type that spans over two registers instead of being a `u64` that points to a tuple on the heap? Is that a thing?

mohammadfawaz:
> 1- While I wasn't sure, I thought we would want to have slices and operations on them recognized at the IR level to create more optimization opportunities. Is this not true? Some elaboration would be helpful here.

Regardless of optimizations, it does seem that we need special handling the IR level. At the end of the day, we are doing something novel here where we're trying to figure out a way to return a chunk of heap memory to an external context that doesn't have access to that memory. This is typically done via a "Global Memory" pointer (Think Cuda/OpenCL) but we don't have that option here because there is no global memory. And even then, those methods don't support returns except via a pointer argument pointing to the actual external memory (via a special address space). All of that to say that I don't see how we can support direct returns of heap types from scripts without additions to the IR. We could start with this approach and build on it, assuming the user-facing APIs don't have to change.
 
> 2- Where in the compiler we would make `Vec::<b256>::get(my_vec, 0)` compile into `asm(ptr: my_vec) { ptr: b256 }` and not this whole thing below?
> 
> ```rust
> asm(ptr: my_vec, item_size: 32, index: 0) {
>   mul index index item_size; // You don't need the `mul` if index <= 1
>   add ptr ptr index; // You don't need the `add` if index = 0
>   ptr: b256
> }
> ```

Well, you first need to try to constant fold (an optimization we don't have yet) the index. That plus DCE will get rid of the logic that "add"s to the pointer and so we're left with the original pointer. If we're doing things right, all of those operations will be IR instructions (and NOT `asm` blocks in IR), which would allow the IR to optimize away and by the time we get to codegen, there would be very little work to do, if any.

AlicanC:
So I've created a new PR that implements the bare minimum required to fix the fuels-ts issue: https://github.com/FuelLabs/sway/pull/3288

For the rest, I'll be working on an RFC.

PR #3253, opened 2022-11-02
Some micro-optimizations to `U128` and `U256`

### LINKED ISSUE #3245: Reenable the `u128_root_test` that was disabled in #3181
https://github.com/FuelLabs/sway/pull/3181 disables `u128_root_test` because the test was running out of gas with the new gas schedule in `fuel-core 0.13`. We can increase the gas limit but I think a better approach would be to figure out a better implementation of the `sqrt` function. I think the division operation takes too much gas now so we can change some of the divisions by 2 to shift rights.

### LINKED ISSUE #3256: Investigate expensive `mul` operation for `U256`
https://github.com/FuelLabs/sway/pull/3253 introduces some optimizations to `U128` and `U256` and enables some `U256` tests that were previously not running. The `u256_mul_test` runs out of gas so I had to disable one test case before enabling the test. We should look into whether we can make this better.

Note that `fuel-core 0.13` introduced a new gas schedule that causes most ops to go up in price so running out of gas in these tests with `fuel-core 0.13` is not surprising.

mohammadfawaz:
Once we have a fully fledged instruction combining optimization pass, these types of optimizations won't be needed anymore (for the most part). For now, this allows us to re-enable the `u128_root_test` that was disabled in #3181.
* Switch division by 2 to a shift left by 1
* Switch to using the lower part of a `U128` in situations where the upper part is obviously not impacted.

This cuts down the gas cost for `sqrt` by around 30-40% and the `u128_root_test` now consumes ~77M gas. Previously, the test was running out of gas (our default limit is 100M). 

I also made the same optimization in `U256` but notice that the `u256_div_test` was disabled (doesn't have a `test.toml`!) so I enabled. The test was actually running out of gas before my change but now it just barely passes with like 98M gas, which is dangerous.

Finally, I replaced the method of multiplication in both `U128` and `U256` and enabled two previously disabled tests:
* `u256_mul_test` which used to run out of gas but now takes ~21M (so that's a least 5 X improvement).
* `u256_ops_test` which runs fine.

In short, div and mul for both `U128` and `U256` are now much cheaper and so is `sqrt` for `U128` (`U256` doesn't have a `sqrt` yet).

Closes #3256 
Closes #3245


mohammadfawaz:
I really wanted this algorithm to work (from https://www.geeksforgeeks.org/divide-two-integers-without-using-multiplication-division-mod-operator/):
```C++
  // test down from the highest bit and
  // accumulate the tentative value for
  // valid bit
  for (int i = 31; i >= 0; --i) {
 
    if (temp + (divisor << i) <= dividend) {
      temp += divisor << i;
      quotient |= 1LL << i;
    }
  }
```
because it's much faster and consumes a lot less gas. However, as is, it gives the wrong result due to a few issues related to overflow and such. I couldn't get it to work just yet. cc @nfurfaro

Voxelot:
Maybe it would make sense to split test cases into multiple scripts for ops that are gas heavy?

mohammadfawaz:
> Maybe it would make sense to split test cases into multiple scripts for ops that are just gas heavy? Seems like we shouldn't restrict the number of test cases based on the gas limit of running all the cases in a single call.

That makes sense. A better way of testing is probably using a contract + the SDK so that each contract call gets its own gas limit and we get to test many more combinations from the Rust SDK itself which is much easier.

otrho (dismissed):
Hopefully many of these sorts of opts will be done in IR/ASM gen, even for aggregate types like these.   One day. :daydream:

mohammadfawaz:
> Hopefully many of these sorts of opts will be done in IR/ASM gen, even for aggregate types like these. One day. :daydream:

My money is on SROA! That plus some constant folding and some basic instcombine stuff like turning divisions by 2 into shifts.

sezna (approved):

Issue #3246, opened 2022-11-02
Disallow illegal opcodes in `asm` blocks in type checking instead of the parser


mohammadfawaz:
Currently, the parser disallows opcodes like `ji`, `jnei`, `ret`, etc in `asm` blocks. This is fine except that the UX is not great. The parser will just say unrecognized opcode which is not very descriptive. Instead, the parser should allow these opcodes and when analyzing an `asm` block in type checking, we would emit an error message saying that the opcode is intentionally disallowed with a clear reason.

PR #3273, opened 2022-11-03
VM instructions cleanup in the compiler and exposing `scwq`

mohammadfawaz:
I started working on https://github.com/FuelLabs/sway/issues/3246 and found myself trying to fix too many things so I decided to make a separate PR for cleanup:
* Removed some dead `asm` instructions in the parser.
* Added some missing `asm` instructions in the parser and the backend, even if they're not used.
* Sectioned the instructions everywhere by following the spec.
* Reordered the instructions everywhere to follow the order in the spec, which is alphabetical.
* I removed a `skip_fmt` annotation because the code started looking overwhelming a large and we're not really doing this anywhere else, so it kinda sticks out.
* I added back the instructions we disallowed in the parser (like `jmp`, `ret`, etc.) but commented out. They will be uncommented in my next PR which will enable them in the parser but disallow them in semantic analysis for better UX.

**This is effectively a non-functional change except for one thing which is exposing the `scwq` opcode in the parser.**


sway-core/src/asm_lang/mod.rs
```diff
@@ -1375,6 +1416,7 @@ impl fmt::Display for VirtualOp {
     fn fmt(&self, fmtr: &mut fmt::Formatter<'_>) -> fmt::Result {
         use VirtualOp::*;
         match self {
+            /* Arithmetica/Logic (ALU) Instructions */
```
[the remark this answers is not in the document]
mohammadfawaz: You'd think after copying this so many times, I'd notice the typo.. but nops

sway-core/src/asm_lang/allocated_ops.rs
```diff
@@ -67,46 +68,42 @@ pub(crate) enum AllocatedOpcode {
...
+
+    /* Conrol Flow Instructions */
+    JMP(AllocatedRegister),
     JI(VirtualImmediate24),
+    JNE(AllocatedRegister, AllocatedRegister, AllocatedRegister),
     JNEI(AllocatedRegister, AllocatedRegister, VirtualImmediate12),
     JNZI(AllocatedRegister, VirtualImmediate18),
-    JMP(AllocatedRegister),
-    JNE(AllocatedRegister, AllocatedRegister, AllocatedRegister),
     RET(AllocatedRegister),
```
otrho, line 99: If you're going to move `RETD` then you might as well move `RET`, they're pretty much the same.  I had `JMP` and `JNE` together because they're dynamic jumps while the others are all immediate.  If you're going for alpha then `JMP` comes after `JI`.

otrho (approved):

sway-core/src/asm_lang/allocated_ops.rs
mohammadfawaz: I honestly just following the order and the grouping in the spec, which is not 100% alphabetical.

PR #3274, opened 2022-11-03
Address Clippy warnings due to Rust 1.65

mohammadfawaz:
So far I'm addressing all Clippy warnings except for `result_large_err`:

Example:
```console
warning: the `Err`-variant returned from this function is very large
   --> sway-core/src/semantic_analysis/namespace/items.rs:130:56
    |
130 |     pub(crate) fn check_symbol(&self, name: &Ident) -> Result<&ty::TyDeclaration, CompileError> {
    |                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the `Err`-variant is at least 224 bytes
```
Addressing this seems like it will take some time. Might have to throw in `Box` in various places. To unblock everyone from working, I'm allowing `clippy::result_large_err` in CI for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,7 +207,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-features --all-targets -- -D warnings
+          args: --all-features --all-targets -- -D warnings --allow clippy::result_large_err
 
   cargo-toml-fmt-check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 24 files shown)


kayagokalp (approved):

emilyaherbert (approved):

mohammadfawaz:
> @mohammadfawaz Trying this locally but still seeing a ton of warnings, you only seem to have added `--allow clippy::result_large_err` to CI, but we need to deal with it locally too, right?
> 
> I was checking the docs, seems like we can add a `clippy.toml` on the root to configure clippy. Tried to disable this warning but it's not working, however we can configure the threshold, which might be a better solution.
> 
> `clippy.toml`:
> 
> ```
> large-error-threshold = 256
> ```
> 
> Seems to do the trick (128 is the default).

That seems reasonable. Would you like to do this? cc @tritao

PR #3277, opened 2022-11-03
Prevent an illegal use of a generic from crashing the compiler

### LINKED ISSUE #3270: Illegal use of a generic crashes compiler
As a victim of "Find and Replace", I have stumbled upon a case that crashes the compiler.

The illegal use of trying to call a function from a generic such as `new()`  will make the compiler panic:

```
    fn test_function(self, value: T) {
        T::new();
    }
```
illegal_use_of_generic

emilyaherbert:
This PR removes an `unreachable!` that we had inside one of the impl blocks for `Spanned`.

In implementing this PR, I noticed that we hadn't been using `ty::TyDeclaration::ErrorRecovery` anywhere! Rather than remove it, I've inserted places where we now use it, to hopefully remove some cascading errors.

Closes #3270


mohammadfawaz (approved):
Seems reasonable. Thanks!.

Question: what is the reason for changing `nextln` to `check` the `test.toml` files you've updated?

emilyaherbert:
> Question: what is the reason for changing `nextln` to `check` the `test.toml` files you've updated?

The `nextln:` directive appears to have some whitespace sensitivity issues that were causing it to fail with the updated errors being reported. https://github.com/bytecodealliance/filecheck/issues/2

sway-core/src/semantic_analysis/ast_node/declaration/mod.rs
```diff
@@ -1,4 +1,6 @@
 mod abi;
+#[allow(clippy::module_inception)]
+mod declaration;
```
otrho, line 3: In the source on my machine this is still in `ast_node/declaration.rs` and it seems like the new `declaration.rs` code taken from `ast_node/mod.rs` could go in there or here in this `mod.rs` since it's so small and the module inception could be avoided.

otrho (approved):
Nit on the code reorg, but cool otherwise.

PR #3279, opened 2022-11-04
Disallow control flow opcodes in `asm` blocks

### LINKED ISSUE #3246: Disallow illegal opcodes in `asm` blocks in type checking instead of the parser

mohammadfawaz:
Closes #3246
These are now disallowed in type checking instead of parsing because that's better UX.


mohammadfawaz:
This is based of https://github.com/FuelLabs/sway/pull/3273.

otrho (approved):

vaivaswatha (approved):

PR #3280, opened 2022-11-04
Prevent stack overflow when defining a trait method with certain types

### LINKED ISSUE #3271: When trait and implementation functions differ by an `Option` with a generic, stack overflow occurs
When implementing a trait function definition where the function implementation has an `Option` and the definition does not, a stack overflow occurs. 

This can be seen in the following example:

```
pub struct MyStruct<T> {
    val: T
}

pub trait MyTrait<T> {
    fn set(self, val: T);
}

impl<T> MyTrait<T> for MyStruct<T> {
    // This implementation uses an Option, but the definition does not
    fn set(self, val: Option<T>) {
        
    }
}
```

option_stack_overflow

___________________________________________________________________________________________________________________________________

**NOTE:** Only including the `Option` in the implementation causes the stack overflow, the following code errors as expected:

```
pub struct MyStruct<T> {
    val: T
}

pub trait MyTrait<T> {
    fn set(self, val: Option<T>);
}

impl<T> MyTrait<T> for MyStruct<T> {
    // This implementation does not use an Option, but the definition does
    fn set(self, val: T) {

    }
}
```

expected_error_trait

This has been tested with forc v0.30.0

emilyaherbert:
This PR fixes a bug that was causing a stack overflow when a trait method impl differentiated from a trait method definition. The fix was to draw a direct comparison between the expected type (the type from the definition) and the found type (the type from the impl) instead of trying to unify the found type into the expected type.

Closes #3271


sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs
```diff
@@ -687,6 +683,23 @@ fn type_check_trait_implementation(
                 });
                 continue;
             }
+
+            // let (new_warnings, new_errors) = unify_right_with_self(
```
mohammadfawaz, line 687: Is this code meant to be uncommented later?

sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs
```diff
@@ -728,6 +737,23 @@ fn type_check_trait_implementation(
             continue;
         }
 
+        // let (new_warnings, new_errors) = unify_right_with_self(
```
mohammadfawaz, line 740: Same question about this commented code.

mohammadfawaz (dismissed):

otrho:
@emilyaherbert Got a lock file thing.

mohammadfawaz:
@emilyaherbert converting to a draft until this is 100% ready for review and conflicts are resolved.

otrho (approved):
utACK

vaivaswatha (approved):

PR #3285, opened 2022-11-04
Tests revert in an if branch

### LINKED ISSUE #2876: `revert` in an `if` branch causes an IR verification error

esdrubal:
This is a test that was posted in #2876

This test case now runs without causing the internal error mentioned in the issue.

Pushing this test case which appears not to be covered in the existing ones.

Closes #2876

### DIFF
--- a/test/src/e2e_vm_tests/test_programs/should_pass/stdlib/if_type_revert/Forc.lock
+++ b/test/src/e2e_vm_tests/test_programs/should_pass/stdlib/if_type_revert/Forc.lock
@@ -4,7 +4,7 @@ source = 'path+from-root-6FF6432D2813E4FB'
 
 [[package]]
 name = 'if_type_revert'
-source = 'root'
+source = 'member'
 dependencies = ['std']
 
 [[package]]
--- a/test/src/e2e_vm_tests/test_programs/should_pass/stdlib/if_type_revert/Forc.lock
+++ b/test/src/e2e_vm_tests/test_programs/should_pass/stdlib/if_type_revert/Forc.lock
@@ -1,0 +1,13 @@
+[[package]]
+name = 'core'
+source = 'path+from-root-6FF6432D2813E4FB'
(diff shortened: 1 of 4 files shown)


kayagokalp:
Hey, it seems like the test is built without fetching https://github.com/FuelLabs/sway/commit/76b1ee9d4cd2500898f702e4725d483e71b23a46, so it fails as the lock file is outdated. Building `forc` from master and rebuilding the test should fix the CI failure

esdrubal:
> Hey, it seems like the test is built without fetching [76b1ee9](https://github.com/FuelLabs/sway/commit/76b1ee9d4cd2500898f702e4725d483e71b23a46), so it fails as the lock file is outdated. Building `forc` from master and rebuilding the test should fix the CI failure

Thank you for letting me know what the issue was.

mohammadfawaz:
I bet https://github.com/FuelLabs/sway/pull/3210 fixed it :) Thanks!

mohammadfawaz (approved):

kayagokalp (approved):

PR #3288, opened 2022-11-04
Kinda allow returning `Vec`s from scripts

AlicanC:
This PR is a minimal cut of https://github.com/FuelLabs/sway/pull/3251 that just implements the bare minimum required to fix https://github.com/FuelLabs/fuels-ts/issues/467:
- A new type `raw_slice` and support for returning it from scripts.
- Support for conversion between `raw_slice` and `Vec`.

```rs
script;

fn main() -> raw_slice {
    // Create a vec
    let vec = Vec::new();
    vec.push(42);
    vec.push(1337);

    // Return it
    vec.as_slice()
}
```


mohammadfawaz:
To be honest, I feel like this is a perfectly reasonable use model that addresses the user's requirements with almost no compiler magic involved. There is an actual conversion needed when returning a `Vec` from a script and there is no reason to hide that from users. When someone returns a `Vec` directly, they should just get what they ask for (or not allow them as we do today), but when the want the heap data, have them make it explicit via `as_slice()`. I guess what I'm suggesting is that this is a good enough solution and, in fact, the best solution.

This also lets users do something like `return (vec1.as_slice(), vec2.as_slice())` with no issues.

sway-lib-std/src/vec.sw
```diff
@@ -248,3 +249,27 @@ impl<T> Vec<T> {
...
+        Self {
+            buf,
+            len: buf.cap,
+        }
+    }
+
+    fn into(self) -> raw_slice {
+        asm(ptr: (self.buf.ptr(), self.len)) { ptr: raw_slice }
+    }
+}
```
mohammadfawaz, line 275: I like all of this.

mohammadfawaz: I also like that `Vec` doesn't have to change. It's nice to keep it somewhat in sync with how Rust's `Vec` looks like.

mohammadfawaz: Also this is a non-breaking change so that's a bonus :D

AlicanC: Yeah I was so happy to be able to avoid that :D

AlicanC:
> This also lets users do something like `return (vec1.as_slice(), vec2.as_slice());` with no issues (or does it? if not, can we make this work? I feel like we can).

This is the tricky part. If you do that you'll get an error as tested here: https://github.com/FuelLabs/sway/pull/3288/files#diff-c298aeb7cf82037eecb5c67af39936b08be7074d518d536086b999e8f8d46635R4

The question "How do you serialize variable length data?" is not answered anywhere in spec or implementation. This is also why there is no `raw_slice` support for `main()` args where we would have to support multiples. We can get away with returning a single one because it just fits in a RETD.

I think we first need to add DSTs to the spec and define how to serialize them. ([Ethereum's SSZ](https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md#vectors-containers-lists) can do this.) I'll mention this in the RFC.

mohammadfawaz:
> > This also lets users do something like `return (vec1.as_slice(), vec2.as_slice());` with no issues (or does it? if not, can we make this work? I feel like we can).
> 
> This is the tricky part. If you do that you'll get an error as tested here: https://github.com/FuelLabs/sway/pull/3288/files#diff-c298aeb7cf82037eecb5c67af39936b08be7074d518d536086b999e8f8d46635R4
> 
> The question "How do you serialize variable length data?" is not answered anywhere in spec or implementation. This is also why there is no `raw_slice` support for `main()` args where we would have to support multiples. We can get away with returning a single one because it just fits in a RETD.
> 
> I think we first need to add DSTs to the spec and define how to serialize them. ([Ethereum's SSZ](https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md#vectors-containers-lists) can do this.) I'll mention this in the RFC.

Ah of course.. the tricky part is decoding the data on the other side.. I keep forgetting about that 😆.

otrho (approved):
Hrmm, I still think that we could just special case a struct with a pointer and len for `RETD`.  It wouldn't make sense that 'sometimes' we want to `RETD` that struct literally, since returning a pointer doesn't make sense.

But that isn't something we couldn't switch to later on if need be.

vaivaswatha (approved):

sway-lib-core/src/raw_slice.sw
```diff
@@ -0,0 +1,34 @@
+library raw_slice;
+
+dep raw_ptr;
+
+use ::raw_ptr::*;
+
+pub trait AsSlice {
+    fn as_slice(self) -> raw_slice;
+}
```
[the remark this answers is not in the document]
AlicanC: What would the use for `raw_slice` be when we have `&[T]`? I thought it would just replace `raw_slice` so I even removed the space I was leaving for it :D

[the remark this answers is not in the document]
AlicanC: IIUC, we don't want `raw_ptr` or `raw_slice` in Sway because they are too unsafe/low level, but we are just implementing them to unblock certain things until we can get an RFC on something better. I personally haven't worked on any Sway code that you'd want these instead of their typed variants.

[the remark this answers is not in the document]
AlicanC: TBH, I don't even want this documented, but naming it `raw` would make that less likely so I've made the changes :D

otrho (approved):

PR #3289, opened 2022-11-04
Fixes "Unknown attribute: inline." warnings.

### LINKED ISSUE #3286: Inline annotation gives a warning `Unknown attribute: "inline".`
The annotation _does_ work but gives a warning.

esdrubal:
When #[inline(always)] or #[inline(never)] were used the warning above would be displayed.

Closes #3286


mohammadfawaz (approved):
Great, thanks!

PR #3291, opened 2022-11-04
Adds more opcodes to check_invalid_opcodes for predicates.

### LINKED ISSUE #2550: Enhance purity checking to include specific checks for predicates

esdrubal:
All the contract specific opcodes now throw an error if used within a predicate. Except for RVRT which is used currently in tests.

Closes #2550

### DIFF
--- a/sway-core/src/asm_generation/checks.rs
+++ b/sway-core/src/asm_generation/checks.rs
@@ -14,15 +14,14 @@ use sway_error::error::CompileError;
 pub fn check_invalid_opcodes(asm: &FinalizedAsm) -> CompileResult<()> {
     match asm.program_kind {
         ProgramKind::Contract | ProgramKind::Library => ok((), vec![], vec![]),
-        ProgramKind::Script | ProgramKind::Predicate => {
-            check_for_contract_opcodes(&asm.program_section.ops[..])
-        }
+        ProgramKind::Script => check_script_opcodes(&asm.program_section.ops[..]),
+        ProgramKind::Predicate => check_predicate_opcodes(&asm.program_section.ops[..]),
     }
 }
 
-/// Checks if an opcode is one that can only be executed from within a contract. If so, throw an
-/// error.
-fn check_for_contract_opcodes(ops: &[AllocatedOp]) -> CompileResult<()> {
+/// Checks if an opcode is one that cannot be executed from within a script.
+/// If so, throw an error.
+fn check_script_opcodes(ops: &[AllocatedOp]) -> CompileResult<()> {
     use AllocatedOpcode::*;
     let default_span =
         sway_types::span::Span::new("no span found for opcode".into(), 0, 1, None).unwrap();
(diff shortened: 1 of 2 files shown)


esdrubal:
Do we want to disallow RVRT in predicates and remove it usage from tests?

mohammadfawaz:
According to https://fuellabs.github.io/fuel-specs/master/vm/index.html?highlight=predicate#predicate-verification, all contract instructions (including `rvrt`) should be disallowed. In addition, backward jump instructions are also not allowed:
* Jump immediate instructions should be relatively easy to check.
* Dynamic jump instructions are a bit harder (`jmp` and `jne`). I think we just have to disallow them completely.

mohammadfawaz:
> 

+1 to @sezna's comment. I had assumed we can blindly follow the spec here, but there is a subtlety regarding `rvrt` related to the standard library.

esdrubal:
New commits applies all @Centril requested changes.

Added test that uses every testable opcode and checks for message errors.

The RVRT opcode was not included in the list of disallowed opcodes.

sway-core/src/asm_generation/checks.rs
```diff
@@ -68,6 +75,115 @@ fn check_for_contract_opcodes(ops: &[AllocatedOp]) -> CompileResult<()> {
...
+            }
+            JI(imm) => {
+                if imm.value <= opcode_index {
+                    invalid_backward_jump!("JI");
+                }
+            }
+            JMP(..) => invalid_opcode!("JMP"),
+            JNE(..) => invalid_opcode!("JNE"),
+            JNEI(_, _, imm) => {
+                if u32::from(imm.value) <= opcode_index {
```
mohammadfawaz, line 156: Is index of the `opcode` in the list `ops` the same as what jumps instructions use? Maybe that's true for `AllocatedOp` but I'll let @otrho confirm.

test/src/e2e_vm_tests/test_programs/should_fail/predicate_invalid_opcodes/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+license = "Apache-2.0"
+name = "predicate_invalid_opcodes"
+entry = "main.sw"
+
+[dependencies]
+core = { path = "../../../../../../sway-lib-core" }
```
mohammadfawaz, line 8: Not really needed. Instead, add `implicit-std = false` under `[project]` above (and then rerun `forc fmt`)

esdrubal: With `implicit-std = false`, I get the following error:
predicate_invalid_opcodes
The while loop at the end of the test is using operators from core::ops, so it looks like at least core is necessary.

mohammadfawaz: 🤦‍♂️ sorry you're right!

sway-core/src/asm_generation/checks.rs
otrho: Yeah, nah, it's an address at this stage.

The `VirtualOpcode`s and `AllocatedOpcode`s have addresses as args, though I don't think we use the `VirtualOpcode` ones.  Prior to address realisation we have `ControlFlowOp` which does jumping to symbolic labels.  So we're never jumping to an index in an array.

To catch backwards jumps you'll need to know the address of the jump instruction.

esdrubal: @otrho what do you think of changing [map_label_offsets](https://github.com/FuelLabs/sway/blob/a781aa4d990472cf664dc4c5a330988bbb0aa237/sway-core/src/asm_generation/allocated_abstract_instruction_set.rs#L307) to also set a field with the address of opcodes. The new field `address` would be in `AllocatedAbstractOp`, `RealizedOp` and `AllocatedOp`.

The other option we have is to compute address in place but this would duplicate the places we handle with offsets.

otrho: It only makes sense to perhaps add the address to `RealizedOp`, not the others.

But I've taken a closer look, and at the point where you're making these checks you have a fully formed binary blob which is basically an array of 32bit instructions, so calculating the address of each one is trivial -- not that different to what you're doing now, as you should be able to take the index of the instruction you're checking and just mul by 4.  I think?

esdrubal: In the opcode docs it says for [JI](https://fuellabs.github.io/fuel-specs/master/vm/instruction_set.html#ji-jump-immediate) that `$pc = $is + imm * 4;` imm is multiplied by 4 to update the address so maybe we can compare imm with opcode_index directly.

otrho: Oh, actually, I think you're right!  So at this stage, when everything is fully baked, the address _is_ the index and the code was OK all along.  Sorry.  It _might_ be more useful to rename `opcode_index` to `opcode_addr` just to document this.

otrho (approved):

mohammadfawaz (dismissed):
Great! Thanks :)

sway-core/src/asm_generation/checks.rs
```diff
@@ -67,61 +75,115 @@ fn check_script_opcodes(ops: &[AllocatedOp]) -> CompileResult<()> {
...
+            ($name_str:literal) => {{
+                errors.push(CompileError::InvalidBackwardJumpFromPredicate {
+                    opcode: $name_str.to_string(),
+                    span: op
+                        .owning_span
+                        .clone()
+                        .unwrap_or_else(|| default_span.clone()),
+                });
+            }};
+        }
```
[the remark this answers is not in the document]
esdrubal: I tried to change this to closures but both of them are borrowing `errors`.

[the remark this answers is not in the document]
esdrubal: Changed most recent version to use closures.

mohammadfawaz (approved):

Issue #3060, opened 2022-10-18
Refactor the E2E test config to not bail out on the first failure


mohammadfawaz:
It would be much better if there was some error recovery so that if one test fails, we continue to run the other tests, and at the end, we just list all the failing tests. This is similar to what `cargo test` does.

mohammadfawaz:
> I will see about cleaning it all up and send a PR once I got some free cycles.

Oh nice. That would be extremely helpful :)

mohammadfawaz:
@tritao note https://github.com/FuelLabs/sway/pull/3079 - not sure if there's any overlap between your work and #3079 but just an FYI.

mohammadfawaz:
I do not yet if #3709 is addressing error recovery or not.

PR #3294, opened 2022-11-04
Refactor E2E CLI runner.

tritao:
Now output handling is done in an uniform manner for all kinds of tests and makes it possible to implement better developer experience.

Now the output is hidden by default for all tests, and only shown up when a particular test is failing (indented so its easier to parse).

Failure handling is also more robust, with all tests being executed by default, and failures being reported as they happen and in the summary report at end of execution. There are still a couple panics here and there for more esoteric errors, but those can be easily be transformed to errors in the future as need arises.

Previous behavior of showing all output can be obtained by passing the `--verbose` flag to the runner.

The only exception are the IR tests which still panic. However their summary report was updated to match the improvements from above.

Closes https://github.com/FuelLabs/sway/issues/3060.

Example of new output with failed test report because `fuel-core` is not running:

Screenshot from 2022-11-09 17-53-05

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4119,6 +4119,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a507befe795404456341dfab10cef66ead4c041f62b8b11bbb92bffe5d0953e0"
 
 [[package]]
+name = "smawk"
+version = "0.3.1"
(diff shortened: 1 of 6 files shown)


mohammadfawaz (dismissed):
Amazing, thank you! This will make things so much easier.

otrho (approved):

mohammadfawaz (approved):

PR #3307, opened 2022-11-06
Update lock files for all examples

mohammadfawaz:
And adding a missing semicolon in `subcurrency`.

### DIFF
--- a/examples/arrays/Forc.lock
+++ b/examples/arrays/Forc.lock
@@ -1,6 +1,6 @@
 [[package]]
 name = 'array'
-source = 'root'
+source = 'member'
 dependencies = ['std']
 
 [[package]]
--- a/examples/asm_return_tuple_pointer/Forc.lock
+++ b/examples/asm_return_tuple_pointer/Forc.lock
@@ -1,6 +1,6 @@
 [[package]]
 name = 'asm_return_tuple_pointer'
-source = 'root'
+source = 'member'
 dependencies = ['std']
 
 [[package]]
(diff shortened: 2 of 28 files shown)


kayagokalp (approved):

mitchmindtree (approved):
Good catch!

Perhaps we can add a `--locked` flag to `scripts/examples-checker` to make sure these don't get missed in the future?

Specifically the 
```rust
fn run_forc_build(path: &Path) -> bool {
    run_forc_command(path, &["build", "--path"])
}
```
could be changed to:
```rust
fn run_forc_build(path: &Path) -> bool {
    run_forc_command(path, &["build", "--locked", "--path"])
}
```
Approving either way in case we want to leave it for a follow-up PR :)

mohammadfawaz:
> Good catch!
> 
> 
> 
> Perhaps we can add a `--locked` flag to `scripts/examples-checker` to make sure these don't get missed in the future?
> 
> 
> 
> Specifically the 
> 
> ```rust
> 
> fn run_forc_build(path: &Path) -> bool {
> 
>     run_forc_command(path, &["build", "--path"])
> 
> }
> 
> ```
> 
> could be changed to:
> 
> ```rust
> 
> fn run_forc_build(path: &Path) -> bool {
> 
>     run_forc_command(path, &["build", "--locked", "--path"])
> 
> }
> 
> ```
> 
> Approving either way in case we want to leave it for a follow-up PR :)

Ah good idea. Let me do that in a follow up PR asap.

Issue #3321, opened 2022-11-08
Nested generic type inference error 


mohammadfawaz:
Trying to implement `transpose()` from https://doc.rust-lang.org/src/core/option.rs.html#1864 in Sway as follows:
```rust
impl<T, E> Option<Result<T, E>> {
    pub fn transpose(self) -> Result<Option<T>, E> {
        match self {
            Option::Some(Result::Ok(x)) => Result::Ok(Option::Some(x)),
            Option::Some(Result::Err(e)) => Result::Err(e),
            Option::None => Result::Ok(Option::None),
        }
    }
}
```
compiles fine in the library but when I try to use it, it gives me weird errors:
Use case:
```rust
let y: Option<Result<u32, u8>> = Option::Some(Result::Ok(5));
let z = y.transpose();
```
gives a list of errors as follows:
![image](https://user-images.githubusercontent.com/59666792/200469069-43b891b1-0f39-4e8e-a159-307ab3b2e578.png)

I also tried:
```rust
let z = Option<Result<u32, u8>>::transpose(y);
```
and that gave me:
![image](https://user-images.githubusercontent.com/59666792/200469182-f7ad5ff6-04dd-4cda-b22b-9f9076c30c1f.png)

@emilyaherbert I'm not sure much we're expecting out of the type system at the moment with regards to patterns like the one above. Is this something we're meant to support or is this a bug?

PR #3322, opened 2022-11-08
Enhance docs and add a few methods for `option.sw`

mohammadfawaz:
A lot of this is copied from https://doc.rust-lang.org/src/core/option.rs.html#1864

Now that we have `forc-doc`, we want to spend some time enhancing all of our standard library docstrings.

There are a few more methods we can implement. I attempted `transpose()` but faced an [issue](https://github.com/FuelLabs/sway/issues/3321). There are all the Boolean methods such as `and` and `or` as well.

- [x] Write tests for `unwrap_or` and `ok_or`.
- [x] Enhance tests for `Option` in general


sway-lib-std/src/option.sw
```diff
@@ -1,53 +1,221 @@
...
     /////////////////////////////////////////////////////////////////////////
-    /// Returns `true` if the result is [`Some`].
-    fn is_some(self) -> bool {
+
+    /// Returns `true` if the option is a [`Some`] value.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// let x: Option<u32> = Some(2);
```
SwayStar123, line 105: I dont think Sway has the variants of Option in its prelude, so `Option::Some(2)` would be more accurate

mohammadfawaz: Yup you're right.. I was fixing those but clearly forgot this one. In fact, we can't import enum variants yet either which is unfortunate, but that's coming!

sway-lib-std/src/option.sw
```diff
@@ -1,18 +1,86 @@
-//! Error handling with the `Option` type.
+//! Optional values.
 //!
-//! [`Option<T>`][`Option`] is the type used for representing the existence or absence of a value. It is an enum with the variants, [`Some(T)`], representing
-//! some value, and [`None()`], representing
-//! no value.
+//! Type [`Option`] represents an optional value: every [`Option`]
+//! is either [`Some`] and contains a value, or [`None`], and
+//! does not. [`Option`] types are very common in Rust code, as
```
AlicanC, line 5: Should this say "Rust" or "Sway"?

AlicanC (approved):

sway-lib-std/src/option.sw
mohammadfawaz: Ah of course. Updated.

sway-lib-std/src/option.sw
```diff
@@ -1,18 +1,86 @@
...
 //!
-//! [`Option<T>`][`Option`] is the type used for representing the existence or absence of a value. It is an enum with the variants, [`Some(T)`], representing
-//! some value, and [`None()`], representing
-//! no value.
+//! Type [`Option`] represents an optional value: every [`Option`]
+//! is either [`Some`] and contains a value, or [`None`], and
+//! does not. [`Option`] types are very common in Sway code, as
+//! they have a number of uses:
+//!
+//! * Initial values
```
emilyaherbert, line 8: Maybe expand this to "Initial values can be set to [`None`]"  or something like that.

sway-lib-std/src/option.sw
```diff
@@ -1,18 +1,86 @@
...
+//! ```
+//! fn divide(numerator: u64, denominator: u64) -> Option<u64> {
+//!     if denominator == 0 {
+//!         Option::None
+//!     } else {
+//!         Option::Some(numerator / denominator)
+//!     }
+//! }
+//!
+//! // The return value of the function is an option
```
emilyaherbert, line 26: Maybe put the remaining block of code in its own function, like `call_divide` or something like that?

emilyaherbert (dismissed):
Very well written!

emilyaherbert (approved):

otrho (approved):

PR #3323, opened 2022-11-08
Document workspaces in The Sway Book

### LINKED ISSUE #1777: Add support for `[workspace]`s to `forc`

### LINKED ISSUE #3272: Document workspace support in the Sway Book
After following PRs we will have working workspace building/deploying/checking/updating. We should have some explanation and couple of examples for them:

1. https://github.com/FuelLabs/sway/pull/3134
2. https://github.com/FuelLabs/sway/pull/3146
3. https://github.com/FuelLabs/sway/pull/3248
4. https://github.com/FuelLabs/sway/pull/3255 - closed as already fixed

kayagokalp:
closes #3272.
closes #1777.

blocked by:
- [x] #3248


docs/src/forc/workspaces.md
```diff
@@ -0,0 +1,29 @@
+# Workspaces
+
+A *workspace* is a collection of one or more packages, namely *workspace members*, that are managed together.
+
+The key points for workspaces are:
+
+* Common `forc` commands available for a single package can also be used for a workspace, like `forc build` or `forc deploy`.
+* All packages share a common [`Forc.lock`] file which resides in the root directory of the workspace.
+
+For workspaces, `Forc.toml` accepts following fields:
```
mitchmindtree, line 10: Workspace manifests are declared within `Forc.toml` files and support the following fields:

docs/src/forc/workspaces.md
```diff
@@ -0,0 +1,29 @@
...
+
+## The `members` field
+
+The `members` field define which packages are members of the workspace:
+
+```toml
+members = ["member1", "path/to/member2"]
+```
+
+The `members` field accepts entries to be given in relative path with respect to the workspace root.
```
mitchmindtree, line 22:
```suggestion
The `members` field accepts entries to be given in relative path with respect to the workspace root.
Packages that are located within a workspace directory but are *not* contained within the `members` set are ignored.
```

mitchmindtree (commented):
Just a few nitpicks, otherwise I think this is a good start! We can flesh out descriptions about build order, deployment behaviour etc in future PRs.

mitchmindtree (dismissed):

mohammadfawaz (commented):
Just some nits!

mohammadfawaz (approved):

JoshuaBatty (approved):

PR #3331, opened 2022-11-09
Compiler error when using private methods of library.

### LINKED ISSUE #3327: Impl self methods should respect the `pub` keyword
Currently, the `pub` keyword is ignored for impl self methods. Example

Library code:
```rust
library lib;

pub struct MyStruct {
    x: u64,
}

impl MyStruct {
    fn foo(self) {}

    fn bar() {}
}
```
Script code:
```rust
script;

dep lib;

use lib::*;

fn main() {
    MyStruct { x: 42 }.foo();
    MyStruct::bar();
}
```
This compiles fine. Instead, it should complain that `foo` and `bar` are not `pub`.

esdrubal:
The compiler now returns an error when using private methods of impl defined in another file.

Closes #3327

### DIFF
--- a/sway-core/src/control_flow_analysis/dead_code_analysis.rs
+++ b/sway-core/src/control_flow_analysis/dead_code_analysis.rs
@@ -265,7 +265,12 @@ fn connect_node(
         ty::TyAstNodeContent::SideEffect => (leaves.to_vec(), exit_node),
         ty::TyAstNodeContent::Declaration(decl) => {
             // all leaves connect to this node, then this node is the singular leaf
-            let decl_node = graph.add_node(node.into());
+            let cfg_node: ControlFlowGraphNode = node.into();
+            // check if node for this decl already exists
+            let decl_node = match graph.get_node_from_decl(&cfg_node) {
+                Some(node) => node,
+                None => graph.add_node(cfg_node),
+            };
             for leaf in leaves {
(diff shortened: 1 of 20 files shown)


mohammadfawaz:
Thanks for doing this! I think we're violating this requirement in the standard library. For example, https://github.com/FuelLabs/sway/blob/e7f4bd3b61b1085913da1b918cc7256dcdf94ad3/sway-lib-std/src/result.sw#L28 should be marked `pub`. Do you mind going through the `core` and `std` and making sure all methods are updated? Note that impl trait methods should not be updated, only impl self.

esdrubal:
> Do you mind going through the `core` and `std` and making sure all methods are updated?

Ok, I will do that.

esdrubal:
`test/src/sdk-harness/test_projects/evm_ec_recover` is fetching lib-std from github, any reason for it to not do `implicit-std=false` and using local std dependency?

mohammadfawaz:
> `test/src/sdk-harness/test_projects/evm_ec_recover` is fetching lib-std from github, any reason for it to not do `implicit-std=false` and using local std dependency?

Nops, that's probably an oversight. Feel free to change :)

mohammadfawaz:
Merge conflicts :(

mohammadfawaz (approved):

PR #3336, opened 2022-11-09
Simplify compose and decompose

nfurfaro:


sway-lib-core/src/ops.sw
```diff
@@ -621,32 +621,12 @@ fn rsh_with_carry(word: u64, shift_amount: u64) -> (u64, u64) {
...
-    let res: b256 = 0x0000000000000000000000000000000000000000000000000000000000000000;
-    asm(w1: word_1, w2: word_2, w3: word_3, w4: word_4, result: res) {
-        sw result w1 i0;
-        sw result w2 i1;
-        sw result w3 i2;
-        sw result w4 i3;
-        result: b256
-    }
+/// Build a single b256 value from a tuple of 4 u64 values.
+fn compose(words: (u64, u64, u64, u64)) -> b256 {
```
mohammadfawaz, line 625: I was worried for a second that this would not work if we call `compose` with literals as in `compose((0, 0, 0, 0))`, but I think that's fine given that `(0, 0, 0, 0)` is a reference type. However, will you please double check? :)

otrho: At this stage we can assume the words are contiguous and in the same order in memory, but I'm not sure we should have an ASM block depending on that behaviour.  In this specific case it's so simple I think it's probably OK.

sway-lib-core/src/ops.sw
```diff
@@ -621,32 +621,31 @@ fn rsh_with_carry(word: u64, shift_amount: u64) -> (u64, u64) {
...
-    let w2 = get_word_from_b256(val, 8);
-    let w3 = get_word_from_b256(val, 16);
-    let w4 = get_word_from_b256(val, 24);
-    (w1, w2, w3, w4)
+#[test]
+fn test_decompose()  {
+    let initial: b256 = 0x0000000000000001_0000000000000002_0000000000000003_0000000000000004;
+    let expected = (1, 2, 3, 4);
+    let decomposed = decompose(initial);
+    if decomposed.0.neq(expected.0) && decomposed.1.neq(expected.1) && decomposed.2.neq(expected.2) && decomposed.3.neq(expected.3) {
```
mohammadfawaz, line 648: BTW, is this formatted?

mohammadfawaz (approved):

sway-lib-core/src/ops.sw
nfurfaro: It not formatted by the formatter, but is formatted consistently as compared to https://github.com/FuelLabs/sway/blob/7a8ab39e0881e298e763d50a7436ac4cae5db325/sway-lib-std/src/u256.sw#L213

otrho (approved):

PR #3339, opened 2022-11-10
Fix for requiring data sections with more than 2^12 entries.

### LINKED ISSUE #3161: Investigate issue with data section addressing limits.
@vaivaswatha mentioned that while implementing `mem2reg` he was testing against https://github.com/sway-libs/amm (which has become the de facto gargantuan library for testing Sway performance) and it failed:

```
'Unable to offset into the data section more than 2^12 bits. Unsupported data section length.', sway-core/src/asm_lang/allocated_ops.rs:506:19.
```

The data section uses a single register as a base and indexes elements using `lw`, which has a 12-bit index immediate.  If the data section gets too large then this won't work any more.

We should look at both allowing a large DS and also optimising the DS for size by removing any redundancy.

- Having multiple data sections isn't out of the question.  Loading their offsets at runtime would add a minor cost.
- As the data section is read-only there should never be any repeated values as they can be safely re-used.  (This should already be the case with the current compiler, but we should double check.)
- Word values like zero or one should never be in the data section as they can be taken directly from `$zero` and `$one`.
- Values smaller than 2^18 should never be in the data section as they can be created with `MOVI`.

otrho:
Adding this WIP now to run it through CI.

Closes #3161.


otrho:
I'm abandoning this approach in favour of fixing #3347 and #3346 first.

PR #3340, opened 2022-11-10
Use `fuels-types` instead of `sway-types` for JSON ABI types

mohammadfawaz:
* Replace `sway-types` with `fuels-types` when forming the JSON ABI.
* Update how type de-duplication in done in `forc-pkg` to not use a `HashSet` because `fuels_types::TypeDeclaration` does not implement `Hash`. I think the new solution is fine but we can consider implementing `Hash` for `fuels_types::TypeDeclaration` if we see any performance issues.


forc-pkg/src/pkg.rs
```diff
@@ -2556,22 +2556,30 @@ pub fn build(
...
+        let mut deduped_types: Vec<fuels_types::TypeDeclaration> = Vec::new();
+
+        // Insert values in `deduped_types` if they haven't been inserted before. Otherwise, create
+        // an appropriate mapping between type IDs in the HashMap `old_to_new_id`.
+        for decl in json_abi_program.types.iter() {
+            if let Some(ty) = deduped_types.iter().find(|d| {
+                d.type_field == decl.type_field
+                    && d.components == decl.components
+                    && d.type_parameters == decl.type_parameters
+            }) {
```
otrho, line 2579: Does `fuels_types::TypeDeclaration` need `Hash`?  Is it too fiddly to introduce?

otrho (dismissed):

forc-pkg/src/pkg.rs
mohammadfawaz: Yeah we'd need to have custom implementations of `Hash` and `PartialEq` in `fuels-types` (like I had in `sway-types`).

emilyaherbert (approved):
LGTM

otrho:
Conflicts @mohammadfawaz

mohammadfawaz:
> Conflicts @mohammadfawaz

Fixed.

otrho (approved):

sezna (approved):

PR #3344, opened 2022-11-10
Fixes method Eq not found when used with where clause.

### LINKED ISSUE #3324: Small trait constraints and the `Eq` trait while testing `ok_or`
Code example:
```rust
script;

use core::ops::*;

impl<T> Option<T> {
        match self {
        }
    }
}

fn test_ok_or<T, E>(val: T, default: E) where T: Eq {
    match Option::Some(val).ok_or(default) {
        Result::Ok(inner) => assert(inner == val),
        Result::Err(_) => revert(0),
    }
}

fn main() {}
```
Error:
```console
15 |     match Option::Some(val).ok_or(default) {
16 |         Result::Ok(inner) => assert(inner == val),
   |                                           ^^ No method named "eq" found for type "T".
17 |         Result::Err(_) => revert(0),
18 |     }
```
Interestingly, when I change the assert to `assert(val == inner)` (just flip the values), the script compiles fine.

esdrubal:
Solution was to change how `TypeInfo::UnknownGeneric` are compared in `are_equal_minus_dynamic_types`. It now compares the span of the name which is unique in the source code. If the span is equal then it assumes it is safe to consider the TypeIds refer to the same generic parameter and that using the methods of one another is also safe.

Closes #3324

### DIFF
--- a/sway-core/src/semantic_analysis/namespace/trait_map.rs
+++ b/sway-core/src/semantic_analysis/namespace/trait_map.rs
@@ -651,7 +651,6 @@ fn are_equal_minus_dynamic_types(left: TypeId, right: TypeId) -> bool {
         (TypeInfo::Unknown, TypeInfo::Unknown) => false,
         (TypeInfo::SelfType, TypeInfo::SelfType) => false,
         (TypeInfo::Numeric, TypeInfo::Numeric) => false,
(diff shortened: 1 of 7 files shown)


esdrubal:
Some test are failing, it appears this does not work because generic parameters with the same source span cannot be used interchangeably between monomorphized generic types.

esdrubal:
Closing for now as this appears to not be possible to solve this way.

PR #3345, opened 2022-11-10
Add `forc edit`

### LINKED ISSUE #2369: Add a subcommand to forc to `edit` third party dependencies similar to Cargo
## Summary

Cargo allows you to [add, remove and upgrade](https://github.com/killercup/cargo-edit) dependencies.

It would be nice, and has been requested externally, that a similar command is added to `forc`.
Moreover, it would be nice to allow the user to search available dependencies which can then be selected and automatically added to the manifest file so that the user does not have to manipulate the dependencies themselves.

gr00vytvniks:
WIP: still trying to figure out clap, need to add logic for adding and removing dependencies.

Closes #2369


eureka-cpu:
I think you could separate the `edit.rs` file into an `add.rs` and `remove.rs` to fix the error you're having with the `Args` trait. This would also be in line with the already existing `update.rs`. You'd also have to do the same the existing functions for `add` and `remove`, and the commands in the `cli` mod file. Other than that I think this is looking good!

forc/src/cli/commands/add.rs
```diff
@@ -0,0 +1,37 @@
...
+    forc remove [OPTIONS] <DEP_PATH> ...
+")]
+pub struct Command {
+    /// Reference to a package to add or remove as a dependency
+    ///
+    /// You can reference a packages by:
+    /// - `<path>`, like `forc add ./crates/parser/` or `forc remove ./crates/parser/`
+    ///
+    #[clap(value_name = "DEP_ID")]
+    pub crates: Vec<String>,
```
kayagokalp, line 26: One small suggestion that I could make here would be maybe using an enum that could hold either an url or path. Because dependencies could be added as path or git (or registry index as we are starting to plan that) so in terms of code quality/maintainability it could make sense to have an enum of possible sources and we can then have different add/remove logic based on the enum type. I am not sure how well clap would interact with an enum but it might be worth playing with 👍

forc/src/cli/commands/remove.rs
```diff
@@ -0,0 +1,37 @@
...
+    forc remove [OPTIONS] <DEP_PATH> ...
+")]
+pub struct Command {
+    /// Reference to a package to add or remove as a dependency
+    ///
+    /// You can reference a packages by:
+    /// - `<path>`, like `forc add ./crates/parser/` or `forc remove ./crates/parser/`
+    ///
+    #[clap(value_name = "DEP_ID")]
+    pub crates: Vec<String>,
```
kayagokalp, line 26: ditto^

kayagokalp (commented):
Hey there, thanks for the contrib @gr00vytvniks! Overall it seems fine to me but left a little suggestion in terms of improving the maintainability of these new features w.r.t different dependency sources that forc supports (and will support).

forc/src/ops/forc_edit.rs
```diff
@@ -0,0 +1,67 @@
...
+        rev: None,
+    });
+
+    // 2. How will we find the dependency block in the forc.toml?
+    //
+    // Todo: can I insert a new key value pair directly into this BTreeMap?
+    if let Some(mut deps) = pkg_manifest.dependencies.clone() {
+        // 3. Write new dependency to that block (name and path)
+        deps.insert(key, value);
+    }
```
mitchmindtree, line 61: This currently clones the package manifest's dependencies set and inserts the dependency to the cloned dependencies set, but does nothing with the cloned dependencies. There are a few issues with this:

1. This doesn't update the original `pkg_manifest` instance. We can fix this with something like:
   ```rust
       if let Some(deps) = pkg_manifest.dependencies.as_mut() {
           deps.insert(key, value);
       }
   ```
2. We should also check whether or not the dependency already exists first. E.g. if the user already has the dependency entry under their dependencies, we probably don't want to implicitly override an existing entry that might point to an alternative source.
3. We don't write the result to the manifest file. Even if we do make the change mentioned in `1.`, we still need to write the updated manifest to the user's `Forc.toml` file. We should do so in a way that preserves the original formatting and layout of the manifest too, so we might need to consider using a crate like `toml_edit` for this (as mentioned [here](https://github.com/FuelLabs/sway/issues/2369#issuecomment-1193609370)).

forc-pkg/src/manifest.rs
```diff
@@ -191,13 +191,13 @@ pub enum Dependency {
...
-    pub(crate) git: Option<String>,
-    pub(crate) branch: Option<String>,
-    pub(crate) tag: Option<String>,
-    pub(crate) package: Option<String>,
-    pub(crate) rev: Option<String>,
+    pub git: Option<String>,
+    pub branch: Option<String>,
+    pub tag: Option<String>,
+    pub package: Option<String>,
+    pub rev: Option<String>,
```
mitchmindtree, line 200: Rather than making all these fields `pub` which would allow users to construct invalid `DependencyDetails` more easily, we might want to consider exposing dedicated `pub` constructor functions for each particular use case, e.g.
```rust
impl DependencyDetails {
    pub fn from_path(path: &Path) -> Self {
        // ...
    }

    pub fn from_git(repo_url: &Url, git_ref: GitReference) -> Self {
        // ...
    }

    // ... 
```

mitchmindtree (changes requested):
Thanks a lot for the contribution!

I've left some initial feedback, and it looks like a lot of the existing feedback you have received still has not yet been addressed. If you could have ago at addressing all of the existing feedback before requesting another review that would be appreciated!

forc-pkg/src/manifest.rs
gr00vytvniks: Thanks for the feedback Mitch, quick question:
For the `from_git` method, are the `Url` and `GitReference` types here already existing in the codebase or library you are referencing, or was your intention that I make them?

eureka-cpu:
As far as `add` goes, I think taking a look at how the old `cargo add` command worked would be a good idea. It seems pretty well structured and we could essentially do the same: https://github.com/withoutboats/cargo-add/blob/master/src/main.rs

mitchmindtree:
Hmmm one issue with implementing `cargo add` today is that we don't have a registry from which to fetch the latest version of a dependency. In that sense, it might be safe to say `cargo add` is blocked by #3752.

eureka-cpu:
> Hmmm one issue with implementing `cargo add` today is that we don't have a registry from which to fetch the latest version of a dependency. In that sense, it might be safe to say `cargo add` is blocked by #3752.

Right, it should only be used as a reference

mitchmindtree:
Ahhh sorry no I meant to say, it might be safe to say that *`forc add`* is blocked in general as a result. I think one of the main benefits of `cargo add foo` is that the command can infer the entire dependency from the name `foo` (including source and version), however without a registry, we can't really do either of these in a `forc add` command :sweat_smile:

eureka-cpu:
> Ahhh sorry no I meant to say, it might be safe to say that _`forc add`_ is blocked in general as a result. I think one of the main benefits of `cargo add foo` is that the command can infer the entire dependency from the name `foo` (including source and version), however without a registry, we can't really do either of these in a `forc add` command sweat_smile

Would we not be able to at least use the path method for now like `forc add [options] --path path`?

mitchmindtree:
> Would we not be able to at least use the path method for now like forc add [options] --path path?

Ah does `cargo add` allow this? If so, I guess we can start with that here too :+1:

kayagokalp:
Hey @gr00vytvniks, thanks for your PRs so far! I am closing this one as it has been stale for couple of months now. Please feel free to re-open if you still want to get this in.

PR #3353, opened 2022-11-12
Nix the global `TYPE_ENGINE`

Centril:
First part of https://github.com/FuelLabs/sway/issues/2063.

The goal here is to remove all the functions that use the lazy static `TYPE_ENGINE`, and once that's done, we can safely remove it.

Unfortunately, we have to introduce `PartialEqWithTypeEngine` and friends, but eventually we should be able to revert back to good ol' `PartialEq` once https://github.com/FuelLabs/sway/issues/3115 is done.


mohammadfawaz (dismissed):

sway-parse/src/lib.rs
```diff
@@ -37,3 +37,12 @@ pub fn parse_file(
     let ts = lex(handler, &src, 0, src.len(), path)?;
     Parser::new(handler, &ts).parse_to_end().map(|(m, _)| m)
 }
+
+pub fn parse_module_kind(
```
emilyaherbert, line 41: Is this a related change?

sway-core/src/type_system/type_engine.rs
```diff
@@ -1,29 +1,48 @@
...
 use crate::{
     concurrent_slab::ConcurrentSlab, declaration_engine::*, language::ty, namespace::Path,
     type_system::*, Namespace,
 };
 
-use lazy_static::lazy_static;
 use sway_error::{error::CompileError, type_error::TypeError, warning::CompileWarning};
 use sway_types::{span::Span, Ident, Spanned};
 
-lazy_static! {
```
emilyaherbert, line 13: Yesss

emilyaherbert (approved):
> Unfortunately, we have to introduce `PartialEqWithTypeEngine` and friends, but eventually we should be able to revert back to good ol' PartialEq once #3115 is done.

Yes, adding `PartialEqWithTypeEngine` and such is not ideal but is needed for this PR in the meantime.

emilyaherbert (approved):

mohammadfawaz (approved):

PR #3361, opened 2022-11-14
feat: Implement workspace testing

### LINKED ISSUE #3358: Update forc test logic to support workspaces
Currently forc supports workspaces while building, deploying, running and checking but not while testing. We should also be able to run all the unit tests in the workspace.

kayagokalp:
closes #3358.

## Summary

This PR adds workspace support to `forc-test`. Executing `forc test` in a workspace runs all tests coming from the members of the workspace and outputs them formatted separately. An example of the output can be seen below. I also added workspace unit test support to testing suite and added a `should_pass/unit_tests/` test with a workspace.

image


forc-test/src/lib.rs
```diff
@@ -208,10 +228,12 @@ fn run_tests(built: BuiltTests) -> anyhow::Result<Tested> {
         })
         .collect::<anyhow::Result<_>>()?;
 
-    let built = built_pkg;
-    let tested_pkg = TestedPackage { built, tests };
-    let tested = Tested::Package(Box::new(tested_pkg));
-    Ok(tested)
+    let tested_pkg = TestedPackage {
+        built: Box::new(built_pkg.clone()),
```
mitchmindtree, line 232: Rather than cloning the whole package here, it looks like we can take `built_pkg` by value in the function input rather than by reference, i.e. `built_pkg: BuiltPackage`.

forc/src/cli/commands/test.rs
```diff
@@ -44,56 +45,73 @@ pub(crate) fn exec(cmd: Command) -> Result<()> {
...
-                    info!(
-                        "      - test {}, {:?}:{} ",
-                        failed_test_name, path, line_number
-                    );
-                }
-                info!("\n");
+        forc_test::Tested::Workspace(pkgs) => {
+            for pkg in pkgs {
+                let built = &pkg.built.pkg_name;
+                info!("\n   testing -- {built}\n");
```
mitchmindtree, line 51: "testing" might imply that the testing begins after this output, but testing has already completed by the time we reach this point. Perhaps this should be:

```suggestion
                info!("\n   tested -- {built}\n");
```

test/src/e2e_vm_tests/mod.rs
```diff
@@ -264,16 +264,21 @@ impl TestContext {
...
+                            .enumerate()
+                            .map(|(ix, test)| match test.state {
+                                ProgramState::Revert(code) => {
+                                    Some(format!("Test {ix} failed with revert {code}\n"))
+                                }
+                                _ => None,
+                            })
+                            .collect()
                     })
                     .collect();
```
mitchmindtree, line 283: It looks like we might want to use `flat_map` here instead?

We might also want to include the package name now, otherwise there could be multiple tests with the same index. On that note, we should be able to provide the test name rather than just its index too.

Also, it looks like this only checks for the `Revert` state, but tests can now fail if they were decorated with `should_revert` but don't actually revert - maybe this should be updated?

```rust
let failed: Vec<String> = tested_pkgs
    .into_iter()
    .flat_map(|tested_pkg| {
        tested_pkg
            .tests
            .into_iter()
            .filter(|test| !test.passed())
            .map(|test| format!(
                "{}: Test '{}' failed with state {:?}, expected: {:?}",
                tested_pkg.built.name,
                test.name,
                test.state,
                test.condition,
            ))
    })
    .collect();
```

mitchmindtree (changes requested):
Nice, excited for this one! Just a couple small suggestions then should be good

JoshuaBatty (approved):

mitchmindtree (approved):

PR #3364, opened 2022-11-15
`U128` Multiplication optimisation

rostyslavtyshko:
Gas usage dropped about 70 times


sway-lib-std/src/u128.sw
```diff
@@ -304,18 +304,18 @@ impl core::ops::Subtract for U128 {
...
-        while y != zero {
-            if (y & one).lower != 0 {
-                result += x;
-            }
-            x <<= 1;
-            y >>= 1;
+        // in case both of the `U128` upper parts are bigger than zero,
+        // it automatically means overflow, as any `U128` value
+        // is upper part multiplied by 2 ^ 64 + lower part
+        assert(!(self.upper != 0 && other.upper != 0));
```
otrho, line 310:
```suggestion
        assert(self.upper == 0 || other.upper == 0);
```
Waaay more readable. :)

Braqzen (commented):
Gas? Do you mean bytecode size? How is the gas measured?

sway-lib-std/src/u128.sw
rostyslavtyshko: I'd leave as it is, as what I wanted to say is that I want to assure a situation in which both values don't have upper values, and not that I want at least one of them to zero. From the point of view of mathematic logic it's the same thing, but semantically, no.

Braqzen:
> > Gas? Do you mean bytecode size? How is the gas measured?
> 
> It's approximate value, I got it from testing.

_How_ did you get it from testing?

rostyslavtyshko:
> > > Gas? Do you mean bytecode size? How is the gas measured?
> > 
> > 
> > It's approximate value, I got it from testing.
> 
> _How_ did you get it from testing?

Ran some tests with different values with 2 versions of the code

Braqzen:
> > > > Gas? Do you mean bytecode size? How is the gas measured?
> > > 
> > > 
> > > It's approximate value, I got it from testing.
> > 
> > 
> > _How_ did you get it from testing?
> 
> Ran some tests with different values with 2 versions of the code

... Can you _literally_ tell me the steps, the versions used, where you saw the output in order to make that comparison.

rostyslavtyshko:
> > > > > Gas? Do you mean bytecode size? How is the gas measured?
> > > > 
> > > > 
> > > > It's approximate value, I got it from testing.
> > > 
> > > 
> > > _How_ did you get it from testing?
> > 
> > 
> > Ran some tests with different values with 2 versions of the code
> 
> ... Can you _literally_ tell me the steps, the versions used, where you saw the output in order to make that comparison.
I was using `0.31.1` version, ran tests using `forc-run` got the gas usage from the result of the run

sway-lib-std/src/u128.sw
otrho: Semantically it's pretty much the same, really.  I looked at your original version for a minute just to work out what it was actually doing because of the double-negatives.

mohammadfawaz:
@Braqzen when you run a script using `forc run`, you get something like
image
which tells you how much `gas` was used.

sway-lib-std/src/u128.sw
mohammadfawaz: Yeah I think Toby is right. Saying "We need at least one `upper` value to be 0 to avoid overflow" is pretty clear and requires less operators.

sway-lib-std/src/u128.sw
```diff
@@ -304,18 +304,18 @@ impl core::ops::Subtract for U128 {
...
+        // is upper part multiplied by 2 ^ 64 + lower part
+        assert(self.upper == 0 || other.upper == 0);
+
+        let mut result = self.lower.overflowing_mul(other.lower);
+        if self.upper == 0 {
+            let self_lower_other_upper = self.lower * other.upper; // will panic in case of overflow
+            result.upper += self_lower_other_upper; // will panic in case of overflow
+        } else if other.upper == 0 {
+            let self_upper_other_lower = self.upper * other.lower; // will panic in case of overflow
+            result.upper += self_upper_other_lower; // will panic in case of overflow
```
Braqzen, line 318:
```suggestion
            // panic in case of overflow
            result.upper += self.lower * other.upper;
        } else if other.upper == 0 {
            // panic in case of overflow
            result.upper += self.upper * other.lower;
```

The additional variable in the block does not convey anything significant imo and it allocates memory for a one time use of a variable, moreover, the comments are a little repetitive / make the line long. Imo it's easier to read from top to bottom and automatically filter out the commented line as soon as you see it rather than having to read the entire line to see why it's so long.

Braqzen (commented):
Approving after suggestion.

sway-lib-std/src/u128.sw
rostyslavtyshko: Fixed accordingly

mohammadfawaz (approved):

Issue #3283, opened 2022-11-04
Introduce `thiserror` to forc-pkg


kayagokalp:
related to https://github.com/FuelLabs/sway/pull/3134#discussion_r1010055662

mohammadfawaz:
Is this a duplicate of https://github.com/FuelLabs/sway/issues/1201?

eureka-cpu:
Closing in favor of #4817

PR #3367, opened 2022-11-15
Output manifest file parsing errors/warnings for workspace members

### LINKED ISSUE #3365: Invalid manifest file parsing errors are silenced by ManifestFile checks
While checking for the type of a manifest file (workspace or package) the parsing warnings and errors are silenced, meaning that trying to parse any invalid manifest file just prints: 

`Error: Cannot find a valid Forc.toml at "/Users/kayagokalp/fuel/test_projects/virtual/test_project"`

kayagokalp:
closes #3365.

This PR re-enables parsing errors/warnings of manifest files (if there are errors/warnings emitted while parsing the manifest file) and this includes individual members of a workspace. Before this PR, members with invalid manifest files are just omitted without throwing any error. Once we have #3283 we could potentially improve these places.


mitchmindtree (approved):
A bit messy, but agreed #3283 would help clean this up.

This is maybe also further motivation for #3437, as if it were implemented there would be no ambiguity here at all.

JoshuaBatty (approved):

PR #3368, opened 2022-11-15
Prevent I/O while sourcing each path dependency

### LINKED ISSUE #3282: Remove unnecessary I/O operation while checking dep_path in forc-pkg
related to https://github.com/FuelLabs/sway/pull/3134#discussion_r1013555605.

kayagokalp:
closes #3282.


bingcicle (dismissed):

mitchmindtree (dismissed):

kayagokalp:
I did a little bit of investigation [this CI](https://github.com/FuelLabs/sway/actions/runs/3561601315/jobs/5982647372) failure, seems like the dependency graph of the failing test was created with duplicate nodes by the master version of forc and this re-use of manifest file PR also fixes that (examples below). This resulted the order of the printed dependency cycle to be reversed. I am planning to modify the test's expected output (as the following prompt for dependency cycle essentially the same a->b->a and b->a->b) and open up an issue for this as I also want to change kosaraju to tarjan's for SCC detection as it seems to be tarjan's should be faster in most cases (although their asymptotical complexities are the same kosaraju is doing 2 DFS while tarjan's is doing only 1). This might make a performance difference in big dependency graphs. While doing the mentioned change we can also invert the order.


## Failing test dependency graph before this pr:
image

## Failing test dependency graph after this pr:
image

mitchmindtree (approved):

bingcicle (approved):

PR #3370, opened 2022-11-15
Add `rust-toolchain` file

AlicanC:
The [toolchain file](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) is used to pin a project to a certain version of the Rust toolchain.

We currently do not specify an actual version anywhere but just `stable` so I have followed that, but pinning it to an actual version like `1.65` could prevent new Clippy lints, etc. breaking our CI process when there's a new release.

Though, the file is useful even with just `stable` since it helps us clean the CI YAMLs.

Note: The current format of this file is actually TOML, but the GH action we're using [doesn't support it yet](https://github.com/actions-rs/toolchain/pull/209) so I had to use the legacy format. (The GH action also seems to be unmaintained: https://github.com/actions-rs/toolchain/issues/216)


AlicanC:
cc @ControlCplusControlV

otrho (approved):

mohammadfawaz (approved):

otrho (approved):

mohammadfawaz (approved):

AlicanC:
I'm actually going to close this because I just couldn't find a convenient way of doing it. The action we currently use doesn't fully support our case and [it's unmaintained](https://github.com/actions-rs/toolchain/issues/216) so there's nothing we can do besides forking. There's [an alternative](https://github.com/dtolnay/rust-toolchain) but maintainer of that is refusing to implement support for the file. (This comment summarizes the situation well: https://github.com/actions-rs/toolchain/issues/126#issuecomment-1249931329)

I'll revisit if the situation changes.

PR #3371, opened 2022-11-15
Introduce code relocation to avoid the need for far jumps.

### LINKED ISSUE #3347: Relocate code to avoid far jumps in ASMgen.
Very large binaries, larger than 1MB of bytecode, require using far jumps for control flow using `JNE` as implemented in #3109.

But this blew out the data section as noted in #3161.  One solution is relative jumps in the FuelVM https://github.com/FuelLabs/fuel-specs/issues/430

Another not-quite-as-elegant solution is to move all non-control flow code to beyond the 1MB boundary and to keep control flow in that lower region.  Basic blocks could be replaced with a single `JI` instruction, where essentially all control flow logic remains at addresses which fit in 18bits and other data manipulation code would lie beyond.

The largest BBs could be moved first, incrementally, until all control flow lies under 1MB.  If replacing _every_ BB still doesn't fit in 1MB then we can fall back to using data section addressing which we have now (as per #3109).

otrho:
The existing solution for far jumps which can't fit in an immediate is to convert them to dynamic jumps, storing the destination address in the data section.

This change helps avoid far jumps altogether by moving large blocks to beyond the limit to keep immediate mode control flow.

It doesn't strictly fix the #3161 problem but it makes it very much less likely to occur.  E.g., the `sway-libs/amm` library will compile now with this change.

Closes #3347


mohammadfawaz (dismissed):

sway-core/src/asm_generation/allocated_abstract_instruction_set.rs
```diff
@@ -156,12 +167,143 @@ impl AllocatedAbstractInstructionSet {
         self
     }
 
+    /// Relocate code to keep as much control flow below the magic 1MB address space accessible by
+    /// a `JNZI` instruction.
+    pub(crate) fn relocate_control_flow(mut self, data_section: &DataSection) -> Self {
+        // Do an analysis pass, gathering basic block offsets and whether any jumps are going to be
```
emilyaherbert, line 173: Can you describe what it would mean for a jump to be a problem?

emilyaherbert (changes requested):
> It doesn't strictly fix the https://github.com/FuelLabs/sway/issues/3161 problem but it makes it very much less likely to occur.

Can you help me get an intuition for what the circumstances would be in which this issue would occur still?

Just a few questions on my end, otherwise---awesome work!

sway-core/src/asm_generation/allocated_abstract_instruction_set.rs
emilyaherbert: (Document this in the code I mean)

sway-core/src/asm_generation/allocated_abstract_instruction_set.rs
```diff
@@ -156,12 +167,143 @@ impl AllocatedAbstractInstructionSet {
         self
     }
 
+    /// Relocate code to keep as much control flow below the magic 1MB address space accessible by
```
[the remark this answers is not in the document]
mohammadfawaz: Toby clarified this in his latest commit. Basically it's just the size of the bytecode that can be addressed using the 18-bit immediate of `jnzi`.

PR #3373, opened 2022-11-15
Add markdown formatting & support to `forc doc`

### LINKED ISSUE #3175: `forc doc` - render doc comments using markdown (not just raw HTML)
Currently we render documentation output by `forc doc` to `HTML` via `horrorshow`. Ideally, while the page layout can be handled with HTML, we should be rendering the doc comments themselves using markdown.

See sway-lsp's implementation of hover documentation as an example https://github.com/FuelLabs/sway/pull/3084.

We could also investigate what `rustdoc` uses for this.

eureka-cpu:
Some things that are nice about this PR: The library I've chosen to handle markdown to HTML rendering is based on `cmark` and has some really great customization options. I've gone ahead and chosen the ones that I think make sense, and it looks like this library is also well maintained. I've made the `format_docs` function public from the LSP to get some formatting before rendering to HTML.

Closes #3175


bingcicle (approved):
:shipit:

forc-plugins/forc-doc/src/render.rs
```diff
@@ -348,19 +350,31 @@ fn qualified_file_path(module_prefix: &Vec<String>, file_name: String) -> String
...
+    let mut docs = String::new();
+
     if let Some(vec_attrs) = attributes {
         for attribute in vec_attrs {
             for ident in &attribute.args {
-                attr_strings.push_str(ident.as_str())
+                docs.push_str(&format!("{}\n", ident.as_str()))
             }
         }
     }
```
kayagokalp, line 363: Maybe could be re-written like the following to omit extra allocations 

```rust
if let Some(vec_attrs) = attributes {                                                                                                                                                                                                
         for ident in vec_attrs.iter().flat_map(|attribute| &attribute.args) {
             writeln!(docs, "{}", ident.as_str());
         }
     }
```

kayagokalp (approved):
Just left a single nit, otherwise utACK

forc-plugins/forc-doc/src/render.rs
eureka-cpu: I'll make a follow up with this, nice!

PR #3390, opened 2022-11-18
add option to preserve lazy statics on compilation

sezna:
It has been made clear to me that the LSP can't be adopted by app devs until it can run multiple projects without crashing. @mitchmindtree suggested we just preserve the lazy statics if it is coming from the LSP so the type ids and decl ids remain valid.

Since @emilyaherbert recently reduced the number of type ids we generate significantly, this is probably a good enough workaround until we actually remove the lazy statics. Perhaps the devs can get a couple of hours out of this before the engine overflows. 😅 

This is totally just a hacky workaround and should be removed when we remove lazy statics.


sezna:
no longer necessary due to #3353

PR #3391, opened 2022-11-18
Fixes issue of trait implementation not beeing used.

### LINKED ISSUE #3326: Trait constraints issues with `Eq` for custom types
Code:
```rust
script;

use core::ops::*;

impl<T> Option<T> {
        match self {
        }
    }
}

enum Error {
    BoolError: bool,      
    U8Error: u8,      
}

impl Eq for Error {
        match (self, other) {
            (Error::BoolError(val1), Error::BoolError(val2)) => val2 == val1,
            (Error::U8Error(val1), Error::U8Error(val2)) => val2 == val1,
            _ => false,
        }
    }
}

fn test_none_ok_or<T, E>(_val: T, default: E) where E: Eq {
    match Option::None::<T>().ok_or(default) {
        Result::Ok(_) => revert(0),
        Result::Err(e) => assert(default == e),
    }
}

fn main() {
    test_none_ok_or(true, Error::BoolError(true));
}
```
This compiles but reverts. When I change `test_none_ok_or(true, Error::BoolError(true));` to `test_none_ok_or(true, Error::BoolError(0));` for example, the test does not revert. 

In `test_none_ok_or`, if I log `default` and `e`, I get the same value so the issue is with the comparison `default == e` which evaluates to `false` for some reason, even though `Eq` is implemented correctly. In fact, even if I change the implementation of `Eq` for `Error` to just be `return true`, the test still reverts 🤔

esdrubal:
The issue was caused by constraints used in generic functions having trait name that does not contain any prefixes. This was causing `get_methods_for_type_and_trait_name` to return no methods to be used.

This PR changes `get_methods_for_type_and_trait_name` so when a `trait_name` without prefixes is provided it still tries to match using the suffixes.

Closes #3326


emilyaherbert (changes requested):
Can you expand the PR description? I'm not understanding how the changes in your PR have solved the problem based on the PR description.

esdrubal:
@emilyaherbert I rephrased and expanded the PR description.

sway-core/src/semantic_analysis/namespace/trait_map.rs
```diff
@@ -559,8 +559,10 @@ impl TraitMap {
                 suffix: map_trait_name.suffix.0.clone(),
                 is_absolute: map_trait_name.is_absolute,
             };
-            if &map_trait_name == trait_name && are_equal_minus_dynamic_types(type_id, *map_type_id)
-            {
+            let are_same_name_traits = (trait_name.prefixes.is_empty()
+                && map_trait_name.suffix == trait_name.suffix)
```
emilyaherbert, line 563: It seems that this would allow for false positives between traits with the same suffix and different prefixes. Some local `MyAdd` and a different `my_library::MyAdd`.

emilyaherbert (changes requested):
> The issue was caused by constraints used in generic functions having trait name that does not contain any prefixes.

I think what I'm seeking to understand is what the conditions are that are causing this case to come up? My hunch is that the bug is not in how we are drawing equivalency between trait names but rather how we are performing call path and namespace resolution for trait names.

esdrubal:
When I comment out [trait_constraint.rs#L82](https://github.com/FuelLabs/sway/blob/81b920db4e10bd8cf57292c1951cd2182e423d2c/sway-core/src/type_system/trait_constraint.rs#L82) and use `where E: core::ops::Eq` I no longer get this error.
It looks like when `get_methods_for_type_and_trait_name` receives the trait name from `TraitConstraint` with module path this error no longer occurs.

The issue originates in [type_parameter.rs#L203](https://github.com/FuelLabs/sway/blob/81b920db4e10bd8cf57292c1951cd2182e423d2c/sway-core/src/type_system/type_parameter.rs#L203) where the trait name is not absolute.

@emilyaherbert is there a way of resolving a local `CallPath` to an absolute `CallPath`?

mohammadfawaz:
Conflicts

sway-core/src/semantic_analysis/namespace/trait_map.rs
esdrubal: In the most recent version it no longer compares suffixes instead before calling `get_methods_for_type_and_trait_name` it converts the trait name call path to contain prefixes when it is possible.

sway-core/src/type_system/type_parameter.rs
```diff
@@ -220,8 +220,25 @@ impl TypeParameter {
                     type_arguments: trait_type_arguments,
                 } = trait_constraint;
 
+                // Use trait name with module path as this is expected in get_methods_for_type_and_trait_name
+                let mut full_trait_name = trait_name.clone();
+                if trait_name.prefixes.is_empty() {
+                    if let Some(vec) = ctx.namespace.use_synonyms.get(&trait_name.suffix) {
+                        full_trait_name = CallPath {
+                            prefixes: vec.0.clone(),
```
emilyaherbert, line 228: Why only the first element from `vec`?

sway-core/src/type_system/type_parameter.rs
```diff
@@ -220,8 +220,25 @@ impl TypeParameter {
                     type_arguments: trait_type_arguments,
                 } = trait_constraint;
 
+                // Use trait name with module path as this is expected in get_methods_for_type_and_trait_name
+                let mut full_trait_name = trait_name.clone();
+                if trait_name.prefixes.is_empty() {
+                    if let Some(vec) = ctx.namespace.use_synonyms.get(&trait_name.suffix) {
+                        full_trait_name = CallPath {
```
emilyaherbert, line 227: Why only the first element from vec?

emilyaherbert (changes requested):
I think that this is closer to the best solution. Just one question.

sway-core/src/type_system/type_parameter.rs
esdrubal: The vec variable is in fact `(Vec<Ident>, GlobImport)`. I will change its name to cause less confusion.

mohammadfawaz (approved):

emilyaherbert (approved):

PR #3392, opened 2022-11-21
Fix CEI analysis storage map/vec false positives

### LINKED ISSUE #3355: `insert()` into `StorageMap` triggers two CEI pattern check warnings
Example:
```rust
contract;

dep external;

use external::*;

use std::auth::msg_sender;

abi MyContract {
    #[storage(read, write)]
    fn withdraw();
}

storage {
    balances: StorageMap<Identity, u64> = StorageMap {}
}

    #[storage(read, write)]
    fn withdraw() {
        let bal = storage.balances.get(sender);

        assert(bal > 0);

        let caller = abi(OtherContract, 0x3dba0a4455b598b7655a7fb430883d96c9527ef275b49739e7b0ad12f8280eae);
        caller.external_call();

        storage.balances.insert(sender, 0);
    }
}
```
image

The reason is that `storage.balances.insert()` requires first "reading" `storage.balances`. However, because `balances` is an empty struct (being a `StorageMap {}`), no storage reads ever show up in the bytecode for it. 

We should modify the CEI pattern check analysis to check if a `StorageAccess` is of type `struct { }`, in which case, we shouldn't consider it a _real_ storage read.

anton-trunov:
Also fixes the effect of the `__get_storage_key` intrinsic.

closes #3355


emilyaherbert (approved):

sway-core/src/semantic_analysis/cei_pattern_analysis.rs
```diff
@@ -462,7 +462,10 @@ fn effects_of_expression(expr: &ty::TyExpression) -> HashSet<Effect> {
         | AbiName(_) => HashSet::new(),
         // this type of assignment only mutates local variables and not storage
         Reassignment(reassgn) => effects_of_expression(&reassgn.rhs),
-        StorageAccess(_) => HashSet::from([Effect::StorageRead]),
+        StorageAccess(_) => match look_up_type_id(expr.return_type) {
+            crate::TypeInfo::Struct { .. } => HashSet::new(),
```
mohammadfawaz, line 466: As discussed in the morning, this seems too liberal. I would have expected an extra check here to make sure that the struct has no fields (i.e. `fields.is_empty()`). And also, maybe empty enums should be checked as well because I suspect they don't generate any storage opcodes either.

mohammadfawaz (changes requested):

sway-core/src/semantic_analysis/cei_pattern_analysis.rs
anton-trunov: Thanks! Added a fix for this. I added a check for empty enums as well, although, one should not be able to construct a value of an empty enum, so I cannot test this check, but it would not hurt in the analysis :)

anton-trunov:
@emilyaherbert @Centril @mohammadfawaz Could you please check that the refactoring to accommodate the newest changes in the type engine are the right ones?

mohammadfawaz (approved):
All makes sense to me. Thanks!

sezna (approved):

Issue #3357, opened 2022-11-14
Disallow `while` loops in predicates in the frontend


mohammadfawaz:
This shouldn't compile:
```rust
predicate;

fn main() -> bool {
    let x = 0;
    while true { x += 1; }

    true 
}
```
https://github.com/FuelLabs/sway/pull/3291 will disallow this in codegen. It would be better to disallow this in the frontend, potentially along with other things like storage intrinsics etc.

anton-trunov:
Looks like this could be achieved by refactoring the language into a *core*, which could be somewhat pure language (no loops, no storage effects, no contract calls(?), only local state modification is allowed, etc.) and the rest of it. This would greatly simplify developing static analyses for Sway, formal verification for Sway etc.

mohammadfawaz:
https://github.com/FuelLabs/sway/pull/3400 Closes this

PR #3400, opened 2022-11-21
Disallows while loops in predicates.

esdrubal:
Now when a while loop is used in a predicate an error is thrown.

Ref #3357


sway-core/src/semantic_analysis/type_check_context.rs
```diff
@@ -71,6 +74,7 @@ impl<'ns> TypeCheckContext<'ns> {
             self_type: insert_type(TypeInfo::Contract),
             mode: Mode::NonAbi,
             purity: Purity::default(),
+            kind: None,
```
emilyaherbert, line 77:
```suggestion
            kind: TreeType::Contract,
```

Would you change this to `TreeType::Contract` to follow `self_type: insert_type(TypeInfo::Contract),`?

test/src/e2e_vm_tests/test_programs/should_fail/predicate_while/src/main.sw
```diff
@@ -0,0 +1,8 @@
+predicate;
+
+fn main() -> bool {
+    let mut x = 0;
+    while true { x += 1; }
+
+    true 
+}
```
emilyaherbert, line 8:
```suggestion

}
```

test/src/e2e_vm_tests/test_programs/should_fail/predicate_while/Forc.toml
```diff
@@ -0,0 +1,9 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+license = "Apache-2.0"
+name = "predicate_while"
+entry = "main.sw"
+implicit-std = false
+
+[dependencies]
+core = { path = "../../../../../../sway-lib-core" }
```
emilyaherbert, line 9:
```suggestion
core = { path = "../../../../../../sway-lib-core" }

```

test/src/e2e_vm_tests/test_programs/should_fail/predicate_while/test.toml
```diff
@@ -0,0 +1,4 @@
+category = "fail"
+
+# check: while true { x += 1; }
+# nextln: $()Using "while" in a predicate is not allowed.
```
emilyaherbert, line 4:
```suggestion
# nextln: $()Using "while" in a predicate is not allowed.

```

emilyaherbert (changes requested):
Just a few nits.

esdrubal:
@emilyaherbert Thanks for the review, all requested changes were applied.

test/src/e2e_vm_tests/test_programs/should_fail/predicate_while/src/main.sw
```diff
@@ -0,0 +1,8 @@
+predicate;
+
+fn main() -> bool {
+    let mut x = 0;
+    while true { x += 1; }
```
mohammadfawaz, line 5: I'm curious to know if this still throws an error if the the while loop is used in a library function and that function is used in a predicate. If so, it's worth adding a test for it.

esdrubal: Current solution was not throwing the error when the predicate uses a library with loops.
We still got the `The JI opcode cannot jump backwards in a predicate.` but it is not very useful because it has no span available.

We could check for while loops in other modules too but `type_check` checks all declarations so even functions that are not used would thrown an error if they contain loops. Ideally we want to throw an error only if the predicate uses a function with a while loop.

The solution was to perform the while in predicate check on `compile_while_loop` as at this time only used functions are compiled.

mohammadfawaz (approved):

emilyaherbert (approved):

PR #3408, opened 2022-11-22
Add forc-index plugin

ra0x3:
### Description
- This PR adds the long-awaited Forc Indexer plugin that is invoked via `forc index`
- `forc index` is pretty much based on forc in terms of structure
- There may have been a few things that I left out (it's been a minute since I've committed to forc)
- I'm not really sure how to "hook this in" to forc -- so I just built the binary `cargo build --release -p forc-index` and manually added it to my `~/.fuelup/bin`

### Testings steps
- [ ] I would recommend you pull down the https://github.com/FuelLabs/fuel-indexer/ project
- [ ] And build a `fuel-indexer` binary via `cargo build --release -p fuel-indexer`
  - This will help with the `forc index start` and `forc index deploy` commands
  - This `fuel-indexer` binary has been merged into `fuelup` but haven't had a chance to use it yet 
- [ ] Then maybe just start playing around by following [the `forc index` README](https://github.com/FuelLabs/sway/pull/3408/files#diff-f2a224153badb56732262b7d0bcbf0ab38ea97660a497b02be28408310918dd1)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -299,8 +299,8 @@ dependencies = [
  "futures-core",
  "futures-io",
  "rustls 0.18.1",
- "webpki",
- "webpki-roots",
+ "webpki 0.21.4",
+ "webpki-roots 0.20.0",
 ]
 
 [[package]]
@@ -1114,7 +1114,7 @@ version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c1a816186fa68d9e426e3cb4ae4dff1fcd8e4a2c34b781bf7a822574a0d0aac8"
 dependencies = [
- "sct",
+ "sct 0.6.1",
 ]
 
 [[package]]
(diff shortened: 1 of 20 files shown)


mitchmindtree:
Wicked stuff @ra0x3!

The first thought that comes to mind: perhaps this would be better hosted under the `fuel-indexer` repo instead?

We've slowly been moving plugins out of this repo (at least, plugins unrelated to Sway) to try and cut down on bloat here a little. As an example, we recently moved out the `forc-explorer` command into a dedicated repo [here](https://github.com/FuelLabs/forc-explorer). It sets a bit of an unmaintainable precedent if we end up hosting every forc plugin under the sun under this repo :sweat_smile:

By hosting `forc-index` under the `fuel-indexer` repo, I imagine it might make it easier to integrate with the fuel-indexer integration testing too, which could lower the chance of the plugin falling out of sync with the tool?

> I'm not really sure how to "hook this in" to forc

Nothing to be done! `forc` will automatically try to find `forc-index` via the user's `PATH` when it receives an otherwise unknown `forc index` command, so it should work however you install it (whether via `cargo install`, fuelup or your distro's package manager, etc).

ra0x3:
@mitchmindtree 
- Totally fine with moving this to the `fuel-indexer` repo
- Question. If I do move this to the `fuel-indexer` repo. How is `forc` supposed to be aware of it? Would we just include the plugin binary(ies) on releases -- then have `fuelup` pull in the plugin binary as well as the service binary? 🤔

mitchmindtree:
> Would we just include the plugin binary(ies) on releases -- then have fuelup pull in the plugin binary as well as the service binary?

Yeah exactly this :100: This is how `forc-wallet` and `forc-explorer` work today too cc @bingcicle

ra0x3:
@mitchmindtree Awesome, thanks for that. I'll close this now. @bingcicle be on the lookout for me with another ask coming soon 🥳

PR #3416, opened 2022-11-23
Fix error message for arity mismatch on method calls

### LINKED ISSUE #2949: Fix the error message for incorrect number of arguments on method calls
This Sway code:

```rust
script;

struct Data {
    value: u64
}

impl Data {
    fn add_values(self, other: Data) -> u64 {
        self.value + other.value
    }
}

fn main() -> u64 {
    let data1 = Data { 
        value: 42u64
    };
    let data2 = Data {
        value: 1u64
    };
    data1.add_values()
}
```

gives this error message:

```bash
   |
18 | 
19 |     };
20 |     data1.add_values()
   |           ^^^^^^^^^^ Function "add_values" expects 2 arguments but you provided 1.
21 | }
   |
____

  Aborting due to 1 error.
```

When it should say "expects 1 argument".

anton-trunov:
close #2949


sway-error/src/error.rs
```diff
@@ -413,21 +413,25 @@ pub enum CompileError {
     TraitNotFound { name: String, span: Span },
     #[error("This expression is not valid on the left hand side of a reassignment.")]
     InvalidExpressionOnLhs { span: Span },
-    #[error(
-        "Function \"{method_name}\" expects {expected} arguments but you provided {received}."
+    #[error("{} \"{method_name}\" expects {expected} {} but you provided {received}.",
```
mohammadfawaz, line 416: How do you feel about using the exact wording of `rustc`?
```
this function takes 1 argument but 0 arguments were supplied
```
Ref: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6e72e1cd8e1c6fab26cd2803ff46d9ca

mohammadfawaz (commented):
Just a small suggestion, but not a strong opinion.

mohammadfawaz (approved):

sway-error/src/error.rs
anton-trunov: Yeah, sure, let's do this. Let me clarify two things, though: 
- do we want to still distinguish between functions and methods here?
- do we want the function/method name to be mentioned in the error message or just 'this' suffices?

mohammadfawaz: Ah yeah good question. I wonder why Rust doesn't use "method" here. 
* I say we continue using "method" instead of "function"
* I also like the method name to be mentioned. 
I guess that brings us mostly back to how the error was 😆 , except for the minor rewording. Maybe let's just keep it as is now as this is ready to merge, and we can do a full audit for all error messages in the future if needed.

PR #3424, opened 2022-11-24
Add a check in codegen to ensure all registers are initialised before use.

### LINKED ISSUE #2272: Sanity check for the intermediate assembly
It is possible that the intermediate assembly generated uses registers that are not assigned. This showed up several times in the past and usually points to an actual functional bug. We should probably introduce a sanity check after generating intermediate assembly (before running the register allocator) to make sure that this pattern does not arise. We can emit an internal compiler error in that case which would help us easily detect that something is wrong and tell us where to look.

otrho:
It isn't really possible to add tests, at least not easily, as we need to get the IR -> ASM code to generate bad code deliberately.

Closes #2272

### DIFF
--- a/sway-core/src/asm_generation/abstract_instruction_set.rs
+++ b/sway-core/src/asm_generation/abstract_instruction_set.rs
@@ -6,7 +6,13 @@ use crate::{
     },
 };
 
-use std::{collections::HashSet, fmt};
+use sway_error::error::CompileError;
+use sway_types::Span;
+
+use std::{
+    collections::{BTreeSet, HashSet},
+    fmt,
+};
 
 use either::Either;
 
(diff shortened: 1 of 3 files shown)


sway-core/src/asm_generation/abstract_instruction_set.rs
```diff
@@ -115,6 +121,37 @@ impl AbstractInstructionSet {
...
+        }
+
+        let mut use_regs = BTreeSet::new();
+        let mut def_regs = BTreeSet::new();
+        for op in &self.ops {
+            add_virt_regs!(op.use_registers(), use_regs);
+            add_virt_regs!(op.def_registers(), def_regs);
+        }
+
+        if def_regs.is_superset(&use_regs) {
```
mohammadfawaz, line 145: Very cool.

mohammadfawaz (approved):

sezna (approved):

Issue #3067, opened 2022-10-18
Multiple `impl` blocks with methods of the same name shadow each other in dead code analysis


emilyaherbert:
For example this Sway code:

```rust
script;

trait Get1 {
    fn get(self) -> u64;
}

trait Get2 {
    fn get(self) -> u64;
}

struct Data1 {
    value: u64
}

struct Data2 {
    value: u64
}

impl Get1 for Data1 {
    fn get(self) -> u64 {
        self.value
    }
}

impl Get2 for Data2 {
    fn get(self) -> u64 {
        self.value
    }
}

fn main() -> u64 {
    let a = Data1 {
        value: 7
    };
    let b = Data2 {
        value: 8
    };
    let c = a.get();

    0
}
```

generates these compiler warnings:

```bash
 --> /Users/emilyherbert/fuel/sway/test/src/e2e_vm_tests/test_programs/should_pass/language/tmp_test/src/main.sw:3:7
  |
1 | 
2 | 
3 | trait Get1 {
  |       ---- This trait is never implemented.
4 |     fn get(self) -> u64;
5 | }
  |
____

warning
  --> /Users/emilyherbert/fuel/sway/test/src/e2e_vm_tests/test_programs/should_pass/language/tmp_test/src/main.sw:12:5
   |
10 | 
11 | struct Data1 {
12 |     value: u64
   |     ---------- This struct field is never accessed.
13 | }
   |
____

warning
  --> /Users/emilyherbert/fuel/sway/test/src/e2e_vm_tests/test_programs/should_pass/language/tmp_test/src/main.sw:19:1
   |
17 |   
18 |   
19 | / impl Get1 for Data1 {
20 | |     fn get(self) -> u64 {
21 | |         self.value
22 | |     }
23 | | }
   | |_- This declaration is never used.
24 |   
25 |   impl Get2 for Data2 {
   |
____
```

The bug is coming from the fact that there *is* a compiler warning for an unused declaration `impl Get1 for Data1 {` and that there *is not* a compiler warnings for an unused declaration `impl Get2 for Data2 {`. The Sway code actually does use the first declaration and does not use the second declaration.

This is happening because this call to `insert_function` overwrites any function with the same name that was already in the map.
https://github.com/FuelLabs/sway/blob/9f31df0dc9219e991fdea90e8978ec10a2638f01/sway-core/src/control_flow_analysis/dead_code_analysis.rs#L574

mohammadfawaz:
[...]

PR #3427, opened 2022-11-25
Fix DCA warnings for impl trait methods.

tritao:
We now connect the edge of methods referenced in function call expressions to the edge of the corresponding node of the implemented trait method.

For this FunctionDeclaration was extended to keep track of the implementing type where it is declared.

Also previously we were not correctly connecting edges in `add_edge_from_entry` due to the function using `self.entry_points` which is only computed as the last stage of the DCA process. Fix this by deferring these edge node indices and only making the connection after they are computed.

Aditionally now we use an `IdentUnique` to keep track of functions in the CFG namespace which helps disambiguate functions shadowing each other.

Closes https://github.com/FuelLabs/sway/issues/3067.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2021,9 +2021,9 @@ dependencies = [
 
 [[package]]
 name = "fuel-vm"
-version = "0.22.4"
+version = "0.22.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0dcee37cc71091466c4dcbd5feebd689573e34a404fef7f02946afd0923e200a"
+checksum = "1521a9ecf42e1d133b29e3785ca6f46bf74ae9bf2509fc3daed49b731f530560"
 dependencies = [
  "fuel-asm",
  "fuel-crypto",
--- a/docs/book/src/forc/manifest_reference.md
+++ b/docs/book/src/forc/manifest_reference.md
@@ -182,3 +182,12 @@ fn main() {
   let foo_id = foo::CONTRACT_ID;
 }
 ```
+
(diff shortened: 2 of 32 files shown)


sway-core/src/declaration_engine/declaration_wrapper.rs
```diff
@@ -113,6 +113,23 @@ impl ReplaceDecls for DeclarationWrapper {
...
+            DeclarationWrapper::Unknown => {}
+            DeclarationWrapper::Function(decl) => decl.set_implementing_type(implementing_type),
+            DeclarationWrapper::Trait(_) => {}
+            DeclarationWrapper::TraitFn(_) => {}
+            DeclarationWrapper::ImplTrait(_) => {}
+            DeclarationWrapper::Struct(_) => {}
+            DeclarationWrapper::Storage(_) => {}
+            DeclarationWrapper::Abi(_) => {}
+            DeclarationWrapper::Constant(_) => {}
+            DeclarationWrapper::Enum(_) => {}
```
otrho, line 128:
```suggestion
            DeclarationWrapper::Function(decl) => decl.set_implementing_type(implementing_type),
            DeclarationWrapper::Unknown
            | DeclarationWrapper::Trait(_)
            | DeclarationWrapper::TraitFn(_)
            | DeclarationWrapper::ImplTrait(_)
            | DeclarationWrapper::Struct(_)
            | DeclarationWrapper::Storage(_)
            | DeclarationWrapper::Abi(_)
            | DeclarationWrapper::Constant(_)
            | DeclarationWrapper::Enum(_) => {}
```
This actually doesn't make much difference I guess, but I'll throw it up anyway.

otrho (commented):
I don't quite follow the specifics, but the tests look good and @emilyaherbert can comment on stuff like adding `implementing_type` to `TyFunctionDeclaration`.

I've mentioned a couple of Rusty nits, but otherwise LGTM.

otrho (approved):

otrho:
@tritao Got a weird LSP test failure now.

otrho (approved):

mohammadfawaz (approved):

PR #3428, opened 2022-11-25
Update the E2E test runner to report shell exit code on failure

tritao:
This updates the E2E test runner to report shell exit code on failure, which it never did before but is now required due to the whole testing process not panicking anymore.

Unfortunately this broke the E2E tests step in CI, which means we have not been correctly reporting tests failures on PRs since https://github.com/FuelLabs/sway/commit/4fe5a388ed894aca00e4a3995e085d63bdab169c landed. :hear_no_evil: 

Some test failures have been introduced to master since, thankfully nothing major, which this also fixes.


mohammadfawaz (approved):
😅

otrho (approved):

PR #3434, opened 2022-11-25
feat: add no-span-debug feature for omitting span for debug purposes

### LINKED ISSUE #3422: Provide a way to debug compiler related structs without their `Span`
While debugging compiler related issues (before opening this I was trying to understand if #3406 is easy to tackle or not) it is hard to keep track of things (printing stuff here and there) as default debug implementation of `Span` is pretty long (file path + related source code itself). I thought i was the only one having this problem (as I am not touching the compiler regularly) but after last language sync I realized that this might be useful to others too 😄. So maybe providing a conditional debug implementation of `Span` that doesn't output these long statements when compiled with a flag could help debugging flow of the compiler.

kayagokalp:
closes #3422.

Building the workspace with `cargo build` and trying to print a debug declaration:
```console
[ConstantDeclaration(DeclarationId(6403, Span { src (ptr): 0x600001fe81c0, path: Some("/Users/kayagokalp/fuel/test_projects/test_debug_library/src/lib.sw"), start: 111, end: 121, as_str(): "test_const" }))]
```

After building the workspace with `cargo build --features sway-types/no-span-debug` and trying to do the same thing:

```console
[ConstantDeclaration(DeclarationId(6403, ))]
```

This is fairly limited as it is completely removing all the span but it is enough for my limited debugging around the compiler. If some flexibility (like keeping the as_str field etc with a different feature) is needed I can also introduce it with this PR.

### DIFF
--- a/sway-types/Cargo.toml
+++ b/sway-types/Cargo.toml
@@ -14,3 +14,6 @@ fuel-crypto = "0.6"
 fuel-tx = "0.23"
 lazy_static = "1.4"
 serde = { version = "1.0", features = ["derive"] }
+
+[features]
+no-span-debug = []
--- a/sway-types/src/span.rs
+++ b/sway-types/src/span.rs
@@ -191,6 +191,7 @@ impl Span {
 }
 
 impl fmt::Debug for Span {
+    #[cfg(not(feature = "no-span-debug"))]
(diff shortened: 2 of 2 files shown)


mohammadfawaz (approved):
Seems reasonable.

sezna (approved):

PR #3435, opened 2022-11-26
Add code block formatting to `forc doc`

eureka-cpu:
This PR adds a formatting step to documentable code blocks which also removes user comments. So something like 
```rust
pub   struct    Hello     { \n\n
      // hello, a comment good sir
    my_field : u32, // another unwanted comment
    // oops this is a comment
        my_other_field: u32 ,
 }
```

will get formatting and have its comments removed only for the main code block show casing the `struct`.

Along the way, I found a small problem in `sway_parse` in `item_struct`. Looks like the parser was looking for the `struct` keyword first instead of the `pub` keyword. I just added a `let` binding for the visibility keyword before we check for `struct`.


swayfmt/src/items/item_struct/tests.rs
```diff
@@ -63,14 +63,14 @@ macro_rules! fmt_test_inner {
 }
 
 fmt_test!(  annotated_struct
-"struct Annotated {
+"pub struct Annotated {
```
eureka-cpu, line 66: Found the parser issue while testing this cc @mohammadfawaz

sway-parse/src/item/item_struct.rs
```diff
@@ -7,6 +7,7 @@ use sway_types::Spanned;
 
 impl Parse for ItemStruct {
     fn parse(parser: &mut Parser) -> ParseResult<ItemStruct> {
+        let visibility = parser.take();
```
eureka-cpu, line 10: Here was my solution based on how `item_enum` looked, and that it was passing the test that `struct` was not. Let me know if this needs to change, I'm not well versed in parser-land requirements. @mohammadfawaz

mohammadfawaz: This does seem reasonable. I'm just trying to understand how this wasn't causing issues. I quickly tried importing a private struct (without this change) and I couldn't so clearly the `pub` keyword is being respected somewhere.

eureka-cpu: Not sure how you could recreate this scenario other than adding the `pub` keyword like I did to the test I've changed here

eureka-cpu: I think this would only be important in instances where you are trying to parse the `struct` standalone or as a snippet, like I am in the tests and in `forc doc`

mohammadfawaz: Hmm notice that the visibility qualifier is collected in https://github.com/FuelLabs/sway/blob/master/sway-parse/src/item/mod.rs#L37-L39 which is a bit weird 🤔

eureka-cpu: Should I open an issue and put this PR on hold?

mohammadfawaz: I would say just continue with this PR as is. We'll clean up all that stuff later.

eureka-cpu: OK, it's working as intended if you wouldn't mind a review!

mohammadfawaz (dismissed):
Compiler stuff seems reasonable to me. As discussed, we should clean up that stuff for structs and enums and whatever else has duplicate visibility parsing.

forc-plugins/forc-doc/src/render.rs
```diff
@@ -463,7 +464,7 @@ impl Renderable for TyStorageDeclaration {
             attributes,
         } = &self;
         let name = "Contract Storage".to_string();
-        let code_span = span.as_str().to_string();
+        let code_span = parse::parse_format::<sway_ast::ItemStorage>(span.as_str());
```
kayagokalp, line 467: Hmm looks like `parse_format` returns the formatted code as a string rather than a span, if that is the case it might make sense to change variable names for readability.

kayagokalp (dismissed):
Left a single nit for readability: nothing serious and it can be addressed alongside a future PR - utACK ✅

kayagokalp (approved):

mitchmindtree (approved):

PR #3450, opened 2022-11-28
Add `#[payable]` annotation to ABI methods

### LINKED ISSUE #1608: Add annotation for ABI method payability
Add an annotation to ABI method _declarations_, `#[payable]`. The lack of this annotation implies the method is nonpayable.

When calling an ABI method that is nonpayable, the compiler must emit an error if the amount of coins to forward is not guaranteed to be zero.

anton-trunov:
## Description

The lack of this annotation implies the method is non-payable. When calling an ABI method that is non-payable, the compiler must emit an error if the amount of coins to forward is not guaranteed to be zero.

The compiler also emits an error if the method signature and the method implementation have different `#[payable]` annotations.

## Assumptions

Currently, the analysis of non-zero coins is very simple and only checks if the special `coins:` contract call parameter is the zero `u64` literal directly or can be reached through a chain of variable/constant definitions.

## Tests

- [x] Must fail when scripts call non-payable methods with non-zero coins as a literal
- [x] Must fail when scripts call non-payable methods with non-zero coins as a constant
- [x] Must fail when scripts call non-payable methods with non-zero coins as a variable definition
- [x] Must fail when contracts call non-payable methods with non-zero coins
- [x] Must fail when there is an extra `#[payable]` annotation in method implementation (not mentioned in its signature)
- [x] Must fail when the `#[payable]` annotation in method implementation is missing (but mentioned in its signature)

close #1608

## TODOs

- [x] Fix `#[payable]` annotations for the standard library
- [x] Fix the empty parens issue for formatting attributes: #3451
- [x] Parser should allow us write annotations like so: `#[storage(read, write), payable]`: #3452 
- [x] Refactor `#[payable]` annotations in this PR in the style above
- [x] Add more complex formatter unit tests with `payable`
- [x] As pointed out by @mohammadfawaz, the SDK can bypass payable annotations and one option would be to add function attributes to the [JSON ABI schema](https://fuellabs.github.io/fuel-specs/master/protocol/abi/json_abi_format.html) and process it in the SDK (see these issues: https://github.com/FuelLabs/fuel-specs/issues/446 and https://github.com/FuelLabs/fuels-rs/issues/742)
- [x] Bump `fuels-rs`'s version


otrho (approved):
Nice!

anton-trunov:
@otrho Thank you

sezna (dismissed):

test/src/e2e_vm_tests/test_programs/should_pass/payable_non_zero_coins/src/main.sw
```diff
@@ -0,0 +1,18 @@
...
+dep wallet_abi;
+
+use std::constants::ZERO_B256;
+use wallet_abi::Wallet;
+
+fn main() {
+    let contract_address = 0x9299da6c73e6dc03eeabcce242bb347de3f5f56cd1c70926d76526d7ed199b8b;
+    let caller = abi(Wallet, contract_address);
+    // `coins:` is missing (its default value is zero) and hence this is not an error
+    // even that `send_funds` is not payable
```
mohammadfawaz, line 12: The comment here does not match what the test is doing.

sway-error/src/error.rs
```diff
@@ -694,6 +694,23 @@ pub enum CompileError {
     CallingPrivateLibraryMethod { name: String, span: Span },
     #[error("Using \"while\" in a predicate is not allowed.")]
     DisallowedWhileInPredicate { span: Span },
+    #[error("Possibly non-zero amount of coins transferred to non-payable contract method \"{fn_name}\".")]
+    PossiblyNonZeroAmountOfCoinsPassedToNonPayableContractMethod { fn_name: Ident, span: Span },
```
mohammadfawaz, line 698: That's a very long and descriptive error name 😆

test/src/e2e_vm_tests/test_programs/should_fail/non_payable_non_zero_coins_let_binding/src/main.sw
```diff
@@ -0,0 +1,23 @@
...
+
+fn main() {
+    let contract_address = 0x9299da6c73e6dc03eeabcce242bb347de3f5f56cd1c70926d76526d7ed199b8b;
+    let caller = abi(Wallet, contract_address);
+    let amount_to_send = 200;
+    let recipient_address = Address::from(0x9299da6c73e6dc03eeabcce242bb347de3f5f56cd1c70926d76526d7ed199b8b);
+    // calling a non-payable method with a non-zero amount `coins`
+    // this must result in an error
+    // we use a chain of let-bindings that ends in a non-zero literal
+    let fortytwo = 42;
```
mohammadfawaz, line 16: Can we also check that if this values is 0, then we don't emit an error? This would prove that the despite the fact that the number of coins is a _variable_, we're still able to deduce that it's actually just a 0 and not conservatively emit an error.

mohammadfawaz (dismissed):
Looks great! Just some comments on the tests but everything else looks good. Not going to block merging the PR though, so feel free to address the comments in a separate PR.

test/src/e2e_vm_tests/test_programs/should_fail/non_payable_non_zero_coins_let_binding/src/main.sw
anton-trunov: Done in https://github.com/FuelLabs/sway/pull/3450/commits/c4065f8e326f6755a128aedbc7418100705c98c7

sway-error/src/error.rs
anton-trunov: Lol, fixed in https://github.com/FuelLabs/sway/pull/3450/commits/0e659c633a40f0e56b4350b299b1a8eff9968df0

test/src/e2e_vm_tests/test_programs/should_pass/payable_non_zero_coins/src/main.sw
anton-trunov: Good catch! Fixed in https://github.com/FuelLabs/sway/pull/3450/commits/32be2e8bedbed1b9de7e6cf3b63b9be272781487

sezna (approved):

mohammadfawaz (approved):
Excellent work @anton-trunov. :shipit:

anton-trunov:
Thanks everyone for the very helpful feedback!

PR #3454, opened 2022-11-28
New `Bytes` type module for the stdlib

### LINKED ISSUE #3078: Introduce a `Bytes` type in the standard library
This would the same as ` Vec<u8>` without all the padding. So it would be a heap type like `Vec` but not generic, and allows and inserts bytes contiguously (so 8 separate bytes per words, etc.).

nfurfaro:
Adds a new heap type `Bytes` to the stdlib which holds an arbitrary number of packed bytes.

TODO:

- [x] Add docstrings for all functions
- [x] Move helpers to appropriate libs
- [x] add `to_vec_u8()` & `from_vec_u8()` methods
- [x] add `split` & `join` methods

closes #3078


Cargo.toml
```diff
@@ -30,3 +30,6 @@ exclude = [
 
 [profile.dev.package.sway-lsp]
 debug = 2
+
+[patch.crates-io]
```
nfurfaro, line 34: not merging this !

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,564 @@
+library bytes;
+
+use ::{alloc::{alloc, realloc}, vec::Vec};
+use ::assert::assert;
+use ::option::Option;
+use ::logging::log;
+use ::intrinsics::size_of_val;
+
+impl raw_ptr {
```
nfurfaro, line 9: Should these be added to the raw_pointer lib itself?

mohammadfawaz: image

otrho: Woah, emphatic!

sway-lib-core/src/raw_ptr.sw
```diff
@@ -49,4 +49,43 @@ impl raw_ptr {
...
         }
     }
+
+    /// Writes the given byte to the address.
+    pub fn write_byte(self, val: u8) {
+        let val_ptr = asm(r1: val) { r1: raw_ptr };
+        asm(ptr: self, val: val_ptr) {
+            sb ptr val i0;
+        };
+    }
```
otrho, line 59: This worries me a bit.  Has this been tested?

Line 55 works for casting a reference type to an address, because `val` will actually be an implicit pointer.  But when `val` is a copy type, like `u8`, line 55 is just converting the value into a pointer (e.g., if `val` is 10 then you will have an address of 10, not the address of `val`).  At least that's how I think it's working, I'd need to double check.

sway-lib-std/src/alloc.sw
```diff
@@ -43,3 +43,25 @@ pub fn realloc<T>(ptr: raw_ptr, count: u64, new_count: u64) -> raw_ptr {
...
+    if new_count > count {
+        let new_ptr = alloc_bytes(new_count);
+        if count > 0 {
+            ptr.copy_bytes_to(new_ptr, count);
+        };
+        new_ptr
+    } else {
+        ptr
+    }
+}
```
otrho, line 67: Yeah, this is annoying.  `__size_of<u8>` really should be 1.  Ugh.

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,717 @@
...
+#[test()]
+fn test_pop() {
+    let (mut bytes, a, b, c) = setup();
+    assert(bytes.len() == 3);
+
+    let first = bytes.pop();
+
+    assert(first.unwrap() == c);
+    assert(bytes.len() == 2);
+}
```
nfurfaro, line 551: @otrho This tests that we can push and pop bytes. 
`push` is using that `Raw_ptr::write_byte()` under the hood.

otrho: But you don't have a test which is pushing, then popping and confirming the value is correct.  I only saw a test which pushes and then checks the len has incremented..?

sway-lib-core/src/raw_ptr.sw
nfurfaro: This seems to work according to the in-language tests in std::bytes, but ideally I could use `_addr_of` instead of this cast to a `raw_ptr`.

sway-lib-std/src/alloc.sw
nfurfaro: Yeah, I don't love this. It works but could be cleaner. 
I considered adding some conditional branching to the existing types, but there's no simple way to know if the function should treat `u8` as a single byte or as 8 bytes like everywhere else.

sway-lib-std/src/bytes.sw
nfurfaro: Ah. the `setup` function in the test pushes 3 values.
```
fn setup() -> (Bytes, u8, u8, u8) {
    let mut bytes = Bytes::new();
    let a = 5u8;
    let b = 7u8;
    let c = 9u8;
    bytes.push(a);
    bytes.push(b);
    bytes.push(c);
    (bytes, a, b, c)
}
```

otrho: Oh, OK.  I still don't think it should work, but I'll take a look at the final asm and/or reassess how it works in my head. :slightly_smiling_face:

nfurfaro: Update: I've tried to improve the testing for this; the new test for pop is:
```
#[test()]
fn test_pop() {
    let (mut bytes, a, b, c) = setup();
    assert(bytes.len() == 3);
    bytes.push(42u8);
    bytes.push(11u8);
    bytes.push(69u8);
    bytes.push(100u8);
    bytes.push(200u8);
    bytes.push(255u8);
    bytes.push(180u8);
    bytes.push(17u8);
    bytes.push(19u8);
    assert(bytes.len() == 12);

    let first = bytes.pop();
    assert(first.unwrap() == 19u8);
    assert(bytes.len() == 11);

    let second = bytes.pop();
    assert(second.unwrap() == 17u8);
    assert(bytes.len() == 10);

    let third = bytes.pop();
    assert(third.unwrap() == 180u8);
    assert(bytes.len() == 9);

    bytes.pop();
    bytes.pop();
    bytes.pop();
    bytes.pop();
    bytes.pop();
    bytes.pop();
    assert(bytes.len() == 3);

    assert(bytes.pop().unwrap() == c);
    assert(bytes.pop().unwrap() == b);
    assert(bytes.pop().unwrap() == a);

    assert(bytes.pop().is_none() == true);
    assert(bytes.len() == 0);
}
```

otrho: OK, great!  I need to take a closer look and work out why I was mistaken. :+1:

nfurfaro: Hrmm. If I modify `fn write_byte` to look like this, it doesn't appear to break any tests:
```
pub fn write_byte(self, val: u8) {
        asm(ptr: self, val: val) {
            sb ptr val i0;
        };
    }
```

nfurfaro:
I see that the `Vec` lib has `impl<T> AsRawSlice for Vec<T> ` in it. Do we want/need that for `Bytes` ?

nfurfaro:
Re: testing
I've added a bunch of unit tests directly to the lib, and have already enabled running `forc test` in CI in another PR (merged).
So, do we still need to add "e2e" tests for this lib? I feel like unit tests are probably sufficient.

nfurfaro:
I think having generic `from` & `into` traits implemented for `Bytes` would be great.
That said, I'm not sure how I would go about implementing that. I think it would end up achieving something like solidity's `abi.encodePacked`.

otrho:
> So, do we still need to add "e2e" tests for this lib? I feel like unit tests are probably sufficient.

If it's getting fully tested in the CI one way or another then that's fine IMO. :slightly_smiling_face: Unit testing is usually a better fit for libraries anyway.

otrho:
> I think having generic `from` & `into` traits implemented for `Bytes` would be great. That said, I'm not sure how I would go about implementing that. I think it would end up achieving something like solidity's `abi.encodePacked`.

Maybe in a followup issue/PR?

nfurfaro:
Opened #3561 to track future improvements to this.

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
+//! The bytes type is used when a colection of tightly-packed arbitrary bytes is needed.
+library bytes;
+
+use ::{alloc::{alloc_bytes, realloc_bytes}, vec::Vec};
+use ::assert::assert;
+use ::option::Option;
+use ::intrinsics::size_of_val;
```
mohammadfawaz, line 7: Alphabetize :)

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+            i += 1;
+        };
+        vec
+    }
+
+    pub fn split(self, index: u64) -> (Bytes, Bytes) {
+        // TODO: handle edge cases:
+        // index == 0 (also handles the self.len() == 1 case)
+        // index == self.len() - maybe we say index must be < self.len - 1
+        // @todo check rust impl for this type of thing
```
mohammadfawaz, line 515: Should these todos be tracked in an issue?

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+        let mut i = 0;
+        while i < index {
+            first.push(self.get(i).unwrap());
+            i += 1;
+        };
+
+        let mut i2 = index;
+        while i2 < self.len {
+            second.push(self.get(i2).unwrap());
+            i2 += 1;
```
mohammadfawaz, line 528: Isn't `i == index` at this point so might as well use it instead of creating another index `i2`?

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+}
+
+impl core::ops::Eq for Bytes {
+    fn eq(self, other: Self) -> bool {
+        if self.len() == other.len() {
+            let mut i = self.len();
+            while i > 0 {
+                assert(self.get(i - 1).unwrap() == other.get(i - 1).unwrap());
+                i -= 1;
+            };
```
mohammadfawaz, line 559: What is the reasoning for looping backwards here? Isn't this less readable than looping forward?

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+            true
+        } else {
+            false
+        }
```
mohammadfawaz, line 563: You may want to exit early instead of `if/else` to reduce indentation for the `while` above. So:
```rust
if self.len() != other.len() {
    return false;
}

while ...
```

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+    pub fn join(ref mut self, other: self) -> Self {
+        let mut joined = Bytes::with_capacity(self.len + other.len);
+
+        let mut i = 0;
+        while i < self.len() {
+            joined.push(self.get(i).unwrap());
+            i += 1;
+        };
+
+        let mut i2 = 0;
```
mohammadfawaz, line 542: I think name shadowing is fine here, or just re-initialize `i` to 0. I just find `i2` not as elegant, but that's a personal opinion so feel free to ignore it :)

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+        let mut i = 0;
+        let length = self.len;
+        while i < length {
+            vec.push(self.get(i).unwrap());
+            i += 1;
+        };
+        vec
+    }
+
+    pub fn split(self, index: u64) -> (Bytes, Bytes) {
```
mohammadfawaz, line 511: `split` doesn't have doc string comments

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+
+        let mut i2 = index;
+        while i2 < self.len {
+            second.push(self.get(i2).unwrap());
+            i2 += 1;
+        };
+        (first, second)
+    }
+
+    pub fn join(ref mut self, other: self) -> Self {
```
mohammadfawaz, line 533: `split` doesn't have doc string comments

Cargo.toml
```diff
@@ -30,3 +30,4 @@ exclude = [
 
 [profile.dev.package.sway-lsp]
 debug = 2
+
```
mohammadfawaz, line 33: Maybe keep this file untouched by reverting it to how it is on `master`.

sway-lib-std/src/alloc.sw
```diff
@@ -43,3 +43,25 @@ pub fn realloc<T>(ptr: raw_ptr, count: u64, new_count: u64) -> raw_ptr {
...
+
+/// Reallocates the given area of memory in individual bytes
+pub fn realloc_bytes(ptr: raw_ptr, count: u64, new_count: u64) -> raw_ptr {
+    if new_count > count {
+        let new_ptr = alloc_bytes(new_count);
+        if count > 0 {
+            ptr.copy_bytes_to(new_ptr, count);
+        };
```
mohammadfawaz, line 62:
```suggestion
        }
```

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+    /// let b = 7u8;
+    /// bytes.push(a);
+    /// bytes.push(b);
+    /// assert(bytes.len() == 2);
+    /// ```
+    pub fn push(ref mut self, byte: u8) {
+        // If there is insufficient capacity, grow the buffer.
+        if self.len == self.buf.capacity() {
+            self.buf.grow();
+        };
```
mohammadfawaz, line 123:
```suggestion
        }
```

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+    ///
+    /// bytes.push(5);
+    /// let res = bytes.pop();
+    /// assert(res.unwrap() == 5);
+    /// assert(bytes.is_empty());
+    /// ```
+    pub fn pop(ref mut self) -> Option<u8> {
+        if self.len == 0 {
+            return Option::None;
+        };
```
mohammadfawaz, line 157:
```suggestion
        }
```

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+    /// let item = bytes.get(1).unwrap();
+    /// assert(item == 10u8);
+    /// let opt = bytes.get(10);
+    /// assert(opt.is_none()); // index out of bounds
+    /// ```
+    pub fn get(self, index: u64) -> Option<u8> {
+        // First check that index is within bounds.
+        if self.len <= index {
+            return Option::None;
+        };
```
mohammadfawaz, line 186:
```suggestion
        }
```

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+    /// assert(bytes.get(2).unwrap() == c);
+    /// ```
+    pub fn from_vec_u8(ref mut vec: Vec<u8>) -> Self {
+        let mut bytes = Bytes::new();
+        let mut i = 0;
+        let length = vec.len();
+        while i < length {
+            bytes.push(vec.get(i).unwrap());
+            i += 1;
+        };
```
mohammadfawaz, line 472:
```suggestion
        }
```

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+    /// assert(vec.get(2).unwrap() == c);
+    /// ```
+    pub fn into_vec_u8(self) -> Vec<u8> {
+        let mut vec = Vec::new();
+        let mut i = 0;
+        let length = self.len;
+        while i < length {
+            vec.push(self.get(i).unwrap());
+            i += 1;
+        };
```
mohammadfawaz, line 507:
```suggestion
        }
```

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+        // index == self.len() - maybe we say index must be < self.len - 1
+        // @todo check rust impl for this type of thing
+        assert(index < self.len());
+        let mut first = Bytes::with_capacity(index);
+        let mut second = Bytes::with_capacity(self.len() - index);
+        let mut i = 0;
+        while i < index {
+            first.push(self.get(i).unwrap());
+            i += 1;
+        };
```
mohammadfawaz, line 523:
```suggestion
        }
```

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+
+        let mut i2 = index;
+        while i2 < self.len {
+            second.push(self.get(i2).unwrap());
+            i2 += 1;
+        };
```
mohammadfawaz, line 529:
```suggestion
        }
```

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+    }
+
+    pub fn join(ref mut self, other: self) -> Self {
+        let mut joined = Bytes::with_capacity(self.len + other.len);
+
+        let mut i = 0;
+        while i < self.len() {
+            joined.push(self.get(i).unwrap());
+            i += 1;
+        };
```
mohammadfawaz, line 540:
```suggestion
        }
```

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+
+        let mut i2 = 0;
+        while i2 < other.len() {
+            joined.push(other.get(i2).unwrap());
+            i2 += 1;
+        };
```
mohammadfawaz, line 546:
```suggestion
        }
```

sway-lib-std/src/bytes.sw
mohammadfawaz, line 559:
```suggestion
            }
```

sway-lib-std/src/bytes.sw
```diff
@@ -0,0 +1,849 @@
...
+
+    let mut joined = bytes.join(bytes2);
+    assert(joined.len() == bytes.len() + bytes2.len());
+    assert(joined.capacity() == bytes.len() + bytes2.len());
+    let values = [a, b, c, d, e, f];
+    let mut i = 0;
+    while i < 6 {
+        assert(joined.get(i).unwrap() == values[i]);
+        i += 1;
+    };
```
mohammadfawaz, line 848:
```suggestion
    }
```

mohammadfawaz (commented):
This is great! Thanks @nfurfaro. Just some small comments.

sway-lib-std/src/bytes.sw
nfurfaro: I've been working under the assumption that a semicolon after an `if` statement is needed _unless_ I want to implicitly return the resulting value?

sway-lib-std/src/bytes.sw
nfurfaro: Ah, I will review these and either remove or track them in issues as needed.

sway-lib-std/src/bytes.sw
mohammadfawaz: This is no longer a problem and these semicolons are only required if you have something like `let x = if cond { 5 } else { 6 };`

nfurfaro:
>Is there any way to form a Bytes (or a Vec for that matter) without instantiating an empty one and pushing values one by one?

As far as I know, no.

fwiw, `Bytes::push` is using the `sb` opcode under the hood (via `Raw_vec::write_byte`), so it's already pretty low-level.
We basically have to handle the creation of a `Bytes` type one byte at a time or else we're back to padded values again.

mohammadfawaz:
> > Is there any way to form a Bytes (or a Vec for that matter) without instantiating an empty one and pushing values one by one?
> 
> As far as I know, no.
> 
> fwiw, `Bytes::push` is using the `sb` opcode under the hood (via `Raw_vec::write_byte`), so it's already pretty low-level. We basically have to handle the creation of a `Bytes` type one byte at a time or else we're back to padded values again.

Yeah we'd need to be able to support something like `let vec: Vec<u64> = vec![1, 2, 3]` which we don't have yet.

sway-lib-std/src/bytes.sw
nfurfaro: yes, I can reuse `i` here.

sway-lib-std/src/bytes.sw
nfurfaro: There was no good reason. 
I found a better way to compare without looping.

nfurfaro:
I've addressed all comments.
- `Eq` has been refactored to use `meq` rather than looping, and I added a test for it.
- Docstrings added to `split` & `join`
- various other refactorings as suggested.

mohammadfawaz (dismissed):
Looks great! Thank you @nfurfaro.

mohammadfawaz (approved):

otrho (approved):
Am loving the distinct lack of ASM blocks. :+1:

Issue #3449, opened 2022-11-28
Cannot perform bitwise operations on implicitly upcasted u32 -> u64


tkporter:
To my understanding, there is only implicit upcasting from u32 -> u64. I also believe bitwise operations are only possible for u64.

If I have a u32 that I want to upcast to u64 so I can perform bitwise operations, I'm unable to do so. Despite explicitly declaring a variable as u64, the compiler still thinks it's a u32 somewhere under the hood:

```
contract;

pub fn test_function() {
    let some_u32: u32 = 4u32;

    let some_u64: u64 = some_u32;
    let foo = some_u64 & 1u64;
}
```

When compiling, this gives:
```
$ forc build
  Compiled library "core".
  Compiled library "std".
error
 --> /Users/trevor/fuel/u32-bitwise/src/main.sw:7:24
  |
5 |
6 |     let some_u64: u64 = some_u32;
7 |     let foo = some_u64 & 1u64;
  |                        ^ No method named "binary_and" found for type "u32".
8 | }
  |
____

  Aborting due to 1 error.
Error: Failed to compile u32-bitwise
```

mohammadfawaz:
> To my understanding, there is only implicit upcasting from u32 -> u64

This is correct.

> I also believe bitwise operations are only possible for u64.

This should be rectified. There is no good reason for not allow bitwise operations for `u32`, `u16`, and `u8`.

>  Despite explicitly declaring a variable as u64, the compiler still thinks it's a u32 somewhere under the hood.

This, I believe, is a bug that can also be exercised with this simple example:
```rust
fn main() -> u8 {
    let x: u8 = 4u32;
}
```
I think we're not updating the type of `x` in the type engine, as per [this comment](https://github.com/FuelLabs/sway/blob/master/sway-core/src/type_system/unify.rs#L555-L560). @emilyaherbert can you shed some light here as to why we're not updating the concurrent slab at all?

mohammadfawaz:
Actually just found https://github.com/FuelLabs/sway/pull/1540

mohammadfawaz:
So I'm proposing something in https://github.com/FuelLabs/sway/pull/3458

PR #3458, opened 2022-11-28
Fix type inference for integer variables and constants with type ascriptions

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/3449

Basically, when deciding what the type of a `VariableExpression` or a `ConstantExpression` is, rely on a new field called `return_type`. This `return_type` is set to:
* The type of the value assigned to the variable/constant if the type ascription is not an integer. This assumes that the type ascription has already been checked against the value assigned.
* The type ascription otherwise.

The reasoning for this is that integers are special because they support implicit casting. While the type of the body contain useful information, it is not always sufficient. For example for `let x: u8 = 5u16;`, the compiler currently infers the type `u16` for `x`, ignoring `u8`. The type ascription is _checked_ against `u16` (hence generating a downcast warning etc.), but not actually used as the actual type of `x`.


sway-core/src/language/ty/declaration/variable.rs
```diff
@@ -7,6 +7,7 @@ pub struct TyVariableDeclaration {
     pub name: Ident,
     pub body: TyExpression,
     pub mutability: VariableMutability,
+    pub return_type: TypeId,
     pub type_ascription: TypeId,
```
mohammadfawaz, line 11: I'm actually questionning the usefulness of `type_ascription` here 🤔  - we don't seem to have it for `ConstantDeclaration` and it's been consumed already. All we care about is the actual type now which `return_type` will hold. cc @emilyaherbert what do you think?

otrho (approved):

vaivaswatha (approved):

anton-trunov:
`let x2: u8 = 4u16` looks really confusing to me, sorry.

Rust, for instance, complains about this piece of code:

```rust
fn foo() {
    let x2: u8 = 4u16;
}
```

Here is the error message:
```
  |
2 |     let x2: u8 = 4u16;
  |             --   ^^^^ expected `u8`, found `u16`
  |             |
  |             expected due to this
  |
help: change the type of the numeric literal from `u16` to `u8`
  |
2 |     let x2: u8 = 4u8;
  |                   ~~
```

mohammadfawaz:
> `let x2: u8 = 4u16` looks really confusing to me, sorry.
> 
> Rust, for instance, complains about this piece of code:
> 
> ```rust
> fn foo() {
>     let x2: u8 = 4u16;
> }
> ```
> 
> Here is the error message:
> 
> ```
>   |
> 2 |     let x2: u8 = 4u16;
>   |             --   ^^^^ expected `u8`, found `u16`
>   |             |
>   |             expected due to this
>   |
> help: change the type of the numeric literal from `u16` to `u8`
>   |
> 2 |     let x2: u8 = 4u8;
>   |                   ~~
> ```
> 
> (sorry I disabled auto-merge just to have one more chance to discuss this; I might have totally misunderstood it)

@anton-trunov this makes a lot of sense. As discussed, the right approach here is completely remove implicit casting though there are some things to figure out with that related to math operations. https://github.com/FuelLabs/sway/issues/3470 now tracks that work. For now, this is merged to fix the current bug.

PR #3460, opened 2022-11-29
Do not progress with IR generation if `deterministically_aborts`

### LINKED ISSUE #3303: ICE on simple code with returns statements in `if/else`
Code:
```rust
contract;

abi MyContract {
    fn foo() -> u64;
}

    fn foo() -> u64 {
        let x: u64 = if true {
            return 0;
        } else {
            return 1;
        };

        return 2;
    }
}
```
This is a reduced version of a project provided by someone in Discord. The intent was to write this:
```rust
    fn foo() -> u64 {
        let x: u64 = if true {
            0
        } else {
            1
        };

        return 2;
    }
```
which works, but the original code should not cause an ICE either.
ICE:
```console
error: Internal compiler error: Verification failed: Store value and pointer type mismatch.
```

vaivaswatha:
Fixes #3303 

The existing check for `deterministically_aborts` assumes that function applications are inlined. This shouldn't be so for the newly added call to `determinisitically_aborts`. So I've added a new parameter to the `DeterministicallyAborts` trait to switch considering "calls are inlined" or not semantics. I'm not sure if this is the best way, so I'm open to suggestions.


otrho:
What actual difference does it make to the algorithm when the calls are inlined or not?  I would've thought at that stage inlining is a hidden implementation detail.

vaivaswatha:
> What actual difference does it make to the algorithm when the calls are inlined or not? I would've thought at that stage inlining is a hidden implementation detail.

On `master`
```rust
fn diverge_in_let_body() -> u64 {                                                                                                                                                                            
    let x: bool =  {                                                                                                                                                                                         
        return 5;                                                                                                                                                                                            
    };                                                                                                                                                                                                       
    123                                                                                                                                                                                                      
} 
```
compiles fine because, the `x = ...` assignment isn't even compiled down to the IR. This is because  it passes the [init_val.is_diverging()](https://github.com/FuelLabs/sway/pull/3460/files#diff-bfae318752d54fae537a316de96038bab0ff0e1eebe44643d63ce9bcca784de7L1476) check.

In the same way (for the same reason), we shouldn't proceed for IR generation if the body `deterministically_aborts`, such as in this code:

```rust
        let x: u64 = if true {
            return 0;
        } else {
            return 1;
        };
```

Now the problem is, this code could as well be

```rust
        let x: u64 = if true {
           foo(x);
        } else {
           foo(y);
        };
```
and `foo` can have a return inside it. But that doesn't mean that *that* return makes this caller `deterministically_aborts`. That's what happens without my change to `deterministically_aborts` to ignore function bodies at a call site.

I experimented with modifying `deterministically_aborts` to unconditionally return `false` for function applications. That seemed to cause a lot of other things to fail, so decided to go for this approach.

vaivaswatha:
> What actual difference does it make to the algorithm when the calls are inlined or not? I would've thought at that stage inlining is a hidden implementation detail.

I'm guessing that the name I gave that parameter may be what's causing the confusion. Maybe I can change it to `check_call_body` ?

otrho:
Hrmm, right OK.  Feels a bit hacky, but it sounds like it's the best fix.

> > What actual difference does it make to the algorithm when the calls are inlined or not? I would've thought at that stage inlining is a hidden implementation detail.
> 
> I'm guessing that the name I gave that parameter may be what's causing the confusion. Maybe I can change it to `check_call_body` ?

Yep, I think that's a bit better.

vaivaswatha:
> Hrmm, right OK. Feels a bit hacky, but it sounds like it's the best fix.
> 
> > > What actual difference does it make to the algorithm when the calls are inlined or not? I would've thought at that stage inlining is a hidden implementation detail.
> > 
> > 
> > I'm guessing that the name I gave that parameter may be what's causing the confusion. Maybe I can change it to `check_call_body` ?
> 
> Yep, I think that's a bit better.

Done.

otrho (approved):

PR #3463, opened 2022-11-29
LSP: Refactor the API

### LINKED ISSUE #3459: LSP: Refactor the server API
When showing @sdankel the current API during the offsite it became obvious that it could use a bit of restructuring to make it more intuitive. 

For example, we could wrap our `TokenMap` type in a struct so we can define methods that deal directly with the map on the type directly, rather than having them live in `session.rs`.

JoshuaBatty:
No new methods or functionality have been added. This PR just cleans up the API so everything is a little more obvious and also improves the documentation.

closes #3459


sway-lsp/src/core/token.rs
```diff
@@ -91,3 +62,113 @@ pub enum SymbolKind {
...
+        Self {
+            parsed: token,
+            typed: None,
+            type_def: None,
+            kind,
+        }
+    }
+
+    /// Return the `Ident` of the declaration of the provided token.
+    pub fn declared_token_ident(&self, type_engine: &TypeEngine) -> Option<Ident> {
```
sdankel, line 94: Is the need to pass `type_engine` because the compiler no longer uses lazy statics (this tells it which session we're using)?

sdankel (commented):
Nice! This makes the interface between a session and its token map much clearer.

sway-lsp/src/core/token.rs
JoshuaBatty: Yeah pretty much. Internally we call `ident_of_type_id` which converts from a `TypeId` to a `TypeInfo`. There is a method on the type_engine that does this `type_engine.look_up_type_id(*type_id)`. So in order to have the `TypeId` return the correct `TypeInfo`, we need to pass in the type_engine associated with the current session.

sway-lsp/src/core/token_map.rs
```diff
@@ -0,0 +1,124 @@
...
+            Some(ident) => self.get(&token::to_ident_key(&ident)).map(|item| {
+                let ((ident, _), token) = item.pair();
+                (ident.clone(), token.clone())
+            }),
+            None => None,
+        }
+    }
+
+    /// Uses the TypeId to find the associated TypedDeclaration in the TokenMap.
+    pub(crate) fn declaration_of_type_id(
```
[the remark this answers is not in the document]
JoshuaBatty: This is handy when dealing with tokens that are of the `Expression` type in the AST.

```
pub struct TyExpression {
    pub expression: TyExpressionVariant,
    pub return_type: TypeId,
    pub span: Span,
}
```

In this instance, the return_type field is a `TypeId`. We can then use that `TypeId` to retrieve the declaration Token.

sway-lsp/src/core/token_map.rs
```diff
@@ -0,0 +1,124 @@
...
+        }
+        None
+    }
+
+    /// Check if the code editor's cursor is currently over one of our collected tokens.
+    pub(crate) fn token_at_position(
+        &self,
+        uri: &Url,
+        position: Position,
+    ) -> Option<(Ident, Token)> {
```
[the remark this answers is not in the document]
JoshuaBatty: Good question! The `Token` type might contain multiple Idents or Spans. So it's not clear exactly which one is to be considered to correct one. The only real place we can infer the correct `Ident` is during the AST traversal when we populate the `TokenMap`. As such, this method is returning the `Token` & the sort of, "source of truth" `Ident` within that token.

sway-lsp/src/capabilities/document_symbol.rs
```diff
@@ -53,3 +38,17 @@ pub(crate) fn symbol_kind(symbol_kind: &SymbolKind) -> lsp_types::SymbolKind {
...
+// TODO: the "deprecated: None" field is deprecated according to this library
+fn symbol_info(ident: &Ident, token: &Token, url: Url) -> SymbolInformation {
+    let range = get_range_from_span(&ident.span());
+    SymbolInformation {
+        name: ident.as_str().to_string(),
+        kind: symbol_kind(&token.kind),
+        location: Location::new(url, range),
+        tags: None,
+        container_name: None,
+        deprecated: None,
```
sdankel, line 52: Could we use `..Default::default()` here and get rid of the warning?

sway-lsp/src/capabilities/formatting.rs
```diff
@@ -1,9 +1,8 @@
+use crate::error::LanguageServerError;
 use std::sync::Arc;
 use swayfmt::Formatter;
 use tower_lsp::lsp_types::{Position, Range, TextEdit};
 
-use crate::error::LanguageServerError;
```
sdankel, line 5: I'm surprised clippy allowed this to get checked in 😅

sway-lsp/src/capabilities/semantic_tokens.rs
```diff
@@ -36,15 +35,15 @@ pub fn semantic_tokens_full(session: Arc<Session>, url: &Url) -> Option<Semantic
 /// Tokens are encoded relative to each other.
 ///
 /// This is taken from rust-analyzer which is also a direct port of <https://github.com/microsoft/vscode-languageserver-node/blob/f425af9de46a0187adb78ec8a46b9b2ce80c5412/server/src/sematicTokens.proposed.ts#L45>
-pub(crate) struct SemanticTokensBuilder {
+struct SemanticTokensBuilder {
```
sdankel, line 38: 👍

sdankel (approved):
🚀

sway-lsp/src/capabilities/document_symbol.rs
JoshuaBatty: Unfortunately `SymbolInformation` doesn't have a impl of `Default` so we need to do this :(

emilyaherbert:
LGTM. Just a few comments---I'm happy to approve once those are resolved.

JoshuaBatty:
Hi @emilyaherbert I can't seem to locate your comments, unfortunately. :(

sway-lsp/src/capabilities/completion.rs
```diff
@@ -1,10 +1,29 @@
...
+    token::{AstToken, SymbolKind, Token, TypedAstToken},
+    token_map::TokenMap,
+};
 use tower_lsp::lsp_types::{CompletionItem, CompletionItemKind};
 
-pub fn to_completion_items(token_map: &TokenMap) -> Vec<CompletionItem> {
+pub(crate) fn to_completion_items(token_map: &TokenMap) -> Vec<CompletionItem> {
     let mut completion_items = vec![];
 
+    let is_initial_declaration = |token_type: &Token| -> bool {
```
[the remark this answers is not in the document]
JoshuaBatty: Sure that's fine, can do.

sway-lsp/src/capabilities/inlay_hints.rs
```diff
@@ -38,6 +41,7 @@ pub(crate) fn inlay_hints(
     let type_engine = session.type_engine.read();
 
     let hints: Vec<lsp_types::InlayHint> = session
+        .token_map
```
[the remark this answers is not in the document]
JoshuaBatty: Good catch, this was just an oversight on my behalf. I've made it so you can only access it using `token_map()` now. thanks.

sway-lsp/src/core/session.rs
```diff
@@ -47,19 +45,19 @@ pub struct Session {
     pub token_map: TokenMap,
     pub runnables: DashMap<RunnableType, Runnable>,
     pub compiled_program: RwLock<CompiledProgram>,
-    pub sync: SyncWorkspace,
     pub type_engine: RwLock<TypeEngine>,
+    pub sync: SyncWorkspace,
 }
 
 impl Session {
```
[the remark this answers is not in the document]
JoshuaBatty: Absolutely :)

JoshuaBatty:
Thanks, @emilyaherbert have just addressed your feedback. Hoping to get this in soon so it doesn't clash with your DE lazy static removal PR.

sdankel (approved):

PR #3464, opened 2022-11-29
Add code action for generating ABI impl

sdankel:
Closes https://github.com/FuelLabs/sway/issues/3518

Adds a code action for generating an impl from an ABI definition.

For example, 

image

```
abi FizzBuzz {
    fn fizzbuzz(input: u64) -> FizzBuzzResult;
}
```

generates:

```
impl FizzBuzz for Contract {
    fn fizzbuzz(input: u64) -> FizzBuzzResult {}
}
```

It works for methods with attributes and missing return types as well:

```
abi Ownership {
    #[storage(read, write)]
    fn set_owner(owner: Option<Identity>);

    #[storage(read)]
    fn action();
}
```

generates:

```
impl Ownership for Contract {
    #[storage(read, write)]
    fn set_owner(owner: Option<Identity>) {}

    #[storage(read)]
    fn action() {}
}
```


JoshuaBatty (commented):
This is really cool @sdankel 🥳  Excited to have this scaffolding in the server!

sway-lsp/src/capabilities/code_actions/abi_impl.rs
```diff
@@ -0,0 +1,96 @@
...
+                })
+        })
+        .collect::<Vec<String>>()
+        .join("\n")
+}
+
+fn get_contract_impl_string(abi_decl: TyAbiDeclaration) -> String {
+    let contract_name = abi_decl.name.to_string();
+    format!(
+        "\nimpl {} for Contract {{\n{}\n}}",
```
JoshuaBatty, line 92: Can we change this line to
```
"\nimpl {} for Contract {{{}\n}}",
```

Otherwise, there is an unnecessary empty line of code between the `impl` and the attributes of the first method

sway-lsp/src/capabilities/code_actions/abi_impl.rs
```diff
@@ -0,0 +1,96 @@
...
+        edit: Some(WorkspaceEdit {
+            changes: Some(text_edit_map),
+            ..Default::default()
+        }),
+        data: Some(Value::String(uri.to_string())),
+        ..Default::default()
+    })
+}
+
+fn get_function_signatures(abi_decl: TyAbiDeclaration) -> String {
```
JoshuaBatty, line 41: Excellent, this is exactly what I had in mind!

sway-lsp/src/capabilities/code_actions/abi_impl.rs
sdankel: Good catch!

sdankel:
Can't use `return_type_span` until this is resolved: https://github.com/FuelLabs/sway/issues/3594

JoshuaBatty:
> Can't use `return_type_span` until this is resolved: #3594

Can we match on the type_info returning an empty tuple instead? This still means the function has no return type and we have access to that information in the compiler currently.

sdankel:
> > Can't use `return_type_span` until this is resolved: #3594
> 
> Can we match on the type_info returning an empty tuple instead? This still means the function has no return type and we have access to that information in the compiler currently.

Yep, updated the PR to do that :)

JoshuaBatty (approved):
:shipit:

Voxelot (approved):

PR #3469, opened 2022-11-29
Allow comma-separated annotations: `#[ann1, ann2, ...]`

### LINKED ISSUE #3452: Parser does not allow us to write annotations as #[ann1, ann2, ...]
For instance, instead of

```
#[storage(read, write), payable]
fn foo ();
```

I have to write it as

```
#[storage(read, write)]
#[payable]
fn foo ();
```

in PR #3450

### LINKED ISSUE #3475: Format #[doc(foobar)] annotations in its original syntax, not as ///foobar?
### Preamble

Rust uses `#[doc(foobar)]` annotations to control how code docs are rendered and it uses another syntax as an alternative to the triple slash `///`: `#[doc = " This is a doc comment."]`.

### Issue

In Sway, currently, I believe, it's not possible to write useful comments using the `#[doc(foobar)]` annotations, because the parser expects `foobar` to be an identifier, and the user cannot put a string there.

Moreover, `#[doc(foo,bar,baz)]` gets printed as `///baz` by the Sway formatter.

### Possible solution

My proposal would be to
- stop printing `#[doc(foobar)]` as `///foobar` and print it in its original syntax,
- add a new attribute kind (let's say `doc-comment`) and parse `/// foo bar baz` into this new attribute kind, hence decoupling doc comments from doc controlling annotations.

Any comments on this proposal are more than welcome.

anton-trunov:
close #3452

### TODOs

- [x] Fix sway-parse tests
- [x] See if we can resolve #3475 as proposed there
- [x] Deal with the new `.pop()`s


swayfmt/src/utils/language/attribute.rs
```diff
@@ -40,41 +40,43 @@ impl Format for AttributeDecl {
...
-                    };
-                    // `]\n`
-                    Self::close_square_bracket(formatted_code, formatter)?;
-
-                    Ok(())
-                },
-            )?;
         }
-
+        formatted_code.pop();
```
eureka-cpu, line 76: If you can find a way around using `.pop()` or destructive means it will narrow down instances of mysterious formatter failures creating unparsable code! There may be other instances that are in the formatter at present, but we should try to avoid adding any more and look to review the others at some point.

swayfmt/src/utils/language/attribute.rs
```diff
@@ -40,41 +40,43 @@ impl Format for AttributeDecl {
...
-        let attr = self.attribute.get();
-        if attr.name.as_str() == "doc" {
-            if let Some(Some(doc_comment)) = attr
-                .args
-                .as_ref()
-                .map(|args| args.inner.final_value_opt.as_ref())
-            {
-                writeln!(formatted_code, "///{}", doc_comment.as_str().trim_end())?;
+        // `#`
+        write!(formatted_code, "{}", self.hash_token.span().as_str())?;
```
eureka-cpu, line 44: The formatter check is failing here because the hash token is being written before checking if the attribute is a `doc` type

swayfmt/src/utils/language/attribute.rs
```diff
@@ -40,41 +40,43 @@ impl Format for AttributeDecl {
...
-        } else {
-            formatter.with_shape(
-                formatter.shape.with_default_code_line(),
-                |formatter| -> Result<(), FormatterError> {
-                    // `#`
-                    write!(formatted_code, "{}", self.hash_token.span().as_str())?;
-                    // `[`
-                    Self::open_square_bracket(formatted_code, formatter)?;
-                    // name e.g. `storage`
-                    write!(formatted_code, "{}", attr.name.span().as_str())?;
```
eureka-cpu, line 61: I think the part you would need to change is here if you're trying to have more than one attribute

test/src/e2e_vm_tests/test_programs/should_pass/language/test_multiple_attributes/src/main.sw
```diff
@@ -2,4 +2,3 @@ library test_multiple_attributes;
 
 #[test, inline(always)]
 fn foo() {}
```
sezna, line 4: maybe we should add some more complicated formatter tests that involve more attributes or duplicate attributes?

anton-trunov: Added a bit better unit test for the formatter and refactored the simple one in `swayfmt/src/formatter/mod.rs`. We don't deal with duplicate attributes at the moment, unfortunately -- I opened an issue on that https://github.com/FuelLabs/sway/issues/3598

anton-trunov: Also, once we have the `payability` attribute, I can add more formatter tests (there is not too many attributes to make complex examples right now)

otrho (approved):

PR #3476, opened 2022-11-30
ci: use cross to fix arm64 release step

### LINKED ISSUE #2726: CI: Packed binaries for apple arm arch seems to be in intel arch
Realized while installing on a fresh m1 system without any rosetta installed. For more information please see https://github.com/FuelLabs/fuelup/issues/177.

I think I narrowed down the issue to sway repo CI since looking at `file` output of the binaries I get from
https://github.com/FuelLabs/sway/releases/download/v0.23.0/forc-binaries-darwin_arm64.tar.gz
results:

```
forc: Mach-O 64-bit executable x86_64
```

But if I built them myself and check `file forc` I can see:

```
forc: Mach-O 64-bit executable arm64
```

So I am guessing the binaries placed in the arm64 zip seems to be in Intel arch. Because of this default installation with `fuelup` fails if rosetta2 is not installed and even if it is installed by default normally, it would be better to have native mac os arm binaries working as many people use these devices now.

bingcicle:
Closes #2726 

The issue was with the `cargo install` step, which resulted in the binaries being built and compiled on the `macos:latest` GitHub runner (which is on x86-apple-darwin ([source](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources))). This is why our "arm64" binaries were actually still in "amd64" format.

A direct fix would be to update the step to use `cargo build --target ${{ matrix.job.os }}` instead, but I opted to go for an indirect fix which is to use `cross` for 2 reasons: 1) standardization, since we're using that for all other repos anyway, and 2) this allows us to skip some of the setup steps as well.

I have a personal repo setup specifically to test this and the releases are here (albeit with older binary versions because I haven't yet pulled changes there): https://github.com/bingcicle/sway/releases/tag/nightly-2022-11-30. 

Output of `file forc` for each of the tarballs:

amd64 Linux:
`forc: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=3f51b5c41c1c915d117cb12214808a417d695dd7, with debug_info, not stripped`

arm64 Linux:
`forc: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, BuildID[sha1]=44bc7da687b369f5bf6bc8890ea97b31c6c4023a, with debug_info, not stripped`

amd64 Darwin:
`forc: Mach-O 64-bit x86_64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE|HAS_TLV_DESCRIPTORS>`

arm64 Darwin:
`forc: Mach-O 64-bit arm64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE|HAS_TLV_DESCRIPTORS>`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -541,8 +541,8 @@ jobs:
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_NOTIFY_BUILD }}
 
-  install-and-upload-forc-binaries:
-    name: Install and upload forc binaries to release
+  build-and-release-forc-binaries:
+    name: Build and upload forc binaries to release
     runs-on: ${{ matrix.job.os }}
     if: github.event_name == 'release' && github.event.action == 'published'
     needs: publish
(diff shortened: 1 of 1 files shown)


kayagokalp (approved):
🚢

mitchmindtree (approved):
Nice work on tracking this down @bingcicle, looks like a tricky one!

PR #3484, opened 2022-12-01
add file locks to prevent git checkout races

### LINKED ISSUE #919: Add a `Filesystem` abstraction similar to cargo's for managing access to the `forc` user directory

### LINKED ISSUE #2603: Running multiple instances of `forc` concurrently can (very occasionally) cause a git checkout race
I'm working on a project that compiles multiple sway projects concurrently spawning several forc processes.

Locally everything worked. When I first ran the CI I got the following error ([link](https://github.com/FuelLabs/fuels-rs/runs/7939488743?check_suite_focus=true)):
```
Error: Following Sway projects could not be built:
- foo_caller_contract, Running command: '"/usr/local/bin/forc" "build" "--silent" "--output-directory" "/home/runner/work/fuels-rs/fuels-rs/fuel-e2e-tests/xtask/../assets/foo_caller_contract" "--path" "/home/runner/work/fuels-rs/fuels-rs/fuel-e2e-tests/sway_projects/foo_caller_contract"' failed with status: false.   Fetching git+https://github.com/fuellabs/sway?tag=v0.19.2#6808861389966f99887f71476918a562a9edd90e
Error: could not open '/home/runner/.forc/git/checkouts/std-57e47d01450577b4/6808861389966f99887f71476918a562a9edd90e/sway-core/src/ir_generation/function.rs' for writing: No such file or directory; class=Os (2)
 :
- auth_testing_abi, Running command: '"/usr/local/bin/forc" "build" "--silent" "--output-directory" "/home/runner/work/fuels-rs/fuels-rs/fuel-e2e-tests/xtask/../assets/auth_testing_abi" "--path" "/home/runner/work/fuels-rs/fuels-rs/fuel-e2e-tests/sway_projects/auth_testing_abi"' failed with status: false.   Fetching git+https://github.com/fuellabs/sway?tag=v0.19.2#6808861389966f99887f71476918a562a9edd90e
Error: failed to stat '/home/runner/.forc/git/checkouts/std-57e47d01450577b4/6808861389966f99887f71476918a562a9edd90e/scripts/prism/components.json': No such file or directory; class=Os (2)
 :
- contract_output_test, Running command: '"/usr/local/bin/forc" "build" "--silent" "--output-directory" "/home/runner/work/fuels-rs/fuels-rs/fuel-e2e-tests/xtask/../assets/contract_output_test" "--path" "/home/runner/work/fuels-rs/fuels-rs/fuel-e2e-tests/sway_projects/contract_output_test"' failed with status: false.   Fetching git+https://github.com/fuellabs/sway?tag=v0.19.2#6808861389966f99887f71476918a562a9edd90e
Error: failed to stat '/home/runner/.forc/git/checkouts/std-57e47d01450577b4/6808861389966f99887f71476918a562a9edd90e/docs/src/sway-program-types/index.md': No such file or directory; class=Os (2)
```

There seems to be a race on the checked-out sway stdlib. I don't know the exact cause but [this code](https://github.com/FuelLabs/sway/blob/0929300e967425f4ef1903de7b9c3ff1efb1af7e/forc-pkg/src/pkg.rs#L1308) seems to be doing git checkouts and deletions on deterministic folder names without any inter-process-locking surrounding it.

bingcicle:
Closes #2603 
Related #919

~Introduces Filesystem abstraction as well as locking. Mostly inspired by `cargo`'s flock but using [file-lock](https://docs.rs/file-lock/latest/file_lock/) instead.~

EDIT: We've abandoned the attempt to introduce Filesystem abstraction (see [comment](https://github.com/FuelLabs/sway/pull/3484#pullrequestreview-1201893997) below). This PR will introduce locks only, and instead of [file-lock](https://docs.rs/file-lock/latest/file_lock/), we will use [fd-lock](https://docs.rs/fd-lock/3.0.8/fd_lock/).

Similar to `cargo` we open a `.forc-lock` file at the root of the directory we want to lock - this is unlocked on Drop.

Tested this on a fork of `fuels-rs` in my personal repo [here](https://github.com/bingcicle/fuels-rs/actions/runs/3591002787/jobs/6045027347#step:9:18). This is using the same setup in the issue raised here: https://github.com/FuelLabs/sway/issues/2603

On first glance in terms of speed, it doesn't seem to offer much improvement over the time spent building contracts when ran with `-n 1`, at least from just glancing at the duration of the build step in the fuels-rs repo. Improvements in speed aside, introducing `fd-lock` at least seems to allow concurrent builds.


forc-fs/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.31.3"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+homepage = "https://fuel.network/"
+license = "Apache-2.0"
+repository = "https://github.com/FuelLabs/sway"
+description = "Filesystem abstraction for Forc"
+
+[dependencies]
+file-lock = "2.1"
```
mitchmindtree, line 12: Out of curiosity, what prompted the choice to use `file-lock` here? It looks like there are a few alternatives that have a lot more use like [fslock](https://crates.io/crates/fslock) or [fd-lock](https://crates.io/crates/fd-lock) - maybe it's worth switching to one of these more established options?

forc-pkg/src/pkg.rs
```diff
@@ -1802,21 +1805,23 @@ pub fn git_commit_path(name: &str, repo: &Url, commit_hash: &str) -> PathBuf {
...
-pub fn fetch_git(fetch_id: u64, name: &str, pinned: &SourceGitPinned) -> Result<PathBuf> {
+pub fn fetch_git(fetch_id: u64, name: &str, pinned: &SourceGitPinned) -> Result<Filesystem> {
     let path = git_commit_path(name, &pinned.source.repo, &pinned.commit_hash);
     // Checkout the pinned hash to the path.
     with_tmp_git_repo(fetch_id, name, &pinned.source, |repo| {
         // Change HEAD to point to the pinned commit.
         let id = git2::Oid::from_str(&pinned.commit_hash)?;
         repo.set_head_detached(id)?;
+
+        let _lock = path.open_rw(".forc-lock")?;
```
mitchmindtree, line 1816: This should be a `const`.

mitchmindtree (commented):
Thanks for tackling this @bingcicle!

I'm not sure the `Filesystem` abstraction is actually providing us much extra safety here, as its trivial for future contributions to ignore the lock and read/write files as they see fit, and Rust doesn't give us any other way to guard against I/O :thinking:

That said, I'm also not sure we really need that level of safety at the moment, as our point of contention appears to be in a single place - git checkouts.

What are your thoughts on removing `forc-fs` (and the `Filesystem` abstraction), and simply using `fslock` or `fd-lock` directly within `fetch_git`? We should still link your commit with `forc-fs` to #919 as it might make for a good foundation if we need a more sophisticated `Filesystem` abstraction in the future.

forc-fs/Cargo.toml
bingcicle: Ahh I saw fslock but not fd-lock. `file-lock` seemed at the time to me was more mature than `fslock`, but it seems like `fd-lock` is a better alternative since its from a core maintainer. Will re-write using that.

bingcicle:
> I'm not sure the `Filesystem` abstraction is actually providing us much extra safety here, as its trivial for future contributions to ignore the lock and read/write files as they see fit, and Rust doesn't give us any other way to guard against I/O thinking

Yeah, while looking into locks it seems like advisory locks are only as strong as the maintainers choose to abide by them. We would have to enforce using `forc-fs` and extending that over using the standard `fs`. From looking through the code that `forc-fs` could replace there weren't much places that would benefit from abstraction as well (for now).

> What are your thoughts on removing `forc-fs` (and the `Filesystem` abstraction), and simply using `fslock` or `fd-lock` directly within `fetch_git`? We should still link your commit with `forc-fs` to #919 as it might make for a good foundation if we need a more sophisticated `Filesystem` abstraction in the future.

This sounds good, I'll edit the original PR content to close only #2603 then and link it to #919 instead.

forc-pkg/src/pkg.rs
```diff
@@ -1809,10 +1818,16 @@ pub fn fetch_git(fetch_id: u64, name: &str, pinned: &SourceGitPinned) -> Result<
...
-            let _ = std::fs::remove_dir_all(&path);
-        }
-        std::fs::create_dir_all(&path)?;
+
+        let _ = std::fs::create_dir_all(&path)?;
+        let mut lock = RwLock::new(
+            fs::OpenOptions::new()
+                .write(true)
+                .create(true)
+                .open(path.join(FORC_FILE_LOCK_NAME))?,
```
mitchmindtree, line 1827: Could we move this expression into a re-usable function? E.g.

```rust
/// Given the path to a directory, lock the directory using a file lock.
///
/// The directory is "unlocked" when the returned `RwLock` is dropped.
///
/// This was introduced to avoid racing on the creation of git checkout directories.
fn lock_dir(dir_path: &Path) -> io::Result<RwLock<File>> {
    let file = fs::OpenOptions::new()
        .write(true)
        .create(true)
        .open(path.join(FORC_FILE_LOCK_NAME))?;
    Ok(RwLock::new(file))
}

File: forc-pkg/src/pkg.rs [hunk 4]
```diff
@@ -1609,9 +1610,17 @@ where
...
+    let _ = std::fs::create_dir_all(&repo_dir);
+    let mut lock = RwLock::new(
+        fs::OpenOptions::new()
+            .write(true)
+            .create(true)
+            .open(repo_dir.join(FORC_FILE_LOCK_NAME))?,
+    );
+    let _ = lock.write()?;
+
+    let _ = std::fs::remove_dir_all(&repo_dir);
```
mitchmindtree, line 1623: Wait, if we remove the directory here, don't we also remove our lock file?

forc-pkg/src/pkg.rs
```diff
@@ -1809,10 +1818,16 @@ pub fn fetch_git(fetch_id: u64, name: &str, pinned: &SourceGitPinned) -> Result<
...
+
+        let _ = std::fs::create_dir_all(&path)?;
+        let mut lock = RwLock::new(
+            fs::OpenOptions::new()
+                .write(true)
+                .create(true)
+                .open(path.join(FORC_FILE_LOCK_NAME))?,
+        );
+        let _ = lock.write()?;
+        let _ = std::fs::remove_dir_all(&path);
```
mitchmindtree, line 1830: Same here?

mitchmindtree (changes requested):

forc-pkg/src/pkg.rs
bingcicle: I should probably convert this to a draft - I've realized now that in my original PR it was incorrectly implemented as well. Also, `file-lock` is based on `fnctl()` which is different from `fd-lock` which is based on `flock()`. For whatever reason initializing the lock at the same line did not work for `fd-lock` - still looking into this and trying to understand how flock() compares against fnctl().

bingcicle:
Closing this in favor of @mitchmindtree 's #3592.

PR #3494, opened 2022-12-02
Fixes error when dependency variable is used as function parameter.

### LINKED ISSUE #3406: Call paths cannot be resolved as function parameters
Call paths are resolving when they are assigned to a variable but not as a function parameter.

## Working example:

lib:
```sway
library;
pub const CONTRACT_ID = 0x79fa8779bed2f36c3581d01c79df8da45eee09fac1fd76a5a656e16326317ef0;
```

contract that depends on that lib:
```sway
fn test_function() -> bool {
  let id = test_lib::CONTRACT_ID;
  let contract_id = ContractId::from(id);
}
```

## Failing example:

lib:
```sway
library;
pub const CONTRACT_ID = 0x79fa8779bed2f36c3581d01c79df8da45eee09fac1fd76a5a656e16326317ef0;
```

contract that depends on that lib:
```sway
fn test_function() -> bool {
  let contract_id = ContractId::from(test_lib::CONTRACT_ID);
}
```

esdrubal:
When a dependency variable was passed into a function as first parameter such as `ContractId::from(contract_c::CONTRACT_ID);` an error was thrown by `type_check_method_application` while it tried to resolve the variable as a local one.

The solution was to only resolve the variable if the function is expecting a mutable parameter.

Closes #3406

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/expression/typed_expression/method_application.rs
+++ b/sway-core/src/semantic_analysis/ast_node/expression/typed_expression/method_application.rs
@@ -205,33 +205,35 @@ pub(crate) fn type_check_method_application(
         Some(ty::TyFunctionParameter { is_mutable, .. }),
     ) = (args_buf.get(0), method.parameters.get(0))
     {
-        let unknown_decl = check!(
-            ctx.namespace.resolve_symbol(name).cloned(),
-            return err(warnings, errors),
-            warnings,
-            errors
-        );
+        if *is_mutable {
+            let unknown_decl = check!(
+                ctx.namespace.resolve_symbol(name).cloned(),
+                return err(warnings, errors),
+                warnings,
+                errors
+            );
 
-        let is_decl_mutable = match unknown_decl {
-            ty::TyDeclaration::ConstantDeclaration(_) => false,
-            _ => {
-                let variable_decl = check!(
-                    unknown_decl.expect_variable().cloned(),
-                    return err(warnings, errors),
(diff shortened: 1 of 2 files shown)


mohammadfawaz:
> The solution was to only resolve the variable if the function is expecting a mutable parameter.

What is the relationship between mutability and this particular bug?

esdrubal:
> What is the relationship between mutability and this particular bug?

The bug is caused by trying to resolve by name a variable that is declared in another module. But this can be avoid altogether by changing the order we check mutability.

`type_check_method_application` performs a check to see if the argument is a variable and if both the argument variable and corresponding parameter are mutable. This was fixed by only checking if the variable is mutable after checking the parameter is mutable. I think we cannot use mutable variables from dependencies only const so when the function parameter is mutable we should never get this error because we have to use a local variable.

mohammadfawaz:
> > What is the relationship between mutability and this particular bug?
> 
> The bug is caused by trying to resolve by name a variable that is declared in another module. But this can be avoided altogether by changing the order we check mutability.
> 
> `type_check_method_application` performs a check to see if the argument is a variable and if both the argument variable and corresponding parameter are mutable. This was fixed by only checking if the variable is mutable after checking the parameter is mutable. I think we cannot use mutable variables from dependencies only const so when the function parameter is mutable we should never get this error because we have to use a local variable.

Ah gotcha. That makes sense. Thanks!

mohammadfawaz (approved):

PR #3495, opened 2022-12-02
Use borrowed types instead of owned types so that `TyProgram` is not consumed after IR generation

### LINKED ISSUE #3492: Consider using borrows instead of owned types to not to consume `TyProgram` after IR generation
Since we are generally using owned types while passing `Ty` structs, they are consumed while generating the IR. After our quick investigation with @mitchmindtree earlier today, it seems they can be converted into borrowed types without a major problem. This is required so that tooling can work on the `TyProgram` even after the IR is generated. It is needed for #3260 and probably lsp would need this in the near future as well.

kayagokalp:
closes #3492.

### DIFF
--- a/sway-core/src/ir_generation/function.rs
+++ b/sway-core/src/ir_generation/function.rs
@@ -113,7 +113,7 @@ impl<'te> FnCompiler<'te> {
                 Some(ast_node) => ast_node,
                 None => break Ok(Constant::get_unit(context)),
             };
-            match self.compile_ast_node(context, md_mgr, &ast_node) {
+            match self.compile_ast_node(context, md_mgr, ast_node) {
                 Ok(Some(val)) => break Ok(val),
                 Ok(None) => (),
                 Err(err) => break Err(err),
@@ -327,7 +327,14 @@ impl<'te> FnCompiler<'te> {
                 elem_to_access_num: idx,
                 elem_to_access_span: span,
                 resolved_type_of_parent: tuple_type,
-            } => self.compile_tuple_elem_expr(context, md_mgr, prefix, *tuple_type, *idx, span.clone()),
+            } => self.compile_tuple_elem_expr(
+                context,
+                md_mgr,
+                prefix,
+                *tuple_type,
+                *idx,
+                span.clone(),
+            ),
             ty::TyExpressionVariant::AbiCast { span, .. } => {
                 let span_md_idx = md_mgr.span_to_md(context, span);
                 Ok(Constant::get_unit(context).add_metadatum(context, span_md_idx))
@@ -351,7 +358,9 @@ impl<'te> FnCompiler<'te> {
             ty::TyExpressionVariant::UnsafeDowncast { exp, variant } => {
                 self.compile_unsafe_downcast(context, md_mgr, exp.to_owned(), variant)
             }
-            ty::TyExpressionVariant::EnumTag { exp } => self.compile_enum_tag(context, md_mgr, exp.to_owned()),
+            ty::TyExpressionVariant::EnumTag { exp } => {
+                self.compile_enum_tag(context, md_mgr, exp.to_owned())
+            }
             ty::TyExpressionVariant::WhileLoop { body, condition } => self.compile_while_loop(
                 context,
                 md_mgr,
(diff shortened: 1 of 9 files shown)


sway-core/src/ir_generation/function.rs
```diff
@@ -128,40 +128,40 @@ impl<'te> FnCompiler<'te> {
...
-        match ast_node.content {
+        match &ast_node.content {
             ty::TyAstNodeContent::Declaration(td) => match td {
                 ty::TyDeclaration::VariableDeclaration(tvd) => {
-                    self.compile_var_decl(context, md_mgr, *tvd, span_md_idx)
+                    self.compile_var_decl(context, md_mgr, tvd, span_md_idx)
                 }
                 ty::TyDeclaration::ConstantDeclaration(decl_id) => {
-                    let tcd = declaration_engine::de_get_constant(decl_id, &ast_node.span)?;
+                    let tcd = declaration_engine::de_get_constant(decl_id.clone(), &ast_node.span)?;
```
kayagokalp, line 140: I left declaration engine API intact to not cause a massive breaking change, but because of that I had to clone declaration ids. I think it should be fine since we were `clone` ing entire nodes before this PR but if it is preferred to have declaration engine API accepting borrowed types as well I can continue there too.

sway-core/src/ir_generation/function.rs
```diff
@@ -128,40 +128,40 @@ impl<'te> FnCompiler<'te> {
...
                     self.compile_const_decl(context, md_mgr, tcd, span_md_idx)?;
                     Ok(None)
                 }
                 ty::TyDeclaration::FunctionDeclaration(_) => {
                     Err(CompileError::UnexpectedDeclaration {
                         decl_type: "function",
-                        span: ast_node.span,
+                        span: ast_node.span.clone(),
```
kayagokalp, line 147: Also while returning error's I am `clone`ing spans. I feel like this should be ok because of the same reason with previous comment but open to suggestions again 🙌

kayagokalp (commented):
I can continue to work with `TyProgram` in `forc-pkg` after compiling the package to asm with this PR. I think we can continue this type changes (accepting borrowed types instead of owned ones) in public APIs that compiler exposes to public.  

While doing so I think I needed to touch a little deeper than I thought at the first place but it is possible that I missed something and went further than required 😅

sway-core/src/lib.rs
```diff
@@ -319,22 +319,22 @@ pub fn compile_to_asm(
...
+    ast_to_asm(type_engine, &ast_res, &build_config)
 }
 
 /// Given an AST compilation result,
 /// try compiling to a `CompiledAsm`,
 /// containing the asm in opcode form (not raw bytes/bytecode).
 pub fn ast_to_asm(
     type_engine: &TypeEngine,
-    ast_res: CompileResult<ty::TyProgram>,
+    ast_res: &CompileResult<ty::TyProgram>,
```
otrho, line 330: This is a bit weird, even prior to your change.  I wonder if we need to take a look at the flow through `sway-core/src/lib.rs` and reassess the API.  I think it's grown fairly organically over the months and needs some chopping.

sway-core/src/ir_generation/function.rs
otrho: It's a `usize` and a `Span`, not ideal but not the end of the world.

sway-core/src/ir_generation/function.rs
otrho: Cloning for errors is fine IMO. :+1:

otrho (approved):
Man, this must've been quite laborious!  LGTM.

sway-core/src/lib.rs
mitchmindtree: Yeah, I think a *slightly* more idiomatic way of writing this would be:
```suggestion
    ast_res: CompileResult<&ty::TyProgram>,
```
We might want to add an `as_ref` method to `CompileResult` if we don't already have one (similar to [`std::result::Result::as_ref`](https://doc.rust-lang.org/std/result/enum.Result.html#method.as_ref)).

All that said, this function probably shouldn't take a `CompileResult` as an argument anyway, and would likely be better off taking the `&ty::TyProgram` directly. But we can address that in the future :)

mitchmindtree (approved):
Great work in peeling this apart @kayagokalp!

PR #3499, opened 2022-12-02
Introduce additional spans in Array and String types

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/2486
- [x] Spans for array element types
- [x] Spans for array size
- [x] ~Spans for types in nested tuples~ Already available.. Just have to dig into each type ID recursively.
- [x] Spans for string size

The gist of the change is the following:
- Add a new type `struct Length { val: usize, span: Span }` to represent the length of an array or string. This is useful to preserve the `Span` of the raw integer length.
- Change `TypeInfo::Array(TypeId, usize, TypeId)` into `TypeInfo::Array(TypeArgument, Length)`. This helps preserve the span of the element type and the raw integer length.
- Change `TypeInfo::Str(u64)` to `TypeInfo::Str(Length)`. Again, this is useful to preserve the `Span` of the raw integer length.
- Everything else in this PR is just a fallout of the above.

### DIFF
--- a/sway-core/src/type_system/type_info.rs
+++ b/sway-core/src/type_system/type_info.rs
@@ -115,9 +115,7 @@ pub enum TypeInfo {
     Contract,
     // used for recovering from errors in the ast
     ErrorRecovery,
-    // Static, constant size arrays. The second `TypeId` below contains the initial type ID
-    // which could be generic.
-    // TODO: change this to a struct instead of a tuple
+    // Static, constant size arrays.
     Array(TypeArgument, Length),
     /// Represents the entire storage declaration struct
     /// Stored without initializers here, as typed struct fields,
--- a/sway-core/src/type_system/type_info.rs
+++ b/sway-core/src/type_system/type_info.rs
@@ -29,6 +29,12 @@ impl fmt::Display for AbiName {
     }
 }
 
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+pub struct Length {
+    pub len: usize,
+    pub span: Span,
+}
+
 /// A slow set primitive using `==` to check for containment.
 #[derive(Clone)]
 pub struct VecSet<T>(pub Vec<T>);
@@ -112,7 +118,7 @@ pub enum TypeInfo {
     // Static, constant size arrays. The second `TypeId` below contains the initial type ID
     // which could be generic.
     // TODO: change this to a struct instead of a tuple
-    Array(TypeArgument, usize),
+    Array(TypeArgument, Length),
     /// Represents the entire storage declaration struct
     /// Stored without initializers here, as typed struct fields,
     /// so type checking is able to treat it as a struct with fields.
@@ -405,7 +411,7 @@ impl DisplayWithTypeEngine for TypeInfo {
                         .unwrap_or_else(|| "None".into())
                 )
             }
-            Array(elem_ty, count) => format!("[{}; {}]", type_engine.help_out(elem_ty), count),
+            Array(elem_ty, count) => format!("[{}; {}]", type_engine.help_out(elem_ty), count.len),
             Storage { .. } => "contract storage".into(),
             RawUntypedPtr => "raw untyped ptr".into(),
             RawUntypedSlice => "raw untyped slice".into(),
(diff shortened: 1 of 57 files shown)


sway-lsp/src/server.rs
```diff
@@ -956,7 +956,7 @@ mod tests {
         let uri = init_and_open(&mut service, doc_comments_dir()).await;
         let _ = go_to_definition_request(&mut service, &uri, 44, 19, 1).await;
         let _ = did_change_request(&mut service, &uri).await;
-        let _ = go_to_definition_request(&mut service, &uri, 45, 19, 2).await;
+        let _ = go_to_definition_request(&mut service, &uri, 45, 20, 2).await;
```
mohammadfawaz, line 959: @JoshuaBatty I've had to make this change here but I don't know why 🤔

eureka-cpu: is it possible it was passing incorrectly before?

eureka-cpu (commented):
Looking much more readable now imo

sway-core/src/type_system/type_info.rs
```diff
@@ -29,6 +29,12 @@ impl fmt::Display for AbiName {
     }
 }
 
+#[derive(Debug, Clone, Hash)]
+pub struct Length {
+    pub val: usize,
```
sezna, line 34: I have a slight preference for using `.val()` and `.span()` accessor methods, since I've done so many refactors where I wish there was a defined API 😓 but I think this is probably low-risk.  Worth considering, though.

mohammadfawaz: Reasonable suggestion. I will do that.

sway-core/src/type_system/type_mapping.rs
```diff
@@ -202,10 +202,17 @@ impl TypeMapping {
...
+                TypeMapping::from_type_parameters_and_type_arguments(
+                    vec![type_parameter]
+                        .iter()
+                        .map(|x| x.type_id)
+                        .collect::<Vec<_>>(),
+                    vec![type_argument]
+                        .iter()
+                        .map(|x| x.type_id)
+                        .collect::<Vec<_>>(),
+                )
```
otrho, line 215:
```suggestion
                TypeMapping::from_type_parameters_and_type_arguments(
                    vec![type_parameter.type_id],
                    vec![type_argument.type_id],
                )
```

sway-core/src/type_system/type_mapping.rs
```diff
@@ -350,10 +357,13 @@ impl TypeMapping {
...
-                self.find_match(ary_ty_id, type_engine).map(|matching_id| {
-                    type_engine.insert_type(TypeInfo::Array(matching_id, count, initial_elem_ty))
-                })
+            TypeInfo::Array(mut elem_ty, count) => {
+                if let Some(type_id) = self.find_match(elem_ty.type_id, type_engine) {
+                    elem_ty.type_id = type_id;
+                    Some(type_engine.insert_type(TypeInfo::Array(elem_ty, count)))
+                } else {
+                    None
+                }
```
otrho, line 366:
```suggestion
            TypeInfo::Array(mut elem_ty, count) => self
                .find_match(elem_ty.type_id, type_engine)
                .map(|type_id| {
                    elem_ty.type_id = type_id;
                    type_engine.insert_type(TypeInfo::Array(elem_ty, count))
                }),
```
This could've stayed as a `map`.

sway-core/src/type_system/type_mapping.rs
mohammadfawaz: 😆

sway-core/src/type_system/type_info.rs
```diff
@@ -528,7 +546,7 @@ impl TypeInfo {
         match self {
             Unknown => "unknown".into(),
             UnknownGeneric { name, .. } => name.to_string(),
-            Str(x) => format!("str[{}]", x),
+            Str(x) => format!("str[{}]", x.val),
```
sezna, line 549: Now that we have `.val()`, I think we should make the fields on `Length` private and use the method everywhere. Same for `span()`.

mohammadfawaz: Oh I've done that already, but `Length` lives in `type_info.rs` so `x.val` still works here, so I missed this one. I guess the right thing to do would be to move `Length` into a separate module.

sezna (approved):

otrho (approved):

sway-lsp/src/server.rs
JoshuaBatty: I'll need to dig into this a bit more but it shouldn't stand in the way of getting the PR merged.

JoshuaBatty (approved):
Thanks for this @mohammadfawaz really excited to use these changes in the server.

Issue #3425, opened 2022-11-25
Incorrect dead code warning for constant used as the condition of a `while` loop


mohammadfawaz:
Reproducer:
```rust
script;

const COND = false;

fn main() {
    while COND { }
}
```
Warning:
```console
3 | const COND = false;
  | ------------------- This declaration is never used.
4 |
5 | fn main() {
```

PR #3504, opened 2022-12-03
Fix DCA warnings for expressions reachable from while loop conditions.

tritao:
This connects the edges from expressions in while loop conditions to the entry node of the while loop.

Closes https://github.com/FuelLabs/sway/issues/3425.


mohammadfawaz (approved):
great, thanks!

otrho (approved):

PR #3506, opened 2022-12-05
ci: pin linux release steps to ubuntu-20.04

bingcicle:
Recently Github Actions began [migrating `ubuntu-latest` from `ubuntu-20.04` to `ubuntu-22.04`](https://github.com/actions/runner-images/issues/6399) on a rolling basis. To prevent issues with libc and to keep our binaries as backward compatible as possible, let's pin the release steps to use `ubuntu-20.04` instead.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -566,12 +566,12 @@ jobs:
     strategy:
       matrix:
         job:
-          - os: ubuntu-latest
+          - os: ubuntu-20.04
             platform: linux
             target: x86_64-unknown-linux-gnu
             arch: amd64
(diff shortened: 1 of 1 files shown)


mitchmindtree (approved):
Nice, pinning to a specific version is probably for the best either way.

There appears to be a large mix of folk pinning to `20.04` and others pinning to `22.04` in https://github.com/actions/runner-images/issues/6399. It looks like newer versions of glibc at least attempt to maintain backwards compatibility, so perhaps pinning to 20.04 might indeed ensure slightly broader compatibility than pinning to 22.04.

Issue #3266, opened 2022-11-03
Include test function span information (file, module path, line number) in `forc test` output


mitchmindtree:
As of #2985, we only produce the function name for each test in our output, e.g.

![Screenshot from 2022-11-03 19-00-57](https://user-images.githubusercontent.com/4587373/199695004-14610ab6-8c97-47f0-a330-26f061a834c4.png)

We should be showing the full file path, module path and line number like cargo does.

This was omitted from #2985, only because it wasn't immediately obvious to me where to retrieve the span from during the ASM generation stage, and the PR is big enough :sweat_smile:

kayagokalp:
I think cargo only prints line number, file path and module path when a test is ignored (maybe also for failing ones too), I think that makes sense if a test is passing it results less noise in the terminal and only if something that may require your attention those details are printed. I am leaning towards doing the same thing.

File path and line number should not be a problem but it is not really clear how can we get the module path. Do we even have modules? If a test named `foo` is declared under `baz.sw`, is there any other possilibity than `baz::foo` in sway right now?

PR #3509, opened 2022-12-05
Introduce a way for entries generated by tests to be associated with their related function declaration

### LINKED ISSUE #3508: Introduce a way for entries generated by tests to be associated with their related function declaration
For introducing lots of interesting stuff to `forc-test`, we need to get the function declaration for the tests we are running but currently the way they are working is only exposing the `FinalizedEntryPoint` to `forc-test` which does not provide any information that we can use to retrieve the function declaration itself.

This will unblock:
- #3260 
- #3266

kayagokalp:
closes #3508.

To continue with improving in-language testing (like the linked issues), we need to have a way of getting the original function declaration for the test that we are running from `forc-test`. Currently `forc-test` only works on `FinalizedEntry`s generated by the tests and we cannot get the function declaration from the entry point. Here I am introducing an optional field to the `FinalizedEntry` for tests. If the `FinalizedEntry` is generated from a test function, corresponding `FinalizedEntry`'s declaration id is populated and can be retrieved. To pass this declaration id information I needed to preserve it in the IR generation step to do that I added a metadata for it and populate it for the test function declarations. So for test function declarations we are inserting their declaration index to the metadata. I did not insert the `DeclarationId` since it does not implement `Hash`, and we are already inserting the span for the function declaration. Since `DeclarationId` is basically the index + span, storing only the index was enough. After the IR generation before creating an entry point I construct the `DeclarationId` from index and span (which comes from the metadata) and pass that to `FinalizedEntry`. This way in the IR generation step we are not losing the declaration index + span (and thus, the declaration id) and can retrieve the test function's declaration later on.

unblocks #3490.
unblocks #3266.


kayagokalp (commented):
I ended up collecting declaration ids only for test functions, so I added an optional test decl id field to `FinalizedEntry`. With this PR, I can retrieve `DeclarationId` for a specific test function from `forc-test` which makes it possible for later PRs to get the `TyFunctionDeclaration` from the declaration engine.

sway-core/src/metadata.rs
```diff
@@ -61,6 +63,33 @@ impl MetadataManager {
...
+                    .unwrap_struct("decl_index", 2)
+                    .and_then(|fields| {
+                        // First check if the read metadata corresponds to Some(_) or None
+                        let is_index_some = fields[0].unwrap_integer().map(|index| index != 0)?;
+                        let index = if is_index_some {
+                            Some(fields[1].unwrap_integer().map(|index| index as usize)?)
+                        } else {
+                            None
+                        };
+                        self.md_decl_index_cache.insert(md_idx, index);
```
kayagokalp, line 85: I ended up doing some juggling to read `Option<>` value from the metadata like this, curious about whether there is a more idiomatic way to replace this stuff.

sway-core/src/metadata.rs
```diff
@@ -198,6 +227,35 @@ impl MetadataManager {
...
+            .copied()
+            .or_else(|| {
+                let md_idx = MetadataIndex::new_struct(
+                    context,
+                    "decl_index",
+                    vec![
+                        Metadatum::Integer(is_decl_index_some),
+                        Metadatum::Integer(decl_index_val as u64),
+                    ],
+                );
```
kayagokalp, line 252: Related to the above comment this is how I store an `Option<u64>` in metadata ^

kayagokalp:
Looks like the metadata addition needs to be somehow reflected in the ir generation test checks. I am trying to understand the structure there right now.

sway-core/src/metadata.rs
kayagokalp: solved by 24ba46465c61d05914280e00881d310e211ce7f8

mohammadfawaz:
Two things:
* Do you mind adding a detailed description of the change? You seem to be adding a new metadata here. It would be nice to have that metadata well described.
* It'd be nice to see some IR tests for this such as ones in https://github.com/FuelLabs/sway/tree/master/sway-ir/tests/serialize

kayagokalp:
cc @mohammadfawaz Of course to continue with improving in-language testing (like the linked issues), we need to have a way of getting the original function declaration for the test that we are running from `forc-test`. Currently `forc-test` only works on `FinalizedEntry`s generated by the tests and we cannot get the function declaration from the entry point. Here I am introducing an optional field to the `FinalizedEntry` for tests. If the `FinalizedEntry` is generated from a test function, its declaration id is populated and can be retrieved. To pass this declaration id information I needed to preserve it in the IR generation step to do that I added a metadata for it and populate it for the test function declarations. This way in the IR generation step we are not losing the declaration id and can retrieve the test function's declaration later on.

I am looking for how to (and where to) add a test for this 👍

mohammadfawaz:
> 

Excellent, thank you! Just wanted to see this description in the Description field for this PR for future reference.

mohammadfawaz:
> cc @mohammadfawaz Of course, to continue with improving in-language testing (like the linked issues), we need to have a way of getting the original function declaration for the test that we are running from `forc-test`. Currently `forc-test` only works on `FinalizedEntry`s generated by the tests and we cannot get the function declaration from the entry point. Here I am introducing an optional field to the `FinalizedEntry` for tests. If the `FinalizedEntry` is generated from a test function, corresponding `FinalizedEntry`'s declaration id is populated and can be retrieved. To pass this declaration id information I needed to preserve it in the IR generation step to do that I added a metadata for it and populate it for the test function declarations. So for test function declarations we are inserting their declaration index to the metadata. I did not insert the `DeclarationId` since it does not implement `Hash`, and we are already inserting the span for the function declaration. Since `DeclarationId` is basically the index + span, storing only the index is enough. After the IR generation before creating an entry point I construct the `DeclarationId` from index and span (which comes from the metadata) and pass that to `FinalizedEntry`. This way in the IR generation step we are not losing the declaration index + span (and thus, the declaration id) and can retrieve the test function's declaration later on.
> 
> I am looking for how to (and where to) add a test for this 👍

Excellent, thank you! Just wanted to see this description in the Description field for this PR for future reference.

test/src/ir_generation/mod.rs
```diff
@@ -141,7 +141,7 @@ pub(super) async fn run(filter_regex: Option<&regex::Regex>) -> Result<()> {
                 let tree_type = typed_program.kind.tree_type();
 
                 // Compile to IR.
-                let include_tests = false;
+                let include_tests = true;
```
kayagokalp, line 144: To have a test to check metadata generated for the tests I need it to be compiled with `include_tests` enabled. I can change this to to be true if the file name starts with `test_` or something similar to not the change default behavior for all IR tests if it is preferred over this. cc @otrho

otrho: I don't see why we shouldn't try to have `include_tests` as true.  @mitchmindtree was probably being conservative, unless it overly complicates things somehow?  Most of the existing tests won't have any unit tests so it wouldn't make any difference.

test/src/ir_generation/tests/test_metadata.sw
```diff
@@ -0,0 +1,15 @@
...
+
+#[test()]
+fn my_test_func() {
+  my_func();
+}
+
+// check: fn main() -> (), $(main_md=$MD) {
+// check: fn my_test_func() -> (), $(test_md=$MD) {
+
+// check: $test_md = ($MD $MD)
```
otrho, line 14: This could probably be a little more explicit, perhaps matching the `decl_id` tag is definitely referenced by the `test_md` index.

otrho (commented):
This is looking really good overall. :+1:

kayagokalp:
Thanks for the comments @otrho! I added a `serialize` test and made the `ir_generation` test more explicit (added checks related to the `decl_id`). I think this should be good go

otrho (approved):

mohammadfawaz (approved):
Great, thanks! Thanks for adding the test.. it just makes it clear what you're adding in this change and also prevents regression in case someone makes a change that drops the metadata.

It seems that this metadata is attached to the top level test function so there is no risk of it being lost since those functions are never changed (never inlined for example).

@otrho it would be nice to have a place where all the metadata are explained along with their legal use cases, such as https://llvm.org/docs/LangRef.html#dicompileunit

mitchmindtree (approved):
:ok_hand: :shipit:

PR #3510, opened 2022-12-05
`U256` Subtraction fix

rostyslavtyshko:
I found a bug in case when one of the words of the rhs `U256` values if bigger than the corresponding words in the lhs `U256` values, but the next to the left value of the lhs `U256` word is zero.


sway-lib-std/src/u256.sw
```diff
@@ -4,6 +4,7 @@ use ::assert::assert;
 use ::convert::From;
 use ::result::Result;
 use ::u128::U128;
+use ::logging::log;
```
mohammadfawaz, line 7:
```suggestion
```

sway-lib-std/src/u256.sw
```diff
@@ -356,34 +357,51 @@ impl core::ops::Add for U256 {
...
         }
-
         let mut result_c = 0;
         if word_3 < other_word_3 {
             result_c = u64::max() - (other_word_3 - word_3 - 1);
-            result_b -= 1;
+            if result_b > 0 {
+                result_b -= 1;
+            } else {
+                result_a -= 1;
```
mohammadfawaz, line 384: Do we know that `result_a > 0` here?

sway-lib-std/src/u256.sw
```diff
@@ -356,34 +357,51 @@ impl core::ops::Add for U256 {
...
         if word_4 < other_word_4 {
             result_d = u64::max() - (other_word_4 - word_4 - 1);
-            result_c -= 1;
+            if result_c > 0 {
+                result_c -= 1;
+            } else {
+                if result_b > 0 {
+                    result_b -= 1;
+                } else {
+                    result_a -= 1;
```
mohammadfawaz, line 400: Same question. Do we know that `result_a > 0` here?

rostyslavtyshko: In case `result_a <= 0` means that lhs of the operation is smaller than rhs, which we ruled out at the beginning of the function.

sway-lib-std/src/u256.sw
rostyslavtyshko: Fixed

sway-lib-std/src/u256.sw
mohammadfawaz: Great! The original code is somewhat hard to understand, so if you want to be a good samaritan and comment out the thought process (and your comment above), that'll be much appreciated.

mohammadfawaz (approved):
Great! Thanks.

otrho (approved):

PR #3523, opened 2022-12-06
Correctly compute predicate data pointer in codegen

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/3563

## Description
Currently, the predicate data pointer is computed in the same way as the script data pointer which is incorrect. Predicates and scripts have completely different addresses at which their data starts. Predicates are also a bit more complicated because the base pointer depends on whether the input type is a `Coin` or `Message`. Therefore, we need to introduce a runtime check that decides what pointer to use based on the input type. The `gtf` opcode gives us the input type as well as the required pointers. We also need to first get the index of the predicate via the `gm` opcode and use that in the subsequent `gtf` instructions.

### Single arg case, copy type:
```rust
fn main(x: u64) -> bool {
    true
}
```
```asm
.program:
.0                                      ; --- start of function: main ---
gm   $r2 i3                             ; get predicate index
gtf  $r1 $r2 i257                       ; get input type
jnzi $r1 .3
gtf  $r3 $r2 i269                       ; get input coin predicate data pointer
ji  .2
.3
movi $r0 i2                             ; register containing 2
eq   $r0 $r1 $r0                        ; input type is message(2)
xori $r0 $r0 i1                         ; input type is not message(2)
jnzi $r0 .4
gtf  $r3 $r2 i288                       ; input message predicate data pointer
ji  .2
.4
ret  $zero                              ; return false
.2
lw   $r3 $r3 i0                         ; load main fn parameter
ret  $one
```

### Single arg case, reference type:
```rust
fn main(x: b256) -> bool {
    x == 0x0000000000000000000000000000000000000000000000000000000000000000
}
```
```asm
.program:
.0                                      ; --- start of function: main ---
gm   $r2 i3                             ; get predicate index
gtf  $r1 $r2 i257                       ; get input type
jnzi $r1 .3
gtf  $r3 $r2 i269                       ; get input coin predicate data pointer
ji  .2
.3
movi $r0 i2                             ; register containing 2
eq   $r0 $r1 $r0                        ; input type is message(2)
xori $r0 $r0 i1                         ; input type is not message(2)
jnzi $r0 .4
gtf  $r3 $r2 i288                       ; input message predicate data pointer
ji  .2
.4
ret  $zero                              ; return false
.2
lw   $r1 data_0                         ; literal instantiation
addi $r0 $zero i32                      ; asm block
meq  $r0 $r3 $r1 $r0                    ; asm block
ret  $r0
```

### Multiple arg case
```rust
fn main(x: u64, y: u64) -> bool {
    x == 0 && y == 0 
}
```
```asm
.program:
.0                                      ; --- start of function: main ---
gm   $r2 i3                             ; get predicate index
gtf  $r1 $r2 i257                       ; get input type
jnzi $r1 .3
gtf  $r3 $r2 i269                       ; get input coin predicate data pointer
ji  .2
.3
movi $r0 i2                             ; register containing 2
eq   $r0 $r1 $r0                        ; input type is message(2)
xori $r0 $r0 i1                         ; input type is not message(2)
jnzi $r0 .4
gtf  $r3 $r2 i288                       ; input message predicate data pointer
ji  .2
.4
ret  $zero                              ; return false
.2
lw   $r0 $r3 i0                         ; get arg x
lw   $r1 $r3 i1                         ; get arg y
eq   $r0 $r0 $zero
jnzi $r0 .5
ji  .6
.5
eq   $r0 $r1 $zero
.6
ret  $r0
```

## Testing
I added two unit tests here but full E2E tests are difficult to add without SDK support. That being said, @hal3e has been experimenting with this branch in his open PR https://github.com/FuelLabs/fuels-rs/pull/726. Once we have that full support, we can then add SDK tests in the sway repo to help test this change.


test/src/e2e_vm_tests/test_programs/should_pass/language/main_args/main_args_predicate/test.toml
```diff
@@ -1,4 +1,4 @@
-category = "run"
+category = "disabled"
```
mohammadfawaz, line 1: This test is disabled because it uses script data even though it's a predicate. That was passing because our E2E test config makes no distinction between predicates and scripts, and so the predicate was run a script. This now falls apart when predicates have main args.

otrho: So do we need to update the harness to make that distinction and re-enable this test?

otrho (dismissed):

mohammadfawaz:
This only looks at input at index 0, which doesn't seem to be enough. Turning into a draft until the generalized case is resolved.

test/src/e2e_vm_tests/test_programs/should_pass/language/main_args/main_args_predicate/test.toml
mohammadfawaz: Yeah exactly. Doing this in the SDK is much easier. In E2E tests, we'd have to manually craft a transaction with an input predicted by the predicate.

mohammadfawaz:
Ok this is now ready for review again.

otrho:
I have this in `sway/.git/hooks/pre-commit`.  Sometimes it's a bit aggressive but you can use `git commit -n` to turn it off.
```
#!/usr/bin/env bash

declare -a rust_files=()
files=$(git diff-index --name-only HEAD)

for file in $files; do
    if [ ! -f "${file}" ]; then
        continue
    fi
    if [[ "${file}" == *.rs ]]; then
        rust_files+=("${file}")
    fi
done

if [ ${#rust_files[@]} -ne 0 ]; then
    if ! rustfmt --check --files-with-diff ${rust_files[@]} ; then
      echo
      echo 'Rustfmt failed!'
      exit 1
    fi
fi
```

otrho (approved):

sway-core/src/asm_generation/asm_builder/functions.rs
```diff
@@ -433,6 +439,149 @@ impl<'ir> AsmBuilder<'ir> {
...
+        });
+
+        // Label to jump to if the input type is *not* zero, i.e. not "coin". Then do the jump.
+        let input_type_not_coin_label = self.reg_seqr.get_label();
+        self.cur_bytecode.push(Op::jump_if_not_zero(
+            input_type.clone(),
+            input_type_not_coin_label,
+        ));
+
+        // If the input is indeed a "message", then use `GTF` to get the "input coin predicate data
```
hal3e, line 486:
```suggestion
        // If the input is indeed a "coin", then use `GTF` to get the "input coin predicate data
```

hal3e (dismissed):

otrho (approved):

hal3e (approved):

PR #3531, opened 2022-12-06
Add a way to specify salt while declaring contract dependencies

### LINKED ISSUE #2970: Add a way to specify salt with `forc-build`

kayagokalp:
closes #2970.

## Summary
This PR introduces a way of declaring `Salt` with contract dependencies. To specify a salt for a contract dependency, users can basically provide the salt next to the source of the dependency. Example:

```toml
[contract-dependencies]
foo = { path = "../foo", salt = "0x11111111111111111111111111111111111111111111111111111111111111111" }
```
Declared salts are taken into account while injecting the contract dependency ids into the dependee's namespace.

## Implementation Details
I am adding the `Salt` information to the edges as conceptually they are only meaningful between two nodes in the graph (it is a property of the dependency relation just like dependency type). I started implementing this with Salt being part of the `Pinned` package, but that creates situations where two different salt declarations for the same package creates two different nodes in the graph and create double compilations so I decided to add them to the edges as it both solves double compilations and made more sense conceptually.

I added an optional field for `Salt` in the edges of the graph which gets populated only if the dependency kind described by that edge is `DepKind::Contract` (a contract dependency). For contract dependencies declared under `[contract-dependencies]` table in `Forc.toml`, we have the ability of specifying the salt. If no salt is provided `Salt::zeroed` is used instead. An example for the created dependency graph can be seen below:
success_case

I am also checking for conflicting salt declarations, meaning that if two different contracts that depends on the same contract declares different salt for that contract, I am producing an error. An example dependency graph for that case can be seen below (see the edges going to contract_c):
error_case

Tests for both cases are added. Graphs seems a little small but can be expanded if clicked on it. (I wonder if there is a good solution for hosting dot output with markdown here 😄)


forc-pkg/src/pkg.rs
```diff
@@ -55,6 +55,7 @@ pub struct Edge {
     /// This may differ from the package name as declared under the dependency package's manifest.
     pub name: String,
     pub kind: DepKind,
+    pub salt: Option<fuel_tx::Salt>,
```
mitchmindtree, line 58: Rather than making this an `Option<fuel_tx::Salt>` on the `Edge` type, let's use the type system to guarantee that contract dependencies *always* have a salt, and library dependencies *never* have a salt.

We can do so by moving the `salt` field to the `Contract` variant of the `DepKind`, e.g.

```rust
pub enum DepKind {
    Library,
    Contract {
        salt: fuel_tx::Salt,
    }
}
```

mitchmindtree (changes requested):
Awesome work Kaya!

Great job on identifying that this information belongs on the edge, and not the node.

> I am also checking for conflicting salt declarations, meaning that if two different contracts that depends on the same contract declares different salt for that contract, I am producing an error.

This certainly sounds like a good strategy for avoiding a lot of human error around adding salt to some locations but forgetting others!

However, I'm not sure that it's correct to always assume this is an *error*. Could there not be cases where the user intentionally depends on multiple unique deployments of the same contract, each distinguished by their unique salts?

### Allowing differing salts

Perhaps we can produce an error by default, and add an extra field to `ContractDependency` along the lines of `allow-different-salt` that defaults to `false`? Then, a contract dependency may only specify a salt that differs to another instance of the same contract within the graph *if* `allow-different-salt = true`. This way perhaps we can still allow the niche use-case where a user might need to depend on multiple different deployments of the same contract?

### Overriding transitive salts?

Another use-case we might want to consider is, what if a user wants to override the salt used in a transitive contract dependency? E.g. it's not hard to imagine a case where a user might want to specify a salt at the top-level and ensure that all instances of the same contract within the dependency graph refer to the same salt. For these cases, perhaps we could add another field to the `ContractDependency` type along the lines of `override-transitive-salts` that defaults to `false`?

Either way, I'm happy for us to simply error in this PR, and post issues for both of these cases so that we can tackle them as follow-up / in the future.

I've left one suggestion w.r.t. moving the salt onto the `Contract` variant of `DepKind`, otherwise I think this is looking great!

test/src/e2e_vm_tests/test_programs/should_fail/contract_dependencies_conflicting_salt/contract_a/Forc.lock
```diff
@@ -0,0 +1,26 @@
...
+source = 'path+from-root-C28B6153CE3E3FEB'
+
+[[package]]
+name = 'core'
+source = 'path+from-root-C28B6153CE3E3FEB'
+
+[[package]]
+name = 'std'
+source = 'path+from-root-C28B6153CE3E3FEB'
+dependencies = ['core']
```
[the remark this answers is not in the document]
kayagokalp: Ahh yeah this was mistakenly copied from passing test.

test/src/e2e_vm_tests/test_programs/should_fail/contract_dependencies_conflicting_salt/contract_b/Forc.lock
```diff
@@ -0,0 +1,8 @@
+[[package]]
+name = 'contract_b'
+source = 'member'
+contract-dependencies = ['contract_c']
```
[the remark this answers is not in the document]
kayagokalp: This is also solved while working on moving the salt into `DepKind`

forc-pkg/src/lock.rs
```diff
@@ -91,7 +97,7 @@ impl PkgLock {
             .collect();
         let mut contract_dependencies: Vec<String> = all_dependencies
             .iter()
-            .filter(|(_, dep_kind)| *dep_kind == DepKind::Contract)
+            .filter(|(_, dep_kind)| matches!(*dep_kind, DepKind::Contract { salt: _ }))
```
mitchmindtree, line 100:
```suggestion
            .filter(|(_, dep_kind)| matches!(*dep_kind, DepKind::Contract { .. }))
```

forc-pkg/src/lock.rs
```diff
@@ -250,30 +270,56 @@ fn pkg_dep_line(
     dep_name: Option<&str>,
     name: &str,
     source: &pkg::SourcePinned,
+    salt: Option<&fuel_tx::Salt>,
```
mitchmindtree, line 273: Perhaps we can take a `&DepKind` directly here rather than an `Option<&Salt>`? This could remove the need to convert to an `Option` before calling `pkg_dep_line`.

forc-pkg/src/pkg.rs
```diff
@@ -62,7 +62,16 @@ pub enum DepKind {
...
     /// The dependency is a contract and declared under `[contract-dependencies]`.
-    Contract,
+    Contract { salt: fuel_tx::Salt },
+}
+
+impl fmt::Display for DepKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            DepKind::Library => write!(f, "library"),
+            DepKind::Contract { salt: _ } => write!(f, "contract"),
```
mitchmindtree, line 72: Should we include the salt in this? I don't quite remember where this `Display` impl for `DepKind` is used :thinking:

forc-pkg/src/pkg.rs
```diff
@@ -2510,14 +2509,37 @@ pub fn build_with_options(build_options: BuildOpts) -> Result<Built> {
...
+        let salt_declarations: HashSet<fuel_tx::Salt> = graph
+            .edges_directed(node, Direction::Incoming)
+            .filter_map(|e| match e.weight().kind {
+                DepKind::Library => None,
+                DepKind::Contract { salt } => Some(salt),
+            })
+            .collect();
+        if salt_declarations.len() > 1 {
+            bail!(
+                "There are conflicting salt declarations for contract dependency named: {}",
```
mitchmindtree, line 2537: Perhaps we can also print the list of salts in this msg too?

forc-pkg/src/pkg.rs
```diff
@@ -2578,7 +2600,7 @@ pub fn build(
         if plan
             .graph()
             .edges_directed(node, Direction::Incoming)
-            .any(|e| e.weight().kind == DepKind::Contract)
+            .any(|e| matches!(e.weight().kind, DepKind::Contract { salt: _ }))
```
mitchmindtree, line 2603:
```suggestion
            .any(|e| matches!(e.weight().kind, DepKind::Contract { .. }))
```

test/src/e2e_vm_tests/test_programs/should_fail/contract_dependencies_conflicting_salt/contract_a/src/main.sw
```diff
@@ -0,0 +1,16 @@
...
+    fn test_function();
+}
+
+impl MyContract for Contract {
+    fn test_function() {
+        let CONTRACT_B = CONTRACT_B_ID;
+        let contract_b_id = ContractId::from(CONTRACT_B);
+        let contract_c_id = ContractId::from(contract_c::CONTRACT_ID);
+    }
+}
```
mitchmindtree, line 16: It looks like this test could be reduced a little?

```rust
contract;

abi MyContract {
    fn test_function();
}

    fn test_function() {
        let contract_b_id = ContractId::from(contract_b::CONTRACT_ID);
        let contract_c_id = ContractId::from(contract_c::CONTRACT_ID);
    }
}
```

I guess it doesn't matter as we never get to the point of building this particular test.

test/src/e2e_vm_tests/test_programs/should_fail/contract_dependencies_conflicting_salt/contract_a/.gitignore
```diff
@@ -0,0 +1,2 @@
+out
+target
```
mitchmindtree, line 2: I think we can remove the `.gitignore`s as the top-level `.gitignore` should cover these.

test/src/e2e_vm_tests/test_programs/should_fail/contract_dependencies_conflicting_salt/contract_b/.gitignore
```diff
@@ -0,0 +1,2 @@
+out
+target
```
mitchmindtree, line 2: Same here

test/src/e2e_vm_tests/test_programs/should_fail/contract_dependencies_conflicting_salt/contract_c/.gitignore
```diff
@@ -0,0 +1,2 @@
+out
+target
```
mitchmindtree, line 2: Same here

forc-pkg/src/lock.rs
```diff
@@ -191,7 +197,14 @@ impl Lock {
                 .as_ref()
                 .into_iter()
                 .flatten()
-                .map(|contract_dep| (contract_dep, DepKind::Contract));
+                .map(|contract_dep| {
+                    (
+                        contract_dep,
+                        DepKind::Contract {
+                            salt: fuel_tx::Salt::zeroed(),
```
mitchmindtree, line 204: I'm a bit weary of using a fake `Salt` here, even if it's only temporary. We should either:

- Comment very clearly explaining that this value is overridden later on in the function or
- Make a dedicated enum type just for distinguishing between the dep kinds here. E.g.
  ```rust
  enum UnparsedDepKind { Library, Contract }
  ```
  then this could be:
  ```rust
  .map(|contract_dep| (contract_dep, UnparsedDepKind::Contract))
  ```

My preference would be for the latter, as it ensures a bit better type safety, and avoids the possibility of accidentally using the invalid salt in future refactors.

mitchmindtree (commented):
Nice, much better with salt in the `DepKind`!

forc-pkg/src/pkg.rs
kayagokalp: This is only used for displaying errors when trying to depend on a program with invalid type so I explicitly left salt out to not cause ultra verbose errors. 

https://github.com/FuelLabs/sway/blob/master/forc-pkg/src/pkg.rs#L817

docs/src/forc/manifest_reference.md
```diff
@@ -182,3 +182,12 @@ fn main() {
   let foo_id = foo::CONTRACT_ID;
 }
 ```
+
+Since contract id is created deterministcly, building the same contract would always result with the same contract id. Since two contracts with same contract id cannot be deployed, a salt is needed to add randomness to the contract id. For each contract dependency declared under `[contract-dependencies]`, a salt can be specified by providing it with the declaration, an example can be seen below.
```
mohammadfawaz, line 186:
```suggestion
Because the ID of a contract is computed deterministically, rebuilding the same contract would always result in the same contract ID. Since two contracts with same contract ID cannot be deployed on the blockchain, a "salt" factor is needed to modify the contract ID. For each contract dependency declared under `[contract-dependencies]`, `salt` can be specified. An example is shown below:
```

docs/src/forc/manifest_reference.md
```diff
@@ -182,3 +182,12 @@ fn main() {
...
+
+```toml
+[contract-dependencies]
+foo = { path = "../foo", salt = "0x1000000000000000000000000000000000000000000000000000000000000000" }
+```
+
+For contract dependencies that does not declare any salt, default salt is implicitly added (which is 32 bytes of 0).
```
mohammadfawaz, line 193:
```suggestion
For contract dependencies that do not specify any value for `salt`, a default of all zeros for `salt` is implicitly applied.
```

mohammadfawaz (commented):
Just some nits on the documentation. Will leave the review of the code to @FuelLabs/tooling.

forc-pkg/src/lock.rs
```diff
@@ -296,30 +296,25 @@ fn pkg_dep_line(
...
+        let mut iter = s.split('(');
+        let pkg_str = iter
+            .next()
+            .ok_or_else(|| anyhow!("missing pkg string"))?
+            .trim();
+        let salt = iter
+            .next()
+            .map(|s| s.trim())
+            .map(|s| &s[..s.len() - 1])
+            .and_then(|s| fuel_tx::Salt::from_str(s).ok());
```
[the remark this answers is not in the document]
mitchmindtree: Looking a little closer I see there's some code duplication where we check for and parse the salt here, as well as later on in the function.

The purpose of this condition where we check `!s.starts_with('(')` is only to check for the `dep_name`.

Rather than parsing the `pkg_str` and `salt` in this condition and returning early, how about we instead do something like this:

```rust
    let (dep_name, s) = match s.starts_with('(') {
        false => (None, s),
        true => {
            // If we have the open bracket, grab everything until the closing bracket.
            let s = &s["(".len()..];
            let mut iter = s.split(')');
            let dep_name = iter
                .next()
                .ok_or_else(|| anyhow!("missing closing parenthesis"))?;
            // The rest is the unique package string and possibly the salt.
            let s = &s[dep_name.len() + ")".len()..];
            (Some(dep_name), s)
        }
    };
```

This should help to avoid the early return and the code duplication related to salt parsing.

forc-pkg/src/lock.rs
```diff
@@ -296,30 +296,25 @@ fn pkg_dep_line(
 type ParsedPkgLine<'a> = (Option<fuel_tx::Salt>, Option<&'a str>, &'a str);
 // Parse the given `PkgDepLine` into its dependency name and unique string segments.
 //
-// I.e. given "(<salt>) (<dep_name>) <name> <source>", returns ("<salt>", "<dep_name>", "<name> <source>").
+// I.e. given "(<dep_name>) <name> <source> (<salt>)", returns ("<salt>", "<dep_name>", "<name> <source>").
```
mitchmindtree, line 303: Perhaps we should change the order of the returned tuple to match the order of the input, e.g.
```suggestion
// I.e. given "(<dep_name>) <name> <source> (<salt>)", returns ("<dep_name>", "<name> <source>", "<salt>").
```

mitchmindtree (changes requested):
Almost there! Just a couple more small things

test/src/e2e_vm_tests/test_programs/should_fail/contract_dependencies_conflicting_salt/contract_a/src/main.sw
```diff
@@ -1,16 +1,12 @@
...
-
 abi MyContract {
     fn test_function();
 }
 
 impl MyContract for Contract {
     fn test_function() {
-        let CONTRACT_B = CONTRACT_B_ID;
-        let contract_b_id = ContractId::from(CONTRACT_B);
+        let contract_b_id = ContractId::from(contract_b::CONTRACT_B);
```
mitchmindtree, line 9: woops
```suggestion
        let contract_b_id = ContractId::from(contract_b::CONTRACT_ID);
```

forc-pkg/src/lock.rs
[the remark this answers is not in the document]
kayagokalp: Yeah, that `.ok()` part was there for handling `salt` prepended to the `pkg_dep_line` (to distinguish from `dep_name`) since we moved it to the end there is no need for it anymore. Also removed the duplication as you suggested. Thanks!

test/src/e2e_vm_tests/test_programs/should_fail/contract_dependencies_conflicting_salt/contract_a/src/main.sw
kayagokalp: Looks like I should stop working at 3 AM.. 😵‍💫

mitchmindtree (approved):
:ok_hand:

otrho (approved):

Issue #3306, opened 2022-11-06
Defining constants multiple times should be a compile error


mohammadfawaz:
Unlike local variables where shadowing is allowed, `const` variable should not allow shadowing. Here's [a playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a0dff8464892240889d09407e9674138). Currently, we only get a warning in Sway:
```rust
script;

const X = 0;
const X = 1;

fn main() -> u64 {
    X
}
```
Warning:
```console
 1 |
2 |
3 | const X = 0;
  | ------------ This declaration is never used.
4 | const X = 1;
  |
____

  Compiled script "subcurrency" with 1 warning.
```

mohammadfawaz:
Of course, this should also apply to configuration-time constants, but I think that's documented in the RFC.

Issue #2585, opened 2022-08-19
Storage accesses in libraries


mohammadfawaz:
We need a way to support reading and writing storage variables in libraries. Storage variables in a contract behave like global variables. They are accessible from everywhere in the contract and don't need to be passed around. In libraries, this is different. We want to be able to write a library function that is able to read a storage variable or update it. For example, this is a rough idea of what we want:
```rust
// Some contract
...
storage {
    var: u64 = 0,
}
...
update_storage_variable(storage.var);
...
```
```rust
// Some library
fn update_storage_variable(storage x: u64) { // storage somewhat implies `ref mut`?
    storage.x = 42;
}
```
The syntax above is just an example... it may not be the best idea as is.

mohammadfawaz:
Or,
```rust
// Some contract
...
storage {
    var: u64 = 0,
}
...
update_storage_variable(storage.var);
...
```
```rust
// Some library
fn update_variable(ref mut x: u64) {
    x = 42;
}
```
and the fact that we're passing in a storage variable via `storage.var` means "modify _that_ variable"? The downside here is that the function `update_storage_variable` will sometimes be pure and sometimes be impure which will complicate storage annotations. Also, the lack of the `storage` keyword on `x = 42` is a deviation of our philosphy of making storage variables and. storage access as explicit as possible.

adlerjohn:
Could we do something like attach storage to a library import?

E.g.

```rust
library foo;

// To import this library, you must "attach" storage variables to it.
storage {
    bar: u64,
}
```

then

```rust
contract MyContract;

storage {
    s: u64,
}

// Attach `storage.s` to `bar`, syntax obviously subject to change.
use foo { bar: storage.s };
```

vaivaswatha:
> Could we do something like attach storage to a library import?
> 
> E.g.
> 
> ```rust
> library foo;
> 
> // To import this library, you must "attach" storage variables to it.
> storage {
>     bar: u64,
> }
> ```
> 
> then
> 
> ```rust
> contract MyContract;
> 
> storage {
>     s: u64,
> }
> 
> // Attach `storage.s` to `bar`, syntax obviously subject to change.
> use foo { bar: storage.s };
> ```

I vote for this approach.

mohammadfawaz:
To summarize: there seems to be multiple potential directions here:
1. Function-level `storage` similar to my initial suggestions where `storage` is added to a function argument.
[...]

PR #3536, opened 2022-12-07
Introduce Configuration-time Constants

### LINKED ISSUE #1498: Add support for configuration-time constants

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/1498

This PR adds configuration-time constants as described in Ref https://github.com/FuelLabs/sway-rfcs/pull/19. There is still some work to be done, namely:
- https://github.com/FuelLabs/sway/issues/3680
- https://github.com/FuelLabs/sway/issues/3679
- https://github.com/FuelLabs/sway/issues/3678
- https://github.com/FuelLabs/sway/issues/3306 (expanding that issue to consider all types of constants)
as well as some work in the SDKs and in `forc` to override these constants at deploy-time.

The design of this feature is relatively simply:

### Key Points
- Configuration-time constants are declared using a `configurable` block which is quite similar to a `storage` block. It also requires that each config constant is initialized (just like storage variables).
- Configuration-time constants show up in the JSON ABI as described in the [ABI spec](https://fuellabs.github.io/fuel-specs/master/protocol/abi/json_abi_format.html)
- There is a 1:1 mapping between each configuration-time constant and a particular location in the data section. The amount of bytes consumed in the data section is dependent on the type of the constant.
- The optimizer guarantees that configuration-time constants are not optimized in any way.
- The IR represents configuration-time constants are global scope (see IR tests)
- Configuration-time constants are not allowed in libraries for the time being.

### Parser
Converts a `configurable` block into a list of constant declaration with a new field that indicates whether the constant is configurable or not. In other words, the concept of a configurable block does not exist after `convert_parse_tree`.

### Frontend
The type checker is virtually unchanged.

### IR Gen and IR
Simply special casing global constants that have `is_configurable = true` and converting those to a special type of `Value`s called `Configurable`. This makes sure that the optimizer knows not to touch this type of values in any way.

I also had to add a new metadata to keep track of the name of each config. This is important to know what data value corresponds to what config const in codegen because that information is useful for the JSON object.

### Codegen
Mostly the same as regular constants except that we need to be careful about merging data values that are the same. We continue to merge values for regular consts but we make sure that configuration consts are uniqued and get their own unshared spot.

### JSON ABI
The configuration consts and their types are added to the JSON object in the frontend as usual. We later modify the object and add the actual bytecode offsets as computed in codegen.

### Raw Tasks:
- [x] Parser
- [x] Frontend
  - [ ] (deferred) Disallow multiple `configurable` blocks: will do this separately in https://github.com/FuelLabs/sway/issues/3678
- [x] new IR `Value` type: `Configurable`
  - [x] Printing
  - [x] Parsing
  - [x] New metadata `config_name` 
- [x] IR Gen, especially on `const_eval`
  - [x] Unit tests 
- [x] ASM gen
  - [ ] (deferred) Make sure unused configurable variables are still included in the data section: See https://github.com/FuelLabs/sway/issues/3680
  - [x] Unit tests
- [x] Handling of data section and returning a map: config name -> bytecode offset
- [x] JSON ABI generation + `forc-pkg` changes
- [x] E2E tests
- [ ] (deferred) Formatter. See: https://github.com/FuelLabs/sway/issues/3679


test/src/e2e_vm_tests/test_programs/should_pass/language/configurable_consts_in_lib/src/lib.sw
```diff
@@ -0,0 +1,32 @@
...
+configurable {
+    C0: bool = true,
+    C1: u64 = 42,
+    C2: b256 = 0x1111111111111111111111111111111111111111111111111111111111111111,
+    C3: MyStruct = MyStruct { x: 42, y: true },
+    C4: MyEnum = MyEnum::A(42),
+    C5: MyEnum = MyEnum::B(true),
+    C6: str[4] = "fuel",
+    C7: [u64; 4] = [1, 2, 3, 4],
+}
```
mitchmindtree, line 32: Perhaps we should be a little conservative here and disallow `configurable` blocks in `library`s? I think `configurable` in libraries raises many of the same questions as we have in #2585.

E.g. what if a library with a `configurable` block is depended on by two other library dependencies? Does it show up in the data section once for each time its depended on? Or only once altogether? How should `forc` handle `configurable` sections in transitive dependencies? We have ideas [here](https://github.com/FuelLabs/sway-rfcs/pull/19#issuecomment-1286351753) on how to handle top-level dependencies, but it's not yet clear how we'd handle transitive deps.

I think only allowing configurable in scripts, predicates and contracts for now is probably safe - it covers the only motivating use-case we came up with in RFC 19 ([contract factories](https://github.com/FuelLabs/sway-rfcs/pull/19/files#diff-47c0979605623f4b3f39c63e1b462abe798abcb7a775e833d4e91817f1c2bbe3R20)).

mohammadfawaz: Thank you @mitchmindtree, this makes sense and I'm happy to disallow this for now. You're right that we can cover the required use cases without libraries - I was strictly thinking of top level dependencies here and I suspect this PR disallows problematic transitive dependencies via the global check I have in program.rs, but I haven't tested that yet so I could be wrong 😆. Either way, I will disallow libraries for now and we can handle top level dependencies separately as part of the `forc` side of this feature.

mohammadfawaz: On that note, I think I will wrap up the RFC before marking this PR as ready for review. I won't make much progress this week though as I'm mostly ooo.

sway-core/src/metadata.rs
```diff
@@ -21,13 +21,15 @@ pub(crate) struct MetadataManager {
     md_storage_key_cache: HashMap<MetadataIndex, u64>,
     md_inline_cache: HashMap<MetadataIndex, Inline>,
     md_test_decl_index_cache: HashMap<MetadataIndex, usize>,
+    md_config_const_name_cache: HashMap<MetadataIndex, Arc<str>>,
```
otrho, line 24: I would use `Rc<str>` instead of `Arc`.  We don't need thread safety at this stage.

sway-ir/src/constant.rs
```diff
@@ -7,6 +7,12 @@ use crate::{
     value::Value,
 };
 
+/// A [`Type`] and constant value, including [`ConstantValue::Undef`] for uninitialized constants.
```
otrho, line 10:
```suggestion
/// A boot time configurable constant value.  I.e., a constant which may be initialised at program start.
```
Or something... 🙂

otrho: Actually, I don't think this is used?  You wrap a `ValueDatum::Configurable` around a `Constant` and don't seem to need this wrapper?

sway-ir/src/parser.rs
```diff
@@ -47,14 +48,26 @@ mod ir_builder {
                 / "predicate" _ { Kind::Predicate }
 
             rule contract() -> IrAstModule
-                = "contract" _ "{" _ fn_decls:fn_decl()* "}" _ metadata:metadata_decls() {
+                = "contract" _ "{" _ configs:init_config()* _ fn_decls:fn_decl()* "}" _ metadata:metadata_decls() {
```
otrho, line 51:
```suggestion
                = "contract" _ "{" _ configs:init_config()* fn_decls:fn_decl()* "}" _ metadata:metadata_decls() {
```

sway-ir/src/printer.rs
```diff
@@ -294,7 +334,7 @@ fn maybe_constant_to_doc(
     maybe_const_val: &Value,
 ) -> Doc {
     // Create a new doc only if value is new and unknown, and is a constant.
-    if !namer.is_known(maybe_const_val) && maybe_const_val.is_constant(context) {
+    if !namer.is_known(maybe_const_val) && (maybe_const_val.is_constant(context)) {
```
otrho, line 337: Extra parens?  Doesn't clippy usually warn about this?

sway-ir/src/value.rs
```diff
@@ -34,6 +34,7 @@ pub struct ValueContent {
 #[derive(Debug, Clone, DebugWithContext)]
 pub enum ValueDatum {
     Argument(BlockArgument),
+    Configurable(Constant),
```
otrho, line 37: Did you consider just parameterising `Constant` to be configurable?  It feels like it'd be simpler than creating another `ValueDatum` variant, but perhaps there are blockers to this approach?

test/src/ir_generation/tests/configs_1.sw
```diff
@@ -0,0 +1,25 @@
...
+// check: $(c0=$CONFIG) = config u64 42, $(config_mds=$MD)
+
+// check: ret u64 $c0
+// check: $(config_name=$MD) = config_name "X"
+// check: $config_mds = ($MD $config_name)
+
+// ::check-asm::
+
+// regex: IMM=i\d+
+// regex: REG=\$r\d+
```
otrho, line 25: This isn't doing anything.

test/src/ir_generation/tests/configs_4.sw
```diff
@@ -0,0 +1,29 @@
...
+
+// regex: DATA=data_\d+
+// regex: REG=\$r\d+
+
+// There should only be 3 data entries here. One shared by `A` and `B` and the
+// other two are for `X` and `Y` respectively
+
+// check: $DATA .word 42
+// check: $DATA .word 42
+// check: $DATA .word 42
```
otrho, line 29: Why don't you give them different values just to be absolutely sure the constants are merged and not the configs.

otrho (commented):
Cool.  I can see this conflicting with at least two other current PRs... 🙂 

I don't see a test to absolutely confirm that the configurables are immutable, not sure how important that is.  Otherwise it looks pretty OK, other than a couple of things below.

sway-ir/src/constant.rs
mohammadfawaz: True. Remnant of an earlier attempt. Removed.

sway-ir/src/printer.rs
mohammadfawaz: Weird indeed. Fixed.

sway-ir/src/value.rs
mohammadfawaz: My reasoning was that, if someone is writing an optimization pass that may touch constants, they have to know to check a flag called `is_configurable` if we were to parametrize `Constant`, and they have to do that everywhere. By making `Configurable` its own thing that is _initialized_ by a const but _not really_ a const, we force future developers to handle `Configurable` separately, or not at all, and existing passes continue to behave as they should (i.e. ignoring `Configurable`).

sway-core/src/metadata.rs
mohammadfawaz: Done

mohammadfawaz:
> I don't see a test to absolutely confirm that the configurables are immutable

Added a test for that and addressed all comments. Thanks!

otrho (approved):

forc-pkg/Cargo.toml
```diff
@@ -13,7 +13,7 @@ anyhow = "1"
 fd-lock = "3.0"
 forc-tracing = { version = "0.33.0", path = "../forc-tracing" }
 forc-util = { version = "0.33.0", path = "../forc-util" }
-fuels-types = "0.33"
+fuel-abi-types = "0.1"
```
mitchmindtree, line 16: Looks like we just need to update `forc-client` to point to `fuel-abi-types` too, then I think all of the sway repo crates will be on `fuel-abi-types`.

sway-core/src/lib.rs
```diff
@@ -222,7 +222,7 @@ fn module_path(parent_module_dir: &Path, dep: &sway_ast::Dependency) -> PathBuf
 pub struct CompiledAsm(pub FinalizedAsm);
 
 /// The bytecode for a sway program.
-pub struct CompiledBytecode(pub Vec<u8>);
+pub struct CompiledBytecode(pub (Vec<u8>, BTreeMap<String, u64>));
```
mitchmindtree, line 225: Could be nice to have a type alias for this for readability, e.g.

```rust
pub type ConfigConstOffsets = BTreeMap<String, u64>;
```
Or alternatively add field names
```rust
pub struct CompiledBytecode {
    pub bytecode: Vec<u8>,
    pub config_const_offsets: BTreeMap<String, u64>,
}
```

mitchmindtree (commented):
:rocket:  

Is the plan to document these in a follow-up PR?

forc-pkg/Cargo.toml
mohammadfawaz: Looks like `forc-client` still needs `fuels-types` for `Bech32Address`: https://github.com/FuelLabs/sway/blob/847cbccd32ba0a1a5cc6399e837f43b9b0e4726c/forc-plugins/forc-client/src/ops/tx_util.rs#L14

which is not in `fuel-abi-types` unfortunately.

sway-core/src/lib.rs
mohammadfawaz: Great idea! I went with the second solution and also moved the same logic to `finalized_asm.rs`.

mohammadfawaz:
> Is the plan to document these in a follow-up PR?

Indeed. Only because these config const are not very useful right now without additional changes in `forc` and/or the SDK to allow us to actually change them.

mitchmindtree (approved):

Issue #3012, opened 2022-10-12
Unnecessary warnings for logged structs


mohammadfawaz:
Logging events is a common pattern in smart contracts. For example, [our subcurrency example](https://fuellabs.github.io/sway/v0.25.2/examples/subcurrency.html) has the following:
```rust
log(Sent {
    from: sender,
    to: receiver,
    amount: amount,
});
```
where the struct `Sent` is not used anywhere else and its fields are not accessed. Currently, the compiler will emit the following warning when compiling subcurrency:
```console
im srwarning
  --> /Users/mohammad/Desktop/fuel/sway/examples/subcurrency/src/main.sw:15:5
   |
13 |
14 | struct Sent {
15 |     from: Address,
   |     ------------- This struct field is never accessed.
16 |     to: Address,
17 |     amount: u64,
   |
____

warning
  --> /Users/mohammad/Desktop/fuel/sway/examples/subcurrency/src/main.sw:16:5
   |
14 |
15 |     from: Address,
16 |     to: Address,
   |     ----------- This struct field is never accessed.
17 |     amount: u64,
18 | }
   |
____

warning
  --> /Users/mohammad/Desktop/fuel/sway/examples/subcurrency/src/main.sw:17:5
   |
15 |
16 |     to: Address,
17 |     amount: u64,
   |     ----------- This struct field is never accessed.
18 | }
   |
____

  Compiled contract "subcurrency" with 3 warnings.
  Bytecode size is 3856 bytes.
❯ vim src/main.sw
```
These warnings are not very useful in this particular case and the user has no way of getting rid of them, especially that the user is not doing anything wrong. In fact, the code follows best practices. I think the solution here might be to rely on the `__log` intrinsic and make sure that we don't emit warning in the situation above.

PR #3538, opened 2022-12-07
Fix DCA warnings for log intrinsic and stdlib function

tritao:
This adds special handling for the log intrinsic, consuming the fields of all structs used inside the inner expressions.

It also does the same for functions external to the current analysis, to solve the same thing for the stdlib log function, which could lead to some false negatives when it comes to the unused fields warning, but it's the best we can do for now without considerable effort refactoring the DCA engine to work inter-proceduraly/module.

Closes https://github.com/FuelLabs/sway/issues/3012.


mohammadfawaz (commented):
Looks good to me! Will approve after @otrho's comment is addressed.

mohammadfawaz (approved):

otrho (approved):

PR #3545, opened 2022-12-08
Adds function returning only `fn signature` in function code blocks

### LINKED ISSUE #3540: Functions in `forc doc` include function body
We should remove the function body from documentable functions, like cargo doc.

eureka-cpu:
This PR adds a function `trim_fn_body()` that returns only the function signature for documentable function code blocks.

Result:
Screen Shot 2022-12-08 at 1 25 46 PM

Closes #3540


mitchmindtree (approved):

mohammadfawaz (approved):

forc-plugins/forc-doc/src/render.rs
```diff
@@ -480,6 +480,13 @@ fn attrsmap_to_html_string(attributes: &AttributesMap) -> String {
     options.parse.default_info_string = Some("sway".into());
     markdown_to_html(&format_docs(&docs), &options)
 }
+/// Takes a formatted String fn and returns only the function signature.
+fn trim_fn_body(f: String) -> String {
+    match f.find('{') {
+        Some(index) => f.split_at(index).0.to_string(),
+        None => f,
+    }
+}
```
mitchmindtree, line 489: Not important, but this might be better written as:

```rust
fn trim_fn_body(f: &str) -> &str {
    f.split('{').next().unwrap_or(f)
}
```

This way we:
1. don't require taking ownership of the `String`
2. don't require allocating a new `String` upon returning

eureka-cpu: Ah nice! I'll fix that now

eureka-cpu: Ah, actually the function that's receiving the output is taking a String. I was planning on opening an issue to reduce the amount of String allocations soon. I'll tackle this then!

mitchmindtree: When writing functions in Rust, it's generally good practise to make the abstraction itself minimal cost, and allow the caller of the function to make the choice of taking on extra costs like resource usage or dynamic allocation (e.g. `trim_fn_body(...).to_string()`).

That said, it really doesn't matter in this case, and sometimes it is worth taking on costs for code ergonomics! Just a tip to keep in mind :)

PR #3546, opened 2022-12-08
Include test function span information (file, line number) in forc test output

### LINKED ISSUE #3266: Include test function span information (file, module path, line number) in `forc test` output

kayagokalp:
closes #3266.

If there are any visual improvement recommendations that would be really cool.

## Summary

This PR adds details for failing tests. An example can be found below. The details are only printed for failing tests. Unlike cargo, our line number for the failing tests shows the test function declaration whereas cargo shows the failing assert statement's line number.

image

## Implementation Details

I am not sure how to find out line number for the specific `assert` but other than that (and also we don't have module paths as they are not available right now: https://github.com/FuelLabs/sway/issues/3266#issuecomment-1341821771). I collected file path and line number for failing tests. To find out the line number, I had to read the file declaring the test as `Span` only contains beginning and ending char indices not the line numbers.


forc-test/src/lib.rs
```diff
@@ -98,6 +107,26 @@ impl TestResult {
...
+            .path()
+            .ok_or_else(|| anyhow::anyhow!("Missing span for test function"))?
+            .to_owned();
+        let span_start = self.span.start();
+        let file_str = fs::read_to_string(&*file_path)?;
+        let line_number = file_str[..span_start]
+            .chars()
+            .into_iter()
+            .filter(|&c| c == '\n')
+            .count();
```
otrho, line 124: I would love for spans to have the line number info, as provided by the parser, but if this is working and accurate then it's good enough. :+1:

otrho (approved):

mitchmindtree (approved):

mohammadfawaz (approved):
Very nice.

Issue #3543, opened 2022-12-08
Non-existent enum variant does not show where the illegal enum was used


bitzoic:
When using an enum and attempting to use a non-exising variant, the compiler currently points to the enum definition rather than where the non-existing enum variant is used.

For example, in the following `my_lib.sw` file, there is an enum called `MyEnum` and it contains two variants:

```rust
library my_lib;

pub enum MyEnum {
    State1: (),
    State2: (),
}

impl core::ops::Eq for MyEnum {
    fn eq(self, other: Self) -> bool {
        match (self, other) {
            (MyEnum::State1, MyEnum::State1) => true,
            (MyEnum::State2, MyEnum::State2) => true,
            _ => false,
        }
    }
}
```

In the following contract, I attempt to use `MyEnum::State3` which does not exist:

```rust
contract;

dep my_lib;

use my_lib::MyEnum;

abi MyContract {
    fn test_function(my_enum: MyEnum) -> bool;
}

impl MyContract for Contract {
    fn test_function(my_enum: MyEnum) -> bool {
        require(my_enum == MyEnum::State3, "Not state 3");

        true
    }
}
```

This is the resulting error:

![EnumError](https://user-images.githubusercontent.com/54727135/206480854-57fc3c0f-725c-4dc6-a7b1-0e4fda05da2f.png)

Rather than stating where in my project I've used the invalid enum, I am directed to the enum definition and now have to hunt down where I've misused the enum type. This should instead point to the illegal statement.

PR #3551, opened 2022-12-09
Fix span for unknown enum variant error

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/3543

The new span now points to the actual enum instantiation where the error is, as opposed to the enum declaration.

image


otrho (approved):

JoshuaBatty (approved):
nice, thanks.

PR #3557, opened 2022-12-09
formatter: fix handling of statements ending with `;` in `if else` blocks

### LINKED ISSUE #3555: formatter: fmt tries to incorrectly fmt `if-else` into a single line
This piece of code:

```rust
script;

fn main() {
    if foo {
        let x = 1;
    } else {
        bar(y);
    }
}
```

formats into:
```rust
script;

fn main() {
    if foo { let x = 1;
 } else { bar(y); }
}
```

bingcicle:
unblocks #3556
closes #3555 

~The situation occurring in the issue seems to happen when the formatter tries to (incorrectly) assume that an if expression can be in-lined when there's a newline character inside. During this check, the line kind (multi-line vs inline) is decided solely depending on the width of the line, ignoring the existence of a `\n` character.~

~This PR introduces something like a 'catch-all' during the above check to default to `LineKind::Multiline` if a `\n` is found during the look-ahead formatting done by the temporary formatter, which forces the formatter to use multi-line style, so that at the very least we don't end up with a broken inline format as seen in the issue.~

~This catch-all might unblock the comment formatting PR above as well.~

The above was my misunderstanding - the mistake was further upstream instead within statement formatting. The inline style was not being handled for let statements.

### DIFF
--- a/swayfmt/src/utils/language/expr/conditional.rs
+++ b/swayfmt/src/utils/language/expr/conditional.rs
@@ -71,7 +71,6 @@ fn get_full_width_line_style(
         |temp_formatter| -> Result<LineStyle, FormatterError> {
             let mut if_expr_str = FormattedCode::new();
             format_if_expr(if_expr, &mut if_expr_str, temp_formatter)?;
-
             let if_expr_width = if_expr_str.chars().count();
 
             temp_formatter.shape.code_line.update_width(if_expr_width);
--- a/swayfmt/src/utils/language/expr/conditional.rs
+++ b/swayfmt/src/utils/language/expr/conditional.rs
@@ -72,10 +72,6 @@ fn get_full_width_line_style(
             let mut if_expr_str = FormattedCode::new();
             format_if_expr(if_expr, &mut if_expr_str, temp_formatter)?;
 
-            if if_expr_str.contains('\n') {
-                return Ok(LineStyle::Multiline);
-            }
-
             let if_expr_width = if_expr_str.chars().count();
 
             temp_formatter.shape.code_line.update_width(if_expr_width);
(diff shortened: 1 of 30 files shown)


swayfmt/src/utils/language/expr/conditional.rs
```diff
@@ -71,6 +71,11 @@ fn get_full_width_line_style(
         |temp_formatter| -> Result<LineStyle, FormatterError> {
             let mut if_expr_str = FormattedCode::new();
             format_if_expr(if_expr, &mut if_expr_str, temp_formatter)?;
+
+            if if_expr_str.contains('\n') {
+                return Ok(LineStyle::Multiline);
+            }
```
mitchmindtree, line 77: Intuitively I'd expect that, if the formatter determines the length is small enough for the expr to be inlined, it should remove the newlines. :thinking:

In #3555 it looks like the newlines were correctly removed for the else block and the first newline in the "then" block, but the final newline wasn't formatted correctly.

I guess we can land this as a patch, but perhaps we should open an issue to revisit handling this properly?

mitchmindtree (approved):

swayfmt/src/utils/language/expr/conditional.rs
bingcicle: Let me look into this again, maybe the fix should be somewhere upstream instead

bingcicle: So I relooked at it again and the correct place to look at was indeed upstream when formatting statements [here](https://github.com/FuelLabs/sway/pull/3557/files#diff-b8982d881172082dbaa5ac05c3b19c79ebaa8a7800c719169c630b41d28fef68L71) - inline was not handled

kayagokalp (approved):
Looks nice to me!

eureka-cpu (approved):
Good catch @bingcicle ! Thanks for this

Issue #1305, opened 2022-04-19
Unnecessary warning when a struct member is not accessed but the struct is returned


mohammadfawaz:
Compiling the following causes a warning:
```rust
contract;

abi MyContract {
    fn foo(s: S) -> S;
}

struct S {
    x: u:64,
}

impl MyContract for Contract {
    fn foo(s: S) -> S {
        s 
    }
}
```
Result:
```bash
❯ forc build
  Compiled library "core".
  Compiled library "std".
warning
  --> proj/src/main.sw:15:5
   |
13 |
14 | struct S {
15 |     x: u64,
   |     - This struct field is never accessed.
16 | }
   |
____

  Compiled contract "proj" with 1 warning.
  Bytecode size is 76 bytes.
```
The warning is out of place here since this is perfectly legal code and even though `x` is never accessed in `foo` directly, it is being implicitly used.

mohammadfawaz:
@tritao can https://github.com/FuelLabs/sway/pull/2468 be extended to avoid the warning for struct/enum fields as well? I just tested the above and still seems to be an issue. I don't think we should be emitting a warning.

mohammadfawaz:
My reasoning is that there is nothing wrong with the example above. I see a warning as something that the user has to address which usually means that they did something wrong, which is not the case here.

> Or are you thinking we just enable it in the cases of contract ABIs, like https://github.com/FuelLabs/sway/pull/2468 was originally implemented before we generalized it? In that case, then I think it would work.

Maybe! I think that actually makes sense. Basically, `x` does not need to be accessed inside the contract because it may actually be accessed outside. This applies to both input structs and output structs. Now that I'm thinking about it, `s` here is _kind_ of `pub` or `external`, not in the sense that it can be used in other Sway files, but in the sense that the SDK can access it.

mohammadfawaz:
I think the same reasoning applies to library functions. So, a library function that takes a struct and just returns it without accessing any of its element should not cause a warning because the elements _may_ be accessed outside the library.

mohammadfawaz:
> The only thing that comes to mind is we can assume structs to always have public visibility for contracts (but not libraries) which would fix this particular case. What do you think?

@tritao If we do this, then unused structs in contract won't cause a warning right?

I think the issue here is that you have to assume that `return s;` _consumes_ all of `s`'s fields, so no warning is required. Maybe we just need to special case returns and implicit returns to connect all struct fields.

otrho:
[...]

PR #3562, opened 2022-12-09
Fix DCA warnings for unused struct fields in contracts.

tritao:
If a struct type is used as a return type in the interface surface of the contract, then assume that any fields inside the struct can be used outside of the contract.

Closes https://github.com/FuelLabs/sway/issues/1305.

### DIFF
--- a/sway-core/src/control_flow_analysis/dead_code_analysis.rs
+++ b/sway-core/src/control_flow_analysis/dead_code_analysis.rs
@@ -3,12 +3,12 @@ use crate::{
     declaration_engine::{declaration_engine::*, DeclarationId},
     language::{parsed::TreeType, ty, CallPath, Visibility},
     type_system::TypeInfo,
-    TypeEngine,
+    TypeEngine, TypeId,
 };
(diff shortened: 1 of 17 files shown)


sway-core/src/control_flow_analysis/dead_code_analysis.rs
```diff
@@ -456,13 +456,14 @@ fn connect_struct_declaration(
         .iter()
         .map(|field| (field.name.clone(), graph.add_node(field.into())))
         .collect::<Vec<_>>();
-    // If this is a library or smart contract, and if this is public, then we want to connect the
+    // If this is a smart contract, or a library and the struct is public, then we want to connect the
```
mohammadfawaz, line 459: Wouldn't this mean that a struct that is only partially used in a contract will no longer generate a warning? The issue in #1305 is that a struct that is _returned_ from a contract should not generate a warning even if the contract does not directly accesses its fields.

vaivaswatha (dismissed):

mohammadfawaz (approved):

vaivaswatha (approved):

Issue #3457, opened 2022-11-28
Introduce an `__smo` intrinsic and messages types in the JSON ABI


mohammadfawaz:
This is a wrapper around `smo` that also encode the ID of the type of the message as the first word of the message data. Similarly to `log`, this type ID is also emitted to the JSON ABI (see https://github.com/FuelLabs/fuel-specs/issues/435). 

We can then create an alternative version of `std::messages::send_message` that takes a value of an arbitrary (non-heap) type and calls `__smo` under the hood.

PR #3564, opened 2022-12-09
Introduce `__smo` intrinsic

mohammadfawaz:
Closes https://github.com/FuelLabs/sway/issues/3457

Following how `__log` works for the most part.

Tasks:
- [x] Parser
- [x] Frontend
- [x] JSON ABI (following https://github.com/FuelLabs/fuel-specs/pull/444)
- [x] IR
- [x] IR Gen
  - [x] Unit tests
- [x] Codegen
  - [x] Unit tests
- [x] Standard library change
  - [x] SDK test 
- [x] E2E tests
- [x] Documentation for `__smo`

Nothing particularly special about this change except for IR gen. The new intrinsic has the following signature:
```rust
__smo<T>(recipient: b256, data: T, output_index: u64, coins: u64)
```
while the IR instruction has the following signature:
```rust
smo recipient_and_message, message_size, output_index, coins
```
which matches the `smo` opcode. IR gen builds `recipient_and_message` by creating a new struct where the first field is the recipient (a `b256`), the second field is the `smo` ID (just like log IDs, but we don't have a special register to use here so we need to use the message data), and the third field is the actual user data passed in via the intrinsic:
```rust
struct recipient_and_message {
    recipient: b256,
    smo_id: u64,
    user_data: T
}
```

The `smo_id` above is the same as the one used in the JSON ABI for each `smo`.

Assembly builder simply convert the IR instruction into an `smo` opcode directly.

Finally, add a `send_typed_message` to the standard library that allows sending arbitrary types and added an SDK test for it. The SDK can't decode the types yet so I had to check for raw bytes.


mohammadfawaz:
This is ready but cannot be merged until we have a new release of `fuels-rs` that includes the changes I need `fuels-types`

mohammadfawaz:
Due to various dependency issues with `fuels-rs`, I had to bump the `fuel-core` version as well as run `cargo update` here.

mohammadfawaz:
Sorry for the trivial JSON ABI file diffs, again

otrho (commented):
Do we need to add an SDK test to make sure the message actually gets through correctly?  Is that possible?

vaivaswatha (dismissed):
Looks fine to me (after Toby's comments). So approving in advance.

sway-core/src/lib.rs
```diff
@@ -402,6 +409,10 @@ pub(crate) fn compile_ast_to_ir_to_asm(
         errors.extend(e);
     }
 
+    if build_config.print_ir {
+        tracing::info!("{}", ir);
+    }
+
```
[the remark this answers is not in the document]
mohammadfawaz: Ouf yes.. Added this for debugging and forgot to remove it.

sway-lib-std/src/message.sw
```diff
@@ -43,3 +43,26 @@ pub fn send_message(recipient: b256, msg_data: Vec<u64>, coins: u64) {
     }
     revert(FAILED_SEND_MESSAGE_SIGNAL);
 }
+
+/// Sends a message `msg_data` of type T to `recipient` with a `coins` amount of the base asset
```
nfurfaro, line 47: Would it be accurate to say that `send_typed_message` is what should be used if you want to be able to index the messages?

nfurfaro (dismissed):
I just left 1 question but looks great overall!

sway-lib-std/src/message.sw
mohammadfawaz: Yup! Added that as a comment. I also changed `send_typed_message` to look somewhat like `transfer_to_address` in the sense it now takes an `Address` and the recipient is the last argument instead of the first.

otrho (approved):

nfurfaro (approved):

PR #3566, opened 2022-12-10
fix: reserved keywords are checked before creating dir with forc new

### LINKED ISSUE #3565: Executing `forc new` with reserved keyword still creates the folder
Raised by @deekerno in slack:

Executing `forc new test` throws an error regarding `test` being a reserved keyword but it still creates the folder `test`.

```console
~/test-index/contracts
❯ forc new test
Error: the name `test` cannot be used as a project name, it conflicts with Sway's built-in test library

~/test-index/contracts
❯ ls
drwxr-xr-x - deekerno  9 Dec 17:43 test
```

kayagokalp:
closes #3565.

### DIFF
--- a/forc/src/cli/commands/new.rs
+++ b/forc/src/cli/commands/new.rs
@@ -1,7 +1,8 @@
 use crate::{cli::init::Command as InitCommand, ops::forc_init::init};
-use anyhow::{bail, Result};
+use anyhow::{anyhow, bail, Result};
 use clap::Parser;
-use std::path::Path;
+use forc_util::validate_name;
+use std::path::{Path, PathBuf};
 
 /// Create a new Forc project at `<path>`.
 #[derive(Debug, Parser)]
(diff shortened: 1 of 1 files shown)


forc/src/cli/commands/new.rs
```diff
@@ -43,6 +44,20 @@ pub(crate) fn exec(command: Command) -> Result<()> {
         path,
     } = command;
 
+    match &name {
```
bingcicle, line 47: An alternative way of doing this is to handle it with clap's value_parser ([fuelup example](https://github.com/FuelLabs/fuelup/blob/07381e1597568d6f200d50c97e3780bab8edd4a3/src/commands/toolchain.rs#L29-L30)). However the changes look good as they are so this is just an alternative way of doing this.

bingcicle (approved):

forc/src/cli/commands/new.rs
```diff
@@ -43,6 +44,20 @@ pub(crate) fn exec(command: Command) -> Result<()> {
...
+        Some(name) => validate_name(name, "project name")?,
+        None => {
+            // If there is no name specified for the project directory name will be used by default
```
bingcicle, line 50:
```suggestion
            // If there is no name specified for the project, the 'path' input will be used by default
```

minor docs nit here

bingcicle (changes requested):

forc/src/cli/commands/new.rs
kayagokalp: I explicitly stated directory name as the name of the created dir is what we need to check not the path (that is why we are getting the 'file_name' from the path before checking.

bingcicle: Ah, got it. I think maybe I got a bit confused myself reading it😅

bingcicle (approved):

deekerno (approved):

PR #3569, opened 2022-12-10
Check for errors after type checking type arguments.

### LINKED ISSUE #3196: Omitting `Option` type when calling `std::storage::get` results in internal compiler error
When calling the `get` method from the standard library's storage library and omitting the `Option` type will crash the compiler. 

For example, the following function results in the error shown below:

```
    #[storage(read)]
    fn test_function() {
        let result = get::<Option>(VAL);
    }
```

generic_crash

This has been tested with forc v0.25.2

vaivaswatha:
Closes #3196

### DIFF
--- a/sway-core/src/type_system/type_binding.rs
+++ b/sway-core/src/type_system/type_binding.rs
@@ -154,6 +154,10 @@ impl TypeBinding<CallPath> {
             );
         }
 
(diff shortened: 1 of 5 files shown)


mohammadfawaz (approved):
Awesome :)

PR #3581, opened 2022-12-13
Fix CI failure by running `cargo update` and bumping `fuel-core`

mohammadfawaz:
This unfortunately happened because I made a non-breaking release in `fuels-rs` that was in fact breaking. The breaking was non-breaking from the point of view of the user, but it was breaking from the point of view of the sway repo. 
- Had to run `cargo update` to update the `fuels-rs` version
- Had to bump `fuel-core` to `0.15` due to some transitive dependencies and conflicting types.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -254,7 +254,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       fuel-core:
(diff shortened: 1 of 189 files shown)


bingcicle (approved):

otrho (approved):
183 `json_abi_oracle.json` files, all very much indifferent to `messagesTypes`... :grin:

PR #3588, opened 2022-12-13
Remove duplicate gas from std

### LINKED ISSUE #2464: `std` contains a duplicate function for getting the gas in the current context
## Summary

Two functions exist under different names in two different places. I will argue that the first one should remain (because of the explicit name that differentiates from the global gas) and the one at the bottom should be removed. Although "context_gas" may not be the best name imo.

https://github.com/FuelLabs/sway/blob/a3ffd94c413ca62a8fc130a55b296ee61d6ab1e5/sway-lib-std/src/context/registers.sw#L54-L66

https://github.com/FuelLabs/sway/blob/a3ffd94c413ca62a8fc130a55b296ee61d6ab1e5/sway-lib-std/src/context.sw#L23-L28

jtriley-eth:
Removed `gas` function in favor of `context_gas`.

Resolves #2464

### DIFF
--- a/sway-lib-std/src/context.sw
+++ b/sway-lib-std/src/context.sw
@@ -18,11 +18,6 @@ pub fn balance_of(asset_id: ContractId, target: ContractId) -> u64 {
     }
 }
 
-/// Get the remaining gas in the context.
-pub fn gas() -> u64 {
-    asm() { cgas: u64 }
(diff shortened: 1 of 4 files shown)


mohammadfawaz (approved):
Looks good! Thanks.

PR #3589, opened 2022-12-13
Jtriley eth/bitwise trait impls

### LINKED ISSUE #3456: Implement bitwise operations for `u8`, `u16`, `u32` in `sway-lib-core`
The traits `BitwiseAnd`, `BitwiseOr`, and `BitwiseXor` are only implemented for `u64` (and `b256`) but not for `u8`, `u16`, and `u32`. I don't see a good reason for not implementing those.

jtriley-eth:
Draft PR for bitwise impls.

Progress: Created boilerplate for new impls and tests.

Resolves #3456


otrho (approved):
This is good, thanks.

In https://github.com/FuelLabs/sway/issues/3456#issuecomment-1350215833 I actually meant we could have an `ops` parent directory and then subdirs for each type, as in `ops/u64_ops`, `ops/u32_ops`, etc.  But this is OK too.

mohammadfawaz (approved):
Great work! Thank you :)

PR #3604, opened 2022-12-15
Add a `cargo package` CI job to check pkgs are publish-friendly

mitchmindtree:
`cargo package` assembles a local package into a distributable tarball. It can be thought of as doing a dry-run for the publish command, however unlike `cargo publish --dry-run`, the `cargo package` command is compatible with running at the workspace level (as introduced in this PR) and will automatically package each project in the workspace one-at-a-time.

The main benefit of this job is to catch errors caused by `cargo package`'s sandboxed build, e.g. attempting to statically link to assets outside of the project's directory.

This job *might* take an excessive amount of time due to building each package one-at-a-time in the sandboxed environment. I'm not sure if the Rust cache that we use accounts for caching projects built via `cargo package`. We'll have to keep an eye on the times and see if it's worth it!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -348,6 +348,7 @@ jobs:
     # those are local packages.
     if: github.event_name != 'release' && github.event.action != 'published'
     strategy:
+      fail-fast: false
       matrix:
         # Include all workspace members we want to publish to crates.io.
         pkg: [
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -341,7 +341,7 @@ jobs:
           command: udeps
           args: --locked --all-targets
 
-  cargo-publish-dry-run:
+  cargo-package:
     strategy:
       matrix:
         # Include all workspace members we want to publish to crates.io.
@@ -376,8 +376,8 @@ jobs:
       - uses: Swatinem/rust-cache@v1
       - uses: actions-rs/cargo@v1
         with:
-          command: publish
-          args: --locked --dry-run -p ${{ matrix.pkg }}
+          command: package
+          args: --locked -p ${{ matrix.pkg }}
 
   notify-slack-on-failure:
     needs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -341,6 +341,20 @@ jobs:
           command: udeps
           args: --locked --all-targets
 
+  cargo-package-workspace:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+      - uses: Swatinem/rust-cache@v1
+      - uses: actions-rs/cargo@v1
+        with:
+          command: package
+          args: --locked
+
   notify-slack-on-failure:
     needs:
       [
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -342,6 +342,11 @@ jobs:
           args: --locked --all-targets
 
   cargo-package:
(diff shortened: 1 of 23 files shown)


mitchmindtree:
Hmmmmm it looks like `cargo package` is taking quite a bit longer than the rest of our jobs.

Perhaps it is worth setting up the job as a matrix that checks each package in parallel? If we want to take this route, seeing as we'd be testing each package individually, we could potentially just use `cargo publish --dry-run` (e.g. `cargo publish --dry-run -p ${{ matrix.pkg }}`).

mitchmindtree:
Ahhh it looks like the dry-runs have run before the crates have all been fully published following #3606. Will re-run the workflow in a bit!

mitchmindtree:
Thinking on this more, the dry-run publish jobs will *always* fail when we do version bumps to publish new versions. This is because the dependencies all need to be published before a dry-run can go ahead :thinking: I'll switch these parallel jobs back to `cargo package` in an attempt to avoid this issue.

bingcicle:
Perhaps we could run `cargo package` only when a PR is labelled as `release`?

mitchmindtree:
> Thinking on this more, the dry-run publish jobs will always fail when we do version bumps to publish new versions. This is because the dependencies all need to be published before a dry-run can go ahead thinking

Hmmmmmmmmm it appears we have the same issue with `cargo package`.

I'll switch this back into draft mode until we can come up with a nice solution here.

mohammadfawaz:
> > Thinking on this more, the dry-run publish jobs will always fail when we do version bumps to publish new versions. This is because the dependencies all need to be published before a dry-run can go ahead thinking
> 
> Hmmmmmmmmm it appears we have the same issue with `cargo package`.
> 
> I'll switch this back into draft mode until we can come up with a nice solution here.

Ah yes I noticed the same locally. Running `cargo package` after the version bump did not work.

mitchmindtree:
> Perhaps we could run cargo package only when a PR is labelled as release?

Ahhhh good thinking! I think we actually want the *opposite* of this - that is, only do the `cargo package` check when we're **not** publishing a new release, as the `release` PRs will always have dependencies set to versions that are not yet published.

bingcicle:
> I think we actually want the opposite of this

Wouldn't that mean each new PR takes about 30~mins to run?

> as the `release` PRs will always have dependencies set to versions that are not yet published.

We might be able to do this: check out the repo separately in a job, use `cargo-edit` to bump the version back down in a PR labelled as `release`, run `cargo package`. This is what the [nightly workflow](https://github.com/FuelLabs/sway-nightly-binaries/blob/bd65b219715fc7bb89e2e90117868f2d5d55d6e7/.github/workflows/nightly-forc-release.yml#L127-L137) does to add nightly metadata to the versions

mitchmindtree:
> Wouldn't that mean each new PR takes about 30~mins to run?

This is why I parallelized these jobs (they finish far more quickly now).

> We might be able to do this: check out the repo separately in a job, use cargo-edit to bump the version back down in a PR labelled as release, run cargo package. This is what the [nightly workflow](https://github.com/FuelLabs/sway-nightly-binaries/blob/bd65b219715fc7bb89e2e90117868f2d5d55d6e7/.github/workflows/nightly-forc-release.yml#L127-L137) does to add nightly metadata to the versions

This sounds interesting - would this require always maintaining the previous version in the CI file somewhere?

Another problem I'm foreseeing is that, these cargo-package checks would also make it impossible to add two new crates (where one depends on the other) to the workspace at once, as the dependency crate will not be published yet. We could remedy this by ensuring that new crates are only ever added one-at-a-time, and that we cut a new release each time a new crate is added, but this could also add a lot of unnecessary churn/delay if both crates are closely related.

Issue #3575, opened 2022-12-12
Enhance docs and expand the `Result` library


mohammadfawaz:
Similarly to https://github.com/FuelLabs/sway/pull/3322, we should expand and enhance the `Result` library and make it look at lot like https://doc.rust-lang.org/src/core/result.rs.html.

PR #3608, opened 2022-12-15
Result enhancements

jtriley-eth:
Enhanced `Result` documentation and add `unwrap_ok` method.

> Notice: Transforms via `ok(self)` and `err(self)` are not possible to implement until import
> limitations are resolved (option and result modules cannot currently import one another).

Resolves: #3575


test/src/e2e_vm_tests/test_programs/should_pass/stdlib/result/src/main.sw
```diff
@@ -35,3 +36,12 @@ fn test_unwrap_ok() {
         revert(0);
     }
 }
+
+fn test_unwrap_or() {
```
mohammadfawaz, line 40: Do you mind expanding this whole test for `Result` into a more elaborate one similarly to what we did for `Option` here: https://github.com/FuelLabs/sway/tree/master/test/src/e2e_vm_tests/test_programs/should_pass/stdlib/option?

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/result/src/main.sw
```diff
@@ -7,6 +7,7 @@ fn main() -> bool {
     test_ok();
     test_err();
     test_unwrap_ok();
+    test_unwrap_or();
```
mohammadfawaz, line 10: Do you mind expanding this whole test for `Result` into a more elaborate one similarly to what we did for `Option` here https://github.com/FuelLabs/sway/pull/3322?

sway-lib-std/src/result.sw
```diff
@@ -4,6 +4,55 @@
...
+//! ## Extracting the contained value
+//! 
+//! These methods exctract the contained value in a [`Result<T,E>`] when it is
+//! the [`Ok`] variant. If the [`Result`] is [`Err`]:
+//! 
+//! * [`unwrap`] reverts
+//! * [`unwrap_or`] returns the default provided value
+//! 
+//! [`unwrap`] Result::unwrap
+//! [`unwrap_or`] Result::unwrap_or
```
mohammadfawaz, line 55: Similarly to https://github.com/FuelLabs/sway/pull/3322, it would be useful to follow `result.rs` and expand the doc comments for all the methods in this file. This should look very much like Rust's `result.rs` to the extent possible.

jtriley-eth: Added examples to each method.

sway-lib-std/src/result.sw
```diff
@@ -25,6 +74,20 @@ impl<T, E> Result<T, E> {
...
+    /// enum Error {
+    ///     NotFound,
+    ///     Invalid,
+    /// }
+    /// 
+    /// let x: Result<u64, Error> = Result::Ok(42);
+    /// assert(x.is_ok());
+    /// 
+    /// let y: Result<u64, Error> = Result::Err(Error::NotFound));
+    /// assert(!x.is_ok());
```
otrho, line 90: I think you'll need to close the code blobs with '```' too, I imagine `forc-doc` will complain one day.

jtriley-eth: Corrected. Thank you

sway-lib-std/src/result.sw
```diff
@@ -4,6 +4,55 @@
 //! errors. It is an enum with the variants, [`Ok(T)`], representing
 //! success and containing a value, and [`Err(E)`], representing error
 //! and containing an error value.
+//! 
+//! Functions return [`Result`] whenever errors are expected and recoverable. In
+//! the `std` crate, [`Result`] is most prominently used for `Identity`
+//! interactions and cryptographic operations.
+//! 
+//! A simple function returning Result might be defined and used like so:
```
mohammadfawaz, line 12:
```suggestion
//! A simple function returning [`Result`] might be defined and used like so:
```

mohammadfawaz (approved):

otrho (approved):

PR #3610, opened 2022-12-15
Deep bug fix to type unification

### LINKED ISSUE #3330: Initializing Options and misdirection via Generics
#### Code

Forc: `0.30.0`

1. Storage variable `game` uses `Option::None` without issues
2. The contract function `new_game()` inserts twice into storage
    - First insertion requires "complex" syntax in order to work (not user friendly)
    - Second insertion is there to demonstrate the error when used in the preferred sense
3. Note how point `1` works but point `2` requires different syntax

```rust
contract;

struct Game {
    winner: Option<Identity>,
}

abi TicTacToe {
    #[storage(write)]
    fn new_game();
}

storage {
    // NOTE: we can use `Option::None` here and it just works
    game: Game = Game { winner: Option::None },

    game_boards: StorageMap<u64, Option<Identity>> = StorageMap {},
}

impl TicTacToe for Contract {
    #[storage(write)]
    fn new_game() {
        // Note: "Complex" syntax required for `Option::None` ...
        storage.game_boards.insert(1, Option::None::<Identity>());

        // Note: Will fail to compile
        storage.game_boards.insert(1, Option::None);
    }
}
```

### Rust

Tangentially, note that the following is acceptable in Rust

### LINKED ISSUE #3530: Failure to return a proper type error when using mixed generic types.
error: Internal compiler error: Verification failed: Type mismatch found for call to 'push_181'.

```#[storage(read)]
    pub fn to_vec(self, key: K) -> Vec<V> {
        // The length of the vec is stored in the sha256((key, __get_storage_key())) slot
        let k = sha256((key, __get_storage_key()));
        let len = get::<u64>(k);
        let mut i = 0;
        let mut vec = Vec::new();
        while len > i {
            let k = sha256((key, i, __get_storage_key()));
            let item = get::<K>(k);
            vec.push(item);
            i += 1;
        }
        vec
    }
```
    
This code seems to have caused the error.

emilyaherbert:
This PR introduces a new type variant called `TypeInfo::Placeholder`---we can think of this type as the `_` type. At a high level, before this PR, for any new types that were generated in the type system that had type parameters, they would be generated with those type parameters directly. A fresh `Result` would generate in the type system initially as `Result<T, E>`. This forced us to add some complexities (i.e. questions like, how do we unify two generics of different names? See the removal of `unify_right` for an example of this). Now, new types that are generated in the type system that have type parameters are generated with placeholder types (i.e. `_`), which are then resolved to more specific types at the earliest logical opportunity (see the [new `should_fail/generics_in_contract` test](https://github.com/FuelLabs/sway/blob/c7f8e4af2211de7142a11eb9fc05cda6ded0d59f/test/src/e2e_vm_tests/test_programs/should_pass/language/generics_in_contract/src/main.sw) for an example of this). This allows us to perform better type unification (see the [new type unification compatibility check](https://github.com/FuelLabs/sway/blob/c7f8e4af2211de7142a11eb9fc05cda6ded0d59f/sway-core/src/type_system/unify_check.rs#L92) that [*happens before*](https://github.com/FuelLabs/sway/blob/c7f8e4af2211de7142a11eb9fc05cda6ded0d59f/sway-core/src/type_system/type_engine.rs#L290) the [new simplified type unification](https://github.com/FuelLabs/sway/blob/c7f8e4af2211de7142a11eb9fc05cda6ded0d59f/sway-core/src/type_system/unify.rs#L78)).

The placeholder type indicates _no type at all_, and it is more general than the `TypeInfo::UnknownGeneric`. It is perhaps equally as general as the `TypeInfo::Unknown` type, but the `TypeInfo::Placeholder(..)` variant holds a `TypeParameter` of the original `TypeParameter` found in that spot that the placeholder type is in so that we can do pretty printing.

Future work:
- Expose the placeholder type to users.
- Consider combining the new concept of unification compatibility with type subsets, as the algorithms and concepts are similar.

Closes #3530
Closes #3330


sway-core/src/engine_threading.rs
```diff
@@ -42,6 +42,14 @@ impl<'a> Engines<'a> {
             engines: *self,
         }
     }
+
+    /// Helps out some `thing: T` by adding `self` as context.
+    pub fn help_out_with_self<T>(self, thing: T) -> WithEngines<'a, T> {
```
otrho, line 47: I see `help_out()` and this function... is there a better name for it?  'Help out' is a bit ambiguous.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression/struct_instantiation.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    for struct_field in struct_fields.iter() {
+        if let Some(typed_field) = typed_fields.iter().find(|x| x.name == struct_field.name) {
+            check!(
+                CompileResult::from(type_engine.unify_adt(
+                    declaration_engine,
+                    typed_field.value.return_type,
+                    struct_field.type_id,
+                    &typed_field.value.span,
+                    "Struct field's type must match up with the type specified in its declaration.",
```
otrho, line 206:
```suggestion
                    "Struct field's type must match the type specified in its declaration.",
```

test/src/e2e_vm_tests/test_programs/should_fail/bad_generic_var_annotation/test.toml
```diff
@@ -1,13 +1,8 @@
...
-# nextln: $()expected: bool
-# nextln: $()found:    str[3].
-# nextln: $()help: The argument that has been provided to this function's type does not match the declared type of the parameter in the function declaration.
-
-# check: let g: str[3] = three_generics(true, "foo", 10);
-# nextln: $()Mismatched types.
-# nextln: $()expected: str[3]
-# nextln: $()found:    bool.
+# nextln: $()expected: B
+# nextln: $()found:    A.
```
otrho, line 7: This is great, though would it be possible to have both the generic name and the concrete name in the error?

test/src/e2e_vm_tests/test_programs/should_fail/type_mismatch_error_message/test.toml
```diff
@@ -1,3 +1,15 @@
 category = "fail"
 
-# check: $()No method named "does_not_exist" found for type "Result<Data<bool>, str[4]>".
+# check: $()error
+# check: should_fail/type_mismatch_error_message/src/main.sw:18:5
```
otrho, line 4:
```suggestion
# check: type_mismatch_error_message/src/main.sw:18:5
```
We're planning on removing the `should_pass` and `should_fail` dirs one day I think.  IMO it's etter not to encode them in the tests.

otrho (commented):
Looks good!  A couple of minor comments below.

test/src/e2e_vm_tests/test_programs/should_fail/bad_generic_var_annotation/test.toml
emilyaherbert: I don't think we would want that. The equivalent Rust error for the same code is:

```
error[[E0308]](https://doc.rust-lang.org/stable/error-index.html#E0308): mismatched types
 --> src/main.rs:9:18
  |
6 | fn three_generics<A, B, C>(a: A, b: B, c: C) -> B {
  |                   -  - expected type parameter
  |                   |
  |                   found type parameter
...
9 |   let new_a: B = a;
  |              -   ^ expected type parameter `B`, found type parameter `A`
  |              |
  |              expected due to this
  |
  = note: expected type parameter `B`
             found type parameter `A`
```

See: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a68bc92d1e7dfd75cbee763153d42830

sway-core/src/engine_threading.rs
emilyaherbert: I changed it to `with_thing`---`help_out` is from a separate PR. Let me know if you'd prefer something other than `with_thing` 😄

otrho: I guess!

otrho (approved):

PR #3617, opened 2022-12-16
Use `try_get` methods on `DashMap` to avoid potential deadlocks

JoshuaBatty:
This PR simply changes calling `.get` to `.try_get` and `.get_mut` to `.try_get_mut` on types that use a `DashMap` in the language server. These functions ensure that a result is only returned if the shard is not locked. 

There was an issue where an LSP test was hanging in CI only rarely a while ago and it was because an internal lock wasn't getting released in `DashMap`. This PR makes sure we are handling this situation in all places we are calling these methods in the server.


otrho (approved):
Seems fair enough.

mohammadfawaz (approved):
:shipit:

PR #3619, opened 2022-12-16
Reorder block creation for `while` loops.

### LINKED ISSUE #3580: Internal compiler error: Program erroneously uses uninitialized virtual registers.
Getting this error when trying to compile the `vault` module.

image

**Reproduce**: https://github.com/Reimao/Kitsch/tree/refactor/system-design

otrho:
Closes #3580.

Unfortunately the register allocator is a little bit fragile and requires all register uses are listed after their definitions, regardless of topological order.  When compiling `while` loops we have put the 'end' block before the body which could cause these sort of out of order definitions.  This change ensures `while` bodies are between the conditional and end blocks.

Improving the allocator (perhaps including the ability for spilling) should be done after we refactor the ASM generation, in particular after we introduce SSA and basic blocks to the abstract program.  Performing a true topological traversal of the ASM without BBs would be much, much harder otherwise.  The ASM gen refactor is mostly covered in #2505.

### DIFF
--- a/sway-core/src/ir_generation/function.rs
+++ b/sway-core/src/ir_generation/function.rs
@@ -1461,23 +1461,25 @@ impl<'te> FnCompiler<'te> {
         // We're dancing around a bit here to make the blocks sit in the right order.  Ideally we
         // have the cond block, followed by the body block which may contain other blocks, and the
         // final block comes after any body block(s).
+        //
+        // NOTE: This is currently very important!  There is a limitation in the register allocator
+        // which requires that all value uses are after the value definitions, where 'after' means
+        // later in the list of instructions, as opposed to in the control flow sense.
+        //
+        // Hence the need for a 'break' block which does nothing more than jump to the final block,
+        // as we need to construct the final block after the body block, but we need somewhere to
+        // break to during the body block construction.
 
         // Jump to the while cond block.
         let cond_block = self.function.create_block(context, Some("while".into()));
-
         if !self.current_block.is_terminated(context) {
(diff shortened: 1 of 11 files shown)


sway-core/src/ir_generation/function.rs
```diff
@@ -1461,23 +1461,25 @@ impl<'te> FnCompiler<'te> {
         // We're dancing around a bit here to make the blocks sit in the right order.  Ideally we
         // have the cond block, followed by the body block which may contain other blocks, and the
         // final block comes after any body block(s).
+        //
+        // NOTE: This is currently very important!  There is a limitation in the register allocator
```
mohammadfawaz, line 1465: Damn it, is this hitting us again? Do we plan on enforcing SSA eventually or just looking for a register allocator algorithm that is less fussy?

otrho: This is exactly the problem I mentioned to you a couple of weeks ago, about the allocator not using topological traversal for liveness analysis.  But before we fix that it would really make sense to translate/preserve the basic blocks from the IR to the abstract ASM.

vaivaswatha (approved):

otrho:
The new 'break' blocks will disappear with `simplify-cfg` either deleting them or possibly merging them.

I've also snuck in a couple of other changes -- Vaivas improved the inliner, I added a little check in the IR verifier which caught a bug at some point, plus an update to the E2E test harness to make it easier to fix the hard coded contract IDs when they go bad.

mohammadfawaz (approved):

Issue #3609, opened 2022-12-15
Test `lsp_syncs_with_workspace_edits` is flaky


sdankel:
It failed here: https://github.com/FuelLabs/sway/actions/runs/3704653621/jobs/6277535455

```
---- server::tests::lsp_syncs_with_workspace_edits stdout ----
thread 'server::tests::lsp_syncs_with_workspace_edits' panicked at 'assertion failed: `(left == right)`
  left: `Ok(Some(Response { jsonrpc: Version, result: Object {"range": Object {"end": Object {"character": Number(11), "line": Number(19)}, "start": Object {"character": Number(7), "line": Number(19)}}, "uri": String("file:///home/runner/work/sway/sway/test/src/e2e_vm_tests/test_programs/should_pass/language/doc_comments/src/main.sw")}, id: Number(2) }))`,
 right: `Ok(Some(Response { jsonrpc: Version, result: Object {"range": Object {"end": Object {"character": Number(11), "line": Number(20)}, "start": Object {"character": Number(7), "line": Number(20)}}, "uri": String("file:///home/runner/work/sway/sway/test/src/e2e_vm_tests/test_programs/should_pass/language/doc_comments/src/main.sw")}, id: Number(2) }))`', sway-lsp/src/server.rs:775:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

And passed after rerunning.

PR #3620, opened 2022-12-16
Only parse project if the changes could be written to disk

JoshuaBatty:
There was a [CI failure earlier](https://github.com/FuelLabs/sway/actions/runs/3704653621/jobs/6277535455) on one of the LSP tests.

On further investigation, we seemed to reparse the project on `did_change` events, even if the changes were unable to be written to disk. This PR updates the logic so this can't happen. 
  
should hopefully solve #3609

### DIFF
--- a/sway-lsp/src/error.rs
+++ b/sway-lsp/src/error.rs
@@ -33,6 +33,10 @@ pub enum DocumentError {
     ManifestsLockPathFailed { dir: String },
     #[error("Document is already stored at {:?}", path)]
     DocumentAlreadyStored { path: String },
+    #[error("File wasn't able to be created: {:?}", err)]
+    UnableToCreateFile { err: String },
+    #[error("Unable to write string to file: {:?}", err)]
+    UnableToWriteFile { err: String },
 }
 
 #[derive(Debug, Error, PartialEq, Eq)]
--- a/sway-lsp/src/error.rs
+++ b/sway-lsp/src/error.rs
@@ -33,10 +33,10 @@ pub enum DocumentError {
     ManifestsLockPathFailed { dir: String },
     #[error("Document is already stored at {:?}", path)]
     DocumentAlreadyStored { path: String },
-    #[error("File wasn't able to be created: {:?}", err)]
-    UnableToCreateFile { err: String },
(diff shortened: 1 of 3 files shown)


sway-lsp/src/error.rs
```diff
@@ -33,6 +33,10 @@ pub enum DocumentError {
     ManifestsLockPathFailed { dir: String },
     #[error("Document is already stored at {:?}", path)]
     DocumentAlreadyStored { path: String },
+    #[error("File wasn't able to be created: {:?}", err)]
+    UnableToCreateFile { err: String },
+    #[error("Unable to write string to file: {:?}", err)]
+    UnableToWriteFile { err: String },
```
mitchmindtree, line 39:
```suggestion
    UnableToWriteFile { err: io::Error },
```
Might be better to propagate the original error (which still impls `Debug` and `Display`) for these, but this could be done in a future PR as it looks like many of the other error variants are also currently "stringly typed".

mitchmindtree (dismissed):

sway-lsp/src/error.rs
JoshuaBatty: Nice, that's a good idea. Just made the suggested change.

sway-lsp/src/error.rs
```diff
@@ -33,9 +35,13 @@ pub enum DocumentError {
     ManifestsLockPathFailed { dir: String },
     #[error("Document is already stored at {:?}", path)]
     DocumentAlreadyStored { path: String },
+    #[error("File wasn't able to be created: {:?}", err)]
+    UnableToCreateFile { err: Error },
```
sdankel, line 39: nit: could be good to include the filepath in the error

JoshuaBatty: good idea.

sway-lsp/src/error.rs
JoshuaBatty: Actually, I just reverted this change as we need `Eq` and `PartialEq` implemented for our Error types so we can compare them in the LSP tests.

sdankel (approved):

kayagokalp (approved):

PR #3625, opened 2022-12-16
Switch to `Bytes` type for `send_message` data param

nfurfaro:


otrho (approved):
Strictly speaking 'recipient_heap_buffer' is on the stack, not the heap, but :shrug:.

sway-lib-std/src/message.sw
```diff
@@ -1,46 +1,44 @@
...
 ///
 /// ### Arguments
 ///
 /// * `recipient` - The address of the message recipient
-/// * `msg_data` - Arbitrary length message data
-/// * `coins` - Amount of base asset to send
-pub fn send_message(recipient: b256, msg_data: Vec<u64>, coins: u64) {
+/// * `msg_data` - arbitrary length message data
+/// * `coins` - Amount of base asset sent
+pub fn send_message(recipient: b256, msg_data: Bytes, coins: u64) {
```
mohammadfawaz, line 16: Do you find the updated description better? I've actually changed it recently in my `__smo` PR but I see you're now changing it back to how it was before. My reasoning at the time was that the definition of this function has no `output` nor `msg_len`.

sway-lib-std/src/message.sw
```diff
@@ -1,46 +1,44 @@
...
+    }
 
     let mut index = 0;
     let outputs = output_count();
 
     while index < outputs {
         let type_of_output = output_type(index);
         if let Output::Message = type_of_output {
-            asm(r1: recipient_heap_buffer, r2: size * 8, r3: index, r4: coins) {
+            asm(r1: recipient_heap_buffer, r2: size, r3: index, r4: coins) {
```
mohammadfawaz, line 34: Maybe we should rename `recipient_heap_buffer` to `recipient_and_msg_data_pointer`?

mohammadfawaz (commented):
Thanks for doing this @nfurfaro 🙌

mohammadfawaz (approved):
LGTM! Thanks 🙏

sway-lib-std/src/message.sw
nfurfaro: I restored the improved comments, they are indeed better. :)

test/src/sdk-harness/test_projects/token_ops/mod.rs
```diff
@@ -387,11 +387,8 @@ async fn can_send_message_output_with_data() {
     assert_eq!(*fuelcoin_id, **message_receipt.sender().unwrap());
     assert_eq!(&recipient_address, message_receipt.recipient().unwrap());
     assert_eq!(amount, message_receipt.amount().unwrap());
-    assert_eq!(24, message_receipt.len().unwrap());
-    assert_eq!(
-        vec![0, 0, 0, 0, 0, 0, 0, 100, 0, 0, 0, 0, 0, 0, 0, 75, 0, 0, 0, 0, 0, 0, 0, 50],
-        message_receipt.data().unwrap()
-    );
+    assert_eq!(8, message_receipt.len().unwrap());
+    assert_eq!(vec![100, 75, 50], message_receipt.data().unwrap());
```
Braqzen, line 391: If the type is changed why is the test also changed? 
Shouldn't this be the same as before?

I see that the `token_ops` has changed so I assume that has something to do with it.

test/src/sdk-harness/test_projects/token_ops/src/main.sw
```diff
@@ -53,6 +53,13 @@ impl TestFuelCoin for Contract {
     }
 
     fn send_message(recipient: b256, msg_data: Vec<u64>, coins: u64) {
-        send_message(recipient, msg_data, coins);
+        let mut data = Bytes::new();
+        if msg_data.len() > 0 {
+            data.push(msg_data.get(0).unwrap());
+            data.push(msg_data.get(1).unwrap());
+            data.push(msg_data.get(2).unwrap());
```
Braqzen, line 60: Why do we hard code an assumption of 3 data points instead of looping over any amount of data in the vec?

test/src/sdk-harness/test_projects/token_ops/mod.rs
nfurfaro: This is because of the difference in the way `u8` and `u64` are encoded in Rust and Sway.
In Sway, both are 8 bytes long, whereas in Rust a `u8` is only a single byte.
Previously the tests were passing in `vec![100, 75, 50]` as the data. This was a `Vec<u64>` in sway, and so we had to expect the padded `vec![0, 0, 0, 0, 0, 0, 0, 100, 0, 0, 0, 0, 0, 0, 0, 75, 0, 0, 0, 0, 0, 0, 0, 50]` in our assertion.
Now that the data has the type `Bytes`, there is no more padding and so we can assert using the unpadded 
`vec![100, 75, 50]`.

test/src/sdk-harness/test_projects/token_ops/src/main.sw
mitchmindtree: Unrolled loop for extreme test efficiency :smile:

mitchmindtree (approved):

test/src/sdk-harness/test_projects/token_ops/src/main.sw
nfurfaro: No good reason. The assumption of 3 data points in a contract which only exists for testing was good enough to get the job done but could be made more robust with a while loop.

PR #3626, opened 2022-12-16
Low level call std-lib functionality

simonr0204:
Low level call wrapper. Closes https://github.com/FuelLabs/sway/issues/2144

Can be used to call contract B from contract A, when the ABI for the former was unknown at compile time of the latter (so using `abi(Interface, contract_id)` is not possible.

e.g `call_with_function_selector(target: ContractId, function_selector: Bytes, calldata: Bytes, call_params: CallParams)`

For example, this can be used by a DAO contract or a multisig where arbitrary call execution (without ABI) is necessary.

Notes: 

- An additional function is provided which converts between `Vec<u8>` and `Bytes` in Sway. This will be deprecated once the SDK supports `Bytes`.
- The current tests use `call_with_function_selector` with manually crafted selector and calldata. A solution to support `call_with_raw_payload` directly from the SDK is being worked on [here](https://github.com/FuelLabs/fuels-rs/issues/775).


otrho:
So far you have a contract with a single word-sized arg working here.  FYI for contracts with larger-than-word-sized-args then you need to pass a pointer, and for contracts with multiple args you need to put each arg into a struct and ABI encode that, and pass its pointer.

simonr0204:
> Even the word-sized example here takes a pointer to a `Bytes` representation of the word.

simonr0204:
> So far you have a contract with a single word-sized arg working here. FYI for contracts with larger-than-word-sized-args then you need to pass a pointer, and for contracts with multiple args you need to put each arg into a struct and ABI encode that, and pass its pointer.

Even the word-sized example here takes a pointer to a `Bytes` representation of the word.

otrho:
> Even the word-sized example here takes a pointer to a `Bytes` representation of the word.

Oh, you're right, I thought it was encoding the `42` into bytes and then just `join()`ing that to the payload.  But it's taking the `.buf.ptr` out of the calldata.  I would not expect this to work.  Is it passing the test?

simonr0204:
It takes the `buf.ptr` out of the _payload_, which is composed of the contract id, parameter 1 (function selector) and parameter 2 (pointer to Encoded function args). Test passes!

simonr0204:
However, a basic test for a type larger than 1 word panics on a `MemoryOverflow`. I'll get back to this in the new year.

simonr0204:
So it seems a single-word copy type argument is actually the special case in that it is provided by value (as it fits in the 8 bytes of `second parameter`). For reference type, `second parameter` is a pointer.

For functions of multiple parameters, IIUC they're encoded as though the function takes a single struct argument (so a pointer to that struct will be sufficient).  i.e. `f(a: u64, b: u64)` is equivalent to `f({a: u64, b: u64})`. 

We'll need some way of distinguishing these two cases in order to handle all possible function calls.

sway-lib-std/src/low_level_call.sw
```diff
@@ -0,0 +1,105 @@
...
+    };
+
+    payload
+}
+
+/// Call a target contract with a function selector and calldata, provided as `Bytes`
+/// `target` : The contract ID of the contract to be called
+/// `function_selector` : The function selector of the function to be called. E.g. bytes8(sha256("my_func(u64)"))
+/// `calldata` : The encoded arguments with which to call the function
+/// `single_value_type_arg` : Whether the function being called takesa single value-type argument
```
otrho, line 78:
```suggestion
/// `single_value_type_arg` : Whether the function being called takes a single value-type argument
```

sway-lib-std/src/low_level_call.sw
```diff
@@ -0,0 +1,105 @@
...
+    } else {
+        payload = payload.join(ptr_as_bytes(calldata.buf.ptr)); // When calldata is reference type, need to get pointer as bytes
+    };
+
+    payload
+}
+
+/// Call a target contract with a function selector and calldata, provided as `Bytes`
+/// `target` : The contract ID of the contract to be called
+/// `function_selector` : The function selector of the function to be called. E.g. bytes8(sha256("my_func(u64)"))
```
otrho, line 76: Is `bytes8(sha256("my_func(u64)"))` an actual thing that works?  If not it might be confusing.

otrho (commented):
Looks good to me.

sway-lib-std/src/low_level_call.sw
simonr0204: This is just pseudo-code. I can write it out long form if it could be misleading.

simonr0204: Resolved [`fd47f11`](https://github.com/FuelLabs/sway/pull/3626/commits/fd47f114981a5ee6c63276b7dce4668861e13940)

sway-lib-std/src/low_level_call.sw
```diff
@@ -0,0 +1,105 @@
...
+fn create_payload(
+    target: ContractId,
+    function_selector: Bytes,
+    calldata: Bytes,
+    single_value_type_arg: bool,
+) -> Bytes {
+    // packs args according to spec (https://github.com/FuelLabs/fuel-specs/blob/master/src/vm/instruction_set.md#call-call-contract) :
+    /*
+    bytes	type	    value	description
+    32	    byte[32]	to	    Contract ID to call.
```
nfurfaro, line 57: nit: everything is aligned so nicely except this line :)

sway-lib-std/src/low_level_call.sw
```diff
@@ -0,0 +1,105 @@
...
+fn contract_id_to_bytes(contract_id: ContractId) -> Bytes {
+    let mut target_bytes = Bytes::with_capacity(32);
+    target_bytes.len = 32;
+
+    __addr_of(contract_id).copy_bytes_to(target_bytes.buf.ptr, 32);
+
+    target_bytes
+}
+
+/// Represent a raw pointer as a Bytes
```
nfurfaro, line 26: Perhaps an additional note describing this data structure is in order? It still seems a bit like a staircase designed by M.C Escher even though we know it works :)

sway-lib-std/src/low_level_call.sw
```diff
@@ -0,0 +1,105 @@
...
+
+    // Need to copy pointer to heap so it has an address and can be copied onto the bytes buffer
+    let mut ptr_on_heap = Vec::new();
+    ptr_on_heap.push(ptr);
+    ptr_on_heap.buf.ptr.copy_bytes_to(bytes.buf.ptr, 8);
+
+    bytes
+}
+
+/// Call a target contract with an already-encoded payload
```
nfurfaro, line 40: Could you possibly add a simple 1-liner comment with a common use-case  for each of these functions to help make it super clear when you would choose one option over another?

test/src/sdk-harness/test_projects/low_level_call/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+macro_rules! fn_selector {
+    ( $fn_name: ident ( $($fn_arg: ty),* )  ) => {
+         ::fuels::core::code_gen::function_selector::resolve_fn_selector(stringify!($fn_name), &[$( <$fn_arg as ::fuels::core::Parameterize>::param_type() ),*]).to_vec()
+    }
+}
+macro_rules! calldata {
+    ( $($arg: expr),* ) => {
+        ::fuels::core::abi_encoder::ABIEncoder::encode(&[$(::fuels::core::Tokenizable::into_token($arg)),*]).unwrap().resolve(0)
+    }
+}
```
nfurfaro, line 16: Are these macros going to be no longer needed once there's better SDK support for these operations?

nfurfaro (commented):
Just a few small comments, but looks great overall :)

sway-lib-std/src/low_level_call.sw
simonr0204: Currently this function is private, and only called by  `call_with_function_selector` after `create_payload` has been used to assemble the payload. In a future iteration (SDK support permitting) this function can be `pub` and called using a raw payload directly generated by the SDK.

test/src/sdk-harness/test_projects/low_level_call/mod.rs
simonr0204: These are actually super convenient in their own right (shout out to @segfault-magnet ). I wonder if they merit being added to the SDK as helper functions? 

Once we have full payload generation in the SDK, they _probably_ won't be necessary.

sway-lib-std/src/low_level_call.sw
simonr0204: Resolved [`73ff24c`](https://github.com/FuelLabs/sway/pull/3626/commits/73ff24bc9612bca45298c80d90236f8146efa55f)

sway-lib-std/src/low_level_call.sw
simonr0204: Resolved [`73ff24c`](https://github.com/FuelLabs/sway/pull/3626/commits/73ff24bc9612bca45298c80d90236f8146efa55f)

nfurfaro (approved):

otrho (approved):

PR #3628, opened 2022-12-16
Add `--stdlib` option to `forc new` and `forc init`

### LINKED ISSUE #3627: forc new/init should allow overriding the path to standard library
It could look something like this

```
forcc new <project-directory> --stdlib path/to/sway-lib-std
```

I'd propose to also relativize the path to stdlib w.r.t. the new project root, i.e. if one creates a new `forc` project deep inside the Sway `test` directory, it gives the following manifest file:

```
cd fuel/sway
forc new test/src/e2e_vm_tests/test_programs/should_pass/static_analysis/cei_pattern_violation --stdlib ./sway-lib-std
cat test/src/e2e_vm_tests/test_programs/should_pass/static_analysis/cei_pattern_violation/Forc.toml
```

```
[project]
<skipped>
[dependencies]
std = { path = "../../../../../../../sway-lib-std" }
```

Notice `std = { path = "../../../../../../../sway-lib-std" }` and not `std = { path = "./sway-lib-std" }`.

And if the user really wants an absolute path, `forc` should respect that, so we get the following result:

```
cd fuel/sway
forc new test/src/e2e_vm_tests/test_programs/should_pass/static_analysis/cei_pattern_violation --stdlib $(realpath ./sway-lib-std)
cat test/src/e2e_vm_tests/test_programs/should_pass/static_analysis/cei_pattern_violation/Forc.toml
```
and here is the contents of the manifest file in this case:
```
[project]
<skipped>
[dependencies]
std = { path = "/Users/anton/fuel/sway/sway-lib-std" }
```

anton-trunov:
It lets the user to specify the path to the standard library. If the path to stdlib is relative, `forc` makes it relative w.r.t. the project directory. If the path is absolute, forc uses it as is.

close #3627

### DIFF
--- a/forc/src/cli/commands/new.rs
+++ b/forc/src/cli/commands/new.rs
@@ -23,9 +23,15 @@ pub struct Command {
     /// Adding this flag creates an empty workspace.
     #[clap(long)]
     pub workspace: bool,
-    /// Set the package name. Defaults to the directory name
+    /// Set the package name. Defaults to the directory name.
     #[clap(long)]
     pub name: Option<String>,
+    /// The path at which the standard library is located.
+    /// Does not override the std dependency if missing.
+    /// If the path is relative, forc makes it relative w.r.t. the project directory.
+    /// If the path is absolute, forc uses it as is.
+    #[clap(long)]
+    pub stdlib: Option<String>,
     /// The path at which the project directory will be created.
     pub path: String,
 }
@@ -41,6 +47,7 @@ pub(crate) fn exec(command: Command) -> Result<()> {
         library,
         workspace,
         name,
+        stdlib,
         path,
     } = command;
 
@@ -77,6 +84,7 @@ pub(crate) fn exec(command: Command) -> Result<()> {
         predicate,
         library,
         workspace,
+        stdlib,
         name,
     };
 
--- a/forc/src/cli/commands/new.rs
+++ b/forc/src/cli/commands/new.rs
@@ -27,7 +27,8 @@ pub struct Command {
     #[clap(long)]
     pub name: Option<String>,
     /// The path at which the standard library is located.
-    /// Does not override the std dependency if missing.
+    /// If unspecified, the std will be implicitly included as a git dependency
+    /// via the git tag that matches `forc`'s version.
     /// If the path is relative, forc makes it relative w.r.t. the project directory.
     /// If the path is absolute, forc uses it as is.
     #[clap(long)]
--- a/forc/src/cli/commands/new.rs
+++ b/forc/src/cli/commands/new.rs
@@ -26,11 +26,11 @@ pub struct Command {
     /// Set the package name. Defaults to the directory name.
     #[clap(long)]
     pub name: Option<String>,
-    /// The path at which the standard library is located.
+    /// The local directory path or URL at which the standard library is located.
     /// If unspecified, the std will be implicitly included as a git dependency
     /// via the git tag that matches `forc`'s version.
-    /// If the path is relative, forc makes it relative w.r.t. the project directory.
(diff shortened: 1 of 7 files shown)


nfurfaro (dismissed):
LGTM.

forc/src/cli/commands/new.rs
```diff
@@ -23,9 +23,15 @@ pub struct Command {
     /// Adding this flag creates an empty workspace.
     #[clap(long)]
     pub workspace: bool,
-    /// Set the package name. Defaults to the directory name
+    /// Set the package name. Defaults to the directory name.
     #[clap(long)]
     pub name: Option<String>,
+    /// The path at which the standard library is located.
+    /// Does not override the std dependency if missing.
```
mitchmindtree, line 30:
```suggestion
    /// If unspecified, the std will be implicitly included as a git dependency
    /// via the git tag that matches `forc`'s version.
```

Perhaps it's worth clarifying that we're talking about the case where `stdlib` is *unspecified* like this? Otherwise, "missing" might be misinterpreted as though we fall back if the given std path doesn't exist?

forc/src/cli/commands/new.rs
```diff
@@ -23,9 +23,15 @@ pub struct Command {
...
+    /// If the path is relative, forc makes it relative w.r.t. the project directory.
+    /// If the path is absolute, forc uses it as is.
+    #[clap(long)]
+    pub stdlib: Option<String>,
```
mitchmindtree, line 34: Perhaps we should consider naming this flag `--std-path`, as it only allows overriding the implicit `std` with a path dependency (e.g. you can't use this to specify a `git` URL).

anton-trunov: This is a very good point! I added a commit to allow URLs as `--stdlib` values. It does not support branches and things like that, but it could be a step in the right direction. If the need arises, we can later modify it to support branches/tags with a syntax like this:
```
forc new foo --stdlib https://github.com/FuelLabs/sway-libs#branch-name
```

WDYT? (If it does not feel right, I can revert that commit and rename `--stdlib` into `--std-path`)

mitchmindtree: I'm not quite sure of the motivation behind #3627, but it seems like the idea is to make it easier for us to create new projects that point to our local version of `std` from our E2E tests directory, as it can be awkward to manually write the correct depth of parent path nesting (i.e. `../../../../../../../../../sway-lib-std`) in the `Forc.toml` due to how deeply those tests are nested - does that sound right?

If so, it seems to me like this flag has a very niche use that is only really relevant to us as Sway compiler devs. I imagine it will be very rare that:

1. Downstream (non-Sway compiler devs) will ever need to override the std dependency manually and
2. For those rare few that do need to specify std, they can just edit the manifest directly like they do for all other dependencies, or use a future `cargo-edit` style command as described in [this issue](https://github.com/FuelLabs/sway/issues/2369) if they need a single shell command, e.g. something like `forc new my-proj && forc add std --path ./path/to/std`.

For these reasons, I'm a bit weary about including the flag in forc's public interface at all, as it may just add noise to the `forc new --help` output for the vast majority of devs.

If we do still want to add it to make our own lives as sway compiler devs easier, maybe we're better off keeping it as simple as possible and stick to supporting just paths? Let me know if I'm missing the point!

anton-trunov: That was precisely my motivation, you got it right.

Indeed, it makes more sense to go with something like

```
forc new my-proj && forc add std --path ./path/to/std
```

didn't know it was in the works :).

Btw, would it be a good idea to add this *path relativization* feature from this PR (which I'm going to close) to the future `forc add` command?

PR #3630, opened 2022-12-16
Impl From<b256> for Bytes

### LINKED ISSUE #3561: Impl `From<b256>` for `Bytes`
This will be reusable for converting `Address` & `ContractId` types to `Bytes` as well.

nfurfaro:
WIP.

Closes #3561


simonr0204 (approved):
Nice!  Thanks for this, will clean up my code quite a bit.

sway-lib-std/src/u256.sw
```diff
@@ -491,7 +491,11 @@ impl core::ops::Divide for U256 {
 
         assert(divisor != zero);
 
-        if self.a == 0 && self.b == 0 && divisor.a == 0 && divisor.b == 0 {
+        if self.a == 0
```
nfurfaro, line 494: dang. I guess this needed to be formatted but it doesn't belong in this pr,

otrho (commented):
Great!  Except for the name. :slightly_smiling_face: In Rust the '-able' trait names are avoided in favour of verbs or nouns, like `Debug`, `Clone`, `Display`.  I think we should try and do the same.

So I'd prefer this was the `Pack` trait in `pack.rs`.

mohammadfawaz:
> Great!  Except for the name. :slightly_smiling_face: In Rust the '-able' trait names are avoided in favour of verbs or nouns, like `Debug`, `Clone`, `Display`.  I think we should try and do the same.
> 
> 
> 
> So I'd prefer this was the `Pack` trait in `pack.rs`.

I also like `Pack` better.

otrho (dismissed):
Thanks.

JoshuaBatty (dismissed):

nfurfaro:
Marking as a draft temporarily as a new trait may not actually be needed. It was suggested on Slack that the existing `From` trait is probably sufficient.

mohammadfawaz:
[Someone asked on the forum](https://forum.fuel.network/t/how-can-i-transform-b256-into-u8-32/1124) for methods to convert a `b256` into an array of bytes so I suggested the type `Bytes` and a code similar to what you have here. Maybe we should start by implementing conversion between `b256` and `Bytes` instead of `ContractId` and `Bytes`? That way we can cover more ground and more use cases.

nfurfaro:
I think if we modify our `From` trait first to remove the `into` method, then this is a pretty easy task.

The reason I started with `ContractId` is that @simonr0204 needed that conversion and was performing it locally, so I thought to both use it as a test case and give him a library function he could use right away.

ref: https://github.com/FuelLabs/sway/issues/3637

nfurfaro:
closing as this is replaced by #3889

PR #3631, opened 2022-12-16
Demonstrating bugfix to trait constraints

### LINKED ISSUE #3325: Multiple trait constraints on the same function causing issues
Code:
```rust
script;

use core::ops::*;

impl<T> Option<T> {
        match self {
        }
    }
}

fn test_ok_or<T, E>(val: T, default: E) where T: Eq, E: Eq{
    match Option::Some(val).ok_or(default) {
        Result::Ok(inner) => assert(val == inner),
        Result::Err(_) => revert(0),
    }
}

fn main() {
    test_ok_or(true, 0);
}
```
The idea is that I want both `T` and `E` in `test_ok_or` to implement `Eq` so that I can add another test to compare errors. I currently get the following error:
```console
error: Internal compiler error: Verification failed: Type mismatch found for call to 'eq_3'.
```
The error goes away where I remove the trait constraint `E: Eq` on `test_ok_or`.

emilyaherbert:
I wasn't able to reproduce the error shown in #3325---I believe it to be resolved in the form shown in #3325. This PR adds the code from that issue to the test suite.

Closes #3325

### DIFF
--- a/test/src/e2e_vm_tests/test_programs/should_pass/language/where_clause_functions/src/main.sw
+++ b/test/src/e2e_vm_tests/test_programs/should_pass/language/where_clause_functions/src/main.sw
@@ -1,5 +1,7 @@
 script;
 
+use core::ops::*;
+
 use std::{
     assert::assert,
     logging::log,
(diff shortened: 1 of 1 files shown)


mohammadfawaz (approved):

otrho:
Merging before it gets out of date....

PR #3642, opened 2022-12-21
Extract FuelVM-specific instructions to umbrella instruction

### LINKED ISSUE #3071: Consider moving FuelVM specific IR operations to a `fuelvm` specifc wrapper.
There are some IR instructions like `gtf`, `get_storage_key`, `log`, `read_register`, `revert` and the `state_` operations which are specific to the Fuel VM.

They stick out a bit as being non-generic, and as we prepare to support other targets like the EVM it'd be good for them to be grouped together somehow.

A parent instruction like `fuelvm` could wrap them so logically we have a distinction between generic IR (for arithmetic, control flow and memory R/W) and these special ops.

arzg:
This PR fixes #3071. I’ve extracted the following IR instructions:

- `Instruction::GetStorageKey`
- `Instruction::Gtf`
- `Instruction::Log`
- `Instruction::ReadRegister`
- `Instruction::Revert`
- `Instruction::Smo`
- `Instruction::StateLoadQuadWord`
- `Instruction::StateLoadWord`
- `Instruction::StateStoreQuadWord`
- `Instruction::StateStoreWord`

Each of these instructions has been moved to a new `FuelVmInstruction` enum. The main `Instruction` enum now has a new `FuelVm` variant, which contains a single `FuelVmInstruction` instance.

I’ve named these `FuelVm` instead of `FuelVM` to [follow the Rust API Guidelines / RFC 430](https://rust-lang.github.io/api-guidelines/naming.html#c-case) – let me know if I should change this to `FuelVM` though :)

I’ve tried to maintain the existing alphabetical ordering of `Instruction` variants, so apologies for the confusing diff 😅

### DIFF
--- a/sway-core/src/asm_generation/asm_builder.rs
+++ b/sway-core/src/asm_generation/asm_builder.rs
@@ -183,22 +183,81 @@ impl<'ir> AsmBuilder<'ir> {
                 Instruction::ExtractValue {
                     aggregate, indices, ..
                 } => self.compile_extract_value(instr_val, aggregate, indices),
-                Instruction::GetStorageKey => {
-                    check!(
-                        self.compile_get_storage_key(instr_val),
+                Instruction::FuelVm(fuel_vm_instr) => match fuel_vm_instr {
+                    FuelVmInstruction::GetStorageKey => {
+                        check!(
+                            self.compile_get_storage_key(instr_val),
+                            return err(warnings, errors),
+                            warnings,
+                            errors
+                        )
+                    }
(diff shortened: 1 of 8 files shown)


otrho (approved):
Great, thanks.  This is a good first step towards isolating FuelVM specific operations.

Next we'll want to probably put these into their own modules, at least in the `sway-ir` crate in preparation for EVM support (which would also go in their own crate).  We may even want to use conditional compilation for back-end target features.  But that can happen when we actually introduce a new target.

mohammadfawaz (approved):
Great! Thank you so much for this.

PR #3644, opened 2022-12-21
Add `Delimiter` tokens to the parser

### LINKED ISSUE #3594: `function_decl.return_type.span` should be the span of the opening delimiter
For example:

```
abi Ownership {
    #[storage(read, write)]
    fn set_owner(owner: Option<Identity>);
}
```

In this case, `function_decl.return_type.span` is `fn set_owner(owner: Option<Identity>)`
and `function_decl.return_type` is a `TypeId` that resolves to `TypeInfo::Tuple([])`

The bug is that `return_type.span` is set to the full function signature. I would expect it to be an empty/dummy Span or `None` when there is no return type (though `return_type.span` is currently not optional).

### LINKED ISSUE #3634: `TyEnumVariant` `type_span` contains the entire span of the variant
I may be assuming incorrectly, but should this not be the span of only the type of the variant? If not there is already a `span` field which has the entire span.

Found while working on #3607 the expected result is `UnauthorizedUser: ()` but instead got `UnauthorizedUser: UnauthorizedUser: ()`

Source: https://github.com/FuelLabs/sway/blob/master/sway-core/src/language/ty/declaration/enum.rs#L110

### LINKED ISSUE #3635: `TyTraitFn` `return_type_span` contains the entire function signature
Similar problem to #3634 I may also be assuming incorrectly here as well, but shouldn't the `return_type_span` be only a span of the type the function returns? In the instance I found it, I assumed the `return_type_span` would give me `()` since it was a function without a return block but instead I got the entire function signature, `fn store_something() -> fn store_something()`.

Source: https://github.com/FuelLabs/sway/blob/master/sway-core/src/language/ty/declaration/trait_fn.rs#L15

eureka-cpu:
Main issue is #5140 
Adds `Delimiter` tokens, and updates their corresponding use cases.

Ref #3634 #3734 
Closes #3635 
Closes #3594


mohammadfawaz:
Looks good! But you have some E2E test failures that need to be addressed. Specifically, `test.toml` for:
```
    should_fail/trait_method_signature_mismatch ... failed
    should_fail/abi_method_signature_mismatch ... failed
```
should be updated.

mohammadfawaz:
Ok so I will write my thoughts on the failing test here:
for `trait_method_signature_mismatch`, we used to emit the following error:
image
Now, we just emit this (not pointing to any location):
image
We previously used to point to the function span itself because the return type is missing (i.e. () implicitly). Now that we’re using `Span::dummy()`, the error is different and doesn’t point to anything which is not good. I thought we didn’t need this span for anything but clearly we do.
So I looked at what Rust does… it seems to point to the span of the following character as in:
image
So I suggest we do the same instead of returning a dummy span

sdankel (dismissed):

eureka-cpu:
Going to spend some time addressing this after #4097 & #4095

sway-parse/src/item/mod.rs
```diff
@@ -144,20 +145,34 @@ impl Parse for FnArg {
...
+                FnReturnType::TypedReturn((right_arrow_token, ty))
+            }
+            None => {
+                where_clause_opt = parser.guarded_parse::<WhereToken, _>()?;
+                let Some(open_curly_brace) = parser.peek::<OpenCurlyBraceToken>();
+                FnReturnType::Implicit(ImplicitReturn {
+                    span: open_curly_brace.span(),
+                })
+            }
+        };
```
eureka-cpu, line 167: @mohammadfawaz (: making some progress here finally

eureka-cpu: need to add the error handling there, but just excited to finally finish this PR up

eureka-cpu:
@IGI-111 @sdankel Going to spend some time getting this back into shape this coming week

sway-ast/src/token.rs
```diff
@@ -67,45 +67,83 @@ impl PunctKind {
...
-            Delimiter::Brace => '}',
-            Delimiter::Bracket => ']',
+            Self::Parenthesis => ClosingDelimiter::Parenthesis,
+            Self::CurlyBrace => ClosingDelimiter::CurlyBrace,
+            Self::SquareBracket => ClosingDelimiter::SquareBracket,
+            Self::AngleBracket => ClosingDelimiter::AngleBracket,
+        }
+    }
+}
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
JoshuaBatty, line 130: This enum has the same variants as `OpeningDelimiter`. I feel like we should be able to keep the original `Delimiter` enum and make that work. Perhaps we could rename it to. 

```rust
pub enum DelimiterType {
    Parenthesis,
    CurlyBrace,
    SquareBracket,
    AngleBracket,
}
```

sway-ast/src/item/mod.rs
```diff
@@ -115,11 +115,22 @@ impl Spanned for FnSignature {
...
 }
+
+#[derive(Clone, Debug, Serialize)]
+pub enum FnReturnType {
+    Implicit(ImplicitReturn),
+    TypedReturn((RightArrowToken, Ty)),
+}
+
+#[derive(Clone, Debug, Serialize)]
+pub struct ImplicitReturn {
```
JoshuaBatty, line 134: Can we add a doc comment outining what this `Span` is pointing to when we have an implicit return type.

sway-ast/src/keywords.rs
```diff
@@ -232,3 +233,99 @@ define_token!(DoublePipeToken, "`||`", [Pipe, Pipe], [Pipe]);
...
+    /// Creates the keyword from the given `span`.
+    fn new(span: Span) -> Self;
+
+    /// Returns an identifier for this keyword.
+    fn ident(&self) -> Ident;
+
+    /// The sequence of punctuations that make up the token.
+    const DELIMITER_KIND: &'static [OpeningDelimiter];
+}
+macro_rules! define_opening_delimiter_token (
```
JoshuaBatty, line 247: Again, my intuition is that we don't need to have unique traits for closing or opening delimeters and we can find a sollution that is generic over either type. Is it possible to rename this to `define_delimiter_token` and make it work for both types?

eureka-cpu:
Putting this on hold while I work on https://github.com/FuelLabs/fuelup/issues/457 for the next beta release.

sdankel:
We should rethink this solution in light of  https://github.com/FuelLabs/sway/issues/4981

eureka-cpu:
> We should rethink this solution in light of #4981

In this instance, we don't want the span to be none since it would mean that we couldn't point to where the error actually happens

eureka-cpu:
Closing due to how big of a PR this became, and I no longer have time to try and finish it. There is some good stuff in here, for anyone who may decide to tackle the issue of adding delimiters to the parser and how that could affect the resulting AST, error handling and so on.

PR #3647, opened 2022-12-22
Adds interface for generating links between `items` & `modules`

### LINKED ISSUE #3553: Reduce the amount of `String` allocations in `forc doc`
Specifically, `render.rs` being the primary offender here

### LINKED ISSUE #3646: Craft path creation interface
Redesign `module_prefix` to fully support path creation for each item & parent module.

This is a large part of #3173

eureka-cpu:
A redesign of `module_prefix`. This PR introduces the `ModuleInfo` struct with methods for auto generating paths for use in the rendering process. This also reduces the amount of `String` allocations, but `horrorshow` requires that the lifetimes must outlive `'static` so some instances couldn't be avoided. 

Closes #3646 
Closes #3553


forc-plugins/forc-doc/src/doc.rs
```diff
@@ -119,11 +118,76 @@ impl Document {
...
+            .last()
+            .expect("There will always be at least the project name")
+    }
+    /// The location of the parent of the current module.
+    ///
+    /// To be used in path navigation between modules.
+    pub(crate) fn _parent(&mut self) -> &str {
+        self.0.pop();
+        self.location()
+    }
```
kayagokalp, line 145: I guess we are not using this function, maybe we can remove it?

```suggestion
```

kayagokalp (commented):
Overall looks good to me just left a nit. Reading refactor PRs feels oddly satisfying 😄

forc-plugins/forc-doc/src/doc.rs
eureka-cpu: So currently, all paths lead back to the `AllDoc` (`all.html`) path. But this isn't the end goal. How it _should_ be is that if you are on an item page the `parent` module should be the desired path. I'd like to keep this fn since my follow up PR will be tackling the aforementioned problem if that's ok!

eureka-cpu:
> Overall looks good to me just left a nit. Reading refactor PRs feels oddly satisfying smile

I've never agreed with something more. There's something refreshing about it haha

forc-plugins/forc-doc/src/doc.rs
JoshuaBatty: sounds reasonable enough to me.

JoshuaBatty (approved):

kayagokalp (approved):

forc-plugins/forc-doc/src/doc.rs
mitchmindtree: For future reference, we probably don't want this method to mutate `self` as it currently does, otherwise consecutive calls to `_parent` will produce different results. We might want to return an `Option` too, as if this is the root module, I guess there won't be a parent?

forc-plugins/forc-doc/src/render.rs
```diff
@@ -620,72 +603,57 @@ impl Renderable for AllDoc {
...
-
-    file_path.to_str().unwrap().to_string()
-}
-/// Create a path prefix string for navigation from the `module_depth`
-fn module_depth_to_path_prefix(module_depth: usize) -> String {
-    (1..module_depth).map(|_| "../").collect::<String>()
-}
 /// Creates an HTML String from an [AttributesMap]
-pub(crate) fn attrsmap_to_html_string(attributes: &AttributesMap) -> String {
+pub(crate) fn attrsmap_to_html_str(attributes: AttributesMap) -> String {
```
mitchmindtree, line 656: Normally we use the name `str` if talking about the `str` slice type. I think the original name ending with `_string` is probably better as it's clearer that we return a `String` and not a `str`.

mitchmindtree (commented):
Nice one @eureka-cpu, looking cleaner!

Just left some follow-up review notes as I was away when this landed - nothing major!

forc-plugins/forc-doc/src/doc.rs
```diff
@@ -16,26 +15,24 @@ pub(crate) type Documentation = Vec<Document>;
...
             StorageDeclaration(_) => None,
-            _ => Some(&self.item_header.item_name),
+            _ => Some(self.item_header.item_name.as_str()),
         };
 
-        Document::create_html_file_name(self.item_body.ty_decl.doc_name(), name.map(|s| &**s))
+        Document::create_html_file_name(self.item_body.ty_decl.doc_name(), name)
     }
-    fn create_html_file_name(ty: &str, name: Option<&str>) -> String {
+    fn create_html_file_name<'name>(ty: &'name str, name: Option<&'name str>) -> String {
```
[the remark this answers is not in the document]
eureka-cpu: Ah yeah, It was initially a slice but I must've changed it and forgotten since it didn't give me a clippy warning. Thanks!

forc-plugins/forc-doc/src/render.rs
eureka-cpu: Ah, another part where I had changed the return type to slice when i was refactoring but forgot to change it back.

PR #3648, opened 2022-12-22
Add balance tree and output message effects to CEI analysis

### LINKED ISSUE #3395: Add new kind of effects (balance tree modification) to CEI analysis
The following description is from an internal discussion.

It looks like there are 3 more opcodes we have to think about for the CEI pattern analysis: `tro`, `tr`, and `smo`. These opcodes are considered “effects”. We need to document, in the Sway Book, what operations are considered effects and what the warning looks like etc.

A few notes:
- `tr` and `tro` are are not exactly storage accesses. They do update the “state” but not the general purpose state tree. They update the *balance tree*. They should have their own type of warning. Same with `bal` which only reads from the balance tree.
- `smo` might also needs its own warning because it doesn’t update any trees but may have serious effects if the user sends arbitrary bytes that “do something”. While `smo` definitely has effects, it looks like they would be write only. It doesn’t write to contract storage though; just creates a new `Output::Message` which could have implications on Ethereum if a system has been implemented to receive these messages.

anton-trunov:
close #3395

Summary:
- Process `smo` intrinsic
- Process `smo` asm instruction
- Process `tr` and `tro` instructions
- Process `bal` instruction
- Refactor warnings to incorporate new effects and stop duplicating warning constructors for each type of effect
- Refactor warning reporting for while loops
- Tests for detecting CEI pattern violations in:
  - asm block with `bal` instruction
  - asm block with `smo` instruction
  - asm block with `tr` instruction, indirectly via `std::token::force_transfer_to_contract`
  - asm block with `tro` instruction, indirectly via `std::token::transfer_to_address`
  - `smo` intrinsic, indirectly via `std::message::send_typed_message`
- Document the new effects in the Sway book


test/src/e2e_vm_tests/test_programs/should_pass/static_analysis/cei_pattern_violation_in_asm_block_smo/test.toml
```diff
@@ -0,0 +1,3 @@
+category = "compile"
+
+# check: $()Output messages modification after external contract interaction in function or method "deposit". Consider making all output messages modifications before calling another contract
```
mohammadfawaz, line 3: "Output messages modification" doesn't seem to capture the real problem with the `smo` opcode used after an interaction. I'd like to hear what @simonr0204 @nfurfaro @adlerjohn would have to say about the exact wording.

mohammadfawaz (commented):
This looks great! Thanks @anton-trunov. Just a small comment on the wording for the `smo` effect. We may need help from some app devs to craft an accurate warning.

test/src/e2e_vm_tests/test_programs/should_pass/static_analysis/cei_pattern_violation_in_asm_block_smo/test.toml
anton-trunov: Thanks @mohammadfawaz! Just in case, the test case with the `__smo` intrinsic is here: https://github.com/FuelLabs/sway/blob/1d59ec97466d884dae1fb543e32d34576f68f9a6/test/src/e2e_vm_tests/test_programs/should_pass/static_analysis/cei_pattern_violation_smo_intrinsic/src/main.sw. The standard library `send_typed_message` function calls the intrinsic inside its `while`-loop.

nfurfaro: It seems like the main issue with `smo` is that for something like a bridge, if there's a re-entrancy vulnerability in a function / chain of functions that sends a message allowing withdrawal of tokens from fuel, multiple messages could possibly be sent when there is supposed to be only a single message, allowing withdrawal of more than a users balance of tokens.
Perhaps changing the wording of "Output messages modification" to something like:
"Output Message sent after external contract interaction ..."
"Consider sending all output messages before calling another contract."

anton-trunov: @nfurfaro Indeed this seems much nicer, thank you! Done

mohammadfawaz (approved):
Looks good! Thanks.

otrho (approved):
:+1:

PR #3663, opened 2022-12-31
Remove pointers from the Sway IR.

otrho:
## Argh, this is hard

This is the result of several attempts (made over several weeks) to attack #2819 and is only the very first stage.

- Initially @vaivaswatha introduced a PR which removed `Aggregate` and `extract_value`/`insert_value`, replacing them with proper pointer types and GEP.  This was great, but was too hard to make work with the ASM generation.  There are too many assumptions around reference types vs copy types and what is a 'pointer' or not for a change like this to go in _yet_.
- I then tried paring it back to only introduce proper pointer types, leaving GEP for later, but still making it work with ASM gen (especially contract calls and storage operations) was too hard.

## The problem

`get_ptr` was a bit nasty, as we use it to get the address of locals, but also to cast them to different types and also to index, a bit like GEP will. 

Also, using `Pointer` for `ref mut` function args was a mistake as they were designed only for local variables.

## Solution

This PR does as little as possible.  **The baby steps approach is absolutely required for this task** and this is the first one.

To remove any ambiguity around how pointers should work this PR removes them altogether from IR, as they were actually not strictly doing anything.  In particular it removes `Pointer` from `Type` and `get_ptr` is now `get_local` which is actually what 'pointers' were.

We will re-introduce pointers as actual types in an upcoming PR.

So now we have `get_local` which does nothing more than get a local variable address.  Casting and indexing has been shunted to a new `cast_ptr` instruction as a stopgap.  When we introduce GEP we can remove `cast_ptr`, or at least formalise it into a better/general cast.

This PR also uses the newer quad storage operations which support multiple slots, so the `get_ptr` -style indexing isn't actually required AFAICT.

## Future stages

- Next we can convert `Type` into an arena and remove `Aggregate`.  This is another ambiguity around 'reference types' which needs to be removed.  **The IR should not have any concept of copy types or reference types.** This is a Sway thing and isn't portable to different back-ends.
- Then we can introduce `Pointer` as an actual `Type`.  This will let us be explicit and strict about memory accesses and genuinely remove the notion of 'reference types'.  We should consider opaque pointers as an alternative to pointer casting, which is required for dealing with storage slots and arrays of `b256`s.
- Then we replace `extract_value` and `insert_value` with GEP.
- Also work out how casting will work.


vaivaswatha (approved):
The change looks good to me. I didn't understand the change to `fn compile_union_or_string_storage_(read/write)` though. Hopefully the 2nd reviewer can check that.

sway-core/src/ir_generation/function.rs
```diff
@@ -2774,59 +2720,44 @@ impl<'eng> FnCompiler<'eng> {
...
+                .ins(context)
+                .cast_ptr(value_val_b256, Type::B256, 0)
                 .add_metadatum(context, span_md_idx);
 
-            let one_value = convert_literal_to_value(context, &Literal::U64(1));
+            let count_value = convert_literal_to_value(context, &Literal::U64(number_of_elements));
             self.current_block
                 .ins(context)
-                .state_load_quad_word(value_ptr_val_b256, *key_ptr_val, one_value)
+                .state_load_quad_word(indexed_value_val_b256, *key_val, count_value)
```
mohammadfawaz, line 2756: Great! Thanks.

sway-core/src/ir_generation/function.rs
```diff
@@ -2774,59 +2720,44 @@ impl<'eng> FnCompiler<'eng> {
...
-                    .add_metadatum(context, span_md_idx);
-            }
-
+        if number_of_elements > 0 {
             // Get the b256 from the array at index iter
-            let value_ptr_val_b256 = self
+            let value_val_b256 = self
                 .current_block
                 .ins(context)
-                .get_ptr(value_ptr, Type::B256, array_index)
```
mohammadfawaz, line 2820: Yeah I remember one of the main reasons for adding `get_ptr` was this particular use case.

mohammadfawaz (approved):
Awesome. The direction of this seems very reasonable to me.

PR #3668, opened 2023-01-01
Small fixes for incorrect rebasing in 3411

JoshuaBatty:
Small fixes to a few things that got reverted in the language server while merging #3411 into master.

Specifically, versions were downgraded from 0.32.2 to 0.32.1. And the changes introduced in #3620 were omitted.

### DIFF
--- a/sway-lsp/Cargo.toml
+++ b/sway-lsp/Cargo.toml
@@ -11,19 +11,19 @@ description = "LSP server for Sway."
 [dependencies]
 anyhow = "1.0.41"
 dashmap = "5.4"
-forc-pkg = { version = "0.32.1", path = "../forc-pkg" }
-forc-tracing = { version = "0.32.1", path = "../forc-tracing" }
+forc-pkg = { version = "0.32.2", path = "../forc-pkg" }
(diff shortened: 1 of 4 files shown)


otrho (dismissed):
Why wasn't this picked up by CI?

JoshuaBatty:
> Why wasn't this picked up by CI?

CI is just checking that the version of `sway-lsp` is the correct version which it still was, just the dependencies that it uses were bumped down.

otrho (approved):

otrho:
> > Why wasn't this picked up by CI?
> 
> CI is just checking that the version of `sway-lsp` is the correct version which it still was, just the dependencies that it uses were bumped down.

But those new errors from #3620  should have tests or something which would fail?  Is testing some of this stuff outside of an editor environment too hard?

JoshuaBatty:
The 2 errors that were introduced `UnableToCreateFile` and `UnableToWriteFile` are just mapping errors from rusts `File::create` and `writeln!` so they are compatible with our custom `LanguageServerError` type. I think it's probably fine to not need specific tests for these error types as an error will be reported in the `did_chage` test if these fail for any reason.

mohammadfawaz (approved):

Issue #3622, opened 2022-12-16
Add a way to omit tests at parsing step


kayagokalp:
Currently tests are parsed and type checked even if we have tests disabled. This is both unnecessary and also blocks me on #3262.

mitchmindtree:
> This is both unnecessary and also blocks me on https://github.com/FuelLabs/sway/issues/3262.

Just to clarify: Currently, when tests are not enabled, we still include them during type checking, but omit them from IR. This is intentional, to ensure that the tests functions are available to DCA to avoid producing unnecessary warnings about dead code in the case that some items are only used within the test functions.

That said, as we need the contract's ID in order to build the contract's unit tests (seeing as we wish to call into that contract), we cannot type-check unit tests on the first build that produces the contract's ID. As discussed on slack, adding a flag to sway-core's BuildConfig like `omit_tests_from_parsed` or similar could be the way to go here, but we'll need to silence the aforementioned unnecessary dead code warnings on this contract ID build.

mohammadfawaz:
> Just to clarify: Currently, when tests are not enabled, we still include them during type checking, but omit them from IR. This is intentional, to ensure that the tests functions are available to DCA to avoid producing unnecessary warnings about dead code in the case that some items are only used within the test functions.

[This](https://github.com/FuelLabs/sway/issues/3586) is somewhat related. There are some DCA warnings are get generated today from just running `forc build` even though they have nothing to do with the actual non-test code being built. Is this expected behavior? Shouldn't we not care about the tests unless we're running `forc test`? To your point, it's important that types that are only used in tests do not cause a DCA warning but also DCA warnings that are generated from _inside_ tests (such as unused return value) should not be emitted when running `forc build`.

mitchmindtree:
Ahhhh originally, I thought Rust was also type-checking tests (regardless of `#[test]`), but after double checking Rust's behaviour it seems they only *parse* the tests unless `cargo build --tests` or `cargo test` is run, which makes sense in retrospect!

@mohammadfawaz I think you're right in that we should likely stop including tests during type-checking and DCA in general (unless running `forc test` or `forc build --tests`).Perhaps we _can_ make this a more general change to the way tests are included in compilation.

One nicety that we're missing that Rust provides in aid of testing and avoiding lots of dead-code warnings is conditional compilation, i.e. `#[cfg(test)]`. E.g. a user can decorate an item or module with `#[cfg(test)]` and know that it won't interfere with the regular, non-test compilation.

[...]

Issue #3444, opened 2022-11-28
Add `forc contract-id` and `forc predicate-id` commands


mitchmindtree:
These would be super simple commands on top of `forc build` that produce the associated contract/predicate ID.

An optional `--salt` flag can be provided for each.

Each should check that the program type is valid for the given command and produce a friendly error otherwise.

Configuration time constants ala https://github.com/FuelLabs/sway-rfcs/pull/19 would make the `forc contract-id` command trickier as we would need to specify values for each of the constants (or fall back to defaults) to produce a useful ID.

mohammadfawaz:
I think the terminology is
- Contract ID
- Script bytecode hash (not too important)
- Predicate root

mitchmindtree:
Thanks for clarifying @mohammadfawaz!

Perhaps `forc predicate-root` makes more sense here.

Then again, I wonder if there's a good case for the differing terminology here? I've seen ID, address, root, bytecode hash. I wonder if there's a chance to improve on naming consistency here, or if these are necessarily different for each of the program types?

PR #3672, opened 2023-01-02
feat: Add support for calling into contract ABIs from unit tests

### LINKED ISSUE #3262: Add support for calling into contract ABIs from unit tests
When unit tests are written within a contract module, it would be great if the user could run the test assuming the contract is "deployed" and can be interacted with.

We could either make this the default behaviour for contract unit tests, or we could add a special nested attribute like `#[test(script)]` or `#[test(with_contract)]` or similar.

We could provide the contract's ID in a similar manner to how we provide contract dependency IDs.

This will require a slightly more sophisticated testing setup in `forc-test`. Currently, we construct our own `fuel-vm` `Interpreter` and execute unit tests directly as if they were a script.

Calling into contracts will instead require first *deploying* the contract and then interacting with the contract from a script via a client.

This is an updated version of the #1832 issue, but specifically for supporting calling into contracts following landing #2985.

kayagokalp:
closes #3262.

# About this PR

For unit tests in contracts, we are now deploying the contract before executing the tests so that we can support calling into contract ABI. 

With this PR testing a contract looks like:

1. Compile the contract with tests
2. Compile the contract without the tests
3. Deploy the contract (the one that we compiled without the tests)
4. Run each test as a script

Currently there is no way to obtain the contract id without deploying the contract (related #3444) and automatic `CONTRACT_ID` injection is not available as we cannot inject it until we solve #3622.

I am planning to tackle #3622  and add the documentation after that as this is not directly usable by our users until we have `CONTRACT_ID` injection or #3444.

## TODO

- [ ] Add tests


JoshuaBatty (approved):
Nice work!

docs/book/src/testing/unit-testing.md
```diff
@@ -53,7 +53,44 @@ Tests with `#[test(should_revert)]` considered to be passing if they are reverti
 
 ## Calling Contracts
 
-***Coming Soon***
+Unit tests can call contract functions an example for such calls can be seen below.
```
bingcicle, line 56:
```suggestion
Unit tests can call contract functions, like so:
```

forc-test/src/lib.rs
```diff
@@ -151,16 +343,61 @@ impl BuiltTests {
...
+        contract_id
+    );
+
+    // Setup the interpreter for deployment.
+    let params = tx::ConsensusParameters::default();
+    let storage = vm::storage::MemoryStorage::default();
+    let mut interpreter = vm::interpreter::Interpreter::with_storage(storage, params);
+
+    // Create the deployment transaction.
+    let mut rng = rand::rngs::StdRng::seed_from_u64(0x7E57u64);
```
bingcicle, line 376: This u64 seed is being used in two places, maybe we want a const for this?

bingcicle (approved):
some small nits, but overall LGTM!

docs/book/src/testing/unit-testing.md
kayagokalp: Nice suggestions thanks a lot! I will have a follow-up PR anyways (for injecting `CONTRACT_ID` automatically) will mark these resolved for now and fix them in the follow-up not to dismiss josh's review.

forc-test/src/lib.rs
```diff
@@ -80,8 +102,163 @@ pub struct Opts {
...
+            secret_key: rng.gen(),
+            utxo_id: rng.gen(),
+            amount: 1,
+            asset_id: rng.gen(),
+            tx_pointer: rng.gen(),
+            maturity: 1,
+            block_height: (u32::MAX >> 1) as u64,
+        }
+    }
+}
```
mitchmindtree, line 258: Rather than implementing `Distribution` here, we should probably use a dedicated constructor function. We don't randomise *all* fields of the `TxMetadata` here, I think implementing `Distribution` is a little dubious.

forc-test/src/lib.rs
```diff
@@ -95,6 +272,15 @@ impl Opts {
...
     }
+
+    /// Patch this set of test options, so that it will build the package at the given `path`.
+    pub(crate) fn patch_opts(self, path: &std::path::Path) -> Opts {
+        let mut opts = self;
+        let mut pkg_opts = opts.pkg;
+        pkg_opts.path = path.to_str().map(|path_str| path_str.to_string());
+        opts.pkg = pkg_opts;
+        opts
+    }
```
mitchmindtree, line 283: If this method took `mut self`, this whole implementation could be:

```rust
        self.pkg.path = path.to_str().map(ToString::to_string);
        self
```

forc-pkg/src/pkg.rs
```diff
@@ -434,6 +435,14 @@ impl Built {
             Built::Workspace(built_workspace) => Ok(built_workspace),
         }
     }
+
+    /// Tries to retrieve the `Built` as a `BuiltPackage`, panics otherwise.
```
mitchmindtree, line 439: This doc comment says the method panics, but it returns an `Err`

mitchmindtree (commented):
Great stuff @kayagokalp!

Just thought I'd leave a post-review for some follow-up seeing as I was away when this landed.

Originally I thought deploying the contracts would be a little more complicated, but I can't yet think of why we shouldn't just insert them directly into the interpreter with dedicated create transactions. as you've done here. Good stuff!

> Currently there is no way to obtain the contract id without deploying the contract (related https://github.com/FuelLabs/sway/issues/3444) and automatic CONTRACT_ID injection is not available

As mentioned in [this comment](https://github.com/FuelLabs/sway/issues/3262#issuecomment-1305204912), we will likely want to build the contract *without* tests first. This will be necessary in order to be able to determine the contract ID so that it may be inserted into the namespace that is used when compiling the contract *with* tests. This might require a bit of a refactor of the approach taken in this PR (which compiles the contract *with* tests first) but it shouldn't be too major.

forc-test/src/lib.rs
```diff
@@ -151,16 +343,61 @@ impl BuiltTests {
...
+            metadata.asset_id,
+            metadata.tx_pointer,
+            metadata.maturity,
+        )
+        .add_output(tx::Output::contract_created(contract_id, state_root))
+        .maturity(metadata.maturity)
+        .finalize_checked(metadata.block_height, &params);
+
+    // Deploy the contract.
+    interpreter.transact(tx)?;
```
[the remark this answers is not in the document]
kayagokalp: My initial approach was like that but interpreter instance does not let transacting 2 different types of transaction (`crate` and `script`) to be transact with the same instance. Because the instance is taking the type of the transaction it is runnnig. (https://github.com/FuelLabs/fuel-low-level/blob/bca7c506cdb1168266626b019ad95535a264292a/fuel-vm/src/transactor.rs#L149 might explain thing better. `fn transact()` creates the interpreter instance with the transactions type.) Not sure why that is implemented that way but to walk around that I had to create storages with contract deployed for each tests that requires them.

kayagokalp:
Thanks for the comments @mitchmindtree. Welcome back! I was under the impression that blocker I had in the compiler (#3622) for automatic injection of `CONTRACT_ID` (#3673) would take longer because of that I wanted to get this in so that it is somehow possible to use contract calls in the tests. But #3684 turned out to be smaller than I thought at the beginning and I am now preparing the refactor you discussed in your comment for #3673.

PR #3675, opened 2023-01-03
Return the lexing stage produced by `sway_parse` from `forc-pkg::check()`

### LINKED ISSUE #3524: Seperate the `sway_parse` and `ParseProgram` generation in the compiler.
Calling `forc-pkg::check()` currently returns a tuple containing a `ParseProgram` and `Option<ty::TyProgram>`.

However, it would also be beneficial if we could return the initial lexing stage produced by `sway_parse`, before it is converted into a `ParseProgram`. Currently, this is all happening in one step in `sway_core::parse()`. I'm proposing that we separate this into 2 functions so `forc-pkg::check()` can also return the `Module` kind produced by `sway_parse::parse_file` so it can be used upstream in LSP and the formatter.

JoshuaBatty:
Previous to this PR, `forc_pkg::check()` would only return the `ParseProgram` and `TyProgram` AST's. In the language server, we now also need access to the initial lexing stage produced by `sway_parse` in order to get access to the sway keyword tokens. 

This PR now returns the lexing stage of a program along with the parsed and typed stages. 

closes #3524

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -31,17 +31,14 @@ use sway_core::{
         fuel_crypto,
         fuel_tx::{self, Contract, ContractId, StorageSlot},
     },
-    language::lexed,
-    Engines, TypeEngine,
-};
-use sway_core::{
     language::{
+        lexed::LexedProgram,
         parsed::{ParseProgram, TreeType},
         ty,
     },
     semantic_analysis::namespace,
     source_map::SourceMap,
-    CompileResult, CompiledBytecode, FinalizedEntry,
+    CompileResult, CompiledBytecode, Engines, FinalizedEntry, TypeEngine,
 };
 use sway_error::error::CompileError;
 use sway_types::Ident;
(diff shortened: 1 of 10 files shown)


forc-pkg/src/pkg.rs
```diff
@@ -2799,8 +2797,26 @@ fn update_json_type_declaration(
...
+        Programs {
+            lexed,
+            parsed,
+            typed,
+        }
+    }
+}
 
 /// Compile the entire forc package and return the parse and typed programs
 /// of the dependancies and project.
```
kayagokalp, line 2822: Should we also update this doc comment 👀

```suggestion
/// Compile the entire forc package and return the lexed, parsed and typed programs
/// of the dependancies and project.
```

kayagokalp (approved):
🚢

forc-pkg/src/pkg.rs
JoshuaBatty: Good catch, i'll up date that comment now

kayagokalp (approved):

sdankel (approved):

PR #3676, opened 2023-01-03
Fixes issue with generic trait impls using the same method names.

### LINKED ISSUE #3633: Bug with duplicate impl blocks defined for the same type with different type parameters.
The following code won't compile. The error is:
image

```
library bug;

pub trait Convert<T> {
    fn convert(t: T) -> Self;
}

pub struct A {
    a: u64,
}

pub struct B {
    b: u64,
}

pub struct C {
    c: u64,
}

impl Convert<B> for A {
    fn convert(t: B) -> Self {
        A {
            a: t.b
        }
    }
}

impl Convert<C> for A {
    fn convert(t: C) -> Self {
        A {
            a: t.c
        }
    }
}
```

esdrubal:
When two generic trait implementation use the same method name with a different type parameter an error such as `Duplicate definition for method "convert" for type "A"` would occur.

This fix does remove the error when a method with diferent type parameters is added to another trait implementation with different type parameters.
The fix also ensures that when possible find_method_for_type returns a method whose parameters types matches the given arguments expressions return types.

Closes #3633.


emilyaherbert:
Upon further investigation, #3633 isn't a bug after all, but instead a matter of showing the user a bad error message. Apologies for adding this additional context after have submitted this PR 😭 

I'm going to close this PR and remove your assignment from #3633 for bookkeeping, but please feel free to re-add your assignment to #3633 if you'd like to continue working on that GH issue.

emilyaherbert:
Hi @esdrubal, apologies for the confusion, I misread the example in #3633---this PR was closed in error. I'll reopen this PR and add you back to that issue to correct those changes.

sway-core/src/semantic_analysis/namespace/trait_map.rs
```diff
@@ -191,8 +191,14 @@ impl TraitMap {
                     second_impl_span: impl_span.clone(),
                 });
             } else if types_are_subset {
```
emilyaherbert, line 193: I think that your changes in this file might not be necessary. I'm wondering if changing this conditional check to `else if types_are_subset && is_impl_subset` would be sufficient?

emilyaherbert (changes requested):

sway-core/src/semantic_analysis/namespace/trait_map.rs
esdrubal: There is currently no `is_impl_subset` defined in the file. Could you please elaborate on this?

emilyaherbert: Sorry I meant `is_impl_self`.

sway-core/src/semantic_analysis/namespace/trait_map.rs
```diff
@@ -792,7 +823,11 @@ impl TraitMap {
     }
 }
 
-fn are_equal_minus_dynamic_types(type_engine: &TypeEngine, left: TypeId, right: TypeId) -> bool {
+pub fn are_equal_minus_dynamic_types(
```
emilyaherbert, line 826: Please change this to `pub(crate)` if this function is not being use anywhere outside of this crate.

sway-core/src/semantic_analysis/namespace/namespace.rs
```diff
@@ -220,14 +225,62 @@ impl Namespace {
...
             }
         }
 
+        let matching_method_decl_id = match matching_method_decl_ids.len().cmp(&1) {
+            Ordering::Equal => Some(matching_method_decl_ids[0].clone()),
+            Ordering::Greater => {
+                // Case where multiple methods exist with the same name
+                // This is the case of https://github.com/FuelLabs/sway/issues/3633
+                // where multiple generic trait impls use the same method name but with different parameter types
+                let mut matching_method_decl_ids2: Vec<DeclarationId> = vec![];
```
emilyaherbert, line 238: This doesn't need to be a list---could you refactor this to an `Option<DeclarationId>`?

This can be refactored because the only functions that this variable is used for is to check that is (1) has been set and (2) get the first element from the list. What you could do instead is to use an `Option<DeclarationId>` called `maybe_method_decl_id`. When `maybe_method_decl_id` is `None`, then we know that we haven't found an end condition yet, but when we find that end condition (i.e. the large conditional in the `if` expression below) we can set this to `Some(decl_id)`, when that happens we can break out of the loop and avoid wasted compute.

emilyaherbert (changes requested):
This PR looks very close! Just a few comments left to address.

sway-core/src/semantic_analysis/namespace/trait_map.rs
esdrubal: I tried to do that, but then the case below no longer throws an error.

https://github.com/FuelLabs/sway/blob/a930f01755c67c4b189090e2ecb922571856ffd1/test/src/e2e_vm_tests/test_programs/should_fail/multiple_methods_with_the_same_name/src/main.sw#L86-L90

sway-core/src/semantic_analysis/namespace/namespace.rs
esdrubal: Good point. Applied the changes suggested.

sway-core/src/semantic_analysis/namespace/trait_map.rs
emilyaherbert: Ah I see what you're saying. In that case I have some remaining comments about the additions here, but I think your overall approach is good.

sway-core/src/semantic_analysis/namespace/trait_map.rs
```diff
@@ -191,8 +191,14 @@ impl TraitMap {
                     second_impl_span: impl_span.clone(),
                 });
             } else if types_are_subset {
+                let type_args_are_equal = trait_type_args.len() == map_trait_type_args.len()
```
emilyaherbert, line 194: This check should be encapsulated by the `types_are_subset` check, so you should need to add this specifically. If its not, then there may be a bug with that check.

emilyaherbert (changes requested):

sway-core/src/semantic_analysis/namespace/trait_map.rs
esdrubal: This check is encapsulated by the `traits_are_subset` but not by `types_are_subset`.

The most recent commit replaces `type_args_are_equal` usage by `!traits_are_subset && !is_impl_self`.

nfurfaro (dismissed):
Approved based on the testing added to demonstrate the effectiveness of the fix, but needs a compiler team member for final review.

sway-core/src/semantic_analysis/namespace/trait_map.rs
emilyaherbert: Looking at all of our review comments, I'm going to dig deeper on this still---I still feel that there is a better way to write this code block. For example here we have the `else if types_are_subset` expression, and inside it contains a `if !traits_are_subset && !is_impl_self` expression. I believe that the if expression relation works out such that we can move the `!traits_are_subset && !is_impl_self` check to accompany the `else if types_are_subset` check.

However, this is the same spot that we were in before, when you reported in your previous comment that this still threw a "duplicate definition" error. What this is telling me is that the check of `method.parameters` against `map_trait_method.parameters` is doing quite a bit of work in this algorithm to determine if this is a duplicated definition. I see that this check is using the `are_equal_minus_dynamic_types` algorithm which [compares generic types by directly comparing the names and the trait constraints](https://github.com/FuelLabs/sway/blob/dd9578008b989a6000ab659d34657d75fc859a8e/sway-core/src/semantic_analysis/namespace/trait_map.rs#L809). On the other hand, the check to `traits_are_subset` is using the `TypeInfo::is_subset_of` algorithm, which only [compares the trait constraints and not the names](https://github.com/FuelLabs/sway/blob/dd9578008b989a6000ab659d34657d75fc859a8e/sway-core/src/type_system/info.rs#L1432). What this is telling me is that a better fix to this problem would be to create an alternative to `traits_are_subset` (called `traits_are_subset_2` or something) that instead uses a different algorithm [(I suggest the updated unify_check algorithm) to compare both the names and the trait constraints](https://github.com/FuelLabs/sway/blob/dd9578008b989a6000ab659d34657d75fc859a8e/sway-core/src/type_system/unify_check.rs#L119). Then much of the changes in the part of the PR can be streamlined to just a check to `else if types_are_subset && !traits_are_subset_2 && !is_impl_self`.

The reason that I'm being so picky😊 about this PR is to avoid antipatterns and to emphasize separation of concerns. If the compiler is performing static analysis to determine if the user has written conflicting `impl`s for traits and types, then that should not rely on the method information. e.g. what if the methods of the conflicting impl block themselves have type errors? Or the user forgot to provide methods? In this case checking the types of the method parameters would lead to a false report.

emilyaherbert (changes requested):

emilyaherbert:
(I wrote a review comment but it's showing up under the fold of a previous resolved comment. CC @esdrubal)

sway-core/src/semantic_analysis/namespace/trait_map.rs
esdrubal: > For example here we have the `else if types_are_subset` expression, and inside it contains a `if !traits_are_subset && !is_impl_self` expression. I believe that the if expression relation works out such that we can move the `!traits_are_subset && !is_impl_self` check to accompany the `else if types_are_subset` check.

We do `if !traits_are_subset && !is_impl_self` inside a loop that compares methods.
The `for` block needs to skip methods that are different on a method by method basis so the error `CompileError::DuplicateMethodsDefinedForType` is only printed for methods that are effectively duplicates.

> Then much of the changes in the part of the PR can be streamlined to just a check to `else if types_are_subset && !traits_are_subset_2 && !is_impl_self`.

Not sure how we can refactor this to be outside the `for` loop. Comparing just names inside the `for` loop is not sufficient. We still need somehow to check in place if the methods signatures are different and just print the error when they are equal.

emilyaherbert: > We do if !traits_are_subset && !is_impl_self inside a loop that compares methods.
The for block needs to skip methods that are different on a method by method basis so the error CompileError::DuplicateMethodsDefinedForType is only printed for methods that are effectively duplicates.

The values of `traits_are_subset` and `is_impl_self` do not change between iterations of the loop, so if any one method fails the check to ` !traits_are_subset && !is_impl_self && (method.parameters.len() != ...` then all methods will fail that check. Thus we don't need to enter the for loop at all if `traits_are_subset || is_impl_self`.

> Not sure how we can refactor this to be outside the for loop. Comparing just names inside the for loop is not sufficient. We still need somehow to check in place if the methods signatures are different and just print the error when they are equal.

Backing up from this code block. The bug in #3633 is misleading. It's not the case that this bug should be solved by determining if the methods in `Convert<B>` and `Convert<C>` contain any duplicate method definitions. It should be solved by determining if the traits `Convert<B>` and `Convert<C>` are conflicting and if therefore the implementations are conflicting. The fact that the compiler is _entering the code path of LOC 193 at all is the bug_.

Look at this sample Rust code that shows conflicting impl blocks. It's not the case that the error message is displaying that the methods are conflicting but instead that the implementations of those conflicting traits are conflicting: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8dcc799aa2c3bb2355134905b0757f5e

emilyaherbert (changes requested):

sway-core/src/semantic_analysis/namespace/trait_map.rs
esdrubal: @emilyaherbert Thank you for the review, I was trying to fit in a wrong use case and your comment made me understand it.
I fixed the LOC 193 to your suggestion.

emilyaherbert (approved):
This looks really good @esdrubal 🎉 I appreciate your persistence with this PR through the re-review process.

mohammadfawaz (approved):

PR #3682, opened 2023-01-03
Introduce a `context` to `convert_parse_tree`

### LINKED ISSUE #1839: Rethink the use of atomic variables in #1542 and #1829
https://github.com/FuelLabs/sway/pull/1829 and https://github.com/FuelLabs/sway/pull/1542 introduced unique compiler-generated names for certain variable declarations that rely on an atomic variable. This is fine for now because the compiler is single-threaded. If we decide to switch to a multi-threaded compiler one day, this may cause issues because we cannot guarantee that those generated names will be stable, which can lead to nondeterministic bytecode.

### LINKED ISSUE #3678: Disallow multiple `configurable` blocks in a given project (and introduce a `context` to `convert_parse_tree.rs`)
Configuration-time constants are being added in https://github.com/FuelLabs/sway/pull/3536. We want to disallow multiple `configurable` blocks in a given project. This means disallow `configurable` in the top level module as well as in any dependent libraries (which is already disallowed in https://github.com/FuelLabs/sway/pull/3536)

Due to the implementation in https://github.com/FuelLabs/sway/pull/3536, this requires a bit of work in `convert_parse_tree`. Basically, we need the concept of a `context` which will keep track of whether a `configurable` block has been found already. The `context` can also be used to get rid of the atomic variables in that same file.

mohammadfawaz:
Closes #3678
Closes #1839

- Add a `Context` to `convert_parse_tree` which is useful to keep track of global information such as unique names for destructured tuples or to count how many configurable blocks there are in a module. Added the `context` everywhere even when it's not really used, in preparation for any potential future uses of it.
- Use the `context` to remove uses of atomics
- Use the `context` to detect when a `configurable` block is used multiple times in a given module
- Add tests for the two cases above.


sway-core/src/transform/to_parsed_lang/context.rs
```diff
@@ -0,0 +1,66 @@
+use core::cell::RefCell;
+
+/// A context containing global information used during `convert_parse_tree"
+#[derive(Default)]
+pub struct Context {
+    /// The inner context.
+    /// This construction is used to avoid `&mut` all over `convert_parse_tree`.
+    inner: RefCell<ContextInner>,
+}
```
otrho, line 9: Using `RefCell` gives me the heebie-jeebies as they can accidentally produce _runtime_ borrow conflicts.  It looks below like it's used pretty safely, but I'm not sure I follow the justification of avoiding passing `&mut`?  What's wrong with a mutable context?

otrho (commented):
Question below.  Plus it seems there's some `configurable` tests which have been borrowed from your previous PR?  Probably doesn't hurt if there's no conflict or whatever.

sway-core/src/transform/to_parsed_lang/context.rs
mitchmindtree: Perhaps we could also keep a `&Handler` in the `Context` for the traversal? Could always be done in a follow-up PR though too.

mohammadfawaz:
> Question below. Plus it seems there's some `configurable` tests which have been borrowed from your previous PR? Probably doesn't hurt if there's no conflict or whatever.

Hmm there is only one new test in this PR called `multiple_config_blocks`. That wasn't in my previous PR I believe?

sway-core/src/transform/to_parsed_lang/context.rs
mohammadfawaz: To be entirely honest, `Context` here is modelled after the [`Handler`](https://github.com/FuelLabs/sway/blob/master/sway-error/src/handler.rs) which uses `RefCell`. I don't really mind one way over another.. just went with `RefCell` for consistency. Thoughts? 

@mitchmindtree moving `Handler` into `Context` in this file seems like a good idea!

otrho:
> > Question below. Plus it seems there's some `configurable` tests which have been borrowed from your previous PR? Probably doesn't hurt if there's no conflict or whatever.
> 
> Hmm there is only one new test in this PR called `multiple_config_blocks`. That wasn't in my previous PR I believe?

Oh, I might've jumped to the wrong conclusion.  I saw that a) this PR used to be branched off another PR/branch, b) multiple config blocks didn't seem to  be related to adding a context to conversion and c) your last PR was the configurable stuff.

sway-core/src/transform/to_parsed_lang/context.rs
otrho: > To be entirely honest, `Context` here is modelled after the [`Handler`](https://github.com/FuelLabs/sway/blob/master/sway-error/src/handler.rs) which uses `RefCell`. I don't really mind one way over another.. just went with `RefCell` for consistency. Thoughts?

I see.  Personally I'd rather avoid them unless there's no (sane) alternative.

mohammadfawaz: Alright, I'm no longer using `RefCell` for `Context`. I can update `Handler` in a separate PR along with moving `Handler` into `Context`.

mitchmindtree (approved):
Nicce :ok_hand:

mitchmindtree (approved):

JoshuaBatty (approved):
awesome :) :shipit:

Issue #3673, opened 2023-01-02
Inject `CONTRACT_ID` automatically while testing contracts


kayagokalp:
blocked by #3622.

After #3672, we will have the ability to call into contract abis but to do that we need to provide the contract id of the contract we are testing at that moment. Since we already know the contract id we can inject it automatically like we do for contract dependencies as mentioned by @mitchmindtree in the original issue (#3262). This will make the following code piece usable:

```rust
let caller = abi(MyContract, CONTRACT_ID);
let result = caller.test_function {}();
assert(result == true)
```

With #3672 we require the following manual entry:

```rust
let contract_id = 0xa8f18533afc18453323bdf17c83750c556916ab183daacf46d7a8d3c633a40ee;
let caller = abi(MyContract, contract_id);
let result = caller.test_function {}();
assert(result == true)
```

PR #3684, opened 2023-01-04
Exclude tests from `ParseTree` after parsing if they are not enabled

### LINKED ISSUE #3586: Do not emit warnings from in-language unit tests when compiling with `forc build`
Compiler the standard library emits the following warnings now:
```console
warning
   --> /Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/bytes.sw:664:5
    |
662 |
663 |     assert(bytes.len() == 9);
664 |     bytes.pop();
    |     ----------- This returns a value of type Option<u8>, which is not assigned to anything and is ignored.
665 |     bytes.pop();
666 |     bytes.pop();
    |
____

warning
   --> /Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/bytes.sw:665:5
    |
663 |
664 |     bytes.pop();
665 |     bytes.pop();
    |     ----------- This returns a value of type Option<u8>, which is not assigned to anything and is ignored.
666 |     bytes.pop();
667 |     bytes.pop();
    |
____

warning
   --> /Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/bytes.sw:666:5
    |
664 |
665 |     bytes.pop();
666 |     bytes.pop();
    |     ----------- This returns a value of type Option<u8>, which is not assigned to anything and is ignored.
667 |     bytes.pop();
668 |     bytes.pop();
    |
____

warning
   --> /Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/bytes.sw:667:5
    |
665 |
666 |     bytes.pop();
667 |     bytes.pop();
    |     ----------- This returns a value of type Option<u8>, which is not assigned to anything and is ignored.
668 |     bytes.pop();
669 |     bytes.pop();
    |
____

warning
   --> /Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/bytes.sw:668:5
    |
666 |
667 |     bytes.pop();
668 |     bytes.pop();
    |     ----------- This returns a value of type Option<u8>, which is not assigned to anything and is ignored.
669 |     bytes.pop();
670 |     assert(bytes.len() == 3);
    |
____

warning
   --> /Users/mohammad/Desktop/fuel/sway/sway-lib-std/src/bytes.sw:669:5
    |
667 |
668 |     bytes.pop();
669 |     bytes.pop();
    |     ----------- This returns a value of type Option<u8>, which is not assigned to anything and is ignored.
670 |     assert(bytes.len() == 3);
671 |     assert(bytes.pop().unwrap() == c);
    |
____

  Compiled library "std" with 6 warnings.
  Bytecode size is 28 bytes.
```
I'm getting rid of the warning in https://github.com/FuelLabs/sway/pull/3587 but we should not be emitting warnings from tests when compiling with `forc build`.

### LINKED ISSUE #3622: Add a way to omit tests at parsing step

kayagokalp:
closes #3622.
closes #3586.

unblocks #3673.

## About this PR

After parsing, I added a check to remove tests from `AstTree` if tests are not enabled. Since all the other checks are done after this exclusion of tests, this PR also prevents warnings and errors coming from test functions to be emitted while using `forc build`. They are still visible with `forc test` as they should be.


mohammadfawaz (dismissed):
Seems reasonable to me.

mohammadfawaz (dismissed):

JoshuaBatty (changes requested):
Nice, this seems reasonable. Would it be possible to set a custom config in LSP so that tests *are* included in the ParseProgram by default? Otherwise we will lose any LSP capabilities for test functions.

kayagokalp:
@JoshuaBatty ooh makes sense, adding it now, thanks!

kayagokalp:
I feel like e6505fc2561612aa3b574d66863321d7851ae9a4 should be enough but I couldn't test it thoroughly, will test this again in my morning

kayagokalp:
LSP capabilities looks intact to me after latest changes. I think this ready for re-review.

mohammadfawaz (approved):

JoshuaBatty (approved):

JoshuaBatty:
> LSP capabilities looks intact to me after latest changes. I think this ready for re-review.

Looks good, cheers.

forc/src/cli/commands/check.rs
```diff
@@ -23,6 +23,9 @@ pub struct Command {
     /// Terse mode. Limited warning and error output.
     #[clap(long = "terse", short = 't')]
     pub terse_mode: bool,
+    /// Disable checking unit tests.
+    #[clap(long = "disable-tests")]
+    pub disable_tests: bool,
```
mitchmindtree, line 28: Rather than having a dedicated `disable_tests` flag, I think we should probably disable tests by default, and only enable tests if a `--tests` flag is passed like `cargo check` (briefly mentioned [here](https://github.com/FuelLabs/sway/issues/3622#issuecomment-1362261173)).

PR #3685, opened 2023-01-04
Bootstrap the EVM backend.

tritao:
This bootstraps the EVM backend via the following:

- [x] Add build target support. This is implemented as a `--build-target / --target` CLI flag in Forc.
- [x] Allow for multiple asm builders The existing architecture was refactored to be trait-based, and the code was split into two directories, for `fuel` and `evm` backends.
- [x] Add minimal EVM assembly output. 
- [x] Add `evm` target testing support.
- [x] Implement EVM smart contract memory loader
- [x] Implement basic Solidity ABI generation


otrho (commented):
This is a great start!

This is a bit contentious but I would much rather not use `mod.rs` in the new `asm_builder` modules and use the newer way.  And that way we could probably avoid the redundant `evm/evm_asm_builder.rs` type layout.

I'm also thinking that the encapsulation around the different codegen targets might be made stricter as I think we can't assume parallels between the different workflows to always fit.  So the stuff under `programs/` will probably need to be split up, but I'm not sure, and I'd be happy to keep this and worry about that change when addressing #2505.

The test stuff with `revm` is great too. 👍

otrho:
OK, cool. 👍 New conflicts @tritao

otrho (approved):

emilyaherbert (approved):
Should we move asm generation to its own crate?

Otherwise LGTM. I largely defer to @otrho the more complex design details of the VM structure, but I like the approach you have taken here @tritao regarding interaction with multiple build targets and the initial EVM support.

otrho:
> > Should we move asm generation to its own crate?
> 
> Yeah, that's a great idea.

Definitely @emilyaherbert, I'm hoping to get into refactoring ASM gen ASAP.

mohammadfawaz:
Late review but this is great! Thanks @tritao 🚀

PR #3687, opened 2023-01-04
Adds a better error message in the case of using `ref` and `deref` expressions

### LINKED ISSUE #3677: Improve error message when `ref` is used incorrectly
with this code
```
contract;

abi MyContract {
    fn test_function() -> bool;
}

    fn test_function() -> bool {
        foo(ref Vec::new());
        true
    }
}

pub fn foo(ref mut vec: Vec<u64>) {
    vec.push(1);
}
```

This error is recieved
```
thread 'main' panicked at 'not implemented', /home/swaystar123/.cargo/registry/src/github.com-1ecc6299db9ec823/sway-core-0.32.2/src/transform/to_parsed_lang/convert_parse_tree.rs:1571:29
```

emilyaherbert:
Instead of using the Rust `unimplemented!()` macro, Sway users now see an "unimplemented" error from the Sway compiler when attempting to use `ref` or `deref` expressions.

Closes #3677

### DIFF
--- a/sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
+++ b/sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
@@ -1584,8 +1584,18 @@ fn expr_to_expression(
             }),
             span,
         },
-        Expr::Ref { .. } => unimplemented!(),
-        Expr::Deref { .. } => unimplemented!(),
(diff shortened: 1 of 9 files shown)


mohammadfawaz (approved):
Great, thanks.

JoshuaBatty (approved):
🚢

PR #3689, opened 2023-01-04
Add an error message about nested function definitions.

### LINKED ISSUE #3097: Disallow functions defined inside other functions
Example
```rust
script;

fn main() {
    fn bla() { }
}
```
This fails to compile with an ICE:
```console
Internal compiler error: Unexpected function declaration found.
```

It would be better if this was a clean parser or frontend error instead.

emilyaherbert:
This PR finds cases of nested function definitions like:

```rust
script;

fn main() {
    fn bla() { }
}
```

and gives an "unimplemented" error. (as opposed to an ICE).

Closes #3097

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/declaration/function.rs
+++ b/sway-core/src/semantic_analysis/ast_node/declaration/function.rs
@@ -1,7 +1,10 @@
 mod function_parameter;
 
 pub use function_parameter::*;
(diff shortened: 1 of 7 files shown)


JoshuaBatty (approved):

mohammadfawaz (approved):
Very cool. Thanks.

PR #3690, opened 2023-01-04
Disallow repeated symbols where applicable

### LINKED ISSUE #3306: Defining constants multiple times should be a compile error

mohammadfawaz:
Closes #3306 

This PRs disallows shadowed symbols where applicable. Specifically, and similarly to Rust, the following rules are applied:
- types/traits/abis are not allowed to shadow other types/traits/abis with the same name
- constants are not allowed to shadow other constants with the same name
- constants are not allowed to shadow variables with the same name
- variables are not allowed to shadow constants with the same name

Enforcing shadowing rules now looks at both the local and imported items.

There are still some additional rules to take care of related to functions with the same name as other declarations (which I think are handled somewhere else right now). Also, I think shadowing between items imported from two different libraries is overly aggressive. Just starting with this PR for now and will do more clean up in the future.

I added a lot of tests to check for these rules. I also removed some obsolete tests due to my new tests. One of the tests was checking for global const shadowing (by a local const) which is technically allowed in Rust. It was allowed for us but somewhat by luck since shadowing was allowed for consts in any context. This should be addressed separately by checking that the second const is in a "sub-scope" of the first const.


sway-core/src/semantic_analysis/namespace/items.rs
```diff
@@ -101,31 +98,51 @@ impl Items {
...
+                | (
+                    StructDeclaration { .. }
+                    | EnumDeclaration { .. }
+                    | TraitDeclaration { .. }
+                    | AbiDeclaration { .. },
+                    StructDeclaration { .. }
+                    | EnumDeclaration { .. }
+                    | TraitDeclaration { .. }
+                    | AbiDeclaration { .. },
+                ) => errors.push(CompileError::NameDefinedMultipleTimes { name: name.to_string(), span: name.span() }),
```
mohammadfawaz, line 127: The formatter is not able to format this closure properly 🤷

sway-core/src/semantic_analysis/ast_node/declaration/declaration.rs
```diff
@@ -70,7 +70,12 @@ impl ty::TyDeclaration {
                         type_ascription,
                         type_ascription_span,
                     }));
-                ctx.namespace.insert_symbol(name, typed_var_decl.clone());
+                check!(
+                    ctx.namespace.insert_symbol(name, typed_var_decl.clone()),
+                    return err(warnings, errors),
+                    warnings,
+                    errors
+                );
```
mitchmindtree, line 78: Ahh I wonder how we missed these in the past :thinking: Perhaps we should decorate `CompileResult` with `#[must_use]` so that we are forced to handle results like this, or explicitly ignore them (like `std::result::Result`)? Not the responsibility of this PR of course, just a thought!

sway-core/src/semantic_analysis/namespace/items.rs
mitchmindtree: Good to know it's not just us struggling with formatter correctness :sweat_smile:

mitchmindtree: I love how the pattern matching here clearly lays out our shadowing rules btw :heart:

test/src/e2e_vm_tests/test_programs/should_fail/shadowing/shadowed_glob_imports/src/main.sw
```diff
@@ -0,0 +1,39 @@
...
+struct MyAbi14 {}
+enum MyAbi24 {}
+trait MyAbi34 {}
+abi MyAbi44 {}
+
+
+fn main() {
+    // var shadowing an imported const with glob
+    let X2 = 4;
+}
```
mitchmindtree, line 39: It looks like this example passes, but is under the `should_fail` test_programs. Should we create another `shadowing` dir under the `should_pass` directory for those expected to pass?

sway-core/src/semantic_analysis/ast_node/declaration/declaration.rs
mohammadfawaz: Hmm that's very interesting. I'll create an issue. That being said, if we end up implementing https://github.com/FuelLabs/sway/issues/2734, then we won't have that particular issue I think.

test/src/e2e_vm_tests/test_programs/should_fail/shadowing/shadowed_glob_imports/src/main.sw
mohammadfawaz: I can do that, but honestly the distinction between `should_pass` and `should_fail` is not very relevant anymore and we should just merge them. I think there's an issue tracking that.

mitchmindtree (approved):

anton-trunov (approved):

PR #3691, opened 2023-01-05
Change `expect` messages to `Result` in `forc doc`

### LINKED ISSUE #3655: Convert `expect()` messages in `forc doc` to `anyhow::Result`
import `anyhow::Result` where necessary and remove the `expect` messages in favor of `Result`

gr00vytvniks:
Work in progress, implementing `anyhow::Result` and handling `expect` & `unwrap` where necessary

Closes #3655


bingcicle (changes requested):
Thanks for the PR! :) Just a few nits

forc-plugins/forc-doc/src/descriptor.rs
```diff
@@ -81,8 +85,14 @@ impl Descriptor {
...
-                    let attrs_opt = (!enum_decl.attributes.is_empty())
-                        .then(|| attrsmap_to_html_str(enum_decl.attributes));
+                    let attrs_opt = if !enum_decl.attributes.is_empty() {
+                        match attrsmap_to_html_str(enum_decl.attributes) {
+                            Ok(s) => Some(s),
+                            Err(e) => bail!("{}", e),
+                        }
+                    } else {
+                        None
+                    };
```
kayagokalp, line 95: ^^ Ditto

forc-plugins/forc-doc/src/descriptor.rs
```diff
@@ -112,11 +122,17 @@ impl Descriptor {
...
-                    let attrs_opt = (!trait_decl.attributes.is_empty())
-                        .then(|| attrsmap_to_html_str(trait_decl.attributes));
+                    let attrs_opt = if !trait_decl.attributes.is_empty() {
+                        match attrsmap_to_html_str(trait_decl.attributes) {
+                            Ok(s) => Some(s),
+                            Err(e) => bail!("{}", e),
+                        }
+                    } else {
+                        None
+                    };
```
kayagokalp, line 132: ^^ Ditto

forc-plugins/forc-doc/src/descriptor.rs
```diff
@@ -143,11 +159,17 @@ impl Descriptor {
...
-                let attrs_opt = (!abi_decl.attributes.is_empty())
-                    .then(|| attrsmap_to_html_str(abi_decl.attributes));
+                let attrs_opt = if !abi_decl.attributes.is_empty() {
+                    match attrsmap_to_html_str(abi_decl.attributes) {
+                        Ok(s) => Some(s),
+                        Err(e) => bail!("{}", e),
+                    }
+                } else {
+                    None
+                };
```
kayagokalp, line 169: ^^ Ditto

forc-plugins/forc-doc/src/descriptor.rs
```diff
@@ -173,8 +195,14 @@ impl Descriptor {
...
-                let attrs_opt = (!storage_decl.attributes.is_empty())
-                    .then(|| attrsmap_to_html_str(storage_decl.attributes));
+                let attrs_opt = if !storage_decl.attributes.is_empty() {
+                    match attrsmap_to_html_str(storage_decl.attributes) {
+                        Ok(s) => Some(s),
+                        Err(e) => bail!("{}", e),
+                    }
+                } else {
+                    None
+                };
```
kayagokalp, line 205: ^^ Ditto

forc-plugins/forc-doc/src/descriptor.rs
```diff
@@ -229,8 +257,14 @@ impl Descriptor {
...
-                    let attrs_opt = (!fn_decl.attributes.is_empty())
-                        .then(|| attrsmap_to_html_str(fn_decl.attributes));
+                    let attrs_opt = if !fn_decl.attributes.is_empty() {
+                        match attrsmap_to_html_str(fn_decl.attributes) {
+                            Ok(s) => Some(s),
+                            Err(e) => bail!("{}", e),
+                        }
+                    } else {
+                        None
+                    };
```
kayagokalp, line 267: ^^ Ditto

kayagokalp (changes requested):
Thanks for the contribution👌, I left some comments, suggestions to clean up some of the things a little, also agree with @bingcicle's comments in the review. 

As a discussion point:

I generally tend to use `anyhow::Result` where the source of error is user related and `expect` in places where I expect something to be there and if it is not it is due to an error in the code logic. So some of the things here might be better of left as `expect` imo but I am good with it either way 👍 I am not aware if this is the idiomatic way of deciding between `expect` and `Result` more like a rule I tend to follow without being explicit about it

gr00vytvniks:
@kayagokalp @bingcicle  It doesn't look like I can request both of your reviews at the same time so i'll just tag you both here

bingcicle (changes requested):

forc-plugins/forc-doc/src/doc.rs
```diff
@@ -119,34 +120,41 @@ impl Document {
...
 #[derive(Clone)]
 pub(crate) struct ModuleInfo(pub(crate) Vec<ModulePrefix>);
 impl ModuleInfo {
     /// The current module.
-    pub(crate) fn location(&self) -> &str {
+    pub(crate) fn location(&self) -> Result<&String> {
         self.0
             .last()
-            .expect("There will always be at least the project name")
+            .ok_or_else(|| anyhow::anyhow!("Expected Some module location, found None"))
```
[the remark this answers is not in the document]
kayagokalp: Yepp, seems like I missed it sorry for this one @gr00vytvniks, let's go back to `match` and +1 on the otter comment too

eureka-cpu:
Found a bug that is unrelated to this #3923

eureka-cpu:
We should probably get this in sooner rather than later lest we want more merge conflicts. PR is up for the bug mentioned above: #3930

forc-plugins/forc-doc/src/doc.rs
```diff
@@ -160,26 +161,35 @@ impl Document {
...
 pub(crate) struct ModuleInfo(pub(crate) Vec<ModulePrefix>);
 impl ModuleInfo {
     /// The current module.
+    ///
+    /// Panics if there are no modules.
     pub(crate) fn location(&self) -> &str {
         self.0
             .last()
-            .expect("There will always be at least the project name")
+            .expect("Expected Some module location, found None")
```
bingcicle, line 184: Sorry if I wasn't clear with what I meant by "going back to our original matching", but this could be what you originally went with in your initial PR:

```rust
match self.0.last() {
  Some(location) => Ok(location),
  None => bail!("There will always be at least the module location"),
```

Note also the err message change, as pointed out [here](https://github.com/FuelLabs/sway/pull/3691#discussion_r1073313586).

forc-plugins/forc-doc/src/doc.rs
```diff
@@ -160,26 +161,35 @@ impl Document {
...
     }
     /// The name of the project.
+    ///
+    /// Panics if the project root is missing.
     pub(crate) fn project_name(&self) -> &str {
-        self.0.first().expect("Project name missing")
+        self.0
+            .first()
+            .expect("Expected root module, project root missing")
```
bingcicle, line 192: Same as above:

```suggestion
        match self.0.first() {
          Some(name) => Ok(name),
          None => bail!("Project name missing"),
        }
```

bingcicle (changes requested):
Thank you for your patience @gr00vytvniks ! Just last few things here.

forc-plugins/forc-doc/src/doc.rs
eureka-cpu: Hmm I think for now the expect message here is okay. Implementing result there means you'd have to worry about how you handle every call to location in the render macro which doesn't return a result. There's a few for loops in there where I can see it would be difficult to find a non-hacky way of resolving the results before using them in the `box_html!` macro.

bingcicle: that's reasonable!

bingcicle (dismissed):
Thank you @gr00vytvniks, great work! Giving this the :+1:

eureka-cpu (approved):
🚢

bingcicle (approved):
:shipit:

mohammadfawaz:
@kayagokalp this is waiting on a re-review from you :)

kayagokalp:
Ahh must have missed this one! thanks for pinging @mohammadfawaz 👍

kayagokalp (approved):

PR #3700, opened 2023-01-05
Demonstrate that a bug with numeric methods is resolved

### LINKED ISSUE #832: `numeric` type cannot look up any methods
With the introduction of a `numeric` type, which is castable to any unsigned integer type, we have been able to infer a lot about integer widths and remove a lot of downcasting warnings.

Unfortunately, we are unable to look up methods for this type. 
```rust
29 |
30 |     // should return 15
31 |     if let Result::Ok(y) = x { y + 10 } else { 1 }
   |                                  ^ No method named "add" found for type "numeric".
32 | }
```

I think it would be reasonable to either cast `numeric` to a `u64` when looking up methods, or, check all numeric types starting from `u64` and going down, then casting it to the first type that had a matching method.

For anybody who encounters this issue, the workaround is to just define the number  as a concrete type. E.g. `let x = 5u64` or `let x: u64 = 5`.

emilyaherbert:
This PR adds a test case to demonstrate that the bug described in #832 is resolved.

Closes #832

### DIFF
--- a/test/src/e2e_vm_tests/test_programs/should_pass/language/enum_if_let/src/main.sw
+++ b/test/src/e2e_vm_tests/test_programs/should_pass/language/enum_if_let/src/main.sw
@@ -12,6 +12,10 @@ const B1 = Address {
 };
 
 fn main() -> u64 {
+    let a = Result::Ok::<u64, u64>(100);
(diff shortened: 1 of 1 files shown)


mohammadfawaz (approved):
Nice.

sezna (approved):

PR #3705, opened 2023-01-06
Fixes nested generic type inference errors.

### LINKED ISSUE #3321: Nested generic type inference error

esdrubal:
The issue was that `TraitMap::filter_by_type_inner` monomorphization was not working properly for nested generic types, some type_ids were not properly replaced thus we would get type inference errors.

The solution was to fix `TypeMapping::from_superset_and_subset` to be recursive.

Closes #3321

### DIFF
--- a/sway-core/src/type_system/type_mapping.rs
+++ b/sway-core/src/type_system/type_mapping.rs
@@ -156,7 +156,8 @@ impl TypeMapping {
                     .iter()
                     .map(|x| x.type_id)
                     .collect::<Vec<_>>();
-                TypeMapping::from_type_parameters_and_type_arguments(
+                TypeMapping::from_superset_and_subset_helper(
+                    type_engine,
                     type_parameters,
                     type_arguments,
                 )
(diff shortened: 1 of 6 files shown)


emilyaherbert (approved):
Great PR!

PR #3706, opened 2023-01-06
Adds docstring to core::never::Never

esdrubal:


sway-lib-core/src/never.sw
```diff
@@ -2,6 +2,25 @@ library never;
 
 use ::ops::{Not, Eq, Ord};
 
+/// `Never` is assigned to a block that is deterministically aborted. This can be
```
mohammadfawaz, line 5: I was hoping we'd copy the documentation from https://doc.rust-lang.org/std/primitive.never.html verbatim (to the extent possible). That's what we've tried to do in other spots in the standard library. Is that reasonable here or are there some major differences between our `Never` type and Rust's `Never` type?

sway-lib-core/src/never.sw
```diff
@@ -2,6 +2,25 @@ library never;
...
+///         return 5;
+///     };
+///     123
+/// }
+/// ```
+///
+/// In the example above the assigning of `return 5` to `b` is pointless because
+/// `return 5` returns from the entire function. Nonetheless internally the return
+/// block will be considered of type `Never`. Furthermore this example shows also
+/// that `Never` type can be coerced into any other type, here `b` is of type `u32`
```
emilyaherbert, line 22: "this example shows also that `Never` type can be coerced into any other type". This bit is pretty important, can you add docstrings to the `impl`'s for `Never` and include this as a note?

emilyaherbert (commented):
Could you add a short PR description? Otherwise LGTM aside from @mohammadfawaz's comments.

sway-lib-core/src/never.sw
esdrubal: Copied the link above to the possible extend. Rust has more use cases to
the Never type than Sway currently has, so the sections were not included in the
docstring as they are not applicable to the Sway Never type.

sway-lib-core/src/never.sw
esdrubal: Looked into Sway and Rust documentation and usually the docstring are only attached to traits not to impls.
Added note about `Never` type coercion.

mohammadfawaz (approved):
LGTM! Thanks.

PR #3708, opened 2023-01-06
std-lib wrapper around BHSH (block header hash) opcode

simonr0204:


mohammadfawaz:
@nfurfaro is this something you can take on?

nfurfaro:
Yes, assigned to me now.

PR #3709, opened 2023-01-06
Fix bug in calculating missing patterns in nested enum matching in match expressions

### LINKED ISSUE #2878: Bug in calculating missing patterns in nested enum matching in match expressions, throwing ICE
This code generates an internal compiler error, when it shouldn't:

```rust
contract;

use std::{chain::auth::{AuthError, msg_sender}};

abi MyContract {
    fn test_function() -> bool;
}

const ORACLE_1: b256 = 0x6b63804cfbf9856e68e5b6e7aef238dc8311ec55bec04df774003a2c96e0418e;
const ORACLE_2: b256 = 0x54944e5b8189827e470e5a8bacfc6c3667397dc4e1eef7ef3519d16d6d6c6610;
const ORACLE_3: b256 = 0xe10f526b192593793b7a1559a391445faba82a1d669e3eb2dcd17f9c121b24b1;

    fn test_function() -> bool {
        let sender: Result<Identity, AuthError> = msg_sender();
        match sender {
            Result::Ok(Identity::Address(Address { value: ORACLE_1 })) => true,
            Result::Ok(Identity::Address(Address { value: ORACLE_2 })) => true,
            Result::Ok(Identity::Address(Address { value: ORACLE_3 })) => true,
            // Result::Ok(_) => false,
            Result::Err(_) => false,
        }
    }
}
```

and this also creates an ICE:

```rust
    fn test_function() -> bool {
        let sender: Result<Identity, AuthError> = msg_sender();
        match sender.unwrap() {
            Identity::Address(Address { value: ORACLE_1 }) => true,
            Identity::Address(Address { value: ORACLE_2 }) => true,
            Identity::Address(Address { value: ORACLE_3 }) => true,
			// Identity::Address(_) => false,
            Identity::ContractId(_) => false,
        }
    }
}
```

Appropriate error messages should look something like: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=faedbae316077ef66f9f79b3ff6b1be9

emilyaherbert:
This PR fixes a bug with the match expression exhaustivity checker that was causing an ICE in specific conditions. The fix required a change to how witness reports are combined in the wildcard case. This PR also has minor updates to how ranges are displayed to the user in an error messages to eliminate odd `[MIN..0]` error messages, and also small updates to rustdoc documentation.

One thing to note. There is a limitation with the match expression exhaustivity algorithm that Sway (and Rust!) is currently using. In cases where the user has provided match arms with varying levels of nested patterns, the algorithm defaults to displaying the missed case at the top most level of relevant pattern matching. Here is an example:

```rust
const ORACLE_1: u64 = 1;
const ORACLE_3: u64 = 3;

struct MyAddress {
    inner: u64,
}

struct MyContractId {
    inner: bool
}

enum MyIdentity {
    Address: MyAddress,
    ContractId: MyContractId,
}

enum MyAuthError {
    ReasonOne: (),
    ReasonTwo: (),
}

pub fn enum_match_exp_bugfix_test() {
    let a: Result<MyIdentity, MyAuthError> = Result::Ok(
        MyIdentity::Address(
            MyAddress { inner: 7 }
        )
    );

    // should fail with non-exhaustive
    let e = match a {
        Result::Ok(MyIdentity::Address(MyAddress { inner: ORACLE_1 })) => 1,
        // missing Result::Ok(MyIdentity::ContractId(_))
        Result::Err(_) => 5,
    };

    // should fail with non-exhaustive
    let f = match a {
        Result::Ok(MyIdentity::Address(MyAddress { inner: ORACLE_1 })) => 1,
        Result::Ok(MyIdentity::ContractId(_)) => 2,
        // missing Result::Ok(MyIdentity::Address(MyAddress { inner: 0, 2..MAX }))
        Result::Err(_) => 5,
    };
}
```

Both of these cases would fail with a "non-exhaustive" error, but the former would only display to the user that they are missing the case `Result::Ok(MyIdentity::ContractId(_))`. When the user adds the case `Result::Ok(MyIdentity::ContractId(_))`, then a different "non-exhaustive" error is displayed with the cases `Result::Ok(MyIdentity::Address(MyAddress { inner: 0, 2..MAX }))`.

Rust and Sway use the same exhaustivity algorithm, and Rust mimics this exact same behavior. See: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6b1e99c1e281b4115c9994674ac388de

In my opinion, this isn't a bug, as the non-exhaustive code will fail appropriately. It is more just a thing to be aware of.

Closes  #2878


otrho (approved):
Nice!

JoshuaBatty (approved):

PR #3726, opened 2023-01-09
LSP: Save the `LexedProgram` AST to a temp file  

JoshuaBatty:
This PR saves the `LexedProgram` AST to a temp file so it can be displayed in VScode.

I have a PR open in the VScode plugin that adds a "Sway: Show Lexed AST" command to the command palette. When that is executed, it loads the file saved to temp memory that the server has saved. https://github.com/FuelLabs/sway-vscode-plugin/pull/114

We currently have a command for showing the parsed and typed ASTs of the currently open file, this just adds functionality for viewing the initial lexing stage.

### DIFF
--- a/sway-lsp/src/server.rs
+++ b/sway-lsp/src/server.rs
@@ -18,7 +18,7 @@ use std::{
     path::{Path, PathBuf},
     sync::Arc,
 };
-use sway_types::Spanned;
+use sway_types::{Ident, Spanned};
 use tower_lsp::lsp_types::*;
 use tower_lsp::{jsonrpc, Client, LanguageServer};
 use tracing::metadata::LevelFilter;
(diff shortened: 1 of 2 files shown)


sway-lsp/src/server.rs
```diff
@@ -532,52 +532,58 @@ impl Backend {
...
 
+                // Returns true if the current path matches the path of a submodule
+                let path_is_submodule = |ident: &Ident, path: &Option<PathBuf>| -> bool {
+                    ident.span().path().map(|a| a.deref()) == path.as_ref()
+                };
+
                 {
                     let program = session.compiled_program.read();
                     match params.ast_kind.as_str() {
+                        "lexed" => {
```
eureka-cpu, line 543: just out of curiosity, is there a reason we match on the string slice here rather than the `ast_kind` as is? That is, assuming it's an enum.

eureka-cpu (approved):

sway-lsp/src/server.rs
JoshuaBatty: Yeah that `ast_kind` is a `String`. Converting it to a `str` here just means there is less of `"lexed".to_string()` needed when matching for example.

sdankel (approved):

Issue #21, opened 2021-03-19
Support multi-span errors and warnings


sezna:
Often, you want to display help text or render multiple spans in an error.

vaivaswatha:
> A good one to start might be "not declared as mutable... but used as mutable here ^". What do you think?

Another error that might benefit from this: https://github.com/FuelLabs/sway/issues/751. "Function ^ is being redefined. It was already previously defined ^"

anton-trunov:
One more issue that would benefit from this: #3532. The CEI pattern analysis would highlight just two lines: the one with the external call and the one with storage modification or some other effect.

mohammadfawaz:
I recall that we tried to do this but then reverted the change for some reason? Is this too difficult to implement? I do find myself needing this often. cc @sezna. Also, if it's too complex, we should probably remove the `good first issue` tag 😄

sezna:
There was a bug previously where the compiler would panic if the two errors were from different files, which was usually the case, so we reverted the old implementation.

ironcev:
Here is the progress so far and two open questions to agree on. Please provide your feedback on the questions.

# Before and after :-)

![01A Constant shadowing - Before](https://github.com/FuelLabs/sway/assets/4142833/97e70a69-de62-4c86-b4a4-9e03deaf3a22)

![01B Constant shadowing - After](https://github.com/FuelLabs/sway/assets/4142833/88ddb46e-31ee-41da-98e8-006f3f71ea94)

# Open Questions
## Structure of compile messages

What do we want to support in our compile messages and how much freedom, or restriction do we want to offer. For the sake of consistency, I am for a limited, clearly defined number of elements that we can embed in a compile message. This way we can impose consistency across all our messages. So far I support these elements, and enforce them via the compile message (CM) API:

- CM can have a _label_.
- CM must have exactly one (looking for a better term) _message_ that can be an error or a warning. This is what we have as a single error/warning right now.
- CM can provide additional _info_. These are references to other places in code related to the core message. They can span over different files.
- CM can provide _help_. _Help_ is a plain text. We can provide arbitrary many _help_ strings.
- _Info_ refer to places in files, whereas the _message_ itself can (and mostly will) but does not have to (we have such situation now).

You can spot all of these elements on the above image.

What do you think of this approach? Do you see any other elements that we could need?

## Declarative API

At the moments messages are declared imperatively, pretty much like the way we do warnings right now. This means a lot of clutter and boilerplate. E,g,:

```
[...]

PR #3728, opened 2023-01-09
Improves trait map insertion and getting.

### LINKED ISSUE #3623: Generic Result Method Failure
Generic Result method fails when explicitly two assertions are in the same test function.

```rs
enum Result<T, E> {
    Ok: T,
    Err: E,
}

impl<T, E> Result<T, E> {
    pub fn unwrap_or(self, default: T) -> T {
        match self {
            Result::Ok(inner_value) => inner_value,
            Result::Err(_) => default,
        }
    }
}

// REVERTS
pub fn test_unwrap_or<T>(val: T, default: T)
where
    T: Eq
{
    assert(Result::Ok::<T, T>(val).unwrap_or(default) == val);
    assert(Result::Err::<T, T>(val).unwrap_or(default) == default);
}

// PASSES
pub fn test_unwrap_or_ok<T>(val: T, default: T)
where
    T: Eq
{
    assert(Result::Ok::<T, T>(val).unwrap_or(default) == val);
}

// PASSES
pub fn test_unwrap_or_err<T>(val: T, default: T)
where
    T: Eq
{
    assert(Result::Err::<T, T>(val).unwrap_or(default) == default);
}
```

## Reproduce

https://github.com/jtriley-eth/test-proj

## Version Information

Forc version: 0.32.2
Cargo Version: 1.68.0-nightly (cc0a32087 2022-12-14)
Cargo Generate version: 0.17.4
Mac OS Montery (M1) Version 12.6

esdrubal:
`insert_trait_implementation` and
`get_methods_for_type_and_trait_name` now use a new method
 `trait_name.to_fullpath` this improves consistency between inserting
and getting by using a full path representation in stored trait implementations.

Adds test generic_result_method which covers issue #3623.

Closes #3623

This is a subset of PR #3621


sway-core/src/semantic_analysis/namespace/module.rs
```diff
@@ -41,6 +41,8 @@ pub struct Module {
     pub(crate) submodules: im::OrdMap<ModuleName, Module>,
     /// The set of symbols, implementations, synonyms and aliases present within this module.
     items: Items,
+    /// Name of the module, package name for root module, dependency name for other modules
```
emilyaherbert, line 44: Would you be able to add some examples of what this would look like in the rustdoc comment here? I'm kind of struggling a little bit to solidify what this field is going to mean concretely.

sway-core/src/semantic_analysis/namespace/namespace.rs
```diff
@@ -290,6 +290,43 @@ impl Namespace {
...
+        )
+    }
+
+    pub(crate) fn get_methods_for_type_and_trait_name(
+        &mut self,
+        engines: Engines<'_>,
+        type_id: TypeId,
+        trait_name: &CallPath,
+    ) -> Vec<DeclarationId> {
+        // Use trait name with module path as this is expected in get_methods_for_type_and_trait_name
```
emilyaherbert, line 324: Can you add this same comment to the `insert_trait_implementation` method above?

sway-core/src/language/call_path.rs
```diff
@@ -45,7 +45,11 @@ impl<T: Spanned> Spanned for CallPath<T> {
         if self.prefixes.is_empty() {
             self.suffix.span()
         } else {
-            let prefixes_spans = self.prefixes.iter().map(|x| x.span());
+            let prefixes_spans = self
+                .prefixes
+                .iter()
+                .map(|x| x.span())
+                .filter(|x| x.path() == self.suffix.span().path());
```
emilyaherbert, line 52: Will you add a code comment above this to document that this LOC should be removed when #21 goes in?

sway-core/src/semantic_analysis/namespace/module.rs
```diff
@@ -144,6 +146,8 @@ impl Module {
 
     /// Insert a submodule into this `Module`.
     pub fn insert_submodule(&mut self, name: String, submodule: Module) {
+        let mut submodule = submodule;
```
emilyaherbert, line 149: Code style nit: remove this LOC and add `mut` in front of the argument `submodule`.

emilyaherbert (changes requested):

emilyaherbert (approved):

emilyaherbert:
I went ahead and approved because all of my request are code quality ones 😄 but please do correct those if this PR gets merged before you get the opportunity to change them.

sway-core/src/semantic_analysis/namespace/namespace.rs
esdrubal: Added comment to both methods.

emilyaherbert (approved):
LGTM

PR #3729, opened 2023-01-09
feat: inject contract id into the namespace for tests automatically

### LINKED ISSUE #3673: Inject `CONTRACT_ID` automatically while testing contracts

kayagokalp:
closes #3673.

## About this PR

This PR adds `CONTRACT_ID` injection while building contracts with `forc build --test`. `CONTRACT_ID` is the id of the contract without the tests added. To find out that id, we first compile the contract without tests enabled. 

The rough outline of stuff happening here:

1. We iterate over `BuildPlan` members to find out contracts and collect their contract id into an injection map. In this step only the contracts are built. To determine contract id we need to compile the contract without tests. Since we also need the bytecode of the contract without tests, we are collecting them as we come across them while iterating over members.
2. With the injection map build and execute all the tests, so basically after first step we are just doing the old `forc-test` behaviour.

So basically I added a pre-processing step for contract id collection for those members that require it (contracts).

This enables the following test structure:

```rust
let caller = abi(MyContract, CONTRACT_ID);
let result = caller.test_function {}();
assert(result == true)
```

## TODO

- [x] Documentation


forc-pkg/src/pkg.rs
```diff
@@ -85,7 +85,21 @@ pub struct PinnedId(u64);
...
+
+/// The result of successfully compiling a contract package.
+#[derive(Debug, Clone)]
+pub struct BuiltContract {
+    pub without_tests: BuiltPackageDetails,
+    pub with_tests: Option<BuiltPackageDetails>,
+}
+
+/// The details of a built package.
+#[derive(Debug, Clone)]
```
kayagokalp, line 101: Not really sure if this is the cleanest way to do this. I added this struct to hold common information that we have for contract and non-contract packages. Open to any suggestions 🙌

mitchmindtree (changes requested):
Excited for this!

That said, I don't think we should be distinguishing between contract and non-contract in `forc-pkg` itself like this, as the difference is not particularly relevant to `forc-pkg` and is only a concern of `forc-test`.

We should aim to make the minimum changes possible in `forc-pkg`. Namely, we should only need to add the ability to optionally add an item to a package's namespace when building it (so that we can add the `CONTRACT_ID`).

Rather than having `forc-pkg` special-case contracts, I think we should instead do something like the following:

1. In `forc-test`, rather than building the whole workspace, iterate over members (in the correct order) and only build the contract members without tests.
2. Add an option to `forc-pkg` build functionality that allows for injecting item(s) into a package's namespace.
3. In `forc-test`, rather than building the whole workspace with tests at once, iterate over the members (in the correct order) and inject the CONTRACT_ID into contract members as necessary.

This might end up being slightly more verbose, but should help to keep a clearer separation of concerns between `forc-pkg` and `forc-test`. Thoughts?

kayagokalp:
I wasn't really happy with how this turned out tbh, thanks for the comment @mitchmindtree! I think the proposal seems it would be a lot cleaner. I will come-up with the implementation tomorrow morning 👍

kayagokalp:
This should be good to go for another review, I will be introducing a general refactor to `forc-test` to address some of the left over review comments from earlier PRs.

forc-test/src/lib.rs
```diff
@@ -365,9 +375,49 @@ impl BuiltTests {
...
+                .include_tests(false);
+            let built_contract_without_tests =
+                pkg::build_with_options(build_opts_without_tests)?.expect_pkg()?;
+            let contract_id =
+                pkg::contract_id(&built_contract_without_tests, &fuel_tx::Salt::zeroed());
+
+            built_contracts.insert(pinned_member.name.clone(), built_contract_without_tests);
+
+            // Construct namespace with contract id
+            let contract_id_constant_name = "CONTRACT_ID".to_string();
```
mitchmindtree, line 405: Do we have a `const` for the `"CONTRACT_ID"` str literal? I think we also use it in `forc_pkg`, maybe we can import it from there?
```suggestion
            let contract_id_constant_name = "CONTRACT_ID".to_string();
```

forc-test/src/lib.rs
```diff
@@ -365,9 +375,49 @@ impl BuiltTests {
...
+            let contract_id_value = format!("0x{contract_id}");
+            let contract_id_constant = ConfigTimeConstant {
+                r#type: "b256".to_string(),
+                value: contract_id_value.clone(),
+                public: true,
+            };
+            let constant_declarations = vec![(contract_id_constant_name, contract_id_constant)];
+            inject_map.insert(pinned_member, constant_declarations);
+        }
+    }
```
mitchmindtree, line 415: Perhaps we can abstract this section out into a function, e.g.

```rust
fn build_contracts_with_tests(opts: &Opts, build_plan: &BuildPlan) -> BuiltContractsWithTests
```

mitchmindtree (changes requested):
Nice, this new approach is looking much better! Just a few nits we should address first.

forc-test/src/lib.rs
```diff
@@ -134,23 +143,21 @@ struct TestSetup {
...
     ///
-    /// If the `built` is a workspace, `opts` is patched for the members of the workspace that
-    /// is a `Contract` so that only that contract is re-built.
-    ///
-    /// If the `built` is a package, `PackageTests::from_built_pkg` is used.
-    pub(crate) fn from_built(built: Built, opts: Opts) -> anyhow::Result<BuiltTests> {
+    /// Contracts are already compiled once without tests included to do `CONTRACT_ID` injection. `built_contracts` map holds already compiled contracts so that they can be matched with their "tests included" version.
+    pub(crate) fn from_built(
+        built: Built,
+        built_contracts: HashMap<String, BuiltPackage>,
```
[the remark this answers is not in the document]
kayagokalp: After realizing cargo does not allow package name duplication based on package name for the build plan, I opened-up #3840 and since duplicate names would not be possible in our build plan too, I though we could use String as key but if you feel like `Pinned` would be more robust here, happy to go that way too 🙌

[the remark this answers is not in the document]
kayagokalp: I ended up adding a field (`BuiltPackageDescriptor`) to `BuiltPackage` so that we can retrieve some information about the source of the package from `BuiltPackage`. This enabled me to use `Pinned` as a key as I have a way to get `Pinned` of a pkg built with tests and the `Pinned` of a pkg built with tests and built without tests is the same (since they are the same package).

forc-test/src/lib.rs
```diff
@@ -362,12 +372,67 @@ impl BuiltTests {
...
+                .clone()
+                .patch_opts(pkg_path)
+                .into_build_opts()
+                .include_tests(false);
+            let built_pkg =
+                pkg::build_with_options(build_opts_without_tests).and_then(|pkg| pkg.expect_pkg());
+            (pinned_pkg, built_pkg)
+        })
+        .collect()
+}
```
kayagokalp, line 402: I wanted to return an iterator over the elements rather than a `Vec` but couldn't manage to get pass the compiler due to strange lifetime complains. I ended up collecting them this dynamically allocates a `Vec` but if I could return an iterator over `BuiltPackage` references, I would need need to clone them in `build()` as I need the owned version to store in `built_contracts` since I would `clone` all `BuiltPackage` references, I feel like performance wise these two options should be the same but wanted to hear your thoughts on this @mitchmindtree

mitchmindtree: In this case there likely won't be a noticeable performance impact as when using this function, we end up storing all of the `BuiltPackage`s either way. However, from an API design perspective it's almost always better to return an Iterator if feasible as it gives the function user more control over their costs.

> couldn't manage to get pass the compiler due to strange lifetime complains

Sometimes when there are multiple references as function inputs, Rust can't safely infer the correct lifetime for the output, so using `'_` in something like this doesn't work:

```rust
fn build_contracts_without_tests(
    opts: &Opts,
    build_plan: &pkg::BuildPlan,
) -> impl '_ + Iterator<Item = (pkg::Pinned, anyhow::Result<BuiltPackage>)> {
```

You can work around this by explicitly naming the lifetime, so it's clear to Rust what inputs the lifetime is bound to, e.g.

```rust
fn build_contracts_without_tests<'a>(
    opts: &'a Opts,
    build_plan: &'a pkg::BuildPlan,
) -> impl 'a + Iterator<Item = (pkg::Pinned, anyhow::Result<BuiltPackage>)> {
```

mitchmindtree (approved):
Nice, exciting stuff! :rocket:

I've left a couple of optional suggestions, but none are blockers.

forc-test/src/lib.rs
```diff
@@ -134,23 +143,21 @@ struct TestSetup {
 impl BuiltTests {
     /// Constructs a `PackageTests` from `Built`.
     ///
-    /// If the `built` is a workspace, `opts` is patched for the members of the workspace that
-    /// is a `Contract` so that only that contract is re-built.
-    ///
-    /// If the `built` is a package, `PackageTests::from_built_pkg` is used.
-    pub(crate) fn from_built(built: Built, opts: Opts) -> anyhow::Result<BuiltTests> {
+    /// Contracts are already compiled once without tests included to do `CONTRACT_ID` injection. `built_contracts` map holds already compiled contracts so that they can be matched with their "tests included" version.
```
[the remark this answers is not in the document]
kayagokalp: That's weird usually my nvim setup breaks the line as I type not sure how this turned out like this 🤔 Anyways I have some other refactorings in mind (some left-over comments from initial PR) so I will open a general clean-up PR. I could also address these reviews there

JoshuaBatty (approved):

PR #3730, opened 2023-01-09
Merge `Type` and `Aggregate` into a copyable `Type`

vaivaswatha:
This is the second step towards achieving #2819.

Combining Type and Aggregate into a single data structure enables offset computations for complex indexing (i.e., a nesting of arrays and structs) which will be needed when we introduce GEPs. The new function `get_indexed_type` does this computation.

The core of this PR is in `irtype.rs`. Everything else is just to accommodate that.


otrho (commented):
This is great, but... 🙂 

I assume you've introduced the 'basic' types thing to avoid adding the semi-redundant types to the `Instruction`s so that we can continue to pass an immutable context to `get_type()`?  It feels a bit clunky, like an unwanted dependency, where if we have a new instruction with a new type (who knows, `u128` or something?) then we need to be sure to add this type to the basic types too else it could panic.

I still prefer the method of putting the return type for each instruction in the `Instruction`, created with a mutable context when the instruction is created.  Then remove all the proposed `Type::get_X()` methods and rename the `Type::new_X()` to `get_X()`, having them take a mutable context and call `Type::get_or_create...`

But I could be convinced otherwise..?

vaivaswatha:
>where if we have a new instruction with a new type (who knows, u128 or something?) then we need to be sure to add this type to the basic types too else it could panic.

No it wouldn't panic, unless you add a `get_u128` method too AND you don't add that to the `create_basic_types`. and even then, it's a simple fix. Without a `get` method, you're forced to use a `new` method with a mutable context.

This is a whole lot simple than every instruction tracking its type (which is a redundant thing to do). Type is (most often) a static property of the op code, not dynamic.

vaivaswatha:
Moreover, it's also about having easy (i.e., with just an immutable context) to types, especially in places such as the verifier (or another analysis tomorrow, which doesn't modify things, just analyses). In all these cases, we want to be able to, in an easy way, handle different basic types without needing a mutable context, independent of instructions.

otrho:
Hrrm, I'm still not convinced that this is simpler than just adding the types to the instructions (as most of them carry their type already anyway).  It niggles at me that we're priming the context for some special cases, which as we add more functionality and more backend targets like EVM and MidenVM that type list could/should grow.

I say simpler because it's

* adding some obvious, invariant, semi-redundant type fields to some instructions.

versus

* having all the `Type::get_X()` _as well as_ `Type::new_X()`
* having those `Type::get_u64()`, `Type::get_u8()` special case methods when `Type::get_uint(n)` should be enough.
* having a list of primed types in `Type::create_basic...` and then priming the `Context` in `default()`.
* needing to be sure that if we allow other ways to create a `Context` in the future, via `new()` or something, that they all prime it too.

I'm not sure what you mean with the verifier.  It wouldn't need to create types via `Type::get_X()`, would it?  It's only asking if a type matches, via `Type::get_content()` or `Type::is_X()`, no?

vaivaswatha:
This is the smallest way to combine `Aggregate` into `Type`, which is the goal. Having types as part of Instructions, if we find that really necessary, can always be done later. It wouldn't be wise to touch `Instruction` too as part of this PR. That would be going into the same set of large unmanageable changes we stumbled upon previously.

otrho (approved):
OK, I still don't like priming the `Context` but approving for now.

vaivaswatha:
> OK, I still don't like priming the `Context` but approving for now.

I understand. I'd be happy to revisit this later once we're done with #2819.

Re-requesting for review as I had to fix a merge conflict.

mohammadfawaz (approved):
Approving as there seems to be agreement on this solution in the short term.

That being said, I actually agree with Toby that adding types to instructions is cleaner. In principle, adding a new instruction should be as simple as adding an enum variant for it and very little to no changes in `Instruction.rs` or `irtypes.rs`. We shouldn't need to rely on a function `get_type`, that will keep growing forever, to get the type of each instructions/intrinsic.

vaivaswatha:
>In principle, adding a new instruction should be as simple as adding an enum variant for it and very little to no changes in `Instruction.rs` `irtypes.rs.`

That's how it'll be now also. Adding a new instruction doesn't need changes in `type.rs`. It's only when we add a new `Type` that there's a difference. More :point_down: 

>We shouldn't need to rely on a function `get_type` that will keep growing forever, to get the type of each instructions/intrinsic.

In either approach, we will still need a `get_type` added if we add a new `Type` to the type system of the IR (not when we add new instructions or intrinsics). The only difference would be, if we add a new primitive `Type`, we'll have to add it to the list of primitive types that are [initialised](https://github.com/FuelLabs/sway/blob/9b85da99c3179251d3001aecfe258954b338de92/sway-ir/src/irtype.rs#L48) when `Context` is created. Otherwise we cannot give it a `get_type` method that takes in a non-mutable context. We can only provide a `new_type` (read as `get_or_create_type`) method that takes in a `&mut`.

vaivaswatha:
I realised that you meant the `get_type` method on Instruction (and not the get[*type*] methods that are there in irtype.rs). In which case the second part of my explanation doesn't address that.
If we indeed want an IR where types are stored as part of the IR, we can do that separately (as I mentioned earlier). But when Toby suggested it, that wasn't the goal. It was a workaround to not need `&mut` context to just get `Type`s. So having `Type`s as part of the IR, even though adding redundancy was a solution with that tradeoff.

PR #3731, opened 2023-01-09
Where clauses on `impl` blocks

### LINKED ISSUE #1162: Handle `where` clause for impl traits

emilyaherbert:
Closes #1162


esdrubal:
Resolved merging conflicts, and did some minor changes so the test case, with where clauses on impl blocks, added by Emily now passes.

IGI-111 (changes requested):
LGTM but I think we also want a negative test here to make sure the trait bounds are being processed correctly.

PR #3736, opened 2023-01-10
Allow LSP to start with partial config

sdankel:
Closes https://github.com/FuelLabs/sway.vim/issues/6

Makes it so the LSP will use partial config when provided, and defaults otherwise.

Confirmed it's working using partial config:
```
       init_options = {
         logging = { level = 'trace' },
         --inlayHints = { maxLength = 25, renderColons = true, typeHints = true },
         -- trace = {},
         --trace = { extension = true, server = 'messages' },
         -- debug = { showCollectedTokensAsWarnings = 'off' }
       },
```


JoshuaBatty (approved):
LGTM

mohammadfawaz (approved):

PR #3743, opened 2023-01-10
Disables usage of parenthesis in unit enum variants.

### LINKED ISSUE #3375: Should `Option::None()` compile?
The following code compiles:
```rust
script;

fn main() {
    let n: Option<u64> = Option::None(); 
}
```
but does not in Rust. The `()` syntax is somewhat justified here because the `None` variant does not require a value and this is equivalent to requiring a unit type. That being said, should we still disallow this, just like Rust?

### LINKED ISSUE #3585: Cannot use turbofish generic syntax
Take the following code:

```rust
script;

pub enum Option<T> {
    None: (),
    Some: T,
}

impl<T> Option<T> {
    pub fn none(self) -> Option<T> {
        Option::<T>::None
    }
}

fn main() {
    Option::Some(5).none();
}
```

Works fine in Rust but gives the following error in Sway:

```
error
  --> /sway/test-case/src/main.sw:10:17
   |
 8 | 
 9 |     pub fn none(self) -> Option<T> {
10 |         Option::<T>::None
   |                 ^^^ generics are not supported here
11 |     }
12 | }
   |
____
```

esdrubal:
With these changes it is no longer possible to do `Option::None()` instead it should be done with `Option::None`.

This changes also fixes enum turbofish without parenthesis so `Option::None::<T>` can be used.

Closes #3585
Closes #3375


sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -1161,6 +1162,7 @@ impl ty::TyExpression {
             (false, Some(func_decl), None, None) => {
                 warnings.append(&mut function_probe_warnings);
                 errors.append(&mut function_probe_errors);
+                let args = args.unwrap_or_default();
```
emilyaherbert, line 1165: Is there any case in which this assumption would be incorrect? i.e. is there any code that the user could write that would trigger this to fail silently?

sway-error/src/error.rs
```diff
@@ -416,6 +416,8 @@ pub enum CompileError {
     MoreThanOneEnumInstantiator { span: Span, ty: String },
     #[error("This enum variant represents the unit type, so it should not be instantiated with any value.")]
     UnnecessaryEnumInstantiator { span: Span },
+    #[error("`{ty}` is a unit enum variant, and does not take parentheses to be constructed.")]
```
emilyaherbert, line 419: Can we reword this to:

"The enum variant `{ty}` is of type `unit`, so its constructor does not take arguments or parenthesis. Try removing the `()`."

sway-core/src/semantic_analysis/ast_node/expression/typed_expression/enum_instantiation.rs
```diff
@@ -35,6 +35,19 @@ pub(crate) fn instantiate_enum(
         errors
     );
 
+    if type_engine
```
emilyaherbert, line 38: Would you add a line comment above this indicating what this block of code is doing?

sway-lsp/src/traverse/parsed_tree.rs
```diff
@@ -624,8 +624,9 @@ impl<'a> ParsedTree<'a> {
                     self.collect_type_arg(type_arg, &token);
                 }
 
-                for exp in args {
-                    self.handle_expression(exp);
+                let args_vec = args.clone().unwrap_or_default();
+                for exp in args_vec {
```
emilyaherbert, line 628: To eliminate the `.clone()`, change this block of code to something like:

```rust
args.map(|args_vec| {
  args_vec.iter().for_each(|exp| {
    self.handle_expression(&exp);
  })
});
```

emilyaherbert (changes requested):
Great solution to these two GH issues!

Would you add additional tests to cover these cases:
1. Test when enums are specified with their fully qualified path including the library names both with and without the `()` (`lib_a::inner_lib::my_enum::MyEnum::VariantA()`)
2. Test function calls and method calls that take no arguments both with and without the `()` and both with and without their fully qualified paths including the library names

Thanks!

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
esdrubal: It was possible to have args set to None and we could have functions called with no parenthesis. Thanks for noticing. Recent changes address this issue, and add a test case as well.

sway-error/src/error.rs
esdrubal: Applied the suggested change.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression/enum_instantiation.rs
esdrubal: Added a line comment.

sway-lsp/src/traverse/parsed_tree.rs
esdrubal: Followed the suggestion but then clippy told me to change it a bit.

esdrubal:
> * Test when enums are specified with their fully qualified path including the library names both with and without the `()` (`lib_a::inner_lib::my_enum::MyEnum::VariantA()`)
> * Test function calls and method calls that take no arguments both with and without the `()` and both with and without their fully qualified paths including the library names

Added the tests mentioned above in most recent change. Thanks to this I found an issue where a fully qualified path method with no parameters could be called without parenthesis. This issue was fixed with a new error thrown.

sway-error/src/error.rs
```diff
@@ -442,6 +444,8 @@ pub enum CompileError {
         expected: usize,
         received: usize,
     },
+    #[error("The function \"{method_name}\" was called without parenthesis.")]
```
emilyaherbert, line 447: Can you change this to:

"The function \"{method_name}\" takes no arguments, but was called without parenthesis. Try adding ()."

Thanks!

emilyaherbert (commented):
Great fixes! Just one slight rewording to the new error message, then I'm happy to approve 😄

emilyaherbert (dismissed):

sway-error/src/error.rs
esdrubal: Changed to "The function "{method_name}" was called without parenthesis. Try adding ()."
Removed the "takes no arguments" as the method may or may not take arguments. This error is thrown when the method is called with no arguments but it may still take arguments.

sway-error/src/error.rs
```diff
@@ -416,6 +416,8 @@ pub enum CompileError {
     MoreThanOneEnumInstantiator { span: Span, ty: String },
     #[error("This enum variant represents the unit type, so it should not be instantiated with any value.")]
     UnnecessaryEnumInstantiator { span: Span },
+    #[error("The enum variant `{ty}` is of type `unit`, so its constructor does not take arguments or parenthesis. Try removing the ().")]
```
mohammadfawaz, line 419:
```suggestion
    #[error("The enum variant `{ty}` is of type `unit`, so its constructor does not take arguments or parentheses. Try removing the ().")]
```

sway-error/src/error.rs
```diff
@@ -442,6 +444,8 @@ pub enum CompileError {
         expected: usize,
         received: usize,
     },
+    #[error("The function \"{method_name}\" was called without parenthesis. Try adding ().")]
+    MissingParenthesisForFunction { span: Span, method_name: Ident },
```
mohammadfawaz, line 448:
```suggestion
    #[error("The function \"{method_name}\" was called without parentheses. Try adding ().")]
    MissingParenthesesForFunction { span: Span, method_name: Ident },
```

sway-error/src/error.rs
```diff
@@ -813,10 +817,12 @@ impl Spanned for CompileError {
             ImportMustBeLibrary { span, .. } => span.clone(),
             MoreThanOneEnumInstantiator { span, .. } => span.clone(),
             UnnecessaryEnumInstantiator { span, .. } => span.clone(),
+            UnitVariantWithParenthesesEnumInstantiator { span, .. } => span.clone(),
             TraitNotFound { span, .. } => span.clone(),
             InvalidExpressionOnLhs { span, .. } => span.clone(),
             TooManyArgumentsForFunction { span, .. } => span.clone(),
             TooFewArgumentsForFunction { span, .. } => span.clone(),
+            MissingParenthesisForFunction { span, .. } => span.clone(),
```
mohammadfawaz, line 825:
```suggestion
            MissingParenthesesForFunction { span, .. } => span.clone(),
```

sway-core/src/semantic_analysis/ast_node/expression/typed_expression/function_application.rs
```diff
@@ -22,6 +22,15 @@ pub(crate) fn instantiate_function_application(
     let decl_engine = ctx.decl_engine;
     let engines = ctx.engines();
 
+    if arguments.is_none() {
+        errors.push(CompileError::MissingParenthesisForFunction {
```
mohammadfawaz, line 26:
```suggestion
        errors.push(CompileError::MissingParenthesesForFunction {
```

mohammadfawaz (commented):
Looks great! Just some nits and then I'll approve.

mohammadfawaz (approved):

JoshuaBatty (approved):

PR #3744, opened 2023-01-10
Perform monomorphization and constraint solving after type checking

### LINKED ISSUE #1267: Consider refactoring monomorphizing into a separate step following `type_check`?
At the moment, the process of monomorphizing generics is intertwined with the type-checking process.

I'm curious whether this is out of necessity, i.e. do we need to monomorphize in order to successfully type check? Or is it more the case that we _can_ do it at the same time so we might as well?

In the case that it's the latter, I wonder if it might be worth splitting it into a separate step that immediately follows type checking?

The type checking process is quite complex, and as a newcomer to `sway-core` I've found grokking how the namespace is updated to be particularly tricky (could be just me!).

Building a mental model of how the namespace is updated requires following two axes:

1. through the creation of new scopes and populating the scope's contents and
2. subtle inner mutation within methods like `namespace.find_method_for_type` due to the inner `resolve_type_with_self` call that may `monomorphize` internally.
 
`1.` is quite intuitive to follow, but `2.` seems a little shoe-horned in and is making #1213 particularly finicky (though I think that PR's still worth tackling before more global mutation sites start cropping up).

I'd imagine having distinct steps might make debugging issues related to type-checking and monomorphizing a little easier too? This is purely speculation though as I'm still new to `sway-core`.

I'm also unsure just how much work would be involved in such a refactor. I'd imagine at the very least the `Namespace::resolve_*` methods, and the semantics around `TypeId`s would need some rethinking.

---

Ahh I just noticed https://github.com/FuelLabs/sway/issues/862 which seems related.

### LINKED ISSUE #2636: Introduce a monomorphization cache

### LINKED ISSUE #3403: Consider reframing internal compiler typing logic to be "locally-based constraint solving" instead of "globally-based"
~~Previously the compiler has relied on a global `TypeEngine` in order to complete type inference. Now that there is a PR to remove this (#3353), we can begin to think about introducing an internal change to the how the `TypeEngine` that will make understanding and contributing to the type system much easier and will reduce the scope of new changes significantly.~~

~~Instead of approaching type constraint solving "globally", we should change it to be more "local" (I'm making these phrases up as a go so I'll try my best to explain). Essentially what I'm trying to say is that exposing all types across all scopes of the program is adding unnecessary complexity. For example, when doing type inference in a function body, the types that you need are 1) from the scope surrounding the function body and 2) from the scope of the function body itself. You don't need to drag around the types from inside the declaration in the previous declaration block, which is what the global `TypeEngine` does.~~

~~I think that this issue manifests in the fact that local constraint solving is not fully solved right now. For example, #3324 demonstrates that there is an issue resolving types with match expressions (although there is a fix in #3348).~~

Notes:
- https://rustc-dev-guide.rust-lang.org/ty.html#question-why-not-substitute-inside-the-adtdef
- https://rustc-dev-guide.rust-lang.org/generics.html
- https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/subst/type.SubstsRef.html

emilyaherbert:
This PR:
1. Make the `DeclId`s unique with a 1:1 ratio for data structure definition. This means that we can compare `DeclId`s directly, reduces the number of `DeclId`s that we produce, and reduces the number of times that we need to perform lookups in the `DeclEngine`
    a. Modify `PartialEqWithEngines` to no longer compare the `DeclId`s in the AST directly
    b. Remove the "parents" abstraction in the `DeclEngine`
2. Refactors monomorphization into a separate step following type checking
    a. This include trait constraint solving

Closes #3403 
Closes #1267
Closes #2636

Blocked by:
- #4082
- #4089
- #3762
- #4213


emilyaherbert:
> Do you think we can keep the parents abstraction? I am using it in https://github.com/FuelLabs/sway/pull/3878
> The problem there is that I need to go from the monomorphized type back to the parent non-monomorph type. Or should we come up with something else for that?

Thanks for letting me know @tritao! Currently I'm planning to remove the parents abstraction as it will no longer be necessary for monomorphization or type checking. I think we could either come up with something else clever, or in the interim we could construct a slimmed down version of the parents abstraction during monomorphization.

PR #3755, opened 2023-01-11
fix: forc checks validity of parent manifests before assuming it is a workspace manifest

### LINKED ISSUE #3535: When checking for a parent workspace manifest, `forc` gives a confusing error if it fails to parse a parent manifest as a workspace
Raised by @segfault-magnet in slack:

> Hey team. Not sure whether to write here or in tooling. If I build a sway project located in /home/some_user/some_folder/some_folder_2/the_project it will fail if there is a Forc.toml in any of the parent directories. All my sway builds started failing due to the existence of ~/Forc.toml. If I remove it the builds pass again. The errors were something like:
> 
> Error: failed to parse manifest:"...,
> Error: failed to parse manifest: missing field `members` at line 7 column 1.

We should either:

1. Provide a much more concise explanation of the error or
2. Warn that a parent `Cargo.toml` was detected that was not a valid workspace or
3. Silently ignore parent manifests that are _not_ workspaces with a members set containing the current package.

We should investigate cargo's behaviour here, it appears to have changed recently as previously cargo used to provide a warning in this situation (option 2), whereas currently it appears to silently ignore (option 3).

kayagokalp:
I ended-up implementing the 3rd option discussed in the relevant issue: non workspace manifests residing at a parent directory:

> Silently ignore parent manifests that are not workspaces with a members set containing the current package.

There is one catch though, if we filter the ones that does not declare the current package as a member we are also losing a valuable error message which states the workspace manifest file does not list the current package as a member. Should we do that? cc @mitchmindtree.

Currently this fixes the situation where there is a pkg manifest in some parent directory, just like the cause of the original issue of @segfault-magnet.

again this would be a little bit clearer if we had https://github.com/FuelLabs/sway/issues/3283 so might refactor after that
closes #3535

### DIFF
--- a/forc-pkg/src/manifest.rs
+++ b/forc-pkg/src/manifest.rs
@@ -726,7 +726,18 @@ impl WorkspaceManifestFile {
             // Check if the found manifest file is a workspace manifest file or a standalone
             // package manifest file.
             let possible_path = possible_manifest_dir.join(constants::MANIFEST_FILE_NAME);
-            Self::from_file(possible_path).is_ok()
+            // We should not continue to search if the given manifest is a workspace manifest with
+            // some issues.
+            //
+            // If the error is missing field `workspace` (which happens when trying to read a
+            // package manifest as a workspace manifest), look into the parent directories for a
+            // legitimate workspace manifest. If the error returned is something else this is a
+            // workspace manifest with errors, classify this as a workspace manifest but with
+            // errors so that the erros will be displayed to the user.
+            Self::from_file(possible_path)
+                .err()
+                .map(|e| !e.to_string().contains("missing field `workspace`"))
+                .unwrap_or_else(|| true)
         })
         .ok_or_else(|| manifest_file_missing(manifest_dir))?;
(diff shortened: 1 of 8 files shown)


forc-pkg/src/manifest.rs
```diff
@@ -722,8 +722,24 @@ impl WorkspaceManifestFile {
...
+            // some issues.
+            //
+            // If the error is missing field `workspace` (which happens when trying to read a
+            // package manifest as a workspace manifest), look into the parent directories for a
+            // legitimate workspace manifest. If the error returned is something else this is a
+            // workspace manifest with errors, classify this as a workspace manifest but with
+            // errors so that the erros will be displayed to the user.
+            Self::from_file(possible_path)
+                .err()
+                .map(|e| !e.to_string().contains("missing field `workspace`"))
```
mitchmindtree, line 739: This is certainly a bit messy, but should do the trick until we address #3283 :+1:

mitchmindtree (approved):

bingcicle (approved):

Issue #1667, opened 2022-05-25
Address the problem with `OrdEq` and bogus dead code warnings discovered during #1600


otrho:
There are a couple of notes in the PR #1600 which mustn't be forgotten.

mohammadfawaz:
Closing in favour of https://github.com/FuelLabs/sway/issues/1682

@otrho are there other issues here that need to be addressed other than the bogus dead code warnings?

adlerjohn:
@mohammadfawaz yes, this one https://github.com/FuelLabs/sway/pull/1600#discussion_r881196724

mohammadfawaz:
Have we decided that this is blocked by the [declarations engine](https://github.com/FuelLabs/sway/issues/1692)?

PR #3760, opened 2023-01-11
Added non-strict comparison functions

chlenc:


otrho:
Hi, @chlenc.  The `le` and `ge` methods are a part of the `OrdEq` trait.  @rostyslavtyshko Do you recall why we haven't added it already?

mohammadfawaz:
> Hi, @chlenc. The `le` and `ge` methods are a part of the `OrdEq` trait and should be added that way. @rostyslavtyshko Do you recall why we haven't added it already?

Unfortunately, we are blocked by #1667 if I'm not mistaken. That's why we haven't done added `ge` and `le` to these two libraries yet. I suggest that we wait until #1667 is fixed and doing this the right way, especially that there is are temporary workarounds.

PR #3764, opened 2023-01-12
Bug fix to type inference with placeholder types for nested types

### LINKED ISSUE #3741: Type inference fails for nested types
Some tests in the Rust SDK fail to compile with the release of forc 0.33.0

Here is a minimal example:

```rust
fn main() -> bool {
   let mut exp_vec_in_a_vec_in_a_struct_in_a_vec = Vec::new();

   let mut inner_vec_1 = Vec::new();

   let inner_inner_vec_1 = vec_from([0, 1, 2]);
   inner_vec_1.push(inner_inner_vec_1);

   exp_vec_in_a_vec_in_a_struct_in_a_vec.push(SomeStruct { a: inner_vec_1 });

   true
}
```

Which leads to:
```
inner_vec_1 : This parameter was declared as type Vec<u32>, but argument of type SomeStruct<Vec<Vec<u32>>> was provided.
```

The example complies without issues if we manually set the type of `exp_vec_in_a_vec_in_a_struct_in_a_vec` to `Vec<SomeStruct<Vec<Vec<u32>>>>`

emilyaherbert:
This PR fixes a bug relating to type substitution for placeholder types. The fix was to update the code path for placeholder type substitution from "no substitution" to "substitute if a match is found".

This PR also changes the internal data structure of the `TypeMapping` type from a `Vec<(TypeId, TypeId)>` to a `BTreeMap<TypeId, TypeId>` because during debugging I noticed that there were duplicates and this could lead to wasted compute.

Closes #3741


mohammadfawaz (approved):
Awesome, thanks for the quick fix!

PR #3766, opened 2023-01-12
Update error reporting for mismatched ABI method

### LINKED ISSUE #3740: Fix the error message for mismatched return types
image

The error should be something like “Mismatched return types”, but instead it says the function is missing entirely.

Considering we don't allow multiple functions with the same name we should at least say there's a mismatch rather than that it's missing.

emilyaherbert:
This PR fixes this bug by creating a skeleton method definition in the event that type checking the method produces an error.

Closes #3740


otrho (commented):
Can you also add a test or two for mismatching argument types?  Or do we already have that covered?

emilyaherbert:
> Can you also add a test or two for mismatching argument types? Or do we already have that covered?

There might already be one but I'll add it as a test case as well.

otrho (approved):

Issue #3746, opened 2023-01-10
Change compiler error/warning to the span of the name rather than whole block


sdankel:
Currently some compiler warnings/errors return the span of the entire block. For example, if there's a an unused function, it returns the span of the entire function including comments. Ideally, it would return the span of the function name (ident) so that the editor/terminal only highlights the function name.

PR #3769, opened 2023-01-12
Make compiler dead code warning spans use name instead of entire block

sdankel:
Closes https://github.com/FuelLabs/sway/issues/3746

This PR changes the highlighted portion for dead code warnings to show only the declaration name (ident) rather than the entire code block for the LSP.

It also adds some test helper functions:
- an async function that spawns a thread, uses the server socket to capture server-initiated requests, and asserts that the messages received match the expected messages
- a simple function to read a json file into memory to be used as a test fixture

And adds a dead code warning for Enum, allowing us to highlight only the enum name, rather than the entire block.

Before:

image

After:

image


sway-lsp/src/utils/test.rs
```diff
@@ -16,8 +20,8 @@ pub(crate) fn e2e_test_dir() -> PathBuf {
         .join("struct_field_access")
 }
 
-pub(crate) fn sway_example_dir() -> PathBuf {
```
sdankel, line 19: This wasn't being used

sway-lsp/src/core/token_map.rs
```diff
@@ -22,18 +22,16 @@ impl TokenMap {
         &'s self,
         uri: &'s Url,
     ) -> impl 's + Iterator<Item = (Ident, Token)> {
-        self.iter()
-            .filter(|item| {
-                let (_, span) = item.key();
-                match span.path() {
-                    Some(path) => path.to_str() == Some(uri.path()),
-                    None => false,
+        self.iter().flat_map(|item| {
```
sdankel, line 25: Slightly easier to read with flat_map, imho

mohammadfawaz (approved):
Changes in `sway-core` seem reasonable to me! This does indeed the behavior of `rustc` which I really like.

sway-lsp/test/fixtures/diagnostics/dead_code/Forc.toml
```diff
@@ -0,0 +1,7 @@
+[project]
+authors = ["Sophie"]
```
otrho, line 2: Just noticed this.  We use `authors = ["Fuel Labs <contact@fuel.sh>"]` in the compiler tests, which is automatic if you set `FORC_INIT_MANIFEST_AUTHOR='Fuel Labs <contact@fuel.sh>'` in your env.

John used to be really strict with it but perhaps it isn't so important these days.

otrho (dismissed):

sway-lsp/test/fixtures/diagnostics/dead_code/Forc.toml
mohammadfawaz: Actually yeah it's probably better to be consistent and use `Fuel Labs`.

sway-lsp/src/utils/test.rs
JoshuaBatty: Can we put this function back in? Although it's not being used in any current CI tests, it is handy to have when testing locally.

sway-lsp/test/fixtures/diagnostics/dead_code/Forc.toml
JoshuaBatty: Agree with the above.

JoshuaBatty (changes requested):
This is really great work @sdankel! just one tiny nit and I'm happy to approve.

mohammadfawaz (approved):

JoshuaBatty (approved):
:shipit:

Issue #3077, opened 2022-10-19
Contract ID constants should not use `ConfigTimeConstant` and should construct AST directly


kayagokalp:
With #3016, we will have `[contract-dependencies]` section for the manifest file which is used to automatically insert `CONTRACT_ID`s to the namespace while doing so it is using config time constants and we can improve that section such that related AST node can be constructed directly.

Related to https://github.com/FuelLabs/sway/pull/3016#discussion_r998914116

mitchmindtree:
I just got the all clear from @sezna a little earlier that we are free to remove the old `ConfigTimeConstant`s now.

For anyone picking up this issue (whether it's you @kayagokalp or someone else), feel free to remove the `ConfigTimeConstant` type stuff at the same time and strip the logic down to just what we need for `CONTRACT_ID`s.

bingcicle:
After some initial attempt to replace the injection of contract IDs, I decided to drop this issue for now - and I think this also deserves a rename, since `ConfigTimeConstant` is no longer relevant following #4298.

The issue here is the 'hack' of parsing and injecting of a `CONTRACT_ID` as a `ConstDeclaration` into the dependency namespace's `SymbolMap` to allow a contract access to its own `CONTRACT_ID` or other dependent `CONTRACT_ID`s. This is not ideal, but it seems like we do not really have a viable alternative at the moment that can allow `forc-pkg` to relieve the compiler of this process.

To expand on the end goal, there are 2 situations where the compiled contract IDs are required:

1) when declared under the `[contract-dependencies]` table in `Forc.toml`, and
2) when running `forc test`.

Initially I attempted a solution which kind of compiles a 'virtual' library that declares only the contract ID post-compilation of a contract, and compiles that 'virtual' library itself. This works for use case 1) above of course, but it doesn't really solve the issue of 2) - which, after some investigation, would still require us to do some hacky work in the namespace itself to inject some sort of constant declaration to let the current contract gain access to its own compiled contract ID.

Having to maintain 2 separate ways of contract ID support seems like unnecessary complexity at this point - even though the current method may be 'hacky', but I think there aren't any better solutions atm :/

Instead of creating a new issue, I'm changing the name of the issue so that we still have context of what went on in the past, and I am un-assigning myself for now.

bingcicle:
[...]

PR #3770, opened 2023-01-12
Introduce a way to inject items into project namespaces through `forc-pkg`

### LINKED ISSUE #3763: Provide a way to inject items into project namespaces through `forc-pkg`
related to https://github.com/FuelLabs/sway/pull/3729#pullrequestreview-1244708714

kayagokalp:
This PR, refactor the `BuildPlan` generation a little to be able to re-use some of the parts from other crates that requires to create `BuildPlan` from build opts.

On top of that this PR also introduces a way to inject list of items into the namespace of any member of the workspace. 

This is done to unblock #3729

closes #3763.


forc-pkg/src/pkg.rs
```diff
@@ -2490,16 +2518,12 @@ fn build_profile_from_opts(
...
-        PathBuf::from(path)
-    } else {
-        std::env::current_dir()?
-    };
-
+/// Builds a project with given BuildOptions while injecting provided items to the namepsace of
+/// relevant package's namespace.
+pub fn inject_and_build_with_options(
+    build_options: BuildOpts,
+    inject_map: HashMap<NodeIx, Vec<(String, ConfigTimeConstant)>>,
```
mitchmindtree, line 2525: Perhaps we can include the `inject_map` in the `BuildOpts` to remove the need for an extra constructor function?

Can we really use `NodeIx` safely as the key? Isn't it possible for the `NodeIx` to change if members are added/removed to/from the workspace? Curious to see what usage looks like. Is it possible we want to use `Pkg` as a key instead?

Also, perhaps we could use some type aliases to make this type a little easier to read? e.g.

```rust
type ConstName = String;
type ConstInjectionMap = HashMap<NodeIx, Vec<(ConstName, ConfigTimeConstant)>>;
```

We should really change the name from `ConfigTimeConstant` to `Constant` or something, especially now that *actual* config time constants are being introduced (#3536). Maybe it's time to prioritize #3077?

mitchmindtree (changes requested):
Looking good for the most part! Just a few nits

forc-pkg/src/pkg.rs
kayagokalp: It seems to me that `NodeIx` stays the same but using `pkg` might be better here as it would easier to understand. I planning to tackle #3077 right after #3729 👍

kayagokalp: I ended up using `Pinned` since I need the get a `Path` reference to convert `Pinned` to `Pkg` as `Pinned` is the struct that I can directly reach from `BuildPlan`. This feels more coherent to me as well since we are injecting stuff to the nodes of `BuildPlan`'s `graph` (which are `Pinned`s). To be able to use the new injection map easily from outside of `forc-pkg` I also added a util function that returns an iterator over `Pinned`s of member packages.

mitchmindtree: Nice, this looks a little more fool-proof! Looking forward to seeing how this gets used in `forc-test` :ok_hand:

mitchmindtree (approved):

bingcicle (approved):

PR #3773, opened 2023-01-13
Fix command for updating 'forc' from cargo

mehtavishwa30:
The **Update 'forc' from cargo** section in the installation docs currently lists the following incorrect command:
`cargo install forc fuel-core`

Instead, it should be as follows:
`cargo update forc fuel-core`

This PR replaces `install` with `update` to add the correct command for updating 'forc' using cargo.


kayagokalp (approved):
Thanks for the contribution, LGTM!

mohammadfawaz (approved):

PR #3777, opened 2023-01-13
Error out when `Vec` is used in a `storage` block

### LINKED ISSUE #3305: Error out when `Vec` is used in a `storage` block
This can probably be done by checking if a type has `raw_ptr` in it. We have a similar check for disallowing types containing `raw_ptr` from being returned from scripts.

emilyaherbert:
This PR creates an error when `Vec` is used in a `storage` block by checking for instances in `raw_ptr`. Right now there is only one simple test, so I can add more if you feel it necessary.

Closes #3305


otrho (approved):
LGTM

mohammadfawaz (approved):

PR #3778, opened 2023-01-15
Add index file navigation to `forc doc`

### LINKED ISSUE #3170: `forc doc` - add a sidebar for quick navigation
Currently, `forc doc` presents items in a single page. E.g. here's the doc root of `std`:

Screenshot from 2022-10-27 23-27-35

We should present a sidebar for easier navigation between items / sections at a higher level like `docs.rs`, e.g.

Screenshot from 2022-10-27 23-37-32

cc @eureka-cpu as I feel like I remember seeing something about a sidebar in #2869? Is this maybe implemented but just not showing up in certain contexts?

### LINKED ISSUE #3658: Add parent module index file creation
Currently we only have the `AllDoc` and `RenderedDocument`s, but we should also be creating the index files for each of the parent modules. 

This is also related to #3173

### LINKED ISSUE #3660: Refactor `AllDocIndex`'s `AllDocs` hyperlink variable instantiation
The link variable creation for each type, e.g `struct`, `enum`, etc. is messy and difficult to read. This could be simplified by integrating it into some method or field maybe.

https://github.com/FuelLabs/sway/blob/2faff70d3d452181dee771b0c2c82cee48b33c6c/forc-plugins/forc-doc/src/render.rs#L490

eureka-cpu:
Closes #3658 
Closes #3170 
Closes #3660 

Updated video using `std`:

https://user-images.githubusercontent.com/57543709/213319718-c8dd9d40-a29b-4598-8db5-89c8d6a5a837.mov

Here's what the files look like, the only difference to rust is that they create a folder for the parent module as where we omit it. This will likely need to change once we have support for workspaces.
Screen Shot 2023-01-17 at 7 51 32 PM


mitchmindtree (commented):
Nicce, digging the all items page too - it's cool to see an overview!

I just did a quick tests with `std` and noticed that on the main index.html many of the modules show up duplicated:

Screenshot from 2023-01-18 18-39-05

Any ideas what might be going on here?

eureka-cpu:
> Nicce, digging the all items page too - it's cool to see an overview!
> 
> I just did a quick tests with `std` and noticed that on the main index.html many of the modules show up duplicated:
> 
> Screenshot from 2023-01-18 18-39-05
> 
> Any ideas what might be going on here?

I'll check into this tomorrow morning, I may have a clue to what might be the cause

eureka-cpu:
@mitchmindtree just ran it and I'm certain the issue has to do with how I collected child modules, likely the issue didn't show up in my other tests because there were only singular references to any one module.

forc-plugins/forc-doc/src/doc.rs
```diff
@@ -21,23 +21,31 @@ pub(crate) struct Document {
 }
 impl Document {
     /// Creates an HTML file name from the [Document].
-    pub(crate) fn html_file_name(&self) -> String {
+    pub(crate) fn html_filename(&self) -> String {
```
mitchmindtree, line 24: Fwiw, the Rust std uses an underscore separator, e.g. https://doc.rust-lang.org/std/path/struct.Path.html#method.file_name, but I don't really mind either way as long as we're consistent :+1:

mitchmindtree (approved):
Looking good!

It would be cool to have the start of the docstring to the right of the item in each of the module indexes (like in docs.rs), e.g.

Screenshot from 2023-01-19 13-14-25

It would be cool to have _The ContractId type, a struct wrapper around the inner b256 value._ show up to the right of the `ContractId` listing here.

do we have an issue for this yet?

Also, I notice that Rust orders items and modules alphabetically - perhaps we should do this to? I think it does make it a little easier to navigate.

Happy for these to be addressed in follow-up either way  :+1:

forc-plugins/forc-doc/src/doc.rs
eureka-cpu: Gotcha, as I was working through this it just bothered my OCD brain for some reason haha

eureka-cpu:
> Looking good!
> 
> It would be cool to have the start of the docstring to the right of the item in each of the module indexes (like in docs.rs), e.g.
> 
> It would be cool to have _The ContractId type, a struct wrapper around the inner b256 value._ show up to the right of the `ContractId` listing here. Do we have an issue for this yet?
> 
> Also, I notice that Rust orders items and modules alphabetically - perhaps we should do this too? I think it does make it a little easier to navigate.
> 
> Happy for these to be addressed in follow-up either way 👍

Just filed #3820 and I'll look into it as a follow up!

JoshuaBatty (approved):

PR #3785, opened 2023-01-16
fix(formatter): fix else indentation after comment between if-else

### LINKED ISSUE #2649: Problem with newline not being inserted after comments when a token follows it

bingcicle:
closes #2649

Added a few test cases with comments in between, including:
- single-line comment with overindentation,
- multi-line comment with underindentation,
- single-line comment, from inline to multiline if-else

Updated `get_post_context()` with a better way to format `else` after comment. This is a bit of a hacky fix but since the post-context concept is currently only used for formatting `else` blocks after comment, I was thinking this might tide us through for now (and the test suite passes as well).

I wrote a long-form comment explaining what the new get_post_context() does there especially because fixing the off-by-one error with subtracting by 1 is going to look like a hacky solution (and it is, unfortunately)


kayagokalp (approved):
utACK ✅

bingcicle:
So this doesn't actually fix the case where the comments themselves are unformatted. Example:

```rust
contract;

    fn is_blue() -> bool {
        if self == PrimaryColor::Blue {
            true
        }
        // TODO
            // remove this else when exhaustive ifs are checked for
    else {
            false
        }
    }
}"#;
```

I'd say this PR is still worth merging since it fixes the linked issue, and I guess misindented comment would be rarer than a misindented `else` block. I'll open a new issue for the above.

swayfmt/src/formatter/mod.rs
```diff
@@ -1375,6 +1375,115 @@ fn foo() {}
...
+    #[test]
+    fn test_comment_between_if_else_overindented() {
+        let sway_code_to_format = r#"contract;
+
+impl MyContract for Contract {
+    fn is_blue() -> bool {
+        if self == PrimaryColor::Blue {
+            true
+        }
+        // TODO remove this else when exhaustive ifs are checked for
```
sdankel, line 1395: It seems like the new logic is assuming that the comment is correctly aligned. Do we have a test case where both the comment and the else block are misaligned? Something like:

```
    fn is_blue() -> bool {
        if self == PrimaryColor::Blue {
            true
        }
                      // TODO remove this else when exhaustive ifs are checked for
                else {
            false
        }
    }
}"#;
```

sdankel (approved):

swayfmt/src/formatter/mod.rs
bingcicle: Can confirm that this is still an issue, output:

```rust
    fn is_blue() -> bool {
        if self == PrimaryColor::Blue {
            true
        }
                // TODO remove this else when exhaustive ifs are checked for
                else {
            false
        }
    }
}"#;
```

PR #3788, opened 2023-01-16
PassManager: Refactor / separate into a module

vaivaswatha:
This PR moves code in preparation for a full fledged pass manager for the IR.

Reference: https://github.com/FuelLabs/sway/discussions/3596


mohammadfawaz (approved):
Great! Thanks.

PR #3791, opened 2023-01-16
fuels-types to fuel-abi-types

### LINKED ISSUE #3638: Remove the dependency on `fuels` in `sway-core` and `forc-pkg`
Ref https://github.com/FuelLabs/fuels-rs/issues/752

Remove the dependency of `sway-core` and `forc-pkg` on `fuels-rs` and replace it with a dependency on the shared repo `fuels-types`. This may _just_ work given that the dependency is already on a package named `fuels-types` (the one from `fuels-rs` currently).

Salka1988:
replaced fuels-types with fuel-abi-types
closes #3638

### DIFF
--- a/forc-plugins/forc-client/Cargo.toml
+++ b/forc-plugins/forc-client/Cargo.toml
@@ -19,7 +19,7 @@ fuel-gql-client = { version = "0.15", default-features = false }
 fuel-tx = { version = "0.23", features = ["builder"] }
 fuels-core = "0.33"
 fuels-signers = "0.33"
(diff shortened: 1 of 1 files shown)


Salka1988:
fuels_types::bech32::Bech32Address is not exported in fuel-abi-types

mohammadfawaz:
Right yeah, that particular dependency is fine to have. What's really important to keep `forc-pkg` and `sway-core` (and everything they depend on) free of `fuels-rs` dependencies.

PR #3792, opened 2023-01-17
Display documentation for sway keywords

### LINKED ISSUE #3515: Add support for displaying documentation for sway keywords
In rust, you can hover over keywords and documentation appears with an explanation for what the keyword means in the language. 

pub keyword

It would be cool if we can also offer the same functionality. We could mostly copy this information directly from rust and modify it where required. Instead of adding this documentation to `sway-ast` it probably is easier to just add it to the language server itself.

JoshuaBatty:
This PR provides documentation when hovering over keywords in sway. I copied over the documentation for keywords that are shared with rust and tweaked them so they made sense to sway [from here](https://github.com/rust-lang/rust/blob/master/library/std/src/keyword_docs.rs
).

It would be great to get some compiler engineers to look over the documentation to make sure nothing is being incorrectly defined. 

There are still a number of keywords that need documentation written up for them but I've opened up  #3939 so these can be added by someone on the compiler team at a later date. 

will close #3515


sway-lsp/src/capabilities/hover.rs
```diff
@@ -14,9 +16,31 @@ use sway_types::{Ident, Span, Spanned};
...
     let range = get_range_from_span(&ident.span());
+
+    // check if our token is a keyword
+    if token.kind == SymbolKind::Keyword {
+        let name = ident.as_str();
+        let documentation = keyword_docs.get(name).unwrap();
+        let prefix = format!("\n```sway\n{name}\n```\n\n---\n\n");
+        let formatted_doc = format!("{prefix}{documentation}");
+        let content = Markup::new()
+            .maybe_add_sway_block(None)
```
sdankel, line 35: This line looks like a no-op

sway-lsp/src/utils/keyword_docs.rs
```diff
@@ -0,0 +1,828 @@
...
+// Ported from https://github.com/rust-lang/rust/blob/master/library/std/src/keyword_docs.rs
+
+/// Documentation for sway keywords.
+/// Primarily used for showing documentation on LSP hover requests.
+#[derive(Debug)]
+pub struct KeywordDocs {
+    /// Key = keyword
+    /// Value = documentation
+    inner: HashMap<String, String>,
+}
```
sdankel, line 17: Could we use the same pattern as token_map here?

```suggestion
pub struct KeywordDocs(HashMap<String, String>);
```

sway-lsp/src/utils/keyword_docs.rs
```diff
@@ -0,0 +1,828 @@
...
+        };
+
+        let fn_keyword: ItemMod = parse_quote! {
+            /// Functions are the primary way code is executed within Sway. Function blocks, usually just
+            /// called functions, can be defined in a variety of different places and be assigned many
+            /// different attributes and modifiers.
+            ///
+            /// Standalone functions that just sit within a module not attached to anything else are common,
+            /// but most functions will end up being inside [`impl`] blocks, either on another type itself, or
+            /// as a trait impl for that type.
```
sdankel, line 265: The formatting gets messed up on this line apparently because the line is 1 character too long. Moving "or" to the next line fixes it.


image

```suggestion
            /// but most functions will end up being inside [`impl`] blocks, either on another type itself,
            /// or as a trait impl for that type.
```

image

Is there a way to detect this with clippy/rustfmt? It could be a problem in other places.

sway-lsp/src/utils/keyword_docs.rs
```diff
@@ -0,0 +1,828 @@
...
+        };
+
+        // TODO
+        let configurable_keyword: ItemMod = parse_quote! {
+            mod configurable_keyword {}
+        };
+
+        let mut keyword_docs = HashMap::new();
+
+        let keywords = vec![
```
sdankel, line 752: It would be cool if there was a way to ensure that this list of keywords matches the keywords defined in `keywords.rs`, so we either get a compilation error or a test failure when a new keyword is added without docs.

sway-lsp/src/capabilities/hover.rs
sdankel, line 35: This line looks like a no-op

sway-lsp/src/utils/keyword_docs.rs
sdankel: Maybe it's not possible with the way macros work...

sway-lsp/src/utils/keyword_docs.rs
JoshuaBatty: sure, i'm with that.

sway-lsp/src/capabilities/hover.rs
JoshuaBatty: good spot! :)

sway-lsp/src/utils/keyword_docs.rs
JoshuaBatty: great call, i'll add a test to ensure this is always the case.

sway-lsp/src/utils/keyword_docs.rs
JoshuaBatty: thanks for finding this, turns out there was a bug in the `extract_lit` function. I've updated the logic so this can't happen anymore. :)

sway-lsp/src/utils/keyword_docs.rs
```diff
@@ -0,0 +1,850 @@
...
+            mod continue_keyword {}
+        };
+
+        // TODO
+        let str_keyword: ItemMod = parse_quote! {
+            mod str_keyword {}
+        };
+
+        // TODO
+        let for_keyword: ItemMod = parse_quote! {
```
sdankel, line 689: Are the rust docs for `for` different from sway?

sway-lsp/src/utils/keyword_docs.rs
```diff
@@ -0,0 +1,850 @@
...
+            /// is added as the type describes, for example `Option::Some(123)`. The same follows with
+            /// struct-like variants, with things looking like `ComplexEnum::LotsOfThings { usual_struct_stuff:
+            /// true, blah: 245, }`. Empty Enums are similar to [`!`] in that they cannot be
+            /// instantiated at all, and are used mainly to mess with the type system in interesting ways.
+            ///
+            /// [ADT]: https://en.wikipedia.org/wiki/Algebraic_data_type
+            mod enum_keyword {}
+        };
+
+        let self_keyword: ItemMod = parse_quote! {
```
sdankel, line 209: I see the hover for `self` when it's a function parameter, but not when it's used inside of a function block (e.g. `self.bar`)

image

sway-lsp/src/utils/keyword_docs.rs
```diff
@@ -0,0 +1,850 @@
...
+            ///     fn get_number(self) -> u32 {
+            ///         self.number
+            ///     }
+            /// }
+            /// ```
+            mod impl_keyword {}
+        };
+
+        let const_keyword: ItemMod = parse_quote! {
+            /// ## Compile-time constants.
```
sdankel, line 379: nit: the formatting here with a title is different from the others

sway-lsp/src/utils/keyword_docs.rs
```diff
@@ -0,0 +1,850 @@
...
+            ///     i *= 2;
+            ///     if i == 64 {
+            ///         break; // Exit when `i` is 64.
+            ///     }
+            /// }
+            /// ```
+            mod while_keyword {}
+        };
+
+        let true_keyword: ItemMod = parse_quote! {
```
sdankel, line 613: I don't see hover docs for true/false
image

sdankel (approved):
Looks great! My comments are all nits/can be addressed later.

sway-lsp/src/utils/keyword_docs.rs
```diff
@@ -0,0 +1,850 @@
...
+            let s = s.trim_start_matches("r\""); // remove the r" sequence at the beginning
+            let s = s.trim_end_matches('\"'); // remove the " at the end
+            res.push_str(s);
+        }
+    }
+    res
+}
+
+#[tokio::test]
+async fn keywords_in_sync() {
```
sdankel, line 835: Awesome!

sway-lsp/src/utils/keyword_docs.rs
JoshuaBatty: Yeah, we don't have support for iterators in sway yet so we don't have `for` loops. The only place `for` is used is when implementing traits for a type.

sway-lsp/src/utils/keyword_docs.rs
```diff
@@ -0,0 +1,850 @@
...
+        .iter()
+        .map(|s| s.to_string())
+        .collect();
+
+    for keyword in &compiler_keywords {
+        if !lsp_keywords.contains(&keyword) {
+            let err =
+                format!("Error: Documention for the `{keyword}` keyword is not implemented in LSP");
+            panic!("{err}");
+        }
```
sdankel, line 848: nit: assert does the same thing

```suggestion
        assert!(
            lsp_keywords.contains(&keyword),
            "Error: Documention for the `{keyword}` keyword is not implemented in LSP"
        );
}
```

sdankel (approved):

sway-lsp/src/utils/keyword_docs.rs
JoshuaBatty: Hmm interesting. I'll open an issue and follow up on this.

sway-lsp/src/utils/keyword_docs.rs
JoshuaBatty: nice, i'll add this change to the next PR

sway-lsp/src/utils/keyword_docs.rs
JoshuaBatty: Nice find, i'll open an issue for this as well.

sway-lsp/src/utils/keyword_docs.rs
JoshuaBatty: opened #3966

sway-lsp/src/utils/keyword_docs.rs
JoshuaBatty: opened #3967

PR #3795, opened 2023-01-17
Adds support for `Enum::<T>::Variant`.

### LINKED ISSUE #3585: Cannot use turbofish generic syntax

esdrubal:
This commit changes `TypeBinding` to contain also `prefix_type_arguments` this allows us to have both `Enum::<T>::Variant` compiled with `prefix_type_arguments` and `Enum::Variant::<T>` compiled with `type_arguments`. In case both are used an error is thrown.

Errors are also thrown when `prefix_type_arguments` is used with constants or functions. Tests included for this.

Also added an error when `type_arguments` is used with constants.

Closes #3585


sway-core/src/type_system/binding.rs
```diff
@@ -73,6 +73,7 @@ use super::{TypeArgument, TypeId};
 pub struct TypeBinding<T> {
     pub inner: T,
     pub type_arguments: Vec<TypeArgument>,
+    pub prefix_type_arguments: Vec<TypeArgument>,
```
anton-trunov, line 76: As far as I understand (please correct me if I'm wrong), `type_arguments` and `prefix_type_arguments` cannot be both non-empty. If that's the case, can we then use a different representation that ensures this by construction? Something isomorphic to

```
enum TypeArgs {
    Regular(Vec<TypeArgument>),
    Prefix(Vec<TypeArgument>),
}

pub struct TypeBinding<T> {
...
    pub type_arguments: TypeArgs,
...
}
```

perhaps?

anton-trunov (commented):
Excited to have this feature! Left a couple suggestions

sway-error/src/error.rs
```diff
@@ -713,6 +713,8 @@ pub enum CompileError {
         missing_impl_attribute: bool,
         span: Span,
     },
+    #[error("Enum instantiated with multiple generic type parameters. Try removing generic type parameters before or after the enum variant name.")]
+    EnumMultipleTurbofish { span: Span },
```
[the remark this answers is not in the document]
esdrubal: This error message was replaced in most recent version by `ConvertParseTreeError::MultipleGenericsNotSupported` which is also tested in [generics_multiple_type_arguments/src/main.sw](https://github.com/FuelLabs/sway/pull/3795/files#diff-57579ac2123e6b5841f47d0bf9bd259a26704a62d1c220648b87fc09f15faee5)

sway-core/src/type_system/binding.rs
esdrubal: Thanks for the review, most recent version applies this suggestion.

sway-core/src/type_system/binding.rs
```diff
@@ -74,10 +74,48 @@ use super::{TypeArgument, TypeId};
 #[derive(Debug, Clone)]
 pub struct TypeBinding<T> {
     pub inner: T,
-    pub type_arguments: Vec<TypeArgument>,
+    pub type_arguments: TypeArgs,
     pub span: Span,
 }
 
+#[derive(Debug, Clone)]
+pub enum TypeArgs {
```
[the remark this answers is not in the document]
esdrubal: Added comments.

emilyaherbert (commented):
Just a few nits, otherwise LGTM.

In the long term, we should consider replacing `TypeBinding` with something more ergonomic. See the discussion here for some more context: https://github.com/FuelLabs/sway/pull/3883#discussion_r1087229296

For this reason, @IGI-111 would you review this PR as well? Thanks!

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -1188,6 +1184,16 @@ impl ty::TyExpression {
             (false, None, None, Some((const_decl, span))) => {
                 warnings.append(&mut const_probe_warnings);
                 errors.append(&mut const_probe_errors);
+                if !unknown_call_path_binding.type_arguments.to_vec().is_empty() {
```
IGI-111, line 1187: I don't think you should use `unknown_call_path_binding` here given we know it's for a const declaration at that point. I remember setting the name specifically to distinguish it from the later moves. Better to use the same pattern as the function and pass it through the pattern as a tuple that you get to call `call_path_binding`

IGI-111 (changes requested):

esdrubal:
Thanks for the reviews. Applied all the review suggestions.

IGI-111 (approved):
LGTM

Issue #3747, opened 2023-01-10
Show unused code warnings as reduced opacity


sdankel:
Similar to other languages, we should show unused code as "dimmed" or reduced opacity in the editor. Investigate rust-analyzer's implementation of this and make it work for sway.

PR #3800, opened 2023-01-18
Show dead code with reduced opacity

sdankel:
Closes https://github.com/FuelLabs/sway/issues/3747

I realized that I missed a couple declaration types in https://github.com/FuelLabs/sway/pull/3769, so I updated those to use the name spans as well.
- const declarations
- variable declarations

Before:

image

After:

image


sdankel:
There is still a problem where used variable declarations are flagged as dead code if they're inside of a dead function (e.g. `everything` in the screenshot below). This wasn't noticeable before since previously the entire function block was highlighted as dead code. Fixing this seems pretty involved, since it's a problem with how we determine if code is dead or not.

We could simply update the code to filter out any dead code warnings inside of a dead function block (like we do for `UnreachableCode` inside a dead function block), but then even true dead code warnings wouldn't show up inside dead function blocks. I don't think that would be any better than what we have now.

mohammadfawaz (approved):
Looks good!

sway-lsp/src/capabilities/diagnostic.rs
```diff
@@ -24,6 +24,7 @@ fn get_warning_diagnostics(warnings: &[CompileWarning]) -> Vec<Diagnostic> {
         range: get_range(warning.span().line_col()),
         severity: Some(DiagnosticSeverity::WARNING),
         message: warning.to_friendly_warning_string(),
+        tags: get_warning_diagnostic_tags(&warning.warning_content),
```
JoshuaBatty, line 27: sweet :)

JoshuaBatty (approved):
:shipit:

PR #3808, opened 2023-01-18
Add a flag to print `Log` and `LogData` receipts emitted from tests

### LINKED ISSUE #3807: Collect `Log` and `LogData` receipts when running `forc test`
Currently, `Log` and `LogData` receipts are not collected when running `forc test`. It would be useful to actually display those receipts for debugging purposes, and maybe hide this behavior behind a debug flag.

kayagokalp:
closes #3807.

- Moved `Log` and `LogData` formatting code from `forc-client` to `forc-util` as `forc-test` needs it too.
- Added `--logs` and `--pretty-print` flags to `forc-test`

## Sample outputs:

```console
fuel/test_projects/deploy [⏱ 4s]
❯ mforc test --log --pretty-print
  Compiled library "core".
  Compiled library "std".
  Compiled library "deploy".
  Bytecode size is 36 bytes.
   Running 1 tests
[
  {
    "Log": {
      "id": "0000000000000000000000000000000000000000000000000000000000000000",
      "is": 10344,
      "pc": 10368,
      "ra": 1,
      "rb": 0,
      "rc": 0,
      "rd": 0
    }
  }
]
      test my_test ... ok (600.166µs)
   Result: OK. 1 passed. 0 failed. Finished in 600.166µs.
```
```console
fuel/test_projects/deploy [⏱ 4s]
❯ mforc test --log         
  Compiled library "core".
  Compiled library "std".
  Compiled library "deploy".
  Bytecode size is 36 bytes.
   Running 1 tests
[{"Log":{"id":"0000000000000000000000000000000000000000000000000000000000000000","is":10344,"pc":10368,"ra":1,"rb":0,"rc":0,"rd":0}}]
      test my_test ... ok (405.875µs)
   Result: OK. 1 passed. 0 failed. Finished in 405.875µs.
```


nfurfaro (approved):
LGTM.

mitchmindtree (approved):
LGTM too!

My only question is - how does the output look when there are multiple tests? Is it clear which test the log data is associated with?

mitchmindtree:
Btw @nfurfaro do you know if that log data is designed to be presented in a nicer manner?

E.g. is it enough to print the raw struct like that? Or would you normally expect it to be formatted in a nicer way?

kayagokalp:
> My only question is - how does the output look when there are multiple tests? Is it clear which test the log data is associated with?

I just realized this question and I think I should fix it as it is not very clear which log is coming from which test. Will open an issue for this.

mohammadfawaz:
> > Btw @nfurfaro do you know if that log data is designed to be presented in a nicer manner?
> > E.g. is it enough to print the raw struct like that? Or would you normally expect it to be formatted in a nicer way?
> 
> Seeing the full raw struct like this is what I've grown used to. That said, there's no reason why we couldn't just log the value of `ra` by default , or `data` in the case of a `logd` receipt. The user could always specify a `--verbose` option to get the full thing if needed, but 99% of the time I'm trying to sort through a mess of debugging logs to find the value of `ra`.

If we're feeling fancy, we can also use the JSON ABI file to actually decode the log data and print it nicely 😉

nfurfaro:
>If we're feeling fancy, we can also use the JSON ABI file to actually decode the log data and print it nicely

That could be nice to have...
For unit testing I'm 100% just using `log` in the way that I would use `println!` for debugging, so usually just simple values that are easy to parse in the receipts. I definitely miss being able  to use string templates, ie:
`println!("Value of thing 1: {:?}", thing_1);`

I usually end up just having to manually count log statements, and then go through the receipts looking fo the `ra` value of the 9th `log` receipt for example.

PR #3811, opened 2023-01-18
Move to our own published crate instead of Git repo for ethabi crate.

tritao:


mohammadfawaz (approved):
Great! Thanks for the quick fix :)

esdrubal (approved):

PR #3814, opened 2023-01-18
update installation instructions to be more clear

camiinthisthang:
- Remove instructions to run fuelup-init
- Add text that clearly states that installing via precompiled binaries is the recommended way
- Add console output
- Add instructions to optionally install distributed toolchains besides latest

Before Screen Shot 2023-01-18 at 1 55 00 PM

After Screen Shot 2023-01-18 at 1 54 25 PM

@mohammadfawaz I added the console output for the commands but it makes the page really long. I like the output because it tells the user what to expect, but should we consider shortening it with ... so the steps don't look as long?


mohammadfawaz (approved):
Looks good to me! Thanks. Hoping @bingcicle can also take a look.

docs/book/src/introduction/installation.md
```diff
@@ -2,35 +2,132 @@
...
+   ```
+
+   `fuelup-init` will ask for permission to add `~/.fuelup/bin` to your PATH. Otherwise, you can also pass `--no-modify-path` so that `fuelup-init` does not modify your PATH:
+
+   ```sh
+   curl --proto '=https' --tlsv1.2 -sSf \
+   https://fuellabs.github.io/fuelup/fuelup-init.sh | sh -s -- --no-modify-path
+   ```
+
+2. Once `fuelup` is installed, run the following command to install the latest Fuel toolchain:
```
bingcicle, line 25: Ah I see why you asked about `fuelup-init`! You actually run `fuelup-init` in the first step here:

```sh
curl --proto '=https' --tlsv1.2 -sSf \
   https://fuellabs.github.io/fuelup/fuelup-init.sh | sh -s -- --no-modify-path
```
and that does everything for you, including the [toolchain installation step](https://github.com/FuelLabs/fuelup/blob/bd4855bfbca5df0886ab0f9966e7fd2044b89ad4/fuelup-init.sh#L130) - `fuelup-init` invokes `fuelup` to do the installation for you by default for convenience if you didn't provide `--skip-toolchain-installation`, so maybe the original documentation wasn't clear enough about that.

I'd suggesting reverting back to the original line or some variation that keeps the original meaning here, and we'll include the missing info above.

docs/book/src/introduction/installation.md
```diff
@@ -2,35 +2,132 @@
...
-
-You can run the same command at a later time to update the toolchain.
+1. Start by installing `fuelup` with the following command:
+
+   ```sh
+   curl --proto '=https' --tlsv1.2 -sSf \
+   https://fuellabs.github.io/fuelup/fuelup-init.sh | sh
+   ```
+
+   `fuelup-init` will ask for permission to add `~/.fuelup/bin` to your PATH. Otherwise, you can also pass `--no-modify-path` so that `fuelup-init` does not modify your PATH:
```
bingcicle, line 18:
```suggestion
    This downloads `fuelup-init` to a temp directory on your machine and starts the installation for `fuelup`. `fuelup-init` will ask for permission to add `~/.fuelup/bin` to your PATH. Otherwise, you can also pass `--no-modify-path` so that `fuelup-init` does not modify your PATH:
```

Perhaps here is where we include the additional info for clarity

bingcicle (changes requested):

docs/book/src/introduction/installation.md
camiinthisthang: Ok cool thanks. I'm not clear on if the user has to run `fuelup-init`, or if the users has to run `fuelup toolchain install latest` or neither and these get called automatically once you install fuelup? @bingcicle

bingcicle: > curl --proto '=https' --tlsv1.2 -sSf \
   https://fuellabs.github.io/fuelup/fuelup-init.sh | sh
   
   this command invokes `fuelup-init`, then

> fuelup toolchain install latest

this is invoked by `fuelup-init` automatically once `fuelup` is done installing. So yes, it is automatic :)

camiinthisthang:
I did a bad git thing, closing this PR in favor of https://github.com/FuelLabs/sway/pull/3890 cc @mohammadfawaz @bingcicle

Issue #3693, opened 2023-01-05
Refactor runnables to use standard LSP codeLens API


sdankel:
The "run" button for the sway VSCode plugin was implemented using a custom "runnables" method. We should refactor this to be returned from the standard codeLens API so that it works in all LSP-supporting IDEs.

PR #3819, opened 2023-01-19
Replace runnables custom fn with LSP codeLens

sdankel:
Related https://github.com/FuelLabs/sway/issues/3693

This PR moves the logic from the custom `runnables` method into the LSP `codeLens` method. From here, we'll add other types of runnables, e.g. sway tests.

Added a test for `codeLens` as well as a missing test for `codeActions`. 

Unfortunately there is currently no support for tooltips on server-supplied code lenses (https://github.com/microsoft/language-server-protocol/issues/1644) because they aren't part of the LSP spec. So, the tooltip will no longer show up for the time being.

The upside is that the run button could be implemented for other IDEs with less effort (not entirely free, still need to define the run command in the client-side plugin code).

There will be two follow-up PRs to this:
- Removing the runnables code in the vscode client & updating the run command
- Removing the `runnables` custom method in the server.


sway-lsp/src/server.rs
```diff
@@ -659,8 +677,8 @@ mod tests {
         let params = json!({ "capabilities": capabilities() });
         let initialize = build_request_with_id("initialize", params, 1);
         let response = call_request(service, initialize.clone()).await;
-        let ok = Response::from_ok(1.into(), json!({ "capabilities": capabilities() }));
-        assert_eq!(response, Ok(Some(ok)));
+        let expected = Response::from_ok(1.into(), json!({ "capabilities": capabilities() }));
+        assert_json_eq!(expected, response.ok().unwrap());
```
sdankel, line 681: `assert_json_eq` has nicer formatting, making it easy to tell what the difference is when they're not equal.

sway-lsp/src/server.rs
```diff
@@ -1037,17 +1140,19 @@ mod tests {
...
-    lsp_capability_test!(format, format_request);
-    lsp_capability_test!(hover, hover_request);
-    lsp_capability_test!(highlight, highlight_request);
+    lsp_capability_test!(semantic_tokens, semantic_tokens_request, doc_comments_dir);
+    lsp_capability_test!(document_symbol, document_symbol_request, doc_comments_dir);
+    lsp_capability_test!(format, format_request, doc_comments_dir);
+    lsp_capability_test!(hover, hover_request, doc_comments_dir);
+    lsp_capability_test!(highlight, highlight_request, doc_comments_dir);
+    lsp_capability_test!(code_action, code_action_request, doc_comments_dir);
+    lsp_capability_test!(code_lens, code_lens_request, e2e_test_dir);
```
sdankel, line 1157: This one has to be tested using a sway script in order to get any runnables back.

sway-lsp/src/server.rs
JoshuaBatty: Oh nice find

sway-lsp/src/server.rs
JoshuaBatty: Makes sense

JoshuaBatty (approved):
LGTM

mitchmindtree (approved):

PR #3822, opened 2023-01-19
Add sha256 and keccak256 hash functions to `Bytes` type

### LINKED ISSUE #3809: Hashing of `Bytes` type
The `Bytes` type was recently introduced as a way to tightly pack bytes. A common use case for this is hashing as a way to avoid padding of types such as `u8` to a full word. This is explicitly done in the [Merkle Proof library](https://github.com/FuelLabs/sway-libs/blob/27215d1829d2fd03d78d31a7d00cd70675cbc8fc/sway_libs/src/merkle_proof/binary_merkle_proof.sw#L22) where the concatenation of a single `u8` to the front of a `b256` is needed to follow the RFC-6962 specification.

However, the current [`sha256()`](https://github.com/FuelLabs/sway/blob/f18197f0c040d1223968ca046edf12e053a3568a/sway-lib-std/src/hash.sw#L4) function in the hash library does not allow for hashing of the tightly packed bytes. The `Bytes` type stores a length along with the `RawBytes` type which contains a pointer to the bytes. 
For example, when calling `sha256(my_bytes)` where `my_bytes` is a `Bytes` struct, the `sha256()` method will call `__size_of` upon the struct. As a result, I did not just hash the tightly packed bytes but instead hashed the struct itself. 

To avoid developers having to create their own asm blocks, there should be an easy method to hash these bytes. 

I believe the best way to handle this would be to add a hashing function to the `Bytes` type. A developer would then call the function in the same manor that `.join()` and `.push()` are called. 

cc @nfurfaro

bitzoic:
Due to the need to hash the elements in the `Bytes` type, the sha256() and keccak256() functions have been added to the `Bytes` library.

Closes #3809


sway-lib-std/src/bytes.sw
```diff
@@ -913,3 +955,29 @@ fn test_eq() {
...
+fn test_sha256() {
+    use ::hash::sha256;
+    let (mut bytes, a, b, c) = setup();
+    bytes.push(0u8);
+    bytes.push(0u8);
+    bytes.push(0u8);
+    bytes.push(0u8);
+    bytes.push(0u8);
+
+    assert(sha256(362268190631264256) == bytes.sha256());
```
nfurfaro, line 969: Overall this looks good. Can you comment on how you derived this value `362268190631264256` ?

mohammadfawaz:
I wonder if eventually we need `Sha256` and `Keccak256` traits that allows us custom implementations of hash functions. We can then have identical APIs for hash functions regardless of the type.

sway-lib-std/src/bytes.sw
```diff
@@ -435,6 +435,48 @@ impl Bytes {
...
+    /// use std:bytes::Bytes;
+    ///
+    /// let bytes = Bytes::new();
+    /// bytes.push(1);
+    /// bytes.push(2);
+    /// bytes.push(3);
+    /// let sha256_hash = bytes.sha256();
+    /// ```
+    pub fn sha256(self) -> b256 {
+        let mut result_buffer: b256 = b256::min();
```
Braqzen, line 453:
```suggestion
        let mut result_buffer = b256::min();
```

Unless the compiler requires it we should avoid explicit annotation for types that can be inferred when reading the code

sway-lib-std/src/bytes.sw
```diff
@@ -435,6 +435,48 @@ impl Bytes {
...
+    /// use std:bytes::Bytes;
+    ///
+    /// let bytes = Bytes::new();
+    /// bytes.push(1);
+    /// bytes.push(2);
+    /// bytes.push(3);
+    /// let keccak256_hash = bytes.keccak256();
+    /// ```
+    pub fn keccak256(self) -> b256 {
+        let mut result_buffer: b256 = b256::min();
```
Braqzen, line 474:
```suggestion
        let mut result_buffer = b256::min();
```

sway-lib-std/src/bytes.sw
```diff
@@ -913,3 +955,29 @@ fn test_eq() {
     other.swap(0, 1);
     assert(bytes != other);
 }
+
+#[test()]
+fn test_sha256() {
```
Braqzen, line 960: Should we test each `type` or is this fine?

bitzoic: Since we don't have the `From` trait implemented these would all need to be manually added.

bitzoic:
> I wonder if eventually we need `Sha256` and `Keccak256` traits that allows us custom implementations of hash functions. We can then have identical APIs for hash functions regardless of the type.

This may be needed if hashes of the contents of `Vec`, `String`, or anything which contains a pointer is desired

Braqzen (approved):

mohammadfawaz (approved):

PR #3823, opened 2023-01-19
Align asm opcodes to 4 chars in formatter

### LINKED ISSUE #590: Formatter should format inline asm opcodes aligned to 4 chars
The maximum size of each opcode is 4 characters. The formatter should pad with extra spaces such that the "opcode" part of each line is 4 characters.

anton-trunov:
close #590


swayfmt/src/utils/language/expr/asm_block.rs
```diff
@@ -140,20 +140,37 @@ impl Format for AsmRegisterDeclaration {
...
+        formatted_code: &mut FormattedCode,
+        _formatter: &mut Formatter,
+    ) -> Result<(), FormatterError> {
+        write!(formatted_code, "{:<4}", &self.op_code_ident().as_str())?;
+        for arg in self.register_arg_idents() {
+            write!(formatted_code, " {}", arg.as_str())?
+        }
+        if let Some(ident) = self.immediate_ident_opt() {
+            write!(formatted_code, " {}", ident.as_str())?
+        }
```
mohammadfawaz, line 155: Is this assuming that `asm` instructions always look like `<opcode> <r> <r> .. <i>`, so a list of registers followed by at most 1 immediate? While that may be true today, I find that a bit limiting? Can we not simply loop over all the args and write them regardless of their type?

anton-trunov: Good point! I took a stab at it

mohammadfawaz (approved):
Alright that seems general enough for now :) I suspect we won't ever have opcodes that alternate registers and immediates.

JoshuaBatty (approved):

Issue #2552, opened 2022-08-16
Add conditional compilation to Sway


sezna:
This will enable us to support multiple backends in the standard library among other things. We should minimally support something like `arch = EVM` vs `arch = FuelVM` in annotations (syntax is pseudo, follow Rust syntax)

Other cfg flags for conditional compilation can come later.

AlicanC:
Some notes from my research are below. cc @mohammadfawaz 

---

So it looks like we want to go for `target_arch` first: https://doc.rust-lang.org/reference/conditional-compilation.html#target_arch

It would look like this:
```rs
// The function is only included in the build when compiling for EVM
#[cfg(target_arch = "evm")]
fn evm_only() {
  // ...
}
```

This uses attributes which we have a very limited implementation of. This could be a good excuse to improve it and fix issues like https://github.com/FuelLabs/sway/issues/3576.

There is also a short form which we could implement along with the long form or just by itself:
```rs
#[cfg(evm)]
fn evm_only() {
  // ...
}
```

The short form shouldn't require any changes to the parser, so we could just go for this and not touch the parser at all, but I'm still for improving attr parsing a bit before we do this.

So the steps would be:
- Improve attribute support. This can be delivered early on a separate PR. This *should* serve all the parser/syntax needs.
- Add `cfg` attribute: https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg-attribute
- Add `cfg_attr` attribute: https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg_attr-attribute
- Add `cfg` macro: https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg-macro
- Add `target_arch` configuration option: https://doc.rust-lang.org/reference/conditional-compilation.html#target_arch

AlicanC:
I think the least we have to do is:
- Improve attr parsing until building on it doesn't feel like a crime.
- Add `cfg` attr and macro, no `cfg_attr`, no `target_arch`, just `cfg(evm)`/`cfg(fvm)`. (Names should be bikeshed.)

PR #3829, opened 2023-01-19
Make revert() generic

### LINKED ISSUE #3639: Consider making `revert` generic or implementing `log_and_revert`
This is based on the discussion in https://forum.fuel.network/t/provide-a-log-and-revert-function-in-the-standard-library/958

nfurfaro:
Closes #3639 

TODO: 
- [ ] open issue with fuels-rs to watch for `REVERT_WITH_LOG_SIGNAL` similar to `FAILED_REQUIRE_SIGNAL`


nfurfaro:
Hmm. an unintended consequence of this change is that the use of the `revert` function anywhere in a dependency of a predicate _can_ lead to an attempt to call `log`, which gives the following error (as seen in the CI failures):
```
error
 --> /home/runner/work/sway/sway/sway-lib-std/src/logging.sw:8:5
  |
6 | 
7 | pub fn log<T>(value: T) {
8 |     __log::<T>(value);
  |     ^^^^^^^^^^^^^^^^^ Internal compiler error: Unable to determine ID for log instance.
9 | }
  |
____
```
i.e: using `ec_recover` makes use of the `input_type()` function in std::inputs which can call revert.
likewise, using `assert` makes use of `revert`. Even though `log` will never be called if we pass a copy type to revert, the compiler won't build a predicate which uses revert anywhere now.

At least I think this is what is happening.
cc @mohammadfawaz

nfurfaro:
It was mentioned on Slack by @sezna that conditional compilation could be helpful here. 
ref: #2552

PR #3832, opened 2023-01-19
Fixed `self` in trait methods

### LINKED ISSUE #3805: `forc doc`: Problem formatting `self` in function parameters of trait methods
This has to do with #3657 

I hastily made a formatter for trait methods, and in doing so inadvertently birthed this issue. I plan to tackle this while handling the aforementioned issue, but I'm opening this for consistency.

gr00vytvniks:
Closes #3805


kayagokalp (commented):
👌 Hmm this makes me wonder if we can have some tests for `forc-doc` for catching this sort of regressions. I am not sure how to test the PR, so would be great to have @eureka-cpu's review/input :)

gr00vytvniks:
> 👌 Hmm this makes me wonder if we can have some tests for `forc-doc` for catching this sort of regressions. I am not sure how to test the PR, so would be great to have @eureka-cpu's review/input :)

Here's how I tested:
```sh
cd ~/sway/forc-plugins/forc-doc
cargo install --path .
cd ~/sway/sway-lib-std
forc doc --open
```
just do this process on master, click on any method that includes self and then do the same on this branch to compare

kayagokalp:
Ahh so this is a fix for all methods, got it 👍

eureka-cpu (approved):
Just tested, looks like that was all that was needed to fix it lol

kayagokalp (approved):

PR #3836, opened 2023-01-20
Adds support for where clauses in impl self methods.

### LINKED ISSUE #3333: Can't use generic type parameters from `impl<>` in impl self methods
Code:
```rust
script;

use core::ops::*;


struct Vec<T> {
    x: T
}

impl<T> Vec<T> {
    fn contains(self, other: T) -> bool where T: Eq { 
        true
    }
}

fn main() { }
```
I expect the above to work so that I can compare `self` with `other` in the body of `contains`. Instead, I get:
```console
10 | impl<T> Vec<T> {
11 |     fn contains(self, other: T) -> bool where T: Eq {
   |                                               ^ cannot find type "T" in this scope
12 |         true
13 |     }
   |
```

esdrubal:
When where clauses were used in impl self methods with a generic parameter from
 impl<T> an error was thrown saying `cannot find type "T" in this scope`

Parent type parameters that are used in method where clauses are now added to `FunctionDeclaration.type_parameters`.
This is needed for `TypeParameter::gather_decl_mapping_from_trait_constraints` to work properly as we need to have a type parameter with constraints. The generated decl_mapping is used by `type_check_method_application` to replace decls from trait constraints.

Closes #3333

### DIFF
--- a/sway-core/src/language/call_path.rs
+++ b/sway-core/src/language/call_path.rs
@@ -110,29 +110,28 @@ impl CallPath {
             // combination of the package name in which the symbol is defined and the path to the
             // current submodule.
             let mut synonym_prefixes = vec![];
-            let mut submodule_name_in_synonym_prefixes = false;
+            let mut is_external = false;
 
             if let Some(use_synonym) = namespace.use_synonyms.get(&self.suffix) {
                 synonym_prefixes = use_synonym.0.clone();
                 let submodule = namespace.submodule(&[use_synonym.0[0].clone()]);
                 if let Some(submodule) = submodule {
-                    if let Some(submodule_name) = submodule.name.clone() {
-                        submodule_name_in_synonym_prefixes =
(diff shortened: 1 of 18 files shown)


emilyaherbert (changes requested):
For bookkeeping---requesting changes on this PR for this reason:

https://github.com/FuelLabs/sway/pull/3621#issuecomment-1402757205

esdrubal:
Reopen the PR up for review.

The PR still uses the `TypeParameter` with `is_from_parent`. This looks like a cleaner approach than moving `TraitConstraint` out of `TypeParameter`. Because a trait constraint is always relative to a type parameter.

sdankel:
Looks like there are some merge conflicts. It would be nice to have this fixed!

esdrubal:
The merge conflicts are now resolved.

IGI-111 (approved):

JoshuaBatty (approved):

PR #3840, opened 2023-01-20
feat: error out if more than one package declares the same package name in a workspace

### LINKED ISSUE #3650: `forc build` does not build all workspace members if two members share the same name
When running `forc build` and using a workspace, if members have the same name in their local Forc.toml file, both will not compile. 

In the following example this is the project structure:

```
test_project/
├──  member1/
|     └── src/
|     |     └── main.sw
|     └── Forc.toml
├── member2/
|     └── src/
|     |     └── main.sw
|     └── Forc.toml
└── Forc.toml
```

This is the parent directory Forc.toml file: 

```rust
[workspace]
members = [
  "./member1",
  "./member2",
]
```

Both members of the workspace have the following identical local Forc.toml file:

```rust
[project]
authors = ["Cam Fuel"]
entry = "main.sw"
license = "Apache-2.0"
name = "member"

[dependencies]
```

Despite both being listed as separate members in the workspace, because they have the same name in the local Forc.toml file only one will compile. 

Tested with forc v0.32.2

kayagokalp:
closes #3650

### DIFF
--- a/forc-pkg/src/manifest.rs
+++ b/forc-pkg/src/manifest.rs
@@ -836,19 +836,14 @@ impl WorkspaceManifest {
             }
             let member_manifest_file = PackageManifestFile::from_file(member_path.clone())?;
             let pkg_name = member_manifest_file.manifest.project.name;
-            if let Some(paths) = pkg_name_to_paths.get_mut(&pkg_name) {
-                paths.push(member_path);
-            } else {
-                pkg_name_to_paths.insert(pkg_name, vec![member_path]);
-            }
+            pkg_name_to_paths.entry(pkg_name).or_default().push(member_path);
         }
 
         // Check for duplicate pkg name entries in member manifests of this workspace.
         let duplciate_pkg_lines = pkg_name_to_paths
             .iter()
             .filter(|(_, paths)| paths.len() > 1)
-            .map(|(pkg_name, _)| pkg_name)
-            .map(|pkg_name| {
+            .map(|(pkg_name, _)| {
                 let duplicate_paths = pkg_name_to_paths
                     .get(pkg_name)
                     .expect("missing duplicate paths");
--- a/forc-pkg/src/manifest.rs
+++ b/forc-pkg/src/manifest.rs
@@ -4,7 +4,7 @@ use forc_tracing::println_yellow_err;
 use forc_util::{find_manifest_dir, validate_name};
 use serde::{Deserialize, Serialize};
 use std::{
-    collections::BTreeMap,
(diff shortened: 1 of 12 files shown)


forc-pkg/src/manifest.rs
```diff
@@ -833,6 +834,34 @@ impl WorkspaceManifest {
                     member_path
                 );
             }
+            let member_manifest_file = PackageManifestFile::from_file(member_path.clone())?;
+            let pkg_name = member_manifest_file.manifest.project.name;
+            if let Some(paths) = pkg_name_to_paths.get_mut(&pkg_name) {
+                paths.push(member_path);
+            } else {
+                pkg_name_to_paths.insert(pkg_name, vec![member_path]);
+            }
```
mitchmindtree, line 843:
```suggestion
            pkg_name_to_paths.entry(pkg_name).or_default().push(member_path);
```

forc-pkg/src/manifest.rs
```diff
@@ -833,6 +834,34 @@ impl WorkspaceManifest {
...
+        }
+
+        // Check for duplicate pkg name entries in member manifests of this workspace.
+        let duplciate_pkg_lines = pkg_name_to_paths
+            .iter()
+            .filter(|(_, paths)| paths.len() > 1)
+            .map(|(pkg_name, _)| pkg_name)
+            .map(|pkg_name| {
```
mitchmindtree, line 851:
```suggestion
            .map(|(pkg_name, _)| {
```

forc-pkg/src/manifest.rs
```diff
@@ -833,6 +834,34 @@ impl WorkspaceManifest {
...
+                let duplicate_paths = pkg_name_to_paths
+                    .get(pkg_name)
+                    .expect("missing duplicate paths");
+                format!("{}: {:?}", pkg_name, duplicate_paths)
```
mitchmindtree, line 855:
```suggestion
                format!("{}: {:#?}", pkg_name, duplicate_paths)
```
^ I _think_ this should print each path on a new line to improve readability a little, but I could be wrong.

mitchmindtree (commented):
Looking good! Just some small cleanup then happy to land.

forc-pkg/src/manifest.rs
kayagokalp: Just tested with and without this change and pretty print is easily 10x better, thanks!

mitchmindtree (approved):

JoshuaBatty (approved):

PR #3847, opened 2023-01-21
Add missing tokens inside turbofish

### LINKED ISSUE #3612: LSP missing tokens: turbo fish
We aren't able to get unique span and type_info information from types within turbo fish syntax. eg.

```rust
let x = local_panic::<u64>();
```

In the above example, we are missing the `u64`.

IGI-111:
Expose type arguments inside turbofish for function application, method application and enum instantiation by populating the typed AST with relevant instances of TypeBinding and having the LSP collect them.

All three of the following lines will now generate proper tokens that point to real type definitions inside turbofish:

```sway
fun::<Option<u32>>(Option::None);
s.meth::<Option<u32>>(Option::None);
let a = Option::None::<Option<u32>>;
```

Fix #3612

### DIFF
--- a/sway-lsp/src/server.rs
+++ b/sway-lsp/src/server.rs
@@ -792,42 +792,77 @@ mod tests {
         document_symbol
     }
 
-    async fn go_to_definition_request(
+    fn definition_request(uri: &Url, token_line: i32, token_char: i32, id: i64) -> Request {
+        let params = json!({
+            "textDocument": {
+                "uri": uri,
+            },
+            "position": {
+                "line": token_line,
+                "character": token_char,
+            }
+        });
+        build_request_with_id("textDocument/definition", params, id)
+    }
+
+    fn definition_response(
+        uri: &Url,
+        start_line: i32,
+        start_char: i32,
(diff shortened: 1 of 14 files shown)


JoshuaBatty (approved):
Nice work! Looks good.

otrho:
Is it possible to add a test?

IGI-111:
> Is it possible to add a test?

Added some.

mohammadfawaz (approved):
Awesome, thanks!

sway-lsp/src/server.rs
```diff
@@ -792,44 +792,79 @@ mod tests {
...
+                        "line": start_line,
                     }
                 },
                 "uri": uri,
             }),
-        );
+        )
+    }
+
+    async fn doccomments_definition_check(
```
JoshuaBatty, line 834: nit: can we rename this to `doc_comments_definition_check`

JoshuaBatty (approved):
Thanks! :shipit:

otrho (approved):

PR #3848, opened 2023-01-21
Add docstring previews to index files

### LINKED ISSUE #3820: `forc doc`: add docstring previews to index files
ref #3778

eureka-cpu:
Closes #3820 

I've added a limit of 200 characters so that anything longer would be 75 characters with an ellipsis (not pictured):
Screen Shot 2023-01-20 at 7 16 37 PM

### DIFF
--- a/forc-plugins/forc-doc/src/descriptor.rs
+++ b/forc-plugins/forc-doc/src/descriptor.rs
@@ -1,7 +1,7 @@
 //! Determine whether a [Declaration] is documentable.
 use crate::{
     doc::{Document, ModuleInfo},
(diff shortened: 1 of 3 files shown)


mitchmindtree (approved):
Nice! utACK

It would be great if we could also format those lines with basic markdown as they look a bit noisy with the raw unformatted text. Happy to leave that for a future issue if it's tricky to land in this though.

JoshuaBatty (approved):

eureka-cpu:
> Nice! utACK
> 
> It would be great if we could also format those lines with basic markdown as they look a bit noisy with the raw unformatted text. Happy to leave that for a future issue if it's tricky to land in this though.

Will follow up on this!

Issue #3320, opened 2022-11-08
LSP: collect `#[test]` functions and display a run button in editor


JoshuaBatty:
Might be worth looking at the method that was added in #3152 for populating our HashMap of runnables with these test methods.

sdankel:
We should do this as a prerequisite: https://github.com/FuelLabs/sway/issues/3693

PR #3849, opened 2023-01-21
Add run button for sway tests

sdankel:
Closes https://github.com/FuelLabs/sway/issues/3320

This shouldn't be merged until the front end change is released: https://github.com/FuelLabs/sway-vscode-plugin/pull/119

Jan-20-2023 18-04-56


sway-core/src/transform/attribute.rs
```diff
@@ -69,3 +69,26 @@ pub(crate) fn generate_json_abi_attributes_map(
         )
     }
 }
+
+pub trait First {
```
JoshuaBatty, line 73: I usually avoid making a trait if it will only be implemented for a single type. I feel like it's usually better to implement the behaviour on the type itself in these situations. Perhaps something like the following might be more appropriate? 

```rust
/// Stores the attributes associated with the type.
#[derive(Default, Clone, Debug, Eq, PartialEq)]
pub struct AttributesMap(Arc<HashMap<AttributeKind, Vec<Attribute>>>);

impl AttributesMap {
    /// Create a new attributes map.
    pub fn new(attrs_map: Arc<HashMap<AttributeKind, Vec<Attribute>>>) -> AttributesMap {
        AttributesMap(attrs_map)
    }

    /// Returns the first attribute by span, or None if there are no attributes.
    pub fn first(&self) -> Option<(&AttributeKind, &Attribute)> {
        let mut first: Option<(&AttributeKind, &Attribute)> = None;
        for (kind, attrs) in self.iter() {
            for attr in attrs {
                if let Some((_, first_attr)) = first {
                    if attr.span.start() < first_attr.span.start() {
                        first = Some((kind, attr));
                    }
                } else {
                    first = Some((kind, attr));
                }
            }
        }
        first
    }
}

impl std::ops::Deref for AttributesMap {
    type Target = Arc<HashMap<AttributeKind, Vec<Attribute>>>;
    fn deref(&self) -> &Self::Target {
        &self.0
    }
}
```

JoshuaBatty (commented):
This is super cool, can't wait to see run buttons for all of the tests! 

just have one comment about the trait. Looks good otherwise nice work.

sway-lsp/src/capabilities/runnable.rs
```diff
@@ -1,25 +1,62 @@
...
+#[derive(Debug, Eq, PartialEq, Clone)]
 pub struct Runnable {
+    /// The kind of runnable
+    pub kind: RunnableKind,
     /// The location in the file where the runnable button should be displayed
     pub range: Range,
     /// The program kind of the current file
     pub tree_type: TreeType,
+    /// Additional arguments to use with the runnable command
+    pub arguments: Option<Vec<Value>>,
```
JoshuaBatty, line 23: Is it possible to document the specific values that are expected for arguments? Since Value can represent 6 unique types it's a bit unclear what is valid or not to send in here.

sdankel: Done! Now, there is a Runnable trait and 2 structs with implementations for Runnable. Each struct is responsible for implementing it's own `arguments`, making it really clear which arguments are needed.

JoshuaBatty (approved):
Nicely done!

JoshuaBatty (approved):

mitchmindtree (approved):

sway-core/src/transform/attribute.rs
```diff
@@ -48,7 +48,39 @@ pub enum AttributeKind {
...
+
+impl AttributesMap {
+    /// Create a new attributes map.
+    pub fn new(attrs_map: Arc<HashMap<AttributeKind, Vec<Attribute>>>) -> AttributesMap {
+        AttributesMap(attrs_map)
+    }
+
+    /// Returns the first attribute by span, or None if there are no attributes.
+    pub fn first(&self) -> Option<(&AttributeKind, &Attribute)> {
+        let mut first: Option<(&AttributeKind, &Attribute)> = None;
```
otrho, line 62: You can avoid imperative mutable state-updating by using `iterator::fold` or `iterator::reduce` in cases like this.

otrho (approved):
utACK.

sway-core/src/transform/attribute.rs
sdankel: Thanks, I'll try this out - not going to block merging this PR on it though :)

Issue #3789, opened 2023-01-16
forc-fmt: Consider parsing with comments


kayagokalp:
We are facing lots of small issues regarding comment insertion. At the time of https://github.com/FuelLabs/sway/pull/2311 we were still exploring the issue and after maintaining this way of handling comments for couple of months, I think it seems like a better way of handling should be discussed. I am not linking all comment related hot-fix PRs but we had couple of them 😄

## Proposal
Let's parse comments and produce a `CommentedParseTree` just like we are lexing comments (we have `CommentedTokenStream` vs `TokenStream` etc). This way re-constructing the comments section would not need:

1. Search the unformatted code
2. Search the formatted code

We would be simply formatting comments just like we are formatting other stuff like `Ty`s, `expr`s etc. I believe this would be a little bit faster and most importantly it could lead to less problems. Currently once we have some edge case stuff happening related to the comments we tend to introduce rather "hacky" solutions as the interaction between formatter and comment handler is not simple enough.

With parsed comments, if we can also collect some sort of context we could still be sure that we are not altering the intentional comment styles (rust classifies comments like [this](https://github.com/rust-lang/rust/blob/master/compiler/rustc_ast/src/util/comments.rs#L8-L18))

Any thoughts on this? cc @FuelLabs/tooling

bingcicle:
+1 on this, swayfmt v2 seems mature enough to work well for most cases but it seems like the search-and-insert method is leading to a few weird behavior surrounding formatting with comments. Right now fixing these edge cases are just small changes but this could become hard to maintain over time.

mitchmindtree:
Sounds good to me!

This would also address https://github.com/FuelLabs/sway/issues/2357.

Originally, we avoided trying to produce a `Commented` alternative to the `sway_ast::Module` due to the sheer number of places that comments can appear within the AST (between every leaf node!) however as we're discovering, the unstructured alternative leaves a *lot* of edge cases that are difficult to reliably handle in a non-destructive way.

I just had a chat with @bingcicle and after checking rustfmt again for better ideas, he'll start looking into how feasible this is.

The gist of the idea is to mirror the existing AST of `sway_ast` under a new `commented` module and replace each field with a commented alternative along the lines of the following:
```rust
pub struct Commented<T> {
    pub node: T,
    pub comments: Vec<Comment>,
}

pub struct Comment {
    pub span: Span,
}
```
Where `comments` is a list of comments that occur between this node and the previous node.

[...]

Issue #2888, opened 2022-09-28
Sway formatter moves comment outside of empty struct


Braqzen:
`forc -V` = `0.24.5`.

The following

```rust
library cramps;

struct MultiSignatureWallet {
    // code
}

trait MetaData {
    // code
}
```

turns into

```rust
library cramps;

struct MultiSignatureWallet {}

    // code
trait MetaData {
    // code
}
```

when running `forc fmt`.

cc @eureka-cpu @kayagokalp

PR #3854, opened 2023-01-23
feat(fmt): introduce formatting comments together with src

### LINKED ISSUE #3574: fmt: if-else with comments inline formatting interaction
source: https://github.com/FuelLabs/sway/pull/3556#issuecomment-1345820171

The code here:

```rust
fn main() {
    if foo {
        let x = true;
    }
    // comment
    else {
        bar(x);
    }
}
```

 is formatted into:

```rust
script;

fn main() {
    if foo { let x = true; }
    // comment
    else { bar(x); }
}
```

This looks acceptable but IMO it probably shouldn't be inlined.

### LINKED ISSUE #3853: formatter: if-else formatting indentation issue
In the referenced PR, we discovered that the method of inserting comments post-formatting led to some fundamental issues, mostly because inserting comments post-formatting the source code results in a lack of contextual knowledge for inserting comments at that point in formatting, causing confusion when trying to insert comments (is the formatted code changed from inline to multiline, or vice versa? Are there comments between places we don't expect?)

```rust
contract;

    fn is_blue() -> bool {
        if self == PrimaryColor::Blue {
            true
        }
        // TODO
            // remove this else when exhaustive ifs are checked for
    else {
            false
        }
    }
};
```

> The formatter entirely ignores the above misindentation.

_Originally posted by @bingcicle in https://github.com/FuelLabs/sway/issues/3785#issuecomment-1384291535_

bingcicle:
closes #3853
closes #3574

## Overview

Currently, comment formatting is done post-formatting of the source code, which can make it difficult to reason about formatting comments in certain scenarios. There have been a few cases (https://github.com/FuelLabs/sway/issues/3574, https://github.com/FuelLabs/sway/issues/2888, https://github.com/FuelLabs/sway/issues/2649) where comments are often wrongly indented.

In #3789, there is extensive discussion on an alternative way of handling comment formatting. It seems like we're at a similar problem that rustfmt originally encountered since comments are not part of the AST. 

## Approach

Instead of formatting comments _after_ formatting the source code as a whole w/o comments, we now format comments locally within each `Format` implementation of each AST node. Few advantages of doing this:

1) We now have the context of the formatter at the point of which the comments are being added, e.g. we can use `formatter.shape.indent` to indent comments, especially in odd locations.

2) Easier handling of  `LineStyle::Multiline` <> `LineStyle::Inline` conversions, when comments are involved. You can see this from how we can check if comments are present and we can decide which `LineStyle` to use in the `}` and `else` formatting of comments.

To make sure that we don't do too much extra work, we use `comment_map_from_src` to init a `CommentMap` at the point where formatting begins instead of during `handle_comments` - we then reuse this `CommentMap` for the entire pass of the formatting. This is temporary and the end goal is to deprecate comment insertion post-formatting the source code, and to do it all in one pass.

To ensure backward compatibility, we _still_ call `handle_comments` for now until we refactor everything to use the new method. Thankfully we have extensive tests that will prevent regression while this refactor is happening.

This will be a gradual refactor and the above 2 issues were selected since they were difficult to fix with the old model of comment insertion, but much easier with this new model.


swayfmt/src/comments.rs
```diff
@@ -0,0 +1,66 @@
+use std::fmt::Write;
+use sway_ast::token::Comment;
+use sway_types::Spanned;
+
+use crate::{formatter::FormattedCode, Formatter, FormatterError};
+
+// Given a start, an end and a CommentMap, return references to all comments
+// contained within a given start and end of a span, in an exclusive range.
+pub fn get_comments_between(start: usize, end: usize, formatter: &mut Formatter) -> Vec<&Comment> {
```
mitchmindtree, line 9:
```suggestion
pub fn get_comments_between(start: usize, end: usize, formatter: &Formatter) -> Vec<&Comment> {
```
This function does not mutate the `Formatter`, so we should not take a `&mut`.

swayfmt/src/comments.rs
```diff
@@ -0,0 +1,66 @@
...
+    let mut comments = vec![];
+    let iter = formatter.comment_map.clone().into_keys();
```
mitchmindtree, line 11: This clones the entire `comment_map` just to iterate over the keys. Instead, you could use the `keys()` iterator to iterate over the keys by reference, e.g. `formatter.comment_map.keys()`.

swayfmt/src/comments.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    for bs in iter {
+        if bs.start > start && bs.end < end {
```
mitchmindtree, line 14: The `comment_map` is a `BTreeMap`. The `BTreeMap` stores its elements using ordered keys, and actually [provides a dedicated method](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.range) for efficiently retrieving elements within a given range.

swayfmt/src/comments.rs
```diff
@@ -0,0 +1,66 @@
...
+            if let Some(comment) = formatter.comment_map.get(&bs) {
+                comments.push(comment)
+            }
+        }
+    }
+
+    comments
+}
```
mitchmindtree, line 22: This function currently collects all comments into a `Vec` before returning, which requires dynamically allocating the `Vec`. To avoid this, we could rewrite this function to return an `Iterator`, e.g. `-> impl Iterator<Item = &'_ Comment>`.

However, seeing as `comment_map` already provides a `range` method, we can likely omit this function altogether in favour of using something like:

```rust
formatter.comment_map.range(start..end)
```

Also just a note: `get` is generally reserved for datastructures that can `get` a reference to some element given an index or key. Here, it's safe to omit it from the function name (i.e. `comments_between`) as it's still clear what is being returned.

swayfmt/src/comments.rs
```diff
@@ -0,0 +1,66 @@
...
+    for bs in iter {
+        if bs.start > start && bs.end < end {
+            if let Some(comment) = formatter.comment_map.remove(&bs) {
+                comments.push(comment)
+            }
+        }
+    }
+
+    comments
+}
```
mitchmindtree, line 39: `BTreeMap` does not have an equivalent for `Range` that also removes the yielded elements, however it does have a [`drain_filter` method](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.drain_filter).

A more efficient (performance and code-wise) alternative to this function using `drain_filter` might look like:

```rust
formatter.comment_map.drain_filter(|bs, _v| bs.start >= start && bs.end <= end)
```

swayfmt/src/comments.rs
```diff
@@ -0,0 +1,66 @@
...
+
+// Writes comments between a given start and end and removes them from the formatter's CommentMap.
+// This returns Ok(true) on successful execution AND comments were written, but returns
+// Ok(false) on successful execution without any comments written. Returns Err on failure.
+pub fn maybe_write_comments_from_map(
+    formatted_code: &mut FormattedCode,
+    start: usize,
+    end: usize,
```
mitchmindtree, line 47: Rather than specifying `start` and `end` separately, another option is to accept a range, e.g.

```suggestion
    range: std::ops::Range<usize>,
```

mitchmindtree (changes requested):
Nice, this looks like we're on the right track :+1: Just a few things to clean up first.

swayfmt/src/comments.rs
bingcicle: We can't borrow the `Formatter` immutably here, since in the `Format` impls the formatter is always borrowed mutably. Is there a way to work around this?

mitchmindtree: You can always pass a `&mut T` instance to a `&T` argument of a function, but you cannot pass a `&T` to a `&mut T` argument, as that would break Rust's ownership rules.

kayagokalp, line 9: If I understood the usage [here](https://github.com/FuelLabs/sway/blob/d6d2fa945bd74878cdbe5b5b6720c541e34b52d4/swayfmt/src/utils/language/expr/conditional.rs#L28) correctly, I wonder if it would make sense to accept a `std::Range` and also provide a way to convert `Span` to `Range` so that it would be easy to use the function while searching in an item.

swayfmt/src/comments.rs
```diff
@@ -0,0 +1,66 @@
+use std::fmt::Write;
+use sway_ast::token::Comment;
+use sway_types::Spanned;
+
+use crate::{formatter::FormattedCode, Formatter, FormatterError};
+
+// Given a start, an end and a CommentMap, return references to all comments
+// contained within a given start and end of a span, in an exclusive range.
```
kayagokalp, line 8: Maybe we should consider converting them into doc comments i.e `//` -> `///`

swayfmt/src/comments.rs
```diff
@@ -0,0 +1,66 @@
...
+                comments.push(comment)
+            }
+        }
+    }
+
+    comments
+}
+
+// Given a start, an end and a CommentMap, removes and returns all comments in a
+// CommentMap contained within a given start and end of a span, in an exclusive range.
```
kayagokalp, line 25: Maybe we should consider converting them into doc comments i.e // -> ///

swayfmt/src/comments.rs
```diff
@@ -0,0 +1,66 @@
...
+            }
+        }
+    }
+
+    comments
+}
+
+// Writes comments between a given start and end and removes them from the formatter's CommentMap.
+// This returns Ok(true) on successful execution AND comments were written, but returns
+// Ok(false) on successful execution without any comments written. Returns Err on failure.
```
kayagokalp, line 43: Maybe we should consider converting them into doc comments i.e // -> ///

kayagokalp (commented):
Very excited for this 🚀 left some nits nothing vital

swayfmt/src/comments.rs
bingcicle: `drain_filter` seems to be unstable atm - we can consider adding this when it's stabilized. Will create a good first issue for this

mitchmindtree: Ahh apologies, I missed the nightly label!

swayfmt/src/comments.rs
bingcicle: So to be able to get a range over `CommentMap` I had to implement `RangeBounds` for `ByteSpan`, but this means the range over the `CommentMap` takes a starting `ByteSpan` and an ending `ByteSpan`. We not only want the comments between `ByteSpans` but we also might want comments within a `ByteSpan`. Would this still be appropriate here? eg. within the `IfExpr` i'd like to check if there are comments self-contained between `self.span().start` and `self.span().end`.

mitchmindtree: Ahh good point! For some reason I was thinking of the keys as indices rather than the full byte spans. In that case, perhaps we are best to keep this `comments_between` function (though have it take a `std::ops::Range<usize>` as input and return an iterator).

swayfmt/src/comments.rs
bingcicle: Updated this to `comments_between()` - creating an iterator and returning it instead as you suggested below.

swayfmt/src/comments.rs
bingcicle: Comment from above:

> Updated this to `comments_between()` - creating an iterator and returning it instead as you suggested below.

swayfmt/src/comments.rs
bingcicle: We now accept a `std::ops::Range` in `comments_between()`.

swayfmt/src/comments.rs
```diff
@@ -0,0 +1,67 @@
+use std::{fmt::Write, ops::Range};
+use sway_ast::token::Comment;
+use sway_types::Spanned;
+
+use crate::{formatter::FormattedCode, Formatter, FormatterError};
+
+/// Given a range, return an iterator to comments contained within the range.
+pub fn comments_between<'a>(
+    range: &'a Range<usize>,
```
mitchmindtree, line 9:
```suggestion
    range: Range<usize>,
```
`Range` is just copy data and its more common to take it by value. That said, `Range` notably doesn't actually implement `Copy`! The reason `Copy` was omitted for `Range` is that `Range` is also an `Iterator` type, and its easy to make confusing mistakes during iteration when `Iterator`s are `Copy` (e.g. it's easy to accidentally call `.next()` on a copy rather than the original iterator, causing infinite loops or unexpected results).

swayfmt/src/comments.rs
mitchmindtree: I wonder if it's worth making `CommentMap` an actual type, e.g.

```rust
struct CommentMap(BTreeMap<ByteSpan, Comment>);
```
Then we could implement `Deref` and `DerefMut` targeting the inner `BTreeMap` to provide all the original methods, while adding functions like `comments_between` as actual *methods* on the comment map :thinking:

mitchmindtree (approved):
:ship:

swayfmt/src/comments.rs
bingcicle: correct me if there's a better way but I'm taking to take a reference here because `range` is being used within the `filter_map()` closure (which is trying to capture its value). Would its usage here in this case be a footgun?

bingcicle:
I'll create an issue about making `CommentMap` a type :+1:

kayagokalp (approved):
🚀

PR #3855, opened 2023-01-23
Fix command for updating 'forc' from cargo

mehtavishwa30:
Replace 'install' with 'update' to correct the command for updating 'forc' using cargo


mohammadfawaz (approved):

bingcicle (approved):

bingcicle:
Hmm, I think the merge got rid of the changes here...?

PR #3858, opened 2023-01-23
feat(fmt): allow a single file for `--path`

bingcicle:
Allows you to format a single file with defaults for `Formatter` and `BuildConfig`.

Usage: `forc fmt --path my_file.sw`


forc-plugins/forc-fmt/src/main.rs
```diff
@@ -46,16 +46,43 @@ fn main() {
...
         Some(path) => PathBuf::from(path),
         None => std::env::current_dir()?,
     };
-    let manifest_file = forc_pkg::manifest::ManifestFile::from_dir(&dir)?;
-    for (_, member_path) in manifest_file.member_manifests()? {
-        let member_dir = member_path.dir();
-        let mut formatter = Formatter::from_dir(member_dir)?;
-        format_pkg_at_dir(&app, &dir, &mut formatter)?;
+
+    if path.is_dir() {
```
mitchmindtree, line 54: Previously, I believe `--path` worked for both the manifest directory *as well as* a path directly to a manifest file, e.g. `--path /path/to/project/Forc.toml`. This change would instead result in an error if a project manifest was specified, which is probably not what we want.

Perhaps we can keep the old behaviour as well with something like:

```rust
    if is_sway_file(&path) {
        format_single_file(&path)?;
    } else {
        let manifest_file = forc_pkg::manifest::ManifestFile::from_dir(&path)?;
        // ...
```

bingcicle: Ahh, I was not aware of this. Should we adopt rustfmt's defaults then?

```console
$ rustfmt --help
usage: rustfmt [options] <file>...
```

So,

`forc fmt` and `forc fmt --path <project>`  will format the project,
but `forc fmt <file>` targets a single file.

bingcicle: Also, if this is a bad idea we can ditch this PR as well - I created it on a whim because I was testing out the formatting and trying to execute `forc fmt` on a single file but realized such an option wasn't possible.

bingcicle:
Updated the implementation here to be more alike `rustfmt`:

`forc fmt` will format the project in the current working directory, and `forc fmt --path <project>` will format the project at path, like before.

`forc fmt <file>` will format a single file, if the `file` argument is given.

sdankel (approved):
lgtm!

mitchmindtree (approved):

Issue #3318, opened 2022-11-07
Revamp the `storage` library to use the new capabilities of storage opcodes


mohammadfawaz:
Specifically, the `$rB` registers now tells whether the given storage slot has already been set or not (See [`SRW`](https://fuellabs.github.io/fuel-specs/master/vm/instruction_set.html#srw-state-read-word), [`SRWQ`](https://fuellabs.github.io/fuel-specs/master/vm/instruction_set.html#srwq-state-read-sequential-32-byte-slots), [`SWW`](https://fuellabs.github.io/fuel-specs/master/vm/instruction_set.html#sww-state-write-word), and [`SWWQ`](https://fuellabs.github.io/fuel-specs/master/vm/instruction_set.html#swwq-state-write-sequential-32-byte-slots)). This is useful because we can now return an `Option` from storage operations in the standard library. Tasks required:

### Update Storage Intrinsics
Currently, storage intrinsics look like:
* `__state_load_word(key: b256) -> u64;`
* `__state_store_word(key: b256, value: u64);`
* `__state_load_quad(key: b256, addr: u64);`
* `__state_store_quad(key: b256, addr: u64);`
Change them to the following:
* `__state_load_word(key: b256) -> (u64, bool);` or `__state_load_(key: b256, addr: u64) -> bool;`
* `__state_store_word(key: b256, value: u64) -> bool;`
* `__state_load_quad(key: b256, addr: u64) -> bool;`
* `__state_store_quad(key: b256, addr: u64) -> bool;`
In each case, the `bool` represent the registers `$rB`.

### Update `store` and `get` in `std::storage`
Update `store` and `get` to look like:
```rust
pub fn store<T>(key: b256, value: T) -> Option<T>;
pub fn get<T>(key: b256) -> Option<T>;
```
The `bool` value returned from the intrinsics have to be ANDed (or ORed?) and the result is used to decide whether to return `Some` or `None`. Typically, `store` would return `Some(previous_value)` if the key was written before, and `None` otherwise, but this is potentially too expensive because we would have to always read the storage location first before writing. One alternative is to make two versions of `store`: one that doesn't return anything and one that returns `Option<T>`. We can also have `store` return a bool:
```rust
pub fn store<T>(key: b256, value: T) -> bool; // The `bool` tells us whether the `key` is already paired with a value or not.
```

For `get`, the `Option` is very useful and doesn't require any extra reads or writes to generate.

### Update `StorageMap`
[...]

PR #3862, opened 2023-01-23
Update `std::storage::get` and `std::storage::StorageMap::get` to return an `Option`

mohammadfawaz:
API change:
Previous:
```rust
pub fn get<T>(key: b256) -> T;
pub fn get(self, key: K) -> V;
```
Now:
```rust
pub fn get<T>(key: b256) -> Option<T>;
pub fn get(self, key: K) -> Option<V>;
```
`Option::None` indicates that the storage slot requested in not available.

- Updated all storage intrinsics to return a `bool` except for `__state_load_word` because it already returns the value loaded. The `bool` returned represents the previous state of the storage slots as described in the specs. I was not able to update `__state_load_word` to return both the value loaded and the `bool` unfortunately because I hit various hiccups along the way, particularly in IR/codegen.
- Updated `get` and `StorageMap::get` in the standard library to return an `Option` based on the Boolean mentioned above. In the copy type case, I had to use `asm` blocks instead of `__state_load_word` until we're able to return a struct from an intrinsic.
- I did not update `store` and `StorageMap::insert` to return an `Option`, for now, because that would involve an extra storage read to return the previous value, if available. We can think about whether this is worth it at some point.
- I added `unwrap` and `unwrap_or` where it makes sense in `StoargeVec`.
- Update various docs, examples, and tests.

Closes https://github.com/FuelLabs/sway/issues/3318

I will open separate issues for updating `__state_load_word` and thinking about making `store` and `insert` also return an `Option`.


Braqzen (approved):
Looks reasonable

jtriley-eth (approved):
LGTM!

PR #3867, opened 2023-01-23
Remove decl engine cloning for built package generation

### LINKED ISSUE #3859: Enormous memory usage while building a large workspace
Recently @Salka1988 raised an issue over slack stating that building their workspace located in this [pr](https://github.com/FuelLabs/fuels-rs/pull/792)  taking up more than 32 gb of ram and failing thus failing to build. After @mitchmindtree's investigation with valgrind unvailed that a heavy `DeclEngine::clone` was happening. After digging through the following code piece was flagged https://github.com/FuelLabs/sway/blob/313f6b6ea99d04a27221728657d1a8f66116f53d/forc-pkg/src/pkg.rs#L2517.

This is probably introduced to get `test_pass_condition` detection compiling in `forc-test` while removing the `lazy_static` declaration engine. We should remove the clone and collect necessary information from `forc-pkg` and pass on-to `forc-test`. Adding critical as it is probably blocking for the linked PR.

kayagokalp:
I couldn't test it yet as i need modify tests from @Salka1988's branch but feel like this should fix the issue
closes #3859.


forc-pkg/src/pkg.rs
```diff
@@ -2523,6 +2555,70 @@ pub fn compile(
     }
 }
 
+impl PkgEntry {
+    /// Tries to retrieve the `PkgEntry` as a `PkgTestEntry`, panics otherwise.
+    pub fn expect_test_entry(&self) -> Result<PkgTestEntry> {
```
mitchmindtree, line 2560: We can let the method user choose whether or not they incur the cost of the `clone`.

```suggestion
    pub fn expect_test_entry(&self) -> Result<&PkgTestEntry> {
```

It's worth keeping in mind, the Rust std sets the convention that `expect` is generally used to indicate that the method will `panic!` if the expected outcome isn't the case (e.g. `Option::expect`, `Result::expect`), whereas here we return a `Result`. 

It might be better to add this method to the `PkgEntryKind` type as that is the only data that is required to check this.

Something like the following might give the method user more flexibility:
```rust
impl PkgEntryKind {
    /// Returns `Some` if the `PkgEntryKind` is `Test`.
    pub fn test(&self) -> Option<PkgEntryTest> {
        match self {
            PkgEntryKind::Test(test) => Some(test),
            _ => None,
        }
    }
}
```
We can still use it to expect a value:
```rust
let test_entry = pkg_entry.kind.test().expect("expected test entry point");
```
or return an error:
```rust
let test_entry = pkg_entry.kind.test().ok_or(anyhow!("expected test entry point"))?;
```
It also removes the need for the `PkgEntry::is_test` method as we can do this:
```rust
pkg_entry.kind.test().is_some()
```

None of these notes are super important, but I just wanted to point out these patterns as they can show up a lot in Rust library design around these sorts of `*Kind` enums!

forc-test/src/lib.rs
```diff
@@ -197,13 +190,13 @@ impl<'a> PackageTests {
         let pkg_with_tests = self.built_pkg_with_tests();
         // TODO: We can easily parallelise this, but let's wait until testing is stable first.
         let tests = pkg_with_tests
-            .entries
+            .pkg_entries
             .iter()
             .filter(|entry| entry.is_test())
             .map(|entry| {
```
mitchmindtree, line 196: Another option could be something like:

```suggestion
            .filter_map(|entry| entry.kind.test().map(|test| (entry, test)))
            .map(|(entry, test_entry)| {
```
This might allow us to remove the second unnecessary `entry.expect_test_entry()` check below

forc-pkg/src/pkg.rs
```diff
@@ -92,14 +93,41 @@ pub struct BuiltPackage {
     pub json_abi_program: ProgramABI,
     pub storage_slots: Vec<StorageSlot>,
     pub bytecode: Vec<u8>,
-    pub entries: Vec<FinalizedEntry>,
+    pub pkg_entries: Vec<PkgEntry>,
```
mitchmindtree, line 96: I think `entries` is fine as a field name, as the association with pkg is inferred from both the type and likely also the variable names when it's used, e.g.
```rust
built_pkg.entries
```
rather than
```rust
built_pkg.pkg_entries
```

forc-pkg/src/pkg.rs
```diff
@@ -2511,10 +2544,9 @@ pub fn compile(
                 storage_slots,
                 bytecode,
                 tree_type,
-                entries,
+                pkg_entries,
                 source_map: source_map.to_owned(),
                 pkg_name: pkg.name.clone(),
-                decl_engine: engines.de().clone(),
```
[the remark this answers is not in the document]
kayagokalp: Yep, just removed it. Thanks for pointing it!

mitchmindtree (approved):
:ship:

mohammadfawaz (approved):

PR #3869, opened 2023-01-23
feat: Adds Markdown formatting to item previews in `forc doc`

### LINKED ISSUE #3868: Add Markdown formatting to item previews

eureka-cpu:
Closes #3868 

Adds markdown support and a few layers of logic to figure out where a preview should end.
Screen Shot 2023-01-23 at 5 18 57 PM


JoshuaBatty (dismissed):

forc-plugins/forc-doc/src/render.rs
```diff
@@ -1231,3 +1231,46 @@ pub(crate) fn trim_fn_body(f: String) -> String {
...
+    const H2: &str = "<h2>";
+    const H3: &str = "<h3>";
+    const H4: &str = "<h4>";
+    const H5: &str = "<h5>";
+    if raw_html.contains(H1) {
+        let v: Vec<_> = raw_html.split(H1).collect();
+        match v.first() {
+            Some(s) => s,
+            None => panic!("expected some &str"),
+        }
```
mitchmindtree, line 1248: Rather than branching and `panic!`ing manually, we can use `expect`:
```suggestion
        v.first().expect("expected a non-empty str")
```
Same goes for the branches below.

mitchmindtree (commented):
Nice, preview's looking good! Just a couple small suggestions.

forc-plugins/forc-doc/src/doc.rs
```diff
@@ -49,17 +49,27 @@ impl Document {
...
-        match &self.raw_attributes {
-            Some(description) => {
-                if description.len() > 200 {
-                    let (first, _) = description.split_at(75);
-                    Some(format!("{}...", first.trim_end()))
-                } else {
-                    Some(description.to_string())
+        self.raw_attributes.as_ref().map(|description| {
+            let preview = split_at_markdown_header(description);
+            if preview.len() > 100 && preview.contains("</p>") {
```
mitchmindtree, line 54: We should make these magic values like `100` and `5` consts, e.g. `const PREVIEW_LEN_LIMIT: usize = 100;`, especially for those that are used multiple times.

Also, keep in mind that checking `len` only checks the *byte* length. The byte length is often not the best way to check visible length, as in utf8 some characters are represented with more bytes than others. Often it's a bit more reliable to check against `.chars().count()`, e.g.
```suggestion
            if preview.chars().count() > MAX_PREVIEW_CHARS && preview.contains("</p>") {
```

forc-plugins/forc-doc/src/doc.rs
```diff
@@ -49,17 +49,35 @@ impl Document {
...
+            let preview = split_at_markdown_header(description);
+            if preview.chars().count() > MAX_PREVIEW_CHARS
+                && preview.contains(CLOSING_PARAGRAPH_TAG)
+            {
+                match preview.find(CLOSING_PARAGRAPH_TAG) {
+                    Some(index) => {
+                        // We add 1 here to get the index of the char after the closing tag.
+                        // This ensures we retain the closing tag and don't break the html.
+                        let (preview, _) =
+                            preview.split_at(index + CLOSING_PARAGRAPH_TAG.chars().count() + 1);
```
mitchmindtree, line 65: Careful: `split_at` expects a byte index, not a character index: https://doc.rust-lang.org/std/primitive.str.html#method.split_at. Here, `index` is a byte index produced by `find`, but the right-hand-side of the add is in `chars`.

In general, working with `str` requires being careful about distinguishing between bytes and chars, so it's always worth checking what methods like `split_at` or `find` expect. It gets even more complicated when working with graphemes and emoji :sweat:

eureka-cpu: Ah thanks, I'll read more into this tomorrow morning!

eureka-cpu: Okay, so I took a look just now. I've altered it to check in `chars` for the conditional statement but then use the `bytes` for splitting since it's checking for a specific index. I can't check how it looks just yet because the `forc` building issue, but this is essentially what I had before, save that the initial conditional statement now checks for the length in `chars`.

eureka-cpu: Just checked with `--locked` and looks like it's functioning as intended

mitchmindtree (approved):

bingcicle (approved):
This looks awesome :rocket:

PR #3870, opened 2023-01-24
Refactor doc strings in `sway-lib-std`

### LINKED ISSUE #3850: Make docstrings more consistent in `sway-lib-std`
Some docstrings start with caps, other don't. Some code snippets are wrapped in backticks while others are not. There are also plenty of typeos.

gr00vytvniks:
I went through the std library and fixed any inconsistencies, typos, punctuation errors and fixed up the markdown where applicable.

Closes #3850


gr00vytvniks:
Also thanks for the help @eureka-cpu :)

sway-lib-std/src/outputs.sw
```diff
@@ -49,8 +48,8 @@ pub fn output_type(index: u64) -> Output {
     }
 }
 
-/// Get a pointer to the Ouput at `index`
-/// for either tx type (transaction-script or transaction-create).
+/// Get a pointer to the ouput at `index`
```
otrho, line 51:
```suggestion
/// Get a pointer to the output at `index`
```
Missed one. :wink:

sway-lib-std/src/vec.sw
```diff
@@ -20,8 +20,8 @@ impl<T> RawVec<T> {
     }
 
     /// Creates a `RawVec` (on the heap) with exactly the capacity for a
-    /// `[T; capacity]`. This is equivalent to calling `RawVec::new` when
-    /// `capacity` is `0`.
+    /// `T; capacity`. This is equivalent to calling `RawVec::new` when
```
otrho, line 23: Hrmm, we either need to put the `[...]` back, escaped maybe, or word this better.  `T; capacity` by itself doesn't make sense.

otrho (commented):
Nice!  Generally approve of the changes here.  Not sure if `forc doc` will support the e.g., `[Option]` style hyperlinking, but it's better to have it consistent for now, one way or the other.  We can add the brackets back consistently if needed later.

Braqzen (commented):
Generally an improvement. Agree with the `[...]` comment. Will approve after.

sway-lib-std/src/token.sw
```diff
@@ -43,16 +43,16 @@ pub fn mint_to(amount: u64, to: Identity) {
...
-/// # WARNING
-///
-/// This will transfer coins to a contract even with no way to retrieve them
-/// (i.e: no withdrawal functionality on the receiving contract), possibly leading to
-/// the PERMANENT LOSS OF COINS if not used with care.
+/// > **_WARNING:_**
+/// >
+/// > This will transfer coins to a contract even with no way to retrieve them
+/// > (i.e: no withdrawal functionality on the receiving contract), possibly leading to
+/// > the PERMANENT LOSS OF COINS if not used with care.
```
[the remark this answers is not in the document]
eureka-cpu: Looks like GH is formatting this, but it's the same as on L20

sway-lib-std/src/vec.sw
eureka-cpu: To elaborate, the syntax used is unique to vecs so I can see this accidentally getting removed. Should be `[T; capacity]`, where `T` is the type and `capacity` is the size it can contain.

eureka-cpu (changes requested):
All in all, great job!

sway-lib-std/src/alloc.sw
```diff
@@ -1,11 +1,10 @@
...
-/// Allocates zeroed memory on the heap
+/// Allocates zeroed memory on the heap.
 ///
-/// In FuelVM, the heap begins at `VM_MAX_RAM - 1` and grows downward.
-/// Heap pointer `$hp` will always point to unallocated space.
+/// In the FuelVM, the heap begins at `VM_MAX_RAM - 1` and grows downward.
+/// The heap pointer, `$hp`, will always point to unallocated space.
 ///
 /// Initially the heap will look like this:
 /// ... 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 |
```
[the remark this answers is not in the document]
eureka-cpu: Referring to L10, `... 00 00 ...`

eureka-cpu (approved):
Approving to dismiss my previous request for changes. Looks like something weird is going on with CI, don't worry about it though. Doesn't seem to be related from what I can tell.

otrho (approved):

PR #3876, opened 2023-01-24
Introduce `__state_clear`, `std::storage::clear`, and `std::storage::StorageMap::remove`

### LINKED ISSUE #3160: Support the new `scwq` opcode in the compiler
Ref https://github.com/FuelLabs/fuel-specs/pull/415

We need to figure out how to best use and expose this to users.

mohammadfawaz:
* New intrinsics `__state_clear` that wraps `scwq`.
* Add `std::storage::clear` which simply clears a variable stored in storage... it really just wipes out the underlying storage slots (i.e. set them to `None`).
* Add `std::storage::StorageMap::remove` to remove a `(key, value)` pair from a `StorageMap`.

Closes #3160


sway-lib-std/src/storage.sw
```diff
@@ -169,6 +196,36 @@ impl<K, V> StorageMap<K, V> {
         let key = sha256((key, __get_storage_key()));
         get::<V>(key)
     }
+
+    /// clears a value previously stored using a key
```
Braqzen, line 200:
```suggestion
    /// Clears a value previously stored using a key
```

sway-lib-std/src/storage.sw
```diff
@@ -108,6 +108,33 @@ pub fn get<T>(key: b256) -> Option<T> {
...
+/// let five = 5_u64;
+/// store(ZERO_B256, five);
+/// let cleared = clear::<u64>(ZERO_B256);
+/// assert(cleared);
+/// assert(get::<u64>(ZERO_B256).is_none());
+/// ```
+#[storage(write)]
+pub fn clear<T>(key: b256) -> bool {
+    // Get the number of storage slots needed based on the size of `T`
+    let number_of_slots = (__size_of::<T>() + 31) >> 5;
```
Braqzen, line 132: Is there a reason that these numbers have been selected?

mohammadfawaz: So `(N + 31) >> 5` is a standard way of computing the ceiling of `N / 32`. I made that clearer.

sway-lib-std/src/storage.sw
```diff
@@ -71,11 +71,11 @@ pub fn get<T>(key: b256) -> Option<T> {
...
         // NOTE: we should eventually be using `__state_load_word` here but we are currently unable 
         // to make that intrinsic return two things due to some limitations in IR/codegen.
-        let result_ptr = alloc::<u64>(16); //  
-        asm(key: key, result_ptr: result_ptr, loaded_word, previously_set) {
+        asm(key: key, result_ptr, loaded_word, previously_set) {
+            move result_ptr sp; // Make `result_ptr` point to the current top of the stack
+            cfei i16; // Grow stack by 2 word
             srw  loaded_word previously_set key;
             sw   result_ptr previously_set i0;
             sw   result_ptr loaded_word i1;
```
otrho, line 81:
```suggestion
        let temp_pair = (false, 0_u64);    // Using a `u64` as a placeholder for copy-type `T` value.
        asm(key: key, result_ptr: temp_pair, loaded_word, previously_set) {
            srw  loaded_word previously_set key;
            sw   result_ptr previously_set i0;
            sw   result_ptr loaded_word i1;
```

I think this is right, my ASM-block-fu is a bit rusty.  I _think_ you can init a block arg with a ref type and it uses the address...

Anyway, using a `CFEI` without a `CFSI` is asymmetrical and will break the call stack.  Instead we can use a temporary, populate that, and then cast it to `(bool, T)`.  Ideally we'd use `T::default()` instead of a `u64` in the placeholder but :shrug: .

otrho (commented):
I can't tell -- is there testing to ensure the storage annotations are enforced for functions which only use clear() (and not also write())?

mohammadfawaz:
> I can't tell -- is there testing to ensure the storage annotations are enforced for functions which only use clear() (and not also write())?

Just added one!

otrho (approved):

JoshuaBatty (approved):

otrho (approved):

Issue #3671, opened 2023-01-02
DCA issue when traits are define in library and used in generic function


esdrubal:
When traits are define in library and used in `main.sw` within a generic function we get a `This declaration is never used` for the trait implementation that is used.

**main.sw**
``` rust
script;

dep traits;

use std::assert::assert;
use traits::*;

fn test_my_eq<T>(x: T, y: T) -> bool where T: MyEq {
  x.my_eq(y)
}

fn main() {
  assert(test_my_eq(42, 42));
}
```

**traits.sw**
``` rust
library traits;

pub trait MyEq {
    fn my_eq(self, other: Self) -> bool;
}

impl MyEq for u64 {
    fn my_eq(self, other: Self) -> bool {
        self == other
    }
}
```

```
warning
      --> /home/iota/sway/test/src/e2e_vm_tests/test_programs/should_pass/language/trait_dep/src/traits.sw:3:11
       |
     1 | 
     2 | 
     3 | pub trait MyEq {
       |           ---- This trait is never implemented.
     4 |     fn my_eq(self, other: Self) -> bool;
     5 | }
       |
     ____

     warning
       --> /home/iota/sway/test/src/e2e_vm_tests/test_programs/should_pass/language/trait_dep/src/traits.sw:7:1
        |
      5 |   
      6 |   
      7 | / impl MyEq for u64 {
      8 | |     fn my_eq(self, other: Self) -> bool {
      9 | |         self == other
     10 | |     }
     11 | | }
        | |_- This declaration is never used.
        |
     ____
```

PR #3878, opened 2023-01-24
Fix DCA for generic functions

tritao:
Closes https://github.com/FuelLabs/sway/issues/3671.

There were a bunch of different fixes necessary to fix this one, which are separated into individual commits to make it easier to review. The main gist is that the DCA now keeps track of functions by their signature instead of just their name and when encountering monomorphized functions in the DCA engine, we now look for their counterpart fully generic version and link the two together.

Previously:

bug

Now:

bug3

### DIFF
--- a/docs/book/README.md
+++ b/docs/book/README.md
@@ -23,30 +23,30 @@ You must also install forc plugins that are already documented within the book. 
 ```sh
 cargo install --path ./forc-plugins/forc-fmt
 cargo install --path ./forc-plugins/forc-lsp
-cargo install --path ./forc-plugins/forc-explore
+cargo install forc-explore
 cargo install --path ./forc-plugins/forc-doc
 ```
 
 To build book:
 
 ```sh
-mdbook build
+mdbook build docs/book
 ```
 
 To build the book on strict mode to check if pages should be removed or added within the Forc Reference:
 
 ```sh
-MDBOOK_preprocessor__FORC_documenter__STRICT="true" mdbook build docs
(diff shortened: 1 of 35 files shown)


sway-core/src/language/ty/declaration/declaration.rs
```diff
@@ -466,8 +484,32 @@ impl TyDeclaration {
         }
     }
 
-    /// friendly name string used for error reporting.
-    pub fn friendly_name(&self) -> &'static str {
+    /// friendly type name string used for error reporting.
```
IGI-111, line 487: The doc comments should probably explain the difference between `friendly_name and `friendly_type_name`.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -1107,14 +1111,24 @@ impl ty::TyExpression {
         // Check if this could be a function
         let mut function_probe_warnings = Vec::new();
         let mut function_probe_errors = Vec::new();
-        let maybe_function = {
+
+        let maybe_function_decl = {
```
IGI-111, line 1115: I think this being a `maybe_` is confusing since those are used in the match to differentiate the types and it looks like it's only relevant in the function case.

The decl fetching should probably be moved into the `maybe_function` block and then later actually use `_fun_decl` instead of ignoring it.

Moreover, though that's out of scope, this entire function should probably be broken up so there's a clear interface and enforced pattern instead of mixing scopes.

IGI-111 (approved):
LGTM

vaivaswatha (approved):

PR #3882, opened 2023-01-24
Impl From<raw_slice> for Bytes

### LINKED ISSUE #3844: Add methods for converting between `Bytes` and `raw_slice`
Follow the example set by the `Vec` library.
This should wait until #3637 is resolved to avoid implementing and testing an `.into()` method that is destined to be removed.

nfurfaro:
Follows the implementation of this trait for Vec.

I had to add some duplicate functions to `std::raw_slice` to make this work, which I also had to do in `std::raw_ptr` when I originally implemented the Bytes lib. The reason is that some functions I needed to use from those modules are generic, and use the `__size_of` intrinsic. But, when working with bytes, there's no appropriate type we can pass to a generic function, `u8` being the closest candidate, but its 8 bytes instead of 1.

closes #3844


sway-lib-core/src/raw_slice.sw
```diff
@@ -22,6 +22,11 @@ impl raw_slice {
         from_parts((ptr, __mul(count, __size_of::<T>())))
     }
 
+    /// Forms a slice from a pointer and a length.
+    pub fn from_raw_bytes_parts(ptr: raw_ptr, count: u64) -> Self {
```
nfurfaro, line 26: hmm. I realized that I only use this function for testing.

sway-lib-core/src/raw_slice.sw
```diff
@@ -31,4 +36,8 @@ impl raw_slice {
     pub fn len<T>(self) -> u64 {
         __div(into_parts(self).1, __size_of::<T>())
     }
+
+    pub fn raw_bytes_len(self) -> u64 {
```
nfurfaro, line 40: I don't like this name.

otrho (dismissed):
I see we still use `asm` blocks for casts quite a bit.  We need to think again about introducing `as` to replace many of these.

mohammadfawaz (dismissed):

nfurfaro:
Dang. Blocked by https://github.com/FuelLabs/sway/issues/3633
cc @esdrubal

mohammadfawaz:
Moving to a draft until #3633 goes in.

nfurfaro:
This is still blocked, but #4180 should clear the way to landing this.

PR #3883, opened 2023-01-24
Better token handling inside turbofish

IGI-111:
This builds on top of #3847 and addresses two main issues:

1. This adds support for struct instantiation, expressions like `B::<Option<u32>>{}` will now produce proper tokens inside turbofish.

2. The granularity of tokens is improved, instead of only producing one token for the final top level type arguments, we resolve types recursively inside of type arguments and produce appropriate tokens. Expressions like `fun::<Option<Result<u32, u32>>>(Option::None);` will produce different tokens for `Option` and `Result` that both point to valid definitions.

As a sidenote, this also fixes an issue where the information produced by type checking of call path bindings would get thrown away.


JoshuaBatty:
Nice! This looks good from the LSP side. Happy to approve once CI passes.

JoshuaBatty (approved):

sway-core/src/type_system/engine.rs
```diff
@@ -205,6 +200,45 @@ impl TypeEngine {
         }
     }
 
+    /// Recursively resolve type arguments
+    #[allow(clippy::too_many_arguments)]
+    fn resolve_type_arguments(
```
emilyaherbert, line 205: Would you restructure this so that we can more efficiently capture errors when we use this method? change the function signature to:

```rust
    fn resolve_type_arguments(
        &self,
        type_arguments: &mut [TypeArgument],
        decl_engine: &DeclEngine,
        enforce_type_arguments: EnforceTypeArguments,
        namespace: &mut Namespace,
        mod_path: &Path,
    ) -> CompileResult<()>
```

Instead of this being a recursive function, would you change `resolve_type_arguments` to iterate through the elements of `type_arguments`? Then at the end of `resolve_type_arguments`, will you return either an `ok((), warnings, errors)` or an `err(warnings, errors)`, depending on if `errors` is empty? This way we can append any warnings or errors found and return an `err` in that case, while still attempting to resolve each type argument.

sway-core/src/type_system/engine.rs
```diff
@@ -172,22 +172,17 @@ impl TypeEngine {
...
-                            namespace,
-                            mod_path
-                        ),
-                        self.insert(decl_engine, TypeInfo::ErrorRecovery),
-                        warnings,
-                        errors
-                    );
-                }
+
+                self.resolve_type_arguments(
```
emilyaherbert, line 176: Note my other comment about restructuring the `resolve_type_arguments` method. --- if the changes from that comment are implemented, then at this call site you can use the `check!` macro again like so:

```rust
check!(
  self.resolve_type_arguments(......),
  return err(warnings, errors),
  warnings,
  errors,
);
```

This ensures that we only return will `err` if the whole of `resolve_type_arguments` returns an error.

However---the previous source code didn't account for this, so this request is just a code style one really 😃 I'm happy to discuss.

sway-core/src/language/ty/expression/expression_variant.rs
```diff
@@ -51,6 +51,7 @@ pub enum TyExpressionVariant {
         struct_name: Ident,
         fields: Vec<TyStructExpressionField>,
         span: Span,
+        type_binding: TypeBinding<()>,
```
emilyaherbert, line 54: Would it makes sense to combine the `struct_name` and `type_binding` fields?

sway-core/src/type_system/type_argument.rs
```diff
@@ -56,6 +58,8 @@ impl From<&TypeParameter> for TypeArgument {
             type_id: type_param.type_id,
             initial_type_id: type_param.initial_type_id,
             span: type_param.name_ident.span(),
+            name_span: None,
+            type_arguments: vec![],
```
emilyaherbert, line 62: Can you elaborate on why you have added this field? Or give a code example that would trigger this condition?

emilyaherbert (changes requested):
Can you elaborate on why you have added the `type_arguments` field to `TypeArgument`?

Otherwise the rest are just code style suggestions.

sway-core/src/type_system/engine.rs
IGI-111: I suppose I could turn this into a stack based implementation, I was concerned that it would be harder to understand what it's doing at a glance but it would indeed make error handling clearer.

We have to evaluate the whole tree of arguments either way though.

sway-core/src/language/ty/expression/expression_variant.rs
IGI-111: I've pondered this but I think it makes more sense to keep them separate, `TypeBinding<()>` is pretty clearly meant to be consumed by the LSP as used so far and making generic over random other related stuff seems like a bad idea.

In fact, I've been thinking about doing away with the generic argument for it altogether and splitting  `TypeBinding<()>`, `TypeBinding<CallPath<(TypeInfo, Span)>>` and `TypeBinding<CallPath>` into properly separate types or at least separate struct fields. The composition doesn't seem very ergonomic to me as it is.
What do you think?

sway-core/src/type_system/type_argument.rs
IGI-111: Yeah this requires some explanation.

So let's say we're compiling `fun::<Option<Result<u32, u32>>>();` We want individual tokens that point to definitions for both `Option` and `Result` but the type system as it stands is only interested in figuring out what the type of `Option<Result<u32, u32>>` is. There is only a single instance of TypeArgument.

What we need is a way to store the entire tree so we can evaluate the type of every node (and store the associated `name_span`s). This is what this field allows: a `TypeArgument` potentially stores children that will have to be resolved as well.

The need for recursive evaluation of the arguments elsewhere might make more sense in light of that.

sway-core/src/language/ty/expression/expression_variant.rs
emilyaherbert: > In fact, I've been thinking about doing away with the generic argument for it altogether and splitting TypeBinding<()>, TypeBinding<CallPath<(TypeInfo, Span)>> and TypeBinding<CallPath> into properly separate types or at least separate struct fields. The composition doesn't seem very ergonomic to me as it is.
What do you think?

The original design of `TypeBinding` was intended to be modular and """left-associated""". So instead of both the `StructExpression` variant, the `FunctionApplication` variant, etc, keeping track of the fact that their calling conventions could take a turbofish, this would be """left-associated""" with the """name""" field. For this reason, I prefer `TypeBinding<CallPath>` over a specific `type_binding` field on `StructExpression`.

But you're right, `TypeBinding` is not very ergonomic. I'm wondering if there is a better compromise squashing `TypeBinding` into `CallPath`? I'm curious to look at other designs---take a crack at it if you're interested!

If you do decide to rework `TypeBinding`, I recommend that we merge this PR and that you open a new issue and a new PR for this specifically.

IGI-111: Agreed, I think this is out of scope here but I'll definitely look into refactoring it in a later PR.

sway-core/src/type_system/type_argument.rs
emilyaherbert: Adding a `type_arguments` field on `TypeArgument` will create duplicated information---the type information in `type_id` contains the same types that would be `type_arguments`, granted without the `Span` information as you mentioned.

Greater than this problem though---is it accurate to say that the LSP is unable to find tokens for most types, in general?

IGI-111: >Greater than this problem though---is it accurate to say that the LSP is unable to find tokens for most types, in general?

I don't believe it's accurate, there are missing tokens in some edge cases but it does produce the correct info in general.

>Adding a type_arguments field on TypeArgument will create duplicated information---the type information in type_id contains the same types that would be type_arguments, granted without the Span information as you mentioned.

I think I might be able to get away with just storing the spans in the `TypeArgument` and reconstruct the tree in the LSP, which should simplify this PR a lot if it works.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression/struct_instantiation.rs
```diff
@@ -21,13 +21,18 @@ pub(crate) fn struct_instantiation(
     let decl_engine = ctx.decl_engine;
     let engines = ctx.engines();
 
+    // Populate call_path_binding with real types
```
emilyaherbert, line 24: Can you add a longer comment here explaining what this block of code is doing? LOC 25 - 74? In regards to this being for the LSP, the `type_binding` object being used for the LSP, etc.

In fact, I'm pleasantly surprised that LOC 25 isn't conflicting with the hack on LOC 36 for resolving the self type. LOC 36 was added because the `Self` keyword was unable to be parsed in this position (at some point in the past).

emilyaherbert (approved):
LGTM from the compilers side. I appreciate the thoughtful design of the new field on `TypeArgument`.

Just one nit about adding a code comment.

mohammadfawaz (approved):

PR #3892, opened 2023-01-25
Add `ops::*` to the prelude in `core`

### LINKED ISSUE #3891: Incorrect compiler error
While trying to implement the following implementation of an `assert_eq` function, I'm getting the following error:
image

The code is:
```
library assert;

use core::ops::Eq;
use ::logging::log;
use ::revert::revert;


fn assert_eq<T>(v1: T, v2: T) where T: Eq {
    if (v1 != v2) {
        log(v1);
        log(v2);
        revert(0xffff_ffff_ffff_0004);
    }
}

#[test()]
fn test_assert_eq_u64() {
    let a = 42;
    let b = 40 + 2;
    assert_eq(a, b);
}
```

emilyaherbert:
This PR fixes the bug described in #3891 by adding `ops` to the prelude in `core`. This change is congruent with Rust's approach as well.

This is a _breaking change_---if any users have any traits manually implemented for which core already has an existing implementation, this will now create an error. A rightful one, IMO, but a breaking change nonetheless.

A few of the tests were written in a way that is not compatible with this change---so this PR rewrites these. The only one of note is that a old `should_pass/language/trait_override_bug` test was refactored to a new `should_fail/redefine_method_from_core` test.

This PR also refactors the trait finding algorithm in `check_if_trait_constraints_are_satisfied_for_type` to reduce it from O(n^2) to O(n).

Closes #3891


nfurfaro (dismissed):
LGTM, but will defer to someone from the compiler team.

Braqzen (approved):
Unsure about `trait_map.rs` but looks reasonable otherwise. Dropping an approval to bump up the count

emilyaherbert:
> Unsure about `trait_map.rs` but looks reasonable otherwise. Dropping an approval to bump up the count

Would it make more sense if I moved that to a different PR @Braqzen? That change came about while debugging, and I didn't want to erase a big-O improvement 😃

Braqzen:
> > Unsure about `trait_map.rs` but looks reasonable otherwise. Dropping an approval to bump up the count
> 
> Would it make more sense if I moved that to a different PR @Braqzen? That change came about while debugging but turned out to be unnecessary, but I didn't want to erase a big-O improvement smiley

No, it's okay. Someone who knows what's going on there should approve though. I was just saying that that my approval is mostly for the other things

otrho (approved):

nfurfaro (approved):
Thank you!

nfurfaro (approved):
LGTM!

Braqzen (approved):

PR #3893, opened 2023-01-25
Better document reentrancy_guard()

### LINKED ISSUE #3216: Rename stdlib reentrancy guard
Either rename or better document the reentrancy guard. The guard only protects against control flow hijacking on a single contract, not multiple trusted contracts working together. While the latter case may be less common on Fuel due to a larger contract size limit, it will still happen.

See more: https://inspexco.medium.com/cross-contract-reentrancy-attack-402d27a02a15

nfurfaro:
closes #3216 

cc @adlerjohn

### DIFF
--- a/sway-lib-std/src/reentrancy.sw
+++ b/sway-lib-std/src/reentrancy.sw
@@ -7,6 +7,9 @@ use ::assert::assert;
 use ::call_frames::*;
 use ::registers::frame_ptr;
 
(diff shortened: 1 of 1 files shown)


mohammadfawaz:
@jtriley-eth is in the process of moving the guard to `sway-libs` in https://github.com/FuelLabs/sway-libs/pull/83. Should we just make this same change there instead?

PR #3895, opened 2023-01-25
Bytes: refactor & rename `join` to `append`

nfurfaro:
This renames both `split` and `join` to `split_at` and `append`, which is consistent with WIP changes to the Vec library, and the rust naming.
This also updates the behaviour and implementation to be both more performant(presumably) and consistent with the implementation of the same functions in the Vec library.


sway-lib-std/src/bytes.sw
```diff
@@ -888,19 +906,19 @@ fn test_bytes_limits() {
 }
 
 #[test()]
-fn test_split() {
+fn test_split_at() {
```
otrho, line 909: Can you add another test or two for edge cases where `mid` is zero and `mid` is `self.len`?

sway-lib-std/src/bytes.sw
```diff
@@ -611,26 +621,34 @@ impl Bytes {
...
+    /// let second_cap = bytes2.capacity();
+    /// bytes.append(bytes2);
+    /// assert(bytes.len() == first_length + second_length);
+    /// assert(bytes.capacity() == first_length + first_length);
     /// ```
-    pub fn join(ref mut self, other: self) -> Self {
-        let mut joined = Bytes::with_capacity(self.len + other.len);
+    pub fn append(ref mut self, ref other: self) {
+        let both_len = self.len + other.len;
+        let other_start = self.len;
```
otrho, line 634: It may be worth returning here if `other.len` is zero?  A rare occurrence but a fair optimisation probably.

Another optimisation would be if `self.len` is zero, instead of copying the bytes, they can just be 'moved' to self, by copying the pointer and avoiding a new alloc.  This might be less rare if the user is building something starting with an empty `Bytes`.

And then adding tests for those cases too.

sway-lib-std/src/bytes.sw
```diff
@@ -631,6 +635,15 @@ impl Bytes {
...
     /// ```
     pub fn append(ref mut self, ref other: self) {
+        if other.len == 0 {
+            return
+        };
+
+        // optimization for when starting with empty bytes and appending to it
+        if self.len == 0 {
+            self.buf.ptr = other.buf.ptr;
+        };
```
otrho, line 645:
```suggestion
        if self.len == 0 {
            self.buf.ptr = other.buf.ptr;
            self.buf.cap = other.buf.cap;
            self.len = other.len;
            other.buf.ptr = alloc_bytes(0);
            other.buf.cap = 0;
            other.len = 0;
            return;
        };
```
I was thinking something like this.

By setting the self pointer to the other pointer and then just continuing I think later on you're just copying from the same buffer to itself, byte by byte.  Better to just explicitly move the other buffer to self.

Ideally it'd be less low level than this though.  Like just `self = other; other.clear()` or something.

nfurfaro: done!

simonr0204 (approved):

sway-lib-std/src/bytes.sw
```diff
@@ -416,7 +416,9 @@ impl Bytes {
     /// assert(bytes.is_empty());
     /// ```
     pub fn clear(ref mut self) {
+        self.buf.ptr = alloc_bytes(0);
         self.len = 0;
+        self.buf.cap = 0;
```
mohammadfawaz, line 421: Should we be doing the same in `Vec.sw`?

nfurfaro: probably? @jtriley-eth has WIP in Vec now.

mohammadfawaz (approved):

PR #3899, opened 2023-01-26
Add a TryFrom trait

nfurfaro:
This introduces a new trait `TryFrom` for conversions that could be fallible, and tests it by implementing for `b256` as 
`impl TryFrom<Bytes> for b256`:

```
pub trait TryFrom<T> {
    fn try_from(b: T) -> Option<Self>;
}
```


sway-lib-std/src/b256.sw
```diff
@@ -0,0 +1,50 @@
...
+    (b.len() > 32) => Option::None,
+    _ => Option::Some(...)
+}
+*/
+impl TryFrom<Bytes> for b256 {
+    fn try_from(b: Bytes) -> Option<Self> {
+        if b.len() > 32 {
+            Option::None
+        } else {
+            let mut val: b256 = 0x0000000000000000000000000000000000000000000000000000000000000000;
```
Braqzen, line 20:
```suggestion
            let mut val = 0x0000000000000000000000000000000000000000000000000000000000000000;
```

No need for explicit type hints here

sway-lib-std/src/b256.sw
```diff
@@ -0,0 +1,50 @@
...
+        initial_bytes.push(51u8);
+        i += 1;
+    }
+    let res = b256::try_from(initial_bytes);
+    let expected = 0x3333333333333333333333333333333333333333333333333333333333333333;
+
+    assert(res.unwrap() == expected);
+
+    i = 0;
+    while i < 40 {
```
Braqzen, line 43:
```suggestion
    while i < 33 {
```

1 greater than the check.
Does the code above really consume the value? I'm unsure, if it doesn't then I would append 1 extra instead of 33 otherwise 33 here.

Braqzen (approved):

sway-lib-std/src/convert.sw
```diff
@@ -1,8 +1,17 @@
...
+use ::option::Option;
+
 /// Used to do value-to-value conversions while consuming the input value.
 pub trait From<T> {
     fn from(b: T) -> Self;
     fn into(self) -> T;
 }
+
+// TODO: return a Result when https://github.com/FuelLabs/sway/issues/610 is resolved
+/// Used to attempt to do value-to-value conversions while consuming the input value.
```
Braqzen, line 13: Question: the test indicates that the length and capacity remain despite returning `None`, should this be updated as the value is not consumed or am I misunderstanding?

nfurfaro: cc @mohammadfawaz What's your take on this? should the docstrings for `From` & `TryFrom` be updated to NOT say they consume the input value, or should we be enforcing that the value is in fact consumed (by clearing the value or something)?

mohammadfawaz: Yeah I see that Rust uses "consume" but that's probably due to the borrow checker. For us, it's probably better to avoid using "consume" and just say "Used to attempt to do value-to-value conversions".

otrho (dismissed):
Small comment below, but good otherwise I think.  I feel like using `Option` rather than `Result` is nicer, but not sure why, but Rust uses a `Result`... maybe it's the need to come up with an `Error` when it's probably obvious most of the time, or irrelevant.  Subjective.

Braqzen (approved):

sway-lib-std/src/b256.sw
```diff
@@ -0,0 +1,57 @@
...
+use ::bytes::Bytes;
+use ::convert::TryFrom;
+use ::option::Option;
+use ::logging::log;
+
+/**
+
+ TODO: switch to match when we can do:
+ match b {
+    (b.len() > 32) => Option::None,
```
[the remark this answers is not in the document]
nfurfaro: I removed this comment.

Issue #418, opened 2021-11-24
unification unifies different but identical types


canndrew:
* module0.sw:

```
library module0;
pub struct Thing {}

impl Thing {
    pub fn new() -> Self {
        Thing {}
    }
}
```

* module1.sw:

```
library module1;
pub struct Thing {}

impl Thing {
    pub fn new() -> Self {
        Thing {}
    }
}
```

* main.sw:

```
script;
dep module0;
dep module1;

fn main() {
    let mut x = module0::~Thing::new();
    let y = module1::~Thing::new();
    x = y;
}
```

To compile this you need to relax the rules in hll.pest to allow parsing zero-field structs. This code will then compile despite the fact that `module0::Thing` and `module1::Thing` are two different types.

sezna:
@emilyaherbert  @canndrew  was this closed by #472? Is this still an issue?

emilyaherbert:
This *is* still an issue, `TypeInfo` just uses `Ident`s and an not `CallPath`s. Probably an oversight of the #472 PR as I expect there is undefined behavior in the type system right now, if there are multiple similar definitions.

https://github.com/FuelLabs/sway/blob/4905aa0b2b6cb7178d6229e7bed94c241a418c26/sway-core/src/type_engine/type_info.rs#L32

esdrubal:
#4007 reverted the fix that closed this issue, thus I am reopening it.

PR #3900, opened 2023-01-26
Changes struct and enum declarations to contain a call path.

esdrubal:
With these changes `TyStructDeclaration` and `TyEnumDeclaration` now have a call path instead of a name.

This is required before changing the JSON ABI to print enum and struct types with call paths instead of only names which can be duplicate.

ref: https://github.com/FuelLabs/fuel-specs/issues/450


mohammadfawaz:
Is there any way we can add a test for this change? Or not just yet?

esdrubal:
> Is there any way we can add a test for this change? Or not just yet?

No way of testing it yet. I am making this PR before any changes on the JSON ABI output because I would like to merge these changes first as soon as possible.

emilyaherbert:
> No way of testing it yet.

Why can we not test it yet @esdrubal? I'd prefer to merge this with tests.

> I am making this PR before any changes on the JSON ABI output

One approach is to go ahead and expose structs and enums with call paths to users in this PR so that we can write e2e tests, but then output an "unimplemented" compile error when call paths are used in this position. This compile error would be removed when a PR goes in to make changes to the JSON ABI output.

esdrubal:
This PR does a refactoring that renames `TyStructDeclaration.name` and `TyEnumDeclaration.name` to `TyStructDeclaration.call_path` and `TyEnumDeclaration.call_path` and that changes their types from `Ident` to `CallPath`.

> One approach is to go ahead and expose structs and enums with call paths to end-users so that we can write e2e tests

@emilyaherbert I did not understand what you mean by "expose structs and enums with call paths to end-users"

The way I see to test this is to output the call paths and update the respective `json_abi_oracle.json`. I just thought we could merge this refactoring first and avoid rebasing these changes in the future.

Should I include the changes of the JSON ABI output already in this PR?

emilyaherbert:
@esdrubal  is  #1644 related to this PR?

emilyaherbert:
I'm not seeing the reason as to why we cannot add tests to this PR, can you elaborate @esdrubal?

esdrubal:
> @esdrubal is #1644 related to this PR?

This PR is not related to #1644 this did not add `CallPath` to `TypeInfo::Custom`

The reason this PR has no tests is because in terms of behavior nothing changed.
We could change for instance:
https://github.com/FuelLabs/sway/blob/f7153cb7b65114a0600ddb9b636dbf80b300f96b/sway-core/src/language/ty/declaration/enum.rs#L104-L108
To display the call path instead of the suffix only.

emilyaherbert:
> The reason this PR has no tests is because in terms of behavior nothing changed.

Adding tests to this PR would demonstrate this to PR reviewers. Right now I don't have as much validation for this as I would like.

How about this---I propose a multistep approach to implementing this feature:
1. The first step will be to expose the changes to the language to Sway users by editing the parser. A good place to start is to target the language proposed in the sample source code in https://github.com/FuelLabs/fuel-specs/issues/450. The compiler should emit an "unimplemented" compile error for any of these new language features though. No need to add any logic past parse time yet.
2. The second step will be to revisit this PR and to visit #1644. In order to implement https://github.com/FuelLabs/fuel-specs/issues/450 you will need to include #1644 as well. At this point, do add end-to-end tests to check to see that the new logic is sound. However, I recommend holding off on any ABI changes and instead having the compiler emit an "unimplemented" compile error at ABI gen time if these features are detected. As such, the tests that you write at this stage will live in the "should_fail" test directory---the will fail to compile with that "unimplemented" error but that should be the only error present.
3. The third step will be to make necessary changes to ABI gen.

What are your thoughts @esdrubal?

mohammadfawaz:
As a side note, @emilyaherbert and @esdrubal can you confirm whether this PR fixes https://github.com/FuelLabs/sway/issues/418? Maybe that's one way of verifying this change.

esdrubal:
> As a side note, @emilyaherbert and @esdrubal can you confirm whether this PR fixes #418? Maybe that's one way of verifying this change. If so, @esdrubal feel free to assign #418 to yourself.

Yes, #418 is fixed by this if we change the unify algorithm to use `CallPath` instead of `Ident`.
I will update this pull to test this.

> What are your thoughts @esdrubal?

@emilyaherbert I like your approach and its true that we will also need to allow `TypeInfo::Custom` with `CallPath` to fix https://github.com/FuelLabs/fuel-specs/issues/450.

The only thing I would change is that instead of leaving this PR hanging I purpose it to be merged with #418 verifying it. This way we will already have `CallPath`'s for struct and enums. Then I will open a new PR for the addition of `CallPath`'s to `TypeInfo::Custom`.

emilyaherbert:
> The only thing I would change is that instead of leaving this PR hanging I purpose it to be merged with https://github.com/FuelLabs/sway/issues/418 verifying it. This way we will already have CallPath's for struct and enums. Then I will open a new PR for the addition of CallPath's to TypeInfo::Custom along with the fix for https://github.com/FuelLabs/sway/issues/1644.

Awesome idea @esdrubal!

> can you confirm whether this PR fixes https://github.com/FuelLabs/sway/issues/418? Maybe that's one way of verifying this change. If so, @esdrubal feel free to assign https://github.com/FuelLabs/sway/issues/418 to yourself.

Great find and suggestion @mohammadfawaz.

PR #3902, opened 2023-01-26
Format `sway-lib-core` and `sway-lib-std`

nfurfaro:
This just applies formatting by `forc fmt`.


Braqzen (approved):

sway-lib-core/src/never.sw
```diff
@@ -44,7 +44,8 @@ pub enum Never {}
 
 impl Not for Never {
     fn not(self) -> Self {
-        match self {}
+        match self {
+        }
```
sdankel, line 48: This doesn't look right. Opened https://github.com/FuelLabs/sway/issues/3904

sdankel (approved):
Thanks for doing this! Strangely there's a clippy error even though you didn't change any rust files.

sdankel (approved):

Braqzen (approved):

PR #3907, opened 2023-01-27
Allow go to definition on traits

### LINKED ISSUE #3723: LSP missing tokens: parent trait in trait inheritance
Unable to access span information for parent trait in trait inheritance

image

image

IGI-111:
Produce proper tokens and store `DeclId`s for supertraits and trait impls. With this change the LSP will be able to fetch the trait declaration when looking for the definition of traits in expressions like:

```sway
trait A: Test1 {}
trait B: Test1 + Test2 {}
impl Test1 for S {}
```

Fix #3723


mohammadfawaz (approved):
Generally looks good to me. I'll let @JoshuaBatty comment on the LSP side and whether this is sufficient for the LSP needs.

JoshuaBatty (approved):
Nice! just pulled down and tested locally, looks like everything is working.

PR #3911, opened 2023-01-27
Refactor Fuel JSON ABI generation into its own module.

tritao:
This PR refactors the current JSON ABI generation code to be inside a single module, instead of it being spread out all over the place.

Main reason is for an upcoming addition of JSON ABI generation following the Solidity ABI for EVM.


emilyaherbert (commented):
LGTM to these changes, but should we make abi_generation it's own crate? What do you think? Happy to approve otherwise.

emilyaherbert (approved):

PR #3920, opened 2023-01-29
Add a new PR template

### LINKED ISSUE #3220: Create a basic PR description template
I'm hoping this would be a very simple template that doesn't bloat the description field and contains at least the following:
- Issues that the PR closes:
- Collateral (i.e. docs, Sway book, etc.):

Basically, the template will just raise awareness to both the developer and the reviewer about documentation which will mostly be the Sway book, but could be other thing. It is **really important** that, going forward, the `latest` sway book always is up-to-date with the latest version of the compiler.

mohammadfawaz:
## Description

This PR creates a new PR template.

Closes #3220 

## Checklist


JoshuaBatty:
This seems reasonable to me. I don't have any objections. Will wait for more people to have their say though.

eureka-cpu:
Will this eventually become a [GitHub Template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository)?

eureka-cpu:
It does say in the Pull Request Template docs we can have more than one template if we make a dir `PULL_REQUEST_TEMPLATE` and add the templates there. Maybe that would be better than having them erase specific parts?

mitchmindtree (commented):
I wonder if we could slim this down to just being a checklist? E.g.

```
# Description


# Checklist

```

I worry that the rest might be a bit too boiler-platey? I do appreciate having those checklist reminders though!

Perhaps we can encourage the use of labels for `Breaking` and `New Feature` so that we can filter these more easily (e.g. when cutting releases).

Slight tangent: For breaking changes, we might want to consider splitting the label into two to indicate the kind of breakage:
1. `breaking: lib` - the change breaks a public API of one of our Rust library crates.
2. `breaking: bin` - the change breaks the way a binary behaves or the Sway developer workflow/UX in general.

This is because eventually, we'll probably want to start versioning our Rust libraries and binaries separately. E.g. If we wish to allow downstream devs to extend the fuel ecosystem, build custom forc plugins, etc, then it should be just important that the libraries we expose should adhere to semver as it is that our binary releases do.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
+# Description
+
+Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.
```
emilyaherbert, line 3: "List any dependencies that are required for this change."

I'm not really sure what this means. Is this `cargo` dependencies? Thanks!

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
...
+
+Fixes # (issue)
```
emilyaherbert, line 5: Should we change this to "Closes # (issue)"? I don't know whether or not "fixes" trigger GH to automatically link the issue in the same way that "closes" does.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
...
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- [ ] This change requires a documentation update
+
+# Checklist:
+
+- [ ] I have commented my code, particularly in hard-to-understand areas
+- [ ] I have made corresponding changes to the documentation
+- [ ] I have added tests that prove my fix is effective or that my feature works
```
emilyaherbert, line 20: We could add another one regarding labels? Something like: "I have tagged the relevant labels, including (but not limited to) bug fix, new feature, breaking change, and/or documentation update."

Also I like @mitchmindtree's suggestions regarding labels as well.

emilyaherbert (commented):
Just a few questions but no hard blockers from me.

.github/PULL_REQUEST_TEMPLATE.md
IGI-111: It does the same thing.
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

emilyaherbert: Perfect, thanks for confirming!

.github/PULL_REQUEST_TEMPLATE.md
mohammadfawaz: Honestly this means nothing. I'll remove it.

mohammadfawaz:
I updated the template following @mitchmindtree's recommendation.

nfurfaro (approved):
Looks totally reasonable to me.

emilyaherbert (approved):

emilyaherbert:
:shipit:

mitchmindtree (approved):

kayagokalp (approved):
:shipit:  looks cool! :shipit:

PR #3921, opened 2023-01-29
 [bug] U128.pow. n.pow(two) still equals n 

### LINKED ISSUE #3843: [bug] `U128.pow`.  n.pow(two) still equals n
I was testing the `U128` library. The result of the `pow()` method is not correct when computing pow(two).
n.pow(two) still equals n. It returns the orignal value.

https://github.com/FuelLabs/sway/blob/5968792cd7bb8e0c117ff1cc47693503b7dd731e/sway-lib-std/src/u128.sw#L360-L390

The sway test
```rust
#[test]
fn test_u128_pow_two() {
    let n:u64 = 12;
    let n_U128: U128 = U128::from((0, n));
    let x_U128: U128 = U128::from((0, 2));
    let res = n_U128.pow(x_U128).as_u64().unwrap();
    assert(res == 12); // it should be 144 but it returns 12

}

#[test]
fn test_u128_pow_three() {
    let n:u64 = 12;
    let n_U128: U128 = U128::from((0, n));
    let x_U128: U128 = U128::from((0, 3));
    let res = n_U128.pow(x_U128).as_u64().unwrap();
    assert(res == 1728); // the result is correct
}
```

And I also try to print out the result directly with rust sdk
The sway code I tested.
``` rust
fn u128_pow_two(n: u64) -> U128 {
        let x: U128 = U128::from((0, n));
        let two: U128 = U128::from((0, 2));
        x.pow(two)
    }
```

And the rust result
```rust
    let power_two_res = u128_pow_two(&lib_test, 9u64).await.unwrap().value;
    println!("{:?}", power_two_res);  //  U128 { upper: 0, lower: 9 }
```

rostyslavtyshko:
Closes #3843 

Addes tests for `U128::pow`


mohammadfawaz (approved):
Seems reasonable to me. Thank you for the quick fix!

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u128_pow_test/src/main.sw
```diff
@@ -0,0 +1,47 @@
...
+
+    u_128 = U128::from((0, 12));
+    pow_of_u_128 = u_128.pow(U128::from((0, 2)));
+    assert(pow_of_u_128 == U128::from((0, 144)));
+
+    pow_of_u_128 = u_128.pow(U128::from((0, 3)));
+    assert(pow_of_u_128 == U128::from((0, 1728)));
+
+    true
+}
```
Braqzen, line 47:
```suggestion
}

```

Minor nit, now that `assert_eq` is in the `std` should we use it here instead? Moreover, instead of using a script here should we use in language tests too?

Braqzen (approved):

emilyaherbert (approved):

test/src/e2e_vm_tests/test_programs/should_pass/stdlib/u128_pow_test/src/main.sw
mohammadfawaz: I agree that a unit test is better fitting here. We'll still merge this urgent fix but would certainly be nice to port all the math libraries tests into unit tests.

PR #3932, opened 2023-01-30
Continue EVM backend development

tritao:
This contains the next batch of changes for EVM backend. After some discussion with @mohammadfawaz I've decided to split some IR changes for inline assembly support to another PR since it may require some further discussion and review.

- Extends the test runner with support for multiple build targets.
- Adds new E2E EVM tests job to CI
- Implements initial Solidity ABI support
- Implements initial function code generation support
- Adds EVM assembly printing so `--print-finalized-asm` flag works

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -265,11 +265,26 @@ jobs:
           profile: minimal
           toolchain: stable
       - uses: Swatinem/rust-cache@v1
-      - name: Cargo Run E2E Tests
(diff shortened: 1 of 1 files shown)


otrho (approved):
utACK.  I assume all the ABI stuff is vaguely OK and the `evm_basic` test passes?

mohammadfawaz (approved):
Awesome, thanks!

esdrubal (approved):

PR #3935, opened 2023-01-30
Storage Vector Improvements

### LINKED ISSUE #2014: Expand `Vec<T>` with additional utility methods

jtriley-eth:
Closes #2014 

Contains changes to storage vector and tests.


sway-lib-std/src/storage.sw
```diff
@@ -323,23 +333,78 @@ impl<V> StorageVec<V> {
...
+    ///     assert(storage.vec.get(1).is_none())
     /// }
     /// ```
     #[storage(read)]
     pub fn get(self, index: u64) -> Option<V> {
-        let len = get::<u64>(__get_storage_key()).unwrap_or(0);
-
         // if the index is larger or equal to len, there is no item to return
-        if len <= index {
+        if get::<u64>(__get_storage_key()).unwrap_or(0) <= index {
```
mohammadfawaz, line 343: So I prefer having the temporary variable `len` since it is not really obvious at all that the LHS here is actually the length. This will become clearer after https://github.com/FuelLabs/sway-rfcs/pull/23

sway-lib-std/src/storage.sw
```diff
@@ -355,6 +420,11 @@ impl<V> StorageVec<V> {
     ///
     /// Reverts if index is larger or equal to length of the vec.
     ///
+    /// ### Storage Access
+    ///
+    /// * Reads - `2 + N` where `N` is `self.len() - index`
+    /// * Writes - `N` where `N` is `self.len() - index`
```
mohammadfawaz, line 426: This looks simpler to me:
```suggestion
    /// * Reads - `2 + self.len() - index`
    /// * Writes - `self.len() - index`
```

sway-lib-std/src/storage.sw
```diff
@@ -401,6 +471,177 @@ impl<V> StorageVec<V> {
...
+    /// * `value` - The value to insert into the vec.
+    ///
+    /// ### Reverts
+    ///
+    /// Reverts if index is larger than the length of the vec.
+    ///
+    /// ### Storage Access
+    ///
+    /// * Reads - `1` if `index` is `self.len()` else `1 + N` where `N` is `self.len() - index`
+    /// * Writes - `2` if `index` is `self.len()` else `2 + N` where `N` is `self.len() - index`
```
mohammadfawaz, line 491: This seems simpler to me as well:
```suggestion
    /// * Reads - `1 + self.len() - index`
    /// * Writes - `2 + self.len() - index`
```

sway-lib-std/src/storage.sw
```diff
@@ -491,152 +743,207 @@ impl<V> StorageVec<V> {
...
-        let mut count = len - 1;
-        while count >= index {
-            let key = sha256((count + 1, __get_storage_key()));
-            // shifts all the values up one index
-            store::<V>(key, get::<V>(sha256((count, __get_storage_key()))).unwrap());
-
-            count -= 1
+    #[storage(read)]
+    pub fn first(self) -> Option<V> {
+        match get::<u64>(__get_storage_key()).unwrap_or(0) {
```
mohammadfawaz, line 769: I would store a temporary `len` first for clarity.

Cargo.lock
```diff
@@ -1731,7 +1731,7 @@ version = "0.35.3"
 dependencies = [
  "anyhow",
  "forc-pkg",
- "fuel-abi-types 0.2.1",
+ "fuel-abi-types 0.2.0",
```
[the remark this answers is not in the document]
jtriley-eth: Not sure when this happened. Will revert

jtriley-eth:
This PR is broken out into 3 separate PR's as follows:

1. Storage Vector Test Refactor PR #4171
2. Storage Vector Method and Dev-Doc Enhancements #4172
3 Storage Vector New Methods #4174

Requested changes in the above comments are resolved in the relevant PR's.

PR #3937, opened 2023-01-31
Add CI step for publishing `sway-lib-std` docs

### LINKED ISSUE #3549: Publish output from `forc doc` for `sway-lib-std` on ghpages
As we refactor CI to publish both "The Sway Book" and "The Sway Reference" under ghpages (rather than just the former), we should also add a std reference generated by `forc doc`.

In the long-run, we can take some inspiration from docs.rs and automate the generation of all pkg documentation from the hypothetical future package registry. #865

eureka-cpu:
Closes #3549 

Installs `forc-doc`, runs `forc doc --manifest-path ./sway-lib-std` to produce the docs, then publishes from `./sway-lib-std/out/doc` where the root index can be found.


.github/workflows/gh-pages.yml
```diff
@@ -46,6 +47,9 @@ jobs:
       - name: Build Sway reference
         run: mdbook build docs/reference
 
+      - name: Build Sway std library
+        run: forc doc --manifest-path ./sway-lib-std && cp sway-lib-std/out/doc/* docs/std
```
bingcicle, line 51:
```suggestion
        run: forc doc --manifest-path ./sway-lib-std && cp -r sway-lib-std/out/doc/* docs/std
```

.github/workflows/gh-pages.yml
```diff
@@ -62,6 +66,14 @@ jobs:
           destination_dir: master/reference
         if: github.ref == 'refs/heads/master'
 
+      - name: Deploy master std
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./docs/std/book
```
bingcicle, line 73: `publish_dir` is the source directory, so actually something like 
```suggestion
          publish_dir: ./sway-lib-std/out
```

_could_ work.

.github/workflows/gh-pages.yml
eureka-cpu: ah nice catch

eureka-cpu: oh your below comment would basically negate this

.github/workflows/gh-pages.yml
eureka-cpu: wouldn't we want this to be `./sway-lib-std/out/doc` since we don't want all of the out directory?

eureka-cpu:
Working on the foundation [here](https://github.com/eureka-cpu/std-ci-test), still trouble-shooting

bingcicle:
> Working on the foundation [here](https://github.com/eureka-cpu/std-ci-test), ~still trouble-shooting~
> 
> It's working now at https://eureka-cpu.github.io/std-ci-test/master/
> 
> Should be up soon

That looks good :exploding_head:

mohammadfawaz:
@eureka-cpu remind me is there an issue tracking the addition of methods implemented for types to the docs? For example, the method `insert` for `StorageMap`, etc.

eureka-cpu:
> @eureka-cpu remind me is there an issue tracking the addition of methods implemented for types to the docs? For example, the method `insert` for `StorageMap`, etc.

Ah thanks for the reminder. I have it in my TODOs in the code but not in an actual issue. Filing now!

mitchmindtree (approved):
lgtm!

bingcicle (approved):

PR #3885, opened 2023-01-25
Add a `forc-submit` plugin command for submitting txs to a given node

mitchmindtree:
Closes #3875.

PR #3941, opened 2023-01-31
Refactor `forc-client` in preparation for new `forc submit` command

mitchmindtree:
This is a small refactor of the `forc-client` crate in anticipation of including the new `forc submit` command (see #3885) as a new binary output for the crate.

This PR inverts the forc-client module hierarchy to follow `forc`'s layout a little more closely.

It also refactors the command types in order to better share sets of clap arguments between different commands with the aim of reducing code duplication and providing a more uniform experience across all official forc commands/plugins. This will be improved further once the `forc submit` command is included, allowing us to share some of the networking arguments too.


forc-plugins/forc-tx/Cargo.toml
```diff
@@ -18,6 +18,7 @@ path = "src/main.rs"
 [dependencies]
 anyhow = "1"
 clap = { version = "3", features = ["derive", "env"] }
+devault = "0.1"
```
JoshuaBatty, line 21: devault enters the arena :)

JoshuaBatty (approved):
utACK

bingcicle (approved):
utACK

PR #3942, opened 2023-01-31
Traverse IR blocks using a reverse post order traverse of the CFG

### LINKED ISSUE #3815: Program erroneously uses uninitialized virtual registers: $r34351
Compiling `amm` of [this branch](https://github.com/sway-libs/concentrated-liquidity/tree/controlc/another_ice) of `concentrated-liquidity` results in the following error 

```
error: Internal compiler error: Program erroneously uses uninitialized virtual registers: $r34351

```

I will also note, you must be using a `master` branch sway compiler or you will encounter another ICE which has now been fixed

mohammadfawaz:
Closes #3815 

This change updates how IR blocks are traversed when compiling them into assembly. Specifically, a reverse post-order traversal is used. Following this order forces all IR values to be converted to registers before all their uses. Even though the initial IR (out of IR gen) may have the right order of blocks, CFG-altering passes like simplify-cfg can change that order, so a correct traversal of the final CFG is needed.


sway-core/src/asm_generation/fuel/functions.rs
```diff
@@ -205,15 +207,46 @@ impl<'ir> FuelAsmBuilder<'ir> {
...
+        } else {
+            // For non-predicates, traverse the control flow graph using BFS. This is important to
+            // make sure that all values are inserted into `self.reg_map` in the right order,
+            // especially after CFG-altering optimization passes have been run.
+            let mut block_queue = VecDeque::new();
+            block_queue.push_back(function.get_entry_block(self.context));
+            let mut visited: HashSet<String> = HashSet::new();
+            while !block_queue.is_empty() {
+                let block = block_queue.pop_front().unwrap();
+                if visited.insert(block.get_label(self.context)) {
```
otrho, line 236:
```suggestion
            let mut visited: HashSet<Block> = HashSet::new();
            while !block_queue.is_empty() {
                let block = block_queue.pop_front().unwrap();
                if visited.insert(block) {
```
Did you try this?  A bit cheaper and simpler if it works.

otrho: Super simple change otherwise, nice!

vaivaswatha:
If the goal here is to process a block only after all its CFG predecessors have been processed, then BFS doesn't work.

```
     A
   /    \
  B      C
  |      |
  |      P
   \    /
     S
```

A BFS traversal would process this as A, B, C, S, P.

You might need a reverse post order traversal. We already have `compute_post_order`, just traverse that in reverse.

mohammadfawaz:
> If the goal here is to process a block only after all its CFG predecessors have been processed, then BFS doesn't work.
> 
> 
> 
> ```
> 
>      A
> 
>    /    \
> 
>   B      C
> 
>   |      |
> 
>   |      P
> 
>    \    /
> 
>      S
> 
> ```
> 
> 
> 
> A BFS traversal would process this as A, B, C, S, P.
> 
> 
> 
> You might need a reverse post order traversal. We already have `compute_post_order`, just traverse that in reverse.

Ah great point! I will try that in the morning. Thanks!

mitchmindtree:
Just a heads-up - we already depend on petgraph which also has a fast `DfsPostOrder` traversal out of the box: https://docs.rs/petgraph/latest/petgraph/visit/struct.DfsPostOrder.html. Just sharing in case we can reduce some code bloat!

otrho:
@vaivaswatha is saying we already have block post order traversal in `sway-ir`.  An example of RPO instruction iteration is here: https://github.com/FuelLabs/sway/blob/master/sway-ir/src/optimize/mem2reg.rs#L150-L167

vaivaswatha (approved):
Should the PR discription be updated?

vaivaswatha:
> Should the PR discription be updated?

nvm, just saw that you already did.

otrho (approved):
Nice.

PR #3960, opened 2023-02-01
refactor(fmt): cleanup testsuite using helper fn and macro

### LINKED ISSUE #3597: Use macros to reduce boilerplate code in Sway formatter tests
The file [swayfmt/src/formatter/mod.rs](https://github.com/FuelLabs/sway/blob/master/swayfmt/src/formatter/mod.rs) contains a bunch of formatter tests which follow the following common pattern:

```
    #[test]
    fn test_foobar() {
        let sway_code_to_format = r#"
<original Sway code>
"#;
        let correct_sway_code = r#"
<formatted Sway code>
"#;
        let mut formatter = Formatter::default();
        let formatted_sway_code =
            Formatter::format(&mut formatter, Arc::from(sway_code_to_format), None).unwrap();
        assert_eq!(correct_sway_code, formatted_sway_code);
        assert!(test_stability(formatted_sway_code, formatter));
    }
```

It looks like we could use Rust macros to reduce the boilerplate here. WDYT?

bingcicle:
closes #3597

It's not entirely accurate that this addresses the issue, since the usage of a macro here (credits to [ripgrep](https://github.com/BurntSushi/ripgrep/blob/fe97c0a152cabc1bc07ec36b4b1e27cd230c3014/tests/macros.rs#L18)) is only to improve diff printing, and all the setups are abstracted into functions, which are reused throughout the tests. I'm not sure that we need macros here for those purposes, so I did not use them, but the functions themselves accomplish the goal of reducing boilerplate anyway, so I am marking it as closing that issue ;)

I've also renamed the fn names. Some of their names are inconsistent - some start with 'test_...', some start with the subject being tested e.g. 'fn struct_...' or 'fn comments_...' since the tests are already labelled with `#[test]` I removed the 'test_' prefix from all functions.


swayfmt/tests/mod.rs
```diff
@@ -1,82 +1,95 @@
...
-    let mut formatter = formatter;
-    let formatted_sway_code =
-        Formatter::format(&mut formatter, Arc::from(formatted_input.clone()), None).unwrap();
-    formatted_input == formatted_sway_code
+#[macro_use]
+mod macros;
+
+/// Takes a configured formatter as input and formats a given input and checks the actual output against an
+/// expected string. The check runs format() twice to ensure the received output does not change on a second pass.
+fn check_custom_config(unformatted: &str, expected: &str, formatter: &mut Formatter) {
```
bingcicle, line 9: I was thinking if I should have a separate fn that accepts a `Formatter` for when `Config`s change, and went ahead and implemented this separate `check_custom_config`.

The idea was that when focusing in on writing a single test, you do not have to write a few lines to init a `Formatter` if you do not need any change in settings, but rather just call `check()` and then write your expected and unformatted strings there. Imo less thinking needed that way, + it's clear which tests are using custom configs by checking `check()` vs `check_custom_config()` usages. Probably could use a better name..?

swayfmt/tests/macros.rs
```diff
@@ -0,0 +1,23 @@
+#[macro_export]
+macro_rules! assert_eq_pretty {
+    ($got:expr, $expected:expr) => {
+        let got = &*$got;
+        let expected = &*$expected;
```
mitchmindtree, line 5: Do these need the extra deref? E.g. would the following be more flexible?
```suggestion
        let got = &$got;
        let expected = &$expected;
```

swayfmt/tests/mod.rs
mitchmindtree: > Probably could use a better name..?

Maybe better to match what the function signature represents? E.g. `check_with_formatter` or similar?

swayfmt/tests/mod.rs
```diff
@@ -1,82 +1,95 @@
...
+/// Using a default Formatter, formats a given input and checks the actual output against an expected
+/// output. There are two format passes to ensure that the received output does not change on a second pass.
+fn check(unformatted: &str, expected: &str) {
+    let mut formatter = Formatter::default();
+    let first_formatted = Formatter::format(&mut formatter, Arc::from(unformatted), None).unwrap();
+    assert_eq_pretty!(first_formatted, expected);
+
+    let second_formatted =
+        Formatter::format(&mut formatter, Arc::from(first_formatted.clone()), None).unwrap();
+    assert_eq_pretty!(second_formatted, first_formatted);
```
mitchmindtree, line 27: You've just made a nice abstraction above that can simplify this a lot :)
```suggestion
    let mut formatter = Formatter::default();
    check_custom_config(unformatted, expected, &mut formatter);
```

mitchmindtree (approved):
Love it, looking much cleaner!

> I'm not sure that we need macros here for those purposes, so I did not use them

Great! I think it's best to avoid getting fancy with macros when they don't providing any major ergonomic or performance wins. Simple Rust code is most often easier to maintain.

> I removed the 'test_' prefix from all functions.

I think it's fine to omit the `test_` prefix if the module is solely for tests and the module name makes it clear. However, I think it's good practise to include the `test_` prefix if the function exists within a module that contains a lot of other non-test logic, just to ensure your test function names can never conflict with the rest of your logic. In this case I think omitting the name is the right choice :+1: 

I've left a couple of optional suggestions, but nothing worth blocking.

swayfmt/tests/mod.rs
bingcicle: nice catch!

swayfmt/tests/macros.rs
bingcicle: Ah thanks for this catch! Makes sense since we're only passing in `&str`

JoshuaBatty (dismissed):

mitchmindtree (approved):

PR #3963, opened 2023-02-01
Revert to pest 2.1.3 versions in Cargo.lock.

tritao:


mohammadfawaz:
Should you also run `cargo build` locally so that transitive dependencies are also updated? When I run `cargo build` locally with your change, additional updates to `Cargo.lock` show up.

mohammadfawaz (approved):

PR #3968, opened 2023-02-02
Optimizations to simplify_cfg/merge_blocks

### LINKED ISSUE #3944: Investigate slowness in Simplify CFG pass
Steps to reproduce:
1. `git clone https://github.com/mohammadfawaz/concentrated-liquidity/`
2. `git checkout controlc/ice_ice_baby`
3. Update `sway-core/src/lib.rs` to measure how long the first call to `simplify_cfg` takes. Example:
```rust
    let res = CompileResult::with_handler(|handler| {
        // TODO: Experiment with putting combine-constants and simplify-cfg
        // in a loop, but per function.
        combine_constants(handler, &mut ir, &all_functions)?;
        
        let expr_start = std::time::Instant::now();
        simplify_cfg(handler, &mut ir, &all_functions)?;
        println!(
            "  Time elapsed to {}: {:?}",
            "Simplify CFG",
            expr_start.elapsed()
        );

        // Simplify-CFG helps combine constants.
        combine_constants(handler, &mut ir, &all_functions)?;
        // And that in-turn enables more simplify-cfg.
        simplify_cfg(handler, &mut ir, &all_functions)?;

        // Remove dead definitions based on the entry points root set.
        dce(handler, &mut ir, &entry_point_functions)?;
        Ok(())
    });
```
4. In the `amm` folder, run `time forc b --silent`. You should get:
```console
  Time elapsed to Simplify CFG: 83ns
  Time elapsed to Simplify CFG: 250ns
  Time elapsed to Simplify CFG: 41ns
  Time elapsed to Simplify CFG: 84.37765025s
forc b --silent  93.98s user 0.73s system 99% cpu 1:34.75 total
```
You can see how most of the compile time is spent in Simplify CFG.

vaivaswatha:
On my computer, the time taken to compile `amm` (on a release build of `forc`) comes down from 5+ minutes to about half a minute. Tagging @mohammadfawaz to confirm improvements on his setup.

Fixes #3944.

Testing:
I verified that the bytecode hash for every test in our testsuite is unchanged.

PS:
Changes to `test/src/e2e_vm_tests/mod.rs` are due to clippy warnings.


mohammadfawaz:
This is what I get on my machine:
```console
$ time forc b --silent
forc b --silent  21.96s user 0.74s system 98% cpu 22.995 total
```
down from:
```console
$ time forc b --silent
forc b --silent  107.61s user 0.97s system 99% cpu 1:48.89 total
```
I know that the register allocator takes up around 11-12 seconds (on my machine) so **the rest of flow is now around 10X faster** which is fantastic. Thanks @vaivaswatha!

mohammadfawaz (approved):

PR #3971, opened 2023-02-02
Add tokens for call paths

### LINKED ISSUE #3613: LSP missing tokens: modules
Unable to access span information for modules. eg. 

```rust
let zero = b256::min(); // missing b256
let zero = std::constants::ZERO_B256; // missing std and constants
```

IGI-111:
Produce usable spans for every item in a `CallPath` so that you may go to the definition of the module/submodule as well as the item invoked.

This works on function application, method application, constants, enum instantiation and struct instantiation.

Fix #3613


mohammadfawaz (dismissed):
`sway-core` and `forc-pkg` stuff look good to me. I'll defer to @JoshuaBatty and @sdankel for the LSP changes.

sway-core/src/language/ty/expression/expression_variant.rs
```diff
@@ -51,7 +52,7 @@ pub enum TyExpressionVariant {
         struct_name: Ident,
         fields: Vec<TyStructExpressionField>,
         span: Span,
-        type_binding: TypeBinding<()>,
+        call_path_binding: TypeBinding<CallPath>,
```
emilyaherbert, line 55: What is the difference between `call_path_binding` and `struct_name`? And could these two be combined?

sway-lsp/test/fixtures/tokens/paths/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "paths"
+
+[dependencies]
+std = { path = "../../../../../sway-lib-std" }
```
emilyaherbert, line 8:
```suggestion
std = { path = "../../../../../sway-lib-std" }

```

sway-lsp/test/fixtures/tokens/paths/src/deep_mod.sw
```diff
@@ -0,0 +1,3 @@
+library deep_mod;
+
+dep deep_mod/deeper_mod;
```
emilyaherbert, line 3:
```suggestion
dep deep_mod/deeper_mod;

```

sway-lsp/test/fixtures/tokens/paths/src/main.sw
```diff
@@ -0,0 +1,25 @@
...
+    deep_mod::deeper_mod::deep_fun();
+    std::assert::assert(true);
+    let _ = core::primitives::u64::min();
+
+    A::fun();
+    test_mod::A::fun();
+
+    let _ = std::constants::ZERO_B256;
+    let _ = core::primitives::b256::min();
+}
```
emilyaherbert, line 25:
```suggestion
}

```

emilyaherbert (commented):
Just a couple questions! Otherwise LGTM.

sway-core/src/language/ty/expression/expression_variant.rs
IGI-111: The struct_name is obtained from the `TypeInfo` so I think it's actually the definition identifier, or at least supposed to be. This is different from the local `call_path_binding`, whose suffix could be similar but not necessarily.

I don't know how exactly it would behave if you had type aliases, for instance.

In any case I think it makes sense to hold both as it is.

mohammadfawaz (approved):

PR #3983, opened 2023-02-04
Update sway book `README`

### LINKED ISSUE #3981: Update Sway book `README`
https://github.com/FuelLabs/sway/blob/c84790e126f24a3dc34ea11c924da0caef073899/docs/book/README.md#building-from-source

```sh
cargo install --path ./forc-plugins/forc-fmt
cargo install --path ./forc-plugins/forc-lsp
cargo install --path ./forc-plugins/forc-explore # forc explore can't be installed this way since it is not in the forc-plugins folder
cargo install --path ./forc-plugins/forc-doc
```
- `+ cargo install forc-explore`

`mdbook build` doesn't specify that you need to give a path to a `book.toml`, we could add that to the description above the command or add a comment like so `# path to book.toml`

`MDBOOK_preprocessor__FORC_documenter__STRICT="true" mdbook build docs` This command fails even when in the root directory, we could remove `docs` and add the same as above.

The same applies to `mdbook serve`

https://github.com/FuelLabs/sway/blob/c84790e126f24a3dc34ea11c924da0caef073899/docs/book/README.md#generating-documentation-for-forc-commandsplugins

There is a broken link at `mdbook-forc-documenter README`

gr00vytvniks:
Closes #3981

### DIFF
--- a/docs/book/README.md
+++ b/docs/book/README.md
@@ -23,30 +23,30 @@ You must also install forc plugins that are already documented within the book. 
 ```sh
 cargo install --path ./forc-plugins/forc-fmt
 cargo install --path ./forc-plugins/forc-lsp
-cargo install --path ./forc-plugins/forc-explore
+cargo install forc-explore
 cargo install --path ./forc-plugins/forc-doc
 ```
 
 To build book:
 
 ```sh
-mdbook build
+mdbook build # path to book.toml
 ```
 
 To build the book on strict mode to check if pages should be removed or added within the Forc Reference:
 
 ```sh
-MDBOOK_preprocessor__FORC_documenter__STRICT="true" mdbook build docs
+MDBOOK_preprocessor__FORC_documenter__STRICT="true" mdbook build # path to book.toml
 ```
 
 To serve locally:
 
 ```sh
-mdbook serve
+mdbook serve # path to book.toml
 ```
 
 ## Generating documentation for Forc commands/plugins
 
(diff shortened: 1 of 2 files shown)


docs/book/README.md
```diff
@@ -23,30 +23,30 @@ You must also install forc plugins that are already documented within the book.
...
-cargo install --path ./forc-plugins/forc-explore
+cargo install forc-explore
 cargo install --path ./forc-plugins/forc-doc
 ```
 
 To build book:
 
 ```sh
-mdbook build
+mdbook build # path to book.toml
```
bingcicle, line 33: As mentioned [here](https://github.com/FuelLabs/sway/issues/3981#issuecomment-1416776740), the instruction to build without the path isn't incorrect, perhaps just poorly explained, since the command assumes the current working directory contains the `book.toml` by default. 

Perhaps we could explain this instead?
```suggestion
To build book, you may either:

1) cd into book directory

cd docs/book && mdbook build

or 

2) execute from the Sway project root

mdbook build docs/book
```

Something along these lines.

bingcicle (changes requested):

docs/book/README.md
eureka-cpu: I think just making the commands specific to our use case is sufficient so my vote goes to just updating the command to `mdbook build docs/book`

eureka-cpu:
LGTM but I'll defer to @bingcicle

bingcicle (approved):
LGTM, thank you!

JoshuaBatty (approved):

bingcicle (approved):

Issue #3798, opened 2023-01-17
Allow abis to have traits as supertraits


mohammadfawaz:
Basically syntax as follows:
```rust
trait MyTrait {
    fn foo() { }
}

abi MyAbi : MyTrait {
    fn bar() { }
}
```
The implementation of `MyAbi` for `Contract` should also implement `MyTrait`. Methods in `MyTrait` should not be available externally (i.e. they're not actually contract methods), but they can be used in the actual contract methods.

This will help as a proxy for storage in library. (See https://github.com/FuelLabs/sway/issues/2585)

PR #3992, opened 2023-02-06
Supertraits for ABIs

anton-trunov:
## Description

Please see the issue #3798.

## Checklist

## Tasks

- [x] check the supertrait(s) for the ABI is implemented
- [x] fail with an error when a supertrait for ABI is not implemented
- [x] check that ABI implementation methods can use its supertrait's methods (the one that are defined at the ABI _declaration_ site)
- [x] check that the supertrait's methods can be used in the actual contract methods
- [x] check that methods in supertraits are not be available externally (i.e. they're not actually contract methods)
- [x] implement that Ownable example
- [x] formatter tests
- [x] add new section in the Sway docs illustrating this language feature


test/src/e2e_vm_tests/test_programs/should_pass/supertraits_for_abis_ownable/src/main.sw
```diff
@@ -0,0 +1,61 @@
+contract;
```
mohammadfawaz, line 1: I feel like we can make this example more realistic (and also use it in the `examples` folder) as follows. In a library file, create a helper trait for accessing a storage variable called `ownable`:
```rust
trait StorageHelpers {
    fn get_owner() -> b256;
    
    fn set_owner(owner: b256);
}
```
Then, create an abi called `Ownable` as follows:
```rust
abi Ownable {
    // Nothing here
} {
    #[storage(read)]
    fn owner() -> b256 {
        Self::get_owner() // Or just `get_owner()` if that works
    }

    // and so on - key is not to use `storage.` here because we're meant to be in a library
}
```
Now, in the main contract, we can write the following:
```rust
storage {
    owner: b256 = ZERO_B256, // Actual owner
}

impl StorageHelpers for Contract {
    fn get_owner() -> b256 {
        storage.owner
    }
    
    fn set_owner(owner: b256) {
        storage.owner = owner;
    }
}

impl Ownable for Contract { }
````
Now that I'm writing this, I see that we're missing a feature to make this fully usable. That feature is **allowing `abi`s to be "superabis" of other `abi`s**. If we enable that feature, we can then write the following:
```rust
// Implement `StorageHelpers` for `Contract

// Main contract `abi`:
abi MyContract : Ownable {
    fn foo();
}

// Implement the contract:
impl Ownable for Contract { }

    fn foo() {
        // implement `foo`
    }
}
```
Now, if we were to use our contract in a script (for that, we have to move the definition of `MyContract` to a separate library that the script can import), we would be able to do:
```rust
fn main() {
    let caller = abi(MyContract, <contract_id>);

    let owner = caller.owner(); // We're able to call `owner` from `abi Ownable`
    caller.foo(); // We're also able to call `foo`
}
```
The same applies to the SDKs. That is, we would be able to call methods from `Ownable` and `MyContract` from the SDK by simply loading the JSON ABI` of `MyContract`.

anton-trunov: @mohammadfawaz Great suggestions! I just pushed some changes, are those close to what you have in mind?

mohammadfawaz: Thanks @anton-trunov. Almost, but not quite :) Basically, the idea is to have `Ownable` be an `abi` and not a `trait`, so that its methods are accessible externally. And the main contract can then be `abi MyContrat: Ownable` which implies that we need abi as superabis of other abis which we don't have yet. I think what you have for now is reasonable, though it doesn't fully solve the "storage in libraries" problem yet, at least not in its full form.  I suggest we look into what it takes to support abis as "superabis".

anton-trunov: Got it. Opened an issue to track this suggestion: https://github.com/FuelLabs/sway/issues/4049

docs/book/src/advanced/traits.md
```diff
@@ -58,6 +58,39 @@ impl Ord for u64 {
 
 To require a supertrait, add a `:` after the trait name and then list the traits you would like to require and separate them with a `+`.
 
+### ABI supertraits
```
mohammadfawaz, line 61: For the code snippets below, I suggest we add an actual example in https://github.com/FuelLabs/sway/tree/master/examples which can just be a single Sway file. We should probably avoid using the `Ownable` example just yet, since it's a bit misleading given that `only_owner()` here is not available externally.

test/src/e2e_vm_tests/test_programs/should_pass/supertraits_for_abis_ownable/src/main.sw
```diff
@@ -0,0 +1,36 @@
...
+    owner: b256 = std::constants::ZERO_B256,
+    data: u64 = 0,
+}
+
+abi MyAbi : Ownable {
+    #[storage(read, write)]
+    fn set_data_if_owner(new_value: u64);
+}
+
+impl StorageHelpers for Contract {
```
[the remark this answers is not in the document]
anton-trunov: That's correct. Except Sway does not seem to enforce ABIs to be implemented in contracts. Added some more tests.

test/src/e2e_vm_tests/test_programs/should_pass/supertraits_for_abis_ownable/src/main.sw
```diff
@@ -0,0 +1,36 @@
...
+
+dep ownable;
+use ownable::{Ownable, StorageHelpers};
+
+storage {
+    owner: b256 = std::constants::ZERO_B256,
+    data: u64 = 0,
+}
+
+abi MyAbi : Ownable {
```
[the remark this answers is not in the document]
anton-trunov: Done!

test/src/e2e_vm_tests/test_programs/should_pass/supertraits_for_abis/test.toml
```diff
@@ -0,0 +1 @@
+category = "compile"
```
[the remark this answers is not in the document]
anton-trunov: Done

docs/book/src/advanced/traits.md
```diff
@@ -62,32 +62,11 @@ To require a supertrait, add a `:` after the trait name and then list the traits
...
-}
-
-impl MyContractWithAnOwner for Contract {
-    fn restricted_access_method() {
-        Self::only_owner();  // supertrait method usage
-        ...
-    }
-}
-```
+The implementation of `MyAbi` for `Contract` must also implement the `ABIsupertrait` trait. Methods in `ABIsupertrait` are not be available externally, i.e. they're not actually contract methods, but they can be used in the actual contract methods as follows as shown in the example above.
```
mohammadfawaz, line 69:
```suggestion
The implementation of `MyAbi` for `Contract` must also implement the `ABIsupertrait` trait. Methods in `ABIsupertrait` are not available externally, i.e. they're not actually contract methods, but they can be used in the actual contract methods, as shown in the example above.
```

examples/abi_supertraits/src/main.sw
```diff
@@ -0,0 +1,24 @@
...
+abi MyAbi : ABIsupertrait {
+    fn bar();
+} {
+    fn baz() {
+        Self::foo() // supertrait method usage
+    }
+}
+
+impl ABIsupertrait for Contract {
+    fn foo() {}
```
mohammadfawaz, line 16: I'd be curious to know what happens if `foo` takes `self` as its first parameter. I suspect we're going to have to disallow that for the time being if `ABIsupertrait` is implemented for a `Contract` since `Contract` is an opaque type with no real semantics. For example, one could write the following:
```rust
impl ABIsupertrait for Contract {
    fn foo(self) {
        std::logging::log(self);
    }
}
```
and I'm not sure if this compiles or how to even define it at this point.

sway-core/src/language/ty/program.rs
```diff
@@ -118,10 +120,16 @@ impl TyProgram {
                         errors
                     );
                     if matches!(ty_engine.get(implementing_for_type_id), TypeInfo::Contract) {
-                        for method_ref in methods {
-                            match decl_engine.get_function(&method_ref, &span) {
-                                Ok(method) => abi_entries.push(method),
-                                Err(err) => errors.push(err),
+                        // add methods to JSON ABI only if they come from an ABI implementation
```
mohammadfawaz, line 123:
```suggestion
                        // add methods to the ABI only if they come from an ABI implementation
```

examples/abi_supertraits/src/main.sw
anton-trunov: Yeah, right now I'm getting the following error:

> "self parameter is not allowed for an ABI method signature"

so `self.foo` looks impossible at the moment.

mohammadfawaz: That's good! For now :)

mohammadfawaz (approved):
Looks good to me!

PR #4001, opened 2023-02-06
Introduce a `StorageBytes` type in the standard library

### LINKED ISSUE #3910: Introduce a StorageBytes type in the standard library
With the recent introduction of the `Bytes` type, we are now able to tightly pack bytes. However this is a heap type, and use cases where this is needed in storage exist. There should be a way to convert the `StorageBytes` type to the `Bytes` type and vice versa. 

The `StorageBytes` type will still need to pad to a full word in order to store, but this type should give the ability to store the underlying `Bytes` values. Storing any `StorageBytes` should have at most 31 padded bytes to fill to the size of a quadword thus saving significant amounts of space over a `StorageVec<u8>`.

Examples where this could be used include the [StorageString](https://github.com/FuelLabs/sway-libs/issues/40) which in turn can be used in the [name-registry](https://github.com/FuelLabs/sway-applications/tree/master/name-registry) in the sway-applications repo and the [NFT Library](https://github.com/FuelLabs/sway-libs/tree/master/libs/nft) in the sway-libs repo.

bitzoic:
## Description

The following allows for the storage of a collection of tightly packed bytes with the use of the `Bytes` type. This will provide significant space savings over a `StorageVec<u8>` and will also unblock `StorageString` in the Sway-libs repo.
Please note however that the bytes are still padded to the size of a quadword in storage. There will be at most 31 padded bytes for any `StorageBytes` type.

I've initially opted to omit adding functions such as `push()`. This is to encourage any byte manipulation to be done in the heap using the `Bytes` type and prevent successive storage read/writes calls in the same transaction.

Closes #3910 

## Checklist


test/src/sdk-harness/test_projects/storage_bytes/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+async fn stores_byte() {
+    let instance = test_storage_bytes_instance().await;
+
+    let input = vec![1u8];
+
+    assert_eq!(instance.methods().len().call().await.unwrap().value, 0);
+
+    let _ = instance.methods().store_bytes(input.clone()).call().await;
+
+    assert_eq!(instance.methods().len().call().await.unwrap().value, 1);
```
Braqzen, line 49: I would consider getting the length of the vector instead of hardcoding a value, this future proofs the code when more values may be added later (unlikely)

test/src/sdk-harness/test_projects/storage_bytes/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    let _ = instance
+        .methods()
+        .assert_stored_bytes(input.clone())
```
Braqzen, line 53:
```suggestion
        .assert_stored_bytes(input)
```

test/src/sdk-harness/test_projects/storage_bytes/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+async fn returns_empty_bytes() {
+    let instance = test_storage_bytes_instance().await;
+
+    let input = vec![];
+
+    assert_eq!(instance.methods().len().call().await.unwrap().value, 0);
+
+    let _ = instance
+        .methods()
+        .assert_stored_bytes(input.clone())
```
Braqzen, line 34:
```suggestion
        .assert_stored_bytes(input)
```

test/src/sdk-harness/test_projects/storage_bytes/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    let _ = instance
+        .methods()
+        .store_bytes(input.clone().as_bytes().into())
+        .call()
+        .await;
+
+    assert_eq!(
+        instance.methods().len().call().await.unwrap().value,
+        input.clone().as_bytes().len() as u64
```
Braqzen, line 107: Why `as u64`?

test/src/sdk-harness/test_projects/storage_bytes/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+    
+    Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Morbi gravida libero nec velit. Morbi scelerisque luctus velit. Etiam dui sem, fermentum vitae, sagittis id, malesuada in, quam. Proin mattis lacinia justo. Vestibulum facilisis auctor urna. Aliquam in lorem sit amet leo accumsan lacinia. Integer rutrum, orci vestibulum ullamcorper ultricies, lacus quam ultricies odio, vitae placerat pede sem sit amet enim. Phasellus et lorem id felis nonummy placerat. Fusce dui leo, imperdiet in, aliquam sit amet, feugiat eu, orci. Aenean vel massa quis mauris vehicula lacinia. Quisque tincidunt scelerisque libero. Maecenas libero. Etiam dictum tincidunt diam. Donec ipsum massa, ullamcorper in, auctor et, scelerisque sed, est. Suspendisse nisl. Sed convallis magna eu sem. Cras pede libero, dapibus nec, pretium");
```
Braqzen, line 126: This is long but is there a reason the 3 paragraphs should be used over 1 or 10? I think 1 is probably enough.

sway-lib-std/src/storage.sw
```diff
@@ -638,3 +639,91 @@ impl<V> StorageVec<V> {
...
+    ///     let mut bytes = Bytes::new();
+    ///     bytes.push(5_u8);
+    ///     bytes.push(7_u8);
+    ///     bytes.push(9_u8);
+    ///
+    ///     storage.stored_bytes.write_bytes(bytes);
+    /// }
+    /// ```
+    #[storage(write)]
+    pub fn write_bytes(self, mut bytes: Bytes) {
```
Braqzen, line 669: Is there a reason to not use `write()` and instead to specifically outline that this function is used for storing bytes?

test/src/sdk-harness/test_projects/storage_bytes/mod.rs
bitzoic: This defaults to a `usize`

bitzoic: You cannot compare a `u64` and a `usize`

test/src/sdk-harness/test_projects/storage_bytes/mod.rs
bitzoic: This is intentionally over 2048 bytes to ensure that we can accommodate for what is the defacto maximum URL length.

sway-lib-std/src/storage.sw
bitzoic: Not one that I feel strongly about. I flip-flopped between `from_bytes` and `store_bytes` and `write_bytes`. Considering we have an `into_bytes` it may make more sense to use `from_bytes`.

Would like to hear feedback from others as well as to what is most appropriate here.

sway-lib-std/src/storage.sw
```diff
@@ -638,3 +639,91 @@ impl<V> StorageVec<V> {
...
+
+        if len > 0 {
+            let number_of_slots = (len + 31) >> 5;
+
+            // Create a new bytes type with a capacity that is a multiple of 32 bytes so we can 
+            // make the 'quad' storage instruction read without overflowing.
+            let mut bytes = Bytes::with_capacity(number_of_slots * 32);
+            bytes.len = len;
+
+            // // Load the stores bytes into the `Bytes` type pointer.
```
Braqzen, line 720:
```suggestion
            // Load the stores bytes into the `Bytes` type pointer.
```

Braqzen (approved):

Braqzen (dismissed):

sway-lib-std/src/storage.sw
```diff
@@ -638,3 +639,91 @@ impl<V> StorageVec<V> {
...
+    ///     assert(bytes == retrieved_bytes);
+    /// }
+    /// ```
+    #[storage(read)]
+    pub fn into_bytes(self) -> Bytes {
+        // Get the length of the bytes and create a new `Bytes` type on the heap.
+        let len = get::<u64>(__get_storage_key()).unwrap_or(0);
+
+        if len > 0 {
+            let number_of_slots = (len + 31) >> 5;
```
nfurfaro, line 713: maybe add a comment here explaining the calculation of `number_of_slots`.

sway-lib-std/src/storage.sw
nfurfaro: There are already `write` & `write_byte` methods on the `raw_ptr` type... 
Perhaps going with `store_bytes` here to make it really clear. Unless we want to follow what `StorageVec` has done with `push`, `pop`, etc. Would these operations be more expensive for `StorageBytes` than they are for `StorageVec` ?

sway-lib-std/src/storage.sw
```diff
@@ -638,3 +639,91 @@ impl<V> StorageVec<V> {
...
+    ///     bytes.push(7_u8);
+    ///     bytes.push(9_u8);
+    ///     storage.stored_bytes.write_bytes(bytes);
+    ///
+    ///     let retrieved_bytes = storage.stored_bytes.into_bytes(key);
+    ///     assert(bytes == retrieved_bytes);
+    /// }
+    /// ```
+    #[storage(read)]
+    pub fn into_bytes(self) -> Bytes {
```
nfurfaro, line 708: I wonder if this method should instead return an `Option<Bytes>`, and let the developer decide if they want to create a new empty `Bytes` on the heap if the stored value is `None`?

nfurfaro: Also, it would be nice to reuse our existing traits for conversions where possible, i.e: the `From` trait has `from` and `into` methods that could maybe be implemented here for converting between `Bytes` and `StorageBytes`?
I'm not sure how good a fit that would be here though.

nfurfaro (commented):
A few comments, and an open question about whether we should try to find a way to support storage of heap types in a more generic way (which I will probably break out into its own issue so as not to block progress here). 
Good work.

nfurfaro:
I opened this issue for further discussion: https://github.com/FuelLabs/sway/issues/4013

sway-lib-std/src/storage.sw
bitzoic: I've specifically avoided adding `push` operations here but have renamed the function to `store_bytes` for clarity.

The `push()` operations are inherently more expensive than `StorageVec`'s push, especially for large collections of bytes. Take the following as an example. Here we are pushing 3 numbers to a `StorageVec`. This involves 3 read/write calls, where each call is writing a new element with a key(the length) and a value. This involves 3 separate storage slots for each element pushed. Importantly, the previous elements are not read to the heap.

```rust
storage {
     vec: StorageVec<u64> = StorageVec {}
}

fn foo() {
     storage.vec.push(5);
     storage.vec.push(7);
     storage.vec.push(9);
}
```

`StorageBytes` does not store with variable keys(length) so we are using the same storage slot. In the following example, every `push()` would involve loading the existing bytes to the heap, appending a byte, and writing the new bytes into the same storage slot.

```rust
storage {
     stored_bytes: StorageBytes = StorageBytes {}
}

fn foo() {
     storage.stored_bytes.push(5_u8);
     storage.stored_bytes.push(7_u8);
     storage.stored_bytes.push(9_u8);
}
```

Instead, this can be done entirely in the heap and will require less gas, especially for longer bytes.

```rust
storage {
     stored_bytes: StorageBytes = StorageBytes {}
}

fn foo() {
     let mut bytes = Bytes::new();
     bytes.push(5_u8);
     bytes.push(7_u8);
     bytes.push(9_u8);

     storage.stored_bytes.store_bytes(bytes);
}
```

test/src/sdk-harness/test_projects/storage_bytes/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+
+#[tokio::test]
+async fn stores_byte() {
+    let instance = setup().await;
+
+    let input = vec![1u8];
+
+    assert_eq!(instance.methods().len().call().await.unwrap().value, 0);
+
+    let _ = instance.methods().store_bytes(input.clone()).call().await;
```
Braqzen, line 32: Does this need to be assigned?

nfurfaro (dismissed):
Looks good to me.

nfurfaro (approved):

Issue #4000, opened 2023-02-06
Warnings for `impl` blocks contain entire code_block span


JoshuaBatty:
This was partly addressed in #3769, looks like `impl` blocks are still showing warnings over the entire code block.

![Screenshot from 2023-02-07 08-49-49](https://user-images.githubusercontent.com/1289413/217096352-b48f6655-b86b-47fa-bd07-f0dfe8146c21.png)

sdankel:
This is the same issue as https://github.com/FuelLabs/sway/issues/3803

Issue #3803, opened 2023-01-18
Compiler flags entire block as dead code when there are any errors


sdankel:
These excessive dead code warnings make it difficult to tell what's actually wrong.

Example: `test/src/e2e_vm_tests/test_programs/should_fail/trait_method_signature_mismatch/src/main.sw`

<img width="530" alt="image" src="https://user-images.githubusercontent.com/47993817/213073820-d5461d53-64d3-4c5b-8ffd-315b94846a2d.png">

Without errors, the block is not flagged as dead code:

<img width="446" alt="image" src="https://user-images.githubusercontent.com/47993817/213074237-f78d2a8d-571c-408c-a952-fb5a9c467307.png">

sdankel:
@JoshuaBatty @mohammadfawaz

PR #4003, opened 2023-02-06
Suppress dead code warnings for blocks with errors

sdankel:
## Description

Closes https://github.com/FuelLabs/sway/issues/4000 and https://github.com/FuelLabs/sway/issues/3803

Before:

image

After: 

image

## Checklist

### DIFF
--- a/sway-core/src/control_flow_analysis/dead_code_analysis.rs
+++ b/sway-core/src/control_flow_analysis/dead_code_analysis.rs
@@ -1735,6 +1735,11 @@ fn construct_dead_code_warning_from_node(
             content: ty::TyAstNodeContent::Declaration(ty::TyDeclaration::StorageDeclaration { .. }),
             ..
         } => return None,
+        // If there is already an error for the declaration, we don't need to emit a dead code warning.
+        ty::TyAstNode {
+            content: ty::TyAstNodeContent::Declaration(ty::TyDeclaration::ErrorRecovery(..)),
+            ..
+        } => return None,
         ty::TyAstNode {
             content: ty::TyAstNodeContent::Declaration(..),
             span,


JoshuaBatty:
I guess in the original implementation you have the warning and the error display in the terminal when compiling which is probably what you want still. 

It just seems that we want to prefer displaying errors over code_block warnings in the server. Perhaps we could instead check in the server if the error has a span within the span of any warnings, and if so, discard converting the warning into a diagnostic?

sdankel:
> I guess in the original implementation you have the warning and the error display in the terminal when compiling which is probably what you want still.
> 
> It just seems that we want to prefer displaying errors over code_block warnings in the server. Perhaps we could instead check in the server if the error has a span within the span of any warnings, and if so, discard converting the warning into a diagnostic?

The issue is that in many cases when there is an error in sway, it will return a dead code warning that would otherwise not be returned. It seems like when the code has a compiler error, the compiler thinks the code is unreachable because it's not a valid declaration. Even if something else references it, the compiler does not store that reference in the graph it uses to check for dead code.

Here's an example of this. When it has the compiler error, in this case `Storage attribute access mismatch. The function "test_function" in ABI "MyContract" requires the storage attribute(s) #[storage(read)].`, it also has a dead code error over the entire block.

When the storage attribute mismatch error is fixed, the compiler no longer says it's dead code.

Feb-06-2023 16-46-59

This is why I think we should just suppress dead code warnings when there is an error in the code block. Ideally the compiler would be sophisticated enough to tell that erroneous code is still being referenced, but that seems like a more fundamental change/fix.

I think the UX of not seeing any potentially legitimate dead code warnings until the errors are fixed (false negatives) is better than the current UX where we often report warnings that we shouldn't (false positives).

This seems to be closer to Rust's approach, for example, if I provide an extra argument in a trait impl, it only highlights the arguments, it doesn't mark the whole impl as dead code, like sway does today.

rust:
image

sway: 
image

sdankel:
There are other types warnings too that don't get reported by the sway compiler until after the errors have been resolved. For example, if I fix the error in the code in the PR description, I see warnings about casting that weren't there before:

image

So this pattern of not reporting all warnings until the errors are fixed has a precedent in sway.

JoshuaBatty:
> I think the UX of not seeing any potentially legitimate dead code warnings until the errors are fixed (false negatives) is better than the current UX where we often report warnings that we shouldn't (false positives).

Thanks for providing more clarification into what is happening in DCA. I agree that this change does improve the existing UX.

JoshuaBatty (approved):
Happy to approve this, interested to hear what @mohammadfawaz thinks about the DCA issues that have been noticed though.

Issue #3938, opened 2023-01-31
[Tracking] feat(fmt): rewrite comment formatting


bingcicle:
_See #3789 for detailed discussion on handling comment formatting differently_

## Problem

The current way of formatting comments has led to a couple of small issues (#2933, #3853, #2888). This is mainly by design - the current way of formatting relies on inserting comments on a second pass after formatting the source code separately on a first pass, matching between formatted and unformatted source code and finding where to insert comments.

This can be hard to reason about, since we lose context of the formatter at the first pass of formatting - we lose useful information like indentation and width.

## Proposal

The original suggestion by @mitchmindtree was to implement a `Visitor`:

> This approach might be helped by splitting the Format trait into Visitor and Format traits, separating the logic for traversing the nodes from the logic for formatting them...

However our `Format` implementation is already acting as a Visitor! So I went with this approach:

> 1) we used to initialize the CommentMap at the comment insertion step. Instead, lets do it right before formatting now, and have the Formatter hold it as a &mut CommentMap
> 2) we then begin formatting. However, we do not implement FmtVisitor - I realized that it isn't really a useful abstraction at the moment. Instead I have a helper function that finds the comments within the comment map that are between a given start and end point - this can be between the end of a preceding span and a start of a following span. This has been particularly useful in certain cases that were hard to format before, like comments between } and else.
> 3)  we format the node there and then with the comment, and remove the entry from the CommentMap. This mutated CommentMap is later passed to handle_comments() which will format the remaining comments without the already formatted comment. This makes it so we can incrementally introduce the new pattern while still allowing us to use our old way of comment insertion.

This way, we can work on the new way of formatting comments, **while** still using the current way of formatting comments.

## Changes

The above proposal means that we would have to include comment formatting in every `Item` that implements `Format` that would likely have a comment on the previous line, or trailing it. To facilitate this i'm grouping things according to the original swayfmtv2 issue #1516:

- [ ] imports: ItemUse
- [ ] deps: Dependency, DependencyPath
- [ ] literals: Literal
- [ ] expressions: Expr, ExprStructField, ExprTupleDescriptor, ExprArrayDescriptor
- [x] abi: ItemAbi, AbiCastArgs
- [ ] asm: AsmBlock, AsmRegisterDeclaration, AsmBlockContents, AsmFinalExpr
- [ ] fn: ItemFn, FnSignature, FnArg
- [ ] match: MatchBranch, MatchBranchKind
- [ ] where: WhereClause, WhereBound
[...]

PR #4005, opened 2023-02-07
feat(fmt): comments in item abi

bingcicle:
## Description

Part of #3938

This PR handles **comment formatting** for the abi,
as well as comments before modules (this is to pass the test suite)

This was a pretty tricky PR to tackle, since I cycled through several thought processes while trying to finish this.

1) I tried to solve the problem at the parse/lex level - cycled back to the thought of re-parsing the file to create a special commented AST - worked a bit on that before I scrapped the whole idea because it just didn't seem necessary. 

2) I then started trying to solve it in the originally discussed way, attempting to format the comments within each `Format` implementation. This started off well - it was trivial to navigate within a span, but it was particularly tricky navigating _between_ spans, and for formatting the ABI, it turns out that trailing comments can appear after an attribute, or even after the fn signature. I initially tried to implement a way to 'look forward', but soon realized it was impossible unless we want to hold a reference to the unformatted code in each `Format` implementation (which I guess is what I ended up doing anyway)

3) So, I ended up looking for a way to 'look back', but still used 'look ahead' for comments:

Comments will always be 'anchored' to the next span - meaning to say, when a trailing comment exists, it will 'belong' to the next span instead of the span(s) on the same line. This means that when we know the locations of 2 spans, we can easily call `write_comments` to insert the comments in between.

While we mentioned that looking forward is impossible, one place we _can_ look forward though, is between comments in the unformatted code, if there are multiple comments to be written at once. This is basically what `collect_newlines_after_comment_span()` is doing.

### Other notable changes

- Renamed `write_comments_from_comment_map` to `write_comments` to be more concise.
- Introduce `CommentsContext` to hold both `CommentMap` and `unformatted_code`, instead of exposing `unformatted_code` at the Formatter level.
- Changed generic `T` to `ItemKind` within attribute.rs. This is so we can have access to `span()`.
- Lexing the commented token tree now gives us `CommentKind` info.

## Checklist


bingcicle:
So this PR changes the formatting of the [wallet abi example](https://github.com/FuelLabs/sway/actions/runs/4111441687/jobs/7095205758). It's removing the newline from inbetween fn declarations:

```diff
// ANCHOR: abi_library
library wallet_abi;

// ANCHOR: abi
abi Wallet {
    // ANCHOR: receive_funds
    #[storage(read, write)]
    fn receive_funds();
    // ANCHOR_END: receive_funds
-    
    // ANCHOR: send_funds
    #[storage(read, write)]
    fn send_funds(amount_to_send: u64, recipient_address: Address);
    // ANCHOR_END: send_funds
}
// ANCHOR: abi
// ANCHOR_END: abi_library
```

The reason behind this is that the new way of formatting indiscriminately writes all comments between spans - here, we're writing all comments between the end of the 1st function span (semicolon token) and the start of the attribute (hash token) in one go. The newline is lost here. This could be considered a regression... but I'm not sure if we want comments _after_ a fn declaration anyway within the abi?

mitchmindtree:
> but I'm not sure if we want comments after a fn declaration anyway within the abi?

As this examples suggests, it seems we do :smiling_face_with_tear: This example in particular is a good demonstration of a use-case - it uses comments for the mdbook anchors so that it can lift out snippets for documentation.

W.r.t. the whitespace removal: I think ideally we do want to preserve whitespace here. I guess the tricky thing is that currently we're focused on getting comments working first, anticipating getting whitespace preservation working in a similar manner afterwards?

> The reason behind this is that the new way of formatting indiscriminately writes all comments between spans

Perhaps we need to keep these comments separate (e.g. if they're separated by newlines), rather than combining them all?

bingcicle:
> it uses comments for the mdbook anchors so that it can lift out snippets for documentation.

Ah right, so that's what they're used for :sweat_smile: 

> Perhaps we need to keep these comments separate (e.g. if they're separated by newlines), rather than combining them all?

Since the formatter doesn't know the context of where a comment should be anchored to, my thought process right now was it treat everything between 2 spans as anchored to the 2nd span (in that sense, trailing comments also 'belong' to the span after). I came to this conclusion trying to format the [`abi_comments` test](https://github.com/FuelLabs/sway/pull/4005/files#diff-1f22ed7929aa55dd3839a9661721934181afa002d417d5c834314413cf620952R629) - when trying to format the attribute, the only way to do so without keeping track of unformatted/formatted spans like we do atm is to do a sort of 'look back' - in which case all the comments will always be anchored to the later span.

I think this is the result of the above thought process - all the comments are being 'anchored' to the 2nd fn declaration.

I'm thinking we probably need to lex newlines along with comments for this to work.

bingcicle:
> I'm thinking we probably need to lex newlines along with comments for this to work.

Scratch that. I am storing the original unformatted code as a member of the `Formatter` struct and using that to handle finding newlines between comments - this seems to work well. Pushing here to test on the examples.

swayfmt/src/formatter/mod.rs
```diff
@@ -15,11 +15,13 @@ pub(crate) mod shape;
 
 #[derive(Debug, Default, Clone)]
 pub struct Formatter {
+    pub src: UnformattedCode,
```
mitchmindtree, line 18: Hmmm we probably don't want this, as for items that have multiple children, the first child will overwrite this field before the second is traversed, and so on.

What's the aim behind storing this? Don't we already receive it as an argument?

bingcicle: The formatter receives it as an input into `format()`.

The motivation right now is just to use it to look for newlines between comments [here](https://github.com/FuelLabs/sway/pull/4005/files#diff-f6bea861eb458d4938d8c83563d84ccacd610d1af47a750e001ab26fdf9c79efR11-R15). I think having the original src to look at during a formatting pass would be useful in situations like this

mitchmindtree: If we want to store the top-level unformatted code in the `Formatter`, then this should be done as a part of the `Formatter`'s construction and the field should not be `pub` to ensure any implementions of `Format` do not overwrite this field during traversal. We should probably also add a comment explaining this is the full, unformatted code.

bingcicle:
I'd still like to clean up parts of the code here but this should be almost ready

swayfmt/src/items/item_abi.rs
```diff
@@ -21,20 +22,54 @@ impl Format for ItemAbi {
...
+        } else {
+            abi_items.first().unwrap().0.value.span().start()
+        };
+
+        write_comments(
+            formatted_code,
+            std::ops::Range {
+                start: self.name.span().end(),
+                end,
+            },
```
bingcicle, line 44: I'm starting to think using `std::ops::Range` as an input doesn't look as nice, because in most cases it's formatted into this and takes up a lot of vertical space. Is it worth considering going back to just providing `usize, usize`?

bingcicle:
This should be gtg for reviews!

swayfmt/src/utils/language/attribute.rs
```diff
@@ -12,21 +13,42 @@ use sway_ast::{
 };
 use sway_types::{constants::DOC_COMMENT_ATTRIBUTE_NAME, Spanned};
 
-impl<T: Format> Format for Annotated<T> {
+impl<ItemKind: Format + Spanned> Format for Annotated<ItemKind> {
```
mitchmindtree, line 16: Just to clarify, this declares a generic type `ItemKind` bound by `Format + Spanned` and then implements `Format` for this `Annotated` generic type. Was this the intent here?

If you just want to implement for `Annotated<ItemKind>` (i.e. the `ItemKind` type from sway-ast), then that would be:
```suggestion
impl Format for Annotated<ItemKind> {
```
but if you still want to keep this generic and just add the `Spanned` requirement, this might be clearer as:
```suggestion
impl<T: Format + Spanned> Format for Annotated<T> {
```

swayfmt/src/comments.rs
```diff
@@ -1,26 +1,77 @@
...
+        &self.unformatted_code
+    }
+}
 
 pub fn has_comments<I: Iterator>(comments: I) -> bool {
     comments.peekable().peek().is_some()
 }
 
+/// This function collects newlines to insert after the comment span to preserve them.
+pub fn collect_newlines_after_comment_span(unformatted_code: &str) -> String {
```
mitchmindtree, line 37: Perhaps we should clarify in the comment that the given `unformatted_code` should only be the unformatted code of the slice between the comment end and the next item or comment?

mitchmindtree (commented):
Nice, new tests are looking good, just a couple Qs first!

sway-parse/src/token.rs
```diff
@@ -130,7 +130,19 @@ pub fn lex_commented(
         if character == '/' {
             match l.stream.peek() {
                 Some((_, '/')) => {
-                    token_trees.push(lex_line_comment(&mut l, end, index));
+                    let has_newline = src[..index]
```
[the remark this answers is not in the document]
bingcicle: Hmm, the thought behind this is that `take_while()` is only doing work while the predicate is true (`c.is_whitespace()`), and counting the newlines within that iteration, so its likely that the iterator isn't checking the entire `src` until the start (start since i call `rev()`)

swayfmt/src/comments.rs
```diff
@@ -32,23 +83,50 @@ pub fn maybe_write_comments_from_map(
...
-            // to support trailing comments.
-            writeln!(
-                formatted_code,
-                "{}{}",
-                formatter.shape.indent.to_string(&formatter.config)?,
-                comment.span().as_str(),
-            )?;
+        while let Some(comment) = comments_iter.next() {
+            let newlines = collect_newlines_after_comment_span(
+                &formatter.comments_context.unformatted_code()[comment.span().end()..],
```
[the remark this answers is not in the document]
bingcicle: This is the same as the above comment, but in the opposite direction - we're doing a `take_while()`, so I don't think this would happen. I will add some tests though just to keep this behaviour consistent + show that its working as intended

swayfmt/src/comments.rs
bingcicle: Hmm actually, let me rename this function. It doesn't really make sense to be named `after_comment_span` if i'm not passing in a comment here.

bingcicle: renamed + refactored it to take in a `&CommentsContext` and a `&Comment` instead of a raw `&str`

sway-parse/src/token.rs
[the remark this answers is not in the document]
mitchmindtree: I mean that `src[..index]` will start at the very beginning of all of the `src` for the module, no? The provided `..index` range means "everything up until this index" - is that actually what's intended?

swayfmt/src/comments.rs
```diff
@@ -1,26 +1,81 @@
...
+        &self.unformatted_code
+    }
+}
 
 pub fn has_comments<I: Iterator>(comments: I) -> bool {
     comments.peekable().peek().is_some()
 }
 
+/// This function collects newlines to insert after the comment span to preserve them.
+/// The given 'unformatted_code' should be
```
mitchmindtree, line 37: Looks like comment was cut short?

mitchmindtree (commented):
Lgtm! One last nit

sway-parse/src/token.rs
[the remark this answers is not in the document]
bingcicle: Hm, at the time of writing it was the quick and dirty way (to me) to get a new reversed iterator to start looking backwards for newline tokens.

I'm not sure if this has cost implications, if this is what the question is about, but in terms of correctness it should stop at the first non-whitespace character regardless within the source code.

I added a new `search_end` in my latest commit to end the search at the last token's `span().end()` instead, but I'm not sure if this overcomplicates the problem.

swayfmt/src/comments.rs
```diff
@@ -32,24 +86,136 @@ pub fn maybe_write_comments_from_map(
...
+                            "{}{}{}",
+                            formatter.shape.indent.to_string(&formatter.config)?,
+                            comment.span().as_str(),
+                            newlines
+                        )?;
+                    }
+                    CommentKind::Trailing => {
+                        write_trailing_comment(formatted_code, comment)?;
+                        if comments_iter.peek().is_none() {
+                            write!(
```
sdankel, line 106: Wouldn't this write `formatted_code` twice since `write_trailing_comment` also writes it? Could you add a comment here?

sdankel (approved):
LGTM, just one question

swayfmt/src/items/item_abi.rs
```diff
@@ -28,22 +29,72 @@ impl Format for ItemAbi {
...
+        let end = if abi_items.is_empty() {
+            self.span().end()
+        }
+        // if there are existing abi items, we want to end before the hash token
+        // This is OK to unwrap since we've checked is_empty() above.
+        else if let Some(first_attr) = abi_items.first().unwrap().0.attribute_list.first() {
+            first_attr.span().start()
+        } else {
+            abi_items.first().unwrap().0.value.span().start()
+        };
```
kayagokalp, line 44: Just a tiny suggestion if you prefer not using `unwrap`. I am good either way

```rust
let end = if let Some(first_abi_item) = abi_items.first() {
  let fn_signature = &first_abi_item.0;
  if let Some(first_attr)  = fn_signature.attribute_list.first() {
      first_attr.span().start()
  }else {
      fn_signature.value.span().start()
  }
}else {
  self.span().end()
};
```

kayagokalp (approved):
:rocket: :shipit: 

I left an unimportant suggestion/personal preference, please feel free to resolve it as I am good with this either way

bingcicle:
BTW, a note here: I realized it may not be clear how I'm testing if this method truly works - but what I'm doing is commenting out `handle_comments` which contains the entire logic that is currently in-charge of comment formatting, and checking if the tests pass without that logic.

mitchmindtree (approved):

swayfmt/src/comments.rs
bingcicle: I'm actually going to fix this in a follow up PR - this behaviour probably isn't what we want, its an indirection to indent within write_comments(), but we should we do it one-level-up instead.

kayagokalp (approved):

sway-parse/src/token.rs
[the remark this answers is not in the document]
bingcicle: I marked this as resolved for now so I could get this merged since it changes some of the key code that causes conflicts - we could address this in a followup PR if it's an issue! @mitchmindtree

PR #4007, opened 2023-02-07
Fixes regression caused by enum and struct callpaths.

### LINKED ISSUE #3998: Regression after struct and enum changed to use call paths.
The following test case stop working:

https://github.com/FuelLabs/fuels-rs/tree/master/packages/fuels/tests/types/vectors

The regression was caused by https://github.com/FuelLabs/sway/commit/3fbad7b5f0935a19522cb50c6e3979211fe8939f.

Printing errors with full call paths looks like:
```
120 |         assert(arg == expected);
    |                       ^^^^^^^^ This parameter was declared as type
std::vec::Vec<vectors::eq_impls::data_structures::SomeStruct<std::vec::Vec<std::vec::Vec<u32>>>>, but argument of type
std::vec::Vec<vectors::data_structures::SomeStruct<std::vec::Vec<std::vec::Vec<u32>>>> was provided.
121 |     }
```

esdrubal:
## Description

The issue is that struct or enum can be imported with different callpaths depending the module they are imported from.

For instance we could have a struct with callpath
`my_script::data_structures::SomeStruct` in a module and in another module as `my_script::eq_impls::data_structures::SomeStruct`. This makes callpaths unreliable to compare declarations.

Closes #3998

Reopens #418

Disables tests multiple_enums_with_the_same_name and multiple_structs_with_the_same_name because they relied in callpath comparisons to work. 

## Checklist


mohammadfawaz:
Thanks for the quick fix @esdrubal 

Is the issue here that the call path used in type unification does not always refer to _actual_ struct/enum declaration? I would have assumed that there is a unique path for the struct/enum declaration. Where is the other path coming from? Is it from the `impl trait` block?

emilyaherbert:
Hi @esdrubal, I think what you are saying is that even though two `TypeInfo::Enum`s may point to the same type syntactically, relying on the call paths inside `TypeInfo::Enum` is unreliable as these can be relative call paths? Do I have this right?

If so, you might be interested in taking a look at #3744, among other things this PR changes `TypeInfo::Enum` and others to remove the call path field and instead use `DeclId`s directly. That would be a fix to this problem.

CC @mohammadfawaz

mohammadfawaz:
> Hi @esdrubal, I think what you are saying is that even though two `TypeInfo::Enum`s may point to the same type syntactically, relying on the call paths inside `TypeInfo::Enum` is unreliable as these can be relative call paths? Do I have this right?
> 
> If so, you might be interested in taking a look at #3744, among other things this PR changes `TypeInfo::Enum` and others to remove the call path field and instead use `DeclId`s directly. That would be a fix to this problem.
> 
> CC @mohammadfawaz

Oh excellent. In that case, I suggest we merge this PR to fix the regression until #3744 is complete.

emilyaherbert (approved):
One of the temporary approaches we could take would be to add "unimplemented" compile errors for using types with a call path. i.e. the case defined here #1644. But if my hunch is right, then this unimplemented error will catch the case in #418, thereby disallowing any syntax like `module0::Thing::new()`, which is a breaking change and undesirable.

I can think of some temporary hacks to prevent #418, but I'm more inclined to block that bug behind #3744 instead. Approving this patch right now though.

mohammadfawaz (approved):

esdrubal:
> I think what you are saying is that even though two `TypeInfo::Enum`s may point to the same type syntactically, relying on the call paths inside `TypeInfo::Enum` is unreliable as these can be relative call paths? Do I have this right?

Yes that's exactly the problem. If we have two modules A and B using an enum in module C, in module A we will have the enum declaration with call path `my_script::A::C::MyEnum` and in module B we will have the enum declaration with call path `my_script::B::C::MyEnum`. Then the unifier fails when using call paths.

PR #4014, opened 2023-02-07
Add `HashWithEngines` impls for more types

emilyaherbert:
## Description

This PR is a subset of #3744. It adds a `HashWithEngines` for the types required in #3744. Note---this isn't a necessary change into `master` right at this current moment, just simply a way of breaking up #3744 😄 

This PR also makes a few corrections to some of the existing implementations for `HashWithEngines` which were incorrectly comparing `TypeId`s.

## Checklist


sway-core/src/decl_engine/wrapper.rs
```diff
@@ -189,6 +232,21 @@ impl From<ty::TyEnumDeclaration> for (DeclWrapper, Span) {
...
+            DeclWrapper::Trait(_) => 2,
+            DeclWrapper::TraitFn(_) => 3,
+            DeclWrapper::ImplTrait(_) => 4,
+            DeclWrapper::Struct(_) => 5,
+            DeclWrapper::Storage(_) => 6,
+            DeclWrapper::Abi(_) => 7,
+            DeclWrapper::Constant(_) => 8,
+            DeclWrapper::Enum(_) => 9,
+        }
+    }
```
otrho, line 248: Can [std::mem::discriminant](https://doc.rust-lang.org/std/mem/fn.discriminant.html) be used here, since `Discriminant` implements `Hash`?

sway-core/src/decl_engine/wrapper.rs
```diff
@@ -56,6 +53,52 @@ impl PartialEqWithEngines for DeclWrapper {
...
+                decl.hash(state, engines);
+            }
+            DeclWrapper::TraitFn(decl) => {
+                state.write_u8(self.discriminant_value());
+                decl.hash(state, engines);
+            }
+            DeclWrapper::ImplTrait(decl) => {
+                state.write_u8(self.discriminant_value());
+                decl.hash(state, engines);
+            }
```
otrho, line 77: This might be subjective, but I'd whip up a little macro for most of these (and the others in other types in this PR).

sway-core/src/decl_engine/wrapper.rs
mitchmindtree: Another common pattern is to add another `enum DeclKind { Unknown, Function, Trait, etc... }` that can be used for distinguishing between decl kinds, implement `Hash` etc.

That said I've never seen discriminant used in the wild - looks like an interesting option!

sway-core/src/language/ty/declaration/constant.rs
```diff
@@ -29,7 +31,15 @@ impl PartialEqWithEngines for TyConstantDeclaration {
...
+    }
+}
+
+impl HashWithEngines for TyConstantDeclaration {
+    fn hash<H: Hasher>(&self, state: &mut H, engines: Engines<'_>) {
+        let type_engine = engines.te();
+        self.name.hash(state);
+        self.value.hash(state, engines);
+        self.visibility.hash(state);
+        type_engine.get(self.return_type).hash(state, engines);
```
mohammadfawaz, line 43: Should we also hash `is_configurable` here?

mohammadfawaz: In fact, with all of these custom implementations of `hash`, how do we decide what fields should be hashed? And how do we make sure that this is future proof when new fields are added? (hopefully this is rare but `is_configurable` is an example of a new field that may have been missed, even in `eq` above).

emilyaherbert: > how do we decide what fields should be hashed?

There is not any perfect science to it... although there should be. In general there is poor separation between the information needed for static analysis + type checking (only some stuff should be hashed) and the information needed for the consumers of the compiler, LSP, etc, (other stuff should to be hashed). The implementations of `Hash` and such have historically accommodated the former case, leading to workarounds like [`BaseIdent`](https://github.com/FuelLabs/sway/commit/80a8d1b7382fd70a7e417c30a94d30778d56e762) being introduced.

Regarding this larger problem, I'm not yet sure what the best course of action should be. I'll need to do some research to learn about different approaches and the current landscape.

> And how do we make sure that this is future proof when new fields are added?

So for what to do right now, I propose that I change this PR slightly so that adding new fields will give a Rust error unless you decide if the new field should be hashed or not. And I'll add little explanation blurbs for the fields left out. What are your thoughts?

mohammadfawaz: Your proposal seems reasonable to me. And since we using a custom hash trait instead of the standard `Hash`, maybe we need two separate traits with two separate implementations to be used in two different scenarios. One would call the other.

mohammadfawaz (approved):
This is fantastic. Great idea destructuring `self` like that and I appreciate the extract comments.

PR #4026, opened 2023-02-08
Changes TypeInfo::Custom to have call paths.

### LINKED ISSUE #1644: Allow `TypeInfo::Custom`s with `CallPath`s

esdrubal:
## Description

With these changes TypeInfo::Custom have a `CallPath` instead of an `Ident`.

Changes `TypeEngine::resolve` to resolve TypeInfo::Custom by `CallPath` instead of `Ident`.

Changes `typed_scrutinee.rs` `type_check_enum` to support resolve with `CallPath`.

Closes #1644

## Checklist


test/src/e2e_vm_tests/test_programs/should_pass/language/typeinfo_custom_callpath/src/main.sw
```diff
@@ -0,0 +1,21 @@
...
+dep foo;
+
+struct Bar {
+  baz: foo::Foo
+}
+
+
+fn main() {
+    let x = Bar {
+        baz: foo::Foo::A
```
mohammadfawaz, line 12: so being able to write fully qualified paths here is new right? If so, we should probably add some more tests. For example
- A negative test to show that an invalid paths leads to an error
- A path that starts with `::` if that's allowed.
- A test that both imports `A` and uses the paths `foo::Foo::A`.

esdrubal: Added tests for the cases you mentioned. Also tested when enum is private instead of public which we could in the last revision write fully qualified paths to private enum without any error.

mohammadfawaz (approved):
Looks good to me! Thanks @esdrubal

PR #4038, opened 2023-02-10
bug(fmt): do not iteratively call forc-fmt over members on the root dir

### LINKED ISSUE #4027: bug(fmt): forc-fmt is inefficently formatting workspaces
While reformatting parts of `forc-fmt` I discovered this inefficiency:

In forc-fmt when formatting the workspace, we're currently starting the formatting in the root of the workspace while iterating over the members ([link](https://github.com/FuelLabs/sway/blob/b6f19a3be7b2fb5ef88e358a926854dac10cb281/forc-plugins/forc-fmt/src/main.rs#L72-L75)). In each iteration we call `format_pkg_at_dir()` and within that we're doing [`get_sway_files()`](https://github.com/FuelLabs/sway/blob/b6f19a3be7b2fb5ef88e358a926854dac10cb281/forc-plugins/forc-fmt/src/main.rs#L103). This means that we're formatting the same files X number of times for X members declared in the workspace.

What we want to do is to call `format_pkg_at_dir()` on each **subdirectory** of the workspace that contains a forc manifest because [not every workspace declares all subdirs as members](https://github.com/FuelLabs/sway-applications/blob/0132a4a89bb0f9c1a5d016db609adead87c43626/AMM/Forc.toml), as `libraries` are excluded here even though it contains .sw files. We would also want to collect any possible .sw files on the root itself.

Have a fix on the way for this!

bingcicle:
## Description

closes #4027

This is a fix for the formatter being called to format multiple times.

This implements the suggested changes described in the issue, and refactors some of the code:

- `format_file()` is now renamed to `write_file_formatted()`, since all it does is write to a file.
- Extract logic of formatting a single file into `format_file()`. This means its reusable at 3 levels: formatting a single file, formatting a package and formatting a workspace.
- Handles formatting by workspace vs package - formatting a workspace is handled slightly differently. Formatting a package has no changes - it still works the same. However, for workspaces, we want to format all files at the root, then find for each subdirectories (inclusively), all the nested directories with a manifest inside. If it exists, we are interested in formatting it. If the subdirectory happens to have a `swayfmt.toml` inside it, we will prioritize that configuration file. Generally, we should prefer member config > workspace config > default. This nuance is [explained in a comment](https://github.com/FuelLabs/sway/pull/4038/files#diff-bf9b5d7023fff817f83ae31bd01e66d2788178a0705a2c638ce79d7990cc374aR190-R193).
- Extract logic of formatting a manifest into `format_manifest()`

It's faster by quite a bit, with some good ol' primitive testing:

image

new (fmt.sh):
```
for i in `seq 1 13`; do
  ./forc-fmt --path AMM
done
```

The above `forc-fmt` is built on `--release` and I moved it to the sway apps repo for testing. Looped it 13 times because that's the no. of projects in sway-applications.

old (fmt-old.sh):
```
for i in `seq 1 13`; do
  forc fmt --path AMM
done
```

This one is just the current formatter.

## Checklist


kayagokalp (dismissed):
utACK

forc-plugins/forc-fmt/src/main.rs
```diff
@@ -51,9 +52,16 @@ fn run() -> Result<()> {
     let app = App::parse();
 
     if let Some(f) = app.file.as_ref() {
+        let mut formatter = Formatter::default();
```
sdankel, line 55: Should we be using `Formatter::from_dir` instead of the default formatter here?

forc-plugins/forc-fmt/src/main.rs
```diff
@@ -69,29 +77,177 @@ fn run() -> Result<()> {
...
+    while let Some(next_dir) = dirs_to_search.pop() {
+        if let Ok(read_dir) = fs::read_dir(next_dir) {
+            for entry in read_dir.filter_map(|res| res.ok()) {
+                let path = entry.path();
+
+                if path.is_dir() {
+                    dirs_to_search.push(path.clone());
+                    if path.join(constants::MANIFEST_FILE_NAME).exists() {
+                        dirs_to_format.push(path);
+                    }
```
sdankel, line 108: Is this a valid directory structure? Say I have a project called `fizzbuzz` with the following structure:

```
.
├── Forc.lock
├── Forc.toml
├── proj1
│   ├── Forc.toml
│   └── src
│       └── main.sw
└── src
    └── main.sw
```

If so, I think this will still format sub projects twice for nested sway projects. In this example,

`get_sway_dirs` will return: `[fizzbuzz, proj1]`

`proj1` will get formatted twice, once with the config from `proj1`'s manifest and once with the config from `fizzbuzz`'s manifest.

One way to prevent this would be, in `format_pkg_at_dir`, to keep track of which files have been formatted and skip them if so. I think you'd also need to iterate through the result of `get_sway_dirs` in reverse to ensure the nested subdirectory manifest was used for formatting before the outer directory's manifest.

```
for sub_dir in get_sway_dirs(dir.to_path_buf()).reverse()
```

Same thing for `format_workspace_at_dir`

forc-plugins/forc-fmt/src/main.rs
```diff
@@ -104,49 +260,13 @@ fn format_pkg_at_dir(app: &App, dir: &Path, formatter: &mut Formatter) -> Result
...
-                } else if formatted_content != file_content {
-                    contains_edits = true;
-                    error!("\nManifest Forc.toml improperly formatted");
-                    display_file_diff(&file_content, &formatted_content)?;
-                } else {
-                    info!("\nManifest Forc.toml properly formatted")
-                }
+            if let Ok(edited) = format_manifest(app, manifest_file) {
+                contains_edits = edited;
             }
```
sdankel, line 271: It would be great to add a test for a sway workspace with multiple projects (and nested projects, if those are valid) to verify that each sway file is only getting formatted once.

sdankel (commented):
This is a great improvement! Happy to approve with a test!

forc-plugins/forc-fmt/src/main.rs
bingcicle: Good catch! If there are nested packages within a package when handling `format_pkg_at_dir()` this would certainly be an issue, though I'm not sure if it's valid to do that. `forc` allows me to build a random project i spun up, so that means it must be allowed, and we should probably handle this scenario. I'll go with your suggestion to fix that!

However, in the case of workspace formatting, since we're starting at with `workspace_dir` already inside `dirs_to_search`, `workspace_dir` itself isn't added to `dirs_to_format`! ;) Of course, this assumes that there are no .sw files in the root, hence the code block [here](https://github.com/FuelLabs/sway/pull/4038/files#diff-bf9b5d7023fff817f83ae31bd01e66d2788178a0705a2c638ce79d7990cc374aR174-R188) that handles that assumption. So I think the possible multiple formats can occur at the package level and not the workspace level, which `get_sway_dirs()` doesn't handle.

forc-plugins/forc-fmt/src/main.rs
```diff
@@ -69,29 +77,177 @@ fn run() -> Result<()> {
...
+
+    // Format subdirectories. Note that we do not call format on members directly here, since
+    // in workspaces, it is perfectly valid to have subdirectories containing Sway files,
+    // yet not be a member of the workspace.
+    for sub_dir in get_sway_dirs(dir.to_path_buf()) {
+        // Here, we cannot simply call Formatter::from_dir() and rely on defaults
+        // if there is not swayfmt.toml in the sub directory because we still want
+        // to use the swayfmt.toml at the workspace root (if any).
+        // In order of priority: member > workspace > default.
+        if members.contains(&sub_dir.join(constants::MANIFEST_FILE_NAME)) {
```
bingcicle, line 198: Hmm, it also just occurred to me after your comment @sdankel that this doesn't handle the case where a subdir is not a member but has a `swayfmt.toml`. The [AMM](https://github.com/FuelLabs/sway-applications/tree/0132a4a89bb0f9c1a5d016db609adead87c43626/AMM) is one such example of a project - the forc projects within `libraries` are not declared as members.

forc-plugins/forc-fmt/src/main.rs
bingcicle: Opened https://github.com/FuelLabs/sway/issues/4050 to track this

forc-plugins/forc-fmt/src/main.rs
bingcicle: Perhaps we just want to add timing metrics to our [examples-checker script](https://github.com/FuelLabs/sway/tree/master/scripts/examples-checker)? Think maybe checking that each file is only formatted once may be too specific to check :thinking: 

In this case this seems like a regression that could've been prevented with closer review (which I was a reviewer for...)

forc-plugins/forc-fmt/src/main.rs
```diff
@@ -63,35 +76,178 @@ fn run() -> Result<()> {
...
+            // One or more files are not formatted, exit with error
+            bail!("Files contain formatting violations.");
+        } else {
+            // All files are formatted, exit cleanly
+            Ok(())
+        }
+    } else {
+        Ok(())
+    }
+}
```
sdankel, line 222: nit

```suggestion
    if app.check && contains_edits {
        // One or more files are not formatted, exit with error
        bail!("Files contain formatting violations.");
    }
    Ok(())
}
```

forc-plugins/forc-fmt/src/main.rs
```diff
@@ -50,10 +51,22 @@ fn main() {
 fn run() -> Result<()> {
     let app = App::parse();
 
+    let dir = match app.path.as_ref() {
+        Some(path) => PathBuf::from(path),
+        None => std::env::current_dir()?,
+    };
+
     if let Some(f) = app.file.as_ref() {
+        let mut formatter = Formatter::from_dir(&dir)?;
```
sdankel, line 60: nit: could put this above on line 58 and use it again on line 87

forc-plugins/forc-fmt/src/main.rs
```diff
@@ -63,35 +76,178 @@ fn run() -> Result<()> {
...
+                f,
+                BuildTarget::default(),
+            )
+        });
+        match Formatter::format(formatter, file_content.clone(), build_config.as_ref()) {
             Ok(formatted_content) => {
-                format_file(path, &formatted_content)?;
+                if app.check {
+                    if *file_content != formatted_content {
+                        info!("\n{:?}\n", file);
```
sdankel, line 143: Did you mean to leave this in here? If so, maybe add a prefix to the log message

```suggestion
                        info!("File was edited by formatter: \n{:?}\n", file);
```

forc-plugins/forc-fmt/src/main.rs
```diff
@@ -63,35 +76,178 @@ fn run() -> Result<()> {
...
+                } else {
+                    write_file_formatted(&file, &formatted_content)?;
+                }
+
+                return Ok(edited);
             }
             Err(err) => {
+                // there could still be Sway files that are not part of the build
+                error!("\nThis file: {:?} is not part of the build", file);
                 error!("{}\n", err);
```
sdankel, line 156:
```suggestion
                error!("\nThis file: {:?} is not part of the build\n{}\n", file, err);
```

forc-plugins/forc-fmt/src/main.rs
```diff
@@ -63,35 +76,178 @@ fn run() -> Result<()> {
...
             }
         }
     }
 
-    Ok(())
+    bail!("Could not read file")
```
sdankel, line 161:
```suggestion
    bail!("Could not read file: {:?}", file)
```

forc-plugins/forc-fmt/src/main.rs
```diff
@@ -63,35 +76,178 @@ fn run() -> Result<()> {
...
+    for member_path in workspace.member_paths()? {
+        members.push(member_path)
+    }
+
+    // Format files at the root - we do not want to start calling format_pkg_at_dir() here,
+    // since this would mean we format twice on each subdirectory.
+    if let Ok(read_dir) = fs::read_dir(dir) {
+        for entry in read_dir.filter_map(|res| res.ok()) {
+            let path = entry.path();
+            if path.is_file() && is_sway_file(&path) {
```
sdankel, line 179: Consider moving the `is_file` check inside `is_sway_file`, since it seems like we'd always want to check these together:

```
pub fn is_sway_file(file: &Path) -> bool {
    file.is_file() && Some(OsStr::new(constants::SWAY_EXTENSION)) == file.extension()
}
```

forc-plugins/forc-fmt/src/main.rs
```diff
@@ -63,35 +76,178 @@ fn run() -> Result<()> {
...
+        let taplo_alphabetize = taplo_fmt::Options {
+            reorder_keys: true,
+            ..Default::default()
+        };
+        let formatted_content = taplo_fmt::format(&manifest_content, taplo_alphabetize);
+        if !app.check {
+            write_file_formatted(&manifest_file, &formatted_content)?;
+        } else if formatted_content != manifest_content {
+            edited = true;
+            error!("\nManifest Forc.toml improperly formatted");
```
sdankel, line 241: Is this really an error? Seems like a warning/info to me

```suggestion
            error!("Improperly formatted manifest file: {}", manifest_file);
```

forc-plugins/forc-fmt/src/main.rs
```diff
@@ -63,35 +76,178 @@ fn run() -> Result<()> {
...
+            display_file_diff(&manifest_content, &formatted_content)?;
+        } else {
+            info!("\nManifest Forc.toml properly formatted")
```
sdankel, line 244: Is it necessary to put newlines in tracing error calls?

```suggestion
            info!("Manifest file was formatted correctly: {}", manifest_file);
```

forc-plugins/forc-fmt/src/main.rs
```diff
@@ -104,49 +260,13 @@ fn format_pkg_at_dir(app: &App, dir: &Path, formatter: &mut Formatter) -> Result
...
-                    format_file(&manifest_file, &formatted_content)?;
-                } else if formatted_content != file_content {
-                    contains_edits = true;
-                    error!("\nManifest Forc.toml improperly formatted");
-                    display_file_diff(&file_content, &formatted_content)?;
-                } else {
-                    info!("\nManifest Forc.toml properly formatted")
-                }
+            if let Ok(edited) = format_manifest(app, manifest_file) {
+                contains_edits = edited;
```
sdankel, line 269: If the `format_file` returns true and `format_manifest` returns false, we still want `contains_edits` to be true, right?

```suggestion
            if let Ok(edited) = format_manifest(app, manifest_file) {
                contains_edits = contains_edits || edited;
```

forc-plugins/forc-fmt/src/main.rs
sdankel: Fair enough! The important thing is that the execution time of `forc fmt` stays low.

forc-plugins/forc-fmt/src/main.rs
bingcicle: good catch!

forc-plugins/forc-fmt/src/main.rs
bingcicle: This is if the formatter was run with `--check` i.e. `app.check == true`, which should error out if the manifest is not properly formatted.

forc-plugins/forc-fmt/src/main.rs
bingcicle: Created https://github.com/FuelLabs/sway/issues/4073 to track this

sdankel (approved):

kayagokalp (approved):

PR #4068, opened 2023-02-13
beta-3 updates

mehtavishwa30:
## Description

Ongoing PR for beta-3 changes in the Sway repo:
1. Made minor changes in the Sway quickstart for beta-3

### DIFF
--- a/docs/book/src/introduction/sway_quickstart.md
+++ b/docs/book/src/introduction/sway_quickstart.md
@@ -188,7 +188,7 @@ where `--url` parameter is the URL where your local node is hosted.
 This should produce some output in `stdout` that looks like this:
 
 ```console
-$ forc deploy --unsigned
+$ forc-deploy --unsigned
   Compiled library "core".
   Compiled library "std".
   Compiled contract "counter_contract".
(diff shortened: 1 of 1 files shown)


kayagokalp (changes requested):
Left a single nit otherwise lgtm ✅ 

Some explanation around `forc-deploy` and `forc deploy` as it seems like popping-up around in slack frequently:

# Forc plugins

Forc basically only have the native commands such as `build`, `test`, `check`, `clean`, `new/init`, `update` maybe one or two more. These are baked into `forc` and packed with it all the time.  On top of this forc supports plugins just like cargo. So anyone can write plugins to forc that does something meaningful. There are some plugins that are maintained by us, the sway tooling team, such as `forc-fmt`, `forc-doc`, `forc-tx`, `forc-client`. These plugins are distributed with `fuelup` and installed automatically. Once you type `forc x` if x is not a native command, forc interprets it as a plugin name and tries to call `forc-x`. If a binary named `forc-x` can be found in your `PATH` than `forc-x` plugin would start working otherwise you would see an error. So executing `forc deploy` and `forc-deploy` is equivalent and the `forc deploy` one is the preferred way as it provides a better UX.

mehtavishwa30:
@kayagokalp thanks so much for the explanation! there was indeed some initial confusion around it for me but that sounds reasonable. Committed the suggestion and asked for a review. :)

mohammadfawaz (approved):

PR #4074, opened 2023-02-14
ci(fmt): time examples-checker and do CI checks

### LINKED ISSUE #4073: ci(fmt): time the formatter in `examples-checker`
To prevent a similar issue like #4027 from happening again we should time the `examples-checker` to check for regressions in the performance of the formatter. While the formatter is pretty fast already, i'd still rather the formatter take 0.4s than 1s to format multiple projects/a big project ;)

Ideally, we should have some action that triggers a bot to post the metrics via a comment on any PR that changes files in `swayfmt` or `forc-fmt`. I don't think we should hard reject any PR that causes a significant regression, that should probably be left up to the reviewers/code owners to decide.

If we do not have this automated action, any reviewer who reviews a PR labelled formatter should be aware of this timing check, and check the CI job for the metrics.

bingcicle:
## Description

Closes #4073

## Checklist


bingcicle:
This adds about ~5 minutes to install a master version of `forc-fmt` vs the current version of `forc-fmt`, runs it through the examples and prints some diagnostics for it: https://github.com/FuelLabs/sway/actions/runs/4190149757/jobs/7263258070

I don't think there's a way to auto-run this, so it seems like if we want to use this, we would have to manually go into the job to see the results :thinking:

bingcicle:
Going to close this for now because I don't think the added CI time is really worth it for this.

PR #4081, opened 2023-02-14
Support inner doc attributes with `#!` & `//!` syntax

### LINKED ISSUE #3171: Support nested module attributes `#![attr]` - i.e. attributes intended for the enclosing module
Not sure if this is the right name for them, but this will likely be necessary for module-level documentation. E.g.

```rust
//! These are some module docs talking about `foo`.

library foo;

// ...
```

The `//!` doc comment at the top should convert into an attribute that looks something like:
```rust
#![doc("These are some module docs talking about `foo`.")]
```
which would be added to the set of attributes associated with the module `foo`.

eureka-cpu:
## Description
This adds the `HashBangToken` (`#!`) which will support nested module doc comments as well as inner doc comments for annotated items.

- [x] Add `HashBangToken`
- [x] Differentiate hash tokens on `AttributeDecl`
- [ ] Fix directional attribute parsing for `Inner` & `Outer` doc comments ( This is optional, the only downside of not implementing this right away is that inner and outer docs are treated equally )
- [x] Add attributes to `Module`
- [x] Complete support for module level attributes

## Checklist

- [x] I have linked to any relevant issues. Closes #3171 
- [ ] I have added (or requested a maintainer to add) the necessary `Breaking*` or `New Feature` labels where relevant. @mohammadfawaz 

### DIFF
--- a/sway-core/src/lib.rs
+++ b/sway-core/src/lib.rs
@@ -136,14 +136,7 @@ fn module_attrs_to_map(
                 DOC_COMMENT_ATTRIBUTE_NAME => Some(AttributeKind::DocComment),
                 _ => None,
             } {
-                match attrs_map.get_mut(&attr_kind) {
-                    Some(old_args) => {
-                        old_args.push(attribute);
-                    }
-                    None => {
-                        attrs_map.insert(attr_kind, vec![attribute]);
-                    }
-                }
+                attrs_map.entry(attr_kind).or_default().push(attribute);
             }
         }
     }
--- a/sway-core/src/lib.rs
+++ b/sway-core/src/lib.rs
@@ -114,7 +114,11 @@ fn parse_in_memory(
         module.clone(),
     )?;
     let submodules = Default::default();
-    let root = parsed::ParseModule { tree, submodules };
+    let root = parsed::ParseModule {
+        tree,
+        submodules,
+        // attributes,
+    };
     let lexed_program = lexed::LexedProgram::new(
         kind.clone(),
         lexed::LexedModule {
@@ -225,6 +229,7 @@ fn parse_module_tree(
     let parsed = parsed::ParseModule {
         tree,
         submodules: submodules.parsed,
+        // attributes,
     };
     Ok((kind, lexed, parsed))
 }
--- a/sway-core/src/lib.rs
+++ b/sway-core/src/lib.rs
@@ -31,12 +31,15 @@ use metadata::MetadataManager;
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
+use sway_ast::AttributeDecl;
 use sway_error::handler::{ErrorEmitted, Handler};
 use sway_ir::{
     register_known_passes, Context, Kind, Module, PassManager, PassManagerConfig,
     CONSTCOMBINE_NAME, DCE_NAME, FUNC_DCE_NAME, INLINE_NONPREDICATE_NAME, INLINE_PREDICATE_NAME,
     MEM2REG_NAME, SIMPLIFYCFG_NAME,
 };
+use sway_types::constants::DOC_COMMENT_ATTRIBUTE_NAME;
+use transform::{Attribute, AttributeKind, AttributesMap};
 
 pub use semantic_analysis::namespace::{self, Namespace};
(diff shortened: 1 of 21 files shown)


eureka-cpu:
I made a test branch to try out the module level docs and........


https://user-images.githubusercontent.com/57543709/218918905-ebe1b46f-1d14-4cf0-b700-5b47d5cb0c46.mov

sway-parse/src/item/mod.rs
```diff
@@ -206,9 +206,9 @@ mod tests {
         let item = parse_item(
             r#"
             // I will be ignored.
-            //! I will be ignored.
+            //! I can't feel the way I did before
```
mitchmindtree, line 209: :smiling_face_with_tear:

sway-parse/src/module.rs
```diff
@@ -23,16 +30,50 @@ impl Parse for ModuleKind {
...
+            // TODO: Use a Literal instead of an Ident when Attribute args
+            // start supporting them and remove `Ident::new_no_trim`.
+            let value = Ident::new_no_trim(doc_comment.content_span.clone());
+            let hash_kind = match &doc_comment.doc_style {
+                DocStyle::Inner => Ok(AttributeHashKind::Inner(HashBangToken::new(
+                    doc_comment.span.clone(),
+                ))),
+                DocStyle::Outer => Err(parser.emit_error(ParseErrorKind::ExpectedModuleDocComment)),
+            }
+            .unwrap();
```
mitchmindtree, line 48: Do we want to `unwrap` here? Or should we return an error somehow? (Not exactly sure how `ParseResult` works off the top of my head).

mitchmindtree (commented):
Awesome stuff @eureka-cpu! Very exciting, I think this will do wonders for the accessibility of our docs. That `std` root module is begging for a welcome msg :smiling_face_with_tear: Just a couple tiny suggestions, otherwise looks great.

sway-parse/src/item/mod.rs
eureka-cpu: https://youtu.be/LYU-8IFcDPw 🤘🏻

eureka-cpu:
> Awesome stuff @eureka-cpu! Very exciting, I think this will do wonders for the accessibility of our docs. That `std` root module is begging for a welcome msg 🥲 Just a couple tiny suggestions, otherwise looks great.

I'll open an issue for it but if we want to have the project root module to have a description I think we should add a description option in the project section of the `forc.toml`. What do you think?

sway-parse/src/module.rs
eureka-cpu: I'll look into it. I was using `Result` in the heat of the moment as a wrapper type.

sway-core/src/lib.rs
```diff
@@ -100,6 +103,53 @@ pub fn parse_tree_type(input: Arc<str>) -> CompileResult<parsed::TreeType> {
...
+                _ => None,
+            } {
+                match attrs_map.get_mut(&attr_kind) {
+                    Some(old_args) => {
+                        old_args.push(attribute);
+                    }
+                    None => {
+                        attrs_map.insert(attr_kind, vec![attribute]);
+                    }
+                }
```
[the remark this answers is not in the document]
eureka-cpu: ah nice, thanks (:

mitchmindtree:
> I'll open an issue for it but if we want to have the project root module to have a description I think we should add a description option in the project section of the forc.toml. I could then use that description as the project index attribute. What do you think?

Fwiw, this would differ a little to how cargo and Rust work today. E.g.  `description`s in `Cargo.toml` files tend to be quite short and concise, and crates.io favours rendering the project's README on the project page. docs.rs on the other hand (closer to our target) first presents the crate-root-level documentation (i.e. the root module documentation, so `main.rs` or `lib.rs`). Often these docs can be large and detailed and give an overview of how to navigate the package ([std](https://doc.rust-lang.org/stable/std/) is a good example). I think modelling this behaviour probably makes the most sense.

Oh btw, I think it's totally fine that we add std root docs in a follow-up PR! Just getting excited about it as all :)

mitchmindtree (approved):
utACK

otrho (approved):

Issue #3804, opened 2023-01-18
Refactor `forc-tx` `input`/`output` subcommand handling to use a manual `clap` trait implementation


mitchmindtree:
Currently `clap` appears to lack support for multiple trailing subcommands that contain arguments. As a result, `forc-tx` (introduced in #3582) provides a custom `Command::try_parse` command that first parses the transaction type, then manually parses the remaining `Input` and `Output`s.

This approach has the downside that the `input` and `output` subcommands don't show up in the `forc tx --help`.

We should investigate whether or not it's possible to provide a custom implementation for parsing the trailing inputs and outputs in a manner that `clap` can present in it's `--help` UI. This might involve implementing some clap traits manually for some wrapper type around the inputs and outputs, however with my experiments so far I've been unable to find a solution implementing either `Subcommand` or `ValueParser` that actually allows for parsing multiple trailing subcommands.

Opening this as a reminder to investigate further, and potentially open a discussion around parsing options for multiple trailing subcommands at the clap repo.

PR #4085, opened 2023-02-15
`forc-tx`: Improve error handling using `thiserror`. Fix `panic!`ing version/help output.

### LINKED ISSUE #3886: `forc tx --help` prints help successfully but within a `panic!` message
This is due to the current way we attempt to parse the trailing chain of `input`s and `output`s.

Until this is addressed, we'll need to omit this from the Sway book's Forc Plugins section as the mdbook-forc-documenter plugin uses the `--help` output to generate docs.

This is further motivation for #3804, ideally so that we can just call `Command::parse` at the top-level like a normal clap application and get regular coloured help output.

mitchmindtree:
## Description

This introduces a `Command::parse` constructor that emulates the behaviour of the `clap` parse constructor in order to provide a more consistent clap-like CLI experience. For context, we cannot use the generated `parse` command due to limitations in clap's ability to handle trailing subcommands as mentioned in #3804.

We switch to using `thiserror` in the library in order to allow for handling error cases (necessary for the new `parse` constructor) and to provide cleaner, more detailed error cause traces.

The help and version output no longer panic, and now behave like regular clap applications. Closes #3886.

Also adds an example command invocation to the end of the `--help` output.

## Checklist

### DIFF
--- a/forc-plugins/forc-tx/src/lib.rs
+++ b/forc-plugins/forc-tx/src/lib.rs
@@ -377,19 +377,19 @@ pub enum ConvertScriptTxError {
 /// Errors that can occur during transaction input conversion.
 #[derive(Debug, Error)]
 pub enum ConvertInputError {
-    #[error("failed to read message data from {path:?}")]
+    #[error("failed to read `--msg-data` from {path:?}")]
     MessageDataRead {
         path: PathBuf,
         #[source]
         err: std::io::Error,
     },
-    #[error("failed to read predicate from {path:?}")]
+    #[error("failed to read `--predicate` from {path:?}")]
     PredicateRead {
         path: PathBuf,
         #[source]
         err: std::io::Error,
     },
-    #[error("failed to read predicate data from {path:?}")]
+    #[error("failed to read `--predicate-data` from {path:?}")]
     PredicateDataRead {
         path: PathBuf,
         #[source]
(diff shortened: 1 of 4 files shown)


bingcicle (approved):
utACK

forc-plugins/forc-tx/src/lib.rs
```diff
@@ -306,29 +306,231 @@ pub struct OutputContractCreated {
...
+/// Errors that can occur during transaction input conversion.
+#[derive(Debug, Error)]
+pub enum ConvertInputError {
+    #[error("failed to read message data from {path:?}")]
+    MessageDataRead {
+        path: PathBuf,
+        #[source]
+        err: std::io::Error,
+    },
+    #[error("failed to read predicate from {path:?}")]
```
bingcicle, line 386:
```suggestion
    #[error("failed to read --predicate from {path:?}")]
```

Should we be consistent with printing the error here? Above, in the `BytecodeRead` error, we are printing the option flag as is, but here, we're not showing `--predicate`. IMO in this case, explicitly showing the option flag works better since it's a long command.

bingcicle (changes requested):

JoshuaBatty (approved):

bingcicle (approved):
utACK

otrho (approved):
utACK

PR #4086, opened 2023-02-15
feat: workspace level patch table added

### LINKED ISSUE #3958: `forc-pkg`: Support workspace-level `[patch]` table
Currently, I believe our `[patch]` table support is limited to individual packages (@kayagokalp correct me if I'm wrong :sweat_smile:).

We should match Rust's behaviour, where `[patch]` is applied at the workspace level. This is because the workspace is responsible for the overall package graph and lock file, and when a patch needs to be applied to a package in the graph it is most often applicable for all members of the workspace.

kayagokalp:
## Description
closes #3958.

This is a handy feature for tests with workspaces. We were already supporting a `patch` table feature for packages. This PR adds the same feature for workspaces. Basically we can patch every instance of a package with a different version. One main motivation behind this can be patching the standard library while working with an unreleased version of `forc`. With this feature we can pin the `std` version for the whole workspace so that all members in the workspace depending on `std` would depend on master version of the `std`.

## Checklist


forc-pkg/src/manifest.rs
```diff
@@ -240,6 +240,35 @@ impl PackageManifestFile {
         Ok(Self { manifest, path })
     }
 
+    /// Apply patch declarations from the workspace (which declares this package manifest as a
+    /// member) to this package manifest.
+    ///
+    /// This checks whether there is such workspace present before attempting to apply patches so
+    /// it is safe to use with standalone package manifests as well.
+    pub fn with_workspace_patches(&self) -> Self {
```
bingcicle, line 248: Would it be possible to have a patch in both a package and a workspace at the top-level? If yes, wouldn't the workspace patch take precedence over the package patch?

For example, if at the top-level, we're patching with `my-branch`. But in a specific member, we declare another patch.

forc-pkg/src/manifest.rs
```diff
@@ -368,6 +397,7 @@ impl PackageManifestFile {
                 if e.to_string().contains("could not find") {
                     return Ok(None);
                 } else {
+                    println!("{e:?}");
```
bingcicle, line 400:
```suggestion
```

docs/book/src/forc/workspaces.md
```diff
@@ -25,6 +26,23 @@ members = ["member1", "path/to/member2"]
 The `members` field accepts entries to be given in relative path with respect to the workspace root.
 Packages that are located within a workspace directory but are *not* contained within the `members` set are ignored.
 
+## The `[patch]` section
+
+The `[patch]` section can be used to override any dependency in the workspace dependency graph. The usage is the same with package level `[patch]` section and details can be seen [here](./manifest_reference.md#the-patch-section).
```
bingcicle, line 31: Even though it may seem obvious to us, I think we might want to add a note here that this will only take effect for packages that are declared as members. The [AMM project](https://github.com/FuelLabs/sway-applications/blob/master/AMM/project/Forc.toml) for example does not declare `libraries` as a member.

bingcicle (changes requested):

forc-pkg/src/manifest.rs
kayagokalp: Yes it is possible and it was intentionally done that way so that you can override any specific patch applied to individual packages from a single point (from the workspace that declares them as members). I had it documented but apparently deleted it while refactoring.

bingcicle: I see, so the patches declared in the workspace take higher priority over whatever the patches declared are downstream?

bingcicle: > I think it makes sense for the formatter to prioritize the package level one so that you can have specific formatting logic for each member. But at the same time, is it something that we should allow? While developing the mvp for the formatter we decided to be somewhat opinionated and does not allow a lot of configurability like the go formatter. 

Good question! Honestly, I myself am against too much configurability. I think having similar formatting enforced on a ecosystem-wide level seems better than enforcing formatting per-project (i like zig for this as well). I was under the assumption that we were going for a `rustfmt`-like experience (which tbh, I don't fancy) :sweat_smile: If it is the case that we want to push for a more opinionated formatter, I'd +1 for that, but that's a whole another topic.

As for patching, I do not really have a strong opinion on whether a package-level patch should take precedence over a workspace-level patch, it was just the order that felt natural to me, and was just asking questions to clarify my own understanding. Besides, I think it would probably be trivial to update this behavior further down the line if we find that users don't like this.

bingcicle (approved):
:shipit:

docs/book/src/forc/workspaces.md
```diff
@@ -25,6 +26,27 @@ members = ["member1", "path/to/member2"]
...
+
+If a member of a workspace has a patch table in its package manifest, the workspace level patch declarations are merged with package level declarations. If there are conflicting declarations (patching the same source with different target) between workspace level patch table and a member's patch table, workspace level patch declarations.
```
mitchmindtree, line 33: I think we may want to follow suit with cargo here and *disallow* package `[patch]` tables if there a workspace `[patch]` table exists.

Patching is specifically relevant to the package graph (represented by the Forc.lock file), and the package graph is always the same within a workspace *regardless* of whether you are only building one member, or many (it's just that when we are building one package, we skip nodes in the graph that are not relevant to the individual package).

Like the `Forc.lock` file, I think we might want to limit the `[patch]` table use to the workspace-level in the case that a workspace exists.

docs/book/src/forc/workspaces.md
```diff
@@ -25,6 +26,27 @@ members = ["member1", "path/to/member2"]
...
+
+> **NOTE:** Workspace level patch declarations only apply to members declared in `members` field.
```
mitchmindtree, line 35: Just to clarify, does this note refer to the current `[patch]`-table-merging behaviour, or the application of patches?

Patches from a `[patch]` table should be applied to *all* occurrences of the specified package within the package graph (not only to the top-level members). I *think* we already do this in our package `[patch]` table support, but just want to clarify.

kayagokalp: Yeah exactly, we already do this for workspace patch table as well but the wording is pretty confusing. Removed it

forc-pkg/src/manifest.rs
```diff
@@ -240,6 +240,36 @@ impl PackageManifestFile {
...
+            .ok()
+            .flatten()
+            .and_then(|workspace| workspace.patch.clone());
+        let package_patches = self.patch.clone();
+        let patch = match (workspace_patches, package_patches) {
+            (Some(_), Some(_)) => bail!("Found [patch] table both in workspace and member package's manifest file. Consider removing [patch] table from package's manifest file."),
+            (Some(workspace_patches), None) => anyhow::Ok(Some(workspace_patches)),
+            (None, Some(pkg_patches)) => anyhow::Ok(Some(pkg_patches)),
+            (None, None) => Ok(None),
+        }?;
```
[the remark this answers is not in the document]
kayagokalp: Ahhh yeah, thanks for catching that, i was returning `anyhow!("")` and to remove it i switched to `bail!` in the first place. Looks like I forgot to remove `Ok`s 🤦

mitchmindtree (commented):
Mostly LGTM, just a couple small final suggestions!

sdankel (approved):
utACK

JoshuaBatty (approved):

forc-pkg/src/source/mod.rs
```diff
@@ -179,19 +179,19 @@ impl Source {
...
+    fn dep_patch(
         &self,
         dep_name: &str,
-        manifest: &'manifest PackageManifestFile,
-    ) -> Option<&'manifest manifest::Dependency> {
+        manifest: &PackageManifestFile,
+    ) -> Result<Option<manifest::Dependency>> {
         if let Source::Git(git) = self {
-            if let Some(patches) = manifest.patch(&git.repo.to_string()) {
+            if let Some(patches) = manifest.resolve_patch(&git.repo.to_string())? {
```
mitchmindtree, line 188: Internally, this calls calls the `PackageManifest::workspace` method which does the I/O work of finding the workspace, reading the manifest then deserializing it every time we call this `resolve_patch` method.

I think this approach is fine for now, though in a follow-up PR we may want to consider refactoring this so that we find, read and deserialize the resolved patch table only *once*, as if the file is edited during the compilation process (not uncommon while running sway-lsp) we might occasionally encounter some strange inconsistencies. E.g. it could be worth doing this upon loading the manifest file itself, or immediately after. Then, our `PatchMap` type (or similar abstraction) could be the type that gets passed in here to check for patches.

mitchmindtree (approved):
:ship:

PR #4090, opened 2023-02-15
Update expected revert string in `storage_vec` tests to fix CI failure

mohammadfawaz:
Seems that the format of the revert receipts is now different in the newest versions of `fuel-core` and `fuels-rs`.

Also ran `cargo fmt` just as I'm passing by :)


eureka-cpu (approved):
:shipit:

PR #4093, opened 2023-02-16
fix: empty fn comments

### LINKED ISSUE #4092: Formatter moving comments below empty function
I'm writing simple examples and defined functions with only a comment in the body. On save, the formatter moves the comment under the function and moves the closing curly brace next to the opening one. For example:

```
fn on_even(num: u64) {
 // do something with even numbers
}
```

to
```
fn on_even(num: u64) {}
    // do something with even numbers
```

I'm on the nightly version of the toolchain, using` forc-lsp : 0.35.0+nightly.20230215.f7cbf83bd5`

bingcicle:
## Description

closes #4092

Note that this PR also moves the block indentation logic up one level into the `ItemFn` level instead of at the `CurlyBrace` impl.

## Checklist


swayfmt/src/items/item_fn/mod.rs
```diff
@@ -35,10 +36,18 @@ impl Format for ItemFn {
...
                     Self::close_curly_brace(formatted_code, formatter)?;
                 } else {
-                    write!(formatted_code, " {{}}")?;
+                    Self::open_curly_brace(formatted_code, formatter)?;
+                    let range: Range<usize> = self.span().into();
+                    let comments = formatter.comment_map.comments_between(&range);
+                    if has_comments(comments) {
+                        formatter.shape.block_indent(&formatter.config);
+                        maybe_write_comments_from_map(formatted_code, range, formatter)?;
+                    }
```
bingcicle, line 49: Honestly, the logic to check if a range has comments, and then write the comments + some other logic, could probably be better written. The use case of having to check if comments exist before writing them has come up twice now, because there are times when we might need to write a newline or do an indent before writing the comments (like here).

kayagokalp (dismissed):
utACK. Tests looks great, I wonder what happens if we have something like

```rust
fn single_comment_same_line() { /*this is a comment*/ }
```

bingcicle:
> utACK. Tests looks great, I wonder what happens if we have something like
> 
> ```rust
> fn single_comment_same_line() { /*this is a comment*/ }
> ```

This actually pushes it to the next line, but I don't think this is much of a concern, since it's not a breaking format.

JoshuaBatty (approved):

eureka-cpu (approved):
🚢

PR #4105, opened 2023-02-16
Support for configurable blocks in the formatter

### LINKED ISSUE #3679: Support `configurable` in the formatter
Ref: https://github.com/FuelLabs/sway/pull/3536

This would be fairly similar to how `storage` blocks are formatted.

mohammadfawaz:
## Description
Closes #3679

This is based on how `storage` blocks are handled.. Most of the code is copied from `item_storage`. We should probably consider making as much as that code common between the two because `storage` and `configurable` blocks look pretty much the same.

Making this change now so that the formatter does not panic when we add an example using `configurable` under `sway/examples`.

## Checklist

### DIFF
--- a/swayfmt/src/items/item_configurable/mod.rs
+++ b/swayfmt/src/items/item_configurable/mod.rs
@@ -10,13 +10,13 @@ use crate::{
     },
 };
 use std::fmt::Write;
-use sway_ast::{keywords::Token, token::Delimiter, ItemStorage, StorageField};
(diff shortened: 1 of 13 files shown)


mitchmindtree (approved):
Thanks Mo! Looks fine to me, we can consider refactoring some of the common logic in a follow-up PR.

kayagokalp (approved):
👌

PR #4106, opened 2023-02-16
Implements json abi with call paths for structs and enums.

esdrubal:
## Description

Reference to external dependency struct now looks like: `"type": "struct ::abi_with_tuples::Person",`

Note that the `::` is only present when accessing external dependencies this includes std:
`"type": "enum ::std::option::Option",`

Struct reference in root module still looks like:
`"type": "struct MyStruct",`

Struct reference in sub modules looks like:
`"type": "struct folder1::dep_1::MyStruct1",`

I decided to use `folder_name::lib_name::struct_name` instead of a module path because we can have multiple module paths to the same final module. When two deps `A` and `B` in the root dir both use a third dep `C`. The module path to `C` is `A::C` and `B::C` respectively while the path used now would be only `C` as `C` is also in the root dir. Overall using the paths with folder names removes ambiguous cases.

This is similar to how rust works where a folder can have a mod.rs with `pub mod file_name` and then the call paths to the sub modules (in files) of the folder will be `folder_name::file_name::type_name`.

This is related to:
https://github.com/FuelLabs/fuel-specs/issues/450
https://github.com/FuelLabs/fuels-rs/issues/791

## Checklist


esdrubal:
Closing because this was replaced by:https://github.com/FuelLabs/sway/pull/4137

Issue #4112, opened 2023-02-16
Add parsing of inner doc attributes for `Item`s


eureka-cpu:
With #4097 closed, the `//!` syntax is only considered a doc comment attribute if it is at the top of the file. If that syntax is used elsewhere it is considered just a normal comment to the parser.

mitchmindtree:
I'll add that we've intentionally omitted parsing inner doc attributes for `Item`s as we're not certain we actually want this behaviour. That said, it's good to have a tracking issue seeing as some users coming from Rust might expect the behaviour.

PR #4113, opened 2023-02-16
Limit inner doc comment attributes to `Module`

### LINKED ISSUE #4097: Limit inner doc comment attributes to `Module`
As of right now this:
```rust
struct Foo {
    //! this is a doc comment about Foo
    field: u32,
}
```
will collect the attribute into `field` and not `Foo`.
We can either skip collection for anything but the collection into `Module` using `//!` or finish implementing it for all `Item`s so that the above will collect the attribute into `Foo` as intended. Ref #4112 

This is a follow up to #4081 which managed to get merged before I could address this fully. The only trade off is that both inner and outer doc comment attributes are treated as outer for now.

eureka-cpu:
## Description
Ran on test branch & working as intended. Only beginning of file `//!` is collected as `DocComment` with `DocStyle::Inner` which is only parsed at the beginning of the `Module` parsing stage. This ensures that any `//!` found after the module docs will be treated as normal comments for now. Ref #4112 for tracking further progress in allowing inner doc comments for `Item`.

## Checklist

- [x] I have linked to any relevant issues. Closes #4097 

### DIFF
--- a/sway-parse/src/item/mod.rs
+++ b/sway-parse/src/item/mod.rs
@@ -206,9 +206,9 @@ mod tests {
         let item = parse_item(
             r#"
             // I will be ignored.
-            //! I can't feel the way I did before
+            //! I will be ignored.
             /// This is a doc comment.
-            //! Don't turn your back on me, I won't be ignored.
+            //! I will be ignored.
             // I will be ignored.
             fn f() -> bool {
                 false
@@ -218,17 +218,7 @@ mod tests {
         assert!(matches!(item.value, ItemKind::Fn(_)));
         assert_eq!(
             attributes(&item.attribute_list),
-            vec![
-                [(
-                    "doc-comment",
-                    Some(vec![" I can't feel the way I did before"])
-                )],
-                [("doc-comment", Some(vec![" This is a doc comment."]))],
-                [(
-                    "doc-comment",
-                    Some(vec![" Don't turn your back on me, I won't be ignored."])
-                )]
-            ]
+            vec![[("doc-comment", Some(vec![" This is a doc comment."]))]]
         );
     }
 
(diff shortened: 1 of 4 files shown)


sway-parse/src/item/mod.rs
```diff
@@ -206,9 +206,9 @@ mod tests {
         let item = parse_item(
             r#"
             // I will be ignored.
-            //! I can't feel the way I did before
+            //! I will be ignored.
```
eureka-cpu, line 209: Changing this back oddly makes me sad @mitchmindtree

mitchmindtree (approved):
Yeah nice, I think this conservative approach and enforcing some consistency like this is the right one :+1:  We can reconsider enabling for all items if someone comes across an actual good use-case to have them anywhere else.

sway-parse/src/token.rs
```diff
@@ -120,17 +120,34 @@ pub fn lex_commented(
...
 
     let mut parent_token_trees = Vec::new();
     let mut token_trees = Vec::new();
     while let Some((mut index, mut character)) = l.stream.next() {
         if character.is_whitespace() {
+            // if the beginning of a file starts with whitespace
+            // we must keep track to ensure that the module level docs
+            // will get inserted into the tree correctly
+            if index - file_start_offset == 0 {
+                file_start_offset += character.len_utf8();
```
eureka-cpu, line 134: hey @mitchmindtree just wanted to make sure this was the correct approach to managing the offset of whitespace at the beginning of a file

`.len_utf8()` returns the byte span in `usize`

JoshuaBatty (approved):

sway-parse/src/token.rs
mitchmindtree: Hmm I'm not sure I fully understand why we need to keep track of the whitespace at the beginning. How does the whitespace affect whether or not the module level docs are inserted into the tree correctly?

mitchmindtree: If in doubt, maybe we can just add more tests with weird whitespace gaps, e.g.

This one should probably be valid
```rust
// <comment or whitespace here>
//! some docs
```

Should this fail or merge with the top? Not sure what Rust does here:
```rust
//! some docs

//! more docs
```
Although it should *probably* merge, particularly if gapped by a normal comment, e.g. I guess the following should probably pass?
```rust
//! some docs
// //! some more commented out docs
//! some more docs

Comment (eureka-cpu) [reply sway-parse/src/token.rs:134]: Well the only way I could think of how to track module doc comments was to use the index. If the index - offset of whitespace is 0 then the bool `gather_module_docs` will be true so it will start to collect them as `DocStyle::Inner` `DocComment`s until the next char it finds is not a comment. It will continue to process through other normal comments just in case. So 
```rust
//! hello...
// hello
/* hi again */
//! world!
```
will produce two inner doc comments even though there are other comments in between.

eureka-cpu: It will work in all cases you provided, I believe.

PR #4116, opened 2023-02-17
feat: Add salt flag to `forc deploy`, maturity flag to `deploy` and `run`

mitchmindtree:
## Description

This allows for specifying a `--salt` for the top-level contract member when running `forc deploy`. Right now, we only enable this for *packages*, and not for workspaces with more than one package due to some ambiguities that arise (see comment note in code). I'll open an issue to track relaxing this constraint a little. Edit: opened https://github.com/FuelLabs/sway/issues/4117.

I noticed we also had no way of specifying maturity, so have added a `--maturity` flag for this as well.

These flags are shared with the `forc-tx` crate in order to try and provide some consistency (in help output and flag representation) across our CLI tooling.

Ideally we'd like to ship this feature in the Sway release that ends up in the fuelup beta-3 toolchain - otherwise users following the quick-start guide will be unable to re-deploy the same contract to the network as it will already exist - they'll need to be able to provide a salt.

## Checklist

### DIFF
--- a/forc-plugins/forc-client/src/op/deploy.rs
+++ b/forc-plugins/forc-client/src/op/deploy.rs
@@ -9,7 +9,7 @@ use anyhow::{bail, Context, Result};
 use forc_pkg::{self as pkg, PackageManifestFile};
 use fuel_core_client::client::types::TransactionStatus;
 use fuel_core_client::client::FuelClient;
-use fuel_tx::{Output, Salt, TransactionBuilder};
+use fuel_tx::{Output, TransactionBuilder};
 use fuel_vm::prelude::*;
 use futures::FutureExt;
 use pkg::BuiltPackage;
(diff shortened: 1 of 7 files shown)


otrho (approved):

JoshuaBatty (dismissed):

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -64,7 +84,7 @@ pub async fn deploy_pkg(
     let client = FuelClient::new(node_url)?;
 
     let bytecode = compiled.bytecode.clone().into();
-    let salt = Salt::new([0; 32]);
+    let salt = command.salt.salt.unwrap_or_default();
```
Voxelot, line 87: Would it make sense to support the ability to generate a random salt for the user? Would make things like QuickStart doc examples easier if users could just copy paste commands without needing to customize arguments etc

mitchmindtree: Yeah that sounds useful :+1: I'll add a `--random-salt` flag now unless there's any objections?

mitchmindtree:
Okydoke, rebased and added @Voxelot's suggestion for a `--random-salt` option!

kayagokalp (approved):
utACK ✅

Voxelot (approved):

PR #4118, opened 2023-02-17
create workspace for examples folder

mehtavishwa30:
## Description

The example folder currently has several smart contracts that serve as reference material for new devs. A common `fuel-toolchain.toml` file that globally overrides user's local dependencies would help devs to try out these examples without worrying about version mismatches or upgrades. 

This enhancement add the examples under a workspace to be enable a single `fuel-toolchain.toml` file. It covers the following changes:

1. Created a workspace for the examples folder
2. Added a `fuel-toolchain.toml` file that automatically overrides local installations of the toolchain dependencies.


mitchmindtree:
Ahhhh, while I think the workspace is a great addition (nice!!), I'm not that we also to add a `fuel-toolchain.toml` to the directory as well :thinking:

This is because we use these examples as tests against the local copies of `forc`, `forc-fmt`, etc. It's important that these `examples` stay up-to-date with master as they're also referred to within the book.

It *might* be OK to add a `fuel-toolchain.toml` that purely specifies the `fuel-core` component (and a `channel` or any of `forc` components), but we might need to keep an eye out for this in our downstream `fuelup` testing (which pulls in this repo to run tests).
cc @bingcicle for thoughts on this.

Side-note: we could also add a `beta-3` git tag to the commit that we settle on for the beta 3 release. This way during tutorials and workshops we could safely clone the repo, checkout the `beta-3` git tag and reliably use these examples as a starting point for a beta-3 compatible demo app.

mehtavishwa30:
Ahh, I see your point. My thought here was to make sure the examples, at latest the ones that are referenced in the Sway book make use of a `fuel-toolchain.toml` file so that new devs who are trying them out don't have to worry about version mismatches or upgrading their local dependencies. Is there a solution that could just serve this then?

bingcicle:
Yeah I'd +1 on the thought that having `fuel-toolchain.toml` may not be what we want here for 2 reasons:

1) As @mitchmindtree mentioned the examples are being tested against `master` - there might be cases where changes to `master` has 'broken' a previous release's behavior and cause the examples (on master) to not be compatible with a component (on the declared version within `fuel-toolchain.toml`). It would also add 1 more component to look out for during releases.

2) Semantically, if I browse the `examples` folder of a project on master, I'd expect the examples to be up-to-date with the `master` build, and not some other tag IMO.

bingcicle:
@mehtavishwa30 I think the problem is kinda half solved by us having a book at the tag,  eg. going to https://fuellabs.github.io/sway/v0.33.0/book/examples/fizzbuzz.html to see `fizzbuzz` on `forc v0.33.0`, but there is indeed a lack of a solution where we serve this example directly to devs wanting to try the examples out. :thinking: 

I think it could be great value add especially for our testnet-specific toolchains, eg. `beta-3`, but I'm not too sure on the approach we should take here to make such a resource available and convenient for devs.

mitchmindtree:
> My thought here was to make sure the examples, at latest the ones that are referenced in the Sway book make use of a fuel-toolchain.toml file so that new devs who are trying them out don't have to worry about version mismatches or upgrading their local dependencies. Is there a solution that could just serve this then?

If the user has cloned the examples from this repo, ideally they'll try running the examples using the `forc` tools that are also a part of this repo (unfortunately we can't stop them from trying to use a different versions), however I think the big unknown is whether those examples will work with whatever version of `fuel-core` they wish to deploy to.

### Targeting a beta fuel-core

We could add a `beta-3` git tag to the commit that we settle on for beta 3 to the Sway repo. This would allow users to checkout that git tag before trying to run the examples to have a higher chance that they're compatible with the beta-3 network.

### Running a local node

For users targeting a local `fuel-core`, the only way to be sure of compatibility of the `examples` is to use the same version of `fuel-core` that our CI tests against, which is conveniently buried [here in our ci.yml](https://github.com/FuelLabs/sway/blob/729f8b3bf8bb10d0e887946909515184bfd88fee/.github/workflows/ci.yml#L239) :sweat_smile: 

Perhaps in the mean-time it could be worth adding a README to the `examples/` directory that explains some of this?

mehtavishwa30:
> Yeah I'd +1 on the thought that having fuel-toolchain.toml may not be what we want here for 2 reasons:
> 
> As @mitchmindtree mentioned the examples are being tested against master - there might be cases where changes to master has 'broken' a previous release's behavior and cause the examples (on master) to not be compatible with a component (on the declared version within fuel-toolchain.toml). It would also add 1 more component to look out for during releases.
> 
> Semantically, if I browse the examples folder of a project on master, I'd expect the examples to be up-to-date with the master build, and not some other tag IMO.

Thanks for your inputs! That makes a lot of sense, @bingcicle.

> We could add a beta-3 git tag to the commit that we settle on for beta 3 to the Sway repo.

I like the idea of the `beta-3` git tag on the final commit for the Sway repo @mitchmindtree. FWIW, a substantial difference between an approach like that is still an additional check for new users as opposed to using `fuel-toolchain.toml` where _**things just work**._ When I was starting off, a notable pain-point was to check for outdated components and upgrading that created another barrier to a smooth devex.

Let me think this through with the devrel team and propose a solution that could potentially add value to new users, (especially around testnet-specific toolchains) without running compatibility issues while testing the examples against the `master` branch. :)

mehtavishwa30:
Updated the PR by removing the `fuel-toolchain.toml` file for now until further discussion with the team regarding the need for a solution to have channel specific dependencies for the workspace using the fuel-toolchain.toml

kayagokalp (approved):
thanks for adding this 🚀

mohammadfawaz (approved):

PR #4119, opened 2023-02-17
Check for storage types in impl method params

### LINKED ISSUE #4040: `StorageVec` not disallowed as argument to methods and trait declarations.
Normally, when trying to pass a struct like `StorageVec` to a public function we expect an error informing the user that it may only be declared as a storage field. An example function of this works properly is:

```rust
#[storage(read, write)]
pub fn takes_storage_struct(my_struct: StorageVec<u64>) {
    my_struct.push(5);
}
```
The code above produces the following error:
`Type StorageVec<u64> can only be declared directly as a storage field`

However, if attempting to do the same in an impl block for another struct and passing `self` as an argument, we get an internal compiler error. For example:

```rust
pub struct MyStruct {}

impl MyStruct {
    #[storage(read, write)]
    pub fn takes_storage_struct(self, my_struct: StorageVec<u64>) {
        my_struct.push(5);
    }
}
```
This code above produces the following internal compiler error:
internal_compiler_error

It should be noted that removing the `self` argument of `takes_storage_struct` will not produce any errors and compile. For example:

```rust pub struct MyStruct {}

impl MyStruct {
    #[storage(read, write)]
    pub fn takes_storage_struct(my_struct: StorageVec<u64>) {
        my_struct.push(5);
    }
}
```

This was tested on forc v0.35.0

vaivaswatha:
## Description
Fixes #4040 

There is an exception for `self` parameters, to avoid the error being thrown for impl of the actual storage types (in storage.sw). I'm not sure if that's the best way to add that exception. Reviewers please advice. [Edit: This is fixed, thanks @anton-trunov ]

## Checklist

### DIFF
--- a/sway-core/src/semantic_analysis/storage_only_types.rs
+++ b/sway-core/src/semantic_analysis/storage_only_types.rs
@@ -268,7 +268,7 @@ fn decl_validate(engines: Engines<'_>, decl: &ty::TyDeclaration) -> CompileResul
                             errors
                         );
                         for param in method.parameters {
-                            if param.name.as_str() != "self" {
+                            if !param.is_self() {
                                 check!(
                                     check_type(
                                         engines,
--- a/sway-core/src/semantic_analysis/storage_only_types.rs
+++ b/sway-core/src/semantic_analysis/storage_only_types.rs
@@ -267,6 +267,21 @@ fn decl_validate(engines: Engines<'_>, decl: &ty::TyDeclaration) -> CompileResul
                             warnings,
                             errors
                         );
+                        for param in method.parameters {
+                            if param.name.as_str() != "self" {
+                                check!(
+                                    check_type(
+                                        engines,
+                                        param.type_argument.type_id,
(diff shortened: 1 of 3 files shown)


sway-core/src/semantic_analysis/storage_only_types.rs
```diff
@@ -267,6 +267,21 @@ fn decl_validate(engines: Engines<'_>, decl: &ty::TyDeclaration) -> CompileResul
                             warnings,
                             errors
                         );
+                        for param in method.parameters {
+                            if param.name.as_str() != "self" {
```
anton-trunov, line 271: Maybe something like this?

```suggestion
                            if !param.is_self() {
```

vaivaswatha: Thanks @anton-trunov , that works. I didn't know that the method existed.

mohammadfawaz (approved):
Thanks!

anton-trunov (approved):

PR #4121, opened 2023-02-17
Add Google Analytics ID to book.toml

mehtavishwa30:
## Description

mdBook lets you track traffic on your generated website or docs simply by adding the tracking ID to the `book.toml` file.

This PR adds the Google Analytics tracking ID to `book.toml` for Sway Book.


mehtavishwa30:
Converted to draft. The tracking ID needs to be replaced.

docs/book/book.toml
```diff
@@ -4,6 +4,7 @@ language = "en"
 multilingual = false
 src = "src"
 title = "The Sway Programming Language"
+google_analytics_id: "G-NEWVJ5V9T7"
```
kayagokalp, line 7:
```suggestion
google_analytics_id= "G-NEWVJ5V9T7"
```

kayagokalp (changes requested):
TOML format does not allow for initialization with `:` I think that is why the CI is failing added a suggestion changing it with `=`

I am not familiar with any google analytics stuff, should this ID be kept as a secret or is it fine to add it like this?

docs/book/book.toml
mehtavishwa30: Ahh, you're right!! thanks for the catch. Committing the suggestion.

mehtavishwa30:
Re: the tracking ID, it would be a good practice to be kept as a secret because it's essentially the tag that is used by Google to identify a page as a source for their analytics. Is it okay to .gitignore book.toml?

kayagokalp:
@mehtavishwa30 Hmm about adding book.toml to gitignore, I think that wouldn't be something we would prefer as we need that file to build the book 🤔. Also it is important to point out that the book is built and published by our CI. So if you would like to keep the id as a secret we may need to enter that as a secret to our repo and find a way to inject it into book.toml.

The thing is this might not be very easy to achieve. I guess we can we can add fetching from env variable step to our mdbook preprocessor. But it is an open question if it would allow us to inject stuff into book.toml. I see that [there](https://lib.rs/crates/mdbook-variables) is a preprocessor that enables usage of env variable in markdown files but again not sure if it would allow us to inject into book.toml, i am guessing it would not allow it.

I guess we could have a CI step that runs script which rewrites that toml file with the repo secret before building the book.

Finally if we need to keep it as a secret, I think once we have a solution to this we may need to regenerate it as the id is exposed in this PR history.

cc @bingcicle

bingcicle:
From a quick search on github it seems like many projects expose their ID so i'm assuming it's fine(?) but i'm not an expert on this. And yes, the book.toml has to be checked in for building the book.

kayagokalp:
> From a quick search on github it seems like many projects expose their ID so i'm assuming it's fine

Oh that is good to hear than let's get this in if it is ok with you too @mehtavishwa30

kayagokalp (approved):

mehtavishwa30:
@kayagokalp @bingcicle thanks for the clarity. I think we should be good to merge this in that case. :)

mohammadfawaz (approved):

PR #3986, opened 2023-02-05
fix: prevent duplicate sway unit test execution

kayagokalp:
## Description
This PR adds a step to execution of unit tests so that duplicate unit tests are omitted. Duplicate tests are generated from dependencies declared with `dep` keyword in a duplicate manner (see #3947). Until that is fixed this PR de-duplicates sway unit tests based on their `span`. Once the fix for #3947 is ready we can remove the changes from this PR. This is also commented in the code.

closes #3813.

## Checklist

- [x] I have linked to any relevant issues.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation where relevant (API docs, the reference, and the Sway book).
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have added (or requested a maintainer to add) the necessary `Breaking*` or `New Feature` labels where relevant.
- [x] I have done my best to ensure that my PR adheres to [the Fuel Labs Code Review Standards](https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/external-contributors.md).
- [x] I have requested a review from the relevant team or maintainers.

Issue #3813, opened 2023-01-18
Sway unit tests are run multiple times


nfurfaro:
In my current project, some(but not all) of the unit tests run twice.
One thing to note with my project (a library) is that the tests are currently spread across multiple files/libraries.
I haven't managed to create a minimal repro yet, but here is a link to my branch: https://github.com/FuelLabs/sway-applications/tree/furnic/add-abi-and-contract/chess-game 

Here is a sample of the output. Note that there are actually only 5 unique tests being run here:
```
Bytecode size is 1700 bytes.
   Running 16 tests

   tested -- lib_chess

      test test_new_bitstack ... ok (229µs)
      test test_new_bitstack ... ok (99.875µs)
      test test_to_u8 ... ok (321.125µs)
      test test_to_u8 ... ok (321.25µs)
      test test_multimask ... ok (2.139291ms)
      test test_new_board ... ok (164.291µs)
      test test_increment_half_move_counter ... ok (211.958µs)
      test test_to_u8 ... ok (352.75µs)
      test test_new_bitstack ... ok (215.166µs)
      test test_to_u8 ... ok (347.708µs)
      test test_to_u8 ... ok (338µs)
      test test_multimask ... ok (2.069458ms)
      test test_new_board ... ok (138.083µs)
      test test_increment_half_move_counter ... ok (156.333µs)
      test test_to_u8 ... ok (312.375µs)
      test test_to_u8 ... ok (310.208µs)
   Result: OK. 16 passed. 0 failed. Finished in 7.726871ms.
```

 cc @kayagokalp

kayagokalp:
Hey @nfurfaro thanks for the issue. I am trying to debug this but having some troubles. I think your repo does not build with latest(`0.33`) `forc`, right? Do you have the version of `forc` you are using?

nfurfaro:
@kayagokalp Hmm, it does for me?
<img width="731" alt="image" src="https://user-images.githubusercontent.com/17053076/213578564-23bf192a-aa2a-4e97-9715-a5a54e7c299f.png">

<img width="1088" alt="image" src="https://user-images.githubusercontent.com/17053076/213578624-b3f0639d-7921-4298-9fcc-af25a8a3dd2c.png">

mitchmindtree:
Ahh, I wonder if this has something to do with picking up tests not only in the top-level project, but in dependencies too? I thought I'd addressed this in the original implementation, but perhaps not.

nfurfaro:
[...]

PR #4127, opened 2023-02-19
feat: disallow duplicate submodule declaration

### LINKED ISSUE #3947: Duplicate dep declarations should generate a compile error
If I declare something like
```
dep pkg_b;
dep pkg_c;
```
in `pkg_a.sw` , and
```
dep pkg_c;
```
in `pkg_b.sw`
entry functions that are declared in `pkg_c` appears to be listed as entry functions under `pkg_a` 2 times (one coming from 
  `pkg_b->pkg_c` other directly from `pkg_c` `dep` declaration). Rather than creating duplicate entries, @sezna mentioned that we should not allow this at the first place and generate a compile error.

### LINKED ISSUE #3988: Remove changes introduced at #3986 once #3947 is addressed
To fix duplicate sway unit tests filtering is added with #3986. This won't be necessary and should be removed once #3947 is fixed.

kayagokalp:
## Description

This PR is heavily WIP, I am opening to discuss both the problem and solution. 
closes #3947.
closes #3988.

## TODO

- [ ] Generate an actual compilation error
- [ ] Remove filters added to `forc-test`
- [ ] Fix this situation in standard library
- [ ] Tests
- [ ] Documentation

## Problem

Duplicate `dep` declarations lead to duplicate entry points in the `FinalizedProgram` with different `imm` values and same `Span`. This was causing test entries to get duplicated. I guess this wasn't a problem until addition of unit testing, as other than test functions there is only the main function as an entry point(?). Also duplicate main declaration is something already checked. 

So basically duplicate submodule declarations (through dependencies) are causing tests to be duplicated in the top level module. An example can be seen below:

```mermaid
graph TD;
    A-->B;
    A-->C;
    B-->C;
```
A, B and C are separate modules where in with A we have following declarations:

```rust
dep B;
dep C;
```

and in B there is:

```rust
dep C;
```

In a scenario like this entry points of `C` are duplicated in `A`. So any tests that are placed in `C`, is duplicated in the top most level `A`. This caused #3813. To solve that I added a hot-fix (#3986  and #4069) but in a small discussion over slack, sezna mentioned that this is something that should be a compiler error.

## Solution

I constructed a graph from `dep` declarations. Such as the example given in the problem section and checked if any given node has more than 1 incoming edge. If there is one I am marking this `dep` declarations invalid and (will be) producing a compile error.

edit: I realized I need to add one more check as only checking incoming edge number also forbids the following pattern (which is perfectly fine):

```mermaid
graph TD;
    A-->C;
    B-->C;
```
I wonder if converting the graph into undirected graph and check for cycles is enough

## Checklist

### DIFF
--- a/sway-core/src/submodules/dependency_graph.rs
+++ b/sway-core/src/submodules/dependency_graph.rs
@@ -1,0 +1,87 @@
+use std::collections::HashMap;
+
+use petgraph::{stable_graph::StableGraph, Directed};
+use sway_types::Span;
+
+use crate::language::{ty::TyProgram, DepName};
+
+type GraphIx = u32;
+type NodeIx = petgraph::graph::NodeIndex<GraphIx>;
+
+#[derive(Debug, Hash, PartialEq, Eq, Clone)]
+pub(crate) struct Node {
+    pub(crate) dep_decls: Vec<NodeIx>,
+    pub(crate) dep_name: DepName
+}
+
+pub(crate) type DepGraph = StableGraph<Node, (), Directed, GraphIx>;
+
+#[derive(Debug)]
+pub(crate) struct SubmodulePlan {
+    pub(crate) graph: DepGraph,
+}
+
+impl SubmodulePlan {
+    /// Constructs a new `SubmodulePlan` from given `TyProgram`.
+    pub(crate) fn from_ty_program(ty_program: &TyProgram) -> Self {
+        let root = &ty_program.root;
+        let mut graph = DepGraph::new();
+        let mut node_indices = HashMap::new();
+
+        // Add non-root nodes to the graph.
+        for (dep_name, submodule) in root.submodules_recursive() {
+            let dep_decls = &submodule.module.submodules;
+            let dep_decls: Vec<NodeIx> = dep_decls.iter().map(|(dep_decl, _)| node_indices[dep_decl]).collect();
+            let node = Node::new(dep_name.clone(), dep_decls);
+            node_indices.entry(dep_name).or_insert_with(|| {
+                graph.add_node(node.clone())
+            });
+        }
+
+        // Create the root node.
+        let root_node_name = DepName::new(Span::dummy());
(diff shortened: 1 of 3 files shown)


nfurfaro:
In the above example, if we had module re-exports, `B` could simply re-export (i.e: `pub use C`) `C` so `A` doesn't need to also import `C`. 
I don't know wif this actually solves the issue at its root, but would be nice to have it regardless.

sway-core/src/submodules/dependency_graph.rs
```diff
@@ -0,0 +1,87 @@
...
+use sway_types::Span;
+
+use crate::language::{ty::TyProgram, DepName};
+
+type GraphIx = u32;
+type NodeIx = petgraph::graph::NodeIndex<GraphIx>;
+
+#[derive(Debug, Hash, PartialEq, Eq, Clone)]
+pub(crate) struct Node {
+    pub(crate) dep_decls: Vec<NodeIx>,
```
mitchmindtree, line 13: If we're to go the graph route, we probably want to represent the `dep_decls` on the edges, rather than storing a separate list on the node.

mitchmindtree (commented):
One potentially simpler approach to this check could be to use a `BTreeMap<DepPath, ParentPath>` (rather than a full `Graph`), where each "path" is relative to the root. This way, you could detect duplicates during construction of the tree, as [`BTreeMap::insert`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.insert) returns `Some` if there was already an entry for the given key, or in other words, if there was already a parent for the given dependency.

mohammadfawaz:
Note that [this RFC](https://github.com/FuelLabs/sway-rfcs/blob/master/rfcs/0006-structured-modules.md) proposes a full rework of how modules work and removes the `dep` keyword altogether in favor of `mod`, similarly to Rust.

kayagokalp:
> Note that [this RFC](https://github.com/FuelLabs/sway-rfcs/blob/master/rfcs/0006-structured-modules.md) proposes a full rework of how modules work and removes the dep keyword altogether in favor of mod, similarly to Rust.

Really nice one, closing in favor of that as once that RFC is implemented, I guess this won't be needed.

PR #4129, opened 2023-02-20
feat(fmt): format comments supertraits

bingcicle:
## Description

Followup to #4005

makes `abi_supertraits` test pass comment formatting without `handle_comments`.

In fixing this, this also inadvertently led to me fixing other formatting issues:

- There may be comments before the module declaration - the first one was tagged as a `Trailing` comment, even though it should not be. Added a 0 check within lexing for this.
- Massively simplified `comments.rs` - before, we used to arbitrarily compare to the previous token, but now that we have the concept of `Trailing` and `Newlined` comments, we do not need to do this. We also used to write an indent after writing a trailing comment, but I think this should be moved up-one-level to whereever the trailing comment was written. There is also an update to a test that retains the trailing nature of a trailing comment in an existing test.

For the rest of changes, they seem largely unrelated to what this PR claims to solve, so I'll go change by change:
- within `item_fn/mod.rs`: This fixes the supertraits test example (which had a weird spacing for expected comment output)
- within `module/mod.rs`: We format comments between items (this fixes the the tests like `abi_comments` and `abi_supertrait` that have comments written between items.

## Checklist


bingcicle:
Ideally would like https://github.com/FuelLabs/sway/pull/4005 to go in first

bingcicle:
There might be a bug introduced with my previous PR to forc-fmt - one of the examples is badly formatted but `--check` isn't error-ing out. Investigating this.

bingcicle:
Opened #4147 to fix the issue with `forc-fmt --check` not failing. The abi_supertraits example by right should fail here, since there are extra whitespaces unformatted.

bingcicle:
I've fixed the example and ran `cargo run --bin=forc-fmt -- --check --path examples/` to ensure that the examples aren't broken - opened an issue $4152 to address the bug to do with `forc-fmt` not picking up the manifest within the examples workspace. This PR should be good to review on the formatting front though!

sway-parse/src/token.rs
```diff
@@ -159,7 +159,7 @@ pub fn lex_commented(
                         .count()
                         > 0;
 
-                    let comment_kind = if has_newline {
+                    let comment_kind = if has_newline || (search_end == 0 && index == 0) {
```
kayagokalp, line 162: Hmm not sure I got why we needed this. Is it for detecting comments at the beginning of the file as `newlined` or somehting else? If it is not something obvious that I am missing rn and you would like, maybe could use a single comment 👍

kayagokalp (commented):
Tests looks nice 👌, I was curious about a small change

swayfmt/src/comments.rs
```diff
@@ -57,6 +57,23 @@ fn write_trailing_comment(
     Ok(())
 }
 
+/// Writes an indentation if the given `predicate` evaluates to true.
+pub fn write_indent_if(
+    predicate: bool,
```
mitchmindtree, line 62: It seems a little odd to include an evaluated condition as the first argument: can we instead have this be `write_indent`, and then use `if` or `predicate.then(||...` to first check the condition before writing the indent?

mitchmindtree (commented):
Nice! Just one minor nit, plus I'll second Kaya's suggestion.

sway-parse/src/token.rs
bingcicle: >  Is it for detecting comments at the beginning of the file

Yes! Will add a comment to this

swayfmt/src/comments.rs
bingcicle: Actually, that was how I wrote [this](https://github.com/FuelLabs/sway/pull/4129/files#diff-43ff8fe4b8153ef64b5a5ee8c2b1cf50b5922a96de80d571ce74736ee3994153R41-R42) at first but it ended up taking a lot of vertical space:

```
 write_comments(formatted_code, end..self.value.span().start(), formatter)
                .and_then(|written|  {
                                    if written {
                                         write_indent(..)
                                    }
                                 }
```

I guess your suggested method does make it clearer :thinking:

mitchmindtree (approved):

kayagokalp (approved):

PR #4132, opened 2023-02-20
Add usage docs for Salt feature

mehtavishwa30:
## Description

Added usage docs for `--salt` and `--random-salt` flags


docs/book/src/introduction/sway_quickstart.md
```diff
@@ -202,6 +202,34 @@ Note the contract ID — you will need it if you want to build out a frontend to
...
+
+```bash
+cargo add forc
+```
+
+Alternately, you can add the following to your `cargo.toml` file:
+
+```toml
+forc = "0.35.2"
+```
```
kayagokalp, line 219: Hmm I have a couple of small concerns here:

If we follow the quickstart we can see that step by step we are describing creating a new project with `forc new` which creates a basic sway project that _does not_ contain a `Cargo.toml` by default. Executing `cargo add forc` would fail on such project as it is not a rust project that we can use cargo to add a dependency. 

Also I don't think we should ever depend on `forc` in a `Cargo.toml` as `forc` is a binary that you need to install. If you want to install `forc` via cargo, you can use `cargo install forc`. 

This opens up the next point: unless we have a reason not to, we should use `fuelup` for the installation process. For example installing newest `forc` with `cargo install forc` would not install `forc-client` which is the package that enables `forc deploy` and `forc run`. You have to manually install `forc-client` with `cargo install forc-client` to be able to use them. Installing via `fuelup` takes care of this automatically and installs both of them. So we may want to point them to update via `fuelup`. Or for some reason we need to point them to installation with cargo we need to mention `cargo install forc-client`.

docs/book/src/introduction/sway_quickstart.md
```diff
@@ -202,6 +202,34 @@ Note the contract ID — you will need it if you want to build out a frontend to
...
+
+Once `forc` is updated, can now add a `--salt` to your contractID while deploying the contract as follows:
```
kayagokalp, line 221: On a similar comment with above, updating `forc` alone would not suffice as what install `forc-deploy` is `forc-client`

kayagokalp (changes requested):
Thanks for adding this! Left some suggestions otherwise lgtm 👍

docs/book/src/introduction/sway_quickstart.md
mehtavishwa30: That's a very fair point. Taking both first and second comment into account, is it safe to skip any installation steps in this section all together and assuming that the reader has already installed the latest dependencies at the beginning of the quickstart during the installation step?

kayagokalp: I think it would be perfectly safe to assume that as we have an entire section for installation as the previous section. If you want to be extra cautious, you could add a note that mentions for to be able to use `forc deploy` or `forc run`  installing `forc-client` is needed.

mehtavishwa30: Awesome!! thanks for the notes. Will add final changes now.

docs/book/src/introduction/sway_quickstart.md
```diff
@@ -202,6 +202,26 @@ Note the contract ID — you will need it if you want to build out a frontend to
...
+
+### Randomised Salt
+
+Salt is 0x00000000000000000000000000000000 by default. If you do not want to pass a salt manually, you can also use the --random-salt flag to randomise your salt as follows:
+
+```bash
+forc deploy --random-salt --node-url <key> --gas-price 1
+```
+
+> Note: To be able to use the `forc deploy` and `forc run` commands, you require the `forc client` plugin that lets you interact with a Fuel node. You can read more about `forc client` [here](https://fuellabs.github.io/sway/v0.35.2/book/forc/plugins/forc_client/index.html).
```
mohammadfawaz, line 223: We should not point to a published version of the book. Instead, a relative path here is better.

docs/book/src/introduction/sway_quickstart.md
```diff
@@ -202,6 +202,26 @@ Note the contract ID — you will need it if you want to build out a frontend to
...
+
+To add a Salt to your ContractID while deployment, you can run the following command:
+
+```bash
+forc deploy --salt <Salt> --node-url <key> --gas-price 1
+```
+
+### Randomised Salt
+
+Salt is 0x00000000000000000000000000000000 by default. If you do not want to pass a salt manually, you can also use the --random-salt flag to randomise your salt as follows:
```
mohammadfawaz, line 217:
```suggestion
Salt is `0x0000000000000000000000000000000000000000000000000000000000000000` by default. If you do not want to pass a salt manually, you can also use the --random-salt flag to randomise your salt as follows:
```

docs/book/src/introduction/sway_quickstart.md
```diff
@@ -202,6 +202,26 @@ Note the contract ID — you will need it if you want to build out a frontend to
...
+
+```bash
+forc deploy --random-salt --node-url <key> --gas-price 1
```
mohammadfawaz, line 220: We seem to be deploying with `--unsigned` in this document and none of the other flag. Maybe we should follow that convention here as well, as in:
```
forc deploy --unsigned --random-salt
```
That way, we are _building_ on what's already been explained before.

docs/book/src/introduction/sway_quickstart.md
```diff
@@ -202,6 +202,26 @@ Note the contract ID — you will need it if you want to build out a frontend to
...
 If `forc-deploy` emits a `no such subcommand` error, try to install it with [fuelup](../introduction/installation.md) or [manually](https://github.com/FuelLabs/sway/releases).
 
+## Deploying the contract with Salt
+
+You can pass a Salt using the `--salt` flag in your `forc deploy` command to redeploy a contract that has been deployed already. This is useful for multiple deployments of the same contract.
+
+To add a Salt to your ContractID while deployment, you can run the following command:
+
+```bash
+forc deploy --salt <Salt> --node-url <key> --gas-price 1
```
mohammadfawaz, line 212: Same comment about the extra flag. Simply write:
```
forc deploy --unsigned --salt <Salt>
```

docs/book/src/introduction/sway_quickstart.md
```diff
@@ -202,6 +202,26 @@ Note the contract ID — you will need it if you want to build out a frontend to
 
 If `forc-deploy` emits a `no such subcommand` error, try to install it with [fuelup](../introduction/installation.md) or [manually](https://github.com/FuelLabs/sway/releases).
 
+## Deploying the contract with Salt
+
+You can pass a Salt using the `--salt` flag in your `forc deploy` command to redeploy a contract that has been deployed already. This is useful for multiple deployments of the same contract.
+
+To add a Salt to your ContractID while deployment, you can run the following command:
```
mohammadfawaz, line 209:
```suggestion
To add a "Salt" while deploying your contract, you can run the following command:
```

mohammadfawaz (changes requested):

kayagokalp:
I pushed the changes requested so far to keep the PR alive and it is ready to review again 👍

mohammadfawaz (approved):

mitchmindtree (approved):

kayagokalp:
Ahh looks like my previous change requested holds this, approving to remove it

kayagokalp (approved):

PR #4133, opened 2023-02-20
Fixes struct and enum call path comparison to use suffix.

### LINKED ISSUE #4125: Struct is being treated as two different types
Take a look at the following error (forc versions 0.35.1 and 0.35.2)

On the left lower portion of the screenshot is the `main.sw` and on the right is the `abi` deceleration. They use a struct called `ExecutionRange` which is defined in a 3rd file. From this 3rd file we import and use it in both the `abi` and `main.sw` files. The type is exactly the same; however, it is being treated as two different types.

Here is a sample repro:

### main.sw

```rust
contract;

dep interface;
dep data_structures;

use interface::MyContract;
use data_structures::MyStruct;

storage {
    a: StorageMap<u64, MyStruct> = StorageMap {}
}

    #[storage(read)]
    fn test_function() -> Option<MyStruct> {
        storage.a.get(1)
    }
}
```

### data_structures.sw

```rust
library data_structures;

pub struct MyStruct {}
```

### interface.sw

```rust
library interface;

dep data_structures;

use data_structures::MyStruct;

abi MyContract {
    #[storage(read)]
    fn test_function() -> Option<MyStruct>;
}
```

esdrubal:
## Description
The issue is that currently call paths of structs and enums are based on module path which can differ depending on the order of imports.

The solution was to revert to the old behavior of comparing only the type name contained in the suffix.

Closes #4125

## Checklist

### DIFF
--- a/sway-core/src/type_system/info.rs
+++ b/sway-core/src/type_system/info.rs
@@ -264,7 +264,7 @@ impl PartialEqWithEngines for TypeInfo {
                     type_parameters: r_type_parameters,
                 },
             ) => {
-                l_name == r_name
+                l_name.suffix == r_name.suffix
                     && l_variant_types.eq(r_variant_types, engines)
                     && l_type_parameters.eq(r_type_parameters, engines)
             }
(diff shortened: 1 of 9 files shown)


mohammadfawaz (approved):
Great! Thanks for the quick fix.

Issue #3576, opened 2022-12-12
Add support for parsing argument of an argument for attributes


kayagokalp:
To unblock https://github.com/FuelLabs/sway/issues/3544 we need to be able to somehow specify a revert code with `#[test(should_revert)]` annotations.  This requires some parser tweak and I have some possible options in my mind but it would be nice to discuss and get it in separately.

I think `should_revert` part of the `#[test(should_revert)]`  is called argument so basically I need a way to add an argument to an argument 😅 

### To specify a revert code of 0
```rust
#[test(should_revert(0))]
fn test_foo() {

}
```
### To specify any revert code (parser allows this now)
```rust
#[test(should_revert)]
fn test_foo() {

}
```

### To specify a revert code of 0
```rust
#[test(should_revert = "0")]
fn test_foo() {

}
```

### To specify any revert code (parser allows this now)
```rust
#[test(should_revert)]
fn test_foo() {

}
```

I think both options requires a little tweaking with the parser which shouldn't be a problem but curios about your thoughts cc @mitchmindtree @mohammadfawaz. 
I am leaning towards option 2 as Rust also has it, example:
```rust
#[cfg(target_os = "linux")]
```

mohammadfawaz:
What about going full Rust and having 
```
    #[should_revert(expected = "Revert(0)")]
```
which similar to 
![image](https://user-images.githubusercontent.com/59666792/207128958-09c7b3d4-214d-42ee-a4f0-c9db170b28d6.png)

mohammadfawaz:
> I will try that but I am assuming rust does that check.

I would imagine!

mitchmindtree:
I think I still prefer the original proposal of allowing for recursively "nesting" of attributes.

> What about going full Rust

Rust's support for both nesting *and* assignment always struck me as a little unnecessary, and it's not always clear which to choose when designing macro APIs in Rust. E.g.

```rust
#[test(should_panic(expected("Revert(0)")))]
```
```rust
#[cfg(target_os("linux"))]
```

I think these are just as readable and clear, and both require less syntax and keep the macro model a little simpler. 

> In that case we might want to add a check to make sure that#[should_panic] is not used without #[test].

I think this is a good argument against using a totally separate attribute for `should_revert`. If we require that the attribute is nested within a test attribute (as we currently do), it's easier to check it's used within a valid context, and it's easier for us to work with the whole set of test-related attributes within a single structure (rather than having to check separately for `test` and `should_revert`).

[...]

Issue #3544, opened 2022-12-08
Add a way to specify expected revert value to `forc-test`


kayagokalp:
Related to https://github.com/FuelLabs/sway/pull/3490#pullrequestreview-1209486732.

While implementing this we can also clean the two suggestions from:

1. https://github.com/FuelLabs/sway/pull/3490#discussion_r1042948459
2. https://github.com/FuelLabs/sway/pull/3490#discussion_r1042816789

kayagokalp:
Adding "blocked" as this requires modifying the parser too. I will open-up a separate issue for that.

PR #4134, opened 2023-02-20
Improve attribute parsing

AlicanC:


otrho (approved):
It's a tiny bit confusing having the parser `AttributeArg` and the transform `AttributeArg` with the same name, but I'm guessing there are many precedents before it...

mohammadfawaz (dismissed):

AlicanC:
This was marked ready for some reason. I got more to do, putting it back to draft.

sway-ast/src/attribute.rs
```diff
@@ -52,10 +52,26 @@ pub enum AttributeHashKind {
...
+        }
+    }
+}
+
 #[derive(Clone, Debug)]
 pub struct Attribute {
     pub name: Ident,
-    pub args: Option<Parens<Punctuated<Ident, CommaToken>>>,
+    pub args: Option<Parens<Punctuated<AttributeArg, CommaToken>>>,
 }
```
mitchmindtree, line 75: Perhaps we should consider making `Attribute` a tree, as we'll likely end up wanting more than just one or two depths of nesting. I'd imagine it might also simplify the implementation a little. Briefly mentioned [here](https://github.com/FuelLabs/sway/issues/3576#issuecomment-1352863930).

I'm also not so sure that supporting the `=` operator like Rust does is really necessary when we can just use nesting/args for these cases too (also touched-on in the aforementioned comment).

otrho: I was kinda thinking the same thing when looking at the code but then thought, nah this is fine.  But after looking at those examples in #3544 you linked, I agree.  And we don't really need both types.

AlicanC: AFAIK Rust doesn't even parse the args and `cfg(a b c)` is like "calling" `cfg` with a TokenStream containing `a b c`. (There's support for `cfg = a b c` but that's not what we are doing here.)

If we wanted to go full Rust it would be something like:
```rs
struct Attribute {
    path: Path, // Not a name, but a proper path
    tokens: TokenStream, // No args, but just raw tokens
}
```

We were already parsing these tokens as args so I just went with it, but parsing them as whole attributes with nesting takes us a lot further away from what Rust does.

For example this:
```rs
#[cfg(target = "evm", feature = "test")]
// ...
```

becomes this:
```rs
#[cfg(target("evm"), feature("test"))]
// ...
```

which makes args of `cfg` attributes and `target` and `feature` become paths which sounds weird.

mitchmindtree: > AFAIK Rust doesn't even parse the args

> If we wanted to go full Rust it would be something like:

> but parsing them as whole attributes with nesting takes us a lot further away from what Rust does.

My suggestion is an attempt to come up with something simpler than Rust's approach.

> For example this:
> 
> #[cfg(target = "evm", feature = "test")]
> // ...
> 
> becomes this:
> 
> #[cfg(target("evm"), feature("test"))]
> // ...

IMO they're equally as readable, but with the latter we get a simpler macro model.

> which makes args of cfg attributes and target and feature become paths which sounds weird.

Fwiw I'm not convinced paths are actually that useful for Rust macros either :sweat_smile:  I know they're most popularly used by clippy and rustfmt, but they could probably also just use nesting rather than complicate the macro model further too. All that said, even if we did have a tree of paths, I personally don't think `target` and `feature` being paths under the hood makes this example any stranger or less readable.

I decidedly didn't block this as I know our policy is to follow Rust by default unless we can come up with something unambiguously nicer. This is just one of those areas where I think Rust is a bit unnecessarily messy, but if you think the assignment syntax adds value I don't mind if we include it!

mohammadfawaz (approved):

PR #4142, opened 2023-02-20
Add flag to reverse errors

### LINKED ISSUE #3860: Add a flag to forc to reverse the order of printed errors
Requested by @nfurfaro in a recent sway sync. Since most of the time, lots of errors propagates from the first error (meaning that once the first error is fixed, all of them goes away) it would make sense to be able to revert the order so that users do not need to scroll.

gr00vytvniks:
closes #3860


gr00vytvniks:
hey @kayagokalp I made the changes you requested, I tried running `forc build` after adding some errors to `examples/counter` and I got this error.
```sh
error
  --> /Users/mitchmackert/Code/sway/sway-lib-core/src/ops.sw:11:9
   |
 9 | 
10 |     fn add(self, other: Self) -> Self {
11 |         __add(self, other)
   |         ^^^^^ Identifiers cannot begin with a double underscore, as that naming convention is reserved for compiler intrinsics.
12 |     }
13 | }
   |
____
```
would you mind giving me steps to test this correctly? I think I am looking at this the wrong way

kayagokalp:
hey @gr00vytvniks thanks for the PR, the error looks odd. Can you make rebase/merge master you shouldn't have any lock file is the counter folder as we recently changed that folder to be a workspace, [see](https://github.com/FuelLabs/sway/tree/master/examples/counter). This type of core compilation errors can be seen with master forc depending on a released version of std/core. Can you make sure that in `examples/counter/Forc.toml` you actually depend on local std. You should have something like this in the `Forc.toml`:

```toml
[dependencies]
std = { path = "../../sway-lib-std" }TOML
``` 

You should be able to test this with a template project by adding a couple of compile errors. Let me know if you are unsure/stuck 👍

gr00vytvniks:
> hey @gr00vytvniks thanks for the PR, the error looks odd. Can you make a rebase/merge master as you shouldn't have any lock file is the counter folder as we recently changed that folder to be a workspace, [see](https://github.com/FuelLabs/sway/tree/master/examples/counter). This type of core compilation errors can be seen with master forc depending on a released version of std/core. Can you make sure that in `examples/counter/Forc.toml` you actually depend on local std. You should have something like this in the `Forc.toml`:
> 
> ```toml
> [dependencies]
> std = { path = "../../sway-lib-std" }
> ```
> 
> You should be able to test this with a template project by adding a couple of compile errors. Let me know if you are unsure/stuck 👍

Hey @kayagokalp, The error came from a collision between an old fuelup binary and a cargo binary.

gr00vytvniks:
Hey @kayagokalp,  I added a debug build profile like the book instructed but it is giving me this warning.

```
 ~/Code/sway/examples/counter/ [gr00vytvniks/reverse-errors*] forc build
  WARNING! unused manifest key: build-profiles
  WARNING! unused manifest key: build-profiles
  WARNING! unused manifest key: build-profiles
  WARNING! unused manifest key: build-profiles
  WARNING! unused manifest key: build-profiles
  WARNING! unused manifest key: build-profiles
  Compiled library "core".
  Compiled library "std".
[forc-util/src/lib.rs:243] reverse_errors = false
warning
  --> /Users/mitchmackert/Code/sway/examples/counter/src/main.sw:19:9
   |
17 | 
18 |         storage.counter = value;
19 |         value
   |         ----- This cast, from integer type of width sixty four to integer type of width thirty two, will lose precision.
20 |     }
   |
____

warning
  --> /Users/mitchmackert/Code/sway/examples/counter/src/main.sw:26:9
   |
24 | 
25 |         storage.counter = incremented;
26 |         incremented
   |         ----------- This cast, from integer type of width sixty four to integer type of width eight, will lose precision.
27 |     }
28 | }
   |
____

warning
  --> /Users/mitchmackert/Code/sway/examples/counter/src/main.sw:12:5
   |
10 | 
11 | storage {
12 |     counter: u64 = 0,
   |     ------- This storage declaration is never accessed and can be removed.
13 | }
   |
____

error
  --> /Users/mitchmackert/Code/sway/examples/counter/src/main.sw:17:42
   |
15 | 
16 |     #[storage(write)]
17 |     fn initialize_counter(value: u64) -> u32 {
   |                                          ^^^ expected: u64 
found:    u32 
help:     The definition of this function must match the one in the ABI "TestContract" declaration.
18 |         storage.counter = value;
19 |         value
   |
____

error
  --> /Users/mitchmackert/Code/sway/examples/counter/src/main.sw:23:42
   |
21 | 
22 |     #[storage(read, write)]
23 |     fn increment_counter(amount: u64) -> u8 {
   |                                          ^^ expected: u64 
found:    u8 
help:     The definition of this function must match the one in the ABI "TestContract" declaration.
24 |         let incremented = storage.counter + amount;
25 |         storage.counter = incremented;
   |
____

  Aborting due to 2 errors.
Error: Failed to compile counter

```

gr00vytvniks:
I went ahead and just changed the default in `debug()` to `true` and this was the output :) 
```
error
  --> /Users/mitchmackert/Code/sway/examples/counter/src/main.sw:23:42
   |
21 | 
22 |     #[storage(read, write)]
23 |     fn increment_counter(amount: u64) -> u8 {
   |                                          ^^ expected: u64 
found:    u8 
help:     The definition of this function must match the one in the ABI "TestContract" declaration.
24 |         let incremented = storage.counter + amount;
25 |         storage.counter = incremented;
   |
____

error
  --> /Users/mitchmackert/Code/sway/examples/counter/src/main.sw:17:42
   |
15 | 
16 |     #[storage(write)]
17 |     fn initialize_counter(value: u64) -> u32 {
   |                                          ^^^ expected: u64 
found:    u32 
help:     The definition of this function must match the one in the ABI "TestContract" declaration.
18 |         storage.counter = value;
19 |         value
   |
____

  Aborting due to 2 errors.
Error: Failed to compile counter
```

examples/counter/Forc.lock
```diff
@@ -0,0 +1,14 @@
...
+
+[[package]]
+name = 'counter'
+source = 'root'
+dependencies = ['std']
+
+[[package]]
+name = 'std'
+source = 'path+from-root-D8BC155605882939'
+dependencies = ['core']
```
kayagokalp, line 14: We should not have individual lock files as examples folder is a workspace now. Did you manually add this file? How did it produced?

kayagokalp (changes requested):
We should not revert it in any build profile imo. This is not something everyone would be using it is more niche command that some people may end-up using. 

So we should not revert by default as it would be unexpected for the users. We should be able to use this feature with the flag `forc build --reverse-errors` etc.

examples/counter/Forc.lock
gr00vytvniks: I figured that by adding the `reverse_errors` field to the `BuildProfile` the idiomatic way for the user would be able to add it to their build profile instead of passing a flag, that way they can just run `forc build`. 
The lock file is being generated probably because I need to update.

gr00vytvniks: @kayagokalp

kayagokalp: Hey @gr00vytvniks, I am waiting you to remove the lock file to do a final review as it shouldn't be there. After removing it and building the example again, if it is generated again we might have a problem and we should deal with it before merging this one.

gr00vytvniks: Hey @kayagokalp I removed the lock file and it looks like I just had an outdated forc version. 👍

nfurfaro (dismissed):
Thanks!

forc-pkg/src/pkg.rs
```diff
@@ -1721,9 +1721,11 @@ pub fn compile(
         "produce `sway_core::BuildConfig`",
         sway_build_config(pkg.manifest_file.dir(), &entry_path, pkg.target, profile)?
     );
+    let terse_mode = build_profile.terse;
+    let reverse_errors = build_profile.reverse_errors;
     let terse_mode = profile.terse;
```
eureka-cpu, line 1726: `terse_mode` is defined twice here

forc-pkg/src/pkg.rs
```diff
@@ -2290,6 +2292,10 @@ pub fn build(
             engines,
         ) {
             Ok(o) => o,
+            Err(errs) => {
+                print_on_failure(profile.terse, &[], &errs, profile.reverse_errors);
+                bail!("Failed to compile {}", pkg.name);
+            }
             Err(errs) => return fail(&[], &errs),
```
eureka-cpu, line 2299: Looks like you're returning `Err` twice here

forc-util/src/lib.rs
```diff
@@ -227,13 +227,23 @@ pub fn print_warnings(
...
+    errors: &[CompileError],
+    reverse_errors: bool,
+) {
     let e_len = errors.len();
     let w_len = warnings.len();
 
     if !terse_mode {
-        warnings.iter().for_each(format_warning);
-        errors.iter().for_each(format_err);
+        if dbg!(reverse_errors) {
```
eureka-cpu, line 240: remove this `dbg!` macro if this is ready

eureka-cpu:
Looks like there's a problem with the formatting of the `examples/counter/Forc.toml` file. Also some of the `should_fail` e2e tests are failing incorrectly (not sure if it's related, may just need to re-run the job).

forc-pkg/src/manifest.rs
```diff
@@ -608,6 +609,7 @@ impl BuildProfile {
             include_tests: false,
             json_abi_with_callpaths: false,
             error_on_warnings: false,
+            reverse_errors: true,
```
kayagokalp, line 612:
```suggestion
            reverse_errors: false,
```
I think we should not reverse the errors by default even if we are using a debug profile. This is pretty specialized feature and users might be surprised if errors are reverted out of the box. Users can pass the flag if they want to have it explicitly.

Also this is why some/all of the should fail tests failing as we are looking for specific errors with them reversing the order breaks the tests.

kayagokalp (changes requested):
Hey @gr00vytvniks, looking good! Left a simple thing to address which may fix your CI issues as well (I suppose e2e-tests failing due to this change)

forc-pkg/src/manifest.rs
eureka-cpu: Does `forc build --reverse-errors` work, or does this only work when creating a custom build profile?

eureka-cpu: Willing to approve once this is resolved. I think it would be okay to keep this as a feature that you have to specify in a build profile, since it's likely that those who would use it would always want to use it. I think it would be redundant to have both, as it would defeat the purpose of having the option for custom build profiles. If there is a flag it may lead to people using it under a command alias, instead.

eureka-cpu: cc @nfurfaro thoughts?

nfurfaro: I've never used a build profile yet... a flag would be sufficient for me, but don't see any harm in supporting both.

kayagokalp (commented):
Hey @gr00vytvniks current changes looks perfect! I only have a single nit and a very simple addition request so that your new feature is useable through our CLI and lets merge this afterwards. 

To be able to do that we are just missing cli bindings for the new `reverse_error` bool, we could have a field under `Print` in `forc/src/cli/shared.rs` and it could look like this:

```rust
/// Options related to printing stages of compiler output.
#[derive(Args, Debug, Default)]
pub struct Print {
    /// Print the generated Sway AST (Abstract Syntax Tree).
    #[clap(long)]
    pub ast: bool,
    /// Print the computed Sway DCA graph.
    #[clap(long)]
    pub dca_graph: bool,
    /// Print the finalized ASM.
    ///
    /// This is the state of the ASM with registers allocated and optimisations applied.
    #[clap(long)]
    pub finalized_asm: bool,
    /// Print the generated ASM.
    ///
    /// This is the state of the ASM prior to performing register allocation and other ASM
    /// optimisations.
    #[clap(long)]
    pub intermediate_asm: bool,
    /// Print the generated Sway IR (Intermediate Representation).
    #[clap(long)]
    pub ir: bool,
    /// Output the time elapsed over each part of the compilation process.
    #[clap(long)]
    pub time_phases: bool,
    /// Output build errors and warnings in reverse order.
    #[clap(long)]
    pub reverse_order: bool,
}
```

and in `forc/src/ops/forc_build.rs` we need to add the cli flag which would make that place something like:

```rust
fn opts_from_cmd(cmd: BuildCommand) -> pkg::BuildOpts {
    pkg::BuildOpts {
        pkg: pkg::PkgOpts {
            path: cmd.build.pkg.path,
            offline: cmd.build.pkg.offline,
            terse: cmd.build.pkg.terse,
            locked: cmd.build.pkg.locked,
            output_directory: cmd.build.pkg.output_directory,
            json_abi_with_callpaths: cmd.build.pkg.json_abi_with_callpaths,
        },
        print: pkg::PrintOpts {
            ast: cmd.build.print.ast,
            dca_graph: cmd.build.print.dca_graph,
            finalized_asm: cmd.build.print.finalized_asm,
            intermediate_asm: cmd.build.print.intermediate_asm,
            ir: cmd.build.print.ir,
            reverse_order: cmd.build.print.ast
        },
        time_phases: cmd.build.print.time_phases,
        minify: pkg::MinifyOpts {
            json_abi: cmd.build.minify.json_abi,
            json_storage_slots: cmd.build.minify.json_storage_slots,
        },
        build_profile: cmd.build.profile.build_profile,
        release: cmd.build.profile.release,
        error_on_warnings: cmd.build.profile.error_on_warnings,
        binary_outfile: cmd.build.output.bin_file,
        debug_outfile: cmd.build.output.debug_file,
        build_target: cmd.build.build_target,
        tests: cmd.tests,
        member_filter: Default::default(),
    }
}
```

forc-util/src/lib.rs
```diff
@@ -227,13 +227,23 @@ pub fn print_warnings(
...
+    errors: &[CompileError],
+    reverse_errors: bool,
+) {
     let e_len = errors.len();
     let w_len = warnings.len();
 
     if !terse_mode {
-        warnings.iter().for_each(format_warning);
-        errors.iter().for_each(format_err);
+        if reverse_errors {
```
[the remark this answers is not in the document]
gr00vytvniks: Thanks for the review @kayagokalp! I'll get right on this :)

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -163,6 +163,7 @@ fn build_opts_from_cmd(cmd: &cmd::Deploy) -> pkg::BuildOpts {
             finalized_asm: cmd.print.finalized_asm,
             intermediate_asm: cmd.print.intermediate_asm,
             ir: cmd.print.ir,
+            reverse_order: cmd.print.ast,
```
kayagokalp, line 166: Hmm this looks like an error should we be using cmd.build.print.reverse_order?

forc-plugins/forc-client/src/op/run.rs
```diff
@@ -159,6 +159,7 @@ fn build_opts_from_cmd(cmd: &cmd::Run) -> pkg::BuildOpts {
             finalized_asm: cmd.print.finalized_asm,
             intermediate_asm: cmd.print.intermediate_asm,
             ir: cmd.print.ir,
+            reverse_order: cmd.print.ast,
```
kayagokalp, line 162: ^

forc/src/cli/commands/test.rs
```diff
@@ -168,6 +168,7 @@ fn opts_from_cmd(cmd: Command) -> forc_test::Opts {
             finalized_asm: cmd.build.print.finalized_asm,
             intermediate_asm: cmd.build.print.intermediate_asm,
             ir: cmd.build.print.ir,
+            reverse_order: cmd.build.print.ast,
```
kayagokalp, line 171: ^

forc/src/ops/forc_build.rs
```diff
@@ -24,6 +24,7 @@ fn opts_from_cmd(cmd: BuildCommand) -> pkg::BuildOpts {
             finalized_asm: cmd.build.print.finalized_asm,
             intermediate_asm: cmd.build.print.intermediate_asm,
             ir: cmd.build.print.ir,
+            reverse_order: cmd.build.print.ast,
```
kayagokalp, line 27: ^

kayagokalp (changes requested):
Thanks for the updates! I think we have some copy paste errors leaked here but otherwise lgtm 👍

forc/src/ops/forc_build.rs
gr00vytvniks: Gotcha! Must have slipped past me but I will fix right away :)

eureka-cpu:
Happy to approve this once merge conflicts are resolved and @kayagokalp gives his approval

eureka-cpu:
As a follow up PR, you could update the documentation so that developers will know how to use this feature. ie as part of a `BuildProfile` or just via `--reverse-order`(Not sure what name you guys landed on).

forc-pkg/src/manifest.rs
```diff
@@ -223,6 +223,9 @@ pub struct BuildProfile {
     pub json_abi_with_callpaths: bool,
     #[serde(default)]
     pub error_on_warnings: bool,
+    #[serde(default)]
+    pub reverse_results: bool,
+    pub experimental_private_modules: bool,
```
eureka-cpu, line 228: I think this field `experimental_private_modules` was removed at some point as it isn't on `master` anymore. I also think the `serde` annotation above `reverse_results` was intended for that field and can be removed as well. cc @kayagokalp

forc-pkg/src/manifest.rs
```diff
@@ -666,6 +669,8 @@ impl BuildProfile {
             include_tests: false,
             json_abi_with_callpaths: false,
             error_on_warnings: false,
+            experimental_private_modules: false,
```
eureka-cpu, line 672: It can be removed here as well

forc-pkg/src/manifest.rs
```diff
@@ -682,6 +687,8 @@ impl BuildProfile {
...
```
eureka-cpu, line 690: and here

eureka-cpu (changes requested):
Haven't tested yet but there's one thing that seems out of place if you wouldn't mind addressing it. After that I'll pull this branch and test it out!

forc-pkg/src/manifest.rs
gr00vytvniks: Ah I see. Fixing :)

forc-pkg/src/manifest.rs
```diff
@@ -223,6 +223,8 @@ pub struct BuildProfile {
     pub json_abi_with_callpaths: bool,
     #[serde(default)]
     pub error_on_warnings: bool,
+    #[serde(default)]
```
eureka-cpu, line 226: I don't think you need to have the `#[serde(default)]` annotation here. Here's the doc that explains what it does, though I think it was just leftover from the field you had removed here.

gr00vytvniks: Just got that done and committed, thank you for the patience :)

eureka-cpu (dismissed):
🚢

JoshuaBatty:
Thanks for this. just some last merge conflicts @gr00vytvniks

JoshuaBatty (approved):

eureka-cpu (approved):

kayagokalp (approved):

PR #4151, opened 2023-02-21
Implementing basic type aliases

### LINKED ISSUE #869: Implement type aliases
As described in FuelLabs/sway#1064, it would be nice to have type aliases using the `type` keyword [just like Rust](https://doc.rust-lang.org/reference/items/type-aliases.html).

mohammadfawaz:
Closes #869 

### Description
This introduces basic type aliases without generics. For example
```rust
type AssetId = ContractId;

let id = AssetId::from(ZERO_B256);
```
I tried to chase down all instances where aliases are meant to work seamlessly and added examples for all these cases. There may be others. I have a list of raw tasks below but the summary of the change is as follows:
- Lexer + Parser changes
- Type System: new `TypeInfo::Alias` that coerces with the type it encapsulates.
- Semantic Analysis: new `Declaration` called `TypeAliasDeclaration` and also a typed version of it. Used in the declaratin engine. A lot of work here to make sure that the alias is resolved when used in various situations such as scrutinees, struct field accesses, struct and enum expressions, `impl` blocks, etc.
- DCA work is skipped for now. We just don't emit any warnings for unused type aliases right now. Doing this properly is non-trivial. (not hard, but not trivial).
- Formatter
- LSP
- Handling in the JSON ABI. For now, all aliases are completely resolved to their original types before inserting in the JSON ABI. If we don't do this, then we would need spec changes, particularly for things like aliases of tuples or integers, etc.
- Documentation + example
- `std` lib changes to add `AssetId` as an alias for `ContractId`

### Raw tasks
- [x] Aliases in `match` statements on enums
- [x] Array alias indexing
- [x] Tuple alias indexing
- [x] `impl` for aliases
- [ ] ~~DCA~~ - deferred
- [x] LSP
- [x] JSON ABI types
- [x] Formatter
- [x] Error out on aliases of generic types + tests
- [x] Error out on aliases of non-existing type when the aliases are not used.
- [x] Documentation
- [x] Correct handling of the function selector

### Checklist


sway-ast/src/keywords.rs
```diff
@@ -81,6 +81,7 @@ define_keyword!(FalseToken, "false");
 define_keyword!(BreakToken, "break");
 define_keyword!(ContinueToken, "continue");
 define_keyword!(ConfigurableToken, "configurable");
+define_keyword!(TypeToken, "type");
```
eureka-cpu, line 84: 🚀

otrho (dismissed):
All looks very thorough, and as always there's a lot of tests, but I can't really comment on the code itself.   But if it's passing all those tests, good enough for me.

sway-lib-std/src/contract_id.sw
```diff
@@ -25,3 +25,7 @@ impl From<b256> for ContractId {
         self.value
     }
 }
+
+/// The `AssetId` type is simply an alias for `ContractId` that represents the ID of a native asset
+/// which matches the ID of the contract that implements that asset.
+pub type AssetId = ContractId;
```
mitchmindtree, line 31: :ok_hand: Noice

mitchmindtree (dismissed):

otrho (approved):

Issue #4155, opened 2023-02-22
Contract dependency id injection should be determined with tests disabled


kayagokalp:
Currently we are determining contract id for injecting them with our `contract-dependencies` feature with the same build profile as the root package. This means that if we are building with tests enabled, our contract dependencies will also compiled with tests enabled. Since enabling tests alters the bytecode of the contract, this will yield a different contract id than the one while deploying them. This shouldn't be allowed an ideally we should build contract-dependencies with tests disabled no mather the passed flags to `forc build` command. 

Adding high as this is kind of related/blocking #3571.

PR #4156, opened 2023-02-22
feat: multi contract calls in unit tests for contracts

### LINKED ISSUE #3571: Add support for multi-contract calls to unit testing
On slack it was mentioned that unit testing would be limited to calling into the current contract, however this doesn't need to be the case.

It should be possible to call into *all* contracts that were declared under `[contract-dependencies]` whether directly or transitively. We have their src, and can build and deploy them just like we can build and deploy a local contract.

This would make for a good followup to #3262 cc @kayagokalp. Rather than only deploying the project contract, we'd deploy all contract dependencies that exist within the graph.

We will still be limited in the case that a contract attempts to call into another contract that is *not* declared under `[contract-dependencies]`, e.g. if a contract is constructed from a raw contract ID within Sway code.

This is something we should communicate in the `[contract-dependencies]` docs and under the Sway contract docs in general - to encourage users to always declare their contract dependencies under `[contract-dependencies]` *if possible* as it gives `forc` the power to do cool stuff like this.

It *might* be possible to support private contracts (where the src or salt is not known) by extending the `[contract-dependencies]` feature a little. I'll open a new issue with some thoughts. Edit: see #3572.

kayagokalp:
## Description
closes #3571.

This PR adds the ability of calling multiple contracts from sway unit tests if they are added as `[contract-dependencies]`. This is limited with contracts currently but I will be having a follow-up which builds upon this to introduce this support to scripts as well. 

As these contracts are already declared under `[contract-dependencies]` their contract ids are injected into their namespace by `forc-pkg`. A bug related to this step is fixed in #4159. 

image

### Follow-ups
- #4161
- ~#4162~
 
## Checklist


kayagokalp:
I am planning to move [d63fe03](https://github.com/FuelLabs/sway/pull/4156/commits/d63fe039d35e143bae27dcf6774078e5e11017ae) into another PR which will close #4155

kayagokalp:
This is ready for review. Unfortunately the shadowing of consts with the same name that i mentioned in the last sync call is blocking us to use `contract_dependency::CONTRACT_ID` and for that I added a manual id printing so that until we solve that issue this is still useable.

To give a little bit of more context: `CONTRACT_ID` injected into the current contract's namespace is taking over the `CONTRACT_ID` injected into contract dependency's namespace even though I am trying to access explicitly (`contract_dependency::CONTRACT_ID`). Basically `contract_dependency::CONTRACT_ID` returns the value of `CONTRACT_ID` and they are not the same as the first one is the id of the dependency second one is the id of the current contract.

forc-test/src/lib.rs
```diff
@@ -353,16 +451,41 @@ impl BuiltTests {
...
+                .map(|contract_depency_node_ix| graph[contract_depency_node_ix].clone())
+                .collect();
+            let contract_dependencies = pinned_contract_dependencies
+                .iter()
+                .filter_map(|pinned| built_members.remove(pinned))
+                .collect();
+
+            (pinned_member, contract_dependencies)
+        })
+        .collect();
```
mitchmindtree, line 475: It looks like the produced `member_contract_dependencies` might not always contain all dependencies in the case that a member contract is depended on by two other contracts, as it will be removed from the `built_members` set the first time it appears in a member's `pinned_contract_dependencies`?

Also, it looks like this `clone`s all the built members at the start, which is quite a heavy operation (i.e. all bytecode, etc). Ideally we shouldn't have to clone built packages at all, and can reference the original members via keys or similar.

It looks like we might benefit from a refactor between this `build` constructor and the `BuiltTests::from_built` constructor to achieve what we want much more efficiently (and without missing contract deps for some members).

mitchmindtree (changes requested):
Nice work! I've left a couple suggestions I think might be worth addressing, but in general it looks like we're on the right track!

W.r.t. contract ID shadowing, thinking more on this, I think it's probably worth solving before attempting to land this. Could you possibly share a link to your attempt at adding the contract dependencies to the namespace so I can get a better idea of what might be going on with the shadowing issues? Let me know if you'd like to chat it out over a call if that's easier!

forc-pkg/src/pkg.rs
```diff
@@ -665,6 +671,22 @@ impl BuildPlan {
...
+            .iter()
+            .cloned()
+            .filter(move |&n| n != node)
+            .filter(|&n| {
+                graph
+                    .edges_directed(n, Direction::Incoming)
+                    .any(|edge| matches!(edge.weight().kind, DepKind::Contract { .. }))
+            })
+            .filter(move |&n| has_path_connecting(graph, node, n, None))
+    }
```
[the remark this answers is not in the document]
kayagokalp: Makes a lot of sense! Thanks 👍

forc-test/src/lib.rs
kayagokalp: Thanks for pointing it, I wasn't happy how this turned out but couldn't found a way to avoid clones just by refactoring `forc-test`. To solve it I made a small refactor in `pkg::Built`.

While trying to avoid cloning I realized the way we store `Built::Package` (which was a `Box<BuiltPackage>`) and `Built::Workspace` (which was a `HashMap<String, BuiltPackage>`) made this very hard to deal with.  Also storing both the name and built package felt wrong as built package already contains its pinned counterpart.

In a870681e43a48869e58f82a32b252b27fa12e33f i refactored them to hold `Arc<_>` and `Vec<Arc<_>>` and by doing that I am avoiding entire clones of `BuiltPackage` struct. Also refactored `inner_member()` just return an iterator so that the caller can decide if they want to collect it or not. 

After this change I tested with my simple workspace test containing couple of contracts and although I wasn't expecting it to be the effect was visible 😄 it is about ~%10 faster in the total time whole test operation takes. I think the effect is visible since std and core is considerably big so cloning them was taking some time.

docs/book/src/testing/unit-testing.md
```diff
@@ -92,3 +92,37 @@ fn test_fail() {
 ```
 
 > **Note:** When running `forc test`, your contract will be built twice: first *without* unit tests in order to determine the contract's ID, then a second time *with* unit tests with the `CONTRACT_ID` provided to their namespace. This `CONTRACT_ID` can be used with the `abi` cast to enable contract calls within unit tests.
+
+Contracts can call other contract's functions if they are declared unedr [`contract-dependencies`](../forc/manifest_reference.md#the-contract-dependencies-section) table. An example for such calls can be seen below:
```
mohammadfawaz, line 96:
```suggestion
Unit tests can call methods of external contracts if those contracts are added as contract dependencies, i.e. in the the [`contract-dependencies`](../forc/manifest_reference.md#the-contract-dependencies-section) section of the manifest file. An example of such calls is shown below:
```

docs/book/src/testing/unit-testing.md
```diff
@@ -92,3 +92,37 @@ fn test_fail() {
...
+
+```sway
```
mohammadfawaz, line 98: I'd love for these examples to be moved into the `examples` directory if possible.

docs/book/src/testing/unit-testing.md
```diff
@@ -92,3 +92,37 @@ fn test_fail() {
...
+
+  let should_be_true  = caller.test_true {}();
+  let should_be_false = caller2.test_false {}();
+
+  assert(should_be_true == true);
+  assert(should_be_false == false);
+}
+```
+
+> **Note:** In order for this example to work, the package hosting this contract must declare the package implementing the `MyContract2` ABI as a contract dependency.
```
mohammadfawaz, line 128: Instead of this note, can you show what the `[contract-dependencies]` section of the manifest file looks like? Again, preferable by _including_ the manifest from the example and using anchors, instead of copying it here.

forc-pkg/src/pkg.rs
```diff
@@ -2039,7 +2065,7 @@ pub fn build_with_options(build_options: BuildOpts) -> Result<Built> {
     let outputs = member_filter.filter_outputs(&build_plan, outputs);
 
     // Build it!
-    let mut built_workspace = HashMap::new();
+    let mut built_workspace = Vec::new();
```
mitchmindtree, line 2068: Nice, the `Vec` is probably a better fit here as we get to retain order, deduplicate the pkg name, and keying by pkg name probably isn't useful anyways.

forc-test/src/lib.rs
```diff
@@ -115,24 +117,117 @@ pub struct TestPrintOpts {
...
 
 impl BuiltTests {
     /// Constructs a `PackageTests` from `Built`.
     ///
-    /// Contracts are already compiled once without tests included to do `CONTRACT_ID` injection. `built_contracts` map holds already compiled contracts so that they can be matched with their "tests included" version.
-    pub(crate) fn from_built(built: Built) -> anyhow::Result<BuiltTests> {
+    /// `contract_dependencies` represents ordered (by deployment order) packages that needs to be deployed for each package, before executing the test.
+    pub(crate) fn from_built(
+        built: Built,
+        contract_dependencies: &HashMap<pkg::Pinned, Vec<Arc<pkg::BuiltPackage>>>,
```
mitchmindtree, line 220: Perhaps we should use a type alias for this?

forc-test/src/lib.rs
```diff
@@ -352,16 +450,43 @@ impl BuiltTests {
...
+            .member_nodes()
+            .map(|member_node| {
+                let graph = build_plan.graph();
+                let pinned_member = graph[member_node].clone();
+                let pinned_contract_dependencies: Vec<pkg::Pinned> = build_plan
+                    .contract_dependencies(member_node)
+                    .map(|contract_depency_node_ix| graph[contract_depency_node_ix].clone())
+                    .collect();
+                let contract_dependencies = pinned_contract_dependencies
+                    .iter()
```
mitchmindtree, line 469:
```suggestion
```
It looks like we might not need to collect the `pinned_contract_dependencies`, and can `filter_map` in this same iteration?

test/src/e2e_vm_tests/test_programs/should_pass/unit_tests/multi-contract-calls/contract_multi_test/src/main.sw
```diff
@@ -0,0 +1,41 @@
...
+#[test]
+fn test_contract_multi_call() {
+  let caller = abi(MyContract, CONTRACT_ID);
+  let caller2 = abi(MyContract2, contract2::CONTRACT_ID);
+
+  let should_be_true  = caller.test_true {}();
+  let should_be_false = caller2.test_false {}();
+
+  assert(should_be_true == true);
+  assert(should_be_false == false);
```
mitchmindtree, line 40:
```suggestion
  assert(should_be_true);
  assert(!should_be_false);
```

forc-test/src/lib.rs
```diff
@@ -115,24 +117,117 @@ pub struct TestPrintOpts {
...
+            None
+        }
+    }
+
+    /// Produces an iterator yielding all contract ids required to be included in the transaction
+    /// for this test setup.
+    fn contract_ids(&self) -> impl Iterator<Item = tx::ContractId> + '_ {
+        self.contract_dependency_ids()
+            .cloned()
+            .chain(self.root_contract_id().into_iter())
```
mitchmindtree, line 158:
```suggestion
            .chain(self.root_contract_id())
```
`chain` should except any type that implements `IntoIterator`, which `Option` does

forc-test/src/lib.rs
```diff
@@ -115,24 +117,117 @@ pub struct TestPrintOpts {
...
+        // Setup the interpreter for deployment.
+        let params = tx::ConsensusParameters::default();
+        let storage = vm::storage::MemoryStorage::default();
+        let mut interpreter =
+            vm::interpreter::Interpreter::with_storage(storage, params, GasCosts::default());
+
+        // Root contract is the contract that we are going to be running the tests of, after this
+        // deployment.
+        let (root_contract_id, root_contract_tx) =
+            deployment_transaction(&self.pkg, &self.without_tests_bytecode, params);
```
mitchmindtree, line 182: It looks like we don't need these until the final `interpreter.transact` call - perhaps we should move this call to after the contract dependency transactions have executed?

forc-test/src/lib.rs
```diff
@@ -115,24 +117,117 @@ pub struct TestPrintOpts {
 
 /// The storage and the contract id (if a contract is being tested) for a test.
 #[derive(Debug)]
-struct TestSetup {
+enum TestSetup {
+    ContractSetup(ContractTestSetup),
+    NonContractSetup(vm::storage::MemoryStorage),
+}
```
mitchmindtree, line 123: I wonder if we should remove this distinction?

E.g. a `script` type program might have contract dependencies and need the contract setup anyway? Maybe this should look something like what `ContractTestSetup` currently looks like, but with `root_contract_id` as an `Option` (i.e. `Some` if the root is a contract, `None` otherwise)? I.e.

```rust
struct TestSetup {
    storage: vm::storage::MemoryStorage,
    contract_dependency_ids: Vec<tx::ContractId>,
    root_contract_id: Option<tx::ContractId>,
}
```

For libraries and predicates that cannot have contract dependencies, perhaps the `contract_dependency_ids` `Vec` can just remain empty?

mitchmindtree (approved):
Nice one @kayagokalp, looking good! I've left a few small Qs and suggestions, but I don't mind if we address these now or in a follow-up PR.

mitchmindtree (approved):
:ok_hand:

JoshuaBatty (approved):

PR #4159, opened 2023-02-22
fix: calculate contract dependency ids with tests disabled

### LINKED ISSUE #4155: Contract dependency id injection should be determined with tests disabled

kayagokalp:
## Description
closes #4155.
unblocks #4156.

Since we are using the build profile derived from build flags of forc for nodes in the dependency graph, contract dependencies can end-up getting compiled with tests (we are fine if we compile with `forc build` but this issue surfaces once we use `forc build --tests`). This is not something we want because of two reasons:

1. In terms of `forc-pkg`: We shouldn’t allow this as inserted contract id and deployed contract id should be the same for the feature to work all the time. (This is the case if you run forc build today but if you run forc build --tests the inserted contract ids are essentially wrong, if you have some tests in those packages)

2. In terms of `forc-test`: The contract deployment’s are done without the tests so their contract ids are calculated without the tests, since forc-pkg  inserts the contract id calculated with tests included. The actual contract id in the interpreter instance and injected contract id is different. This means calls using the following abi declaration is failing: `let abi = abi(MyContractABIinContractDependency, contract_dependency::CONTRACT_ID)` . Basically we cannot use contract id namespace injection feature with the unit tests.

`forc-test` related problems are not apparent yet since multi contract calls are not enabled (they will be with #4156). 
## Checklist


kayagokalp:
Converting to draft as I realized I need to address the fact that this PR removes capability of executing tests from a package that is a contract dependency to some other package.

kayagokalp:
Realized this is not happening here at this PR, so good to go

mitchmindtree (commented):
Ahh good job spotting this issue!

Another more general way of describing this issue is that as the `BuildProfile` is an input to the code-gen process, it is in turn also an input to the contract ID calculation for contract-dependencies.

With the approach in this PR, I am a bit concerned that when the user performs a build of multiple workspace contract members with `include_tests` enabled, it might be surprising that only the absolute-top-level contract member binary actually has tests enabled.

Perhaps if tests are enabled, we should always produce two binaries for each workspace contract member - 1. with tests disabled (for interpreter deployment) and another with tests enabled (for script execution)?

For all non-workspace-member contract dependencies, it's probably fine to just disable tests as you've done here. In fact, it's probably fine to disable `include_tests` for **all** non-workspace-member packages *in general* (not just for contracts)?

Let me know your thoughts!

kayagokalp:
@mitchmindtree Sounds perfect to me and that was the initial thought I had while inserting [this comment](https://github.com/FuelLabs/sway/pull/4159#issuecomment-1440441426) but very surprisingly I can run the tests of a contract workspace member which is a contract dependency to another member. It was my expectation (and one of the problems I ran into while hacking on #4156) that I would lose the tests of pkgs that are both a workspace member and a contract dependency to any pkg.

Anyways, it is highly possible that I am missing something as I tested this after I sunk around ~10 hours into debugging different injection related errors popped out during #4156 😢. I will look here again with a fresh mind.

As a summary, I am %100 agreeing with you on the solution proposed (producing two binaries for contract dependencies that are also a workspace member) but couldn't generate the problem that should happen with this PR that would be solved with that solution.

forc-pkg/src/pkg.rs
```diff
@@ -99,6 +100,9 @@ pub struct BuiltPackage {
     source_map: SourceMap,
     pub pkg_name: String,
     pub built_pkg_descriptor: BuiltPackageDescriptor,
+    /// If this pkg is a contract dependency of another pkg in the build graph, this points to the
+    /// version of this pkg compiled without tests.
+    pub pkg_without_tests: Option<Arc<BuiltPackage>>,
```
kayagokalp, line 105: I am not really happy how this turned out, I wanted to break `BuiltPackage` into `BuiltContract` and `BuiltNonContract` to overcome this but I remember having a discussion not do that to keep different logic of building for different program types in only `forc-pkg`.

kayagokalp (commented):
I have a open quesiton but this is ready for review. I tested without the new addition and in-fact we were losing some member tests in the test execution. After the proposed soltuion we have all the tests for all members. One thing to note here is the test contract id injeciton is now done in `forc-pkg` rather than `forc-test`. That is because with this PR, we are compiling every contract (if tests are enabled) twice to get the version without tests included. If tests are disabled we only double compile contracts that are contract dependencies. To insert contract ids `forc-test` was recompiling with injection map. If I were to keep that we would be compiling contracts 3 times which is not ideal.

Also I have some new insights about #4162 that I understood while working on this PR will insert some comments there.

forc-pkg/src/pkg.rs
```diff
@@ -2971,6 +2985,67 @@ pub fn build(
...
+                    value: contract_id_value.clone(),
+                    public: true,
+                };
+                let constant_declarations = vec![(contract_id_constant_name, contract_id_constant)];
+                const_inject_map.insert(pkg.clone(), constant_declarations);
+            }
+            Some(built_package_without_tests)
+        } else {
+            None
+        };
```
mitchmindtree, line 3048: This section copies a lot of the same logic from below already used to construct a `BuiltPackage`. Can we abstract this out into a function and share some of the logic with the section below?

mitchmindtree: Also, it looks like this only builds *contracts* without tests, but I suspect we want to build *all* non-member dependencies without tests (as mentioned [here](https://github.com/FuelLabs/sway/pull/4159#pullrequestreview-1310438754)).

forc-pkg/src/pkg.rs
mitchmindtree: Rather than storing this inside of `BuiltPackage` and creating a recursive data structure, perhaps we can just store a second `BuiltPackage` in the `built_packages` `Vec` that is produced by the `build` function?

kayagokalp: Sounds good, I ended up retuning `Vec` of `(BuiltPackage, Option<BuiltPackage>)` as non-contract members are not recompiled without tests.

forc-pkg/src/pkg.rs
```diff
@@ -2044,6 +2058,14 @@ fn validate_contract_deps(graph: &Graph) -> Result<()> {
     Ok(())
 }
 
+/// Represents the result of building a package.
+///
+/// For non-contract packages second element of the tuple will always be `None`.
+///
+/// For contract packages second element of the tuple will hold the contract re-compiled without
+/// thte tests included.
+type BuildResult = (BuiltPackage, Option<BuiltPackage>);
```
mitchmindtree, line 2067: Hmmm I'm not sure `BuildResult` is a good name for this. `Result` is normally used for result types, i.e. types that might be an error.

Perhaps we're better off splitting up `BuiltPackage` into something like the following?

```rust
/// The result of successfully compiling a package.
#[derive(Debug, Clone)]
pub struct BuiltPackage {
    pub name: String,
    pub descriptor: BuiltPackageDescriptor,
    pub tree_type: TreeType,
    pub build_target: BuildTarget,
    pub json_abi_program: ProgramABI,
    pub storage_slots: Vec<StorageSlot>,
    source_map: SourceMap,
    pub bytecode: BuiltPackageBytecode,
    /// `Some` for contract member builds where tests were included. This is
    /// required so that we can deploy once instance of the contract (without
    /// tests) with a valid contract ID before executing the tests as scripts.
    ///
    /// For non-contract members, this is always `None`.
    pub bytecode_without_tests: Option<BuiltPackageBytecode>,
}

/// The bytecode associated with a built package along with its entry points.
struct BuiltPackageBytecode {
    pub data: Vec<u8>,
    pub entries: Vec<PkgEntry>,
}
```

This also helps to avoid duplicating stuff that stays the same (name, descriptor, etc).

forc-pkg/src/pkg.rs
```diff
@@ -2055,7 +2077,47 @@ pub fn build(
     profile: &BuildProfile,
     outputs: &HashSet<NodeIx>,
     const_inject_map: &ConstInjectionMap,
-) -> anyhow::Result<Vec<(NodeIx, BuiltPackage)>> {
+) -> anyhow::Result<Vec<(NodeIx, BuildResult)>> {
+    #[allow(clippy::too_many_arguments)]
+    fn compile_util(
```
mitchmindtree, line 2082: Can we call this `compile_pkg` instead? I think it's good to try to avoid "util" in favour of something more descriptive when we can (I know we have a bunch of util modules all over the place, but this is probably something we should aim to improve).

forc-pkg/src/pkg.rs
```diff
@@ -2055,7 +2077,47 @@ pub fn build(
...
+        lib_namespace_map: &HashMap<NodeIx, namespace::Module>,
+        compiled_contract_deps: &HashMap<NodeIx, BuiltPackage>,
+        profile: &BuildProfile,
+        plan: &BuildPlan,
```
mitchmindtree, line 2086: Perhaps we can take the hint from clippy here and put some of these immutable arguments in a `CompileCtx` struct  or similar to cut down on the number of arguments? It could also include the `engines` and `target` too. Something like:

```rust
struct CompilePkgCtx<'a> {
    lib_namespace_map: &'a HashMap<NodeIx, namespace::Module>,
    compiled_contract_deps: &'a HashMap<NodeIx, BuiltPackage>,
    profile: &'a BuildProfile,
    plan: &'a BuildPlan,
    engines: Engines<'a>,
    target: &'a BuildTarget,
}
```

mitchmindtree (approved):
Looking better!

I might do some follow-up refactoring of the build/compile parts of forc-pkg as it's starting to get a bit hairy, but I don't want to block you any further and am happy to see this land as is in the meantime :+1:

JoshuaBatty (approved):

PR #4169, opened 2023-02-23
DCA now detects dead code when all branches return explicitly.

### LINKED ISSUE #2313: Enhance DCA to warn on redundant `return` statements
Example:
```rust
script;

fn main() -> u64 {
    if true {
        return 0;
    } else {
        return 1;
    }
    return 2;
}
```
This code only compiles with https://github.com/FuelLabs/sway/pull/2312 but still does not give any warnings that `return 2;` is unreachable.

vaivaswatha:
Fixing `If` was straightforward. But that didn't automatically the fix `match` scenario (see test case). Turns out that `match` always has a covering "ImplicitReturn" expression, and with implicit returns in DCA, we add an edge from the beginning to the end, which I couldn't figure out why. Just removing that got it working. I request the reviewers to validate this part of the change carefully.

Closes #2313

## Checklist


mohammadfawaz (approved):
Nice! Very clean.

otrho (approved):

Issue #513, opened 2021-12-28
Show byte and gas costs


AlicanC:
To ease the burden of resource management on developers, we can provide info on resource costs such as the cost of importing a library (in bytes) and executing a line (or any piece) of code (in gas).

Bonus: Display dollar costs (or in any fiat)

# Prior Art

VSCode extension for JS:
https://github.com/wix/import-cost
![image](https://user-images.githubusercontent.com/412180/147516276-38db4195-c8e6-4fb8-af79-923095e7c5f5.png)

# Expectation

We should have byte costs like above and also gas costs. I assume the following data would be available, but not sure which should be displayed:
![image](https://user-images.githubusercontent.com/412180/147519165-19b11dda-85cb-4d30-b812-563b2c7097e0.png)

AlicanC:
A few notes:

- It doesn't have to be as detailed as in the screenshot. Less info still could be useful.
- It is assumed that all problems like "how do we know how many times this loop will execute" can be solved like "lets assume they all run once".

PR #4178, opened 2023-02-24
feat: display gas used for unit tests

### LINKED ISSUE #2411: [Tracking] Gas profiling for unit tests
This is a tracking issue for supporting gas profiling of unit tests.

Blocked by: #1832

---

The aim here is to enable devs to be able to reason about gas usage by writing unit tests for their APIs and being able to easily see the gas used.

To begin, this can be as simple as outputting the result of the gas usage from the receipt while running the unit tests.

Related: #513

kayagokalp:
## Description
closes #2411.

This PR adds used gas information to the output of `forc test`. An example run can be seen below:

Screen Shot 2023-02-24 at 4 47 29 PM

This is done automatically without any flags. I felt like this should be the default way but open to suggestions if you feel like it should be behind a flag like `--gas-usage`.

Also by accessing `TestResult` struct's new "used gas" field, we can add this information into vscode plugin just like #513. We do not have a line by line gas cost information just yet but we could show the whole test's used gas info next to the run button or something in vscode. Thoughts @JoshuaBatty @sdankel 

Please let me know if anyone has any requests around the flag and the way we are printing the used gas. Maybe @nfurfaro could have some ideas as the prime user of `forc-test` 😄 
## Checklist

### DIFF
--- a/forc-test/src/lib.rs
+++ b/forc-test/src/lib.rs
@@ -211,7 +211,10 @@ impl<'a> PackageTests {
                     .find_map(|receipt| match receipt {
                         tx::Receipt::ScriptResult { gas_used, .. } => Some(gas_used),
                         _ => None,
-                    }).ok_or_else(|| anyhow::anyhow!("missing used gas information from test execution"))?;
+                    })
+                    .ok_or_else(|| {
+                        anyhow::anyhow!("missing used gas information from test execution")
+                    })?;
 
                 // Only retain `Log` and `LogData` receipts.
                 let logs = receipts
--- a/forc-test/src/lib.rs
+++ b/forc-test/src/lib.rs
@@ -47,7 +47,7 @@ pub struct TestResult {
     pub name: String,
     /// The time taken for the test to execute.
     pub duration: std::time::Duration,
-    /// The span for the function declaring this tests.
+    /// The span for the function declaring this test.
     pub span: Span,
     /// The resulting state after executing the test function.
     pub state: vm::state::ProgramState,
@@ -55,6 +55,8 @@ pub struct TestResult {
     pub condition: pkg::TestPassCondition,
     /// Emitted `Recipt`s during the execution of the test.
     pub logs: Vec<fuel_tx::Receipt>,
+    /// Gas used while executing this test.
+    pub gas_used: u64,
 }
 
 const TEST_METADATA_SEED: u64 = 0x7E57u64;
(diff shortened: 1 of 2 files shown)


forc/src/cli/commands/test.rs
```diff
@@ -85,10 +85,11 @@ fn print_tested_pkg(pkg: &TestedPackage, test_print_opts: &TestPrintOpts) -> Res
             false => ("FAILED", Colour::Red),
         };
         info!(
-            "      test {} ... {} ({:?})",
+            "      test {} ... {} ({:?}) - (gas used: {})",
```
mitchmindtree, line 88:
```suggestion
            "      test {} ... {} ({:?}, {} gas)",
```
Maybe could be slightly more concise by putting time cost and gas cost together like this? This would look something like:
```
    test test_foo ... ok (4.3s, 1234 gas)
```

forc-test/src/lib.rs
```diff
@@ -204,6 +206,16 @@ impl<'a> PackageTests {
...
 
+                let gas_used = *receipts
+                    .iter()
+                    .find_map(|receipt| match receipt {
+                        tx::Receipt::ScriptResult { gas_used, .. } => Some(gas_used),
+                        _ => None,
+                    })
+                    .ok_or_else(|| {
+                        anyhow::anyhow!("missing used gas information from test execution")
+                    })?;
```
mitchmindtree, line 217: Do we know what cases under which we can get a receipt that has no `gas_used` field?

mitchmindtree (commented):
Niccce, this will be very handy!

forc-test/src/lib.rs
kayagokalp: Hmm I looked for this and it is not stated [in the specs](https://fuellabs.github.io/fuel-specs/master/protocol/abi/receipts.html#scriptresult-receipt). As far as i understood from the specs and quick search, every script will return a script result that contains gas used. I tested also tested this with `should_revert` tests. Since each of our tests runs as scripts we should be able to get it all the time.

nfurfaro:
Perhaps the gas used can be printed as the default, but there's a "--silent" flag to suppress anything but the pass/fail result... on the other hand, that's inconsistent with printing logs, for which the default is **_not_** printing them. 

I think the best approach is probably to require passing a flag to print the gas, which keeps the behaviour consistent with other logs that are printable.

mitchmindtree (approved):
Nice, lgtm!

I think it's probably fine to print gas alongside the elapsed time like this by default. Unlike the logs, it's always on the same line as the test result and doesn't add too much noise (especially with [this tweak](https://github.com/FuelLabs/sway/pull/4178#discussion_r1118339378)).

:shipit:

JoshuaBatty (approved):
Nice :shipit:

PR #4192, opened 2023-02-27
feat(fmt): fmt comments with local insertions

### LINKED ISSUE #3938: [Tracking] feat(fmt): rewrite comment formatting

bingcicle:
## Description

Closes #3938

### Problem

I've submitted a few PRs (#4129, #4093, #4005) implementing a new way of formatting comments in tackling #3938, but after submitting #4129 and following that working on formatting storage comments in a local branch realized that the method of formatting previously by writing `write_comments(..)` between every span is getting very unwieldy. That method results in a lot of repeated calls to `write_comments()`, and in some cases where we have to check if some value exists, it would lead to even more vertical space taken up because we have to do something like this:

```rust
...
if let Some(item) = my_item {
  write_comments(
    formatted_code,
    start..end,
    formatter,
  )
  // do other things
}
...
```

My sense is that this will probably lead to the formatter code being hard to maintain in future because of the length of each `Format` implementations, and having to work your way through a lot of conditional code and repeated calls to `write_comments(..)` isn't a very nice experience. In #4005, the comment formatting for `ItemAbi` alone led to **56 additional line insertions** in that PR (previously it was 54 LoC, which means the LoC doubled!)

### Solution

We adapt the original approach of inserting comments, but this time on a local scale. That means we parse each node on its own, compare its unformatted and formatted versions, and find out where to insert comments found between spans. Unlike the original approach, this approach will parse and format while the AST node is being visited and formatted - that means the entire `handle_comments(..)` that reparses the entire `Module` can be deprecated once this PR is in.

This is done mainly through the function `rewrite_with_comments(..)` that is called at the end of a `impl Format`. Essentially what this function does is it parses the unformatted and formatted versions of the currently visited node with `parse_snippet::<P: sway_parse::Parse + Format>(..)`, finding the comments between the unformatted spans, and inserting them into the formatted spans.

Based on the `CommentKind` of the found comments, we do some formatting to decide how they should be written.

Of course, this also means that unlike the previously proposed method of using `write_comments()` everywhere, where we have the indentation context, this new approach means that we yet again do not have the local context. To figure out indentation, we assume that each comment will always be 'pinned' to some node below it - if this node does not exist we try to 'pin' to a node above it instead.

## Checklist


bingcicle:
As of this comment this PR is able to pass tests and check examples without regression without the 2nd pass entirely. Going to check for regressions on sway-applications now.

swayfmt/tests/mod.rs
```diff
@@ -624,9 +620,9 @@ pub enum Bazz { // Here is a comment
 pub enum Bazz { // Here is a comment
     // Trying some ASCII art
     baz: (),
-    bazzz: (),//-----
-              //--D--
-              //-----
+    bazzz: (), //-----
+    //--D--
+    //-----
```
bingcicle, line 625: This is unindented because it would be difficult to support this kind of comment given the current heuristics/assumptions, but I'm also not sure if we even want this sort of alignment to begin with? Unless there is some context i am missing here.

swayfmt/tests/mod.rs
```diff
@@ -504,24 +503,23 @@ pub struct Foo { // Here is a comment
...
+    //  ________ ___  ___  _______   ___               ___       ________  ________  ________
+    // |\  _____\\  \|\  \|\  ___ \ |\  \             |\  \     |\   __  \|\   __  \|\   ____\
+    // \ \  \__/\ \  \\\  \ \   __/|\ \  \            \ \  \    \ \  \|\  \ \  \|\ /\ \  \___|_
+    //  \ \   __\\ \  \\\  \ \  \_|/_\ \  \            \ \  \    \ \   __  \ \   __  \ \_____  \
+    //   \ \  \_| \ \  \\\  \ \  \_|\ \ \  \____        \ \  \____\ \  \ \  \ \  \|\  \|____|\  \
+    //    \ \__\   \ \_______\ \_______\ \_______\       \ \_______\ \__\ \__\ \_______\____\_\  \
+    //     \|__|    \|_______|\|_______|\|_______|        \|_______|\|__|\|__|\|_______|\_________\
+    //                                                                                  \|_________|
 }
-// This is a comment
```
bingcicle, line 524: Adding a new `post_module_comments()` test to specifically test for post module comments, so i'm removing this comment here in this test case.

I also believe for this test case that the ascii should follow indentation of the previous line, so that's changed here too.

swayfmt/tests/mod.rs
```diff
@@ -504,24 +503,23 @@ pub struct Foo { // Here is a comment
              //     \|__|    \|_______|\|_______|\|_______|        \|_______|\|__|\|__|\|_______|\_________\
              //                                                                                  \|_________|
 }
-// This is a comment
 "#,
         r#"contract;
 // This is a comment, for this one to be placed correctly we need to have Module visitor implemented
 pub struct Foo { // Here is a comment
+
```
bingcicle, line 510: A heuristic I stuck to while rewriting the logic was to allow at most 1 newline if there are multiple newlines between comments. Should we actually squash multiline comments that are newline separated?

bingcicle:
This should be good for review and actually should address the big issue #3938, since in this PR we completely deprecate the entire `handle_comments` 2nd pass.

Aside from passing the test suite, i've also made sure to run this version of the formatter against the existing `sway-applications` projects and made sure there weren't any regressions, though there are a few formatting changes that probably has to do with using a newer `forc-fmt` instead.

swayfmt/tests/mod.rs
mitchmindtree: > Should we actually squash multiline comments that are newline separated?

I think what you currently have (leaving just one empty line if one or more empty lines previously existed) is the right approach :+1:

swayfmt/tests/mod.rs
mitchmindtree: Yeah I think I'm content with not supporting this kind of aligned trailing-yet-multiline comment - it's pretty niche, and I think we've come to terms with the fact we may need to be at least slightly opinionated on what we allow for comments.

swayfmt/tests/mod.rs
```diff
@@ -586,9 +584,7 @@ fn multiline_comments() {
 }"#,
         r#"contract;
 
-fn single_comment_same_line() {
-    /* a comment */
-}
+fn single_comment_same_line() { /* a comment */ }
```
mitchmindtree, line 587: I'm curious - if this comment were a `// a comment` instead of `/* a comment */`, would it behave the same? Or does the formatter know this is collapsible because it's a "block" comment?

mitchmindtree (approved):
Thanks so much for diving into this Bing, great work in digging in and coming to grips with such a tricky problem!

I agree this approach seems like a happier balance between the previous approach to comment re-insertion and integration into the initial traversal. Another benefit that comes to mind is that, if we do come across a strange edge case interaction between the comment `rewrite` behaviour and a particular item type, at least we can more easily manually replace the re-insertion approach at the local level with a manual approach if we need to, without affecting the behaviour of the rest of the items.

utACK, thanks for also testing with the applications!

I wonder if we should try and plug this into the passing E2E test workflow as a follow-up, and to better catch any regressions that might pop up? This might also help to ensure new language items are supported in the formatter when initially introduced?

swayfmt/tests/mod.rs
bingcicle: Hmm good catch. So this:

```rust
fn single_comment_same_line() { // a comment
}
```

actually doesn't newline it atm. Working on a fix

bingcicle: Maybe this isn't an actual issue - rustfmt has the same result. Perhaps we can merge this first and reconsider semantics again?

mitchmindtree: In my original comment I was just checking that this:
```rust
fn single_comment_same_line() {
    // a comment
}
```
doesn't get turned into this
```rust
fn single_comment_same_line() { // a comment }
```

If not, and if we behave the same as rustfmt, then feel free to disregard :)

bingcicle:
I messed up the commit for a bit there but we should be good now

> I wonder if we should try and plug this into the passing E2E test workflow as a follow-up, and to better catch any regressions that might pop up? This might also help to ensure new language items are supported in the formatter when initially introduced?

By the passing E2E test workflow, do you mean in `cargo-run-e2e-test` job?

mitchmindtree:
Yeah I was referring to the `should_pass` tests in `test/src/e2e_vm_tests/test_programs/should_pass/`. My thinking is that, we know they should already at least always be parsing successfully, so we should be able to reliably check their formatting too.

bingcicle:
> Yeah I was referring to the `should_pass` tests in `test/src/e2e_vm_tests/test_programs/should_pass/`. My thinking is that, we know they should already at least always be parsing successfully, so we should be able to reliably check their formatting too.

Good suggestion - I'll create an issue to track this

swayfmt/src/utils/map/byte_span.rs
```diff
@@ -44,7 +44,7 @@ impl ByteSpan {
     /// by a given Range<usize>, where 'start' is inclusive. This is to account for the start
     /// of the file.
```
kayagokalp, line 45: Maybe we can also mention the end is inclusive too in this doc comment, since stating start is inclusive explicitly feels like end is not?

kayagokalp (approved):
Looks good to me! It is so nice to see comment handling getting better 🚀 🚢 . Just left an unimportant nit for a (i guess) outdated doc comment

swayfmt/src/utils/map/byte_span.rs
bingcicle: good catch, thank you!

swayfmt/tests/mod.rs
eureka-cpu: I think we had some kind of threshold before that the user could set to change the number of newlines between comments and items if they wanted to. But looking back on it, I don't really see anyone actually wanting to use that feature so this is good as it is.

examples/config_time_constants/src/main.sw
```diff
@@ -29,9 +29,9 @@ abi TestContract {
 }
 
 impl TestContract for Contract {
-// ANCHOR: using_configurables
+    // ANCHOR: using_configurables
```
eureka-cpu, line 32: nice

examples/option/src/main.sw
```diff
@@ -14,7 +14,7 @@ fn main() {
     match result {
         // The division was valid
         Option::Some(x) => std::logging::log(x),
-         // The division was invalid
+        // The division was invalid
```
eureka-cpu, line 17: 🚢

examples/storage_variables/src/main.sw
```diff
@@ -38,7 +38,7 @@ impl StorageExample for Contract {
         storage.var2.z = true;
     }
     // ANCHOR_END: storage_write
-    
+
```
eureka-cpu, line 41: 🚢 🚢

swayfmt/tests/mod.rs
```diff
@@ -504,24 +503,23 @@ pub struct Foo { // Here is a comment
...
-             //     \|__|    \|_______|\|_______|\|_______|        \|_______|\|__|\|__|\|_______|\_________\
-             //                                                                                  \|_________|
+    //  ________ ___  ___  _______   ___               ___       ________  ________  ________
+    // |\  _____\\  \|\  \|\  ___ \ |\  \             |\  \     |\   __  \|\   __  \|\   ____\
+    // \ \  \__/\ \  \\\  \ \   __/|\ \  \            \ \  \    \ \  \|\  \ \  \|\ /\ \  \___|_
+    //  \ \   __\\ \  \\\  \ \  \_|/_\ \  \            \ \  \    \ \   __  \ \   __  \ \_____  \
+    //   \ \  \_| \ \  \\\  \ \  \_|\ \ \  \____        \ \  \____\ \  \ \  \ \  \|\  \|____|\  \
+    //    \ \__\   \ \_______\ \_______\ \_______\       \ \_______\ \__\ \__\ \_______\____\_\  \
+    //     \|__|    \|_______|\|_______|\|_______|        \|_______|\|__|\|__|\|_______|\_________\
+    //                                                                                  \|_________|
```
eureka-cpu, line 521: 🚢 🚢 🚢

eureka-cpu (approved):
LGTM other than the one weird formatting of `// comment` after same line functions

examples/storage_variables/src/main.sw
eureka-cpu: really glad to see this fixed

kayagokalp (approved):

bingcicle:
> LGTM other than the one weird formatting of `// comment` after same line functions

Which one is this? Are you referring to:

```rust

fn single_comment_same_line_trailing() {  // a comment
}
```

?

Issue #1950, opened 2022-06-13
Stricter SemVer adherence. Version library and binary crates separately.


mitchmindtree:
# Motivation

Currently, we consider SemVer across all of the crates within the Sway repo only with respect to *binary compatibility* and not library compatibility (i.e. `pub` APIs and observable side-effects).

Now that we are open-source and publishing all library crates to crates.io, we should re-think how we treat SemVer in order to better adhere to the crates.io versioning expectations and to avoid burning the trust of downstream, 3rd-party tooling developers that wish to depend on our library crates.

# Potential Solution

## Version `lib` and `bin` crates separately.

This would allow us to more clearly distinguish between library-compatibilty and binary-compatibility in our semver versioning across our crates.

We could consider splitting our repo into `lib` and `bin` directories to make it clearer which set of crates are libs, which are user-facing bins. This might make the distinct version-bump process a little easier?

For our crates that expose both binary and library targets, we should consider splitting out the binary crate so that it can be versioned separately. E.g. in the case of `forc`, we could split the `bin` target (which is the command-line tool) into a `forc-cli` crate with *only* a `bin` target, and change the `forc` crate to only expose a `lib` target.

## Stricter SemVer for all `lib` crates

We should attempt to adhere to [Cargo's SemVer Compatibility reference](https://doc.rust-lang.org/cargo/reference/semver.html) for our library versions.

Without this, it will be impossible for downstream users who want to build tools with our libraries to have any guarantees of stability from our semver versions. This would also bring us more in alignment with the vast majority of crates on crates.io.

---

cc @mohammadfawaz @adlerjohn @sezna for thoughts.

PR #4193, opened 2023-02-27
Make AST serializable

AlicanC:
- Implements `serde::Serialize` for `sway_ast::Module` and such.
- Adds a simple integration test to `sway_parse` which uses `insta` for RON snapshot testing.

This should allow us to dump AST in more convenient formats and also use tools like `insta` to do testing on the serialized output.

Note: Deserialization was left for later.


AlicanC:
@kayagokalp will it be necessary to keep `no-span-debug` after this? If not I can remove that here.

kayagokalp:
> will it be necessary to keep `no-span-debug` after this? If not I can remove that here.

I think the use case for the `--no-span-debug` still exists: I originally needed it while printing out entire namespaces for debugging and spans were taking %90 of the output. So I would say we should keep it but good with it either way.

AlicanC:
> Is the idea to add more AST tests in this style? And what is the idea around adding deserialization?
> 
> 
> 
> Overall this seems fine, but I'd be wary about building more testing infrastructure on this as it exposes all the internal details of the AST. And since it's all opaque to the compiler, every little refactoring will cause these tests to fail, making them a bit brittle.

Not planning to add any tests in this E2E style (maybe one for each module type?), but I've found some unit tests for parsing complicated stuff like PathExpr and UseItem necessary.

I'm aware how brittle this is, but: is this okay because we eventually want a clean/stable AST serialization and this will push us towards that? Do I want to mess freely with the AST or do I want that test to keep me in check so when I'm working on AST/parsing I'm forced to be conscious on how it's serialized?

On deserialization: I had to just discard some info during serialization to avoid refactoring too much so the format is lossy. Refactoring Span and such should open the way for it.

sway-ast/Cargo.toml
```diff
@@ -10,6 +10,7 @@ description = "Sway's parser"
 
 [dependencies]
 extension-trait = "1.0.1"
-num-bigint = "0.4.3"
+num-bigint = { version = "0.4.3", features = ["serde"] }
 num-traits = "0.2.14"
+serde = { version = "1.0", features = ["derive"] }
```
kayagokalp, line 15: Hmm I wonder if we should have a `serde` feature that enables this just like most of the crates we depend on. The serde feature would enable the actual serialization implementations. 

I am imagining serialization is not needed by the compiler itself but it is a good option to offer to downstream crates (for example lsp).

kayagokalp (commented):
Looking good! I am not sure where to stand in terms of consuming this serialization in tests as it would be hard to maintain. At the same time it seems like a good idea to somehow enforce AST stabilization as it could lead us to be aware of breaking changes from library perspective (although we are not doing that right now, lately I started dreaming about [this](https://github.com/FuelLabs/sway/issues/1950) 😄) 

That being said I think this is an improvement even we end up not using the tests as downstream crates may want to (de*)serialize the AST and if we let them have this option with a feature, I feel like it would be a good practice overall

sway-ast/Cargo.toml
AlicanC: Compiler could have a parse cache, no?

kayagokalp: Hmm that makes sense 👌

sway-types/src/ident.rs
```diff
@@ -81,6 +83,18 @@ impl BaseIdent {
 /// often be different.
 pub type Ident = BaseIdent;
 
+impl Serialize for Ident {
+    // Serialize an `Ident` struct with two fields: `to_string` and `span`.
```
kayagokalp, line 87: Maybe these were intended to be doc comments? 
```suggestion
    /// Serialize an `Ident` struct with two fields: `to_string` and `span`.
```

sway-types/src/span.rs
```diff
@@ -76,6 +78,18 @@ pub struct Span {
     path: Option<Arc<PathBuf>>,
 }
 
+impl Serialize for Span {
+    // Serialize a tuple two fields: `start` and `end`.
```
kayagokalp, line 82: ^
```suggestion
    /// Serialize a tuple two fields: `start` and `end`.
```

kayagokalp (commented):
lgtm, left 2 non-important nits deferring to compiler team for the approvals

IGI-111:
>I'm aware how brittle this is, but: is this okay because we eventually want a clean/stable AST serialization and this will push us towards that? Do I want to mess freely with the AST or do I want that test to keep me in check so when I'm working on AST/parsing I'm forced to be conscious on how it's serialized?

I'm not sure we want the AST to be stable to the point of serializing identically across versions. That seems like a lot more coupling than necessary to me.

More likely we want the structure to remain the same but the node content will very likely have to evolve.

AlicanC:
I think we need to start keeping our AST stable after some point. There will be tools like [jscodeshift](https://github.com/facebook/jscodeshift) that work on ASTs that would break. Even today, a simple change in the parser will have you refactor all the way down to the formatter.

IGI-111 (approved):

mohammadfawaz (approved):

Issue #2408, opened 2022-07-29
Add autocomplete support for `expr.` 


JoshuaBatty:
The suggested completion items should be aware of the current context and only suggest correct fields related to what is being typed. 

For example, say we have:

```rust
struct Foo {}

impl Foo {
    pub fn new() -> Self { Self {} }
    pub fn bar() {}
    pub fn baz() {}
}

fn main () {
    let f = Foo::new();
    let x = f. // <- only possible suggestions should be `new` `bar` and `baz`
}

JoshuaBatty:
This is currently blocked by #1674

We need access to all of type checked tokens in a file while the user is typing in order to provide semantic aware completions.

PR #4207, opened 2023-03-01
Limited autocomplete for expressions

sdankel:
## Description

This PR adds autocomplete functionality for expressions to the LSP. It's triggered by entering `.` after a variable. 

Closes https://github.com/FuelLabs/sway/issues/2408

### Use cases

- Works inside of `main` functions
- Works inside of other functions outside of an impl
- Supports function chaining
- Finds all methods for a struct anywhere in the namespace

### Limitations

- Does not work inside of impl methods. This is because "incomplete" impl methods apparently cause the compiler to error out before generating tokens for the file
- Can be slow to populate. This slowness is due to the compile time after every keystroke.
- Doesn't show up in certain cases, depending on what other errors are inside the file. This is because the compiler won't generate tokens when certain errors are present.

### Screen captures

Feb-28-2023 16-36-04

method chaining

Feb-28-2023 16-39-35

## Checklist


mohammadfawaz:
This is very impressive 😍

sway-lsp/src/core/token.rs
```diff
@@ -21,29 +21,30 @@ use tower_lsp::lsp_types::{Position, Range};
 /// useful to the language server.
 #[derive(Debug, Clone)]
 pub enum AstToken {
+    Attribute(Attribute),
```
JoshuaBatty, line 24: Nice, i've been meaning to reorder these :)

sway-lsp/src/core/token_map.rs
```diff
@@ -57,30 +57,72 @@ impl TokenMap {
...
+                    .try_unwrap()
+                    .map(|item| {
+                        let ((ident, _), token) = item.pair();
+                        (ident.clone(), token.clone())
+                    })
+            })
+    }
+
+    /// Returns all collected tokens that are at the cursor position.
+    pub fn tokens_at_position(
```
JoshuaBatty, line 91: I'm not sure I fully understand the intended context of this change. There should only ever be a single token at a given position. Can you explain a bit more about what your intention is here?

sdankel: Sometimes there are multiple idents at a position. For instance, with the expression `something.`, where the cursor position is before the `.`, the tokens returned are `something` and `.`

This is because we're doing an inclusive comparison of the tokens' ranges. Previously `ident_at_position` would just return the first ident found, discarding the rest. I changed it to `idents_at_position` to return all of the idents found, and updated the comment of `token_at_position` to reflect this.

The reason for the `tokens_at_position` function is that we want to find tokens that have this position in the _body_ of the token, not necessarily the ident of the token. If we want to get the token for the function declaration that the cursor position is inside of, searching with the span of the ident of the function declaration won't be enough. 

For example:

```
fn fun() {
  let something = MyStruct::new();
  something. // <-- cursor position is before the `.`
}
```

I'm using `tokens_at_position` to find the declaration of `fun` so that I can look inside its body and parameters to see what type `something` is. If I used `idents_at_position` I would only get back the ErrorRecovery tokens (`something` and `.`), but I need to know the contextual token that contains it (the function declaration).

sway-lsp/src/server.rs
```diff
@@ -370,12 +369,16 @@ impl LanguageServer for Backend {
         &self,
         params: CompletionParams,
     ) -> jsonrpc::Result<Option<CompletionResponse>> {
+        let trigger_char = params
+            .context
+            .unwrap()
```
JoshuaBatty, line 374: can we handle this case instead of unwrapping?

sway-lsp/src/core/session.rs
```diff
@@ -243,10 +246,33 @@ impl Session {
...
+        let (ident_to_complete, _) = self.token_map.token_at_position(uri, shifted_position)?;
+        let fn_tokens = self
+            .token_map
+            .tokens_at_position(uri, shifted_position, Some(true));
+        let (_, fn_token) = fn_tokens.first()?;
+        let compiled_program = &*self.compiled_program.read();
+
+        if let Some(TypedAstToken::TypedFunctionDeclaration(fn_decl)) = fn_token.typed.clone() {
+            return Some(capabilities::completion::to_completion_items(
+                &compiled_program.typed.clone().unwrap().root.namespace,
```
JoshuaBatty, line 268: can we if let instead of unwrapping. Perhaps the rwlock failed for some reason to write then this could return `None`

sway-lsp/src/core/token_map.rs
JoshuaBatty: Ok that makes sense. Maybe it's worth elaborating on the doc comment the intended use case just so it's a bit clearer.

JoshuaBatty:
Nice this is looking really good. Just pulled down and compared our suggestions to rust-analyzer 

sway-lsp
Screenshot from 2023-03-08 12-33-59

rust-analyzer
Screenshot from 2023-03-08 12-34-58

Looks like the only difference is rust-analyzer is replacing `Self` with the name of the type self refers to. Is it possible to change our impl so it matches this behaviour also?

sdankel:
> Nice this is looking really good. Just pulled down and compared our suggestions to rust-analyzer
> 
> sway-lsp Screenshot from 2023-03-08 12-33-59
> 
> rust-analyzer Screenshot from 2023-03-08 12-34-58
> 
> Looks like the only difference is rust-analyzer is replacing `Self` with the name of the type self refers to. Is it possible to change our impl so it matches this behaviour also?

Good catch! Updated.

JoshuaBatty (approved):
Nice work on this :shipit:

mitchmindtree (approved):

PR #4210, opened 2023-03-01
Check for regressions in warnings thrown

### LINKED ISSUE #3189: Upgrade E2E testing to check for compile warnings, to prevent unnoticed regression in testing
There have been a fair number of GH issues opened recently regarding the appearance of incorrect compile warnings: #3093, #3067, #3012. My memory is telling me (but haven't confirmed) that some of these incorrect cases are new regressions.

Even if these particular cases are not new regressions in specific compile warnings, we should consider upgrading E2E testing to check to compile warnings, to prevent any potential unnoticed regression in the future.

From my understanding, the way that we have set up most of the `test.toml` files with the [`filecheck`](https://docs.rs/filecheck/latest/filecheck/) crate that we are using for failure testing will check to see if there are regressions in the output of compile warnings (depending on the directives given to `filecheck` it seems ?), but I haven't looked into it deeply.

We should consider solving this problem by introducing a user-facing compile-time flag to promote compile warnings to compile errors. Although I imagine that it is also possible to check with `filecheck` directly on the compile warnings themselves.

vaivaswatha:
## Description
This PR introduces a `test.toml` entry called "expected_warnings = count", and checks that warnings thrown by a test is not more than that (i.e., checks against new regressions in warnings thrown).

It updates the existing tests to enter the current counts in their `test.toml` files.

Closes #3189 

## Checklist


vaivaswatha:
> This looks like a great start, but looks like you're only checking if the warnings produced are more than the expected warnings.
> 
> Shouldn't we also deal with the case when we produce less warnings than expected?

If we're actually expecting a warning, then that must taken care of by FileCheck directives  right? (and is already done, DCA tests for example all work like this).

mohammadfawaz (approved):
Looks good, thanks!

PR #4215, opened 2023-03-03
Implement usize type, using u32 for miden

p0ry-g0n:
## Description

This PR adds a `usize` type that is a `u64` for the Fuel VM, a `u32` for the Miden VM, and panics for the EVM (😄 -- sorry. Not sure if I can just make it 256 bits without rewiring some other parts of the compiler to use a bigint, and I felt that was out of the scope of what I was doing).

I don't see an issue for a `usize` type but it was mentioned to me as needing to be done. 

## Checklist

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -1504,8 +1504,10 @@ pub fn dependency_namespace(
     node: NodeIx,
     constants: BTreeMap<String, ConfigTimeConstant>,
     engines: Engines<'_>,
+    build_target: BuildTarget,
 ) -> Result<namespace::Module, vec1::Vec1<CompileError>> {
-    let mut namespace = namespace::Module::default_with_constants(engines, constants)?;
+    let mut namespace =
+        namespace::Module::default_with_constants(engines, constants, build_target)?;
 
     let node_idx = &graph[node];
     namespace.name = Some(Ident::new_no_span(Box::leak(
@@ -1540,7 +1542,7 @@ pub fn dependency_namespace(
                     public: true,
                 };
                 constants.insert(CONTRACT_ID_CONSTANT_NAME.to_string(), contract_id_constant);
-                namespace::Module::default_with_constants(engines, constants)?
+                namespace::Module::default_with_constants(engines, constants, build_target)?
             }
         };
         namespace.insert_submodule(dep_name, dep_namespace);
@@ -2137,6 +2139,7 @@ pub fn build(
         node: NodeIx,
         source_map: &mut SourceMap,
         compile_ctx: CompilePkgCtx,
+        target: BuildTarget,
     ) -> Result<(BuiltPackage, sway_core::namespace::Root)> {
         let plan = compile_ctx.plan;
         let graph = plan.graph();
@@ -2150,6 +2153,7 @@ pub fn build(
             node,
             compile_ctx.constants.clone(),
             compile_ctx.engines,
+            target,
         ) {
             Ok(o) => o,
             Err(errs) => {
@@ -2218,7 +2222,7 @@ pub fn build(
                 without_tests_bytecode: None,
             };
             let (built_contract_without_tests, _) =
-                compile_pkg(node, &mut source_map, compile_pkg_context)?;
+                compile_pkg(node, &mut source_map, compile_pkg_context, target)?;
             // If this contract is built because tests are enabled we need to insert CONTRACT_ID
             // for the contract.
             if is_contract_dependency {
@@ -2273,7 +2277,7 @@ pub fn build(
             without_tests_bytecode,
         };
         let (mut built_package, namespace) =
-            compile_pkg(node, &mut source_map, compile_pkg_context)?;
+            compile_pkg(node, &mut source_map, compile_pkg_context, target)?;
         if let TreeType::Library { ref name } = built_package.tree_type {
             let mut namespace = namespace::Module::from(namespace);
             namespace.name = Some(name.clone());
@@ -2448,6 +2452,7 @@ pub fn check(
             node,
             constants,
             engines,
+            build_target,
         )
         .expect("failed to create dependency namespace");
 
--- a/sway-ast/src/literal.rs
+++ b/sway-ast/src/literal.rs
@@ -29,6 +29,7 @@ pub enum LitIntType {
     I16,
     I32,
     I64,
+    Usize,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Hash)]
--- a/sway-core/src/abi_generation/evm_json_abi.rs
+++ b/sway-core/src/abi_generation/evm_json_abi.rs
@@ -80,6 +80,7 @@ pub fn json_abi_str(type_info: &TypeInfo, type_engine: &TypeEngine) -> String {
             IntegerBits::Sixteen => "uint16",
             IntegerBits::ThirtyTwo => "uint32",
             IntegerBits::SixtyFour => "uint64",
+            IntegerBits::Usize => todo!(),
         }
         .into(),
         Boolean => "bool".into(),
@@ -127,6 +128,7 @@ pub fn json_abi_param_type(type_info: &TypeInfo, type_engine: &TypeEngine) -> et
             IntegerBits::Sixteen => ethabi::ParamType::Uint(16),
             IntegerBits::ThirtyTwo => ethabi::ParamType::Uint(32),
             IntegerBits::SixtyFour => ethabi::ParamType::Uint(64),
+            IntegerBits::Usize => ethabi::ParamType::Uint(256),
         },
         Boolean => ethabi::ParamType::Bool,
         B256 => ethabi::ParamType::Uint(256),
--- a/sway-core/src/abi_generation/fuel_json_abi.rs
+++ b/sway-core/src/abi_generation/fuel_json_abi.rs
@@ -631,6 +631,8 @@ impl TypeInfo {
                 IntegerBits::Sixteen => "u16",
                 IntegerBits::ThirtyTwo => "u32",
                 IntegerBits::SixtyFour => "u64",
+                // Usize is u64 for fuel
+                IntegerBits::Usize => "u64",
             }
             .into(),
             Boolean => "bool".into(),
--- a/sway-core/src/lib.rs
+++ b/sway-core/src/lib.rs
@@ -299,7 +299,12 @@ pub fn parsed_to_ast(
         value: typed_program_opt,
         mut warnings,
         mut errors,
-    } = ty::TyProgram::type_check(engines, parse_program, initial_namespace);
+    } = ty::TyProgram::type_check(
+        engines,
+        parse_program,
+        initial_namespace,
+        build_config.map(|x| x.build_target).unwrap_or_default(),
+    );
     let mut typed_program = match typed_program_opt {
         Some(typed_program) => typed_program,
         None => return err(warnings, errors),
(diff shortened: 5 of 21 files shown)


otrho:
I'm still not convinced we need a `usize` type in Sway.  Can you give some examples of when a Sway developer would need to use it, or which libraries/APIs would require it?  We already have a `u32` type, as well as the `u64`.

p0ry-g0n:
> I'm still not convinced we need a `usize` type in Sway. Can you give some examples of when a Sway developer would need to use it, or which libraries/APIs would require it? We already have a `u32` type, as well as the `u64`.

In Miden, the main numeric type is `u32`. To write platform-independent code, we need an abstraction over this primitive type.

As an example, a function written `fn foo(a: u64) -> u64` would have to be re-written for both Miden and Fuel, while a function `fn foo(a: usize) -> usize` would be portable.

otrho:
Yep, I understand how `usize` works traditionally, but I was a C++ dev for a long time and I reckon a variable sized type is an anti-feature.  I really want to make sure it's necessary before we make it a built-in type.

On the one hand, if I'm writing platform-independent code, for various blockchains, I should be either aware of each target and use the right types (e.g., `u32` when it is appropriate) or I'm unaware and expect code to work on any target then I should use the smallest type that will fit the data, like `u32` or `u16` for small numbers.

Or perhaps the next step, after introducing `usize` do we need to replace all of the uses of `u64` which are currently in the standard libraries, within reason?  If this was the case I doubt we should call it `usize` as that's mostly an anachronism from C, it should be just `uint` or `native` or something.

And there are other options.  Sway currently defaults to `u64` for unknown `Numeric` constants, but it might be feasible for this to be `u32` for Miden.  Also, there are IR passes which can analyse the use of integers and demote types automatically.  So if analysis can determine that a `u64` value never actually uses more than 32 bits then a `u32` can be used instead.  Considering most blockchains have quite limited VMs, the arithmetic need for a whole 64-bit number is practically non-existent.  Sway chose `u64` by default originally because the architectures running the Fuel VM will all be 64-bit and so it might as well also have 64bit values, not because 64 bits is necessary for arithmetic.

So pinging @mohammadfawaz and @jjcnn for your thoughts?

p0ry-g0n:
> On the one hand, if I'm writing platform-independent code, for various blockchains, I should be either aware of each target and use the right types (e.g., u32 when it is appropriate) or I'm unaware and expect code to work on any target then I should use the smallest type that will fit the data, like u32 or u16 for small numbers. 

In Sway's case, the platforms are wildly different. The EVM has 256 bit native numbers, and Miden has 32. This would mean that generic code would be, for no reasons other than organizationally, limited to 1/8th its capacity. In smart contract programming, I believe that notion is too low-level for the programmer to have to consider.

> Or perhaps the next step, after introducing usize do we need to replace all of the uses of u64 which are currently in the standard libraries, within reason? If this was the case I doubt we should call it usize as that's mostly an anachronism from C, it should be just uint or native or something. 

That would be the next step. I didn't mean to take a stance on naming, I was just following Rust. But after this type is introduced, it would make sense to use it throughout the existing standard library.

> And there are other options. Sway currently defaults to u64 for unknown Numeric constants, but it might be feasible for this to be u32 for Miden. Also, there are IR passes which can analyse the use of integers and demote types automatically. So if analysis can determine that a u64 value never actually uses more than 32 bits then a u32 can be used instead. Considering most blockchains have quite limited VMs, the arithmetic need for a whole 64-bit number is practically non-existent. Sway chose u64 by default originally because the architectures running the Fuel VM will all be 64-bit and so it might as well also have 64bit values, not because 64 bits is necessary for arithmetic.

Would adding a front end type annotation for `Numeric` and using that instead be sufficient here? It accomplishes the same goal. To be frank, it is almost the same as this `usize` type, but with the distinction happening in the inference system instead of prior to that.

mohammadfawaz:
> Sway currently defaults to u64 for unknown Numeric constants, but it might be feasible for this to be u32 for Miden.

This is something we should certainly do. For example, when writing `let x = 5;`, `x` should be a `u32` when targeting MidenVM.

It seems to me that the debate here is whether we want to allow target-independent code or not, particularly interfaces or functions that take or return integers such as `fn foo(a: usize) -> usize`. @otrho do you think `usize` an anti-pattern in Rust as well?

@p0ry-g0n one thing to pay attention to is type unification for integers [here](https://github.com/FuelLabs/sway/blob/0125b3851f48116393f59ced0f52cb284b2d74f8/sway-core/src/type_system/unify.rs#L318). Sway has implicit casting of integer type (though that may change soon) and we may need to take care of `usize` in type unification by emitting the appropriate warnings. For example, upcasting `usize` to `u64` should cause a warning when targeting MidenVM, but not for the FuelVM.

otrho:
> do you think `usize` an anti-pattern in Rust as well?

Yeah, it's supposed to represent the size of a pointer.  The only times I ever come across `usize` in Rust is as indices into slices or arrays and half the time I'm casting them from other integers.  If that's when we'd use them in Sway then it's fairly pointless as there aren't blockchains with more than 4G of linear memory, is there?  They could always just be cast to a native integer by the backend.

`int` and `long int` are different sizes in C depending on both the architecture and operating system and all it does it confuse things.  It makes it _harder_ to write platform independent code.  That's why Rust has the bit size of the integers in the types themselves, `i16`, `i32`, etc., there's no ambiguity.  I really do wonder how far we'd get by defaulting to `u32` for `Numeric` on Miden and/or fiddling with their use in IR and the backend.  I think the IR could demote them even to `u63` 99% of the time.

jjcnn:
I've had a busy weekend, so I only started thinking about this last night. I might very well change my opinion later on, so suitable caveats apply. I should also stress that I speak from the longer-term perspective for MidenVM, whereas @p0ry-g0n is speaking from a shorter-term perspective due to the specific tasks that have been assigned to them.

First off: In designing a programming language I believe the programmer should not have to worry about anything other than business logic. In the ideal world this means there is no `u32` or `u64` - there is just `uint`, and the compiler makes sure that that behaves like an unsigned integer.

We do not live in an ideal world and we're not getting rid of `u32` and `u64`, so the question boils down to "Given that the programmer does need to worry about the size of integer representations, what allows the programmer to think as little as possible about the underlying architecture?" I would argue that the answer is `u32` and `u64`, because that specifically tells the user that this is the size, regardless of architecture.

Second: Is there a performance hit on FuelVM from using `u32` over `u64`? And is there likely to ever be a performance hit on any target architecture? 

Say that `usize` is added, and say that it translates to different sizes on different architectures, specifically `u32` for MidenVM and `u64` for FuelVM. A programmer who uses `usize` must be in a situation where an overflow won't happen in `u32`, because if it did then the program would behave differently on FuelVM and MidenVM (for no good business-logic reason that I can see). So why does the programmer not just use `u32` instead of `usize`? That would ensure the same behavior across all platforms (including aborting due to 32-bit overflows on 64-bit architectures, should the programmer have made a mistake), at no performance cost.

There is definitely a case to be made for `usize` in C-style languages where you may want to do your own memory management, but I would question the sense in having those kinds of features in a smart contract language... :-)

Third: From a MidenVM point of view it probably won't make much of a difference what you end up deciding. My main worry is the performance hit we will get from `u64` arithmetic, and the (smaller, but still worrying) performance hit we will get from overflow checks. We will almost certainly implement some sort of static analysis to determine which operations can be changed into field operations and which overflow checks can be eliminated, and I would be surprised if the difference in `usize` between architectures became an issue.

As long as `usize` becomes either `u32` or field integers when targeting MidenVM, then I don't think we'll notice.

otrho:
OK, I think I agree with all of the above @jjcnn. :slightly_smiling_face: When targeting the Fuel VM we promote all integers to a native `u64` anyway (even `u8`, which was easier initially, but we'll change sometime soon).  I feel like most of the time the Miden backend could do similarly, except to more efficient types for that architecture.  The main problem is the assumption in the existing Sway standard library that `u64` is cheap and fast.

So for this PR I feel like it should wait until we definitely have a genuine use-case, rather than adding it now assuming it will be required in the future.  Using library types or the appropriately sized builtin types and codegen manipulations may deem `usize` as unnecessary.

bitwalker:
My two cents: I think dismissing `usize` as a convenience for programmers to avoid thinking about bit-widths is mischaracterizing its purpose. The `usize` type expresses a few key properties (in my mind):

1. It represents the desire for an unsigned integer type of unspecified width, but at least as wide as the pointer-width of the target. This is useful for indices and such where the range corresponds to the amount of memory available to the program. As @jjcnn mentioned, if your maximum addressable memory is only 4GB, you don't need 64-bit pointers, array indices, etc.
2. It behaves the same way on all targets with regards to overflow/underflow, only the boundaries change.
3. The fact that the width is unspecified means that the representation chosen by the backend can be optimal for the target (and even context). For Miden, this choice might vary between u32 and a field integer, depending on what the context is and what information is available via static analysis. There are certainly many cases where you want to control the representation, but plenty where you don't really care, or you specifically want what is most natural for the target.

That said, if `usize`/`isize` aren't the "default" integer types for indices and otherwise unspecified integral values, then the value is diminished a bit. Still, I would think there are more than a few source languages that would want to defer the choice of integer representation to the backend, rather than needing to specify a type when they may not be in a position to make the best choice there.

> Say that usize is added, and say that it translates to different sizes on different architectures, specifically u32 for MidenVM and u64 for FuelVM. A programmer who uses usize must be in a situation where an overflow won't happen in u32, because if it did then the program would behave differently on FuelVM and MidenVM (for no good business-logic reason that I can see). So why does the programmer not just use u32 instead of usize? That would ensure the same behavior across all platforms (including aborting due to 32-bit overflows on 64-bit architectures, should the programmer have made a mistake), at no performance cost.

Because on a platform with more addressable memory, you want to be able to support using those additional resources without having to write the same function N times for N-many "native" bit-widths, or use conditional compilation, or be wasteful by using 64-bit values on targets that will never use the full range. Sure, if you know that regardless of what the target supports, the code needs a specific integer representation, then you'd use that rather than `usize`, but there are plenty of situations where you _don't_ know that, or you know that the values that need to be represented are guaranteed to fit in whatever the native integral range for the target is (e.g. array indices).

> As long as usize becomes either u32 or field integers when targeting MidenVM, then I don't think we'll notice.

I think that is the purpose of this PR, really, provide an integral type that specifically "allows" the compiler to choose which underlying type is used at runtime based on its own heuristics. In other words, it is safe to optimize `usize` by choosing the most efficient representation for the VM (i.e. a field integer), but you are much less free to do so with a specific bit-width, because you must assume that a specific bit-width was chosen for a reason (particularly in the case of 64-bit integers).

In any case, I see value in having such a type, but only if it has a natural place in the language. If it is treated like it shouldn't be used, or isn't used as a default integral type, then it probably isn't worth it. I'd really use Rust here as a reference moreso than C/C++, which are much more clunky in this regard (IMO).

otrho:
So, we often use a silly mantra when designing Sway features -- WWRD?  (What would Rust do?)  And as mentioned above, `usize` in Rust seems to me to be used either for very low level FFI/unsafe type areas or more commonly as indices for slices or arrays.

If we agree that indexing is the primary use for `usize` then we can add it (and update any library functions or intrinsics to use it for indexing) and make it a fixed 32bit size.  I agree that 64-bit indexing is not something that we need on blockchain today, although perhaps you never know, in 10+ years Sway may still be used where this is a thing.

But OTOH, there's no point in making a new type which is just  a `u32`.  So is the simple solution just to assume all indexing in Sway is 32bit, via `u32`s?

IGI-111:
I'm going to close this since the PR is getting stale, but if you feel like this really belongs in Sway, I think an RFC for it would be appropriate so people can evaluate the idea on the merits.

Issue #4162, opened 2023-02-22
Enable usage of injected contract ids from contract dependencies in multi contract calls 


kayagokalp:
from #4156:

> There is contract_a which declares contract_b as a contract dependency. Because of this relation forc-pkg inserts CONTRACT_ID constant into namespace and we can address the contract id of contract_b from contract_a using let id = contract_b::CONTRACT_ID;. Since these are both contracts that we want to test, forc-test injects CONTRACT_ID into their namespaces for them to be able to make contract calls. After this step using let id = contract_b::CONTRACT_ID; in contract_a would result id taking the contract id of contract_a. This looks like a bug with the way we resolve constants or the way we inject stuff into namespaces.

mitchmindtree:
> Contract dependency is an information we are holding in the edges of the build plan graph as it is a feature of a package relative to another package. This means that contract id of a contract dependency is relative to the package it is a contract dependency to. On the other hand the id we are injection for test execution is not relative to any other package.

Ahhhh I think I see the issue.

I think this indicates that we want to build each contract not just *once* without tests, but once for each time it is uniquely depended on by an edge from a dependent contract. Almost always this will still only be once, as even if a contract is depended on by more than one other contract, they'll almost certainly use the same salt anyway.

I don't think we should need a separate `TEST_CONTRACT_ID` - as long as we use the salt specified in the manifest, there should be no reason for the contracts (when built without tests) to have different IDs right?

kayagokalp:
> Almost always this will still only be once, as even if a contract is depended on by more than one other contract, they'll almost certainly use the same salt anyway. (Edit: I guess technically we shouldn't have to do the whole build again for each unique edge, just create a unique contract ID for each edge 🤔).

Yeah exactly! 

Currently this is not a problem as (if i recall correctly) we do not allow separate salt declarations for the same contract but I think we should, as a user might want to interact with two different deployments of the same contract. Each contract id injection should be done relative to the calling package so calling `pkg_a::CONTRACT_ID` from `pkg_b` which declares `pkg_a` as a contract dependency with zeroed salt, and from `pkg_c` which declares `pkg_a` as a contract dependency with non zero salt should yield different ids. Because that might be actually intentional. 

For example I would imagine there will be some fuel standards for NFTs and people deploy these contracts lots of time, I guess I might want to interact with two different NFTs deployed with different salts from my scripts?


[...]

PR #4219, opened 2023-03-04
Introduce `CallPath` to `TyConstantDeclaration`

### LINKED ISSUE #4217: Constants in the root module override other constants with the same name in the submodules
## Problem 

Constants in submodules overriden by the root module if they share the same name. To give an example:

main.sw:

```rust
contract;

dep pkga;

const TEST_CONST: u64 = 20;

abi MyContract {
    fn test_function() -> bool;
}

    fn test_function() -> bool {
        true
    }
}

#[test(should_panic)]
fn test_success() {
  assert(TEST_CONST == pkga::TEST_CONST)
}
```

pkga.sw:
```rust
library pkga;

pub const TEST_CONST: u64 = 10;
```

If `forc test` is run with this setup , the test would be failing but it shouldn't as it is a should revert test and `TEST_CONST` had to be different. Logging them shows that both of them is `20`, the value at the root submodule.

## Possible cause

We had a chat with @mitchmindtree yesterday and made a quick investigation in the compiler. It might be possible that problem is arising from `compile_const_decl` function which compiles const expression to a `Value`. It only accepts a constant name and some look-up variables which contains the namespace. The function then searches the namespace only using the name but in the example above although we are explicitly using  the constant from submodule the name is still `TEST_CONST`. This causes the symbol checking in the namespace to found the `TEST_CONST` in the root module (instead of `pkga::TEST_CONST`)

## Possible solution

`compile_const_decl`  should also be accepting the `path`, and module look-up function should accept the path and the name of the variable so that it can differentiate between the symbols that has the same name in root and submodules.

This is blocking me in #4156. Related to #4162

mohammadfawaz:
## Description
Closes #4217 

Main changes: 
* `TyConstantDeclaration` now contains a `CallPath` instead of an `Ident` to identify the path to and the name of the constant declaration.
* Use the full path, converted to a vector of strings, when looking up and adding constants in `const_eval.rs`. This helps identifying constants with the same name more uniquly.
* Amend the function `to_fullpath` to always complete the path even in situations where only some of the prefixes are present. This is useful for when expressions use a delineated path, especially in internal libraries. 

## Checklist


kayagokalp (approved):
Thank you very much for this! Tested by merging with #4156 and everything works! This also closes #4162 I opened it as I was thinking to merge #4156 before this fix

otrho (approved):

sway-lsp/src/traverse/typed_tree.rs
```diff
@@ -519,16 +519,10 @@ impl<'a> TypedTree<'a> {
                     self.collect_call_path_prefixes(&call_path.prefixes);
                 }
 
-                let span = if let Some(call_path) = call_path {
```
JoshuaBatty, line 522: Is there any reason to remove this check? We want to make sure if the call path looks something like `foo::bar::const` that the span only contains the `const` token and not the `foo::bar` span info.

mohammadfawaz: I see your point. What I did is update the span of VariableExpression to always point to the name of the const in the expression even if a call path is provided, so the check is no longer necessary here. But I'm now wondering it this is the right approach. The span of the suffix of the call paths may point to the const declaration after my changes. I think I can do better here though.

mohammadfawaz: Merging for now and will deal with the span thing tonight or tomorrow.

JoshuaBatty: Ok great, happy for this to be addressed in a follow up PR so @kayagokalp can be unblocked.

PR #4221, opened 2023-03-06
ci: check for `fuels` deps in `forc-pkg`

### LINKED ISSUE #3447: Add CI check to ensure `forc-pkg` never depends on any `fuels` crates (besides `fuels-types`)
This was a part of the resolution of the sway <-> fuels repo cycle.

Specifically, we determined that it's OK for the SDK to use `forc-pkg` to build its test suite under the condition that we are careful to ensure we never depend on `fuels` or any of its subcrates from `forc-pkg`. The exception is `fuels-types` as it has been moved to a dedicated repo and no longer causes a cycle.

The check can be something like:

1. Use something like `cargo-deps` to retrieve the dependency graph specifically for `forc-pkg`.
2. Check that none of the dependency crate names match `fuels` or `fuels-*` (allowing `fuels-types`).

We should document the check to explain why it is necessary, as the inter-repo cycle issue is perhaps not immediately obvious to new fuel devs.

bingcicle:
## Description

closes #3447

Failed the first commit deliberately just to [demonstrate the failure case](https://github.com/FuelLabs/sway/actions/runs/4341593516/jobs/7581384965) (there's a mistyped newline in that job, fixed in the following commit), added a little message just to explain why we may not want `fuels` deps in `forc-pkg` - let me know if it is adequately explained!

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -314,8 +314,8 @@ jobs:
           deps=$(cargo deps --manifest-path forc-pkg/Cargo.toml)
 
           case "$deps" in
-            *fuel*)
-              echo "\nFound dependency on a 'fuels' crate.
+            *fuels*)
+              echo -e "\nFound dependency on a 'fuels' crate.
           To avoid cyclic dependencies between 'fuels' and 'forc-pkg',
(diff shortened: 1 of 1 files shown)


eureka-cpu (approved):
LGTM

PR #4241, opened 2023-03-08
Adds visibility check for fully qualified function

### LINKED ISSUE #4231: Visibility qualifier ignored when fully qualified paths are used
For example, say we have a library `pkga` with a private function `foo`:
```rust
library pkga;

fn foo() { }
```
The following script fails to compile because `foo` is not `pub`:
```rust
script;

dep pkga;
use pkga::foo;

fn main() {
    foo();
}
```
However, this compiles fine:
```rust
script;

dep pkga;

fn main() {
    pkga::foo();
}
```

We should also check if this is a problem for other items that accept a visibility qualifier such as structs and enums.

esdrubal:
## Description
The issue was that it was possible to call private methods in another module by calling them with fully qualified paths.

The fix was to use resolve_call_path_with_visibility_check instead of resolve_call_path.

Closes #4231.

## Checklist


mohammadfawaz (dismissed):
Nice, thanks!

test/src/e2e_vm_tests/test_programs/should_fail/private_function_fully_qualified/json_abi_oracle.json
```diff
@@ -0,0 +1 @@
+[]
```
mohammadfawaz, line 1: Actually this file can probably be removed?

esdrubal: Good point, removed file.

mohammadfawaz (approved):

nfurfaro (approved):
LGTM.

PR #4256, opened 2023-03-09
Introduce `TypeSubstList` and add it to `TyDeclaration` and `DeclRef`.

emilyaherbert:
## Description

`TypeSubstList` is the necessary component for performing monomorphization after type checking and for strengthening trait constraint solving. At a high level, after all of these changes are complete, `DeclId`'s will be unique to each declaration, so for each declaration that the user writes in source code, there is only ever one `DeclId` that is used to refer to this declaration. Generic types will be extracted to the `TypeSubstList`, so for one declaration in source code, there will be one `DeclId`, but potentially multiple `TypeSubstList`. In particular `DeclRef` will hold a `TypeSubstList` with the generics for a particular `DeclId`:

```rust
/// Represents the use of / syntactic reference to a declaration. A
/// smart-wrapper around a [DeclId], containing additional information about a
/// declaration.
#[derive(Debug, Clone)]
pub struct DeclRef<I> {
    /// The name of the declaration.
    // NOTE: In the case of storage, the name is "storage".
    name: Ident,

    /// The index into the [DeclEngine].
    id: I,

    /// The type substitution list to apply to the `id` field for type
    /// monomorphization.
    subst_list: TypeSubstList,

    /// The [Span] of the entire declaration.
    decl_span: Span,
}
```

And the type system enforces that the `TyDeclaration` AST nodes have templated `TypeSubstList` with the `Template` type: 

```rust
/// An object that is a template for copies from the template.
///
/// This is predominantly used with
/// [TypeSubstList](crate::type_system::TypeSubstList) and
/// [TyDeclaration](crate::language::ty::TyDeclaration). The various
/// variants of [TyDeclaration](crate::language::ty::TyDeclaration) contain
/// fields `type_subst_list: Template<TypeSubstList>`. This type indicates that
/// the [TypeSubstList](crate::type_system::TypeSubstList) contained in this
/// field is simply a template for usages of the
/// declaration declared in that particular
/// [TyDeclaration](crate::language::ty::TyDeclaration) node.
#[derive(Clone, Debug)]
pub struct Template<T>(T)
where
    T: Clone;

impl<T> Template<T>
where
    T: Clone,
{
    pub(crate) fn new(value: T) -> Template<T> {
        Template(value)
    }

    pub(crate) fn into_inner(self) -> T {
        self.0
    }

    pub(crate) fn fresh_copy(&self) -> T {
        self.0.clone()
    }
}
```

```rust
    StructDeclaration {
        name: Ident,
        decl_id: DeclId<TyStructDeclaration>,
        type_subst_list: Template<TypeSubstList>,
        decl_span: Span,
    },
```

## Checklist


IGI-111 (commented):
Is it wise to add this to every DeclRef instead of creating a wrapper type for the type substitution?

I understand the goal is to make use of these substitutions everywhere but has thought been given to those cases where we do not want that?
One pointer may not be a lot but this is a very hot type.

We may also want to explicit cases where type substitution is present or not through the type. For generic vs actual types.

emilyaherbert:
> a wrapper type for the type substitution

Hmm, I'm not sure how I feel about this idea yet. Are you thinking something like this:

```rust
struct Typeable<I> {
  decl_ref: DeclRef<I>,
  subst_list: TypeSubstList
}

enum TyExpressionVariant {
  // ..,
  FnApplication {
    fn_ref: Typeable<DeclRef<DeclId<TyFunctionDeclaration>>>,
    // ..,
  },
  // ..,
}
```

> I understand the goal is to make use of these substitutions everywhere but has thought been given to those cases where we do not want that?

Can you give an example of a case like this? Some additional context---the "self type" is to abstracted to the `TypeSubstList` and the `TypeInfo::SelfType` to be removed. This means that structs, enums, traits, and abis, will always have at least one member in their `TypeSubstList`.

> We may also want to explicit cases where type substitution is present or not through the type. For generic vs actual types.

This is the distinction between `DeclRef` and `DeclId`. `DeclRef` contains a `TypeSubstList` holding the (potential) generics which may, at any point in compilation, be unified or substituted with other types. And `DeclId` does not. Once a `DeclId` is created, there will be no substitution to the contents of that `DeclId`.

IGI-111 (approved):
Okay that explanation makes sense. In that case LGTM.

mohammadfawaz (approved):

PR #4260, opened 2023-03-10
More intrinsics and `const_eval`

### LINKED ISSUE #407: Constant value evaluation in the AST needed.

vaivaswatha:
## Description
New intrinsics and constant evaluations supported by this PR can be seen in the additions to [const_inits](https://github.com/FuelLabs/sway/compare/vaivaswatha/const_eval_407?expand=1#diff-5886ff784ee10cb93be850eba9966c80d539bf6c08eaeb7d286183494690c610) test.

 Closes #407.

## Checklist


mohammadfawaz (commented):
This seems very reasonable, thanks!

Do you mind adding documentation for the new intrinsics in https://github.com/FuelLabs/sway/blob/master/docs/book/src/reference/compiler_intrinsics.md?

mohammadfawaz (approved):

PR #4268, opened 2023-03-12
ci: strip release binaries

### LINKED ISSUE #4236: ci: strip release binaries
```console
$ file forc
forc: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=06be24deeabfc888d1d7274358f488ab8b922e47, with debug_info, not stripped
```

Currently, the forc binaries distributed within this repo are not stripped like in the other repos ([forc-wallet](https://github.com/FuelLabs/forc-wallet/blob/41a0e30ac425e6b5c294cc877867d69be8c7e47b/.github/workflows/ci.yml#L188-L202), for example) This step should be added to the CI.

fgimenez:
## Description

Closes #4236 

Follows https://github.com/FuelLabs/forc-wallet/blob/41a0e30ac425e6b5c294cc877867d69be8c7e47b/.github/workflows/ci.yml#L188-L202 as suggested. The strip command is applied to the same binaries that are later archived in the `prep_assets` steps, I see a `forc-submit` binary in the target directory that is not included there, let me know it is should be stripped too. 

## Checklist


mitchmindtree (commented):
> I see a forc-submit binary in the target directory that is not included there, let me know it is should be stripped too.

That would be great, thanks! That binary is one of those produced by `forc-client` alongside `forc-deploy` and `forc-run`.

bingcicle:
Thank you for the PR!

> I see a `forc-submit` binary in the target directory that is not included there, let me know it is should be stripped too.

And thanks for catching this, seems like we did not include `forc_submit` in the last few latest releases since we added that as well.  :sweat_smile: We should probably also add that in the release step too, [here](https://github.com/FuelLabs/sway/pull/4268/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR675-R682)

bingcicle:
Thank you for the PR!

> I see a `forc-submit` binary in the target directory that is not included there, let me know it is should be stripped too.

And thanks for catching this, seems like we did not include `forc_submit` in the last few latest releases since we added that as well.  :sweat_smile: We should probably also add that in the release step too: https://github.com/FuelLabs/sway/blob/f6128f1c10d0a76e40b30b0eedc17cd8cd1f8ed0/.github/workflows/ci.yml#L675-L682

fgimenez:
> We should probably also add that in the release step too

Absolutely, done here https://github.com/FuelLabs/sway/pull/4268/commits/6347231bf9bb0b693e1842928371f1eeafd4dc85

bingcicle (approved):
LGTM - thank you!

otrho (approved):

PR #4277, opened 2023-03-14
Add LSP renaming functionality

### LINKED ISSUE #1418: Sway LSP should not allow renaming keywords or tokens from std
When a rename request is made, the `textDocument/prepareRename` method should check if the token belongs to the `std` library, eg `assert` or if the token is a language keyword. If so, then send a notification back to the client to cancel the rename request. 

We still want to gather the tokens from `std` as we need these for `go_to` requests.

### LINKED ISSUE #2261: Support the LSP rename capability

JoshuaBatty:
## Description
Enable the LSP renaming functionality for all tokens within the users workspace. Makes sure to not allow the renaming of keywords, intrinsics, or types defined outside of the workspace. 

Note, renaming of modules is not yet supported but I have an issue to follow up on this at #4289.

closes #2261
closes #1418

## Checklist


sdankel (commented):
This looks great! 

A few things I noticed while testing:

1. Sometimes it seems to time out while renaming and I get an error message that says "Canceled: Canceled". This happened if I click anywhere in the file while it's renaming.
1. Renaming a trait method does not rename the method inside impls
1. Same thing for renaming an ABI method
1. I tried renaming `assert` in `sway-lib-std`. It successfully completed the rename pretty quickly, but got stuck for 30s running the formatter.
1. I tried renaming a primitive type (`b256`) and it ran for a while before timing out. This seems like an easy fix (left a comment about it)
1. I tried renaming `copy_to` from within `sway-lib-std/src/bytes.sw`. It says the element can't be renamed, but I expected it to work because it's defined within my VSCode workspace (within the sway repo), even though it's in a different module. Is this expected?
1. I renamed `Option` from within `sway-lib-std/src/bytes.sw`, which I expected to rename Option everywhere because it's defined in `sway-lib-std`, but it only renamed it within the file.
1. Similar thing for renaming enum variants - I tried renaming `None` in `Option::None` and it only renamed within the file.

sway-lsp/src/capabilities/rename.rs
```diff
@@ -1,43 +1,136 @@
...
-    let mut map_of_changes = HashMap::new();
-    session.sync.to_workspace_url(url).map(|url| {
-        map_of_changes.insert(url, edits);
-        WorkspaceEdit::new(map_of_changes)
-    })
+    // Sort the TextEdits by their range in reverse order so the client
+    // applies edits from the end of the document to the beginning, preventing issues with offset changes.
+    for edits in map_of_changes.values_mut() {
+        edits.sort_unstable_by(|a, b| b.range.start.cmp(&a.range.start));
+    }
```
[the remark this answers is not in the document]
JoshuaBatty: Thanks this is much nicer!

sway-lsp/src/core/token_map.rs
```diff
@@ -44,11 +44,13 @@ impl TokenMap {
...
     ) -> impl 's + Iterator<Item = (Ident, Token)> {
         let current_type_id = token.declared_token_span(type_engine, decl_engine);
-
         self.iter()
             .filter(move |item| {
-                let ((_, _), token) = item.pair();
-                current_type_id == token.declared_token_span(type_engine, decl_engine)
+                let ((_, span), token) = item.pair();
+                let decl_span = token.declared_token_span(type_engine, decl_engine);
+                // This condition is true when the current item is the declaration of the token.
```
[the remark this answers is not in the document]
JoshuaBatty: sure, i'm happy with this. Does make it easier to read.

sway-lsp/src/capabilities/rename.rs
```diff
@@ -1,43 +1,136 @@
...
     session: Arc<Session>,
     new_name: String,
     url: Url,
     position: Position,
-) -> Option<WorkspaceEdit> {
-    let (_, token) = session.token_map().token_at_position(&url, position)?;
-    let mut edits = Vec::new();
+) -> Result<WorkspaceEdit, LanguageServerError> {
+    // Make sure the new name is not a keyword
+    if sway_parse::RESERVED_KEYWORDS.contains(&new_name) {
```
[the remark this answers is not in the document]
JoshuaBatty: rust-analyzer seems to allow you to do this so I think this is fine for now

JoshuaBatty:
> 1. Sometimes it seems to time out while renaming and I get an error message that says "Canceled: Canceled". This happened if I click anywhere in the file while it's renaming.

I've opened up #4418 so we can handle that in a separate PR.

> 2. Renaming a trait method does not rename the method inside impls
> 3. Same thing for renaming an ABI method

These should be working now. I've added these cases into the renaming test.

> 4. I tried renaming assert in sway-lib-std. It successfully completed the rename pretty quickly, but got stuck for 30s running the formatter.

Seems like the formatter is having to do a lot of work when a key function like `assert` is changed. I'll raise this during the tooling meeting. 

> 5. I tried renaming a primitive type (b256) and it ran for a while before timing out. This seems like an easy fix (left a comment about it)

I've made the `parse_int_suffix` function from `sway_parse` `pub` so we can make sure the new name isn't a literal type. 

```rust
if sway_parse::RESERVED_KEYWORDS.contains(&new_name)
        || sway_parse::parse_int_suffix(&new_name).is_some()
{
    ...
}
```

> 6. I tried renaming copy_to from within sway-lib-std/src/bytes.sw. It says the element can't be renamed, but I expected it to work because it's defined within my VSCode workspace (within the sway repo), even though it's in a different module. Is this expected?

Yeah this expected behaviour. We only allow renaming to happen at the local level of the users `Forc.toml` and not the whole workspace. 

> 7. I renamed Option from within sway-lib-std/src/bytes.sw, which I expected to rename Option everywhere because it's defined in sway-lib-std, but it only renamed it within the file.
> 8. Similar thing for renaming enum variants - I tried renaming None in Option::None and it only renamed within the file.

So both these cases actually work when I test them, i'm able to rename both variants and the Option type itself. Interestingly doing so seems to break any further LSP capabilities.

sway-lsp/src/core/token_map.rs
```diff
@@ -104,7 +119,7 @@ impl TokenMap {
         tokens
             .filter_map(|(ident, token)| {
                 let span = match token.typed {
-                    Some(TypedAstToken::TypedFunctionDeclaration(decl)) => decl.span(),
+                    Some(TypedAstToken::TypedDeclaration(decl)) => decl.span(),
```
sdankel, line 122: This breaks autocompletion. The test should pass if we change this to:

```suggestion
                    Some(TypedAstToken::TypedDeclaration(decl)) => decl.span(),
                    Some(TypedAstToken::TypedFunctionDeclaration(decl)) => decl.span(),
```

sway-lsp/src/traverse/typed_tree.rs
```diff
@@ -233,7 +233,7 @@ impl<'a> TypedTree<'a> {
                 // impl is registered
                 self.collect_type_id(
                     implementing_for.type_id,
-                    &TypedAstToken::TypedDeclaration(declaration.clone()),
+                    &TypedAstToken::TypedArgument(implementing_for.clone()),
```
sdankel, line 236: This change apparently breaks some of the struct code actions. Specifically, `all_references_of_token` used to return the `TypedAstToken::TypedDeclaration(TyDecl::ImplTrait)` given a struct decl. For example, the reason this test is failing:

```
struct A {
    a: u64,
    b: u64,
}

impl A {
    fn f(ref mut self) {
        self.a = 42;
        self.b = 77;
    }
}
```

We check if there are any existing impls for `A`, to see if we can insert the generated `new` function inside the existing impl ([code](https://github.com/FuelLabs/sway/blob/4e0291c48b23e047bbfab5a52ee4c4842e6cc1a6/sway-lsp/src/capabilities/code_actions/struct_decl/struct_new.rs#L21-L33)).

What's happening now seems to be that the token for the `TyImpl` is not getting inserted into the token map, so `all_references_of_token` is not returning it and we can't find it. The result is that the code action can only generate a `new` function inside of a new impl block, which causes the tests to fail.

I'm not sure of the best way to fix this - I see that renaming impls works on your branch, so maybe we could use the same mechanism for finding existing impls in `struct_new.rs` to make it work for code actions as well?

sdankel: Here's the difference:

failing tests

image

working (by changing this line back)

image

sdankel: Does changing this back break renaming?

sway-lsp/src/core/token_map.rs
JoshuaBatty: Thanks, i've changed this to

```rust
let span = match token.typed {
    Some(TypedAstToken::TypedFunctionDeclaration(decl)) if functions_only == Some(true) => decl.span(),
    Some(TypedAstToken::TypedDeclaration(decl)) => decl.span(),
    _ => ident.span(),
};
```

sway-lsp/src/traverse/typed_tree.rs
JoshuaBatty: thanks for looking into this! I've got the tests working again now. 

We now pass in a different typed_token in the `collect_type_id` depending on if `trait_decl_ref.is_some()`. If it isn't (in the case of a struct impl) then we pass through the original decl which contains the `ImplTrait`. Otherwise, we mark this token a `TypeArgument`.

So if we take the following:

```rust
abi MyContract {
    fn test_function();
}

    fn test_function() {}
}
```

In this case the `MyContact` token is the declaration, while the `Contract` token is the `TypeArgument` token. This allows us to one have one Declaration for the code block which is needed for the renaming logic, plus I think it makes more sense conceptually as well.

sdankel (approved):
🚢

mitchmindtree (approved):

PR #4287, opened 2023-03-14
feat: Introduce git SSH support to forc

### LINKED ISSUE #1636: Add proper git authentication handling to `forc-pkg`
Git is used by `forc` to enable support for remote git dependencies and remote project templates. The fetching and handling of git repositories is performed in `forc-pkg`.

## Problem

Currently we don't do any sort of authentication handling for git within `forc-pkg`.

So far this hasn't caused too many issues as all existing forc projects only have public repository dependencies and they all use `http`/`https` URLs. However, once we start opening up to more users I can see us quickly getting a bunch of issues about this.

The main issues with not handling git authentication are:

1. **URL types other than http/https don't seem to work without it**. I don't think git URLs like `git@github.com:FuelLabs/sway.git` work at the moment, and I believe this might be due to our lack of authentication handling.
2. **No support for private dependencies**. Normally git can support fetching and updating private repositories via ssh as long as the git URL format is used, however trying to do so with forc just panics at the moment.
3. **Some git URL rewrite rules don't work**. Some users use re-write rules so that URLs starting with `http`/`https` automatically get rewritten to git URLs. As a result of 1. forc will just panic. This is the source of this issue:
   - https://github.com/FuelLabs/sway/issues/998.

## Potential Solution

We should investigate how `cargo` handles this.

Some high-level docs about cargo's support for git authentication can be found here: https://doc.rust-lang.org/cargo/appendix/git-authentication.html.

Investigate cargo's [`git::utils::with_authentication` function](https://github.com/rust-lang/cargo/blob/b1636fc1e049e977ae542ec2b2191ffab075b049/src/cargo/sources/git/utils.rs#L448) which provides a scope with valid git auth. This function seems to be the primary part of `cargo`'s code-base responsible for handling git authentication.

kayagokalp:
## Description
closes #1636.

This PR introduces ssh authentication support for git sources. I also started to maintain `git2_auth` which I aim to make a feature complete authentication handler for git2-rs that supports non ssh-agent interaction as well. That would take a lot more time so I pulled the ssh-agent support and injected into `forc-pkg`. As I find out useful things while experimenting on `git2_auth`, will update our authentication handler with those as well.

1. We can now use dependencies in git ssh format 
Example:
```TOML
std = { git = "git@github.com:FuelLabs/sway.git" } 
```
2. We can also point to private repos

Follow-up:

- #4300

## Checklist


forc-pkg/src/source/git/mod.rs
```diff
@@ -310,10 +327,24 @@ impl From<Pinned> for source::Pinned {
     }
 }
 
+impl FromStr for GitUrl {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        if s.starts_with("git") {
+            // TODO: We should probably parse for and detect any errors with this as well.
+            Ok(Self::Ssh(s.to_string()))
```
mitchmindtree, line 336: Ahh, we might want to be careful about our variant names for this `GitUrl` type.

E.g. I think normal git also supports ssh URLs like `ssh://`, however it looks like these would currently end up in the `Http` variant which would be a little unintuitive :sweat_smile:

Git's URL handling is pretty diverse - here's their URL normalization testing: https://github.com/git/git/blob/77bd3ea9f54f1584147b594abc04c26ca516d987/t/t0110-urlmatch-normalization.sh

Perhaps we could use the [gitoxide URL representation](https://docs.rs/gix-url/0.16.0/gix_url/struct.Url.html) for validation rather than make our own `GitUrl` type?

forc-pkg/src/source/git/auth.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// Shows if the `AuthHandler` tried to make `SSH` authentication so far.
+    ssh_authentication_attempt: bool,
+    /// Shows if the `AuthHandler` tried to make `USER_PASS_PLAINTEXT` authentication so far.
+    plain_user_pass_attempt: bool,
+}
+
+impl AuthHandler {
+    /// Creates a new `AuthHandler` from all fields of the struct. If there are no specific reasons
+    /// not to, `default_with_config` should be prefered.
+    pub(crate) fn new(
```
mitchmindtree, line 13:
```suggestion
    fn new(
```

Let's make this private if it's not intended to be used elsewhere

forc-pkg/src/source/git/auth.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// Creates a new `AuthContext` with provided `git2::Config` and default values for other
+    /// context used during handling authentication callbacks.
+    pub(crate) fn default_with_config(config: git2::Config) -> Self {
+        let ssh_authentication_attempt = false;
+        let plain_user_pass_attempt = false;
+        Self::new(config, ssh_authentication_attempt, plain_user_pass_attempt)
+    }
+
+    pub fn handle_callback(
```
mitchmindtree, line 33: Could we maybe add some docs here about when the given callback function will be called? Judging by the use of the `ssh_authentication_attempt` and `plain_user_pass_attempt` `bool`s it looks like this can get called multiple times - is that right?

mitchmindtree (commented):
Great job digging into this @kayagokalp!

I haven't yet tested this with a private repo myself, but have left a few minor suggestions in the meantime.

forc-pkg/src/source/git/mod.rs
kayagokalp: Ahh that makes sense, I will look into it 👍

forc-pkg/src/source/git/auth.rs
kayagokalp: Yeah, the the remote callback that we register to `FetchOpts` calls this handler until it receives a valid authentication method or an explicit error. So without the flags in place users can end-up in infinite loops waiting to be obtain a sucessful SSH authentication

forc-pkg/src/source/git/mod.rs
```diff
@@ -228,6 +225,19 @@ impl source::DepPath for Pinned {
     }
 }
 
+impl AsRef<gix_url::Url> for Url {
+    fn as_ref(&self) -> &gix_url::Url {
+        &self.url
+    }
+}
```
mitchmindtree, line 232: Do we use this implementation anywhere? If not, perhaps we can leave this out? Might be good to avoid exposing the inner `gix_url::Url` if we don't need to.

mitchmindtree (approved):
Nice one Kaya, LGTM :rocket: Left one optional suggestion, but will leave it up to you.

forc-pkg/src/source/git/mod.rs
kayagokalp: I am using it for the `display` implementation but I can remove it as well, I originally added it since `gix_url::Url` provides nice utilities such as [this](https://docs.rs/gix-url/latest/gix_url/struct.Url.html#method.user) one, which might be useful in the future for more edge-case ssh authentication purposes (especially we might end-up checking the url for a username before trying the agent, I haven't added it as I need to experiment with it)

mitchmindtree: Yeah nice, I can see that maybe being useful. I would probably still omit this until we need it, but I'm happy either way

kayagokalp: All right, I am converting to draft and will remove it first thing. Thanks for the comment 🙌

JoshuaBatty (dismissed):

forc-pkg/src/source/git/mod.rs
```diff
@@ -12,14 +13,18 @@ use std::{
...
+    url: gix_url::Url,
+}
 
 /// A git repo with a `Forc.toml` manifest at its root.
 #[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Deserialize, Serialize)]
 pub struct Source {
     /// The URL at which the repository is located.
     pub repo: Url,
-    /// A git reference, e.g. a branch or tag.
+    // A git reference, e.g. a branch or tag.
```
bingcicle, line 27: Was this changed intentionally?

bingcicle (dismissed):
Looks nice! Just have 1 question

forc-pkg/src/source/git/mod.rs
kayagokalp: Oh nice catch thanks 👍

bingcicle (approved):

JoshuaBatty (approved):

PR #4292, opened 2023-03-15
Conditional compilation

AlicanC:
Adds support for conditional compilation based on `target`.

```rs
// Only compiled if we are targeting FVM
#[cfg(target = "fuel")]
const VALUE: u64 = 40;

// Only compiled if we are targeting EVM
#[cfg(target = "evm")]
const VALUE: () = ();
```


sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
```diff
@@ -101,6 +101,28 @@ fn item_to_ast_nodes(
...
     let attributes = item_attrs_to_map(context, handler, &item.attribute_list)?;
 
+    // Walk all the cfg attributes and check if the item should be included
+    if let Some(cfg_attrs) = attributes.get(&AttributeKind::Cfg) {
+        for cfg_attr in cfg_attrs {
+            for arg in &cfg_attr.args {
+                match arg.name.as_str() {
+                    "target" => {
+                        if let Some(sway_ast::Literal::String(value)) = &arg.value {
+                            if value.parsed != "evm" {
```
AlicanC, line 111: @mohammadfawaz, so here, `"evm"` shouldn't be hardcoded, but read from the [BuildConfig](https://github.com/FuelLabs/sway/blob/a0150c0c5e2ff3a6fed54a300a131f829a4b9887/sway-core/src/build_config.rs#L17-L28). Should I just put the config in [Context](https://github.com/FuelLabs/sway/blob/a0150c0c5e2ff3a6fed54a300a131f829a4b9887/sway-core/src/transform/to_parsed_lang/context.rs#L2-L14)? Maybe this isn't the right place to evaluate the `cfg` attribute? (The [test](https://github.com/FuelLabs/sway/blob/a0150c0c5e2ff3a6fed54a300a131f829a4b9887/test/src/e2e_vm_tests/test_programs/should_pass/conditional_compilation/src/main.sw) passes so it works, but maybe there's a better place?)

mohammadfawaz: Passing in `BuildConfig` to `Context` is not a bad idea I think. It's not unusual that you'll have to read flags or fields from the manifest file and pass them to the compiler in various places. [Here's](https://github.com/FuelLabs/sway/blob/1a2d323e9c26b7f65bcc81d66a5acbb4c5459a6c/sway-core/src/lib.rs#L359) an example.

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
```diff
@@ -3933,3 +3988,54 @@ fn error_if_self_param_is_not_allowed(
...
+                        } else {
+                            let error = ConvertParseTreeError::ExpectedCfgTargetArgValue {
+                                span: arg.span(),
+                            };
+                            return Err(handler.emit_err(error.into()));
+                        }
+                    }
+                    _ => {
+                        // Already checked with `AttributeKind::expected_args_*`
+                        unreachable!("cfg attribute should only have the target argument");
```
mohammadfawaz, line 4034: Would certainly be nice to have negative tests here to make sure we error out nicely.

test/src/e2e_vm_tests/test_programs/should_pass/conditional_compilation/src/main.sw
```diff
@@ -0,0 +1,102 @@
...
+  CFG_VALUE: () = (),
+}
+
+#[cfg(target = "fuel")]
+const VALUE: u64 = 40;
+#[cfg(target = "evm")]
+const VALUE: () = ();
+#[cfg(target = "fuel")]
+#[cfg(target = "evm")]
+const VALUE: () = ();
```
mohammadfawaz, line 19: What is the desired behavior when both targets are specified?

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
mohammadfawaz, line 4034: Would certainly be nice to have negative tests here to make sure we error out nicely.

AlicanC: Added two fail tests.

test/src/e2e_vm_tests/test_programs/should_pass/conditional_compilation/src/main.sw
AlicanC: They are AND'd so the attributed never compiles. Added comments in the test to clear it.

test/src/e2e_vm_tests/test_programs/should_pass/conditional_compilation/Forc.toml
```diff
@@ -0,0 +1,7 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+implicit-std = false
+license = "Apache-2.0"
+name = "conditional_compilation"
+target = "evm"
```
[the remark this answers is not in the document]
AlicanC: Added a test, but not as you described. I think it does the same thing tho.

otrho (approved):

mohammadfawaz (approved):

PR #4294, opened 2023-03-15
Manually clone `self` in `subtract` for `U256` and `pow` for `U128`

### LINKED ISSUE #4293: `U256` subtract method causes an overflow
The problem shows up in this design: https://github.com/FossilFrank/reproduce-fuel-issue which has all the information needed to build and test. The symptom is a `MemoryOverflow` panic receipt. 

I tracked this down and it seems to be coming from [this line](https://github.com/FossilFrank/reproduce-fuel-issue/blob/d7bc820c664a92a53edb8e00605fd2da6aeb52f9/src/stableswap.sw#L116). Interestingly, changing that line to:
```rust
let x = _y - _dy;
_y = x;
```
fixes the problem. Therefore, this tells me that there is an issue in https://github.com/FuelLabs/sway/blob/ee0a0fd9ea0f401904cd3dbff95268a3c7bf21fb/sway-lib-std/src/u256.sw#L369 related to reading and writing the same variable `_y`.

mohammadfawaz:
## Description
Closes #4293

Basically, `self` has to be cloned before returned. I found that `U128::pow` has a similar problem so I fixed that too.

Anyways, I opened a separate issue to think about the compiler side of this and whether this behavior makes sense. It's certainly a bit unintuitive.

## Checklist


Braqzen (approved):

otrho (approved):
Why did this manifest now?  This will also be interesting to watch with my branch where we're much stricter with pointers and copying.

PR #4296, opened 2023-03-15
Make string representations of build targets consistent

AlicanC:
We had two implementations for converting a string to a `BuildTarget` and they didn't match. (One accepted "miden-vm" or "midenvm" and the other "miden" or "midenvm".) Now only "midenvm" is accepted and we use the same codepath everywhere.

### DIFF
--- a/sway-core/src/build_config.rs
+++ b/sway-core/src/build_config.rs
@@ -1,14 +1,34 @@
 use std::{path::PathBuf, sync::Arc};
 
 use serde::{Deserialize, Serialize};
+use strum::EnumString;
 
 #[derive(
-    Clone, Copy, Debug, Default, Eq, PartialEq, Hash, Serialize, Deserialize, clap::ValueEnum,
+    Clone,
+    Copy,
+    Debug,
+    Default,
+    Eq,
+    PartialEq,
+    Hash,
+    Serialize,
+    Deserialize,
+    clap::ValueEnum,
+    EnumString,
 )]
 pub enum BuildTarget {
     #[default]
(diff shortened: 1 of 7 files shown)


sway-core/src/build_config.rs
```diff
@@ -1,14 +1,34 @@
...
     Fuel,
+    #[serde(rename = "evm")]
+    #[clap(name = "evm")]
+    #[strum(serialize = "evm")]
     EVM,
+    #[serde(rename = "midenvm")]
+    #[clap(name = "midenvm")]
+    #[strum(serialize = "midenvm")]
     MidenVM,
 }
```
AlicanC, line 33: @mohammadfawaz, this looks really crowded TBH. I thought it would be cool if we could just configure Serde and have `clap` and `strum` use that, but couldn't find a good way. Thoughts?

mohammadfawaz: Can we not write this instead?
```rust
#[default, serde(rename = "fuel"), clap(name = "fuel"), strum(serialize = "fuel")]
```

AlicanC: I don't think Rust does that. Not a fan that Sway supports it TBH.

mohammadfawaz: Oh - did we make that up in Sway? 😆 

Anyways, the main reason for doing this in Sway was that many contract methods needed `storage()` annotations and `payable` so having those on line seems a lot cleaner than the alternative.

In your case here, I don't mind the three annotations like that.. If it works, go for it. This is not user facing.

PR #4297, opened 2023-03-15
Experimental: `StorageKey` feature

mohammadfawaz:
## Description
This PR introduces `StorageHandle` to the `core` as an **experimental** feature. `StorageHandle` is a descriptor of a location in storage containing two fields:
1. A key pointing to a storage slot `s`.
2. An offset pointing to a particular word in `s` or in its subsequent slots. This offset is new and is not described in the RFC. I will rectify that shortly.
The standard library introduces helper methods `read`, `try_read`, and `write` to `StorageHandle` that allow reading from and writing to the location pointed to by the handle. `try_read` returns an `Option` wrapping the data while `read` automatically internally unwraps the data and could panic.

The summary of this change is as follows:
* New struct `core::experimental::storage::StorageHandle`
* New storage library `std::experimental::storage::*` that should eventually replace `std::storage::*`. This new library mirrors the old one to the extent possible and introduces the helper methods for `StorageHandle`. Other data structures such as `StorageVec` and `StorageBytes` will be introduced in a separate PR.
* Add an experimental flag `--experimental-storage` to `forc` that enables all the required code in the compiler to support `StorageHandle` as described in the https://github.com/FuelLabs/sway-rfcs/pull/23.
  * Type checking phases assumes the existence of `StorageHandle` and monomorphizes it as needed.
  * Storage accesses now always return `StorageHandle` and storage reassignment are no longer relevant.
  * IR gen handles storage accesses by "filling" the key and the offset in an aggregate representing the `StorageHandle`. The key and the offset are statically determined based on the index of the storage variable in the `storage` block and the field accessed, if applicable.
  * Storage initializers are now handled based on the new model
  * The new approach packs struct fields by default but does not pack across storage variables. This offers the most amount of flexibility. This is a deviation from the RFC which I will update shortly.
  * To implement `StorageMap` and other dynamic storage data structures, we now write `impl StorageHandle<StorageMap<K, V>>` instead of `impl StorageMap<K, V>` directly. Also, note that the `get` method now returns a `StorageHandle` instead of the actual data. To get the actual data, `read()` or `try_read()` should be called on the resulting handle. This is needed for multiple reasons including proper support for nested dynamic storage types. Rust also does the same, in a way (where `get` actually returns ` &` which happens to coerce to the real data in certain places).
* I added various tests but they're not comprehensive. Some tests on my list:
  * Extensive tests for storage maps in structs (which now work!)
  * Extensive tests for storage accesses with various types and struct layouts

I still need to figure out how to do nested dynamic storage types with this approach. The stuff I have in `map_in_map_access` in `test_projects/experimental_storage/src/main.sw` is just an attempt but not ergonomic at all of course.

## Checklist


sway-core/src/ir_generation/function.rs
```diff
@@ -2499,17 +2515,39 @@ impl<'eng> FnCompiler<'eng> {
...
+                if (field_idx as u64) >= *index {
+                    ty_at_idx = field_type;
+                    break;
+                }
+                offset += ir_type_size_in_bytes(context, &field_type);
+            }
+        }
+        Ok(offset / 8)
+    }
+
```
otrho, line 2539: This looks a lot like `aggregate_idcs_to_field_layout()` in `asm_generation`.  The `ir_type_size_in_bytes()` is already stolen from there so you might as well re-use the former.

It's a bit unfortunate that this type-sizing stuff is in IR though, but I understand it's already there, nothing new from this branch.  Ideally we wouldn't be thinking about memory layout until `asm_generation`.   This is a Fuel specific operation though, so assumptions can be made, for now.

sway-core/src/ir_generation/function.rs
```diff
@@ -2579,39 +2635,100 @@ impl<'eng> FnCompiler<'eng> {
...
+                        .current_block
+                        .ins(context)
+                        .get_local(storage_handle_ptr)
+                        .add_metadatum(context, span_md_idx);
+
+                    // Insert the key as the first field in the `StorageHandle` struct
+                    storage_handle = self
+                        .current_block
+                        .ins(context)
+                        .insert_value(storage_handle, storage_handle_aggregate, key_val, vec![0])
```
otrho, line 2676: Argh, this is going to conflict _hard_ with my refactor...

otrho (approved):

sway-core/src/ir_generation/function.rs
mohammadfawaz: With regards to target-specific memory layout in IR, LLVM has an analysis pass called [`DataLayout`](https://llvm.org/doxygen/classllvm_1_1DataLayout.html) which is target dependent (relies on the "target triple" in IR) and allows you to query any all kinds of information about how various data types, such as structs, are laid out in memory. This can be quite useful for many optimization passes (even SROA). 

I guess what I'm saying is that querying memory layout in IR (or even in IRGen) is not unheard of.

otrho (dismissed):

mohammadfawaz:
Ah conflicts.. let me fix.

otrho (approved):

sway-lib-core/src/raw_ptr.sw
```diff
@@ -37,6 +37,14 @@ impl raw_ptr {
         };
     }
 
+    /// Copies `count * size_of<T>` bytes from `self` to `dst`.
```
nfurfaro, line 40: This comment looks out of place?

sway-lib-std/src/experimental/storage.sw
```diff
@@ -0,0 +1,257 @@
...
+/// let five = 5_u64;
+/// write(ZERO_B256, 2, five);
+/// let stored_five = read::<u64>(ZERO_B256, 2);
+/// assert(five == stored_five);
+/// ```
+#[storage(read)]
+pub fn read<T>(key: b256, offset: u64) -> Option<T> {
+    // NOTE: we are leaking this value on the heap.
+    // Get the number of storage slots needed based on the size of `T`
+    let number_of_slots = (offset * 8 + __size_of::<T>() + 31) >> 5;
```
nfurfaro, line 73: Why `offset * 8` here?
Isn't `offset` already in words?

nfurfaro (commented):
looks great! I just left a couple comments/questions.

sway-lib-std/src/experimental/storage.sw
mohammadfawaz: `offset` is in words - multiplying it by 8 converts it to bytes which is the same unit as `__size_of::<T>()`.

otrho (approved):

sway-lib-std/src/experimental/storage.sw
```diff
@@ -0,0 +1,287 @@
...
+    ///         offset: 2,
+    ///     };
+    ///
+    ///     // Reads the third word from storage slot with key 0x000...0
+    ///     let x:u64 = r.read();
+    /// }
+    /// ```
+    #[storage(read)]
+    pub fn read(self) -> T {
+        read::<T>(self.key, self.offset).unwrap()
```
mitchmindtree, line 141:
```suggestion
        self.try_read().unwrap()
```

sway-lib-std/src/experimental/storage.sw
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    /// Reads a value of type `T` starting at the location specified by `self`. If the value
+    /// crosses the boundary of a storage slot, reading continues at the following slot.
+    ///
+    /// Returns `Option(value)` if a the storage slots read were valid and contain `value`.
```
mitchmindtree, line 147:
```suggestion
    /// Returns `Some(value)` if a the storage slots read were valid and contain `value`.
```

sway-lib-std/src/experimental/storage.sw
```diff
@@ -0,0 +1,287 @@
...
+    /// None
+    ///
+    /// ### Examples
+    ///
+    /// ```sway
+    /// fn foo() {
+    ///     let r: StorageHandle<u64> = StorageHandle {
+    ///         key: 0x0000000000000000000000000000000000000000000000000000000000000000,
+    ///         offset: 2,
+    ///     };
```
mitchmindtree, line 133: Do we want to allow users to construct arbitrary storage handles? It might limit our ability to provide some more ergonomic API for known-valid storage keys in the future? I guess that ship has already sailed though as we publicly expose the `write` and `clear` functions (without any kind of `unsafe` indicator) anyways.

Maybe we should at least keep the fields conservatively private for now and offer a `StorageHandle::from_key_and_offset` constructor, in case we decide we need to add something like a `_type: core::marker::PhantomData<T>` or similar in the future (which would otherwise be a breaking change)?

sway-lib-std/src/experimental/storage.sw
```diff
@@ -0,0 +1,287 @@
...
+    ///
+    ///     // Reads the third word from storage slot with key 0x000...0
+    ///     let x:Option<u64> = r.try_read();
```
mitchmindtree, line 164:
```suggestion
    ///     let x: Option<u64> = r.try_read();
```
I *think* both the rust formatter and ours would make this change.

sway-lib-std/src/experimental/storage.sw
```diff
@@ -0,0 +1,287 @@
...
+    /// storage {
+    ///     map: StorageMap<u64, bool> = StorageMap {}
+    /// }
+    ///
+    /// fn foo() {
+    ///     let key = 5_u64;
+    ///     let value = true;
+    ///     storage.map.insert(key, value);
+    ///     let retrieved_value = storage.map.get(key);
+    ///     assert(value == retrieved_value);
```
mitchmindtree, line 219:
```suggestion
    ///     let retrieved_value = storage.map.get(key).read();
    ///     assert(value == retrieved_value);
```

sway-lib-std/src/experimental/storage.sw
```diff
@@ -0,0 +1,287 @@
...
+    /// fn foo() {
+    ///     let key = 5_u64;
+    ///     let value = true;
+    ///     storage.map.insert(key, value);
+    ///     let retrieved_value = storage.map.get(key).read();
+    ///     assert(value == retrieved_value);
+    /// }
+    /// ```
+    #[storage(read)]
+    pub fn get(self: StorageHandle<Self>, key: K) -> StorageHandle<V> {
```
mitchmindtree, line 251: Nice, I think returning the handle here makes a lot of sense and probably provides the most flexibility :+1:

mitchmindtree (commented):
Epic work! Left a few small Qs and suggestions, but mostly lgtm.

Also just a reminder we can make contract calls in unit tests now which might allow us to avoid some of the sdk harness boilerplate in our testing - that said I'm sure there will still be cases where we want to test with the SDK either way.

sway-lib-core/src/experimental/storage.sw
```diff
@@ -0,0 +1,9 @@
+library;
+
+/// Describes a location in storage specified by the `b256` key of a particular storage slot and an
+/// offset, in words, from the start of the storage slot at `key`. The parameter `T` is the type of 
+/// the data to be read from or written to at `offset`.
+pub struct StorageHandle<T> {
```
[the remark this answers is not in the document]
mohammadfawaz: Yeah I'm on the fence about this. Indeed the offset is what tripped me. We could argue that `StorageKey` is shorter though and probably clearer so I'll just revert to that.

sway-core/src/semantic_analysis/ast_node/declaration/declaration.rs
```diff
@@ -209,20 +212,84 @@ impl ty::TyDecl {
...
+                // Set of all type IDs that have corresponding items implemented for them in this
+                // `impl` block. In most cases, this is a single type ID that is equal to
+                // `impl_trait.implementing_for.type_id`.
+                let mut processed_ids = HashSet::new();
+
+                // This is a map from type IDs to a vector of impl items. In most cases, all the
+                // items are available to `impl_trait.implementing_for.type_id` (i.e. `Self`).
+                // However, there are situations where `self`, in some of the impl methods, has a
+                // type ascription that is different from `Self`. In that case, we have to keep
+                // track of which impl items are available to which type ID.
```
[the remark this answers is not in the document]
mohammadfawaz: I was somewhat under the impression that this does not work in Sway just yet 🤔 but, to my surprise, it does seem to work fine without any issues. As you said, we can introduce rules around allowing `impl` for upstream generic types as long as their generic input type is filled with a type from the current package. It might just be okay to deviate from Rust in that regard.

Let me nuke all the code related to `self` type ascriptions, or at least shelve it in case we need it in the future.

sway-lib-std/src/experimental/storage.sw
mohammadfawaz: I don't believe we have the ability to keep struct fields private at the moment, though I certainly agree with you that we shouldn't allow constructing arbitrary storage keys in the future.

sway-lib-std/src/experimental/storage.sw
mohammadfawaz: Ugh I wish this would work. Methods referencing each other from the same `impl` block does not work yet 😢

sway-core/src/semantic_analysis/ast_node/declaration/function/function_parameter.rs
```diff
@@ -41,6 +46,37 @@ impl ty::TyFunctionParameter {
...
+                TypeInfo::Struct(decl_ref) => {
+                    let struct_decl = decl_engine.get_struct(&decl_ref);
+                    if !(struct_decl.call_path.prefixes
+                        == vec![
+                            Ident::new_no_span("core".into()),
+                            Ident::new_no_span("experimental".into()),
+                            Ident::new_no_span("storage".into()),
+                        ]
+                        && struct_decl.call_path.suffix
+                            == Ident::new_no_span("StorageHandle".into()))
```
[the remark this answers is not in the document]
mohammadfawaz: We hardcode similar paths in the compiler already such as `core::ops::*`. We should probably see how Rust handles the situation you mentioned.

sway-core/src/semantic_analysis/ast_node/declaration/declaration.rs
```diff
@@ -209,20 +212,84 @@ impl ty::TyDecl {
...
+                            // Other items are only available for `Self`
+                            (impl_trait.implementing_for.type_id, item)
+                        }
+                    })
+                    .fold(
+                        BTreeMap::new(),
+                        |mut acc: BTreeMap<TypeId, Vec<_>>, (type_id, item)| {
+                            acc.entry(type_id)
+                                .and_modify(|v| v.push(item.clone()))
+                                .or_insert(vec![item.clone()]);
```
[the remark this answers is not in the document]
mohammadfawaz: No longer relevant 😄

otrho (approved):

otrho (approved):

sway-lib-core/src/experimental/storage.sw
```diff
@@ -0,0 +1,9 @@
+library;
+
+/// Describes a location in storage specified by the `b256` key of a particular storage slot and an
+/// offset, in words, from the start of the storage slot at `key`. The parameter `T` is the type of 
+/// the data to be read from or written to at `offset`.
+pub struct StorageKey<T> {
+    key: b256,
```
mitchmindtree, line 7: Maybe we should rename this to `slot`?

```suggestion
    slot: b256,
```

Perhaps this way we can more clearly distinguish between the terms "slot" and "key" throughout our code and documentation? E.g.

- **storage slot**: the `b256`-addressed storage slot.
- **storage key**: a location within a specific storage slot as described by `offset`.

sway-lib-std/src/experimental/storage.sw
```diff
@@ -0,0 +1,287 @@
...
+/// ### Examples
+///
+/// ```sway
+/// let five = 5_u64;
+/// write(ZERO_B256, 2, five);
+/// let stored_five = read::<u64>(ZERO_B256, 2).unwrap();
+/// assert(five == stored_five);
+/// ```
+#[storage(read, write)]
+pub fn write<T>(key: b256, offset: u64, value: T) {
```
mitchmindtree, line 26:
```suggestion
pub fn write<T>(slot: b256, offset: u64, value: T) {
```

sway-lib-std/src/experimental/storage.sw
```diff
@@ -0,0 +1,287 @@
...
+/// ### Examples
+///
+/// ```sway
+/// let five = 5_u64;
+/// write(ZERO_B256, 2, five);
+/// let stored_five = read::<u64>(ZERO_B256, 2);
+/// assert(five == stored_five);
+/// ```
+#[storage(read)]
+pub fn read<T>(key: b256, offset: u64) -> Option<T> {
```
mitchmindtree, line 70:
```suggestion
pub fn read<T>(slot: b256, offset: u64) -> Option<T> {
```

sway-lib-std/src/experimental/storage.sw
```diff
@@ -0,0 +1,287 @@
...
+///
+/// ```sway
+/// let five = 5_u64;
+/// write(ZERO_B256, 0, five);
+/// let cleared = clear::<u64>(ZERO_B256);
+/// assert(cleared);
+/// assert(read::<u64>(ZERO_B256, 0).is_none());
+/// ```
+#[storage(write)]
+pub fn clear<T>(key: b256) -> bool {
```
mitchmindtree, line 106:
```suggestion
pub fn clear<T>(slot: b256) -> bool {
```

mitchmindtree (dismissed):
Looking good to me! Left some minor suggestions, but will leave it up to you if you think they're worth addressing.

sway-lib-core/src/experimental/storage.sw
mohammadfawaz: I will address this in a follow up PR where I'm adding the rest of the storage library.

JoshuaBatty (approved):

mitchmindtree (approved):

PR #4298, opened 2023-03-16
refactor: deprecate `ConfigTimeConstant`s

### LINKED ISSUE #3681: Deprecate the first version of configuration-time constant
This can be done once https://github.com/FuelLabs/sway/pull/3536 is merged and the tooling around it is supported (via `forc` and the SDK).

bingcicle:
## Description

Closes #3681 

I meant to tackle #3077 first, but tackling #3681 first cleans up a lot of the code needed to make #3077 easier.

This PR is mostly cleanup:
- Completely removes usage of `ConfigTimeConstant` everywhere (in forc-pkg, sway-core, sway-type)
- **(BREAKING)** Completely removes support for `[constants]` table in the manifest, which was using the `ConfigTimeConstant` struct. This change is breaking for app devs ([open issue](https://github.com/FuelLabs/sway-applications/issues/375)) and from a [quick search](https://github.com/FuelLabs/sway-applications/search?l=TOML&q=%5Bconstants%5D) on our sway-applications repo, there's a number of manifests still using the old version of configuration time constants.

Consequentially, the above change also allowed us to cut down on other constructs like `ConstInjectMap` within forc-pkg, since the only thing we need to inject now are const CONTRACT_IDs - this cuts down quite a bit of now-redundant code.

- As a result of the removal of `ConfigTimeConstant`, a few tests using that feature were also deleted.
- Also updated the `configurable` example to be known as `configurable_constants` instead of `config_time_constant`.

**This PR does not:**
- Fix the manual creation of `CONTRACT_ID` const. To create the const, we still undergo manual parsing and type-checking and finally inserting into a `SymbolMap` manually. This will be handled in a separate PR to tackle #3077 

Other than removal of a lot of code to do with CTC, the main change here that allows contract dependencies to still work here is instead of relying on the `ConstInjectMap` which we used to inject both constants and contract dependencies, there is now only a single contract ID value that we require (this is abstracted as `ContractIdConst`) to inject into the contract for tests.

## Checklist


sway-core/src/semantic_analysis/namespace/module.rs
```diff
@@ -52,110 +51,99 @@ pub struct Module {
...
-            // Temporarily disallow non-literals. See https://github.com/FuelLabs/sway/issues/2647.
-            let has_literal = match &const_decl.value {
-                Some(value) => {
-                    matches!(value.kind, ExpressionKind::Literal(_))
-                }
-                None => false,
-            };
+        // FIXME(Centril): Stop parsing. Construct AST directly instead!
+        // parser config
+        let const_item = format!("pub const CONTRACT_ID: b256 = {contract_id_value};");
```
bingcicle, line 83: This line is probably the only notable change in this entire block, but the unindentation (since we're no longer looping through a bunch of constants) is displaying this entire block as diff'd

forc-pkg/src/pkg.rs
```diff
@@ -1518,14 +1505,18 @@ pub fn dependency_namespace(
     compiled_contract_deps: &CompiledContractDeps,
     graph: &Graph,
     node: NodeIx,
-    constants: BTreeMap<String, ConfigTimeConstant>,
     engines: Engines<'_>,
+    contract_id_value: Option<ContractIdConst>,
```
mitchmindtree, line 1509: Could we add to this function's doc comment an explanation of when to specify this value? E.g. something like:

> `contract_id_value` should only be `Some` when producing the `dependency_namespace` for a contract with tests enabled. This allows us to provide a contract's `CONTRACT_ID` constant to its own unit tests.

sway-core/src/semantic_analysis/namespace/module.rs
```diff
@@ -52,110 +51,97 @@ pub struct Module {
 }
 
 impl Module {
-    pub fn default_with_constants(
+    pub fn default_with_contract_id(
```
mitchmindtree, line 54: Let's also add a similar comment here, otherwise I imagine this might confuse compiler-devs quite a lot, as sway-core really shouldn't know anything about contract IDs :sweat_smile:! We might also want to add a link to the issue which states how we plan to refactor this out of sway-core in favour of creating temporary package dependencies for providing these `CONTRACT_ID`-containing modules.

sway-error/src/error.rs
```diff
@@ -582,10 +582,10 @@ pub enum CompileError {
...
     #[error("\"continue\" used outside of a loop")]
     ContinueOutsideLoop { span: Span },
-    #[error("Configuration-time constant value is not a constant item.")]
-    ConfigTimeConstantNotAConstDecl { span: Span },
-    #[error("Configuration-time constant value is not a literal.")]
-    ConfigTimeConstantNotALiteral { span: Span },
+    #[error("Contract ID is not a constant item.")]
+    ContractIdConstantNotAConstDecl { span: Span },
+    #[error("Contract ID value is not a literal.")]
+    ContractIdValueNotALiteral { span: Span },
```
mitchmindtree, line 588: Let's also add a small comment that says these will be removed in the future, linking to our issue.

mitchmindtree (commented):
Nice, looking good to me! Just a few suggestions for doc comments that I think could help.

Also would be good to double check with @kayagokalp that we can remove the const inject map? I think we previously used it to insert `CONTRACT_ID`s for tests, but as of a recent refactor related to multi-contract calls, I think we no longer need it?

bingcicle:
> Also would be good to double check with @kayagokalp that we can remove the const inject map? I think we previously used it to insert CONTRACT_IDs for tests, but as of a recent refactor related to multi-contract calls, I think we no longer need it?

Good call! Requested @kayagokalp specifically for a review too. At a glance it seems like the PR for multi-contract calls don't conflict with removal of the const inject map, but i'll wait for Kaya's input on this.

kayagokalp:
Yeah, a recent refactor to re-use compiled packages between `forc-pkg` and `forc-test` made the `injection_map` removable. I will have a closer look today for review 👌

mohammadfawaz (approved):
Looks good to me. Thanks!

forc-pkg/src/pkg.rs
```diff
@@ -1513,19 +1500,26 @@ pub const CONTRACT_ID_CONSTANT_NAME: &str = "CONTRACT_ID";
...
+/// `contract_id_value` should only be Some when producing the `dependency_namespace` for a contract with tests enabled.
+/// This allows us to provide a contract's `CONTRACT_ID` constant to its own unit tests.
 pub fn dependency_namespace(
     lib_namespace_map: &HashMap<NodeIx, namespace::Module>,
     compiled_contract_deps: &CompiledContractDeps,
     graph: &Graph,
     node: NodeIx,
-    constants: BTreeMap<String, ConfigTimeConstant>,
     engines: Engines<'_>,
+    contract_id_value: Option<ContractIdConst>,
```
kayagokalp, line 1512: If you prefer removing the `Option` in the function params we can introduce something like:

```rust
enum NamespaceKind {
  Build,
  Test(ContractIdConst)
}
```
and then we can refactor `pub fn dependency_namespace()` to 

```rust
pub fn dependency_namespace(
  lib_namespace_map: &HashMap<NodeIx, namespace::Module>,
  compiled_contract_deps: &CompiledContractDeps,
  graph: &Graph,
  node: NodeIx,
  engines: Engines<'_>,
  namespace_type: NamespaceKind
)
```

and match on `NamepsaceKind::Test(contract_id)` like:

```rust
let mut namespace = if let NamepsaceKind::Test(contract_id_value) = namespace_type {
  namespace::Module::default_with_contract_id(engines, name.clone(), contract_id_value)?
} else {
  namespace::Module::default()
};
```

kayagokalp (approved):
🚀  Looks good to me! It feels so nice to get rid of that scary "const" table in the manifest 😄

Left a simple suggestion to make `default_with_contract_id()`'s interface subjectively cleaner but I am perfectly fine with the current way as well as it seems like a personal preference thingy. 

And there seems to be some conflicts, I will re-approve after this is ready to go in

mitchmindtree (approved):

mohammadfawaz (approved):

forc-pkg/src/pkg.rs
bingcicle: I'm thinking we can maybe do without this since inputting the contract ID _might_ be a temporary thing we do here :thinking: if it's the case that my followup PR to tackle #3077 does not change the way we feed the contract ID into the namespace, i'll implement this approach

mitchmindtree (approved):
:ship:

Issue #3752, opened 2023-01-11
forc: Package Hosting and Registry Design (using IPFS and the Fuel Network)


mitchmindtree:
The following is a proposed design around decentralised package hosting and registration for `forc`.

# Goal

The aim is to provide a user experience akin to Cargo’s package registry in a manner that avoids some of the issues associated with its centralised approach to hosting, registering and indexing packages.

### Rust’s Approach

In Rust, users can easily publish packages with `cargo publish`, and depend on packages with simple, semver declarations e.g. `serde = "1.0"`.

By default, packages are published to and fetched from [crates.io](https://crates.io). Crates.io acts as a canonical **registry** (the official record of all packages), **host** (src code and manifest for each version) and **index** (allows searching / exploring packages and discoverability).

The caveat with this design is that when crates.io is down for whatever reason, projects cannot be published or fetched by Rust users. It also means that crates.io is responsible for the content that it hosts, and must respond to licensing challenges, moderation requests, name-sitters, and so on. Cargo users must trust whoever happens to have control over crates.io and their ability to secure it from from malicious actors.

### Proposed Approach

This design aims to improve upon Rust’s approach by avoiding the use of a central authority to handle package registration and hosting. Instead, we opt for delegating the registry and hosting roles to suitable decentralised tooling.

At a high level, the intent is to:

1. **Host** package content (manifest, lock and src) with **IPFS**, and 
2. **Register** packages (publishing their content addresses) with **Fuel Contracts**. 

Package **indexing** could be built upon these two components, but in order to narrow the scope of this work, implementation details around package indexing (website for search-ability, discoverability, etc) are omitted.

# Package Hosting - IPFS

The core appeal of using IPFS is that hosted content is **content addressed**. This means that, only knowing the content address of a package (i.e. the SHA256 of the package contents), we can reliably fetch exactly that package, ***as long as that package is hosted by an IPFS node***.

To ensure that at least one node always hosts all packages, we should host an IPFS node under Fuel Labs. This node would simply query/monitor the registry for all known packages, and ensure that all are pinned and available from at least one location on the network. This ensures that, at a minimum, we always offer *at least* the level of persistence offered by crates.io.

[...]

PR #4299, opened 2023-03-16
feature(forc-pkg): Support IPFS-sourced dependencies

### LINKED ISSUE #3926: Add support to `forc-pkg` for `ipfs`-sourced dependencies
This is one of the first implementation steps in the new package registry design discussed in #3752.

IPFS dependencies will be the first remote dependency alternative to `git` supported by `forc-pkg`.

Its usage should look something like the following:

```toml
foo = { ipfs = "QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG" }
```
where `QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG` is a CID for a directory that contains the package `foo`.

`forc-pkg`'s `pkg.rs` module is getting a bit cluttered - it's likely worth first splitting up the different `Source` kinds into their own modules before adding in `ipfs` support. I'll open a dedicated issue for this.

mitchmindtree:
## Description
Closes #3926.

(edited by @kayagokalp to reflect latest status of the PR)

This PR adds IPFS support for forc-pkg. To do so, forc relies on existing ipfs local node without embedding a node to forc itself. If local system does not have an ipfs node running, a public gateway (https://ipfs.io) is used as a fallback mechanism and packages are fetched through that. 

## Some Context and Future Work

For our package registry stuff to work properly we will eventually need to host an IPFS node to make sure that our published packages are pinned by at least one node in the network. That is the case as long as we don't have a way to incentivize people pinning the packages they published. What will happen is that a package will be published by a developer and after we detect the published package our own IPFS node will pin it to make sure it is always accessible.

Also some benchmarks are done for public gateway during development. It looks like for initial fetch operations from public gateway spends some time looking for the package pinned by the IPFS node simulating our incoming fuel IPFS node which is explained above. It is manageable but fetching from the fuel IPFS node is instant as it already got the package pinned. We can consider falling back to our own node's gateway api rather than a public one to smooth the process.

Finally maybe we can explore embedded ipfs node option as a follow-up, due to the status of IPFS with Rust, it is not very obvious (maybe not event possible atm) how to be able to be fully compatible with kubo, for fetching/publishing our packages. But this is still an open question worth exploring.

### To test with a local node:

1. You can install kubo, you can follow the instructions in their website while installing https://docs.ipfs.tech/install/command-line/
2. Run the `ipfs daemon` service with `ipfs daemon`
3. Add an ipfs source, I have a package already pinned by my IPFS node that you can use:
```toml
[dependencies]
test_lib = { ipfs = "QmfZ3uH7dFEDkYN5RQfyu4m7L8uk8kGiLkNwzHqsrormSj" }
```
4. Run `forc build`

### To test public gateway fallback:

1. Either do this before starting `ipfs daemon` or stop your local daemon with `ipfs shutdown`.
Add an ipfs source, I have a package already pinned by my IPFS node that you can use:
```toml
[dependencies]
test_lib = { ipfs = "QmfZ3uH7dFEDkYN5RQfyu4m7L8uk8kGiLkNwzHqsrormSj" }
```
2. Run `forc build`
## TODO:

- [x] Finish the `pin`, `fetch` and `dep_path` implementations.
- [x] Add tests.
- [x] Add documentation

## Checklist


kayagokalp:
created #4550, #4549 as follow-ups to this one

forc-pkg/src/source/ipfs.rs
```diff
@@ -0,0 +1,233 @@
...
+    forc_util::user_forc_directory().join("ipfs")
+}
+
+fn cache_dir() -> PathBuf {
+    ipfs_dir().join("cache")
+}
+
+fn pkg_cache_dir(cid: &Cid) -> PathBuf {
+    cache_dir().join(format!("{}", cid.0))
+}
```
bingcicle, line 212: These should maybe be placed in a separate file, or at least be at the top of this file so that it's easier to find

forc-pkg/src/source/ipfs.rs
```diff
@@ -0,0 +1,233 @@
...
+/// A client that can interact with local ipfs daemon.
+pub type IpfsClient = ipfs_api::IpfsClient;
+
+/// Package source at a specific content address.
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
+pub struct Source(pub Cid);
+
+/// Package source at a specific content address.
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
+pub struct Pinned(pub Cid);
```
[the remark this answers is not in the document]
kayagokalp: Nice catch, fixing it now!

bingcicle (dismissed):
:shipit:

mitchmindtree (commented):
Great stuff @kayagokalp! Super exciting to see this working locally :)

---

While the local node seemed to work smoothly (via ipfs-api), when falling back to the gateway I appeared to have some problems: 

```console
$ forc build
  Fetching test_lib ipfs+QmQFGz4Z7gqN79PEYakpMNQSf9j58q5rcmdPbMUZ5PW9wn
    Couldn't fetch from local ipfs node, reason:
hyper client error `error trying to connect: tcp connect error: Connection refused (os error 111)`

Caused by:
    0: error trying to connect: tcp connect error: Connection refused (os error 111)
    1: tcp connect error: Connection refused (os error 111)
    2: Connection refused (os error 111).
 Falling back to https://ipfs.io

# 5-minute wait

Error: failed to iterate over archive
```

I first get the warning about falling back to the gateway, then a 5-minute hang, and then finally the `failed to iterate over archive` message. Any ideas what might be going on with the error?

W.r.t. the warning, Rather than printing the full warning and stack trace in yellow, I think it's fine if we just have a single line that indicates the switch over to using the gateway, as this is potentially expected behaviour that we want to support. I'm thinking the verbose yellow warnings might scare the user into thinking something has gone wrong, or potentially also clutter their output? It would be great if we could keep the alignment/style of formatting consistent here too.

---

On another note kind of related to the inline comment - I haven't looked to closely at our newly added dependencies yet, but it would be great to try and make sure we use the minimal set of features where possible, and make sure that in the case a crate offers multiple backends (e.g. tokio, actix, hyper, something else) we try to pick those that are shared to minimize the number of dependencies we introduce.

---

Again, wicked stuff, I think using the local ipfs node via the http API like this will get us a long way before needing to embed a node into forc ourselves :+1:

kayagokalp:
Hey @mitchmindtree, 

Are you pinning the package (`QmQFGz4Z7gqN79PEYakpMNQSf9j58q5rcmdPbMUZ5PW9wn`) with your local node? The public gateway cannot find the package you tried to install from the browser as well, I guess that is why it took 5 min before timing out, you can try that with this: http://ipfs.io/ipfs/QmQFGz4Z7gqN79PEYakpMNQSf9j58q5rcmdPbMUZ5PW9wn/

I came across this issue as well, if your 5001 port is not open, other nodes cannot discover your pinned stuff. I remember reading it from [here](https://discuss.ipfs.tech/t/problem-acessing-my-own-files-through-the-public-gateway/4563). That being said we definitely, need to have better output to let the user know we are waiting a response from public gateway.

The error message you see there occurred since gateway returned a 504, and forc tried to decode that response as a package.

If you don't want to play with your local network firewall & ports just to test this, I have a test server running with that port open. That server hosts an ipfs node which pins a library package with CID: `QmfZ3uH7dFEDkYN5RQfyu4m7L8uk8kGiLkNwzHqsrormSj`, you can check it from https://ipfs.io/ipfs/QmfZ3uH7dFEDkYN5RQfyu4m7L8uk8kGiLkNwzHqsrormSj/ that it loads and forc should be able to download it.

---
For dependencies I might have some unnecessary stuff since I had lots of trials so I will have a second look.

---
I will also take care of the warning message 👍

forc-pkg/src/source/ipfs.rs
```diff
@@ -0,0 +1,235 @@
...
+                    "  {} {} {}",
+                    ansi_term::Color::Green.bold().paint("Fetching"),
+                    ansi_term::Style::new().bold().paint(ctx.name),
+                    self
+                );
+                let cid = &self.0;
+                let ipfs_client = ipfs_client();
+                let dest = cache_dir();
+                let handle = tokio::runtime::Handle::current();
+                let _ = handle.enter();
```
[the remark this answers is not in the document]
kayagokalp: Yeah turns out it is left over from a previous trial :) and we need a better async approach to forc as you mentioned. I will open a seperate issue for it 👍

sdankel:
This is super cool! A few questions since I don't have full context on this.

1. Are we still using github to fetch dependencies as a fallback if both modes of IPFS fail?
2. Does the local node run on a fixed port or is the port dynamically selected?
3. Is the node running all the time or only when `forc build` is run? Does it stay running or shut down? At what point does it shut down?
4. You mentioned benchmarks but I can't find them - could you please add them to the PR description? 😃 
5. Any issues with RAM or CPU usage? About how much CPU and RAM does the local node use when it's running?
6. How do we authenticate with the gateway node? (would be helpful to add to the PR desc)
7. What kind of tests did you add? (helpful in the PR desc)
8. Do we need to test this on different OS's? (i.e. make sure `forc build` still works on Windows, Linux, Mac)

It would also be helpful to add to the PR description details about how you manually tested different scenarios so I could try it myself:
- IPFS local node
- IPFS gateway fallback (do you kill the local node to trigger this?)
- github only (no IPFS)

Some failure cases I'd like to try:
- Something is already running on the port where the local node would start (if it's a fixed port).
- Local node fails and IPFS gateway is down
- Incorrect package name / package not found
- Package version not found
- Incompatible package versions

kayagokalp:
Hey @sdankel, thanks for the comment & questions!

1. 

> Are we still using github to fetch dependencies as a fallback if both modes of IPFS fail?

IPFS support is an alternative way for dependencies (rather than a replacement) so there is no fallback mechanism. As there is no way `forc` can associate a CID with a package without fetching it first. So without accepting a fallback source for a package in manifest file we cannot have a fallback mechanism. I will open an issue to discuss this, I think it makes sense to enable people add multiple sources for the same package so that `forc` can try each of them.

2 & 3.

> Does the local node run on a fixed port or is the port dynamically selected?

> Is the node running all the time or only when forc build is run? Does it stay running or shut down? At what point does it shut down?


One thing to note that is, with this PR, forc does not start a local node automatically, as we couldn't find a good, embeddable rust ipfs node implementation. So instead forc tries to find a local node running (by looking to the default port), if it can find the node, forc will do the fetch operation using the local node. If there is no node running in the system, forc fallbacks into using a public gateway, which is https://ipfs.io. So how the local node selects is port and how and when it stops running depends on the user and the implementation they are using. The most well known implementation is with go and it is called kubo. Also there is a js implementation that is widely popular. So as a user you can choose to use any of them or you can just rely on public gateway if you don't want to have a local node.

4. 

> You mentioned benchmarks but I can't find them - could you please add them to the PR description?

Of course! let me time them in different scenarios and add it to description.

5.

> Any issues with RAM or CPU usage? About how much CPU and RAM does the local node use when it's running?

This again changes with the implementation you are using, this PRs effect on forc is negligible as forc is simply a client to this stack.  

6. 

> How do we authenticate with the gateway node? (would be helpful to add to the PR desc)

That is the nice thing about public gateway, you don't! Since it is read only anyone on the internet can go and use the gateway. For example here is a link to a random file using ipfs.io public gateway. https://ipfs.io/ipfs/QmcKi2ae3uGb1kBg1yBpsuwoVqfmcByNdMiZ2pukxyLWD8/  So as a forc user you don't do anything special. You just need to provide the CID under the dependencies section in manifest file, and forc will fetch it using your local node, and if it does not exists, forc will get it using ipfs.io without any authentication.

7. 

> What kind of tests did you add? 

Just a unit test for parsing. Most importantly we have 0 e2e test for fetching. This is the tricky part, since we don't have a IPFS node operated by Fuel yet, we cannot introduce a test fetching a package. Because to be able to fetch a file at least one node needs to be pinning it. That would mean I would have to let my local node running indefinitely for entire sway repo CI to work 😄. So I left full e2e tests to the end since we will eventually get to a point, where we would also have the publishing infra as a forc plugin. Once we got that far, we will likely host an IPFS node as Fuel so that every package published by our users are pinned at least once by us, this would also mean we will have packages pinned by a node that we operate/rely on which would enable the e2e tests for whole fetch operations.

8.

> Do we need to test this on different OS's? (i.e. make sure forc build still works on Windows, Linux, Mac)

I don't think anything special needed for this PR, but in general if we have the resources that would be great! I don't have a windows machine but I work on a mac and linux workstation. So I tested on those two.


> It would also be helpful to add to the PR description details about how you manually tested different scenarios so I could try it myself:
> 
> IPFS local node
> IPFS gateway fallback (do you kill the local node to trigger this?)
> github only (no IPFS)
> 

Of course,

To test with a local node:

1. You can install kubo, you can follow the instructions in their website while installing https://docs.ipfs.tech/install/command-line/
2. Run the ipfs daemon service with `ipfs daemon`
3. Add an ipfs source, I have a package already pinned by my IPFS node that you can use:
```toml
[dependencies]
test_lib = { ipfs = "QmfZ3uH7dFEDkYN5RQfyu4m7L8uk8kGiLkNwzHqsrormSj" }
```
4. Run `forc build`

To test public gateway fallback:

1. Either do this before starting `ipfs daemon` or stop your local daemon with `ipfs shutdown`.
2. Add an ipfs source, I have a package already pinned by my IPFS node that you can use:
```toml
[dependencies]
test_lib = { ipfs = "QmfZ3uH7dFEDkYN5RQfyu4m7L8uk8kGiLkNwzHqsrormSj" }
```
3. Run `forc build`

> github only (no IPFS): 

Like i mentioned in the questions above, IPFS sources are not related with git sources, so there is no fallback to git. I guess there is nothing to test in that sense but if you do not specify `implicit-std = false` in your manifest file you are already fetching std using git. So while testing the above two steps, you fetch a package using git and then another package using ipfs.

> Something is already running on the port where the local node would start (if it's a fixed port).

This depends on your local ipfs node choice, it is handled by them so it is hard to say something. But as vague as it gets ipfs has a standard which by default starts the local node gateway in 5001 port. You can configure your node to use a different port. Which we will support once we have: https://github.com/FuelLabs/sway/issues/4550

> Local node fails and IPFS gateway is down

I feel like you can test this with cutting your internet and shutdown your local ipfs node if it is running, but not sure will look into this but since we cannot stop public gateway i feel like it is the only way we can test this.

> Incorrect package name
This is a local check forc is already doing, if the package's name does not match the name from manifest forc will error out. So this is done after fetching before building, i.e not dependent on how we fetch the package.

>  / package not found
This part is tricky, we do not have a hashing mechanism to check this, basically we need to fetch the package to see if it is really a package.

> Package version not found
> Incompatible package versions

I am a little lost about these scenarios, can you provide an example for this?

sdankel:
> > Package version not found
> > Incompatible package versions
> 
> I am a little lost about these scenarios, can you provide an example for this?

I was thinking if you specify a dependency version that is incompatible with another dependency, or a version that doesn't exist. This is not really related to the IPSF change though, and is hopefully already handled by forc.

forc-pkg/src/pkg.rs
```diff
@@ -1461,9 +1464,10 @@ fn fetch_deps(
         })?;
```
sdankel, line 1465: Could we change the error message on line 1419 from `Failed to source dependency` to `Failed to source dependency {dependency name}` ?

forc-pkg/src/source/ipfs.rs
```diff
@@ -0,0 +1,235 @@
...
+fn cache_dir() -> PathBuf {
+    ipfs_dir().join(IPFS_CACHE_DIR_NAME)
+}
+
+fn pkg_cache_dir(cid: &Cid) -> PathBuf {
+    cache_dir().join(format!("{}", cid.0))
+}
+
+/// Returns a `IpfsClient` instance ready to be used to make requests to local ipfs node.
+/// TODO: Accept user specified local ipfs node port.
```
sdankel, line 217: Is this done?

sdankel (commented):
We should do more to match the CLI UX of the rest of forc.

image

This could look like:

```
  Fetching test_lib ipfs+QmfZ3uH7dFEDkYN5RQfyu4m7L8uk8kGiLkNwzHqsrormSj
   Checking for local IPFS node
   Fetching from https://ipfs.io Note: this could take several minutes.
     Unpacking IPFS archive
```

with "Fetching", "Checking", and "Unpacking" highlighted green. Also, if there are any other intermediate status messages we could print after `Fetching from https://ipfs.io`, rather than just having it hang for several minutes, we should print them.

kayagokalp:
> Thanks for adding the testing details. Do you have any other packages in the public IPFS node that I could test with?
> 
> A few suggestions:
> 
> ### UX
> We should do more to match the CLI UX of the rest of forc.
> 
> image
> This could look like:
> 
> ```
>   Fetching test_lib ipfs+QmfZ3uH7dFEDkYN5RQfyu4m7L8uk8kGiLkNwzHqsrormSj
>    Checking for local IPFS node
>    Fetching from https://ipfs.io Note: this could take several minutes.
>      Unpacking IPFS archive
> ```
> 
> with "Fetching", "Checking", and "Unpacking" highlighted green. Also, if there are any other intermediate status messages we could print after `Fetching from https://ipfs.io`, rather than just having it hang for several minutes, we should print them.
> 
> ### Latency
> I'm a little concerned about how long it takes to fetch from IPFS. Do we know where that latency is going? If I had 10 dependencies to fetch from IPFS, would they be fetched in parallel or would it take 10+ minutes to fetch them all?
> 
> ### Specifying versions
> I'm not sure I understand how to specify a pinned version and have it fetch from IPFS. If want to fetch `test_lib` v1.0.0. do I need to know the IPFS hash string to include it in my project?
> 
> I tried this:
> 
> ```
> test_lib = { ipfs = "QmfZ3uH7dFEDkYN5RQfyu4m7L8uk8kGiLkNwzHqsrormSj", tag = "v1.0.0" }
>  and 
> test_lib = { ipfs = "QmfZ3uH7dFEDkYN5RQfyu4m7L8uk8kGiLkNwzHqsrormSj", tag = "v1.0.1" }
> ```
> 
> Both seemed to succeed, but I couldn't tell if it was actually using a different version or not.

Thanks for the review.

## UX

Sounds good! I did not go that way to prevent showing duplicate fetching statements for the same package but if it feels more cohesive that way will make the changes 👌 

## Latency

This greatly depends on the network itself and there is nothing we can do about it, except for hoping it to get a lot better once we have our own hosted IPFS. That should greatly improve the latency as our own node will be pinning all the packages itself. Thus it won't have to look it up from peers. That is something we can see once we have the node operational. My initial tests felt a lot faster than ipfs.io but this is a server that I am hosting in my local network with very minimal hardware/bandwidth specs.

## Specifying versions
Tag is something git specific, so it does not change anything with the ipfs source. In IPFS, each different package (and thus each different versions of the same package) will have a different CID. I would suggest taking a peek at the general design at #3752. You can see from the design that we are not going to be specifying a package directly once we have all the components ready (It will be still possible, but it wouldn't be preferred as much as it is pretty manual). Instead we will specify the address of the publisher and desired version and after that forc will resolve the CID from the repo contract. Basically this is the barebones of ipfs support for forc, first of many PRs to come. The implementation is split into multiple parts so that it can be reviewed 😅 

Example:

```toml
foo = { publisher = "mindtree", version = "1" }
```

On a separate note we should generate an error, if users provide git specific stuff with path or IPFS sources to prevent this type of stuff happening with users. I will open an issue for it.

mitchmindtree:
I just took a final quick look - LGTM!

Again, nice work on this Kaya :) Looking forward to seeing where you take this registry!

---

W.r.t. the versions discussion, just wanted to echo Kaya's point about most folks likely not using `ipfs` dependencies directly - rather they're more of a stepping stone toward enabling the design in #3752.

> Instead we will specify the address of the publisher and desired version and after that forc will resolve the CID from the repo contract

@kayagokalp I'd recommend aiming to avoid the need for including `publisher` as a required field for registry packages if at all if possible, i.e. aiming to allow folks to specify packages solely by version like we do for cargo, e.g. `foo = "1"`.

In the design, the idea of having a `publisher` field was a potential workaround for the name-spacing/name-squatting issues, however it does complicate the design and usage in other ways, and raises *more* issues when we start thinking the need for being able to transfer packages between different owners etc.

[This blogpost](https://samsieber.tech/posts/2020/09/registry-structure-influence/) is a nice insight into why cargo's flat namespace is a feature (and not a bug). There are likely better ways of dealing with name-squatting issues, and lessons to be learned from folks dealing with similar problems (e.g. ENS and `.eth` domains, lens protocol and `.lens` handles, the fuel nomen project, etc).

forc-pkg/src/source/ipfs.rs
```diff
@@ -0,0 +1,242 @@
...
+
+/// A pinned instance of an ipfs source
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
+pub struct Pinned(pub Cid);
+
+impl Pinned {
+    pub const PREFIX: &'static str = "ipfs";
+}
+
+const PUBLIC_GATEWAY: &str = "https://ipfs.io";
```
sdankel, line 36: Discussed offline: let's make this public gateway the default, and allow the user to override the URL in the `Forc.toml` with an optional field called `ipfs-node-url`

forc-pkg/src/source/ipfs.rs
```diff
@@ -0,0 +1,242 @@
...
+        let cid = &self.0;
+        let pinned = Pinned(cid.clone());
+        let path = pkg_cache_dir(cid);
+        Ok((pinned, path))
+    }
+}
+
+impl source::Fetch for Pinned {
+    fn fetch(&self, ctx: source::PinCtx, repo_path: &Path) -> Result<PackageManifestFile> {
+        // TODO: implement local cache search for ipfs sources.
```
sdankel, line 61: nit: Is there an issue for this? It would be good to put the issue number in the TODO comment

kayagokalp:
Opened #4659 to track ipfs node deployment

sdankel (commented):
I don't see `--ipfs-node` in the list if options in `forc build --help` - what am I missing?

kayagokalp:
> I don't see `--ipfs-node` in the list if options in `forc build --help` - what am I missing?

Hmm that is weird., are you sure you are building this PR and using the freshly built binary? You might have a `forc` in your path which is shadowing the new forc binary. Here is the output I have with `forc build --help` from this PR:

```console
❯ fuel/dev/sway/target/debug/forc build --help
forc-build 
Compile the current or target project.

The output produced will depend on the project's program type.

- `script`, `predicate` and `contract` projects will produce their bytecode in binary format
`<project-name>.bin`.

- `script` projects will also produce a file containing the hash of the bytecode binary
`<project-name>-bin-hash` (using `fuel_cypto::Hasher`).

- `predicate` projects will also produce a file containing the **root** hash of the bytecode binary
`<project-name>-bin-root` (using `fuel_tx::Contract::root_from_code`).

- `contract` and `library` projects will also produce the public ABI in JSON format
`<project-name>-abi.json`.

USAGE:
    forc build [OPTIONS]

OPTIONS:
        --ast
            Print the generated Sway AST (Abstract Syntax Tree)

        --build-profile <BUILD_PROFILE>
            Name of the build profile to use.
            
            If unspecified, forc will use debug build profile.

        --build-target <BUILD_TARGET>
            Build target to use for code generation
            
            [default: fuel]
            [possible values: fuel, evm, midenvm]

        --dca-graph <DCA_GRAPH>
            Print the computed Sway DCA graph. DCA graph is printed to the specified path. If
            specified '' graph is printed to stdout

        --dca-graph-url-format <DCA_GRAPH_URL_FORMAT>
            Specifies the url format to be used in the generated dot file.
            Variables {path}, {line} {col} can be used in the provided format.
            An example for vscode would be:
              "vscode://file/{path}:{line}:{col}"

        --error-on-warnings
            Treat warnings as errors

        --finalized-asm
            Print the finalized ASM.
            
            This is the state of the ASM with registers allocated and optimisations applied.

    -g, --output-debug <DEBUG_FILE>
            If set, outputs source file mapping in JSON format

    -h, --help
            Print help information

        --intermediate-asm
            Print the generated ASM.
            
            This is the state of the ASM prior to performing register allocation and other ASM
            optimisations.

        --ipfs-node <IPFS_NODE>
            The IPFS Node to use for fetching IPFS sources.
            
            Possible values: PUBLIC, LOCAL, <GATEWAY_URL>

        --ir
            Print the generated Sway IR (Intermediate Representation)

        --json-abi
            By default the JSON for ABIs is formatted for human readability. By using this option
            JSON output will be "minified", i.e. all on one line without whitespace

        --json-abi-with-callpaths
            Outputs json abi with callpaths instead of names for struct and enums

        --json-storage-slots
            By default the JSON for initial storage slots is formatted for human readability.
            By using this option JSON output will be "minified", i.e. all on one line without
            whitespace

    -L, --log-level <LOG_LEVEL>
            Set the log level

        --locked
            Requires that the Forc.lock file is up-to-date. If the lock file is missing, or it needs
            to be updated, Forc will exit with an error

        --metrics-outfile <METRICS_OUTFILE>
            Output compilation metrics into file

    -o, --output-bin <BIN_FILE>
            If set, outputs a binary file representing the script bytes

        --offline
            Offline mode, prevents Forc from using the network when managing dependencies. Meaning
            it will only try to use previously downloaded dependencies

        --output-directory <OUTPUT_DIRECTORY>
            The directory in which the sway compiler output artifacts are placed.
            
            By default, this is `<project-root>/out`.

    -p, --path <PATH>
            Path to the project, if not specified, current working directory will be used

        --release
            Use release build plan. If a custom release plan is not specified, it is implicitly
            added to the manifest file.
            
            If --build-profile is also provided, forc omits this flag and uses provided
            build-profile.

    -s, --silent
            Silence all output

    -t, --terse
            Terse mode. Limited warning and error output

        --tests
            Also build all tests within the project

        --time-phases
            Output the time elapsed over each part of the compilation process

    -v, --verbose
            Use verbose output
```

sdankel:
> > I don't see `--ipfs-node` in the list if options in `forc build --help` - what am I missing?
> 
> Hmm that is weird., are you sure you are building this PR and using the freshly built binary? You might have a `forc` in your path which is shadowing the new forc binary. 

Ah yes, that was my bad. I've installed it correctly now.

When I do forc build --ipfs-node "127.0.0.1:7000"  (random port with nothing hosted on it), it still successfully compiles. It should give an error message in that case, right?

In the Forc.lock the dependency is there:
```
[[package]]
name = 'test_lib'
source = 'ipfs+QmfZ3uH7dFEDkYN5RQfyu4m7L8uk8kGiLkNwzHqsrormSj'
```

sdankel (approved):
Great work!! 🚢

forc/src/cli/shared.rs
```diff
@@ -115,6 +116,11 @@ pub struct Pkg {
     /// Outputs json abi with callpaths instead of names for struct and enums.
     #[clap(long)]
     pub json_abi_with_callpaths: bool,
+    /// The IPFS Node to use for fetching IPFS sources.
+    ///
+    /// Possible values: PUBLIC, LOCAL, <GATEWAY_URL>
+    #[clap(long)]
+    pub ipfs_node: Option<IPFSNode>,
```
sdankel, line 123: Is there a way we can define this CLI arg once and have it be part of the CLI for `build`, `check`, etc?

JoshuaBatty:
Nice work Kaya, tested locally and it feels great. 

Happy to approve once CI is passing again.

forc/src/cli/shared.rs
kayagokalp: That is the main reasoning behind using `shared.rs` but some of the downstream tooling does not use that, will open a separate issue for that since it is not only for this field but other common fields are also repeated in some downstream tooling.

kayagokalp: Opened #4684

forc-pkg/src/source/mod.rs
```diff
@@ -45,6 +46,33 @@ trait DepPath {
...
 
+#[derive(Clone, Debug)]
+pub enum IPFSNode {
+    Local,
+    WithUrl(String),
+}
+
+impl Default for IPFSNode {
+    fn default() -> Self {
+        Self::Local
```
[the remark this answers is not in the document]
kayagokalp: Nice idea! [5865461](https://github.com/FuelLabs/sway/pull/4299/commits/5865461d8387656f2395c21d7f8e03805e365d15)

forc-pkg/src/pkg.rs
kayagokalp: #4685

kayagokalp (approved):
Shameless self approve as @sdankel is OOO and we need a re-approval after merge conflicts fixed

JoshuaBatty (approved):
:shipit:

PR #4310, opened 2023-03-20
Adds urls to graph.visualize generated dot file.

esdrubal:
## Description

graph.visualize now includes urls that are clickable in generated svg files. The urls point to a vscode scheme, and opens vscode in the proper file, line and column number.

Also adds edge labels.
Also changes entry node shape to be different, a double octagon instead of an ellipse.

## Checklist


sway-core/src/control_flow_analysis/flow_graph/mod.rs
```diff
@@ -206,7 +206,26 @@ impl<'cfg> ControlFlowGraph<'cfg> {
...
+                    let node = &self.graph[nr.0];
+                    let mut shape = "";
+                    if self.entry_points.contains(&nr.0) {
+                        shape = "shape=doubleoctagon";
+                    }
+                    let mut url = "".to_string();
+                    if let Some(span) = node.span() {
+                        if let Some(path) = span.path_str() {
+                            let (line, col) = span.start_pos().line_col();
+                            url = format!("URL = \"vscode://file/{path}:{line}:{col}\"");
```
mohammadfawaz, line 223: Because not everyone uses VSCode, baking this into the code directly is a bit odd. However, we could amend the flag `--dca-graph` to accept a graph format or something similar so that people who use VSCode can then use the new flag (or maybe pass a value for it as in `--dca-graph vscode-url` to get the VSCode URL.

esdrubal: Added a new flag, we can now call forc with the argument `--dca-graph-url-format`.
For vscode we would use: `--dca-graph-url-format 'vscode://file/{path}:{line}:{col}'`

mohammadfawaz (dismissed):
Great!

mohammadfawaz:
Conflicts :(

PR #4325, opened 2023-03-22
Fix semantic definitions for `Self` and `self` tokens

### LINKED ISSUE #4058: `self` keyword: incorrect syntax highlighting
Just noticed that `self` keyword isn't having the same syntax highlighting as the other sway keywords. 

Screenshot from 2023-02-13 14-21-32

### LINKED ISSUE #4062: `Self` has incorrect syntax highlighting
Currently `Self` has the semantic token definition of `struct`. In rust, the `Self` token has a semantic token definition of `selfTypeKeyword`.

Screen Shot 2023-02-13 at 4 02 29 pm

JoshuaBatty:
## Description
Fixes incorrect semantic token definitions for `Self` and `self` tokens so they match how `rust-anaylzer` treats these tokens. Also adds simple docs for the variants in `SymbolKind`. 

closes #4058 
closes #4062

## Checklist

### DIFF
--- a/sway-lsp/src/traverse/typed_tree.rs
+++ b/sway-lsp/src/traverse/typed_tree.rs
@@ -1231,8 +1231,7 @@ impl<'a> TypedTree<'a> {
         let type_engine = self.ctx.engines.te();
         let decl_engine = self.ctx.engines.de();
         let type_info = type_engine.get(type_id);
-        let symbol_kind = 
-            type_info_to_symbol_kind(type_engine, &type_info, Some(&type_span));
+        let symbol_kind = type_info_to_symbol_kind(type_engine, &type_info, Some(&type_span));
         match &type_info {
             TypeInfo::Array(type_arg, ..) => {
                 self.collect_type_argument(type_arg);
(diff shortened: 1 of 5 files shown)


sway-lsp/src/traverse/typed_tree.rs
```diff
@@ -1231,7 +1231,15 @@ impl<'a> TypedTree<'a> {
         let type_engine = self.ctx.engines.te();
         let decl_engine = self.ctx.engines.de();
         let type_info = type_engine.get(type_id);
-        let symbol_kind = type_info_to_symbol_kind(type_engine, &type_info);
+        // This is necessary because the type engine resolves `Self` & `self` to the type it refers to.
+        // We want to keep the semantics of these keywords.
+        let symbol_kind = if type_span.as_str() == "Self" {
+            SymbolKind::SelfTypeKeyword
+        } else if type_span.as_str() == "self" {
+            SymbolKind::SelfKeyword
```
sdankel, line 1239: Should this check be moved inside of `type_info_to_symbol_kind`?

JoshuaBatty: I did think about doing that but decided to keep the original use case of that function as is. The problem is that TypeInfo during parsing == `TypeInfo::SelfType` but then after type checking it is changed to be the type it refers so, so for example it could now be `TypeInfo::Struct`. This is what was causing the incorrect definitions.

JoshuaBatty: On second thought though it can't hurt moving them into this function as it is used elsewhere and guarantees nothing falls through the cracks. I'll update now. thanks :)

eureka-cpu (approved):
utACK :shipit:

sdankel (approved):

PR #4331, opened 2023-03-23
Experimental: `index`, `index_assign`, and nested storage maps

mohammadfawaz:
## Description
Introduce `index` and `index_assign` to `StorageMap` and showcase how storage maps and nested storage maps could look like with and without the `[]` operator. These two methods should really be part of traits `Index` and `IndexAssign` but we're not ready to implement those yet because we do not have associated types for traits yet.

In any case, de-sugaring to `index` was implemented but de-sugaring to `index_assign` needed some work in how reassignments are handled.

Note that `IndexAssign` is inspired by https://github.com/rust-lang/rfcs/pull/1129. This is probably the best we can do given that we don't want references in the language just yet (and probably ever).

## Checklist


mitchmindtree (commented):
I'm a bit weary of forging ahead with specialized indexing for the storage types while we don't have a clear idea of how we'll approach the `Index` trait just yet.

E.g. these implementations take `self: StorageHandle<Self>`, which would likely differ from the actual method signature we are likely to want on an `Index` trait. If we're to end up allowing specialized indexing on storage types that differs from an Index trait behaviour but both use the same `[]` syntax sugar, I'm worried we'll just cause more confusion for devs?

One alternative could be similar to the suggestion mentioned [here](https://github.com/FuelLabs/sway/pull/4297#discussion_r1149963104), and allow for implementing the `Index` trait *for* `StorageHandle<StorageMap<K, V>>` (rather than using self type ascription). E.g.

```sway
trait Index<Ix> {
    type Output;
    fn index(self, ix: Ix) -> Self::Output;
}

impl<K, V> Index<K> for StorageHandle<StorageMap<K, V>> {
    type Output = StorageHandle<V>;
    fn index(self, ix: K) -> Self::Output {
        self.get(ix)
    }
}
```
This would not incur the same issues w.r.t. orphan rules as mentioned at the linked comment as the `Index` trait would likely be declared in the same library that provides the implementation (i.e. sway-lib-std).

PR #4337, opened 2023-03-24
Implements aliases support in DCA.

### LINKED ISSUE #4283: Handle DCA for type aliases
https://github.com/FuelLabs/sway/pull/4151 introduces type aliases but does not handle DCA for them just yet.

### LINKED ISSUE #4328: Wrong DCA warning while using struct method defined in impl self.
While compiling the following:
```rust
struct MyStruct1 {}

impl MyStruct1 {
    fn foo() {}
}

fn main() {
    MyStruct1::foo();
}
```
We are getting this DCA warnings:
```
warning
 --> /home/marcos/Documents/dev/sway/test/src/e2e_vm_tests/test_programs/should_pass/dca/impl_self/src/main.sw:3:8
  |
1 | 
2 | 
3 | struct MyStruct1 {}
  |        --------- This struct is never used.
```

In rust there are no warnings emitted, as shown in this [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6a7bcdc59c8733b4b2de7f124f5737cc).

The current DCA graph to the example is as follows:
bug3

esdrubal:
## Description

Adds support for aliases in DCA.

Fixes #4328 where a DCA warning was emmited while using struct method defined in an impl self.

Closes #4283
Closes #4328

## Checklist


mohammadfawaz:
I did not understand the purpose of `call_path_decl`. What does it contain exactly? And why do we only need it in certain situations? (example: only for enums but not structs)

esdrubal:
> I did not understand the purpose of `call_path_decl`. What does it contain exactly? And why do we only need it in certain situations? (example: only for enums but not structs)

`call_path_decl` is needed for the DCA to be able to retrieve a type alias. For instance in the example below we have EnumScrutinee which contains `enum_ref: DeclRefEnum` already resolved to an enum, it is impossible to infer if we used a type alias from that. So I added `call_path_decl` which may be an alias or an enum decl.

https://github.com/FuelLabs/sway/blob/9b80160f681df58138079e841920e987518268b6/sway-core/src/language/ty/expression/scrutinee.rs#L27-L31

Structs should also have it, I probably forgot to add tests cases for structs, I will check that.

esdrubal:
Fixed merging conflicts. And added one more test case and a fix for it.

The reason we have `call_path_decl` in `EnumScrutinee` and not in `StructScrutinee` is because we build an `UnsafeDowncast` from `EnumScrutinee`.  The DCA uses `UnsafeDowncast` to connect match statements to the enum variants. In case we use an alias in the enum we need `call_path_decl` to contain it.

sdankel (approved):

IGI-111 (approved):

Issue #4117, opened 2023-02-17
Relax `--salt` flag restrictions for the `forc deploy` command. Allow specifying contract member.


mitchmindtree:
#4116 introduces the ability to specify `--salt` for `forc deploy` however there are some ambiguities when attempting to deploy a full workspace, so we only allow the flag when deploying a single package:

https://github.com/FuelLabs/sway/blob/2b6859e8ad735b94fe765f6329baa6493202ac73/forc-plugins/forc-client/src/op/deploy.rs#L43-L59

Perhaps we should enable specifying `--salt` for workspace deployment, but with the condition that the workspace member contract name must be prefixed?

E.g. here we deploy all contracts in the workspace, but specify the salt for workspace member contract `foo`:

```console
forc deploy --salt foo:0x12341234123412341234123412341234
```

We could also allow multiple `--salt` flags, e.g. here we salt both `foo` and `bar`:

```console
forc deploy \
  --salt foo:0x12341234123412341234123412341234 \
  --salt bar:0x56875678567856785678567856785678
```

NOTE: These `--salt` flags should **not** override salt specified in `[contract-dependencies]`, and may only override deployment of top-level contract members.

If salt is specified for a contract member, but this member *also* appears within `[contract-dependencies]` of another contract member with a different salt, we should either 1. provide an error or 2. the contract should be deployed twice - once for each salt.

PR #4338, opened 2023-03-24
feat: add `forc contract-id` and `forc predicate-id` commands

### LINKED ISSUE #3444: Add `forc contract-id` and `forc predicate-id` commands

kayagokalp:
## Description
closes #3444.

## Checklist


forc-plugins/forc-id/src/op/contract_id.rs
```diff
@@ -0,0 +1,75 @@
...
+use tracing::info;
+
+pub fn contract_id(command: ContractIdCommand) -> Result<()> {
+    let build_options = build_opts_from_cmd(&command);
+    let build_plan = pkg::BuildPlan::from_build_opts(&build_options)?;
+    // If a salt was specified but we have more than one member to build, there
+    // may be ambiguity in how the salt should be applied, especially if the
+    // workspace contains multiple contracts, and especially if one contract
+    // member is the dependency of another (in which case salt should be
+    // specified under `[contract- dependencies]`). Considering this, we have a
```
bingcicle, line 14:
```suggestion
    // specified under `[contract-dependencies]`). Considering this, we have a
```

forc-plugins/forc-id/src/op/contract_id.rs
```diff
@@ -0,0 +1,75 @@
...
+    // simple check to ensure that we only accept salt when working on a single
+    // package. In the future, we can consider relaxing this to allow for
+    // specifying a salt for workspacs, as long as there is only one
+    // root contract member in the package graph.
+    if command.salt.salt.is_some() && build_plan.member_nodes().count() > 1 {
+        bail!(
+            "A salt was specified when attempting to contract id detection \
```
bingcicle, line 21:
```suggestion
            "A salt was specified when attempting to detect the contract id \
```

Also the behaviour for deploy in workspaces will change because of the issue i'm working on now (#4117). Perhaps this could be a follow-up issue to that.

bingcicle (changes requested):
We should also probably add the `forc-id` plugin in the CI:

https://github.com/FuelLabs/sway/blob/adc5c1a9271cb24f91a5e1e68341ae1e4671115b/.github/workflows/ci.yml#L646-L682

mitchmindtree (commented):
I'm thinking maybe we should just add these as subcommands to `forc` for simplicity?

I don't think these subcommands require any special dependencies beyond what we already have in `forc`, so I don't think it would add much bloat to the original binary. Thoguhts?

kayagokalp:
> I'm thinking maybe we should just add these as subcommands to forc for simplicity?
> 
> I don't think these subcommands require any special dependencies beyond what we already have in forc, so I don't think it would add much bloat to the original binary. Thoguhts?

Hmm that makes sense, this commands are indeed only uses subset of `forc`'s dependencies other than `forc-tx`. Semantically i thought it would be weird for `forc` to depend on a plugin which sounds like a piece of optional code piece. I am good with either way, and it is also important to note that I only use `Salt` from `forc_tx` which is a wrapper around original `Salt`. I can move that wrapper into somewhere like `forc-util` and re-use it from `forc_tx` and the integrated version of this plugin.

mitchmindtree:
> How does that sound?

Ahh I didn't notice the use of forc-tx! Yeah moving the salt type to `forc-util` sounds good for now. Perhaps in the future we can create a  `forc-cli-types` crate or similar that provides a bunch of helper types for building commands that can be shared between different crates :thinking:

kayagokalp:
@mitchmindtree Moved the `forc-id` plugin into `forc`. We now have `contract-id` and `predicate-root`  subcommands. The `salt` is moved to `forc-util` and re-used from the plugin (`forc-tx`) and `forc` itself. Having a `forc-cli-types` sounds great btw.

@bingcicle addressed your requested changes as well but since the plugin is removed I haven't touched the CI.

forc-pkg/src/pkg.rs
```diff
@@ -756,6 +766,24 @@ impl BuildPlan {
                 .map(|(n, p)| (n.clone(), p.clone()))
         })
     }
+
+    /// Return the salt for the given pinned pkg, for none contract members returns `None`.
```
bingcicle, line 770:
```suggestion
    /// Returns a salt for the given pinned package if it is a contract and None for libraries.
```

forc/src/ops/forc_contract_id.rs
```diff
@@ -0,0 +1,75 @@
+use crate::cli::ContractIdCommand;
+use anyhow::{bail, Result};
+use forc_pkg::{self as pkg, build_with_options};
+use sway_core::{fuel_prelude::fuel_tx, BuildTarget};
+use tracing::info;
+
+pub fn contract_id(command: ContractIdCommand) -> Result<()> {
+    let build_options = build_opts_from_cmd(&command);
+    let build_plan = pkg::BuildPlan::from_build_opts(&build_options)?;
+    // If a salt was specified but we have more than one member to build, there
```
bingcicle, line 10: We might want to create a new issue to handle multiple salts when https://github.com/FuelLabs/sway/pull/4362 goes in, but for now this seems OK to me

forc-pkg/src/pkg.rs
```diff
@@ -346,6 +346,16 @@ impl MemberFilter {
         }
     }
 
+    /// Returns a new `BuildFilter`, that only builds predicates.
```
bingcicle, line 349:
```suggestion
    /// Returns a new `MemberFilter`, that only builds predicates.
```

Should this be `MemberFilter` instead? Same for the other methods found above this.

forc/src/ops/forc_contract_id.rs
kayagokalp: Created #4338.

bingcicle (approved):
:shipit:

JoshuaBatty (approved):

PR #4341, opened 2023-03-26
feat: enable raw pointer return from scripts

iqdecay:
- This PR enables returning heap types from scripts.
- In https://github.com/FuelLabs/fuels-rs/pull/848, support was added for
  returning vectors from smart contracts in the SDK, this safeguard can now be
  removed so the same can happen with scripts.


otrho:
Hrmm, I'm not sure this is necessarily what we want.  Returning a `Vec` can be special-cased, and should be returned using a `RETD` of the vector contents, I assume?  Is that right?

But just returning a `raw_ptr` using `RET` doesn't make sense and was banned for a reason.  A better solution might be to convert the `Vec` to a `raw_slice` which we already allow, and to return that..?

iqdecay:
I'm just trying to support returning `Vec` and `Bytes` types from a script, which is not currently supported because of the safeguard that I remove in this PR. @mohammadfawaz had told me that this safeguard was here just because the SDK didn't support it, so I naively assumed that just removing the safeguard was enough.

iqdecay:
Closing in favor of an upcoming PR.

iqdecay:
Reopening the PR after a discussion with Toby. Currently, contracts can return `raw_ptr`, and for coherence there is no reason to prevent scripts from doing so too, this way the SDK can handle scripts returning vectors.

mohammadfawaz (approved):

PR #4348, opened 2023-03-27
Implement OR match pattern

### LINKED ISSUE #769: Implement "or" match patterns in match expressions.
The "or" pattern is the one shown below:

```rust
let x = 1;
match x {
  0 | 1 => true,
  _ => false
}
```

IGI-111:
## Description

Allow users to specify match patters such as:

```sway
let x = 1;
match x {
  0 | 1 => true,
  _ => false
}
```

We also check that all patterns in a disjunction declare the same set of
variables (not doing so is an error).

Fix https://github.com/FuelLabs/sway/issues/769

This requires a change in the pattern matching analysis to remove the
assumption that a specialized matrix of a vector pattern is always a
vector, which is now no longer true because or patterns can generate
multiple branches and therefore multiple rows.

## Checklist


examples/match_statements/src/main.sw
```diff
@@ -37,6 +37,11 @@ fn main(num: u64) -> u64 {
         Weather::Snowy => 20,
     };
 
+    let isSunny = match currentWeather {
```
JoshuaBatty, line 40:
```suggestion
    let is_sunny = match current_weather {
```

sway-core/src/semantic_analysis/ast_node/expression/match_expression/analysis/witness_report.rs
```diff
@@ -49,6 +49,7 @@ impl WitnessReport {
     ) -> CompileResult<(Pattern, Self)> {
         let mut warnings = vec![];
         let mut errors = vec![];
+        dbg!(&witness_report, &c);
```
JoshuaBatty, line 52: can this be removed?

examples/match_statements/src/main.sw
IGI-111: I'd rather follow the convention within the file, but maybe we should change the whole file to use snake case for identifiers.

JoshuaBatty: my vote is to change all occurrences to snake case.

JoshuaBatty (approved):

PR #4361, opened 2023-03-27
Move display impls to debug

sdankel:
## Checklist

### DIFF
--- a/sway-core/src/type_system/info.rs
+++ b/sway-core/src/type_system/info.rs
@@ -411,7 +411,7 @@ impl OrdWithEngines for TypeInfo {
     }
 }
 
-impl DisplayWithEngines for TypeInfo {
+impl DebugWithEngines for TypeInfo {
     fn fmt(&self, f: &mut fmt::Formatter<'_>, engines: Engines<'_>) -> fmt::Result {
         use TypeInfo::*;
         let s = match self {
@@ -434,7 +434,7 @@ impl DisplayWithEngines for TypeInfo {
             Tuple(fields) => {
                 let field_strs = fields
                     .iter()
-                    .map(|field| engines.help_out(field).to_string())
+                    .map(|field| format!("{:?}", engines.help_out(field)))
                     .collect::<Vec<String>>();
                 format!("({})", field_strs.join(", "))
             }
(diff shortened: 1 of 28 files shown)


sway-core/src/type_system/info.rs
```diff
@@ -412,6 +412,65 @@ impl OrdWithEngines for TypeInfo {
...
+                    decl.type_parameters.iter().map(|x| x.type_id),
+                )
+            }
+            ContractCaller { abi_name, .. } => format!("ContractCaller<{abi_name}>"),
+            Array(elem_ty, count) => {
+                format!("[{:?}; {}]", engines.help_out(elem_ty), count.val())
+            }
+            Storage { .. } => "storage".into(),
+            RawUntypedPtr => "pointer".into(),
+            RawUntypedSlice => "slice".into(),
```
sdankel, line 466: I'm not sure what the best user-friendly label for these is. Any example code using these types would be helpful!

sway-core/src/type_system/info.rs
```diff
@@ -412,6 +412,65 @@ impl OrdWithEngines for TypeInfo {
...
+            }
+            Struct(decl_ref) => {
+                let decl = engines.de().get_struct(decl_ref);
+                print_inner_types(
+                    engines,
+                    decl.call_path.suffix.as_str().to_string(),
+                    decl.type_parameters.iter().map(|x| x.type_id),
+                )
+            }
+            ContractCaller { abi_name, .. } => format!("ContractCaller<{abi_name}>"),
```
sdankel, line 460: This is the fix for `ContractCaller`

JoshuaBatty (approved):
Much nicer!

mitchmindtree (approved):
Nice!

One alternative to the trait approach we might want to consider in sway-core could be to use a type, something like:

```rust
pub struct WithEngines<'a, T> {
    ty: T,
    engines: &'a Engines<'a>,
}
```

Then we make use of the regular `fmt` traits, e.g.

```rust
impl<'a> fmt::Debug for WithEngines<'a, TyAstNode> {
    // ...
}

impl<'a> fmt::Display for WithEngines<'a, TyAstNode> {
    // ...
}
```

Rust's `Path::display` methods are a good demo of this: https://doc.rust-lang.org/std/path/struct.Path.html#method.display.

But anyways this is probably worth it's own issue!

LGTM :ship:

sway-core/src/type_system/info.rs
mohammadfawaz: Yeah I think this fine!

PR #4362, opened 2023-03-28
feat: allow multiple salt declarations in `forc deploy`

### LINKED ISSUE #4117: Relax `--salt` flag restrictions for the `forc deploy` command. Allow specifying contract member.

bingcicle:
## Description

Closes #4117 

Allow multiple salt declarations in `forc deploy`, with parsing and validation involved. The gist of it is that it collects all the salt inputs by the user as a `Vec<String>`, parses each input as a `Salt` and store them in a `ContractSaltMap` matching contract->salt, and later compares them against the contract dependencies to make sure that duplicate salts do not exist.

This throws an error if:
1) multiple salts are present (if the same salt is declared but also found within a contract dependency), 
2) if a salt is invalid, 
3) if the salt input is invalid (there's a difference - the salt input is the string `<CONTRACT>:<SALT>`, while the salt above is `fuel_tx::Salt`, which has its own validation.
4) duplicate salts for the same contract were provided in the CLI.

Tests consist of dummy folders (not entire full package initialized by forc) with only manifests to keep it the bare minimum required for these tests.

## Checklist


bingcicle:
This could conflict with #4338 with the way salts are handled

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -9,20 +9,70 @@ use anyhow::{bail, Context, Result};
...
 
+#[derive(Debug)]
 pub struct DeployedContract {
     pub id: fuel_tx::ContractId,
 }
 
+type ContractSaltMap = BTreeMap<String, Salt>;
+
+/// Takes the salt input passed via the --salt option, validates them against
+/// the manifests and returns a ContractSaltMap (BTreeMap of contract names to salts).
```
mitchmindtree, line 30:
```suggestion
/// Takes the contract member salt inputs passed via the --salt option, validates them against
/// the manifests and returns a ContractSaltMap (BTreeMap of contract names to salts).
```

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -9,20 +9,70 @@ use anyhow::{bail, Context, Result};
...
+) -> Result<ContractSaltMap> {
+    let mut contract_salt_map = BTreeMap::default();
+
+    // Parse all the salt arguments first, and exit if there are errors in this step.
+    for salt_arg in salt_args {
+        if let Some((given_contract_name, salt)) = salt_arg.split_once(':') {
+            if contract_salt_map.contains_key(given_contract_name) {
+                bail!("2 salts provided for contract '{given_contract_name}'");
+            }
+            contract_salt_map.insert(given_contract_name.to_string(), salt_from_str(salt)?);
```
mitchmindtree, line 43:
```suggestion
            let salt = salt_from_str(salt)?;
            if let Some(old) = contract_salt_map.insert(given_contract_name.to_string(), salt) {
                bail!("2 salts provided for contract '{given_contract_name}':\n  {old}\n  {salt}");
            }
```

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -9,20 +9,70 @@ use anyhow::{bail, Context, Result};
...
+        } else {
+            bail!("Invalid salt provided - salt must be in the form <CONTRACT_NAME>:<SALT> when deploying a workspace");
+        }
+    }
+
+    for manifest in manifests {
+        for dep in manifest.contract_deps() {
+            if let Some(declared_salt) = contract_salt_map.get(dep.0) {
```
mitchmindtree, line 51: This only checks for the dependency name, but we actually want to check the dependency's *package* name. Often these are the same, but it's possible that the user might use the `package` field to give a different name to the dependency itself for local use.

```suggestion
        for (dep_name, dep) in manifest.contract_deps() {
            let dep_pkg_name = dep.package().unwrap_or(&dep_name);
            if let Some(declared_salt) = contract_salt_map.get(dep_pkg_name) {
```

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -9,20 +9,70 @@ use anyhow::{bail, Context, Result};
...
+    Ok(contract_salt_map)
+}
+
+/// Decodes a hexadecimal number into a fuel_tx::Salt.
+fn salt_from_str(salt_str: &str) -> Result<Salt> {
+    let mut bytes = [0u8; 32];
+    hex::decode_to_slice(salt_str.replace("0x", ""), &mut bytes)?;
+
+    Ok(Salt::new(bytes))
+}
```
mitchmindtree, line 74: I _think_ `Salt` impls `FromStr`, so we should be able to remove this in favour of using:
```rust
let salt: Salt = salt_str.parse()?;
```

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -75,6 +148,7 @@ pub async fn deploy_pkg(
     command: &cmd::Deploy,
     manifest: &PackageManifestFile,
     compiled: &BuiltPackage,
+    contract_salt_map: Option<&ContractSaltMap>,
```
mitchmindtree, line 151: Rather than passing in the whole `ContractSaltMap` here, perhaps we should determine the salt before calling `deploy_pkg` and pass the package's `Salt` as an arg?

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -179,3 +259,117 @@ fn build_opts_from_cmd(cmd: &cmd::Deploy) -> pkg::BuildOpts {
...
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    fn setup_manifest_files() -> BTreeMap<String, PackageManifestFile> {
+        let mut contract_to_manifest = BTreeMap::default();
+        let manifests_dir = std::env::current_dir().unwrap().join("test/data/manifests");
```
mitchmindtree, line 269: It looks like this assumes the location of the `current_dir` which can change depending on from where the user is running `cargo test`. We probably want to use `env!("CARGO_MANIFEST_DIR");` instead? I don't recall if this gives us the package dir or the workspace dir though :thinking:

forc-plugins/forc-client/src/cmd/deploy.rs
```diff
@@ -17,8 +17,8 @@ pub struct Command {
     pub gas: Gas,
     #[clap(flatten)]
     pub maturity: Maturity,
-    #[clap(flatten)]
-    pub salt: Salt,
+    #[clap(long)]
+    pub salt: Option<Vec<String>>,
```
mitchmindtree, line 21: Let's add some docs here with a usage example if we can?

mitchmindtree (commented):
Nice one @bingcicle! Just a few small suggestions.

forc-plugins/forc-client/src/op/deploy.rs
bingcicle: This is much cleaner, thanks!

forc-plugins/forc-client/src/op/deploy.rs
bingcicle: Thanks for this catch!

forc-plugins/forc-client/src/cmd/deploy.rs
```diff
@@ -17,8 +17,18 @@ pub struct Command {
...
+    /// Optional 256-bit hexadecimal literal(s) to redeploy contracts.
+    ///
+    /// For a single contract, use `--salt <SALT>`, eg.: forc deploy --salt 0x0000000000000000000000000000000000000000000000000000000000000001
+    ///
+    /// For a workspace with multiple contracts, use `--salt <CONTRACT_NAME>:<SALT>`
+    /// to specify a salt for each contract, eg.:
+    ///
+    /// forc deploy --salt contract_a:0x0000000000000000000000000000000000000000000000000000000000000001
+    /// --salt contract_b:0x0000000000000000000000000000000000000000000000000000000000000002
+
```
mitchmindtree, line 29:
```suggestion
```

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -179,3 +256,105 @@ fn build_opts_from_cmd(cmd: &cmd::Deploy) -> pkg::BuildOpts {
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    fn setup_manifest_files() -> BTreeMap<String, PackageManifestFile> {
+        let mut contract_to_manifest = BTreeMap::default();
+
+        let manifests_dir = PathBuf::from(format!(
+            "{}/test/data/manifests",
```
mitchmindtree, line 268: Always good to prefer using `join` over `format!` - it's just as easy and supports more platforms, e.g.
```rust
Path::from(env!("CARG_MANIFEST_DIR")).join("test").join("data").join("manifests")
```

mitchmindtree (dismissed):
Couple minor nits, but not important - LGTM

:shipit:

forc-plugins/forc-client/src/cmd/deploy.rs
bingcicle: Oops! Thanks for this catch

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -4,10 +4,25 @@ use anyhow::Result;
...
+    pub fn built_package(&self) -> &Arc<BuiltPackage> {
+        &self.0
+    }
+
+    /// Returns an immutable reference into the PackageManifestFile.
+    pub fn package_manifest_file(&self) -> &PackageManifestFile {
+        &self.1
+    }
+}
+
```
kayagokalp, line 21: This is a nice abstractions but it made me think a little and I think returning `Vec` of `(PackageManifestFile, Arc<BuiltPackage>)` might be unnecessary at the first place from `built_pkg_with_manifest` function. Every built package contains a package descriptor which contains the related package manifest so returning a tuple is kind of a data duplication. 

We can also get this in and think about it later, maybe refactor it slightly such that we do not need to return duplicated package manifests.

kayagokalp (approved):
left a comment/discussion point but other than that looks good to me ✅

forc-plugins/forc-client/src/util/pkg.rs
bingcicle: Good catch, going to create a new issue to address this

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -40,30 +87,66 @@ pub async fn deploy(command: cmd::Deploy) -> Result<Vec<DeployedContract>> {
...
+                    .map(|bpwm| bpwm.package_manifest_file()),
+            )?;
+
+            Some(map)
+        } else {
+            if salt_input.len() > 1 {
+                bail!("More than 1 salt was specified when deploying a single contract");
+            }
+
+            // OK to index into salt_input ahd built_pkgs_with_manifest here,
```
sdankel, line 107: Do we need to handle the case where `built_pkgs_with_manifest.len() == 0`?

sdankel (approved):
LGTM, just had 1 question.

forc-plugins/forc-client/src/op/deploy.rs
bingcicle: This shouldn't be able to return 0 technically, I guess adding a check upstream within built_pkgs_with_manifest would be good here

mohammadfawaz (approved):

PR #4381, opened 2023-04-01
Allow importing enum variants

### LINKED ISSUE #614: Allow importing enum variants
In Rust, you can import enum variants like this: `use Result::*`. We want that.

IGI-111:
## Description

Fix #614

This change allows imports of enum variants either individually or with a wildcard, like so:

```sway
use std::option::Option::None;
use std::option::Option::*;
```

Variants can then be used without specifying the enum, both in scrutinees and expressions:

```sway
let a = Some(10);
match a {
  Some(_) => {}
  None => {}
}
```

This creates a novel ambiguity: `Some(10)` would previously have been unambiguously a function application, but it can now be a variant. We resolve this through the existing mechanism for ambiguous call paths which now handles not just expressions like `Foo::Bar` but also expressions like `Foo`.

These variant imports point to a special kind of `TyDecl` and are eventually resolved to a proper enum declaration at the type checking phase similarly to other "special" `TyDecl`s.

## Checklist


JoshuaBatty (commented):
Nice! LGTM, just a conflict that needs addressing before I can approve.

JoshuaBatty (approved):

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -968,6 +968,37 @@ impl ty::TyExpression {
     ) -> CompileResult<ty::TyExpression> {
         let decl_engine = ctx.decl_engine;
 
+        // is it a singleton?
+        let before = if let Some(b) = before {
+            b
+        } else {
```
otrho, line 974: This pattern can be simplified to `before.unwrap_or_else(|| ...)`.

sway-core/src/semantic_analysis/namespace/module.rs
```diff
@@ -400,6 +403,183 @@ impl Module {
...
+
+        let decl_engine = engines.de();
+
+        let src_ns = check!(
+            self.check_submodule(src),
+            return err(warnings, errors),
+            warnings,
+            errors
+        );
+        match src_ns.symbols.get(enum_name).cloned() {
```
otrho, line 430: Similarly here, for the rest of this match statement.  It's often more concise to use `map()`, `and_then()`, `unwrap_or_else()`, etc. for `Option` and `Result` rather than matching `Some` or `None`.

otrho (approved):

Issue #3659, opened 2022-12-27
Simplify `forc doc` `constants` to a single file


eureka-cpu:
This is to avoid duplicates and have all `const`s in one place

eureka-cpu:
This was closed by #4382

Issue #3656, opened 2022-12-27
Reorganize the module tree of `forc doc`


eureka-cpu:
The `doc` and `render` files in `forc doc` are growing quite large and will continue to do so. It may be a good idea to reorganize this to simplify file navigation.

eureka-cpu:
This was closed by #4382

PR #4382, opened 2023-04-01
chore(`forc-doc`): Reorganizes the file tree

eureka-cpu:
## Description
A clean up PR to make finding & understanding things much easier.

## Checklist

- [x] I have linked to any relevant issues. #3656 #3659


kayagokalp (dismissed):
Left some optional stuff, feel free to disregard

forc-plugins/forc-doc/src/render/index.rs
```diff
@@ -0,0 +1,243 @@
...
+                        //                     placeholder="Searchbar unimplemented, see issue #3480...",
+                        //                     type="search"
+                        //                 );
+                        //                 div(id="help-button", title="help", tabindex="-1") {
+                        //                     button(type="button") { : "?" }
+                        //                 }
+                        //             }
+                        //         }
+                        //     }
+                        // }
```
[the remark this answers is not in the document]
eureka-cpu: I would like to keep it since it's technically going to be a feature soon, if possible.

forc-plugins/forc-doc/src/render/util/format/code_block.rs
```diff
@@ -0,0 +1,7 @@
+/// Takes a formatted String fn and returns only the function signature.
```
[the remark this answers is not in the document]
eureka-cpu: Sure that makes sense

JoshuaBatty (dismissed):

forc-plugins/forc-doc/src/render/util/format/docstring.rs
```diff
@@ -0,0 +1,91 @@
...
+        v.first().expect("expected a non-empty str")
+    } else if raw_html.contains(H4) {
+        let v: Vec<_> = raw_html.split(H4).collect();
+        v.first().expect("expected a non-empty str")
+    } else if raw_html.contains(H5) {
+        let v: Vec<_> = raw_html.split(H5).collect();
+        v.first().expect("expected a non-empty str")
+    } else {
+        raw_html
+    }
```
Braqzen, line 90: hm, I have a feeling that there's probably some slick way of refining this into a few lines

forc-plugins/forc-doc/src/render/title.rs
```diff
@@ -0,0 +1,94 @@
...
+        match self {
+            Self::Modules => "mod",
+            Self::Structs => "struct",
+            Self::Enums => "enum",
+            Self::Traits => "trait",
+            Self::Abi => "abi",
+            Self::ContractStorage => "storage",
+            Self::Constants => "constant",
+            Self::Functions => "fn",
+            _ => unimplemented!("These titles are unimplemented, and should not be used this way."),
```
Braqzen, line 63: Shouldn't this use a `panic!` instead? Wouldn't that be more accurate since it's not meant to be implemented?

forc-plugins/forc-doc/src/render/title.rs
```diff
@@ -0,0 +1,94 @@
...
+        match self {
+            TyDecl::StructDecl { .. } => BlockTitle::Structs,
+            TyDecl::EnumDecl { .. } => BlockTitle::Enums,
+            TyDecl::TraitDecl { .. } => BlockTitle::Traits,
+            TyDecl::AbiDecl { .. } => BlockTitle::Abi,
+            TyDecl::StorageDecl { .. } => BlockTitle::ContractStorage,
+            TyDecl::ConstantDecl { .. } => BlockTitle::Constants,
+            TyDecl::FunctionDecl { .. } => BlockTitle::Functions,
+            _ => {
+                unreachable!("All other TyDecls are non-documentable and will never be matched on")
```
Braqzen, line 90: Also wondering about why `unreachable!` was used here.

forc-plugins/forc-doc/src/render/index.rs
```diff
@@ -0,0 +1,243 @@
...
+    module_info: ModuleInfo,
+    module_docs: DocLinks,
+}
+impl ModuleIndex {
+    pub(crate) fn new(
+        version_opt: Option<String>,
+        module_info: ModuleInfo,
+        module_docs: DocLinks,
+    ) -> Self {
+        ModuleIndex {
```
Braqzen, line 111:
```suggestion
        Self {
```

forc-plugins/forc-doc/src/render/index.rs
```diff
@@ -0,0 +1,243 @@
...
+            self.module_docs.clone(),
+        )
+    }
+}
+impl Renderable for ModuleIndex {
+    fn render(self, render_plan: RenderPlan) -> Result<Box<dyn RenderBox>> {
+        let doc_links = self.module_docs.clone().render(render_plan.clone())?;
+        let sidebar = self.sidebar().render(render_plan)?;
+        let title_prefix = match self.module_docs.style {
+            DocStyle::ProjectIndex(ref program_type) => format!("{program_type} "),
```
Braqzen, line 137: I don't remember the different between using `ref` vs `&`, I always forget. Is there a reason for `ref` over the common `&`?

forc-plugins/forc-doc/src/render/index.rs
```diff
@@ -0,0 +1,243 @@
...
+            DocStyle::ModuleIndex => "Module ".to_string(),
+            _ => unreachable!("Module Index can only be either a project or module at this time."),
```
Braqzen, line 139: Unreachable over panic?

forc-plugins/forc-doc/src/render/item/context.rs
```diff
@@ -0,0 +1,354 @@
...
+    pub(crate) fn new(module_info: ModuleInfo, context_type: ContextType) -> Self {
+        Self {
+            module_info,
+            context_type,
+        }
+    }
+}
+impl Renderable for Context {
+    fn render(self, render_plan: RenderPlan) -> Result<Box<dyn RenderBox>> {
+        let mut rendered_list: Vec<String> = Vec::new();
```
Braqzen, line 46: Perhaps inside each match arm before the loop consider reallocating to the size of the field/variant/method so that the pushing does not continue to reallocate? This would be a single change vs multiple changes depending on the number of items

forc-plugins/forc-doc/src/render/item/context.rs
```diff
@@ -0,0 +1,354 @@
...
+pub(crate) enum ContextType {
+    /// structs
+    StructFields(Vec<TyStructField>),
+    /// storage
+    StorageFields(Vec<TyStorageField>),
+    /// enums
+    EnumVariants(Vec<TyEnumVariant>),
+    /// traits and abi, this can be split
+    /// at a later date if need be
+    RequiredMethods(Vec<TyTraitFn>),
```
Braqzen, line 343: Unsure about how useful this documentation is. Perhaps remove / provide more context? Pun unintended

forc-plugins/forc-doc/src/render/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// Final rendering of a [Document] HTML page to String.
+    fn from_rendered_content(rendered_content: Box<dyn RenderBox>) -> Self {
+        let markup = html! {
+            : doctype::HTML;
+            html {
+                : rendered_content
+            }
+        };
+
+        Self(markup.into_string().unwrap())
```
Braqzen, line 264: Can we unwrap here or should we expect?

Braqzen (approved):
I would avoid as many glob imports as possible

forc-plugins/forc-doc/src/render/util/format/docstring.rs
eureka-cpu: I could put them in a vec, then loop through the vec and perform the action that's being repeated here. That would reduce the amount of lines for sure. I'm honestly not sure why I did it this way initially, I think I was just trying to rush past this part and finish the main portion of the PR this came from.

forc-plugins/forc-doc/src/render/title.rs
eureka-cpu: Since it's only for use in forc doc, I think it makes sense to have them as unimplemented or unreachable. cc @JoshuaBatty

forc-plugins/forc-doc/src/render/title.rs
eureka-cpu: I believe I was told to use unreachable for instances like this one. Initially, I would have used panic. I forget what logic was presented to me, but it was enough to make me change it to unreachable.

forc-plugins/forc-doc/src/render/index.rs
eureka-cpu: The `ref` keyword here is used since you can't use `&`, they are essentially equivalent. I'm not sure why rust made the syntax this way.

forc-plugins/forc-doc/src/render/index.rs
eureka-cpu: Again, used in instances where I used a match statement and all other types to match on will never be reached

forc-plugins/forc-doc/src/render/item/context.rs
eureka-cpu: Makes sense

forc-plugins/forc-doc/src/render/item/context.rs
eureka-cpu: Not sure why my brain thought this was enough info for someone else to understand what this does lol thanks

forc-plugins/forc-doc/src/render/mod.rs
eureka-cpu: I can probably do you one better and add an `anyhow!` error message

forc-plugins/forc-doc/src/render/item/context.rs
eureka-cpu: After looking at this a bit more, the size of the `String` created in each iteration cannot be known until it's created. The field/variant/methods are not the same size as the `String` created from them.

Braqzen (approved):

forc-plugins/forc-doc/src/render/title.rs
JoshuaBatty: I'm happy with it staying as `unimplemented!`

JoshuaBatty (approved):

Issue #4218, opened 2023-03-03
Local constants do not require initializers


mohammadfawaz:
E.g.
```rust
script;

fn main() {
    const X:u64;
}
```
This should fail to compile but it doesn't. Moving the declaration to module scope causes the compiler to emit an error, but the local case doesn't seem to be handled.

mohammadfawaz:
I believe this is a regression in https://github.com/FuelLabs/sway/pull/4140. Assigning to you @tritao.

PR #4385, opened 2023-04-03
Disallow block-level const declarations without a value expression.

tritao:
Closes https://github.com/FuelLabs/sway/issues/4218.

## Description

As the title says, disallows block-level const declarations without a value expressions.

## Checklist

### DIFF
--- a/sway-lib-std/src/u256.sw
+++ b/sway-lib-std/src/u256.sw
@@ -478,15 +478,27 @@ impl core::ops::Multiply for U256 {
                     result_d_d.lower,
                 ))
             } else {
+                // note, that `self.a`, `self.b`, `other.a`, `other.b` are all equal to 0
                 let result_c_c = other.c.overflowing_mul(self.c);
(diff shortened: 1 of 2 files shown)


sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
```diff
@@ -97,6 +104,7 @@ fn item_to_ast_nodes(
     engines: Engines<'_>,
     item: Item,
     is_root: bool,
+    is_inside_code_block: bool,
```
mohammadfawaz, line 107: I prefer something like `is_associated_const_in_trait` which is only `true` when the `const` is an associated constant in a trait definition, and `false` otherwise.

mohammadfawaz:
> Updated PR with a simpler approach.

Does the new approach still allow associated consts in traits to not be initialized?

mohammadfawaz (approved):
Great! Thanks.

PR #4393, opened 2023-04-04
Add conditional compilation by `program_type`

AlicanC:
This is a followup for #4292 which adds the `program_type` argument for the `cfg` attribute.

Should unblock https://github.com/FuelLabs/sway/pull/3829#issuecomment-1398751998


mohammadfawaz:
Very cool!

otrho (commented):
Cool.  Can we add some should-fail tests to confirm a mismatch?

otrho (approved):

Issue #2690, opened 2022-09-01
Remove the `should_fail`/`should_pass` separation in E2E tests


AlicanC:
Our E2E test suite separates tests into two directories: `should_fail` and `should_pass`. IIRC we were using these directories to determine what to expect, but we now do that with `test.toml`s and it doesn't really matter which directory a test sits in.

My suggestion is that we should not organize tests in this way. If you are working on a feature, all tests, failing or not, should be in the same directory.

So now it looks like:
```
test/src/e2e_vm_tests/test_programs/should_pass/passing_test_for_feature_a
test/src/e2e_vm_tests/test_programs/should_fail/failing_test_for_feature_a
```

And my suggestion is:
```
test/src/e2e_vm_tests/test_programs/feature_a/passing_test
test/src/e2e_vm_tests/test_programs/feature_a/failing_test
```

otrho:
I considered this when introducing the `test.toml` testing, but it would've been too big a diff to do it then.  But I agree, we really don't need to organise based on the test type, but rather what it is actually testing.

PR #4394, opened 2023-04-04
Begin phasing out test directories `should_pass` and `should_fail`

AlicanC:
This PR is an attempt to fix https://github.com/FuelLabs/sway/issues/2690.

Challenges I've identified:
- These paths are hot and modifying them can cause a lot of conflicts.
- It isn't easy to figure out where everything should go.

The idea here is:
- We begin a migration with the merge of this PR.
- We put a notice in the repo describing what's up.
- We assign team leads as code-owners so they watch the progress. They prevent new code being added and encourage migration.
- When all is done we remove the notice and unassign code-owners.


AlicanC:
@otrho @mohammadfawaz asked for reviews to get your initial thoughts on this.

otrho:
Sure, sounds good to me.  It's not super high priority though, so it can be a gradual change.

PR #4407, opened 2023-04-06
feat: disallow nested packages

### LINKED ISSUE #4072: Conservatively disallow nesting a package within a package
In today's tooling sync there was some discussion w.r.t. some ambiguities in forc behaviour around handling packages that are nested within another package.

E.g. currently, there's nothing stopping a user nesting a package within the src directory of another package. This makes it tricky for tools like `forc fmt` to reliably only operate on workspace *members* without actually parsing the sway files to learn what modules are a part of the project, and which ones aren't.

For now, we may wish to conservatively disallow nesting a forc package within another package to avoid these ambiguous situations.

We could use a simple function where given a path to a package's forc.toml, we walk the directory and ensure that there are no other occurences of a `forc.toml` throughout the package.

We should avoid doing this a lot unnecessarily as some users might have large package directories and the walk could be expensive. We probably only want to call this once for each member at the beginning of a command invocation, after the initial workspace or package manifest is loaded.

kayagokalp:
## Description

closes #4072 

Disallows package structure such as the following:

```console
proj1
├── proj2
│   ├── Forc.lock
│   ├── Forc.toml
│   ├── out
│   └── src
├── Forc.lock
├── Forc.toml
├── out
│   └── debug
└── src
    └── main.sw
```

## Checklist


kayagokalp:
blocked by #4412

forc-util/src/lib.rs
```diff
@@ -82,22 +102,22 @@ pub fn find_parent_dir_with_file(starter_path: &Path, file_name: &str) -> Option
     None
 }
 /// Continually go up in the file tree until a Forc manifest file is found.
-pub fn find_manifest_dir(starter_path: &Path) -> Option<PathBuf> {
+pub fn find_manifest_dir_in_parent(starter_path: &Path) -> Option<PathBuf> {
```
mitchmindtree, line 105:
```suggestion
pub fn find_manifest_dir_in_parents(starter_path: &Path) -> Option<PathBuf> {
```
Maybe we should use plural here to as the manifest might not be in the immediate parent, but some grandparent.

mitchmindtree: Thinking further, maybe a simpler name would be:
```suggestion
pub fn find_parent_manifest_dir(starter_path: &Path) -> Option<PathBuf> {
```
This might be a little more consistent with the `find_parent_dir` and `find_nested_manifest_dir` function names? Maybe we should change the `with_check` function too for consistency? I'm picturing:
- `find_parent_dir_with_file`
- `find_nested_dir_with_file`
- `find_parent_manifest_dir`
- `find_nested_manifest_dir`
- `find_parent_manifest_dir_with_check`

forc-util/src/lib.rs
```diff
@@ -64,6 +64,26 @@ pub fn format_log_receipts(receipts: &[fuel_tx::Receipt], pretty_print: bool) ->
...
+        .into_iter()
+        .filter_map(|e| e.ok())
+        .filter(|entry| entry.file_name().to_string_lossy() == constants::MANIFEST_FILE_NAME)
+        .map(|entry| {
+            let mut entry = entry.path().to_path_buf();
+            entry.pop();
+            entry
+        })
+        .nth(1)
+}
```
mitchmindtree, line 85: Perhaps we should split this into `find_nested_dir_with_file` and `find_nested_manifest_dir` like we do for the parent traversal functions?

forc-util/src/lib.rs
```diff
@@ -64,6 +64,26 @@ pub fn format_log_receipts(receipts: &[fuel_tx::Receipt], pretty_print: bool) ->
...
+    WalkDir::new(starter_path)
+        .into_iter()
+        .filter_map(|e| e.ok())
+        .filter(|entry| entry.file_name().to_string_lossy() == constants::MANIFEST_FILE_NAME)
+        .map(|entry| {
+            let mut entry = entry.path().to_path_buf();
+            entry.pop();
+            entry
+        })
+        .nth(1)
```
mitchmindtree, line 84: I'm not sure that `.nth(1)` is what we want here, as I think it assumes `starter_path` will always contain a manifest file :thinking: This might break in the case that this function is used to search for a nested manifest directory starting from a directory that does not already contain a manifest. I realize the doc comment mentions this, but perhaps we can make this function a little more intuitive/flexible by removing this constraint?

Instead, perhaps we should use a filter or something to check that the `entry.path()`'s directory is not equal to the starter path directory?

mitchmindtree (commented):
Nice, this will be handy! Just a few small suggestions, otherwise lgtm

forc-util/src/lib.rs
kayagokalp: Sounds great!

mitchmindtree (commented):
Nice! One last nit

forc-util/src/lib.rs
```diff
@@ -65,26 +65,31 @@ pub fn format_log_receipts(receipts: &[fuel_tx::Receipt], pretty_print: bool) ->
...
+/// Continually go down in the file tree until a specified file is found.
+///
+/// Starts the search from child dirs of `starter_path`.
+pub fn find_nested_dir_with_file(starter_path: &Path, file_name: &str) -> Option<PathBuf> {
     use walkdir::WalkDir;
     WalkDir::new(starter_path)
         .into_iter()
         .filter_map(|e| e.ok())
-        .filter(|entry| entry.file_name().to_string_lossy() == constants::MANIFEST_FILE_NAME)
+        .filter(|entry| entry.path() != starter_path.join(file_name))
```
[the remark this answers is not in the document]
kayagokalp: Ohh makes sense!

mohammadfawaz (approved):

mitchmindtree (approved):
:ok_hand:

Issue #4413, opened 2023-04-07
Improve latency in `parse_project`


sdankel:
In codespaces, parse_project consistently takes the same amount of time as on my local machine (between 1.2-1.8 seconds). I noticed we’re calling it more than once per change. I know we do it on every keystroke, but sometimes I’m seeing it happen 5+ times after making just 1 keystroke. Since each call is happening in a separate thread, it makes sense that the latency is so much better on 16 cores vs 2 cores. If we implemented something that cancels a running parse_project  call when it’s been usurped by a new one, or otherwise prevents multiple parse_projects  to run at the same time, I think it would help a lot with latency overall. Another benefit would be that the server would be sending fewer publish_diagnostics  notifications so from the user’s perspective, the squiggly lines won’t be jumping around as much while they’re typing.

I also noticed that the inlay hints load pretty quickly when the project is first opened, but they seem to take progressively longer the more I type. Since we call parse_project  on save as well as on change, I tried disabling autosave to see if it would reduce the number of calls. It did reduce the number of calls and improved latency significantly (inlay hints after less than 2 seconds). I wonder if we could remove the call to parse_project on save entirely, since it should be captured by on_change already.
I’m still not sure why the perf is worse in codespaces. I think it has something to do with the throughput between the codespaces machine and the extension host (which is currently a black box). There is a big delay between when the client thinks it’s sending a message and when the server starts executing the API handler function. It seems like whatever is being used to relay the stdin/stdout to the remote extension is doing it in a more synchronous way than how it works locally.

If we make the LSP more efficient overall (fewer messages passed, especially time consuming & resource blocking ones like parse_project) I think it will help immensely. I’ll start looking into how to queue and cancel `parse_project` threads.

sdankel:
https://github.com/FuelLabs/sway/issues/4275

Issue #3525, opened 2022-12-06
LSP Optimization: Cancel pending futures on new requests.


JoshuaBatty:
The language server should cancel any pending futures that have yet to yield a result if a new request comes in from the client. For example, on every did_change event, we should stop compilation and start again immediately. 

As this event happens on every keystroke, this optimization in itself should lead to a drastic speed improvement.

JoshuaBatty:
Here you can see just how backed up the server can get in it's current state. 

https://user-images.githubusercontent.com/1289413/230806840-d950882c-6703-4d78-9766-3b6de09a6eed.mov

JoshuaBatty:
A good stress test for this is to rename the `Option` type from within `sway-lib-std/src/option.sw`. In doing so, the client sends out a `didChange` notification for each of the files that are impacted, so roughly 10 notifications one after the other.

If we implement this correctly, each subsequent `didChange` notification will cancel any current work done in `parse_project` until the final one comes in, allowing us to complete parsing of the project and be ready to receive new requests from the Client.

JoshuaBatty:
This was mostly addressed in #5429 

We can now bail out of compilation if a newer did_change notification message is received

PR #4414, opened 2023-04-07
Optimize LSP `parse_project`

sdankel:
## Description

Closes https://github.com/FuelLabs/sway/issues/4413

This change drastically improves the UX of the VSCode plugin in github codespaces. It's slightly noticeably faster running locally too. The key optimizations are:

#### Only one thread of `parse_project` can run at a time

- A write lock taken on `diagnostics` function forces the program to wait until the previous `parse_project` execution is finished
- A semaphore with 2 permits accessed with `try_acquire` means that only 1 thread of `parse_project` can wait on the mutex. When the first one finishes, the waiting one will run against the latest version of the user's code. All others will not be able to acquire a permit and will exit immediately.

This was a big improvement because previously, as the user was typing, 5+ threads of `parse_project` would kick off simultaneously, all operating on the same data stores (taking turns with read & write locks). We only really care that the most recent version of the user's code is compiled & indexed.

#### Hot swapping the engines with write lock

Previously, we only used a write lock to clear the engines, then read locks to do the rest of the parsing. After making the change to limit the # of parsing threads and the subsequent latency improvement, I encountered an issue where we'd sometimes request stale TypeIds from the engines. To prevent this, I'm using a new set of engines for compilation (`pkg::check`), and only storing them in the session once compilation is complete. The write lock only applies to the swapping process, but ensures that no readers will read incomplete data.

- This also means we clear the data stores _after_ compilation is complete.
- I had to replace the `read` lock from `create_runnables` since it caused a deadlock (`create_runnables` is called in the section of code that is now holding a write lock)

#### Spawn blocking for parse_project

We're now calling the synchronous `session::parse_project` in a [spawn_blocking](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) thread so that it doesn't block tokio's runtime, allowing other requests like `inlay_hints` to be processed on whatever runtime threads are available.

#### Publish compiler diagnostics only when they're fresh

Since we have "passive" threads of `parse_project` that simply wait for the active threads to finish, there's no need for the passive threads of `parse_project` to publish their results, since they'll be the same as what the active threads before them have gathered. In practice this cuts the number of `publishDiagnostic` messages roughly in half, saving resources for the server to handle to other requests. It also reduces the amount of squiggly lines jumping around reporting warnings/errors while the user is typing.

#### Performance

Here's a video of it running in github codespaces on a **2-core** machine - the smallest option available. Previously it was taking upwards of 1 minute to render inlay hints, mostly due to requests timing out and getting cancelled. Now it takes 2ms  plus the ~1 second of compile time :-)

https://user-images.githubusercontent.com/47993817/230686772-a521c072-19c5-4729-8dad-a48630b8735d.mp4

#### Notes

I had to increase the timeout on a test that waits for diagnostics to be published. I suspect this means that parsing takes slightly longer because of the locks, but overall the LSP performs better because it's doing less throwaway work.

## Checklist


sway-lsp/src/server.rs
```diff
@@ -62,23 +63,30 @@ impl Backend {
...
                 }
             }
-        };
-        self.publish_diagnostics(&uri, &workspace_uri, session, diagnostics)
-            .await;
+        })
+        .await
+        .unwrap_or_default();
+
+        if let Some(diagnostics) = diagnostics {
```
sdankel, line 86: Previously we were publishing empty diagnostics if there were any errors. This meant that all compile errors/warnings would temporarily disappear. We should only publish empty diagnostics when we get them from the compiler.

sway-lsp/src/core/session.rs
```diff
@@ -109,8 +122,15 @@ impl Session {
     }
 
     pub fn parse_project(&self, uri: &Url) -> Result<Diagnostics, LanguageServerError> {
-        self.token_map.clear();
-        self.runnables.clear();
+        // Acquire a permit to parse the project. If there are none available, exit early.
```
JoshuaBatty, line 125: I think we might want to approach this differently. If we return here, then the workspace is out of sync with the server. Ideally, we would cancel any on going work and reparse with the most up to date information from the client. 

This definitely isn't a trivial thing to implement. Ideally, we would be able to take advantage of the [cancelRequest](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#cancelRequest) part of the LSP spec to drop any current work. Unfortunately, `parse_project` is getting triggered by `didOpen` which is a notification and not an LSP request. It looks like `rust-analyzer` also bumped into this problem with notifications, see [666 ](https://github.com/microsoft/language-server-protocol/issues/666) and [584](https://github.com/microsoft/language-server-protocol/issues/584)

There is also a really good discussion [here](https://github.com/ebkalderon/tower-lsp/issues/284) where `tower-lsp` is considering ditching concurrent execution and also notes the same issues above. 

This is work that I was planning on trying to solve in #3525.

JoshuaBatty:
Awesome to see this will unlock codespaces! Huge win. Just ran the branch locally. When pressing 'enter' at the beginning of a sway file, the semantic highlighting for the tokens becomes incorrect. Seems like now the client, server and our hacky current temp file system are becoming out of sync. 

Screen Shot 2023-04-10 at 11 34 22 am

sdankel:
> Awesome to see this will unlock codespaces! Huge win. Just ran the branch locally. When pressing 'enter' at the beginning of a sway file, the semantic highlighting for the tokens becomes incorrect. Seems like now the client, server and our hacky current temp file system are becoming out of sync.
> 
> Screen Shot 2023-04-10 at 11 34 22 am

Thanks for catching this! It should be fixed now if you want to try again. The solution was to add a `wait_for_parsing` that waits for a read lock on `diagnostics`. Tested in codespaces as well.

sway-lsp/src/server.rs
```diff
@@ -62,23 +63,26 @@ impl Backend {
...
-        self.publish_diagnostics(&uri, &workspace_uri, session, diagnostics)
-            .await;
+        })
+        .await
+        .unwrap_or_default();
+
+        if should_publish {
+            self.publish_diagnostics(&uri, &workspace_uri, session)
+                .await;
+        }
```
JoshuaBatty, line 85: Perhaps something like this is cleaner?

```rust
async fn parse_project(&self, uri: Url, workspace_uri: Url, session: Arc<Session>) {
    let should_publish = run_blocking_parse_project(uri.clone(), session.clone()).await;
    if should_publish {
        self.publish_diagnostics(&uri, &workspace_uri, session).await;
    }
}
```

and then have a free function declared outside of `Backend`

```rust
async fn run_blocking_parse_project(uri: Url, session: Arc<Session>) -> bool {
    task::spawn_blocking(move || {
        match session.parse_project(&uri) {
            Ok(should_publish) => should_publish,
            Err(err) => {
                tracing::error!("{}", err);
                matches!(err, LanguageServerError::FailedToParse)
            }
        }
    })
    .await
    .unwrap_or_default()
}
```

JoshuaBatty (dismissed):
Just left one suggestion, happy to approve this either way.

Pulled down and tested locally, response times are so much faster. Really awesome work! 🚀

JoshuaBatty (approved):

sway-lsp/src/core/session.rs
```diff
@@ -60,6 +61,11 @@ pub struct Session {
     pub type_engine: RwLock<TypeEngine>,
     pub decl_engine: RwLock<DeclEngine>,
     pub sync: SyncWorkspace,
+    // Limit the number of threads that can wait to parse at the same time. One thread can be parsing
+    // and one thread can be waiting to start parsing. All others will return the cached diagnostics.
+    parse_permits: Arc<Semaphore>,
+    // Cached diagnostic results that require a lock to access. Readers will wait for writers to complete.
+    diagnostics: Arc<RwLock<Diagnostics>>,
```
mitchmindtree, line 68: As joshy mentioned in the sync, it'd be nice if tower-lsp had a sync option (rather than assuming an async API) so we wouldn't have to manually sync access like this! Alas, this looks fine in the meantime :+1:

mitchmindtree (approved):

sway-lsp/src/core/session.rs
sdankel: It turns out there is kind of a sync option: we can set max_concurrency=1. When I tried it out locally, the LSP ran much slower. I also read that `cancelRequest` only works when tower-lsp is running with multiple threads. 

I think we'd need to bring compilation time down a lot for the LSP to be usable running synchronously. From what I can tell, most of the problems with async LSP stem from the ~1 second it takes to compile the project, since LSP needs to constantly recompile when anything changes in the editor.

PR #4423, opened 2023-04-11
Allows configurable consts to be used in storage.

### LINKED ISSUE #4357: Allow `configurable` constants to be used in `storage`
The manifest file allowed users to set the previously termed `config-time constants` and use them in storage blocks for initialization.

The new `configuration {}` block does not contain this feature meaning you cannot set a value in the `configuration` block (which is treated as a `const`) and then initialize it inside `storage` as well.

To reproduce this on `forc v0.35.5` you may `forc build` the following code to reproduce the following error.

__Repro__

```rust
contract;

configurable {
  INITIAL_OWNER: Option<Identity> = Option::None,
}

storage {
  owner: Option<Identity> = INITIAL_OWNER,
}

abi MyContract {
    fn test_function() -> bool;
}

    fn test_function() -> bool {
        true
    }
}
```

__Error__

esdrubal:
## Description

Before this fix configurables were not found because we were doing the lookup of only constants.

This commit changes the lookup to also get configurables.

Closes #4357

## Checklist


mohammadfawaz (approved):
Looks great! Thanks.

PR #4428, opened 2023-04-11
chore: bump to fuel-vm 0.30 and latest (0.39-0.38) fuels-*

### LINKED ISSUE #4402: Update `sway/Cargo.toml` dependencies
[The following dependencies in `Cargo.toml`](https://github.com/FuelLabs/sway/blob/6d8b7de00aca951ed5934f181f13b781534aa790/Cargo.toml#L35) need to be updated. Almost all of the changes required to make this update possible are in `forc-tx` and `forc test`.

kayagokalp:
## Description
closes #4402.

## Checklist

### DIFF
--- a/forc-plugins/forc-client/src/op/run.rs
+++ b/forc-plugins/forc-client/src/op/run.rs
@@ -9,7 +9,7 @@ use anyhow::{anyhow, bail, Context, Result};
 use forc_pkg::{self as pkg, fuel_core_not_running, PackageManifestFile};
 use forc_util::format_log_receipts;
 use fuel_core_client::client::FuelClient;
-use fuel_tx::{ContractId, Transaction, TransactionBuilder, UniqueIdentifier};
+use fuel_vm::prelude::*;
 use futures::TryFutureExt;
 use pkg::BuiltPackage;
 use std::time::Duration;
@@ -20,7 +20,7 @@ use tokio::time::timeout;
 use tracing::info;
 
 pub struct RanScript {
-    pub receipts: Vec<fuel_tx::Receipt>,
+    pub receipts: Vec<Receipt>,
 }
 
 /// Builds and runs script(s). If given path corresponds to a workspace, all runnable members will
@@ -100,7 +100,7 @@ async fn try_send_tx(
     tx: &Transaction,
     pretty_print: bool,
     simulate: bool,
-) -> Result<Vec<fuel_tx::Receipt>> {
+) -> Result<Vec<Receipt>> {
     let client = FuelClient::new(node_url)?;
 
     match client.health().await {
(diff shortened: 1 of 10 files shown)


forc-plugins/forc-tx/src/lib.rs
```diff
@@ -234,7 +234,6 @@ pub struct TxPointer {
 pub enum Output {
     Coin(OutputCoin),
     Contract(OutputContract),
-    Message(OutputMessage),
```
kayagokalp, line 237: Looks like newest fuel-tx release removede `Message` from output.

forc-plugins/forc-tx/src/lib.rs
```diff
@@ -784,36 +790,41 @@ impl TryFrom<Input> for fuel_tx::Input {
...
-                        recipient,
-                        amount,
-                        nonce,
-                        data,
-                        witness_index,
-                    },
-                    (None, Some(predicate), Some(predicate_data)) => {
-                        fuel_tx::Input::MessagePredicate {
-                            message_id,
+                    (Some(witness_index), None, None) => fuel_tx::Input::MessageDataSigned(
```
kayagokalp, line 793: Looks like we now have `MessageCoinSigned` and `MessageDataSigned` introduced to `fuel-tx`. 

As far as I understood if data is present we should be creating `MessageDataSigned` as `MessageCoinSigned` needs an empty tuple for the data field. 

Should we make data optional and add it to match statement so that we also generate `MessageCoinSigned`?

kayagokalp (commented):
I have some unresolved issues so this is still not building. I resolved most of the stuff but having some troubles in `forc-client`. It throws bunch of `= note: perhaps two different versions of crate `fuel_tx` are being used?` I am only using `fuel_vm::fuel_tx::*` from vm prelude but it looks like `fuels-*` uses `fuel_tx` as well and they do not support the latest version yet. If that is the case and we don't have a known work around to that I might be blocked for now until a new SDK release comes with `fuel-tx 0.30` support.

Also some small stuff changed in the client which broke `forc-tx`. I added a couple of comments to the places that I am not sure.

 cc @mitchmindtree @mohammadfawaz

forc-plugins/forc-client/src/op/run.rs
```diff
@@ -119,8 +119,8 @@ async fn send_tx(
     tx: &Transaction,
     pretty_print: bool,
     simulate: bool,
-) -> Result<Vec<fuel_tx::Receipt>> {
-    let id = format!("{:#x}", tx.id());
+) -> Result<Vec<Receipt>> {
+    let id = format!("{:#x}", tx.id(&ConsensusParameters::default()));
```
kayagokalp, line 123: It looks like id now requires `ConsensusParameters` not sure why but I am providing the default. Not sure if that is how we should handle this.

kayagokalp:
blocked by https://github.com/FuelLabs/fuels-rs/issues/933

kayagokalp:
https://github.com/FuelLabs/fuels-rs/issues/933 requires a new client release. So this can be continued after:

1. Client releases
2. SDK bump happens
3. SDK releases

Issue #1066, opened 2022-01-06
Add dosctrings everywhere in `std` and `core` libraries


sezna:
While `forc doc` is not fully functional yet, we really should be adding docstrings to everything in the standard library.

mitchmindtree:
Now that #2869 is implemented, now would be a great time to get started on this!

The docs are still looking quite raw today, but having fleshed out documentation in `core` and `std` would act as a great real-world test case for landing new features like doc tests, inter-module linking, etc.

eureka-cpu:
Is this still open? I believe the docs in at least std are looking pretty good

mitchmindtree:
Loads of the items and modules in std are still missing docs, and most of those that have docs have something like a single sentence that doesn't give much context about the purpose of the type, how to use it correctly, no examples, etc.

I think the "standard" of std docs we should hold ourselves to is Rust's std: https://doc.rust-lang.org/std/index.html. It totally outclasses us at the moment.

That said, this issue is a bit vague - maybe we're better off opening a tracking issue that sets a standard for how we expect std items to be documented, and a checklist for going through them? All methods/functions should likely have examples at least, however we'd likely be better off tackling #3176 before adding too many examples, otherwise they'll almost certainly fall out of date without close attention.

eureka-cpu:
> Loads of the items and modules in std are still missing docs, and most of those that have docs have something like a single sentence that doesn't give much context about the purpose of the type, how to use it correctly, no examples, etc.
> 
> I think the "standard" of std docs we should hold ourselves to is Rust's std: https://doc.rust-lang.org/std/index.html. It totally outclasses us at the moment.
> 
> That said, this issue is a bit vague - maybe we're better off opening a tracking issue that sets a standard for how we expect std items to be documented, and a checklist for going through them? All methods/functions should likely have examples at least, however we'd likely be better off tackling #3176 before adding too many examples, otherwise they'll almost certainly fall out of date without close attention.

@gr00vytvniks this could be a good issue for you to take on.

gr00vytvniks:
If this is still a relevant issue I'd be happy to take it on :)

bitzoic:
Closed by https://github.com/FuelLabs/sway/pull/4830

PR #4433, opened 2023-04-12
refactor: `forc-doc`- move headers above descriptions and add extra vertical spacing between items

### LINKED ISSUE #4432: Improve readability of forc doc
Right now the std-lib docs generated with forc doc are difficult to read due to the styling format where sub-headings are on the left and descriptions are on the right.

std-lib-docs

Making the headings go above the descriptions with vertical spacing in between each section would improve the readability of the docs significantly.

sarahschwartz:
close #4432

forc-doc-updates


eureka-cpu:
Hey @sarahschwartz would you mind including a screen cap of the changes?

eureka-cpu:
Ah I see what you mean now! I'm not sure if we want the previews under since it will make the page much longer vertically but I'll defer to  @mitchmindtree & @JoshuaBatty

mitchmindtree:
I could go either way on this one, though my inclination is to stick to the table-style approach following suit with docs.rs ([e.g.](https://docs.rs/rand/latest/rand/)), mostly because these pages can become very dense depending on the number of items in the module, and the idea is more to show an overview/list of the items that exist within the module where the user can click in to learn more. 

I guess the problem with the table style approach is that sometimes you have to look close to see which preview is associated with each line. Perhaps we should use some kind of visual cue like a colour offset for every second row to make it a bit clearer?

Happy to go with what you recommend @sarahschwartz as you probably get the most feedback about the forc-doc UX!

eureka-cpu:
> I could go either way on this one, though my inclination is to stick to the table-style approach following suit with docs.rs ([e.g.](https://docs.rs/rand/latest/rand/)), mostly because these pages can become very dense depending on the number of items in the module, and the idea is more to show an overview/list of the items that exist within the module where the user can click in to learn more. 
> 
> I guess the problem with the table style approach is that sometimes you have to look close to see which preview is associated with each line. Perhaps we should use some kind of visual cue like a colour offset for every second row to make it a bit clearer?
> 
> Happy to go with what you recommend @sarahschwartz as you probably get the most feedback about the forc-doc UX!

I like the idea about offsetting the color of each second line but keeping the old layout

sarahschwartz:
Generally tables are used to display charts of data or statistics like a spreadsheet with column and row headers. I don't think it makes sense to use a table format for this type of content, but I do see the possible need to split some of the content up into different pages if they are too long. Pushing the content into a smaller space may allow it to fit in a smaller space, but this becomes less accessible / readable as well.

I also understand the desire to mirror Rust, but I don't believe this type of UI design jn particular is something that was well done or something developers really love about Rust.

kayagokalp:
@sarahschwartz I changed the description of the PR so that it is a little more descriptive in `git log` once it is merged. Feel free to revert it back if you feel like it.

mitchmindtree:
> I don't think it makes sense to use a table format for this type of content, but I do see the possible need to split some of the content up into different pages if they are too long.

Ohhh there might be some confusion w.r.t. the intentions for these module pages!

Originally, these were actually just *lists* of items to act as a point of easy navigation from the module level to the items it contains. The one-liner preview of the docs was added on the right just to show a little more context on what you can navigate into. Most of those item-level docs are very small right now and just *happen* to almost fit into the one line, however eventually (once sway-lib-std is documented _properly_) they be should be much more extensive with multiple tested examples, information about performance trade-offs, context on when and where to use them, etc.

Does this help to clarify the original intent a little? Knowing this, do you still think we might be better with the line breaks and larger previews? I think I'm still open to it, though I think we'd want to have some nice heuristic for trimming previews to a couple lines, or just the first sentence or something so that we don't make it too hard to quickly scan through all the items.

eureka-cpu:
Hm, I still think it may be better keep the original formatting since it aligns with how `cargo doc` produces documentation. If it's a matter of being able to find what you're looking for there is an issue open for organising links alphabetically? In general, `forc doc` is immature in comparison to the `cargo` documenter, and with the growth it will have in the coming month I believe it will make the user interface more friendly and keeping it familiar to rust users is definitely the priority.

In general, I think the right approach is to follow in `cargo doc`'s footsteps. I appreciate the want to make `forc doc` better though, and there are a good number of issues I think would be good first issues to tackle such as the aforementioned alphabetization one.

sarahschwartz:
I'm not confused about this, but I appreciate you explaining your logic. 

For the Dev Rel team, we frequently receive requests to better document the standard library. 
We believe that the official Fuel documentation should eventually (soon) have a consistent Fuel branding and design system, which won't exactly mirror Rust's. Unless you think we should create an entirely separate documentation resource for the standard library, this documentation will eventually have to align with that design system.

This PR was meant to be a sort of quick fix to improve the accessibility of these docs for the time being. I'm definitely open to different designs and solutions, but I strongly believe a table format is not the correct format here.

eureka-cpu:
> Unless you think we should create an entirely separate documentation resource for the standard library, this documentation will eventually have to align with that design system.


If that's the case it may make more sense to do that

> This PR was meant to be a sort of quick fix to improve the accessibility of these docs for the time being. I'm definitely open to different designs and solutions, but I strongly believe a table format is not the correct format here.

I understand the Dev rel team has a goal in mind for how they would like the documentation for the standard library to look, but I don't think it makes sense to change the look of the documentation produced by `forc doc` just to serve this purpose.

I would recommend working out the design for the standard library documentation in an issue and if it's decided that the structure of `forc doc` should stray from `cargo`'s we can go from there.

mitchmindtree (approved):
> For the Dev Rel team, we frequently receive requests to better document the standard library.
We believe that the official Fuel documentation should eventually (soon) have a consistent Fuel branding and design system, which won't exactly mirror Rust's. Unless you think we should create an entirely separate documentation resource for the standard library, this documentation will eventually have to align with that design system.

Thanks for sharing some more context! I think no-one is in a better position to make these kinds of design/style/feedback judgements than you guys, so if you feel it's best I'm happy to see this land - especially if it's a part of a wider effort for consistency across all Fuel docs.

I'm excited to see devrel dig into the std doc content, it's definitely sorely lacking atm :smiling_face_with_tear: fwiw, I left some thoughts [here](https://github.com/FuelLabs/sway/issues/1066#issuecomment-1442685135) recently.

> I understand the Dev rel team has a goal in mind for how they would like the documentation for the standard library to look, but I don't think it makes sense to change the look of the documentation produced by forc doc just to serve this purpose.

I think if the std docs can benefit, then likely the wider `forc doc`-generated docs might also benefit anyways. Eventually, I'm imagining we'll host *all* ecosystem package docs under some `docs.rs` style domain like `docs.fuel.network/std`, `docs.fuel.network/core`, `docs.fuel.network/nft`, etc (or something along these lines) and we'll likely want some consistency.

---

Just tested locally, LGTM

JoshuaBatty (approved):

kayagokalp (approved):

PR #4454, opened 2023-04-17
Bugfix: use the index param in std::tx::tx_witness_pointer

### LINKED ISSUE #4429: `std::tx::tx_witness_pointer` does not use `index`
See: https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/tx.sw#L114-L119. Not sure why `index` is not used but it probably should be according to the spec of `gtf`.

nfurfaro:
## Checklist

Closes #4429


mohammadfawaz:
Awesome. Is it possible to add a test for this? Or is this too complicated to do in the SDK test?

mohammadfawaz:
> >Awesome. Is it possible to add a test for this?
> 
> 
> 
> I could add another test. The current test was passing `0` as the index, and while the index param was not being used, `0` was hardcoded. So this was already tested, and coincidentally the change to the stdlib didn't require any changes to the test.
> 
> https://github.com/FuelLabs/sway/blob/abddf732125d8391f37ee5459c4ddad2c0abdb38/test/src/sdk-harness/test_projects/tx_fields/mod.rs#L309
> 
> 

Yeah I'm just thinking of the non-zero index case that should pass now (but would have failed before)!

nfurfaro:
Got it. 
I'll work out how to sign the tx a second time with another wallet. 
Otherwise, there's only 1 witness and any index > 0 fails.

Braqzen (approved):
It would be good to do something about the magic number `11032` in the assert since idk what that is or why it changed.

PR #4455, opened 2023-04-17
refactor: Introduce `ForcResult` and `ForcError` to propagate return code with error message 

### LINKED ISSUE #4420: Provide a way to propagate custom exit status with forc operations
Realized while working on #4417.  

We are using `anyhow::Result` return types all around forc cli but maybe we should introduce a `ForcError` or something like that to contain some custom return status. After propagating the custom error to the entry point (main function in forc) we could then return `ExitCode` so that necessary clean-up/destructions are executed or we could also implement `Termination` for `ForcError` and return `ForcError`.

Note that directly using `std::process:exit()` is not preferable as it disrupts the current execution and directly terminates without necessary destructions executed (see the docs [here](https://doc.rust-lang.org/std/process/fn.exit.html)).  But by returning a struct that implements `Termination` we can make sure that necessary clean-ups are done before terminating the process and we could still exit with custom error code.

kayagokalp:
## Description
closes #4420.

This PR introduces `ForcResult` and `ForcError` which enables us to return with custom exit code via propagating the desired exit code alongside the error message. Using `ForcResult` we can exit with different return code for different cases. Cargo uses similar approach for similar purposes.

## Checklist

### DIFF
--- a/forc-util/src/lib.rs
+++ b/forc-util/src/lib.rs
@@ -70,6 +70,19 @@ impl Termination for ForcError {
     }
 }
 
+#[macro_export]
+macro_rules! forc_result_bail {
+    ($msg:literal $(,)?) => {
+        return $crate::ForcResult::Err(anyhow::anyhow!($msg).into())
+    };
+    ($err:expr $(,)?) => {
+        return $crate::ForcResult::Err(anyhow::anyhow!($err).into())
+    };
+    ($fmt:expr, $($arg:tt)*) => {
+        return $crate::ForcResult::Err(anyhow::anyhow!($fmt, $($arg)*).into())
+    };
+}
+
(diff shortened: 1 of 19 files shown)


eureka-cpu (approved):
Nice! Would we be able to use this in the forc plugins as well?

JoshuaBatty (approved):

kayagokalp:
> Nice! Would we be able to use this in the forc plugins as well?

@eureka-cpu  Yes we can use it, this should work as long as the entry point in the plugin binary returns an `ForcResult` 👌 

I will be using this to return different return codes between failing tests vs failing `forc-test` execution. If you have some plugins in mind that requires a difference like that we can open tracking issue for them.

PR #4459, opened 2023-04-18
Fixes `Self` not found in match statement.

### LINKED ISSUE #4263: `Self` does not work when qualifying an enum variant in a `match` statement
Example:
```rust
script;
enum Foo {
    X: u64,
}

impl Foo {
    fn foo(self) -> bool {
        match self {
            Self::X(_) => true, // Complains that `Self` does not exist
        } 
    }
}

fn main() {}
```
Rust equivalent that works: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fb615cf6f202e2e307c9e6d4bb7b6fe4

esdrubal:
## Description

`Self` was not working when qualifying an enum variant in a match statement.

The solution was to add the implementing type declaration as an impl_trait namespace symbol named "Self".

Closes #4263.

## Checklist


anton-trunov:
I might be wrong but I'm thinking maybe reusing the `replace_self_type` function is the way to go here?

esdrubal:
> I might be wrong but I'm thinking maybe reusing the `replace_self_type` function is the way to go here?

It looks like we are trying to resolve the enum scrutinee from its call path during its `type_check` which occurs before `replace_self_type` is called.
https://github.com/FuelLabs/sway/blob/30f869a3a845671263e5a70102bc31644e4bf29a/sway-core/src/semantic_analysis/ast_node/expression/match_expression/typed/typed_scrutinee.rs#L272-L278

IGI-111 (approved):

JoshuaBatty (approved):

Issue #4304, opened 2023-03-16
Split `storage.sw` into modules


Braqzen:
## Motivation

The file has multiple types and it's starting to become large.
We should split the file into modules and reexport to maintain functionality but compartmentalise the code.

mohammadfawaz:
It would be nice to do this after https://github.com/FuelLabs/sway/issues/3487 is closed so that importing items from the storage sub-libraries is not too verbose (assuming we're splitting `storage.sw` into submodules inside a `storage` folder).

Issue #3419, opened 2022-11-23
Document how storage slots are assigned


mohammadfawaz:
Document how the compiler assigns storage slots given:
* The data type.
* The storage index in the `storage` block.
* Whether the data is in a `StorageMap`/`StorageVec` or regular `storage` block. This _shouldn't_ really matter but currently does matter unfortunately due to a variety of reasons that are going to be resolved soon once we have range storage.

This is blocked until range storage is fully implemented and supported in the compiler.

Issue #3202, opened 2022-10-30
Reconsider the design of `StorageMap` and `StorageVec` to avoid the use of `__get_storage_key` (while staying backward compatible)


mohammadfawaz:
When storage references get implemented, we should move away from using `__get_storage_key`. This was a _hacky_ way to get `StorageMap` working but now that we're about to have storage references, I think we can do better. 

Currently, the method insert looks like:
```rust
    fn insert(self, key: K, value: V) {
        let key = sha256((key, __get_storage_key()));
        store::<V>(key, value);
    }
```
and is called using `storage.map.insert(key, value)`, which is equivalent to `~StorageMap::insert(storage.map, key, value)`.

When we have storage references, we can imagine something like:
```rust
    fn insert(ref self, key: K, value: V) { // ref self or `StorageRef<Self>` or something like that
        let key = /* extract this from the reference to self */
        store::<V>(key, value);
    }
```
and the syntax for calling `insert` would remain the same: `storage.map.insert(key, value)` and would be equivalent to `~StorageMap::insert(storage.map.ref(), key, value)`.

Of course, a lot of details and the syntax have to be worked out, but I think this is **a much more robust way of implementing `StorageMap` that does not actually break existing code**. This also enables passing storage maps around to functions and also writing:
```rust
let map_ref = storage.map.ref();
map_ref.insert(key, value);
```

One more motivation for the above is that I actually have doubts about the intrinsic `__get_storage_key` if the method using it (like `insert()` or `get()`) is **not** inlined 🤔 

### Note
Equivalent Rust code for a simple struct would look like
```rust
struct S {
    x: u64,
}

impl S {
    fn foo(&self) -> u64 {
        self.x
    }
}

fn main() {
    let s = S { x: 5 };

    // These two are equivalent
    let x = s.foo(); // Effectively taking the ref here before calling `foo()`
    let x = S::foo(&s);
}
```

mohammadfawaz:
@mitchmindtree touched on this subject a bit here https://github.com/FuelLabs/sway/issues/2585

sezna:
I like this -- the only handwavy part we need to sort out is the "extract the storage key from the reference". I suppose that'd be through `StorageRef`'s stdlib API, which would perhaps use `__get_storage_key()`?

mohammadfawaz:
[...]

Issue #3043, opened 2022-10-14
Introduce storage references (or `StorageKey`)


mohammadfawaz:
## Motivation

While the current use model model for storage is quite ergonomic, it is missing an important element to enable important features such as:

1. The ability to read/modify an arbitrary storage variable in functions.
2. The ability to read/modify an arbitrary storage variable in libraries.
3. The ability to compose storage collections such as `StorageVec<StorageVec<T>>`.
4. The ability to have a storage collection in a struct.

We always seem to come back to the idea of a "storage reference" every time any of the above use cases are discussed.

## Proposals

There are two ways of introducing storage references.

### Method 1

The first approach changes the expression `storage.<>` to return a storage reference. The reference can then be passed to function arguments and can be de-referenced via `deref()`. So, code that previously looked like
```rust
storage.x += 1;
```
would now look like
```rust
storage.x.deref() += 1;
```

The type of `storage.x` in this case would be some generic type called `StorageRef<T>` which contains enough information about the "location" of the storage variable. This could be a simple index or a list of indices in the case of a nested struct (i.e. whatever is needed to generate the storage keys for the variable).

### Method 2

The second approach is very similar to the the first one except that it does not modify the meaning of `storage.<>`. Instead, we make storage references opt-in via an `as_ref()` call or a `ref` prefix such as:
```rust
let some_ref = storage.x.as_ref();
```
or
```rust
let some_ref = ref storage.x;
```

## Implementation Details

A lot of the implementation details still have to be worked out, but we will need a special type called `StorageRef<T>` that represents the reference type in question. We have to decide on what it contains exactly and how "special" it is. 

## Code examples

Here's how accessing a storage variable in a library would look like:

```rust
// Library
library lib;

fn increment_var(storage_ref: StorageRef<u64>) {
    storage_ref.deref() += 1;
} 
```

```rust
// Contract

storage {
    x: u64,
}

fn some_contract_method() {
    // Local updates
    storage.x += 1; // or storage.x.deref() += 1 if we go with Method 1
    
    // Library update
    increment_var(storage.x.as_ref()); // or increment_var(storage.x) if we go with Method 1
}
```

Edit: I've started working on [this RFC](https://github.com/FuelLabs/sway-rfcs/pull/23) which follows mostly Method 2 above. Before we fully implement that RFC, we need to implement the following:
- [x] #4045
- [ ] #4010
- [x] #4046
- [x] #4047

estill01:
How practical would it be to do this?:

[...]

PR #4464, opened 2023-04-18
Replace old storage API with the new one explained in the `StorageKey` RFC

mohammadfawaz:
## Description
Implement https://github.com/FuelLabs/sway-rfcs/pull/23

Closes https://github.com/FuelLabs/sway/issues/3419 (technically via the RFC)
Closes https://github.com/FuelLabs/sway/issues/3202
Closes https://github.com/FuelLabs/sway/issues/3043
Closes https://github.com/FuelLabs/sway/issues/2639
Closes https://github.com/FuelLabs/sway/issues/2465
Closes https://github.com/FuelLabs/sway/issues/4304

This is a big one but mostly removes stuff. It's hard to review but the summary of changes below should be sufficient. Most of the changes here are things that I just had to do to make CI pass.

- Removed the `--experimental-storage` flag and made its [behavior](https://github.com/FuelLabs/sway/pull/4297) the default. Also removed everything that was required for the previous storage APIs in favour of the new ones.
- Break down the `std::storage` into multiple submodules:
 - `storage_key` implements the API for `std::core::StorageKey`
 - `storage_api` implements the free functions `read` (previously `get`), `write` (previously `store`), and `clear`.
 - 4 more modules for the dynamic storage types which now use the new `StorageKey` API.
- `#[storage(write)]` now allows reading from storage as well. This is needed because the way we pack structs in storage now requires that we sometimes read from storage first if we were to write a portion of a slot.
- Removed the "storage only types" checks and the corresponding tests.
- Removed the `__get_storage_key` intrinsic and the `get_storage_key` IR instruction and all corresponding tests. Also removed the `state_index` metadata as it is no longer required.
- Added tests and example to showcase nested storage maps and nested storage vectors.

## Checklist


sway-lsp/tests/lib.rs
```diff
@@ -913,95 +913,6 @@ async fn go_to_definition_for_traits() {
     shutdown_and_exit(&mut service).await;
 }
 
-#[tokio::test]
```
JoshuaBatty, line 916: I've opened #4482 as a note to add this test back in, but I'll do that later so we don't hold this PR up.

mohammadfawaz: I thought this test was testing for StorageReassignment since that's the only place the `storage` symbol kind was used and I removed that now that's it not used. 🤔

JoshuaBatty: I just pulled down and installed this branch and it does seem like there has been a regression somewhere. Previously this test was testing go to definition for `storage` , `var1`, and `x`. 

```rust
storage.var1.x = 42;
```

I installed your branch and manually tested go_to for the updated line but it doesn't find anything for each of the 3 tokens. 

```rust
storage.var1.x.write(42);
```

mohammadfawaz: Hmm mainly because storage reassignments are completely removed so the LSP logic that I removed now has to be rewritten with the new API?

sway-lsp/src/traverse/typed_tree.rs
```diff
@@ -1069,88 +1066,6 @@ impl Parse for ty::TyReassignment {
...
-                            .fields
-                            .into_iter()
-                            .find(|f| f.name.as_str() == head_field.name.as_str())
-                    })
-                {
-                    token.type_def = Some(TypeDefinition::Ident(storage_field.name));
-                }
-            }
-            // collect the rest of the idents as fields of their respective types
-            for (field, container_type_id) in tail_fields
```
JoshuaBatty, line 1114: Yeah it looks like it's this block that had the go to logic working. Inspecting the AST and it seems we can move this logic when looping over the fields in `TyExpressionVariant::StorageAccess`

```rust
TyStorageAccess {
    fields: [
        TyStorageAccessDescriptor {
            name: BaseIdent {
                name_override_opt: None,
                span: Span {
                    src (ptr): 0x000000010ef27890,
                    path: Some(
                        "/private/var/folders/3z/79mvj_fn5298rz6zb4gnctw00000gn/T/SWAY_LSP_TEMP_DIRQs3zQJ/storage/src/main.sw",
                    ),
                    start: 345,
                    end: 349,
                    as_str(): "var1",
                },
                is_raw_ident: false,
            },
            type_id: TypeId(
                67840,
            ),
            span: Span {
                src (ptr): 0x000000010ef27890,
                path: Some(
                    "/private/var/folders/3z/79mvj_fn5298rz6zb4gnctw00000gn/T/SWAY_LSP_TEMP_DIRQs3zQJ/storage/src/main.sw",
                ),
                start: 345,
                end: 349,
                as_str(): "var1",
            },
        },
        ..
    ]
}
```

test/src/e2e_vm_tests/test_programs/should_fail/generics_in_contract/src/main.sw
```diff
@@ -1,21 +1,21 @@
 contract;
 
-use std::{hash::sha256, storage::get};
+use std::{hash::sha256, storage::storage_api::read};
 
 struct MyStorageMap<K, V> {}
 
-impl<K, V> MyStorageMap<K, V> {
+impl<K, V> StorageKey<MyStorageMap<K, V>> {
```
mitchmindtree, line 7: Cool to see this working!

mitchmindtree (approved):
Nice one @mohammadfawaz, great to see how much this simplifies! Loving these detailed docs with examples too - good motivation to address #3176. I just left a couple Qs/suggestions but no blockers.

sway-lsp/tests/lib.rs
JoshuaBatty: All good, I'm happy to add this functionality back in a follow up PR, no need to hold this up.

JoshuaBatty (approved):
Super impressive work, thanks boss :)

PR #4465, opened 2023-04-18
Sort forc-doc links alphabetically, add std-lib docs definitions

### LINKED ISSUE #3995: Alphabetically order links in `forc doc`

sarahschwartz:
## Description

This PR sorts forc doc links in alphabetical order and adds some missing definitions to the std-lib docs.

Close #3995 

## Checklist

- [X] I have linked to any relevant issues.
- [X] I have commented my code, particularly in hard-to-understand areas.
- [X] I have updated the documentation where relevant (API docs, the reference, and the Sway book).
- [X] I have added (or requested a maintainer to add) the necessary `Breaking*` or `New Feature` labels where relevant.
- [X] I have done my best to ensure that my PR adheres to [the Fuel Labs Code Review Standards](https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/external-contributors.md).
- [X] I have requested a review from the relevant team or maintainers.


sway-lib-std/src/math.sw
```diff
@@ -84,13 +87,14 @@ impl Power for u8 {
     }
 }
 
-// Trait for exponential functions
+/// Trait for exponential functions.
 // Should exist for UFP64, UFP128 and their signed versions
```
kayagokalp, line 91: Not sure if this intended but, should this also be a doc comment?

```suggestion
/// Should exist for UFP64, UFP128 and their signed versions
```

kayagokalp (approved):
forc-doc changes looks good to me, left a small q about a doc comment 👍

sdankel (approved):

sway-lib-std/src/math.sw
sdankel: I think we should.

```suggestion
/// This should exist for UFP64, UFP128 and their signed versions.
```

sway-lsp/tests/lib.rs
```diff
@@ -824,7 +824,7 @@ async fn go_to_definition_for_paths() {
         req_uri: &uri,
         req_line: 24,
         req_char: 31,
-        def_line: 5,
+        def_line: 9,
```
sdankel, line 827: 👍

sway-lib-std/src/math.sw
kayagokalp: Hmm, that is a good question. I think the way it is right now might not be entirely suitable for exposing as a doc comment but it doesn't matter that much. I guess we can refactor them later to be more like proper standard library documentation?

If you prefer we can address these stuff later module by module? That could be easier to manage, for example we also have a `Logarithm` trait in the same file which does not have any doc comments on it. If we do a pass over the modules it is easier to catch those stuff. But totally your call, I am good with this either way 🚀

kayagokalp (approved):

sdankel (approved):

forc-plugins/forc-doc/src/render/link.rs
```diff
@@ -23,9 +23,28 @@ pub(crate) struct DocLinks {
...
+        // for the AllDoc page, sort based on the module prefix
+        match self.style {
+            DocStyle::AllDoc(_) => {
+                for (block_title, mut doc_link) in self.links {
+                    doc_link.sort_by(|a, b| {
+                        a.module_info.module_prefixes[1].cmp(&b.module_info.module_prefixes[1])
+                    });
+                    links_vec.push((block_title, doc_link));
+                }
+            }
```
eureka-cpu, line 37: I tested locally and I think there may a be a problem here. It seems that it only does a shallow alphabetization for the all doc items so you end up with things like `alloc::realloc` coming before `alloc::alloc_bytes`. Not a big deal really, but would you mind making a follow up issue/PR for it?

Also thank you for tackling this!

eureka-cpu: Was there a reason just using `.sort()` didn't work? Just curious

PR #3916, opened 2023-01-28
Add `logging::log` & `assert::assert_eq` to `prelude.sw`

Braqzen:
## Changes

- Added `log` & `assert_eq` to `prelude.sw` and to docs
- Removed `prelude` imports from files
- Changed logging doc to explicitly state that it only logs "arbitrary `stack` types"
- Minor manual formatting here and there
- Updated `LSP` test since it relies on hardcoded values in a file which had an import that no longer exists (and thus the number of lines is different)

## Related

Closes #3914

PR #4483, opened 2023-04-21
Added to msg_sender to prelude and removed from swayfiles

### LINKED ISSUE #4478: Add `msg_sender()` into the prelude
It's used often, worth adding into the prelude.

calldelegation:
## Description
Fixes #4478

## Checklist


docs/reference/src/code/examples/access-control/ownership/src/main.sw
```diff
@@ -1,8 +1,5 @@
 contract;
 
-// ANCHOR: identity
```
kayagokalp, line 3: Should we keep the anchor? Looks like we are still using it

calldelegation:
Missing LSP modifications to support these changes
[Similar PR by Braqzen](https://github.com/FuelLabs/sway/pull/3916/files#diff-54dcd495bcabacd52ae149a0d941f40561fb7e9bdec3878ed66377ec113f7183)

docs/reference/src/code/examples/access-control/ownership/src/main.sw
calldelegation: Added back in! @kayagokalp

Braqzen:
I suggest doing a global search for `auth::` as that will come up with additional imports to remove

calldelegation:
> I suggest doing a global search for `auth::` as that will come up with additional imports to remove

Good call I think I got them all now! 😄

Braqzen (commented):
[Here](https://github.com/FuelLabs/sway/blob/add/msg_sender-to-prelude/test/src/e2e_vm_tests/test_programs/should_fail/supertraits_for_abis_ownable1/src/ownable.sw#L28) is another.

It's fine to clean up files within the related area so if you search for `auth::*` you'll find a bunch of imports too which seem to be entirely unused at this point. I would consider which file they are in and if they still serve some demonstration purpose - if not I would remove them.
One common usage is to still import the `AuthError` and use it as an explicit type hint for the user; however, unless that's used in the book (i.e. just in regular code) it's likely that it's better to omit it.

docs/reference/src/code/operations/call_data/src/lib.sw
Braqzen, line 1: This change is related to the following [doc](https://github.com/FuelLabs/sway/blob/master/docs/reference/src/documentation/operations/call-data/msg-sender.md). Could this doc be updated to remove the comment stating that we must import and then also remove the following code block which imports?

Braqzen (commented):
After the following suggestion we can likely merge

docs/reference/src/code/operations/call_data/src/lib.sw
calldelegation: Also updated the [prelude doc here](https://github.com/FuelLabs/sway/blob/master/docs/reference/src/documentation/misc/prelude.md) to match that of this one [here](https://github.com/FuelLabs/sway/blob/master/docs/reference/src/documentation/misc/prelude.md)

Braqzen (approved):
That's cool. I haven't really checked the reference since it was written. It's probably quite out of date and it needs more work.

kayagokalp (approved):

PR #4492, opened 2023-04-24
Bump to v0.37.2

kayagokalp:
Waiting for:

1. ~https://github.com/FuelLabs/sway/pull/4488~ we can get this later on it is not urgent.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1720,7 +1720,7 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
 name = "forc"
-version = "0.37.1"
+version = "0.37.2"
 dependencies = [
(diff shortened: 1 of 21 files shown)


IGI-111 (approved):

rostyslavtyshko (approved):

mitchmindtree:
This release includes quite a few breaking changes (including Mohammad's storage key refactor) and likely should be rolled out under a new minor release, 0.38.0.

I'm unsure if we have the ability to yank 0.37.2, but if we can we should and re-release as 0.38.0 to avoid mysteriously breaking things for users who d epend on 0.37.1.

mitchmindtree:
It looks like yanking might not be an option, so we might need to:

1. Make a branch from 0.37.2.
2. Create a new patch release from this branch for 0.37.3.
3. Create a new minor release with these changes as 0.38.0.

PR #4493, opened 2023-04-24
Add block_header_hash() function

nfurfaro:
## Description
This exposes the `BHSH` opcode in the stdlib.

## Checklist


test/src/sdk-harness/test_projects/block/src/main.sw
```diff
@@ -19,4 +19,12 @@ impl BlockTest for Contract {
     fn get_block_and_timestamp() -> (u64, u64) {
         (height(), timestamp())
     }
+
+    fn get_block_header_hash(h: u64) -> b256 {
```
Braqzen, line 23: I'd consider whether it would be better to return the Result instead and test for the variant and value

sway-lib-std/src/block.sw
```diff
@@ -21,3 +30,51 @@ pub fn timestamp_of_block(block_height: u64) -> u64 {
...
+        Result::Ok(header_hash)
+    }
+}
+
+////////////////////////////////////////////////////////////////////
+// Tests
+////////////////////////////////////////////////////////////////////
+ 
+#[test(should_revert)]
+fn test_block_header_hash_err_current_height() {
```
Braqzen, line 56: If the condition for `height >= current height` holds then aren't both of these tests doing the same thing despite one passing in a `+1`? I would keep the first test

nfurfaro: Well, having 2 tests could prevent a regression in the VM; the first tests the `=` condition, and the second tests the `>` condition.

Braqzen (approved):

matt-user (approved):
lgtm

PR #4494, opened 2023-04-25
Re-publish 0.37.2 as 0.38.0 due to breaking changes

mitchmindtree:
See here for context: https://github.com/FuelLabs/sway/pull/4492#issuecomment-1520949069

This follows a re-release of 0.37.1 as 0.37.3: https://github.com/FuelLabs/sway/releases/tag/v0.37.3

The outcome of all this is that ideally, users previously on 0.37.1 will now pull 0.37.3 next time they run `cargo update` and will not encounter any breakage. Those ready to update can point to 0.38.0 after this is published.

Also runs `cargo update` in order to update deps to their latest patch release.


bingcicle (approved):
LGTM, great catch on the breaking changes!

JoshuaBatty (approved):

PR #4496, opened 2023-04-25
ci: extract out verify_tags step from publish job into its own job

### LINKED ISSUE #4495: extract out verifying tags step to receive faster feedback
We have a step to verify tags in the workflow, but it's hidden behind the `publish` job which needs all other jobs to run first before attempting to publish:

https://github.com/FuelLabs/sway/blob/384646c9914ed8efcc9b30851301f2d6f95b72ba/.github/workflows/ci.yml#L391-L420

Under normal circumstances, this wouldn't really matter, but sometimes, something minor goes wrong and this could mean about 20-30 minutes of delay prior to getting feedback if the publish step would succeed. Example: https://github.com/FuelLabs/sway/actions/runs/4792365151/jobs/8524237487#step:4:142

We could instead extract out the verify tags step into a separate job that doesn't have the above `[needs]` and run it first so we can fail fast if something went wrong.

bingcicle:
Closes #4495 

## Description

Further explanation is on #4495, but the gist of it is that sometimes the verify tags step may fail but we do not get this feedback until all other jobs are already run since the check has all the other jobs as a pre-requisite.

Extracting this step out would give us immediate feedback instead of requiring us to wait 20-30 minutes. We may also add more pre-publish checks here.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -388,26 +388,11 @@ jobs:
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_NOTIFY_BUILD }}
 
-  publish:
-    # Only do this job if publishing a release
-    needs:
-      [
-        build-forc-test-project,
-        build-sway-examples,
-        build-sway-lib-core,
-        build-sway-lib-std,
-        cargo-build-workspace,
-        cargo-clippy,
-        cargo-fmt-check,
-        cargo-run-e2e-test,
-        cargo-run-e2e-test-evm,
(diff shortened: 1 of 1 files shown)


JoshuaBatty (approved):

mitchmindtree (approved):
Lgtm!

I might do a follow-up PR to consolidate the `needs` check under a `verifications-complete` job similar to how we do for [the fuel-vm CI](https://github.com/FuelLabs/fuel-vm/blob/d04931842420adbeddc7d0bbab5e8cf8c80441c2/.github/workflows/ci.yml#L99-L105).

PR #4498, opened 2023-04-25
Fixes `star_import_with_reexports`.

### LINKED ISSUE #2780: Add a way to disginguish between "crate-relative" and external paths
Currently, `::foo::bar` means crate-relative. There should be a way, like in Rust, to refer to `foo::bar` where `foo` is the external crate. Though this should probably be part of a complete overhaul of Sway's module system.

### LINKED ISSUE #3946: Issue with trait constraints related to the `Eq` trait used in `assert_eq`
The following script should revert but doesn't:
```rust
script;

fn main() {
    assert_eq(1, 2);
}
```
If we investigate the initial IR (need to comment out all optimization passes for that), we can see that the implementation of `eq` "discovered" by the trait constraint solver is:
```rust
    pub fn eq_3(self !24: u64, other !25: u64) -> bool, !26 {
        entry(self: u64, other: u64):
        v0 = const bool undef
        ret bool v0
    }
```
which does nothing. This explains why the `assert_eq` doesn't fail above. 

I wonder if any of the work that was recently done helps with this. cc @esdrubal @emilyaherbert

esdrubal:
## Description

`star_import_with_reexports` was not able to handle external paths.

This was causing an issue #3946 where `get_items_for_type_and_trait_name` was called with trait name `std::core::ops::Eq` instead of `core::ops::Eq`.

With these changes we bring back `assert_eq` and fix `star_import_with_reexports` to handle properly external paths so it no longer adds the src prefix to `use_synonyms`.

Closes #3946
Closes #2780

## Checklist


IGI-111 (approved):
LGTM

PR #4522, opened 2023-05-02
Parse `__ptr[T]` and `__slice[T]`

AlicanC:
Just parsing for typed pointers and slices. Went with long and distinct names for now but we could change them to something like `&T` and `&[T]` in the future.

The only tricky part was this, where I had to fix how we parse `impl` items: https://github.com/FuelLabs/sway/pull/4522/files#diff-51ff5479fa74960c097173172a99b15be1dbd4d858960e619eca1ae740816552


sway-core/src/ir_generation/convert.rs
```diff
@@ -145,6 +145,9 @@ fn convert_resolved_type(
         }
         TypeInfo::RawUntypedPtr => Type::get_uint64(context),
         TypeInfo::RawUntypedSlice => Type::get_slice(context),
+        // TODO: Are these ok?
```
IGI-111, line 148: You'll want to remove this or clarify what's left to do here.

AlicanC: Removed.

IGI-111 (approved):

PR #4532, opened 2023-05-03
Add links to hover documentation

sdankel:
## Checklist


sway-core/src/language/call_path.rs
```diff
@@ -100,7 +100,7 @@ impl CallPath {
     ///
     /// Paths to _external_ libraries such `std::lib1::lib2::my_obj` are considered full already
     /// and are left unchanged since `std` is a root of the package `std`.
-    pub fn to_fullpath(&self, namespace: &mut Namespace) -> CallPath {
+    pub fn to_fullpath(&self, namespace: &Module) -> CallPath {
```
IGI-111, line 103: Why is this no longer mutable?
Not saying it's wrong but I'd like to understand the reasoning for the change.

sdankel: It didn't need to be mutable in the first place :)

sway-core/src/semantic_analysis/namespace/items.rs
```diff
@@ -167,6 +170,27 @@ impl Items {
...
 
+    pub fn get_impl_spans_for_decl(&self, engines: Engines<'_>, ty_decl: &TyDecl) -> Vec<Span> {
+        ty_decl
+            .return_type(engines)
+            .value
+            .map(|type_id| {
+                self.implemented_traits
+                    .get_impl_spans_for_type(engines, &type_id)
+            })
+            .unwrap_or_default()
```
bingcicle, line 181: Q: Don't have the full context here, but would `unwrap_or_default()` here lead to propagating unwanted or incorrect information?

forc-pkg/src/pkg.rs
```diff
@@ -2579,7 +2579,7 @@ pub fn check(
                 )
                 .unwrap(),
             );
-            lib_namespace_map.insert(node, namespace);
+            lib_namespace_map.insert(node, namespace.module().clone());
```
bingcicle, line 2582: Perhaps `lib_namespace_map` should be renamed here?

bingcicle (commented):
Nice! A few questions:

sway-lsp/src/core/session.rs
```diff
@@ -330,6 +330,13 @@ impl Session {
         None
     }
 
+    /// Returns the [Namespace] from the compiled program if it exists.
+    pub fn namespace(&self) -> Option<Namespace> {
```
JoshuaBatty, line 334: Is it possible to return a reference to `Namespace` instead? Cloning the namespace is pretty heavy, would be nice to avoid if possible.

sway-core/src/semantic_analysis/namespace/items.rs
kayagokalp: Hmm, to add on top of Bing's comment, maybe we can consider returning `impl Iterator<Item= Span>` from these type of functions so that their consumer have more freedom to collect them or not. I guess that would also eliminate the need of using `unwrap_or_default()` as well (since an empty iterator would be returned in case of `value` being `None`).

But that being said while reading inner parts of the compiler, I tend to come across returning `Vec` as the default behaviour for these type of functions so it might not make sense to break cohesion just for these new ones. Maybe, later on, we can consider exploring a compiler api clean-up?

sway-core/src/semantic_analysis/namespace/trait_map.rs
```diff
@@ -71,10 +71,17 @@ impl OrdWithEngines for TraitKey {
 /// Map of name to [TyImplItem](ty::TyImplItem)
 type TraitItems = im::HashMap<String, TyImplItem>;
 
+#[derive(Clone, Debug)]
+struct TraitValue {
+    trait_items: TraitItems,
+    // The span of the entire impl block.
```
kayagokalp, line 77:
```suggestion
    /// The span of the entire impl block.
```

kayagokalp (approved):
lgtm! left couple of notes, nothing to block 🚢

sway-core/src/semantic_analysis/namespace/trait_map.rs
JoshuaBatty: We already have a few functions in the server that return `impl Iterator<Item= Span>` so I think it's a good idea to keep this practice going.

JoshuaBatty: For example our `tokens_for_file` function...

```rust
pub fn tokens_for_file<'s>(
    &'s self,
    uri: &'s Url,
    self.iter().flat_map(|(ident, token)| {
        ident.span().path().and_then(|path| {
            if path.to_str() == Some(uri.path()) {
                Some((ident.clone(), token.clone()))
            } else {
                None
            }
        })
    })
}
```

sway-core/src/semantic_analysis/namespace/items.rs
sdankel: Good call outs!

`unwrap_or_default` is fine here because we want it to return an empty array when there are no results. 

I think it's more important to use iterators when dealing with larger collections. Here, even in the std lib, there are typically less than 5 results.

forc-pkg/src/pkg.rs
sdankel: I could, but I see "namespace" used as the variable name for `namespace::Module` throughout the compiler, so I think it's fine to keep. I'd rather not change this code more than necessary.

sway-lsp/src/core/session.rs
sdankel: Good catch!

sway-core/src/semantic_analysis/namespace/trait_map.rs
sdankel: I agree in principle, but I'm not sure if this is necessary when dealing with such small collections of data. Thoughts?

JoshuaBatty: I agree it's probably not necessary but it would be nice to go this direction from a unified API perspective. Just an idea I don't have any stronger opinions on this other than that.

sway-lsp/src/core/session.rs
JoshuaBatty: Would something like this work?

```rust
pub fn namespace<'a>(&'a self) -> Option<&'a Namespace> {
    let compiled_program = &*self.compiled_program.read();
    let program = compiled_program.typed.as_ref()?;
    Some(&program.root.namespace)
}
```

sdankel: With that I get this error: `cannot return value referencing temporary value`

I've tried a few different things but it seems like the function has to return the value for it to exist outside of the function.

sway-core/src/semantic_analysis/namespace/trait_map.rs
sdankel: I actually need the results as a vector anyway, so making this function return an iterator would add some overhead (converting to iter and then back to vec). In the compiler, the existing functions that are similar to this are returning vector, so I think it makes more sense to stick with that in this case.

JoshuaBatty (approved):
Nice work

kayagokalp (approved):

PR #4543, opened 2023-05-05
Fix `StorageVec` bugs with use of `field_id`

bitzoic:
## Description

Only some of the read/writes in `StorageVec` were updated to use `field_id` in https://github.com/FuelLabs/sway/pull/4537. For example, there are two writes when inserting, the length and value itself. Only the value was updated meaning the lengths would still be overwriting one another.
 
Tests to ensure both values and lengths are updated separately have been added.

## Checklist


IGI-111 (approved):
I left those thinking StorageVec wouldn't have an issue just storing the length but I failed to take into account that it's also defined as zero size.

Braqzen (approved):

PR #4546, opened 2023-05-06
`forc doc`: Generate dependency documentation 

### LINKED ISSUE #4533: `forc doc`: Add support for dependencies
> Is there a way to get the dependencies from parsing a Sway program? I’m trying to get `core` from `std` so that I can create links for `core` implementations in `std` eg `impl Eq for MyType` where you’d be able to click on `Eq` and it takes you to `core::ops::Eq`

### LINKED ISSUE #4544: `forc doc`: Check for and remove `doc` folder when generating new docs
I noticed that sometimes extra files will remain in the `out/doc` folder between versions of `forc doc` and that `cargo doc` will actually remove the folder entirely before generating the new docs.

### LINKED ISSUE #4545: `forc doc`: Change `assets` folder to `static.files`
`cargo doc` used to include all of its logos and CSS in the root of the `doc` directory, but it recently changed to including them in a folder titled `static.files`. `forc doc` currently has it under `assets` which could potentially conflict with documentation produced by `forc doc` should the user have a Sway project called `assets`. The `cargo` way seems a bit better, as it's unlikely a user would use this specific name for a project and we can add error handling to deter its use.

eureka-cpu:
## Description

Generates the documentation for dependencies & small improvements to the API. 

What's new so far:
- [x] Programs get their own folders, just like `cargo doc`
- [x] Generates documentation for dependencies by default: Closes #4533 
- [x] Fixes possible conflict with `assets` folder: Closes #4545 
- [x] Removes old docs before generating new ones: Closes #4544 

## Checklist


eureka-cpu:
Not sure why the CI build for `std` can't find the canonical path for `core`. cc @kayagokalp do you have an idea? I could just add the `--no-deps` flag to quiet it, but then I'm not sure we are getting the most out of that test.

sdankel:
Could you add some examples of how this works to this PR description?

I looked at `Address` since it implements `Eq` from core, but there is no link to core. Is that expected?

image

eureka-cpu:
> Could you add some examples of how this works to this PR description?
> 
> I looked at `Address` since it implements `Eq` from core, but there is no link to core. Is that expected?
> 
> image
> 
> I also looked at functions in the `low_level_call` module that use `Bytes` from core, and there are no links to core.
> 
> image

Hey @sdankel yes, this is expected. This PR simply allows the dependencies to be built at all, but it does not yet implement the logic to link them together.

If you take a look at the file structure in `out/doc` you should now notice 3 changes when building. 

1. Each program has its own folder. So the `index.html` used to look something like: `out/doc/index.html`, but is now `out/doc/program_name/index.html`.
2. `assets` is now `static.files`. This removes the probability of potential conflicts in naming conventions for programs and the assets directory.
3. The `doc` directory is cleared before rebuilding the new docs. This prevents strange artifacts from appearing from older versions of `forc doc`.

eureka-cpu:
> I tracked down the issue, if you print the path in the `build_deps` function you can see that before canonicalizing it is `../sway-lib-core` but the relative path should be taken into account from sway-lib-std's folder.
> 
> CI is running the command like `forc-doc --manifest-path ../sway-lib-std` in the root folder which tries to canonicalize `/sway/../sway-lib-core` which does not exist. You need to join the path of the dependency with the package's path (the package that declares the dependency).
> 
> You can verify this is the case by going into the sway-lib-std folder and running `forc-doc` directly there. It is working as `/sway/sway-lib-std/../sway-lib-core` is a valid path.
> 
> That being said, I see that `forc-doc` has a custom logic for building which requires bunch of common code with forc-pkg. I guess the reason for not using `forc-pkg` was the fact that forc-doc requires type and decl engines. I would suggest adding a nice function to forc-pkg, which uses all the established logic like`BuildPlan` like our `build_with_opts` end point, but at the same time accepts the engines from caller. This way you would loose duplicated building steps, automatically support dependencies declared by git and ipfs as well, also this path issues will be gone as `forc-pkg` can give you an iterator of correct paths for each member in the `BuildPlan` and finally you would get the recursive dependency logic for free.

Nice, I will take a look 🙂 Thanks @kayagokalp

kayagokalp:
> Odd that it is working as intended on my machine though!

Are you trying in the sway repo's root folder? It doesn't work in my system if I run `forc-doc --manifest-path ./sway-lib-std`, but if i go into sway-lib-std and run `forc doc`, it works.

eureka-cpu:
> > Odd that it is working as intended on my machine though!
> 
> Are you trying in the sway repo's root folder? It doesn't work in my system if I run `forc-doc --manifest-path ./sway-lib-std`, but if i go into sway-lib-std and run `forc doc`, it works.

Ah I see now, that is what I was doing. Thanks again @kayagokalp 🙂

sdankel (dismissed):

forc-plugins/forc-doc/src/doc/mod.rs
```diff
@@ -6,107 +6,59 @@ use anyhow::Result;
...
+#[derive(Default)]
+pub(crate) struct Documentation(pub(crate) Vec<Document>);
+impl Documentation {
     /// Gather [Documentation] from the [TyProgram].
     pub(crate) fn from_ty_program(
         decl_engine: &DeclEngine,
         project_name: &str,
         typed_program: &TyProgram,
-        no_deps: bool,
+        _no_deps: bool,
```
kayagokalp, line 24: is this intentional? Should we remove the param all together?

forc-plugins/forc-doc/src/main.rs
```diff
@@ -121,61 +198,73 @@ pub fn main() -> Result<()> {
...
+                        &dep_manifest,
+                        dep_pkg_manifest,
+                        doc_path,
+                        build_instructions,
+                    )?;
+                } else {
+                    println!("a path variable was not set for {dep_name}, which is currently the only supported option.")
+                }
+            } else {
+                println!("{dep_name} is a simple format dependency,\nsimple format dependencies don't specify a path to a manfiest file and are unsupported at this time.")
```
kayagokalp, line 249: Should we use bail instead? So this function returns an error, if simple format is used.

forc-plugins/forc-doc/src/render/sidebar.rs
```diff
@@ -59,9 +62,12 @@ impl Renderable for Sidebar {
...
         };
-        let root_path = self
-            .module_info
-            .to_html_shorthand_path_string(INDEX_FILENAME);
+        let root_path = self.module_info.to_html_shorthand_path_string(
+            PathBuf::from(self.module_info.project_name())
+                .join(INDEX_FILENAME)
+                .to_str()
+                .unwrap(),
+        );
```
kayagokalp, line 70: Nit: is it possible to use `ok_or_else` + `?` rather than `unwrap()`? If somehow `project_name()` contains non-valid unicode (not sure if that is checked anywhere, if so feel free to disregard this comment), it would be great to return an error rather than `panic`.

If the check is done while building, we can use `.expect("Invalid project name: containing ")` to panic with some context, if you would like to

kayagokalp (commented):
Looks nice! left couple of nits. I opened #4573, after that we can remove all the custom build effort for `forc-doc`, and use the usual route of `forc-pkg` 👌

forc-plugins/forc-doc/src/main.rs
eureka-cpu: Since it's for dependencies, I thought it would make more sense to just let the user know but still build the main docs. What do you think?

sdankel (approved):

forc-plugins/forc-doc/src/main.rs
kayagokalp: That sounds fine to me!

kayagokalp (approved):

PR #4548, opened 2023-05-08
feat(fmt): improve newline insertion when comments are in between leaf spans

### LINKED ISSUE #4185: Weird formatting with commented out functions and docs
When commenting out a function and there is a doc above the function, we see the following weird formatting behavior. 

This is our code before formatting:

```rust
    /// This is documentation for a commented out function
    // fn commented_out_function() {
    //}

    fn test_function() -> bool {
        true
    }
```

After formatting we have the following:

```rust
    /// This is documentation for a commented out function

        // fn commented_out_function() {
    //}
fn test_function() -> bool {
        true
    }
```

Tested with v0.35.3

bingcicle:
Closes #4185

## Description

The solution we adopted as a start would have us take the space between leaf spans, except a problem was that a doc comment would be a leaf span but a regular comment would not be a leaf span. Usually this would not be an issue, but in certain cases where we perhaps want to comment code out temporarily to test changes, the formatter would move comments around unintentionally.

For example, a doc comment followed by a commented out function (as shown in the issue) would cause the commented out function to move away from the doc comment:

```rust
contract;

abi MyContract {
    fn test_function() -> bool;
}

    /// This is documentation for a commented out function <-- a newline is inserted after this comment
    // fn commented_out_function() {
    //}

    fn test_function() -> bool {
        true
    }
}
```

Situations like these could ruin developer experience.

The main problem is that the entire whitespace gap between 2 leaf spans could have multiple comments with multiple newline sequences, and all these would be squished to the end of the first leaf span when inserting the newlines.

The fix is to take 'snippets' while we're scanning for newline sequences to insert and try to detect comments within these snippets. We then split insertion of newline sequences into two phases: one before the first comment (which is described above, which caused the issue above), one after the last comment.

## Checklist


swayfmt/src/utils/map/newline.rs
```diff
@@ -149,18 +143,65 @@ fn add_newlines(
...
+        if previous_unformatted_newline_span.end < unformatted_newline_span.start {
+            let snippet = &unformatted_code
+                [previous_unformatted_newline_span.end..unformatted_newline_span.start];
+
+            let start = previous_unformatted_newline_span.end;
+
+            // Here, we will try to insert newlines that occur before and after comments.
+            // The reason we do this is because comments aren't a part of the AST, and so they aren't
+            // collected as leaf spans - they simply exist between whitespaces.
+            if let Some(start_first_comment) = snippet.find("//") {
```
kayagokalp, line 155: Hmm does this check with `//` breaks anything related to the comments starting with `/*`?

kayagokalp (commented):
Overall lgtm, left a single question just to make sure 👍

swayfmt/src/utils/map/newline.rs
```diff
@@ -149,18 +143,65 @@ fn add_newlines(
...
+                    },
+                    &newline_map,
+                ) {
+                    let at = previous_formatted_newline_span.end + offset;
+                    offset +=
+                        insert_after_span(at, newline_sequence, formatted_code, newline_threshold)?;
+                }
+
+                // Insert newlines that occur after the last comment here
+                if let Some(start_last_comment) = snippet.rfind("//") {
```
sdankel, line 170: Do we need to handle `///` as well here? Otherwise the third `/` would be considered the start of the comment.

bingcicle: `///` are considered a valid leaf span in the AST, so it won't be picked up here since we're searching between leaf spans.

swayfmt/src/utils/map/newline.rs
bingcicle: Good catch!

I added some naive checks for now that handles catching this, but in future this should probably be extended as the formatter matures to handle other comment formats as well.

kayagokalp (approved):
Thanks for the fix, LGTM :)

sdankel (approved):

Issue #970, opened 2022-03-19
Implement trait constraints


mohammadfawaz:
Trait constraints using `where` seem to be parsed but not yet handled in later stages: https://github.com/FuelLabs/sway/blob/dce836d9e78841dc35a184c918e15e8161cd1882/sway-core/src/semantic_analysis/ast_node/mod.rs#L537

The `where` keyword can be applied in many contexts. See `trait_bounds` and its uses in https://github.com/FuelLabs/sway/blob/dce836d9e78841dc35a184c918e15e8161cd1882/sway-core/src/sway.pest#L211

- [x] #1163
- [x] #1159
- [x] #1162
- [x] #4560

adlerjohn:
Can you clarify what specifically is missing?

emilyaherbert:
Blocked by #1821.

emilyaherbert:
*copying this from slack for record keeping*

My best recommendation for solving this is this:

imagine that we have this example:

```rust
trait Double<T> {
  fn double(self) -> T;
}

trait Triple<T> {
  fn triple(self) -> T;
}

impl Double<u64> for u64 {
  fn double(self) -> u64 {
    self + self
  }
}

impl Triple<u64> for u64 {
  fn triple(self) -> u64 {
    self + self + self
  }
}

struct Foo<T> {
  x: T
}

impl<T> Foo<T> where T: Double + Triple {
  fn math(self) -> Foo<T> {
    let a = self.x.double();
    let b = self.x.triple();
    Foo {
      x: a + b
    }
  }
}
```

on the inside of the compiler we (in pseudocode) change this to:

```rust
impl<T> Foo<T> {
  //                       ↓↓↓↓↓↓↓
  fn math(self) -> Foo<T> where T: Double + Triple {
    let a = self.x.double();
    let b = self.x.triple();
    Foo {
      x: a + b
    }
  }
}
```

During type checking of the `impl<T> Foo<T>` block, we assume that `T` does implement `Double` and `Triple` and we proceed as we normally would for nested generic functions.

Then, during monomorphization of `Foo` for some specific `T`, we check to see if `T` implements `Double` and `Triple`. If it does, then we proceed as normal.

The benefits of this are:
1. low overhead in terms of developer time
2. we can create better error message. "X does not implement Double" is a better error message than "X does not have this method"
3. We can eliminate the use of dummy functions
4. the same logic could be used for enums, structs, and functions

[...]

PR #4557, opened 2023-05-10
Implements where clauses for Structs and Enums.

### LINKED ISSUE #1159: Handle `where` clause for struct.

esdrubal:
This commit enables where clauses for Structs and Enums and also adds two kinds of error handling.

One for when a type does not implement a trait and is used in a data structures that requires that trait.

And another for when a generic type parameter should have the trait constraint.

Lets use the example below for demonstrating the two kinds of error handling.
```rust
enum MyEnum<T> where T: MyAdd {
  X: T,
}
```
The enum above has a constraint where `T` should implement the trait `MyAdd`.

When the type `MyEnum<u64>` is used and if `u64` does not implement the trait `MyAdd` then the error `CompileError::TraitConstraintNotSatisfied` is thrown.

The second type of error handling performed results in an error for the example below.
```rust
fn add2<G>(e: MyEnum<G>) -> G {
}
```
When `MyEnum<G>` is used and the type parameter `G` does not have the trait constraint `where G: MyAdd` then the error `CompileError::TraitConstraintMissing` is thrown.

Fixes #1159

## Checklist

### DIFF
--- a/sway-core/src/type_system/id.rs
+++ b/sway-core/src/type_system/id.rs
@@ -400,19 +400,16 @@ impl TypeId {
                     }
                 }
             } else {
-                // TODO(Esdrubal): Use callpath directly instead of suffix,
-                // this will require TraitConstraint.trait_name to have an absolute CallPath
                 let generic_trait_constraints_trait_names = ctx
                     .namespace
                     .implemented_traits
-                    .get_trait_names_for_type(engines, *structure_type_id)
-                    .iter()
-                    .map(|c| c.suffix.clone())
-                    .collect::<Vec<_>>();
+                    .get_trait_names_for_type(engines, *structure_type_id);
                 for structure_trait_constraint in structure_trait_constraints {
-                    if !generic_trait_constraints_trait_names
-                        .contains(&structure_trait_constraint.trait_name.suffix)
-                    {
+                    if !generic_trait_constraints_trait_names.contains(
+                        &structure_trait_constraint
+                            .trait_name
+                            .to_fullpath(ctx.namespace),
+                    ) {
                         errors.push(CompileError::TraitConstraintNotSatisfied {
                             ty: structure_type_info_with_engines.to_string(),
                             trait_name: structure_trait_constraint.trait_name.suffix.to_string(),


sway-core/src/semantic_analysis/ast_node/declaration/enum.rs
```diff
@@ -27,7 +27,7 @@ impl ty::TyEnumDecl {
         // Type check the type parameters. This will also insert them into the
         // current namespace.
         let new_type_parameters = check!(
-            TypeParameter::type_check_type_params(ctx.by_ref(), type_parameters, true),
+            TypeParameter::type_check_type_params(ctx.by_ref(), type_parameters, false),
```
IGI-111, line 30: With this, are there any invocations that still require `disallow_trait_constraints` to be true?

sway-core/src/type_system/id.rs
```diff
@@ -317,16 +325,108 @@ impl TypeId {
...
+                    {
+                        errors.push(CompileError::TraitConstraintMissing {
+                            param: structure_type_info_with_engines.to_string(),
+                            trait_name: structure_trait_constraint.trait_name.suffix.to_string(),
+                            span: span.clone(),
+                        });
+                    }
+                }
+            } else {
+                // TODO(Esdrubal): Use callpath directly instead of suffix,
```
IGI-111, line 403: Does  this mean that we would miscompile on similarly named traits at different paths or is it just a performance improvement?

esdrubal: Yes as it was similarly named traits could be used interchangeably to avoid the error. The most recent version uses the trait name with full path so this should not occur.

sway-core/src/semantic_analysis/ast_node/declaration/enum.rs
IGI-111: I added a new issue for this as #4560 as part of #970
Traits are out of scope for this PR but not for implemented trait constraints, please update this PR to reflect that it doesn't close #970

esdrubal: I removed #970 from the list of fixed issues by this PR.

IGI-111 (approved):

PR #4566, opened 2023-05-16
feat: implement `forc-test` filter

### LINKED ISSUE #3268: Add support for "filtering" unit tests using a string
I.e. only run tests whose functions contain the given string (same as cargo).

kayagokalp:
## Description
closes #3268.

This PR adds filtering capability to `forc-test` based on test names. Tests are executed if their name contains the given filter phrase.

## Checklist


forc-test/src/lib.rs
```diff
@@ -396,6 +397,14 @@ impl<'a> PackageTests {
...
                 .par_iter()
                 .filter_map(|entry| entry.kind.test().map(|test| (entry, test)))
+                .filter(|(entry, _)| {
+                    // If a test filter is specified, only the tests containing the filter phrase in
+                    // their name are going to be executed.
+                    match &test_filter {
+                        Some(filter) => entry.finalized.fn_name.contains(filter),
+                        None => true,
+                    }
+                })
```
sdankel, line 407: We should unit test this functionality:
- No filter
- Filter matches nothing
- Filter matches something
- Exact match filter

forc-test/src/lib.rs
```diff
@@ -641,17 +672,23 @@ fn deployment_transaction(
...
+/// Returns the result of test execution.
+fn run_tests(
+    built: BuiltTests,
+    test_runners: &rayon::ThreadPool,
+    test_filter: Option<String>,
+) -> anyhow::Result<Tested> {
     match built {
         BuiltTests::Package(pkg) => {
-            let tested_pkg = pkg.run_tests(test_runners)?;
+            let tested_pkg = pkg.run_tests(test_runners, test_filter.as_deref())?;
```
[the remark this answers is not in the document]
kayagokalp: Nice point! It should not (and also does not as we have our unit test e2es still passing without a filter), you can see that from [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ff0cf2537b83f7d43340d6a02218f34b) playground. 

It is because, in core, `as_deref` is implemented for `Option` like the following:

```rust
pub const fn as_deref(&self) -> Option<&T::Target>
where
    T: ~const Deref,
{
    match self.as_ref() {
        Some(t) => Some(t.deref()),
        None => None,
    }
}
```

[reference](https://doc.rust-lang.org/src/core/option.rs.html#1267-1277)

forc-test/src/lib.rs
kayagokalp: Good idea, added tests for:

1. No filter
2. Exact filter matches nothing
3. Exact filter matching a thing
4. Non exact filter matches nothing
5. Non exact filter fetches a single thing
6. Non exact filter matches everything

Cargo.toml
```diff
@@ -26,7 +26,8 @@ members = [
 ]
 exclude = [
     "examples/*",
-    "swayfmt/test_macros"
+    "swayfmt/test_macros",
+    "forc-test/test_data"
```
sdankel, line 30: Just curious, is it best practice to always exclude all test data?

forc-test/src/lib.rs
```diff
@@ -543,34 +561,66 @@ pub enum TestRunnerCount {
     Auto,
 }
 
+pub struct TestCount {
+    pub total: usize,
+    pub filtered: usize,
```
sdankel, line 566: nit: consider renaming to `ignored` or `filtered_out`. I assumed this was the number of tests remaining after filtering, but it looks like this is the count of tests filtered out.

```suggestion
    pub ignored: usize,
```

forc-test/src/lib.rs
```diff
@@ -543,34 +561,66 @@ pub enum TestRunnerCount {
...
+            if ignored {
+                num_ignored += 1;
+            }
+            num_total += 1;
+        }
+
+        TestCount {
+            total: num_total,
+            filtered: num_ignored,
+        }
```
sdankel, line 608: nit: it'd be clearer to read if we did this whole operation with higher order functions, rather than a combination of for-loops and HOFs. This is a good use case for an accumulator.

```suggestion
        pkgs.iter()
            .flat_map(|pkg| {
                pkg.built_pkg_with_tests()
                    .bytecode
                    .entries
                    .iter()
                    .filter_map(|entry| entry.kind.test().map(|test| (entry, test)))
            })
            .fold(TestCount::default(), |acc, (pkg_entry, _)| {
                let num_ignored = match &test_filter {
                    Some(filter) => {
                        if filter.filter(&pkg_entry.finalized.fn_name) {
                            acc.filtered
                        } else {
                            acc.filtered + 1
                        }
                    }
                    None => acc.filtered,
                };
                TestCount {
                    total: acc.total + 1,
                    filtered: num_ignored,
                }
            })
```

forc-test/src/lib.rs
```diff
@@ -760,3 +816,127 @@ fn exec_test(
...
+
+    #[test]
+    fn test_no_filter() {
+        let test_filter = None;
+        let test_results = test_library_test_results(test_filter).unwrap();
+        let tested_package_test_count = test_results.len();
+
+        assert_eq!(tested_package_test_count, 2)
+    }
+}
```
sdankel, line 942: 🏅

forc-test/src/lib.rs
```diff
@@ -760,3 +816,127 @@ fn exec_test(
...
+
+        let test_results = test_library_test_results(Some(test_filter)).unwrap();
+        let tested_package_test_count = test_results.len();
+
+        assert_eq!(tested_package_test_count, 1)
+    }
+
+    #[test]
+    fn test_filter_all_match() {
+        let filter_phrase = "test_b";
```
sdankel, line 922: nit: have one of the tests matchig in the middle of the string rather than matching the prefix.

```suggestion
        let filter_phrase = "est_b";
```

forc/src/cli/commands/test.rs
```diff
@@ -49,21 +52,27 @@ pub struct TestPrintOpts {
...
     let built_tests = forc_test::build(opts)?;
     let start = std::time::Instant::now();
-    info!("   Running {} tests", built_tests.test_count());
-    let tested = built_tests.run(test_runner_count)?;
+    let test_count = built_tests.test_count(test_filter.as_ref());
+    info!(
+        "   Running {} tests, filtered {} tests",
+        test_count.total - test_count.filtered,
+        test_count.filtered
+    );
```
sdankel, line 74: nit: for nicer UX, print `1 test` instead of `1 tests`

```
fn formatted_test_count_string(count: usize) -> String {
    format!("{} {}", count, if count == 1 { "test" } else { "tests" })
}
```

```suggestion
    info!(
        "   Running {}, ignored {}",
        formatted_test_count_string(test_count.total - test_count.filtered),
        formatted_test_count_string(test_count.filtered),
    );
```

sdankel (commented):
Looking good, mostly nits 👍

forc-test/src/lib.rs
kayagokalp: Ahh yeah missed this one, I actually changed the local vars name but forgot to apply it to the struct.

Cargo.toml
kayagokalp: There is a long discussion about cargo's default behaviour once it comes to non-package files and publishing [here](https://github.com/rust-lang/cargo/issues/2063). So I tend to exclude big artifacts like a built package ~mbs from the publishing step. 

There are some downsides to this as well, since we strip the test data from the published package, tools like [crater](https://github.com/rust-lang/crater) maintained by Rust core team cannot use our packages to discover regression in the rust compiler.

kayagokalp:
Thanks for the review @sdankel, should be good to go!

sdankel (approved):
🚢

forc-test/test_data/test_library/src/lib.sw
```diff
@@ -0,0 +1,11 @@
+library;
+
+#[test]
+fn test_bam() {
+  assert(1 == 1)
+}
+
+#[test]
+fn test_bum() {
```
mitchmindtree, line 9: :peach:

forc/src/cli/commands/test.rs
```diff
@@ -32,6 +32,9 @@ pub struct Command {
     /// When specified, only tests containing the given string will be executed.
     pub filter: Option<String>,
     #[clap(long)]
+    /// When specified, only the test exactly matching the given string will be executed.
+    pub exact: bool,
```
mitchmindtree, line 36: Maybe this should be `filter_exact` to make its association with the filter a little more clear?

mitchmindtree (approved):
One tiny optional change but otherwise utACK

forc/src/cli/commands/test.rs
kayagokalp: Good idea!

mitchmindtree (approved):

JoshuaBatty (approved):

PR #4578, opened 2023-05-22
ci: use `--locked` while installing `cargo-generate`

### LINKED ISSUE #4577: ci: `cargo-generate` failing to install blocks CI
https://github.com/cargo-generate/cargo-generate/issues/942, looks like we need to update our cargo-generate install step, to use `--locked`, so that it builds successfully. Once the upstream fixes the issue we can revert it back.

kayagokalp:
## Description
closes #4577.

Because of an upstream issue in `cargo-generate`, which can be seen from https://github.com/cargo-generate/cargo-generate/issues/942, our CI is blocked. Until the fix lands we need the `--locked` flag added to install step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,7 +158,7 @@ jobs:
       # - name: Run test project's test suite
       #   run: (cd test-proj && forc test)
       - name: Install cargo-generate
-        run: cargo install cargo-generate
+        run: cargo install --locked cargo-generate
       - name: Generate Rust Integration Tests
         run: (cd test-proj && cargo generate --init --path ../templates/sway-test-rs --name test-proj)
       - name: Update project cargo manifest with workspace


Braqzen (approved):
Coolio, afaik this will throw an error to say that `cargo update` needs to be run because upstream dependencies have changed

mitchmindtree:
Good catch!

I'd recommend keeping the `--locked` flag going forward - all it does is make sure that the existing `Cargo.lock` file in our workspace is valid, which helps to make sure that we can build entirely from existing, pinned dependencies in a deterministic manner.

Without `--locked`, cargo will sometimes automatically update dependencies if e.g. a new dependency was added, but the lock file was not updated. This is a problem as it might end up pulling in a different version each time it gets run, leading to non-determinism in our CI checks.

I wonder if there's a way we can specify some global cargo configuration so that it uses `--locked` by default in our CI so that we don't have to remember to add `--locked` to just about every cargo invocation :thinking:

kayagokalp:
> Good catch!
> 
> I'd recommend keeping the `--locked` flag going forward - all it does is make sure that the existing `Cargo.lock` file in our workspace is valid, which helps to make sure that we can build entirely from existing, pinned dependencies in a deterministic manner.
> 
> Without `--locked`, cargo will sometimes automatically update dependencies if e.g. a new dependency was added, but the lock file was not updated. This is a problem as it might end up pulling in a different version each time it gets run, leading to non-determinism in our CI checks.
> 
> I wonder if there's a way we can specify some global cargo configuration so that it uses `--locked` by default in our CI so that we don't have to remember to add `--locked` to just about every cargo invocation 🤔

That would be superb, quick search yielded following issues: 

1. https://github.com/rust-lang/cargo/issues/8207
2. https://github.com/rust-lang/cargo/issues/8207

Seems like we don't have a way to do that, but will keep an eye out for a solution we can use 👍

PR #4579, opened 2023-05-22
Adds fully qualified paths for methods.

### LINKED ISSUE #4025: Not possible to use associate function call of trait impl method.
@nfurfaro found an issue while doing https://github.com/FuelLabs/sway/pull/3882 that can be reproduced by the following sway script:

```rust
script;

trait From2<T> {
    fn into(self) -> T;
}

struct S1 {
    s1: u64
}

struct S2 {
    s2: u64
}

impl From2<S1> for u64 {
    fn into(self) -> S1 {
        S1{s1: self}
    }
}

impl From2<S2> for u64 {
    fn into(self) -> S2 {
        S2{s2: self}
    }
}

fn main() {
    let s1: S1  = 42.into();

    // throws Mismatched types.
    let s2: S2  = 42.into();
    
    // Associate function call not supported in sway.
    // throws Unknown type name "From2"
    let s2: S2  = From2::<S2>::into(42);
}
```

The following rust [playgroud](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2efe4d98c686bde483e73ef714cf2312) shows how this works in rust. The error `multiple applicable items in scope` is thrown when calling `into` directly but it is still possible to use the associate function call.

This issue should be addressed by:
  1. ensure that `into` methods can be called by associate function call.
  2. throw a proper error like `multiple applicable items in scope` when `into` is called.

### LINKED ISSUE #4383: Ambiguous trait method should not be allowed
The following piece of code compiles but should not, due to an ambiguous method call at the end of the example:

```sway
contract;

struct S {}

trait MySuperTrait {
    fn method();
}

trait MyTrait : MySuperTrait {
    fn method();
}

impl MySuperTrait for S {
    fn method() { }
}

impl MyTrait for S {
    fn method() { }
}

fn main() {
    S::method()            // ambiguous method call here
}
```

esdrubal:
## Description

With the changes in this PR it is now possible to qualify a trait to disambiguate the method to be used, such as `<S as Trait>::method()`.

Fixes #4383
Fixes #4025

## Checklist


sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
```diff
@@ -2699,15 +2722,19 @@ fn path_root_opt_to_bool(
     _context: &mut Context,
     handler: &Handler,
     root_opt: Option<(Option<AngleBrackets<QualifiedPathRoot>>, DoubleColonToken)>,
+    throw_error_on_qualified_path_root: bool,
```
IGI-111, line 2725: I don't like this kind of specialization because it's hard to see what this does at a glance.
Consider using a wrapper function to let the caller explicitly pick if they want the error or not. Or throwing the error all the time but handling it with `unwrap_or` at the call site.

sway-lsp/src/traverse/parsed_tree.rs
```diff
@@ -405,7 +405,9 @@ impl Parse for AmbiguousPathExpression {
         let AmbiguousPathExpression {
             call_path_binding,
             args,
+            qualified_path_root: _,
         } = self;
+        // TODO: process qualified paths
```
IGI-111, line 410: Is this still to do?

esdrubal: Removed the to-do in the most recent version and added processing of qualified path tokens.

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
esdrubal: Removed the parameter `throw_error_on_qualified_path_root` as suggested and the place that was depending on it now uses `path_root_opt_to_bool_and_qualified_path_root`

IGI-111 (approved):

PR #4582, opened 2023-05-23
Add from_struct and to_struct methods for Bytes

dmihal:
## Description

It is common for smart contract developers to need to serialize & unserialize structured data into byte arrays. In solidity, this is provided by `abi.encode` and `abi.decode`. This feature has been requested by multiple ecosystem projects.

This PR adds 2 functions to the Bytes struct: `from_struct` and `to_struct`, which convert between standard structs and byte arrays by copying memory.

It should be noted that this functionality only works for stack structs, not variable-length heap structs like Strings, Vecs and Byte arrays.

## Checklist

### DIFF
--- a/sway-lib-std/src/bytes.sw
+++ b/sway-lib-std/src/bytes.sw
@@ -104,6 +104,33 @@ impl Bytes {
         }
     }
 
+
+    /// Constructs a new, `Bytes` with the contents of a struct.
+    ///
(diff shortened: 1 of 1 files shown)


IGI-111:
This is a nice hack but it's also extremely unsafe and doesn't work for heap types at all.
Nice idea for a library but I don't think it should be in `std` as is.

PR #4588, opened 2023-05-24
Simplify the lifetimes around engines.

tritao:
## Checklist

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs
+++ b/sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs
@@ -24,6 +24,8 @@ impl ty::TyImplTrait {
         let mut errors = vec![];
         let mut warnings = vec![];
 
+        let engines = &ctx.engines.clone();
+
         let ImplTrait {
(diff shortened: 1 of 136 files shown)


IGI-111:
This is probably more maintainable but is it worth the extra cloning?

sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs
```diff
@@ -1033,12 +1054,13 @@ fn check_for_unconstrained_type_parameters(
 
 fn handle_supertraits(
     mut ctx: TypeCheckContext,
+    engines: &Engines,
```
IGI-111, line 1057: This goes for all other instances of it but what's the point of passing an explicit `Engines` argument when it's already part of the `TypeCheckContext`?

IGI-111: Well then sorry about the mixed messages, but I think cloning is probably better here. This entire PR is about introducing some acceptable runtime overhead for clarity, and this makes things more confusing rather than less so.

PR #4603, opened 2023-05-30
feat: enable providing main arguments to a script with `forc-run`

### LINKED ISSUE #3439: `forc run` - allow for specifying `main` arguments for scripts
Currently, using `forc run` there is no way to provide main arguments for scripts.

## CLI arguments

We couild *at least* allow for specifying primitive arguments via the command line, e.g. 

```console
forc run -- arg1 arg2 arg3
```

## `script-args.json` or similar?

We could also settle on a file format (toml or json) that provides both the argument values and their types so that the user can update their file and do something like `forc run --args script-args.json` or similar.

We might need to support some sort of args file like this either way if we want to allow users to submit more sophisticated types via main arguments. Perhaps we can extract the `fuels` representation into `fuels-types` (if it isn't already) so that we can re-use the same serializable main args format directly.

This is not a big priority for users already using the Rust or TS SDKs directly to run scripts, but a feature like this would be especially useful for users with simpler scripts, or whose Sway projects are embedded in other languages.

cc @nfurfaro @simonr0204 let us know if you have any more thoughts / ideas on this!

kayagokalp:
## Description

closes #3439.

This PR adds `--args` flag and encoding capabilities to `forc-run` so that users can provide inputs to main function of the script they are running.

Tested with the following setup:

1. a fuel-core instance (version 0.19) is running in a terminal - `fuel-core run --chain chainConfig.json --db-type in-memory`
2. local wallet is funded through chain config provided to that fuel-core instance.
3. a test script is created with the following `main.sw`:
```sway
script;

fn main(a: u64) -> u64 {
 a
}
```
4. executed `forc-run --node-url http://127.0.0.1:4000 --args 5`, output receipt shows that script returned 5 as expected (see `val`):
```console
fuel/test_projects/test-script took 3s 
❯ ../../../fuel/dev/sway/target/debug/forc-run --node-url http://127.0.0.1:4000 --args 5
 Compiling script test-script (/Users/kayagokalp/fuel/test_projects/test-script)
  Finished debug in 1.998ms
    script test-script
      Bytecode size: 44 bytes
      Bytecode hash: 0x36f3349cc22b483e97a2af133761bc3f64baba27816252a702604a46c1874402

Please provide the password of your encrypted wallet vault at "/Users/kayagokalp/.fuel/wallets/.wallet":
Do you accept to sign this transaction with fuel1x6pc0g9gwp2c2hzhejlw7l4guqd99732zt2wd68j2ukrdu59kv9qqz9h3k? [y/N]: y
[{"Return":{"id":"0000000000000000000000000000000000000000000000000000000000000000","is":10344,"pc":10384,"val":5}},{"ScriptResult":{"gas_used":68,"result":"Success"}}]
```
5. executed `forc-run --node-url http://127.0.0.1:4000 --args 10`, output receipt shows that script returned 10 as expected (see `val`):
``` console
fuel/test_projects/test-script took 5s 
❯ ../../../fuel/dev/sway/target/debug/forc-run --node-url http://127.0.0.1:4000 --args 10
 Compiling script test-script (/Users/kayagokalp/fuel/test_projects/test-script)
  Finished debug in 2.016958ms
    script test-script
      Bytecode size: 44 bytes
      Bytecode hash: 0x36f3349cc22b483e97a2af133761bc3f64baba27816252a702604a46c1874402

Please provide the password of your encrypted wallet vault at "/Users/kayagokalp/.fuel/wallets/.wallet":
Do you accept to sign this transaction with fuel1x6pc0g9gwp2c2hzhejlw7l4guqd99732zt2wd68j2ukrdu59kv9qqz9h3k? [y/N]: y
[{"Return":{"id":"0000000000000000000000000000000000000000000000000000000000000000","is":10344,"pc":10384,"val":10}},{"ScriptResult":{"gas_used":68,"result":"Success"}}]
```
## Checklist


Cargo.toml
```diff
@@ -39,9 +39,9 @@ fuel-crypto = "0.26"
 fuel-tx = "0.26"
 fuel-core-client = "0.17"
 fuel-vm = "0.26"
-fuels-core = "0.36.1"
-fuels-signers = "0.36.1"
-fuels-types = "0.36.1"
+fuels-core = { version = "0.36.1", path = "../fuels-rs/packages/fuels-core/" }
+fuels-signers = { version = "0.36.1", path = "../fuels-rs/packages/fuels-signers/" }
+fuels-types = { version = "0.36.1", path = "../fuels-rs/packages/fuels-types/" }
```
kayagokalp, line 44: This is for testing locally until I expose the FullProgramABI from SDK

forc-plugins/forc-client/Cargo.toml
```diff
@@ -19,11 +19,13 @@ forc-pkg = { version = "0.40.1", path = "../../forc-pkg" }
 forc-tracing = { version = "0.40.1", path = "../../forc-tracing" }
 forc-tx = { version = "0.40.1", path = "../forc-tx" }
 forc-util = { version = "0.40.1", path = "../../forc-util" }
+fuel-abi-types = "0.3"
 fuel-core-client = { workspace = true, default-features = false }
 fuel-crypto = { workspace = true }
 fuel-tx = { workspace = true, features = ["builder"] }
 fuel-vm = { workspace = true }
 fuels-accounts = { workspace = true }
+fuels-code-gen = "0.42"
```
kayagokalp, line 28: I will be removing this dependency as we don't need it anymore with latest release of `fuel-abi-types`

sdankel:
Looking at the original issue, I think the proposed options were either to pass args like this: `forc run -- arg1 arg2 arg3` or have the `--args` flag take a json file that specifies the arguments. I think we could do both, but reserving the `--args` flag for a json file seems like a good idea in case the arguments are complex types. It might make sense to start with the first option for now.

sdankel (commented):
Does this approach allow you to specify multiple arguments?

Looking at the original issue, the proposed options were either to pass args like this: `forc run -- arg1 arg2 arg3` or have the `--args` flag take a json file that specifies the arguments. I think we could do both, but at least reserving the `--args` flag for a json file seems like a good idea in case the arguments are complex types. It might make sense to start with the first option for now.

kayagokalp:
> Does this approach allow you to specify multiple arguments?
> 
> 
> 
> Looking at the original issue, the proposed options were either to pass args like this: `forc run -- arg1 arg2 arg3` or have the `--args` flag take a json file that specifies the arguments. I think we could do both, but at least reserving the `--args` flag for a json file seems like a good idea in case the arguments are complex types. It might make sense to start with the first option for now.

Hey, yes this allows multiple arguments. You can use forc run --arg 10 --arg 40 -- arg true etc. Also i am planning to add option to provide args via a json descriptor just like you suggested in a future PR. Opening an issue for that

sdankel:
> > Does this approach allow you to specify multiple arguments?
> > Looking at the original issue, the proposed options were either to pass args like this: `forc run -- arg1 arg2 arg3` or have the `--args` flag take a json file that specifies the arguments. I think we could do both, but at least reserving the `--args` flag for a json file seems like a good idea in case the arguments are complex types. It might make sense to start with the first option for now.
> 
> Hey, yes this allows multiple arguments. You can use forc run --arg 10 --arg 40 -- arg true etc. Also i am planning to add option to provide args via a json descriptor just like you suggested in a future PR. Opening an issue for that

If `--arg` accepts strings and it will accept a .json file, how will we distinguish between the literal `args.json` and the file contents as args?

kayagokalp:
> 

we can make --args-json or --args accepting a json file and --arg accepting the actual argument literal values.

JoshuaBatty:
I like `--args-json` for providing a json file and `--args` for specifying 1..n number of arguments. I don't think we need to have an `--arg` flag for single values.

JoshuaBatty (approved):
Nice work, utACK

sdankel (approved):
LGTM

PR #4607, opened 2023-05-30
fix: Change order of master deploy

eureka-cpu:
## Description
My previous PR #4606 is overwriting all other folders in `ghpages/master` since it is the final step. This changes it to the first step in that process.

If it makes more sense, we could instead change the `master/std` path to be `master/lib` and this would allow for a link to both `master/lib/core` and `master/lib/std` instead of `master/core` and `master/std`. This would mean that we would also have to look out for places we use the links for `std` in and update them accordingly.
## Checklist


Braqzen (approved):

eureka-cpu:
@Braqzen do you have a preference wrt my initial comment?

Braqzen:
> @Braqzen do you have a preference wrt my initial comment?

I don't see the benefit of adding another layer of nesting via the `//lib/` step as opposed to then starting from `master`. Moreover, `core` is available and some things should be visible but at the same time there's other code in there used internally and so it shouldn't be linked to.
If we need to start changing things then that's additional work, which is fine if there is a benefit to the distinction in the paths.
I'm somewhat indifferent because I don't feel like I have the full context of why this should be done. We could have a chat internally if you'd like to catch me up on the context

sarahschwartz (approved):

PR #4613, opened 2023-05-31
Span optimizations

tritao:
## Description

This implements a set of optimizations around span usage. The main idea is we stop using paths as strings, and instead replace it using an integer-based id, which is way less expensive for hash map comparisons.

Here are some numbers, from the slowest test we have on the suite (`should_pass/stdlib/vec`).

**Benchmarking Results** (`master`):
  **Time** (mean ± σ):      6.129 s ±  0.080 s    [User: 5.755 s, System: 0.359 s]
  **Range** (min … max):    6.009 s …  6.268 s    10 runs

**Benchmarking Results** (span hash optimization):
  **Time** (mean ± σ):      5.345 s ±  0.161 s    [User: 5.000 s, System: 0.341 s]
  **Range** (min … max):    5.082 s …  5.590 s    10 runss

**Benchmarking Results** (both span hash and SourceEngine optimization):
  **Time** (mean ± σ):      4.376 s ±  0.089 s    [User: 4.065 s, System: 0.309 s]
  **Range** (min … max):    4.255 s …  4.546 s    10 runs

## Checklist

### DIFF
--- a/sway-types/src/source_engine.rs
+++ b/sway-types/src/source_engine.rs
@@ -1,0 +1,59 @@
+use std::{collections::HashMap, path::PathBuf, sync::RwLock};
+
+use crate::SourceId;
+
+/// The Source Engine manages a relationship between file paths and their corresponding
+/// integer-based source IDs. Additionally, it maintains a reserve - a map that traces
+/// back from a source ID to its original file path. The primary objective of this
+/// system is to enable clients that need to reference a file path to do so using an
+/// integer-based ID. This numeric representation can be stored more efficiently as
+/// a key in a hashmap.
+/// The Source Engine is designed to be thread-safe. Its internal structures are
+/// secured by the RwLock mechanism. This allows its functions to be invoked using
+/// a straightforward non-mutable reference, ensuring safe concurrent access.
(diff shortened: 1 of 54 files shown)


vaivaswatha (approved):

vaivaswatha:
I checked the sway-ir parts and it looked ok to me. Would prefer that someone else have a keener look at other places.

sway-types/src/source_engine.rs
```diff
@@ -0,0 +1,46 @@
+use std::{collections::HashMap, path::PathBuf, sync::RwLock};
+
+use crate::SourceId;
+
+#[derive(Debug, Default)]
+pub struct SourceEngine {
```
IGI-111, line 6: I think it might be worth it to add some comments that explain what the source engine does and why the internals are RwLocked.

IGI-111:
I think this needs some good doccomments, but otherwise LGTM.

IGI-111 (approved):

JoshuaBatty (approved):
Nice work!

vaivaswatha (approved):

PR #4616, opened 2023-06-01
Replace recursive build strategy with `BuildPlan` compilation order

### LINKED ISSUE #4604: Use `forc_pkg::BuildPlan` dependency graph for building `forc doc` deps
This change will update the current recursive strategy and prevent building the same dep multiple times.

eureka-cpu:
## Description
Closes #4604 

fixes the redundant re-building of dependencies by using the `BuildPlan`s compilation order


forc-plugins/forc-doc/src/main.rs
```diff
@@ -77,8 +80,83 @@ pub fn main() -> Result<()> {
...
+        engines,
+    )?;
+
+    if !build_instructions.no_deps {
+        let order = plan.compilation_order();
+        let graph = plan.graph();
+        let manifest_map = plan.manifest_map();
+        compile_results.reverse();
+
+        for node in order {
```
mitchmindtree, line 118: Rather than reversing and popping from `compile_results` as we iterate, it looks like we should be able to iterate over the order and results together:

```suggestion
        for (node, compile_res) in order.iter().zip(compile_results) {
```

forc-plugins/forc-doc/src/main.rs
```diff
@@ -77,8 +80,83 @@ pub fn main() -> Result<()> {
...
+        &member_manifests,
+        build_instructions.locked,
+        build_instructions.offline,
+    )?;
+
+    let type_engine = TypeEngine::default();
+    let decl_engine = DeclEngine::default();
+    let query_engine = QueryEngine::default();
+    let engines = Engines::new(&type_engine, &decl_engine, &query_engine);
+    let tests_enabled = true;
```
mitchmindtree, line 103: Perhaps we should only include tests if private items are enabled?

mitchmindtree (commented):
A couple minor suggestions, otherwise looking good!

forc-plugins/forc-doc/src/main.rs
```diff
@@ -77,8 +80,78 @@ pub fn main() -> Result<()> {
...
+        let manifest_map = plan.manifest_map();
+
+        for (node, compile_result) in order.iter().zip(compile_results) {
+            let id = &graph[*node].id();
+
+            if let Some(pkg_manifest_file) = manifest_map.get(id) {
+                let manifest_file = ManifestFile::from_dir(pkg_manifest_file.path())?;
+                let ty_program = match compile_result.value.and_then(|programs| programs.typed) {
+                    Some(ty_program) => ty_program,
+                    _ => bail!("CompileResult returned None"),
```
Braqzen, line 124: I would adjust both `bail` messages for two reasons:

1. They are the same so it's unclear which part of the code triggers the error
2. It's unclear to me what it means for a `CompilationResult` to return `None`. I would change it to be more descriptive to indicate what happened

sdankel (commented):
Can we add a test to verify the fix?

forc-plugins/forc-doc/src/main.rs
```diff
@@ -32,24 +28,31 @@ pub(crate) const ASSETS_DIR_NAME: &str = "static.files";
 
 /// Information passed to the render phase to get TypeInfo, CallPath or visibility for type anchors.
 #[derive(Clone)]
-struct RenderPlan {
+struct RenderPlan<'te, 'de> {
```
JoshuaBatty, line 31:
```suggestion
struct RenderPlan<'a> {
    document_private_items: bool,
    type_engine: &'a TypeEngine,
    decl_engine: &'a DeclEngine,
}
impl<'a> RenderPlan<'a> {
    fn new(
        document_private_items: bool,
        type_engine: &'a TypeEngine,
        decl_engine: &'a DeclEngine,
    ) -> RenderPlan<'a> {
        Self {
            document_private_items,
            type_engine,
            decl_engine,
        }
    }
}
```

We can probably share the same lifetime between the type and decl engines here.

eureka-cpu: > Can we add a test to verify the fix?

Hm, I'm not sure how you would test this since the fix doesn't really change the output but instead changes the strategy of building to use `forc-pkg`'s `BuildPlan`.

eureka-cpu:
> Can we add a test to verify the fix?

Hm, I'm not sure how you would test this since the fix doesn't really change the output but instead changes the strategy of building to use `forc-pkg`'s `BuildPlan`.

forc-plugins/forc-doc/src/main.rs
eureka-cpu: That makes sense, will do.

JoshuaBatty (approved):

mitchmindtree (approved):

PR #4617, opened 2023-06-01
Fixes sha256 returning erroneous results.

### LINKED ISSUE #4491: `sha256` gives erroneous results in some cases
`std::hash::sha256` gives wrong results for some strings, including `"test"`:

```sway
script;

use std::hash::sha256;

fn main() {
  // should be 9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08
  log(sha256("test")); 
  /*
    {
    "LogData": {
      "data": "9134a6432bb8da899b9005e41957835454b289f01b21f6f0373da8bcaac5c251",
      "digest": "4a5c62befb83ded8c020ca5cc13c7f8b059d331ab068a4e123735c9df999cdc1",
      "id": "0000000000000000000000000000000000000000000000000000000000000000",
      "is": 10344,
      "len": 32,
      "pc": 10452,
      "ptr": 10912,
      "ra": 0,
      "rb": 0
    }
  },
  */

  // should be 4a3cd7c8b44dbf7941e55179425f746adeaa97fe2d99b571fffee78e9b41743c
  log("Fastest Modular Execution Layer!");
  /*
    {
    "LogData": {
      "data": "46617374657374204d6f64756c617220457865637574696f6e204c6179657221",
      "digest": "4a3cd7c8b44dbf7941e55179425f746adeaa97fe2d99b571fffee78e9b41743c",
      "id": "0000000000000000000000000000000000000000000000000000000000000000",
      "is": 10344,
      "len": 32,
      "pc": 10456,
      "ptr": 10504,
      "ra": 0,
      "rb": 1
    }
  },
  */
}
```

esdrubal:
## Description

The bug fixed was caused by sha256 being called with a byte size greater than the string provided. This was caused by the intrinsic `__size_of` returning the size padded to the next word alignment.

The bug was fixed by introducing two new intrinsics `__is_str_type` and
`__size_of_str`.

The hashing functions now use `__size_of_str` when the type of `T` is str.

Fixes #4491

## Checklist


IGI-111 (changes requested):
Nice find! I don't think it's a good idea to introduce this intrinsic though.

The bug here is on the standard library side, assuming that the memory footprint of a value is only made of significant data, which it should not assume.
But since padding isn't necessarily contiguous, the newly added intrinsic can't inform you accurately about the layout in every case, and I think that's misleading.

I also think std does not need this kind of help from the compiler, it should be able to figure out the layout of the significant data from the length and type in this case using a simple multiplication of the size of the underlying type and the number of elements with the guarantee that for non struct types, the values are contiguous.

I would however be favorable to introducing an intrinsic that does something like C++'s [`std::has_unique_object_representations`](https://en.cppreference.com/w/cpp/types/has_unique_object_representations) if it helps.

esdrubal:
The new changes fix the current `sha256` which is widely used, including in storage libraries.

I will work on adding a Rust like traits implementation for `Hash` and `Hasher` in another pull request. This will make sure places, where it is critical to have immutability of hashes between versions, can use this which probably is the case for storage libraries.

IGI-111 (approved):
Good enough as a quick fix while we redesign std :+1:

JoshuaBatty (approved):

PR #4630, opened 2023-06-05
initial implementation and tests

### LINKED ISSUE #4626: `to_be_bytes` and `to_le_bytes`
Add methods to convert common integer types into arrays of bytes.

SwayStar123:
## Description
closes #4626 

## Checklist

### DIFF
--- a/sway-lib-core/src/primitives.sw
+++ b/sway-lib-core/src/primitives.sw
@@ -1,5 +1,14 @@
 library;
 
+const OFF: u64 = 0xFF;
+const I: u64 = 8;
+const J: u64 = 16;
+const K: u64 = 24;
+const L: u64 = 32;
+const M: u64 = 40;
+const N: u64 = 48;
+const O: u64 = 56;
+
 impl u64 {
     /// The smallest value that can be represented by this integer type.
     pub fn min() -> u64 {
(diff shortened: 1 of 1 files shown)


sway-lib-core/src/primitives.sw
```diff
@@ -16,6 +25,144 @@ impl u64 {
...
+
+            srl  r1 input n;
+            and  r1 r1 off;
+            sw  output r1 i6;
+
+            srl  r1 input o;
+            and  r1 r1 off;
+            sw  output r1 i7;
+
+            output: [u8; 8]
```
IGI-111, line 64: It seems to me like everything except the SW instructions can be done outside assembly. Which I think would be clearer.

sway-lib-core/src/primitives.sw
```diff
@@ -1,5 +1,14 @@
 library;
 
+const OFF: u64 = 0xFF;
+const I: u64 = 8;
+const J: u64 = 16;
+const K: u64 = 24;
+const L: u64 = 32;
+const M: u64 = 40;
+const N: u64 = 48;
+const O: u64 = 56;
```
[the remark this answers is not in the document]
SwayStar123: What name would you suggest?

sway-lib-core/src/primitives.sw
SwayStar123: If I have to calculate all the values first before inputting them into an assembly block, wouldnt that be worse performance wise?  Would use way more registers im guessing. Right now with no optimizations from the compiler the hash function im writing is gonna be very gas intensive so I need all the optimizations I can get

sway-lib-core/src/primitives.sw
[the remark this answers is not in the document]
IGI-111: Anything more descriptive. I'd probably use hex literals instead of naming each byte mask here though.

sway-lib-core/src/primitives.sw
IGI-111: Fully endorse the assembly if it actually uses more registers. But I'm not sure that would be the case here.

IGI-111 (approved):

esdrubal (approved):

PR #4655, opened 2023-06-12
Sway Ref: Style guide pattern matching

### LINKED ISSUE #4654: Sway Ref: Style guide pattern matching

Braqzen:
## Description

Closes #4654 

## Checklist


docs/reference/src/code/language/style-guide/pattern_matching/src/lib.sw
```diff
@@ -0,0 +1,25 @@
...
+// ANCHOR_END: style_match_unnamed
+
+#[allow(dead_code)]
+// ANCHOR: style_match_named
+fn named_case(variable: u64) {
+    let value = match variable {
+        1 => 1,
+        2 => 3,
+        3 => 5,
+        default => 0,
```
IGI-111, line 22: I'm not very keen on recommending an unused capture like this, if you want a default case I believe you should be using `_` every time unless you need the value.

Braqzen: I hold no strong opinion in this case.
I think it makes sense to generally use descriptive names to provide context. I don't see how it would introduce an issue.
I can remove it if you'd like.

What is your reasoning for not providing a name?

IGI-111: >What is your reasoning for not providing a name?

It is an unused declaration that the compiler will warn you against, for one.

I do believe it might be appropriate to use `_something` to express the particular property of a default, but `_default` would be redundant as `_` already expresses the default case by definition.

Braqzen: What do you think of 04df330c5daaa65410f7ec5020a3beae234415c7 ?

IGI-111: It's not quite accurate. I would rather say:

```
## Alternative

We may apply an appropriate name to provide context to the reader; however, unless it provides additional information the preferred usage is defined in the [`encouraged`](#encouraged) case.

```

As a separate issue, `default` has to be replaced by `_default` here. Style guide examples shouldn't produce compiler warnings.

IGI-111 (approved):

JoshuaBatty (approved):

PR #4671, opened 2023-06-15
Sway Ref: Style Guide hoisting variables

### LINKED ISSUE #4662: Sway Ref: Style Guide hoisting variables

Braqzen:
## Description

Closes #4662 

## Checklist

### DIFF
--- a/docs/reference/src/SUMMARY.md
+++ b/docs/reference/src/SUMMARY.md
@@ -71,6 +71,7 @@
   - [Pattern Matching](./documentation/language/style-guide/pattern-matching.md)
   - [Comments](./documentation/language/style-guide/comments.md)
   - [Getter Functions](./documentation/language/style-guide/getters.md)
+  - [Variable Hoisting](./documentation/language/style-guide/variable-hoisting.md)
 
 ## Common Operations
 
--- a/docs/reference/src/code/Forc.lock
+++ b/docs/reference/src/code/Forc.lock
@@ -1,27 +1,27 @@
 [[package]]
 name = 'annotation_style'
 source = 'member'
-dependencies = ['std']
+dependencies = ['std path+from-root-0F28276288D2432C']
 
 [[package]]
 name = 'annotations'
 source = 'member'
-dependencies = ['std']
+dependencies = ['std path+from-root-0F28276288D2432C']
 
 [[package]]
 name = 'arrays'
 source = 'member'
-dependencies = ['std']
+dependencies = ['std path+from-root-0F28276288D2432C']
 
 [[package]]
 name = 'assertions'
 source = 'member'
-dependencies = ['std']
+dependencies = ['std path+from-root-0F28276288D2432C']
 
 [[package]]
 name = 'asset_operations'
 source = 'member'
-dependencies = ['std']
+dependencies = ['std path+from-root-0F28276288D2432C']
 
 [[package]]
 name = 'associated-consts'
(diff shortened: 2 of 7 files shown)


IGI-111:
I must say I don't really like recommending C-like hoisting in the first place. Like Rust, we have fully scoped variables and there are a number of idiomatic patterns that completely break the assumption that variables are declared in a single place. For good reasons.

In my experience this language style actually lends itself more to declarations that are closer to the usage, but, much like the Rust style guide, I believe we should be agnostic on the matter. Unlike in ANSI C, there are times when hoisting is appropriate and others when it isn't.

Braqzen:
> I must say I don't really like recommending C-like hoisting in the first place. Like Rust, we have fully scoped variables and there are a number of idiomatic patterns that completely break the assumption that variables are declared in a single place such as repeatedly shadowing a particular variable name to conserve const-ness.
> 
> In my experience this language style actually lends itself more to declarations that are closer to the usage, but, much like the Rust style guide, I believe we should be agnostic on the matter. Unlike in ANSI C, there are times when hoisting is appropriate and others when it isn't. I don't think we should take a side and it should be left to the Sway programmer.

That's fine but that is not the reasoning I've provided above. The section has nothing to do with your comments and instead it addresses general code quality which leans towards hoisting. That's about it

IGI-111:
But I am talking about code quality. I simply don't believe that leaning towards hoisting variables is a good idea in general.

I'm unconvinced there is a general rule.
But if there had to be one, I think it would actually be the opposite to what is being recommended here. Variables should be scoped to their usage to help the optimizer do its job.

That's what new-ish PL guidelines usually recommend, see [C++](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es6-declare-names-in-for-statement-initializers-and-conditions-to-limit-scope) for instance.
The reason older style guides used to promote this rule is mostly because K&R said 

>In C, all variables must be declared before they are used, usually at the beginning of the
function before any executable statements.

And people replicated that style. But the cause wasn't that it was clearer or better structured, it was simply an early limitation of C's grammar that segregated declarations and statements.

We have no such limitations. Hoisting actually hurts performance, and the clarity gains are, as per my earlier point, entirely situational.
Hence I see no reason to recommend it.

Braqzen:
> But I am talking about code quality. I simply don't believe that leaning towards hoisting variables is a good idea in general.
> 
> I'm unconvinced there is a general rule. But if there had to be one, I think it would actually be the opposite to what is being recommended here. Variables should be scoped to their usage to help the optimizer do its job.
> 
> That's what new-ish PL guidelines usually recommend, the exact opposite, see [C++](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es6-declare-names-in-for-statement-initializers-and-conditions-to-limit-scope) for instance. The reason older style guides used to promote this rule is mostly because K&R said
> 
> > In C, all variables must be declared before they are used, usually at the beginning of the
> > function before any executable statements.
> 
> And people replicated that style. But the cause wasn't that it was clearer or better structured, it was simply an early limitation of C's grammar that segregated declarations and statements.
> 
> We have no such limitations. Hoisting actually hurts performance, and the clarity gains are, as per my earlier point, entirely situational. Hence I see no reason to recommend it.

From the point of the compiler optimization that is a fair point; however, in terms of readability I think my comments make sense too.

I believe Mohammad said something like "a sufficiently complex compiler should be able to do X" (paraphrasing) with the implication being that this type of optimization is a non-issue (temporary issue at best) and thus we should focus on code quality - readability in the form of some standard which in this case is pretty much "don't scatter your variables randomly". I lean towards hoisting for the reasons in the text.

So, what do you recommend that we do now?

- Remove entirely
- Rephrase content to outline two styles without using language that indicates a preference
- Something else

IGI-111:
Code motion should certainly get you out of some these ordering issues (once we have it implemented) but it doesn't always work.

C++ cares about this because they use scoping to represent the lifetimes of ressources *a lot*.
Rust inherits this, but it also has other idioms that make this less clear.

I think we should be agnostic on this like Rust until patterns emerge through usage.
So I move to just remove this section entirely.

PR #4673, opened 2023-06-16
Refactor the `server.rs` module in the language server.

### LINKED ISSUE #4672: Refactor the `server.rs` module in the language server
We should refactor out a lot of code in the server.rs module in preparation for deprecating tower-lsp in favour of a custom event loop. 

Ideally this module just provides an interface between the LSP protocol and the sway-lsp internals.

JoshuaBatty:
## Description
This PR refactors most of the code that was in `server.rs`. Specifically, the call backs have been moved to `handlers/notification.rs` and `handlers/request.rs`. This was inspired by `rust-analyzer`s approach. 

the `server.rs` module now looks like:

```rust
#[tower_lsp::async_trait]
impl LanguageServer for ServerState {
    async fn initialize(&self, params: InitializeParams) -> Result<InitializeResult> {
        request::handle_initialize(self, params)
    }

    async fn shutdown(&self) -> Result<()> {
        self.shutdown_server()
    }

    async fn did_open(&self, params: DidOpenTextDocumentParams) {
        notification::handle_did_open_text_document(self, params).await;
    }

    async fn did_change(&self, params: DidChangeTextDocumentParams) {
        notification::handle_did_change_text_document(self, params).await;
    }

    async fn did_save(&self, params: DidSaveTextDocumentParams) {
        notification::handle_did_save_text_document(self, params).await;
    }

    async fn did_change_watched_files(&self, params: DidChangeWatchedFilesParams) {
        notification::handle_did_change_watched_files(self, params).await;
    }

    async fn hover(&self, params: HoverParams) -> Result<Option<Hover>> {
        request::handle_hover(self, params)
    }

    async fn code_action(&self, params: CodeActionParams) -> Result<Option<CodeActionResponse>> {
        request::handle_code_action(self, params)
    }

    async fn code_lens(&self, params: CodeLensParams) -> Result<Option<Vec<CodeLens>>> {
        request::handle_code_lens(self, params)
    }
    
    etc ...
}
```

A new type called `ServerState` now holds the config, keyword docs, client and the map of sessions. No new functionality has been added apart from refactoring the code. 

This is needed for the new event loop, I decided to separate that into a follow up PR to make things a bit more manageable. 

closes #4672

## Checklist


sway-lsp/src/global_state.rs
```diff
@@ -0,0 +1,176 @@
...
+        Err(err) => {
+            tracing::error!("{}", err);
+            matches!(err, LanguageServerError::FailedToParse)
+        }
+    })
+    .await
+    .unwrap_or_default()
+}
+
+/// `Sessions` is a collection of `Session`s, each of which represents a project
```
sdankel, line 120:
```suggestion
/// `Sessions` is a collection of [Session]s, each of which represents a project
```

sway-lsp/src/handlers/notification.rs
```diff
@@ -0,0 +1,102 @@
+//! This module is responsible for implementing handlers for Language Server
+//! Protocol. This module specifically handles notifications.
```
sdankel, line 2: In case it's not obvious what LSP notifications are:

```suggestion
//! Protocol. This module specifically handles async messages from the client.
```

sway-lsp/src/handlers/notification.rs
```diff
@@ -0,0 +1,102 @@
...
+
+use crate::{capabilities, core::sync, global_state::GlobalState};
+use forc_pkg::PackageManifestFile;
+use lsp_types::{
+    DidChangeTextDocumentParams, DidChangeWatchedFilesParams, DidOpenTextDocumentParams,
+    DidSaveTextDocumentParams, FileChangeType,
+};
+
+pub(crate) async fn handle_did_open_text_document(
+    state: &GlobalState,
```
sdankel, line 12: My preference is to have these functions encapsulated in a struct that has `&GlobalState` as a member, so that we don't need to pass it into every function.

sway-lsp/src/handlers/request.rs
```diff
@@ -0,0 +1,398 @@
...
+    fs::File,
+    io::Write,
+    path::{Path, PathBuf},
+};
+use sway_types::{Ident, Spanned};
+use tower_lsp::jsonrpc::Result;
+use tracing::metadata::LevelFilter;
+
+pub(crate) fn handle_initialize(
+    state: &GlobalState,
```
sdankel, line 26: Same comment here, if we can avoid having to pass `&GlobalState` to every function, I think it's cleaner.

In the functions below where `snap` is passed as a parameter, that could be replaced by calling `self.state.snapshot()` directly.

We could even make the dynamic `sessions` inside of `GlobalState` private so  handlers are forced to call `snapshot()`.

sway-lsp/src/handlers/request.rs
```diff
@@ -0,0 +1,398 @@
...
+    tracing::info!("Initializing the Sway Language Server");
+    Ok(InitializeResult {
+        server_info: None,
+        capabilities: crate::server_capabilities(),
+        ..InitializeResult::default()
+    })
+}
+
+pub(crate) fn handle_document_symbol(
+    snap: GlobalStateSnapshot,
```
sdankel, line 54: So these request handlers only operate on the state of the sessions at the time they were invoked. This makes sense and will avoid issues where tokens are missing because different functions are operating on a different view of the session. 

This means that `sessions` is cloned every time we call one of these handler functions, right? Is there significant memory overhead in doing this?

sway-lsp/src/global_state.rs
```diff
@@ -0,0 +1,176 @@
...
+
+impl Sessions {
+    fn init(&self, uri: &Url) -> Result<(), LanguageServerError> {
+        let session = Arc::new(Session::new());
+        let project_name = session.init(uri)?;
+        self.insert(project_name, session);
+        Ok(())
+    }
+
+    pub(crate) fn get_uri_and_session(
```
sdankel, line 132: nit: a better name for this function could be something like `from_workspace_uri` - something to indicate it's pulling everything you need from Sessions given a workspace URI.

sway-lsp/src/lib.rs
```diff
@@ -1,23 +1,81 @@
...
+    WorkDoneProgressOptions,
+};
+use tower_lsp::{LspService, Server};
 
-    let (service, socket) = LspService::build(Backend::new)
-        .custom_method("sway/show_ast", Backend::show_ast)
-        .custom_method("textDocument/inlayHint", Backend::inlay_hints)
+pub async fn start() {
+    let (service, socket) = LspService::build(GlobalState::new)
+        .custom_method("sway/show_ast", GlobalState::show_ast)
```
sdankel, line 28: It's a little funky to have the custom capability handlers be part of `GlobalState`.  Might be better to put them in their own module under the handlers directory, like `custom.rs` or `lsp_ext.rs`. Or even put them in with `request.rs` since they are request handlers.

sway-lsp/src/server.rs
```diff
@@ -1,659 +1,120 @@
...
-                    Ok(_) => {
-                        self.parse_project(uri, params.text_document.uri.clone(), session)
-                            .await;
-                    }
-                    Err(err) => tracing::error!("{}", err.to_string()),
-                }
-            }
-            Err(err) => tracing::error!("{}", err.to_string()),
-        }
+        notification::handle_did_change_text_document(self, params).await;
```
sdankel, line 40: Just curious, what happens if we don't await these? Do they still execute fully or do they get interrupted?

sdankel (commented):
This looks good overall and makes sense to me. I left some comments around code organization. Nice work!

sway-lsp/src/server.rs
JoshuaBatty: these are `async` functions that return a `Future` when called. In order for the `Future` to be executed we need to call `.await` on them.

sway-lsp/src/global_state.rs
JoshuaBatty: Nice, i'm happy to make this change. Feels more idiomatic.

sway-lsp/src/lib.rs
JoshuaBatty: Yeah I agree, this is just a temporary work around as tower-lsp requires that these methods are added to the type `LanguageServer` has been implemented for. In this case it's `GlobalState` this won't be the case once we move to the new event loop.

sway-lsp/src/handlers/notification.rs
JoshuaBatty: Ended up going with:

```
//! This module is responsible for implementing handlers for Language Server
//! Protocol. This module specifically handles notification messages sent by the Client.
```

sway-lsp/src/global_state.rs
```diff
@@ -0,0 +1,176 @@
...
+pub(crate) struct GlobalStateSnapshot {
+    pub(crate) config: Arc<RwLock<Config>>, // TODO: remove RwLock once we deprecate tower-lsp
+    pub(crate) keyword_docs: Arc<KeywordDocs>,
+    pub(crate) sessions: Arc<Sessions>,
+}
+
+impl GlobalState {
+    pub fn new(client: Client) -> GlobalState {
+        let sessions = Arc::new(Sessions(DashMap::new()));
+        let config = Arc::new(RwLock::new(Default::default()));
```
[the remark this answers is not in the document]
JoshuaBatty: The idea behind the snapshot to to have a type that holds all of the read-only data from the GlobalState. I'm happy to remove the snapshot all together for now and re-look at it when the eventloop PR goes up and it probably makes more sense within that context.

sway-lsp/src/handlers/request.rs
JoshuaBatty: It's just cloning the Arc which is a pointer so minimal overhead. I'll revert the snapshot until a future PR though.

sway-lsp/src/handlers/request.rs
JoshuaBatty: I agree that it would make more sense for these methods to take self and have them implemented on `GlobalState` directly as the `LanguageServer` trait is implemented for this type. That means that essentially the LSP event loop implemented in tower is operating on `GlobalState`. 

This won't be the case is the follow up PR for the event loop so I would like to keep these as they are for now. We can re look if this is still and issue when the eventloop PR is ready.

sway-lsp/src/global_state.rs
```diff
@@ -0,0 +1,176 @@
...
+use dashmap::DashMap;
+use forc_pkg::PackageManifestFile;
+use lsp_types::Url;
+use parking_lot::RwLock;
+use std::{path::PathBuf, sync::Arc};
+use tokio::task;
+use tower_lsp::{jsonrpc, Client};
+
+/// `GlobalState` is the primary mutable state of the language server
+pub struct GlobalState {
```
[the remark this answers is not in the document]
JoshuaBatty: I've renamed this to `ServerState`

kayagokalp (approved):
utACK

PR #4683, opened 2023-06-19
Additional identity methods

SwayStar123:
## Description
Added some utility methods to the Identity struct

## Checklist

- [] I have linked to any relevant issues.


sway-lib-std/src/identity.sw
```diff
@@ -21,3 +24,53 @@ impl core::ops::Eq for Identity {
         }
     }
 }
+
+impl Identity {
+    pub fn address(self) -> Result<Address, ContractId> {
```
IGI-111, line 29: Should be `as_address`

sway-lib-std/src/identity.sw
```diff
@@ -21,3 +24,53 @@ impl core::ops::Eq for Identity {
...
+        match self {
+            Identity::Address(address) => Result::Ok(address),
+            Identity::ContractId(contract_id) => Result::Err(contract_id),
+        }
+    }
+
+    pub fn contract_id(self) -> Result<ContractId, Address> {
```
IGI-111, line 36: Should be `as_contract_id`

sway-lib-std/src/identity.sw
```diff
@@ -21,3 +24,53 @@ impl core::ops::Eq for Identity {
         }
     }
 }
+
+impl Identity {
```
IGI-111, line 28: Both methods should probably return `Option` instead of `Result`. There is no reason to assume in the type interface that there are only two possibilities. And proper types shouldn't be errors.

IGI-111 (changes requested):

sway-lib-std/src/identity.sw
```diff
@@ -21,3 +24,62 @@ impl core::ops::Eq for Identity {
...
+    pub fn as_address(self) -> Option<Address> {
+        match self {
+            Identity::Address(address) => Option::Some(address),
+            Identity::ContractId(_) => Option::None(),
```
Braqzen, line 32:
```suggestion
            Identity::ContractId(_) => Option::None,
```

sway-lib-std/src/identity.sw
```diff
@@ -21,3 +24,62 @@ impl core::ops::Eq for Identity {
...
+        }
+    }
+
+    pub fn as_contract_id(self) -> Option<ContractId> {
+        match self {
+            Identity::Address(_) => Option::None(),
```
Braqzen, line 38:
```suggestion
            Identity::Address(_) => Option::None,
```

sway-lib-std/src/identity.sw
```diff
@@ -21,3 +24,62 @@ impl core::ops::Eq for Identity {
...
+            Identity::ContractId(_) => true,
+        }
+    }
+
+    pub fn value(self) -> b256 {
+        match self {
+            Identity::Address(address) => address.value,
+            Identity::ContractId(contract_id) => contract_id.value,
+        }
+    }
```
Braqzen, line 62:
```suggestion
```

This should not be added because it defeats the purpose of using the wrapped types, doubly so on the `Identity`.

SwayStar123: I disagree, you can use the wrapper and then still the actual value for some other stuff, like low level instruction calls, this is convenient for that

Braqzen: "I disagree" is not a valid counter argument to type safety.

This does not belong on the `Identity` type. Users can extract it themselves via the wrapped types.

SwayStar123: Its not breaking type safety, you can just not use this function. Users can already access the values yes, and this is just a utility method to make that easier. Address and ContractId also have a into() fn for converting to b256 so I dont see why this is any worse than that

Braqzen: The value is already a private type that at best should only be accessible through the wrapper if it allows it. The `Identity` is another wrapper on top of that, a super type. It should have no concept of what its variants' internal structure is like. That data is encapsulated for a reason. If the user wants the internal value then it needs to query it via the wrapper.

@IGI-111 looking at the field, it is a private field but we do not respect that. Is that intentional or should we raise an issue and force the use of the `pub` keyword on the fields?

Braqzen (changes requested):

sway-lib-std/src/identity.sw
IGI-111: We don't have privacy for struct fields yet. Though that should be fixed.
In any case I don't think we should be writing code into the standard library that assumes fields are public like this.

But the real reason not to do this is that it is a type system escape hatch. This methods allows you to mix the two types of identities without explicitly ascertaining which one you are using, and that's bad design. I vote for removal too. If you want the raw value of something, you should know what it is first. What use is the value otherwise?

IGI-111 (approved):

Braqzen (approved):

PR #4687, opened 2023-06-19
Blanket trait implementations

### LINKED ISSUE #4010: Blanket implementations in Sway
Blanket implementations are important for issues like https://github.com/FuelLabs/sway/issues/3637. They're also useful for storage references so that we can do something like:
```rust
pub struct StorageRef<T> {
    key: b256,
}

pub trait AsRef<T> {
    fn as_ref(self) -> StorageRef<T>;
}

impl AsRef<T> for T {
    #[inline(never)]
    fn as_ref(self) -> StorageRef<T> {
        StorageRef {
            key: __get_storage_key(),
        }
    }
}
```
or something similar.

IGI-111:
## Description

This change allows blanket implementations for traits, i.e.:
```sway
impl<T> SomeTrait for T {}
```
As part of this we now import trait implementations that are in the same namespace as an imported trait.

Ideally we'd want to import all existing trait implementations this way, but this would require another pass or a query, which is probably out of scope for a first implementation.

Traits are deduplicated here using the trait decl span.

Fix #4010

## Checklist

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -1683,6 +1683,7 @@ pub fn dependency_namespace(
         &[CORE, PRELUDE].map(|s| Ident::new_no_span(s.into())),
         &[],
         engines,
+        true,
     );
 
     if has_std_dep(graph, node) {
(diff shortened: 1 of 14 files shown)


JoshuaBatty (approved):

SwayStar123:
lgtm

esdrubal (approved):

Issue #4569, opened 2023-05-18
Distinction between Identity/Contract/Address


MujkicA:
From this discussion https://github.com/FuelLabs/fuels-rs/issues/952 it was suggested that the distinction between Identity/Contract/Address represents a risk factor for people building with sway. 

Unless the developers have the foresight to use Identity as the type for their admin address, they will not be composable with other contracts.

There seems to be a considerable risk associated with as developers coming from other ecosystems are likely to assume that there is no such limitations when just using `Address`.

diyahir:
TLDR: (Highly recommend reading the thread)

1. I was annoyed by not being able to test my contract from a wallet because it was an address and not a contractId. 
2. Using Identity caused some casting issues like when using abi(....) since it expects a contractId
3. Realizing that this is actually a much bigger problem because a multisig is a contractId, so if you use address by accident then you are no longer composable with other contracts who may want to open a position / interface with you
4. I look to see if anyone has made this mistake: prominent project SwayLend, a compound fork, has made it (not faulting them at all) but their current implementation is not composable with other contracts 
[https://github.com/compolabs/sway-lend/blob/master/contracts/market/src/main.sw](https://github.com/compolabs/sway-lend/blob/master/contracts/market/src/main.sw)
6. This system is annoying and introduces incompatibility bugs + makes testing contracts much harder 

Fundamentally, ANY action done by a wallet should be able to be done by a DAO/multisig (full stop, no exceptions). The current scheme means that developers need to have the foresight to allow contractIds (multisgs) to be composable. You can recommend Identity everywhere sure, but this still introduces the potential for composability bugs + makes testing contracts very hard. If this distinction is not a value ADD then please remove it. 

This is a critical mistake, let's ensure that DAOs / Multisigs are not second class citizens in Fuel by removing this distinction.

IGI-111:
Seems like we want two things here:
1. Allow users to express categories of addresses within the type system, to express what is and isn't intended to be used by contracts
2. Prevent common user errors that restrict types in a way that is surprising and not obvious

I would say we should at least:
* Document the distinction better, especially make it clear that `Address` and `ContractId` are less permissive than one would expect
* Structure the standard library in a way that holds these types together so the connection is obvious
* Provide `From` implementations to make the proper conversions easy
* Consider making the improper conversions possible but slightly harder or more explicitly unsafe
* Consider renaming `Address` to something that is more explicit about the restriction

diyahir:
[...]

PR #4694, opened 2023-06-20
Add `GTF_OUTPUT_COIN_ASSET_ID` and `GTF_OUTPUT_COIN_TO` opcode getters

bitzoic:
## Description

Added getters for the `GTF_OUTPUT_COIN_ASSET_ID` and `GTF_OUTPUT_COIN_TO` opcodes. These have been highly requested by external teams for use in predicates.

Note: The `output_asset_to()` getter currently returns a `b256` rather than an `Identity`. This should be updated when https://github.com/FuelLabs/sway/issues/4569 is resolved.

## Checklist

### DIFF
--- a/sway-lib-std/src/outputs.sw
+++ b/sway-lib-std/src/outputs.sw
@@ -99,7 +99,7 @@ pub fn output_asset_id(index: u64) -> Option<ContractId> {
     }
 }
 
-/// If the output's type is `Output::Coin` return the to b256 as an `Some(to)`.
+/// If the output's type is `Output::Coin` return the b256 as `Some(to)`.
 /// Otherwise, returns `None`.
 /// TODO: Update to `Identity` when https://github.com/FuelLabs/sway/issues/4569 is resolved
 pub fn output_asset_to(index: u64) -> Option<b256> {
--- a/sway-lib-std/src/outputs.sw
+++ b/sway-lib-std/src/outputs.sw
@@ -12,13 +12,14 @@ use ::tx::{
     Transaction,
     tx_type,
 };
+use ::option::*;
 
 // GTF Opcode const selectors
 //
 pub const GTF_OUTPUT_TYPE = 0x201;
-// pub const GTF_OUTPUT_COIN_TO = 0x202;
+pub const GTF_OUTPUT_COIN_TO = 0x202;
 pub const GTF_OUTPUT_COIN_AMOUNT = 0x203;
-// pub const GTF_OUTPUT_COIN_ASSET_ID = 0x204;
+pub const GTF_OUTPUT_COIN_ASSET_ID = 0x204;
 // pub const GTF_OUTPUT_CONTRACT_INPUT_INDEX = 0x205;
 // pub const GTF_OUTPUT_CONTRACT_BALANCE_ROOT = 0x206;
 // pub const GTF_OUTPUT_CONTRACT_STATE_ROOT = 0x207;
@@ -86,5 +87,24 @@ pub fn output_amount(index: u64) -> u64 {
                 r1: u64
             }
         },
+    }
+}
+
+/// If the output's type is `Output::Coin` return the asset ID as an `Some(id)`.
+/// Otherwise, returns `None`.
+pub fn output_asset_id(index: u64) -> Option<ContractId> {
+    match output_type(index) {
+        Output::Coin => Option::Some(ContractId::from(__gtf::<b256>(index, GTF_OUTPUT_COIN_ASSET_ID))),
+        _ => Option::None,
+    }
+}
+
+/// If the output's type is `Output::Coin` return the to b256 as an `Some(to)`.
+/// Otherwise, returns `None`.
+/// TODO: Update to `Identity` when https://github.com/FuelLabs/sway/issues/4569 is resolved
+pub fn output_asset_to(index: u64) -> Option<b256> {
+    match output_type(index) {
+        Output::Coin => Option::Some(__gtf::<b256>(index, GTF_OUTPUT_COIN_TO)),
+        _ => Option::None,
     }
 }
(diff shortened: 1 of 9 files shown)


test/src/sdk-harness/test_artifacts/tx_output_predicate/src/main.sw
```diff
@@ -0,0 +1,14 @@
...
+
+use std::outputs::{output_asset_id, output_asset_to};
+
+fn main(index: u64, asset_id: ContractId, to: b256) -> bool {
+
+    let tx_asset_id = output_asset_id(index);
+    let tx_to = output_asset_to(index);
+
+    assert(tx_asset_id.is_some() && tx_asset_id.unwrap() == asset_id);
+    assert(tx_to.is_some() && tx_to.unwrap() == to);
```
Braqzen, line 11: Quality is important but keep in mind that this is a test so unwrapping on a `None` is a valid failure since the test should not be `None` in the 1st place

test/src/sdk-harness/test_projects/tx_fields/mod.rs
```diff
@@ -142,6 +147,41 @@ async fn add_message_input(tx: &mut ScriptTransaction, wallet: WalletUnlocked) {
     tx.tx.inputs_mut().push(message_input);
 }
 
+async fn get_output_predicate() -> (WalletUnlocked, Predicate, AssetId) {
```
Braqzen, line 150: The convention in Rust is to not prepend names with `get_`

test/src/sdk-harness/test_projects/tx_fields/mod.rs
```diff
@@ -726,6 +766,22 @@ mod outputs {
...
+            let balance = predicate.get_asset_balance(&asset_id).await.unwrap();
+
+            predicate
+                .transfer(wallet.address(), 10, asset_id, TxParameters::default())
+                .await
+                .unwrap();
+
+            let new_balance = predicate.get_asset_balance(&asset_id).await.unwrap();
+
+            assert!(balance - 10 == new_balance);
```
Braqzen, line 783: I'd set the `10` to a variable

test/src/sdk-harness/test_projects/tx_fields/mod.rs
Braqzen, line 1: Should there be additional tests for reverting?

test/src/sdk-harness/test_projects/tx_fields/mod.rs
bitzoic: This follows the same syntax as other functions with similar functionality in this file.

sway-lib-std/src/outputs.sw
```diff
@@ -88,3 +89,22 @@ pub fn output_amount(index: u64) -> u64 {
...
+pub fn output_asset_id(index: u64) -> Option<ContractId> {
+    match output_type(index) {
+        Output::Coin => Option::Some(ContractId::from(__gtf::<b256>(index, GTF_OUTPUT_COIN_ASSET_ID))),
+        _ => Option::None,
+    }
+}
+
+/// If the output's type is `Output::Coin` return the to b256 as an `Some(to)`.
+/// Otherwise, returns `None`.
+/// TODO: Update to `Identity` when https://github.com/FuelLabs/sway/issues/4569 is resolved
```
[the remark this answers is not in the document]
bitzoic: I agree, but there is no set date or timeline for the issue to be resolved and this is something external teams need today.

test/src/sdk-harness/test_artifacts/tx_output_predicate/src/main.sw
bitzoic: If either result is `None` the assertion will fail and therefore fail the test, making it a valid failure

K1-R1 (commented):
Changes seem fine, but CI is failing

K1-R1 (dismissed):

Braqzen (commented):
1) In the setup the wallet transfers 2 assets. Do we need to transfer the second asset to the predicate prior to testing or can that transfer be removed?
2) When unpacking variables from the setup function in the tests I would use the same variable names throughout instead of sometimes using `wallet` and then `wallet1` etc.

bitzoic:
> 1. In the setup the wallet transfers 2 assets. Do we need to transfer the second asset to the predicate prior to testing or can that transfer be removed?
> 2. When unpacking variables from the setup function in the tests I would use the same variable names throughout instead of sometimes using `wallet` and then `wallet1` etc.

1. This is a setup function. It is setting up. As more GTFopcodes come online this will be needed.
2. This is done in tests to purposely create distinctions. Nothing is more frustrating than seeing a test fail and not know why because all the variables are named the same.

Braqzen:
> > 1. In the setup the wallet transfers 2 assets. Do we need to transfer the second asset to the predicate prior to testing or can that transfer be removed?
> > 2. When unpacking variables from the setup function in the tests I would use the same variable names throughout instead of sometimes using `wallet` and then `wallet1` etc.
> 
> 1. This is a setup function. It is setting up. As more GTFopcodes come online this will be needed.
> 2. This is done in tests to purposely create distinctions. Nothing is more frustrating than seeing a test fail and not know why because all the variables are named the same.

1) It's a bad practice to add code in anticipation of future features unless you are about to work on those features soon because it creates technical debt as priorities change and the code sits there indefinitely. As things change there is more maintenance on top of polluting the current PR with code that does not serve a function (and thus it should be added later).
2) It's fine to have different names. I'm referring more towards consistency. For example, the last 2 tests that revert use 2 different variable names for the asset. In the top test the 2nd variable is called `asset_id` when it should be called `asset_id2` thus validating the name `asset_id1` in the lower test in the first position. Similarly for any other variable names

Braqzen (approved):

SwayStar123 (approved):

PR #4695, opened 2023-06-20
fix: broken links, ignore DS_Store

sarahschwartz:
fixes a couple broken links in the docs and adds .DS_Store to .gitignore.


kayagokalp (approved):
utAck. We should probably add broken link checking step to CI 🤔. Also would linking to latest versions of the books (instead of master version) make sense?

IGI-111 (approved):

PR #4697, opened 2023-06-21
Fixes trait map issue on the use of absolute path

### LINKED ISSUE #4696: Trait map bug on use with absolute path
I found the following compile error while using the below test case.

```
15 |     let _s = S::<Foo> {};
   |              ^^^^^^^^^^^ Trait "Trait" is not implemented for type "Foo".
```

**main.sw**
```rust
script;

mod r#trait;
mod foo;

use ::foo::*;
use ::trait::*;
use std::assert::*;

struct S<T> where T: Trait {}

fn main() -> u64 {
    assert(Foo::method() == 42);

    let _s = S::<Foo> {};

    1
}
```
**trait.sw**
```rust
library;

pub trait Trait {
    fn method() -> u64;
}
```
**foo. sw**
```rust
library;

use ::trait::*;

pub struct Foo {
    foo: u32,
}

impl Trait for Foo {
    fn method() -> u64 {
        42
    }
}
```

esdrubal:
## Description

The test case use_absolute_path included in this PR was generating an error due to the `CallPath` `to_fullpath` generating `use_absolute_path::foo::trait::Trait` instead of `use_absolute_path::trait::Trait`.

The issue was fixed by including a boolean in `namespace.use_synonyms` that indicates when the path is absolute. With that information `to_fullpath` can behave accordingly.

Fixes #4696

## Checklist

### DIFF
--- a/sway-core/src/semantic_analysis/namespace/module.rs
+++ b/sway-core/src/semantic_analysis/namespace/module.rs
@@ -236,6 +236,7 @@ impl Module {
         src: &Path,
         dst: &Path,
         engines: &Engines,
+        is_absolute: bool,
     ) -> CompileResult<()> {
         let mut warnings = vec![];
         let mut errors = vec![];
@@ -271,7 +272,12 @@ impl Module {
         for symbol_and_decl in symbols_and_decls {
             dst_ns.use_synonyms.insert(
                 symbol_and_decl.0,
-                (src.to_vec(), GlobImport::Yes, symbol_and_decl.1),
+                (
+                    src.to_vec(),
+                    GlobImport::Yes,
+                    symbol_and_decl.1,
+                    is_absolute,
+                ),
             );
         }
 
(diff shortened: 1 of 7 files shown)


sway-core/src/semantic_analysis/namespace/module.rs
```diff
@@ -236,6 +236,7 @@ impl Module {
         src: &Path,
         dst: &Path,
         engines: &Engines,
+        is_absolute: bool,
```
IGI-111, line 239: I think it should be clearer whether the source or destination path is the absolute one.

IGI-111: Would using `CallPath` to bundle that information make sense here maybe?

esdrubal: I renamed `is_absolute` to `is_src_absolute`.
Changing this src to CallPath should probably be done in a dedicated PR as it will likely propagate CallPaths to other places. I created an issue for this: #4700.

IGI-111: Makes sense, I think this is good as is then.

IGI-111 (approved):

JoshuaBatty (approved):

PR #4707, opened 2023-06-27
Handle integer overflow/underflow for non-64-bit integers

### LINKED ISSUE #4646: All integers are treated a u64
Sway currently only handles overflow for u64. Other integer types are just treated like a u64 and can contain larger values than they should be allowed to represent.

For instance this test does *not* panic and instead logs a value that is larger than `U32_MAX`:
```
#[test]
fn test_over_flow_works() {
    let a: u32 = 4294967295;
    let b: u32 = 10000; 

    let result: u32 = a + b;
    log(result);
}
```

anton-trunov:
fixes #4646

## Description

This PR fixes the linked issue by introducing overflow/underflow checks in the `Add`, `Mul`, `Sub` stdlib traits implementations for the `u8`, `u16` and `u32` types.

## Checklist

### DIFF
--- a/docs/book/src/basics/built_in_types.md
+++ b/docs/book/src/basics/built_in_types.md
@@ -42,6 +42,13 @@ Numbers can be declared with binary syntax, hexadecimal syntax, base-10 syntax, 
 <!-- This section should explain the default numeric type in Sway -->
 <!-- default_num:example:start -->
 The default numeric type is `u64`. The FuelVM's word size is 64 bits, and the cases where using a smaller numeric type saves space are minimal.
+
+If a 64-bit arithmetic operation produces an overflow or an underflow,
+computation gets reverted automatically by FuelVM.
+
+8/16/32-bit arithmetic operations are emulated using their 64-bit analogues with
+additional overflow/underflow checks inserted, which generally results in
+somewhat higher gas consumption.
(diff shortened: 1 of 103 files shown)


vaivaswatha:
We'll probably need to do something about `const_eval` too
https://github.com/FuelLabs/sway/blob/48f7125dc0aee62301d1c79c222cbbe7e9138f7c/sway-
core/src/ir_generation/const_eval.rs#L527

The corresponding tests are in `should_pass/language/const_inits`.

IGI-111:
Yeah we probably need something for the const evaluations too.

Are you sure there are no missing operators, what about left shift?

anton-trunov:
> We'll probably need to do something about const_eval too

Noted, will add, thanks.

> Are you sure there are no missing operators, what about left shift?

Yeah, I was going to go over the logical operations as well, probably in a separate PR if any changes are required there, but it also makes sense to do it here in one go.

vaivaswatha (approved):

IGI-111 (commented):
LGTM, but I think the logical operators should also be in this PR if necessary.

This makes me think we might want to add `const fn` as a language construct eventually.

IGI-111 (approved):

vaivaswatha (approved):

PR #4727, opened 2023-07-03
filter out `CompileError::UnknownVariable` errors when `CONTRACT_ID` isn't injected into namespace

JoshuaBatty:
## Description
When `CONTRACT_ID` is used in a test, the compiler returns `CompileError::UnknownVariable` to LSP as `CONTRACT_ID` is only injected into the namespace during building and not checking. As such, we are inserting a dummy value during checking to avoid false error signals being reported in LSP. 

closes https://github.com/FuelLabs/sway-vscode-plugin/issues/154

## Checklist

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -2586,16 +2586,30 @@ pub fn check(
     let compiled_contract_deps = HashMap::new();
 
     let mut results = vec![];
-    for &node in plan.compilation_order.iter() {
+    for (idx, &node) in plan.compilation_order.iter().enumerate() {
         let pkg = &plan.graph[node];
         let manifest = &plan.manifest_map()[&pkg.id()];
+
+        // This is necessary because `CONTRACT_ID` is a special constant that's injected into the
+        // compiler's namespace. Although we only know the contract id during building, we are
+        // inserting a dummy value here to avoid false error signals being reported in LSP.
+        // We only do this for the last node in the compilation order because previous nodes
+        // are dependencies.
+        //
+        // See this github issue for more context: https://github.com/FuelLabs/sway-vscode-plugin/issues/154
+        let contract_id_value = if idx == plan.compilation_order.len() - 1 {
+            Some("0x0000000000000000000000000000000000000000000000000000000000000000".to_string())
+        } else {
+            None
+        };
+
         let dep_namespace = dependency_namespace(
             &lib_namespace_map,
             &compiled_contract_deps,
             &plan.graph,
             node,
             engines,
-            None,
+            contract_id_value,
         )
         .expect("failed to create dependency namespace");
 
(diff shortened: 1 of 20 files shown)


kayagokalp (dismissed):
This LGTM 🚀 we can alternatively inject a dummy `CONTRACT_ID` while checking as well, we currently pass `None` but we could pass all zeros etc too if that is cleaner for LSP. https://github.com/FuelLabs/sway/blob/cd63629167137c2c9d713760cddedcab23da1ec5/forc-pkg/src/pkg.rs#L2592

JoshuaBatty:
> This LGTM 🚀 we can alternatively inject a dummy `CONTRACT_ID` while checking as well, we currently pass `None` but we could pass all zeros etc too if that is cleaner for LSP.
> 
> https://github.com/FuelLabs/sway/blob/cd63629167137c2c9d713760cddedcab23da1ec5/forc-pkg/src/pkg.rs#L2592

Nice thats a good idea and I agree seems like a cleaner approach. Putting this PR in draft mode for now and will make this change tomorrow.

kayagokalp:
> > This LGTM 🚀 we can alternatively inject a dummy `CONTRACT_ID` while checking as well, we currently pass `None` but we could pass all zeros etc too if that is cleaner for LSP.
> > https://github.com/FuelLabs/sway/blob/cd63629167137c2c9d713760cddedcab23da1ec5/forc-pkg/src/pkg.rs#L2592
> 
> Nice thats a good idea and I agree seems like a cleaner approach. Putting this PR in draft mode for now and will make this change tomorrow.

👌 let me know if passing `Some(...)` does not add the dummy constant automatically. In my initial implementation I was inserting dummy `CONTRACT_ID`s for checking but I guess that is removed in a refactor later on.

JoshuaBatty:
Should be ready for review now @kayagokalp

kayagokalp (approved):
LGTM :)

forc-pkg/src/pkg.rs
```diff
@@ -2586,16 +2586,30 @@ pub fn check(
...
+        // inserting a dummy value here to avoid false error signals being reported in LSP.
+        // We only do this for the last node in the compilation order because previous nodes
+        // are dependencies.
+        //
+        // See this github issue for more context: https://github.com/FuelLabs/sway-vscode-plugin/issues/154
+        let contract_id_value = if idx == plan.compilation_order.len() - 1 {
+            Some("0x0000000000000000000000000000000000000000000000000000000000000000".to_string())
+        } else {
+            None
+        };
```
eureka-cpu, line 2604: Feel free to ignore since this is really just a style nit but I think this could be written as
```rust
const DUMMY_CONTRACT_ID: &str = "0x0000000000000000000000000000000000000000000000000000000000000000";
let contract_id_value = (idx == plan.compilation_order.len() - 1).then(|| DUMMY_CONTRACT_ID.to_string());
```

eureka-cpu (approved):
Nice one just left one small nit :shipit:

kayagokalp (approved):

PR #4733, opened 2023-07-03
Add module AST cache.

tritao:
## Description

This adds a module AST cache which speeds up the parsing and type checking stages for modules that have not changed.
To make this work, when first parsing, we save the tree of dependencies and hashes for each module.
Then when compiling again, we check the cache and re-use if possible.

To make this work for LSP, I've had to revert the change that was done earlier to re-create the engines for each compilation step.
This may lead to memory bloating, which was the reason that change was done in the first place IIRC, so I'll be investigating a garbage collection step that LSP can run after each compilation.

On the LSP `did_change` test, that compiles a single file, makes a change, and re-parses again, this drops the time from 6s (3s + 3s for the second parse) to 3s, as expected, since the core and standard library are re-used instead of being parsed and analyzed again.

## Checklist

### DIFF
--- a/sway-lsp/src/core/session.rs
+++ b/sway-lsp/src/core/session.rs
@@ -1,7 +1,7 @@
 use crate::{
     capabilities::{
         self,
-        diagnostic::{get_diagnostics, Diagnostics},
+        diagnostic::DiagnosticMap,
         formatting::get_page_text_edit,
         runnable::{Runnable, RunnableMainFn, RunnableTestFn},
     },
@@ -41,6 +41,7 @@ use sway_core::{
     },
     BuildTarget, Engines, Namespace, Programs,
 };
+use sway_error::{error::CompileError, warning::CompileWarning};
 use sway_types::{SourceEngine, Spanned};
 use sway_utils::helpers::get_sway_files;
 use tokio::sync::Semaphore;
@@ -61,7 +62,7 @@ pub struct CompiledProgram {
 /// the types in [Session] after successfully parsing.
 #[derive(Debug)]
 pub struct ParseResult {
-    pub(crate) diagnostics: Diagnostics,
+    pub(crate) diagnostics: (Vec<CompileError>, Vec<CompileWarning>),
     pub(crate) token_map: TokenMap,
     pub(crate) engines: Engines,
     pub(crate) lexed: LexedProgram,
@@ -85,7 +86,7 @@ pub struct Session {
     // and one thread can be waiting to start parsing. All others will return the cached diagnostics.
     pub parse_permits: Arc<Semaphore>,
     // Cached diagnostic results that require a lock to access. Readers will wait for writers to complete.
-    pub diagnostics: Arc<RwLock<Diagnostics>>,
+    pub diagnostics: Arc<RwLock<DiagnosticMap>>,
 }
 
 impl Default for Session {
@@ -104,7 +105,7 @@ impl Session {
             engines: <_>::default(),
             sync: SyncWorkspace::new(),
             parse_permits: Arc::new(Semaphore::new(2)),
-            diagnostics: Arc::new(RwLock::new(Diagnostics::default())),
+            diagnostics: Arc::new(RwLock::new(DiagnosticMap::new())),
         }
     }
 
@@ -139,8 +140,8 @@ impl Session {
         &self.token_map
     }
 
-    /// Wait for the cached [Diagnostics] to be unlocked after parsing and return a copy.
-    pub fn wait_for_parsing(&self) -> Diagnostics {
+    /// Wait for the cached [DiagnosticMap] to be unlocked after parsing and return a copy.
+    pub fn wait_for_parsing(&self) -> DiagnosticMap {
         self.diagnostics.read().clone()
     }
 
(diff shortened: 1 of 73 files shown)


sway-core/src/query_engine/mod.rs
```diff
@@ -1,4 +1,54 @@
...
+#[derive(Clone, Debug)]
+pub struct ProgramsCacheEntry {
+    pub path: ModulePath,
+    pub compile_result: CompileResult<Programs>,
+}
+
+pub type ProgramsCacheMap = HashMap<ModulePath, ProgramsCacheEntry>;
+
 #[derive(Debug, Default)]
-pub struct QueryEngine {}
```
IGI-111, line 2: This is a really nice start, but I think the query engine needs to be a lot more generic over queries.
Every query should be hashed and possibly have dependencies.

JoshuaBatty:
Hey @tritao just pulled down this branch and gave it a quick play with gc turned on. You can see in the video below that we start off with typed information in the language server. But if we introduce some malformed code we lose access to these (see issue #4866). However, if we then remove the line causing the error, we never get the typed tokens back and the language server is unable to work. See below.

https://github.com/FuelLabs/sway/assets/1289413/0ad640a8-c829-4a9f-a80a-cb4e28b9440c

JoshuaBatty:
It seems to work as expected when garbage collection is turned off.

xunilrj:
Would be interesting to see numbers on `cargo r -p test --release` for both CPU and mem.

sway-lsp/src/core/session.rs
```diff
@@ -71,6 +74,8 @@ pub struct Session {
     parse_permits: Arc<Semaphore>,
     // Cached diagnostic results that require a lock to access. Readers will wait for writers to complete.
     diagnostics: Arc<RwLock<Diagnostics>>,
+    // Enables the experimental GC pass, disabled for now.
+    enable_gc: bool,
```
IGI-111, line 78: Since this has no guarantee to work at this point I think we should add "experimental" to the name of the config field so it's clear in the call as well.

IGI-111:
> For `test`, I think we may be able to re-use the engines across tests, but not really sure if we want to do that, given it means each individual test will not have a fresh engines state. Long-term we will need to serialize/deserialize the state for the type and decl engine to be able to speed up compilations, basically have the cache all primed on the first run.

I also think our tests should run on a clean engine set.

Though what this is missing is a validation test that checks if doing subsequent compilations of some sway that uses a bunch of language features still works.

sway-error/src/handler.rs
```diff
@@ -8,13 +8,13 @@ use core::cell::RefCell;
...
     /// The inner handler.
     /// This construction is used to avoid `&mut` all over the compiler.
-    inner: RefCell<HandlerInner>,
+    inner: RefCell<HandlerData>,
 }
 
 /// Contains the actual data for `Handler`.
 /// Modelled this way to afford an API using interior mutability.
 #[derive(Default, Debug, Clone)]
-struct HandlerInner {
```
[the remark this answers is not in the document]
IGI-111: Yeah I'm just looking for consistency here.

IGI-111 (approved):

sway-core/src/concurrent_slab.rs
```diff
@@ -33,6 +28,47 @@ impl<T> ConcurrentSlab<T> {
...
+        gc.resize(inner.len(), false);
+    }
+
+    pub fn mark_id(&self, index: usize) {
+        let mut gc = self.gc.write().unwrap();
+        gc.insert(index, true);
+    }
+
+    pub fn sweep(&self) {
+        // TODO: for properly implementing sweep we need to change the representation
```
JoshuaBatty, line 45: Perhaps we should open an issue for this and link it here?

sway-core/src/lib.rs
```diff
@@ -200,12 +217,16 @@ fn parse_in_memory(
...
-    lexed: Vec<(Ident, lexed::LexedSubmodule)>,
-    parsed: Vec<(Ident, parsed::ParseSubmodule)>,
+pub struct Submodule {
+    name: Ident,
+    path: ModulePath,
+    lexed: lexed::LexedSubmodule,
+    parsed: parsed::ParseSubmodule,
 }
 
+/// Contains the lexed and parsed submodules 'deps' of a module.
```
JoshuaBatty, line 227: Just out of interest, what's the reason for not also including the typed submodules?

JoshuaBatty (approved):
Thanks @tritao looks like a good first step. Happy to help with #4988 also.

JoshuaBatty (approved):

xunilrj:
Makes sense to move all the GC changes to https://github.com/FuelLabs/sway/issues/4994? I am still confused on how we are going to implement sweep, or if we are going to go with one-slab-per-module and drop it entirely; or even something else.

IGI-111:
This PR has gone for a long time already. I'd say we merge these changes to a stable state and move to the one slab per module approach in a later change.

sway-core/src/language/parsed/module.rs
```diff
@@ -18,6 +20,8 @@ pub struct ParseModule {
     pub module_kind_span: Span,
     /// an empty span at the beginning of the file containing the module
     pub span: Span,
+    /// an hash used for caching the module
```
JoshuaBatty, line 23:
```suggestion
    /// hash used for caching the module
```

JoshuaBatty:
> I just took out the GC changes. We can figure out the best approach in the meanwhile, but am thinking just going to the slab-per-module approach may be the way to go.

Sounds straight forward enough to manage from the LSP side with this approach.

JoshuaBatty (approved):

PR #4736, opened 2023-07-03
Doc storage read unwrap_or clarification.

SilentCicero:
## Description
Document clarification around ensuring .unwrap_or(0) is used for storage read examples.

### DIFF
--- a/examples/storage_variables/src/main.sw
+++ b/examples/storage_variables/src/main.sw
@@ -42,10 +42,10 @@ impl StorageExample for Contract {
     #[storage(read)]
     fn get_something() -> (u64, u64, b256, bool) {
         (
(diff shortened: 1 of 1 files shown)


kayagokalp:
@JoshuaBatty @sdankel this PR seems like failing with an LSP test. Does the failing test makes sense?

JoshuaBatty:
Will look into why this is failing now.

JoshuaBatty:
seems to be working in CI and again running locally now that it's been merged with master. I suspect the fix with just landed with fixing the inlay hints request was what was causing this to fail.

JoshuaBatty (approved):

sdankel (approved):

Issue #4708, opened 2023-06-27
constant expressions evaluator should handle local let-expressions


anton-trunov:
For instance, 
```rust
script;

fn main() -> bool {
    let result: u64 = 0 + 1 + 2 + 3 + 4 + 5;

    log(result);

    true
}
```

currently produces the following IR:

```
script {
    entry fn main() -> bool, !1 {
        entry():
        v0 = const u64 0, !2
        v1 = const u64 1, !3
        v2 = add v0, v1, !4
        v3 = const u64 2, !5
        v4 = add v2, v3, !6
        v5 = const u64 3, !7
        v6 = add v4, v5, !8
        v7 = const u64 4, !9
        v8 = add v6, v7, !10
        v9 = const u64 5, !11
        v10 = add v8, v9, !12
        v11 = const u64 0
        log u64 v10, v11, !16
        v12 = const bool true, !17
        ret bool v12
    }
}
```

but one would expect no `add` instructions for the script shown above.

xunilrj:
I believe I found some limitations in the "const-eval" module. I will create a PR to address them first.

xunilrj:
@anton-trunov, I was unsure if changing the IR generation would be beneficial in this case. So I improved the `constcombine` optimization pass. See https://github.com/FuelLabs/sway/pull/4780

Even if you think the IR generation is a must, I would merge https://github.com/FuelLabs/sway/pull/4780.

Any thoughts?

xunilrj:
The optimizer pass solution is done. I will experiment with `const-eval`-uating expression before IR generation.

PR #4761, opened 2023-07-06
Fix: const_eval bugs

xunilrj:
## Description

This PR is necessary for https://github.com/FuelLabs/sway/issues/4708

It fixes some problems with `const_eval`, most importantly:

1 - It differentiates when an eval cannot be used inside const (there is a new error for that), and when expressions/statements  just do not produce any value;
2 - It always cleans up `eval_codeblock` (it never shortcircuit with `?`);

Another advantage of the explicit error is the exact span of the "offending" expression, which we can use in the future for better error messages.

## Checklist


sway-core/src/ir_generation/const_eval.rs
```diff
@@ -491,15 +548,15 @@ fn const_eval_typed_expr(
             then,
             r#else,
         } => {
-            match const_eval_typed_expr(lookup, known_consts, condition)? {
+            match dbg!(const_eval_typed_expr(lookup, known_consts, condition))? {
```
IGI-111, line 551: Are those uses of `dbg!` leftovers or intended?

xunilrj: Argh! Fixed.

sway-core/src/ir_generation/const_eval.rs
```diff
@@ -188,7 +197,14 @@ pub(super) fn compile_constant_expression(
         module_ns,
         function_compiler,
         const_expr,
-    )?;
+    )
+    .map_err(|err| match err {
+        CompileError::CannotBeEvaluatedToConst { .. } => CompileError::NonConstantDeclValue {
+            span: call_path.span(),
```
anton-trunov, line 203: This gives an impression these two errors are actually the same. Am I missing something? (Usually we `map_err` to lift e.g. parser error into a compiler error or something like that).

anton-trunov:
It would be great to have failing end-to-end tests with e.g. `return` and state-modifying operators somewhere in this directory: https://github.com/FuelLabs/sway/tree/master/test/src/e2e_vm_tests/test_programs/should_fail

sway-core/src/ir_generation/const_eval.rs
xunilrj: I thought about making `CannotBeEvaluatedToConst` local to `const_eval.rs`. I can make it this way if you think more sense. For example:

```rust
enum ConstEvalError {
    CompileError(CompileError),
    CannotBeEvaluatedToConst { span: Span}
}
```

anton-trunov: I think this approach with local errors specific to const evaluation is better, given that it can be extended to more informative error messages in the future, i.e. we could explain better why some initializer fails to be evaluated

xunilrj: Done.

xunilrj:
> It would be great to have failing end-to-end tests with e.g. `return` and state-modifying operators somewhere in this directory: https://github.com/FuelLabs/sway/tree/master/test/src/e2e_vm_tests/test_programs/should_fail

Done.

anton-trunov (approved):

IGI-111 (approved):

PR #4793, opened 2023-07-12
Improve flag library functionality

### LINKED ISSUE #4355: The `U128` library doesn't preserve overflow flags.
The implementation of the `U128` arithmetic library has to disable panic on overflow e.g., for addition, so that it may propagate the carry bit up to the higher word.  It currently will always enable panic on overflow when it is done.

If the user has explicitly turned panics off they will be turned on whether they want them or not.  A better solution is to copy the flags before the `U128` operation and restoring them afterwards.

SwayStar123:
## Description
- Changes `disable_panic_on_overflow()` function to return the flags set prior to calling the function
- Adds a `set_flags` function that simply sets the given value to the flag register
- Changed u64 `overflowing_add` and `overflowing_mul` to use the new functions
- Adds equivalent functions for unsafe-math flag aswell
- Additional documentation for all functions in the library

closes #4355 

## Checklist

### DIFF
--- a/sway-lib-std/src/flags.sw
+++ b/sway-lib-std/src/flags.sw
@@ -264,6 +264,7 @@ fn test_disable_panic_on_unsafe_math_preserving() {
 
     let prior_flags = disable_panic_on_unsafe_math_preserving();
     let _bar = 1 / 0;
+    ::logging::log(error());
     assert(error() == 1);
     set_flags(prior_flags);
 
--- a/sway-lib-std/src/flags.sw
+++ b/sway-lib-std/src/flags.sw
@@ -267,7 +267,7 @@ fn test_disable_panic_on_unsafe_math() {
 #[test]
 fn test_disable_panic_on_unsafe_math_preserving() {
     disable_panic_on_unsafe_math();
-    log(error())
+    log(error());
 
     let prior_flags = disable_panic_on_unsafe_math_preserving();
     let _bar = 1 / 0;
--- a/sway-lib-std/src/flags.sw
+++ b/sway-lib-std/src/flags.sw
@@ -274,6 +274,7 @@ fn test_disable_panic_on_unsafe_math_preserving() {
         r1: u64
     };
     log(error());
+    log(error());
     assert(error() == 1);
     set_flags(prior_flags);
 
--- a/sway-lib-std/src/flags.sw
+++ b/sway-lib-std/src/flags.sw
@@ -138,7 +138,7 @@ fn test_disable_panic_on_overflow_preserving() {
     disable_panic_on_overflow();
 
     let prior_flags = disable_panic_on_overflow_preserving();
-    let _bar = u64::max() + 1;
+    let mut _bar = u64::max() + 1;
     set_flags(prior_flags);
 
     _bar = u64::max() + 1;
--- a/sway-lib-std/src/flags.sw
+++ b/sway-lib-std/src/flags.sw
@@ -267,10 +267,11 @@ fn test_disable_panic_on_unsafe_math() {
 #[test]
 fn test_disable_panic_on_unsafe_math_preserving() {
     disable_panic_on_unsafe_math();
-    log(error());
 
     let prior_flags = disable_panic_on_unsafe_math_preserving();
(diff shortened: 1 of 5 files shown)


sway-lib-std/src/flags.sw
```diff
@@ -30,3 +89,196 @@ pub fn enable_panic_on_overflow() {
...
     }
 }
+
+/// Allows overflowing operations to occur without a FuelVM panic.
+/// More suitable than `disable_panic_on_overflow` for use in functions that are not the entry point of a program.
+///
+/// > **_WARNING:_**
+/// >
+/// > Don't forget to call `set_flags` after performing the operations for which you disabled the default `panic-on-overflow` behavior in the first place!
+///
```
bitzoic, line 99:
```suggestion
/// ### Returns
/// 
/// * [u64] - The flag prior to disabling panic on overflow.
```

sway-lib-std/src/flags.sw
```diff
@@ -30,3 +89,196 @@ pub fn enable_panic_on_overflow() {
...
+/// ### Examples
+///
+/// ```sway
+/// use std::{assert::assert, flags::{disable_panic_on_unsafe_math, enable_panic_on_unsafe_math}, registers::error};
+///
+/// fn main() {
+///     disable_panic_on_unsafe_math();
+///      
+///     let bar = 1 / 0; // Division by zero is considered unsafe math.
```
bitzoic, line 144:
```suggestion
///     // Division by zero is considered unsafe math.
///     let bar = 1 / 0; 
```
This isn't really necessary but bothered me sorry. Same thing is done one line bellow..

sway-lib-std/src/flags.sw
```diff
@@ -30,3 +89,196 @@ pub fn enable_panic_on_overflow() {
...
+///     assert(error() == 1); // Error flag is set to true whenever unsafe math occurs.
+///
+///     set_flags(prior_flags);
+/// }
+/// ```
+pub fn disable_panic_on_unsafe_math_preserving() -> u64 {
+    let prior_flags = flags();
+
+    // Mask first bit, which is `F_UNSAFEMATH`.
```
bitzoic, line 218: I think it would be worthwhile to set these to a `F_UNSAFEMATH` constant instead of defining them in the function itself

sway-lib-std/src/flags.sw
```diff
@@ -30,3 +89,196 @@ pub fn enable_panic_on_overflow() {
...
+
+#[test]
+fn test_disable_panic_on_overflow() {
+    disable_panic_on_overflow();
+    let _bar = u64::max() + 1;
+    enable_panic_on_overflow();
+}
+
+#[test]
+fn test_disable_panic_on_overflow_preserving() {
```
bitzoic, line 239: I like how this test was handled despite not having any `should_panic` attributes. Nicely done

sway-lib-std/src/flags.sw
```diff
@@ -30,3 +89,196 @@ pub fn enable_panic_on_overflow() {
...
+    set_flags(prior_flags);
+
+    let _bar = asm(r2: 1, r3: 0, r1) {
+        div r1 r2 r3;
+        r1: u64
+    };
+    assert(error() == 1);
+
+    enable_panic_on_unsafe_math();
+}
```
bitzoic, line 284:
```suggestion
}

```

sway-lib-std/src/flags.sw
```diff
@@ -1,32 +1,293 @@
...
+///      
+///     // Division by zero is considered unsafe math.
+///     let bar = 1 / 0; 
+///     // Error flag is set to true whenever unsafe math occurs. Here represented as 1.
+///     assert(error() == 1); 
+///
+///     set_flags(prior_flags);
+/// }
+/// ```
+pub fn disable_panic_on_unsafe_math_preserving() -> u64 {
```
xunilrj, line 227: Why the "*_preserving" variants are needed? Can´t we just always do this and ignore the result when not needed?

vaivaswatha (approved):

bitzoic (approved):

PR #4802, opened 2023-07-15
stdlib: remove `decompose` function for `u256`

anton-trunov:
The `From` trait already lets one to decompose a `u256` into four 64-bit words. Also, there is no `compose` for `u256` (unlike `b256`, which has both at the moment).

## Checklist

### DIFF
--- a/sway-lib-std/src/u256.sw
+++ b/sway-lib-std/src/u256.sw
@@ -190,26 +190,6 @@ impl U256 {
     pub fn bits() -> u32 {
         256
     }
-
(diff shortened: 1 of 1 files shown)


Braqzen (approved):
I don't believe this is a breaking change since the method is private and can only be used inside this file otherwise seems good to me.

anton-trunov:
Ah, good catch, didn't notice it lacks the `pub` keyword

IGI-111 (approved):

Issue #4709, opened 2023-06-27
Lots of warnings on `forc init` -> `forc test`, considering cleanup


SilentCicero:
I'm getting a huge set of warnings when I run `forc init`, then `forc test` out of the box.

The default example and test should not have all these warnings and repeated warning statements. Default examples should be always up to date and not produce warnings.

```sh
  WARNING! unused manifest key: project.target
  WARNING! unused manifest key: project.target
  WARNING! unused manifest key: project.target
  WARNING! unused manifest key: project.target
  WARNING! unused manifest key: project.target
  WARNING! unused manifest key: project.target
  WARNING! unused manifest key: constants
  WARNING! unused manifest key: constants
  Creating a new `Forc.lock` file. (Cause: lock file did not exist)
    Adding core
    Adding std git+https://github.com/fuellabs/sway?tag=v0.39.0#7a095280e7e1c0bfbbfc201867896e720aab7209
   Created new lock file at /fuel-oracle/src/contracts/Forc.lock
  WARNING! unused manifest key: project.target
  WARNING! unused manifest key: project.target
  WARNING! unused manifest key: project.target
  WARNING! unused manifest key: project.target
  WARNING! unused manifest key: project.target
  WARNING! unused manifest key: project.target
  WARNING! unused manifest key: constants
  WARNING! unused manifest key: constants
 Compiling library core (.forc/git/checkouts/std-9be0d6062747ea7/7a095280e7e1c0bfbbfc201867896e720aab7209/sway-lib-core)
warning
   --> .forc/git/checkouts/std-9be0d6062747ea7/7a095280e7e1c0bfbbfc201867896e720aab7209/sway-lib-core/src/ops.sw:485:21
    |
483 | 
484 |     fn lsh(self, other: u64) -> Self {
485 |         __lsh(self, other)
    |                     ----- This cast, from integer type of width sixty four to integer type of width thirty two, will lose precision.
486 |     }
487 |     fn rsh(self, other: u64) -> Self {
    |
____

warning
   --> /.forc/git/checkouts/std-9be0d6062747ea7/7a095280e7e1c0bfbbfc201867896e720aab7209/sway-lib-core/src/ops.sw:488:21
    |
486 | 
487 |     fn rsh(self, other: u64) -> Self {
488 |         __rsh(self, other)
    |                     ----- This cast, from integer type of width sixty four to integer type of width thirty two, will lose precision.
489 |     }
490 | }
    |
____

warning
[...]

PR #4808, opened 2023-07-17
chore: add forc-pkg codeowner

kayagokalp:
## Description

We will be fixing #4709, but the one thing more important than fixing it might be ensuring that it does not happen again. Until we have the bandwidth to introduce nice CLI testing suite I am adding myself as code owner to forc-pkg. Since forc-pkg lives between the compiler and forc, it is seeing lots of traffic and I will try to make sure nothing changes in terms of user experience during those changes. Feel free to close this, if you feel like we might have a better way of ensuring this before we got the CLI testing suite ready.


JoshuaBatty (approved):
Sounds like a good plan to me.

anton-trunov (approved):

eureka-cpu (approved):

Issue #4799, opened 2023-07-14
Harmonize error messages for multiple definitions of names


ironcev:
There are several error messages that communicate duplication of symbols/names. Some of them overlap in a confusing or redundant way.

## `MultipleDefinitionsOfName` and `NameDefinedMultipleTimes`

At the moment, there are two compiler errors, [MultipleDefinitionsOfName](https://github.com/FuelLabs/sway/blob/13ed81b6bf0b0d1c21b67ed385b8d246d2402f0a/sway-error/src/error.rs#L71) and [NameDefinedMultipleTimes](https://github.com/FuelLabs/sway/blob/13ed81b6bf0b0d1c21b67ed385b8d246d2402f0a/sway-error/src/error.rs#L641), that print two very similar errors, respectively:

```
Name \"{name}\" is defined multiple times.
The name `{name}` is defined multiple times
```

This looks like an accidental redundancy and not an intentional choice.

## Different errors for multiple definition of constants

There is an additional error specific to constants - [MultipleDefinitionsOfConstants](https://github.com/FuelLabs/sway/blob/13ed81b6bf0b0d1c21b67ed385b8d246d2402f0a/sway-error/src/error.rs#L73)

```
Constant \"{name}\" was already defined in scope.
```

Depending on the context, we can get any of the three errors listed above:

```
const X = 1;
const X = 2; // The name `X` is defined multiple times
```
```
struct S {} // NOTE: Same for enums.
impl S {
    const S_X = 1;
    const S_X = 2; // The name `S_X` is defined multiple times
}
```
```
trait T { // NOTE: Same for ABIs
    const T_X: u32;
    const T_X: u32; // Name "T_X" is defined multiple times.
}
```
```
trait TR {
    const T_X: u32;
}

struct ST {}

impl TR for ST {
    const T_X: u32 = 1;
    const T_X: u32 = 2; // Constant "T_X" was already defined in scope.
}
```

## Specific errors for specific cases of duplicated names

E.g.:

```
struct S {
    s: u32,
    s: u32, // struct field "s" already declared
}
```
```
enum E {
    e: (),
    e: (), // enum variant "e" already declared
}
```
```
fn f() {}
fn f() {} // Function "f" was already defined in scope.
```
```
use lib::LibStruct as LS1;
use lib::LibEnum as LS1; // The name "LS1" shadows another symbol with the same name.
```

## Proposal

In general, use as specific error as practical and applicable and have a single fallback one for more general cases.
This would mean:

[...]

Issue #4587, opened 2023-05-24
Give more helpful error when shadowing `const`


IGI-111:
Shadowing constants is forbidden, but the compiler will currently give an unhelpful error message like so:
 ```
 The name `VAL` is defined multiple times
 ```
We should give a clearer error message in that particular case.

deepak-likes-code:
I can pick this up

ironcev:
Hi, this looks like a _great_ first issue, not only a good one :-) I would love to contribute to it. I have two questions, before starting with the implementation. This is the TL;DR version of the questions. I will also provide a longer one in the comment below. The longer one is more for me, and as a checklist once the feature is implemented. I've started diving deeper into Sway to get a better feeling for the language and the design decisions already done. So I systematically piled up all the cases realted to the shadowing `const`s.
 
## TL;DR

(The terms _item-style shadowing_ and _sequential shadowing_ have the meaning defined in the [Rust RFC 116 (No module shadowing)](https://rust-lang.github.io/rfcs/0116-no-module-shadowing.html).)

:question: Should the new message apply only to _sequential shadowing_, and for the _item-style shadowing_ we keep the existing one?

This is how I understand the motivation behind this issue - to give hint to the programmer that unlike variables `const`s cannot be shadowed on the same or nested scope level.

Having a separate error message for `const` in case of item-style shadowing makes no sense for me, because I don't see why would `const` be an exceptional item and deserve an error message different then other items.

Applied to the [shadowed consts and vars tests](https://github.com/FuelLabs/sway/blob/ac4ff14482471804d80faffe659333f10fa421e5/test/src/e2e_vm_tests/test_programs/should_fail/shadowing/shadowed_consts_and_vars/src/main.sw) this would mean the following test output:

    // const shadowing an imported const
    use lib::X;
    const X = 6; // <existing error> The name `X` is defined multiple times
 
    // const shadowing a local const
    const Y = 7;
    const Y = 8; // <existing error> The name `X` is defined multiple times
    fn main() {
        // variable shadowing an imported const
        let X = 9; // <new error>
 
        // variable shadowing a const
        const Z = 3;
        let Z = 4; // <new error>
 
        // const shadowing a variable
        let W = 2;
        const W = 1; // <new error>
 
        // Variable shadowing a variable - this is okay!
        let P = 7;
        let P = 8;
    }

The detailed reasoning is in the comment below.

:question: How should the new error message exactly look like?                                                                                               

[...]

PR #4819, opened 2023-07-18
Give more helpful error when shadowing constants

ironcev:
## Description

- Introduce three distinguished new error messages on const shadowing as defined in #4587:
  - Variables shadowing constants.
  - Constants shadowing variables.
  - Constants shadowing constants, _item-style_ and _sequentially_.
- _Item-style_ and _sequential shadowing_ are added as an explicit concept to the `TypeCheckContext`. This follows the same pattern that we have in passing _ABI mode_ and _disallow functions_ contextual information. Basically, the parts of the semantic analysis lower in the chain need contextual information available and deducible in the upper parts of the chain. In this case, the deduction is simple - as soon as we are within a function body, we apply _sequential shadowing_. (Note that this simple approach will not work if we ever introduce types local to functions. But this is not an issues. In that case we will need to redefine the const shadowing within functions in general, and also by that time I assume we will have the graph collections which will allow us different approach :-)
- Clean up of redundant error messages as defined in  #4799.

## Checklist

### DIFF
--- a/docs/book/src/basics/constants.md
+++ b/docs/book/src/basics/constants.md
@@ -4,7 +4,7 @@
 <!-- constants:example:start -->
 Constants are similar to variables; however, there are a few differences:
 
-- Constants are always evaluated at compile-time
+- Constants are always evaluated at compile-time.
 - Constants can be declared both inside of a [function](../index.md) and at global / `impl` scope.
 - The `mut` keyword cannot be used with constants.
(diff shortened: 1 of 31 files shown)


IGI-111 (approved):
LGTM

It comes to mind that even this small issue looks like yet more reason to have an explicit step for collecting type checking context.

JoshuaBatty (approved):

PR #4821, opened 2023-07-18
An upgrade to `fuel-core` v0.20.3

xgreenx:
This pull request is an upgrade to a `fuel-core 0.20.1` release.

The biggest changes coming from this new release is a rework of how AssetId is structured to support [granular asset minting](https://github.com/FuelLabs/fuel-specs/pull/491). This means a contract can mint and burn multiple AssetId's, and the ContractId != AssetId.


forc-plugins/forc-client/src/util/encode.rs
```diff
@@ -1,47 +1,11 @@
 use std::str::FromStr;
 
 use fuel_abi_types::abi::full_program::FullTypeApplication;
-use serde::{Deserialize, Deserializer, Serialize};
 
 /// A wrapper around fuels_core::types::Token, which enables serde de/serialization.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, serde::Serialize, serde::Deserialize)]
 pub(crate) struct Token(pub(crate) fuels_core::types::Token);
```
kayagokalp, line 8: I think we can drop the wrapper `Token` since underlying `Token` now implements `Serialize` and `Deserialize`. Just leaving this as a comment, no need to do that here i can tackle that later on 👍

kayagokalp (approved):
LGTM

bitzoic (approved):

Cargo.toml
```diff
@@ -31,18 +31,36 @@ exclude = [
...
 
 [workspace.dependencies]
-fuel-asm = "0.34.1"
-fuel-crypto = "0.34.1"
-fuel-types = "0.34.1"
-fuel-tx = "0.34.1"
-fuel-core-client = { version = "0.19.1", default-features = false }
-fuel-vm = "0.34.1"
+# Depednencies from the `fuel-core` repository:
+fuel-core-client = { version = "0.20.2", default-features = false }
```
Voxelot, line 35: Should we bump this to 0.20.3?

sdankel (approved):

sway-lib-std/src/alias.sw
```diff
@@ -0,0 +1,6 @@
+library;
+
+/// The `AssetId` type is simply an alias for `b256` that represents the ID of a native asset.
+pub type AssetId = b256;
```
Voxelot, line 4: I'm fine with this for now, but we should create a real newtype for AssetId similar to how ContractId is setup in a future release (which is basically also just a b256). This will likely break the SDK though, so it won't be a trivial change.

bitzoic: A defined `AssetId` type would be beneficial. For now, I've added this to distinguish between when AssetId and SubId are needed in token operations.

sway-lib-std/src/asset_id.sw
```diff
@@ -0,0 +1,17 @@
+//! Helper functions for generating an AssetId
+library;
+
+use ::alias::{AssetId, SubId};
+use ::constants::ZERO_B256;
+use ::contract_id::ContractId;
+use ::hash::sha256;
+
+/// Construct an AssetId from a ContractId and SubId by hashing them together with sha256.
+pub fn construct_asset_id(contract_id: ContractId, sub_id: SubId) -> AssetId {
```
bitzoic, line 10: This could be used in some of the functions within `token.sw` file but that can be done in a future PR.

bitzoic (approved):

kayagokalp (approved):
tooling related changes: utACK

PR #4829, opened 2023-07-19
Numeric type inference

### LINKED ISSUE #3470: Removing implicit casting between integer types
Currently, the compiler allows implicit casting between integers. For example, you can assign a `u32` to a `u64` and vice versa. The compiler also emits a warning for a downcast.

@anton-trunov recently suggested that we should follow Rust and disallow implicit casting altogether. Writing `let x:u8 = 5u64` works but is pretty weird. Besides, implicit casts often generate warnings that can't be removed and they can lead to all kinds of problems 

Instead, we would implement `into()` and `from()` between all integer types so that we can write:
```rust
 let x = 5u16;
 let y:u64 = u64::from(x);
 let y:u64 = x.into(); 
```

Now one challenge is that if we remove implicit casts from the type engine today, the following would stop compiling:
```rust
let x:u8 = 1u8 + 1; // or even let x:u8 = 1 + 1;
```
In both cases, the type of `1` starts out as `Numeric` and later because `u64` because that's the default. Rust handles this case well so we may want to look at how that's done.

IGI-111:
## Description

This does two things.

1. Integer types no longer automatically cast to each other. Methods such as `as_u64` and `try_as_u8` have been introduced to require the user to explicit the cast.
Warnings about casts losing precision have been removed as they are no longer possible.

2. Numeric literals (such as `32` without further qualifications) no longer immediately decay to a `u64` and we instead attempt to type them as `TypeInfo::Numeric` as far as possible to allow their usage with any integer type.
Numeric intrinsics unify arguments with the numeric type to enforce their constraints.
Numeric values are forced to decay to `u64` when resolving methods or traits.

This change is a prerequisite to move our representation of sub-byte-sized values to a different memory representation, as numeric types need to be distinct if we are ever to treat them differently.

Fixes #3470

## Checklist

### DIFF
--- a/sway-ir/src/irtype.rs
+++ b/sway-ir/src/irtype.rs
@@ -50,7 +50,6 @@ impl Type {
         Self::get_or_create_unique_type(context, TypeContent::Unit);
         Self::get_or_create_unique_type(context, TypeContent::Bool);
         Self::get_or_create_unique_type(context, TypeContent::Uint(8));
-        Self::get_or_create_unique_type(context, TypeContent::Uint(32));
         Self::get_or_create_unique_type(context, TypeContent::Uint(64));
         Self::get_or_create_unique_type(context, TypeContent::B256);
         Self::get_or_create_unique_type(context, TypeContent::Slice);
@@ -79,11 +78,6 @@ impl Type {
     /// New u8 type
     pub fn get_uint8(context: &Context) -> Type {
         Self::get_type(context, &TypeContent::Uint(8)).expect("create_basic_types not called")
-    }
-
-    /// New u32 type
-    pub fn get_uint32(context: &Context) -> Type {
-        Self::get_type(context, &TypeContent::Uint(32)).expect("create_basic_types not called")
     }
 
     /// New u64 type
--- a/test/src/e2e_vm_tests/test_programs/should_fail/binop_intrinsics/test.toml
+++ b/test/src/e2e_vm_tests/test_programs/should_fail/binop_intrinsics/test.toml
@@ -2,20 +2,67 @@ category = "fail"
 
 # check: $()error
 # check: $()__add(A { a: 32 }, 32);
-# nextln: $()Unsupported argument type to intrinsic "add"
+# nextln: $()Mismatched types
+# nextln: $()expected: numeric
+# nextln: $()found:    A.
+# nextln: $()help: Incorrect argument type
 
 # check: $()error
 # check: $()__add("Hello", 22);
-# nextln: $()Unsupported argument type to intrinsic "add"
+# nextln: $()Mismatched types
+# nextln: $()expected: numeric
+# nextln: $()found:    str[5].
+# nextln: $()help: Incorrect argument type
 
 # check: $()error
 # check: $()__add("Hello", "Hello")
-# nextln: $()Unsupported argument type to intrinsic "add"
+# nextln: $()Mismatched types
+# nextln: $()expected: numeric
+# nextln: $()found:    str[5].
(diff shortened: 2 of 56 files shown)


test/src/e2e_vm_tests/test_programs/should_fail/binop_intrinsics/test.toml
```diff
@@ -2,19 +2,53 @@ category = "fail"
...
+# check: $()__add(A { a: 32 }, 32);
+# nextln: $()Mismatched types
+# nextln: $()expected: numeric
+# nextln: $()found:    A.
+
+# check: $()error
+# check: $()__add("Hello", 22);
+# nextln: $()Mismatched types
+# nextln: $()expected: str[5]
+# nextln: $()found:    numeric.
```
anton-trunov, line 20: This looks confusing: `__add` should not expect a string type, it only supports numeric types according to the docs

test/src/e2e_vm_tests/test_programs/should_pass/language/numeric_constants/src/main.sw
```diff
@@ -1,18 +1,19 @@
 script;
 
 fn main() -> bool {
-    assert(u64::max() == 18446744073709551615);
+    assert(u64::max() == 18446744073709551615u64);
```
anton-trunov, line 4: Will the previous version (without the `u64` suffix) still work?

test/src/sdk-harness/test_projects/generics_in_abi/src/main.sw
```diff
@@ -79,7 +79,7 @@ impl MyContract for Contract {
     }
 
     fn struct_w_generic_in_tuple(arg1: StructWTupleGeneric<u32>) -> StructWTupleGeneric<u32> {
-        let expected = StructWTupleGeneric { a: (1, 2) };
+        let expected = StructWTupleGeneric { a: (1u32, 2u32) };
```
anton-trunov, line 82: Same question as above: as a Sway user I would prefer to write `(1, 2)` instead of `(1u32, 2u32)` and rely on the compiler to infer the type of the numeric literals

IGI-111: It won't work in this case because it's ambiguous.

The way our type checking works right now, the type of `expected` in `let expected = StructWTupleGeneric { a: (1, 2) };` has to be resolved to a concrete type, so it can't be `StructWTupleGeneric<numeric>` because that's not something that can be impl'd for. Hence we decay the numerics in type arguments to `u64`.

Had we a more powerful type checking implementation we could probably do better, but such changes are too large to be in scope for this I think.

anton-trunov: Got it, thanks. Probably would be a good idea to open an issue on things like this after the PR is merged

sway-ir/src/irtype.rs
```diff
@@ -81,11 +80,6 @@ impl Type {
         Self::get_type(context, &TypeContent::Uint(8)).expect("create_basic_types not called")
     }
 
-    /// New u32 type
-    pub fn get_uint32(context: &Context) -> Type {
-        Self::get_type(context, &TypeContent::Uint(32)).expect("create_basic_types not called")
-    }
-
```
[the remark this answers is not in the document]
IGI-111: We will when I get around to changing the memory representation to use byte operations for small types.

anton-trunov (approved):

JoshuaBatty (approved):

PR #4847, opened 2023-07-21
Correct usage of terms "ABI" and "JSON ABI"

AlicanC:
This PR clears up usage of terms "ABI" (our actual ABI) and "JSON ABI" (the JSON representation of our ABI).

What we really actually have here is the "Rust ABI" but we can drop the "Rust" and just call it "ABI".

Other ABIs would be the "TypeScript ABI" generated by fuels-ts, and the "GraphQL ABI" I'm planning to work on.


kayagokalp (approved):
Thanks for the PR! forc-pkg changes seems good to me.

PR #4848, opened 2023-07-21
A-B library documentation

SwayStar123:
## Description
Documentation for library auth, b256, b512, block and bytes

## Checklist

### DIFF
--- a/sway-lib-std/src/block.sw
+++ b/sway-lib-std/src/block.sw
@@ -6,11 +6,28 @@ use ::constants::ZERO_B256;
 use ::result::Result::{self, *};
 use ::logging::log;
 
+/// Error type for when the block hash cannot be found.
 enum BlockHashError {
+    /// Error returned when the block hash cannot be found.
     BlockHeightTooHigh: (),
 }
 
 /// Get the current block height.
+///
+/// # Returns
+///
+/// * [u32] - The current block height.
+///
+/// # Examples
+///
+/// ```sway
+/// use std::block::height;
+///
+/// fn foo() {
+///     let current_height = height();
+///     log(current_height);   
+/// }
+/// ```
 pub fn height() -> u32 {
     asm(height) {
         bhei height;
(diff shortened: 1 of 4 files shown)


sway-lib-std/src/block.sw
```diff
@@ -19,11 +36,57 @@ pub fn height() -> u32 {
 }
 
 /// Get the TAI64 timestamp of the current block.
+///
+/// # Additional Information
+///
+/// The TAI64 timestamp begins at 2^62 seconds before 1970, and ends at 2^62 seconds after 1970,
+/// with a TAI second defined as the duration of 9192631770 periods of the radiation corresponding
+/// to the transition between the two hyperfine levels of the ground state of the cesium atom.
```
bitzoic, line 44: woah

sway-lib-std/src/bytes.sw
```diff
@@ -194,36 +238,38 @@ impl Bytes {
 
     /// Updates an element at position `index` with a new element `value`.
     ///
-    /// ### Arguments
+    /// # Arguments
     ///
-    /// * index - The index of the element to be set
-    /// * value - The value of the element to be set
+    /// * index: [u64] - The index of the element to be set
```
bitzoic, line 243:
```suggestion
    /// * `index`: [u64] - The index of the element to be set
```
These should have ticks, there's a few more below this one as well

sway-lib-std/src/bytes.sw
```diff
@@ -424,34 +517,46 @@ impl Bytes {
 
     /// Returns `true` if the vector contains no elements.
     ///
-    /// ### Examples
+    /// # Returns
+    ///
+    /// * [bool] - `true` if the vector contains no elements, `false` otherwise
```
bitzoic, line 522:
```suggestion
    /// * [bool] - `true` if the type contains no elements, `false` otherwise
```

sway-lib-std/src/bytes.sw
```diff
@@ -424,34 +517,46 @@ impl Bytes {
 
     /// Returns `true` if the vector contains no elements.
```
bitzoic, line 518: Looks like this is wrong too

sway-lib-std/src/bytes.sw
```diff
@@ -80,22 +90,34 @@ impl Bytes {
 
     /// Constructs a new, empty `Bytes` with the specified capacity.
     ///
+    /// # Additional Information
+    ///
     /// The `Bytes` will be able to hold exactly `capacity` bytes without
     /// reallocating. If `capacity` is zero, the `Bytes` will not allocate.
     ///
     /// It is important to note that although the returned `Bytes` has the
     /// capacity specified, the vector will have a zero length.
```
bitzoic, line 99: Mention of vector here as well

bitzoic (commented):
Very minor nit but could you add periods to the end of the sentences? The descriptions and additional information have them but the arguments and returns don't

sway-lib-std/src/block.sw
SwayStar123: is it fine?

bitzoic: I just had no idea that this is how time was defined.

I think including the timestamp spec is useful, so mentioning the TAI64 spec, but instead of writing out the TAI64 definition perhaps including an external link to where TAI64 is defined would be better.

Good thinking about including this!

SwayStar123: I could not find any place to link which had a simple definition, so i tried to describe it shortly myself

bitzoic (approved):

Issue #4812, opened 2023-07-18
Update standard library and core library documentation to follow inline documentation standard


SwayStar123:
Inline documentation standard can be seen here: https://github.com/FuelLabs/sway-standards/tree/master/standards/src_2

Core library:
- [x] never.sw
- [x] ops.sw
- [x] primitive_conversions.sw
- [x] primitives.sw
- [x] raw_ptr.sw
- [x] raw_slice.sw
- [x] storage.sw

Standard library:
- [x] storage/storable_slice.sw
- [x] storage/storage_api.sw
- [x] storage/storage_bytes.sw
- [x] storage/storage_key.sw
- [x] storage/storage_map.sw
- [x] storage/storage_string.sw
- [x] storage/storage_vec.sw
- [x] vm/evm/ecr.sw
- [x] vm/evm/evm_adress.sw
- [x] address.sw
- [x] alloc.sw
- [x] assert.sw
- [x] auth.sw
- [x] b256.sw
- [x] b512.sw
- [x] block.sw
- [x] bytes.sw
- [x] call_frames.sw
- [x] constants.sw
- [x] context.sw
- [x] contract_id.sw
- [x] convert.sw
- [x] ecr.sw
- [x] error_signals.sw
- [x] external.sw
- [x] flags.sw
- [x] hash.sw
- [x] identity.sw
- [x] inputs.sw
- [x] instrinsics.sw
- [x] logging.sw
- [x] low_level_call.sw
- [x] math.sw
- [x] message.sw
- [x] option.sw
- [x] outputs.sw
- [x] registers.sw
- [x] result.sw
- [x] revert.sw
- [x] string.sw
- [x] token.sw
- [x] tx.sw
- [x] u128.sw
- [x] u256.sw
- [x] vec.sw

PR #4851, opened 2023-07-24
Std-lib external, call_frames, and context inline docs updates to meet SRC-2 standard

bitzoic:
## Description

This PR completes the external.sw, call_frames.sw, and context.sw tasks of https://github.com/FuelLabs/sway/issues/4812

## Checklist

### DIFF
--- a/sway-lib-std/src/call_frames.sw
+++ b/sway-lib-std/src/call_frames.sw
@@ -33,10 +33,12 @@ const SECOND_PARAMETER_OFFSET: u64 = 74;
 /// # Examples
 ///
 /// ```sway
-/// use std::call_frames::contract_id;
+/// use std::{call_frames::contract_id, context::this_balance, token::mint_to_contract};
 ///
 /// fn foo() {
 ///     let this_contract = contract_id();
+///     mint_to_contract(50, this_contract);
+///     assert(this_balance(this_contract) == 50);
 /// }
 /// ```
 pub fn contract_id() -> ContractId {
--- a/sway-lib-std/src/call_frames.sw
+++ b/sway-lib-std/src/call_frames.sw
@@ -33,12 +33,13 @@ const SECOND_PARAMETER_OFFSET: u64 = 74;
 /// # Examples
 ///
 /// ```sway
-/// use std::{call_frames::contract_id, context::this_balance, token::mint_to_contract};
+/// use std::{call_frames::contract_id, token::mint};
 ///
 /// fn foo() {
 ///     let this_contract = contract_id();
-///     mint_to_contract(50, this_contract);
-///     assert(this_balance(this_contract) == 50);
+///     mint(50);
+///     // Transfer the newly minted tokens to the ZERO_B256 address from this contract
+///     Address::from(ZERO_B256).transfer(50, this_contract)
 /// }
 /// ```
 pub fn contract_id() -> ContractId {
--- a/sway-lib-std/src/call_frames.sw
+++ b/sway-lib-std/src/call_frames.sw
@@ -20,13 +20,47 @@ const SECOND_PARAMETER_OFFSET: u64 = 74;
 //  Accessing the current call frame
 //
 /// Get the current contract's id when called in an internal context.
+///
+/// # Additional Information
+///
 /// **_Note:_** If called in an external context, this will **not** return a contract ID.
-// @dev If called externally, will actually return a pointer to the transaction ID.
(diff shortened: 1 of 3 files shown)


sway-lib-std/src/call_frames.sw
```diff
@@ -20,13 +20,47 @@ const SECOND_PARAMETER_OFFSET: u64 = 74;
...
+///
+/// * [ContractId] - The contract id of this contract.
+///
+/// # Examples
+///
+/// ```sway
+/// use std::call_frames::contract_id;
+///
+/// fn foo() {
+///     let this_contract = contract_id();
```
SwayStar123, line 39: example seems too barebones, maybe mint and then transfer some tokens by using the this_contract as the assetid

sway-lib-std/src/call_frames.sw
```diff
@@ -20,13 +20,47 @@ const SECOND_PARAMETER_OFFSET: u64 = 74;
...
+///
+/// # Returns
+/// 
+/// * [ContractId] - The asset included in the current call frame.
+///
+/// # Examples
+///
+/// ```sway
+/// use std::call_frames::msg_asset_id;
+/// use std::contsants::BASE_ASSET_ID;
```
SwayStar123, line 56: There should be a single use statement per lib, thats the generally followed convention. Also constants is mispelled
```suggestion
/// use std::{call_frames::msg_asset_id, constants::BASE_ASSET_ID};
```
Please remove the extra use statements in the rest of the code aswell

SwayStar123 (changes requested):

sway-lib-std/src/call_frames.sw
bitzoic: I wasn't sure what else I can add here that is actually relevant to the function.

I also don't think an example with minting tokens and using the contract Id as the asset id is a good idea since we are changing this with the multi token implementation coming soon™

SwayStar123: When multi token comes out it would only change a line wouldnt it? I dont see why that would be a reason to not have that example right now, we can change it later

sway-lib-std/src/call_frames.sw
```diff
@@ -20,13 +20,48 @@ const SECOND_PARAMETER_OFFSET: u64 = 74;
...
+/// * [ContractId] - The contract id of this contract.
+///
+/// # Examples
+///
+/// ```sway
+/// use std::{call_frames::contract_id, context::this_balance, token::mint_to_contract};
+///
+/// fn foo() {
+///     let this_contract = contract_id();
+///     mint_to_contract(50, this_contract);
```
SwayStar123, line 40: I dont think this works? mint already directly mints to the self contract, so this would transfer from self to self. also just makes it confusing even if it works. Sorry to be annoying about this. Maybe just `mint(50)` and then `Address::from(ZERO_B256).transfer(50, this_contract)`

sway-lib-std/src/context.sw
```diff
@@ -6,11 +6,50 @@ use ::call_frames::contract_id;
...
+/// * [u64] - The amount of the asset which the `target` holds.
+///
+/// # Examples
+///
+/// ```sway
+/// use std::{context::balance_of, token::mint, call_frames::contract_id};
+/// 
+/// fn foo() {
+///     mint(50);
+///     assert(this_balance(contract_id(), contract_id()) == 50);
```
SwayStar123, line 50: this example does not even use balance_of?

sway-lib-std/src/context.sw
```diff
@@ -19,6 +58,20 @@ pub fn balance_of(asset_id: ContractId, target: ContractId) -> u64 {
...
+///
+/// * [u64] - The amount of tokens being sent.
+///
+/// # Examples 
+/// 
+/// ```sway
+/// use std::context::msg_amount;
+/// 
+/// fn foo() {
+///     assert(msg_amount() == 0);
```
SwayStar123, line 72:
```suggestion
///     assert(msg_amount() == 0); // Ensures that no assets are being transferred in this context
```

sway-lib-std/src/context.sw
bitzoic: Whoops, I think this is a copy pasta error

sway-lib-std/src/call_frames.sw
bitzoic: This will need some revisions for the new spec but will work for now

SwayStar123 (approved):

PR #4874, opened 2023-07-26
add docs CI, rename docs README to index

sarahschwartz:


docs/book/src/forc/commands/forc_deploy.md
```diff
@@ -1 +0,0 @@
-# forc deploy
```
sdankel, line 1: Should we keep these and write documentation for them? Are they linked to from anywhere?

templates/sway-test-rs/README.md
```diff
@@ -6,6 +6,6 @@ testing within an existing Sway project.
 This template is designed for Rust developers who wish to test integration of
 their Rust application and Sway code.
 
-See the [Testing with Rust](https://fuellabs.github.io/sway/master/testing/testing-with-rust.html) chapter of [the Sway
+See the [Testing with Rust](https://fuellabs.github.io/sway/master/book/testing/testing-with-rust.html) chapter of [the Sway
```
sdankel, line 9: Thanks for fixing these links!

docs/reference/src/documentation/introduction/installation/source.md
```diff
@@ -4,7 +4,7 @@ The `Sway toolchain` can be built directly from the [`Sway repository`](https://
 
 ## Installation & Updating
 
-In the root of the repository `/sway/<here>` build [`forc`](../../forc/commands/index.md) with the following command:
+In the root of the repository `/sway/<here>` build [`forc`](https://fuellabs.github.io/sway/master/book/forc/commands/index.html) with the following command:
```
sdankel, line 7: Is it better to use the relative path, in case we change the DNS at some point?

docs/reference/src/documentation/introduction/installation/index.md
```diff
@@ -18,9 +18,9 @@ The supported operating systems include Linux and macOS; however, Windows is [`u
 
 `Cargo` may be used instead of [`Fuelup`](fuelup.md); however, the user needs to manage the toolchain themselves.
 
-The advantage of using `Cargo` is the installation of [`plugins`](../../forc/plugins/index.md) that have not been added into [`Fuelup`](fuelup.md).
+The advantage of using `Cargo` is the installation of [`plugins`](https://fuellabs.github.io/sway/master/book/forc/plugins/index.html) that have not been added into [`Fuelup`](fuelup.md).
```
sdankel, line 21: It seems to me we should either use the FQDN path or use relative paths linking to .md files everywhere for consistency.

sdankel (approved):
Great clean up of the docs, thank you! I left a few non-blocking comments.

docs/book/src/forc/commands/forc_deploy.md
kayagokalp: These should be kept they are auto generated by mdbook-preproccesor, which lives [here](https://github.com/FuelLabs/sway/tree/master/scripts/mdbook-forc-documenter)

kayagokalp:
Thanks for this! agreeing with @sdankel's comments, happy to approve afterwards

docs/book/src/forc/commands/forc_deploy.md
sarahschwartz: These are unused duplicate pages, there are also ones in the `forc_client` folder.

docs/reference/src/documentation/introduction/installation/source.md
sarahschwartz: It doesn't seem like the relative paths work across books

docs/reference/src/documentation/introduction/installation/index.md
sarahschwartz: see comment above

docs/book/src/forc/commands/forc_deploy.md
kayagokalp: Ahh alright then :)

docs/reference/src/documentation/introduction/installation/source.md
kayagokalp: Just a point to consider: 

Does reference has versions? i feel like linking like this would make versioned books linking to latest always. Example:

Imagine we released 0.43 with this merged. And after the merge we change some stuff in the place this link points. If a user goes to 0.43's documentation they should be taken to that link as it is while we were merging.

sarahschwartz: that's a good point - if I add a version number here though will someone update it for each release?

kayagokalp: That is a good question :) It is highly possible that we will forget that, we can try to do that until we can find a way to link relatively to a different book

sdankel (approved):

kayagokalp (approved):

PR #4889, opened 2023-07-31
Fixes replace_decls issue with nested and multiple types.

### LINKED ISSUE #4852: Compile errors with generic traits used with different types.
The following test case fails to compile in the last 2 asserts:

```rust
script;

use std::assert::*;

trait Trait {
    fn method(self) -> u64;
}

trait Trait2 {
    fn method2(self) -> u64;
}

impl<A, B> Trait for (A, B) {
    fn method(self) -> u64 {
        42
    }
}

impl Trait2 for u64 {
    fn method2(self) -> u64 {
        self
    }
}

struct S {}

impl Trait2 for S {
    fn method2(self) -> u64 {
        2
    }
}

impl<A, B> Trait2 for (A, B) where A: Trait2, B: Trait2 {
    fn method2(self) -> u64 {
        self.0.method2() + self.1.method2()
    }
}

fn main() -> bool {
    assert((1,2).method() == 42);
    assert((1,2).method2() == 3);
    assert((1, S{}).method2() == 3); //Internal compiler error: Verification failed: Type mismatch found for call to 'method2_10'
    assert(((1,2),(1,2)).method2() == 6); //Aborted (core dumped) (stack overflow)

    true
}
```

This issue probably occurs because we use a simple structure in `replace_decls` that can match a single trait method with multiple trait implementations.

esdrubal:
## Description

When a single method uses multiple trait implementation errors are thrown by the compiler similar issue occurs when a method requires a trait implementation and then calls another method that requires another implementation of the same trait.

This PR fixes both issues by filtering traits to be replaced by the arguments type used and trait method self type, and by gathering decl mapping of nested methods.

Fixes #4852

## Checklist

### DIFF
--- a/sway-core/src/decl_engine/mapping.rs
+++ b/sway-core/src/decl_engine/mapping.rs
@@ -103,7 +103,12 @@ impl DeclMapping {
         None
     }
 
-    pub(crate) fn filter(&self, self_type: TypeId, engines: &Engines) -> DeclMapping {
+    /// This method returns only associated item functions that have as self type the given type.
+    pub(crate) fn filter_functions_by_self_type(
+        &self,
+        self_type: TypeId,
+        engines: &Engines,
+    ) -> DeclMapping {
         let mut mapping: Vec<(SourceDecl, DestinationDecl)> = vec![];
         for (source_decl_ref, dest_decl_ref) in self.mapping.iter().cloned() {
             match dest_decl_ref {
--- a/sway-core/src/decl_engine/mapping.rs
+++ b/sway-core/src/decl_engine/mapping.rs
@@ -1,8 +1,11 @@
 use std::fmt;
 
-use crate::language::ty::{TyTraitInterfaceItem, TyTraitItem};
+use crate::{
+    language::ty::{TyTraitInterfaceItem, TyTraitItem},
+    Engines, TypeId, UnifyCheck,
+};
 
-use super::{AssociatedItemDeclId, InterfaceItemMap, ItemMap};
+use super::{AssociatedItemDeclId, DeclEngineGet, InterfaceItemMap, ItemMap};
 
 type SourceDecl = AssociatedItemDeclId;
 type DestinationDecl = AssociatedItemDeclId;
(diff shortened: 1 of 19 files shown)


sway-core/src/decl_engine/mapping.rs
```diff
@@ -99,4 +102,25 @@ impl DeclMapping {
         }
         None
     }
+
+    pub(crate) fn filter(&self, self_type: TypeId, engines: &Engines) -> DeclMapping {
```
IGI-111, line 106: Just naming it `filter` seems too vague. Could be appropriate if it's not ambiguous the sorts of filtering you can do on a `DeclMapping` but it's it's not readily apparent to me.

At the very least this needs a doc comment explaining what it's doing.

xunilrj (dismissed):

anton-trunov:
Looks like there are some merge conflicts to be fixed

sway-core/src/language/ty/expression/expression_variant.rs
```diff
@@ -880,7 +882,14 @@ impl ReplaceSelfType for TyExpressionVariant {
...
 
 impl ReplaceDecls for TyExpressionVariant {
-    fn replace_decls_inner(&mut self, decl_mapping: &DeclMapping, engines: &Engines) {
+    fn replace_decls_inner(
+        &mut self,
+        decl_mapping: &DeclMapping,
+        handler: &Handler,
+        ctx: &TypeCheckContext,
+    ) -> Result<(), ErrorEmitted> {
+        let mut error_emitted = None;
```
anton-trunov, line 891: Same `handler.scope` suggestion as above

anton-trunov (approved):

sway-core/src/language/ty/code_block.rs
```diff
@@ -41,10 +43,29 @@ impl ReplaceSelfType for TyCodeBlock {
...
-        self.contents
-            .iter_mut()
-            .for_each(|x| x.replace_decls(decl_mapping, engines));
+    fn replace_decls_inner(
+        &mut self,
+        decl_mapping: &DeclMapping,
+        handler: &Handler,
+        ctx: &TypeCheckContext,
+    ) -> Result<(), ErrorEmitted> {
+        let mut error_emitted = None;
```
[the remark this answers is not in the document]
esdrubal: Thanks for the suggestion, most recent version now uses scopes.

sway-core/src/language/ty/expression/expression_variant.rs
esdrubal: Thanks for the suggestion, most recent version now uses scopes.

IGI-111 (approved):

PR #4895, opened 2023-08-01
More error handling refactoring

### LINKED ISSUE #4885: Refactor `error_emitted` pattern
As part of #4878 we introduced an imperative error handling pattern where calls to `Handler::emit_err` would get fed into a `error_emitted: Option<ErrorEmitted>` variable to detect error emission within a function's context.

We should develop a facility within/associated with the `Handler` type to refactor that pattern into something more cleaner and more readable.

### LINKED ISSUE #4886: Refactor `ErrorRecovery`
To allow more errors to be produced, our current type system resolution produces an `ErrorRecovery` type in erroneous type checking cases that gets pass through the compiler and has to be handled.

This however breaks our current approach to error handling because it forces us to create a `ErrorEmitted` receipt manually much later than when the error has been detected. We should either find a way to pass it through or refactor our type checking to fix this problem.

IGI-111:
## Description
This changes addresses the issues introduced in #4878

The `error_emitted` pattern is replaced by a scoping method on `Handler` that collects errors within a closure and will error out if new erros are introduced.

All instances of introducing `ErrorEmitted` as a literal value have now been replaced by passing arround the receipt of the error emission through the error representing values. Some types from the AST had to be moved to `sway-types` to avoid circular dependencies introduced by this change.

This also repairs some of the discrepancies to emitted warnings introduced in #4878, as well as rename some of the `Handler` methods.

Fixes #4885
Fixes #4886

## Checklist


sway-core/src/semantic_analysis/ast_node/expression/typed_expression/if_expression.rs
```diff
@@ -81,7 +71,9 @@ pub(crate) fn instantiate_if_expression(
         .unwrap_or_else(|| type_engine.insert(engines, TypeInfo::Tuple(Vec::new())));
     // if there is a type annotation, then the else branch must exist
     if !else_deterministically_aborts && !then_deterministically_aborts {
-        let (new_warnings, new_errors) = type_engine.unify_with_self(
+        let h = Handler::default();
```
anton-trunov, line 74: Could you add a comment on this usage of `Handler::default()`? This looks different from the majority of the error handler things

anton-trunov (approved):

ironcev (dismissed):
Elegant and simple solution for the scoped handlers!

Passing the `ErrorEmitted` downstream approach is a bit different then what I originally thought, essentially other way around, the downstream tells the handler to stop and kind of short-circuit. But now when I think of it, passing the receipt definitely describes the use case better. An error happened before and we are deliberately reporting it later and this is now clearly communicated via `ErrorRecovery`.

anton-trunov (approved):

xunilrj (approved):

PR #4905, opened 2023-08-03
feat: derive first account with forc-deploy after creating a wallet

### LINKED ISSUE #4861: forc-deploy should automatically derive the first account after creating a new wallet
Realized while working on #4771, we need to derive at least a single account after creating the wallet through forc-deploy.

kayagokalp:
## Description
closes #4861.

This PR derives first account after creating a fresh wallet via forc-deploy so that we have at least one account available for users to chose from.

### DIFF
--- a/forc-plugins/forc-client/src/util/tx.rs
+++ b/forc-plugins/forc-client/src/util/tx.rs
@@ -16,7 +16,7 @@ use fuels_core::types::{
 };
 
 use forc_wallet::{
-    account::derive_secret_key,
+    account::{derive_secret_key, new_at_index_cli},
     balance::{
         collect_accounts_with_verification, print_account_balances, AccountBalances,
         AccountVerification, AccountsMap,
@@ -181,7 +181,10 @@ impl<Tx: Buildable + SerializableVec + field::Witnesses + Send> TransactionBuild
                     let accepted = ask_user_yes_no_question(&question)?;
                     if accepted {
                         new_wallet_cli(&wallet_path)?;
-                        println!("Wallet created successfully.")
+                        println!("Wallet created successfully.");
+                        // Derive first account for the fresh wallet we created.
+                        new_at_index_cli(&wallet_path, 0)?;
+                        println!("Account derived successfully.");
                     } else {
                         anyhow::bail!("Refused to create a new wallet. If you don't want to use forc-wallet, you can sign this transaction manually with --manual-signing flag.")
                     }


forc-plugins/forc-client/src/util/tx.rs
```diff
@@ -156,7 +160,9 @@ impl<Tx: Buildable + SerializableVec + field::Witnesses + Send> TransactionBuild
                         let accepted = ask_user_yes_no_question(&question)?;
                         if accepted {
                             new_wallet_cli(&wallet_path)?;
-                            println!("Wallet created successfully.")
+                            println!("Wallet created successfully. Deriving first account");
```
JoshuaBatty, line 163:
```suggestion
                            println!("Wallet created successfully. Setting up your first account...");
```
Perhaps this language is clearer?

kayagokalp: That is a good point to discuss. Because we use this terminology all around the stack. IMO suggestion makes it clear for most of the users but I believe technically correct term is "deriving" so I am perfectly happy to apply the suggestion or leave this as it is.


### A little explanation of how I mapped this stuff in my mind

This terminology is coming from the fact that forc-wallet is a HD-wallet, ([here](https://www.ledger.com/academy/crypto/what-are-hierarchical-deterministic-hd-wallets) is a nice article about what it is) and with hd-wallet we derive accounts. (A really simplified summary of HD wallet's usage of "derivation" comes from the fact that accounts already exist all the time but not discovered by the user yet, thus they are derived using your private key)

Maybe developer facing stuff should use the right technical term but stuff like front-end wallet should mention this as setting up? Anyways let me know what you think and we can apply this suggestion to couple more places derivation is used for account creation.

sdankel: Is "derive" used in any other user-facing contexts at the moment?

forc-plugins/forc-client/src/util/tx.rs
```diff
@@ -156,7 +160,9 @@ impl<Tx: Buildable + SerializableVec + field::Witnesses + Send> TransactionBuild
...
+                            // Derive first account for the fresh wallet we created.
+                            new_at_index_cli(&wallet_path, 0)?;
```
sdankel, line 165: Should we have a success message after the account is derived?

forc-plugins/forc-client/src/util/tx.rs
kayagokalp: Hmm i though we used that more frequently but quick search yielded only the https://github.com/FuelLabs/forc-wallet/blob/master/README.md, might need to look deeper to find all occurrences.

forc-plugins/forc-client/src/util/tx.rs
kayagokalp: Currently it looks like the following:
```
Wallet created successfully. Deriving first account
Please enter your wallet password to derive account 0: 
Wallet address: fuel1dy60cz4kzxxzexdaqkvupcuwvajs6pz859yawv2awrjmh3mj4xzsuk9qfh
```
After the account is derived forc-wallet prints wallet address. Should we make it like:

```
Wallet created successfully. Deriving first account
Please enter your wallet password to derive account 0: 
Wallet address: fuel1dy60cz4kzxxzexdaqkvupcuwvajs6pz859yawv2awrjmh3mj4xzsuk9qfh
Account derived successfully.
```

forc-plugins/forc-client/src/util/tx.rs
JoshuaBatty: Thanks for the explanation Kaya and for the link. In that case I think deriving is more accurately describing what is going on. Happy for us to continue using that terminology.

JoshuaBatty:
I'm happy to approve this once the conflict has been resolved.

JoshuaBatty (approved):

Voxelot (approved):

sdankel (approved):

PR #4918, opened 2023-08-05
feat: enable spending predicates with forc-run

### LINKED ISSUE #3440: `forc-client` - support working with predicates
## Predicates

My understanding of the way predicates work is still a little fuzzy, but I believe the essence is this:

1. **Predicates have addresses.**
    - Basically, the hash of their bytecode, similar to contracts.
2. **A user can send some resources _to_ a predicate.**
    - All that is required is a wallet with funds, the predicate address and a script transaction.
3. **A user can attempt to spend resources _from_ a predicate.**
    - The predicate bytecode must be provided as a part of the script transaction input.
    - Similarly, predicate input data can be provided (basically predicate main function args?).
    - The input predicates will be executed prior to the script.
    - The script execution (and in turn, the spend) only occurs if the predicates evaluate to `true`.

Ideally, we should support all of the above via the `forc-client` CLI.

### 1. Predicate Addresses

This could be a simple `forc predicate-id` command, or outputting the ID to the `out/` directory as a part of `forc build` (or both).

### 2. Sending resources to a predicate

This should be possible today using `forc run` with a custom script that does the send.

Ideally, `forc-wallet` would make it easier to send funds with something like `forc wallet send ...`. I've opened https://github.com/FuelLabs/forc-wallet/issues/65.

### 3. Spending a predicate

To begin, I think the main thing we want to support here is specifying predicates via the `forc run` command. E.g. something like:

```console
forc run --predicate ./out/debug/predicate.bin --predicate-data "arg1 arg2 arg3"
```

- The `--predicate` flag points to some predicate bytecode.
- The `--predicate-data` flag provides some predicate input data. We could expand on this with something similar to #3439.

We could potentially allow for multiple predicates by always requiring that the `--predicate-data` flag immediately follows the `--predicate` flag, and then simply allow for specifying multiple `--predicate` flags.

We could also optionally allow for specifying the predicate address via a `--predicate-id` flag, as a kind of sanity check to ensure that the bytecode pointed to is indeed the correct predicate.

---

cc @simonr0204 I remember we talked very briefly about some ideas for working with predicates via the CLI - feel free to add any thoughts/ideas here!

kayagokalp:
## Description
close #3440.

This PR introduces predicate spending capabilities to `forc-run`. Basically forc-run can be used to spend predicates if run inside a predicate project. There are couple of new flags required by this new flow.

1. `--receive-address <RECEIVE_ADDRESS>`: Receiving address of the spent predicate assets
2. `--spend-amount <SPEND_AMOUNT>`: Amount to unlock for spending a predicate

This works with fuel-core version 0.19.1 and newer.

It is pretty simple to test manually but if you prefer you can test with this script from root of sway repo after checking out this branch: 

**chainConfig.json  should fund your wallet and predicate root so that you can spend it. You can fund your wallet through chainConfig.json and use forc-wallet to transfer funds to the predicate root.**

```sh
#!/bin/bash

current_dir="$(readlink -f "$0" | xargs dirname)"

(cd "$current_dir" && cargo build)
forc_binary="$current_dir/target/debug/forc"
forc_run_binary="$current_dir/target/debug/forc-run"

tmp_dir="$(mktemp -d -p "$current_dir")"
fuel-core run --chain chainConfig.json --db-type in-memory </dev/null &> /dev/null &
fuel_core_pid=$!
cleanup() {
    rm -rf "$tmp_dir";
    kill -9 "$fuel_core_pid"
}
trap cleanup EXIT

cd "$tmp_dir"

"$forc_binary" new --predicate test-predicate
cat > test-predicate/src/main.sw <<LIMIT
predicate;

fn main(a: u32, b: u32) -> bool {
   a + b == 5
}
LIMIT

echo 'std = { path = "../../sway-lib-std" }' >> test-predicate/Forc.toml

cd test-predicate

"$forc_binary" build
"$forc_run_binary" --node-url http://127.0.0.1:4000/ --spend-amount 0 --receive-address "fuel1n75f37x7gdhhrmldq75s689vpsru3svun0ttna33pdduhypfkk5s2p29qh" --args 2 --args 3
```

Example chainConfig.json can be found [here](https://github.com/FuelLabs/fuels-ts/blob/a157551d370b13dbff65f5f85f33a6d0c22a1dc4/.fuel-core/configs/chainConfig.json#L4). Basically get your public key as hex from forc-wallet and in initial states section fund your wallet.

1. `forc-wallet account 0 public-key --as-hex`
2. Add this section to chainConfig:
```json
"initial_state": {
    "coins": [
      {
        "owner": "0x795ac248dd5525d9c8791d0dc28a4721e88753cb407177857bb106b22084245a",
        "amount": "0x00000000FFFFFFFF",
        "asset_id": "0x0000000000000000000000000000000000000000000000000000000000000000"
      }
    ]
  }
```
3. Transfer some funds to predicate root. 
```forc-wallet account 0 transfer --to 0x1f4785e65cde317f0bcb0c314cad72c9d29ea39b9c3e9b8819e8222b5e941747 --amount 100 --asset-id 0x0000000000000000000000000000000000000000000000000000000000000000 --gas-price 1 --node-url http://127.0.0.1:4000```

## Checklist


kayagokalp:
Also i have an issue #4919 to further split this into a separate command, i.e `forc-spend`.

sdankel:
> Also i have an issue #4919 to further split this into a separate command, i.e `forc-spend`.

Would it be better to just do that instead of this PR, so that we're not adding and then removing arguments from `forc-run`?

kayagokalp:
> > Also i have an issue #4919 to further split this into a separate command, i.e `forc-spend`.
> 
> Would it be better to just do that instead of this PR, so that we're not adding and then removing arguments from `forc-run`?

Totally! I opened the issue and linked it here so that we can decide which way to go, original issue wanted this to be part of forc-run but if forc-spend sounds better I will split it here :)

kayagokalp:
closing as there were no consensus of whether this should been part of a forc-run or not

Issue #4911, opened 2023-08-04
parser recovery: `let` statement without an `=` sign fails to return an AST


JoshuaBatty:
Looking at the below example, we can see squiggly lines are drawn under all of the typed tokens that could be parsed in the language server.
 
![Screenshot 2023-07-26 at 12 45 01 pm](https://github.com/FuelLabs/sway/assets/1289413/db7b9ee4-98e8-4fb1-96a8-582ee633940a)

Now, if we add `let` at the top of the function then you notice we lose access to all the tokens.  

![Screenshot 2023-07-26 at 12 49 22 pm](https://github.com/FuelLabs/sway/assets/1289413/9a4daf4f-f685-4f45-8bbb-a28ab643bed8)

The parser fails to recover on the below examples.

```rust
let
```
```rust
let x
```
```rust
let x:
```
```rust
let x: u8
```

and only recovers when an `=` sign is typed leading to an expression

```rust 
let x = 
```

PR #4925, opened 2023-08-08
parser recovery for let statements

xunilrj:
## Description

This PR closes https://github.com/FuelLabs/sway/issues/4911.

It primarily does two things:

1 - Extend a lot of functions that return `CompileError` to return `Vec<CompileError>` allowing multiple errors to return;
2 - Implements the concept of parser recovery.

This happens when using the function `guarded_parse_with_recovery`, exemplified below. When the guard fails, this function return `Ok(None)`; when it succeeds it returns `Ok(item)`. The exciting part is when the parsing fails.

It returns an instance of `Recoverer` which contains a reference to the original parser before any tentative parser was done and the forked as left by the parsing function.

The idea is that it is the caller's responsibility to put the forked parser in a "good position". For that, it offers some helpers function and access to the forked parser.

To avoid boilerplate code one recovery strategy is already implemented `recover_at_next_line_with_fallback_error` which consumes everything at the forked parser's current line and emits an error if none were generated by the tentative parser.

```rust
 match parser.guarded_parse_with_recovery::<LetToken, StatementLet>() {
        Ok(None) => {}
        Ok(Some(item)) => return stmt(Statement::Let(item)),
        Err(r) => {
            let (spans, error) =
                r.recover_at_next_line_with_fallback_error(ParseErrorKind::InvalidStatement);
            return stmt(Statement::Error(spans, error));
        }
    }
```

With this, we have the LSP not "dying" when a strange error happens. The errors themselves are not brilliant, but they will be improved in other PRs.

image

## Checklist

### DIFF
--- a/sway-core/src/control_flow_analysis/analyze_return_paths.rs
+++ b/sway-core/src/control_flow_analysis/analyze_return_paths.rs
@@ -18,17 +18,30 @@ impl<'cfg> ControlFlowGraph<'cfg> {
     pub(crate) fn construct_return_path_graph<'eng: 'cfg>(
         engines: &'eng Engines,
         module_nodes: &[ty::TyAstNode],
-    ) -> Result<Self, CompileError> {
+    ) -> Result<Self, Vec<CompileError>> {
+        let mut errors = vec![];
+
         let mut graph = ControlFlowGraph::default();
         // do a depth first traversal and cover individual inner ast nodes
         let mut leaves = vec![];
         for ast_entrypoint in module_nodes {
-            let l_leaves = connect_node(engines, ast_entrypoint, &mut graph, &leaves)?;
-            if let NodeConnection::NextStep(nodes) = l_leaves {
-                leaves = nodes;
+            match connect_node(engines, ast_entrypoint, &mut graph, &leaves) {
+                Ok(l_leaves) => {
+                    if let NodeConnection::NextStep(nodes) = l_leaves {
+                        leaves = nodes;
+                    }
+                }
+                Err(mut e) => {
+                    errors.append(&mut e);
+                }
             }
         }
-        Ok(graph)
+
+        if !errors.is_empty() {
+            Err(errors)
+        } else {
+            Ok(graph)
+        }
     }
 
     /// This function looks through the control flow graph and ensures that all paths that are
@@ -130,7 +143,7 @@ fn connect_node<'eng: 'cfg, 'cfg>(
     node: &ty::TyAstNode,
     graph: &mut ControlFlowGraph<'cfg>,
     leaves: &[NodeIndex],
-) -> Result<NodeConnection, CompileError> {
+) -> Result<NodeConnection, Vec<CompileError>> {
     match &node.content {
         ty::TyAstNodeContent::Expression(ty::TyExpression {
             expression: ty::TyExpressionVariant::Return(..),
@@ -171,7 +184,7 @@ fn connect_node<'eng: 'cfg, 'cfg>(
         )),
         ty::TyAstNodeContent::Error(spans, _) => {
             let span = Span::join_all(spans.iter().cloned());
-            return Err(CompileError::InvalidStatement { span });
+            return Err(vec![CompileError::InvalidStatement { span }]);
         }
     }
 }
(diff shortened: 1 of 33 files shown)


sway-core/src/control_flow_analysis/analyze_return_paths.rs
```diff
@@ -18,17 +18,30 @@ impl<'cfg> ControlFlowGraph<'cfg> {
...
         for ast_entrypoint in module_nodes {
-            let l_leaves = connect_node(engines, ast_entrypoint, &mut graph, &leaves)?;
-            if let NodeConnection::NextStep(nodes) = l_leaves {
-                leaves = nodes;
+            match connect_node(engines, ast_entrypoint, &mut graph, &leaves) {
+                Ok(l_leaves) => {
+                    if let NodeConnection::NextStep(nodes) = l_leaves {
+                        leaves = nodes;
+                    }
+                }
```
IGI-111, line 33:
```suggestion
                Ok(NodeConnection::NextStep(nodes)) => {
                    leaves = nodes;
                }
                Ok(_) => {}
```

sway-core/src/lib.rs
```diff
@@ -511,6 +513,9 @@ pub fn compile_to_ast(
         metrics
     );
 
+    //  // Fail compilation if we have errors
+    //  handler.ok()?;
```
IGI-111, line 517: Should this be removed?

sway-error/src/handler.rs
```diff
@@ -80,11 +80,40 @@ impl Handler {
...
+        let other = other.inner.borrow();
+
+        for warn in other.warnings.iter() {
+            self.emit_warn(warn.clone());
+        }
+        for err in other.errors.iter() {
+            self.emit_err(err.clone());
+        }
+    }
+
```
IGI-111, line 93: I don't think this is needed when you can just `h1.append(h2.clone())` and get the same result.

sway-lsp/src/utils/debug.rs
```diff
@@ -105,6 +105,7 @@ pub(crate) fn print_decl_engine_types(
                 format!("{expression:#?}")
             }
             ty::TyAstNodeContent::SideEffect(side_effect) => format!("{side_effect:#?}"),
+            ty::TyAstNodeContent::Error(_, _) => todo!(),
```
IGI-111, line 108: Is this still to do?

sway-error/src/handler.rs
```diff
@@ -80,11 +80,40 @@ impl Handler {
...
         inner.warnings = dedup_unsorted(inner.warnings.clone());
     }
+
+    pub fn ok(&self) -> Result<(), ErrorEmitted> {
+        if !self.has_errors() {
+            Ok(())
+        } else {
+            Err(ErrorEmitted { _priv: () })
+        }
+    }
```
IGI-111, line 106: This defeats the purpose of propagating the receipts, and it's not used anywhere that I can see.

sway-parse/src/parser.rs
```diff
@@ -153,6 +223,43 @@ impl<'a, 'e> Parser<'a, 'e> {
...
+        }
+    }
+
+    pub fn clear_errors(&mut self) {
+        self.handler.clear_errors();
+    }
+
+    pub fn clear_warnings(&mut self) {
+        self.handler.clear_errors();
+    }
```
IGI-111, line 254: I'm very much against the idea of clearing all errors in the handlers like this because it introduces an easy way to remove meaningful errors that are outside of one's scope.

If you want to conditionally append errors, you should pass a new empty Handler and append it to the global one or not depending on whether you want to commit to the errors. Mutating the existing error set I don't think is acceptable.

IGI-111 (changes requested):
I think the base structure is sound but the way the error handling is propagated needs to be rewritten, you shouldn't ever need to clear errors and warnings.

sway-error/src/handler.rs
xunilrj: They actually have a slight difference.

```
h1.append(h2.clone())
```

is cloning two `Vec`, two extra allocations. While `append_ref` will allocate only when the variants allocate (in some cases).

But I can simplify for now, given this is only called on error paths.

sway-error/src/handler.rs
xunilrj: Fixed.

sway-core/src/control_flow_analysis/analyze_return_paths.rs
xunilrj: >  you shouldn't ever need to clear errors and warnings.
To recover, you don't need it. I will push improved errors on recovery for another PR, then.

sway-error/src/error.rs
```diff
@@ -286,6 +286,7 @@ pub enum CompileError {
     },
     #[error("This opcode takes an immediate value but none was provided.")]
     MissingImmediate { span: Span },
+
```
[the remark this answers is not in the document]
xunilrj: I will fix these in the next PR to merge this.

Issue #4901, opened 2023-08-02
Benchmark LSP Event Execution Times for Rust Analyzer vs sway-lsp


JoshuaBatty:
It would be beneficial to create identical projects in Sway and Rust, and then benchmark the time taken to execute each Language Server Protocol (LSP) event in rust-analyzer and sway-lsp.

This would give us a quantitative basis for understanding the performance differences between the two implementations. Identifying any significant discrepancies could lead to targeted optimization efforts.

JoshuaBatty:
Ok, in order to create compare the 2 language servers, I first used chatGPT to create a large ~1700 line rust project with lots of types and variable assignments. I then created a new sway project and tweaked the file slightly so it would compile.

The rust and sway projects can be found [here](https://github.com/JoshuaBatty/lsp-test-projects).

I then triggered each of the below LSP features 10 times in both `sway-lsp` and `rust-analyzer` and recorded the time it took to complete. The numbers below are the averaged result for each feature. 

| Request Type         | rust-analyzer | sway-lsp | Speed Difference | Notes                                      |
|----------------------|---------------|----------|------------------|--------------------------------------------|
| didChange            | 305us         | 2270ms   | 7443x           | Due to sway-compiler                      |
| semanticTokens/full  | 45us          | 2300ms   | 51,111x         | Blocks in sway-lsp until compilation is finished |
| hover                | 115us         | 495ms    | 4304x           |                                            |
| documentHighlight    | 83us          | 1070ms   | 12,892x         |                                            |
| definition           | 128us         | 1050ms   | 8203x           |                                            |
| prepareRename        | 47us          | 3520ms   | 74,894x         |                                            |
| rename               | 54us          | 470ms    | 8704x           |                                            |

I was expecting to see such a large difference in `didChange` and `semanticTokens/full` because we are compiling the entire project and all of it's dependencies on each key pressed event, whereas `rust-analyzer` is using incremental compilation. 

I wasn't expecting our requests to take such a long time though. I plan to look into how we can optimise these next.

JoshuaBatty:
Ok I set up a benchmarking framework using criterion so we can get more accurate results.
Here are the results for the LSP requests. 

| Request Type         | duration  |
|----------------------|---------------|
| hover                | 442ms    |
documentHighlight    | 441ms   |
| definition           | 433ms   |
 prepareRename        | 433ms   |
| rename               | 436ms    |

[...]

PR #4927, opened 2023-08-09
Add benchmark framework to `sway-lsp`

JoshuaBatty:
## Description
This PR adds a way to benchmark the LSP requests and `TokenMap` methods using `criterion`. I've also added a large sway benchmark project that can be used when running benchmarks now and in the future to measure performance improvements. 

Here is the result of the benchmarks. 

| Request Type         | duration  |
|----------------------|---------------|
| hover                | 442ms    |
documentHighlight    | 441ms   |
| definition           | 433ms   |
 prepareRename        | 433ms   |
| rename               | 436ms    |
 semantic_tokens        | 464ms   |
| document_symbol               | 447ms    |
 completion        | 893ms   |
| inlay_hints               | 15ms    |
 code_action        | 444ms   |
| code_lens        | 34ms   |
 on_enter               | 572ms    |

| TokenMap method         | duration  |
|----------------------|---------------|
| tokens_for_file                | ~~230ns~~ 10ms    |
idents_at_position    | 438ms   |
| tokens_at_position           | 437ms   |
 token_at_position        | 439ms   |
| parent_decl_at_position               | 437ms    |

| Compilation        | duration  |
|----------------------|---------------|
| compile_and_traverse                | 1.795sec    |

related to #4901 

## Checklist


anton-trunov:
LGTM! Perhaps we should add some kind of developer documentation where it says we actually have benchmarks and how to run those?

sway-lsp/src/core/session.rs
```diff
@@ -86,6 +86,12 @@ pub struct Session {
     pub diagnostics: Arc<RwLock<Diagnostics>>,
 }
 
+impl Default for Session {
```
xunilrj, line 89: Do we need both `default()` and `new()`?

xunilrj:
Some methods like `tokens_for_file` are lazy. Maybe that is why their result is so good. You should `collect()` its results to a `Vec`.

I am also wondering if it is worth to serialize responses to `JSON`. Because this way we also assess the "size" of the response.

sway-lsp/benches/lsp_benchmarks/requests.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    c.bench_function("goto_definition", |b| {
+        b.iter(|| session.token_definition_response(uri.clone(), position))
+    });
+
+    c.bench_function("prepare_rename", |b| {
+        b.iter(|| capabilities::rename::prepare_rename(session.clone(), uri.clone(), position))
+    });
+
+    c.bench_function("rename", |b| {
```
sdankel, line 28: Should we do inlay hints as well, since it's highly visible to users?

sdankel (approved):
Awesome work! 

If the benchmarks don't take too long to run, it would be cool to integrate the request benchmarks into CI with a script like [this](https://gist.github.com/yamafaktory/a3512b379c8caf7fd999ac7c25d0a1df#file-after-success-sh). We could detect any change in the compiler or LSP code that significantly slows things down.

sway-lsp/src/core/session.rs
JoshuaBatty: I only added this to appease clippy. Seems like this lint has recently been added.

JoshuaBatty:
> Some methods like `tokens_for_file` are lazy. Maybe that is why their result is so good. You should `collect()` its results to a `Vec`.

Good point, will modify the benchmark to collect into a Vec for this one. 
I'll have a think about measuring the size and address that in a later PR. Thanks for the review!

sway-lsp/benches/lsp_benchmarks/requests.rs
JoshuaBatty: Yeah good idea, i'll add inlay hints to benchmark also.

JoshuaBatty: I've just added in the missing requests and updated the results

JoshuaBatty:
> If the benchmarks don't take too long to run, it would be cool to integrate the request benchmarks into CI with a script like [this](https://gist.github.com/yamafaktory/a3512b379c8caf7fd999ac7c25d0a1df#file-after-success-sh). We could detect any change in the compiler or LSP code that significantly slows things down.

Nice, yeah something like that could be cool to integrate into our CI down the track. Will have a think about what would be useful for us.

JoshuaBatty:
> LGTM! Perhaps we should add some kind of developer documentation where it says we actually have benchmarks and how to run those?

I've added a page under the Engineering/LSP page in the internal Notion docs for now. We could perhaps add another .MD to this repo at some point if the need arrises. 

If you want to run them yourself just `cd` into the `sway-lsp` directory and run `cargo bench`

sdankel (approved):
:shipit:

sdankel (approved):

xunilrj (approved):

PR #4928, opened 2023-08-09
u256 shifts operators

xunilrj:
## Description

This PR is part of https://github.com/FuelLabs/sway/pull/4794.

It implements left and right shifts. These have a little difference because the `rhs` is always `u64`, which means that the WQOP used the `rhs` is not indirect.

## Checklist


test/src/e2e_vm_tests/test_programs/should_pass/language/u256/src/main.sw
```diff
@@ -26,6 +26,12 @@ fn not_operator() -> u256 {
     !(a - b)
 }
 
+// returns 8
+fn shift_operators() -> u256 {
```
anton-trunov, line 30: maybe add some corner cases here too? like shifting all bits out of a `u256`, making the least significant bit the most significant and vice versa

xunilrj: Done,

anton-trunov: Ah, my bad, I should've used better wording. What I meant is something like `0x0...01 << 255` becomes `0x10....0`, and `0x10..0 >> 255` becomes `0x0...01`

anton-trunov: And for the "shifting out" all bits `0x1...11 << 256` becomes `0x0`, and `0x11..1 >> 256` becomes `0x0` as well

xunilrj: Ah... Done.

sway-core/src/asm_lang/virtual_immediate.rs
```diff
@@ -11,6 +11,8 @@ pub enum WideOperations {
     Add = 0,
     Sub = 1,
     Not = 2,
+    Lsh = 6,
```
IGI-111, line 14: I assume there's an explanation for the jump in value, but what is it?

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
```diff
@@ -885,9 +885,10 @@ fn type_check_shift_binary_op(
         handler,
         ctx.by_ref()
             .with_help_text("Incorrect argument type")
-            .with_type_annotation(return_type),
+            .with_type_annotation(engines.te().insert(engines, TypeInfo::Numeric)),
```
IGI-111, line 888: This doesn't look right, we should be type checking the intrinsics. With this code you could have `let _: u32 = 1u64 << 1u64;` which is incorrect.

What's the reason to relax it here?

sway-ir/src/verify.rs
```diff
@@ -389,11 +389,24 @@ impl<'a, 'eng> InstructionVerifier<'a, 'eng> {
...
+            // Shifts rhs are 64 bits
+            BinaryOpKind::Lsh | BinaryOpKind::Rsh => {
+                if !arg1_ty.is_ptr(self.context)
+                    || !arg2_ty.is_uint64(self.context)
+                    || !result_ty.is_ptr(self.context)
+                {
+                    return Err(IrError::VerifyBinaryOpIncorrectArgType);
+                }
+            }
+            _ => {
```
IGI-111, line 402: You should explicit the other ops so we don't introduce a bug if a new one is added.

sway-ir/src/verify.rs
```diff
@@ -431,8 +444,19 @@ impl<'a, 'eng> InstructionVerifier<'a, 'eng> {
...
-            return Err(IrError::VerifyBinaryOpIncorrectArgType);
+
+        match op {
+            // Shifts can have the rhs with different type
+            BinaryOpKind::Lsh | BinaryOpKind::Rsh => {
+                if !arg1_ty.is_uint(self.context) || !arg2_ty.is_uint(self.context) {
+                    return Err(IrError::VerifyBinaryOpIncorrectArgType);
+                }
+            }
+            _ => {
```
IGI-111, line 455: Same here

sway-core/src/asm_lang/virtual_immediate.rs
xunilrj: This is coming from: http://specs.fuel.network/master/fuel-vm/instruction-set.html#wdop-misc-128-bit-integer-operations

sway-ir/src/verify.rs
xunilrj: Good catch. Done.

sway-ir/src/verify.rs
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
xunilrj: You are right, This is wrong. Fixed.

IGI-111 (approved):

Issue #4959, opened 2023-08-16
Parallelize traversing AST's in the language server.


JoshuaBatty:
We should look to leverage parallel iterators using the `rayon` crate when traversing the AST's in the language server. This should hopefully lead to a nice performance boost.

PR #4961, opened 2023-08-16
Optimize language server performance

JoshuaBatty:
## Description
This PR speeds up all features in the language server, at the cost of a small performance hit during AST traversal. 

The first optimisation is the `Position` type in `sway-types/src/span.rs`. This has been changed to take a reference instead of using clones. The `line_col()` method was getting called a lot in order to convert between the span's start and end to the `lsp_types::Range` type that is used in LSP. As such, we were calculating this for every token in the language server when iterating over the `TokenMap` for all of the features. Removing the clones led to a 40% speed up. 

The last optimisation introduces a new `TokenIdent` type. This is similar to `sway_types::Ident` but holds data that is more useful for LSP, specifically, a `PathBuf` and `Range`. We are now precomputing these _during_ AST traversal and using the new `TokenIdent` as the key into the `TokenMap`. This avoids needing to pass the source engine around to convert `SourceId`'s to paths and also avoids needing to continually compute the range more than once.

I think the small performance hit we take moving this computation to the traversal stage is worth the insane speed increases we get for all other LSP features. I have a plan in #4959 to parallelize traversing the AST's using the `rayon` crate in a follow up PR that overcomes this performance regression. 

Here is the result of running `cargo bench` with these optimisations applied. 

| Type | Previous Duration | Current Duration | Speedup | Improvement % |
|---------------------|-------------------|------------------|---------|---------------|
| on_enter            | 572ms             | 567ns            | 1008818.34x | 100.0%   |
| code_lens           | 34ms              | 16.059µs         | 2117.19x | 99.95%    |
| idents_at_position  | 438ms             | 3.281ms          | 133.50x | 99.25%    |
| prepareRename       | 433ms             | 6.635ms          | 65.26x | 98.47%    |
| token_at_position   | 439ms             | 6.76ms           | 64.94x | 98.46%    |
| definition          | 433ms             | 6.726ms          | 64.38x | 98.45%    |
| hover               | 442ms             | 7.389ms          | 59.82x | 98.33%    |
| document_symbol     | 447ms             | 7.49ms           | 59.68x | 98.32%    |
| semantic_tokens     | 464ms             | 24.48ms          | 18.95x | 94.72%    |
| completion          | 893ms             | 51.03ms          | 17.50x | 94.29%    |
| parent_decl_at_position | 437ms        | 42.49ms          | 10.28x | 90.28%    |
| tokens_at_position  | 437ms             | 42.87ms          | 10.19x | 90.19%    |
| code_action         | 444ms             | 56.97ms          | 7.79x | 87.17%    |
| documentHighlight   | 441ms             | 115ms            | 3.83x | 73.92%    |
| rename              | 436ms             | 115.35ms         | 3.78x | 73.54%    |
| inlay_hints         | 15ms              | 9.75ms           | 1.54x | 35.0%     |
| compile_and_traverse | 1.795sec         | 1.9715sec        | 0.91x | -9.83%    |
| tokens_for_file     | 10ms              | 11.37ms          | 0.88x | -13.7%    |

Related to: #4901

## Checklist


sway-types/src/span.rs
```diff
@@ -11,59 +11,27 @@ lazy_static! {
...
-                Some(c) => {
-                    pos -= c.len_utf8();
-                    line_col = (line_col.0, line_col.1 + 1);
-                }
-                None => unreachable!(),
-            }
-        }
-
-        line_col
+        let slice = &self.input[..self.pos];
```
xunilrj, line 30: This is nice, but I think you can avoid the `collect` here.

```rust
#[inline]
    pub fn line_col(&self) -> (usize, usize) {
        if self.pos > self.input.len() {
            panic!("position out of bounds");
        }

        let (line_count, last_line) = self.input[..self.pos]
            .lines()
            .fold((0, None), |(line_count, _), line| {
                (line_count + 1, Some(line))
            });
        let last_line_len = last_line.map(|x| x.chars().count()).unwrap_or_default();
        (line_count, last_line_len)
    }
```

xunilrj:
>  Removing the clones led to a 40% speed up.

Brilliant!

> The line_col() method was getting called a lot in order to convert between the span's start and end to the lsp_types::Range type that is used in LSP. As such, we were calculating this for every token in the language server when iterating over the TokenMap for all of the features.

I am also having this problem elsewhere. Maybe we should have a data struct `offset -> line number` somewhere? `crates.io` has some interval trees. Not sure which one would be the best fit.

xunilrj (dismissed):

sway-types/src/span.rs
kayagokalp: +1 on usage of `.lines()` instead of splitting with `\n` if there are no drawbacks to do so. We do not officially support Windows but if somehow we decide to support offloading handling different line endings to `std` sounds nice :)

kayagokalp (approved):
This is fantastic!

sway-lsp/src/capabilities/document_symbol.rs
```diff
@@ -1,10 +1,9 @@
-use crate::core::token::{get_range_from_span, SymbolKind, Token};
+use crate::core::token::{SymbolKind, Token, TokenIdent};
 use lsp_types::{self, Location, SymbolInformation, Url};
-use sway_types::{Ident, Spanned};
 
 pub fn to_symbol_information<I>(tokens: I, url: Url) -> Vec<SymbolInformation>
 where
-    I: Iterator<Item = (Ident, Token)>,
+    I: Iterator<Item = (TokenIdent, Token)>,
```
sdankel, line 6: Is the token in TokenIdent the same as the token in the 2nd part of the tuple? A comment here would be helpful.

sway-lsp/src/capabilities/hover/mod.rs
```diff
@@ -126,11 +123,9 @@ fn hover_format(
     session: Arc<Session>,
     engines: &Engines,
     token: &Token,
-    ident: &Ident,
+    token_name: &str,
```
sdankel, line 126: Nice, but would `ident_name` be more accurate here?

sway-lsp/src/capabilities/inlay_hints.rs
```diff
@@ -43,7 +43,7 @@ pub fn inlay_hints(
 
     let hints: Vec<lsp_types::InlayHint> = session
         .token_map()
-        .tokens_for_file(engines.se(), uri)
+        .tokens_for_file(uri)
```
sdankel, line 46: Nice to not have to pass in the source engine everywhere :)

sway-lsp/src/capabilities/rename.rs
```diff
@@ -148,29 +145,32 @@ fn is_token_in_workspace(
...
     let temp_path = &session.sync.temp_dir()?;
-    if let Some(id) = decl_span.source_id() {
-        let path = engines.se().get_path(id);
+    if let Some(path) = decl_ident.path {
         if !path.starts_with(temp_path) {
             return Err(LanguageServerError::RenameError(
                 RenameError::TokenNotPartOfWorkspace,
             ));
         }
     }
```
sdankel, line 160: nit: It would be nice to move this into a helper function in `sync.rs` with a comment. It took me a minute to understand the difference between what this is checking and what `sync.is_path_in_temp_workspace()` does. I believe this is checking that it's in a session subdirectory within the temp directory, right?

Does `TokenIdent.path` always prefer the temp path if it exists?

sway-lsp/src/core/token.rs
```diff
@@ -162,22 +163,60 @@ impl Token {
...
+/// Although this introduces a small amount of overhead while traversing, precomputing this
+/// greatly speeds up performace in all other areas of the language server.
+///
+/// [TokenMap]: crate::core::token_map::TokenMap
+/// [SourceId]: sway_types::SourceId
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct TokenIdent {
+    pub name: String,
+    pub range: Range,
+    pub path: Option<PathBuf>,
```
sdankel, line 188: Shouldn't an ident always have a path?

sway-lsp/src/core/token.rs
```diff
@@ -162,22 +163,60 @@ impl Token {
...
+    pub is_raw_ident: bool,
+}
+
+impl TokenIdent {
+    pub fn new(ident: &Ident, se: &SourceEngine) -> Self {
+        let path = ident
+            .span()
+            .source_id()
+            .map(|source_id| se.get_path(source_id));
```
sdankel, line 197: Consider throwing an error here

sway-lsp/src/core/token.rs
```diff
@@ -247,15 +281,8 @@ pub fn type_info_to_symbol_kind(
...
-    let start_line = start.0 as u32 - 1;
-    let start_character = start.1 as u32 - 1;
-
-    let end_line = end.0 as u32 - 1;
-    let end_character = end.1 as u32 - 1;
-
     Range {
-        start: Position::new(start_line, start_character),
-        end: Position::new(end_line, end_character),
+        start: Position::new(start.0 as u32 - 1, start.1 as u32),
```
sdankel, line 285: Just to be sure I understand, we are shifting the character by 1 to be 0-indexed in line with the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range)?

sway-types/src/span.rs
```diff
@@ -11,59 +11,27 @@ lazy_static! {
...
-impl Position {
-    pub fn new(input: Arc<str>, pos: usize) -> Option<Position> {
-        input.clone().get(pos..).map(|_| Position { input, pos })
+impl<'a> Position<'a> {
+    pub fn new(input: &'a str, pos: usize) -> Option<Position<'a>> {
+        input.get(pos..).map(|_| Position { input, pos })
     }
+
     #[inline]
     pub fn line_col(&self) -> (usize, usize) {
```
sdankel, line 25: This function actually outputs different results now, right? Char is shifted back by 1. Should we write a test for this function? I'm surprised a change like that wouldn't break things outside of LSP.

sdankel (commented):
Awesome results! I left some preliminary comments/questions, and will spend more time testing this locally.

sway-lsp/src/core/token.rs
JoshuaBatty: Yeah I believe we should. I don't think it should be unreasonable to expect that all tokens in the AST's have a path. In another issue i'm running into errors where i have the following

```
std::module::MyType
```

The `std` token in this instance has no `SourceId`. This means I have no way of allowing go_to or hover on this token as the path needs to be known. I would be in favour of have a test in the compiler that makes sure all tokens that make it into the AST have a path. 

Keen to hear compiler team thoughts on this. cc @IGI-111 @tritao

sway-lsp/src/core/token.rs
JoshuaBatty: Ideally `source_id()` wouldn't return an `Option` as outlined above. This should be an error caught in a CI test so upstream tools consuming the AST's always have definite paths to work with.

sway-lsp/src/capabilities/inlay_hints.rs
JoshuaBatty: sure is :)

sway-lsp/src/capabilities/hover/mod.rs
JoshuaBatty: Yep that makes more sense, I change it now.

JoshuaBatty:
> What do you think about the `Position` type taking a `SourceId` instead? Would that provide any benefits or just complicate things for you?

I'm not sure I understand the proposed benefit of doing that. One of the goals of this work is to get away from the `Position` and `SourceId` types as quickly as possible. This way we only need to compute the `Range` once. We also eliminate the need to pass around the source engine throughout the server to get the required `PathBuf`.

Would be keen to hear your thoughts on if you see a better way forward though!

sway-lsp/src/core/token.rs
JoshuaBatty: Yep that's correct. We've been doing this in this function before hand, i've removed the variable allocations now.

sway-lsp/src/capabilities/rename.rs
JoshuaBatty: I guess the idea behind this function is to make sure that we don't allow renaming any tokens that are defined outside of the users project. So types from external libraries don't get renamed. The path in TokenIdent always points to temp memory so that it's in sync with the clients edits. 

This change here just removes the need to call into the source engine to retrieve the path since we have already extracted it in the traversal stage.

sway-types/src/span.rs
JoshuaBatty: Thanks for the suggestion @xunilrj. This version breaks all the LSP tests unfortunately and the numbers seem to be off by quite a bit. Perhaps we can look at modifying this function in a follow up PR..

JoshuaBatty:
> Maybe we should have a data struct `offset -> line number` somewhere? `crates.io` has some interval trees. Not sure which one would be the best fit.

Yeah that could be interesting! Ping me if you come across anything worth looking into

sway-lsp/src/capabilities/document_symbol.rs
JoshuaBatty: I left a doc comment description that explains the differences between the `sway_types::Ident` the the `TokenIdent`. Since this can be pulled up when hovering over the type, and considering it's used throughout the server, maybe that's ok?

sway-types/src/span.rs
JoshuaBatty: You're right good catch. I've changed it back so it outputs the same results as before and then subtracting 1 in the server instead.

sway-lsp/src/core/token.rs
JoshuaBatty: Have been thinking a bit more on this over the weekend. In some situations it actually does make sense for a node to exist in the AST that contains no span, and therefor no path. Namely, any token that is either a keyword or an intrinsic.

Keywords exist in the `LexedProgram` and intrinsics make it to the `ParseProgram` AST. So we already don't expect these tokens to have type definitions from a location. It feels like we should continue with having path  being an `Option<PathBuf>` for this reason. 

However, I would be in favour on still testing that all tokens that make it to the `TypedProgram` AST do in fact contain a valid path by checking if `span.source_id.is_some()`.

kayagokalp (approved):
utACK 🚀

sdankel (commented):
Tested this out in the sway repo and it feels blazingly fast. I didn't find any bugs introduced by this PR. Well done!

PR #4965, opened 2023-08-16
Improve warnings in `match` expressions

### LINKED ISSUE #3949: Emit a warning when all the variants are covered but the user adds a redundant wildcard pattern
Nice!  Are we emitting a warning when all the variants are covered but the user still adds a (redundant) wildcard pattern?  I'm pretty sure Rust does this.

Edit: I just checked, and we don't.  I think this should probably be an issue?

_Originally posted by @otrho in https://github.com/FuelLabs/sway/issues/3790#issuecomment-1407825975_

ironcev:
## Description

At the moment, we have just one warning message in `match` expressions: "This match arm is unreachable."

This PR:
- enriches the existing warning with more contextual information.
- gives specific warning for redundant catch-all last arm as proposed in #3949.
- gives specific warning for catch-all arm preceding other arms; this can easily happen by chance in more complex patterns, e.g., when confusing variable `A` for enum variant `Enum::A`.

What we still want to do is give a hint to user when `A` is (very likely) used instead of `Enum::A`. This will be a part of a separate PR.

The structure of the messages still follow the Elm approach as explained in the [Expressive Diagnostics RFC](https://github.com/FuelLabs/sway-rfcs/pull/30#issuecomment-1682280094). Removing the _issue_ from the title and simplifying the wording will be a part of a different PR.

Closes #3949.

## Demo

Unreachable middle arm

Unreachable last arm

Interior catch-all arm makes arms below it unreachable

Redundant last catch-all arm

## Checklist


ironcev:
@FuelLabs/tooling This PR opens an interesting new case when we are talking about the code that is faded out in IDEs (`DiagnosticTag::UNNECESSARY`).

Up to now, the [`get_warning_diagnostic_tags`](https://github.com/FuelLabs/sway/blob/394ee40eda2f9d5f85253c42a30fa920681c78eb/sway-lsp/src/capabilities/diagnostic.rs#L45) assumes that the span connected to the `CompileWarning` is the span of dead code.

In the case of the `Warning::MatchExpressionCatchAllArmMakesArmsBelowItUnreachable` warning, the span points to the code that makes the code below it dead.

I still see that it makes sense to fade out the dead code, but that's not the code that the warning span points to. The dead code is the part of the warning enum, the `following_arms` span. But showing it as dead would require a different handling as it is now, e.g., passing an information about the span pointing to dead code, which is by default the `CompileWarning` span, but it doesn't have to be.

ironcev:
For the record, in the case of a catch-all arm in the middle, Rust analyzer shows all the unreachable arms below as unreachable, means potentially several warnings. It does point, though, in a hint, to the catch-all arm that causes the issue. This is what I saw differently and thus implemented other way around. Having a catch-all arm somewhere in the middle of a catch expression is actually the issue. Having the arms below unreachable is the consequence.

Here is how it looks like it Rust:

image
image

sway-core/src/semantic_analysis/ast_node/expression/match_expression/typed/typed_scrutinee.rs
```diff
@@ -104,6 +104,44 @@ impl ty::TyScrutinee {
             Scrutinee::Error { err, .. } => Err(err),
         }
     }
+
+    /// Returns true if the `TyScrutinee` consists only of catch-all scrutinee variants, recursively.
+    /// Catch-all varaints are .., _, and variables.
```
xunilrj, line 109: small typo here: `varaints`

sway-core/src/semantic_analysis/ast_node/expression/match_expression/typed/typed_scrutinee.rs
```diff
@@ -104,6 +104,44 @@ impl ty::TyScrutinee {
...
+    /// E.g.: (_, x, Point { .. })
+    /// A catch-all scrutinee matches all the values of the corresponding type.
+    pub(crate) fn is_catch_all(&self) -> bool {
+        match &self.variant {
+            ty::TyScrutineeVariant::CatchAll => true,
+            ty::TyScrutineeVariant::Variable(_) => true,
+            ty::TyScrutineeVariant::Literal(_) => false,
+            ty::TyScrutineeVariant::Constant { .. } => false,
+            ty::TyScrutineeVariant::StructScrutinee { fields, .. } => {
+                let mut is_catch_all_struct_pattern = true;
```
xunilrj, line 119: Is this the same as

```rust
fields.iter().filter_map(|x| x.scrutinee.as_ref()).all(|x| x.is_catch_all())
```

sway-core/src/semantic_analysis/ast_node/expression/match_expression/typed/typed_scrutinee.rs
ironcev: Ups! Spell checker extension just installed, no more typos from nwo on :-)

xunilrj:
This is brilliant, but I am wondering if we can catch all cases with just one error

```rust
 MatchExpressionUnreachableArm {
        match_value: Span,
        preceding_arms: Vec<Span>,
        unreachable_arms: Vec<Span>
    },
```

Because I am not sure the value added by `is_last_arm: bool`, `following_arms_count: usize`

Regarding these multi-line errors, I prefer when texts with same color make a readable sentence. But this is very personal so I leave to your taste which is better.

So for example (ignore the actual text)

1 catch-all, and 1 unreachable
```rust
a => ...,
^ This arm is making... (in yellow)
b => ...
^ ... this arm unreachable  (in yellow)
```

1 catch-all, and n unreachable
```rust
a => ...,
^ This arm is making ... (in yellow)
b => ...
^ ... this arm ...  (in yellow)
c => ...
^ ... and this arm unreachable  (in yellow)
```

n catch-all, and n unreachable
```rust
true => ...,
^ This arm ... (in yellow)
false => ...,
^ and this arm is making ... (in yellow)
b => ...
^ ... this arm ...  (in yellow)
c => ...
^ ... and this arm unreachable  (in yellow)
```

The matched objection blue is a nice touch.

Regarding the help maybe we should propose something? "Consider moving the "Struct { .. }", which is a catch-all arm, to the last position."

sway-core/src/semantic_analysis/ast_node/expression/match_expression/typed/typed_scrutinee.rs
ironcev: Thanks a lot for this, the code is so much simpler and nicer now! I originally conveniently took over the whole matching code from [here](https://github.com/FuelLabs/sway/blob/04a597093e7441898933dd412b8e4dc6ac860cd3/sway-core/src/semantic_analysis/ast_node/expression/match_expression/analysis/pattern.rs#L119) and then pretty brainlessly got caught in the original structure.

ironcev:
> Regarding these multi-line errors, I prefer when texts with same color make a readable sentence. But this is very personal so I leave to your taste which is better.

I like your proposal a lot. I am also constantly thinking if the examples I made so far are too wordy and if they can be made more fluent, as you proposed. Also, I need to purposely step away from the potential limitations that [annotate-snippets](https://crates.io/crates/annotate-snippets) is potentially imposing. If the rendering library does not fit the need we should replace it and not be constrained by it.

Also, I'am trying to follow the example of [Elm](https://elm-lang.org/news/compiler-errors-for-humans) which is perhaps going a bit too far:
image

Anyhow, for this PR, let's leave the wording as it is, and I'll copy your feedback over to the [Expressive Diagnostics RFC](https://github.com/FuelLabs/sway-rfcs/pull/30). Let's agree on the structure of the messages there. And as I said, the fluency that comes from your proposal I like a lot.

ironcev:
> Because I am not sure the value added by is_last_arm: bool, following_arms_count: usize

`is_last_arm` has been added to address #3949 which I agree should be treated a bit differently. It's a common mistake. One starts with "_" at the end, arms get extended "_" is not  needed any more.

`following_arms_count` - good spot :-) It was added to solve the plural issue when building the sentence "This arm", "These arms" and once I put the number in, well, it got used... I also do not see much value in displaying it, but also not in removing it.

xunilrj (dismissed):

ironcev:
Based on the feedback gotten from @sdankel and @JoshuaBatty I'll remove the `Warning::MatchExpressionCatchAllArmMakesArmsBelowItUnreachable` warning and keep just one `Warning::MatchExpressionUnreachableArm` with the appropriate message pointing to the catch-all arm in the middle as the cause of the issue.

ironcev:
@sdankel @xunilrj @JoshuaBatty can you please re-review the PR.

The `Warning::MatchExpressionCatchAllArmMakesArmsBelowItUnreachable` warning is removed. 

The messages still follow the Elm approach as explained in the Expressive Diagnostics RFC.
Removing the _issue_ from the title and simplifying the wording etc. will be a part of a different PR.

sdankel (commented):
The test cases all look fine, except I noticed the diagnostics on this file are very messed up. It seems to be returning the diagnostics from the library files for the main.sw file.

image

This is likely a bug on the LSP side. Opened: https://github.com/FuelLabs/sway/issues/5027

sway-error/src/warning.rs
```diff
@@ -84,7 +84,14 @@ pub enum Warning {
     DeadStorageDeclarationForFunction {
         unneeded_attrib: String,
     },
-    MatchExpressionUnreachableArm,
+    MatchExpressionUnreachableArm {
+        match_value: Span,
+        preceding_arms: Span,
```
sdankel, line 89: Should this be `Vec<Span>`?

sway-error/src/warning.rs
```diff
@@ -84,7 +84,14 @@ pub enum Warning {
...
+        preceding_arm_is_catch_all: bool,
```
sdankel, line 90: Perhaps a name like `follows_catch_all_arm` would be more clear

sdankel (approved):
Just a couple nits/comments

sway-error/src/warning.rs
ironcev: It could be I also thought of it. But since the overall span is used in a few messages I wanted to spare the (one day hopefully macro-based) diagnostic definition of calling `Span::join_all()` every time without having any other benefit of knowing every part of the joint span.

That was the reason I placed the "burden" of getting the joint span to the client.

Interesting question will be how similar situations are handled in general throughout the errors/warnings. Perhaps we come up with a different general guidelines, but so far I would leave it as it is.

sway-error/src/warning.rs
ironcev: It's definitely more clear! I'll rename it in the upcoming PR.

ironcev: Hmm, thinking of `Either` right now :thinking:  That would actually properly solve both comments of yours: either preceding arm*s* or a single catch all arm, and no need for additional boolean flag... I'll think of it.

JoshuaBatty (approved):
utACK, nice work

PR #4972, opened 2023-08-17
feat: forc-client points empty accounts to beta-4 faucet

kayagokalp:
## Description
closes #4862.

waiting for #4905 to be merged first.

With this PR, we are now pointing empty accounts to beta-4 faucet with their address already inserted so that people can click on it and get funded.

```console
Please provide the password of your encrypted wallet vault at "/Users/kayagokalp/.fuel/wallets/.wallet": 
Error: Your wallet does not have any funds to pay for the transaction.

If you are interacting with a testnet consider using the faucet.
-> beta-4 network faucet: https://faucet-beta-4.fuel.network/?address=fuel1l6su7ldka75ntmz7vx05tfznshxjpa4ftt4p537vl2ck45a5p5tsmwjevm
If you are interacting with a local node, consider providing a chainConfig which funds your account.
```

PR #4974, opened 2023-08-17
feat: add beta-4 target, and point --testnet to beta-4 network for forc-deploy

### LINKED ISSUE #4973: add beta-4 target/testnet to forc-client with meaningful default tx params
beta-4 is just around the corner and we should be able to point to that easily as we can point to beta-3 and beta-2

kayagokalp:
## Description
closes #4973.

waiting for #4972 to be merged first.

With this PR, we are now adding `--target beta-4` and alias `--testnet` to `--target beta-4` so that users can deploy to beta-4 with just providing `--tesnet` flag.

```console
forc-deploy --testnet
 Compiling library core (/Users/kayagokalp/.forc/git/checkouts/std-9be0d6062747ea7/04a597093e7441898933dd412b8e4dc6ac860cd3/sway-lib-core)
 Compiling library std (git+https://github.com/fuellabs/sway?tag=v0.44.1#04a597093e7441898933dd412b8e4dc6ac860cd3)
 Compiling contract deploy-to-beta-4 (/Users/kayagokalp/fuel/test_projects/deploy-to-beta-4)
  Finished debug in 4.57878525s
  contract deploy-to-beta-4
      Bytecode size: 68 bytes

Please provide the password of your encrypted wallet vault at "/Users/kayagokalp/.fuel/wallets/.wallet": 

Account 0 -- fuel18caanqmumttfnm8qp0eq7u9yluydxtqmzuaqtzdjlsww5t2jmg9skutn8n:
  Asset ID                                                           Amount
  0000000000000000000000000000000000000000000000000000000000000000 499999960

Please provide the index of account to use for signing: 0
Do you agree to sign this transaction with fuel18caanqmumttfnm8qp0eq7u9yluydxtqmzuaqtzdjlsww5t2jmg9skutn8n? [y/N]: y


Contract deploy-to-beta-4 Deployed!

Network: beta-4.fuel.network/graphql
Contract ID: 0x7e59d984c57b01ef3ffb873781e938fb389f57e14fa4553c54f730a54a6016ec
Deployed in block 0x7d8238d863ece8d3dfb62eae9341ea66b42545b4edcc41265f1f183394d5d717
```


kayagokalp:
We should point this PR to master once we have #4972 merged.

Voxelot:
In the future is there a way we could dynamically update this via a config file we host somewhere (ie. github) instead of needing to make new releases of forc?

forc-plugins/forc-client/src/util/mod.rs
```diff
@@ -40,6 +42,8 @@ impl FromStr for Target {
             Ok(Target::Beta2)
         } else if s == "beta-3" {
             Ok(Target::Beta3)
+        } else if s == "beta-4" {
+            Ok(Target::Beta4)
         } else {
             anyhow::bail!(
                 "invalid testnet name provided. Possible values are 'beta-2', 'beta-3', 'latest'."
```
sdankel, line 49:
```suggestion
                "invalid testnet name provided. Possible values are 'beta-2', 'beta-3', 'beta-4', 'latest'."
```

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -56,20 +56,26 @@ By default `--default-signer` flag would sign your transactions with the followi
...
-While using `forc-deploy` or `forc-run` to interact with the testnet you need to pass the testnet end point with `--node-url`
+To interact with the latest testnet you can pass `--testnet` flag. Passing this flag will send transactions crafted by `forc-deploy` to `beta-4` testnet.
 
 ```sh
-forc-deploy --node-url https://beta-3.fuel.network/graphql
+forc-deploy --testnet
 ```
 
-Since deploying and running projects on the testnet cost gas, you will need coins to pay for them. You can get some using the [testnet faucet](https://faucet-beta-3.fuel.network/).
+Also it is possible to pass exact node url while using `forc-deploy` or `forc-run` which can be done using `--node-url` flag.
```
sdankel, line 65:
```suggestion
It is also possible to pass the exact node url while using `forc-deploy` or `forc-run` which can be done using `--node-url` flag.
```

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -56,20 +56,26 @@ By default `--default-signer` flag would sign your transactions with the followi
...
+
+```sh
+forc-deploy --node-url https://beta-3.fuel.network/graphql
+```
 
-Also the default value of the "gas price" parameter is 0 for both `forc-deploy` and `forc-run`. Without changing it you will get an error complaining about gas price being too low. While using testnet you can pass `--gas-price 1` to overcome this issue. So a complete command for deploying to the testnet would look like:
+Finally you have `--target` option, which provides useful aliases to old testnets. For example if you want to deploy to `beta-3` you can use:
```
sdankel, line 71:
```suggestion
Another alternative is the `--target` option, which provides useful aliases to all targets. For example if you want to deploy to `beta-3` you can use:
```

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -206,14 +206,16 @@ fn apply_target(command: cmd::Deploy) -> Result<cmd::Deploy> {
         if command.target.is_some() {
             bail!("Both `--testnet` and `--target` were specified: must choose one")
```
sdankel, line 207: Do we also need to check if `--node-url` was supplied? It seems like you should only be able to use 1 of the three flags in a single command.

kayagokalp:
> In the future is there a way we could dynamically update this via a config file we host somewhere (ie. github) instead of needing to make new releases of forc?

Yeah that is definitely a better idea.  Opened #4975.

forc-plugins/forc-client/src/op/deploy.rs
kayagokalp: Yeah that is correct. I did not realized this as this was already in place, fixing it now :)

kayagokalp: 9494cf2 should add the checks for this

forc-plugins/forc-client/src/util/mod.rs
```diff
@@ -4,13 +4,14 @@ pub(crate) mod encode;
...
+use crate::default::{BETA_2_ENDPOINT_URL, BETA_3_ENDPOINT_URL, BETA_4_ENDPOINT_URL, NODE_URL};
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 /// Possible target values that forc-client can interact with.
 pub enum Target {
     Beta2,
     Beta3,
+    Beta4,
     LATEST,
```
sdankel, line 15: This actually points to localhost. Should this be `Local` instead? With `latest` it's unclear that it expects you to be running the local node; it sounds like there is a hosted node running the latest fuel-core code somewhere.

forc-plugins/forc-client/src/util/mod.rs
```diff
@@ -40,9 +42,11 @@ impl FromStr for Target {
             Ok(Target::Beta2)
         } else if s == "beta-3" {
             Ok(Target::Beta3)
+        } else if s == "beta-4" {
+            Ok(Target::Beta4)
         } else {
             anyhow::bail!(
-                "invalid testnet name provided. Possible values are 'beta-2', 'beta-3', 'latest'."
+                "invalid testnet name provided. Possible values are 'beta-2', 'beta-3', 'beta-4', 'latest'."
```
sdankel, line 49: nit: 

```rust
impl FromStr for Target {
    type Err = anyhow::Error;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s {
            "beta-2" => Ok(Target::Beta2),
            "beta-3" => Ok(Target::Beta3),
            "beta-4" => Ok(Target::Beta4),
            "latest" => Ok(Target::LATEST),
            _ => anyhow::bail!(
                "'{s}' is not a valid target name. Possible values: '{}', '{}', '{}', '{}'",
                Target::Beta2,
                Target::Beta3,
                Target::Beta4,
                Target::LATEST
            ),
        }
    }
}

impl Display for Target {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        let s = match self {
            Target::Beta2 => "beta-2",
            Target::Beta3 => "beta-3",
            Target::Beta4 => "beta-4",
            Target::LATEST => "latest",
        };
        write!(f, "{}", s)
    }
}
```

JoshuaBatty (dismissed):
utACK, looks good though.

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -206,14 +206,22 @@ fn apply_target(command: cmd::Deploy) -> Result<cmd::Deploy> {
...
+        if command.node_url.is_some() {
+            bail!("Both `--testnet` and `--node-url` were specified: must choose one")
+        }
+        Some(Target::Beta4)
     } else {
         command.target.clone()
     };
 
     if let Some(target) = target {
+        if command.node_url.is_some() {
```
sdankel, line 218: This should be checked above.

sdankel (commented):
Should we also add tests for beta-4?

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -222,17 +230,28 @@ fn apply_target(command: cmd::Deploy) -> Result<cmd::Deploy> {
...
+                // the gas.limit field.
+                let gas_limit = if command.gas.limit
+                    == fuel_tx::ConsensusParameters::DEFAULT.max_gas_per_tx
+                    && target == cmd::deploy::Target::Beta4
+                {
+                    1
+                } else {
+                    command.gas.limit
+                };
+
```
sdankel, line 243: This seems problematic. What if the user intends to use the default max gas?

forc-plugins/forc-client/src/op/deploy.rs
sdankel: I pushed a commit refactoring this to make it easier to read.

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -199,44 +199,55 @@ pub async fn deploy(command: cmd::Deploy) -> Result<Vec<DeployedContract>> {
...
+            // fuel_tx::ConsensusParameters::DEFAULT.max_gas_per_tx is the default value for
+            // the gas.limit field.
+            let gas_limit = if command.gas.limit
+                == fuel_tx::ConsensusParameters::DEFAULT.max_gas_per_tx
+                && target == cmd::deploy::Target::Beta4
+            {
+                1
+            } else {
+                command.gas.limit
+            };
```
sdankel, line 236: What if the user intends to use the default max gas? Why do we do this only for Beta4?

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -484,4 +503,155 @@ mod test {
...
+            testnet: false,
+            ..Default::default()
+        };
+        let expected = cmd::Deploy {
+            target: Some(Target::Beta3),
+            node_url: Some("beta-3.fuel.network/graphql".to_string()),
+            testnet: false,
+            gas: Gas {
+                price: 1,
+                limit: 100000000,
```
sdankel, line 604: Added some unit tests to show the current behavior - still don't understand the logic behind overriding the gad price and limit for the testnets.

kayagokalp: Thanks so much for adding the tests 💯 I am not sure why this is only needed for beta4 but the default max gas limit returns:

```
Error: Decode error: Custom { kind: Other, error: "Response errors; TransactionGasLimit" }
```

This is not the case for other beta networks.

If i change the default value, beta-3 and beta-2 would break, so I patched it only for beta-4 and if users provide `gas_limit` their `gas_limit` will override our preset value (unless they specified the max value, but again providing max value for gas limit returns the error above). This is not ideal but it enables to deploy without specifying any other flags

forc-plugins/forc-client/src/op/deploy.rs
kayagokalp: https://github.com/FuelLabs/sway/pull/4974#discussion_r1299374887 resolving this to prevent duplicate discussions

forc-plugins/forc-client/src/op/deploy.rs
sdankel: I'm just wondering if there's a more appropriate gas limit than 1. It seems so low. We'd want to make sure the default gas limit to work for users.

kayagokalp: We now ask the test net for min gas price if the user left it as 0 by default. The testnet returns the accepted min gas price and we use that instead of 0 so that the transaction is valid.

We do the same thing for gas_limit if they left as the default max value, we ask the chain the valid max value and use it instead, so that we do not exceed the limit set by the node.

I think that while deploying to test net we can override this as we want the transactions to be valid by default and if the user choses to be providing the extremes we can use the extreme set by the node instead. Maybe I can add a warning to notify them?

kayagokalp: On a second thought modiying this rn to make sure we do not override any user given data. If the user wants to override the default data returned by the node they can. Maybe they would want to test some stuff around that

forc-plugins/forc-tx/src/lib.rs
```diff
@@ -624,8 +622,11 @@ impl TryFrom<Create> for fuel_tx::Create {
             .map(|s| fuel_tx::Witness::from(s.as_bytes()))
             .collect();
         let create = fuel_tx::Transaction::create(
-            create.gas.price,
-            create.gas.limit,
+            create.gas.price.unwrap_or(0),
+            create
+                .gas
+                .limit
+                .unwrap_or(fuel_tx::ConsensusParameters::DEFAULT.max_gas_per_tx),
```
sdankel, line 629: Should we be using the defaults from the node here as well?

sdankel: Actually it seems like forc-tx doesn't specify the node, so there's no way to use the node's defaults

Issue #4958, opened 2023-08-16
Re-write the LSP tests to not rely on `tower-lsp`


JoshuaBatty:
Currently we are using `tower-lsp` to create a mock Client for testing in the language server. 

This is a blocker for moving ahead with the custom event loop work as all of our internal types need to be wrapped in `RwLock` to enable interior mutability. 

In `rust-analyzer` they just rely on the expected messages constructed with the `lsp_types` crate directly. We would just need to construct these types ourselves.

JoshuaBatty:
This was mostly done in #4976 . The only tests left that have been untouched are the notification methods that rely on a tokio blocking thread for compilation.

We have shelved moving ahead with a custom event loop at this time and work started in #5057 will remove the reliance on `RwLock` for our types. Closing this for now.

PR #4976, opened 2023-08-18
Use `lsp_types` directly in `sway-lsp` tests instead of simulating a client with `tower-lsp`

JoshuaBatty:
## Description
This PR removes the mock client <> server interaction to test the language server features in favour of using `lsp_types` directly. There are a few places like `did_change` where we are still using the exisiting framework, this is because we still rely on the `tower-lsp` event loop to synchronise notification messages. This will be tackled with the custom event loop work. Should be easy to just update the remaining tests here within that PR.  I've also kept the publish diagnostics test the same as we still rely on the `Client` from `tower-lsp` for that functionality. 

There were also a few tests here that were unnecessary. Namely, `initialized` `initializes_only_once` `shutdown` `refuses_requests_after_shutdown`. These tests are actually duplicated from `tower-lsp` itself, so there is really no reason we need to run them again in `sway-lsp`.
 
working towards #4958

## Checklist


sdankel (approved):
Nice to have this decoupling! LGTM

Issue #4729, opened 2023-07-03
parser recovery: failing on parsing `ItemTraitItem`


JoshuaBatty:
The below code is returning as `ParseErrorKind::ExpectedAnItem` error. This error is causing the `parse` function to bail out of creating a ParseProgram meaning the LSP doesn't have an AST to traverse. 

The compiler should be able to handle an error like this and still be able to continue with creating a parsed and typed AST. 

```sway
contract;

abi MyContract {
    fn test_function() -> bool;
    fssdf
}

impl MyContract for Contract {
    fn test_function() -> bool {
        true
    }
}
```

```
 --> /Users/josh/Documents/rust/fuel_test_projects/no_error_lines/src/main.sw:7:5

6 |     fn test_function() -> bool;
7 |     fssdf
  |     ^ Expected an item.
```

PR #4979, opened 2023-08-18
Parser recovery inside trait

xunilrj:
## Description

Closes https://github.com/FuelLabs/sway/issues/4729.

The trick part of this PR is that `TraitItems` were being parsed as `(TraitItem, SemicolonToken)`. The problem is that to recover from an error we call the `error` method of the `Parser` trait. And it is very hard to come up with a generic implementation for the tuple. Which type of tuple do we call its `error` method, and how do we return a valid tuple in all cases? Impossible.

So that is why I moved the `SemicolonToken` to inside the `TraitItem`. Another benefit is that we can recover from missing semicolons with nice error messages.

## Checklist


JoshuaBatty (approved):
thanks, seems like a reasonable approach.

Issue #4983, opened 2023-08-18
Code lens showing up in weird places


sdankel:
![Screenshot 2023-08-18 at 3 56 19 PM](https://github.com/FuelLabs/sway/assets/47993817/a9c6726e-d639-4ea7-8dac-af12a8b24ee5)

PR #4984, opened 2023-08-19
Fix runnables code lens in multi-file sway projects

sdankel:
## Description

Closes https://github.com/FuelLabs/sway/issues/4983

The issue was that we were returning all of the runnables for the workspace, rather than just the file. We were also calling `get_range_from_span` for every runnable at request time, which @JoshuaBatty discovered is very costly. 

- The runnables map is now mapping [file path] => [vector of runnables for that file]
- Rather than iterating over all of the runnables in the session, we just fetch the vector of runnables indexed by the file path.
- Runnables are now storing the range rather than the span, so we don't need to convert it at request time.
- Added another test that expects _no_ runnables to be returned for a file in a workspace that has runnables in other files. This test was failing before the change. Previously, the runnables test was running on test files from the e2e directory, so I copied that into the LSP fixtures.
- Moved the code_lens logic into a capability and updated the bencher

## Checklist


sway-lsp/src/capabilities/runnable.rs
```diff
@@ -1,21 +1,19 @@
...
 use lsp_types::{Command, Range};
 use serde_json::{json, Value};
 use sway_core::language::parsed::TreeType;
-use sway_types::Span;
 
 #[derive(Debug, Eq, PartialEq, Clone)]
 pub struct RunnableMainFn {
-    /// The location in the file where the runnable button should be displayed
-    pub span: Span,
+    // The range in the file where the runnable button should be displayed.
```
JoshuaBatty, line 7: Can we keep this as a doc comment?

sway-lsp/src/capabilities/runnable.rs
```diff
@@ -1,21 +1,19 @@
...
+    pub range: Range,
     /// The program kind of the current file
     pub tree_type: TreeType,
 }
 
 #[derive(Debug, Eq, PartialEq, Clone)]
 pub struct RunnableTestFn {
-    /// The location in the file where the runnable button should be displayed.
-    pub span: Span,
+    // The range in the file where the runnable button should be displayed.
```
JoshuaBatty, line 15: And here.

sway-lsp/src/core/session.rs
```diff
@@ -341,20 +345,29 @@ impl Session {
...
+        source_engine: &SourceEngine,
+    ) {
         // Insert runnable test functions.
         for (decl, _) in typed_program.test_fns(decl_engine) {
             // Get the span of the first attribute if it exists, otherwise use the span of the function name.
             let span = decl
                 .attributes
                 .first()
                 .map_or_else(|| decl.name.span(), |(_, attr)| attr.span.clone());
+            let path = source_engine.get_path(span.source_id().unwrap());
```
JoshuaBatty, line 361: Can we remove the `unwrap()`?

sway-lsp/src/core/session.rs
```diff
@@ -363,11 +376,15 @@ impl Session {
         } = typed_program.kind
         {
             let span = main_function.name.span();
+            let path = source_engine.get_path(span.source_id().unwrap());
```
JoshuaBatty, line 379: also this `unwrap()`

JoshuaBatty:
Nice optimisation! just some small nits.

sway-lsp/src/core/session.rs
```diff
@@ -332,20 +336,31 @@ impl Session {
...
                 .first()
                 .map_or_else(|| decl.name.span(), |(_, attr)| attr.span.clone());
-            let runnable = Box::new(RunnableTestFn {
-                span,
-                tree_type: typed_program.kind.tree_type(),
-                test_name: Some(decl.name.to_string()),
-            });
-            self.runnables.insert(runnable.span().clone(), runnable);
+            if let Some(source_id) = span.source_id() {
+                let path = source_engine.get_path(source_id);
```
JoshuaBatty, line 353: Now that #4961 is merged, we should have an `Option<PathBuf>` to use in the `TokenIdent`. This also means we don't need to pass the source engine into this function now.

sdankel: Hmm, I'm not sure how to get the TokenIdent or even the file path from this type without using the source engine. Do you have any idea?

JoshuaBatty: nevermind you're correct, we still need the source engine up until the point of creating the type.

JoshuaBatty (approved):

eureka-cpu (approved):

PR #4991, opened 2023-08-21
Add beta-4 target to forc, refactor with tests, and UX improvements

sdankel:
## Description

Closes https://github.com/FuelLabs/sway/pull/4974

Added the beta-4 target and updated documentation (from https://github.com/FuelLabs/sway/pull/4974)

Refactored part of forc-deploy and added unit tests.
- No longer mutating the Command, so the Command is always whatever the user gave us
- Using the default values from the node itself via API call, rather than hardcoded constants
- The values of `Gas` are now optional, so we only override them when the user hasn't specified anything

Other changes:
- deployment works using the urls without `/graphql` prefixes, so I updated the constants to remove the prefix. This is shown to the user and it looks cleaner without it.
- Added coloring and consistent styling for errors and warnings to printed to the console. It looks like this now (previously no color)
- `deploy`, `run`, and `submit` all now have the same options for specifying the node (node_url, testnet, and target), capture in the `TargetNode` struct. They use the same helper functions to extract the node url and determine gas limit & price.

image

## Checklist


JoshuaBatty (dismissed):

xunilrj (approved):

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -197,46 +202,43 @@ pub async fn deploy(command: cmd::Deploy) -> Result<Vec<DeployedContract>> {
...
-                        price: gas_price,
-                        ..command.gas
-                    },
-                    node_url: target_url,
-                    ..command
-                })
-            }
-            cmd::deploy::Target::LATEST => Ok(command),
-        }
+    let limit = if command.gas.limit == 0 {
```
kayagokalp, line 214:
```suggestion
    let limit = if command.gas.limit == fuel_tx::ConsensusParameters::DEFAULT.max_gas_per_tx {
```

command.gas.limit is `fuel_tx::ConsensusParameters::DEFAULT.max_gas_per_tx` by default. So if the user uses `forc-deploy --testnet` currently that if is not checked and following error is returned:

```console
error: Decode error: Custom { kind: Other, error: "Response errors; TransactionGasLimit" }
```

I think we are good to go with only changing that 0 to `fuel_tx::ConsensusParameters::DEFAULT.max_gas_per_tx `, later on once we clean forc-tx stuff too this would be cleaned as well

kayagokalp (changes requested):
Looks really good 🙌 Thanks a lot for the help :)  Just a single change requested so that we can deploy with `forc-deploy --testnet`

forc-plugins/forc-client/src/op/deploy.rs
sdankel: I'm making 0 the default so that we can check if it's 0 to override it. Otherwise, if `fuel_tx::ConsensusParameters::DEFAULT.max_gas_per_tx` is the default, and the user actually wants to use that value, we will not know whether the user provided it or not, so we will override incorrectly. In the previous code, it was impossible for the user to actually run it with `fuel_tx::ConsensusParameters::DEFAULT.max_gas_per_tx`

kayagokalp: 👌 sounds good to me for getting things working, but ideally I feel like they should be optional without default values and if user did not provide a value for them we should get them from the node.

Here is where the gas limit default value is specified:
https://github.com/FuelLabs/sway/blob/149b65726298b008b612612368e2b3e0e9b5997c/forc-plugins/forc-tx/src/lib.rs#L108-L109

sdankel: Good catch, I forgot to set the default to 0. I think it's better to default to 0 so that users can still specify `fuel_tx::ConsensusParameters::DEFAULT.max_gas_per_tx ` if they want, and we won't override it.

kayagokalp:
Ahh looks like https://github.com/FuelLabs/sway-libs/pull/185 changed folder structure in sway-libs. Fixing it in a separate PR.

kayagokalp:
Fix: https://github.com/FuelLabs/sway/pull/4998

kayagokalp (dismissed):
LGTM :) waiting docs fix to merge so that CI unblocks: https://github.com/FuelLabs/sway/pull/4998

forc-plugins/forc-tx/src/lib.rs
```diff
@@ -624,8 +622,11 @@ impl TryFrom<Create> for fuel_tx::Create {
             .map(|s| fuel_tx::Witness::from(s.as_bytes()))
             .collect();
         let create = fuel_tx::Transaction::create(
-            create.gas.price,
-            create.gas.limit,
+            create.gas.price.unwrap_or_default(),
+            create
+                .gas
+                .limit
+                .unwrap_or(fuel_tx::ConsensusParameters::DEFAULT.max_gas_per_tx),
```
sdankel, line 629: Since this command doesn't have a node_url, we are unable to fetch defaults from the node.

JoshuaBatty (approved):

kayagokalp (approved):
LGTM

PR #4992, opened 2023-08-21
Add Appendix A: Keywords to Sway book

### LINKED ISSUE #4932: Add appendix to the Sway book: Keywords
The Rust book has an [appendix that details language keywords](https://doc.rust-lang.org/book/appendix-01-keywords.html) in a table with an example of each keyword's syntax and a description of its use. I think new users of the Sway language could benefit from the same type of index.

andrewvious:
## Description
closes #4932 

## Checklist


docs/book/src/appendix/keywords.md
```diff
@@ -0,0 +1,89 @@
...
+identifiers. Identifiers are names of functions, variables,
+parameters, modules, constants, attributes, types or
+traits, ect.
+
+## [Keywords Currently in Use](#keywords-currently-in-use)
+
+The following is a list of keywords currently in use, with their
+functionality described.
+
+- `as` - perform primitive casting, or rename items in `use` statements
```
Braqzen, line 14: I'm not as up-to-date on new language features so I haven't heard of `as` being in `Sway`. Can you point me at an example which uses this?

If there isn't one then this should be removed.

docs/book/src/appendix/keywords.md
```diff
@@ -0,0 +1,89 @@
...
+- [`abi`](https://fuellabs.github.io/sway/master/book/sway-program-types/smart_contracts.html#the-abi-declaration) - defines a smart contract ABI in a syntactcally similar way to traits
+- [`break`](https://fuellabs.github.io/sway/v0.44.0/book/basics/control_flow.html#break-and-continue) - exit a loop immediately
+- [`const`](https://fuellabs.github.io/sway/v0.44.0/book/basics/constants.html) - define constant items
+- [`continue`](https://fuellabs.github.io/sway/v0.44.0/book/basics/control_flow.html#break-and-continue) - continue to the next loop iteration
+- `dep` - call dependencies outside of the current `Forc.toml` scope
```
Braqzen, line 19: Afaik `dep` has been deprecated in favour of `mod`. There is only 1 place we use it in the repo which is surprising because I'm almost certain that no longer works. I believe that 5 month old file has been forgotten.

I'll double check internally but I'm pretty sure this should be removed.

docs/book/src/appendix/keywords.md
```diff
@@ -0,0 +1,89 @@
...
+- `else` - used in conjunction with `if` conditions for control flow constructs
+- [`enum`](https://fuellabs.github.io/sway/v0.44.0/book/basics/structs_tuples_and_enums.html#enums) - define an enumeration
+- `false` - Boolean false literal
+- [`fn`](https://fuellabs.github.io/sway/master/book/basics/functions.html)- define a function or the function pointer type
+- [`if`](https://fuellabs.github.io/sway/v0.44.0/book/basics/control_flow.html#if-expressions) - branch based on the result of a conditional expression
+- `impl` - implement inherent or trait functionality
```
Braqzen, line 25: inherent?

docs/book/src/appendix/keywords.md
```diff
@@ -0,0 +1,89 @@
...
+- `Self` - a type alias for the type we are defining or implementing
+- `self` - method subject
+- [`struct`](https://fuellabs.github.io/sway/v0.44.0/book/basics/structs_tuples_and_enums.html#structs) - define a structure
+- [`trait`](https://fuellabs.github.io/sway/master/book/advanced/traits.html#declaring-a-trait) - define a trait
+- `true` - Boolean true literal
+- `use` - bring symbols into scope
+- `where` - specifies traits for generic types
+- [`while`](https://fuellabs.github.io/sway/v0.44.0/book/basics/control_flow.html#while) - loop conditionally based on the result of an expression
+
+## [Keywords Reserved for Possible Future Use](#keywords-reserved-for-possible-future-use)
```
Braqzen, line 42: Have you checked that these already cannot be used or are you adding these in apprehension of future language expansion?

If the latter then remove this section.

Braqzen (commented):
1. I have not checked all the links to ensure that they point to where they ought to
2. Unfortunately, I really do not have the bandwidth to make individual suggestions for each definition but I would re-write most / almost all of the definitions to be more accurate / concise. Most aren't bad but ought to be updated imo

docs/book/src/appendix/keywords.md
eureka-cpu: `impl TypeName {}` is the inherent syntax

docs/book/src/appendix/keywords.md
eureka-cpu: Pretty sure we at least have 
```sway
use type::a as alias_name
```

andrewvious:
> 1. I have not checked all the links to ensure that they point to where they ought to
> 2. Unfortunately, I really do not have the bandwidth to make individual suggestions for each definition but I would re-write most / almost all of the definitions to be more accurate / concise. Most aren't bad but ought to be updated imo

Appreciate the review, all the links I found should check out! As far as the definitions I tried to make it as boiled down as possible, using rust's appendix as reference (https://doc.rust-lang.org/book/appendix-01-keywords.html). However, I can change any you feel need to be updated.

docs/book/src/appendix/keywords.md
andrewvious: > Pretty sure we at least have
> 
> ```sway
> use type::a as alias_name
> ```

Just tested and this is the only way it can be used. I went ahead and kept it listed but gave an example of its only use.

eureka-cpu: This code compiles:
```sway
script;

use std::logging::log as logger; // alias here

struct U128 {
    a: u64,
    b: u64,
}

fn main() {
    let x = U128 { a: 0, b: 5 };
    logger(x);
}

```

eureka-cpu (changes requested):
Hey @andrewvious really great work, checked all the links and they're correct.

I've requested just one change and I think this should be good to go.

eureka-cpu (approved):
LGTM nice job

sarahschwartz (changes requested):
Thank you for your contribution! Can you update the links so that they are relative instead of absolute urls? This way the versions won't need to be updated.

docs/book/src/appendix/keywords.md
```diff
@@ -0,0 +1,88 @@
+# [Appendix A: Keywords](#appendix-a-keywords)
```
sarahschwartz, line 1: You can remove the link here and for other headers, as they are automatically applied.

andrewvious:
> Thank you for your contribution! Can you update the links to the sway book so that they are all relative instead of absolute urls? This way the versions won't need to be updated.

I went ahead and set all the urls to the relative path in the book, but lost some precision the links had. Let me know if this works, or if there is a better way of doing it!

Braqzen:
> > Thank you for your contribution! Can you update the links to the sway book so that they are all relative instead of absolute urls? This way the versions won't need to be updated.
> 
> I went ahead and set all the urls to the relative path in the book, but lost some precision the links had. Let me know if this works, or if there is a better way of doing it!

The reference seems to always point to a page instead of a specific heading. I don't recall if I did that because I couldn't link it precisely or because each section was meant to be a small page. If the page is too full then changing that page will inevitably lead to headings being changed and links being broken so it's better to structure content per page in a digestible manner.

I haven't tested the links but I wonder if we should be mixing the book and reference. The reference was meant to entirely replace the book, and hopefully it will before mainnet.

Braqzen (commented):
Quickly glancing through CI the linter fails because the links start with `/docs` instead of `./docs` (I assume, haven't checked)

docs/book/src/appendix/keywords.md
```diff
@@ -0,0 +1,88 @@
...
+parameters, modules, constants, attributes, types or
+traits, ect.
+
+## Keywords Currently in Use
+
+The following is a list of keywords currently in use, with their
+functionality described.
+
+- `as` - rename items in `use` statements, eg `use type::a as alias_name`
+- [`abi`](./docs/book/src/sway-program-types/smart_contracts.md) - defines a smart contract ABI in a syntactcally similar way to traits
```
sarahschwartz, line 15: The links are relative to their location in the book, so you can replace "./docs/book/src/" with "../"

andrewvious:
> Thanks! I have one more nit: can you move this page into the existing Sway Reference section and remove the Appendix section?

Which you mind further elaborating where inside the reference section you would prefer this to be moved to?
Should I create an Appendix in the Reference section?

Braqzen:
> > Thanks! I have one more nit: can you move this page into the existing Sway Reference section and remove the Appendix section?
> 
> Would you mind further elaborating where inside the reference section you would prefer this to be moved to? Should I create an Appendix in the Reference section?

Move entire section from the book into the reference at the bottom

sarahschwartz:
Number 11 in the navigation of the Sway Book is "Sway Reference". You can move the keywords page into that section, add to the existing index.md with a link, and delete the apendix folder.
sway book

sarahschwartz (approved):

docs/book/src/reference/keywords.md
```diff
@@ -0,0 +1,88 @@
...
+Keywords associated with defining the type of Sway program to compile
+
+- [`contract`](../sway-program-types/smart_contracts.md) - analogous to a deployed API with some database state
+- [`library`](../sway-program-types/libraries.md) - Sway code that defines new common behavior
+- [`predicate`](../sway-program-types/predicates.md) - programs that return a Boolean value and which represent ownership of some resource upon execution to true
+- [`script`](../sway-program-types/scripts.md) - a runnable bytecode on the chain, which executes once to preform a task
+
+### Attribute Keywords
+
+Keywords associated with defining the funcitonallity of attributes
```
eureka-cpu, line 81: Something like `, see [attributes](./attributes.md)`

docs/book/src/reference/keywords.md
```diff
@@ -0,0 +1,88 @@
...
+
+- [`allow`](../reference/attributes.md) - overrides checks that would otherwise result in errors or warnings
+- [`doc`](../reference/attributes.md) - specifies documentation
+- [`inline`](../reference/attributes.md) - suggests that a copy of the attributed function should be placed in the caller, rather than generating code to call the function where it is defined
+- [`payable`](../reference/attributes.md) - implies method is payable for compile time
+- [`storage`](../reference/attributes.md) - declaration that contains a list of stored variables
+- [`test`](../reference/attributes.md) - marks a function to be executed as a test
```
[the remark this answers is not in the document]
sarahschwartz: Good points @eureka-cpu ! You can remove change the links that are in the same reference folder from "../reference/attributes.md" to "./attributes.md", and you can also link the a subheading in that page like this: "./attributes.md#allow"

sarahschwartz (changes requested):

eureka-cpu (approved):
LGTM, nice work

sarahschwartz (approved):

PR #5001, opened 2023-08-23
Parallelize traversing AST's in the language server

### LINKED ISSUE #4959: Parallelize traversing AST's in the language server.

JoshuaBatty:
## Description
This PR introduces using rayon's parallel iterators when traversing the lexed and parsed AST's. I initially extended this to also occur on the typed AST but was running into deadlock issues with how DashMap works. 

closes #4959

| Request Type/Method    | Previous  | New      | Improvement % |
|------------------------|-----------|----------|---------------|
| compile_and_traverse   | 2.7732s   | 2.5632s  | 7.57%         |

**edit**

I've split the compilation and traversal stages so they can be benchmarked independently. Here are the actual traversal performance results.

| Request Type/Method    | Previous  | New      | Improvement % |
|------------------------|-----------|----------|---------------|
| traverse   | 422ms   | 211ms  | 50.0%         |

## Checklist

### DIFF
--- a/sway-lsp/src/core/token_map.rs
+++ b/sway-lsp/src/core/token_map.rs
@@ -17,7 +17,7 @@ pub struct TokenMap(DashMap<TokenIdent, Token>);
 impl TokenMap {
     /// Create a new token map.
     pub fn new() -> TokenMap {
-        TokenMap(DashMap::new())
+        TokenMap(DashMap::with_capacity(2048))
     }
 
     /// Create a custom iterator for the TokenMap.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5903,6 +5903,7 @@ dependencies = [
  "parking_lot 0.12.1",
  "proc-macro2",
  "quote",
+ "rayon",
  "ropey",
  "serde",
  "serde_json",
--- a/sway-lsp/Cargo.toml
+++ b/sway-lsp/Cargo.toml
@@ -19,6 +19,7 @@ notify-debouncer-mini = { version = "0.2.0" }
 parking_lot = "0.12.1"
 proc-macro2 = "1.0.5"
 quote = "1.0.9"
+rayon = "1.5.0"
 ropey = "1.2"
 serde = { version = "1.0", features = ["derive"] }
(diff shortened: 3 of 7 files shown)


xunilrj:
A 5% gain is giving me the impression we are parallelizing too much...
Maybe just per module will yield better results? I think there are too much `par_iter()`...

JoshuaBatty:
> Maybe just per module will yield better results? I think there are too much `par_iter()`

Thanks for the tip, I have been trying with a variety of situations. Only traversing in parallel per module didn't lead to any difference in performance. I'll keep investigating.

sway-lsp/src/core/token_map.rs
```diff
@@ -17,7 +17,7 @@ pub struct TokenMap(DashMap<TokenIdent, Token>);
 impl TokenMap {
     /// Create a new token map.
     pub fn new() -> TokenMap {
-        TokenMap(DashMap::new())
+        TokenMap(DashMap::with_capacity(2048))
```
sdankel, line 20: How did you determine the value for the capacity?

sdankel (approved):
Nice work!

sway-lsp/src/core/token_map.rs
JoshuaBatty: I just chose a large power of two number that seemed like it would be a reasonable default. We would likely also be fine using 1024

JoshuaBatty:
Ok this should be ready now. Pre allocating the DashMap gave us a few extra % performance win as well.

JoshuaBatty:
I think this benchmark is a bit misleading as the compiler performance is bundled with the traversal perf. I'm going to put up a PR that splits compilation and traversal into unique stages so we measure each stage separately.

JoshuaBatty:
This PR is ready to go now.

xunilrj (approved):

cr-fuel (approved):

PR #5028, opened 2023-08-24
Update `msg_sender()` to return sender when input type is `Input::Message`

bitzoic:
## Description

This change reflects an issue experienced by an internal contributor where `msg_sender().unwrap()` would revert when sending a message. Currently, the sender is only returned when the input is of type `Input::Coin`.

## Checklist

### DIFF
--- a/sway-lib-std/src/auth.sw
+++ b/sway-lib-std/src/auth.sw
@@ -6,7 +6,7 @@ use ::contract_id::ContractId;
 use ::identity::Identity;
 use ::option::Option::{self, *};
 use ::result::Result::{self, *};
-use ::inputs::{Input, input_count, input_owner, input_type};
+use ::inputs::{Input, input_count, input_message_sender, input_owner, input_type};
 
 /// The error type used when an `Identity` cannot be determined.
 pub enum AuthError {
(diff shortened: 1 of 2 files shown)


Braqzen (commented):
I believe you should use `input_message_recipient` instead of `input_message_sender` because they will be the owner upon receiving the resource

bitzoic:
> I believe you should use `input_message_recipient` instead of `input_message_sender` because they will be the owner upon receiving the resource

Perhaps I'm misunderstanding, but why would we want the recipient when calling `msg_sender()`?

Braqzen:
> > I believe you should use `input_message_recipient` instead of `input_message_sender` because they will be the owner upon receiving the resource
> 
> Perhaps I'm misunderstanding, but why would we want the recipient when calling `msg_sender()`?

Well, the high level method is called `msg_sender()` but the method being adjusted regards the owner of the inputs. When sent the owner of the message may be the sender field but the recipient will become the owner.

There was more talk of this, which yet again I forget the details, from @segfault-magnet for when and what the sender / recipient represent depending on when the message is constructed iirc.

bitzoic:
Resolved by https://github.com/FuelLabs/sway/pull/6231

PR #5038, opened 2023-08-27
[1645] Enhanced PackageManifest parsing

### LINKED ISSUE #1645: Duplicate or conflicting dependency declarations should produce a warning or error
Currently adding a duplicate or conflicting dependency to `Forc.toml` does not produce any error or warning. The behavior is ambiguous (which declaration to use is not clear) thus we may want to produce an error.

**Examples**
---

**Duplicate entries:**

```
[dependencies]
std = { git = "https://github.com/fuellabs/sway", branch = "master" }
std = { git = "https://github.com/fuellabs/sway", branch = "master" }
```

**Conflicting entries:**

```
[dependencies]
std = { git = "https://github.com/fuellabs/sway", branch = "master" }
std = { git = "https://github.com/fuellabs/sway", branch = "kayagokalp/1620" }
```

cr-fuel:
## Description

Fixes #1645 by checking if there are any duplicate keys in sections. Any duplication of keys will generate a parsing error that will bubble up to the final user with some context on how to fix it.

The following `Forc.toml` file will fail:

```
[project]
authors = ["Cesar"]
entry = "main.sw"
license = "Apache-2.0"
name = "invalid"

[dependencies]
foo = "bar"
foo = "xxx"
```

The error that will be bubble up to the CLI output reads as follow:

```
Error: failed to parse manifest: duplicate 'foo' for key `dependencies` at line 7 column 1.
```

## Checklist


kayagokalp:
Merged master into here to get #5043

forc-pkg/tests/invalid/Forc.toml
```diff
@@ -0,0 +1,3 @@
+[dependencies]
```
sdankel, line 1: nit: consider putting this under `forc-pkg/tests/invalid/duplicate_keys/Forc.toml` in case we add more tests of invalid Forc manifests (which we should!)

forc-pkg/src/manifest.rs
```diff
@@ -1153,6 +1212,18 @@ mod tests {
...
+        assert_eq!(
+            "failed to parse manifest: duplicate 'foo' for key `dependencies` at line 1 column 1."
+                .to_owned(),
+            PackageManifest::from_dir("./tests/invalid")
+                .unwrap_err()
+                .root_cause()
+                .to_string()
+        );
+    }
+
```
sdankel, line 1226: nit: consider using `unwrap` + [should_panic](https://doc.rust-lang.org/book/ch11-01-writing-tests.html#checking-for-panics-with-should_panic) to assert error messages

sdankel (commented):
I'm surprised the toml parser doesn't throw this error, since duplicate keys are invalid toml. [Here](https://github.com/toml-rs/toml/blob/ff859198c77a09587b51137cf53d5b3aacb2c31b/crates/toml_edit/src/parser/state.rs#L116) it looks like it should be throwing. Perhaps it's just because we're on such an old version of `toml`? 

With cargo the error is very clear, and it looks like the [error](https://github.com/toml-rs/toml/blob/ff859198c77a09587b51137cf53d5b3aacb2c31b/crates/toml_edit/src/parser/errors.rs#L298) comes directly from the toml parser. I think we should take the same approach rather than checking for this toml error ourselves.

It would be nice if we could get the error to look more like this:

```
$ cargo build
error: failed to parse manifest at `/Users/sophiedankel/Development/sway/forc-plugins/forc-client/Cargo.toml`

Caused by:
  could not parse input as TOML

Caused by:
  TOML parse error at line 13, column 1
     |
  13 | anyhow = "2"
     | ^
  duplicate key `anyhow` in table `dependencies`
```

forc-pkg/src/manifest.rs
cr-fuel: That is pretty cool, I was not aware of should_panic. I'll use it.

forc-pkg/tests/invalid/Forc.toml
cr-fuel: I agree with you @sdankel. Maybe a feature is missing or upgrading. I will try to get the same error upgrading the library. I'll keep you posted.

forc-pkg/src/manifest.rs
```diff
@@ -494,14 +496,15 @@ impl PackageManifest {
     /// If `core` and `std` are unspecified, `std` will be added to the `dependencies` table
     /// implicitly. In this case, the git tag associated with the version of this crate is used to
     /// specify the pinned commit at which we fetch `std`.
-    pub fn from_file(path: &Path) -> Result<Self> {
+    pub fn from_file<P: AsRef<Path>>(path: P) -> Result<Self> {
```
sdankel, line 499: I'm also curious why this is changed

cr-fuel: This was a minor change, I can roll it back, but it makes easier to call any `Path`. It basically will follow what the std library does whenever it deals with paths and it is easier to call the API, because a `&str` or a `String` will match that requirement. Should I roll it back @sdankel ?

forc-pkg/Cargo.toml
```diff
@@ -17,22 +17,26 @@ forc-tracing = { version = "0.45.0", path = "../forc-tracing" }
...
+serde_ignored = "0.1.9"
 serde_json = "1.0"
+serde_with = "3.3.0"
 sway-core = { version = "0.45.0", path = "../sway-core" }
 sway-error = { version = "0.45.0", path = "../sway-error" }
 sway-types = { version = "0.45.0", path = "../sway-types" }
 sway-utils = { version = "0.45.0", path = "../sway-utils" }
 tar = "0.4.38"
-toml = "0.5"
+toml = { version = "0.7", features = ["parse"] }
```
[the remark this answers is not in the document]
cr-fuel: `0.7` [upgrades to 7.6](https://github.com/FuelLabs/sway/pull/5038/files#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87eR6479-R6490) I could, however to pin the 0.7.6 into the toml file as well. That'd be safer now that I think about it.

forc-pkg/src/manifest.rs
sdankel: No this is fine, thanks for explaining!

Cargo.lock
```diff
@@ -6019,7 +6098,7 @@ dependencies = [
  "rowan",
  "semver",
  "smallvec",
- "toml",
+ "toml 0.5.11",
```
JoshuaBatty, line 6101: might be worth opening an issue to bump `toml` to `0.7` anywhere else it's used in the sway repo.

JoshuaBatty (approved):
Great first PR, looks nice a clean.

sdankel (approved):

Cargo.lock
cr-fuel: We would need to upgrade `mdbook`, `proc-macro-crate`, `taplo` and `fuel-etk-ops`. Can this update be addressed in a different PR?

JoshuaBatty: Yeah let's do that in a follow up PR. Thanks

PR #5045, opened 2023-08-28
Ensure generic values used with a `StorageMap` are hashable

bitzoic:
## Description

Currently, any type may be included as a key value to the `StorageMap<K, V>` type. This presents an issue as the type must be hashable in order to generate the key. 

A where clause has been added to the function definition.

This should open the door for using heap types(`String`, `Bytes`) as key values for the `StorageMap` type however, this will require additional compiler changes.

## Checklist

### DIFF
--- a/sway-lib-std/src/storage/storage_map.sw
+++ b/sway-lib-std/src/storage/storage_map.sw
@@ -36,7 +36,7 @@ impl<K, V> StorageKey<StorageMap<K, V>> where K: Hash {
     /// }
     /// ```
     #[storage(read, write)]
-    pub fn insert(self, key: K, value: V) {
+    pub fn insert(self, key: K, value: V) where K: Hash {
         let key = sha256((key, self.field_id));
         write::<V>(key, 0, value);
     }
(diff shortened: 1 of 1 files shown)


Braqzen (approved):
Add breaking label?

I don't really understand the security issue. I see it more as a potential compiler issue.
Don't really see an issue with this constraint though.

bitzoic:
> Add breaking label?
> 
> I don't really understand the security issue. I see it more as a potential compiler issue. Don't really see an issue with this constraint though.

Perhaps not security concern, this has been corrected. But we still see the following:

This compiles:
```rust
struct MyStruct {
    val: u64
}

abi MyContract {
    #[storage(read, write)]
    fn test_function();
}

storage {
    my_map: StorageMap<MyStruct, u64> = StorageMap {},
}

    #[storage(read, write)]
    fn test_function() {
        let my_struct = MyStruct { val: 1 };
        storage.my_map.insert(my_struct, 2)
    }
}
```

When this does not, even though the `insert()` function uses the same `sha256<T>()` function:
```rust
use std::hash::sha256;

struct MyStruct {
    val: u64
}

abi MyContract {
    fn test_function();
}

    fn test_function() {
        let my_struct = MyStruct { val: 1 };
        let result = sha256(my_struct);
    }
}
```

esdrubal (approved):
I created an issue that reports the faulty behavior found in this PR. https://github.com/FuelLabs/sway/issues/5046

Although the added constraints are redundant because the impl self already contains them, because the compiler appears not to check the impl self constraints I think this is a good idea to add until the issue is fixed.

PR #5054, opened 2023-08-29
Fixes assert_eq always passing for U256

### LINKED ISSUE #4989: `assert_eq` always pass for U256
Doing a quick test here I found that this test pass.

```
#[test]
fn test_pow_u256() {
    let five = U256::from((0, 0, 0, 5));
    let two = U256::from((0, 0, 0, 2));
    assert_eq(five, two);
}
```

Maybe it is the same problem as https://github.com/FuelLabs/sway/issues/4868 because I am seeing `fn eq` of `U256` body empty at `compile_fn`. This is returning a `const bool undef` and wreaking havoc from there.

_Originally posted by @xunilrj in https://github.com/FuelLabs/sway/issues/4900#issuecomment-1680766892_

esdrubal:
## Description

The issue was occurring because `get_items_for_type_and_trait_name` was comparing an absolute call path to an external relative call path, when the second should also be absolute.

`compile_fn` now checks if method is trait dummy.

These changes should avoid the silent failures where a dummy method is used,
instead of the proper implementation.

Closes #4989

## Checklist

### DIFF
--- a/sway-core/src/ir_generation/compile.rs
+++ b/sway-core/src/ir_generation/compile.rs
@@ -415,8 +415,16 @@ fn compile_fn(
         visibility,
         purity,
         span,
+        is_trait_method_dummy,
         ..
     } = ast_fn_decl;
+
+    if *is_trait_method_dummy {
+        return Err(vec![CompileError::InternalOwned(
+            format!("Method {name} is a trait method dummy and was not properly replaced."),
+            span.clone(),
+        )]);
+    }
 
     let args = ast_fn_decl
         .parameters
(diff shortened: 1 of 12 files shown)


anton-trunov:
It seems that one more way to fix this for a lot of cases without introducing the `is_trait_method_dummy` flag would be to check if the function body is empty and the return type is not `()`. Although, there is one more case to handle, legitimate functions like this one:
```
fn foo(...) {
}
```

On the other hand those functions are kind of useless and we might as well prohibit functions with the empty body and require the user to write something like this
```
fn foo(...) {
    return ()
}
```

Just some food for thought :)

xunilrj (approved):

sdankel (approved):

PR #4933, opened 2023-08-09
Allow method calls between impl self methods.

tritao:
## Description

This adds support for method calls between impl self methods.

## Checklist

- [x] I have linked to any relevant issues.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation where relevant (API docs, the reference, and the Sway book).
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have added (or requested a maintainer to add) the necessary `Breaking*` or `New Feature` labels where relevant.
- [x] I have done my best to ensure that my PR adheres to [the Fuel Labs Code Review Standards](https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/external-contributors.md).
- [x] I have requested a review from the relevant team or maintainers.

PR #5077, opened 2023-09-01
Split method application monomorphization and introduce type check finalization

tritao:
## Description

This splits off some more changes out of https://github.com/FuelLabs/sway/pull/4933.

With these changes, we can now do type checking in multiple phases (sort of), which is necessary to properly fix ordering issues in the compiler. It introduces a new concept to the type checking pipeline, which is typecheck finalization. Right now this is limited to impl self type checking.

We introduce the concept of a deferred monomorphization for method applications. This means that the final step of the monomorphization, replace decls, is done at a later stage in the finalization pass. 

Typecheck finalization is the final typechecking phase, which is called to resolve and finish monomorphization of deferred function applications, which are function applications which will be left incomplete during the initial type checking phase and have their final monorphization of the function postponed, which will be finalized at a later stage using the dependency graph we previously computed during type check analysis process.

The eventual goal is to move the entire monomorphization step to the finalization pass, already did some work for this but ended up with some test failures. We need to move the entire unification stage of code blocks, as well as unification of return types/return statements to the finalization stage.

Also right now some of the refactored code may arguably appear a bit less readable, but eventually I think it will be more clear once we finally can get the final architecture in place, though it will take a few more incremental refactors to get there.

## Checklist

### DIFF
--- a/deployment/Dockerfile
+++ b/deployment/Dockerfile
@@ -1,5 +1,5 @@
 # Stage 1: Build
-FROM lukemathwalker/cargo-chef:latest-rust-1.67 as chef
+FROM lukemathwalker/cargo-chef:latest-rust-1.72 as chef
 WORKDIR /build/
 # hadolint ignore=DL3008
 RUN apt-get update && \


sway-core/src/semantic_analysis/type_check_finalization.rs
```diff
@@ -0,0 +1,32 @@
...
+//! from the initial type checked tree.
+
+use sway_error::handler::{ErrorEmitted, Handler};
+
+use crate::Engines;
+
+use super::TypeCheckContext;
+
+// A simple context that is used to finish type checking.
+pub struct TypeCheckFinalizationContext<'cx, 'dx> {
```
IGI-111, line 12: Please use full names for lifetimes. It can be `'eng` and `'ctx` but it should be descriptive.

IGI-111 (approved):

PR #5094, opened 2023-09-06
Synchronize code_lens request & provide better UX for `forc-fmt` <> `sway-lsp` interaction.

### LINKED ISSUE #4893: Conflicts between language server and forc-fmt
This was reported by @Braqzen on Slack.

> When I have a script open in VSCode there is a little floating thingy that pops up above the main() function which allows me to run it.
> In a terminal that is external to VSCode when I run the forc formatter it formats the script; however, from that point onward the floating pop up for running the script no longer appears.

I suspect this is caused by `forc-fmt` saving the file after it completes. Because the language server is operating in temporary memory this change might not be getting sync'ed with this location leading to AST's that aren't matching.

JoshuaBatty:
## Description
This PR adds 3 things:

1. Adds a benchmark for the format LSP request.

2. Calls `session.wait_for_parsing();` before computing the `code_lens` request. This fixes the original issue reported in #4893 where run buttons where being placed incorrectly after formatting. 

3. If a file open in a code editor contains unsaved changes we write a lock file to `.forc/lsp_locks/`. This file is removed when the file is saved and there are no pending changes. If `forc-fmt` is run in a terminal we check if the path has a lock file associated with it. If unsaved changes are detected we bail from formatting with an error message instructing the user to save changes before continuing. See video below.

https://github.com/FuelLabs/sway/assets/1289413/75e6fddc-adbc-4796-aeb9-985574ae8dcc

closes #4893 

## Checklist

### DIFF
--- a/sway-lsp/src/core/document.rs
+++ b/sway-lsp/src/core/document.rs
@@ -1,8 +1,10 @@
 #![allow(dead_code)]
-use crate::error::DocumentError;
-use lsp_types::{Position, Range, TextDocumentContentChangeEvent};
+use crate::{
+    error::{DirectoryError, DocumentError, LanguageServerError},
+    utils::document,
+};
+use lsp_types::{Position, Range, TextDocumentContentChangeEvent, Url};
 use ropey::Rope;
-use std::{fs::File, sync::Arc};
 
 #[derive(Debug, Clone)]
 pub struct TextDocument {
@@ -12,7 +14,6 @@ pub struct TextDocument {
     version: i32,
     uri: String,
     content: Rope,
-    is_dirty: Option<Arc<fd_lock::RwLock<File>>>,
 }
 
 impl TextDocument {
@@ -23,7 +24,6 @@ impl TextDocument {
                 version: 1,
                 uri: path.into(),
                 content: Rope::from_str(&content),
-                is_dirty: None,
             })
             .map_err(|_| DocumentError::DocumentNotFound { path: path.into() })
     }
@@ -45,19 +45,6 @@ impl TextDocument {
 
     pub fn get_text(&self) -> String {
         self.content.to_string()
-    }
-
-    pub fn mark_file_as_dirty(&mut self) {
-        match forc_util::path_lock(self.uri.as_ref()) {
-            Ok(lock) => self.is_dirty = Some(lock.into()),
-            Err(err) => {
-                tracing::error!("{}", err.to_string());
-            }
-        }
-    }
-
-    pub fn mark_file_as_clean(&mut self) {
-        self.is_dirty = None;
     }
 }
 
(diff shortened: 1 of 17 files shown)


anton-trunov:
What if the user changes their files on disk using something different from `forc-fmt`, e.g. `sed`? I understand that it's less likely, though. 

I believe, if possible, it would be nice to have `forc-fmt` and `sway-lsp` not aware of each other.

JoshuaBatty:
> I believe, if possible, it would be nice to have `forc-fmt` and `sway-lsp` not aware of each other.

Yeah I agree this would be ideal. I'm going to put this PR in draft mode while I ponder on this a bit more. Thanks Anton

anton-trunov:
Ah, that's interesting, didn't know about that integration. I'd guess they use some kind of file system lock. @tritao can you share some more details how they do it?

sdankel:
We should get thing 1 and thing 2 in regardless! 

> This is the message I see:     Blocking waiting for file lock on build directory

Taking a lock on the build directory is clever because it links it to the specific sway project, doesn't prevent editing the file in other editors at the same time, and should only ever be touched by forc.

Both forc-fmt and forc-client (forc build) would also need to wait for a lock on the build directory. 

We could have LSP take a lock on the build directory on `format_text`, but I'm concerned it would block other operations for a long time because sometimes it's very slow. What results did you get from benchmarking? Is `get_page_text_edit` the reason it's sometimes slow?

JoshuaBatty:
> What results did you get from benchmarking? Is `get_page_text_edit` the reason it's sometimes slow?

It's actually pretty fast. Averaging ~98ms to complete. I think the slowness is coming from the client <> server interaction.

JoshuaBatty:
Thanks for the tips everyone. I'm going to look into a nice solution this week.

JoshuaBatty:
Ok i'm pretty happy with this now. 

Any unsaved changes for a file in LSP now creates a hashed lock file in `~/.forc/.lsp_locks/`. When a file is saved or closed and changes are discarded then the lock file is removed. The formatter checks for the existence of this lock file before formatting. If one exists then the user is instructed to save the file before continuing. See the updated video in the description.

sway-lsp/src/core/document.rs
```diff
@@ -104,6 +106,43 @@ impl TextDocument {
...
+/// Removes the corresponding flag file for the specifed Url.
+///
+/// If the flag file does not exist, this function will do nothing.
+pub fn remove_dirty_flag(uri: &Url) -> Result<(), LanguageServerError> {
+    let path = document::get_path_from_url(uri)?;
+    eprintln!("path to remove dirty flag: {:?}", path);
+    let dirty_file_path = forc_util::is_dirty_path(&path);
+    eprintln!("dirty_file_path: {:?}", dirty_file_path);
+    if dirty_file_path.exists() {
+        eprintln!("Removing dirty flag file: {:?}", dirty_file_path);
```
sdankel, line 136:
```suggestion
    let dirty_file_path = forc_util::is_dirty_path(&path);
    if dirty_file_path.exists() {
```

sdankel (commented):
LGTM, just need to clean up eprintlns

sway-lsp/src/core/document.rs
JoshuaBatty: facepalm. Thanks!

sdankel (approved):

cr-fuel (approved):

PR #5096, opened 2023-09-06
Add `new()` function to `StorageKey`

bitzoic:
## Description

When creating complex storage types, storing of a `StorageKey` is sometimes needed. In these cases a `new()` function is appropriate.

For example, when we have the following storage block:
```rust
storage {
    name: StorageMap<u64, StorageKey<StorageString>> = StorageMap {},
}
```

In order to store a `StorageString` in a `StorageMap`, the `StorageKey<StorageString>` annotation is needed.

From there, we can store a `String` by creating a `StorageKey`, storing a `String` at said key, and then storing the `StorageKey` in the `StorageMap`:

```rust
let string_key: StorageKey<StorageString> = StorageKey {
    slot: sha256(0),
    offset: 0,
    field_id: sha256(1),
};
string_key.write_slice(String::from_ascii_str("my_name"));
storage.name.insert(0, string_key);
```

The introduction of `new()` would change this to the following:
```rust
let string_key: StorageKey<StorageString> = StorageKey::new(
     sha256(0), 
     0, 
     sha256(1)
);
string_key.write_slice(String::from_ascii_str("my_name"));
storage.name.insert(0, string_key);
```

## Checklist


K1-R1 (dismissed):

sway-lib-std/src/storage/storage_map.sw
```diff
@@ -68,11 +68,7 @@ impl<K, V> StorageKey<StorageMap<K, V>> where K: Hash {
     /// }
     /// ```
     pub fn get(self, key: K) -> StorageKey<V> where K: Hash {
-        StorageKey {
-            slot: sha256((key, self.field_id)),
-            offset: 0,
-            field_id: sha256((key, self.field_id)),
-        }
```
anton-trunov, line 75: Admittedly, I don't write too much Sway code myself, maybe that's the reason I think the original code looks a bit easier to read

sway-lib-std/src/storage/storage_key.sw
```diff
@@ -97,4 +97,43 @@ impl<T> StorageKey<T> {
...
         write(self.slot, self.offset, value);
     }
+
+    /// Create a new `StorageKey`.
+    ///
+    /// # Arguments
+    ///
+    /// * `slot`: [b256] - The assigned location in storage for the new `StorageKey`.
+    /// * `offset`: [u64] - The assigned offset based on the data structure `T` for the new `StorageKey`.
+    /// * `field_id`: [b256] - A unqiue identifier for the new `StorageKey`.
```
Braqzen, line 107:
```suggestion
    /// * `field_id`: [b256] - A unique identifier for the new `StorageKey`.
```

Braqzen (commented):
I'm a little confused about your comments in the issue regarding the usage of a `String` and a `StorageString`. The type in the map is `StorageString` but you create a string that gets assigned to some arbitrary key and then via magic it becomes the `StorageString` type

sway-lib-std/src/storage/storage_map.sw
bitzoic: The primary purpose for this is encapsulation. Admittedly the changes in the std-lib to use this should be broken up into new lines, not sure why the formatter didn't flag this. The original method of defining each variable is still possible.

Braqzen (approved):

K1-R1 (approved):

PR #5100, opened 2023-09-08
Adjust existing Diagnostic API according to the Expressive Diagnostics RFC

### LINKED ISSUE #5075: Adjust existing Diagnostic API according to the Expressive Diagnostics RFC
Essentially, this means implementing the changes so that the diagnostics rendered in `forc` CLI look like the new ones shown [here](https://github.com/FuelLabs/sway/issues/5056#issuecomment-1698964837).

ironcev:
## Description

This PR implements the decisions made in the [Expressive Diagnostics RFC](https://github.com/FuelLabs/sway-rfcs/blob/master/rfcs/0011-expressive-diagnostics.md) and adapts the existing diagnostics:
- _Issue_ is removed from the header and placed in code.
- _Help_ can also be in code.
- Wording is simplified.

Closes #5075.

## Demo (Before and After)

Match arm is unreachable - Variant 1

Match arm is unreachable - Variant 2

## Checklist

### DIFF
--- a/forc-util/src/lib.rs
+++ b/forc-util/src/lib.rs
@@ -559,8 +559,8 @@ fn format_diagnostic(diagnostic: &Diagnostic) {
 
     fn get_title_label(diagnostics: &Diagnostic, label: &mut String) {
         label.clear();
-        if diagnostics.reason().is_some() {
-            label.push_str(diagnostics.reason().unwrap().description());
+        if let Some(reason) = diagnostics.reason() {
+            label.push_str(reason.description());
         }
     }
 
--- a/forc-util/src/lib.rs
+++ b/forc-util/src/lib.rs
@@ -561,10 +561,7 @@ fn format_diagnostic(diagnostic: &Diagnostic) {
         label.clear();
         if diagnostics.reason().is_some() {
             label.push_str(diagnostics.reason().unwrap().description());
-            label.push_str(". ");
         }
-        label.push_str(diagnostics.issue().friendly_text());
-        label.push('.');
     }
 
     fn diagnostic_level_to_annotation_type(level: Level) -> AnnotationType {
--- a/sway-error/src/warning.rs
+++ b/sway-error/src/warning.rs
@@ -88,6 +88,7 @@ pub enum Warning {
     },
     MatchExpressionUnreachableArm {
         match_value: Span,
+        match_type: String,
         // Either preceding non catch-all arms or a single interior catch-all arm.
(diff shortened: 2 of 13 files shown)


forc-util/src/lib.rs
```diff
@@ -561,10 +561,7 @@ fn format_diagnostic(diagnostic: &Diagnostic) {
         label.clear();
         if diagnostics.reason().is_some() {
             label.push_str(diagnostics.reason().unwrap().description());
```
xunilrj, line 563: why not `if let Some(reason)` here?

xunilrj (dismissed):

forc-util/src/lib.rs
ironcev: Well, I guess because the line was originally written while I was still catching up with Rust idioms :-) Thanks @xunilrj  the suggestion is applied! Can you please re-approve the PR?

sway-error/src/warning.rs
```diff
@@ -267,58 +268,52 @@ impl ToDiagnostic for CompileWarning {
...
-                    Hint::warning(
-                        source_engine,
-                        unreachable_arm.clone(),
-                        format!("{} arm \"{}\" is unreachable.", if *is_last_arm && *is_catch_all_arm { "Last catch-all match" } else { "Match" }, unreachable_arm.as_str())
-                    ),
                     Hint::info(
                         source_engine,
                         match_value.clone(),
-                        "This is the value to match on.".to_string()
+                        format!("`{}`, of type \"{match_type}\", is the value to match on.", match_value.as_str())
```
anton-trunov, line 299: This should probably be something like this:
```suggestion
                        format!("`{}`, of type \"{match_type}\", is the expression to match on.", match_value.as_str())
```

xunilrj (approved):

sway-error/src/warning.rs
ironcev: After a day of pondering, mixed feelings, and a strong urge to start an interesting bike-shedding round :smile:  I've came to conclusion that I don't like that hint formulation in general. It's too long and clumsy for what it wants to say and the value it brings - pointing to the matched expression and stating its type, which is useful.

I am more for something simple like: _"x" has type "(bool, u64)"._

Or in the longer case: _Expression "x" has type "(bool, u64)"._

But let's do that as a part of an upcoming PR which will also be related to match expressions.

IGI-111 (approved):

PR #5103, opened 2023-09-09
Enhance forc-fmt

cr-fuel:
## Checklist


forc-plugins/forc-fmt/src/main.rs
```diff
@@ -154,7 +154,11 @@ fn format_file(
                 // TODO: Support formatting for incomplete/invalid sway code.
                 // https://github.com/FuelLabs/sway/issues/5012
                 debug!("{}", err);
-                bail!("Failed to compile: {:?}", file);
+                if let Some(file) = file.to_str() {
+                    bail!("Failed to compile {}\n{}", file, err);
+                } else {
+                    bail!("Failed to compile.\n{}", err);
+                }
```
eureka-cpu, line 161: Was this change necessary? Seems the previous code achieves the same outcome in fewer lines.

sway-lib-core/src/ops.sw
```diff
@@ -1019,14 +1018,12 @@ trait OrdEq: Ord + Eq {
...
-    ///     assert(result); 
+    ///     assert(result);
     /// }
     /// ```
     fn ge(self, other: Self) -> bool {
         self.gt(other) || self.eq(other)
-    }
-
-    /// Evaluates if one value of the same type is less or equal to than another.
+    }    /// Evaluates if one value of the same type is less or equal to than another.
```
eureka-cpu, line 1026: There's a missing new line here after the closing brace that the formatter removed. Would you mind making an issue to track it?

sway-lib-std/src/hash.sw
```diff
@@ -137,23 +139,35 @@ impl Hash for Bytes {
     }
 }
 
-impl<A, B> Hash for (A, B) where A: Hash, B: Hash  {
-    #![inline(never)]
+impl<A, B> Hash for (A, B) where
```
eureka-cpu, line 142: `where` clauses by default should format to be on the next line, just like how they are with functions. This should be another issue, easy fix too.

sway-lib-std/src/hash.sw
```diff
@@ -297,8 +337,11 @@ impl<T> Hash for [T; 10] where T: Hash {
 ///     assert(result = 0xa80f942f4112036dfc2da86daf6d2ef6ede3164dd56d1000eb82fa87c992450f);
 /// }
 /// ```
-#![inline(never)]
-pub fn sha256<T>(s: T) -> b256 where T: Hash {
+#[inline(never)]
+pub fn sha256<T>(s: T) -> b256
+where
```
eureka-cpu, line 342: here is an example of what the `where` clause should look like for `impl` blocks by default

sway-lib-std/src/u256.sw
```diff
@@ -241,17 +241,11 @@ impl U256 {
 
 impl core::ops::Ord for U256 {
     fn gt(self, other: Self) -> bool {
-        self.a > other.a ||
-                (self.a == other.a && (self.b > other.b ||
-                                      (self.b == other.b && (self.c > other.c ||
-                                                            (self.c == other.c && self.d > other.d)))) )
-        }
+        self.a > other.a || (self.a == other.a && (self.b > other.b || (self.b == other.b && (self.c > other.c || (self.c == other.c && self.d > other.d)))))
+    }
```
eureka-cpu, line 245: chained booleans should have a newline and indent similarly to `if else` statements, this could also be another issue to open

forc-plugins/forc-fmt/src/main.rs
cr-fuel: @eureka-cpu They are not the same thing. The error was not being printed previously, only logged. Right now the error is part of the output. In fact, adding the error helped me fix the issue with `hash.sw`.

Converting error from 

```
error: Formatting skipped due to error.
error: Failed to compile: "/Users/cr-fuel/projects/fuel/sway/sway-lib-std/src/hash.sw"
```

To 

```
error: Formatting skipped due to error.
error: Failed to compile /Users/cr-fuel/projects/fuel/sway/sway-lib-std/src/hash.sw
Cannot format raw hashbang attribute. If this is intended to be a doc comment try using the `//!` syntax instead
```

sway-lib-core/src/ops.sw
cr-fuel: Please report similar issues. I will fix it now (fix will come in the swayfmt crate)

sway-lib-std/src/hash.sw
cr-fuel: I think in this case, if the where body is too long `where` should be in the next line. I'll improve it.

forc-plugins/forc-fmt/src/main.rs
eureka-cpu: I see, also I wrote that error message haha

Nice to see it there

sway-lib-core/src/ops.sw
eureka-cpu: This may have something to do with the newline handler, which is its own traversal of the formatted string in `formatter/mod.rs`

sway-lib-std/src/hash.sw
eureka-cpu: There is an option `SameLineWhere` in the config settings, and that is the only time `where` should be on the same line. The default config should result in the same kind of formatting as functions, structs and all other types that can have `where` clauses.

sway-lib-std/src/u128.sw
```diff
@@ -95,16 +95,16 @@ impl u64 {
 
         asm(sum, overflow, left: self, right: right, result_ptr: result) {
             // Add left and right.
-            add sum left right;
+            add  sum left right;
```
JoshuaBatty, line 98: the extra spaces in this and the following `asm` blocks don't look correct.

sway-lib-std/src/string.sw
```diff
@@ -147,13 +144,11 @@ impl String {
...
-        let ptr = asm(s: s) {
-            s: raw_ptr
-        };
-        let slice = asm(parts:(ptr, len)) { parts: raw_slice};
+        let len = __size_of_str::<S>();
+        let ptr = asm(s: s) { s: raw_ptr };
+        let slice = asm(parts: (ptr, len)) { parts: raw_slice };
         Self {
-            bytes: Bytes::from(slice)
+            bytes: Bytes::from(slice),
```
JoshuaBatty, line 151: i'm not sure adding a `,` here is necessary if its the last entry

sway-lib-std/src/u128.sw
sdankel: IIRC they're adding padding to make each asm opcode take up 4 characters for alignment. However, I thought that was because none of the opcodes were longer than 4 characters, but I see `right` is 5 chars. I don't see `left` or `right` keywords in the [docs](https://specs.fuel.network/master/fuel-vm/instruction-set.html).

sway-lib-std/src/u256.sw
sdankel: Agreed. Let's open an issue, but I don't think this should block us from enabling the CI checks.

swayfmt/tests/mod.rs
```diff
@@ -1602,3 +1602,61 @@ impl MyContract for Contract {
 "#,
     );
 }
```
sdankel, line 1604: We should add a few test cases for each of the bugs fixed (under `swayfmt/src/items/<item type>/tests.rs`)

swayfmt/src/items/item_impl/mod.rs
```diff
@@ -38,14 +38,27 @@ impl Format for ItemImpl {
...
             formatter.shape.code_line.update_where_clause(true);
         }
-        Self::open_curly_brace(formatted_code, formatter)?;
+
         let contents = self.contents.get();
-        for item in contents.iter() {
-            write!(formatted_code, "{}", formatter.indent_str()?,)?;
-            item.format(formatted_code, formatter)?;
-            writeln!(formatted_code)?;
+        if contents.is_empty() {
```
sdankel, line 43: Let's add a couple test cases for this fix under `swayfmt/src/items/item_impl/tests.rs`

sway-lib-core/src/primitive_conversions.sw
```diff
@@ -647,30 +635,60 @@ impl b256 {
...
+            c[7],
+            d[0],
+            d[1],
+            d[2],
+            d[3],
+            d[4],
+            d[5],
+            d[6],
+            d[7],
+        ];
```
sdankel, line 679: This is a little unsightly. rustfmt would inline them like this:

```rust
    let output = [
        a[0], a[1], a[2], a[3], a[4], a[5], a[6], a[7], b[0], b[1], b[2], b[3], b[4], b[5], b[6],
        b[7], c[0], c[1], c[2], c[3], c[4], c[5], c[6], c[7], d[0], d[1], d[2], d[3], d[4], d[5],
        d[6], d[7],
    ];
```

Perhaps we should open an issue.

eureka-cpu: there may already be one IIRC

sdankel:
As discussed offline, let's split out the bug fixes into separate PRs to get those merged first.

sway-lib-std/src/u128.sw
xunilrj: They are not. See their declaration on line 96.

JoshuaBatty:
@cr-fuel Are we able to close this PR now?

PR #5104, opened 2023-09-10
Fix issues with variables in match arm alternatives

### LINKED ISSUE #5081: Emit compile error if the same match arm variables in alternatives are not of the same type
At the moment there is no error if match arm variables with the same name in match arm alternatives are not of the same type. The type of the variable in the last alternative will win.

```Rust
struct S {
    x: bool,
    y: u64
}

...
let _x = match s {
    S { x: y, y: x, .. } | S { x, y, .. } => x, // No type mismatch error but it should be.
           ^                      ^ y is u64 here
           |
           y is bool here 
    ...
}
```

### LINKED ISSUE #5082: Invalid error message that variable is not bound in all match arm alternatives in case of duplicates in alternatives
If one of the alternatives contains duplicate variables, an error will be reported in other alternatives that the variable is not bound although it is.

```Rust
let _x = match or5 {
    Struct { x, y: x, .. } | Struct { x, .. } => x,
                             ^^^^^^^^^^^^^^^^ Variable "x" is not bound in all patterns
    ...
 };
```

### LINKED ISSUE #5108: Matcher should not short-circuit on first error in match arm variants
At the moment matcher short-circuits on the first error of any kind found in match arm variants (e.g. variable not bound in an alternatives). This leads to not emitting certain errors that are expected to be emitted like discussed in [this PR comment](https://github.com/FuelLabs/sway/pull/5104#discussion_r1321929554).

Adapt matcher to emit as many errors as possible before short-circuiting.

ironcev:
## Description

This PR fixes three issues related to variables in match arm alternatives:
- removes the wrong error message that a variable in match arm alternative is not defined (#5082).
- emits error if the same variable in match arm alternatives is not of the same type in all alternatives (#5081).
- emits all errors in a match arm instead of short-circuiting after the first error (#5108)

Closes #5081.
Closes #5082.
Closes #5108.

## Demo

image

image

## Checklist


ironcev:
This PR builds on top of #5100 so I'll keep it in draft until that one is closed.

test/src/e2e_vm_tests/test_programs/should_fail/match_expressions_or_mismatched_types/src/main.sw
```diff
@@ -0,0 +1,239 @@
...
+    let _x = match gen {
+        (x, y) | (y, x) => 0,
+    };
+}
+
+fn main() {
+    let s1 = Struct::new();
+
+    let _x = match s1 {
+        Struct { x, y } | Struct { x: y, y: x } => if y { 0 } else { x },
```
xunilrj, line 89: Should this be accepted?
At `Struct { x, y }`, `y` is `u64`; but at `Struct { x: y, y: x }`, `y` is `bool`. Rustc correctly complains.

```
   |
17 |     let _x = match s1 {
   |                    -- this expression has type `Struct`
18 |         Struct { x, y } | Struct { x: y, y: x } => if y { 0 } else { x },
   |                     -                 ^ expected `u64`, found `bool`
   |                     |
   |                     first introduced with type `u64` here
   |
   = note: in the same arm, a binding must have the same type in all alternatives
```

ironcev: Good point! The existing code was short-circuiting at the first inconsistency between the checked and the previous alternative and I kept that logic intact. But it's definitely worth questioning, both for this new error and the existing one (which was buggy and also fixed). The question is only if we don't short-circuit immediately, how far do we go. Do we collect all inconsistencies in all the alternatives or only all in the currently checked and then short-circuit?

The latter approach should be straightforward and should also emit error in the case you are pointing.

I have to see if the former approach potentially clashes with the current logic in which (unlike Rust!) the last alternative wins. I personally find that approach confusing because intuitively I see the first alternative as the "source of truth" that others should follow.

Anyhow, putting the PR back to draft until I give it a thought and decide ;-) Thanks!

ironcev:
Back to draft until decision is made on [short-circuiting](#discussion_r1321967091).

test/src/e2e_vm_tests/test_programs/should_fail/match_expressions_or_mismatched_types/src/main.sw
ironcev: Gave it a second thought :-) I definitely want to emit as much as possible but let's do it in a separate PR that will address #5108 and leave this implementation as is.

Essentially, I want to change the original algorithm (remove redundant `if` and `Some`, make the first alternative as the reference and not the ever-changing current one, ...), double check the affected code areas and existing tests, and also use the opportunity to move the existing error for non-bound variable to expressive diagnostics.

I've commented in the PR description that the current algorithm short-circuits.

ironcev:
Back to draft (again 😄 ). I'll put all the remaining work to be done in matcher in this PR.

anton-trunov (approved):

sway-core/src/semantic_analysis/ast_node/expression/match_expression/typed/typed_match_branch.rs
```diff
@@ -31,8 +31,13 @@ impl ty::TyMatchBranch {
         let typed_scrutinee = ty::TyScrutinee::type_check(handler, ctx.by_ref(), scrutinee)?;
 
         // calculate the requirements map and the declarations map
-        let (match_req_map, match_decl_map) =
-            matcher(handler, ctx.by_ref(), typed_value, typed_scrutinee.clone())?;
+        let (match_req_map, match_decl_map) = matcher(
+            handler,
+            ctx.by_ref(),
+            typed_value,
+            typed_value,
```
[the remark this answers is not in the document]
ironcev: You are right! Obvious to me :smile:  but not that obvious at first sight. Thanks for pointing it out I'll include the comment in the next PR there is still that critical bug to fix I stumbled upon.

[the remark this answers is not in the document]
anton-trunov: Thanks Igor!

Issue #4982, opened 2023-08-18
Rename symbol for type aliases isn't working


sdankel:
Renaming the type where it's defined only updates the ident where it's defined. Also, renaming a function that are implemented for a type within the impl block fails without an error (seems to do nothing).

https://github.com/FuelLabs/sway/assets/47993817/9b8591f9-c87c-44dc-9438-c6e66fd29688

https://github.com/FuelLabs/sway/assets/47993817/bdf1f31f-162f-47f5-a9da-f9efba25c92b

PR #5118, opened 2023-09-13
Fix rename for alias and function prefixes

sdankel:
## Description

Closes https://github.com/FuelLabs/sway/issues/4982

There were actually two issues here. 
1. Type aliases were not getting renamed
2. In any function call with a prefix, the prefix was not getting renamed, whether it was a type alias or not

I added a test that covers both. I also noticed that the other rename tests do not assert anything about the response. I think we should at least assert the number of changes is correct to prevent regressions. I can add that to this PR or open a follow up.

This changes the behavior of go to the definition of a function call prefix. Previously it would go to the impl of the prefix where the function was implemented. Now it takes you to the definition of the struct/enum, or the top of the module instead. The new behavior is more similar to rust-analyzer.

## Checklist


JoshuaBatty:
looks like some go to tests are failing now.

JoshuaBatty (approved):
LGTM

kayagokalp (approved):
utACK

Issue #2439, opened 2022-08-01
Cast `Vec` to `StorageVec`


bitzoic:
There is currently no relation between the `Vec` and `StorageVec` in the standard library. There should be a way to cast a `Vec` to a `StorageVec` and vice versa.

An example where this is needed would be passing a `Vec` to a function and then storing the given `Vec`. Or returning a stored `StorageVec` as a `Vec` to the SDK.

We should also consider the case of utility methods within `Vec` as described by #2014 and whether these should be duplicated in `StorageVec` or the end user should cast back and forth to perform certain operations.

mohammadfawaz:
`StorageVec` is really just a `StorageMap` with a length and where the keys are `0, 1, 2, ...`. The only way a cast would work here is trough a method that simply copies the data from a `StorageVec` to a `Vec` or vice versa. For example, returning a `StorageVec` as `Vec` would first require that we read all the elements of the `StorageVec` and store them in a local `Vec` at the same indices. We then return the resulting `Vec` as we usually do (once this is supported of course). Is this what you have in mind?

bitzoic:
This is what I had in mind initially as I see developers needing to do this often. The only problem being that this could get quite expensive.

I bring this up as I have a need for this in my English Auction application. I have a `Vec` of token ids that I will need to store. I may also need something similar when implementing a strings library.

mohammadfawaz:
Yeah this is going to be expensive. Now two things:
* Storing the elements of `Vec` in a `StorageVec` is going to be as expensive as manually copying the elements one by one, and I don't think there's a way around that.
* Returning a `StorageVec` to the SDK _may_ be possible without having to go through `Vec<T>`. We could potentially return the storage keys themselves and have the SDK rebuild the `StorageVec` once its able to access storage slots. At least this moves the burden of reading from the contract to the SDK.

adlerjohn:
The fact that it will be very expensive is pointing toward this _not_ being used often.

jtriley-eth:
Bumping this issue, as we have some possible use cases while updating the `Vec` and `StorageVec` API's (issue #2014).

While the utility of sorting algorithm implementations for storage vectors are limited, it would be good to match API's between the memory and storage vectors. However, any algorithm that requires more than `N` reads or writes on a storage vector of length `N` could benefit from being converted into memory vectors first. So `StorageVec::sort` would ideally create a memory vector, load each value into the memory vector, sort them, then write them to storage.

This may be useful for the following:

- `sort`
[...]

PR #4795, opened 2023-07-12
Optimize Storage read/write api

bitzoic:
## Description

The storage api currently reads in all storage from `0` to `offset`. This has been optimized to only read the required storage slots.

Closes https://github.com/FuelLabs/sway/issues/4789

## Checklist

- [x] I have linked to any relevant issues.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation where relevant (API docs, the reference, and the Sway book).
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have added (or requested a maintainer to add) the necessary `Breaking*` or `New Feature` labels where relevant.
- [x] I have done my best to ensure that my PR adheres to [the Fuel Labs Code Review Standards](https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/external-contributors.md).
- [x] I have requested a review from the relevant team or maintainers.

PR #5123, opened 2023-09-18
Enable storing of a `Vec` with the `StorageVec` type

bitzoic:
## Description

Currently, `StorageVec` and `Vec` do not have any relation with one another. This PR takes advantage of the recent optimization and refactoring done to the Storage API `read()` and `write()` functions in https://github.com/FuelLabs/sway/pull/4795 and enables the storing of a `Vec` using the `StorageVec` type.

To do this, the `StorageVec` now stores things sequentially rather than a different key for every element. Due to the optimizations done to the Storage API, this has become feasible as we can now load a single element of the sequential `StorageVec`.

The storing of elements sequentially mimics the existing `Vec`, allowing us to store it as a `raw_slice` with a ***single*** read/write to storage as opposed to looping over the `Vec` and having ***n*** read/writes. 

It should be noted that due to https://github.com/FuelLabs/sway/issues/409, the storing of a `Vec` is written in the `StorageVec` file. This is the resulting syntax:
```sway
let my_vec = Vec::new();
storage.storage_vec.store_vec(my_vec); // Store the Vec
let other_vec = storage.storage_vec.load_vec(); // Read the Vec
```
When this issue is resolved, this should be changed to a `From` implementation changing the syntax to:
```sway
let my_vec = Vec::new();
storage.storage_vec = my_vec.into(); // Store the Vec
let other_vec = Vec::from(storage.storage_vec); // Read the Vec
```

Closes https://github.com/FuelLabs/sway/issues/2439

## Checklist


sway-lib-std/src/storage/storage_vec.sw
```diff
@@ -755,10 +783,131 @@ impl<V> StorageKey<StorageVec<V>> {
...
+        let mut ptr = slice.ptr();
+
+        // The capacity needs to be a multiple of 32 bytes so we can 
+        // make the 'quad' storage instruction store without accessing unallocated heap memory.
+        ptr = realloc_bytes(ptr, number_of_bytes, number_of_slots * 32);
+
+        // Store `number_of_slots * 32` bytes starting at storage slot `key`.
+        let _ = __state_store_quad(sha256(self.field_id), ptr, number_of_slots);
+
+        // Store the length length, NOT the bytes.
```
K1-R1, line 858:
```suggestion
        // Store the length, NOT the bytes. 
```

sway-lib-std/src/storage/storage_vec.sw
```diff
@@ -755,10 +783,131 @@ impl<V> StorageKey<StorageVec<V>> {
...
+                let ptr = alloc_bytes(number_of_slots * 32);
+                // Load the stored slice into the pointer.
+                let _ = __state_load_quad(sha256(self.field_id), ptr, number_of_slots);
+                Some(Vec::from(asm(ptr: (ptr, len *  __size_of::<V>())) { ptr: raw_slice }))
+            }
+        }
+    }
+}
+
+fn offset_calculator<T>(offset: u64) -> u64 {
```
K1-R1, line 911: Should this be moved to [storage_api.sw](https://github.com/FuelLabs/sway/blob/bitzoic-vec-storage/sway-lib-std/src/storage/storage_api.sw)? Or is it only relevant here?

K1-R1 (approved):
lgtm

sway-lib-std/src/storage/storage_vec.sw
bitzoic: This is a private function that is only relevant here

sway-lib-std/src/storage/storage_vec.sw
```diff
@@ -42,8 +43,9 @@ impl<V> StorageKey<StorageVec<V>> {
         let len = read::<u64>(self.field_id, 0).unwrap_or(0);
 
         // Storing the value at the current length index (if this is the first item, starts off at 0)
```
SwayStar123, line 45:
```suggestion
```

I see you changed the line below so this comment should be updated

sway-lib-std/src/storage/storage_vec.sw
```diff
@@ -191,16 +199,19 @@ impl<V> StorageKey<StorageVec<V>> {
...
-        let removed_element = read::<V>(sha256((index, self.field_id)), 0).unwrap();
+        let key = sha256(self.field_id);
+        let removed_offset = offset_calculator::<V>(index);
+        let removed_element = read::<V>(key, removed_offset).unwrap();
 
         // for every element in the vec with an index greater than the input index,
         // shifts the index for that element down one
         let mut count = index + 1;
         while count < len {
             // gets the storage location for the previous index
```
SwayStar123, line 210:
```suggestion
```
This comment also seems outdated

sway-lib-std/src/storage/storage_vec.sw
```diff
@@ -755,10 +783,131 @@ impl<V> StorageKey<StorageVec<V>> {
...
+    ///     assert(5 == returned_vec.get(0).unwrap());
+    ///     assert(10 == returned_vec.get(1).unwrap());
+    ///     assert(15 == returned_vec.get(2).unwrap());
+    /// }
+    /// ```
+    #[storage(read)]
+    pub fn load_vec(self) -> Option<Vec<V>> {
+        // Get the length of the slice that is stored.
+        match read::<u64>(self.field_id, 0).unwrap_or(0) {
+            0 => None,
```
SwayStar123, line 898: I think you should just match on the Option here, I would expect an empty vec to be returned if there was an empty storagevec (I believe you can store an empty vec using the above store method?).

SwayStar123 (changes requested):

SwayStar123 (dismissed):

IGI-111 (dismissed):

SwayStar123 (approved):

K1-R1 (approved):

PR #5127, opened 2023-09-19
Extend primitive conversions and implement `TryFrom` trait for integer values

bitzoic:
## Description

There have been a number of requests to convert between integer values. This is especially important in some of the recently introduced standards and changes made in the storage optimizations.

This PR extends the primitive conversions to extend values and also implements the `TryFrom` trait for `u64`, `u32`, `u16`, and `u8`.

Note: The `TryFrom` implementations are not included in the prelude and must be imported. The primitive conversions do not need to be imported as they are included in the core. 

## Checklist


SwayStar123 (dismissed):

sway-lib-core/src/primitive_conversions.sw
```diff
@@ -248,6 +270,28 @@ impl u32 {
...
+    ///
+    /// ```sway
+    /// fn foo() {
+    ///     let val = 2u32;
+    ///     let result = val.as_u256();
+    ///     assert(result == 0x0000000000000000000000000000000000000000000000000000000000000002u256);
+    /// }
+    /// ```
+    pub fn as_u256(self) -> u256 {
+        let input = (0, 0, 0, self);
```
xunilrj, line 289: I think this is a little bit dangerous... I would convert `self` to u64 first. This can be typed as (u32, u32, u32, u32) and generate the wrong binary representation.

Same for other integer types.

sway-lib-std/src/u16.sw
```diff
@@ -0,0 +1,84 @@
+library;
+
+use ::convert::TryFrom;
+use ::option::Option;
+
+impl TryFrom<u32> for u16 {
+    fn try_from(u: u32) -> Option<Self> {
+        if u > asm(r1: u16::max()) { r1: u32 } {
+            return Option::None
```
xunilrj, line 9: nit: don't need return here, right?

sway-lib-std/src/u16.sw
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+}
+
+#[test]
+fn test_u16_try_from_u32() {
+    use ::assert::assert;
+    
+    let u32_1: u32 = 2u32;
+    let u32_2: u32 = asm(r1: u16::max()) { r1: u32 } + 1;
```
xunilrj, line 43: You could have used `u16::max().as_u32()` here right?

sway-lib-core/src/primitive_conversions.sw
```diff
@@ -397,6 +419,31 @@ impl u32 {
...
+    /// # Examples
+    ///
+    /// ```sway
+    /// fn foo() {
+    ///     let val = 2u32;
+    ///     let result = val.as_u256();
+    ///     assert(result == 0x0000000000000000000000000000000000000000000000000000000000000002u256);
+    /// }
+    /// ```
+    pub fn as_u256(self) -> u256 {
```
SwayStar123, line 439:
```
pub fn as_u256(self) -> u256 {
    let input = (0, 0, 0, self);
    asm(input: input) {
        input: u256
    }
}
```

This works, you dont need to use another impl block

SwayStar123 (changes requested):
Looks good otherwise. Just remove all the `_u64` and `.into_u64()` for the `as_b256` functions and then move them into the same impl block

SwayStar123 (approved):

sway-lib-std/src/primitive_conversions/u16.sw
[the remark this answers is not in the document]
bitzoic: These already exist and the functions have just been moved.

bitzoic:
> Why do we have `try_as_uX` methods and some `TryFrom` impls? Can we not only use the latter?

These exist prior to the PR in the primitive_conversion.sw file. They will be replaced with `TryInto` once the trait exists.

SwayStar123 (approved):

K1-R1 (approved):

PR #5134, opened 2023-09-20
Type check analysis

tritao:
## Description

This PR introduces a typecheck analysis pass, which works in a similar way to control flow graph analysis, but is specialized to ordering issues.

We use it to build a graph that keeps track of function call dependencies, which we use at a later stage during the type check finalization stage to process nodes in order.

Right now this step is only used for processing impl self items in the right order, but will be useful in a following PR for recursive call chain analysis.

During impl self typechecking, we process each item into a dummy nodes with deferred monomorphization, the nodes are then used for type check analysis, which we use later to build the dependency graph. After that we have the right order and re-type check the nodes with monomorphization enabled.

Another PR will follow after this one is merged that introduces support for impl self methods calling each other based on this code.

## Checklist

### DIFF
--- a/sway-core/src/semantic_analysis/type_check_analysis.rs
+++ b/sway-core/src/semantic_analysis/type_check_analysis.rs
@@ -176,12 +176,6 @@ impl TypeCheckAnalysisContext<'_> {
         }
     }
 
-    #[allow(dead_code)]
-    pub(crate) fn sort_nodes(&self) {
-        let _res = petgraph::algo::toposort(&self.dep_graph, None)
-            .expect("found a cycle in the dependency graph");
-    }
-
     pub(crate) fn get_sub_graph(
         &self,
         node_index: NodeIndex,


sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs
```diff
@@ -1372,4 +1503,39 @@ impl TypeCheckFinalization for TyImplTrait {
             Ok(())
         })
     }
+
+    // fn type_check_finalize_graph(
```
JoshuaBatty, line 1507: Is this commented section meant to be here?

sway-core/src/semantic_analysis/type_check_analysis.rs
```diff
@@ -0,0 +1,249 @@
...
+
+use crate::decl_engine::{DeclId, DeclIdIndexType, DeclRef};
+use crate::engine_threading::DebugWithEngines;
+use crate::language::ty::{self, TyImplItem, TyTraitItem};
+use crate::Engines;
+
+pub type TyNodeDepGraphNodeId = petgraph::graph::NodeIndex;
+
+#[derive(Clone, Debug)]
+pub struct TyNodeDepGraphEdge(pub String);
```
IGI-111, line 19: These should be typed as an enum instead of manipulating raw strings.

IGI-111: Maybe I'm misunderstanding what this is for, but I see literal instantiations of it using those strings and it looks like this is stored in the graph. Even if this is debug information it seems like pure overheat to store it in there as owned strings that are all the same instead of some enum when there's a limited set of possibilities for those messages.

sway-core/src/language/ty/expression/expression_variant.rs
```diff
@@ -898,6 +898,116 @@ impl ReplaceDecls for TyExpressionVariant {
...
+                for elem in contents.iter() {
+                    elem.type_check_analyze(handler, ctx)?
+                }
+            }
+            TyExpressionVariant::ArrayIndex { prefix, index } => {
+                prefix.type_check_analyze(handler, ctx)?;
+                index.type_check_analyze(handler, ctx)?;
+            }
+            TyExpressionVariant::StructExpression { fields: _, .. } => {
+                // for field in fields.iter() {
```
IGI-111, line 938: Does this need to be removed?

sway-core/src/language/ty/expression/expression_variant.rs
```diff
@@ -898,6 +898,116 @@ impl ReplaceDecls for TyExpressionVariant {
...
+            TyExpressionVariant::CodeBlock(block) => {
+                block.type_check_analyze(handler, ctx)?;
+            }
+            TyExpressionVariant::FunctionParameter => {}
+            TyExpressionVariant::MatchExp {
+                desugared,
+                scrutinees: _,
+            } => {
+                desugared.type_check_analyze(handler, ctx)?;
+                // for scrutinee in scrutinees.iter() {
```
IGI-111, line 951: Does this need to be removed?

sway-core/src/lib.rs
```diff
@@ -484,6 +485,14 @@ pub fn parsed_to_ast(
 
     typed_program.check_deprecated(engines, handler);
 
+    // // Analyze the AST for dependency information.
+    let mut ctx = TypeCheckAnalysisContext::new(engines);
+    typed_program.type_check_analyze(handler, &mut ctx)?;
+
+    // // Finalize the AST with the dependency information.
```
IGI-111, line 492: More commented out stuff.

sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs
```diff
@@ -483,7 +537,56 @@ impl TyImplTrait {
...
+
+            // Do a topological sort to compute an ordered list of nodes (by function calls).
+            let sorted = match petgraph::algo::toposort(&sub_graph, None) {
+                Ok(value) => Some(value),
+                // If the dependency graph contains cycles, then this means there are recursive
+                // function calls, which we will report later.
+                Err(_) => None,
+            };
+
+            // let mut impl_trait = decl_engine.get_impl_trait(&impl_self.decl_id);
```
IGI-111, line 578: More commented out stuff.

sway-core/src/semantic_analysis/type_check_analysis.rs
```diff
@@ -0,0 +1,249 @@
...
+                    &|_, er| format!("label = {:?}", er.weight()),
+                    &|_, nr| {
+                        let _node = &self.dep_graph[nr.0];
+                        let shape = "";
+                        // if self.entry_points.contains(&nr.0) {
+                        //     shape = "shape=doubleoctagon";
+                        // }
+                        let url = "".to_string();
+                        if let Some(_url_format) = print_graph_url_format.clone() {
+                            // if let Some(span) = node.span() {
```
IGI-111, line 153: More commented out stuff.

sway-core/src/lib.rs
```diff
@@ -484,6 +485,10 @@ pub fn parsed_to_ast(
 
     typed_program.check_deprecated(engines, handler);
 
+    // // Analyze the AST for dependency information.
```
IGI-111, line 488:
```suggestion
    // Analyze the AST for dependency information.
```

IGI-111 (approved):
:shipit:

IGI-111 (approved):

PR #5137, opened 2023-09-20
fix: take `include_tests` into account for module AST cache

### LINKED ISSUE #5136: `forc test` not recognizing tests in contracts
From @bitzoic 

1. Create a new toolchain fuelup toolchain new v0.46
2. Install Forc fuelup component add forc@0.46.0
3. Install Fuel-core fuelup component add fuel-core@0.20.4
4. Create a new folder mkdir forc_test
5. Init a new project forc init
6. Add a test
7. Run forc test

```sway
contract;

abi MyContract {
    fn test_function() -> bool;
}

    fn test_function() -> bool {
        true
    }
}

#[test]
fn my_test() {
    assert(false);
}
```
This SHOULD fail. Yet I get that no tests were run

kayagokalp:
## Description
closes #5136.

We build contract projects twice, once without tests to get contract bytecode for forc-test's deployment step. (Since contract project's unit tests needs to make contract calls, we need the contract deployed to local vm instance.) And once with the tests included if the user executes `forc test`. This is only needed for contract projects as we only need to deploy them for testing.

Module cache addition in 255350ec5018a4a5cea9f0a2280a62830262201a only takes path into account for key. But same source file with and without tests enabled creates different AST's and thus we should also take `include_tests` variable into account. 

We were missing contract tests because we could not build them with tests. 1st run were building them correctly without the tests and populated the cache, only taking `Path` into account. Second time, although we passed `include_tests = true`, (thus the resulting AST is going to be different for the same file) rather than creating the AST with the tests, we were using the cached version. This was causing contract's to lose tests entries.

## Checklist


kayagokalp:
Curios about what you think @tritao about the fix.

sdankel (approved):
This makes sense to me! Are you planning on adding tests for forc-test in a separate PR?

kayagokalp:
I will add the tests now @sdankel and mark this one ready

forc-test/src/lib.rs
```diff
@@ -838,6 +839,10 @@ mod tests {
     const TEST_DATA_FOLDER_NAME: &str = "test_data";
     /// Name of the library package in the "CARGO_MANIFEST_DIR/TEST_DATA_FOLDER_NAME".
     const TEST_LIBRARY_PACKAGE_NAME: &str = "test_library";
+    /// Name of the contract package in the "CARGO_MANIFEST_DIR/TEST_DATA_FOLDER_NAME".
+    const TEST_CONTRACT_PACKAGE_NAME: &str = "test_contract";
+    /// Name of the predicate package in the "CARGO_MANIFEST_DIR/TEST_DATA_FOLDER_NAME".
+    const TEST_PREDICATE_PACKAGE_NAME: &str = "test_predicate";
```
sdankel, line 845: Should we add one for scripts as well?

kayagokalp: Ahh you are right, I totally forget about it 👍

forc-test/src/lib.rs
```diff
@@ -857,6 +862,42 @@ mod tests {
...
+        let contract_package_dir_string = contract_package_dir.to_string_lossy().to_string();
+        let build_options = Opts {
+            pkg: forc_pkg::PkgOpts {
+                path: Some(contract_package_dir_string),
+                ..Default::default()
+            },
+            ..Default::default()
+        };
+        build(build_options)
+    }
```
sdankel, line 881: nit: we could DRY this up by having one function `test_built_tests(package_name: &str)`

forc-test/src/lib.rs
```diff
@@ -871,6 +912,34 @@ mod tests {
         }
     }
 
+    fn test_contract_test_results(
```
sdankel, line 915: Similarly, this could be one function `get_test_results(package_name: &str, test_filter: Option<TestFilter>)`

sdankel (approved):

PR #5148, opened 2023-09-27
Refactor use of destructive methods in `swayfmt`

### LINKED ISSUE #3473: Parts of the formatter use destructive methods
We should avoid using `.pop()` and the like, and find work arounds for the current use cases that implement it. This will help us avoid creating unparsable code resulting in vague errors.

It might be a good idea if the pop is necessary that we have a function that uses a token as a guard so that we only pop in cases where it is absolutely necessary and we know it is safe to do so.

eureka-cpu:
Closes #3473 
Removes use of `.pop` used directly on the end formatted `String`. Better prioritizes order of operations, and writing into buffers to achieve the same result.


swayfmt/src/utils/language/punctuated.rs
```diff
@@ -78,6 +68,47 @@ where
...
+{
+    let mut ps: Vec<String> = value_separator_pairs
+        .iter()
+        .map(|(_, p)| -> Result<FormattedCode, FormatterError> {
+            let mut buf = FormattedCode::new();
+            p.format(&mut buf, formatter)?;
+            Ok(buf)
+        })
+        .collect::<Result<_, _>>()?;
+    let mut ts: Vec<String> = value_separator_pairs
```
JoshuaBatty, line 88: Is it possible to iterate over `value_separator_pairs` once instead of twice?

JoshuaBatty:
```rust
    let mut ts = Vec::new();
    let mut ps = Vec::new();

    for (t, p) in value_separator_pairs.iter() {
        let mut t_buf = FormattedCode::new();
        let mut p_buf = FormattedCode::new();
        t.format(&mut t_buf, formatter)?;
        p.format(&mut p_buf, formatter)?;
        ts.push(t_buf);
        ps.push(p_buf);
    }
```

or something similar.

cr-fuel (approved):

kayagokalp (commented):
+1 on Josh's comment otherwise looks good to me.

swayfmt/src/utils/language/punctuated.rs
eureka-cpu: It may look nicer to use a `for` loop but it's actually more effective to use `collect` since it doesn't require continual reallocation of memory into a mutable vector. `collect` actually creates a new vec with a known capacity, avoiding those expensive reallocations. 🙂

JoshuaBatty: you could pre-allocate the capacity of the vector upfront then.

JoshuaBatty: like this

```rust
let len = value_separator_pairs.len();
let mut ts = Vec::with_capacity(len);
let mut ps = Vec::with_capacity(len); 
```

JoshuaBatty (approved):

PR #5153, opened 2023-09-28
#4318 `forc-crypto` plugin exposing a CLI for common-use cryptographic operations

### LINKED ISSUE #4318: Add a `forc-crypto` plugin exposing a CLI for common-use cryptographic operations
Raised by @SilentCicero in slack.

The idea is to expose common-use fuel cryptographic functions like `sha256`, `keccak256`, `ecrecover` via forc plugin(s).

The original proposal was for a `forc utils` subcommand with a broader scope, though I'm thinking it could be worth having more descriptive plugins for each set of tools.

A forc-crypto plugin might include commands like the following:

- `forc crypto sha256 ...`
- `forc crypto keccak256 ...`
- `forc crypto ecrecover ...`
- `forc crypto bech32-to-hex ...`
- `forc crypto hex-to-bech32 ...`

Commands like `sha256` and `keccak256` should support both file and string arguments. The same likely goes for the ecrecover message.

The plugin should likely be shipped by `fuelup` as a part of the default set of components.

This plugin should go in its own repo, but we can use this issue to track until then.

cr-fuel:
## Description

Fixes #4318

Supported algorithms:
* [x] keccak256
* [x] sha256
* [ ] ~ecrecover~ (this will be added later)
* [x] bech32-to-hex
* [x] hex-to-bech32
* [x] new-key / parse-secret (from https://github.com/FuelLabs/fuel-core/blob/master/bin/keygen/src/keygen.rs)

## Checklist


.github/workflows/ci.yml
```diff
@@ -204,10 +204,14 @@ jobs:
         with:
           toolchain: ${{ env.RUST_VERSION }}
       - uses: Swatinem/rust-cache@v2
-      - name: 'Build Workspace'
+      - name: Install Protoc
```
JoshuaBatty, line 207: Seems like there a lot of these additions. Can you explain why these were added? Doesn't seem related to this PR but maybe i'm wrong.

forc-plugins/forc-crypto/src/content.rs
```diff
@@ -0,0 +1,63 @@
...
+        if let Ok(text) = String::from_utf8(input.clone()) {
+            let text = text.trim();
+            if let Some(text) = text.strip_prefix("0x") {
+                if let Ok(bin) = hex::decode(text) {
+                    return bin;
+                }
+            }
+            text.as_bytes().to_vec()
+        } else {
+            input
```
JoshuaBatty, line 27: Do we really want to return the input that was passed in here? Perhaps we should be throwing an error instead..

```rust
Err(anyhow!("Invalid input"))
```

forc-plugins/forc-crypto/src/content.rs
```diff
@@ -0,0 +1,63 @@
...
+        if s == "-" {
+            if STDIN_HAS_BEEN_USED.load(Ordering::Acquire) {
+                let mut buffer = vec![];
+                let _ = io::stdin().lock().read_to_end(&mut buffer)?;
+                return Ok(Content::Raw(Self::from_hex_or_raw(buffer)));
+            }
+            STDIN_HAS_BEEN_USED.store(true, Ordering::SeqCst);
+            Ok(Content::Raw(vec![]))
+        } else {
+            let path = std::path::PathBuf::from(s);
```
JoshuaBatty, line 45: match statement might be a little cleaner here. 
```suggestion
            let path = PathBuf::from(s);
            match read(&path) {
                Ok(content) => Ok(Content::Path(path, Self::from_hex_or_raw(content))),
                Err(_) => Ok(Content::Raw(Self::from_hex_or_raw(s.as_bytes().to_vec()))),
            }
```

forc-plugins/forc-crypto/src/content.rs
```diff
@@ -0,0 +1,63 @@
...
+use std::{
+    fs::read,
+    io::{self, Read},
+};
+
+static STDIN_HAS_BEEN_USED: AtomicBool = AtomicBool::new(false);
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Content {
+    Path(std::path::PathBuf, Vec<u8>),
```
JoshuaBatty, line 12:
```suggestion
    Path(PathBuf, Vec<u8>),
```

forc-plugins/forc-crypto/src/content.rs
```diff
@@ -0,0 +1,63 @@
+use std::str::FromStr;
```
JoshuaBatty, line 1:
```suggestion
```

forc-plugins/forc-crypto/src/content.rs
```diff
@@ -0,0 +1,63 @@
...
+use std::sync::atomic::{AtomicBool, Ordering};
```
JoshuaBatty, line 2:
```suggestion
```

forc-plugins/forc-crypto/src/ecrecover.rs
```diff
@@ -0,0 +1,8 @@
+use fuel_crypto::{secp256r1::recover, Signature};
```
JoshuaBatty, line 1: Lets get rid of this whole file as we aren't using it.

forc-plugins/forc-crypto/src/keygen/mod.rs
```diff
@@ -0,0 +1,11 @@
+use clap::ValueEnum;
+
+pub mod new_key;
+pub mod parse_secret;
+
```
JoshuaBatty, line 5: Lets make consts here that can be used by the files below. 

```rust
const BLOCK_PRODUCTION: &str = "block_production";
const P2P: &str = "p2p";
```

forc-plugins/forc-crypto/src/keygen/new_key.rs
```diff
@@ -0,0 +1,66 @@
...
+#[derive(Debug, clap::Args)]
+#[clap(author, version, about = "Creates a new key for use with fuel-core")]
+pub struct Arg {
+    #[clap(long = "pretty", short = 'p')]
+    pretty: bool,
+    #[clap(
+        long = "key-type",
+        short = 'k',
+        value_enum,
+        default_value = "block-production"
```
JoshuaBatty, line 28: Is this supposed to be `block_production` instead of `block-production`?

forc-plugins/forc-crypto/src/keygen/new_key.rs
```diff
@@ -0,0 +1,66 @@
...
+    let public_key = secret.public_key();
+    let secret_str = secret.to_string();
+
+    let output = match arg.key_type {
+        KeyType::BlockProduction => {
+            let address = Input::owner(&public_key);
+            json!({
+                "secret": secret_str,
+                "address": address,
+                "type": "block_production"
```
JoshuaBatty, line 45:
```suggestion
                "type": BLOCK_PRODUCTION
```

forc-plugins/forc-crypto/src/keygen/new_key.rs
```diff
@@ -0,0 +1,66 @@
...
+        KeyType::Peering => {
+            let mut bytes = *secret.deref();
+            let p2p_secret = secp256k1::SecretKey::from_bytes(&mut bytes)
+                .expect("Should be a valid private key");
+            let libp2p_keypair = Keypair::Secp256k1(p2p_secret.into());
+            let peer_id = PeerId::from_public_key(&libp2p_keypair.public());
+            json!({
+                "secret": secret_str,
+                "peer_id": peer_id.to_string(),
+                "type": "p2p"
```
JoshuaBatty, line 57:
```suggestion
                "type": P2P
```

forc-plugins/forc-crypto/src/keygen/new_key.rs
JoshuaBatty: There seems to be inconsistencies here and the next file. Please double check which it's supposed to be.

forc-plugins/forc-crypto/src/keygen/parse_secret.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+pub fn handler(arg: Arg) -> Result<String> {
+    let secret = SecretKey::from_str(&arg.secret)?;
+    let output = match arg.key_type {
+        KeyType::BlockProduction => {
+            let address = Input::owner(&secret.public_key());
+            let output = json!({
+                "address": address.to_string(),
+                "type": "block-production",
```
JoshuaBatty, line 39:
```suggestion
                "type": BLOCK_PRODUCTION,
```

forc-plugins/forc-crypto/src/keygen/parse_secret.rs
```diff
@@ -0,0 +1,61 @@
...
+        }
+        KeyType::Peering => {
+            let mut bytes = *secret.deref();
+            let p2p_secret = secp256k1::SecretKey::from_bytes(&mut bytes)
+                .expect("Should be a valid private key");
+            let libp2p_keypair = Keypair::Secp256k1(p2p_secret.into());
+            let peer_id = PeerId::from_public_key(&libp2p_keypair.public());
+            let output = json!({
+                "peer_id": peer_id.to_string(),
+                "type": "p2p"
```
JoshuaBatty, line 51:
```suggestion
                "type": P2P
```

forc-plugins/forc-crypto/src/main.rs
```diff
@@ -0,0 +1,65 @@
...
+        std::process::exit(1);
+    }
+}
+
+fn run() -> Result<()> {
+    let app = Command::parse();
+    let content = match app {
+        Command::Keccak256(arg) => hex::encode(keccak256::hash(arg.content)?),
+        Command::Sha256(arg) => hex::encode(sha256::hash(arg.content)?),
+        //Command::E => hex::encode(ecrecover::ecrecover(app.content)?),
```
JoshuaBatty, line 57: remove this line

forc-plugins/forc-crypto/src/main.rs
```diff
@@ -0,0 +1,65 @@
+//! A `forc` plugin for converting a given string o path to their hash
+
+use anyhow::Result;
+use clap::Parser;
+use forc_tracing::{init_tracing_subscriber, println_error};
+use std::default::Default;
```
JoshuaBatty, line 6: Is this used? Please remove if not

forc-plugins/forc-crypto/src/maybe_file.rs
```diff
@@ -0,0 +1,65 @@
+use std::str::FromStr;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::{
+    fs::read,
+    io::{self, Read},
+};
+
+static STDIN_HAS_BEEN_USED: AtomicBool = AtomicBool::new(false);
```
JoshuaBatty, line 8: same concerns about the global atomic as mentioned above.

forc-plugins/forc-crypto/src/maybe_file.rs
```diff
@@ -0,0 +1,65 @@
...
+
+/// Pretty simple parser for a string that can be a path, "-" (to read content
+/// from stdin), a hex string or a raw string.
+#[derive(Clone, Debug, PartialEq)]
+pub enum Content {
+    Path(std::path::PathBuf, Vec<u8>),
```
JoshuaBatty, line 14:
```suggestion
    Path(PathBuf, Vec<u8>),
```

forc-plugins/forc-crypto/src/maybe_file.rs
```diff
@@ -0,0 +1,65 @@
+use std::str::FromStr;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::{
```
JoshuaBatty, line 3:
```suggestion
use std::{
    fs::Read,
    io::{self, Read},
    path::PathBuf,
    str::FromStr,
    sync::atomic::{AtomicBool, Ordering},
```

forc-plugins/forc-crypto/src/maybe_file.rs
```diff
@@ -0,0 +1,65 @@
+use std::str::FromStr;
```
JoshuaBatty, line 1:
```suggestion
```

forc-plugins/forc-crypto/src/maybe_file.rs
```diff
@@ -0,0 +1,65 @@
...
+use std::sync::atomic::{AtomicBool, Ordering};
```
JoshuaBatty, line 2:
```suggestion
```

forc-plugins/forc-crypto/src/maybe_file.rs
```diff
@@ -0,0 +1,65 @@
...
+        if s == "-" {
+            if STDIN_HAS_BEEN_USED.load(Ordering::Acquire) {
+                let mut buffer = vec![];
+                let _ = io::stdin().lock().read_to_end(&mut buffer)?;
+                return Ok(Content::Raw(Self::from_hex_or_raw(buffer)));
+            }
+            STDIN_HAS_BEEN_USED.store(true, Ordering::SeqCst);
+            Ok(Content::Raw(vec![]))
+        } else {
+            let path = std::path::PathBuf::from(s);
```
JoshuaBatty, line 47: same thing here where a match statement would be cleaner. 

Even better, these seems to be identical to the logic in `content.rs` can you abstract this logic into a function that can be shared by both files.

JoshuaBatty (changes requested):
Would you mind also adding a video of it being used to the description. thanks.

forc-plugins/forc-crypto/src/content.rs
```diff
@@ -0,0 +1,63 @@
...
+static STDIN_HAS_BEEN_USED: AtomicBool = AtomicBool::new(false);
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Content {
+    Path(std::path::PathBuf, Vec<u8>),
+    Raw(Vec<u8>),
+}
+
+impl Content {
+    pub fn from_hex_or_raw(input: Vec<u8>) -> Vec<u8> {
```
kayagokalp, line 17: I have little concerns about this. 

1. The name makes the impression that this will be returning a `Content`, but it returns an `Vec<u8>`. It is a little confusing.

2. TL;DR: How do we differentiate the prefix bytes from actual bytes if they are `[48, 120]`? IIUC, this function tries to see if the given input is `Raw` or in `Hex` form by checking whether text converted version of the input starts with `0x`. I am not sure how we collect this `Raw` input from the user and why it differs from the hex. My intuition is that they are in the following forms:

*For `0x1a2d`*:

- Hex input: `[48, 120, 49, 97, 50, 100]`
- Raw input: `[26, 45]`

and the expected outcome for this function is to return the same thing for both inputs provided above which is: `[26, 45]`. Is this correct?

If this is correct what happens if the raw input I want to provide starts with byte equivalents of `0x`. For example what happens if the raw input is `[48, 120, 100, 100]`. This will be translated to `0xd` and first if branch will be taken. The output of the function would be `[221]` because 0x will be trimmed and `dd` will be decoded by `hex` crate to `[221]` but the expected output was `[48, 120, 100, 100]`?

Cargo.toml
```diff
@@ -24,11 +25,7 @@ members = [
     "swayfmt",
     "test",
 ]
-exclude = [
-    "examples/*",
-    "swayfmt/test_macros",
-    "forc-test/test_data"
-]
+exclude = ["examples/*", "swayfmt/test_macros", "forc-test/test_data"]
```
kayagokalp, line 28: Not sure how we have this change, did `rustfmt` version changed?

kayagokalp (changes requested):

forc-plugins/forc-crypto/src/main.rs
```diff
@@ -0,0 +1,65 @@
...
+    let content = match app {
+        Command::Keccak256(arg) => hex::encode(keccak256::hash(arg.content)?),
+        Command::Sha256(arg) => hex::encode(sha256::hash(arg.content)?),
+        //Command::E => hex::encode(ecrecover::ecrecover(app.content)?),
+        Command::Address(arg) => address::dump_address(arg.content)?,
+        Command::NewKey(arg) => keygen::new_key::handler(arg)?,
+        Command::ParseSecret(arg) => keygen::parse_secret::handler(arg)?,
+    };
+
+    println!("{}", content);
```
kayagokalp, line 63:
```suggestion
    println!("{}", content);
```

Rather than printing can we use `info` from tracing? Since we are starting tracing subscriber we can use it. This way anyone who wants to filter or print in a different format can do that easily.

forc-plugins/forc-crypto/src/address.rs
```diff
@@ -0,0 +1,30 @@
+use anyhow::anyhow;
+use fuel_crypto::fuel_types::Address;
+use fuels_core::types::bech32::Bech32Address;
+use std::str::FromStr;
+
+/// Takes a valid address in any supported format and returns them in all
+/// supported format. This is meant to be a tool that can be used to convert any
+/// address format to all other formats
+pub fn dump_address<T: AsRef<[u8]>>(data: T) -> anyhow::Result<String> {
```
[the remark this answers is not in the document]
kayagokalp: This looks much readable so +1 here with Josh. I feel like this also increases reusability, I can see `handle_string_conversion` reused somewhere later on.

Also looks like you don't need to go to `String` for the purpose of converting it to a `Bech32Address` or `Address`. You seem like in need of a `&str` since they use `from_str()` and thus you can use [`str::from_utf8()`](https://doc.rust-lang.org/std/str/fn.from_utf8.html) which does not allocate, but uses `mem::transmute` [under the hood](https://doc.rust-lang.org/src/core/str/converts.rs.html#173).

.github/workflows/ci.yml
cr-fuel: @JoshuaBatty that is because protoc (protobuf compiler) is needed in order to build the dependencies needed for the `new-key` and `parse-secret` which were ported from https://github.com/FuelLabs/fuel-core/blob/master/bin/keygen/src/keygen.rs

forc-plugins/forc-crypto/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+fuel-core-types = { version = "0.20.5" }
+fuel-crypto = { workspace = true }
+fuels-core = { workspace = true }
+futures = "0.3"
+hex = "0.4.3"
+libp2p-core = { version = "0.38", features = ["secp256k1"] }
+rand = "0.8"
+serde = "1.0"
+serde_json = "1"
+sha2 = "0.10.8"
```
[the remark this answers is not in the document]
cr-fuel: Probably not, I'll revisit this.

forc-plugins/forc-crypto/src/content.rs
```diff
@@ -0,0 +1,63 @@
+use std::str::FromStr;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::{
+    fs::read,
+    io::{self, Read},
+};
+
+static STDIN_HAS_BEEN_USED: AtomicBool = AtomicBool::new(false);
```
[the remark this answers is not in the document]
cr-fuel: I found a strange issue where STDIN is parsed twice, I couldn't find the reason. I guess I can just make the argument be a file _or_ a raw string (and never caring to tread from stdin). Should I make that happen?

.github/workflows/ci.yml
JoshuaBatty: Perhaps we shouldn't be porting these then. Maybe a better solution is to make the functionality accessible as a public method from fuel-core itself. @Voxelot would this seem reasonable to you?

forc-plugins/forc-crypto/src/content.rs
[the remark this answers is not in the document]
JoshuaBatty: This sounds like another issue then. I would prefer another issue be opened that outlined how to reproduce this behaviour and have it tackled in a follow up PR. Is this happening for all forc CLI plugins?

cr-fuel:
https://github.com/FuelLabs/sway/assets/142530682/350fe6cd-9a80-4edc-8186-b777f080e300

Cargo.toml
cr-fuel: It was my toml fmt. I'll revert this.

forc-plugins/forc-crypto/src/content.rs
cr-fuel: I agree. We should accept valid UTF-8 data *or* any raw binary data encoded as hexadecimal

forc-plugins/forc-crypto/src/main.rs
cr-fuel: It is being used on line 17

forc-plugins/forc-crypto/src/content.rs
cr-fuel: @kayagokalp I improved the logic. Although it is taking a vector of u8, it is expecting it to be a valid UTF-8 string. Anything else will error. Binary data should be encoded as hexadecimal strings.

forc-plugins/forc-crypto/src/address.rs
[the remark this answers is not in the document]
cr-fuel: Done in a837b9c59d

.github/workflows/ci.yml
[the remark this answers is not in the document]
cr-fuel: Done in 4c4b9e491f

forc-plugins/forc-crypto/src/keygen/new_key.rs
cr-fuel: It seems to be `block_production` https://github.com/search?q=org%3AFuelLabs+block_production&type=code vs https://github.com/search?q=org%3AFuelLabs+block-production&type=code

cr-fuel:
@JoshuaBatty I think the copied bits (inside `forc-plugins/forc-crypto/src/keygen/*.rs`) can be merged as is. I will open a follow up MR that will make the original code shareable as a lib, then I will include them in here. 

I will also try to find a way of installing them without adding a protoc dependency. I really don't like that.

.github/workflows/ci.yml
```diff
@@ -125,7 +126,7 @@ jobs:
           mdbook-version: '0.4.25'
       - name: Emit book logs to tmp.txt, fail if build logs contain 'ERROR'
         run: |
-          MDBOOK_preprocessor__FORC_documenter__STRICT="true" mdbook build docs/book &> tmp.txt
+          MDBOOK_preprocessor__FORC_documenter__STRICT="true" mdbook build docs/book 2>&1 | tee tmp.txt
```
cr-fuel, line 129: Using `tee` instead of `&>` will help debug those issues that only happens on the CI by printing and capturing the `stderr` even when `mdbook` exits with any status code other than `0`.

kayagokalp (commented):
CI looks green again, nice 👌 Just left couple of questions and a nit otherwise looking good to me!

forc-plugins/forc-crypto/src/content.rs
```diff
@@ -0,0 +1,54 @@
...
+};
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Content {
+    Path(PathBuf, Vec<u8>),
+    Binary(Vec<u8>),
+}
+
+impl Content {
+    pub fn from_hex_or_utf8(input: Vec<u8>) -> Result<Vec<u8>> {
```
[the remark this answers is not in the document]
JoshuaBatty: +1 on what Kaya's points here.

forc-plugins/forc-crypto/src/keygen/new_key.rs
```diff
@@ -0,0 +1,64 @@
+//! This file is migrated from https://github.com/FuelLabs/fuel-core/blob/master/bin/keygen/src/keygen.rs
```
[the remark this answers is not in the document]
cr-fuel: Yes, it is already done https://github.com/FuelLabs/fuel-core/pull/1397, but I'm waiting for a fix to their Cargo.toml to be able to include the code directly. Until that happens it will be copied. I have a follow up ticket already #5176 5170

forc-plugins/forc-crypto/src/keygen/parse_secret.rs
```diff
@@ -0,0 +1,59 @@
+//! This file is migrated from https://github.com/FuelLabs/fuel-core/blob/master/bin/keygen/src/keygen.rs
```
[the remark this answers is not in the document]
cr-fuel: Yes, it is already done https://github.com/FuelLabs/fuel-core/pull/1397, but I'm waiting for a fix to their Cargo.toml to be able to include the code directly. Until that happens it will be copied. I have a follow up ticket already https://github.com/FuelLabs/sway/pull/5176 5170

forc-plugins/forc-crypto/src/content.rs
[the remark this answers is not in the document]
cr-fuel: @JoshuaBatty @kayagokalp  I think fb73b65e9 makes it better

forc-plugins/forc-crypto/src/content.rs
```diff
@@ -0,0 +1,42 @@
+use std::{convert::Infallible, fs::read, path::PathBuf, str::FromStr};
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Content {
+    Path(PathBuf, Vec<u8>),
+    Binary(Vec<u8>),
+}
+
+impl Content {
+    pub fn from_file_or_binary(input: &str) -> Self {
```
JoshuaBatty, line 10: Can we add a doc comment here? It's not obvious what `input` is supposed to be supplied here. Without reading the internals there is no way to know a `str` with `0x` will be interpreted as a binary. 

Another option would be to split this into 2 functions. 
1. `from_path`
2. `from_binary_str`

forc-plugins/forc-crypto/src/content.rs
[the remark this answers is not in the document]
JoshuaBatty: I've left a comment on the changes, we can resolve this specific conversation and take it up there.

forc-plugins/forc-crypto/src/keygen/parse_secret.rs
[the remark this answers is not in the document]
JoshuaBatty: In this situation I would either like to wait for the upstream changes to be available, or for an issue to be opened and a link to the issue be left in the comments so we don't forget to clean up.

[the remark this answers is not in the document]
JoshuaBatty: I will be happy with just including this URL in the comments for now https://github.com/FuelLabs/sway/issues/5170

forc-plugins/forc-crypto/src/content.rs
cr-fuel: @JoshuaBatty I think 0f3729898 is a much better elegant solution. Do you that this approach is simpler, more readable and better?

JoshuaBatty: Can we add some tests for it?

forc-plugins/forc-crypto/src/keygen/parse_secret.rs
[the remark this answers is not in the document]
cr-fuel: Done in 6c4545c18

forc-plugins/forc-crypto/src/content.rs
cr-fuel: Done in 6c4545c18

JoshuaBatty (approved):
Looks good cheers.

forc-plugins/forc-crypto/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "forc-crypto"
+version = "0.46.0"
```
JoshuaBatty, line 3:
```suggestion
version = "0.46.1"
```

forc-plugins/forc-crypto/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+license.workspace = true
+repository.workspace = true
+
+
+[dependencies]
+anyhow = "1.0.75"
+async-trait = "0.1.58"
+atty = "0.2.14"
+clap = { version = "3", features = ["derive", "env"] }
+forc-tracing = { version = "0.46.0", path = "../../forc-tracing" }
```
JoshuaBatty, line 17:
```suggestion
forc-tracing = { version = "0.46.1", path = "../../forc-tracing" }
```

JoshuaBatty (changes requested):

JoshuaBatty:
looks like CI is complaining about the Cargo.lock needing updating. Happy to approve once CI is green.

JoshuaBatty (dismissed):

forc-plugins/forc-crypto/src/keygen/parse_secret.rs
```diff
@@ -0,0 +1,65 @@
...
+            output
+        }
+    };
+    display_string_discreetly(
+        &(if arg.pretty {
+            serde_json::to_string_pretty(&output)
+        } else {
+            serde_json::to_string(&output)
+        })?,
+        "### Do not share or lose this private key! Press any key to complete. ###",
```
sdankel, line 61: Is this warning needed? In this case we're just showing them the pubic key and type.

forc-plugins/forc-crypto/src/sha256.rs
```diff
@@ -0,0 +1,8 @@
+use fuel_crypto::Hasher;
+
+/// Hashes a given data to Sha256
+pub fn hash<T: Into<Vec<u8>>>(data: T) -> anyhow::Result<Vec<u8>> {
+    let mut hasher = Hasher::default();
```
sdankel, line 5: Same comment as `keccak256.rs` about error handling & input validation.

sdankel (commented):
Looking good overall! I left some nits, questions, and found a few edge cases to fix/test.

forc-plugins/forc-crypto/src/keccak256.rs
```diff
@@ -0,0 +1,8 @@
+use sha3::{Digest, Keccak256};
+
+/// Hashes a given data using Keccak256
+pub fn hash<T: Into<Vec<u8>>>(data: T) -> anyhow::Result<Vec<u8>> {
+    let mut hasher = Keccak256::new();
+    hasher.update(data.into());
+    Ok(hasher.finalize().to_vec())
```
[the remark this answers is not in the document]
cr-fuel: Passing `-1` does not work because the crate  will attempt to parse. I just fixed `""` parsing in cac394f58 (it was attempting to read from stdin when the first argument was empty. Now it'll do if the first argument is `-` or not given).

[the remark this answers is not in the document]
cr-fuel: However `echo -1 |  forc-crypto keccak256` works as expected.

forc-plugins/forc-crypto/src/keygen/new_key.rs
```diff
@@ -0,0 +1,70 @@
...
+            json!({
+                "secret": secret_str,
+                "peer_id": peer_id.to_string(),
+                "type": P2P
+            })
+        }
+    };
+    display_string_discreetly(
+        &(if arg.pretty {
+            serde_json::to_string_pretty(&output)
```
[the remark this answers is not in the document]
cr-fuel: I like that idea. I will make it happen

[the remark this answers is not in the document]
cr-fuel: I like your formatting. I think it would be nicer to print like in your example if we are printing to the stdout and JSON if the output is being piped to another file. What are your thoughts on that @JoshuaBatty ?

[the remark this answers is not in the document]
cr-fuel: That work is done in 360f52cfd. I think that way is better:

1. It will print as yaml to the stdout
2. It will not show the `press a key warning` nor clear the stdout unless there is a `secret` about to be printed
3. All the command handlers uses the same function to output

😄

forc-plugins/forc-crypto/src/sha256.rs
cr-fuel: I missed the context here @sdankel. Would you mind giving it again please? Anyway, the error was that on empty first argument `stdin` was read. That is fixed now.

forc-plugins/forc-crypto/src/keccak256.rs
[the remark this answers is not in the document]
sdankel: What about `forc crypto keccak256 "-1"`? Should it work for any string, in this case a string containing "-1"?

forc-plugins/forc-crypto/src/content.rs
```diff
@@ -0,0 +1 @@
+
```
sdankel, line 1: Should this file be deleted?

forc-plugins/forc-crypto/src/keygen/new_key.rs
```diff
@@ -0,0 +1,60 @@
...
+    let public_key = secret.public_key();
+    let secret_str = secret.to_string();
+
+    let output = match arg.key_type {
+        KeyType::BlockProduction => {
+            let address = Input::owner(&public_key);
+            json!({
+                "secret": secret_str,
+                "address": address,
+                "type": <KeyType as std::convert::Into<&'static str>>::into(KeyType::BlockProduction),
```
sdankel, line 42: nit: Does `arg.key_type.into()` not work?

forc-plugins/forc-crypto/src/args.rs
```diff
@@ -0,0 +1,136 @@
...
+#[derive(Debug, Clone, clap::Args)]
+#[clap(author, version, about = "Hashes the argument or file with this hash")]
+pub struct HashArgs {
+    /// This argument is optional, it can be either:
+    ///
+    ///     1. A path to a file. If that is the case, the content of the file is
+    ///        loaded.
+    ///
+    ///     2. A binary string encoded as a hex string. If that is the case, the
+    ///        hex is decoded a Vec<u8> is passed
```
sdankel, line 18:
```suggestion
    ///     2. A binary string encoded as a hex string. If that is the case, the
    ///        hex is decoded and passed as a Vec<u8>
```

forc-plugins/forc-crypto/src/args.rs
```diff
@@ -0,0 +1,136 @@
...
+    ///
+    ///     3. A string. This is the last option, if the string is "-", `stdin`
```
sdankel, line 20: This seems to be how it works, is this intended? (If the string starts with "-")

```suggestion
    ///     3. A string. This is the last option, if the string starts with "-", `stdin`
```

forc-plugins/forc-crypto/src/keygen/new_key.rs
```diff
@@ -0,0 +1,60 @@
...
+};
+use libp2p_identity::{secp256k1, Keypair, PeerId};
+use serde_json::json;
+use std::ops::Deref;
+
+/// Generate a random new secret & public key in the format expected by fuel-core
+#[derive(Debug, clap::Args)]
+#[clap(author, version, about = "Creates a new key for use with fuel-core")]
+pub struct Arg {
+    /// Key type to generate. It can either be `block-production` or `p2p`.
```
sdankel, line 20: This is what the help message looks like: 
image


It says that `p2p` is a possible value, which doesn't seem to work: 
image

`peering` does work as a value: `forc crypto new-key -k peering`

forc-plugins/forc-crypto/src/keygen/parse_secret.rs
```diff
@@ -0,0 +1,56 @@
...
+#[derive(Debug, clap::Args)]
+#[clap(
+    author,
+    version,
+    about = "Parses a private key to view the associated public key"
+)]
+pub struct Arg {
+    /// A private key in hex format
+    secret: String,
+    /// Key type to generate. It can either be `block-production` or `p2p`.
```
sdankel, line 20: Same issue here.

forc-plugins/forc-crypto/src/keygen/new_key.rs
cr-fuel: It does not, I can however do it in a separated statement. 

```
let type: &str = arg.key_type.into();
```

Should I do that?

forc-plugins/forc-crypto/src/args.rs
cr-fuel: No, the string has to be `-`

forc-plugins/forc-crypto/src/keygen/parse_secret.rs
cr-fuel: My understanding is that peer_id and addresses are derived differently. Technically I _think_ a private key can be either an address or a peer_id.

forc-plugins/forc-crypto/src/keygen/new_key.rs
kayagokalp: I would personally prefer `let type: &str = arg.key_type.into();` happy with it either way though!

kayagokalp (approved):
Looks nice 👌 Haven't tested the latest version locally but changes from the last review looks good to me! utACK

sdankel (approved):
Looks good!

Issue #5052, opened 2023-08-29
Run forc-fmt on standard library in CI


sdankel:
There may be bugs in the formatter that we need to fix before doing this.

eureka-cpu:
@andrewvious

sdankel:
@cr-fuel Would you like to take on one of these related issues? I think any of them would be a good introduction to the formatter.

cr-fuel:
@sdankel absolutely.

eureka-cpu:
relevant PR: #4875

eureka-cpu:
Seems this issue is a duplicate of #2917

PR #5178, opened 2023-10-09
Fix fmt whitespaces between comments and doc comments

cr-fuel:
This is part of #5052

## Description

There was a bug in the formatter whenever a comment as followed by a doc-comment, this was due the `insert_after_span` was not allowed to return a negative offset (due types constraints). But sometimes this function should let their callee know to remove some bytes.

## Checklist

### DIFF
--- a/swayfmt/src/utils/map/newline.rs
+++ b/swayfmt/src/utils/map/newline.rs
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use ropey::Rope;
-use std::{collections::BTreeMap, fmt::Write, path::PathBuf, sync::Arc};
+use std::{collections::BTreeMap, path::PathBuf, sync::Arc};
 use sway_ast::Module;
 use sway_types::SourceEngine;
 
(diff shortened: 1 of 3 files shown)


swayfmt/src/utils/map/newline.rs
```diff
@@ -270,22 +283,31 @@ fn insert_after_span(
...
-    )?;
+) -> Result<i64, FormatterError> {
+    let sequence_string = format_newline_sequence(&newline_sequence, threshold);
+    let mut len = sequence_string.len() as i64;
     let mut src_rope = Rope::from_str(formatted_code);
 
+    // Remove the previous sequence_length, that will be replaced in the next statement
+    let mut remove_until = at;
+    for i in at..at + newline_sequence.sequence_length {
+        if src_rope.get_char(i) != Some('\n') {
```
eureka-cpu, line 294: I think there is a constant you could use for `'\n'` somewhere in the formatter

swayfmt/src/utils/map/newline.rs
```diff
@@ -270,22 +283,31 @@ fn insert_after_span(
...
+            break;
+        }
+        remove_until = i;
+    }
+    if remove_until > at {
+        let _ = src_rope.try_remove(at..remove_until);
```
eureka-cpu, line 300: We shouldn't be ignoring `Result`s, generally, unless there's a good reason for it. In this case if you are trying to avoid a `panic` then maybe the length shouldn't change if you're unable to remove it?
Can you try changing this to:
```rust
if let Err(_) = src_rope.try_remove(at..remove_until) {
    len -= (remove_until - at) as i64;
}
```
I may not have the full context here, feel free to ignore this if my suggestion doesn't work.

eureka-cpu (commented):
Nice one, I left a few suggestions.

JoshuaBatty (approved):
Nice work.

eureka-cpu (approved):
Thanks @cr-fuel

PR #5182, opened 2023-10-11
bug: Fix formatting for `ItemTraitItem` and `Annotated<T>`

### LINKED ISSUE #5179: Fix formatting for `ItemTraitItem` and `Annotated<T>` to be used in both `ItemTrait` and `ItemAbi`
Items that rely on formatting of `ItemTraitItem` and `Annotated<T>` are not using them correctly. The formatting for `ItemTraitItem` is duplicated, and the `Format` implementation isn't actually used. `Annotated<T>` relies on external indentation.

eureka-cpu:
Closes #5179 

What this does:
- Centralizes the use case of `ItemTraitItem`'s format method
- Fixes indentation problems around `Annotated<T>`
- Adds a `write_ident_into_buffer` method which checks if the buffer is already indented
- Adds documentation for methods the interact with the indentation of the formatter to avoid confusion on what they actually do

### DIFF
--- a/swayfmt/src/comments.rs
+++ b/swayfmt/src/comments.rs
@@ -103,7 +103,7 @@ pub fn write_comments(
                     write!(
                         formatted_code,
                         "{}{}{}",
-                        formatter.indent_str()?,
+                        formatter.indent_to_str()?,
                         comment.span().as_str(),
                         newlines
                     )?;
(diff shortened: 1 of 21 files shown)


JoshuaBatty (approved):
utACK

cr-fuel (approved):

PR #5186, opened 2023-10-12
Fix formatting of `abi` & `trait` function signatures

### LINKED ISSUE #4285: Formatter should turn long `abi` methods into multi lines
Example:
```rust
contract;

enum MyEnum<V> {
    Foo: u64,
    Bar: bool,
}
struct MyStruct<T, U> {
    bim: T,
    bam: MyEnum<u64>,
}
struct MyOtherStruct {
    bom: u64,
}

abi MyContract {
    fn complex_function(arg1: MyStruct<[b256; 3], u8>, arg2: [MyStruct<u64, bool>; 4], arg3: (str[5], bool), arg4: MyOtherStruct) -> str[6];
}

    fn complex_function(
        arg1: MyStruct<[b256; 3], u8>,
        arg2: [MyStruct<u64, bool>; 4],
        arg3: (str[5], bool),
        arg4: MyOtherStruct,
    ) -> str[6] {
        "fuel42"
    }
}
```
This is the output of the formatter. Note that `complex_function` is treated differently between the `abi` block and its `impl`. I expect to see the method signature turned into multiple lines in both cases.

eureka-cpu:
Closes #4285 

What this does:
- Adds the necessary information in `Shape` for `FnSignature`'s `Format` implementation, to always format as a `Function`. This bug occurred because `FnSignature` always inherited the parent type's `ExprKind`, meaning it will format correctly so long as the parent type (`ItemFn`) exists. Otherwise it will default to the normal format.
- Adds tests for `ItemAbi` & `ItemTrait`
- Removes an unnecessary indentation from `ItemImpl`


cr-fuel (approved):

swayfmt/src/items/item_impl/mod.rs
```diff
@@ -53,7 +53,6 @@ impl Format for ItemImpl {
             formatter.indent();
             writeln!(formatted_code)?;
             for item in contents.iter() {
-                write!(formatted_code, "{}", formatter.indent_to_str()?,)?;
```
sdankel, line 56: Is this covered by the tests?

sdankel (approved):

swayfmt/src/items/item_impl/mod.rs
eureka-cpu: Yes, since the item is annotated

PR #5193, opened 2023-10-13
bug: Update `swayfmt::Config` & remove partial implementations

### LINKED ISSUE #3472: Certain parts of the `Config` are half-way implemented
There are sections in `swayfmt` that implement parts of the config that are not the default, such as `always_next_line`. We should remove these for the time being or implement them entirely.

eureka-cpu:
Closes #3472 

What this does:
- Removes use of `Config` options that were partially implemented
- Fixes the default implementations for `Config`. This removes the match arms in the implementations of the `CurlyBrace` trait and just handles the default, eg `ItemBraceStyle::SameLineWhere`
- Adds tests for items with `where` clauses, ie `ItemImpl`, `ItemStruct`, `ItemTrait`, which weren't being tested thoroughly enough
- Updates to code quality in the `Config`, and documentation for the default config's options


JoshuaBatty:
Thanks @eureka-cpu. I feel like adding a doc comment specifying what each field defaults too is a little over kill, especially in places where the default impl is already specified by hand. 

Are we able to remove these lines?

swayfmt/src/config/expr.rs
```diff
@@ -4,50 +4,61 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Copy, Clone)]
 pub struct Expressions {
-    /////PUNCTUATION/////
+    // PUNCTUATION
```
sdankel, line 7: Shouldn't this be:

```suggestion
    /// PUNCTUATION
```

swayfmt/src/items/item_abi/mod.rs
```diff
@@ -77,19 +76,10 @@ impl CurlyBrace for ItemAbi {
...
-                // Add opening brace to the next line.
-                write!(line, "\n{open_brace}\n")?;
-            }
-            _ => {
-                // Add opening brace to the same line
-                writeln!(line, " {open_brace}")?;
-            }
-        }
+        // Add opening brace to the same line
+        writeln!(line, " {open_brace}")?;
```
sdankel, line 82: Why does this change?

swayfmt/src/items/item_storage/mod.rs
```diff
@@ -134,19 +134,10 @@ impl CurlyBrace for ItemStorage {
...
-                // Add opening brace to the next line.
-                write!(line, "\n{open_brace}")?;
-            }
-            _ => {
-                // Add opening brace to the same line
-                write!(line, " {open_brace}")?;
-            }
-        }
+        // Add opening brace to the same line
+        write!(line, " {open_brace}")?;
```
sdankel, line 140: Do we have tests for `different brace_style`s?

eureka-cpu: Yeah, the tests are for actual use cases of each type of brace. So for say some array, the test would just be that the array is formatted how you would expect. There isn't any specific test other than that for each implementation of the trait.

sdankel (approved):

sdankel:
Nice simplification! I think there are some failing CI tests, but otherwise looks good!

JoshuaBatty (approved):

JoshuaBatty (changes requested):
whoops didn't mean to approve. @eureka-cpu can we please remove the comments that state what everything defaults to and instead create a markdown file for this information?

JoshuaBatty (approved):
Cheers

sdankel (approved):

PR #5197, opened 2023-10-16
Support for more than 6 arguments in functions, without having to inline

### LINKED ISSUE #2904: Backend: Add support for passing args to functions on the stack.
As of #2843 we can make function calls.  Args to functions are passed in 6 reserved registers.  If a function has more than 6 args those extra args need to be passed via the stack.

We currently forcibly inline these functions to get around this limitation.  Once we add stack based args we can remove this constraint.

vaivaswatha:
## Description
Fixes #2904


sway-core/src/asm_generation/fuel/functions.rs
```diff
@@ -66,21 +70,73 @@ impl<'ir, 'eng> FuelAsmBuilder<'ir, 'eng> {
         args: &[Value],
     ) -> Result<(), CompileError> {
         // Put the args into the args registers.
-        for (idx, arg_val) in args.iter().enumerate() {
-            if idx < compiler_constants::NUM_ARG_REGISTERS as usize {
+        if args.len() <= compiler_constants::NUM_ARG_REGISTERS as usize {
```
xunilrj, line 73: Changing `compiler_constants::NUM_ARG_REGISTERS` will make calling `fn` with different convention an invisible error.

Will this be a problem? Will this affect contract call, library calls etc...

vaivaswatha: What do you mean by a different convention? if someone uses another compiler that uses different conventions for calling, then yes. but otherwise this should be ok. This change impacts all function calls except calls to entry functions (from external sources). Those have a different convention and are already capable of more than NUM_ARG_REGISTERS args.

xunilrj: I mean, if the caller uses `NUM_ARG_REGISTERS=2`, for example, and the the callee `NUM_ARG_REGISTERS=6` or whatever, there will be some calls that will error.

Caller is putting the third argument in the stack, but the callee is actually expecting it as a register.

Just wondering how hard/impactful would be to change these things.

vaivaswatha: >if the caller uses `NUM_ARG_REGISTERS=2`, for example, and the the callee `NUM_ARG_REGISTERS=6`

That would only happen if the caller and the callee were separately compiled with different (versions of the) compiler(s). I'm guessing that that's not a scenario we need to worry about.

IGI-111: This wouldn't be able to happen except through contract calls which use a different convention, correct @vaivaswatha ?

vaivaswatha: That's right, yes. 

We don't have a "linking" phase. We compile the whole program every time. So the whole program will use the same calling convention. And as you mentioned, entry points have a different convention altogether.

xunilrj: That was my concern. If this is not our use case then I am fine. :)

Issue #4705, opened 2023-06-26
`forc-doc`: Add links to dependency types that go to their definition 


eureka-cpu:
As of #4546 we now have dependency docs built and can start adding links to these docs so that something like: 

```rust
// dep
mod foo {
    pub trait Foo { ... }
}

dep foo;

impl Foo for MyType { ... }

```
Will produce a link to `Foo` on the page for `MyType`

JoshuaBatty:
Related to #4371

PR #5205, opened 2023-10-18
Add go-to definition for dependency types: `ImplTraits`

eureka-cpu:
Ref #4705 

What this does:
- Adds go-to definition functionality for external traits implemented for some local type, eg `impl Foo for Bar` where `Foo` is a dependency type
- Adds `no_deps` to `RenderPlan`, when this flag is passed dependencies aren't built, so we also shouldn't produce links to docs that won't be there
- Adds a test harness that uses the Sway -> HTML compiler 

First part of the video shows the output when built with `--no-deps` which does not build dependency docs (in this instance `core`), and thus produces no links to those dependencies. Second shows without the flag.

[Screencast from 2023-10-20 01-54-57.webm](https://github.com/FuelLabs/sway/assets/57543709/fb3a1275-19fa-47ad-a4d3-551211383226)


cr-fuel (dismissed):

JoshuaBatty:
Can we add some tests please?

sdankel (approved):

JoshuaBatty (approved):
nice one

PR #5207, opened 2023-10-19
Attempt to break long expressions into multiple lines

### LINKED ISSUE #3474: Chained methods & fields that should be multiline need logic to handle vertical formatting
```rust
fuelcoin.really_long_field.other_really_long_field.foo().bar().baz().quux();
```
should be:
```rust
fuelcoin
    .really_long_field
    .other_really_long_field
    .foo()
    .bar()
    .baz()
    .quux();
```

To elaborate further, this applies only to `swayfmt`. Logic needs to be added to handle the formatting of Sway code so that chained methods & fields that exceed a certain length (IIRC it's 60 but we are using the `rustfmt` standard which you can find [here](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=chain#chain_width)) are formatted as multiline chains.

`rustfmt` has a file dedicated to handling formatting of chains, [here](https://github.com/rust-lang/rustfmt/blob/master/src/chains.rs).

cr-fuel:
Fixes #3474

## Description

Breaks long expressions (focusing into chained methods calls but will cover all expressions) into multiple lines

## Checklist

### DIFF
--- a/examples/identity/src/main.sw
+++ b/examples/identity/src/main.sw
@@ -54,9 +54,12 @@ impl IdentityExample for Contract {
     fn access_control_with_identity() {
         // ANCHOR: access_control_with_identity
         let sender = msg_sender().unwrap();
-        require(sender == storage
-            .owner
-            .read(), MyError::UnauthorizedUser(sender));
-            // ANCHOR_END: access_control_with_identity
+        require(
+            sender == storage
+                .owner
+                .read(),
+            MyError::UnauthorizedUser(sender),
+        );
+        // ANCHOR_END: access_control_with_identity
     }
 }
--- a/examples/identity/src/main.sw
+++ b/examples/identity/src/main.sw
@@ -54,7 +54,9 @@ impl IdentityExample for Contract {
     fn access_control_with_identity() {
         // ANCHOR: access_control_with_identity
         let sender = msg_sender().unwrap();
-        require(sender == storage.owner.read(), MyError::UnauthorizedUser(sender));
-        // ANCHOR_END: access_control_with_identity
+        require(sender == storage
+            .owner
+            .read(), MyError::UnauthorizedUser(sender));
+            // ANCHOR_END: access_control_with_identity
     }
 }
--- a/examples/vec/src/main.sw
+++ b/examples/vec/src/main.sw
@@ -27,9 +27,7 @@ fn main() {
     let mut i = 0;
     while i < v
         .len() {
-        log(v
-            .get(i)
-            .unwrap());
+        log(v.get(i).unwrap());
         i += 1;
     }
     // ANCHOR_END: vec_iterate
@@ -41,11 +39,17 @@ fn main() {
     }
 
     let mut row = Vec::new();
-    row.push(TableCell::Int(3));
+    row.push(
+        TableCell::Int(3),
+    );
     row
         .push(
-            TableCell::B256(0x0101010101010101010101010101010101010101010101010101010101010101),
+            TableCell::B256(
+                0x0101010101010101010101010101010101010101010101010101010101010101,
+            ),
         );
-    row.push(TableCell::Boolean(true));
+    row.push(
+        TableCell::Boolean(true),
+    );
     // ANCHOR_END: vec_multiple_data_types
 }
--- a/examples/vec/src/main.sw
+++ b/examples/vec/src/main.sw
@@ -38,17 +38,13 @@ fn main() {
     }
 
     let mut row = Vec::new();
-    row.push(
-        TableCell::Int(3),
-    );
+    row.push(TableCell::Int(3));
     row
         .push(
             TableCell::B256(
                 0x0101010101010101010101010101010101010101010101010101010101010101,
             ),
         );
-    row.push(
-        TableCell::Boolean(true),
-    );
+    row.push(TableCell::Boolean(true));
     // ANCHOR_END: vec_multiple_data_types
 }
--- a/examples/vec/src/main.sw
+++ b/examples/vec/src/main.sw
@@ -25,8 +25,7 @@ fn main() {
     // ANCHOR_END: vec_get_oob
     // ANCHOR: vec_iterate
     let mut i = 0;
-    while i < v
-        .len() {
+    while i < v.len() {
         log(v.get(i).unwrap());
         i += 1;
     }
--- a/examples/vec/src/main.sw
+++ b/examples/vec/src/main.sw
@@ -25,8 +25,11 @@ fn main() {
     // ANCHOR_END: vec_get_oob
     // ANCHOR: vec_iterate
     let mut i = 0;
-    while i < v.len() {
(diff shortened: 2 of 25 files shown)


swayfmt/tests/mod.rs
```diff
@@ -1716,3 +1718,175 @@ impl OrdEq for u256 {}
...
+            1123123,
+            12312,
+            312,
+            312,
+            3123,
+            12,
+            31,
+            44,
+            [1, 2],
+            true,
```
eureka-cpu, line 1827: For now this is okay, but should probably emit a warning, eg Rust emits a warning for too many function parameters.

This shouldn't be emitted in the formatter though, and I don't think any users would do anything like this.

would you mind making the number of function parameters here something more reasonable? If the goal is to make it format into multiline for just the method call then I'm sure there is a better approach that won't eat up so many lines.

eureka-cpu (approved):
Nice one @cr-fuel !

Left one small nit, but feel free to ignore.

eureka-cpu (changes requested):
Looks like there are some failing checks.

The format check is failing because the changes you've introduced have not been applied to the examples directory. Also looks like there may be some compiler warnings causing the clippy check to fail.

examples/identity/src/main.sw
```diff
@@ -54,7 +54,9 @@ impl IdentityExample for Contract {
     fn access_control_with_identity() {
         // ANCHOR: access_control_with_identity
         let sender = msg_sender().unwrap();
-        require(sender == storage.owner.read(), MyError::UnauthorizedUser(sender));
-        // ANCHOR_END: access_control_with_identity
+        require(sender == storage
+            .owner
+            .read(), MyError::UnauthorizedUser(sender));
```
eureka-cpu, line 59: This looks a bit strange, but I'm fine with a follow up if you'd prefer to get this PR merged for now.

Ideally it would look like:
```sway
require(
    sender == storage.owner.read(),
    MyError::UnauthorizedUser(sender),
);
```

examples/vec/src/main.sw
```diff
@@ -25,8 +25,11 @@ fn main() {
     // ANCHOR_END: vec_get_oob
     // ANCHOR: vec_iterate
     let mut i = 0;
-    while i < v.len() {
-        log(v.get(i).unwrap());
+    while i < v
+        .len() {
+        log(v
+            .get(i)
+            .unwrap());
```
eureka-cpu, line 32: This one also seems a bit weird, how it was previously looks correct to me

examples/identity/src/main.sw
cr-fuel: I like your format better.

examples/vec/src/main.sw
cr-fuel: I'm on it to fix it.

cr-fuel: Does it look better now @eureka-cpu ?

examples/identity/src/main.sw
cr-fuel: Does it look better now @eureka-cpu ?

examples/vec/src/main.sw
eureka-cpu: They do look better, I think there's a few odd ones here and there but overall it's a big improvement. I'll defer approval to @JoshuaBatty

examples/vec/src/main.sw
```diff
@@ -25,7 +25,8 @@ fn main() {
     // ANCHOR_END: vec_get_oob
     // ANCHOR: vec_iterate
     let mut i = 0;
-    while i < v.len() {
+    while i < v
+        .len() {
```
JoshuaBatty, line 29: yeah this is very strange to see, we should keep this rendering a `v.len()`. having len() on a new line isn't a good look.

JoshuaBatty (changes requested):

examples/vec/src/main.sw
cr-fuel: Any better now @JoshuaBatty ? I added a fix in e656a05d7

examples/wallet_contract_caller_script/src/main.sw
```diff
@@ -7,10 +7,16 @@ fn main() {
     let contract_address = 0x9299da6c73e6dc03eeabcce242bb347de3f5f56cd1c70926d76526d7ed199b8b;
     let caller = abi(Wallet, contract_address);
     let amount_to_send = 200;
-    let recipient_address = Address::from(0x9299da6c73e6dc03eeabcce242bb347de3f5f56cd1c70926d76526d7ed199b8b);
-    caller.send_funds {
```
JoshuaBatty, line 11: I feel like the original looks easier to read. What would cargo fmt produce here?

cr-fuel: @JoshuaBatty Perhaps I should tweak the defaults to allow longer lines? Or perhaps I can aim something like

```rust
    caller
        .send_funds {
            gas: 10000,
            coins: 0,
            asset_id: ZERO_B256,
        }(amount_to_send, recipient_address);
        ````

Comment (JoshuaBatty) [reply examples/wallet_contract_caller_script/src/main.sw:11]: Yeah I think that sounds good. that looks a lot nice IMO

Comment (cr-fuel) [reply examples/wallet_contract_caller_script/src/main.sw:11]: [Any better now](https://github.com/FuelLabs/sway/pull/5207/files#diff-08ef0bbba0955879505c2ae86a078223951e9fafa935c6af77679cf07609cd92R13-R18) @JoshuaBatty ?

Comment (JoshuaBatty) [reply examples/wallet_contract_caller_script/src/main.sw:11]: Yeah that's better, I still think the first part should be on the same line though no?

ie 

```sway
caller.send_funds {
   ....
````   

Here is some unrelated formatted code in rust 

```rust
self.source_to_path_map
    .read()
    .unwrap()
    .get(source_id)
    .unwrap()
    .clone()
```

cr-fuel: This is news to me, but in `cargo fmt` the second element in the chain will be inline if the first element is shorter than 5  letters.


```rust
   fuel.really_long_field
        .other_really_long_field
        .foo()
        .bar()
        .baz
        .quux()
        .yet_another_call()
        .to_go_above_max_line_length();
```

vs 

```rust
    fuelcore
        .really_long_field
        .other_really_long_field
        .foo()
        .bar()
        .baz
        .quux()
        .yet_another_call()
        .to_go_above_max_line_length();
  ```

I'll make this rule happen

cr-fuel: Does it look better now @JoshuaBatty ?

JoshuaBatty: agh ok didn't realise the rule was 5 chars. yep looks good thanks.

JoshuaBatty (approved):

sdankel (approved):
Nice work!

PR #5213, opened 2023-10-20
Refactor match expressions and fix critical issue 

### LINKED ISSUE #5122: Wrong variable binding and missed matching in OR `match` expressions
Consider the following code and the asserts that _all assert to true_:

```Sway
script;

struct Struct {  x: u64, y: u64, z: u64, }

fn main() {
    let s = Struct { x: 1, y: 2, z: 3, };

    let a = match s {
        Struct { x, y: 2, z: 3 } | Struct { x: 0, y: x, z: 0 } | Struct { x: 0, y: 0, z: x } => x,
        _ => 1111,
    };
    
    assert(a == 3); // BUG: Should be 1.

    let a = match s {
        Struct { x, y: 0, z: 0 } | Struct { x: 1, y: x, z: 3 } | Struct { x: 0, y: 0, z: x } => x,
        _ => 1111,
    };
    
    assert(a == 1111); // BUG: Should be 2.

    let a = match s {
        Struct { x, y: 0, z: 0 } | Struct { x: 0, y: x, z: 0 } | Struct { x: 1, y: 2, z: x } => x,
        _ => 1111,
    };

    assert(a == 1111); // BUG: Should be 3.

    let a = match s {
        Struct { x, y: 0, z: 0 } | Struct { x: 0, y: x, z: 0 } | Struct { x: 0, y: 0, z: x } => x,
        _ => 1111,
    };
    
    assert(a == 1111); // OK.
}
```
The reason for the first bug (3 instead of 1) is in the design of the [matcher](https://github.com/FuelLabs/sway/blob/44f46a0a905e3f99b40c2f0019f31ba487e20cb0/sway-core/src/semantic_analysis/ast_node/expression/match_expression/typed/matcher.rs#L173) that returns only the [MatchDeclMap](https://github.com/FuelLabs/sway/blob/44f46a0a905e3f99b40c2f0019f31ba487e20cb0/sway-core/src/semantic_analysis/ast_node/expression/match_expression/typed/matcher.rs#L24) of the last variant.

In other words, no matter which of the variants gets matched the value for the variable will always be extracted from the last variant.

Why the match-arm does not match at all in the other two examples is not immediately clear.

The consequence of this issue on the current design are:
- we need to change the [MatcherResult](https://github.com/FuelLabs/sway/blob/44f46a0a905e3f99b40c2f0019f31ba487e20cb0/sway-core/src/semantic_analysis/ast_node/expression/match_expression/typed/matcher.rs#L26) structure to return all `MatchDeclMap`s for all the alternatives.
- we need to change the desugared code to track which of the variants gets matched and to initialize the match-arm variable accordingly.

ironcev:
## Description

This PR fixes the critical issue #5122. It refactors the initial type checking step in the `matcher` by capturing the variable declarations and requirements in a tree structure that preserves the relationship between the requirements and declarations. This structure is traversed later on in other parts of type checking and desugaring. The desugaring is also fully refactored and produces semantically correct `if-else` chain in case of OR patterns.

The high level overview of the new type checking and desugaring together with examples of desugared `match` expressions can be found in the _match_expression/typed/mod.rs_ file.

Closes #5122.

## Checklist


ironcev:
Draft so far because the _dca/all_paths_return_ test is failing. This is because the DCA does not recognize `__revert()` as a dead end and the new match expression desugaring creates a trailing `else { __revert(KNOWN_CODE) }`.

The proposal is to bring the `__revert()` semantics to DCA and until we do that to temporary comment out the failing part of the _dca/all_paths_return_ test.

Otherwise, the PR is fully implemented.

ironcev:
The small part of the _dca/all_paths_return test_ that is failing is temporarily commented out until #5214 is implemented.

ironcev:
Back to draft. Need to check the `if let` issue generated by the new desugaring.

ironcev:
The [cargo-test-workspace](https://github.com/FuelLabs/sway/actions/runs/6603542041/job/17937466689) job is failing but that looks like a infrastructural issue. Here is the error message:

    The hosted runner: GitHub Actions 2 lost communication with the server.
    Anything in your workflow that terminates the runner process, starves it for CPU/Memory,
    or blocks its network access can cause this error.

All other jobs are passing, so I am confident to mark the PR as ready for review :-)

IGI-111:
Managing to run the tests, looks like there is still some breakage.

ironcev:
> Managing to run the tests, looks like there is still some breakage.

Yes I see it now. Going after it!

sway-core/src/semantic_analysis/ast_node/expression/match_expression/typed/matcher.rs
```diff
@@ -1,38 +1,142 @@
...
+
+/// A single variable in the form `let <ident> = <expression>`
+/// that has to be extracted from the match arm.
+pub(super) type MatchVarDecl = (Ident, ty::TyExpression);
+
+/// A leaf of a match pattern can be either a requirement on the scrutinee or a
+/// variable declaration but not both at the same time.
+/// In the case of the catch-all `_` we will have neither a requirement nor
+/// a variable declaration.
+pub(super) type ReqOrVarDecl = Option<Either<MatchReq, MatchVarDecl>>;
```
IGI-111, line 38: Instead of using `Option<Either<...>>` for a ternary union I think it would be clearer to use an enum with proper names for the cases that you later describe for this.

IGI-111 (commented):
Just one thing but otherwise LGTM.

I really appreciate how well commented this is.

sway-core/src/semantic_analysis/ast_node/expression/match_expression/typed/matcher.rs
ironcev: Agree! Done!

IGI-111 (approved):

PR #5226, opened 2023-10-25
Optimize gas for `AssetId::new()` and `AssetId::default()` functions

bitzoic:
## Description

The following changes remove the use of the `std::hash::sha256` function in place of assembly to hash the `ContractId` and `SubId` of an asset. As the values hashed are of static length, we can perform the hashing in assembly avoiding the conversion to `Bytes` and concatenation of the two values done in `sha256()`

These changes result in a **93%** reduction in gas usage.

## Checklist


K1-R1 (approved):

Braqzen (approved):
I question the addition of asm back into the std. We want it to be readable but at the same time we want it to be cheap.
Consolidation via a single function wrapper is also better for future changes as we only need to change it in 1 place.

PR #5230, opened 2023-10-26
Fixes missing error on type unification.

### LINKED ISSUE #5162: Generics and associated types clash
The following script compiles just fine, but shouldn't.

```sway
script;

trait TypeTrait {
    type T;

    fn method() -> Self::T;
} {
    fn method2<T>() -> T {
        Self::method()
    }
}

fn main() -> u32 {
    1
}
```

It would be nice to remove the "Self" treatment from `replace_self_type` (https://github.com/FuelLabs/sway/blob/103095e1f8332a0bbfaf0f5627973d7ea118050c/sway-core/src/type_system/id.rs#L166) because it conflicts with PR #4962 which is almost done.

### LINKED ISSUE #5212: Generic impl trait compiling when it shouldn't.
The following code compiles when it shouldn't.

```rust
pub trait MyFrom<T> {
    fn from(b: T) -> Self;
}

impl<T,U> MyFrom<T> for U {
    fn from(i: T) -> U {
        i
    }
}
```

esdrubal:
## Description

While using generic blanket implementation no error was thrown when another type other than the generic blanket expected type was used.

This was caused by the typechecker being too permissive in UnknownGeneric unification. The fix changes the unifier to handle unification with the expected self and generics as special cases.

Closes #5212
Closes #5162

## Checklist


sway-core/src/type_system/engine.rs
```diff
@@ -20,6 +20,12 @@ pub struct TypeEngine {
     id_map: RwLock<HashMap<TypeInfo, TypeId>>,
 }
 
+enum UnifyKind {
```
IGI-111, line 23: This needs extensive doccomments to explain that the differences between the unification modes is.

IGI-111:
If we're going to have multiple modes like this, it might be a good idea to refactor the unifier to work like `UnifyCheck`.

IGI-111 (approved):

JoshuaBatty (approved):

Issue #4997, opened 2023-08-22
Internal compiler error: Program erroneously uses uninitialized virtual registers: $r8


Braqzen:
Version `0.44.0`

The issue is the `len_to_hash` in the params of the `asm()`.
If you change `len_to_hash` to `len_to_hash: len_to_hash` it will compile.

```sway
script;

use std::constants::ZERO_B256;

fn main() {
    let len_to_hash = 32;
    let mut buffer = b256::min();

    asm(hash: buffer, prefix: ZERO_B256, len_to_hash) {
        k256 hash prefix len_to_hash;
    }
}
```

The order of the params is irrelevant.

PR #5231, opened 2023-10-26
Error for uninitialized register on asm blocks

xunilrj:
## Description

This PR closes https://github.com/FuelLabs/sway/issues/4997. The problem here is that "asm blocks registers" without initializers are considered uninitialized. There is no fallback to "capture" an available variable. This could be easily done, but that would leave us with no obvious syntax for unitialized registers.

With this in mind this PR solves the compiler bug with two diagnostics:

1 - uninitialized registers cannot be read before they are written. This generates an error:

```
 5 |     let _ = asm(r1) {
   |                 ^^ Unitialized register is being read before being written
 6 |         r1: u64
 7 |     };
```

2 - a warning when a unitialized register shadows an available variable with the same name. This is almost for sure a problem, and it is better the name the registe with something else and avoid all the confusion.

```
 27 |     let r5 = 0;
 28 |     asm(r5) {};
    |         -- This unitialized register is shadowing a variable, you probably meant to also initialize it like "r5: r5".
 29 | 
 30 |     0
```

## Checklist

### DIFF
--- a/sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
+++ b/sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
@@ -3614,13 +3614,6 @@ fn asm_register_declaration_to_asm_register_declaration(
         .value_opt
         .map(|(_colon_token, expr)| expr_to_expression(context, handler, engines, *expr))
         .transpose()?;
-    //     .unwrap_or_else(|| {
-    //         Expression {
-    //             kind: ExpressionKind::AmbiguousVariableExpression(asm_register_declaration.register.clone()),
-    //             span: asm_register_declaration.register.span(),
-    //         }
-    //     });
-    // dbg!(&initializer);
 
     Ok(AsmRegisterDeclaration {
         name: asm_register_declaration.register,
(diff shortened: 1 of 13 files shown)


sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
```diff
@@ -3610,12 +3610,21 @@ fn asm_register_declaration_to_asm_register_declaration(
...
+        .value_opt
+        .map(|(_colon_token, expr)| expr_to_expression(context, handler, engines, *expr))
+        .transpose()?;
+    //     .unwrap_or_else(|| {
+    //         Expression {
+    //             kind: ExpressionKind::AmbiguousVariableExpression(asm_register_declaration.register.clone()),
+    //             span: asm_register_declaration.register.span(),
+    //         }
+    //     });
+    // dbg!(&initializer);
```
IGI-111, line 3623: These comments probably need to be removed.

xunilrj: Done.

vaivaswatha (approved):

IGI-111 (approved):

PR #5234, opened 2023-10-27
Fixes unexpected trait dummy method error.

### LINKED ISSUE #5201: Internal compiler error: Method foo_6 is a trait method dummy and was not properly replaced.
The following example will generate the `Internal compiler error: Method foo_6 is a trait method dummy and was not properly replaced.` error message. 

This error is caused by the `bar()` function. Removing it will result in no errors on compilation.

***Note: This error prevents any implementation of the `Eq` trait on a generic enum or struct as the `not()` function is implemented in the same manner as the `bar()` function in the example.***

```sway
library;

struct MyStruct<T> {
    val: T
}

trait MyTrait {
    fn foo(self, other: Self) -> bool;
} {
    fn bar(self, other: Self) -> bool { // This function implementation is the problem
        self.foo(other)
    }
}

// Implement the trait MyTrait on MyStruct, where the generic also has
// the same trait implemented. This inherently isn't a problem, but the 
// bar() call will make this crash.
impl <T> MyTrait for MyStruct<T> where T: MyTrait {
    fn foo(self, other: Self) -> bool {
        self.val.foo(other.val)
    }
}

#[test]
fn foo() {
    let my_struct_1 = MyStruct { val: 5 };
    let my_struct_2 = MyStruct { val: 9 };

    // Calling foo() gives us the following expected error:
    // Trait "MyTrait" is not implemented for type "u64".
    // assert(my_struct_1.bar(my_struct_2));

    // But calling bar() gives us 
    // Internal compiler error: Method foo_5 is a trait method dummy and was not properly replaced.
    assert(my_struct_1.bar(my_struct_2));
}
```

Tested with forc v0.46.1

esdrubal:
## Description

The unexpected trait dummy method error is now replaced by: `Trait "MyTrait" is not implemented for type "u64".` This is the intended behavior and now throws the equivalent of rust error.

This commit also adds some improvements to the `DebugWithEngines`.

Closes #5201

## Checklist

### DIFF
--- a/sway-core/src/ir_generation/function.rs
+++ b/sway-core/src/ir_generation/function.rs
@@ -12,7 +12,7 @@ use crate::{
         compile_constant_expression, compile_constant_expression_to_constant,
     },
     language::{
-        ty::{self, ProjectionKind, TyConstantDecl},
+        ty::{self, ProjectionKind, TyConstantDecl, TyExpressionVariant},
         *,
     },
     metadata::MetadataManager,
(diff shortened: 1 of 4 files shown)


sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs
```diff
@@ -875,6 +875,29 @@ fn type_check_trait_implementation(
         match item {
             TyImplItem::Fn(decl_ref) => {
                 let mut method = decl_engine.get_function(decl_ref);
+
+                // We need to add impl type parameters to the  method's type parameters
```
ironcev, line 879:
```suggestion
                // We need to add impl type parameters to the method's type parameters
```

ironcev (dismissed):
Just that one supertiny typo otherwise :+1:

IGI-111 (approved):

ironcev (approved):

Issue #5214, opened 2023-10-21
Support `__revert` in dead code analysis


ironcev:
DCA does not support `__revert()` at the moment. E.g., in all of the examples below, dead code will not be reported on returning 42.

```Sway
fn revert_01() -> u64 {
    __revert(0);
 
    42
}
 
fn revert_02() -> u64 {
    __revert(0);
 
    return 42;
}
 
fn revert_03(a: u64) -> u64 {
    if a > 0 {
        __revert(0);
    }
    else {
        return 0;
    }
 
    return 42;
}
```
Question, DCA should then also be aware of `std::revert`?

Note that Rust does recognize `panic!` in its DCA. The equivalent examples in Rust would generate unreachable code warnings.

PR #5235, opened 2023-10-27
Fix DCA handling for `__revert` intrinsic and `std::revert`.

tritao:
## Description

This fixes dead code analysis for `__revert` and `std::revert`.

~~We add a new `NodeReturnOptions` return struct to `connect_expression` which can be used to signal to the parent code block handler that the expression sibling node should not have its leaves connected.~~

Fixes https://github.com/FuelLabs/sway/issues/5214.

## Checklist


ironcev:
On curious question (first time in the DCA :smile:). The solution with the `skip_leaves` is clear, intuitive, and straightforward. I am just curios, wouldn't we get the same effect if we connect `__revert` and `std::revert` calls to the `exit_node` the way `return` is connected? If that would work, it would spare us the need for `NodeReturnOptions` altogether.

test/src/e2e_vm_tests/test_programs/should_pass/dca/revert/src/main.sw
```diff
@@ -0,0 +1,43 @@
+script;
+
+fn revert_01() -> u32 {
+    __revert(0);
```
ironcev, line 4: For the completeness and proof, having the same tests for `std::revert` would be good.

ironcev (approved):
The solution is straightforward, but I am still curios if the same could be achieved if reverts are connected to the exit node? Essentially that we simulate returns.

test/src/e2e_vm_tests/test_programs/should_pass/dca/revert/src/main.sw
ironcev: Ups sorry seen it now! I expected visually to see `std::` :smile:

sway-core/src/control_flow_analysis/dead_code_analysis.rs
```diff
@@ -1263,6 +1268,17 @@ fn connect_expression<'eng: 'cfg, 'cfg>(
             for leaf in current_leaf {
                 graph.add_edge(leaf, fn_exit_point, "".into());
             }
+
+            // check for std::revert and connect to the exit node if thats the case.
```
ironcev, line 1272:
```suggestion
            // check for std::revert and connect to the exit node if that's the case.
```

ironcev (approved):
Trivial typo otherwise :+1: :+1:

ironcev (approved):

Issue #5184, opened 2023-10-11
Compiler should error when semicolons are missing on const declarations


sdankel:
Currently, [this contract](https://github.com/FuelLabs/fuels-ts/blob/caf64651fbd52ae4bab96d7168a68a2f22775547/packages/fuel-gauge/fixtures/forc-projects/multi-token-contract/src/main.sw) compiles without error. It should error out on these lines that are missing semicolons:

```
const TOKEN_1: b256 = 0x4a778acfad1abc155a009dc976d2cf0db6197d3d360194d74b1fb92b96986b00
const TOKEN_2: b256 = 0x0d000e76a67758bbc6861d48ca571876cd480d9df8cf4dfa635c168e1e97f324
const TOKEN_3: b256 = 0xdf78cb1e1a1b31fff104eb0baf734a4767a1b1373687c29a26bf1a2b22d1a3c5
```

This causes strange behavior in the formatter and LSP since the parsed AST is not correct.

PR #5236, opened 2023-10-27
correctly deal with missing semicolons

xunilrj:
## Description

This PR closes https://github.com/FuelLabs/sway/issues/5184, but I have the impression the same problem can happen with const in other places, and maybe even with others items.

The problem is that the parsing of `;` is very inconsistent at the moment. For example, `ItemConst` parses the semicolon, but only as a `peek`. This because a "parent parser" is actually deciding if the semicolon is needed or not.

That would be nice if it were more consistent.

## Checklist


JoshuaBatty (approved):
nice, thanks!

IGI-111 (approved):

PR #5240, opened 2023-10-30
Initialise large enough, same-const-initialised, arrays using loops

### LINKED ISSUE #5229: Optimize array initializations
The follow code
```
   1│script;
   2│
   3│fn main() -> u64 {
   4│    let mut a: [u64;10] = [11;10];
   5│
   6│    let mut i = 0;
   7│    let mut m = 0;
   8│    while i < 5 {
   9│        m += a[i];
  10│        i += 1;
  11│    }
  12│    m
  13│}
```

when compiled, results in an assembly that does the following:
1. Each index from [0,10), all compile time constants are stored in the data area.
2. For each index `i`:
  a. `i` is loaded from its data section slot, multiplied by the size and added to the base array address to compute the actual offset.
  b.  the literal `11` is loaded from its data slot.
  c. Loaded literal is stored to `a[i]`

Much lesser code can be generated if we perform 2a, 2b and 2c in a loop (at least when the array length is beyond a threshold). Starting with an initial index and incrementing it will avoid loading a different index everytime from the data area. So we avoid the loads as well as data area overuse.

With some smartness we can also avoid having to load `11` from its data slot every time. It could be done by LICM (when we write that).

vaivaswatha:
## Description
Closes #5229 

Previously, these arrays were initialised with repeated "load index from data area, store initialise constant to array". Now we just use a loop to store the constant initialise to every element of the array. To avoid overheads of the loop, it's only done, currently, when the array size is greater than 5 (can be tuned later).

### DIFF
--- a/sway-core/src/ir_generation/function.rs
+++ b/sway-core/src/ir_generation/function.rs
@@ -12,7 +12,7 @@ use crate::{
         compile_constant_expression, compile_constant_expression_to_constant,
     },
     language::{
-        ty::{self, ProjectionKind, TyConstantDecl},
+        ty::{self, ProjectionKind, TyConstantDecl, TyExpressionVariant},
         *,
     },
     metadata::MetadataManager,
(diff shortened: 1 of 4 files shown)


IGI-111 (approved):

vaivaswatha:
> Maybe I am misunderstanding, but the test doesn't actually check that the optimization is working right?

You're right, it just checks that, if the optimisation happens, it doesn't end up being incorrect.

> We would need an IR test for that?

Ideally yes. The problem is that this isn't really an optimisation pass. It's just a different IR gen, so I can't test that with `opt` like how we test the other optimisation passes :-(.

PR #5246, opened 2023-11-01
Introduce the `DEFAULT_SUB_ID` constant

bitzoic:
## Description

The following pull request does two things:
1. Introduces the `DEFAULT_SUB_ID` constant. This is equivalent to the `ZERO_B256` constant. This constant makes the Sway code easier to read and comprehend improving the developer's UX. The actual value of the default is something the developer no longer needs to think about.

Before:
```sway
let my_asset = AssetId::new(my_contract, ZERO_B256);
```
Now:
```sway
let my_asset = AssetId::new(my_contract, DEFAULT_SUB_ID);
```

2. Removes the `ContractId` argument from the `AssetId::default()` function. In 90% of cases `default()` is used within the contract that issued the token, making this an unnecessary argument. This function now fetches the contract id of the contract it's called in, thus eliminating the need to import another function and add additional arguments. In the case a developer needs the default of another contract, the `DEFAULT_SUB_ID` can be used instead making the distinction more legible and apparent. 

Before:
```sway
use std::call_frames::contract_id;

fn foo(other_contract: ContractId) {
     let other_asset = AssetId::default(other_contract);
     let my_asset = AssetId::default(contract_id());
}
```
After:
```sway
fn foo(other_contract: ContractId) {
     let other_asset = AssetId::new(other_contract, DEFAULT_SUB_ID);
     let my_asset = AssetId::default();
}
```

## Checklist


sway-lib-std/src/contract_id.sw
```diff
@@ -178,31 +178,24 @@ impl AssetId {
...
-    ///
-    ///     let asset_id = AssetId::default(contract_id);
-    ///
-    ///     assert(asset_id == AssetId::new(contract_id, sub_id));
+    ///     let asset_id = AssetId::default();
+    ///     assert(asset_id == AssetId::new(contract_id(), DEFAULT_SUB_ID));
     /// }
     /// ```
-    pub fn default(contract_id: ContractId) -> Self {
+    pub fn default() -> Self {
```
SwayStar123, line 197: I think it makes more sense to have a 'this_default' function than to repurpose this function

Braqzen (dismissed):
The new const is nice, wondering if we ought to alias instead.

Loving the before / after documentation in the PR.

sway-lib-std/src/contract_id.sw
Braqzen: `default` is canonical

bitzoic:
> The new const is nice, wondering if we ought to alias instead.
> 
> Loving the before / after documentation in the PR.

Alias would just be a `b256` with no value

SwayStar123:
@bitzoic Can you add a `default_for` which works like the old `default`

Braqzen (dismissed):
That's bloat for the lib. Don't add it. `new` can be used instead.

Braqzen (approved):

K1-R1 (approved):

Issue #4600, opened 2023-05-30
Test, benchmark and remove `local_copy_prop_prememcpy` in `memcpyopt.rs`


vaivaswatha:
1. The optimization pass was written during the IR refactor of #4336, and uses an ad-hoc algorithm. A much more systematic (data-flow based, but not a fully data-flow analysis) algorithm was later implemented in #4592. But that does not seem to cover everything that the ad-hoc algorithm did. This needs to be investigated and the function then removed, ensuring correctness and no penalty in generated code sizes.

2. The algorithm implemented in #4592, to optimize a sequence of `memcpy`s, does so by optimizing pairs in a loop. This is inefficient, and can probably be done in one go.

vaivaswatha:
When I worked on #4592, without `local_copy_prop_prememcpy`, the test `u256_ops_test` wouldn't compile. That was because, the new optimization inserted instructions at the beginning of the block, when the use was actually much later, thus increasing register pressure. With #4628 however, this issue is solved. `u256_ops_test` compiles fine now when `local_copy_prop_prememcpy` is removed. So we just need to benchmark and if the change is acceptable, remove it.

vaivaswatha:
@IGI-111 I experiemented with removing `local_copy_prop_prememcpy` and here the numbers.

My suggestion is to still go ahead removing that optimization pass (because it is badly written, at that time for an immediate urgent need). We can have another tracker issue to catch what patterns of memory copy propagation we're missing and try to incorporate that into the current pass (i.e., the one added in #4592).

<details>
<summary> Impact of removing `local_copy_prop_prememcpy` </summary>

test | before | after | % increase
-- | -- | -- | --
should_pass/non_payable_implicit_zero_coins | 240 | 248 | 3.33333333333333
should_pass/supertraits_for_abis_ownable | 2324 | 2316 | -0.344234079173838
should_pass/forc/contract_dependencies/contract_b | 68 | 68 | 0
should_pass/forc/contract_dependencies/contract_c | 68 | 68 | 0
should_pass/forc/contract_dependencies/contract_a | 324 | 348 | 7.40740740740741
should_pass/forc/dependency_package_field | 36 | 36 | 0
should_pass/forc/dependency_patching | 36 | 36 | 0
should_pass/forc/parent_pkg_manifest/contract_a | 68 | 68 | 0
should_pass/evm/evm_basic | 36 | 36 | 0
should_pass/language/basic_func_decl | 36 | 36 | 0
[...]

PR #5253, opened 2023-11-02
Do not memcpy optimize when passed as fn arg

### LINKED ISSUE #5250: Pass-by-value semantics not preserved when using __addr_of
Consider the following test that is passing on `v0.46.0` and the Sway latest master branch:

```Sway
script;

fn main() -> u64 {
   0
}

struct A {
  a: u64,
}

struct B {
  a: A,
  x: (u64, u64),
}

#[test]
fn ptr_issue() -> u64 {
    let a = A { a: 11 };
    let mut b = B { a: a, x: (11, 11) };

    let mut ptr_b = ptr(b);

    ptr_b.write(B { a: A { a: 22 }, x: (22, 22)});

    assert(b.a.a == 22);
    assert(b.x.0 == 22);
    assert(b.x.1 == 22);

    42
}

#[inline(never)]
fn ptr<T>(t: T) -> raw_ptr {
    __addr_of(t)
}
```

Because of the pass-by-value semantics `ptr` should get the copy of `b` and thus a different pointer should be returned from the function. (A pointer to the local value indeed, which is an issue on its own, but different then the one described here.)

Since `ptr` is actually getting the pointer to the original value and no copy is made writing to `ptr_b` overwrites `b`.

Notice that if we don't use `_addr_of` (or references one day) to the immutable parameters, the fact that the copy is not made is a valid optimization.

vaivaswatha:
This fixes a bug in `local_copy_prop_prememcpy` which didn't make an exception when a local gets passed its address into another function, thus making it unsafe to optimize it.

Closes #5250
Related to #4600


ironcev (approved):
Straightforward fix. There was no disabling `memcpy` when passing to `asm` before and now it's the same for function calls.

JoshuaBatty (approved):

PR #5255, opened 2023-11-04
Refactor access to memory layout information

### LINKED ISSUE #5227: Memory Layout
Whilst finishing up https://github.com/FuelLabs/sway/pull/4929 I realized that the memory layout code is scattered through out the code. This poses the obvious problem of having to keep everything in sync, but also will not allow the layout to be controlled using `#repr()` in the future.

This issue will work as a bookmark to places that needs to be refactored to improve and allow customization:

1 - Data Section: at `sway-core/src/asm_generation/fuel/data_section.rs`. Special attention will be needed at the `from_constant` because we don't have a way to describe a enum without hacks. See comments at line 98;

2 - Locals: at `sway-core/src/asm_generation/fuel/functions.rs` inside `alloc_locals` and `init_locals` we also need to take into consideration the memory layout of objects.

3 - Realizing loads: at `sway-core/src/asm_lang/allocated_ops.rs` inside `realize_load`.

4 - Size* Intrinsics: Intrinsics that return the size of types and instances also need to take memory layout into consideration.

5 - Calling conventions: We need to also consider all of our calling conventions:
5.1 - Entry call and return: We have a special case when the argument/return value fits into the VM register or not. Depending on the case the layout of basic data types change;
5.2 - Contract calls at `sway-core/src/ir_generation/function.rs` in `compile_contract_call` : Similiar to above;

6 - Storage: at `sway-core/src/ir_generation/storage.rs` 

7 - Sroa at `sway-ir/src/optimize/sroa.rs` needs to know the offset of each field and end up calling `get_indexed_type` at `sway-ir/src/irtype.rs` . Of course that in this same file `size_in_bytes` also needs to take layout into account.

A difficult question is if this will need to affect anything inside our IR. Maybe `GEP` will need to have access to type information.

ironcev:
## Description

This PR closes #5227 by refactoring access to memory layout information. The focus of the refactoring was removing the redundant code from the places which will be affected when references are introduced. Previously scattered in various places, knowledge about the memory layout of the types is now provided via (mostly) IR `Type` and (a tiny bit) IR `Constant`. Thus, the IR layer becomes the single source of truth for the memory layout. From this perspective, having other IR components like, e.g., SROA being aware of the memory layout is fine, because the memory layout assumptions are not leaking outside of the IR layer.

In addition, the PR fixes two bugs:
- allocation of 8x more space for local string arrays, then necessary
- wrong calculation of an offset of a union variant

The code outside of IR where an assumption on memory layout is still made is in the Storage API. Refactoring storage access code to remove memory layout assumptions would be a questionable effort because we anyhow want to refactor the Storage API. Therefore, a warning TODO-MEMLAY is left in the Storage API code as well as `assert`s.

The `size_bytes_round_up_to_word_alignment` macro is still used in exactly four places, where an alignment is needed, but not of a type size, then rather some other structure in memory like, e.g., locals on the stack.

The PR also fully document memory layout decisions and asserts them with unit test. 

The consumers of the memory layout information like, e.g., `DataSection` of course still need to be aware of the meaning of the layout information and use it according to their own use case. E.g.:
- the `DataSection` places constants according to the expected layout
- `asm_generation` `functions` module places local variables on the stack
- etc

The potential next steps are supporting more optimal layout of aggregates in terms of memory usage, or even providing configurable memory layouts. For this, there is a separate issue #5286. 

Closes #5227.

## Checklist


sway-core/src/asm_generation/fuel/data_section.rs
```diff
@@ -95,45 +74,24 @@ impl Entry {
...
-                        // Return here from our special case.
-                        return Entry::new_collection(vec![tag_entry, val_entry], name, padding);
-                    }
-                }
-            }
-        };
+        // We need a special handling in case of enums.
+        if constant.ty.is_enum(context) {
+            let tag_and_value_with_paddings = constant
+                .elements_of_aggregate_with_padding(context)
```
xunilrj, line 80: This is MUCH better...
but why not naming this like `data_section_layout_as_enum` or something like that.

`elements_of_aggregate_with_padding` is little opaque to me.

sway-core/src/asm_generation/fuel/data_section.rs
```diff
@@ -151,27 +109,12 @@ impl Entry {
...
-                        Entry::from_constant(
-                            context,
-                            el,
-                            None,
-                            Some(Padding::Right { target_size }),
-                        )
-                    })
+            ConstantValue::Array(_) | ConstantValue::Struct(_) => Entry::new_collection(
+                constant
+                    .elements_of_aggregate_with_padding(context)
```
xunilrj, line 114: Here could be `data_section_layout_as_array` etc...

xunilrj (approved):

xunilrj:
Would be possible to have one test that kind asserts all mem layouts decisions? Ideally a rust unit test instead of a e2e in sway.

sway-ir/src/irtype.rs
```diff
@@ -412,43 +431,88 @@ impl Type {
...
+        let field_idx = field_idx as usize;
+        let field_type = self.get_field_types(context)[field_idx];
+        let union_size_in_bytes = self.size(context).in_bytes();
+        let field_size_in_bytes = field_type.size(context).in_bytes();
+
+        // The union fields are at offset (union_size - field_size) due to left padding.
+        Some((union_size_in_bytes - field_size_in_bytes, field_type))
+    }
+
+    pub fn size(&self, context: &Context) -> TypeSize {
```
IGI-111, line 481: I hate long function names and this of all things is probably the best thing to call "size" without qualification, but you should either have a doc comment that clarifies we're talking about the size in the memory representation or make the name more explicit. I think the former is best.

sway-ir/Cargo.toml
```diff
@@ -14,6 +14,7 @@ downcast-rs = "1.2.0"
 filecheck = "0.5"
 generational-arena = "0.2"
 itertools = "0.10.3"
+once_cell = "1.18.0"
```
IGI-111, line 17: This isn't actually used.
```suggestion
```

ironcev: Actually it is: `use once_cell::sync::Lazy;` Used only once :smile: in unit tests to save a line of code in every test setup.

xunilrj (approved):

xunilrj (approved):

IGI-111 (approved):

PR #5263, opened 2023-11-09
Add newer control flow opcodes

dmihal:
## Description

Added newer control flow opcodes


vaivaswatha:
Related to #4804

vaivaswatha:
@dmihal Can you test if this works for you? If you have a unit test, please push to the branch, or alternatively share it here so that i can add it.

dmihal:
Looking into this now, some code is missing, but I'm looking at adding it now.

Some notes:
* Bigint math opcodes like `WQCM` are also missing
* Could we have some CI in the future that automatically checks that new opcodes are available in assembly?

IGI-111:
These should be added to https://github.com/FuelLabs/sway/blob/master/test/src/e2e_vm_tests/test_programs/should_fail/asm_disallowed_opcodes/src/main.sw should they not?

vaivaswatha:
> These should be added to https://github.com/FuelLabs/sway/blob/master/test/src/e2e_vm_tests/test_programs/should_fail/asm_disallowed_opcodes/src/main.sw should they not?

This is a good point. But it looks like @dmihal is actually using these in assembly.

IGI-111:
Well it looks like we have a test that enforced the rule that you're not allowed to use control flow opcodes in asm blocks, which I guess is why we didn't bother to implement these opcodes previously.

Now I'm open to the possibility of loosening that rule given asm blocks are explicitly unsafe, but first I would like to understand why we forbade that inclusion in the first place. The reason for it may be important.

dmihal:
Yes, we need these opcodes for our basic version of LDC

I imagine that control flow opcodes were blocked previously because they essentially allow you to jump "out" of the assembly block. However given that we currently have a real need to use them, and as @IGI-111 mentioned `asm` is already unsafe, I'd recommend us loosening this rule

IGI-111:
I'm closing this in favor of https://github.com/FuelLabs/sway/issues/5563 since it looks like we can do without jumps using that intrinsic

PR #5265, opened 2023-11-10
Fixes error while using generic trait with multiple impls.

### LINKED ISSUE #5262: Unexpected error when using multiple generic trait impls.
```rust
script;

pub trait MyFrom<T> {
    fn from(b: T) -> Self;
}


pub trait MyInto<T> {
    fn into(self) -> T;
}


impl<T, U> MyInto<U> for T
where
    U: MyFrom<T>,
{
    fn into(self) -> U {
        U::from(self)
    }
}

struct Struct3 {
    data: u64,
}

impl MyFrom<u64> for Struct3 {
    fn from(i: u64) -> Struct3 {
        Struct3 {data: i}
    }
}

impl MyFrom<u32> for Struct3 {
    fn from(i: u32) -> Struct3 {
        Struct3 {data: 1u64}
    }
}

fn main() -> bool {
    let _i: Struct3 = 1_u64.into();
    let _i: Struct3 = 1_u32.into();

    true
}
```
The code above throws the following error but it shouldn't.
```
 65 |     let _i: Struct3 = 1_u64.into();
 66 |     let _i: Struct3 = 1_u32.into();
    |                             ^^^^ Trait "MyFrom" is not implemented for type "Struct3".
```

This is one of the errors found in https://github.com/FuelLabs/sway/pull/5259

### LINKED ISSUE #5271: Expecting error on trait with where clause of multiple generics of same traits unimplemented.
This issue was found from a code review of @ironcev:
https://github.com/FuelLabs/sway/pull/5265#discussion_r1390465229
```
script;

trait MyTraitGeneric<T> {
    fn method(self);
}

impl MyTraitGeneric<u64> for u64 {
    fn method(self){
    }
}

trait MyTraitGeneric2<T> where T :  MyTraitGeneric<T> + MyTraitGeneric<u32> {
    fn method2(self);
}

impl MyTraitGeneric2<u64> for u64 {
    fn method2(self){
    }
}

fn main() {
    1_u64.method2();
}
```
The code above should throw an error that  `MyTraitGeneric<u32>` is not implemented for type `u64` but it doesn't.

esdrubal:
## Description

`check_if_trait_constraints_are_satisfied_for_type` now adds trait name and type_id to BTreeSet instead of adding only the trait name as a key to a BTreeMap. This fixes an issue where trait constraints were not considered satisfied when multiple type ids were associated with the same trait name.

`retrieve_interface_surface_and_items_and_implemented_items_for_type` now gets items using trait impl type arguments. This fixes an issue returning dummy methods because the proper implementation was not used.

`check_type_parameter_bounds` and `check_if_trait_constraints_are_satisfied_for_type` now handle generic traits better.

Where clauses such as `T : Trait<u32> + Trait<u64>` are now handled in case of error and in case of proper usage.

Adds more `DebugWithEngines`and `DisplayWithEngines` to facilitate future debugging.

Closes #5262
Closes #5271

## Checklist


sway-core/src/semantic_analysis/namespace/trait_map.rs
```diff
@@ -1065,7 +1073,7 @@ impl TraitMap {
             })
             .collect();
 
-        let relevant_impld_traits: BTreeMap<Ident, TypeId> = all_impld_traits
+        let relevant_impld_traits: BTreeSet<(Ident, TypeId)> = all_impld_traits
```
ironcev, line 1076: This is a question for the line 1052 actually, that was not changed. There we still have the `BTreeMap<Ident, TypeId>` when collecting constraints. Isn't it that with the constraints we can also have the situation of the same trait and different types e.g. `where T: Trait<u64> + Trait<u32>`? Shouldn't that also be `BTreeSet` like the other variables?

sway-core/src/type_system/id.rs
```diff
@@ -483,48 +492,115 @@ impl TypeId {
...
+                },
+            ) {
+                found_error = true;
+                f(structure_trait_constraint);
+            }
+        }
+        found_error
+    }
+
+    fn check_trait_constraints_errors_resolve(
```
ironcev, line 590: My strong preference here would be making the `check_trait_constraints_errors_resolve` a nested function inside of the `check_trait_constraints_errors_resolve` and naming it e.g. `resolve_type`. Why? It is used only within the `check_trait_constraints_errors_resolve` and putting it to the `TypeId` level bloats IMO the private interface of `TypeId`. Also, I think the method actually do not need `&self`.

ironcev (dismissed):
One small suggestion, otherwise :+1:  :+1:  Looking forward to get this integrated, our trait constraints are getting mightier and mightier :smile:

PR #5275, opened 2023-11-14
Bump to v0.48.0

JoshuaBatty:
waiting on 
* #5274 
* #5273
* #5281

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -143,30 +143,30 @@ checksum = "7079075b41f533b8c61d2a4d073c4676e1f8b249ff94a393b0595db304e0dd87"
 
 [[package]]
 name = "anstyle-parse"
-version = "0.2.2"
+version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "317b9a89c1868f5ea6ff1d9539a69f45dffc21ce321ac1fd1160dfa48c8e2140"
+checksum = "c75ac65da39e5fe5ab759307499ddad880d724eed2f6ce5b5e8a26f4f387928c"
 dependencies = [
  "utf8parse",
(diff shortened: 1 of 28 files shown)


kayagokalp (dismissed):

IGI-111:
@ironcev Can you look at that failing test?

JoshuaBatty:
I think it was locked to v0.47 of the std lib. I've just updated it to

```
[[package]]
name = 'trait_cannot_find_in_scope_issue'
source = 'member'
```

kayagokalp:
> I think it was locked to v0.47 of the std lib. I've just updated it to
> 
> ```
> [[package]]
> name = 'trait_cannot_find_in_scope_issue'
> source = 'member'
> ```

I feel like this might not work because std of 0.48 is not released yet but forc will try to get that one. We should always point to std by path in test suite to prevenet issues popping out while releasing.

```
[project]
name = "trait_cannot_find_in_scope_issue"
entry = "main.sw"
license = "Apache-2.0"


[dependencies]
```

Note: I am writing this without having access to the file structure at the moment so amount of `../`s might be invalid in the given example above

kayagokalp:
d6c8c155e73b611372a61864055e959e0bf7bd06 was missing the lock file for the test with std pointing to local. That would also cause test suite to fail as forc will throw warnings. I added the lock file back with std pointed to local version. (simply executed forc build after d6c8c155e73b611372a61864055e959e0bf7bd06) in 2875be862ed39bbdcb88a83efc74809be0575bee.

JoshuaBatty:
Thanks, @kayagokalp :)

IGI-111 (approved):

kayagokalp (approved):

PR #5284, opened 2023-11-16
Add `clear()` to `StorageKey` api

bitzoic:
## Description

The `StorageKey` api is missing the `clear` function. This has been added.

## Checklist


sway-lib-std/src/storage/storage_key.sw
```diff
@@ -137,3 +163,17 @@ fn test_storage_key_new() {
...
+    use ::constants::ZERO_B256;
+    use ::assert::assert;
+
+    let key = StorageKey::<u64>::new(ZERO_B256, 0, ZERO_B256);
+    key.write(42);
+    
+    assert(key.read() == 42);
+    let cleared = key.clear();
+    assert(cleared);
+    assert(key.try_read() == None);
```
Braqzen, line 178:
```suggestion
    assert(key.try_read().is_none());
```

sdankel (dismissed):

sway-lib-std/src/storage/storage_key.sw
```diff
@@ -98,6 +98,39 @@ impl<T> StorageKey<T> {
...
+    ///     assert(cleared);
+    /// }
+    /// ```
+    #[storage(write)]
+    pub fn clear(self) -> bool {
+        if __size_of::<T>() == 0 {
+            // If the generic doesn't have a size, this is an empty struct and nothing can be stored at the slot.
+            // This clears the length value for StorageVec, StorageString, and StorageBytes 
+            // or any other Storage type.
+            clear::<u64>(self.field_id, 0)
```
Braqzen, line 128: I'm unsure about the information flow here. We check a generic `T` and then coerce it into a `u64` because of the specific types that have been found to cause an issue.

Over time this might change and there will be a bug here because we assume a `u64`.
I guess if it passes CI now then it's fine and we'll catch it later if it ever becomes an issue.

Braqzen (approved):

sway-lib-std/src/storage/storage_key.sw
bitzoic: This is done due to a limitation in Rust where function overloading does not exist. Two implementations of `clear()` cannot be done and each case needs to be handled here.
We are unable to check what type the generic passed is. And even if we did call `clear::<T>`, it uses the size of `T` and thus nothing to clear in the first place.

SwayStar123 (approved):

PR #5290, opened 2023-11-20
fix fuel data section alignment

xunilrj:
## Description

This PR solves a problem that Data Section alignment was being calculated even for "internal" entries, like entries inside collections. This creates problems with things like 

```sway
const a: [u8; 4] = [1, 2, 3, 4];
```

See https://github.com/FuelLabs/fuels-ts/blob/309608dc9902a225f685327fc4107deee0349039/apps/docs-snippets/projects/echo-configurables/src/main.sw as an example.

Now the alignment is correctly calculated only for "root" entries , allowing "internal" entries to be as packed as they desire.

## Checklist

### DIFF
--- a/sway-lsp/tests/fixtures/tokens/abi/Forc.toml
+++ b/sway-lsp/tests/fixtures/tokens/abi/Forc.toml
@@ -3,6 +3,7 @@ authors = ["Fuel Labs <contact@fuel.sh>"]
 entry = "main.sw"
 license = "Apache-2.0"
 name = "abi"
+implicit-std = false
 
 [dependencies]
-std = { git = "https://github.com/FuelLabs/sway", tag = "v0.47.0" }
+std = { git = "https://github.com/FuelLabs/sway", commit = "194b5943dfc13e1b4fba7ea27678b4c3ab37ac84" }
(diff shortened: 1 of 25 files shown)


IGI-111 (dismissed):

IGI-111 (approved):

sway-lsp/tests/fixtures/tokens/abi/Forc.toml
```diff
@@ -3,6 +3,7 @@ authors = ["Fuel Labs <contact@fuel.sh>"]
 entry = "main.sw"
 license = "Apache-2.0"
 name = "abi"
+implicit-std = false
 
 [dependencies]
-std = { git = "https://github.com/FuelLabs/sway", tag = "v0.47.0" }
```
JoshuaBatty, line 8: Can we change these back to using the `tag = "v0.47.0"` for all of these?

JoshuaBatty (changes requested):

sway-lsp/tests/fixtures/tokens/abi/Forc.toml
IGI-111: We can not.
The reason the CI wasn't working is because using that version of the std (which still contains the fixtures sources) confuses forc. The commit used is the first one where those are removed.

Issue #5084, opened 2023-09-04
`where` clauses do not automatically import the trait in function definitions


bitzoic:
When importing a function that has a where clause in the definition and the trait is not explicitly imported, the compiler throws an error that the trait is not in scope. 
Instead, if a trait is included as part of a function definition it should be imported automatically.

This can be seen in the `std::hash::sha256` function. It has the following definition:
```rust
pub fn sha256<T>(s: T) -> b256 where T: Hash;
```

In this example, we import only the `sha256` function from above. 
``` rust
library;

use std::{
    constants::ZERO_B256, 
    hash::sha256
};

#[test]
fn test_hash_trait() {
    let my_val = ZERO_B256;

    let digest = sha256(my_val);
}
```

This will give the following error: 

<img width="679" alt="Screenshot 2023-09-04 at 11 53 55 AM" src="https://github.com/FuelLabs/sway/assets/54727135/9109249e-31c2-4382-934b-ac2916ea9561">

PR #5293, opened 2023-11-20
Provide useful error for "Cannot find trait Hash in this scope"

### LINKED ISSUE #5129: Provide temporary fix for the "Cannot find trait Hash in this scope" issue
_(Adapted from our discussion)_

When calling one of the hash functions (such as `sha256`) on a type, error is thrown with span inside the standard library saying that the `Hash` trait is out of scope.

image

What the error really means is that the impl for that particular type is not in scope at the call site. To fix the issue one needs to `use std::hash::Hash`.

The underlying issue in Sway is that we don't have a compile step to collect all the impls for a trait, so when you use one it has to be manually imported. And the only way to import impls is by importing the trait they relate to if and only if the impls are in the same module as the trait definition.

Considering how often the issue appears and how confusing the error message is, we want to provide a temporary solution before implementing collecting impls for a trait.

Potential temporary fixes: 
1. Putting `Hash` (and hence the impls) in the prelude. Cons: Bloating the default namespace is bad practice in general. Removing things from prelude later on is a breaking change. Programmers still need to `use std::hash::*`.
1. Provide helpful error message that will point to the solution. We have to see how much effort would it be to come up with a good heuristic and how helpful the message will really be.

ironcev:
## Description

This PR closes #5129 by providing a detailed guidelines to programmers on how to fix the "Cannot find trait XYZ in this scope". The issue is very prominent in the case of the `Hash` trait as demonstrated in the #5084.

We need to guide programmers at the moment because fixing the root cause will take some time.

The PR:
- replaces the current two misleading and redundant error messages with a clear instruction how to fix the issue at the call site.
- provides the guidelines in a general case, and not only for the prominent issue with the `Hash` trait.
- removes the redundant `SymbolNotFound` error in favor of more specific `TraitNotFound`

## Checklist

### DIFF
--- a/sway-core/src/decl_engine/engine.rs
+++ b/sway-core/src/decl_engine/engine.rs
@@ -261,6 +261,21 @@ impl DeclEngine {
         self.get(index)
     }
 
+    /// Returns all the [ty::TyTraitDecl]s whose name is the same as `trait_name`.
+    /// 
+    /// The method does a linear search over all the declared traits and is meant
+    /// to be used only for diagnostic purposes.
+    pub fn get_traits_by_name(&self, trait_name: &Ident) -> Vec<ty::TyTraitDecl>
+    {
+        self.trait_slab
+            .with_slice(|elems| elems
+                .into_iter()
+                .filter(|trait_decl| trait_decl.name == *trait_name)
+                .cloned()
+                .collect()
+            )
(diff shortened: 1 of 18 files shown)


sdankel (approved):
LGTM 🚀

PR #5306, opened 2023-11-28
AbiEncode trait auto implementation

xunilrj:
## Description

This PR implements "auto implementation" for traits. 

This expression is a little bit ambiguous because Rust has two "things" that can be called "auto impl".

1 - First, is what `rustc` does for `Sync` and `Send`. If the compiler can, it "auto implements" these traits. This is what this PR does;
2 - Second is the https://doc.rust-lang.org/unstable-book/language-features/auto-traits.html. This is *NOT* what this PR implements.

## Checklist


xunilrj:
Some tests are breaking because of https://github.com/FuelLabs/sway/issues/5383. The problem is that when we generate the auto impl, it ends up  resolving to the wrong types.

xunilrj:
The task to test std-lib does not work, because it depends on the SDK parsing the log correctly. Something we still need to do.

jjcnn:
I'm not sufficiently familiar with most of the code being touched here, so I don't feel comfortable giving a review a such.

I did want to raise the issue of whether it's possible to go from the log output back to the data? The type of the data doesn't seem to be part of the encoding, so to map the log back to the data it came from you would need to supply the types separately.

xunilrj:
> whether it's possible to go from the log output back to the data?

Only look at the bytes... no.
For that, and specifically for logging, you will have to look ate the ABI JSON.

```
"loggedTypes": [
    {
      "logId": 0,
      "loggedType": {
        "name": "",
        "type": 5,
        "typeArguments": []
      }
    },
    ...
```

The test that test this is the one broken because we need to update the Rust SDK. What we do in that test is we run a macro on top on this JSON that "recreate" all the types and give you easy access to typed logged data.

sway-core/src/semantic_analysis/program.rs
```diff
@@ -29,10 +29,18 @@ impl TyProgram {
         parsed: &ParseProgram,
         initial_namespace: namespace::Module,
         package_name: &str,
+        build_config: Option<&BuildConfig>,
     ) -> Result<Self, ErrorEmitted> {
         let mut namespace = Namespace::init_root(initial_namespace);
         let ctx =
             TypeCheckContext::from_root(&mut namespace, engines).with_kind(parsed.kind.clone());
+
+        let ctx = if let Some(build_config) = build_config {
```
IGI-111, line 38: this should use let-else

sway-core/src/type_system/id.rs
```diff
@@ -184,10 +185,15 @@ impl TypeId {
         &self,
         engines: &Engines,
         filter_fn: &F,
+        depth: usize,
     ) -> HashMap<TypeId, Vec<TraitConstraint>>
     where
         F: Fn(&TypeInfo) -> bool,
     {
+        if depth >= 128 {
```
IGI-111, line 193: I don't really like hardcoding this, maybe we should have a configurable type recursion limit?

IGI-111 (approved):
A few nitpicks but LGTM

forc/src/cli/commands/build.rs
```diff
@@ -30,6 +30,10 @@ pub struct Command {
     /// Also build all tests within the project.
     #[clap(long)]
     pub tests: bool,
+
+    #[clap(long)]
+    /// Experimental flags for the "new encoding" feature
```
JoshuaBatty, line 35:
```suggestion
    /// Experimental flag for the "new encoding" feature
```

forc/src/cli/commands/contract_id.rs
```diff
@@ -20,6 +20,10 @@ pub struct Command {
     pub build_profile: BuildProfile,
     #[clap(flatten)]
     pub salt: Salt,
+
+    #[clap(long)]
+    /// Experimental flags for the "new encoding" feature
```
JoshuaBatty, line 25:
```suggestion
    /// Experimental flag for the "new encoding" feature
```

forc/src/cli/commands/predicate_root.rs
```diff
@@ -18,6 +18,10 @@ pub struct Command {
     pub build_output: BuildOutput,
     #[clap(flatten)]
     pub build_profile: BuildProfile,
+
+    #[clap(long)]
+    /// Experimental flags for the "new encoding" feature
```
JoshuaBatty, line 23: there are a few of these, feel free to ignore if you want.

JoshuaBatty (approved):
Looks good, excited to see this land.

forc/src/cli/commands/predicate_root.rs
xunilrj: Fixed.

sway-core/src/semantic_analysis/program.rs
xunilrj: Fixed, but went with map_or_else.

sway-core/src/type_system/id.rs
xunilrj: Let us go with CONST for now.

PR #5313, opened 2023-11-29
Changes `ConcurrentSlab` to no longer clone values

esdrubal:
## Description

`ConcurrentSlab` now stores and returns an Arc, and on `get` it only clones an Arc which is much cheaper than cloning the whole values.

The type engine is still cloning every TypeInfo we should be able to also avoid that and get further improvements in speed.

With these changes core + std-lib compilation improved from around 3 secs to 1.5 secs.
`cargo run --bin test --release` is running in around  8min versus 16min in master. 

## Checklist

### DIFF
--- a/sway-core/src/concurrent_slab.rs
+++ b/sway-core/src/concurrent_slab.rs
@@ -53,8 +53,8 @@ where
     T: Clone,
 {
     pub fn len(&self) -> usize {
-        let lock = self.inner.read().unwrap();
-        lock.len()
+        let inner = self.inner.read().unwrap();
+        inner.len()
     }
 
     pub fn insert(&self, value: T) -> usize {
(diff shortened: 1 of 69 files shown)


sway-core/src/concurrent_slab.rs
```diff
@@ -57,63 +52,38 @@ impl<T> ConcurrentSlab<T>
 where
     T: Clone,
 {
+    pub fn len(&self) -> usize {
+        let lock = self.inner.read().unwrap();
```
ironcev, line 56: Proposal: let's use `inner` instead of `lock` here as the variable name. It is more descriptive and also in all other methods it is called `inner`.

ironcev (approved):

JoshuaBatty (approved):
Awesome work!

PR #5365, opened 2023-12-05
Add `InstructionInserter` and make a block's instruction list private

### LINKED ISSUE #4344: Add a better instruction inserter to IR.
There are a few places in the 'demotion' passes introduced in #4336 where we need an instruction inserter as opposed to the instruction appender we have now.

The `InstructionInserter` in `sway-ir/src/instruction.rs` isn't an actual inserter as it only appends an instruction to a block.  And that's just a single instruction at a time.  We currently create a new `InstructionInserter` for each instruction we add.

Ideally we should be able to have a utility which takes a block and either an index or an existing instruction indicating where to insert new instructions (or `None` to indicate appending), and it can then be re-used to insert or append many instructions at a time.

vaivaswatha:
## Description
Closes #4344 

Testing: Compared all tests' compiled hashes against `master`. They all match.


sway-ir/src/optimize/arg_demotion.rs
```diff
@@ -123,21 +123,19 @@ fn demote_fn_signature(context: &mut Context, function: &Function, arg_idcs: &[(
         .collect::<Vec<_>>();
 
     // For each of the old args, which have had their types changed, insert a `load` instruction.
-    let arg_val_pairs = old_arg_vals
```
ironcev, line 126: MUCH nicer with the new API all the places in this file. No need for `rev()` traversal and manual direct manipulation of the `instructions`.

sway-ir/src/optimize/simplify_cfg.rs
```diff
@@ -220,18 +220,14 @@ fn merge_blocks(context: &mut Context, function: &Function) -> Result<bool, IrEr
...
             }
 
-            // Re-get the block contents mutably.
-            let (from_contents, to_contents) = context.blocks.get2_mut(from_block.0, to_block.0);
-            let from_contents = from_contents.unwrap();
-            let to_contents = to_contents.unwrap();
-
             // Drop the terminator from `from_block`.
-            from_contents.instructions.pop();
+            from_block.remove_instruction_at(context, from_block.num_instructions(context) - 1);
```
ironcev, line 224: This pattern occurs several times. Perhaps having a dedicated method `remove_last_instruction`?

ironcev (approved):
Just one small API suggestion, otherwise great the we got safer, simpler, and more expressive API!

ironcev (approved):

xunilrj (approved):

PR #5378, opened 2023-12-11
Refactor Language Server I/O to Use Tokio Async Primitives

JoshuaBatty:
## Description
Aligned the language server's I/O operations with the `tower-lsp` async framework by adopting Tokio's asynchronous I/O primitives instead of ones from `std::fs`.


kayagokalp (approved):
utACK

Issue #528, opened 2021-12-29
Forc can't stop language server


SilentCicero:
Forc needs a command to stop running the language server process nicely, It's very hard to manage currently or at the very least know what the process ID is to stop it manually.

I suggest:
`forc lsp start` and `forc lsp stop` -- to stop the process. And in addition, when it starts to give the process ID so you can stop it manually if needed.

PR #5395, opened 2023-12-15
Make sure that list of plugins is sorted and unique

### LINKED ISSUE #5396: `forc plugins` may list plugins multiples
If the user has multiple versions of force installed locally `forc plugins` show duplicate entries 

```
➜  forc plugins
Installed Plugins:
forc-crypto
forc-doc
forc-submit
forc-deploy
forc-run
forc-explore
forc-wallet
forc-index
forc-lsp
forc-fmt
forc-tx
forc-crypto
forc-doc
forc-submit
forc-deploy
forc-run
forc-explore
forc-wallet
forc-index
forc-lsp
forc-fmt
forc-tx
forc-crypto
forc-fmt
```

crodas:
## Description

Fixes #5396, #5402, [#528](https://github.com/FuelLabs/fuelup/issues/528)

The sorting is performed by the file_name instead of the full text to be printed or full path

## Checklist

### DIFF
--- a/forc/src/cli/commands/plugins.rs
+++ b/forc/src/cli/commands/plugins.rs
@@ -63,7 +63,7 @@ pub(crate) fn exec(command: PluginsCommand) -> ForcResult<()> {
     for plugin in plugins {
         info!("{}", plugin.2);
         if plugin.1.len() > 1 {
-            println_warning(&format!("Multiple path found for {}", plugin.0));
+            println_warning(&format!("Multiple paths found for {}", plugin.0));
             for path in plugin.1 {
                 println_warning(&format!("   {}", path.display()));
             }


kayagokalp (commented):
I haven't tested it locally but looks good! One small suggestion about the error handling so that the output is more understandable on our side, if something happens.

forc/src/cli/commands/plugins.rs
```diff
@@ -18,15 +18,32 @@ pub struct Command {
     describe: bool,
 }
 
+fn get_file_name(path: &Path) -> String {
```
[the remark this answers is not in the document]
kayagokalp: Thanks!

forc/src/cli/commands/plugins.rs
```diff
@@ -18,15 +18,32 @@ pub struct Command {
...
+    if let Some(Some(path_str)) = path.file_name().map(|path_str| path_str.to_str()) {
+        path_str.to_owned()
+    } else {
+        panic!("Failed to read file name from {path:?}")
```
sdankel, line 25: What does the user see when it panics? Curious why we're using `panic!` here instead of `bail!`

forc/src/cli/commands/plugins.rs
```diff
@@ -18,15 +18,32 @@ pub struct Command {
...
     } = command;
 
+    let mut plugins = crate::cli::plugin::find_all()
+        .map(|path| {
+            print_plugin(path.clone(), print_full_path, describe)
+                .map(|info| (get_file_name(&path), info))
+        })
+        .collect::<Result<Vec<_>, _>>()?;
+    plugins.sort_by(|a, b| a.0.cmp(&b.0));
+    plugins.dedup_by(|a, b| a.0 == b.0);
```
sdankel, line 42: How are we deduping here? It should be by what's first in the PATH, or the output of `which {plugin_name}`, right?

sdankel: Perhaps we should put this logic inside of `find_all()`

sdankel: I don't think we have to worry about the version. Fuelup handles that for you by linking your active toolchain binaries to an executable in `~/.fuelup/bin/`. So whatever is linked there is the right version.

forc/src/cli/plugin.rs
```diff
@@ -14,14 +14,35 @@ use std::{
...
     let args = &args[1..];
     let path = find_external_subcommand(cmd);
     let command = match path {
         Some(command) => command,
         None => bail!("no such subcommand: `{}`", cmd),
     };
+
+    if let Ok(forc_path) = std::env::current_exe() {
+        if is_verbose && command.parent() != forc_path.parent() {
+            let bold_yellow_code = "\x1b[1;33m";
```
sdankel, line 31: You can use `println_warning` for consistent warning formatting with the rest of forc

forc/src/cli/commands/plugins.rs
```diff
@@ -18,15 +18,32 @@ pub struct Command {
...
     } = command;
 
+    let mut plugins = crate::cli::plugin::find_all()
+        .map(|path| {
+            print_plugin(path.clone(), print_full_path, describe)
+                .map(|info| (get_file_name(&path), info))
+        })
+        .collect::<Result<Vec<(String, String)>, _>>()?;
+    plugins.sort_by(|a, b| a.0.cmp(&b.0));
+    plugins.dedup_by(|a, b| a.0 == b.0);
```
sdankel, line 42: Discussed offline: we'll show all plugins with the full path. When there's a duplicate, we'll indicate which one will be used based on the $PATH

forc/src/cli/plugin.rs
```diff
@@ -14,14 +14,35 @@ use std::{
...
     let cmd = args.get(0).expect("`args` must not be empty");
     let args = &args[1..];
     let path = find_external_subcommand(cmd);
     let command = match path {
         Some(command) => command,
         None => bail!("no such subcommand: `{}`", cmd),
     };
+
+    if let Ok(forc_path) = std::env::current_exe() {
+        if is_verbose && command.parent() != forc_path.parent() {
```
sdankel, line 30: I think we should always show this warning, not only in verbose mode

JoshuaBatty: This was Cesar's original plan but i asked for it to be behind a verbose flag. Mainly so it doesn't get annoying for people that intentionally have builds in multiple places like we do. 

I guess considering we are the minority, perhaps we could take another option by using a silent flag instead to not show warnings, but have them show by default?

forc/src/cli/commands/plugins.rs
```diff
@@ -18,15 +22,62 @@ pub struct Command {
     describe: bool,
 }
 
+fn get_file_name(path: &Path) -> String {
+    if let Some(Some(path_str)) = path.file_name().map(|path_str| path_str.to_str()) {
```
kayagokalp, line 26:
```suggestion
    if let Some(path_str) = path.file_name().and_then(|path_str| path_str.to_str()) {
```

kayagokalp: Totally personal choice feel free to discard :)

forc/src/cli/plugin.rs
```diff
@@ -14,14 +15,29 @@ use std::{
...
+) -> Result<process::Output> {
     let cmd = args.get(0).expect("`args` must not be empty");
     let args = &args[1..];
     let path = find_external_subcommand(cmd);
     let command = match path {
         Some(command) => command,
         None => bail!("no such subcommand: `{}`", cmd),
     };
+
+    if let Ok(forc_path) = std::env::current_exe() {
```
kayagokalp, line 30: good idea! If understood correctly we are expecting the forc and the plugin to be in the same directory for them to not emit any warnings. At first I thought this might create a problem with plugins that are hosted in different repos. But if users install them with a single installation source they will end up in the same directory. So this will work nicely and hopefully help users find out that they have shadowed plugins etc happening.

forc/src/cli/commands/plugins.rs
```diff
@@ -18,15 +22,62 @@ pub struct Command {
...
+}
+
 pub(crate) fn exec(command: PluginsCommand) -> ForcResult<()> {
     let PluginsCommand {
         print_full_path,
         describe,
     } = command;
 
+    let mut plugins = crate::cli::plugin::find_all()
+        .map(|path| print_plugin(path.clone(), print_full_path, describe).map(|info| (path, info)))
```
kayagokalp, line 40: I know that `print_plugin` is not introduced here, but seeing that it does not print but format the string, it might make sense to rename it?

forc/src/cli/commands/plugins.rs
```diff
@@ -18,15 +22,62 @@ pub struct Command {
...
+            (
+                bin_name,
+                paths
+                    .into_iter()
+                    .filter_map(|path| {
+                        if temp_hashmap.get(&path).is_some() {
+                            return None;
+                        }
+                        temp_hashmap.insert(path.clone(), true);
+                        Some(path)
```
kayagokalp, line 63: I am a little lost about what is happening here,

After this step do we have bin name tupled with path for the first occurance and others are tupled with None? Is that even neccesary? We are not removing them from the vec so that we cna check the len in line 75. So why are we making the duplicate entries `None` explicitly.

kayagokalp (commented):
Left a single nit and a question, looking good!

forc/src/cli/commands/plugins.rs
```diff
@@ -49,22 +51,10 @@ pub(crate) fn exec(command: PluginsCommand) -> ForcResult<()> {
...
-                        temp_hashmap.insert(path.clone(), true);
-                        Some(path)
-                    })
-                    .collect::<Vec<_>>(),
-                content,
-            )
+        .map(|(bin_name, mut paths, content)| {
+            paths.sort();
+            paths.dedup();
+            (bin_name, paths, content)
```
kayagokalp, line 57: nice!! a lot more readable

kayagokalp (dismissed):
🚀

forc/src/cli/commands/plugins.rs
```diff
@@ -18,15 +22,52 @@ pub struct Command {
...
+        .collect::<Vec<_>>();
+    plugins.sort_by(|a, b| a.0.cmp(&b.0));
+
     info!("Installed Plugins:");
-    for path in crate::cli::plugin::find_all() {
-        info!("{}", print_plugin(path, print_full_path, describe)?);
+    for plugin in plugins {
+        info!("{}", plugin.2);
+        if plugin.1.len() > 1 {
+            println_warning(&format!("Multiple path found for {}", plugin.0));
```
sdankel, line 66:
```suggestion
            println_warning(&format!("Multiple paths found for {}", plugin.0));
```

JoshuaBatty:
When running locally I now get this.

```
~ % forc plugins --paths
Installed Plugins:
/Users/joshuabatty/.cargo/bin/forc-doc
/Users/joshuabatty/.cargo/bin/forc-lsp
```

I would expect it should still show all of the forc tools? i.e `forc-tx` forc-fmt` etc...

sdankel (approved):

JoshuaBatty (approved):

PR #5400, opened 2023-12-16
Changes `ConcurrentSlab` to store values in map.

esdrubal:
## Description

With these changes `ConcurrentSlab` now stores values in a map instead of a vector.

This change is necessary to allow us to remove specific ids from the `ConcurrentSlab`.

Before this, removing an index would change the vector and invalidate existing type and decls ids.

This is required for `TypeEngine` and `DeclEngine` `clear_module` methods to work properly.

This change implies a slowdown in the compile time of `cargo run --bin test --release` from ~9min to ~10min. I think the overhead is worth it because we now can remove values from ConcurrentSlab and optimize memory usage, which is mostly concentrated in the slabs.

## Checklist


sway-error/src/error.rs
```diff
@@ -1138,7 +1138,11 @@ impl ToDiagnostic for CompileError {
                     format!("Consider removing the variable \"{variable}\" altogether, or adding it to all alternatives."),
                 ],
             },
-            TraitNotImportedAtFunctionApplication { trait_name, function_name, function_call_site_span, trait_constraint_span, trait_candidates }=> Diagnostic {
+            TraitNotImportedAtFunctionApplication { trait_name, function_name, function_call_site_span, trait_constraint_span, trait_candidates }=> {
+                // Make candidates order deterministic
+                let mut trait_candidates = trait_candidates.clone();
+                trait_candidates.sort();
```
ironcev, line 1144: I am thinking of adding `let trait_candidates = &trait_candidates; // Remove immutability.` here, just to remove the mutability for the rest of the diagnostics and have clear semantics. Safety first, but yes I am sometimes a bit too paranoid :-)

sway-core/src/type_system/engine.rs
```diff
@@ -71,8 +68,15 @@ impl TypeEngine {
 
     /// Removes all data associated with `module_id` from the type engine.
     pub fn clear_module(&mut self, module_id: &ModuleId) {
+        self.slab.retain(|key, _type_info| {
```
ironcev, line 71: A question here, we have two `ConcurrentSlab`s, `slab` and `slab_source_id`, that need to be in sync. Can we have a concurrency issue here on the level of the whole `clear_module` method? Race condition in which the `retain` on `slab` gets executed but on `slab_source_ids` still not and we get a second call on `clear_module`? Should the whole method be atomic? I am not sure how `clear_module` operates with LSD, that's why this ask for clarification.

ironcev (commented):
LGTM just a short question on a potential race condition. I am not sure if we can have it in that case, just checking :-)

sway-core/src/type_system/engine.rs
JoshuaBatty: This shouldn't be a problem anymore since I merged #5428 into this PR. The slab now holds a TypeSourceInfo so we no longer have 2 slabs to clear, just one.

sway-error/src/error.rs
JoshuaBatty: Good idea, added in #[497468b](https://github.com/FuelLabs/sway/pull/5400/commits/497468b545c9ba0d890915bddf0e4af15f575932)

JoshuaBatty (approved):

vaivaswatha (approved):

Issue #5063, opened 2023-08-31
Implement References RFC


IGI-111:
Tracking issue for https://github.com/FuelLabs/sway-rfcs/pull/31

* References
  * [ ] Remove usages of `ref mut`
  * [ ] `const` return values
* Pointers
  * [ ] `*mut` typed pointer
  * [ ] Remove `raw_ptr`
  * [ ] `__ptr_cast` intrinsic
* Slices
  * [ ] New typed slice syntax
  * [ ] `__slice` intrinsic
  * [ ] `__slice_elem` intrinsic
  * [ ] `str` slices
  * [ ] Remove `raw_slice`
* [ ] Update documentation to explain  https://github.com/FuelLabs/sway/issues/5061

DicksonWu654:
I confirmed a concrete instance of the unchecked “Forbid references — In the storage” item on current `master` / v0.72.1 (`dad95cc42b0383b4e3bacdeda9766565aa584a92`).

References nested inside high-level persistent container values are accepted without an experimental flag or warning:

```sway
contract;

use std::storage::storage_vec::*;

struct RefBox { p: &u64 }

storage {
    map: StorageMap<u64, RefBox> = StorageMap::<u64, RefBox> {},
    vec: StorageVec<RefBox> = StorageVec::<RefBox> {},
}

#[storage(write)]
fn store_map(value: u64) {
    storage.map.insert(0, RefBox { p: &value });
}

#[storage(read)]
fn read_map() -> u64 {
    *storage.map.get(0).read().p
}
```

In a local Fuel node, after one committed transaction stored `111`, separate committed transactions returned `0` from `read_map()` in both debug and release. `StorageVec<RefBox>` behaved the same way with a stored value of `222`. The entries existed; the shallow-copied address no longer denoted the original referent. Same-transaction behavior was layout/profile-sensitive.

This matches the `write_quads<T>` documentation: pointers/references are shallow-copied without pointees. The gap is that high-level persistent container declarations and writes admit these types without an unsafe boundary or diagnostic.

The existing recursive contract-storage validation in `sway-core/src/language/ty/program.rs` rejects `TypeInfo::RawUntypedPtr`. Extending the same match to `TypeInfo::Ref { .. }` rejects direct and nested references, including those inside `StorageMap` and `StorageVec` type arguments. I have a small local patch and regression fixture for that approach if useful.

PR #5406, opened 2023-12-18
Add references to Sway

ironcev:
## Description

This PR brings [references](https://github.com/FuelLabs/sway-rfcs/blob/ironcev/amend-references/files/0010-references.sw) to the language. The overall effort is tracked in #5063.

In the below description, when talking about references we mean `&T` references - references to immutable values. `&mut T` will be implemented in upcoming PRs. 

The PR implements the following features:
- References exist in the type system and can be declared, aliased, dereferenced using `*` operator, etc.
- Reference expressions (referencing) are fully supported on all expressions including the nonsensical one like e.g. `&return;`
- Semantic analysis and checks are implemented for the current functionality.
- References can be embedded in aggregates.
- References can reference parts of aggreagates.
- References can be passed to and returned from ASM blocks.
- References can be passed to and returned from functions.
- References can be mutable: `let mut r_x = &x`
- Impls can be implemented for non-generic reference types.
- Traits can be implemented for non-generic reference types.
- References work with generics, means `&T` works as expected.
- References can reference references.
- References can be dereferenced using the `*` operator.
- Dereference expressions (dereferencing) are fully supported on all expressions including the nonsensical one like e.g. `*&return;`

Known limitations:
- When declaring references on references, currently a space (' ') is mandatory between two ampersands. E.g., `&&a` emits error and must be written as `& &a`. Extending parser to support, e.g., arbitrary `&&...&&a` will be done in upcoming PRs.
- Referencing function parameters for copy types works but not with 100% proper semantics (creates a copy of the parameter).

On the IR level, references are pointers stored as `u64` and do not exist as a separate concept. They play well with all the existing IR optimizations.

Since `core::ops::Eq` and `__addr_of()` are still not implemented for references, checking for equality in tests is done by converting references to `raw_ptr`s and checking raw pointers. This is essentially fine and will one day be turned into converting references to typed pointers which are tests we will anyhow want to have.

Next steps are listed in #5063.

The documentation will be done in a separate PR and provided separately once feature is considered ready for public usage.

There are several todos in code marked as `TODO-IG`. They will be resolved in upcoming PRs.

# Demo

For extensive demos of implemented features, see the tests in this PR.

An example of a semantic check:

Expression cannot be dereferenced

## Checklist


Voxelot:
> References can be passed to and returned from functions.

What about ABI methods - are there any special considerations we've taken into account for those? (i.e. no mutable references)

ironcev:
> What about ABI methods - are there any special considerations we've taken into account for those? (i.e. no mutable references)

Good question! I am also asking it in one of the open todos: `// TODO-IG: No references in ABIs according to the RFC. Or we want to have them?`

The current proposal goes for [forbidding references altogether on the boundaries](https://github.com/FuelLabs/sway-rfcs/blob/290f9ed65605b3fc4c9189581e66017f529dbb41/files/0010-references.sw#L528). But if we support `raw_ptr` on the boundaries, and we do, I see no reason not to support references.

Having some good examples for use cases where `raw_ptr` is used in ABIs would be great. I am sure such use cases would get easier to support with references.

Long story short, current proposal bans references on the boundaries, but we will re-discuss it. Any concrete examples in dapps that would be helpful to understand the use cases are appreciated.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -2014,6 +2015,57 @@ impl ty::TyExpression {
...
+                    span,
+                }),
+            ),
+            _ => {
+                let engines = ctx.engines();
+                let type_engine = ctx.engines().te();
+
+                let ctx = ctx
+                    .by_ref()
+                    .with_type_annotation(type_engine.insert(engines, TypeInfo::Unknown, None))
```
xunilrj, line 2037: The type is really unknown here? I  think `TypeContext` may have some tips... this generates problems with `u8` an u64`, for example:

```
let a: &u8 = &1
```

Or something like this, because `1` will be treated as `numeric` until it decays to `u64` sometime later, creating an invalid expression.

sway-core/src/semantic_analysis/node_dependencies.rs
```diff
@@ -907,6 +908,7 @@ fn type_info_name(type_info: &TypeInfo) -> String {
         TypeInfo::Slice(..) => "__slice",
         TypeInfo::Alias { .. } => "alias",
         TypeInfo::TraitType { .. } => "trait type",
+        TypeInfo::Ref(..) => "ref",
```
xunilrj, line 911: Given that this is used as key in trees, shouldn't it have the `referred type`? This method return `String`, so I think you are safe to append it here.

sway-core/src/type_system/info.rs
```diff
@@ -389,6 +393,10 @@ impl PartialEqWithEngines for TypeInfo {
                         .get(*l_trait_type_id)
                         .eq(&type_engine.get(*r_trait_type_id), engines)
             }
+            (Self::Ref(l_ty), Self::Ref(r_ty)) => type_engine
```
xunilrj, line 396: Silly optimization but if type ids match you are done, right?

```
(l_ty.type_id == r_ty.type_id) || (
type_engine.get(l_ty.type_id).eq(&type_engine.get(r_ty.type_id), engines))
```

sway-core/src/type_system/info.rs
```diff
@@ -500,6 +508,9 @@ impl OrdWithEngines for TypeInfo {
             ) => l_trait_type_id
                 .cmp(r_trait_type_id)
                 .then_with(|| l_name.cmp(r_name)),
+            (Self::Ref(l_ty), Self::Ref(r_ty)) => type_engine
```
xunilrj, line 511: Same thing here... others branches are ordering the type id and not the `TypeInfo` struct.

xunilrj (approved):

xunilrj:
I left some comments but I am in favor of merging this and avoid merge/rebase issues.

xunilrj:
I am also curious to know what would happen to references to locals being returned

```
fn f(v: u64) -> &u64 {
    return &v;
}
```

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
ironcev: Good question, assuming that the initial issues, were exactly in the area of type inference and unification. Here is a detailed answer.

Well, `TypeInfo::Unknown` is actually our way of saying `without_type_annotation` ;-). the `type_annotation` TypeId is not an `Option` and setting it to `Unknown` is a way of saying we are fine with any type, no type expectations.This is how type annotation removal is done in all places, and now when I think about it, I would love to have an explicit `without_type_annotation` method to give a clear intent.

And yes in this case we want to remove the expectation that we are getting because if we have one, like in your example where the surrounding context will expect `&8` the expectation will always be for a reference type `&Something`. And in the type check we are actually checking the referenced expression `Something` which will never be `&Something`.

The returned type like `Numeric` in your example is packed into `Ref` and returned, and the later unification in the context (`let` in the above case) does proper unification.

In other words, no worry, the given example works :-)

BDW, since unification issues were so prominent in the beginning I've decided to have `unifying_reference_types` tests but didn't write them because to a good extent unification is implicitly covered in other tests. But the main reason was honestly that I got tired of writing tests. Nevertheless, the `unifying_reference_types` will come in upcoming PRs. Until then I trust my smoke tests, that also have examples like yours :-)

P.S. When implementing `impl<T> &T` or `impl<T> MyTrait for &T` we still get unification issues in some cases. But these features are at the moment considered as not implemented.

sway-core/src/semantic_analysis/node_dependencies.rs
ironcev: The returned string needs to be different from the one used to display `TypeInfo` which is "&". So "ref" should be fine here. Still, to be on the safe side, I've changed it to "reference type" :-)

sway-core/src/type_system/info.rs
ironcev: Good catch! Actually, considering how often this is called, I don't see this optimization as silly at all! Although simple one, it was not used for any variant, so I've added it everywhere where it's applicable. Curious to see those perf graphs of ours dropping a pixel or two more :-) Jokes aside, good catch!

sway-core/src/type_system/info.rs
ironcev: Hmmm, here I am not that sure. Some branches like `TraitType` do compare `TypeId`s but some other like `Alias` or `Array` compare `TypeInfo`s ;-) IMO it makes sense to compare `TypeId`s everywhere but I want to give it a closer look before changing it.

ironcev:
> I am also curious to know what would happen to references to locals being returned

The locals will be placed on heap and the returned reference will contain a heap pointer. This is meant by "Allocating values on the heap when using references" in #5063. Knowing we never free heap allocations this is fine and is a first step. BDW we discussed garbage collection at the offsite and there should not be anything in the reference concept what will stop us one day from implementing garbage collection if we want.

xunilrj (approved):

xunilrj (approved):

xunilrj (approved):

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
ironcev: @xunilrj It turned out that you were right. The above example indeed work, but there were issues in the type inference coming from the fact that `TypeInfo::Unknown` was used. #5643 is addressing those issues.

Issue #5282, opened 2023-11-15
Fix errors when using `return` expression in semantically meaningless places


ironcev:
When using `return` in semantically meaningless places, at the moment we either get the Internal compiler error or some equally meaningless error message. Some examples are given below.

```Sway
script;

fn main() -> u64 {
    let a = [return 0u64];
    //      ^^^^^^^^^^^^^
    //      Internal compiler error: UnknownType { span: Span { src (ptr): 0x5555cdba9440, source_id: Some(SourceId { id: 67 }), start: 40, end: 53, as_str(): "[return 0u64]" } }

    let a = (return 0u64, return 1u64);
    // Internal compiler error: UnknownType { span: Span { src (ptr): 0x55cae80b7b30, source_id: None, start: 0, end: 23, as_str(): "unknown source location" } }

    if return 0u64 {
        0
    };
    //Mismatched types.
    //expected: ()
    //found:    numeric.
    //help: The two branches of an if expression must return the same type.
    
    let a = return 0u64;
    //      ^^^^^^^^^^^
    //      Internal compiler error: UnknownType { span: Span { src (ptr): 0x56496e7d32b0, source_id: Some(SourceId { id: 67 }), start: 677, end: 688, as_str(): "return 0u64" } }
    
    let _ = match return 0u64 {
        _ => 0,
    };
    // Unimplemented feature: matching on this type is unsupported right now
    
    0
}
```

Fix the internal compiler error and replace other errors by clearly stating that `return` cannot be used on such places in code.

While doing it, check the current implementation of the [`gather_return_statements`](https://github.com/FuelLabs/sway/blob/d07a84bbd7767dc385b514910b4343e3454dff3c/sway-core/src/language/ty/expression/expression_variant.rs#L1418) method.

jjcnn:
Fixed by #5416

Issue #5414, opened 2023-12-20
Fix errors when using `break` and `continue` expressions in arbitrary expressions


ironcev:
`break` and `continue` are expressions and allow early return/continuation anywhere from an expression, even in cases that are non-nonsensical like e.g. `let _ = [break];`

At the moment compiler generates internal compiler errors (and perhaps other errors) for some of such usages. E.g.
```Sway
let _ = break; // Ok.
let _ = [break, break]; // Ok.

let _ = continue; // Ok.
let _ = [continue, continue];
//      ^^^^^^^^^^^^^^^^^^^^ Internal compiler error: UnknownType { span: Span { src (ptr): 0x561903e4ac60, source_id: Some(SourceId { id: 3145808 }), start: 287, end: 307, as_str(): "[continue, continue]" } }
    // Please file an issue on the repository and include the code that triggered this error.
```
We should write tests to check for usages of `break` and `continue` in arbitrary expressions to ensure the expected behavior.

PR #5416, opened 2023-12-21
Check errors related to return as an expression

jjcnn:
## Description

Investigating/fixing #5282.

I've added a unit test containing some code where `return` is used in "odd" positions. My interpretation of the test output:

1. For value returns we are generally reporting an "unreachable code" warning, which is the intended behavior.
2. `match return 42 { ... }` reports an "unimplemented feature" error. I haven't looked into which type we're unable to match on, but presumably it's `()`, which seems like an odd oversight.
3. It looks like we have a general problem parsing non-value returns. 

For item 2 I think we can improve the error message to stating the type we're unable to match on.

For item 3 I think we need to investigate if we can parse these cases correctly, or at least give better error messages.

## Checklist


jjcnn:
Re. item 3: Rust struggles with some non-value returns as well:

```
fn main () {
    let _ = match return {
        () => 42,
    };
}
```
gives the following error:
```
error: expected one of `.`, `;`, `?`, `}`, or an operator, found `=>`
 --> return.rs:3:12
  |
3 |         () => 42,
  |            ^^ expected one of `.`, `;`, `?`, `}`, or an operator

error: expected one of `.`, `?`, `{`, or an operator, found `;`
 --> return.rs:4:6
  |
2 |     let _ = match return {
  |             -----
  |             |
  |             while parsing this `match` expression
  |             help: try removing this `match`
3 |         () => 42,
4 |     };
  |      ^ expected one of `.`, `?`, `{`, or an operator
```

The confusing error occurs because Rust tries to group `return` with the curly braces that surrounds the match arms.

On the other hand, Rust is capable of parsing this one fine:

```
fn main () {
    let _ = return + return;
}
```

Instead, the Rust typechecker complains about there being no implementation of `+` for `()`.

test/src/e2e_vm_tests/test_programs/should_fail/return_in_strange_positions/src/main.sw
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// Legal return type. Matching on the type is unimplemented.
+
+fn in_match_scrutinee() -> u64 {
+    match return 42 {
+        _ => 845,
+    }
+}
+
```
ironcev, line 75: This is great coverage already, but for the completeness, let's add the missing expressions like e.g.:

- enum instantiation: `Enum::A(return 42)`.
- function application: `fun(return 42)`.

Ideal would be to double check with `TyExpressionVariant`.

test/src/e2e_vm_tests/test_programs/should_fail/return_in_strange_positions/src/main.sw
```diff
@@ -0,0 +1,159 @@
...
+// Incorrect return types. These should fail because of the 'return'.
```
ironcev, line 76: Well, actually the return type is a correct one here :smile: The `()`, which is what naked `return`s in examples return. So theoretically, in all the below examples we shouldn't have any errors ;-) Rust compiles most of those examples correctly, Sway as well, with the dead code warning and without errors.

I am suggesting here just to change the comment, not to fix the test or the current error message that would be a bit way too much :-)

ironcev, line 76:
```suggestion
// Returning `()`.
```

test/src/e2e_vm_tests/test_programs/should_fail/return_in_strange_positions/src/main.sw
```diff
@@ -0,0 +1,159 @@
...
+
+fn in_init_non_value_return() {
+    let _ = return;
```
ironcev, line 79: Why not having x45 here as well. If the code compiles we should also here have dead code warnings.

test/src/e2e_vm_tests/test_programs/should_fail/return_in_strange_positions/src/main.sw
```diff
@@ -0,0 +1,159 @@
...
+}
+
+fn in_match_scrutinee_non_value_return() {
+    match return {
+        _ => 845,
+    }
+}
+ 
+
+// Copy paste for all expressions, there are not that many :-)
```
ironcev, line 124: The way test methods are structured look clear to me, I guess the comments on the approach are not needed and can be deleted.

test/src/e2e_vm_tests/test_programs/should_fail/return_in_strange_positions/src/main.sw
```diff
@@ -0,0 +1,159 @@
...
+// return 43 + return 42;
+
+// Then copy paste the same for functions returning unit.
+//fn in_array_unit() {
+//   let _ = [return, return];
+//   
+//   assert(false);
+//}
+
+// ...
```
ironcev, line 135: I would still do another one quick copy paste and test the case when the return type is not the expected one. Essentially copying all the functions from the first section and changing the function return type to e.g. `bool`.

The expectation is in that case to have the same errors everywhere, type mismatch.

test/src/e2e_vm_tests/test_programs/should_fail/return_in_strange_positions/src/main.sw
```diff
@@ -0,0 +1,159 @@
...
+
+fn main() {
+   assert(42 == in_init());
```
ironcev, line 138: A slight issue with this is that this code actually never runs because of the compile errors. Out of the current tests we know that the dead code warnings are generated, but strictly speaking not if the proper value is returned. E.g. in the `[return 42, return 43]` example it could be that 43 is returned, or any other value. Since the idea was to indirectly test if the CFG is correct I see the value in those asserts being actually executed at runtime.

If we take a look at the first section only `in_match_scrutinee` does not compile if I a didn't overlook something. To me it makes sense to split this test into two, one that actually fully compiles and executes and one that generates errors.

ironcev (changes requested):
This is great thanks @jjcnn! We finally know where we are when it comes to `return` expression :smile: For the overall question how to proceed I'll comment my view in the conversation section.

ironcev:
> For item 2 I think we can improve the error message to stating the type we're unable to match on.
> For item 3 I think we need to investigate if we can parse these cases correctly, or at least give better error messages.

Generally I agree with both proposals but would be careful in terms of effort. This is just my opinion. The point 2) should be easy to extend. For the point 3) I am not sure if it pays off at this point and would like to explain why I think so.

Essentially, the output of the overall test is much better then I expected. We get dead code where we should and zero ICEs which is a great starting point. Struggle with naked `return` is also not there in all cases and, as well pointed above, even Rust has issues with those. That's why for me, even if we keep it as is for now it's fine.

That's reason talking out of me. But emotionally, of course I would love to see if those generic "Expected an expression." could be turned into something more guiding :-)

Also I am thinking, since with `break` and `continue` we do have ICEs (#5414) I think investing time there would be more beneficial.

jjcnn:
@ironcev : I've added the two cases you mentioned, and I think those are the only two that were missing.

I've also separated the tests out into one passing and one failing test, and I'm now getting ICEs in the tests that are supposed to pass. At least one of them (`let _ = [ return 42, return 43 ];) was also in the original test, where it did not trigger an ICE. Presumably the failing tests cause the compiler pipeline to terminate before the ICEs are hit.

I'll investigate the ICEs after Christmas.

test/src/e2e_vm_tests/test_programs/should_pass/return_in_strange_positions/src/main.sw
```diff
@@ -0,0 +1,97 @@
+script;
+
+// This file tests the reported errors and warnings in various instances when a 'return'
+// occurs in a non-statement position. This is allowed, but will often result in
+// unreachable code or similar warning situations.
+
+pub struct S { x : u64, y : u64, }
+pub enum Enum {
+    A: u64,
```
ironcev, line 9:
```suggestion
    A: (u64, u64),
```

ironcev: Just to stress the CFG more like in other cases where we've used `return 42` and `return 43` wherever we could. Sorry for realizing this only know. When I passed the suggestion for the first time this suggestion didn't cross my mind. And also considering the latest appearance of ICEs I think stressing the compiler and the CFG more makes sense.

test/src/e2e_vm_tests/test_programs/should_pass/return_in_strange_positions/src/main.sw
```diff
@@ -0,0 +1,97 @@
...
+fn in_while_condition() -> u64 {
+    while return 42 {
+        break;
+    };
+    
+    745
+}
+
+fn in_enum() -> u64 {
+    let _ = Enum::A(return 42);
```
ironcev, line 70:
```suggestion
    let _ = Enum::A(return 42, return 43);
```

test/src/e2e_vm_tests/test_programs/should_pass/return_in_strange_positions/src/main.sw
```diff
@@ -0,0 +1,97 @@
...
+    
+    845
+}
+
+
+fn helper_fun(x : u64) -> u64 {
```
ironcev, line 76:
```suggestion
fn helper_fun(x : u64, y: u64) -> u64 {
```

test/src/e2e_vm_tests/test_programs/should_pass/return_in_strange_positions/src/main.sw
```diff
@@ -0,0 +1,97 @@
...
+    x + 1
```
ironcev, line 77:
```suggestion
    x + y
```

test/src/e2e_vm_tests/test_programs/should_pass/return_in_strange_positions/src/main.sw
```diff
@@ -0,0 +1,97 @@
...
+}
+
+fn in_fun_arg() -> u64 {
+    let _ = helper_fun(return 42);
```
ironcev, line 81:
```suggestion
    let _ = helper_fun(return 42, return 43);
```

ironcev (changes requested):

ironcev:
> I've also separated the tests out into one passing and one failing test, and I'm now getting ICEs in the tests that are supposed to pass.

Interesting! I am curious what the root cause will be. Thanks for digging deeper into this!

ironcev:
@jjcnn Perhaps this might be helpful for your analysis of ICEs. While working on supporting nonsensical referencing and dereferencing like e.g. `&return;` or `*&break;` I've realized that ICEs in other places might be coming from the same issue: not checking the `<value>.is_diverging(context)` for the expressions compiled in _sway-core/src/ir_generation/function.rs_.

jjcnn:
> @jjcnn Perhaps this might be helpful for your analysis of ICEs. While working on supporting nonsensical referencing and dereferencing like e.g. `&return;` or `*&break;` I've realized that ICEs in other places might be coming from the same issue: not checking the `<value>.is_diverging(context)` for the expressions compiled in _sway-core/src/ir_generation/function.rs_.

I've found two places where ICEs are thrown, but I'm not sure distinguishing on `is_diverging()` is the correct solution:

- `functions.rs`, function `compile_enum_expr`, line 2522, : A diverging value doesn't have a type, so the calculation of `contents_type` fails with an ICE. Distinguishing on `is_diverging` is the simple solution, but I think the correct solution is to calculate the type based on the enum declaration rather than on the value being used.
- `convert.rs`, function `convert_resolved_typeid`, line 71: The assumption in the type conversion is that `Unknown` doesn't exist as a type at this stage in the compilation (see line 163), which suggests that the typechecker isn't handling `return` (and therefore also `break` and `continue`) correctly.

I'll dig a bit deeper into the typechecker after New Year's, but without knowing much about the IR I think I'll need some help figuring out what needs to be done. 

On a side note, I think we need tests for `return 42 || return 43` and `return 42 && return 43` as well, which I'll add tonight.

ironcev:
> I've found two places where ICEs are thrown, but I'm not sure distinguishing on is_diverging() is the correct solution:

My quick take on both points, without taking a close look, but based on how things are handled on the IR side at the moment:

Ad 1) (`compile_enum_expr`) checking `is_diverging` should be the solution here. In general, out of what I see in code, (almost) all the places where we are calling e.g. `compile_expression_to_value` and not checking if the result `is_diverging` will end up in ICE. I did a quick try on e.g. intrinsics which are not covered, and that is the case e.g.

```Sway
script;

fn main() -> u64 {
    let _ = __size_of_val(return 123u64);   
    42
}
```
will produce this error as expected:
```
Internal compiler error: UnknownType { span: Span { src (ptr): 0x55f793b98c40, source_id: Some(SourceId { id: 67 }), start: 54, end: 67, as_str(): "return 123u64" } }
```
Currently, if any subexpression diverges, the code that checks divergence and returns it without proceeding works fine. Note that this means that we are not generating any IR code for anything that comes after it. E.g., function application checks if arguments are diverging which means the code like:

```Sway
let x = my_function(true, 123, return 42);
```
will generate IR that just returns 42, no `my_function` call.

This is definitely cumbersome and error prone at the moment, that (almost) every call of `compile_expression_to_value/ptr` needs to be checked. I am just documenting the existing behavior here.

Ad 2) (`convert_resolved_typeid`) This is a longer story. `Unknown` is in general not supported on the IR and should never pass the type checking level. Which it doesn't except for diverging expressions where it is (mis)used instead of the `Never` type which is the bottom type which we have in `core` lib but not supported in the type system. But my assumption here is again, that the `convert_resolved_typeid` gets called on a value which is returned from some `compile_expression_to_value` call that has returned diverging result which was not shortcircuted and returned but passed down the chain.

Long story short, on the long term we should support `Never`, but that will not solve the above issues. We would still need to track and cover all the places on IR which return `Never` which are essentially all the `Value`s for which `is_diverging` is true.

jjcnn:
I think the PR is ready for review. There are a couple of outstanding questions I'm aware of:

- The test I've added in the `should_pass` directory is currently only compiling, not end-to-end tested. I don't know what the magical formula is to enable end-to-end testing.
- There is almost certainly an issue with the function `TyExpression::deterministically_aborts()`, which is used at least once during IR conversion. I think it's out of scope to deal with it in this PR, but we should definitely take a look at it.
- We're not getting all the `unreachable code` warnings I'm expecting, but I think it's because we're bailing out on some of the IR conversion once we see a diverging value. I suspect it's better to deal with that problem in a separate PR.
- We need similar tests for `break` and `continue`, and we definitely need to include the test `return break`, which will almost certainly cause an issue when used in the wrong context. I'm pretty sure `deterministically_aborts()` gives the wrong result in that case.

sway-core/src/ir_generation/function.rs
```diff
@@ -485,7 +485,9 @@ impl<'eng> FnCompiler<'eng> {
                 ..
             } => {
                 let enum_decl = self.engines.de().get_enum(enum_ref);
-                self.compile_enum_expr(context, md_mgr, &enum_decl, *tag, contents.as_deref())
+                let res =
```
ironcev, line 488: In other branches we do not use the temporary `res` variable. I don't see a benefit in having it here either and would go for the previous code without `res`.

sway-core/src/ir_generation/function.rs
```diff
@@ -2611,6 +2639,10 @@ impl<'eng> FnCompiler<'eng> {
             // Insert the value too.
             let contents_value =
                 self.compile_expression_to_value(context, md_mgr, contents.unwrap())?;
+            // Only store if the the value does not diverge.
```
ironcev, line 2642:
```suggestion
            // Only store if the value does not diverge.
```

sway-core/src/ir_generation/function.rs
ironcev, line 488:
```suggestion
```

sway-core/src/ir_generation/function.rs
```diff
@@ -485,7 +485,9 @@ impl<'eng> FnCompiler<'eng> {
                 ..
             } => {
                 let enum_decl = self.engines.de().get_enum(enum_ref);
-                self.compile_enum_expr(context, md_mgr, &enum_decl, *tag, contents.as_deref())
+                let res =
+                    self.compile_enum_expr(context, md_mgr, &enum_decl, *tag, contents.as_deref());
```
ironcev, line 489:
```suggestion
                self.compile_enum_expr(context, md_mgr, &enum_decl, *tag, contents.as_deref())
```

sway-core/src/ir_generation/function.rs
```diff
@@ -485,7 +485,9 @@ impl<'eng> FnCompiler<'eng> {
...
+                res
```
ironcev, line 490:
```suggestion
```

test/src/e2e_vm_tests/test_programs/should_pass/return_in_strange_positions/test.toml
```diff
@@ -0,0 +1,72 @@
+category = "compile"
```
ironcev, line 1: To actually run the tests `category = "run"` should be used along with the checks. Here is an example how this can be done: _/sway/test/src/e2e_vm_tests/test_programs/should_pass/language/implicit_casting/test.toml_

ironcev (approved):
Aside from a typo and the unneeded temporary variable LGTM.

Passing tests should definitely be made runnable as noted in a comment.

sway-core/src/ir_generation/function.rs
vaivaswatha: although in this particular instance I don't have a strong opinion either way, my preference is to not have review comments on code style, given that we have clippy to enforce standard practices. if there is a style preference that we want everyone to follow that clippy doesn't enforce, we should have a doc that lists them all. if a particular style requirement cannot be generalized to be defined in th doc, it shouldn't be requested in a review. that'll make reviews smoother without personal preferences coming in the way.

I re-emphasise, my point is, although related to the comment, not very specific to it. I meant to convey it for sometime now and found a place here.

ironcev: Seeing the review suggestions now in the thread - it really blows the comments unnecessary. Shame on me!

To the personal style, I also don't have strong preferences also in this case and don't think we need a separate doc. Slight style differences we have in code are not worth it.

What I do like to keep is the consistency of the code regardless of the preferences. In this particular case, why to introduce a purely stylistic change that is not related to the PR and breaks consistency with all the code around it? This I think makes sense to point out in the review.

But fully agree with you, definitely no need to bloat the review with 5 comments for that one tiny line! It could have been just one small remark.

vaivaswatha: > it really blows the comments unnecessary. Shame on me!

 no no please, that's too strong. I only meant to point out about personal coding style preferences.

overall I think we're in agreement 🤝☺️

jjcnn: In general I agree about code style comments, but in this particular instance the unnecessary local variable is a remnant from some debugging code I introduced, and I forgot to remove the local. I'll remove it.

jjcnn:
> Ad 1) Magical formula 😄 given in the comment.

This causes a revert in the `return 42 + return 43` test. 

I'll need to do some digging to see what's going on, so I'm holding off on the merge.

> 
> Ad 2) Agree, definitely an issue and let's deal with it separately.

I'll open an issue on it.

> 
> Ad 3) We have issue for that #5414. It's also a separate PR.
> 

If you don't mind I'll assign myself to it. I've already got my hands dirty on control-flow expressions, so I should be in a good position to fix that one too.

jjcnn:
> > Ad 1) Magical formula 😄 given in the comment.
> 
> This causes a revert in the `return 42 + return 43` test.
> 
> I'll need to do some digging to see what's going on, so I'm holding off on the merge.
> 

@ironcev : This issue is caused because `return 42 + return 43` is grouped as `return (42 + return 43)`, so the function returns 43 rather than 42, and the assert in the `main` function fails.

This is counter-intuitive, but I think it's actually working as intended, so I'll just add a comment in the test case and make sure the assert works. Then we should be good to go.

ironcev:
> If you don't mind I'll assign myself to it.
Sure go ahead :smile: 

> This is counter-intuitive, but I think it's actually working as intended
Yes exactly. The priority of operations goes as you said: `return (42 + return 43)`.

ironcev (approved):

IGI-111 (approved):

PR #5418, opened 2023-12-21
Fix resolution of local items that shadow imported ones

### LINKED ISSUE #5383: Wrong resolution of items when using glob imports
Local items should shadow glob-importing items with the same name. We have a bug, that an imported item get's resolved instead of the local one, as demonstrated on this example. 

```Sway
library;

pub struct Struct {
    x: u64,
}
```

```Sway
script;

mod lib;

use lib::*; // This imports `Struct` from the library that doesn't have `y` field.

struct Struct { // Local `Struct` that should be resolved below.
   x: u64,
   y: u64,
}

fn main() {
   let _s = Struct { x: 0, y: 0 };
   // WRONG ERROR: ^^^^ Struct "Struct" does not have field "y".
   // It happens because lib::Struct gets resolved. 
}
```

ironcev:
## Description

Closes #5383.

#5383 blocks PR #5306.

## Checklist


test/src/e2e_vm_tests/test_programs/should_pass/resolve_local_items_that_shadow_imports/test.toml
```diff
@@ -0,0 +1,23 @@
...
+#check: $()pub struct PubStruct {
+#nextln $()This struct is never used.
+
+#check: $()warning
+#check: $()pub struct GenericStruct<T> {
+#nextln: $()This struct is never used.
+
+#check: $()warning
+#check: $()pub const X: u64 = 0u64;
+#nextln: $()This declaration is never used.
```
jjcnn, line 23: Are these useful warnings? Shouldn't it be fine to import all declarations from a library (using `use lib::*;`) and then not use all of them?

In this example none of the imported declarations are used, but then the warning should be something like "Unused import 'use lib::*;'".

jjcnn (commented):
This definitely fixes a bug, but there's still an issue when the imported declaration is used before the local declaration:

```
use lib::Enum;

pub const Y: Enum = Enum::A;

enum Enum {
    A: (),
    B: (),
}
```
The type `Enum` in the declaration of `Y` resolves to the locally defined type rather than the imported one.

It's probably intentional that type declarations are in scope in the entire file, but I was surprised to see it resolve this way, so maybe it's worth it to issue a warning when a used type name has multiple resolutions?

In any case the scope of type declarations is unclear from the documentation.

ironcev:
> This definitely fixes a bug, but there's still an issue when the imported declaration is used before the local declaration

This is actually fine. items follow item-style shadowing and not sequential shadowing. They are visible in the whole module no matter where they are declared. E.g.:

```Sway
fn first() {
   second();
}

fn second() {
   let _ = Enum::A;
}

enum Enum {
    A: (),
}
```
Which means `Enum` is visible everywhere and thus shadows the imported one everywhere.

In the above example we will also get an error for the name clash, but the resolution should still be to the local one, because the shadowing is not sequential but item-style.

To item-style shadowing and sequential shadowing, I like the explanation/definitions given in the [Rust RFC 116 (No module shadowing)](https://rust-lang.github.io/rfcs/0116-no-module-shadowing.html).

test/src/e2e_vm_tests/test_programs/should_pass/resolve_local_items_that_shadow_imports/test.toml
ironcev: This is a great question, but out of scope of this PR :smile: It's a question for our DCA and how it works with local libraries, those used in a project that is not only a library. To my understanding, those items are not expected to be reused outside of the project and this if they are public and not used anywhere in the project they trigger dead code warnings. I _think_ it works that way.

BDW I was thinking about the same things while working on making struct fields private by default.

Anyhow, my proposal is to open a separate issue to discuss this topic or get clarifications.

For now, since these warnings are there I am (mis?)using them to additionally prove that those types are really not resolved in the module.

xunilrj (approved):

jjcnn:
> To item-style shadowing and sequential shadowing, I like the explanation/definitions given in the [Rust RFC 116 (No module shadowing)](https://rust-lang.github.io/rfcs/0116-no-module-shadowing.html).

"Due to a certain lack of official, clearly defined semantics and terminology, ..." 

Suffice it to say I don't. :-)

jjcnn (approved):

ironcev:
> Suffice it to say I don't. :-)

Well, as long as we don't have anything better... :smile:

PR #5423, opened 2023-12-28
Added tests for the `sqrt()` method in `math.sw`

### LINKED ISSUE #5330: `math.sw` is missing tests
It appears that most of the functionality in that module is untested, at least in Sway.
If it's tested in our Rust suite then it's not easily findable.

We should test the module.

brandonsurh:
## Description
Closes #5330

Previously, there were no tests for the `sqrt()` method in `math.sw`. This adds square root tests in the Rust test suite.
Other tests are found here:
- [`pow()`](https://github.com/FuelLabs/sway/blob/master/test/src/e2e_vm_tests/test_programs/should_pass/stdlib/exponentiation_test/src/main.sw)
- [`log()` and `log2()`](https://github.com/FuelLabs/sway/blob/master/test/src/e2e_vm_tests/test_programs/should_pass/stdlib/logarithmic_test/src/main.sw)

## Checklist


xunilrj:
Nice, but these can be unit tests inside the stdlib, right? `sway-lib-std/src/math.sw` has an example already.

brandonsurh:
> Nice, but these can be unit tests inside the stdlib, right? `sway-lib-std/src/math.sw` has an example already.

You are correct. I would actually prefer the tests for `math.sw` to be right in there, but the pre-existing ones for `pow()`, `log()`, and `log2()` were made elsewhere. The reason I placed them in the Rust suite was to keep things uniform, but if desired I can move all of them into `math.sw`.

xunilrj:
>  but if desired I can move all of them into math.sw.

If possible I think makes more sense. Thanks! :D

xunilrj (approved):

brandonsurh:
> If possible I think makes more sense. Thanks! :D

Moved the tests for `pow()`, `log()`, and `log2()` into `math.sw`. Everything should be good on that front! However, may need to take a look at the failing checks.

JoshuaBatty (approved):

PR #5427, opened 2024-01-02
New encoding for contract calls 

xunilrj:
## Description

This PR implements the new encoding for contracts/scripts/predicates. https://github.com/FuelLabs/sway/issues/5512 and https://github.com/FuelLabs/sway/issues/5513

### Contract Calls

When the new encoding is turned on using `--experimental-new-encoding`, contract calls like the example below will be "desugarized" differently now.

```sway
let base_asset_id = BASE_ASSET_ID;
let other_contract_id = ContractId::from(0xa38576787f8900d66e6620548b6da8142b8bb4d129b2338609acd121ca126c10);

let test_contract = abi(ContextTesting, other_contract_id.into());
let returned_contract_id = test_contract.get_id { gas: gas, coins: 0, asset_id: BASE_ASSET_ID.value}(1, 2, 3);
```

and will be transformed to

```sway
let base_asset_id = BASE_ASSET_ID;
let other_contract_id = ContractId::from(0xa38576787f8900d66e6620548b6da8142b8bb4d129b2338609acd121ca126c10);

let test_contract = abi(ContextTesting, other_contract_id.into());
let returned_contract_id = contract_call::<ContractId, _>(other_contract_id.into(), "get_id", (1, 2, 3), coins, asset_id, gas);
```

And the important part is the `contract_call` function in the std library. This function does all the encoding as necessary and delegates the actual call to an intrinsic function `__contract_call`. Allowing the protocol to evolve entirely in Sway.

```sway
pub fn contract_call<T, TArgs>(contract_id: b256, method_name: str, args: TArgs, coins: u64, asset_id: b256, gas: u64) -> T
where
    TArgs: AbiEncode
{
    let first_parameter = encode(method_name);
    let second_parameter = encode(args);
    let params = encode(
        (
            contract_id,
            asm(a: first_parameter.ptr()) { a: u64 },
            asm(a: second_parameter.ptr()) { a: u64 },
        )
    );
    __contract_call::<T>(params.ptr(), coins, asset_id, gas)
}
```

### Contracts

On the other side, when the flag `--expiremental-new-encoding` is turned on, the contract specification like the one below is being transformed into all the decoding and encoding necessary. 

The mains points are:

- The compiler generates a function called `__entry` that decodes the method name and its arguments. The method is selected with a bunch of `if`s at the moment, because we don´t have `match` for string slices. Then we `decode` the arguments using the correct type, which is a tuple with all the function arguments, and expand this tuple calling the function;
- All the contract functions are converted to global functions prefixed with `__contract_method`;
- Results are encoded and returned using the intrinsic call `__retd`.

Example:

```sway
abi SomeContract {
    fn some_function(a: u64) -> u64;
}

impl SomeContract for Contract {
    fn some_function(a: u64) -> u64 {
        1
    }
}
```

will be transformed into 

```sway
fn __entry() {
    let method_name = decode_first_parameter();
    if method_name == "some_function" {
        let args = decode_second_parameter::<(u64,)>();
        let result = __contract_method_some_function(args.0);
        __retd(encode(result));
    }
    __revert(0);
}
```

### Scripts and Predicates

The protocol to call scripts and predicates will also change and will be very similar to contracts. See more above. Now when the flag is turned on, the `main` function will not be entry point anymore. The compiler will actually generate an `__entry` function that will decode arguments and encode the result, like contracts. 

For example:

```sway
fn main(a: u64) -> u64 {
    1
}
```

will be transformed into

```sway
fn __entry() {
    let args = decode_script_data::<(u64,)();
    let result = main(args.0);
    __retd(encode(result));
}

fn main(a: u64) -> u64 {
    1
}
```

## Tests

To facilitate testing this PR introduces three changes to our test harness:

1 - A new parameter can be called to update test output files (abi and storage json). This facilitates when we only want to copy and paste these output files to their respective oracles. Brings the framework closer to a snapshot one.

``` 
> cargo r -p test --release -- --update-output-files
```

2 - Depending on the configuration at `test.toml` multiple executions of the same test will be scheduled. At the moment, tests that depend on the encoding will run with the `--experimental-new-encoding` flag automatically. For example:

```
Testing should_pass/language/main_args/main_args_copy_copy ... ok
Testing should_pass/language/main_args/main_args_copy_copy (New Encoding) ... ok
```

3 - A new `script_data_new_encoding` was created because we want to support and run tests with the two encoding for a time. This is also what flags tests to run with the flag on automatically.

## Checklist


Riskypounds:
Very good

sway-core/src/semantic_analysis/program.rs
```diff
@@ -1,23 +1,228 @@
...
+                type_arguments: TypeArgs::Regular(vec![]),
+                span: Span::dummy(),
+            },
+            contract_call_params: vec![],
+            arguments: vec![l, r],
+        })),
+        span: Span::dummy(),
+    }
+}
+
```
jjcnn, line 109: This isn't directly relevant to the code I've marked, or indeed to the PR itself, but it did make me think: What's the semantics for passing parameters in contract calls? Is it always call-by-value (with copying), or can you do call-by-reference?

jjcnn: Same question for values returned by contract calls, BTW: Can you return a reference, or does it have to be a value?

jjcnn (commented):
I found it easier to see what's going on by looking at each commit separately, and I'm done all the way through to the one called "clean tests".

I'll continue with the rest of the commits tomorrow.

sway-core/src/semantic_analysis/program.rs
xunilrj: We are "encoding" all arguments to bytes on the caller side, and decoding it on the "callee" side. So it is "call-by-value" that uses the new encoding to do a "deep copy", instead of a shallow `memcpy`.

Same thing for return values: "deep copy using the new encoding".

> Can you return a reference, or does it have to be a value?

Today the only type that does not have a default impl for AbiEncode/AbiDecode is `raw_pointer` because we dont want to pass pointers across contracts silently. 

Of course you can impl them yourself and decide to pass the pointer, but that means you would be inside "unsafe world".

Same thing for references. They will not be AbiEncode/AbiDecode by default.

xunilrj:
> Are gas and coins implicitly defined variables?

No they are standard variables. Nothing special about them.

> Is it necessary to defined base_asset_id instead of just using BASE_ASSET_ID in the contract call?

It is not necessary, but it is an option that we give when a contract call is made. See https://docs.fuel.network/docs/sway/sway-program-types/smart_contracts/#calling-a-smart-contract-from-a-script

> You also have the option of specifying the following special parameters inside curly braces right before the main list of parameters:
>
> gas: a u64 that represents the gas being forwarded to the contract when it is called.
> coins: a u64 that represents how many coins are being forwarded with this call.
> asset_id: a b256 that represents the ID of the asset type of the coins being forwarded.

sway-lib-core/src/codec.sw
```diff
@@ -386,15 +508,42 @@ where
...
 
+impl<A, B, C, D, E, F> AbiEncode for (A, B, C, D, E, F)
+where
+    A: AbiEncode,
+    B: AbiEncode,
+    C: AbiEncode,
+    D: AbiEncode,
+    E: AbiEncode,
+    F: AbiEncode,
+{
```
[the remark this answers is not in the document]
ironcev: In this case they all actually represent "any type". None of those type variables and types they represent is anyhow special. Almost dummies I would say. That's why I see it difficult to give them more meaningful names in this and other cases of `impl AbiEncode`. The convention in this case is `A`, `B`, etc. C# uses `T1`, `T2`, etc for the same case. That's why I am for staying at the existing convention. In general I am also for more meaningful names when type variables have semantics. E.g. in the `contract_call<T, TArgs>` more specific would be `contract_call<TRet, TArgs>`.

jjcnn:
> > Are gas and coins implicitly defined variables?
> 
> No they are standard variables. Nothing special about them.
> 

So the examples are not self-contained, then? For instance, I see a variable `gas` which is not defined anywhere, and which is used to initialize the `gas` parameter to the contract call.

ironcev:
> So the examples are not self-contained

Exactly, they are more like code snippets.

sway-core/src/ir_generation/function.rs
```diff
@@ -1735,16 +1804,26 @@ impl<'eng> FnCompiler<'eng> {
...
-            callee
-                .parameters
-                .iter()
-                .map(|p| p.type_argument.type_id)
-                .collect(),
-            callee.type_parameters.iter().map(|tp| tp.type_id).collect(),
-        );
-        let new_callee = match self.recreated_fns.get(&fn_key).copied() {
+
+        let (fn_key, item) = if !callee.span().as_str().is_empty() {
```
ironcev, line 1808: Here is how I understand the change done to function generation: if the function is not compiler generated (`!callee.span().as_str().is_empty()`) generate the key and either get already compiled function by that key or create it and insert under that key. But if it is compiler generated, generate it every time it is called.

I guess this works for now because the functions generated for contracts calls, and also `__entry` are called only once. Still, looking into future and possible generation of functions for different purposes (thinking of storage right now), can we think of:
1. Adding `is_compiler_generated()` method to `TyFunctionDecl` and eventually think of other ways of determined if a function is compiler generated. E.g. in the _compiler_generated.rs_ we used name prefixes to determine if variables are compiler generated.
2. Create a unique `fn_key` also for the compiler generated methods. I assume we will still have enough distinguished elements like e.g. unique names and the parameters and type parameters we will anyhow have.

sway-core/src/language/asm.rs
```diff
@@ -10,6 +10,17 @@ pub struct AsmOp {
     pub(crate) immediate: Option<Ident>,
 }
 
+impl AsmOp {
+    pub fn retd(ptr: BaseIdent, len: BaseIdent) -> Self {
```
ironcev, line 14: Is this function ever used? If yes, do we always need to lose span information?

sway-core/src/language/parsed/expression/asm.rs
```diff
@@ -19,3 +19,12 @@ pub struct AsmRegisterDeclaration {
     pub(crate) name: Ident,
     pub initializer: Option<Expression>,
 }
+
+impl AsmRegisterDeclaration {
+    pub fn with_init(name: BaseIdent, initializer: Expression) -> Self {
```
ironcev, line 24: Same here, is it ever used?

sway-core/src/language/parsed/expression/match_branch.rs
```diff
@@ -1,10 +1,33 @@
...
 
 #[derive(Debug, Clone)]
 pub struct MatchBranch {
     pub scrutinee: Scrutinee,
     pub result: Expression,
     pub(crate) span: span::Span,
 }
+
+impl MatchBranch {
+    pub fn catch_all(result: Expression) -> Self {
```
ironcev, line 13: Also these two functions, are they ever used? Sorry for being picky about this, but since they are `pub` I assume they will not be reported as dead by Clippy and the rustc.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression/method_application.rs
```diff
@@ -337,13 +343,141 @@ pub(crate) fn type_check_method_application(
...
+                        .collect(),
+                ),
+                None,
+            );
+            Expression {
+                kind: ExpressionKind::FunctionApplication(Box::new(
+                    FunctionApplicationExpression {
+                        call_path_binding: TypeBinding {
+                            inner: CallPath {
+                                prefixes: vec![],
```
ironcev, line 386: I assume we need to put `["core", "codec"]` here to get a unique and full call path. Otherwise, if there is a local function called "contract_call" that one will be resolved.

ironcev (commented):
Congrats for putting all this together @xunilrj!!

I am still going through the PR and connecting the dots. The overall mechanics is already clear (`contract.method()` being desugared into `contract_call` that in the end calls the intrinsic) but I still didn't touch the part with the heavy code generation.

The comments posted in this review are IMO negligible compared to the overall complexity and importance of the PR and can be ignored for now if there are more pressing issues.

The only one that I think can lead to bugs is the not fully qualified name in the `contract_call` function application.

I'll proceed with the review.

sway-core/src/ir_generation/compile.rs
```diff
@@ -571,16 +591,14 @@ fn compile_abi_method(
...
-    let is_entry = true;
-
     compile_fn(
         engines,
         context,
         md_mgr,
         module,
         &ast_fn_decl,
-        is_entry,
+        // ABI are only entries when the new encoding if off
```
ironcev, line 600:
```suggestion
        // ABI are only entries when the new encoding is off
```

sway-core/src/language/parsed/expression/mod.rs
```diff
@@ -29,6 +29,201 @@ pub struct Expression {
     pub span: Span,
 }
 
+impl Expression {
```
ironcev, line 32: I assume this was used before in the previous approach when the AST was built manually in code. Should we leave it or remove it? That's a lot of code that could be used for desugaring in the future or refactoring existing desugaring. Perhaps moved to a dedicated module, turned into builder pattern, ... I don't know. I am afraid if it just stays where it is now and unused it will remain dead code.

sway-core/src/language/parsed/mod.rs
```diff
@@ -41,6 +47,277 @@ pub struct AstNode {
     pub span: Span,
 }
 
+impl AstNode {
```
ironcev, line 50: Same concern here.

sway-core/src/language/ty/declaration/declaration.rs
```diff
@@ -378,6 +378,17 @@ impl TyDecl {
         }
     }
 
+    pub fn get_trait_decl_ref(&self) -> Option<DeclRefTrait> {
```
ironcev, line 381: Nitpick, the implementation sticks out and breaks the pattern that all other `get_xyz_decl_ref` methods have :wink:

ironcev:
General remark. @tritao's concern about debug symbols and loosing span information when desugaring is valid, especially in this approach when we generate the code by compiling code snippets. It looks to me we will need something like `#line` directive for generated code.

ironcev:
When I compile this simple predicate:
```Sway
predicate;

fn main() -> bool {
   true
}
```
the compiler panics in `aut_impl`:
```
thread 'main' panicked at sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs:679:9:
Handler { inner: RefCell { value: HandlerInner { errors: [SymbolNotFound { name: decode_script_data, span: Span { src (ptr): 0x55b8522472e0, source_id: None, start: 50, end: 68, as_str(): "decode_script_data" } }], warnings: [] } } } pub fn __entry() -> bool {
            let args = decode_script_data::<()>();
            main()
        }
```

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
```diff
@@ -471,404 +396,703 @@ impl<'a, 'b> AutoImplAbiEncodeContext<'a, 'b> {
...
+        let expanded_args = itertools::intersperse(
+            decl.parameters.iter()
+                .enumerate()
+                .map(|(i, _)| format!("args.{i}")),
+            ", ".into()
+        ).collect::<String>();
+
+        let code = format!("pub fn __entry() -> bool {{
+            let args = decode_script_data::<{args_types}>();
+            main({expanded_args})
```
ironcev, line 1060: What will happen if `main` is wrongly declared and does not return `bool`? E.g. `fn main() -> u64`. How will error reporting look like?

Right now, I assume `__entry` will be generated nevertheless and while parsing it we will have an additional type mismatch error in the generated code which will trigger an assert in the `parse_item_fn_to_typed_ast_node`.

sway-core/src/ir_generation/function.rs
```diff
@@ -1247,6 +1247,76 @@ impl<'eng> FnCompiler<'eng> {
...
+                    md_mgr,
+                    &arguments[3]
+                )?);
+
+                let span_md_idx = md_mgr.span_to_md(context, &span);
+
+                let returned_value = self
+                    .current_block
+                    .append(context)
+                    .contract_call("SOMETHING".into(), params, coins, asset_id, gas)
```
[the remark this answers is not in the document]
xunilrj: Good idea. It is optional now.
While we are checking the new encoding, I will check if this is useful or not.

sway-lib-core/src/codec.sw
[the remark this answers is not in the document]
xunilrj: Yeah... I agree, but don't have any ideas how to improve it. Open to suggestions.

This code is now auto generated with a bash script, so it is super easy to change.

sway-core/src/abi_generation/fuel_abi.rs
```diff
@@ -44,8 +44,14 @@ pub fn generate_program_abi(
                 configurables: Some(configurables),
             }
         }
-        TyProgramKind::Script { main_function, .. }
-        | TyProgramKind::Predicate { main_function, .. } => {
+        TyProgramKind::Script {
+            entry_function: main_function,
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/asm_generation/fuel/compiler_constants.rs
```diff
@@ -36,5 +36,5 @@ pub(crate) const EIGHTEEN_BITS: u64 = 0b11_1111_1111_1111_1111;
 pub(crate) const TWELVE_BITS: u64 = 0b1111_1111_1111;
 pub(crate) const SIX_BITS: u64 = 0b11_1111;
 
-/// Some arbitrary values used for error codes.
+// Some arbitrary values used for error codes.
 pub(crate) const MISMATCHED_SELECTOR_REVERT_CODE: u32 = 123;
```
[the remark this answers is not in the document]
xunilrj: Agreed. But this was not inserted by my PR. :)

test/src/e2e_vm_tests/test_programs/should_pass/language/main_args/main_args_empty/src/main.sw
```diff
@@ -1,5 +1,53 @@
...
-    1337
+struct S<T> {
+    a: T
+}
+
+enum E<T> {
+    a: T
+}
+
+pub fn main() {
```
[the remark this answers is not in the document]
xunilrj: Yes. Its fixed now.

sway-core/src/ir_generation/function.rs
xunilrj: I will answer you in the comments because I think this is important.

xunilrj:
> Here is how I understand the change done to function generation: if the function is not compiler generated (!callee.span().as_str().is_empty()) generate the key and either get already compiled function by that key or create it and insert under that key. But if it is compiler generated, generate it every time it is called.

Correct.

> I guess this works for now because the functions generated for contracts calls, and also __entry are called only once. Still, looking into future and possible generation of functions for different purposes (thinking of storage right now), can we think of:
> Adding is_compiler_generated() method to TyFunctionDecl and eventually think of other ways of determined if a function is compiler generated. E.g. in the compiler_generated.rs we used name prefixes to determine if variables are compiler generated.
> Create a unique fn_key also for the compiler generated methods. I assume we will still have enough distinguished elements like e.g. unique names and the parameters and type parameters we will anyhow have.

Currently, in the case of a generated function is called multiple times we will depend on the `dedup` optimization to avoid code bloat.

In the future I think we can actually have the key itself inside the `TyFunctionDecl`, allowing that parsed and generated function to never collide or anything.

sway-core/src/language/asm.rs
xunilrj: We don't need these function at the moment. I will remove them.

sway-core/src/language/parsed/expression/asm.rs
xunilrj: Removed.

sway-core/src/language/parsed/expression/match_branch.rs
xunilrj: Removed.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression/method_application.rs
xunilrj: Will also answer this one in the comments.

xunilrj:
> I assume we need to put ["core", "codec"] here to get a unique and full call path. Otherwise, if there is a local function called "contract_call" that one will be resolved.

This is also a choice we need to make. When I generate code that calls things that live inside `core` I opted to not put the complete path so the code can still compile without the `core` method.

We intend to mark these special functions with attributes `#[entry]`, `#[abi_encode]` etc... So the user could build everything from scratch., but until there, we allow them to write their own with we do not specify the absolute path.

sway-core/src/decl_engine/associated_item_decl_id.rs
```diff
@@ -116,18 +116,30 @@ impl DisplayWithEngines for AssociatedItemDeclId {
             Self::TraitFn(decl_id) => {
                 write!(
                     f,
-                    "decl(trait function {})",
-                    engines.de().get(decl_id).name()
+                    "decl(trait function {:#?})",
```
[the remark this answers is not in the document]
xunilrj: I think you meant `DebugWithEngines`, right? 
But you are right. I kept `DisplayWithEngines` as simpler as possible.

sway-core/src/decl_engine/engine.rs
```diff
@@ -441,11 +441,17 @@ impl DeclEngine {
...
         let mut builder = String::new();
-        let mut list = vec![];
-        for func in self.function_slab.values() {
-            list.push(format!("{:?}", engines.help_out(&*func)));
+        let mut list = String::with_capacity(1024 * 1024);
+        let funcs = self.function_slab.values();
+        for (i, func) in funcs.iter().enumerate() {
+            list.push_str(&format!("{i} - {:?}\n", engines.help_out(&*func)));
+
+            if func.name.as_str().contains("__entry") || func.name.as_str().contains("main") {
```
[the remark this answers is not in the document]
xunilrj: Your are right. I will remove this. It was more for debugging.

sway-core/src/ir_generation.rs
```diff
@@ -63,7 +63,9 @@ pub fn compile_program<'eng>(
     match kind {
         // predicates and scripts have the same codegen, their only difference is static
         // type-check time checks.
-        ty::TyProgramKind::Script { main_function } => compile::compile_script(
+        ty::TyProgramKind::Script {
+            entry_function: main_function,
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/ir_generation/compile.rs
xunilrj: Done.

sway-core/src/language/parsed/mod.rs
xunilrj: Removed.

xunilrj:
> General remark. @tritao's concern about debug symbols and loosing span information when desugaring is valid, especially in this approach when we generate the code by compiling code snippets. It looks to me we will need something like #line directive for generated code.

I will need to dive into our solution for debug, but I do intend for generated code to have spans. I imagine we will need these spans to point to real files to debug to work. 

I have created this issue to track this: https://github.com/FuelLabs/sway/issues/5696

sway-core/src/abi_generation/evm_abi.rs
```diff
@@ -15,8 +15,14 @@ pub fn generate_abi_program(program: &TyProgram, engines: &Engines) -> EvmAbiRes
             .iter()
             .map(|x| generate_abi_function(x, type_engine, decl_engine))
             .collect(),
-        TyProgramKind::Script { main_function, .. }
-        | TyProgramKind::Predicate { main_function, .. } => {
+        TyProgramKind::Script {
+            entry_function: main_function,
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/decl_engine/parsed_engine.rs
```diff
@@ -327,4 +327,22 @@ impl ParsedDeclEngine {
     {
         self.get(index)
     }
+
+    pub fn pretty_print(&self) -> String {
```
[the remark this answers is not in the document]
xunilrj: This one I think we can leave as it is. Because it will help future debug.

sway-core/src/language/parsed/expression/mod.rs
xunilrj: Removed.

xunilrj:
> When I compile this simple predicate:

How were you pointing to the new compiler and the new core?
Because we have similar `predicate` examples that are compiling, even without the `std` lib.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
```diff
@@ -1,47 +1,77 @@
...
+
+        assert!(!p.has_errors());
+        assert!(!p.has_warnings());
+
+        r.unwrap()
+    }
+
+    fn resolve_type(ctx: &mut TypeCheckContext<'_>, call_path: CallPath) -> Option<TypeId> {
         let handler = Handler::default();
         let buffer_type_id = ctx.engines.te().insert(
```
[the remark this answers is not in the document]
xunilrj: This method does not exists anymore.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
xunilrj: Currently is correctly only generating one error:

```
Testing should_pass/language/predicate_while ... failed
     Failed to compile predicate_while
          Compiling should_pass/language/predicate_while ...

          Failed to compile predicate_while
          error
           --> /home/xunilrj/github/sway/test/src/e2e_vm_tests/test_programs/should_pass/language/predicate_while/src/main.sw:3:1
            |
          1 |   
          2 |   
          3 | / fn main() -> u64 {
          4 | |     0
          5 | | }
            | |_^ A predicate's main function must return a boolean.
            |
          ____

            Aborting due to 1 error.
```

What happens is that the module `type_check` fails and the code is never generated.

ironcev:
> This is also a choice we need to make. When I generate code that calls things that live inside core I opted to not put the complete path so the code can still compile without the core method.

I've been thinking about this too when realized the overlook I made that lead to #5522. Essentially, `core` is needed to so many language features that I pondered if it should be always included, like not optional but always, unlike `std` that is optional. Right now I am still more for, `core` is optional and we give a good guiding error message in particular cases (like #5522) saying if you want to use this feature, you must import `core`.

This would be my approach to `contract_call` as well - an error stating `core` must be imported. To justify my thinking, this is the part I do not understand:

> I opted to not put the complete path so the code can still compile without the core method

Essentially, without `core` if I try to call a contract the code will _not compile_, or? It will still tell me that `contract_call` cannot be found because I will surely not have one in my code. And I will for sure not write one in my code nor know how to actually do it.

In addition, if there is a local method called `contract_call` (not impossible!) and considering the fact that most of the Sway programmers will never see the `contract_call` in the `core` library, having the call resolve to local one will cause a plenty of confusion.

That's why, until we introduce the mentioned attributes `#[entry]` etc, I am still for the full function name and error stating `core` must be imported.

ironcev:
> How were you pointing to the new compiler and the new core?

I've just created a new predicate with `forc new --predicate` and then run:

```
repos/sway$ cargo run --bin forc -- build --experimental-new-encoding -p ~/path/to/predicate
```

Of course on the `xunilrj/encoding-contract-call` branch.

ironcev (approved):
As discussed, the changes are safely behind a feature flag and the feature needs refinement a huge part of which can come only if it is used and the feedback is given from the usage. Therefore I am for merging the PR as it is now and proceeding with the work on it in smaller chunks.

forc-pkg/src/pkg.rs
```diff
@@ -1802,7 +1809,10 @@ pub fn compile(
     };
 
     if profile.print_ast {
-        tracing::info!("{:#?}", typed_program);
+        // tracing::info!("{:#?}", &programs.parsed);
+        // tracing::info!("{:#?}", engines.pe().pretty_print());
+        // tracing::info!("{}", engines.de().pretty_print(&engines));
+        // tracing::info!("{:#?}", typed_program);
```
IGI-111, line 1815: This needs to be cleaned up.

sway-core/src/language/ty/program.rs
```diff
@@ -400,6 +416,18 @@ impl TyProgram {
...
+    //     &'b self,
+    //     decl_engine: &'a DeclEngine,
+    //     tree_type: TreeType,
+    // ) -> impl '_ + Iterator<Item = DeclRefFunction> {
+    //     self.root
+    //         .submodules_recursive()
+    //         .flat_map(move |(_, submod)| submod.module.entry_fns(decl_engine, tree_type.clone()))
+    //         .chain(self.root.entry_fns(decl_engine, tree_type.clone()))
+    // }
+
```
IGI-111, line 430: Should this be removed?

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
```diff
@@ -1,875 +1,666 @@
...
-        // for structs inside "core"
-        if matches!(self.ctx.namespace().root_module_name(), Some(x) if x.as_str() == "core") {
-            return false;
-        }
+    fn parse_item_impl_to_typed_ast_node(
+        &mut self,
+        engines: &Engines,
+        code: &str,
+    ) -> Option<TyAstNode> {
+        // println!("{}", code);
```
IGI-111, line 304: This needs to be removed.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
```diff
@@ -1,875 +1,666 @@
...
+                "pub fn __entry() -> raw_slice {{
+                let args = decode_script_data::<{args_types}>();
+                let result: {return_type} = main({expanded_args}); 
+                encode::<{return_type}>(result)
+            }}"
+            )
+        };
+
+        // println!("Generate Script Entry");
+        // println!("---------------------");
```
IGI-111, line 663: This needs to be removed.

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
```diff
@@ -1409,3 +1418,97 @@ fn type_check_smo(
...
+        return Err(handler.emit_err(CompileError::TypeArgumentsNotAllowed { span }));
+    }
+
+    // if arguments.len() != 1 {
+    //     return Err(handler.emit_err(CompileError::IntrinsicIncorrectNumArgs {
+    //         name: kind.to_string(),
+    //         expected: 1,
+    //         span,
+    //     }));
+    // }
```
IGI-111, line 1487: Should this be removed?

test/src/ir_generation/mod.rs
```diff
@@ -257,6 +257,20 @@ pub(super) async fn run(
...
+                    println!("-----------------------");
+                    println!("{}", engines.de().pretty_print(&engines));
+
+                    // println!("Parsed");
+                    // println!("-----------------------");
+                    // dbg!(&programs.parsed);
+
+                    // println!("Typed");
+                    // println!("-----------------------");
+                    // dbg!(&programs.typed);
```
IGI-111, line 271: Might as well display it all if we're verbose.

IGI-111 (commented):
Collected leftover comments and printouts. I think this can be merged once these are cleaned up.

forc-pkg/src/pkg.rs
xunilrj: Done.

sway-core/src/language/ty/program.rs
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
xunilrj: Done.

IGI-111 (approved):

ironcev (approved):

PR #5433, opened 2024-01-04
chore: use sdk v0.54.0 to enable forc-client deployments to beta5

### LINKED ISSUE #5432: Use  SDK v0.54.0 for forc-client
We realized that we cannot deploy to our newest testnet due to a version mismatch between prod fuel-core compatible sdk versions and the ones we are using in forc-client.

kayagokalp:
## Description

closes #5432 

Bumps used forc-wallet version to 0.4.1 and used sdk versions to 0.54.0 so that we can actually start deploying to beta-5. As we need to be using forc-client 0.22 which requires us to go to 0.54.0 of sdk. To do so we need the wallet to use it as well.


kayagokalp:
blocked until #5435

sdankel (dismissed):

JoshuaBatty (approved):

PR #5441, opened 2024-01-08
Add `try_insert` to storage map

### LINKED ISSUE #5147: `StorageMap` insert entry if key has not had a value set
Rust has the following [map.entry("").or_insert();](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html)

It would improve UX if we had a `try_insert()` or `insert_if_empty()` method which checks if a value has been set for a key and if it has not then it inserts it.
This removes the need for higher level code to manually check if there is an entry and then insert it via another line of code if it has not been set (removing the need for an `if` expression at a high level)

SwayStar123:
## Description
Adds the `try_insert` function to the storage map in the standard library. Allows users to only insert when no prior value exists.

Closes #5147

## Checklist


sway-lib-std/src/storage/storage_map.sw
```diff
@@ -125,4 +126,67 @@ where
...
+    ///
+    ///     let key2 = 10_u64;
+    ///     let old_value = storage.map.try_insert(key2, new_value);
+    ///     assert(old_value == Option::None); // No old value is returned.
+    ///     let retrieved_value = storage.map.get(key2).read();
+    ///     assert(new_value == retrieved_value); // New value was inserted, as no value existed prior.
+    /// }
+    /// ```
+    #[storage(read, write)]
+    pub fn try_insert(self, key: K, value: V) -> Option<V>
```
Braqzen, line 174: I'm unsure about the need to return the previous value.
Why are we returning anything?
I think it's better to return a result to indicate if we have or have not successfully inserted so the user can decide what to do.
An option doesn't make sense to me here

test/src/sdk-harness/test_projects/storage_map/try_insert.rs
```diff
@@ -0,0 +1,64 @@
+use super::*;
+
+#[macro_export]
+macro_rules! generate_try_insert_tests {
+    ($input_string:expr, $key:expr, $value1:expr, $value2:expr) => {
+        paste::paste! {
```
Braqzen, line 6: I've spent about 2 minutes looking at the paste docs. What functionality does it provide that macros do not?

test/src/sdk-harness/test_projects/storage_map/try_insert.rs
```diff
@@ -0,0 +1,64 @@
...
+
+                let after = instance.methods().[<get_from_ $input_string _map>]($key).call().await.unwrap().value;
+
+                assert_eq!(after, Some($value2));
+            }
+        }
+    };
+}
+
+generate_try_insert_tests!(u64_to_bool, 1, true, false);
```
Braqzen, line 46: How about a fuzzing macro?

Provide the types to the macro and it goes through all combinations without you having to paste them 1 line at a time below, and have some combinations missing

test/src/sdk-harness/test_projects/storage_map/try_insert.rs
SwayStar123: You cannot make up function names using regular macros

test/src/sdk-harness/test_projects/storage_map/try_insert.rs
SwayStar123: Wouldnt that extend the amount of time taken for the test to run by alot?

sway-lib-std/src/storage/storage_map.sw
SwayStar123: Changed it to result and made the return similar to rust

sway-lib-std/src/storage/storage_map.sw
```diff
@@ -125,4 +133,69 @@ where
         let key = sha256((key, self.slot));
         clear::<V>(key, 0)
     }
+
+    /// Inserts a key-value pair into the map, if a value does not already exist for the key.
```
bitzoic, line 137:
```suggestion
    /// Inserts a key-value pair into the map if a value does not already exist for the key.
```

sway-lib-std/src/storage/storage_map.sw
```diff
@@ -125,4 +133,69 @@ where
...
+    ///
+    /// # Number of Storage Accesses
+    ///
+    /// * Reads: `1`
+    /// * Writes: `1`
+    ///
+    /// # Examples
+    ///
+    /// ```sway
+    /// use std::storage::storage_map::StorageMapError;
```
bitzoic, line 156: You have to import the error separately?

sway-lib-std/src/storage/storage_map.sw
```diff
@@ -125,4 +133,69 @@ where
...
+    ///     assert(old_value == Result::Err(StorageMapError::OccupiedError(value))); // The old value is returned.
+    ///     let retrieved_value = storage.map.get(key).read();
+    ///     assert(value == retrieved_value); // New value was not inserted, as a value already existed.
+    ///
+    ///     let key2 = 10_u64;
+    ///     let returned_value = storage.map.try_insert(key2, new_value);
+    ///     assert(returned_value == Result::Ok(new_value)); // New value is returned.
+    ///     let retrieved_value = storage.map.get(key2).read();
+    ///     assert(new_value == retrieved_value); // New value was inserted, as no value existed prior.
+    /// }
```
bitzoic, line 180:
```suggestion
    ///     let key = 5_u64;
    ///     let value = true;
    ///     storage.map.insert(key, value);
    ///
    ///     let new_value = false;
    ///     let result = storage.map.try_insert(key, new_value);
    ///     assert(result == Result::Err(StorageMapError::OccupiedError(value))); // The old value is returned.
    ///
    ///     let retrieved_value = storage.map.get(key).read();
    ///     assert(value == retrieved_value); // New value was not inserted, as a value already existed.
    ///
    ///     let key2 = 10_u64;
    ///     let returned_value = storage.map.try_insert(key2, new_value);
    ///     assert(returned_value == Result::Ok(new_value)); // New value is returned.
    /// }
```
Just shortening this a little

sway-lib-std/src/storage/storage_map.sw
SwayStar123: Yeah since i use it in the assert

test/src/sdk-harness/test_projects/storage_map/try_insert.rs
Braqzen: Probably but we need to ensure that we have coverage.

If a game crashes... it sucks but you don't lose money. If this causes a $$$ loss of funds then rip

bitzoic (approved):

Braqzen (commented):
We ought to test all k, v combinations.

SwayStar123:
> We ought to test all k, v combinations.

That sounds completely unnecessary to me to be honest. Any failure of storing/reading a type would be a failure of the underlying storage api, so such thorough testing here does not make sense.

Braqzen:
> > We ought to test all k, v combinations.
> 
> That sounds completely unnecessary to me to be honest. Any failure of storing/reading a type would be a failure of the underlying storage api, so such thorough testing here does not make sense.

Perhaps, but then the author would not write a macro to generate various tests.
If that is the belief that the author holds then we ought to remove tests that map 1 key to many values and limit it to 1 key per 1 value.

I lean towards being safe than sorry. Additional tests only slow down CI once written, with minor maint' times.

K1-R1 (approved):

K1-R1:
> > > We ought to test all k, v combinations.
> > 
> > 
> > That sounds completely unnecessary to me to be honest. Any failure of storing/reading a type would be a failure of the underlying storage api, so such thorough testing here does not make sense.
> 
> Perhaps, but then the author would not write a macro to generate various tests. If that is the belief that the author holds then we ought to remove tests that map 1 key to many values and limit it to 1 key per 1 value.
> 
> I lean towards being safe than sorry. Additional tests only slow down CI once written, with minor maint' times.

Not against extensive testing but I agree with @SwayStar123 that those failures would be of the storage api; if we were to add extensive K, V combination testing it should be added to the storage api tests

SwayStar123:
> > > We ought to test all k, v combinations.
> > 
> > 
> > That sounds completely unnecessary to me to be honest. Any failure of storing/reading a type would be a failure of the underlying storage api, so such thorough testing here does not make sense.
> 
> Perhaps, but then the author would not write a macro to generate various tests. If that is the belief that the author holds then we ought to remove tests that map 1 key to many values and limit it to 1 key per 1 value.
> 
> I lean towards being safe than sorry. Additional tests only slow down CI once written, with minor maint' times.

I added the tests for the given types to match the pre existing tests. In order to keep it consistent i will not remove these tests.

PR #5443, opened 2024-01-09
Refactor parsed declaration storage

tritao:
## Description

This PR introduces a new `ParsedDeclEngine` which acts as the storage place for parsed declarations. This is basically a clone of `DeclEngine` but only used for untyped parsed declarations.

Declarations stored in this new engine are referenced by a `ParsedDeclId` which again is analogous to our existing `DeclId`.

The rationale for introducing this is to be able to reference parsed declarations during a multi-stage semantic analysis / type-checking process. A cross-referencing mechanism between untyped/typed AST declarations will still need to be implemented.

Just to note, `rustc` provides a similar architecture with their `NodeId` / `DefId`, which provide a similar role to our `ParsedDeclId` / `DeclId`, except we only support storing top-level AST entities, whereas the system in Rust from what I understand can also be used to reference individual expressions.

## Checklist

### DIFF
--- a/sway-core/src/ir_generation/compile.rs
+++ b/sway-core/src/ir_generation/compile.rs
@@ -499,7 +499,7 @@ fn compile_fn(
         logged_types_map,
         messages_types_map,
     );
-    let mut ret_val = compiler.compile_code_block(context, md_mgr, body)?;
+    let mut ret_val = compiler.compile_code_block_to_value(context, md_mgr, body)?;
 
     // Special case: sometimes the returned value at the end of the function block is hacked
     // together and is invalid.  This can happen with diverging control flow or with implicit
--- a/sway-core/src/ir_generation/function.rs
+++ b/sway-core/src/ir_generation/function.rs
@@ -57,6 +57,33 @@ use std::collections::HashMap;
 /// forced, into a value if that is desired. All the temporary values are manipulated with simple
 /// loads and stores, rather than anything more complicated like `mem_copy`s.
 
+// Wrapper around Value to enforce distinction between terminating and non-terminating values.
+struct TerminatorValue {
+    value: Value,
+    is_terminator: bool,
+}
+
+impl TerminatorValue {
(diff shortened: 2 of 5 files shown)


JoshuaBatty:
It's strange the the LSP tests hang and don't even get a chance to complete before the CI is cancelled. Seems like there could be a deadlock occurring somewhere with this change.

JoshuaBatty:
Ok, i'll pull down you branch and try and figure out whats going wrong now.

JoshuaBatty:
Ok yeah looks like it works if i revert to [35212f7](https://github.com/FuelLabs/sway/pull/5443/commits/35212f74dec504b83311101ed16ddb604e906b63)

We probably still want to be able to perform GC on the parsed_decl_engine from LSP just like we do with the [decl](https://github.com/FuelLabs/sway/blob/c43b57fd3ddafe88c6b5ef5abf1057b05c3e1d09/sway-core/src/decl_engine/engine.rs#L186) and type engines.

JoshuaBatty:
> My first initial try was with the same approach you used for the other engines (which uses `retain`) but it gave significant runtime slowdowns (maybe 10-15%). So i changed to use `clear` instead of `retain` which is faster.

Is there any reason why clear needs to be called from the compiler though? So far the other engines clear is only called from LSP.

JoshuaBatty:
Ok, not sure how to make a PR against your forked repo, but [I made some changes](https://github.com/FuelLabs/sway/commit/d59d06e1178086c1bd53cd2b2582e5fd936c166e) to implement `clear_module` for the `ParsedDeclEngine`.

JoshuaBatty (approved):
Nice work, changes seem good to me. I think I will most likely change `pub lsp_mode: bool` to be something like `pub lsp_mode: Option<LspOptions>` in an upcoming PR. Then we can just check if this is `Some` or not. But this is fine how it is currently.

IGI-111 (approved):

PR #5446, opened 2024-01-10
refactor: update error message for `forc deploy` when password is mismatch

### LINKED ISSUE #5309: `forc deploy` should have a better error message for invalid password
Currently if you enter the wrong password, this is the error message:

image

We should change it to something like "The password provided was incorrect."

Halimao:
Close #5309


kayagokalp (approved):
utACK

Issue #5434, opened 2024-01-04
`forc` panics if the Sway package folder contains dot (`.`)


ironcev:
Steps to reproduce:
```
$ mkdir path_with_._dot
$ cd path_with_._dot
path_with_._dot$ forc init --script
path_with_._dot$ forc build
```
Build will panic with the followinig message:
```
thread 'main' panicked at 'file_name must be either absolute or relative to manifest directory', sway-core/src/build_config.rs:74:17
```
After renaming the folder to a name without dot, the build will pass:
```
path_with_._dot$ cd ..
$ mv path_with_._dot path_without_dot
$ cd path_without_dot
path_without_dot$ forc build
```

PR #5455, opened 2024-01-11
refactor: `forc init` generate unexpected project_name when project dir contains dot

Halimao:
## Description
Refer #5434

Found this issue when I tried to run `forc init --script` under a directory(`/tmp/path_with_._dot`) that contains a dot ".". The value of `project.name` in the generated file `Forc.toml` is truncated by the dot "." as `path_with_`.

Since `project_dir` must be a directory(has checked before), we can use `file_name` instead of `file_stem` to get the last-level dir name. Also, I think it would be better to replace "." with "_", otherwise a `project_name` that contains `.` would fail with validation `validate_name(&project_name, "project name")?;`

#### Before:
image

#### After:
image

## Checklist


kayagokalp (changes requested):
Thanks for the PR! I will test this locally but it seems like this will allow project with `.` in the name, which is not something we should be doing imo. Blocking for now so that I can test

Halimao:
> Thanks for the PR! I will test this locally but it seems like this will allow project with `.` in the name, which is not something we should be doing imo. Blocking for now so that I can test

The `.` will be replaced with `_`. This pr just changed the value of `project_name` by using `path_with___dot` instead of `path_with_`. See the screenshot below:

image

sdankel:
I think we should disallow project names with dots or any special character other than `_`, similar to cargo. Numbers should be allowed but not as a first digit.

Here's a regex we can use: `^([a-zA-Z]([a-zA-Z0-9-_]+)|)$`

Halimao:
> I think we should disallow project names with dots or any special character other than `_`, similar to cargo. Numbers should be allowed but not as a first digit.
> 
> Here's a regex we can use: `^([a-zA-Z]([a-zA-Z0-9-_]+)|)$`

If I understood correctly, the suggested changes are to revert the logic of getting project_name and add a new regex `^([a-zA-Z]([a-zA-Z0-9-_]+)|)$` checker into the function `validate_name`?

sdankel:
> > I think we should disallow project names with dots or any special character other than `_`, similar to cargo. Numbers should be allowed but not as a first digit.
> > Here's a regex we can use: `^([a-zA-Z]([a-zA-Z0-9-_]+)|)$`
> 
> If I understood correctly, the suggested changes are to revert the logic of getting project_name and add a new regex `^([a-zA-Z]([a-zA-Z0-9-_]+)|)$` checker into the function `validate_name`?

Yes, I think that's the best path forward. If it doesn't match the regex, we can show an error like "The project name can be a combination of letters, numbers and underscores, and must start with a letter."

Halimao:
> Yes, I think that's the best path forward. If it doesn't match the regex, we can show an error like "The project name can be a combination of letters, numbers, and underscores, and must start with a letter."

image

1. The function `validate_name` is not only used for `project name` but also for `organization name` and `package name`
2. The param `use_case` is not an enum, there may exist different values for the similar case `project name`, and it's hard to judge whether the case is `project name` or not

For the reason above, I think we should add a new function `validate_project_name` like this:
```rust
pub fn validate_project_name(name: &str) -> Result<()> {
    let re = Regex::new(r"^([a-zA-Z]([a-zA-Z0-9-_]+)|)$").unwrap();
    if !re.is_match(name) {
        bail!(
            "the project name `{name}` cannot be used as a project name.\n\
        project name can be a combination of letters, numbers, and underscores, \
        and must start with a letter."
        );
    }
    validate_name(name, "project name")
}
```

forc-util/src/restricted.rs
```diff
@@ -130,3 +144,42 @@ fn test_invalid_char() {
         std::result::Result::Ok(())
     ));
 }
+
+#[test]
```
JoshuaBatty, line 148: Perhaps something like below is a bit cleaner?

```rust
#[test]
fn test_is_valid_project_name_format() {
    let assert_valid = |name: &str| {
        is_valid_project_name_format(name).expect("this should pass");
    };

    let assert_invalid = |name: &str, expected_error: &str| {
        assert_eq!(
            is_valid_project_name_format(name).map_err(|e| e.to_string()),
            Err(expected_error.into())
        );
    };

    // Test valid project names
    assert_valid("mock_project_name");
    assert_valid("mock_project_name123");
    assert_valid("mock_project_name-123-_");

    // Test invalid project names
    let error_message = "the project name `{}` cannot be used as a project name.\n\
        project name can be a combination of letters, numbers, hyphen, and underscores, \
        and must start with a letter.";

    assert_invalid("1mock_project", &format!(error_message, "1mock_project"));
    assert_invalid("mock_.project", &format!(error_message, "mock_.project"));
    assert_invalid("mock_/project", &format!(error_message, "mock_/project"));
}
```

Halimao: @JoshuaBatty Hi sir, I think `&format!(error_message, "1mock_project")` would issue an error "format argument must be a string literal".

I wrapped the format logic into a function
```rust
    let format_error_message = |name: &str| -> String {
        let formatted_err_str = format!(
            "the project name `{}` cannot be used as a project name.\n\
        project name can be a combination of letters, numbers, hyphen, and underscores, \
        and must start with a letter.",
            name
        );
        formatted_err_str
    };
```

sdankel:
> BTW, is the '-' hyphen valid in the project name?

Yes, good catch! `-` should be allowed.

> The function validate_name is not only used for project name but also for organization name and package name

I don't think there is a difference between package name and project name. I agree that organization name should not have this validation.

forc-util/src/restricted.rs
```diff
@@ -94,6 +95,19 @@ pub fn is_glob_pattern<T: AsRef<str>>(name: T) -> bool {
...
 }
 
+/// Check the project name format.
+pub fn is_valid_project_name_format(name: &str) -> Result<()> {
+    let re = Regex::new(r"^([a-zA-Z]([a-zA-Z0-9-_]+)|)$").unwrap();
+    if !re.is_match(name) {
+        bail!(
+            "the project name `{name}` cannot be used as a project name.\n\
+        project name can be a combination of letters, numbers, hyphen, and underscores, \
+        and must start with a letter."
```
sdankel, line 105:
```suggestion
            "'{name}' is not a valid name for a project. \nThe name may use letters, numbers, hyphens, and underscores, and must start with a letter."
```

forc/src/ops/forc_init.rs
```diff
@@ -78,7 +78,7 @@ pub fn init(command: InitCommand) -> ForcResult<()> {
             .into_owned(),
     };
 
-    validate_name(&project_name, "project name")?;
```
sdankel, line 81: We should also update this in `new.rs`, `forc_template.rs`, and `manifest/mod.rs:553`

Halimao:
> > BTW, is the '-' hyphen valid in the project name?
> 
> Yes, good catch! `-` should be allowed.
> 
> > The function validate_name is not only used for project name but also for organization name and package name
> 
> I don't think there is a difference between package name and project name. I agree that organization name should not have this validation.

Suggestion changes done~

forc-pkg/src/manifest/mod.rs
```diff
@@ -550,7 +550,7 @@ impl PackageManifest {
     /// 2. The validity of the details provided. Makes sure that there are no mismatching detail
     ///    declarations (to prevent mixing details specific to certain types).
     pub fn validate(&self) -> Result<()> {
-        validate_name(&self.project.name, "package name")?;
+        validate_project_and_package_name(&self.project.name, "package name")?;
```
sdankel, line 553: > I don't think there is a difference between package name and project name. I agree that organization name should not have this validation.

What I mean is that this is still checking `self.project.name`. There's no need to pass in `use_case` as a string. We can just use "project name" as the string. The function can be renamed to `validate_project_name`

forc-util/src/lib.rs
```diff
@@ -214,6 +214,11 @@ pub fn lock_path(manifest_dir: &Path) -> PathBuf {
     manifest_dir.join(constants::LOCK_FILE_NAME)
 }
 
+pub fn validate_project_and_package_name(name: &str, use_case: &str) -> Result<()> {
+    restricted::is_valid_project_and_package_name_format(name)?;
+    validate_name(name, use_case)
+}
```
sdankel, line 220:
```suggestion
pub fn validate_project_name(name: &str) -> Result<()> {
    restricted::is_valid_project_and_package_name_format(name)?;
    validate_name(name, "project name")
}
```

forc-pkg/src/manifest/mod.rs
Halimao: @sdankel Done~

sdankel (approved):
Looks good, thanks!

Halimao:
Can we get this merged now? Is there anything else that needs to be added?

JoshuaBatty (approved):

PR #5477, opened 2024-01-16
Sophie/sourcemaps

sdankel:
## Checklist


forc-plugins/forc-debug/src/server/handlers/handle_launch.rs
```diff
@@ -0,0 +1,216 @@
...
+                    test_entry,
+                    name.clone(),
+                ))
+            })
+            .collect();
+        self.state.init_executors(executors);
+
+        // Start debugging
+        self.state.update_vm_breakpoints();
+        while let Some(executor) = self.state.executors.first_mut() {
```
JoshuaBatty, line 45: This looks similar to the logic above in the `handle_continue` function. Can it be abstracted away and called from both places?

forc-plugins/forc-debug/src/server/handlers/handle_launch.rs
```diff
@@ -0,0 +1,216 @@
...
+        )
+        .map_err(|err| AdapterError::BuildFailed {
+            phase: "build plan".into(),
+            reason: err.to_string(),
+        })?;
+
+        let project_name = pkg_manifest.project_name();
+
+        let outputs =
+            std::iter::once(build_plan.find_member_index(project_name).unwrap()).collect();
```
JoshuaBatty, line 113: is it possible to return an error instead of unwraping here? Or at least have it be an expect?

forc-plugins/forc-debug/src/server/handlers/handle_launch.rs
```diff
@@ -0,0 +1,216 @@
...
+                    if let Ok(source) = fs::read_to_string(path_buf) {
+                        Some((path_buf, source))
+                    } else {
+                        None
+                    }
+                })
+                .collect::<HashMap<_, _>>();
+
+            source_map.map.iter().for_each(|(instruction, sm_span)| {
+                let path_buf: &PathBuf = paths.get(sm_span.path.0).unwrap();
```
JoshuaBatty, line 152: Can we handle the unwrap here as well?

forc-plugins/forc-debug/src/server/handlers/handle_launch.rs
```diff
@@ -0,0 +1,216 @@
...
+                })
+            }
+        };
+        let test_setup = pkg_tests.setup().map_err(|err| AdapterError::BuildFailed {
+            phase: "test setup".into(),
+            reason: err.to_string(),
+        })?;
+        self.state.built_package = Some(built_package.clone());
+        self.state.test_setup = Some(test_setup.clone());
+        Ok((built_package.clone(), test_setup))
```
JoshuaBatty, line 214: Not sure but I think this clone isn't necessary?

forc-plugins/forc-debug/src/server/handlers/handle_next.rs
```diff
@@ -0,0 +1,30 @@
+use crate::server::AdapterError;
+use crate::server::DapServer;
+use forc_test::execute::DebugResult;
+
+impl DapServer {
+    /// Handles a `next` request. Returns true if the server should continue running.
+    pub(crate) fn handle_next(&mut self) -> Result<bool, AdapterError> {
+        self.state.update_vm_breakpoints();
+        if let Some(executor) = self.state.executors.first_mut() {
```
JoshuaBatty, line 9: This also looks similar to the above code

forc-plugins/forc-debug/src/server/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+pub const THREAD_ID: i64 = 0;
+pub const REGISTERS_VARIABLE_REF: i64 = 1;
+pub const INSTRUCTIONS_VARIABLE_REF: i64 = 2;
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct AdditionalData {
+    pub program: String,
+}
+
+pub struct DapServer {
```
JoshuaBatty, line 31: Can we add a doc comment?

forc-plugins/forc-debug/src/server/state.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    /// Returns the active [TestExecutor], if any.
+    pub fn executor(&mut self) -> Option<&mut TestExecutor> {
+        self.executors.first_mut()
+    }
+
+    /// Finds the source location matching a VM program counter.
+    pub fn vm_pc_to_source_location(&self, pc: u64) -> Result<(&PathBuf, i64), AdapterError> {
+        // FTry to find the source location by looking for the program counter in the source map.
```
JoshuaBatty, line 57:
```suggestion
        // Try to find the source location by looking for the program counter in the source map.
```

forc-plugins/forc-debug/src/server/util.rs
```diff
@@ -0,0 +1,39 @@
+use dap::types::Source;
+use fuel_vm::fuel_asm::RegId;
+use std::path::Path;
+
+#[derive(Debug, Clone)]
+pub(crate) struct IdGenerator {
```
JoshuaBatty, line 6: Might also be good to add a doc comment explaining what an IdGenerator is and why it's needed

forc-test/src/lib.rs
```diff
@@ -331,7 +355,7 @@ impl<'a> PackageTests {
             pkg_with_tests
                 .bytecode
                 .entries
-                .par_iter()
```
JoshuaBatty, line 334: any reason why this was changed?

JoshuaBatty:
Nice one, looking really good. Just a couple of nits.

forc-test/src/lib.rs
sdankel: Good catch, reverted

forc-plugins/forc-debug/src/server/handlers/handle_next.rs
sdankel: Refactored ✅

JoshuaBatty:
Have just been playing around locally with this and it feels really great! 

Just had a thought, would it be possible to print a message to the debug console saying "this feature is currently not supported" or similar if the user clicks on either step into or step out?

forc-plugins/forc-debug/tests/fixtures/simple/Forc.toml
```diff
@@ -0,0 +1,9 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "simple"
+implicit-std = false
+
+[dependencies]
+std = { git = "https://github.com/FuelLabs/sway", tag = "v0.48.1" }
```
sdankel, line 9:
```suggestion
std = { git = "https://github.com/FuelLabs/sway", tag = "v0.50.0" }
```

forc-plugins/forc-debug/src/server/handlers/handle_set_breakpoints.rs
```diff
@@ -0,0 +1,181 @@
...
+use dap::types::{Breakpoint, StartDebuggingRequestKind};
+use std::path::PathBuf;
+
+impl DapServer {
+    /// Handles a `set_breakpoints` request. Returns the list of [Breakpoint]s for the path provided in `args`.
+    pub(crate) fn handle_set_breakpoints(
+        &mut self,
+        args: &SetBreakpointsArguments,
+    ) -> Result<Vec<Breakpoint>, AdapterError> {
+        // Build the source maps so we can verify breakpoints
```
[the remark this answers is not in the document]
sdankel: > I am not sure about this, as a user I would not expect a breakpoint set to build anything.
It will only build once, and this will only happen after the user has started a debugging session. The user will only see the unverified breakpoints (greyed out) when in a debugging session.

> the breakpoints should be pending until it attaches to / launches the target.
This would work too, but I think the UX is a little better when building the source maps at this point, since they won't flash grey then red again.

forc-plugins/forc-debug/src/server/handlers/handle_launch.rs
```diff
@@ -0,0 +1,202 @@
...
+use forc_test::execute::TestExecutor;
+use forc_test::setup::TestSetup;
+use forc_test::BuiltTests;
+use std::{collections::HashMap, fs, sync::Arc};
+use sway_types::span::Position;
+
+impl DapServer {
+    /// Handles a `launch` request. Returns true if the server should continue running.
+    pub fn handle_launch(&mut self) -> Result<bool, AdapterError> {
+        // Build tests for the given path.
```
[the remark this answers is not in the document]
sdankel: > In my experience, the launch command of a debugger should not be building anything, it should only launch the actual previously built executable.
Agreed, that's my experience too. However, our case requires that the debugger builds the code in order to have access to the VM interpreter to set breakpoints. It also conveniently gives us access to the source maps object, and as you said, makes for a more seamless UX.

forc-plugins/forc-debug/src/server/handlers/handle_stack_trace.rs
```diff
@@ -0,0 +1,55 @@
...
+            .first()
+            .ok_or(AdapterError::NoActiveTestExecutor)?;
+        let name = executor.name.clone();
+
+        let source_location = match self.state.stopped_on_breakpoint_id {
+            // If we stopped on a breakpoint, use the breakpoint's source location.
+            Some(breakpoint_id) => self.state.breakpoints.iter().find_map(|(_, breakpoints)| {
+                breakpoints.iter().find_map(|bp| {
+                    if Some(breakpoint_id) == bp.id {
+                        if let Some(bp_line) = bp.line {
```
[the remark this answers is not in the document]
sdankel: I think it's a bit more readable with the `if let Some`s here

forc-pkg/src/manifest.rs
```diff
@@ -127,6 +128,32 @@ impl ManifestFile {
...
+    fn try_into(self) -> Result<WorkspaceManifestFile> {
+        match self {
+            ManifestFile::Package(_) => {
+                bail!("Cannot convert package manifest to workspace manifest")
+            }
+            ManifestFile::Workspace(workspace_manifest_file) => Ok(workspace_manifest_file),
+        }
+    }
+}
+
```
kayagokalp, line 156: I am a little unsure if we should have `TryInto` impls for this type of operations (converting an enum into its variants). In the compiler codebase we have `as_workspace()` etc for these type of conversions. 

Typically a `TryInto` is for converting a type into another one so it might be a little out of convention to use it for converting an enum to one of its variants.

forc-plugins/forc-debug/src/server/handlers/handle_launch.rs
```diff
@@ -0,0 +1,202 @@
...
+                .try_into()
+                .map_err(|err: anyhow::Error| AdapterError::BuildFailed {
+                    reason: format!("package manifest: {:?}", err),
+                })?;
+        let member_manifests =
+            manifest_file
+                .member_manifests()
+                .map_err(|err| AdapterError::BuildFailed {
+                    reason: format!("member manifests: {:?}", err),
+                })?;
```
kayagokalp, line 72: I haven't looked into much detail in this part but if we already obtain a reference to the `member_manifests` to be able to build, we can also get the package manifest from that with the name of the package without needing to clone manifest file to cast into a package manifest file. `Assuming that we have the project name in place somewhere`.

```rust
let pkg_name = ...;
let pkg_manifest_file = &member_manifests[pkg_name];
```

forc-plugins/forc-debug/src/server/handlers/handle_launch.rs
```diff
@@ -0,0 +1,202 @@
...
+            })
+            .collect();
+        self.state.init_executors(executors);
+
+        // Start debugging
+        self.start_debugging_tests(false)
+    }
+
+    /// Builds the tests at the given [PathBuf] and stores the source maps.
+    pub(crate) fn build_tests(&mut self) -> Result<(BuiltPackage, TestSetup), AdapterError> {
```
kayagokalp, line 48: It is interesting to see that we need this much code to build something with tests enabled 🤔 

Maybe we need to refactor forc-pkg a little sometime to provide easier to use APIs for such cases.

kayagokalp: Also we should visit this function some time later because cloning `BuiltPackage` is a costly event that might lead to huge memory usage stuff that we encountered before. I think we should be fine for now but maybe in a separate PR that refactors `forc-pkg` as mentioned above, we can try to remove all/some clone's on built package's across the repo.

kayagokalp (dismissed):
Huge work! Very excited for this 🚀 I left some nits and comments here and there but we can merge it as it is and discuss next steps in separate issues/PRs :)

forc-plugins/forc-debug/src/server/handlers/handle_launch.rs
sdankel: Great call out. I think the interface of forc-pkg can definitely be improved, but I didn't want to bloat this PR too much. Opened https://github.com/FuelLabs/sway/issues/5620

forc-pkg/src/manifest.rs
sdankel: Does this look better? https://github.com/FuelLabs/sway/pull/5625

forc-plugins/forc-debug/src/server/handlers/handle_launch.rs
sdankel: Addressed in https://github.com/FuelLabs/sway/pull/5625

JoshuaBatty (approved):
Great work Sophie :)

kayagokalp (approved):

Issue #5499, opened 2024-01-21
Incomplete code snippets in compiler output


ironcev:
When displaying additional code snippets in the compiler output (those are hints, and notes that are coming from other parts of the same file or from other files), the snippets are sometimes missing the first line and showing only the last one.

Here is a reproducible example. Create a new script with the following library in the file called _lib.sw_:

```Sway
library;

pub
    const
        LIB_X = 3;
```

Import the `LIB_X` and shadow it in the _main.sw_:

```Sway
script;

mod lib;

fn main() {
    let LIB_X_ALIAS = 7;
}

use lib::LIB_X as LIB_X_ALIAS;
```

The compiler error will look like this. Note that the _info_ snippet is missing the opening `pub` line.

![Multiline issue](https://github.com/FuelLabs/sway/assets/4142833/9c63be71-4709-4cc4-9eeb-5f7ebd230cbf)

A quick experimenting indicates that the issue is there only with the small number of lines of code (three in this example). If the snippet is larger in terms of lines of code, the first line gets properly rendered.

PR #5508, opened 2024-01-23
Struct field privacy

### LINKED ISSUE #4692: Struct field privacy
All struct fields are currently public, we should introduce privacy specifiable with `pub` and make fields private by default.

ironcev:
## Description

This PR makes struct fields private by default and introduces explicit public struct fields:

```
pub struct Struct {
    pub public_field: u8,
    private_field: u8,    
}
```

Private fields can be accessed only withing the module in which their struct is declared.

Error messages are properly orchestrated so that no conflicting or duplicated messages are emitted. Since the change is a breaking change, relevant suggestion on how to fix the particular issue are given.

To avoid an abrupt breaking change, the errors are temporarily turned into warnings. These warnings will become errors in the upcoming versions of Sway. The demo section below demonstrate how the errors will look like, and how a one warning equivalent looks now.

Standard library structs like `Vec` or `String`, are adapted where needed by adding the `pub` keyword to the fields that are accessed outside of the struct declaration module. Some of these fields should very likely remain public, but some, like e.g., `RawBytes::ptr` or `Vec::buf` should be private. Adjusting the standard library to properly utilize private fields is out of scope of this PR and will be done separately. I expect breaking changes in the STD once we start modeling structs keeping encapsulation in mind.

In addition, the PR:
- migrates [annotate_snippets](https://github.com/rust-lang/annotate-snippets-rs) crate to its latest version (breaking changes)
- removes a redundant non-existing field error and orchestrates non-existing field error with privacy errors
- replaces an invalid and misleading error when accessing fields on storage elements that are not structs

Closes #4692.

## Known limitations

There is an issue in displaying multi-line code snippets, which is independent of these changes, but wasn't apparent before. The issue is visible in the demo section below, where the struct bodies are sometimes not fully shown, and they should be. This issue is reported in #5499 and will be solved in a separate PR.

### Private struct field is inaccessible

Private struct field is inaccessible

Private struct field is inaccessible

### Struct cannot be instantiated

Struct cannot be instantiated due to inaccessible private fields

Struct cannot be instantiated due to inaccessible private fields

### Struct pattern must ignore inaccessible private fields

Struct pattern must ignore inaccessible private fields

### Struct pattern has missing fields

Struct pattern has missing fields

### Errors temporarily turned into warnings

Errors temporarily turned into warnings

## Checklist


ironcev:
Two steps still missing:
- implement support for `pub` in `swayfmt` (this [todo](https://github.com/FuelLabs/sway/blob/9987e86de2280d0207c0e08f47835372b3be8874/swayfmt/src/items/item_struct/mod.rs#L22)).
- update documentation

All remaining parts are fully implemented and can be reviewed.

docs/reference/src/documentation/language/functions/associated-function.md
```diff
@@ -4,6 +4,10 @@ Associated functions are similar to methods in that they are also defined in the
 
 Associated functions could be standalone functions, but they are included in a specific type for organizational or semantic reasons.
 
+## Constructors
+
+A distinguished family of associated functions of a specific type are _type constructors_. Constructors are associated functions that construct, or in other words instantiate, new instances of a type. Their return type is always the type itself. E.g., public [structs](../built-ins/structs.md) that have private fields must provide a public constructor, or otherwise they cannot be instantiated outside of the module in which they are declared.
```
IGI-111, line 9:
```suggestion
A distinguished family of associated functions of a specific type are _type constructors_. Constructors are associated functions that construct, or in other words instantiate, new instances of a type. Their return type always includes the type itself. E.g., public [structs](../built-ins/structs.md) that have private fields must provide a public constructor, or otherwise they cannot be instantiated outside of the module in which they are declared.
```

sway-core/src/semantic_analysis/ast_node/expression/match_expression/typed/typed_scrutinee.rs
```diff
@@ -233,55 +235,228 @@ fn type_check_struct(
...
+                    StructScrutineeField::Field {
+                        field: ref field_name,
+                        ..
+                    } => {
+                        let struct_field = struct_decl
+                            .find_field(field_name)
+                            .expect("The struct field with the given field name must exist.");
+
+                        if struct_field.is_private() {
+                            // TODO: Uncomment this code and delete the one with warnings once struct field privacy becomes a hard error.
```
IGI-111, line 311: We should create an issue for changing this and mention the issue number in the relevant comments.

sway-error/src/diagnostic.rs
```diff
@@ -70,8 +70,24 @@ impl Diagnostic {
         self.labels_in_source(self.issue.source_path().unwrap())
     }
 
+    const HELP_NONE: &'static str = "__$$_^^__HELP__NONE__^^_$$__";
```
IGI-111, line 73: I understand the need for something, but this is hacky. Couldn't you make this work with `Option` or empty strings if you can't rework the logic to simply do nothing?

sway-error/src/diagnostic.rs
```diff
@@ -70,8 +70,24 @@ impl Diagnostic {
...
-        self.help.iter().filter(|help| !help.is_empty())
+        self.help
+            .iter()
+            .filter(|help| help.as_str() != Self::HELP_NONE)
+    }
+
+    /// A help text that will never be displayed. Convenient when defining help lines
+    /// that are displayed only when a condition is met.
+    pub fn help_none() -> String {
+        String::from(Self::HELP_NONE)
```
IGI-111, line 84: Just use `String::new` at that point, the constant isn't necessary.

sway-error/src/diagnostic.rs
```diff
@@ -70,8 +70,24 @@ impl Diagnostic {
         self.labels_in_source(self.issue.source_path().unwrap())
     }
 
+    const HELP_NONE: &'static str = "";
+
     pub fn help(&self) -> impl Iterator<Item = &String> + '_ {
-        self.help.iter().filter(|help| !help.is_empty())
+        self.help
+            .iter()
+            .filter(|help| help.as_str() != Self::HELP_NONE)
```
IGI-111, line 78:
```suggestion
            .filter(|help| !help.is_empty())
```

swayfmt/tests/mod.rs
```diff
@@ -79,6 +79,69 @@ pub struct Foo<T, P> {
...
+    check_with_formatter(
+        r#"contract;
+pub struct Foo<T, P> {
+   barbazfoo: u64,
+   pub baz     : bool,
+}
+"#,
+        r#"contract;
+pub struct Foo<T, P> {
+    barbazfoo : u64,
```
JoshuaBatty, line 97: I feel like it should be formatted like this as it would be in rust.

```suggestion
pub struct Foo<T, P> {
    barbazfoo: u64,
    pub baz: bool
}
```

swayfmt/tests/mod.rs
```diff
@@ -79,6 +79,69 @@ pub struct Foo<T, P> {
...
+
+    check_with_formatter(
+        r#"contract;
+pub struct Foo<T, P> {
+   pub barbazfoo: u64,
+   baz     : bool,
+}
+"#,
+        r#"contract;
+pub struct Foo<T, P> {
```
JoshuaBatty, line 112: Same here.
```suggestion
pub struct Foo<T, P> {
    pub barbazfoo: u64,
    baz: bool,
}
```

swayfmt/tests/mod.rs
ironcev: In the case of `FieldAlignment::AlignFields` there is a space between the field name and the `:`. This is clear from the implementation, in _item_struct/mod.rs_, line 116:
```
// Add `:`, ty & `CommaToken`
write!(
    formatted_code,
    " {} ",
    type_field.colon_token.span().as_str(),
)?;
```
This formatting is also expected in the `struct_alignment` test. Adding the `pub` keyword just followed the existing expectations in case of  `FieldAlignment::AlignFields`.

swayfmt/tests/mod.rs
ironcev: See above.

IGI-111 (dismissed):

sway-error/src/warning.rs
```diff
@@ -121,6 +126,35 @@ pub enum Warning {
     UsingDeprecated {
         message: String,
     },
+    // TODO: Remove all these warnings once private struct fields violations become a hard error.
```
sdankel, line 129:
```suggestion
    // TODO: Remove all these warnings once private struct fields violations become a hard error. https://github.com/FuelLabs/sway/issues/5520
```

swayfmt/src/items/item_struct/mod.rs
```diff
@@ -69,9 +69,17 @@ impl Format for ItemStruct {
...
+                        // We need to include the `pub` keyword in the length, if the field is public,
+                        // together with one space character between the `pub` and the name.
                         let variant_length: Vec<usize> = value_pairs
                             .iter()
-                            .map(|(type_field, _)| type_field.name.as_str().len())
+                            .map(|(type_field, _)| {
+                                type_field
+                                    .visibility
+                                    .as_ref()
+                                    .map_or(0, |_| PubToken::AS_STR.len() + 1)
```
sdankel, line 80: Just thinking if we ever had other types of visibility tokens, we shouldn't hardcode this.

```suggestion
                                    .map_or(0, |token| token.span().as_str().len() + 1)
```

sway-ast/src/item/mod.rs
```diff
@@ -66,14 +66,20 @@ impl Spanned for ItemKind {
 
 #[derive(Clone, Debug, Serialize)]
 pub struct TypeField {
+    pub visibility: Option<PubToken>,
```
sdankel, line 69: Is it likely we'll add other visibility tokens in the future, similar to `pub(crate)`?

sdankel (approved):
Awesome work! I left some questions/nits, but nothing blocking.

docs/reference/src/code/language/built-ins/structs/src/lib.sw
```diff
@@ -31,6 +31,18 @@ fn variable_instantiation() {
...
+fn shorthand_instantiation() {
+    // Declare variables with the same names as the fields in `Foo`
+    let bar = 42;
+    let baz = false;
+
+    // Instantiate `foo` as `Foo`
+    let mut foo = Foo { bar, baz };
+
+    // Access and write to "baz"
+    foo.baz = true;
```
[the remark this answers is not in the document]
ironcev: In this case the usage is in the same module _lib.sw_ where the struct `Foo` is declared. So both fields can be accessed.

test/src/e2e_vm_tests/test_programs/should_pass/require_contract_deployment/caller_context_test/src/main.sw
```diff
@@ -39,7 +38,7 @@ fn main() -> bool {
     }(base_asset_id, other_contract_id);
     assert(returned_contract_balance == 0);
 
-    // The checks below don't work (AssertIdNotFound). The test should be
+    // TODO: The checks below don't work (AssertIdNotFound). The test should be
```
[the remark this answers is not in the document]
ironcev: Done: https://github.com/FuelLabs/sway/issues/5527

sway-error/src/warning.rs
```diff
@@ -264,10 +298,25 @@ impl fmt::Display for Warning {
...
+            StructCannotBeInstantiated { struct_name, ..} => write!(f,
+                "Struct \"{struct_name}\" cannot be instantiated here because it has private fields."
+            ),
         }
     }
 }
 
+#[allow(dead_code)]
+const FUTURE_HARD_ERROR_HELP: &str =
+    "*** In the upcoming versions of Sway this warning will become a hard error. ***";
```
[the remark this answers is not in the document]
ironcev: Looks much better and stands out more, indeed :smile: 

I don't see a need for integration tests here. It is only three warnings and we expect them to be turned into errors soon.

sway-core/src/language/parsed/expression/mod.rs
```diff
@@ -321,7 +321,6 @@ pub enum ReassignmentTarget {
 pub struct StructExpressionField {
     pub name: Ident,
     pub value: Expression,
-    pub(crate) span: Span,
```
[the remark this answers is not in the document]
ironcev: Good question! That was the span of the `field_name: <initialization>` used only in the error message shown in struct instantiation that the struct field does not exist. That error is changed to point only at the field name, since having the non-existing field there is actually the issue, and not the `<initialization>` which on it's own does not need to be erroneous at all. This focus on the field as the source of the issue is even more important in the cases where the `<initialization>` spans several lines like in:
```Sway
x: Struct {
  a: 123,
  b: 345,
  ...
}
```
In that case the error span was containing a lot of noise.

I've checked Rust errors and they also mostly focus on the field only, but it's not fully consistent.
Also, this approach, to point to the actual issue I see as a general guidance, also when formulating the error message text.

Long story short, once that error span was changed, it turned out the `pub(crate) span: Span` wasn't used anywhere else. It felt odd to delete it, I must admit. I thought at first marking it with `#[allow(dead_code)]` but in the end opted for removal. If we ever need it, it will be totally trivial to get it back.

sway-ast/src/item/mod.rs
ironcev: This might be the case. If it happens, we will need to change the way we define visibility everywhere, starting from the `TypeField` in the AST.

IGI-111 (approved):

PR #5510, opened 2024-01-24
Explicitly use `u64` operations for safe `u8` arithmetic

### LINKED ISSUE #5449: Overflow behaviour relies on Optimization
While implementing #5385, I found that when optimizations are disabled, (in particular, `constcombine` and/or `mem2reg` for this bug), the IR we're generating for addition doesn't check for overflows correctly.

This is the IR for `should_fail/arith_overflow/u8_add_overflow` (note that there's no `mem2reg` or `constcombine`):

```
script {
    entry fn main() -> bool, !1 {
        local u8 a
        local u8 b
        local u8 res
        local u8 result

        entry():
        v0 = get_local ptr u8, a, !2
        v1 = const u8 255, !4
        store v1 to v0, !2
        v2 = get_local ptr u8, b, !5
        v3 = const u8 1, !6
        store v3 to v2, !5
        v4 = get_local ptr u8, a, !7
        v5 = load v4
        v6 = get_local ptr u8, b, !8
        v7 = load v6
        v8 = add v5, v7, !9
        v9 = get_local ptr u8, res, !12
        store v8 to v9, !13
        v10 = get_local ptr u8, res, !15
        v11 = load v10, !9
        v12 = const u8 255, !4
        v13 = cmp gt v11 v12, !9
        cbr v13, add_1_block0(), add_1_block1(), !17

        add_1_block0():
        v14 = const u64 0, !18
        revert v14, !20

        add_1_block1():
        ...
    }
   ...
```

Because we're using byte stores and loads for `u8`, we end up ignoring the overflowed part in the u64 register, thus not triggering the revert.

When this issue is fixed, we should be enabling the tests `should_pass/arith_overflow/u8_add_overflow` and `should_pass/arith_overflow/u8_mul_overflow` which were both disabled in #5385.

I suspect the bug got introduced with #4929.

vaivaswatha:
Fixes #5449

Without this fix, because we use `u8` load and store instructions, the overflow check (which relies on `u64` arithmetic) gets nullified.


IGI-111:
This doesn't seem like a general purpose fix, shouldn't we be able to have a `u8` specific overflow check?

I'm concerned there may be cases in user code where they use `u8` legally and this problem still shows up. Is that possible still?

vaivaswatha:
> This doesn't seem like a general purpose fix, shouldn't we be able to have a `u8` specific overflow check?
> 
> I'm concerned there may be cases in user code where they use `u8` legally and this problem still shows up. Is that possible still?

This *is* a u8 specific overflow check. It impacts only u8 arithmetic. If someone is using u8 legally, then neither the bug or the fix impacts. It's only when there's an overflow that, because we use byte load and stores, the overflow checks were made ineffective (without optimization). With this, we ensure that the overflow check indeed uses u64 arithmetic, just like how that is ensured automatically for u16 and u32.

IGI-111:
Okay I get it. This is probably good enough for now, but @vaivaswatha, how do you feel about, instead of doing this casting dance, extending the intrinsics to signal overflow based on input type?

vaivaswatha:
> Okay I get it. This is probably good enough for now, but @vaivaswatha, how do you feel about, instead of doing this casting dance, extending the intrinsics to signal overflow based on input type?

Then the compiler will have to generate the same code that we're manually writing here in Sway. The current approach is right imo. It's more maintainable, and gives the user an option to directly use the intrinsic if they don't want overflow checks.

xunilrj:
The `__add` intrinsic returns `Numeric`, if I am not mistaken... so we only needs to do the below, right?

```sway
let res: u64 = __add(self, other);
```

vaivaswatha:
> The `__add` intrinsic returns `Numeric`, if I am not mistaken... so we only needs to do the below, right?
> 
> ```sway
> let res: u64 = __add(self, other);
> ```

No, the type checker ensures that `_add`'s operands match the return type. The intrinsic is specialised.

Specifically,

Just adding this line `let x : u64 = __add(self, other);` in the function (expectedly) causes the below error :

``` 
error
  --> /media/builds/sway2/sway-lib-core/src/ops.sw:84:23
   |
82 | 
83 |     fn add(self, other: Self) -> Self {
84 |         let x : u64 = __add(self, other);
   |                       ^^^^^^^^^^^^^^^^^^ Mismatched types.
expected: u64
found:    u8.
```

Issue #5444, opened 2024-01-10
Enhance `forc_utils::cli_examples`


crodas:
Add more tests to extend the help and add it to the unit tests.

* [ ]  `forc deploy` with --default-signer, --testnet and --node-url
* [ ] `forc build`
* [ ] `forc clean`
* [ ] `forc contract-id`
* [ ] `forc init`
* [ ] `forc new`
* [ ] `forc parse-bytecpde`
* [ ]  `forc template`
* [ ] `forc test`
* [ ] `forc update`

PR #5519, opened 2024-01-26
Attempt to include special code only available for cli_examples

### LINKED ISSUE #5482: CLI tests should not be stateful
We have cli tests introduced in #5388, one problem with the implementation is that some actions are stateful, this is not a problem in CI as instances are killed after each run, but as sway repo developers our local systems should not be affected with tests running. Currently, a wallet is created in the default path  this is not good for developer experience and also creates problems if the developer has a wallet in the default path. We should consider changing the behavour so that anything that effects the system etc are cleaned after tests ran.

crodas:
## Description

The cli_examples are converted also to unit tests. These unit tests will compile a binary and execute it. The main goal is to enforce documentation correctness through the unit tests.

Sometimes a special case should be hard-coded (especially mocking user input) that should only be available for the binaries to be executed from the unit tests but should be excluded otherwise.

This PR adds the ability to include code through `option_env!` macro since the `CLI_TEST` should only be defined in for the binary to be executed from the unit tests.

This PR is a pre-requisite to solve https://github.com/FuelLabs/forc/issues/19. 

This PR also fixes #5482

This PR will include more examples as part of #5444.

## Checklist


forc/tests/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "tests"
```
sdankel, line 5: This sample project should be under `forc/tests/<project_name>`

It should also have a more meaningful name, since all the projects under `tests` are used for testing. 

The binary file should probably not be checked in.

forc-util/src/cli.rs
```diff
@@ -1,38 +1,123 @@
...
+            })
+            .unwrap();
+
+        *build_path = Some(new_build_path.clone());
+        format!("{}/{}", new_build_path, bin_name)
+    }
+}
+
 #[macro_export]
 // Let the user format the help and parse it from that string into arguments to create the unit test
```
JoshuaBatty, line 73: Can we turn this into a doc comment and provide more information about the intended use of setup and teardown?

forc-util/src/cli.rs
```diff
@@ -1,38 +1,143 @@
...
+/// test.
+///
+/// The list of examples is a list of tuples where the first element is the description of the test
+/// (in plain English) followed by the command to be executed and the arguments to be passed to the
+/// command. Optionally, the expected output of the command can be passed as well. This examples are
+/// part of the help message of the CLI.
+///
+/// Each example is also converted into a unit test. The test invokes the CLI command externall
+/// (there is no `#[cfg(test)]` since the command is an external process and unaware of the test
+/// context). The `option_env!("CLI_TEST").is_some()` macro can be used to detect if the command is
```
kayagokalp, line 85:
```suggestion
/// context). The `option_env!("CLI_TEST").is_some()` expression can be used to detect if the command is
```

forc-plugins/forc-client/src/util/tx.rs
```diff
@@ -59,6 +64,9 @@ fn prompt_signature(tx_id: fuel_tx::Bytes32) -> Result<Signature> {
 fn ask_user_yes_no_question(question: &str) -> Result<bool> {
     print!("{question}");
     std::io::stdout().flush()?;
+    if option_env!("CLI_TEST").is_some() {
+        return Ok(true);
+    }
```
kayagokalp, line 69: Not required at all but might make sense to create a macro that expands to this, as this appears a lot. Would bring most of these checks to single line from 3-4 lines.

`execute_if_cli_test!(return Ok(true)))`

forc-plugins/forc-client/src/cmd/deploy.rs
```diff
@@ -30,24 +33,28 @@ pub struct Command {
     pub node: NodeTarget,
     /// Optional 256-bit hexadecimal literal(s) to redeploy contracts.
     ///
-    /// For a single contract, use `--salt <SALT>`, eg.: forc deploy --salt 0x0000000000000000000000000000000000000000000000000000000000000001
+    /// For a single contract, use `--salt <SALT>`, eg.: forc deploy --salt
+    /// 0x0000000000000000000000000000000000000000000000000000000000000001
     ///
```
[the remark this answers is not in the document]
kayagokalp: If these are how these should be formatted I wonder if we can add a check for our doc comments formatting

forc-plugins/forc-client/src/util/tx.rs
kayagokalp: Also I am a little out of the loop so this might be already discussed. What was the reason we are doing these with `if` and not `cfg`s?

kayagokalp (commented):
Looking good, I haven't tested yet but left nitpicks and couple of questions to understand this better.

I see that there is `test_project2` under forc-doc. Is that still necessary. Maybe we can rename that to be more meaningful

forc-plugins/forc-client/src/util/tx.rs
kayagokalp: Arent' these only relevant while we are running `cargo test`? what would be stopping us from using something like `cfg(test)` instead of a seperate env variable `CLI_TESTS`. Do we want to be able to run tests without cli tests or something?

kayagokalp: We can maybe make `if_cli_tests { x }` macro to expand into:

```rust
#[cfg(test)]
{
x
}
```

kayagokalp: Ohh I see if we have other unit tests touching those functions things might get complicated, you may want to test the rest of the function but cli tests insertion would return early etc. A final thought is to make this into a separate flag:

```rust
fn test_me() {
    #[cfg(cli_test)]
    {
    a = 10;
    }
}
```

If you do not want to setup features in cargo tomls, you can set thorough env variables:

```console
RUSTFLAGS="--cfg cli_test" cargo test
```

kayagokalp: Cool

forc-util/src/cli.rs
```diff
@@ -1,38 +1,153 @@
...
+    };
+}
+
+#[macro_export]
+/// Let the user format the help and parse it from that string into arguments to create the unit
+/// test.
+///
+/// The list of examples is a list of tuples where the first element is the description of the test
+/// (in plain English) followed by the command to be executed and the arguments to be passed to the
+/// command. Optionally, the expected output of the command can be passed as well. This examples are
```
JoshuaBatty, line 90:
```suggestion
/// command. Optionally, the expected output of the command can be passed as well. These examples are
```

forc-util/src/cli.rs
```diff
@@ -1,38 +1,153 @@
...
+/// part of the help message of the CLI.
+///
+/// Each example is also converted into a unit test. The test invokes the CLI command externall
```
JoshuaBatty, line 93:
```suggestion
/// Each example is also converted into a unit test. The test invokes the CLI command externally
```

forc-util/src/cli.rs
```diff
@@ -1,38 +1,153 @@
...
+/// (there is no `#[cfg(test)]` since the command is an external process and unaware of the test
+/// context). The `option_env!("CLI_TEST").is_some()` expression can be used to detect if the command is
+/// being executed from the CLI_TEST environment and take a different path (for instance to mock a
+/// user given input response).
+///
+/// This macro also takes a list of examples and a setup and teardown block. The setup code block is
```
JoshuaBatty, line 99: I remember you saying you were thinking of removing the teardown block. Is this still the case? If so we should remove that from the doc comment.

JoshuaBatty (changes requested):

JoshuaBatty (approved):

forc-plugins/forc-client/mint.json
sdankel, line 1: nit: move this under `example/mint.json` or rename to `example-mint.json` so it's clear this is only used for examples

forc-plugins/forc-client/src/cmd/deploy.rs
```diff
@@ -9,8 +9,11 @@ use crate::NodeTarget;
 
 forc_util::cli_examples! {
     [ Deploy a single contract => deploy "bc09bfa7a11a04ce42b0a5abf04fd437387ee49bf4561d575177e2946468b408" => r#".*Error making HTTP request.*"# ]
-    [ Deploy a single contract from a different path => deploy "bc09bfa7a11a04ce42b0a5abf04fd437387ee49bf4561d575177e2946468b408 --path ../tests/" => r#".*Error making HTTP request.*"# ]
-    [ Deploy to a custom network => deploy "--node-url https://beta-5.fuel.network/graphql" => ".*Refused to create a new wallet.*" ]
+    [ Deploy a single contract from a different path => deploy "bc09bfa7a11a04ce42b0a5abf04fd437387ee49bf4561d575177e2946468b408 --path {path}" => r#".*Error making HTTP request.*"# ]
+    [ Deploy to a custom network => deploy "--node-url https://beta-5.fuel.network/graphql" ]
```
sdankel, line 13: Should there be an error for this one?

forc-plugins/forc-client/src/op/submit.rs
```diff
@@ -1,12 +1,22 @@
...
+use forc_tracing::println_warning;
+use forc_util::if_cli_test;
 use fuel_core_client::client::{types::TransactionStatus, FuelClient};
 use fuel_crypto::fuel_types::canonical::Deserialize;
 
 /// A command for submitting transactions to a Fuel network.
 pub async fn submit(cmd: cmd::Submit) -> anyhow::Result<()> {
     let tx = read_tx(&cmd.tx_path)?;
     let node_url = get_node_url(&cmd.network.node, &None)?;
+    if_cli_test! {
```
sdankel, line 12: I think you can do this without an env variable ([link](https://stackoverflow.com/questions/24396293/is-it-possible-to-conditionally-compile-a-code-block-inside-a-function))

```suggestion
    #[cfg(test)]
    {
```

forc/src/cli/mod.rs
```diff
@@ -28,6 +28,32 @@ mod commands;
 mod plugin;
 pub mod shared;
 
+/// Creates a new project and install their dependencies
+#[cfg(any(feature = "test", test))]
```
sdankel, line 32: Isn't this the same as:

```suggestion
#[cfg(test)]
```

forc-plugins/forc-client/src/op/submit.rs
sdankel: Even if you pass `cargo run -f test`? That should make it aware of the feature.

PR #5521, opened 2024-01-28
Initial DWARF debug symbols output code.

vaivaswatha:
This PR adds support to express line number mapping from assembly to source using DWARF object files. Since there are offset details that are specific to the Fuel ISA, I'm using the existing `SourceMap` that's already being built, rather than generating directly from the instructions.

An important point to note is that DWARF uses line-col to express positions rather than an absolute offset-from-start format. To avoid computing line-col from absolute positions (which requires re-reading the source files), I'm changing the source position format everywhere to line-col. This also simplifies our code at many places (for example, in `forc-debug`). The only exception is the `addr2line` tool, where the use of absolute positions is in a function that seemed quite complex to me to attempt changes. So I have, temporarily, added code to compute absolute positions from line-col information and then reuse the existing functions. This is inefficient, but I think that's _okay_ since it's a standalone tool and the whole thing runs just once for every command invocation.

The DWARF information is written to a file specified by the `-g` flag and can be verified using `llvm-dwarfdump --debug-line ./test.obj`. If the argument to the `-g` flag is a filename that ends with `.json`, then the existing JSON format is written out, otherwise it's a DWARF object in an ELF container.


vaivaswatha:
Although this PR still draft (because it isn't fit for merging yet), it would be helpful if various stakeholders would have a look at it.

Primarily, this PR adds support to express line number mapping from assembly to source using DWARF object files. Since there are offset details that are specific to the Fuel ISA, I'm using the existing `SourceMap` that's already being built, rather than generating directly from the instructions.

An important point to note is that DWARF uses line-col to express positions rather than an absolute offset-from-start format. To avoid computing line-col from absolute positions (which requires re-reading the source files), I'm changing the source position format everywhere to line-col. This also simplifies our code at many places (for example, in forc-debug). The only exception is the `addr2line` tool, where the use of absolute positions is in a function that seemed quite complex to me to attempt changes. So I have, temporarily, added code to compute absolute positions from line-col information and then reuse the existing functions. This is inefficient, but I *think* that's ok since it's a standalone tool and the whole thing runs just once for every command invocation (IIUC).

The DWARF information is written to `test.debug` (and hence the draft PR) and can be verified using `llvm-dwarfdump --debug-line ./test.debug`. We'll have to introduce a new flag or use the existing `-g` flag (to write DWARF objects instead of JSONs).

Review requests:

@kayagokalp, @sdankel.
1. This change from using absolute position to using line-col everywhere: https://github.com/FuelLabs/sway/pull/5521/files#diff-0229fd4ea2ee9ab8bd33615a6bf59125aae11b0b492f359a14dae4eb1b163faaR106. 
2. In general, the idea of using DWARF to carry debug information (line number information for the moment).

@Dentosal 
Changes in `addr2line`.

Dentosal:
The `addr2line` changes seem fine in sense that the code likely works. However, as the todo comment notes, it's quite inefficient.

forc-plugins/forc-debug/src/server/handlers/handle_launch.rs
```diff
@@ -118,50 +118,27 @@ impl DapServer {
...
-                                        .entry(line)
-                                        .and_modify(|val| {
-                                            // Store the instructions in ascending order
-                                            match val.binary_search(&instruction) {
-                                                Ok(_) => {} // Ignore duplicates
-                                                Err(pos) => val.insert(pos, instruction),
-                                            }
-                                        })
-                                        .or_insert(vec![instruction]);
+                    let LineCol { line, .. } = sm_span.range.start;
```
sdankel, line 123: This is much better 👍

forc/src/cli/shared.rs
```diff
@@ -29,7 +29,9 @@ pub struct BuildOutput {
     /// If set, outputs a binary file representing the script bytes.
```
sdankel, line 29:
```
/// Output a binary file representing the script bytecode to the provided path.
```

forc-pkg/src/pkg.rs
```diff
@@ -293,7 +293,9 @@ pub struct BuildOpts {
     pub minify: MinifyOpts,
     /// If set, outputs a binary file representing the script bytes.
```
sdankel, line 294:
```suggestion
    /// If set, outputs a a binary file at the given absolute path.
```

forc/src/cli/shared.rs
```diff
@@ -29,7 +29,9 @@ pub struct BuildOutput {
...
     #[clap(long = "output-bin", short = 'o')]
     pub bin_file: Option<String>,
-    /// If set, outputs source file mapping in JSON format
+    /// If set, outputs debug info to the provided file.
+    /// If the argument provided ends with .json, a JSON is emitted,
+    /// otherwise, an ELF file containing DWARF is emitted.
```
sdankel, line 34:
```suggestion
    /// Create a file containing debug information at the provided path.
    /// If the file extension is .json, JSON format is used. Otherwise, an ELF file containing DWARF is emitted.
```

forc/src/cli/shared.rs
sdankel, line 29:
```suggestion
    /// Create a binary file representing the script bytecode at the provided path.
```

sway-core/src/debug_generation/dwarf.rs
```diff
@@ -0,0 +1,128 @@
...
+        .unwrap();
+
+    obj.write_stream(file).map_err(|err| {
+        sway_error::error::CompileError::InternalOwned(err.to_string(), Span::dummy())
+    })?;
+
+    Ok(())
+}
+
+fn build_line_number_program(
```
sdankel, line 75: It would be good to add tests for this function

sway-core/src/debug_generation/dwarf.rs
```diff
@@ -0,0 +1,128 @@
...
+            sway_error::error::CompileError::InternalOwned(err.to_string(), Span::dummy())
+        })?;
+
+    let mut dwarf = DwarfUnit::new(encoding);
+    dwarf.unit.line_program = program;
+    // Write to new sections
+    let mut debug_sections = write::Sections::new(EndianVec::new(BigEndian));
+    dwarf
+        .write(&mut debug_sections)
+        .expect("Should write DWARF information");
```
sdankel, line 48: Should we return `Err` here instead?

sway-core/src/debug_generation/dwarf.rs
```diff
@@ -0,0 +1,128 @@
...
+        None,
+    );
+
+    program.begin_sequence(Some(write::Address::Constant(0)));
+
+    for (ix, span) in &source_map.map {
+        let (path, span) = span.to_span(&source_map.paths, &source_map.dependency_paths);
+
+        let dir = path.parent().expect("Path doesn't have proper prefix");
+        let file = path.file_name().expect("Path doesn't have proper filename");
```
sdankel, line 98: Should we return `Err` instead of using `expect`?

forc-pkg/src/pkg.rs
```diff
@@ -1892,7 +1900,7 @@ pub fn compile(
         "compile asm to bytecode",
         "compile_asm_to_bytecode",
         sway_core::asm_to_bytecode(&handler, asm, source_map, engines.se()),
-        Some(sway_build_config),
+        Some(sway_build_config.clone()),
```
sdankel, line 1903: Does this need to be cloned?

sway-core/src/debug_generation/dwarf.rs
vaivaswatha: To effectively test this, we'll need external dwarf tools. My thought was that once we have code in the debugger to read back this data, we could add tests there. 

If you think we should just use external tools to test it now, I can try adding that (any suggestions on which test in the CI this will go under?). Any other suggestion on how to test it would be helpful too.

sdankel: I thought we might be able to add a test sway project, pass that path and dir into this function, and assert that the resulting LineProgram is translated into the correct JSON and/or dwarf string. Is that possible without adding additional dependencies?

vaivaswatha: >Is that possible without adding additional dependencies?

I cannot see how. Because, this function only generates a DWARF object. It cannot read back that for us to compare. The reading part needs to be added into `forc-debug`. So when we add that, then we'll have an end-to-end flow and then it can be tested. Otherwise to read the DWARF object file, we'll need external tools such as `llvm-dwarf`.

sdankel (approved):

JoshuaBatty (approved):

PR #5530, opened 2024-01-30
Dereferencing operator index `[]`

ironcev:
## Description

This PR implements index operator `[]` for [references](https://github.com/FuelLabs/sway-rfcs/blob/ironcev/amend-references/files/0010-references.sw). The overall effort related to references is tracked in #5063.

`[]` is defined for references using this recursive definition: `<reference>[<index>] := (*<reference>)[<index>]`.

This eliminates the need for the dereferencing operator `*` when working with references to arrays:

```Sway
let array = [1, 2, 3];

let r = &&&array;

assert((***r)[0] == r[0]);
```
```Sway
let r = &&&[ &&[1, 2, 3], &&[4, 5, 6] ];
assert(r[0][2] == 3);
assert(r[1][0] == 4);
```

Additionally, the PR fixes two previously existing issues (see below screenshots):
- the misleading error message on type not implementing the "index" method when indexing a non-indexable type.
- the broken error span and expression text when indexing a non-indexable type in reassignments.

Before:
No method named index found for type

Not an indexable expression - Before

After:
Type is not indexable

## Checklist

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
+++ b/sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
@@ -1799,65 +1799,68 @@ impl ty::TyExpression {
         let type_engine = ctx.engines.te();
         let engines = ctx.engines();
 
-        let prefix_te = {
+        let mut current_prefix_te = Box::new({
             let ctx = ctx
                 .by_ref()
                 .with_help_text("")
                 .with_type_annotation(type_engine.insert(engines, TypeInfo::Unknown, None));
-            ty::TyExpression::type_check(handler, ctx, prefix.clone())?
-        };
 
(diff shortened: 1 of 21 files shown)


IGI-111 (approved):
Though we'll want to formalize it eventually with something like `Deref` it's appropriate to hardcode it for builtin types at this time.

ironcev:
> Though we'll want to formalize it eventually with something like `Deref` it's appropriate to hardcode it for builtin types at this time.

Yes, agree. In this case it will be `Deref` and `Index` traits. Once we introduce them, extending the current solution will be straightforward. The sam will be with the upcoming `.` dereferencing operator for field and element access and method calls.

xunilrj (approved):

PR #5533, opened 2024-01-31
Simplify the DCA handling for return nodes.

tritao:
## Description

This simplifies the handling of return nodes during dead code analysis.

Here is the updated graphs for this example:

```rust
script;
fn main() -> u64 {
    return 1;
    return 0
}
```

Before:

test

After:

test2

## Checklist

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
+++ b/sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
@@ -1799,65 +1799,70 @@ impl ty::TyExpression {
         let type_engine = ctx.engines.te();
         let engines = ctx.engines();
 
-        let prefix_te = {
+        let mut current_prefix_te = Box::new({
             let ctx = ctx
                 .by_ref()
                 .with_help_text("")
                 .with_type_annotation(type_engine.insert(engines, TypeInfo::Unknown, None));
-            ty::TyExpression::type_check(handler, ctx, prefix.clone())?
-        };
 
-        fn get_array_type(ty: TypeId, type_engine: &TypeEngine) -> Option<TypeInfo> {
-            match &*type_engine.get(ty) {
-                TypeInfo::Array(..) => Some((*type_engine.get(ty)).clone()),
-                TypeInfo::Alias { ty, .. } => get_array_type(ty.type_id, type_engine),
-                _ => None,
-            }
+            ty::TyExpression::type_check(handler, ctx, prefix)?
+        });
+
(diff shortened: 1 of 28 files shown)


ironcev:
# TL;DR
- The PR removes the double edge from the `return` node to its contained nodes which was a bug :+1: 
- Moving the "return" edge from the `return` to the content might actually lead to more complex graphs if we have more then one node in the content.
- Can we also remove the "return" edge going from the implicit return to the `"main" fn exit`?
- I see PR as going in the right direction, because to properly handle divergence we will need to dig into the `return`'s content and not only directly connect `return` to the exit node.

# Detailed thoughts
Ha ha, I was thinking about that very same TODO comment today, while pondering on the background thread about our `deterministically_aborts` + `Never` + `match` + `TypeInfo::Unknown` combo of issues :smile: Though my thinking was more on the question of divergence which was not in focus of this PR.

My thoughts. Removing the double edge from the `return` node to its content is a simplification. That was definitely a bug before, adding an edge after it was already added in the `connect_expression`.

Moving the "return" line from the `return` to the content node might be seen as a simplification, but it looks to me more as a matter of taste. For example, if there are more than one node in the `return` content like in this example, one might argue that we actually got a more complex graph with more edges (assuming we remove the double edge):

```Sway
fn main() -> [u64;3] {
    return [1, 2, 3];
    return [1, 2, 3]
}
``` 
Before:
DCA graph before

After:
DCA graph after

But still I think the new approach moves us in the right direction, where we will need to inspect the `return` content and check for divergence. This is what I planed to do after sorting out the `deterministically_abort` issue.

What I want to say, in an example like the one below, we cannot simply connect all the parts of the content to the exit node or mark them as reachable. The variable `a` is not reachable and it should be reported as such, but DCA currently does not marks it as such, because in both the new and old implementation of `return` it assumes that whatever is in `return` will be reachable which must not be the case.

```Sway
fn main() -> [u64;3] {
    let a = 123u64;
    return [0u64, return [1, 2, 3], a];
}
```
For the same example, Rust will properly report that `a` is not reachable because of the preceeding `return [1, 2, 3]`.

But sorry, I went a bit too far, divergence is way out of scope of this PR :smile: 

A detail perhaps being relevant to this PR is that one "return" edge going from the implicit `return` to the `"main" fn exit`. Can we also remove that edge or it needs to stay?

ironcev (approved):
Approved as per the comment. Removing the direct "return" edge between the implicit `return` and the exit might be worth thinking of.

Issue #5504, opened 2024-01-22
Infinite recursion (stack overflow) calling trait function on tuples


xunilrj:
I managed to reproduce this problem with this minimal example on `master`.

```sway
script;

trait T1 {
    fn trait_fn() -> Self;
}

impl T1 for u64 {
    fn trait_fn() -> u64 {
        0
    }
}

impl<A> T1 for (A, ) 
where
    A: T1
{
    fn trait_fn() -> (A, ) {
        (A::trait_fn(), )
    }
}

fn f<T> () 
where
    T: T1
{
    T::trait_fn();
}

fn main() {
    let a = f::<(u64, )>();
}
````

The first interesting code in the stack is at `sway-core/src/language/ty/expression/expression_variant.rs:784` (all the others frames in our codebase point to compiler derived `clone()`s)

```rust
(gdb) where
#0  0x00005555576de047 in alloc::slice::hack::{impl#0}::to_vec<sway_core::language::ty::expression::expression::TyExpression, alloc::alloc::Global> (s=..., 
    alloc=...) at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/alloc/src/slice.rs:124
#1  0x0000555557a16138 in alloc::slice::hack::to_vec<sway_core::language::ty::expression::expression::TyExpression, alloc::alloc::Global> (s=..., alloc=...)
    at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/alloc/src/slice.rs:111
#2  alloc::slice::{impl#0}::to_vec_in<sway_core::language::ty::expression::expression::TyExpression, alloc::alloc::Global> (self=..., alloc=...)
    at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/alloc/src/slice.rs:441
#3  alloc::vec::{impl#10}::clone<sway_core::language::ty::expression::expression::TyExpression, alloc::alloc::Global> (self=0x55555b4b29c0)
    at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/alloc/src/vec/mod.rs:2685
#4  0x0000555557ed22ca in sway_core::language::ty::expression::expression_variant::{impl#11}::clone (self=0x55555b4b29b8)
    at sway-core/src/language/ty/expression/expression_variant.rs:54
#5  0x0000555557b47f08 in sway_core::language::ty::expression::expression::{impl#13}::clone (self=0x55555b4b29b8)
[...]

PR #5541, opened 2024-02-02
Fixes `ReplaceDecls` to no longer depend on filter hack.

esdrubal:
`ReplaceDecls` was using a hack to replace some function refs based on
the self function parameter, this didn't work for associated functions.

Now DeclMapping sources also contain a typeid that guarantees that the
replacement is done to the correct declaration in case we have nested
trait methods with the same name replaced.

Fixes https://github.com/FuelLabs/sway/issues/5504

## Checklist


sway-core/src/decl_engine/associated_item_decl_id.rs
```diff
@@ -14,6 +16,17 @@ pub enum AssociatedItemDeclId {
...
 }
 
+impl AssociatedItemDeclId {
+    pub fn span(&self, engines: &Engines) -> Span {
+        match self {
+            Self::TraitFn(decl_id) => engines.de().get(decl_id).span(),
+            Self::Function(decl_id) => engines.de().get(decl_id).span(),
+            Self::Constant(decl_id) => engines.de().get(decl_id).span(),
+            Self::Type(decl_id) => engines.de().get(decl_id).span(),
+        }
```
IGI-111, line 26: We should probably split the match into a (private) `decl_id` function and make this a oneliner.

ironcev (approved):
LGTM. BDW, with this change implementingthe exact `TyFunctionDecl::is_constructor` method should be straightforward.  The TODO I left in that method was exactly about the difficulty of determining the exact `Self` type for the method. I should give it a try and implement that TODO once this PR gets accepted :smile:

sway-core/src/decl_engine/associated_item_decl_id.rs
IGI-111: Ah yes, never mind.

PR #5552, opened 2024-02-05
Dereferencing operator `.` on tuple element access

ironcev:
## Description

This PR implements tuple element access operator `.` for [references](https://github.com/FuelLabs/sway-rfcs/blob/ironcev/amend-references/files/0010-references.sw). The overall effort related to references is tracked in #5063.

`.` is defined for references to tuples using this recursive definition: `<reference>.<element index> := (*<reference>).<element index>`.

This eliminates the need for the dereferencing operator `*` when working with references to tuples:

```Sway
let t = (1, 2, 3);

let r = &&&t;

assert((***r).0 == r.0);
```
```Sway
let r = &&& ( &&(1, 2), &(3, 4) );
assert(r.0.0 == 1);
assert(r.1.1 == 4);
```

Additionally, the PR:
- creates `Diagnostic` for the `TupleIndexOutOfBounds` error.
- harmonizes the appearance of the `TupleIndexOutOfBounds` error in the element access and reassignments.

## Demo

`TupleIndexOutOfBounds` in element access and reassignments before:

Tuple index out of bounds - Before

New errors:

Tuple index out of bounds - After

Tuple element access requires a tuple

## Checklist

### DIFF
--- a/sway-error/src/error.rs
+++ b/sway-error/src/error.rs
@@ -303,7 +303,7 @@ pub enum CompileError {
         field_name: IdentUnique,
         span: Span,
     },
-    #[error("This expression has type \"{actually}\", which is not a tuple. Elements can be access only on tuples.")]
+    #[error("This expression has type \"{actually}\", which is not a tuple. Elements can only be accessed on tuples.")]
     TupleElementAccessOnNonTuple {
         actually: String,
         span: Span,
--- a/sway-error/src/error.rs
+++ b/sway-error/src/error.rs
@@ -565,12 +565,6 @@ pub enum CompileError {
         index: usize,
         count: usize,
         tuple_type: String,
-        /// Denotes if the tuple element access occurs over
-        /// references. The meaning of the `referencing_level` is:
-        ///   0 - no referencing. Element access directly on a tuple type.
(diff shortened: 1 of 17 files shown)


sway-error/src/error.rs
```diff
@@ -1818,6 +1831,62 @@ impl ToDiagnostic for CompileError {
...
+                        format!("This expression has type \"{tuple_type}\".")
+                    ),
+                    Hint::info(
+                        source_engine,
+                        prefix_span.clone(),
+                        format!("That is {} tuple of only {} element{}.",
+                            match *referencing_level {
+                                0 => "a",
+                                1 => "a reference to a",
+                                _ => "an indirect reference to a",
```
IGI-111, line 1861: I think having this whole infrastructure and concept of a reference depth just to have the right turn of phrase is overengineering.

Devs will expect the dereferencing to work through refs, we shouldn't have to explain it, and if we do we should not do it in terms this specific.

I think this should just say "The tuple has only x elements", provide the type and ditch this last hint along with all references to a referencing level.

IGI-111 (approved):

PR #5575, opened 2024-02-08
Fixes issue of missing type annotations on explicit returns.

### LINKED ISSUE #5518: Broken handling of type annotation for implicit/explicit return type checking
Lets check out this code from standard library `sway-lib-std/src/bytes_conversions/b256.sw`.

```
impl b256 {
    pub fn from_be_bytes(bytes: Bytes) -> Self {
        assert(bytes.len() == 32);
        bytes.into();
    }
}
```

With https://github.com/FuelLabs/sway/pull/5464, this started failing, for the reason that the type annotation in the implicit return expression now gets set no `Unknown`.

While at first, this seems like a bug specific to implicit returns, after some consideration, we can reproduce a similar error with `master` and changing to to use an explicit return instead:

```
impl b256 {
    pub fn from_be_bytes(bytes: Bytes) -> Self {
        assert(bytes.len() == 32);
        **return** bytes.into();
    }
}
```

```
error
   --> sway-lib-std/src/bytes_conversions/b256.sw:145:22
    |
143 | 
144 |         assert(bytes.len() == 32);
145 |         return bytes.into();
    |                      ^^^^ Multiple applicable items in scope. 
  Disambiguate the associated function for candidate #0
    <Bytes as From<b256>>::into
  Disambiguate the associated function for candidate #1
    <Bytes as From<slice>>::into
  Disambiguate the associated function for candidate #2
    <Bytes as From<Vec<u8>>>::into
146 |     }
147 | }
    |
```

So the refactoring PR just surfaced an error that has always been there, and needs to be fixed for both explicit and implicit returns.

Taking a look at the code that handles this, it becomes clear why:

```
            ExpressionKind::Return(expr) => {
                let ctx = ctx
                    // we use "unknown" here because return statements do not
                    // necessarily follow the type annotation of their immediate
                    // surrounding context. Because a return statement is control flow
                    // that breaks out to the nearest function, we need to type check
                    // it against the surrounding function.
                    // That is impossible here, as we don't have that information. It
                    // is the responsibility of the function declaration to type check
                    // all return statements contained within it.
                    .by_ref()
                    .with_type_annotation(type_engine.insert(engines, TypeInfo::Unknown, None))
                    .with_help_text(
                        "Returned value must match up with the function return type \
                        annotation.",
                    );
```

Instead we should be keeping track of current function context and setting up its type as the type annotation for the return.

esdrubal:
## Description

Explicit returns did not have any type annotation causing the method disambiguation to fail.

The solution was to add a new field to the `TypeCheckContext` called `function_type_annotation`. This field is initialized only in the type_check of function declarations. The field is used to set the type_annotation one while doing the type check of explicit return expressions.

With this change `TypeCheckUnification` was also simplified as the unification is now done with `function_type_annotation`. The result was a few functions that are no longer required to be removed.

Closes #5518

## Checklist

### DIFF
--- a/sway-lib-std/src/math.sw
+++ b/sway-lib-std/src/math.sw
@@ -6,6 +6,24 @@ pub trait Root {
     fn sqrt(self) -> Self;
 }
 
+impl Root for u256 {
+    // Integer square root using [Newton's Method](https://en.wikipedia.org/wiki/Integer_square_root#Algorithm_using_Newton's_method).
+    fn sqrt(self) -> Self {
+        let mut x0 = self >> 1;
+        if x0 == 0 {
+            return self;
+        }
+        let mut x1 = (x0 + self / x0) >> 1;
+
+        while x1 < x0 {
+            x0 = x1;
+            x1 = (x0 + self / x0) >> 1;
+        }
+
+        x0
+    }
+}
+
 impl Root for u64 {
     fn sqrt(self) -> Self {
         let index: u64 = 2;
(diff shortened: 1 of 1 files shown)


jjcnn (approved):
Around 200 lines of code removed, around 20 lines of code added. I approve of this. :-)

PR #5576, opened 2024-02-08
Provide `__jmpb_ssp` intrinsic

### LINKED ISSUE #5563: Instrinsic jump to $ssp
To use LDC we need a way to jump into the loaded code like so: https://github.com/FuelLabs/ldc-testing/blob/7d188823b2fa27a027b3733d7deb332f30ff1524/test-contract/src/main.sw#L33

We should add a `__jump_to_ssp` intrinsic to make this possible without having to allow control flow opcodes in asm blocks.

vaivaswatha:
The new intrinsic `__jmpb_ssp(offset: u64)` jumps backwards, from the current value of `$ssp`, by a provided byte offset.

This intrinsic is necessary for dynamically loading and executing a contract.

Here's an example contract (taken from https://github.com/FuelLabs/ldc-testing) for dynamically loading another contract and transferring control to it. The "loadable" contract is the one used in the link above.

```
// to be compiled in release mode
contract;

abi MyContract {
    fn test_function(code_id: ContractId);
}

    fn test_function(code_id_p: ContractId) {
        let length = asm(code_id, length, word, ssp_saved) {
            // Allocate 32 bytes on the heap (we can't use the stack)
            addi word zero i32;
            aloc word;

            lw code_id fp i74;
           
            // Log the ContractID for debugging
            logd zero zero code_id word;

            // Load the entire contract with LDC
            csiz length code_id;
            // Save the old ssp
            move ssp_saved ssp;
            ldc code_id zero length;
            // return the ssp difference, to feed __jmpb_ssp.
            // This need not always be equal to `length` as `ldc` pads the `length`.
            sub length ssp ssp_saved;
            length: u64
        };
        __jmpb_ssp(length)
    }
}
```

The intrinsic generates code to do the following:
1. Subract offset from `$ssp`.
2. Subtract `$is` since `$jmp` adds it back.
3. Divide by 4 since `Jmp` multiplies by 4
4. Jump to the computed value.

Closes #5563


IGI-111 (changes requested):
We're missing a documentation mention in `book/src/reference/compiler_intrinsics.md` but otherwise LGTM

IGI-111 (approved):

IGI-111 (approved):

dmihal (approved):

PR #5588, opened 2024-02-09
cfei: don't generate if frame size is zero

### LINKED ISSUE #5580: Do not generate `cfei i0`
Running the IR generation test `empty.sw` one can see that we generate `cfei i0` (more at https://specs.fuel.network/master/fuel-vm/instruction-set.html#cfei-extend-call-frame-immediate), which is useless because it is expanding the stack by zero bytes.

```
> cargo r -p test --release -- empty.sw --verbose
...
::check-asm::
> .program:
> jmpf $zero i3
> noop
> DATA_SECTION_OFFSET[0..32]
> DATA_SECTION_OFFSET[32..64]
> lw   $ds $is 1
> add  $$ds $$ds $is
> move $$locbase $sp            ; save locals base register
> cfei i0                       ; allocate 0 bytes for locals and 0 slots for call arguments.
> ret  $zero                    ; returning unit as zero
  ^~~~~~~~~~
Matched #3: \bret  \$zero\b
> .data:
  ^~~~~~
Matched #4: \.data:
```

This `cfei`is generated at `sway-core/src/asm_generation/fuel/functions.rs` line 825. One just need to check if the calculated value is zero and not generate this opcode.

A bonus would be to track all places that uses `VirtualOp::CFEI` and check for the immediate value as zero.

The test can be expanded with

```
check: move $$$$locbase $$sp
not: cfei i0
```

sudhackar:
## Description
This small change fixes #5580 as described - only needed 1 additional change and small test fix

Ping @xunilrj 

## Checklist


sudhackar:
the failure looks related - it fails on local too but not on master. not flaky.

vaivaswatha:
Hold on merging this. I vaguely remember there being a reason for it's existence. I'll update back soon.

sudhackar:
hmm.

All the tests seem to pass with this change 
```diff
diff --git a/test/src/e2e_vm_tests/test_programs/should_pass/unit_tests/workspace_test/contract_multi_test/src/main.sw b/test/src/e2e_vm_tests/test_programs/should_pass/unit_tests/workspace_test/contract_multi_test/src/main.sw
index b97f32603..7cdc9a8d0 100644
--- a/test/src/e2e_vm_tests/test_programs/should_pass/unit_tests/workspace_test/contract_multi_test/src/main.sw
+++ b/test/src/e2e_vm_tests/test_programs/should_pass/unit_tests/workspace_test/contract_multi_test/src/main.sw
@@ -17,7 +17,7 @@ fn test_foo() {

 #[test(should_revert)]
 fn test_fail() {
-    let contract_id = 0xa5cd13d5d8ceaa436905f361853ba278f6760da2af5061ec86fe09b8a0cf59b4;
+    let contract_id = CONTRACT_ID;
     let caller = abi(MyContract, contract_id);
     let result = caller.test_function {}();
     assert(result == false)
@@ -25,7 +25,7 @@ fn test_fail() {

 #[test]
 fn test_success() {
-    let contract_id = 0xa5cd13d5d8ceaa436905f361853ba278f6760da2af5061ec86fe09b8a0cf59b4;
+    let contract_id = CONTRACT_ID;
     let caller = abi(MyContract, contract_id);
     let result = caller.test_function {}();
     assert(result == true)
```

`contact_id` for some reason has been hardcoded here.

IGI-111:
We actually want the contract ids hardcoded in those tests. They're meant to detect possible ABI breakage, using `CONTRACT_ID` would render them trivial.

xunilrj:
> There's an assumption ... in the spiller that the `cfei` instruction always exists. It's hard for the spiller to know where to insert a new `cfei`, so it just edits the existing one by growing it by the total spill size.

That makes sense... but what about after the register allocation?

We can move this check to `AllocatedOps::to_fuel_asm` at `sway-core/src/asm_lang/allocated_ops.rs`. This methods already allows a `AllocatedOp` to return a `Vec<FuelOpcode>`, so we could do:

```rs
 CFEI(a) if a.is_zero() => return Left(vec![]),
 CFEI(a) => op::CFEI::new(a.value.into()).into(),
```

vaivaswatha:
> > There's an assumption ... in the spiller that the `cfei` instruction always exists. It's hard for the spiller to know where to insert a new `cfei`, so it just edits the existing one by growing it by the total spill size.
> 
> That makes sense... but what about after the register allocation?
> 
> We can move this check to `AllocatedOps::to_fuel_asm` at `sway-core/src/asm_lang/allocated_ops.rs`. This methods already allows a `AllocatedOp` to return a `Vec<FuelOpcode>`, so we could do:
> 
> ```rust
>  CFEI(a) if a.is_zero() => return Left(vec![]),
>  CFEI(a) => op::CFEI::new(a.value.into()).into(),
> ```

This is acceptable. Good idea !.

sudhackar:
Will follow this feedback and get back.

sudhackar:
> ```
> CFEI(a) if a.is_zero() => return Left(vec![]),
> CFEI(a) => op::CFEI::new(a.value.into()).into(),
> ```

the zero path is triggered quite a few times. This change effectively means we are deleting a line - which could invalidate relations between `AllocatedOp` and `CompiledBytecode` such as offsets - triggering even simple tests `vec_set_index_out_of_bounds` to fail.

I confirmed that 
```rust
            CFEI(a) if a.value == 0 => {
                dbg!(("CFEI with value 0", &a));
                op::NOOP::new().into()
            },
```
instead of deleting - replacing with a NOOP would pass the current tests.

Is there something that I'm missing?

sudhackar:
Can I get some traction here?

vaivaswatha:
Thank you @sudhackar , and apologies for the delay in my response.

> replacing with a NOOP would pass the current tests.

But what do we gain at all by replacing it with a NOOP? The code size remains the same. @xunilrj @IGI-111 any thoughts on this?

> which could invalidate relations between AllocatedOp and CompiledBytecode such as offsets 

I should've seen this coming. There's one more trick you can try. In the function `sway_core::asm_generation::fuel::allocated_abstract_instruction_set::AllocatedAbstractInstructionSet::instruction_size`, if you return the size of `cfei 0` and `cfsi 0` to `0`, then it wouldn't be counted when computing offsets. That way when you actually omit emitting the instruction (instead of a NOOP), the offsets should all add up. If that doesn't work (for some unforeseen reason) we'll go with the current change.

sway-core/src/asm_lang/allocated_ops.rs
```diff
@@ -594,6 +594,7 @@ impl AllocatedOp {
 
             /* Memory Instructions */
             ALOC(a) => op::ALOC::new(a.to_reg_id()).into(),
+            CFEI(a) if a.value == 0 => op::NOOP::new().into(),
```
vaivaswatha, line 597: Shouldn't the same thing be done for `CFSI` also?

test/src/ir_generation/tests/empty.sw
```diff
@@ -10,6 +10,8 @@ fn main() {
 
 // ::check-asm::
 // The data section setup:
+// check: move $$$$locbase $$sp
+// check: cfei i0
```
vaivaswatha, line 14: I'm a little confused. Shouldn't this have been NOOP already?

sudhackar: This is for IR generation. Currently I noop in the asm generation.

sway-core/src/asm_lang/allocated_ops.rs
sudhackar: will do

sudhackar:
thanks @vaivaswatha for the details - right now with this change - the IR will still generate `cfei i0` and `cfsi i0` but while the asm generation - these instructions will be removed.

There are no explicit test cases for this change - I'll try to add this. I have added an explicit `cfei i0` check in the IR generation just to highlight this.

sway-core/src/asm_generation/fuel/allocated_abstract_instruction_set.rs
```diff
@@ -419,6 +419,13 @@ impl AllocatedAbstractInstructionSet {
                 }
             }
 
+            Either::Left(AllocatedOpcode::CFEI(ref op))
```
vaivaswatha, line 422: Please add a comment here on why we're doing this. Something like "cfei 0 and cfsi 0 are omitted from actual asm emission hence we shouldn't count them for offsets"

vaivaswatha (approved):

vaivaswatha:
Thank you @sudhackar

IGI-111 (approved):

PR #5601, opened 2024-02-12
Storage namespace annotation

### LINKED ISSUE #5490: Storage namespacing
In some cases, such as when loading code from a contract within an LDC proxy, it's possible to mix code that uses the same storage context.

For instance, let's say we have a proxy that uses a single storage slot to store the address it's calling into, and a contract it calls that uses an arbitrary amount of storage.

To maintain coherent execution, the callee must leave the first storage slot untouched.

This creates the possibility for a lot of errors, prevents the future extension of the size of the caller storage context and generally is confusing.

We should instead allow contract storage to be optionally parameterized with a namespace string that acts as a salt when calculating the hash of storage slots, thus preventing collision between storage that uses a different namespace argument.

IGI-111:
## Description

Add a `#[namespace(my_namespace)]` annotation to allow developers to avoid storage slot collisions when loading contract code using LDC.

Fixes #5490

## Checklist


xunilrj (approved):

sway-core/src/ir_generation/function.rs
```diff
@@ -2970,7 +2972,7 @@ impl<'eng> FnCompiler<'eng> {
 
         // Do the actual work. This is a recursive function because we want to drill down
         // to load each primitive type in the storage field in its own storage slot.
-        self.compile_storage_read(context, ix, &field_idcs, &base_type, span_md_idx)
+        self.compile_storage_read(context, ns, ix, &field_idcs, &base_type, span_md_idx)
```
ironcev, line 2975: Nitpick. Just a slight inconsistency, `compile_storage_access` accepts `fields`, `ix`, and `ns` in that order, whereas other methods go for `ns`, `ix`, `indices` (gotten from `fields`) essentially other way around (and more logical to me, to start with the namespace).

sway-core/src/semantic_analysis/ast_node/declaration/storage.rs
```diff
@@ -58,6 +59,7 @@ impl ty::TyStorageField {
         md_mgr: &mut MetadataManager,
         module: Module,
         ix: &StateIndex,
+        namespace: Option<&str>,
```
ironcev, line 62: While I am at nit picking :smile: here as well, `namespace` first (or `ns` to be fully consistent?).

test/src/sdk-harness/test_projects/storage_namespace/.gitignore
```diff
@@ -0,0 +1,2 @@
+out
```
ironcev, line 1: Local .gitignore file not needed.

ironcev (approved):
Few nit picks and one redundant `.gitignore` file worth deleting, but otherwise LGTM very straightforward implementation.

ironcev (approved):

JoshuaBatty (approved):

Issue #5560, opened 2024-02-07
Internal compiler error when all `match` branches return `Never`


ironcev:
This code fails with `Internal compiler error: Program erroneously uses uninitialized virtual registers: $r2`.

```Sway
script;

struct S {  x: u8, }

fn revert_me() -> Never {
  __revert(0)
}

fn main() -> bool {
  let s =  S { x: 0 };

  match Option::Some::<S>(s) {
    Some(_) => revert_me(),
    None => revert_me(),
  }

  true
}
```

ironcev:
As expected, the changes done in #5602 and #5643 have indirectly fixed this issue as well. It cannot be reproduced on master anymore so I am closing it.

Of course, since we've replaced `Never` with `!` in the attempt to reproduce the issue the `revert_me` function returns `!`.

PR #5602, opened 2024-02-12
Implements the Never `!` types as a TypeInfo bottom type.

### LINKED ISSUE #5562: Resolve type-system and unification related issues due to missing bottom type and `TypeSystem::Unknown` type annotations
We have several open issues all coming from the same root cause, the fact that we miss the proper bottom type and that we are not passing the correct type annotations in `TypeCheckContext`. Here are some notable examples:

- [ ] #5492
- [ ] #5438
- [ ] #5518
- [ ] #5559
- [ ] #5560
- [ ] #5561

We had similar issues in the past, and always addressed them in isolation. This proposal calls for a systematic fix of all such issues by addressing the root cause.

Proposed steps are the following:
- [ ] Introduce the proper bottom type `!` (Never)
- [ ] Remove the `DeterministicallyAborts` trait and all the calls to the `deterministically_aborts` method
- [ ] Pass the function return type as a part of the `TypeCheckContext`
- [ ] Double-check all the places where we set `TypeInfo::Unknown` as `ctx` type annotation
- [ ] Support `!` and divergence in DCA
- [ ] Optional: Introduce functions like `unreachable` and `todo`

## Introduce the proper bottom type
- [ ] Introduce built-in `!` type with the semantic same to the [Rust never type](https://doc.rust-lang.org/std/primitive.never.html)
- [ ] `break`, `continue`, and `return` have type `!`; consider this in `compute_return_type_and_span`
- [ ] Remove the hack that every empty enum behaves like bottom type in coersion
- [ ] Remove `core::never::Never` and check library functions that need to return `!` (e.g., `revert()`)

## Remove the `DeterministicallyAborts` trait
`DeterministicallyAborts` actually compensates, or tries to compensate, for the missing bottom type. It is erroneous in its current implementation as #5438 explains, but there is no need to fix it. Once we properly introduce `!` and diverging expression have it as a type, the consistent type system will automatically take care of the few places where `deterministically_aborts` was used.

- [ ] As a part if this step, we can check and eliminate double-unifications. E.g. `TyMatchBranch::type_check` calls `branch_ctx.unify_with_type_annotation()` although the type check of the branch result already does the unification.

## Pass the function return type as a part of the context
@tritao already very well explained the need for this in #5518.

- [ ] As a part if this step we can remove `should_unify` from `TyExpression::type_check`.

## Check places where `TypeInfo::Unknown` is set as type annotation
There are parts of semantic analysis where we want `TypeInfo::Unknown` to be set as `ctx` type annotation. Such places should always be documented in a comment. There are other places where we do set it currently, either:
1. because we were forced to due to the lacking of the proper bottom type
1. because we have a bug and need to properly consider which type actually needs to be passed

An example for 1) is #5492. The bug shows us that `if` and `match` expression should not remove the received type annotation from the context by setting it to `Unknown`. But once @vaivaswatha put the proper annotations to `ctx` passed to those expressions we've started getting tons of issues related to the missing bottom type.

An example for 2) is #5559.

## Support Never and divergence in DCA
Currently, `__revert()` and `revert()` are treated as special cases in the DCA. Once we get the `!` type, this special treatment should be replaced by the generic analysis of all functions that return `!`.

Also, once all diverging expressions get `!` as type, we can straightforwardly fully support divergence in all expressions (e.g., #5561).

## Optional: Introduce functions like `unreachable` and `todo`
Once we have `!` we could offer counterparts of Rust `unreachable` and `todo` macros as `std` functions. If I remember well this was even requested during the `std` audit. The function would simply log the message and revert.

However, we should think carefully about this. One day we want to have macros and having these as macros like in Rust would be more suitable. Without macros we cannot offer at the same time `todo()` without arguments and `todo("With argument")`. But then, we can start with functions and turn to macros later on. Decision needed, but nevertheless '!' opens door to this functionality.

esdrubal:
## Description

We now parse the `!` as a TypeInfo::Never, and remove the usage of empty enums as Never type in our code.

This commit removes completely the DeterministicallyAborts and TypeCheckUnificationContext.

The DeterministicallyAborts can be removed because the Never TypeInfo is now propagated using the type checker. Code blocks that return, break, continue, or call an expression that returns Never, are marked as Never.

Partially fixes #5562.

## Checklist


IGI-111 (changes requested):
I do so love PRs with negative line totals. Getting rid of all this complex logic in favor of a defined type is nice. The end result looks a lot cleaner.

But I'm about to ruin it by asking you to write some documentation for it.
`!` is a fairly complex concept and the book should absolutely cover it.

sway-core/src/type_system/unify/unifier.rs
```diff
@@ -137,6 +137,8 @@ impl<'a> Unifier<'a> {
                 )
             }
 
+            (Never, Never) => (),
```
ironcev, line 140: Why this special case here? Isn't that case also covered with the below cases with `Never` and placeholder?

sway-core/src/type_system/unify/unifier.rs
```diff
@@ -181,14 +183,16 @@ impl<'a> Unifier<'a> {
             {
                 self.replace_expected_with_received(expected, r, span)
             }
+
+            // Never type coerces to any other type.
+            // This should be after the unification of self types.
+            (Never, _) => {}
+            (_, Never) => {}
```
ironcev, line 190: Is this valid? `Never` should coerce into any type, but no other type should coerce into `Never` except `Never`, or? I mean:
```Rust
let _: u8 = __revert();      // Ok.  Never -> u8.
let _: u8 = { return 123 };  // Ok.  Never -> u8.
let _: ! = __revert();       // Ok.  Never -> Never.
let _: ! = { return 123 };   // Ok.  Never -> u8.

let _: ! = 123u8;  // ERROR.  u8 -> Never.
```

sway-core/src/type_system/unify/unify_check.rs
```diff
@@ -357,12 +358,10 @@ impl<'a> UnifyCheck<'a> {
                         !OccursCheck::new(self.engines).check(right, left)
                     }
 
-                    // Let empty enums to coerce to any other type. This is useful for Never enum.
-                    (Enum(r_decl_ref), _)
-                        if self.engines.de().get_enum(r_decl_ref).variants.is_empty() =>
-                    {
-                        true
-                    }
+                    // Never coerces to any other type.
```
ironcev, line 361: Same question here.

sway-ir/src/verify.rs
```diff
@@ -1017,7 +1014,15 @@ impl<'a, 'eng> InstructionVerifier<'a, 'eng> {
     // This is a really common operation above... calling `Value::get_type()` and then failing when
     // two don't match.
     fn opt_ty_not_eq(&self, l_ty: &Option<Type>, r_ty: &Option<Type>) -> bool {
-        l_ty.is_none() || r_ty.is_none() || !l_ty.unwrap().eq(self.context, r_ty.as_ref().unwrap())
+        l_ty.is_none()
+            || r_ty.is_none()
+            || (!l_ty.unwrap().is_never(self.context)
```
ironcev, line 1019: Should't this last part just call `ty_not_eq`? But please, first see the comment for `ty_not_eq` below.

sway-ir/src/verify.rs
```diff
@@ -1017,7 +1014,15 @@ impl<'a, 'eng> InstructionVerifier<'a, 'eng> {
...
+                && !r_ty.unwrap().is_never(self.context)
+                && !l_ty.unwrap().eq(self.context, r_ty.as_ref().unwrap()))
+    }
+
+    fn ty_not_eq(&self, l_ty: &Type, r_ty: &Type) -> bool {
+        !l_ty.is_never(self.context) && !r_ty.is_never(self.context) && !l_ty.eq(self.context, r_ty)
```
ironcev, line 1025: The `Type::eq()` is not implemented for `Never` (in _irtype.rs_), that needs to be done. If it were, it looks to me that `ty_not_eq` wouldn't be needed at all. `Never` comparison would also be done in `eq()`.

ironcev (changes requested):
Awesome!! Can you please check the questions regarding the coercion and equality on the IR level?

Also, assuming how crucial `Never` is for the type system, I would expect new dedicated tests. E.g., testing coercion rules could be an example, having `!` as a type in different positions, e.g., as a function parameter, etc.

ironcev:
BDW solving #5560 would be a great test of the new type system (as a separate PR of course). I wouldn't be surprised if the bug is no more reproducible on this branch :smile:

ironcev:
> But I'm about to ruin it by asking you to write some documentation for it.

@IGI-111 Completely agree about the need for the docs! But want to selfishly ask if the documentation could be a part of a separate PR if that will speed up the merge of this PR. I was doing changes in the `Unifier` and `UnifyCheck` and in general in the area of type cheking and would love to get those changes rebased on this PR before pushing them.

IGI-111:
Okay sure, if that blocks some other work we can split it into another PR.

sway-core/src/ir_generation/convert.rs
```diff
@@ -159,6 +159,7 @@ fn convert_resolved_type(
             convert_resolved_typeid(type_engine, decl_engine, context, &ty.type_id, span)?
         }
         TypeInfo::Ref(_) => Type::get_uint64(context),
+        TypeInfo::Never => Type::get_never(context),
```
jjcnn, line 162: Now that `convert_resolved_type` doesn't throw an error on diverging expressions we can simplify `compile_array_expr` in `function.rs`. It currently generates code for the first array element before creating the array, because creating the array requires the element type to be converted which wasn't possible when we used `Unknown` instead of `Never`.

With this change we don't need to use the first element as a guard against an illegal type conversion.

jjcnn (changes requested):

sway-ir/src/irtype.rs
```diff
@@ -197,12 +205,20 @@ impl Type {
             // Unions are special.  We say unions are equivalent to any of their variant types.
             (_, TypeContent::Union(_)) => other.eq(context, self),
             (TypeContent::Union(l), _) => l.iter().any(|field_ty| other.eq(context, field_ty)),
+            // Never type can coerce into any other type.
+            (TypeContent::Never, _) => true,
```
ironcev, line 209: I am not sure about this. `Never` can coerce into any other type but not other way around. Also, I am not sure if `eq` on the IR side is about coercion at all or if it checks strict equality of types. Before analyzing the usage, @vaivaswatha can you please take a look and give your view.

sway-ir/src/verify.rs
```diff
@@ -1017,7 +1017,7 @@ impl<'a, 'eng> InstructionVerifier<'a, 'eng> {
     // This is a really common operation above... calling `Value::get_type()` and then failing when
     // two don't match.
     fn opt_ty_not_eq(&self, l_ty: &Option<Type>, r_ty: &Option<Type>) -> bool {
-        l_ty.is_none() || r_ty.is_none() || !l_ty.unwrap().eq(self.context, r_ty.as_ref().unwrap())
+        l_ty.is_none() || r_ty.is_none() && !l_ty.unwrap().eq(self.context, r_ty.as_ref().unwrap())
```
ironcev, line 1020: Priority of operators will lead to this which is I assume not intended:
```
l_ty.is_none() || ( r_ty.is_none() && !l_ty.unwrap().eq(self.context, r_ty.as_ref().unwrap()) )
```
If `l_ty` is `Some` and `r_ty` is `None` the last operand will be executed leading to unwrapping of `r_ty` which is `None`.

ironcev (changes requested):

jjcnn (approved):

jjcnn:
> Okay sure, if that blocks some other work we can split it into another PR.

As it happens this issue blocks the one I'm currently working on, so it would be nice to have it merged asap.

PR #5613, opened 2024-02-15
Private Struct Fields for Blockchain Types

bitzoic:
## Description

Updates the following blockchain types to have private struct variables:
- `Address`
- `ContractId`
- `AssetId`

Note: This PR merges into a staging branch

## Checklist


K1-R1 (commented):
Changes look good, but with these fields now private how would these types be instantiated as constants or configureable constants?
Currently we are able to do:
```
configurable {
    SIGNER: Address = Address {
        value: ZERO_B256,
    },
}
```
Initialising with `Address::from(ZERO_B256)` would cause the error:
```
^^^^^^^^^^^^^^^^ Could not evaluate initializer to a const declaration.
```

sway-lib-std/src/address.sw
```diff
@@ -7,7 +7,29 @@ use ::hash::{Hash, Hasher};
...
+    ///
+    /// ```sway
+    /// use std::constants::ZERO_B256;
+    ///
+    /// fn foo() -> {
+    ///     let my_address = Address::from(ZERO_B256);
+    ///     assert(my_address.bits() == ZERO_B256);
+    /// }
+    /// ```
+    pub fn bits(self) -> b256 {
```
ironcev, line 30: ## TL;DR

If this is considered to be a type conversion, I would expect using `into()` or dedicated `as_xyz()` or `to_xyz()` methods like in Rust (`to_b256()` in this case).

If it is considered to be an attribute/property/field access, I would expect it to use established terminology, both in the field name and in the method name. E.g. `value()` or `raw_data()`.

## Detailed Reasoning

`bits` felt strange and at first I couldn't pin it down why, but now I know :smile: 

When it comes to type conversions, I would expect Rust approach, means either using `from()` and `into()` or explicit `as_xyz()` or `to_xzy()` methods.

If it is not to be considered a type conversion but as attribute/field/property accessor I would expect the already established terminology. Perhaps `bits` is already established terminology, but I assume it is not, assuming that  `value` was used so far. Also, if `bits` is expected terminology I somehow expect also the comment to use it by stating `The bits of the address.` Out of what I read from the code and the comment, if it is not to be treated as type conversion it should be either `value()` or `raw_data()`.

Additional benefit of naming it the same as the field is as @xunilrj pointed that we can give a better error message if someone tries to access the private field `value` - that the public `value()` could be used. We will have this improved error message in the future.

Just my two cents :smile:

bitzoic: `bits()` was originally selected as we refer to the `b256` as bits in the `From` implementation for `Address`, `ContractId`, and `AssetId` already. https://github.com/FuelLabs/sway/blob/65c19271070ade67d989e8d220f44f00bfa49e87/sway-lib-std/src/address.sw#L40

`bits()` is not a type conversion, we have a `From` implementation as shown above and this is a field accessor.

I don't have a strong opinion on the naming of this function and agree that naming it the same as the field should be prioritized for better error messages. Renaming `value` to `bits` could be another approach.

bitzoic:
> Changes look good, but with these fields now private how would these types be instantiated as constants or configureable constants? Currently we are able to do:
> 
> ```
> configurable {
>     SIGNER: Address = Address {
>         value: ZERO_B256,
>     },
> }
> ```
> 
> Initialising with `Address::from(ZERO_B256)` would cause the error:
> 
> ```
> ^^^^^^^^^^^^^^^^ Could not evaluate initializer to a const declaration.
> ```

What version of forc are you using to test this? I've added tests to add `Address` and `ContractId` using `From` implementation in configurable in [this](https://github.com/FuelLabs/sway/pull/5613/commits/3b3fd5be1f0dd56f824d372426f2363802e72550) commit.

sway-lib-std/src/address.sw
ironcev: Ok now all makes sense. Further renaming `value` to `bits` would make the terminology explicit and improve the error handling of private fields in the future so I would do that as well.

K1-R1:
> What version of forc are you using to test this? I've added tests to add `Address` and `ContractId` using `From` implementation in configurable in [this](https://github.com/FuelLabs/sway/pull/5613/commits/3b3fd5be1f0dd56f824d372426f2363802e72550) commit.

I tried it with forc = "0.50.0". If it's working on this PR though it's a non-issue.

SwayStar123 (approved):

Issue #5622, opened 2024-02-16
Refactor `compile_array_expr` in `function.rs`.


esdrubal:
Now that `convert_resolved_type` doesn't throw an error on diverging expressions we can simplify `compile_array_expr` in `function.rs`. It currently generates code for the first array element before creating the array, because creating the array requires the element type to be converted which wasn't possible when we used `Unknown` instead of `Never`.

With this change we don't need to use the first element as a guard against an illegal type conversion.

_Originally posted by @jjcnn in https://github.com/FuelLabs/sway/pull/5602#discussion_r1487566793_

PR #5631, opened 2024-02-19
Fix ICE in relation to local declarations with erroneous initializers

### LINKED ISSUE #5616: Internal Compiler Error when evaluating nonsensical `const`
Swapping `let` for `const` in places can lead to an ICE among other errors.

The following minimal example:

```sway
contract;

abi Abi1 {
    fn foo() -> b256;
}

abi Abi2{
    fn bar() -> u64;
}

pub fn main() {
    let contract_1 = abi(Abi1, b256::min());
    const invalid_const = contract_1.foo();
    let contract_2 = abi(Abi2, invalid_const);
    let invalid = contract_2.bar();
    log(invalid);
}

```

Produces:

```
error: Internal compiler error: Unable to resolve variable 'invalid'.
```

jjcnn:
## Description

Fixes #5616 and #5622.

When generating the IR for constant and variable declarations the name being declared is not added to the local environment if the initializer cannot be compiled. This can happen when initializing a constant using a non-constant expression.

Since the name isn't added to the environment, and since we attempt to continue compilation after an error, any attempt to use the name later in the code will cause the name to not be found, which in turn causes the IR converter to throw an ICE because it assumes that all used names are in scope.

This PR fixes the issue by not throwing the initializer error until the name has been added to the environment. Note that
1. The initializer must be compiled before the name is added to the environment. The name should not be in scope during its own initialization.
2. #5602 is blocking a more elegant solution to this problem. The lack of a `Never` type means that the call to `convert_resolved_typeid` may fail if the initializer diverges, and since that call must be made before the new name can be added to the environment we check for termination first. The order of operations is therefore currently a) compile the initializer, then b) check that the initializer did not result in an error and did not diverge, then c) resolve the type of the name being declared, then d) add the name to the environment, and finally e) throw an error if the initializer resulted in an error. Once the `Never` type is introduced we can move the terminator check to the end of that sequence.

## Checklist


IGI-111:
#5602 has been merged, so we should be able to get the cleaner solution

jjcnn:
The code has now been simplified based on the introduction of the `Never` type.

jjcnn:
> #5602 has been merged, so we should be able to get the cleaner solution

Actually, I can't. We do have the `Never` type now, but array expressions are still typed using `Unknown` in place of `Never` - I thought that was going to be fixed by the `Never` type PR, but apparently it hasn't.

I've been trying to fix the problem in the typechecker, but something is going wrong and I don't know what it is. If I just change the expected type to `Never`, then I get a bunch of type errors in the core library.

I've also tried doing what we ought to do, which is to determine the type of the array based on the unification of the inferred element types, but then I run into problems with Rust's ownership semantics.

jjcnn:
Fixed. The typechecker still assigned the `Unknown` type to `return`, `continue` and `break` expressions, which screwed things up when generating code for array expressions.

Once they were assigned the `Never` type instead everything works out.

The PR is ready for a final review.

IGI-111 (dismissed):

IGI-111 (approved):

Issue #5597, opened 2024-02-12
Internal compiler error with dereferencing, `Option`, and type ascription


ironcev:
This code fails with `Internal compiler error: Verification failed: Struct field type mismatch.`.

```
script;
 
fn main() {
  let _: Option<u8> = *&Option::Some(123);
}
```

The reason is that type-checking of dereferencing wrongly removes the type annotation from the `TypeCheckContext` by setting it to `TypeInfo::Unknow` instead of passing the expected referenced type as type annotation.

ironcev:
Closed in #5643.

Issue #5581, opened 2024-02-08
Internal compiler error with structs, `Option`, and type ascription


ironcev:
This code fails with `Internal compiler error: Verification failed: Struct field type mismatch.`.

```Sway
script;
 
struct S<T> {
    x: T,
}

fn main() {
    let _: S<Option<u8>>  = S { x: Option::Some(1) };
}
```

ironcev:
Closed in #5643.

Issue #5492, opened 2024-01-17
"Verification failed: Struct field type mismatch" compiler error


dmihal:
The following code is generating the error "Internal compiler error: Verification failed: Struct field type mismatch."

```rust
contract;

use std::{hash::Hash};

storage {
    total_supply: StorageMap<AssetId, u64> = StorageMap {},
}

abi SRC20 {
    #[storage(read)]
    fn decimals(asset: AssetId) -> Option<u8>;
}

impl SRC20 for Contract {
    #[storage(read)]
    fn decimals(asset: AssetId) -> Option<u8> {
        match storage.total_supply.get(asset).try_read() {
            Option::Some(_supply) => Option::Some(0),
            Option::None => Option::None,
        }
    }
}
```

Compiled with `forc 0.48.1+nightly.20231226.a676a3d981`

vaivaswatha:
Simplified examples

```
script;

fn main() {
  let _ = bar(true);
}

#[inline(never)]
fn bar(b: bool) -> Option<u8> {
   if(b) {
     Option::Some(19)
   } else {
     Option::None
   }
}
```

```
script;

fn main() {
  let _ = foo();
}

#[inline(never)]
fn foo() -> Option<u8> {
  match Some(true) {
    Option::Some(_b) => Option::Some(17),
    Option::None => Option::None,
  }
}
```

ironcev:
Blocked by #5562 (only the first two tasks are needed to properly solve this issue).

If urgent, we can provide a workaround and fix this issue before getting the first two tasks of #5562 done.

The workaround would suppress the type unification in case of a `Never` type when type checking `match` branches.

This should solve the issue, but it is an ugly hack in the compiler code. If the issue is not urgent, I am for waiting until the first two tasks of #5562 are done.

ironcev:
Closed in #5643.

Issue #5606, opened 2024-02-13
Wrong type mismatch errors after a mismatched type in `TyExpression::type_check()` that are not reported as `Err`


ironcev:
Currently, if the `TyExpression::type_check()` fails to unify the resulting `TyExpression` with the expected `ctx.type_annotation()` the `TypeError::MismatchedType` will be emitted, but the `TyExpression::type_check()` will still return `Ok(typed_expression)` signaling to the callers that everything went fine.

This leads to issues like the following ones, where the resulting expression of the wrong `result_type` is taken over into the next type-checking step:

```Sway
const X: bool = 123u64;
//              ^^^^^^ Mismatched types. **OK!**
let _ = X == true;
//           ^^^^ This parameter was declared as type u64, but argument of type bool was provided. **WRONG!!**
```
```Sway
let x: bool = 123u64;
//            ^^^^^^ Mismatched types. **OK!**
let _ = x == true;
//           ^^^^ This parameter was declared as type u64, but argument of type bool was provided. **WRONG!!**
```
```Sway
let o: Option<u8> = Option::Some(123u64);
//                               ^^^^^^ Mismatched types. **OK!**
match o {
    Some(x) => x == 123u8,
//                  ^^^^^ Mismatched types.
//                        expected: u64               **WRONG!!**
//                        found:    u8.
    _ => false,
};
```
These are only some examples, similar issues can be found in other kind of expressions.

My assumption is that `TyExpression::type_check()` should return `Err` if the unification fails.

khanhanmeme:
done

PR #5643, opened 2024-02-22
Fix type check and inference issues in references, structs, and enums

### LINKED ISSUE #5559: Internal compiler error with references, `Option`, and type ascription
This code fails with `Internal compiler error: Verification failed: Struct field type mismatch.`.

```Sway
script;
 
fn main() {
  let r: &Option<u8> = &Option::Some(1);
}
```

The reason is the fact that type-checking of references wrongly removes the type annotation from the `TypeCheckContext` by setting it to `TypeInfo::Unknow` instead of extracting the referenced type and passing it as type annotation.

### LINKED ISSUE #5583: Incorrect error messages in enum instantiation
The following example generates incorrect error messages:

```Sway
script;

enum MyOption {
    Some: bool,
    None: (),
}

enum MyOtherOption {
    NoSome: bool,
    None: (),
}

main() {
    let _: MyOption = Option::Some(123);
           //                      ^^^ Mismatched types.
           // expected: bool      // WRONG: Expectation is coming from a type that is not `Option`.
           // found:    numeric.
           // help: Enum instantiator must match its declared variant type.

    let _: MyOtherOption = Option::Some(123);
           //                      ^^^^ Variant "Some" does not exist on enum "MyOtherOption"
           // WRONG: We are type-checking `Option` and `Some` do exist.

    let _: Option<u8> = Option::Some(123u64);
          //            ^^^^^^^^^^^^^^^^^^^^ Mismatched types.
          // expected: u8   // WRONG: Should be `Option<u8>`.
          // found:    u64.   // WRONG: Should be `Option<u64>`.
          // help: Variable declaration's type annotation does not match up with the assigned expression's type.

}
```

The reason for the first two misleading messages is taking the LHS enum as the type annotation source for the RHS without checking that it is the same enum type.

### LINKED ISSUE #5598: Wrong type unification of enums and structs in case of name and structural equality
This example compiles, treating the two enums named `E` as the same type, and two struct `S` respectively.

```
// lib_a.sw
library;
pub enum E { X: u8, }
pub struct S { pub x: u8, }

...

// lib_b.sw
library;
pub enum E { X: u8, }
pub struct S { pub x: u8, }

...

// main.sw
script; 

mod lib_a;
mod lib_b;

use ::lib_a::E as Lib_A_E;
use ::lib_a::S as Lib_A_S;

use ::lib_b::E as Lib_B_E;
use ::lib_b::S as Lib_B_S;

fn main() {
  let _: Lib_A_E = Lib_B_E::X(123);     // Two different enums unified.
  let _: Lib_A_S = Lib_B_S { x: 123 };  // Two different structs unified.
}
```

ironcev:
## Description

This PR:
- fixes #5559, #5597, and #5492 by removing the `TypeInfo::Unknown` and providing the required contextual information to type checking of referencing, dereferencing, `if`, and `match` expressions respectively. The contextual information provided is taken from the `ctx.type_annotation()` but always adapted according to the semantics of the type-checked expression.
- fixes #5583 and #5581 by combining the contextual information coming from the `ctx.type_annotation()` with the one coming from the enum and struct instantiation and declaration.
- fixes #5598 by forcing the name-based and not structure-based identity. In other words, two enums or structs are considered equal only if they whole `call_path`s are equal. Up to now, we were expecting only the enum or struct _names_ to be equal, which was treating types with same names and structures (variants or fields) as equal although they were defined in different modules.

The PR also introduces parsing of references to mutable values (`&mut T`). Since this addition does not overlap with the above bug fixes, it was left as is, and can be fully ignored during the review. Other changes related to references to mutable values are removed from the code from type-checking onward to make this PR only fixing the issues. Continuation on references to mutable values will be done in a separate PR.

Closes #5559, #5597, #5492, #5583, #5581, #5598.

## Checklist


sway-core/src/semantic_analysis/ast_node/declaration/declaration.rs
```diff
@@ -54,14 +54,29 @@ impl TyDecl {
...
+                //       an integer, RHS should properly unify or type check should fail.
+                //       Remove this special case as a part of the initiative of improving type inference.
                 // Integers are special in the sense that we can't only rely on the type of `body`
                 // to get the type of the variable. The type of the variable *has* to follow
                 // `type_ascription` if `type_ascription` is a concrete integer type that does not
                 // conflict with the type of `body` (i.e. passes the type checking above).
                 let return_type = match &*type_engine.get(type_ascription.type_id) {
                     TypeInfo::UnsignedInteger(_) => type_ascription.type_id,
-                    _ => body.return_type,
+                    _ => match &*type_engine.get(body.return_type) {
```
xunilrj, line 67: I am finding this part a little bit confused. 

If a variable declaration has type ascription, the `TyVarDecl` below should always use the type ascription of the variable.

And after that, we need to do is unify and check the type ascription with the return of the init expression and give the best error/warning possible, right?

That will also "kill" the special case for `ints` here.

ironcev: Yes you are right. But at this stage I was already overwhelmed with the PR that took way too long exactly because I was tracking issues that were popping up and not limiting the scope of this PR.

Knowing that if I change the current behavior, there will very likely be additional errors popping from the remaining cases of expressions that can be on the RHS but do not comply with the provided `ctx.type_annotation()` but rather use `TypeInfo::Unknown` internally, I've opted for just writing the above TODOs at the moment and let this be solved properly as a part of #5562 step 4. Also, the `TyExpression::type_check()` currently do not return `Err` if the unification fails (#5606), etc.

That's why I fear touching this as a part of this PR could lead to additional effort hard to estimate upfront. And since this PR is actually blocking for #5562 I am still of the opinion to solve this particular case as a part of the #5562 step 4. 

Is this ok for you?

IGI-111 (approved):
LGTM insofar as we can resolve those TODOs in a later PR

sway-core/src/semantic_analysis/ast_node/declaration/declaration.rs
ironcev: Silence implies consent :smile: so I am resolving this remark. It will be done as part of #5562.

xunilrj (approved):

Issue #4737, opened 2023-07-03
Move `sysinfo` dependency behind a feature flag


kayagokalp:
Since `sysinfo` is only needed for benchmarks we should consider moving it behind a flag. Since it causes this annoying bug we are dealing with in fuel.nix: https://github.com/FuelLabs/fuel.nix/issues/64.

PR #5676, opened 2024-02-28
Implement Pid file locking

### LINKED ISSUE #5633: Handle removing the lsp-lock file in cases of non graceful IDE shutdown
This lock file was originally introduced in #5094

While it does solve the original issue in  #4893 , it can lead to problems when the IDE was shutdown in a non-graceful manner, leaving behind the old lock file. This is the report from Slack:

> forc fmt refuses to format claiming that some files are open in an editor. I'm using nvim. Thought it was looking for swap files but there were none, everything was saved cleanly.
> Turns out the sway lsp marked the file as dirty using a file as a lock over in "$HOME/.forc/.lsp_locks" or something similar. Had to manually remove the file.
> Don't know how robust the lock file removal process is in cases of a ungraceful shutdown of the sway lsp.
> In any case it took some digging around the code to figure out what to remove.
> Maybe it would cause less headaches going forward to lower this from a refusal to format down to a warning?

crodas:
## Description

Fixes #5633

Added functionality to create and manage a lock file containing the process ID (pid) of the running instance of the software. This mechanism prevents multiple instances of the software from running simultaneously by checking the existence and content of the lock file. If the lock file exists and contains a valid pid, the struct will error gracefully to avoid conflicts. If the lock file is missing or contains an invalid pid, the struct will proceed by removing the file.  This ensures that only one instance of the software can run at a time and it avoids stale locking to prevent future instances

## Checklist


forc-util/src/fs_locking.rs
```diff
@@ -0,0 +1,143 @@
...
+    io::{self, Read, Write},
+    path::{Path, PathBuf},
+};
+
+/// Very simple AdvisoryPathMutex class
+///
+/// The goal of this struct is to signal other processes that a path is being used by another
+/// process exclusively.
+///
+/// This struct will self-healh if the process that locked the file is no longer running.
```
sdankel, line 13:
```suggestion
/// This struct will self-heal if the process that locked the file is no longer running.
```

forc-util/src/fs_locking.rs
```diff
@@ -0,0 +1,143 @@
...
+pub struct PidFileLocking(PathBuf);
+
+impl PidFileLocking {
+    pub fn new<X: AsRef<Path>, Y: AsRef<Path>>(path: X, dir: Y, extension: &str) -> PidFileLocking {
```
sdankel, line 17: The path here is actually a filename right?

```suggestion
    pub fn new<X: AsRef<Path>, Y: AsRef<Path>>(filename: X, dir: Y, extension: &str) -> PidFileLocking {
```

forc-util/src/fs_locking.rs
```diff
@@ -0,0 +1,143 @@
...
+        assert!(metadata(&x.0).is_ok());
+
+        // simulate a stale lock file
+        let mut x = File::create(&x.0).unwrap();
+        x.write_all(b"191919191919").unwrap();
+        x.flush().unwrap();
+        drop(x);
+
+        // PID=191919191919 does not exists, hopefully, and this should remove the lock file
+        let x = PidFileLocking::lsp("test");
```
sdankel, line 138: Shouldn't this use the same file name ("stale") to test that it's overwriting the stale file?

forc-util/src/fs_locking.rs
```diff
@@ -0,0 +1,143 @@
...
+#[cfg(test)]
+mod test {
+    use super::PidFileLocking;
+    use std::{
+        fs::{metadata, File},
+        io::{ErrorKind, Write},
+    };
+
+    #[test]
+    fn same_process() {
```
sdankel, line 115:
```suggestion
    fn fs_locking_same_process() {
```

forc-util/src/fs_locking.rs
```diff
@@ -0,0 +1,143 @@
...
+        let x = PidFileLocking::lsp("test");
+        assert!(x.lock().is_ok());
+        // The current process is locking "test"
+        let x = PidFileLocking::lsp("test");
+        assert!(!x.is_locked().unwrap());
+    }
+
+    #[test]
+    fn stale() {
```
sdankel, line 124:
```suggestion
    fn fs_locking_stale() {
```

sway-lsp/src/core/document.rs
```diff
@@ -113,39 +113,31 @@ impl TextDocument {
...
 /// If the flag file does not exist, this function will do nothing.
 pub async fn remove_dirty_flag(uri: &Url) -> Result<(), LanguageServerError> {
     let path = document::get_path_from_url(uri)?;
-    let dirty_file_path = forc_util::is_dirty_path(&path);
-    if dirty_file_path.exists() {
-        // Remove the "dirty" file
-        tokio::fs::remove_file(dirty_file_path)
-            .await
+    let uri = uri.clone();
+    tokio::task::spawn_blocking(move || {
```
sdankel, line 131: Same question here

forc-util/src/fs_locking.rs
```diff
@@ -0,0 +1,153 @@
...
+            return false;
+        }
+        System::new_all().process(Pid::from(pid)).is_some()
+    }
+
+    /// Removes the lock file if it is not locked or the process that locked it is no longer active
+    pub fn remove(&self) -> io::Result<()> {
+        if self.is_locked()? {
+            Err(io::Error::new(
+                std::io::ErrorKind::Other,
```
sdankel, line 50: Maybe `std::io::ErrorKind::ResourceBusy`?

forc-util/src/fs_locking.rs
```diff
@@ -0,0 +1,153 @@
...
+                "Cannot remove a dirty lock file, it is locked by another process",
+            ))
+        } else {
+            self.remove_file()
+        }
```
sdankel, line 55: `is_locked` removes the file if it's not locked, so this is redundant.

```suggestion
        }
        Ok(())        
```

forc-util/src/fs_locking.rs
```diff
@@ -0,0 +1,153 @@
...
+    fn test_fs_locking_same_process() {
+        let x = PidFileLocking::lsp("test");
+        assert!(x.lock().is_ok());
+        // The current process is locking "test"
+        let x = PidFileLocking::lsp("test");
+        assert!(!x.is_locked().unwrap());
+    }
+
+    #[test]
+    fn test_fs_locking_stale() {
```
sdankel, line 134: Might be good to test these cases too:
- empty file
- non-existant file

sway-lsp/src/core/document.rs
```diff
@@ -113,39 +113,31 @@ impl TextDocument {
...
-    }
-    // Create an empty "dirty" file
-    File::create(&dirty_file_path)
-        .await
-        .map_err(|err| DocumentError::UnableToCreateFile {
-            path: uri.path().to_string(),
-            err: err.to_string(),
-        })?;
-    Ok(())
+    tokio::task::spawn_blocking(move || {
```
sdankel, line 116: Why is this done in a separate thread?

forc-util/src/fs_locking.rs
```diff
@@ -0,0 +1,153 @@
...
+                .join(dir)
+                .join(file_name)
+                .with_extension(extension),
+        )
+    }
+
+    /// Create a new PidFileLocking instance that is shared between the LSP and any other process
+    /// that may want to update the file and needs to wait for the LSP to finish (like forc-fmt)
+    pub fn lsp<X: AsRef<Path>>(path: X) -> PidFileLocking {
+        Self::new(path, ".lsp-locks", "dirty")
```
sdankel, line 34: nit:

```suggestion
    pub fn lsp<X: AsRef<Path>>(filename: X) -> PidFileLocking {
        Self::new(filename, ".lsp-locks", "dirty")
```

forc-util/src/fs_locking.rs
```diff
@@ -0,0 +1,153 @@
...
+/// process exclusively.
+///
+/// This struct will self-heal if the process that locked the file is no longer running.
+pub struct PidFileLocking(PathBuf);
+
+impl PidFileLocking {
+    pub fn new<X: AsRef<Path>, Y: AsRef<Path>>(
+        filename: X,
+        dir: Y,
+        extension: &str,
```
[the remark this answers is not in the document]
JoshuaBatty: I think it's useful to have something descriptive. either `.dirty` | `.fmt_lock` | '.lock`

forc-util/Cargo.toml
```diff
@@ -27,6 +27,7 @@ sway-core = { version = "0.51.1", path = "../sway-core" }
 sway-error = { version = "0.51.1", path = "../sway-error" }
 sway-types = { version = "0.51.1", path = "../sway-types" }
 sway-utils = { version = "0.51.1", path = "../sway-utils" }
+sysinfo = "0.30.5"
```
JoshuaBatty, line 30: Unfortunately, the sysinfo crate is problematic for macos builds. See here #4737

Perhaps it's not an issue anymore but it was causing the nix builds to break. Might be worth looking into either clarifying this isn't the case anymore or finding an alternative approach.

sway-lsp/src/core/document.rs
JoshuaBatty: I don't think we should drop async-io as io is one of the perfect use cases for async. Instead of dropping the async I/O model, we could consider abstracting the async-to-sync conversion to make our code cleaner and more maintainable. something like...

```rust
async fn lock_lsp_path_async(path: PathBuf) -> Result<(), DirectoryError> {
    tokio::task::spawn_blocking(move || {
        PidFileLocking::lsp(path)
            .lock()
            .map_err(|_| DirectoryError::LspLocksDirFailed)
    })
    .await
    .map_err(|_| DirectoryError::LspLocksDirFailed)
}
```

JoshuaBatty (changes requested):
Off to a good start, but it would be good to address my concern about the sysinfo crate before this lands.

sway-lsp/src/core/document.rs
JoshuaBatty: I think I misunderstood. The `mark_file_as_dirty` function can probably be written like. 

```rust
pub fn mark_file_as_dirty(uri: &Url) -> Result<(), LanguageServerError> {
    let path = document::get_path_from_url(uri)?;
    Ok(PidFileLocking::lsp(path)
        .lock()
        .map_err(|_| DirectoryError::LspLocksDirFailed)?)
}
```

JoshuaBatty: No need for this function to be async if none of the internal code needs awaiting

forc-util/src/fs_locking.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    /// Checks if the given pid is active
+    #[cfg(not(target = "windows"))]
+    fn is_pid_active(pid: usize) -> bool {
+        use std::process::Command;
+        let output = Command::new("ps")
+            .arg("-p")
+            .arg(pid.to_string())
+            .output()
+            .expect("Failed to execute ps command");
```
sdankel, line 45: We can use the [sysinfo](https://docs.rs/sysinfo/latest/sysinfo/struct.System.html#method.processes) library for this

sdankel: Might be good to add a code comment to that effect, like
```
                // Not using sysinfo here because it has compatibility issues with fuel.nix
                // https://github.com/FuelLabs/fuel.nix/issues/64
```

JoshuaBatty (approved):

sdankel (approved):

PR #5678, opened 2024-03-02
feat: add log type info collection for test entries to enable log decoding at forc-test

kayagokalp:
## Description
Adds necessary capabilities to forc-pkg to collect types of `log`s used in a test. Type information is retrieved from the type engine, and collected for each test entry. This will later on be used for decoding the log receipts at forc-test.


forc-pkg/src/pkg.rs
```diff
@@ -2011,16 +2014,55 @@ impl PkgTestEntry {
...
+            } = &expr.expression
+            {
+                let suffix = call_path.suffix.to_string();
+                if suffix == "log" {
+                    let arg_type = arguments
+                        .iter()
+                        .map(|(_, arg_expr)| arg_expr.return_type)
+                        .map(|type_id| engines.te().get(type_id))
+                        .next();
+                    log_param_types.push((expr.span.source_id().cloned().unwrap(), arg_type));
```
sdankel, line 2040: Should we only store the `arg_type`s that are Some?

forc-pkg/src/pkg.rs
```diff
@@ -2011,16 +2014,55 @@ impl PkgTestEntry {
...
+        }
+
         let file_path = Arc::new(
             engines.se().get_path(
                 span.source_id()
                     .ok_or_else(|| anyhow::anyhow!("Missing span for test function"))?,
             ),
         );
+
+        println!("{log_params:?}");
```
sdankel, line 2060:
```suggestion
```

forc-pkg/src/pkg.rs
```diff
@@ -2774,9 +2814,54 @@ mod test {
...
+            .log_params
+            .iter()
+            .map(|(_, type_info)| type_info)
+            .collect();
+        assert_eq!(log_params.len(), 1);
+        let type_info = log_params[0].clone().unwrap();
+
+        assert!(matches!(
+            type_info.as_ref(),
+            TypeInfo::UnsignedInteger(IntegerBits::SixtyFour)
```
sdankel, line 2856: Is it working for other types? If so we should test them (other numeric types, bool, string, array, tuple, struct, enum).

forc-pkg/src/pkg.rs
```diff
@@ -2011,16 +2014,55 @@ impl PkgTestEntry {
...
+                    log_param_types.push((expr.span.source_id().cloned().unwrap(), arg_type));
+                }
+            }
+        }
+
+        let mut log_params = Vec::new();
+        for ast_node in test_function_decl.body.contents.iter() {
+            if let TyAstNodeContent::Expression(expr) = &ast_node.content {
+                // We found an expression node, try to extract the arguments if this is an assert.
+                collect_log_type_from_decl(expr, &mut log_params, engines);
```
sdankel, line 2049: nit: imho, it would be easier to read if the contents of this function were inline here, unless it's going to be called from somewhere else. Having the helper function right above confused me a little at first because it looked like it was calling itself recursively.

sdankel (commented):
Awesome work! Just a few nits.

forc-pkg/src/pkg.rs
kayagokalp: Yeah, it was recursive at first i designed it, thinking that log can accept multiple params. It looks like it cant i will inline it

forc-pkg/src/pkg.rs
kayagokalp: Yes we should add more test but i think exhaustively trying all the types would go out of scope because:

1. imo at that point we are testing type engine (which should be tested on its own) not this function, as types itself are not coming from this new function but from type engine itself
2. There are a lot of types possible :)

I agree that it makes sense to add more tests that, maybe we can start with one that collects multiple logs that are in different types so that we can test if we can collect them in order?

kayagokalp: On a second thought I am thinking about adding similar tests for:

1. other numeric types, 
2. bool,
3. string,
4. array,
5. tuple, 
6. struct,
7.  enum,

and a function based one that returns one of these types
and one that returns multiple log types

I think this is not exhaustive but common types that will be encountered with logs, also it will make sure that we don't rely on type engine tests as that might not be exhaustive either.

sdankel: Having a test for multiple log types is a good idea! I agree it doesn't need to be exhaustive. It's good to test as much as we can within reason to find bugs that would otherwise surface to users of forc-test.

sdankel (approved):
Nice work! 💯

JoshuaBatty (approved):
LGTM

kayagokalp:
closed in favor of #5812

PR #5685, opened 2024-03-04
Expose external code loading with `std::execution::run_external`

### LINKED ISSUE #1846: Add a module to the stdlib to expose `LDC` opcode
Many common patterns will require this functionality. A few examples include:
- upgradeable contracts
- proxies

I thought we would also expose `SLDC` but found this issue to remove that instruction in the spec: https://github.com/FuelLabs/fuel-specs/issues/313

### LINKED ISSUE #5691: Configurables get elided if only used in LDC
For #5579 we want to let devs build proxies that may look like this:

```sway
contract;

use std::execution::run_external;

configurable {
    TARGET: ContractId = ContractId::from(ZERO_B256),
}

abi RunExternalTest{
    fn foobar() -> u64;
}

impl RunExternalTest for Contract {
    fn foobar() -> u64 {
        run_external(TARGET)
    }
}

fn run_external(load_target: ContractId) -> ! {
    asm(load_target, word, length, ssp_saved) {
        lw load_target fp i74;
        csiz length load_target;
        move ssp_saved ssp;
        ldc load_target zero length;
        addi word zero i64;
        aloc word;
        sw hp ssp_saved i0;
    }
    __jmp_mem()
}
```

However, because `load_target` isn't used in the result of any method of the contract, the `TARGET` configurable gets optimized away, which makes proper configuration impossible.

This still happens if the return value of `run_external` isn't `!`.

We need to either find a way of telling the compiler that `load_target` is consumed by the jump, or fix a bug in the analysis.

IGI-111:
## Description

Fixes #1846
Fixes #5691

## Checklist


xunilrj (approved):

xunilrj (dismissed):

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
```diff
@@ -287,7 +287,19 @@ where
             _ => todo!(),
         };
 
-        assert!(!handler.has_errors(), "{:?} {}", handler, code);
+        if handler.has_errors() {
+            assert!(
+                !handler.has_errors(),
```
ironcev, line 292: This can simply be `false`.

sway-lib-std/src/execution.sw
```diff
@@ -0,0 +1,31 @@
+library;
+
+use ::contract_id::ContractId;
+
+/// Load and run the contract with the provided `ContractId`
```
ironcev, line 5: Nitpick. In other docs we always have `.` at the end.
```suggestion
/// Load and run the contract with the provided `ContractId`.
```

sway-types/src/source_engine.rs
```diff
@@ -123,4 +123,9 @@ impl SourceEngine {
         v.sort();
         v
     }
+
+    pub fn get_source_id_from_module_id(&self, module_id: ModuleId) -> Option<BTreeSet<SourceId>> {
```
ironcev, line 127: `BTreeSet` confused me at first, because I've expected a single `SourceId` based on the method name. Renaming it to plural will remove the confusion and precisely convey that a module can consist of a more then one file.
```suggestion
    pub fn get_source_ids_from_module_id(&self, module_id: ModuleId) -> Option<BTreeSet<SourceId>> {
```

test/src/sdk-harness/test_projects/abi_impl_methods_callable/Forc.toml
```diff
@@ -4,3 +4,6 @@ authors = ["Fuel Labs <contact@fuel.sh>"]
 entry = "main.sw"
 license = "Apache-2.0"
 implicit-std = false
```
ironcev, line 6: Can be removed because it gets overridden by the `core` in `[dependencies]`.

test/src/sdk-harness/test_projects/run_external_proxy/.gitignore
```diff
@@ -0,0 +1,2 @@
+out
```
ironcev, line 1: This `.gitignore` can also be removed.

test/src/sdk-harness/test_projects/run_external_target/.gitignore
```diff
@@ -0,0 +1,2 @@
+out
```
ironcev, line 1: This one also.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
```diff
@@ -298,7 +310,20 @@ where
         )
         .unwrap();
 
-        assert!(!handler.has_errors(), "{:?} {}", handler, code);
+        if handler.has_errors() {
+            assert!(
+                !handler.has_errors(),
```
ironcev, line 315: Same here.

ironcev (commented):
Few nitpicks.

A curious question, what happens in case of mismatch in method signatures of the proxy and the target? Should we also add negative tests?

IGI-111:
Unfortunately, type mismatch in loaded code arguments is undefined behavior, and there isn't really a lot we can do about that.

I would want us to introduce some form of static checking later if we can, but the best practice here is going to be for the programmers to implement the same ABI (defined by a separate library ideally) in their proxy and target.

test/src/sdk-harness/test_projects/abi_impl_methods_callable/Forc.toml
IGI-111: I'd like to keep this one for correctness' sake in case we change the dependencies.

vaivaswatha (approved):

ironcev (approved):

PR #5702, opened 2024-03-06
feat: add clear to storage vec

### LINKED ISSUE #5701: Add clear() to StorageVec
There is no clear for StorageVec

matt-user:
## Description
closes #5701 

## Checklist

I didn't see any StorageVec tests, but I can add a test if wanted.


sway-lib-std/src/storage/storage_vec.sw
```diff
@@ -876,6 +876,32 @@ impl<V> StorageKey<StorageVec<V>> {
...
+    /// fn foo() {
+    ///     storage.vec.push(1_u64);
+    ///     storage.vec.push(2_u64);
+    ///     storage.vec.push(3_u64);
+    ///     storage.vec.clear_vec();
+    ///     assert!(storage.vec.is_empty());
+    /// }
+    /// ```
+    #[storage(write)]
+    pub fn clear_vec(self) {
```
sdankel, line 901: Could we call it simply `clear()`?

sdankel (commented):
It would be good to have unit tests for this function, and for `StorageVec` in general.

ironcev:
> It would be good to have unit tests for this function, and for `StorageVec` in general.

We have existing tests for `StorageVec` in _/test/src/sdk-harness/test_artifacts/storage_vec_. Agree that `clear()` should also be tested there.

sway-lib-std/src/storage/storage_vec.sw
```diff
@@ -876,6 +876,32 @@ impl<V> StorageKey<StorageVec<V>> {
...
+    /// # Examples
+    /// ```sway
+    /// use std::storage::storage_vec::*;
+    ///
+    /// storage {
+    ///     vec: StorageVec<u64> = StorageVec {}
+    /// }
+    ///
+    /// fn foo() {
+    ///     storage.vec.push(1_u64);
```
ironcev, line 893: Nitpick. In all the doc examples for all the methods where we have `storage.vec.push` we do not set the type `_u64` explicitly. Therefore, I am for removing it here as well and have simply `storage.vec.push(1)` etc.

I've noticed three places where `let five = 5_u64;` is used. It is not a part of this PR but it's great to have docs consistent so I would go here also for `let five = 5;`.

A nitpick :smile:

ironcev (changes requested):

IGI-111 (changes requested):
This needs tests.

matt-user:
@ironcev I was unable to name the function clear because it gave a compile error of duplicate clear definitions with the clear function defined in `StorageKey` .  Is there another way to solve this compile error?

ironcev:
> I was unable to name the function clear because it gave a compile error of duplicate clear definitions with the clear function defined in `StorageKey`.

Indeed! A bit of a background info, current version of the storage (v2.0) operates with `StorageKey`s in the background. Essentially, every:

```Sway
storage {
    x: u8 = 0,
    v: StorageVec<u8> = StorageVec::<u8> { }
}
```
is actually:
```Sway
storage {
    x: StorageKey<u8> = 0,
    v: StorageKey<StorageVec<u8>> = StorageVec::<u8> { }
}
```

The compiler magic turns in case of same method names indeed into the error that Matt is talking about:

```
#[storage(write)]
pub fn clear(self) {
           ^^^^^ Duplicate definitions for the method "clear" for type "StorageKey<StorageVec<V>>".
```

@IGI-111 In our new concept for the storage v3.0 where everything will be `Storage` this should never happen.

@matt-user FYI we are working on a new (better, mightier, ...) version of the storage API.

The question is what to do until we get Storage v3.0?

I don't feel comfortable with `clear_vec`. As explained, it breaks the guidelines for method names and now after Matt explained it is obvious that it's a workaround and not an intention.

The possible approaches I see:
- We do start with `clear_vec` and rename it to `clear` once v3.0 is out. Means dealing with a breaking change.
- We insist on `clear` and try to fix the issue in Sway compiler. Potentially high effort for something we anyhow see as temporary and want to change soon.
- We pause this PR for the time being and introduce `clear` once v3.0 is done.

If `clear` is not urgent and heavily demanded I would go with the third option.

@bitzoic  @IGI-111 What do you think?

IGI-111:
Hmm given we're going to rework storage anyways, I think it doesn't hurt much to call this `clear_vec` for now and merge this once it has some tests.

It'll improve the functionality of what's there, and there's precedent for the naming in some of the other `StorageVec` methods. A proper solution will have to wait for v3.

bitzoic:
`clear()` on a `StorageVec` should work as `StorageVec` implements `StorageKey` and `StorageKey` already has `clear()`, hence the name collision. A while ago I added the case that if the generic has a size of zero, aka empty structs like `StorageVec` and `StorageMap`, to clear the field id storage location over the slot. This can be seen [here](https://github.com/FuelLabs/sway/blob/421caefe6c8241b75488f8e515ec41544a35d888/sway-lib-std/src/storage/storage_key.sw#L124).

matt-user:
closing bc as @bitzoic  mentioned this is handled by StorageKey

PR #5238, opened 2023-10-28
feat: Add predicate and script templates

calldelegation:
## Description

When DevRel creates more tutorials on predicates and scripts, it's not ideal to use the same template designed for contracts and ask developers to erase everything and start afresh. It's clearer to have developers use a template tailored for the respective program type. 

These templates are already compatible with existing `forc` commands like 

Predicate
1. `forc new --predicate MyPredicate`
2. `cargo generate --init fuellabs/sway templates/sway-predicate-test-rs --name MyPredicate`

Script
1. `forc new --script MyScript`
2. `cargo generate --init fuellabs/sway templates/sway-script-test-rs --name MyScript`

## Checklist

- [x] I have linked to any relevant issues.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation where relevant (API docs, the reference, and the Sway book).
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have added (or requested a maintainer to add) the necessary `Breaking*` or `New Feature` labels where relevant.
- [x] I have done my best to ensure that my PR adheres to [the Fuel Labs Code Review Standards](https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/external-contributors.md).
- [x] I have requested a review from the relevant team or maintainers.

PR #5720, opened 2024-03-11
feat: change default script and predicate program generated at init

kayagokalp:
## Description
Changes default predicate to:
```sway
predicate;

fn main() -> bool {
    true
}
```
and default script to:
```sway
script;

use std::logging::log;

configurable {
    SECRET_NUMBER: u64 = 0
}

fn main() -> u64 {
    log(SECRET_NUMBER);
    return SECRET_NUMBER;
}
```


calldelegation:
Thank you @kayagokalp 🙏! 

Before the defaults was predicates returning `false` and script was simply empty which means you can't make any meaningful example templated assertions to them. These new examples are for the new templates for scripts and predicates
https://github.com/FuelLabs/sway/pull/5238

calldelegation (approved):

JoshuaBatty (approved):

PR #5721, opened 2024-03-12
make source id optional when auto impl for new encoding

xunilrj:
## Description

This PR makes `source_id` optional when auto generation code. Specifically some contracts don't necessarily must have contract methods.

## Checklist


xunilrj:
> In theory I think we should always be able to get a valid module id, so even if they have no contracts, we should be able to get a module id. Is there maybe a span on the contract itself we could use?

I changed it to use the `module_id` from the `Module` being `type_check`ed. 
But I still think we must not expect a module_id to exist in all cases, given that the data model supports `None`.

@tritao, what do you think?

IGI-111 (approved):

PR #5724, opened 2024-03-13
Tweak the `did_change_stress_test_random_wait` test

JoshuaBatty:
## Description
This test was originally added in #5704. However, compiling 400 times seems to be a bit too ambitious as it was causing issues like stack overflows when all tests were run concurrently. I've reduced this to 60 so it finishes a lot quicker now. It should still be enough to test if any issues affect the garbage collector. 

## Checklist

### DIFF
--- a/sway-lsp/tests/lib.rs
+++ b/sway-lsp/tests/lib.rs
@@ -138,43 +138,43 @@ async fn did_change_stress_test() {
     shutdown_and_exit(&mut service).await;
 }
 
-#[tokio::test]
-#[allow(dead_code)]
-async fn did_change_stress_test_random_wait() {
-    std::env::set_var("RUST_BACKTRACE", "1");
-    let default_panic = std::panic::take_hook();
-    std::panic::set_hook(Box::new(move |panic_info| {
-        default_panic(panic_info); // Print the panic message
-        std::process::exit(1);
-    }));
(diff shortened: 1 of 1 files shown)


xunilrj:
Makes sense to also timeout the test? One the worse things happening before it was that it never failed. Even if it is like 5 minutes or something huge.

kayagokalp (dismissed):

sdankel (dismissed):

JoshuaBatty:
> Makes sense to also timeout the test? One the worse things happening before it was that it never failed. Even if it is like 5 minutes or something huge.

Sure no problem. I'll add a 5 minute timeout just for extra precaution.

kayagokalp (commented):
Do we want timeout to cause test failure or silently pass? I am not sure at all but this feels like it will print the error but test will pass (thus we will not see the error) since we don't have any asserts or panics there?

JoshuaBatty:
> Do we want timeout to cause test failure or silently pass? I am not sure at all but this feels like it will print the error but test will pass (thus we will not see the error) since we don't have any asserts or panics there?

That's a great point. I've changed to an assert if the timeout is hit.

kayagokalp (dismissed):

xunilrj (approved):

Issue #5666, opened 2024-02-26
Fix DCA divergence use cases.


esdrubal:
As reported in the following PR we should fix the TODO-DCA.
 https://github.com/FuelLabs/sway/pull/5657/files#diff-0095b6bff4cf21cd2d7b51905e036b178af0c8179c860b25b50767273b4f131c

Issue #3496, opened 2022-12-02
Consider `(Never, u8)` and `enum A { Foo: (Never, u8) }` -- should these coerce to any other type? Perhaps this should be a more general "is type uninhabited" test?


mohammadfawaz:
Consider `(Never, u8)` and `enum A { Foo: (Never, u8) }` -- should these coerce to any other type? Perhaps this should be a more general "is type uninhabited" test?

_Originally posted by @Centril in https://github.com/FuelLabs/sway/pull/3362#discussion_r1026657101_

mohammadfawaz:
Assigning to you @esdrubal. This is just a comment that may have not been addressed on your `Never` type PR.

PR #5726, opened 2024-03-13
Improves DCA warnings after divergent expressions.

esdrubal:
## Description

DCA warnings were missing when methods call return in some places.

Changes e2e test runner to use file checker in tests with category runs.

Fixes https://github.com/FuelLabs/sway/issues/5666

## Checklist


sway-core/src/type_system/info.rs
```diff
@@ -1156,6 +1156,43 @@ impl TypeInfo {
...
+                .contains_never(engines),
+            TypeInfo::Unknown
+            | TypeInfo::UnknownGeneric { .. }
+            | TypeInfo::Placeholder(_)
+            | TypeInfo::TypeParam(_)
+            | TypeInfo::StringSlice
+            | TypeInfo::StringArray(_)
+            | TypeInfo::UnsignedInteger(_)
+            | TypeInfo::Enum(_)
+            | TypeInfo::Struct(_)
```
ironcev, line 1177: Is this completely correct? We can have uninhabited structs and enums. E.g., this code compiles without errors, as it should:
```Sway
script;

struct S {
    x: !,
}

fn main() -> u64 {
    let s = S { x: { return 0u64; } };    
    0
}
```

sway-core/src/type_system/info.rs
```diff
@@ -1156,6 +1156,43 @@ impl TypeInfo {
...
+            | TypeInfo::ContractCaller { .. }
+            | TypeInfo::Custom { .. }
+            | TypeInfo::B256
+            | TypeInfo::Numeric
+            | TypeInfo::Contract
+            | TypeInfo::ErrorRecovery(_)
+            | TypeInfo::Storage { .. }
+            | TypeInfo::RawUntypedPtr
+            | TypeInfo::RawUntypedSlice
+            | TypeInfo::Ptr(_)
```
ironcev, line 1188: Also here, `Ptr`, `Slice`, `Alias`, and `Ref`. If the types they point to or contain or alias contain `Never` then I think they also cannot be inhabited.

sway-core/src/type_system/info.rs
```diff
@@ -1156,6 +1156,43 @@ impl TypeInfo {
         matches!(self, TypeInfo::Tuple(_))
     }
 
+    pub fn contains_never(&self, engines: &Engines) -> bool {
```
ironcev, line 1159: Perhaps calling this method `is_uninhabited`? `contains_never` sounds to me more like a _way_ to make a type uninhabited, but if I think about a type property the sentence "type.is_uninhabited()` sounds more like what that type _is_, which property it has.

ironcev (changes requested):
To the questions on the `contains_never` method, maybe it can be done as a different PR, but maybe also as the part of this PR, to extend DCA test for uninhabited types and the cases like the one in the code sample provided in the comment where the variable `s` even if used somewhere later will actually always be dead. I expect our DCA with the change in this PR will already behave like that.

In that case we could also close #3496 with this PR.

test/src/e2e_vm_tests/test_programs/should_pass/return_in_strange_positions/src/main.sw
```diff
@@ -90,19 +90,19 @@ fn in_while_condition() -> u64 {
         break;
```
xunilrj, line 90: Shouldn´t this break also be unreachable? Actually.... thee whole body.

xunilrj: This is what rustc returns
```
warning: unreachable block in `if` or `while` expression
 --> src/main.rs:2:22
  |
2 |     while {return 1} {}
  |            --------  ^^ unreachable block in `if` or `while` expression
  |            |
  |            any code following this expression is unreachable
  |
  = note: `#[warn(unreachable_code)]` on by default
```

test/src/e2e_vm_tests/test_programs/should_pass/return_in_strange_positions/src/main.sw
```diff
@@ -90,19 +90,19 @@ fn in_while_condition() -> u64 {
...
 
 fn in_enum() -> u64 {
     let _ = Enum::A((return 42, return 43));
     
-    845 // TODO: Missing unreachable warning
+    845
 }
 
 fn in_enum_multivariant() -> u64 {
     let _ = Enum_multivariant::B((return 42, return 43));
```
xunilrj, line 103: and here.... 

```
warning: unreachable call
 --> src/main.rs:6:5
  |
6 |     A::A({return 1})
  |     ^^^^  -------- any code following this expression is unreachable
  |     |
  |     unreachable call
  |
  = note: `#[warn(unreachable_code)]` on by default
```

Although the "call" is the rustic error, is horrible.

xunilrj (approved):

xunilrj (approved):

IGI-111 (approved):

PR #5741, opened 2024-03-18
Do not initialise zero-sized types

### LINKED ISSUE #5692: MCPI memory ownership issue with mutable zero-sized arrays in release builds
The below example panics when run in `release` mode with the following reason:
```
Reason: PanicInstruction {
    reason: MemoryOwnership,
    instruction: MCPI { dst_addr: 0x11, src_addr: 0x3c, len: 0 } (bytes: 60 47 c0 00)
}
```
A running example can be found in [this branch](https://github.com/FuelLabs/sway/tree/ironcev/mcpi-memory-ownership-issue-with-mut-vars-in-release/test/src/e2e_vm_tests/test_programs/should_pass/language/references/mcpi-memory-ownership-issue). 

My conclusions so far are given below in the code comments. In short, the issue arises only in `release` mode when a mutable zero sized array is allocated within a function and access via pointer.

```Sway
script;

pub trait New {
    fn new() -> Self;
}

impl New for [u64;0] {
    fn new() -> Self {
        []
    }
}

struct EmptyStruct { }

impl New for EmptyStruct {
    fn new() -> Self {
        EmptyStruct { }
    }
}

#[inline(always)]
fn reference_zero_sized_local_var_and_value<T>()
    where T: New
{
    assert(__size_of::<T>() == 0);

    let mut x = T::new();
    // let x = T::new(); // OK if the variable is not mutable.

    // Both examples fail. Taking the address from the reference, or using `__addr_of()`.
    // let x_ptr = asm(r: &x) { r: raw_ptr };
    // let _ = x_ptr.read::<T>();

    let x_ptr = __addr_of(x);
    let _ = x_ptr.read::<T>(); // Fails here with:
                               // Reason: PanicInstruction
                               // { reason: MemoryOwnership, instruction: MCPI { dst_addr: 0x11, src_addr: 0x3c, len: 0 } (bytes: 60 47 c0 00) }
}

#[inline(never)]
fn reference_zero_sized_local_var_and_value_not_inlined<T>()
    where T: New
{
    reference_zero_sized_local_var_and_value::<T>()
}

fn main() -> u64 {
    // OK for empty struct.
    // reference_zero_sized_local_var_and_value_not_inlined::<EmptyStruct>();

    // Fails for empty array when called over `_not_inlined`.
    reference_zero_sized_local_var_and_value_not_inlined::<[u64;0]>();

    // OK if called directly and being inlined.
    // reference_zero_sized_local_var_and_value::<[u64;0]>();

    42
}
```

CC: @vaivaswatha

vaivaswatha:
Fixes #5692

### DIFF
--- a/sway-core/src/asm_generation/fuel/functions.rs
+++ b/sway-core/src/asm_generation/fuel/functions.rs
@@ -851,6 +851,10 @@ impl<'ir, 'eng> FuelAsmBuilder<'ir, 'eng> {
             data_id,
         } in init_mut_vars
         {
+            if var_size.in_bytes() == 0 {
+                // Don't bother initializing zero-sized types.
+                continue;
+            }
             // Load our initialiser from the data section.
             self.cur_bytecode.push(Op {
                 opcode: Either::Left(VirtualOp::LoadDataId(
--- a/test/src/e2e_vm_tests/test_programs/should_pass/language/mut_ref_empty_type/.gitignore
+++ b/test/src/e2e_vm_tests/test_programs/should_pass/language/mut_ref_empty_type/.gitignore
@@ -1,0 +1,2 @@
+out
+target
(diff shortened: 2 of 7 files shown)


ironcev (approved):
A simple fix indeed :smile: 

Removing the `.gitignore` would be nice just to keep the tests clean.

ironcev (approved):

PR #5753, opened 2024-03-19
stdlib: Fix temp alloc size in `try_as_str_array`.

vaivaswatha:
Allocation must be based on the type size with padding.

In addition to the usual testing, I've also tested against https://github.com/FuelLabs/fuel-vm/tree/feature/0.43.3-memory-grow.


sway-lib-std/src/primitive_conversions/str.sw
```diff
@@ -6,17 +6,23 @@ impl str {
...
-            let s: S = asm(str_size: str_size, source: source, dest) {
+            let s: S = asm(
+                str_size: str_size,
+                tmp_alloc_size: tmp_alloc_size,
+                source: source,
+                dest,
+            ) {
                 move dest sp;
-                cfe str_size;
+                cfe tmp_alloc_size;
```
xgreenx, line 20: Have we analyzed other places where we allocate/modify stack/heap? Maybe we have this problem in other places as well.

vaivaswatha: There's a similar code in `from_str_array` in `core::str`, but that already is doing it right, by using `__size_of` for the allocation.

I checked other uses of `cfe` and `alloc` in `std` and `core`, and they all seem fine to me.

IGI-111 (approved):

xgreenx (approved):

PR #5757, opened 2024-03-20
Improve execution time of `should_pass` E2E tests

ironcev:
## Description

This PR improves the execution time of `should_pass` E2E tests by making ~33% of them use a stripped-down version of the `std` library.

This approach brings noticable speed gains as demonstrated in the below comparison table.
(The tests are executed on my local machine in Rust `release` and Sway `debug` build. Running them in Sway `release` build brings additional but small gain.)

| Test filter  | Before (mm:ss) | After (mm:ss) |
| ------------- | ------------- |------------- |
| should_pass/language/ | 14:26 | 04:23 |
| should_pass/language/references | 04:10 | 00:56 |
| should_pass/language/main_args  | 01:10  | 00:12 |
| should_pass/language/match | 00:57 | 00:09 |
| should_pass.*enum | 00:51 | 00:28 |

The improvement approach is based on the following observations:
- most of the time in `should_pass` tests is spent on compiling `std`.
- out of ~480 `should_pass` tests, ~160 use `std` just because of `assert`.
- in addition, ~10 tests need `Option` and/or `Result`.

Thus, the PR provides two stripped-down versions of the `std`, one containing only support for `assert`, and another one containing `assert` and `Option` and `Result`. The compilation time of those two libraries is negligible compared to the compilation time of the whole `std`.

This apporach is:
- easy to use and enforce in new tests
- compatible with whichever solution for running tests in parallel we come up with
- trivial to remove once we implement incremental compilation or caching of `std` in tests

Therefore, proposal is to use this approach for the time being to improve the efficiency of our inner loop.

## Checklist


vaivaswatha (approved):

ironcev:
> I do have some long-term concerns about duplicating the libraries code

Me too! Duplicating and running out of sync is definitely a concern. Having something like cargo features is of course the long term feature great to have. For the tests, I was thinking about a simple approach where the `std` code is not duplicated and checked in, but for each reduced version of the lib simply copied over locally (not checked in) before test start, based on a simple config file. Only the strip down `lib.sw` and `prelude.sw` should be checked in. I might play with this a bit in the second step.

PR #5759, opened 2024-03-21
chore:bump sdk to v0.56, fuel-core to 0.23.0, fuel-vm to 0.47.1

kayagokalp:
## Description

Update rust sdk to v0.56, fuel-core to v0.23.0 and fuel-vm to v0.47.1

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -308,7 +308,7 @@ jobs:
       - name: Install fuel-core from Cargo
         run: cargo install fuel-core-bin --version ${{ needs.get-fuel-core-version.outputs.fuel_core_version }}
       - name: Run fuel-core
-        run: fuel-core run --db-type in-memory --debug 
+        run: fuel-core run --db-type in-memory --debug & 
       - uses: actions/checkout@v3
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@master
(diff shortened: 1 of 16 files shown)


kayagokalp:
Currently points to unreleased version of forc-wallet and it uses some unwraps to quickly see if things work. I will fix the error handling.

.github/workflows/ci.yml
```diff
@@ -304,38 +304,44 @@ jobs:
...
       - uses: actions/checkout@v3
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ env.RUST_VERSION }}
       - uses: Swatinem/rust-cache@v2
       - name: Cargo Run E2E Tests (Fuel VM)
-        run: cargo run --locked --release --bin test -- --locked
+        run: |
+          fuel-core run --db-type in-memory --debug &
```
sdankel, line 321: Why does fuel core need to be run in debug mode for the tests?

forc-plugins/forc-client/src/op/run/mod.rs
```diff
@@ -177,24 +175,34 @@ async fn send_tx(
...
-
-    match outputs {
-        Ok(logs) => {
-            info!("{}", format_log_receipts(&logs, pretty_print)?);
-            Ok(logs)
-        }
-        Err(e) => bail!("{e}"),
+    if !outputs.is_empty() {
+        info!("{}", format_log_receipts(&outputs, pretty_print)?);
     }
```
sdankel, line 204: Should we still bail when receipts are empty?

.github/workflows/ci.yml
kayagokalp: https://github.com/FuelLabs/sway/pull/5759#discussion_r1559961477

forc-plugins/forc-client/src/op/run/mod.rs
kayagokalp: I think output was returning a Result but not anymore, so even if we had empty receipts we were not bailing. So it should be the same behaviour now but I might be missing something.

sdankel (approved):

JoshuaBatty (approved):
thanks Kaya

PR #5792, opened 2024-03-27
Fail nicely on imports without callpaths

jjcnn:
## Description

`use` statements without a call path such as

```
use foo;
```

causes the compiler to crash during semantic analysis.

This PR causes the compiler to fail nicely instead.

These types of imports should be allowed, but I am in the process of reworking the import logic, so I'm postponing an actual fix.

## Checklist

### DIFF
--- a/sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
+++ b/sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
@@ -268,7 +268,7 @@ fn item_use_to_use_statements(
         };
         return Err(handler.emit_err(error.into()));
     };
-    
+
     let mut ret = Vec::new();
     let mut prefix = Vec::new();
     let item_span = item_use.span();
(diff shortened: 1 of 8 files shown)


sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
```diff
@@ -267,7 +267,8 @@ fn item_use_to_use_statements(
             span: pub_token.span(),
         };
         return Err(handler.emit_err(error.into()));
-    }
+    };
```
IGI-111, line 270: Probably no need for this change.

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
```diff
@@ -279,6 +280,18 @@ fn item_use_to_use_statements(
         &mut ret,
         item_span,
     );
+
+    // Check that all use statements have a call_path
+    // This is not the case for `use foo;`, which is currently not supported
```
IGI-111, line 285: I don't think those types of imports were ever part of the spec, so I don't think we have an issue tracking this.
Can you create one and mention it here?

IGI-111 (commented):
Couple nits but LGTM

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
jjcnn: Fixed.

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
jjcnn: Will do.

PR #5812, opened 2024-04-01
feat: add log decoding capabilities to show decoded log values for forc-test

kayagokalp:
## Description

This PR adds log decoding capabilities to forc, so that it can be used for showing actual decoded, human readable values at forc-test output.

For the following unit test:

```sway
script;

fn main() {}

#[test]
fn test_fn() {
	let a = 10;
	log(a);
	let b = 30;
	log(b);
	assert_eq(a, 10)
}
```
Logs are:

```console
zbap@/Users/-/fuel/test_projects/script_main > ../../dev/sway/target/release/forc test --logs --experimental-new-encoding
  Finished debug [unoptimized + fuel] target(s) in 5.14s
      Bytecode hash: 0x562b471bae4323b806a1b2829014ef7078a5b9406a076a54cf0b2f1a61cb49f8
   Running 1 test, filtered 0 tests
      test test_fn ... ok (33.875µs, 200 gas)
Decoded log value: 10, log rb: 0
Decoded log value: 30, log rb: 1
Raw logs:
[{"LogData":{"data":"000000000000000a","digest":"8d85f8467240628a94819b26bee26e3a9b2804334c63482deacec8d64ab4e1e7","id":"0000000000000000000000000000000000000000000000000000000000000000","is":10336,"len":8,"pc":10784,"ptr":67107840,"ra":0,"rb":0}},{"LogData":{"data":"000000000000001e","digest":"48a97e421546f8d4cae1cf88c51a459a8c10a88442eed63643dd263cef880c1c","id":"0000000000000000000000000000000000000000000000000000000000000000","is":10336,"len":8,"pc":11048,"ptr":67106816,"ra":0,"rb":1}}]
   Result: OK. 1 passed. 0 failed. Finished in 33.875µs.
```

Also this PR fixes a bug in test suite which causes test suite to omit the encoding information passed into forc-test from `unit_tests_pass` tests. And adds log decode testing capabilities to the test suite.

Next up is adding actual variable names instead of log id so that decoded log values will show as, this is left to a future PR

`Decoded log: ( a = 10 )`
`Decoded log: ( b = 30 )`


forc/src/cli/commands/test.rs
```diff
@@ -142,6 +147,21 @@ fn print_tested_pkg(pkg: &TestedPackage, test_print_opts: &TestPrintOpts) -> For
...
+                for log in logs {
+                    if let Receipt::LogData {
+                        rb,
+                        data: Some(data),
+                        ..
+                    } = log
+                    {
+                        let decoded_log_data = decode_log_data(*rb, data, &pkg.built.program_abi)?;
+                        let var_value = decoded_log_data.value;
+                        info!("Decoded log value: {}, log rb: {}", var_value, rb);
```
sdankel, line 160: Should we just print the value without anything before it? 

What is `rb`, and how would it be useful?

forc/src/cli/commands/test.rs
```diff
@@ -142,6 +147,21 @@ fn print_tested_pkg(pkg: &TestedPackage, test_print_opts: &TestPrintOpts) -> For
...
+                    }
+                }
+                info!("Raw logs:");
```
sdankel, line 163: Maybe we should have a separate function for printing this, like `raw_log(a)`?

For me, when I use `log(x)`, I'd expect it to only print the value of x.

forc/src/cli/commands/test.rs
kayagokalp: Since we do not have variable names, there is no way for users to differentiate between different logs. It is coming from the raw log and incremented by one with each log, they would have to count number of logs in the output to match the log statements in the code. We can remove it but i was planning to remove it once we have var names implemented.

forc/src/cli/commands/test.rs
kayagokalp: That is correct, the reason i did this way is because decoding logs are very primitive without var names etc, so that is also why this is under a flag `--decode` as well, I am fine with removing raw logs with decoded logs too. Maybe we should have `--raw`, if `--decode` and `--raw` is both provided we can show the current output, with only `--decode` we will only see `Decoded log value ...` lines.

forc/src/cli/commands/test.rs
sdankel: Gotcha. I'm ok with this as a temporary solution until we have var names.

sdankel (approved):

JoshuaBatty (approved):
Nice work

forc/src/cli/commands/test.rs
kayagokalp: Created #6015

Issue #5811, opened 2024-04-01
Including multiple std implementations can fail


IGI-111:
We seem to be including code from test files as part of `std` when compiling user code with a toolchain from `fuelup`, this can result in errors such as this:

```
error
 --> /Users/daniel/.forc/git/checkouts/std-9be0d6062747ea7/98d8f4cadbac34f232926b9a44b1170ec9da425d/test/src/e2e_vm_tests/reduced_std_libs/sway-lib-std-assert/src/lib.sw:6:9
  |
4 |
5 | pub mod revert;
6 | pub mod assert;
  |         ^^^^^^ The file /Users/daniel/.forc/git/checkouts/std-9be0d6062747ea7/98d8f4cadbac34f232926b9a44b1170ec9da425d/test/src/e2e_vm_tests/reduced_std_libs/sway-lib-std-assert/src/assert.sw could not be read: No such file or directory (os error 2)
7 | pub mod prelude;
  |
____

error
 --> /Users/daniel/.forc/git/checkouts/std-9be0d6062747ea7/98d8f4cadbac34f232926b9a44b1170ec9da425d/test/src/e2e_vm_tests/reduced_std_libs/sway-lib-std-assert/src/prelude.sw:7:7
  |
5 |
6 | // Error handling
7 | use ::assert::{assert, assert_eq, assert_ne};
  |       ^^^^^^ Module "assert" could not be found.
8 | use ::revert::{require, revert};
  |
____

error
  --> /Users/daniel/.forc/git/checkouts/std-9be0d6062747ea7/98d8f4cadbac34f232926b9a44b1170ec9da425d/test/src/e2e_vm_tests/reduced_std_libs/sway-lib-std-assert/src/prelude.sw:8:7
   |
 6 |
 7 | use ::assert::{assert, assert_eq, assert_ne};
 8 | use ::revert::{require, revert};
   |       ^^^^^^ Module "revert" could not be found.
 9 |
10 | // Logging
   |
```

This has been triggered by https://github.com/FuelLabs/sway/pull/5763 but we've observed this error in the past when including multiple versions of the standard library in the source for stability purposes.

PR #5819, opened 2024-04-02
fix: make sure to check std is taken from sway-lib-std folder

kayagokalp:
## Description

This PR adds a ad-hoc solution for #5811. Basically checks if the fetched package is 'std' and if so, only package from 'sway-lib-std' folder is accepted for that package.


kayagokalp:
This one is not easy to test, I applied these changes onto 0.52.0 tag and verified to build a project that was not building before.

JoshuaBatty (approved):

PR #5840, opened 2024-04-09
docs: add external code doc

sarahschwartz:
## Description

This PR adds a page for external code execution in the Sway book.

## Checklist


docs/book/src/blockchain-development/external_code.md
```diff
@@ -0,0 +1,61 @@
...
+
+The contract has two functions:
+
+- `set_target_contract` updates the `target_contract` variable in storage with the `ContractId` of an external contract.
+- `double_input` reads the `target_contract` from storage and uses it to run external code. If the `target_contract` has a function with the same name (`double_input`), the code in the external `double_input` function will run.
+In this case, the function will return a `u64`.
+
+## Fallback functions
+
+If the function name doesn't exist in the target contract but a `fallback` function does, the `fallback` function will be triggered.
```
matt-user, line 34: What happens if there is zero or more than 1 fallback functions?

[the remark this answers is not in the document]
IGI-111: Only the attribute matters. The function signature can be anything.

IGI-111: Zero means the default behavior: revert on calling functions that aren't in the contract ABI.

More than one fallback function per contract produces an error.

docs/book/src/blockchain-development/external_code.md
```diff
@@ -0,0 +1,61 @@
...
+```sway
+{{#include ../../../../test/src/sdk-harness/test_projects/run_external_target/src/main.sw:fallback}}
+```
+
+In this case, the `does_not_exist_in_the_target` function will return `_foo * 3`.
+
+## How does this differ from calling a contract?
+
+Unlike a normal contract call, the context of the contract running
+`run_external` is retained for the loaded code.
```
IGI-111, line 52: I feel like this isn't clear enough that the *storage* context is retained. We should insist on that since it's the primary reason to do this, and there probably should be an example here of a proxy that retains storage.

Another missing thing is an explanation of `#[namespace(...)]`, which is meant for that particular use case. Perhaps we should have that as part of such an example and explain why it is needed.

xunilrj (dismissed):

docs/book/src/blockchain-development/external_code.md
```diff
@@ -0,0 +1,74 @@
...
+There are a couple of major differences between calling a contract directly and using the `run_external` method.
+
+First, to use `run_external`, the ABI of the external contract is not required. The proxy contract has no knowledge of the external contract except for its `ContractId`.
+
+### Upgradeable Contract Storage
+
+Second, the storage context of the proxy contract is retained for the loaded code.
+This means that in the examples above, if someone calls the `double_input` function on the proxy contract, the `value` variable will be updated in the proxy contract's storage.
+
+If the proxy contract where to use a normal contract call to call `double_input`, then the `value` variable in the implementation contract would updated, and would not change in the proxy contract.
```
IGI-111, line 43:
```suggestion
If the proxy contract were to use a normal contract call to call `double_input`, then the `value` variable in the implementation contract would updated, and would not change in the proxy contract.
```

docs/book/src/blockchain-development/external_code.md
```diff
@@ -0,0 +1,74 @@
...
+## How does this differ from calling a contract?
+
+There are a couple of major differences between calling a contract directly and using the `run_external` method.
+
+First, to use `run_external`, the ABI of the external contract is not required. The proxy contract has no knowledge of the external contract except for its `ContractId`.
+
+### Upgradeable Contract Storage
+
+Second, the storage context of the proxy contract is retained for the loaded code.
+This means that in the examples above, if someone calls the `double_input` function on the proxy contract, the `value` variable will be updated in the proxy contract's storage.
```
IGI-111, line 41: The example is wrong. The `value` variable in the proxy storage will not get updated, and it's because it's placed at a completely different location in the storage because of the namespace attribute.

From our conversation I think that you understand which storage gets updated, but you seem to be under the impression that a variable needs to exist in the proxy's storage for the target to modify it. This is not the case.

Were the `value` variable not declared in the proxy's storage (and in effect it isn't because the one you have here is moved by the namespace at a different location) the target code would still work and update slots in the proxy's storage *as if* it were laid out in the way that is described by the target's storage specification.

IGI-111 (changes requested):

IGI-111 (approved):

docs/book/src/blockchain-development/external_code.md
```diff
@@ -0,0 +1,73 @@
...
+The proxy contract loads the code and executes it in its own context.
+
+## Fallback functions
+
+If the function name doesn't exist in the target contract but a `fallback` function does, the `fallback` function will be triggered.
+
+> If there is no fallback function, the transaction will revert.
+
+You can access function parameters for fallback functions using the `call_frames` module in the `std-lib`.
+For example, to access the `_foo` input parameter in the proxy function below, you can use the `second_param` method in the `fallback` function:
```
xunilrj, line 55: Although 'second_param' will also somehow work, it is better to use 'called_args', as the example is now doing.

PS: the example has a bunch of logs because we are still fixing a LDC bug.

IGI-111 (approved):

examples/upgradeable_proxy/proxy/src/main.sw
```diff
@@ -0,0 +1,31 @@
...
+    fn set_target_contract(id: ContractId) {
+        storage.target_contract.write(Some(id));
+    }
+
+    #[storage(read)]
+    fn double_input(_value: u64) -> u64 {
+        let target = storage.target_contract.read().unwrap();
+        run_external(target)
+    }
+    // ANCHOR_END: proxy
```
ironcev, line 30: Is this intentional, that the snippet for the impl does not include the closing curly bracket of the impl block?

sarahschwartz: good catch! thanks

IGI-111 (approved):

xunilrj (approved):

PR #5847, opened 2024-04-12
Fix and improve `Unimplemented` compile errors

### LINKED ISSUE #5803: Mutable variable is considered immutable
Forc: `0.52.0`

```sway
contract;

pub struct Balance {
    base: u64,
}

impl Balance {
    pub fn credit(ref mut self, amount: u64) {
        self.base += amount;
    }
}

abi MyContract {
    fn test_function();
}

    fn test_function() {
        let x = Some(Balance { base: 0 });
        match x {
            Some(mut balance) => balance.credit(5),
            None => revert(0),
        };
    }
}
```

### LINKED ISSUE #5846: `Unimplemented` Error not emitted when matching on aliases
In the below example, the `Unimplemented feature: Matching on this type is currently not supported.` error is emitted only in the `array` function, but not in the `array_alias`. Alias is always considered to be matchable no matter what is an alias for, which is a bug.

```Sway
script;

type ArrayAlias = [u64;3];

fn main() {
    let a = [1u64, 2u64, 3u64];
    array(a);
    array_alias(a);
}

fn array(array: [u64;3]) {
    let _ = match array { // <<< OK: Error emitted here.
        _ => true,
    };
}

fn array_alias(array_alias: ArrayAlias) {
    let _ = match array_alias { // <<< ERROR: Error not emitted here,
        _ => true,
    };
}
```

ironcev:
## Description

This PR:
- closes #5803 by providing the proper error message that `mut` is not implemented in pattern matching. Support for `ref`, `mut`, and `ref mut` in pattern matching will be done later as a part of support for references in pattern matching.
- closes #5846 by checking if aliased type can be matched.
- provides `Diagnostic` for the `Unimplemented` error and removes the `UnimplementedWithHelp` error.
- differentiates between unimplemented features and features that are actually semantically not supported. Previously, `Unimplemented` error was used for both which emitted misleading errors in `match` expressions, using module paths in place of expressions, and trait implementations.

Closes #5803, #5846.

# Demo

Before:
Self type of an impl block is not valid - Before

After:
Self type of an impl block is not valid - After

Before:
Module path is not an expression - Before

After:
Module path is not an expression - After

Diagnostic for `Unimplemented` error:
Matching on this type is currently not implemented - After

## Checklist


IGI-111 (dismissed):

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -1506,10 +1506,11 @@ impl ty::TyExpression {
             }
             (true, None, None, None) => {
                 handler.append(module_probe_handler);
-                return Err(handler.emit_err(CompileError::Unimplemented(
-                    "this case is not yet implemented",
+                return Err(handler.emit_err(CompileError::Unimplemented {
+                    feature: "This case".to_string(),
```
jjcnn, line 1510:
```suggestion
                    feature: "Path bindings of modules".to_string(),
```

"this case" refers to compiler code, which is not helpful to the developer.

jjcnn (approved):
A few nits, but otherwise LGTM.

xunilrj (dismissed):

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
ironcev: Thanks for pointing this out @jjcnn! I wouldn't call it a nit, it was rather a blissful ignorance from my side. It turned out in the end that it is one of those places that is semantically meaningless and `Unimplemented` wasn't the right error at all. The new error is added to properly cover this case.

sway-ast/src/pattern.rs
```diff
@@ -15,7 +15,7 @@ pub enum Pattern {
     /// A pattern made of a single ident, which could either be a variable or an enum variant
     AmbiguousSingleIdent(Ident),
     Var {
-        reference: Option<RefToken>, // TODO-IG: Remove this and replace with proper support for references in patterns.
+        reference: Option<RefToken>, // TODO-IG: Implement `ref`, `mut`, and `ref mut` when implementing matching of references.
```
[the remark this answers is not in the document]
ironcev: Agree and I usually marke TODOs with issues (or at least started doing it :smile:). But these references related TODOs are more remainders where to look during the ongoing development of references which is tracked in the tracking issue.

IGI-111 (approved):

xunilrj (approved):

PR #5935, opened 2024-04-30
chore: fund test suite's account by default in local-test node

kayagokalp:
## Description

Funds account associated with  `de97d8624a438121b86a1956544bd72ed68cd69f2c99555b08b1e8c51ffd511c` which is used in the test suite. This is done since locally we tend to rely on ci's initial state which should include this account so that some deployment transactions (required for updating contract ids) can be executed.


JoshuaBatty (approved):

sdankel:
> Deployment failed due to insufficient funds. Please be sure to have enough coins to pay for deployment transaction.

Do we need to add more to the account?

jjcnn (approved):

Issue #5727, opened 2024-03-13
Stabilize encoding v1


IGI-111:
Now that #4769 has been implemented we should take the necessary steps to turn the experimental encoding into the default:

- [ ] Update documentation
- [ ] Improve error messages
- [ ] Resolve any known bugs (notably with SDKs)
- [ ] Improve performance

DicksonWu654:
I confirmed that the remaining custom-`AbiEncode` configurable item causes an observable SDK/runtime mismatch on current `master` (`forc 0.72.1`, `dad95cc42b0383b4e3bacdeda9766565aa584a92`).

Minimal type:

```sway
const MASK: u64 = 0xa5a5a5a5a5a5a5a5;

struct Masked { value: u64 }

impl AbiEncode for Masked {
    fn is_encode_trivial() -> bool { false }
    fn abi_encode(self, buffer: Buffer) -> Buffer {
        (self.value ^ MASK).abi_encode(buffer)
    }
}

impl AbiDecode for Masked {
    fn is_decode_trivial() -> bool { false }
    fn abi_decode(ref mut buffer: BufferReader) -> Self {
        Masked { value: buffer.decode::<u64>() ^ MASK }
    }
}

configurable {
    MASKED: Masked = Masked { value: 1 },
}

fn main() -> u64 { MASKED.value }
```

The JSON ABI exposes `MASKED` only as the structural `Masked { value: u64 }`. The generated Rust configurable setter therefore encodes `Masked { value: 2 }` as `0000000000000002`; it has no metadata for the Sway codec. Applying those exact SDK bytes gives:

```text
compiled default returned:  0x0000000000000001
SDK requested value:        0x0000000000000002
configured script returned: 0xa5a5a5a5a5a5a5a7
```

This appears to be the concrete failure mode of the unchecked tracker item “Correctly error when a type with custom impl for AbiEncode is used in configurables,” rather than a separate issue. A regression test could generate an SDK setter and execute the patched binary, or at compiler level verify that custom codec semantics are rejected before ABI emission. The conservative fix seems to be detecting custom `AbiEncode`/`AbiDecode` implementations recursively in a configurable type and emitting a source diagnostic until the JSON ABI/tooling can represent those semantics.

PR #5942, opened 2024-05-01
Encoding v1 for Configurables

xunilrj:
## Description

This PR uses `AbiEncode` on configurables.  We can imagine that 

```sway
configurable {
    SOMETHING: u64 = 1
}

fn main() -> u64 {
    SOMETHING
}
```

will be desugared into

```sway
configurable {
    SOMETHING: slice = encode(1)
}

fn main() -> u64 {
    abi_decode(SOMETHING)
}
```

To allow this, now the whole `encode` function and all trait impls run inside `const_eval`. To make this work, three new intrinsic were implemented: `EncodeBufferEmpty`, `EncodeBufferAppend`, and `EncodeBufferAsRawSlice`.

`EncodeBufferEmpty` creates an empty "encoding buffer", which is composed of a pointer to the buffer, the buffer capacity, and how many bytes were written already.

`EncodeBufferAppend` appends any primitive data types to the buffer. This intrinsic does not mutate its argument, it returns a new "encoding buffer". If no reallocation is needed, the pointer and the capacity stay the same.

`EncodeBufferAsRawSlice` returns a slice with is composed of the buffer pointer and its length (not the capacity).

### Errors

Some constant expressions cannot live inside the data section because their encoding depends on some instance value, for example: string slices, Vecs, Sttrings, Bytes etc... For these we now we return an error in these cases, like

```
  error
            --> /home/xunilrj/github/sway/test/src/e2e_vm_tests/test_programs/should_pass/language/configurable_consts/src/main.sw:33:5
             |
          31 | 
          32 |     C6: str[4] = __to_str_array("fuel"),
          33 |     C6_2: str = "fuel as str",
             |     ^^^^ This code cannot be evaluated to a configurable because its size is not always limited.
          34 |     C7: [u64; 4] = [1, 2, 3, 4],
          35 |     
             |
          ____
```

## Future TODOs

This PR adds two more tasks to https://github.com/FuelLabs/sway/issues/5727:

- Enable sdk-harness tests that were ignore. SDK needs to update configurable encoding.
- Correctly error when a type with custom impl for AbiEncode is used in configurables
- avoid decoding on each use

## Checklist


test/src/sdk-harness/test_projects/configurables_in_contract/mod.rs
```diff
@@ -45,7 +47,9 @@ async fn contract_uses_default_configurables() -> Result<()> {
     Ok(())
 }
 
+// TODO Remove ignore when SDK supports encoding V1 for configurables
```
IGI-111, line 50: please add an issue number for those TODOs

xunilrj: Done.

ironcev (changes requested):
Before reviewing the implementation, there are two points that IMO needs to be addressed: the serious change in the semantics of configurables as entities and the cost of usage.

## Semantics
Up to now, configurables were treated like other constants, with the difference being deploy-time defined. But unchangable at runtime and in every aspect same as other constants at runtime.

This means, considering:
```
configurable {
    CONF: SomeStruct = SomeStruct { x: 0, y: 0 },
}

fn fun() {
    let p_1 = _addr_of(CONF);
    let p_2 = _addr_of(CONF);

    let r_1 = &CONF:
    let r_2 = &CONF:
}
```
the two pointers were pointing to the same struct, same as the two references.

With the current v1 implementation the code becomes, simplified to demonstrate the new semantics:
```
fn fun() {
    let p_1 = _addr_of(abi_decode(CONF));
    let p_2 = _addr_of(abi_decode(CONF));

    let r_1 = &abi_decode(CONF):
    let r_2 = &abi_decode(CONF):
}
```
This gives us four different entities, means different pointers and different references.

## Cost of usage
Calling `abi_decode` on every use explodes the generated IR code tremendously. The simple line from above:
```
let p_1 = _addr_of(CONF);
```
used to be just one IR instruction (`store c0 to v0`). Understandably it is now dozens of costly instructions like function calls, memcpy, asm, ...

This makes simple expressions like `CONF.x + CONF.y` extremely expensive at run-time, prohibits every function using configurables to be inlined because it will be huge etc. 

## Mitigation
I am thinking how to mitigate both points, to keep the proper semantics and also to remove the hidden runtime cost because it is huge and on every call site.

Could we perhaps do the decoding only once in the `__entry` and also store the decoded values in the `config` section and then use them as we use them now, as direct pointers? I am not sure if we can write to `config` from the code, but if possible, the config would look like e.g.:

```
// Encoded values.
c0 = config [u8; 16] [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0],
c1 = config [u8; 8] [u8 0, u8 0, u8 0, u8 0, u8 0, u8 1, u8 182, u8 105],
...
// Their decoded counterparts filled in `__entry`.
c10 = config { u64, u64 } .... some value here ....,
c11 = config u64 112233,
```

We can explore other options too.

I see it as necessary to resolve the above two issues, semantics and cost, before getting the feature out.

xunilrj:
I agree, but the question is if makes sense to merge this PR as is and improve these things later. Because the SDK needs this PR to finish their migration to encoding V1. 
And we can improve code generation in parallel, without blocking them.

ironcev:
I am fine with merging as is to unblock the SDK team as long as we continue immediately working on the points. Because as mentioned having the feature out as is would have serious unwanted consequences. Also, I can ignore configurables in my work in properly treating consts everywhere in IR and in inlining until the above issues are solved that also shouldn't be a problem.

xunilrj:
> I am fine with merging as is to unblock the SDK team as long as we continue immediately working on the points. 
This is the idea.

sway-core/src/ir_generation/const_eval.rs
```diff
@@ -1149,6 +1313,82 @@ fn const_eval_intrinsic(
                 span: intrinsic.span.clone(),
             })
         }
+        // Buffer starts ((ptr, 1024), 0)
```
ironcev, line 1316: The comment is misleading, because the initial buffer size is zero because of `vec![]`. We should either fix the comment or the initial buffer size.

sway-core/src/ir_generation/const_eval.rs
```diff
@@ -169,6 +171,72 @@ pub(crate) fn compile_const_decl(
...
+                                    ConstantValue::RawUntypedSlice(bytes) => Some(bytes.clone()),
+                                    _ => {
+                                        return Err(CompileError::NonConstantDeclValue {
+                                            span: call_path.span(),
+                                        });
+                                    }
+                                }
+                                .unwrap();
+
+                            // Pad the encoded buffer if needed
```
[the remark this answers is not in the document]
xunilrj: Remember that the idea is that the SDK will write here. We need the space in case they want to write the biggest variant.

sway-core/src/asm_generation/fuel/data_section.rs
```diff
@@ -124,9 +137,10 @@ impl Entry {
                 name,
                 padding,
             ),
+            ConstantValue::RawUntypedSlice(bytes) => Entry::new_slice(bytes.clone(), name, padding),
```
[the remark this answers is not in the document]
xunilrj: Done. Went for the more complete naming scheme.

sway-core/src/asm_generation/fuel/data_section.rs
```diff
@@ -309,6 +330,19 @@ impl fmt::Display for DataSection {
                     }
                     format!(".bytes[{}] {hex_str} {chr_str}", bs.len())
                 }
+                Datum::Slice(bs) => {
```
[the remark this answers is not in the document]
xunilrj: They are actually different. But I managed to extract a common function.

sway-core/src/ir_generation/const_eval.rs
```diff
@@ -169,6 +171,72 @@ pub(crate) fn compile_const_decl(
                             const_val,
                         );
                     } else {
+                        let const_val = if env.context.experimental.new_encoding {
+                            // This expression must be `encode(something)`.
+                            // We want the someting here
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/ir_generation/const_eval.rs
```diff
@@ -650,29 +720,84 @@ fn const_eval_typed_expr(
...
+        ty::TyExpressionVariant::UnsafeDowncast { exp, .. } => {
+            let value = const_eval_typed_expr(lookup, known_consts, exp, allow_configurables)?
+                .map(|x| x.value);
+            if let Some(ConstantValue::Struct(fields)) = value {
+                Some(fields[1].clone())
+            } else {
+                return Err(ConstEvalError::CompileError);
+            }
+        }
+        ty::TyExpressionVariant::WhileLoop {
```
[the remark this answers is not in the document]
xunilrj: Fair point.

We do need loops, because we need to be able to encode arrays, for example.
I will put a limit of 1M iterations at the moment.

We may even put a time-based limit/warning later.

sway-core/src/ir_generation/const_eval.rs
xunilrj: Removed. It is not helping anything actually.

sway-core/src/ir_generation/const_eval.rs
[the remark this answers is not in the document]
ironcev: Still thinking how obvious this is but despite how puzzled I was looking at it in the code. Facepalm definitely, but I would still put that explanation in the code comment. Along the line "Pad the encoded buffer. Configurables can be overridden at deploy time and we need to provide space needed by the largest possible instance of the stored type." The chance someone else will ask himself the same question one day is always there :smile:

sway-core/src/ir_generation/const_eval.rs
```diff
@@ -1149,6 +1324,81 @@ fn const_eval_intrinsic(
...
         }
+        Intrinsic::EncodeBufferEmpty => Ok(Some(to_encode_buffer(lookup, vec![], 0))),
+        Intrinsic::EncodeBufferAppend => {
+            assert!(args.len() == 2);
+
+            let (slice, mut len) = as_encode_buffer(&args[0]).unwrap();
+            let mut bytes = slice.clone();
+
+            use ConstantValue::*;
+            match &args[1].value {
```
ironcev, line 1335: Similar question for `()`. I understand that we have `impl AbiEncode for ()` and that `()` will always be properly encoded when being embedded. Also `().abi_encode(buffer)`. But `__encode_buffer_append` is just an available intrinsic. One can use it by calling `__encode_buffer_append(buffer, ())` and it's not illegal. Same with `bool` as discussed above. I don't see any cost involved in supporting these two types, we avoid general error that expression cannot be evaluated to const, and in general, why to have a special unsupported cases if there is no good reason to.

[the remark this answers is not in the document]
xunilrj: Agree that makes sense. Done.

xunilrj: The problem with supporting `()` is that to make it "complete", it would also need to support any tuple. And we would need to expand to any array, given how "similar" they are.

And my idea was to make it work only with primitive data types. 

So for example both `__encode_buffer_append(buffer, ())` and `__encode_buffer_append(buffer, MyStruct {})` fails as no const. Which I think makes sense.

ironcev: Yes makes perfectly sense, also the example comparing `()` with `EmptyStruct {}` definitely.

sway-core/src/ir_generation/function.rs
```diff
@@ -1354,6 +1480,437 @@ impl<'eng> FnCompiler<'eng> {
...
+                            string_len.val() as u64,
+                        )
+                    }
+                    TypeInfo::StringSlice | TypeInfo::RawUntypedSlice => {
+                        let uint64 = Type::get_uint64(context);
+
+                        let item_ptr = save_to_local_return_ptr(self, context, item);
+                        let addr = calc_addr_to_ptr_u64(&mut self.current_block, context, ptr, len);
+
+                        let addr_ident = Ident::new_no_span("addr".into());
```
ironcev, line 1782: Having a nice comment here like done in other places how the resulting `asm` block looks like would be nice. Seeing the resulting asm is very helpful.

sway-core/src/ir_generation/function.rs
```diff
@@ -1354,6 +1480,437 @@ impl<'eng> FnCompiler<'eng> {
...
+                    let _ = s.current_block.append(context).store(local_var_ptr, value);
+
+                    local_var_ptr
+                }
+
+                let new_len = match &*item_type {
+                    TypeInfo::Boolean => {
+                        assert!(item.get_type(context).unwrap().is_bool(context));
+
+                        // bool is represented as u64. So save to a local
```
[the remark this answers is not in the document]
xunilrj: Not when it is in a register.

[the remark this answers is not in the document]
xunilrj: To make both similar, but not exactly the same, we would need to convert the pointer, so store does not complains.

Not sure yet which is best.

sway-core/src/ir_generation/function.rs
```diff
@@ -1354,6 +1480,437 @@ impl<'eng> FnCompiler<'eng> {
...
+                            .mem_copy_bytes(addr, item_ptr, 1);
+                        increase_len(&mut self.current_block, context, len, 1)
+                    }
+                    TypeInfo::UnsignedInteger(IntegerBits::Eight) => {
+                        assert!(item.get_type(context).unwrap().is_uint8(context),);
+                        let addr = calc_addr_to_ptr_u8(&mut self.current_block, context, ptr, len);
+                        append_u8(&mut self.current_block, context, addr, len, item)
+                    }
+                    TypeInfo::UnsignedInteger(IntegerBits::Sixteen) => {
+                        assert!(item.get_type(context).unwrap().is_uint64(context));
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/ir_generation/function.rs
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/ir_generation/function.rs
```diff
@@ -1354,6 +1480,437 @@ impl<'eng> FnCompiler<'eng> {
...
+                    let addr = current_block
+                        .append(context)
+                        .binary_op(BinaryOpKind::Add, ptr, len);
+
+                    let uint64 = Type::get_uint64(context);
+                    let ptr_uint64 = Type::new_ptr(context, uint64);
+                    current_block.append(context).int_to_ptr(addr, ptr_uint64)
+                }
+
+                fn append_u8(
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/ir_generation/function.rs
```diff
@@ -1354,6 +1480,437 @@ impl<'eng> FnCompiler<'eng> {
...
+                            ty: uint64,
+                            value: ConstantValue::Uint(step),
+                        },
+                    );
+                    current_block
+                        .append(context)
+                        .binary_op(BinaryOpKind::Add, len, step)
+                }
+
+                fn calc_addr_to_ptr_u8(
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/ir_generation/const_eval.rs
```diff
@@ -1269,6 +1519,11 @@ mod tests {
         assert_is_constant(true, "enum Color { Blue: u64 }", "Color::Blue(1)");
         assert_is_constant(true, "struct Person { age: u64 }", "Person { age: 1 }");
         assert_is_constant(true, "struct Person { age: u64 }", "Person { age: 1 }.age");
+        assert_is_constant(
+            true,
+            "struct Person { age: u64 }",
+            "Person { age: { let mut x = 0; x = 1; 1} }",
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/ir_generation/function.rs
```diff
@@ -1354,6 +1480,437 @@ impl<'eng> FnCompiler<'eng> {
...
+                    context: &mut Context,
+                    value: Value,
+                ) -> Value {
+                    let temp_arg_name = s.lexical_map.insert_anon();
+
+                    let value_type = value.get_type(context).unwrap();
+                    let local_var = s
+                        .function
+                        .new_local_var(context, temp_arg_name, value_type, None, false)
+                        .map_err(|ir_error| {
```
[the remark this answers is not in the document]
xunilrj: Done. I am returning Result now.

sway-core/src/ir_generation/function.rs
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
```diff
@@ -92,10 +92,117 @@ impl ty::TyIntrinsicFunctionKind {
...
+    span: Span,
+) -> Result<(ty::TyIntrinsicFunctionKind, TypeId), ErrorEmitted> {
+    let type_engine = ctx.engines.te();
+    let engines = ctx.engines();
+
+    let buffer_expr = {
+        let ctx = ctx
+            .by_ref()
+            .with_help_text("")
+            .with_type_annotation(type_engine.insert(engines, TypeInfo::Unknown, None));
```
ironcev, line 184: Same as above. If we call e.g. `__encode_buffer_append(true, 1u8);` this will pass type check phase and the compiler will, expectantly, panic in both const context and in non-const context.

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
```diff
@@ -92,10 +92,117 @@ impl ty::TyIntrinsicFunctionKind {
...
```
ironcev, line 123: Shouldn't we have `Buffer` here as type annotation? If we call e.g. `__encode_buffer_as_raw_slice(true)` this will pass type check phase and the compiler will, expectantly, panic in both const context and in non-const context.

ironcev (changes requested):

sway-core/src/ir_generation/function.rs
```diff
@@ -1354,6 +1480,458 @@ impl<'eng> FnCompiler<'eng> {
...
+                                    ],
+                                    immediate: None,
+                                    metadata: None,
+                                },
+                            ],
+                            uint64,
+                            Some(new_len_ident),
+                        )
+                    }
+                    _ => return Err(CompileError::EncodingUnsupportedType { span }),
```
ironcev, line 1894: The `span` currently points to the whole intrinsic call. This makes the overall message very confusing because it is not clear that the passed data is of a non-supported type. E.g.:
```
let buffer = __encode_buffer_append(buffer, my_struct);
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Type not supported with __encode_buffer_append.
```
Just pointing to the argument would already be an improvement.

My suggestion is here to use expressive diagnostic and explain to programmers the issue and what is actually supported and expected. This is just a few minutes of extra programming effort and we will gain a lot of clarity in case of a wrong usage.

As example for a similar case, the errors we are giving for unsupported types in `match` expressions can be taken (E.g., This value has type "MyStruct". Appending values of this type to encoding buffer is not supported.). The help can list the supported types, similar to `MatchedValueIsNotValid` error.

IMO, same as `AbiEncode` implementations we have now, programmers will be writing their own impls looking at the examples in the standard library. I expect clarifying questions coming similar to those we had with `AbiEncode` before, and some can surely be avoided if the provided error messages are clear and instructive enough.

ironcev (changes requested):

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
xunilrj: Done

sway-core/src/ir_generation/function.rs
xunilrj: Span: done.

xunilrj: > IMO, same as AbiEncode implementations we have now, programmers will be writing their own impls looking at the examples in the standard library.

In theory, no one else needs to use this intrinsic. As nobody going to have access to the "buffer" field. More likely they will do "1u8.abi_encode(buffer)" like I do for Vecs.

ironcev: In theory :smile: I agree with the intended usage, which is always `<value>.abi_encode(buffer)` and no need for intrinsic. But I have that unsettling feeling of the habit we all devs have, to first look at `codec.sw` and then eventually read the docs :smile: That's why I am always for a good guiding error message, like the new one is.

sway-core/src/type_system/info.rs
```diff
@@ -1526,6 +1526,246 @@ impl TypeInfo {
     pub fn is_unknown_generic(&self) -> bool {
         matches!(self, TypeInfo::UnknownGeneric { .. })
     }
+
+    pub fn new_tuple(engines: &Engines, items: impl IntoIterator<Item = TypeInfo>) -> TypeInfo {
+        let te = engines.te();
+        let items = items
+            .into_iter()
+            .map(|x| te.insert(engines, x, None))
```
ironcev, line 1534: Quite some time ago @JoshuaBatty, @tritao and I were discussing internally, that passing the `source_id` to the `insert` is error prone because it forces all callers to understand the semantics of it and to pass the proper `source_id` which is actually dependent of the type of the `TypeInfo` being inserted etc. We had already places in code where wrong `source_id` information was provided. I wanted to refactor the inserting logic and make it robust but never found time for that task :disappointed: 

My question is here, @JoshuaBatty and @tritao I hope you can provide some more insights, what will happen if `items` contain types that should have `source_id` like e.g. structs. Here we are setting `source_id` to `None` for all types. Can this cause any issues in the garbage collection or can it be ignored?

The `new_tuple` function is new and currently used only to create `(ptr, u64, u64)` where the `source_id` is `None` for all tuple elements. But since it is a general purpose function that can be used in future, I just want to be sure that we are not letting the door ajar for strange bugs down the road.

sway-core/src/type_system/info.rs
```diff
@@ -1526,6 +1526,246 @@ impl TypeInfo {
...
+                type_id,
+                span: Span::dummy(),
+                call_path_tree: None,
+            })
+            .collect();
+        TypeInfo::Tuple(items)
+    }
+
+    /// Calculate the needed buffer for "abi encoding" the self type. If "inside" this
+    /// type there is a custom AbiEncode impl, we cannot calculate the buffer buffer.
```
ironcev, line 1546:
```suggestion
    /// type there is a custom AbiEncode impl, we cannot calculate the buffer.
```

ironcev (changes requested):

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -524,16 +524,95 @@ impl ty::TyExpression {
...
+                                span: Span::dummy(),
+                            },
+                            arguments: vec![the_configurable],
+                        }),
+                        span: Span::dummy(),
+                    };
+                    let as_slice = Expression {
+                        kind: ExpressionKind::Asm(Box::new(AsmExpression {
+                            registers: vec![AsmRegisterDeclaration {
+                                name: Ident::new_no_span("xxx".into()),
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
```diff
@@ -2496,19 +2496,40 @@ fn configurable_field_to_constant_declaration(
...
+
+    let type_ascription = ty_to_type_argument(context, handler, engines, configurable_field.ty)?;
+
+    let value = expr_to_expression(context, handler, engines, configurable_field.initializer)?;
+    let value = if context.experimental.new_encoding {
+        let call_encode =
+            ExpressionKind::FunctionApplication(Box::new(FunctionApplicationExpression {
+                call_path_binding: TypeBinding {
+                    inner: CallPath {
+                        prefixes: vec![],
```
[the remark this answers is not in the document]
xunilrj: The idea is the same as in other places. We want to enable custom implementation of this. Until we have `#[abi_encode]` attribute, this is the best we can do.

If we enforce the full path now, we will be demanding access to `core`. Something that we are trying to avoid when possible.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -524,16 +524,95 @@ impl ty::TyExpression {
...
+                        span,
+                    };
+
+                    // decode it
+                    let r = Self::type_check_function_application(
+                        handler,
+                        ctx,
+                        TypeBinding {
+                            inner: CallPath {
+                                prefixes: vec![],
```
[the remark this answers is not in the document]
xunilrj: Removed the `unwrap`. Now it fails as normal. The error is not great, but we only want the option of being able to do it. I don´t think we actually anyone using this for now (until we implement the special attributes).

```
        48 |     U16: u16 = 16,
             |          ^^^ "abi_decode" does not take type arguments.
```

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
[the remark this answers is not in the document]
ironcev: Ok. I'll stop asking this question in the future I promise :smile: But my bewilderment with this design choice remains **huge**. Dependency on `core` we anyhow have in `contract`s, `script`s, and `predicate`s unless someone wants to write its own `entry` and _everything_ around encoding. We do not know if such need or wish will ever exist, but still opt to render every function named `encode` as invalid at this point, essentially `encode` being a reserved word for function names (`abi_decode` as well, but being not so common word...).

What would cost us to insist on `core::codec` until _eventually_ someone actually really asks for custom implementation and then we either implement and document `#[abi_encode]` or just temporary remove prefixes?

Enough ranting, no need to answer on this one :smile:

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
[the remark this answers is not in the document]
ironcev: Compiler panicking was definitely an issue. This is not great as you mention but much better.

> I don´t think we actually anyone using this for now (until we implement the special attributes).

I am sure we will not have anyone using this for now and in the future (see my ranting above :smile:).

sway-core/src/type_system/info.rs
```diff
@@ -1526,6 +1526,246 @@ impl TypeInfo {
...
+}
+
+impl AbiEncodeSizeHint {
+    fn range_from_min_max(a: AbiEncodeSizeHint, b: AbiEncodeSizeHint) -> AbiEncodeSizeHint {
+        match (a, b) {
+            (AbiEncodeSizeHint::CustomImpl, _) => AbiEncodeSizeHint::CustomImpl,
+            (_, AbiEncodeSizeHint::CustomImpl) => AbiEncodeSizeHint::CustomImpl,
+            (AbiEncodeSizeHint::PotentiallyInfinite, _) => AbiEncodeSizeHint::PotentiallyInfinite,
+            (_, AbiEncodeSizeHint::PotentiallyInfinite) => AbiEncodeSizeHint::PotentiallyInfinite,
+            (AbiEncodeSizeHint::Exact(l), AbiEncodeSizeHint::Exact(r)) => {
```
ironcev, line 1654: If `a > b` we will get a `Range` whose min is larger then max. I assume we count on valid usage of the function and don't want to return `Result` or swap `a` and `b` or.... Suggestion is to have at least an `assert` here to catch an invalid usage? Same for all the arms that return `Range`.

sway-ir/src/optimize/dce.rs
```diff
@@ -58,6 +58,13 @@ fn is_removable_store(
         InstOp::MemCopyBytes { dst_val_ptr, .. }
         | InstOp::MemCopyVal { dst_val_ptr, .. }
         | InstOp::Store { dst_val_ptr, .. } => {
+            // if the pointer comes from unknown sources, it is unsafe to delete the store
```
ironcev, line 61: As discussed, this special case check can now be removed. The handling of `Incomplete` `get_reffered_symbols` result below treats this and other escapes.

ironcev (changes requested):

sway-ir/src/optimize/dce.rs
xunilrj: Done.

sway-core/src/type_system/info.rs
xunilrj: Done. Now all range creation have an assert.

ironcev (dismissed):

sway-core/src/type_system/info.rs
ironcev: Thanks @tritao! On the long term, the solution will very likely be refactoring `insert` to take care of those details plus special behavior for autogenerated code if needed. Right now, since `new_tuple` is used only once for types that do not have source, it looks to me that the easiest way to stay safe is to remove the function from `TypeInfo` and have it as a local private function in the file where it is actually used to create the buffer tuple `(ptr, u64, u64)`.

@xunilrj What do you think?

xunilrj: Done.

JoshuaBatty: Thanks for checking in about the correct usage here @ironcev! Agree with @tritao's advice.

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
```diff
@@ -154,8 +154,9 @@ fn type_check_encode_buffer_empty(
             TypeInfo::UnsignedInteger(IntegerBits::SixtyFour),
             TypeInfo::UnsignedInteger(IntegerBits::SixtyFour),
         ],
+        span.clone(),
     );
-    let return_type = type_engine.insert(engines, return_type, None);
+    let return_type = type_engine.insert(engines, return_type, span.source_id());
```
ironcev, line 159: Hmmm, I am not sure that this is the right solution, actually in this case `None` should be here. This is the very confusing thing about `source_id` parameter, it is not clear out of the parameter itself what should be given here.

The way I understood it, out of the usage and also out of our internal discussion at the time, `source_id` should be specified if the inserted type has a declaration that is in a particular source file, and then the `source_id` of that file should be specified, being a real file or autogenerated. Which is usually something like `decl.name.span().source_id()` or whichever `Span` is in the declaration file. For alias types, it's the file where the alias is declared etc.

Types like `Unknown`, `u64`, ... should have `source_id` set to `None`, as well as tuples combined only of those types, like above.

@tritao We will again need your help :smile:

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
```diff
@@ -166,30 +167,32 @@ fn type_check_encode_buffer_empty(
...
-    let raw_ptr = engines.te().insert(engines, TypeInfo::RawUntypedPtr, None);
+fn encode_buffer_type(engines: &Engines, span: Span) -> TypeInfo {
+    let raw_ptr = engines
+        .te()
+        .insert(engines, TypeInfo::RawUntypedPtr, span.source_id());
     let uint64 = engines.te().insert(
         engines,
         TypeInfo::UnsignedInteger(IntegerBits::SixtyFour),
-        None,
+        span.source_id(),
```
ironcev, line 177: Same here and on other places below. `u64`, `Unknown`, `Error` etc should have `None` as `source_id` . In all of the changes except the last one, I see `None` as the right input. Also, I see no point in specifying the same `span` for all the different type arguments because those types `raw_ptr` and `u64` do not have spans pointing at their declarations.

sway-core/src/type_system/info.rs
```diff
@@ -1527,15 +1527,20 @@ impl TypeInfo {
...
+        engines: &Engines,
+        items: impl IntoIterator<Item = TypeInfo>,
+        span: Span,
+    ) -> TypeInfo {
+        let source_id = span.source_id();
         let te = engines.te();
         let items = items
             .into_iter()
-            .map(|x| te.insert(engines, x, None))
+            .map(|x| te.insert(engines, x, source_id))
```
ironcev, line 1539: As commented above, in a general case of usage of `new_tuple`, all different `x` types can be coming from different source files or like `Unknown` and `u64` have no source file attached. Putting the same `source_id` for potentially completely different types out of how I understood the discussion can cause issues in LSP. What I mean is, each individual `source_id` depends on each individual `x` we are inserting, and not on the resulting tuple.

I am still of opinion that the simplest solution in this case, as long as we do not make `insert` more robust in regard to exact semantics of `source_id`, is removing potentially problematic `new_tuple` from `TypeInfo` and putting it to buffer handling, eventually even renaming to `new_buffer_tuple_type` to make it clear it is not generic.

ironcev (changes requested):

sway-core/src/type_system/info.rs
xunilrj: Not sure what was the idea then. I am reverting the last commit, and making "new_tuple" specific for the encoding case.

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
xunilrj: I reverted this change.

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
xunilrj: I reverted this change

sway-core/src/semantic_analysis/ast_node/expression/intrinsic_function.rs
```diff
@@ -92,10 +92,185 @@ impl ty::TyIntrinsicFunctionKind {
...
+    let kind = ty::TyIntrinsicFunctionKind {
+        kind,
+        arguments: vec![buffer_expr],
+        type_arguments: vec![],
+        span,
+    };
+    Ok((kind, return_type))
+}
+
+pub fn new_encoding_buffer_tuple(
```
ironcev, line 138:
```suggestion
// TODO: Rename to `new_tuple` and move to `TypeInfo` once https://github.com/FuelLabs/sway/issues/5991 is implemented.
fn new_encoding_buffer_tuple(
```

ironcev (approved):
This change will work for now. I've opened an issue that should tackle the `source_id` #5991  and linked it in the TODO. Also, the function shouldn't be `pub`. Otherwise great work and looking forward to get the PR merged and also to the next steps!

ironcev (approved):

Issue #3487, opened 2022-12-01
Allow reexporting items in Sway


nfurfaro:
It would be really nice for code organization to be able to use the
`pub use x;` / `pub use x as y;` type syntax.

IGI-111:
Duplicate of #3113

PR #5963, opened 2024-05-06
Name clash in star imports should result in error when the name is used

### LINKED ISSUE #5940: Ambiguous star imports resolved incorrectly
`a.sw`

```
library;

struct MyStruct { ... }
```

`b.sw`

```
library; 

struct MyStruct { ... }
```

`main.sw`

```
script;

use a::*;
use b::*;  // MyStruct is now ambiguous, which is fine as long as it's not used

fn main() {
  let x = MyStruct { ... };  // This should be an error, but resolves to b::MyStruct
}
```

jjcnn:
## Description

Fixes #5940. 

This PR deals with the situation where the same name is imported by multiple star imports, e.g., 
```
// a.sw
struct X = ...

// b.sw
struct X = ...

// main.sw
use a::*;
use b::*;
```
So far we have resolved this name clash by letting the latter import shadow the former, which is incorrect.

The correct behavior is to allow the import without error, but to report an error if the unqualified name (i.e., `X`) is used (qualified names `a::X` and `b::X` are legal)`. This PR fixes this problem.

Note that if `X` is imported multiple times, but each time is bound to the same entity (e.g., because it is imported both through `core:;prelude` and `core::codec`), then this should not cause an error.

Note also that there is a problem with the implicit imports from the core and std preludes, which in some cases causes the implementation to think that a name is bound to multiple different entities, even though the entities are actually the same. As a workaround anytime a name is imported from either `std` or `core` when it is already imported from `std` or `core`, the new binding replaces the old one instead of being added as a new binding. 

Unfortunately this workaround means that it is not currently possible to redefine and use a name that exists in `std` or `core`, e.g., to create a specialized version of `assert_eq` as is done in one of our tests. To use such a redefinition one must use a qualified name for the entity, which seems like an acceptable workaround. (Note that this only applies if the redefined entity is imported using a star import - if imported using an item import (`use a::X;`), then there is no issue).

Once #3487 is implemented this workaround should no longer be necessary.

## Checklist


IGI-111 (dismissed):
This is mega hacky but it's a step forward and definitely better than what we're doing now.
I appreciate the thoroughness of the tests given how messy imports are.

IGI-111:
Also this is a breaking change and should be labeled as such.

JoshuaBatty (dismissed):

jjcnn:
> Also this is a breaking change and should be labeled as such.

Is it? I think of it as a bugfix.

In any case the change will not change the behavior of programs that successfully compile. Compilation will only fail some programs that used to compile without error.

jjcnn:
Review comments addressed.

There is one outstanding bug: When the name is bound to an enum variant, we report `path_to_enum::variant_name` as a valid path. The correct path is `path_to_enum::enum_name::variant_name`. 

I'll fix this issue later today, and update the failing test accordingly.

jjcnn:
Final issues fixed.

IGI-111 (approved):

PR #5976, opened 2024-05-09
change submodules to use `im::HashMap` instead of `im::OrdMap`

JoshuaBatty:
## Description

The [`im` docs](https://docs.rs/im/latest/im/) state that the performance of `HashMap` is similar to that found in `std` but the `ImOrd` is between 2-3x slower than `BTreeMap`. Changing submodules from using `im::OrdMap` to using `im::HashMap` is giving us a nice performance win. 

I tried compiling the libraries project from fluid protocol as a test, it has 19 seperate libraries. This change gives us a 21.25% speed increase. 

Before: Elapsed time: `27.968193458s`
After:  Elapsed time: `22.02603425s`

Running the compile benchmark in LSP is showing a 29% improvement. 

Before: Elapsed time: `3.8328s`
After: Elapsed time: `2.7086s`

### DIFF
--- a/sway-core/src/semantic_analysis/namespace/module.rs
+++ b/sway-core/src/semantic_analysis/namespace/module.rs
@@ -27,7 +27,7 @@ pub struct Module {
     /// some library dependency that we include as a submodule.
     ///
     /// Note that we *require* this map to be ordered to produce deterministic codegen results.
-    pub(crate) submodules: im::OrdMap<ModuleName, Module>,
+    pub(crate) submodules: im::HashMap<ModuleName, Module>,
     /// Keeps all lexical scopes associated with this module.
     pub lexical_scopes: Vec<LexicalScope>,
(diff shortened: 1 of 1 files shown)


sway-core/src/semantic_analysis/namespace/module.rs
```diff
@@ -27,7 +27,7 @@ pub struct Module {
     /// some library dependency that we include as a submodule.
     ///
     /// Note that we *require* this map to be ordered to produce deterministic codegen results.
```
IGI-111, line 29: I'm just going to point to this comment here so that we don't merge this by mistake.

We may be able to relax this requirement if the codegen evaluation order no longer relies on the container ordering, but if this cannot be ensured, we should very much not do this.

IGI-111 (changes requested):

IGI-111 (approved):

PR #5983, opened 2024-05-10
Make ControlFlowGraph store a reference to Engines to remove clones

JoshuaBatty:
## Description
Using the same performance test as in #5976 and #5978, this change gives a further 21.33% performance improvement. 

Before: Elapsed time: `14.653166459s`
After: Elapsed time: `11.527667375s`


kayagokalp (approved):
🚀

PR #5987, opened 2024-05-10
feat: take dynamic asset id into account with `forc-client`

kayagokalp:
## Description
Base asset ids are dynamic for the newer fuel-core versions. This PR adds capability to query that and use it for forc-client transactions. Also adds a flag `max-fee` which is required to be set for newer versions of fuel-core.

### DIFF
--- a/forc-plugins/forc-client/src/util/tx.rs
+++ b/forc-plugins/forc-client/src/util/tx.rs
@@ -4,9 +4,7 @@ use anyhow::{Error, Result};
 use async_trait::async_trait;
 use forc_tracing::println_warning;
 use fuel_crypto::{Message, PublicKey, SecretKey, Signature};
-use fuel_tx::{
-    field, Address, AssetId, Buildable, ContractId, Input, Output, TransactionBuilder, Witness,
-};
+use fuel_tx::{field, Address, Buildable, ContractId, Input, Output, TransactionBuilder, Witness};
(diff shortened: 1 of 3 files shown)


forc-plugins/forc-client/src/util/tx.rs
```diff
@@ -311,6 +309,8 @@ impl<Tx: Buildable + field::Witnesses + Send> TransactionBuilderExt<Tx> for Tran
                 e
             })?;
 
+        // Set max_fee from consensus_parameters.
```
sdankel, line 312: Is that what the section below is doing? I don't see a reference to `max_fee` in lines 314-326

sdankel (commented):
Could we add a unit test or two for this?

sdankel (dismissed):

forc-plugins/forc-client/src/util/tx.rs
kayagokalp, line 312:
```suggestion
```

kayagokalp: Opps nice catch thanks

kayagokalp:
> Could we add a unit test or two for this?

Opened an issue to track but relaying here as well, unfortunately it is very complicated to write complicated tests to forc-client as it requires multiple different network instantiations, hopefully once we have some headroom we can create this infra and test going forward

JoshuaBatty (approved):

PR #5990, opened 2024-05-11
feat: fee estimation for forc-deploy and point --testnet to devnet

kayagokalp:
## Description

Estimation is hard to test as it is dependent on the network configuration but this is nearly the same thing we have been doing for scripts.

1. `--testnet` points to devnet.
2. For the new `max_fee` requirement, adds an estimation phase with dry runing the transaction
3. Various consts are pointed to relevant devnet endpoints.

Most visible outcome is that `forc-deploy --testnet` can deploy to devnet without any other flags etc.


forc-plugins/forc-client/src/util/gas.rs
```diff
@@ -34,18 +40,73 @@ pub(crate) async fn get_gas_used(mut tx: Script, provider: &Provider) -> Result<
...
-    let gas_used = provider
-        .dry_run_and_get_used_gas(tx.clone().into(), tolerance)
-        .await?;
+    get_gas_used(Transaction::Script(tx), provider).await
+}
 
+/// Returns gas_used for an arbitrary tx, by doing dry run with the provided `Provider`.
+pub(crate) async fn get_gas_used(tx: Transaction, provider: &Provider) -> Result<u64> {
+    let tolerance = 0.1;
+    let gas_used = provider.dry_run_and_get_used_gas(tx, tolerance).await?;
```
kayagokalp, line 57: I initially wanted to use this but SDK either does not support using this routine with Create transactions or they have a bug.

kayagokalp:
Requesting compiler review as well as this is blocking the release.

forc-plugins/forc-client/src/util/gas.rs
sdankel: Is there a tracking issue with the SDK for this? It would be good to add a TODO comment with a link to the issue.

sdankel (approved):

forc-plugins/forc-client/src/util/gas.rs
kayagokalp: There is none, because I just discovered this while working on this. I will open an issue shortly

Voxelot (approved):

PR #5993, opened 2024-05-13
Bump to v0.57.0

IGI-111:


JoshuaBatty (approved):

kayagokalp:
We should ideally also have https://github.com/FuelLabs/sway/pull/5990 as it adjusts defaults for this release. If we cannot solve estimation issue in time, we can cut this one and patch it later on to unblock devrel.

xunilrj (approved):

PR #6002, opened 2024-05-13
Target from chain_info

sdankel:
Abstracts the current "--testnet" network further so we don't need to update it in so many places. Also gets the faucet URL from the network from chain_info rather than always using the testnet faucet URL.


kayagokalp (approved):
Thanks!

PR #6019, opened 2024-05-15
feat: point --testnet flag to latest testnet instead of `devnet`

kayagokalp:
## Description

Changes the target of `--testnet` flag to testnet instead of devnet. Also fixes some missing from end point url conversions for target resolving.


sdankel (dismissed):

kayagokalp:
Ahh thanks for fixing the failing test, looks like i missed it before pushing.

sdankel (approved):

JoshuaBatty (approved):

PR #6025, opened 2024-05-16
Improve IR printing CLI in `forc`

ironcev:
## Description

This PR:
- implements detailed control over printing of IR:
  - initial IR
  - IR after choosen optimization steps, with option to print only the steps that have actually modified the IR
  - final IR
- improves printed IRs by removing the confusing empty modules printed for every external library dependency.
- harmonizes names and descriptions of the optimization passes. The proposed convention is documented in the _sway-ir/src/optimize/README.md_ file.
- harmonizes help comments of the shared compiler options (consistent separation between option title and additional info, wording, etc.)

## Demo

```
'Print initial and final IR together with IR after each optimization step.
--ir all

'Print initial and final IR together with IR after each optimization step that has modified the IR.
--ir all modified

'Print only the final IR.
--ir final

'Print the IR after every "dce", "sroa", and "inline" optimization step.
--ir dce sroa inline

'Print the IR after every "dce", "sroa", and "inline" optimization step that has modified the IR.
--ir dce sroa inline modified
```

### CLI

Instead of the existing `--ir` option without parameters we now have the same CLI option but with parameters. The existing `--ir` is equivalent to the new `--ir final`.

### Build Profile

Instead of the existing `print-ir` of type bool, we now have the same `print-ir` build profile option with parameters. E.g.:

```
print-ir = { initial = true, final = false, modified = true, passes = ["dce", "sroa"]}
```

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2002,6 +2002,7 @@ dependencies = [
  "serde_json",
  "sway-core",
  "sway-error",
+ "sway-ir",
  "sway-types",
  "sway-utils",
  "term-table",
--- a/docs/book/src/forc/manifest_reference.md
+++ b/docs/book/src/forc/manifest_reference.md
@@ -91,12 +91,12 @@ name = "wallet_contract"
 
 [build-profile.debug]
 print-asm = { virtual = false, allocated = false, final = true }
-print-ir = false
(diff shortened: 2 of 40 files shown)


kayagokalp (approved):
Tooling side looks good to me!

JoshuaBatty (approved):

Issue #5498, opened 2024-01-21
Fix performance and semantic problem when creating `Module`s and make `Module::name` mandatory 


ironcev:
The `Namespace::enter_submodule` method creates a new submodule by copying the entire content of the `Namespace::init` into it. `init` contains all the external modules with all their submodules, items, etc.

This causes an obvious performance issue because literally every internal module contains all the modules of all the external modules recursively. Even within a simple project depending only on `std` we will still have all the `core` modules copied as submodules of all the `std` modules. In a typical workspace this will be true for all the external dependencies blowing up the memory and slowing down the compilation time because of extensive copying of `Module`s.

Beside the performance problem, having all the external modules as submodules of each and every module also leads to nonsensical imports like these, which are currently all valid:

```Sway
mod my_internal_mod;

use std::core::ops::Eq;
use std::hash::core::ops::Eq;
use my_internal_mod::my_submod::std::hash::core::ops::Eq;
use some_external_dependency::its_submodule::std::core::ops::Eq;
use my_internal_mod::my_submod::some_external_dependency::its_submodule::std::core::ops::Eq;
etc.
```

To fix this, we need to, obviously, stop copying the entire modules as submodules and refactor the way preludes are imported into namespace/module items.

As a part of that refactoring, one more point should be addressed. Currently, `Module::name`s are `Option`s which does not correspond to the semantics of the modules - we do not have nameless modules. Thus every module must have a name, and this is currently not enforced by the type system. On the contrary, have the `name` being `Option<Ident>` convey the message that we support nameless modules.

Also, the root module must have a name coming from the package name. Making sure that the name actually exists is done by the callers, e.g., tooling and with internal methods like the mentioned `Namespace::enter_submodule`. Still, we had places in tests where root modules are created without a name. Having a root module without a name makes absolute paths ambiguous, because we can have internal libraries with the same name as some external package and the same inner module structure. Currently are such two indistinguishable.

Why do we have then `Module::name`s as `Option` in the first place? After going through the code, my conclusion is that it was done to support the `Namespace::init` being `Module`. Since it was not a real module, there was no sense to have a name for it.

IMO, `Module`a must have `name`, `visibility`, and `is_external` set at creation.

To sum it up, fixing the overall issues with the namespaces and modules should:
- Make `Module::name` being `Ident` instead of `Option<Ident>`
- Remove `Module::default()` and introduce `Module::new(name, visibility, is_external`) and make these fields read-only, accessible through getters.
[...]

Issue #4700, opened 2023-06-22
Replace Path with CallPath in module.rs


esdrubal:
Methods `star_import`, `star_import_with_reexports`, `self_import`, `variant_import`, `variant_star_import` can be changed to receive `CallPath` instead of `Path`.

https://github.com/FuelLabs/sway/blob/00982e477ede7f94bb0bbf2ffd2576be631123b7/sway-core/src/semantic_analysis/namespace/module.rs#L234-L239

The added `CallPath` should also bundle the is_absolute added by #4697. 
Also consider changing `use_synonyms` to use a `CallPath`.

Issue #5864, opened 2024-04-16
Using absolute paths with the new encoding will result in an error


hal3e:
Using absolute paths with the new encoding will result in an error if different types have the same name.

```
contract;

mod contract_a_types;
mod another_lib;

use another_lib::VeryCommonNameStruct;

abi MyContract {
    fn test_function(arg: contract_a_types::VeryCommonNameStruct) -> VeryCommonNameStruct;
}

impl MyContract for Contract {
    fn test_function(_arg: contract_a_types::VeryCommonNameStruct) -> VeryCommonNameStruct {
        VeryCommonNameStruct { field_a: 10u32 }
    }
}
```
generates the following error:
```
3 |
4 |                     let args: (VeryCommonNameStruct,) = decode_second_param::<(VeryCommonNameStruct,)>();
5 |                     let result_test_function: raw_slice = encode::<VeryCommonNameStruct>(__contract_entry_test_function(args.0));
  |                                                                                                                         ^^^^^^ Mismatched types.
expected: VeryCommonNameStruct
found:    VeryCommonNameStruct.
help: The argument that has been provided to this function's type does not match the declared type of the parameter in the function declaration.
```

jjcnn:
Steps to recreate, as described in #6044 by @hal3e :


I tried it with this branch but I still get an error. Unfortunately, I do not get any specifics just: `error: Could not generate the entry method. See errors above for more details.` and there is nothing above :/. 
Steps to recreate: 
- clone `fuels-rs` and checkout `hal3e/name-clash-abi-encode`
- cd into `fuels-rs/e2e/sway/bindings/type_paths`
- `forc build` or `forc build --json-abi-with-callpaths`

P.S. I installed forc with: 
` cargo install --locked forc --force --git https://github.com/FuelLabs/sway --branch "jjcnn/name_clash_struct_def_glob_import"`

PR #6044, opened 2024-05-22
Fix issue with name clash on auto implementation of AbiEncode and AbiDecode

### LINKED ISSUE #5500: Wrong `TyStructDecl::call_path` when using glob imports with local shadowing
#5418 fixed the issue of finding the proper decls in the namespace in case of shadowing, but it turns out we still have issues in the content of the found decls. The concrete example is having `TyStructDecl::call_path` containing wrong path, namely not the one to the local struct but the one to the shadowed struct, although the rest of the decl contains valid information.

To observe the wrong `call_path`, we need to debug the compiler. As a sample tryout project, create a new Sway script with an internal library _lib.sw_ containing the following code:

```Sway
library;

pub struct TestStruct {
    a: u64,
}
```
Put the following code in _main.sw_:

```Sway
script;

mod lib;

use lib::*;

struct TestStruct {
   x: u64,
   y: u64,
}

fn main() {
   let ts = TestStruct { x: 0, y: 0 };
   poke(ts.x);
}

fn poke<T>(_x: T) { }
```
The program will compile without errors, thus resolving the local `TestStruct` as fixed in #5418.

To observe the erroneous behavior, go to _sway-core/src/type_system/info.rs_, to `TypeInfo::apply_subfields` and add the following debugging lines into the second `match` arm `Some(field) =>`:

```Rust
if decl.call_path.suffix.as_str() == "TestStruct" {
    dbg!(&field.name);
    dbg!(&decl.call_path.prefixes);
    for field in decl.fields.iter() {
        dbg!(&field.name);
    }
    dbg!(&decl.span.as_str());
}
```

The output will be as follows:

```
[sway-core/src/type_system/info.rs:1274] &field.name = x
[sway-core/src/type_system/info.rs:1275] &decl.call_path.prefixes = [
    plays,
    lib,    // !!!!!!! Wrong path! This is the path to the lib TestStruct.
]
[sway-core/src/type_system/info.rs:1277] &field.name = x
[sway-core/src/type_system/info.rs:1277] &field.name = y
[sway-core/src/type_system/info.rs:1279] &decl.span.as_str() = "struct TestStruct {\n   x: u64,\n   y: u64,\n}"
```
Note that the fields are valid, and that also the span points to the local `TestStruct`. 

Comment out the `TestStruct` in the _lib.sw_ and rebuild the project. This time, as expected, the `call_path` will contain the valid path to the local `TestStruct` in _main.sw_.

```
[sway-core/src/type_system/info.rs:1274] &field.name = x
[sway-core/src/type_system/info.rs:1275] &decl.call_path.prefixes = [
    plays,
]
[sway-core/src/type_system/info.rs:1277] &field.name = x
[sway-core/src/type_system/info.rs:1277] &field.name = y
[sway-core/src/type_system/info.rs:1279] &decl.span.as_str() = "struct TestStruct {\n   x: u64,\n   y: u64,\n}"
```
Fixing this issue will also need to check:
- [ ] if there are other struct decl fields which are wrongly populated
- [ ] consistency of other items behind structs:
  - [ ] enums
  - [ ] impls
  - [ ] functions
  - [ ] consts

jjcnn:
## Description

Fixes #5500 . Also fixes the issue that caused one of @xunilrj 's tests to fail.

The callpaths generated for struct declarations have so far been incorrect. The `shadowed_glob_imports` test case shows how this causes a compilation error when there is a name clash between a star imported struct and a locally declared struct. This causes a number of issues, in particular that the generated trait impls of `AbiEncode` and `AbiDecode` can refer to the wrong type or to a type that doesn't exist.

The incorrect paths were the result of a combination of two bugs:
1. The `is_external` flag on the containing modules was often set incorrectly. In particular, `core` often had `is_external = false`, leading to the package name being added to the path of anything declared in `core`.
2. When typechecking a struct/enum declaration we would look up paths in the current environment before the name of the struct/enum was bound in the environment. If a type with the same name had already been imported this lookup would result in a path to the import source module rather than the path to current module (which is obviously where the struct/enum is located).

This PR fixes both bugs:
1. When initializing the root namespace we now ensure that all modules that have already been added are marked with `is_external = true`. This happens in `Namespace::init_root()`.
2. When typechecking a type declaration we generate the callpath directly from the name of the type and the path of the current module. This happens in `CallPath::ident_to_fullpath()`.

Step 1 unfortunately adds an extra cloning step in the `namespace` module, but I don't think it's avoidable at the moment. To avoid this extra cloning step we would need to have a better way of building the external module structure than we currently have. #5498 tracks this and other issues.

The improved callpath generation should move us a step closer towards fixing #4700.

## Checklist


ironcev:
@jjcnn Setting `is_external = <current is external>` is a valid change as long as I can see. But it looks like it triggered some other issues which apparently rely on the existing setting of `is_external` to `false`.

I've taken a look a look at the failing tests for references. The failures has nothing to do with references as such. They are caused by the facts that all of those tests have a separate file with traits defined in them (_impls.sw_), as well as trait implementations. Those traits now do not get properly resolved.

Also, there are other E2E tests failing like:
```
should_pass/language/use_absolute_path
should_pass/language/generic_impl_self_where
```
They also have traits defined in separate files and used in _main.sw_.

To the E2E tests, the `go_to_definition_for_paths` LSP tests is also failing perhaps for the same reason.

It looks to me that the simple one line change triggered some other issues related to trait import and resolution.

One thought crossing my mind. The opening sentence in the PR description:
> The callpaths generated for struct declarations are incorrect
reminds me a lot of #5500. Perhaps it makes sense to take a closer look on that issue and fix it first.

jjcnn:
@ironcev : Thanks - that helps quite a bit. I'll take a look at the issue you linked to, and keep digging.

jjcnn:
@ironcev : Sadly, a fix of #5500 does not solve the problem, but I suspect it's just another one of many interlocking issues, so I've pushed a fix and continuing from here.

IGI-111 (approved):
Nice!
I appreciate how well commented this is given how unintuitive the callpaths can be. :+1:

hal3e (commented):
I tried it with this branch but I still get an error. Unfortunately, I do not get any specifics just: `error: Could not generate the entry method. See errors above for more details.` and there is nothing above :/. 
Steps to recreate: 
- clone `fuels-rs` and checkout `hal3e/name-clash-abi-encode`
- cd into `fuels-rs/e2e/sway/bindings/type_paths`
- `forc build` or `forc build --json-abi-with-callpaths`

P.S. I installed forc with: 
` cargo install --locked forc --force --git https://github.com/FuelLabs/sway --branch "jjcnn/name_clash_struct_def_glob_import"`

jjcnn:
> I tried it with this branch but I still get an error. Unfortunately, I do not get any specifics just: `error: Could not generate the entry method. See errors above for more details.` and there is nothing above :/. Steps to recreate:
> 
>     * clone `fuels-rs` and checkout `hal3e/name-clash-abi-encode`
> 
>     * cd into `fuels-rs/e2e/sway/bindings/type_paths`
> 
>     * `forc build` or `forc build --json-abi-with-callpaths`
> 
> 
> P.S. I installed forc with: ` cargo install --locked forc --force --git https://github.com/FuelLabs/sway --branch "jjcnn/name_clash_struct_def_glob_import"`

Doing the above does indeed still fail. I originally tried to recreate your test case within the sway repo where it seemed to compile, but it no longer does, so there's something else going on.

I'll remove the references to #5864 from this PR, and add your recreation steps to the issue.

kayagokalp (approved):

Issue #6068, opened 2024-05-28
Forc deploy should support (and enhance the userflow while) using proxy contracts


kayagokalp:
We have an rfc: https://github.com/FuelLabs/sway-rfcs/pull/39 that describes the desired behaviour.  Basically we want forc-deploy to be aware of contracts with proxy contract requirements and support deploying a proxy contract automatically if it is enabled. If no proxy contract was already deployed for the contract we will deploy a new "default" proxy implementation. Otherwise we will update the already deployed proxy contract.

sdankel:
Blocked on fuel-core 0.27 upgrade in sway

sdankel:
Blocked on https://github.com/FuelLabs/fuels-rs/pull/1396

kayagokalp:
closed by #6069

PR #6069, opened 2024-05-28
feat: create and deploy a reference proxy contract for contracts with `[proxy]` enabled

kayagokalp:
## Description
Part of #6068. 

This PR adds couple of things:

1. A default proxy contract implementation taken from [sway standards](https://github.com/FuelLabs/sway-standards/blob/master/standards/src/src14.sw).
2. Infra for creating, building and deploying the reference implementation for proxy contracts.
3. Deployment procedure such that proxy contract is deployed while working on a contract which enables the `[proxy]` in its forc.toml. In a way that it is owned by the deployer and the target initially points to implementation contract.
4. Infra for making a contract call into the already deployed proxy contracts to update their targets.
5. Adds a `Building` text to the all forc build invocations to better inform the user about what forc is doing behind the scenes.
6. Removes duplicate forc-wallet password prompts which was very frustrating for the users. Now forc-wallet deployment path only asks for password once.
7. Refactors around how secret_key is selected based on user input
8. Updated docs around forc-client
9. Docs around how to use the proxy feature

## How this works
If the user does not have a proxy table in their forc.toml, nothing changes, same old deployment procedure is followed. Only difference is that this PR improves the ux by removing the need of providing the password multiple times.

If the user has a contract with a proxy table but without an address like: 

```TOML
[project]
authors = ["kaya"]
entry = "main.sw"
license = "Apache-2.0"
name = "impl-contract"

[dependencies]

[proxy]
enabled = true
```
Forc automatically creates a proxy contract based on the reference implementation at [SRC14](https://github.com/FuelLabs/sway-standards/pull/94). Sets its target to the implementation contract, whichever contract enabled the proxy and the owner to the deployer (signing account of the transaction).

If the user has a contract with a proxy table and an address specified like:

```TOML
[project]
authors = ["kaya"]
entry = "main.sw"
license = "Apache-2.0"
name = "impl-contract"

[dependencies]

[proxy]
enabled = true
address = "........."
```
 Forc automatically makes a set target conract call to update the proxy contract's target. Pointing it to the newly deployed impl contract which defines the proxy table.

Generated proxy contracts are stored at `~/.forc/.generated_proxy_contracts/project_name` for housekeeping.


forc-plugins/forc-client/src/util/proxy_contract-abi.json
sdankel, line 1: Could this file be generated rather than static? Thinking about when we have to update the proxy contract in the future, we may forget to re-generate this file.

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -1,8 +1,142 @@
...
+/// Forc.toml for the default proxy contract that 'generate_proxy_contract_src()' returns.
+pub const PROXY_CONTRACT_FORC_TOML: &str = r#"
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "proxy_contract"
+
+[dependencies]
+standards = { git = "https://github.com/FuelLabs/sway-standards/" }
```
sdankel, line 23: Should the version be pinned to avoid unplanned automatic upgrades?

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -1,8 +1,142 @@
...
+}
+
+/// Updates the given package manifest file such that the address field under the proxy table updated to the given value.
+/// Updated manifest file is written back to the same location, without thouching anything else such as comments etc.
+/// A safety check is done to ensure the proxy table exists before attempting to udpdate the value.
+pub(crate) fn update_proxy_address_in_manifest(
+    address: &str,
+    manifest: &PackageManifestFile,
+) -> Result<()> {
+    let mut toml = String::new();
```
sdankel, line 89: It might be cleaner to add this logic in a a `write_to_file()` in `PackageManifest`'s impl, as  there is already a `from_file()` there.

Then this function could be simpler: 

```rust
    pub(crate) fn update_proxy_address_in_manifest(
        address: &str,
        manifest: &PackageManifestFile,
    ) -> Result<()> {
           match manifest.proxy() {
               Ok(_) => {
                   let mut new_manifest = manifest.clone();
                   new_manifest.proxy.unwrap().address = address;
                   new_manifest.write_to_file();
                   Ok(())
               }
               Err(_) => // return error?
           }
}
```

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -113,6 +122,74 @@ fn validate_and_parse_salts<'a>(
     Ok(contract_salt_map)
 }
 
+/// Build a proxy contract owned by the deployer.
+/// First creates the contract project at the current dir. The source code for the proxy contract is updated
+/// with 'owner_adr'.
```
sdankel, line 127:
```suggestion
/// with `owner_addr`.
```

sdankel (commented):
Looking good! Left a few initial comments/questions

forc-plugins/forc-client/src/util/proxy_contract-abi.json
kayagokalp: Yeah, initially i wanted that as well. But fuels-rs abigen macro, needs it in compile time :( we can explore updating it in a build script in a build.rs to keep it updated

forc-plugins/forc-client/src/util/pkg.rs
kayagokalp: https://github.com/FuelLabs/sway-standards/issues/100 -- waiting on this

forc-plugins/forc-client/src/util/pkg.rs
kayagokalp: https://github.com/FuelLabs/sway/issues/6074 -- This PR is getting very big, i will do it on a clean-up run after this one https://github.com/FuelLabs/sway/issues/6074

forc-plugins/forc-client/src/util/pkg.rs
kayagokalp: I think due to urgency of this, we can continue as is. As soon as a release is cut on that PR we can make a hotfix. But i totatlly agree that we should pin this version so that breaking changes to the standard (i don't think that should be possible but in any case) won't break our proxy capabilities. cc @JoshuaBatty for visibility

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -91,3 +141,40 @@ forc-deploy saves the details of each deployment in the `out/deployments` folder
...
+license = "Apache-2.0"
+implicit-std = false
+
+[proxy]
+enabled = true
+```
+
+If there is no `address` field present under the proxy table, like the example above, forc automatically creates a proxy contract based on [SRC14](https://github.com/FuelLabs/sway-standards/pull/94) implementation from [sway-standards](https://github.com/FuelLabs/sway-standards). After generating the proxy contract, the target is set to the current contract, and owner of the proxy is set to the account that is signing the transaction for deployment.
+
+This means that if you simply enable proxy in the Forc.toml, forc will automatically deploy a proxy contract for you and you do not need to make anything manually. After deploying the proxy contract the `address` of it is added into the `address` field of the proxy table.
```
sdankel, line 163: I think it will pass spell-check if it's in a code block:

```suggestion
This means that if you simply enable proxy in the `Forc.toml`, forc will automatically deploy a proxy contract for you and you do not need to make anything manually. After deploying the proxy contract the `address` of it is added into the `address` field of the proxy table.
```

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -91,3 +141,40 @@ forc-deploy saves the details of each deployment in the `out/deployments` folder
...
+entry = "main.sw"
+license = "Apache-2.0"
+implicit-std = false
+
+[proxy]
+enabled = true
+address = "0xd8c4b07a0d1be57b228f4c18ba7bca0c8655eb6e9d695f14080f2cf4fc7cd946" # example proxy contract address
+```
+
+If an `address` is present, forc calls into that contract to update its target instead of deploying a new contract. Since a new proxy deployment adds its own `address` into the Forc.toml automatically, you can simply enable the proxy once and after the initial deployment, forc will keep updating the target accordingly for each new deployment of the same contract.
```
sdankel, line 180:
```suggestion
If an `address` is present, forc calls into that contract to update its target instead of deploying a new contract. Since a new proxy deployment adds its own `address` into the `Forc.toml` automatically, you can simply enable the proxy once and after the initial deployment, forc will keep updating the target accordingly for each new deployment of the same contract.
```

forc-plugins/forc-client/src/util/pkg.rs
sdankel: We can pin to the `rev` without needing to release. But it looks like you made a release 👍

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -1,26 +1,76 @@
...
-## Initializing the wallet and adding accounts
+We offer multiple ways to sign the transaction:
 
-If you don't have an initialized wallet or any account for your wallet you won't be able to sign transactions.
+  1. Sign the transaction via your local wallet using forc-client which integrates with our CLI wallet, forc-wallet.
+  2. Use the default signer to deploy to a local node
+  3. Use forc-wallet to manually sign transactions, and copy the signed transaction back to forc-client.
 
-To create a wallet you can use `forc wallet new`. It will ask you to choose a password to encrypt your wallet. After the initialization is done you will have your mnemonic phrase.
+The easiest and recommended way to interact with deployed networks, such as our testnets is the option 1, using forc-client to sign your transactions which reads your default forc-wallet vault. For interacting with local node, we recommend using the second option, which leads forc-client to sign transactions with the a private key that comes pre-funded in local environments.
```
K1-R1, line 11:
```suggestion
The easiest and recommended way to interact with deployed networks such as our testnets is option 1, using forc-client to sign your transactions which reads your default forc-wallet vault. For interacting with local node, we recommend using the second option, which leads forc-client to sign transactions with the a private key that comes pre-funded in local environments.
```

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -91,3 +141,40 @@ forc-deploy saves the details of each deployment in the `out/deployments` folder
   "deployed_block_id": "0x915c6f372252be6bc54bd70df6362dae9bf750ba652bf5582d9b31c7023ca6cf"
 }
 ```
+
+## Proxy Contracts
+
+`forc-deploy` supports deploying proxy contracts automatically if the it is enabled in the `Forc.toml` of the contract.
```
K1-R1, line 147:
```suggestion
`forc-deploy` supports deploying proxy contracts automatically if it is enabled in the `Forc.toml` of the contract.
```

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -91,3 +141,40 @@ forc-deploy saves the details of each deployment in the `out/deployments` folder
...
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+implicit-std = false
+
+[proxy]
+enabled = true
+```
+
+If there is no `address` field present under the proxy table, like the example above, forc automatically creates a proxy contract based on [SRC14](https://github.com/FuelLabs/sway-standards/pull/94) implementation from [sway-standards](https://github.com/FuelLabs/sway-standards). After generating the proxy contract, the target is set to the current contract, and owner of the proxy is set to the account that is signing the transaction for deployment.
```
K1-R1, line 161:
```suggestion
If there is no `address` field present under the proxy table, like the example above, forc automatically creates a proxy contract based on [SRC-14](https://github.com/FuelLabs/sway-standards/blob/master/SRCs/src-14.md) implementation from [sway-standards](https://github.com/FuelLabs/sway-standards). After generating the proxy contract, the target is set to the current contract, and owner of the proxy is set to the account that is signing the transaction for deployment.
```

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -91,3 +141,40 @@ forc-deploy saves the details of each deployment in the `out/deployments` folder
...
+
+This means that if you simply enable proxy in the Forc.toml, forc will automatically deploy a proxy contract for you and you do not need to make anything manually. After deploying the proxy contract the `address` of it is added into the `address` field of the proxy table.
+
+If you want to update the target of a proxy contract rather than deploying a new one, simply add its `address` in the `address` field, like the following example:
```
K1-R1, line 165:
```suggestion
If you want to update the target of an [SRC-14](https://github.com/FuelLabs/sway-standards/blob/master/SRCs/src-14.md) compliant proxy contract rather than deploying a new one, simply add its `address` in the `address` field, like the following example:
```

K1-R1: Perhaps worth specifying that the proxy is assumed to comply with SRC-14?

K1-R1 (commented):
A few suggestions for the docs, but this is a great feature. The auto updating of the proxy address in the manifest is great

forc-plugins/forc-client/src/util/pkg.rs
kayagokalp: Yeah forgot that we had `rev` option 🤦. Updated to tag as we have it now!

K1-R1 (approved):

kayagokalp:
I had to rebase as there were conflicts.

K1-R1 (approved):

sdankel:
I tried this out on `test/src/e2e_vm_tests/test_programs/should_pass/forc/contract_dependencies/contract_c/Forc.toml`

The first deployment was successful, and the Forc.toml was updated correctly.

The second time I tried deploying, I got this error:
```
error: provider: io error: Response errors; InsufficientMaxFee { max_fee_from_policies: 0, max_fee_from_gas_price: 326087 }
```

However, there are sufficient funds in the account.

The other issue is that, for the second request, it's not clear to the user what they are signing. There should be a message like `Updating proxy_contract` before the user is asked to sign the transaction.

image

kayagokalp:
SDK issues are solved should be able to update now. I also opened #6199 and will open a follow-up to this one to further clean forc-client.

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -1,26 +1,76 @@
...
+
+Asset ID : f8f8b6283d7fa5b672b530cbb84fcccb4ff8dc40f8176ef4544ddb1f1952ad07
+Amount   : 2197245
+---------------------------------------------------------------------------
+
+Please provide the index of account to use for signing: 0
+Do you agree to sign this transaction with fuel12pls73y9hnqdqthvduy2x44x48zt8s50pkerf32kq26f2afeqdwq6rj9ar? [y/N]: y
+
+
+Contract impl-contract Deployed!
```
sdankel, line 39: Rather than `Contract contract_name Deployed!` I think it should match the style of the `Deploying` text. 

```
Deployed contract_name
```

Where `Deployed` is in bold-green

image

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -91,3 +141,40 @@ forc-deploy saves the details of each deployment in the `out/deployments` folder
...
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+implicit-std = false
+
+[proxy]
+enabled = true
+```
+
+If there is no `address` field present under the proxy table, like the example above, forc automatically creates a proxy contract based on [SRC-14](https://github.com/FuelLabs/sway-standards/blob/master/docs/src/src-14-simple-upgradeable-proxies.md) implementation from [sway-standards](https://github.com/FuelLabs/sway-standards). After generating the proxy contract, the target is set to the current contract, and owner of the proxy is set to the account that is signing the transaction for deployment.
```
sdankel, line 161:
```suggestion
If there is no `address` field present under the proxy table, like the example above, `forc` will automatically create a proxy contract based on the [SRC-14](https://github.com/FuelLabs/sway-standards/blob/master/docs/src/src-14-simple-upgradeable-proxies.md) implementation from [sway-standards](https://github.com/FuelLabs/sway-standards). After generating and deploying the proxy contract, the target is set to the current contract, and owner of the proxy is set to the account that is signing the transaction for deployment.
```

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -91,3 +141,40 @@ forc-deploy saves the details of each deployment in the `out/deployments` folder
...
+
+This means that if you simply enable proxy in the `Forc.toml`, forc will automatically deploy a proxy contract for you and you do not need to make anything manually. After deploying the proxy contract the `address` of it is added into the `address` field of the proxy table.
```
sdankel, line 163:
```suggestion
This means that if you simply enable proxy in the `Forc.toml`, forc will automatically deploy a proxy contract for you and you do not need to do anything manually aside from signing the deployment transactions for the proxy contract. After deploying the proxy contract, the its address is added into the `address` field of the proxy table.
```

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -91,3 +141,40 @@ forc-deploy saves the details of each deployment in the `out/deployments` folder
...
+entry = "main.sw"
+license = "Apache-2.0"
+implicit-std = false
+
+[proxy]
+enabled = true
+address = "0xd8c4b07a0d1be57b228f4c18ba7bca0c8655eb6e9d695f14080f2cf4fc7cd946" # example proxy contract address
+```
+
+If an `address` is present, forc calls into that contract to update its target instead of deploying a new contract. Since a new proxy deployment adds its own `address` into the `Forc.toml` automatically, you can simply enable the proxy once and after the initial deployment, forc will keep updating the target accordingly for each new deployment of the same contract.
```
sdankel, line 180:
```suggestion
If an `address` is present, `forc` calls into that contract to update its `target` instead of deploying a new contract. Since a new proxy deployment adds its own `address` into the `Forc.toml` automatically, you can simply enable the proxy once and after the initial deployment, forc will keep updating the target accordingly for each new deployment of the same contract.
```

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -1,26 +1,76 @@
...
 
-To create a wallet you can use `forc wallet new`. It will ask you to choose a password to encrypt your wallet. After the initialization is done you will have your mnemonic phrase.
+The easiest and recommended way to interact with deployed networks such as our testnets is option 1, using forc-client to sign your transactions which reads your default forc-wallet vault. For interacting with local node, we recommend using the second option, which leads forc-client to sign transactions with the a private key that comes pre-funded in local environments.
 
-After you have created a wallet, you can derive a new account by running `forc wallet account new`. It will ask your password to decrypt the wallet before deriving an account.
+## Option 1: Sign transactions via forc-client using your local forc-wallet vault
 
-## Signing transactions using `forc-wallet` CLI
+If you used forc-wallet before, a vault which securely holds your private key is created and written to your file-system in a password encrypted format. forc-client is compatible with forc-wallet such that it can read that vault by asking you your password and use your account to sign transactions.
```
sdankel, line 15:
```suggestion
If you've used `forc-wallet` before, you'll already have a secure, password-protected vault holding your private key written to your file-system. `forc-client` is compatible with `forc-wallet` such that it can read that vault by asking you your password and use your account to sign transactions.
```

forc-plugins/forc-client/src/cmd/deploy.rs
```diff
@@ -57,9 +57,6 @@ pub struct Command {
     pub unsigned: bool,
     /// Set the key to be used for signing.
     pub signing_key: Option<SecretKey>,
-    /// Sign the deployment transaction manually.
```
sdankel, line 60: Why is this removed?

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -113,6 +121,54 @@ fn validate_and_parse_salts<'a>(
...
+async fn deploy_new_proxy(
+    pkg: &BuiltPackage,
+    owner_account_address: &mut Bech32Address,
+    impl_contract: &DeployedContract,
+    build_opts: &BuildOpts,
+    command: &cmd::Deploy,
+    salt: Salt,
+    wallet_mode: &WalletSelectionMode,
+) -> Result<DeployedContract> {
+    println!("  {} proxy contract", "Creating".bold().green());
```
sdankel, line 133:
```suggestion
    info!("  {} proxy contract", "Creating".bold().green());
```

Perhaps we should also have it say `Creating proxy contract at ~/.forc/.generated_proxy_contracts/path_to_contract`

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -197,12 +261,80 @@ pub async fn deploy(command: cmd::Deploy) -> Result<Vec<DeployedContract>> {
                     bail!("Both `--salt` and `--default-salt` were specified: must choose one")
                 }
             };
-            let contract_id =
-                deploy_pkg(&command, &pkg.descriptor.manifest_file, &pkg, salt).await?;
-            contract_ids.push(contract_id);
+            let node_url = get_node_url(&command.node, &pkg.descriptor.manifest_file.network)?;
+            println!(
+                "   {} contract: {}",
```
sdankel, line 266: The indentation is off. I think this is right:

```suggestion
                " {} contract: {}",
```

image

We should create a helper function for this is forc-tracing. Something that takes the "action" and puts it in bold-green with the right indentation.

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -197,12 +261,80 @@ pub async fn deploy(command: cmd::Deploy) -> Result<Vec<DeployedContract>> {
                     bail!("Both `--salt` and `--default-salt` were specified: must choose one")
                 }
             };
-            let contract_id =
-                deploy_pkg(&command, &pkg.descriptor.manifest_file, &pkg, salt).await?;
-            contract_ids.push(contract_id);
+            let node_url = get_node_url(&command.node, &pkg.descriptor.manifest_file.network)?;
+            println!(
```
sdankel, line 265:
```suggestion
            info!(
```

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -197,12 +261,80 @@ pub async fn deploy(command: cmd::Deploy) -> Result<Vec<DeployedContract>> {
...
+            .await?;
+            let proxy = &pkg.descriptor.manifest_file.proxy();
+            if let Some(proxy) = proxy {
+                if proxy.enabled {
+                    if let Some(proxy_addr) = &proxy.address {
+                        // Make a call into the contract to update impl contract address to 'deployed_contract'.
+
+                        // Create a contract instance for the proxy contract using default proxy contract abi and
+                        // specified address.
+                        println!("   {} proxy contract", "Updating".bold().green());
```
sdankel, line 286:
```suggestion
                        info!("   {} proxy contract", "Updating".bold().green());
```

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -113,6 +121,54 @@ fn validate_and_parse_salts<'a>(
...
+                bech32_from_secret(&secret_key)
+            }
+        }
+    }?;
+    let user_addr_hex: fuels_core::types::Address = user_addr.into();
+    let user_addr = format!("0x{}", user_addr_hex);
+    let pkg_name = pkg.descriptor.manifest_file.project_name();
+    let contract_addr = format!("0x{}", impl_contract.id);
+    let proxy_contract = build_proxy_contract(&user_addr, &contract_addr, pkg_name, build_opts)?;
+    println!("   {} proxy contract", "Deploying".bold().green());
```
sdankel, line 161:
```suggestion
    info!("   {} proxy contract", "Deploying".bold().green());
```

The indentation here is off as well

forc-plugins/forc-client/src/util/tx.rs
```diff
@@ -81,6 +180,107 @@ async fn collect_account_balances(
...
+        .set_proxy_target(new_target)
+        .call()
+        .await?;
+    println!(
+        "  0x{} proxy contract's target to {}",
+        "Updated".green(),
+        new_target
+    );
+    Ok(())
+}
```
sdankel, line 282: Is this just refactoring the code that was previously in `finalize_signed`? Or is the behavior changing?

It would be great to add unit tests for these functions

forc-plugins/forc-client/src/util/gas.rs
```diff
@@ -55,56 +51,3 @@ pub(crate) async fn get_script_gas_used(mut tx: Script, provider: &Provider) ->
         .await?;
     Ok(estimated_tx_cost.gas_used)
 }
-
-/// Returns an estimation for the max fee of `Create` transactions.
-/// Accepts a `tolerance` which is used to add some safety margin to the estimation.
-/// Resulting estimation is calculated as `(dry_run_estimation * tolerance)/100 + dry_run_estimation)`.
-pub(crate) async fn get_estimated_max_fee(
```
sdankel, line 62: Is this done automatically by the SDK now?

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -1,8 +1,142 @@
...
+/// Creates a proxy contract project at the given path, adds a forc.toml and source file.
+pub(crate) fn create_proxy_contract(
+    addr: &str,
+    impl_contract_id: &str,
+    pkg_name: &str,
+) -> Result<PathBuf> {
+    // Create the proxy contract folder.
+    let proxy_contract_dir = user_forc_directory()
+        .join(PROXY_CONTRACT_FOLDER_NAME)
+        .join(pkg_name);
```
sdankel, line 113: This generated contract directory stores the ABI that is needed to updated the proxy contract, right?

If so, does it need a unique name, since other contracts could have the same package name?

sdankel (commented):
Left a few comments, mostly about UX

forc-plugins/forc-client/src/util/pkg.rs
kayagokalp: We can also add a timestamp based random id to this, but abi for all contracts are going to be the same, every time this command is run the abi is re-generated. So overriding it is not a problem in terms of functionality of the added feature. The reason i am not deleting this after deployment is just to keep a housekeeping data in the disk to troubleshoot, we can add the random identifier to make sure we have separate entries for each deployment.

forc-plugins/forc-client/src/util/tx.rs
kayagokalp: This is basically moving the inside of finalize signed as a separate function. I will try to add tests for this.

forc-plugins/forc-client/src/util/gas.rs
kayagokalp: Yes

forc-plugins/forc-client/src/op/deploy.rs
kayagokalp: It was intentionally indented to keep the old style, but happy to change it.

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -1,26 +1,76 @@
...
-## Initializing the wallet and adding accounts
+We offer multiple ways to sign the transaction:
 
-If you don't have an initialized wallet or any account for your wallet you won't be able to sign transactions.
+  1. Sign the transaction via your local wallet using forc-client which integrates with our CLI wallet, forc-wallet.
+  2. Use the default signer to deploy to a local node
+  3. Use forc-wallet to manually sign transactions, and copy the signed transaction back to forc-client.
 
-To create a wallet you can use `forc wallet new`. It will ask you to choose a password to encrypt your wallet. After the initialization is done you will have your mnemonic phrase.
+The easiest and recommended way to interact with deployed networks such as our testnets is option 1, using forc-client to sign your transactions which reads your default forc-wallet vault. For interacting with local node, we recommend using the second option, which leads forc-client to sign transactions with the a private key that comes pre-funded in local environments.
```
JoshuaBatty, line 11:
```suggestion
The easiest and recommended way to interact with deployed networks such as our testnets is option 1, using `forc-client` to sign your transactions which reads your default `forc-wallet` vault. For interacting with local node, we recommend using the second option, which leads `forc-client` to sign transactions with the a private key that comes pre-funded in local environments.
```

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -1,26 +1,76 @@
 # `forc-client`
 
-Forc plugin for interacting with a Fuel node.
+Forc plugin for interacting with a Fuel node. Since transactions are going to require some gas, you need to sign them with an account that has enough tokens to pay for them.
 
-## Initializing the wallet and adding accounts
+We offer multiple ways to sign the transaction:
 
-If you don't have an initialized wallet or any account for your wallet you won't be able to sign transactions.
+  1. Sign the transaction via your local wallet using forc-client which integrates with our CLI wallet, forc-wallet.
```
JoshuaBatty, line 7:
```suggestion
  1. Sign the transaction via your local wallet using `forc-client` which integrates with our CLI wallet, `forc-wallet`.
```

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -1,26 +1,76 @@
...
+  2. Use the default signer to deploy to a local node
+  3. Use forc-wallet to manually sign transactions, and copy the signed transaction back to forc-client.
```
JoshuaBatty, line 9:
```suggestion
  3. Use `forc-wallet` to manually sign transactions, and copy the signed transaction back to `forc-client`.
```

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -1,26 +1,76 @@
...
+
+
+Contract impl-contract Deployed!
+
+Network: https://testnet.fuel.network
+Contract ID: 0x94b712901f04332682d14c998a5fc5a078ed15321438f46d58d0383200cde43d
+Deployed in block 00114d4d
+```
+
+As it can be seen from the example, forc-client asks for your password to decrypt the forc-wallet vault, and list your accounts so that you can select the one you want to fund the transaction with.
```
JoshuaBatty, line 46:
```suggestion
As it can be seen from the example, `forc-client` asks for your password to decrypt the `forc-wallet` vault, and list your accounts so that you can select the one you want to fund the transaction with.
```

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -1,26 +1,76 @@
...
+
+## Option 2: Using default signer
+
+If you are not interacting with a deployed network, such as testnets, your local fuel-core environment can be structured such that it funds an account by default. Using `--default-signer` flag with forc-client binaries (run, deploy) will instruct forc-client to sign transactions with this pre-funded account. Which makes it a useful command while working against a local node.
```
JoshuaBatty, line 50:
```suggestion
If you are not interacting with a deployed network, such as testnets, your local `fuel-core` environment can be structured such that it funds an account by default. Using `--default-signer` flag with `forc-client` binaries (run, deploy) will instruct `forc-client` to sign transactions with this pre-funded account. Which makes it a useful command while working against a local node.
```

forc-plugins/forc-client/src/cmd/deploy.rs
kayagokalp: That is a good catch! I will remove this in a follow-up PR as this is not a meaningful anymore but it is a breaking change

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -1,8 +1,142 @@
...
+license = "Apache-2.0"
+name = "proxy_contract"
+
+[dependencies]
+standards = { git = "https://github.com/FuelLabs/sway-standards/", tag = "v0.5.0" }
+"#;
+
+/// Generates source code for proxy contract owner set to the given 'addr'.
+pub(crate) fn generate_proxy_contract_src(addr: &str, impl_contract_id: &str) -> String {
+    format!(
```
JoshuaBatty, line 28: This might be a good spot to use the [indoc crate](https://crates.io/crates/indoc) that @rvcas suggested.

JoshuaBatty:
Thanks Kaya, this is great. Appreciate the thoughtful documentation updates. I think adding more tests might be useful before merging this though. Here's a few suggestions:

* Add an integration test simulating the full workflow of enabling, deploying, and updating proxy contracts.
* Include error case testing to ensure proper handling of failure scenarios (e.g., invalid addresses, updating non-existent proxies).
* Test the interaction between proxy and implementation contracts to verify correct function forwarding.
* Verify the proxy contract's ownership controls, ensuring only the owner can update the target.

sdankel:
Currently, we ask for user input for both the proxy and the main contract deployment transactions. We should be able to just ask for the password, account to use, and user agreement only once.

kayagokalp:
> 



> Currently, we ask for user input for both the proxy and the main contract deployment transactions. We should be able to just ask for the password, account to use, and user agreement only once.

I agree that this is more seamless but this way it is possible to use a different account for paying for the update transaction. If the first deployment depletes the first account, a user would have the ability still continue with the transaction. That being said we can do this tradeoff

sdankel:
> > 
> 
> > Currently, we ask for user input for both the proxy and the main contract deployment transactions. We should be able to just ask for the password, account to use, and user agreement only once.
> 
> I agree that this is more seamless but with current way it is possible to use a different account for paying for the update transaction. If the first deployment depletes the first account, a user would have the ability still continue with the transaction. That being said we can do this tradeoff

For these multi-tx deployments, I think we should also build in a mechanism where if an account runs out of gas after the first deployment, forc-deploy is suspended and gives the user a change to fund the account before continuing, at which point forc-deploy will try that transaction again.

We could also give the user the option to choose a different account before retrying.

Even if the user can choose multiple accounts, they still could run into an issue where they run out of gas and don't want to have to redo the tx's that were already successful.

I think it would also be good to show an estimate of the total amount of gas for all transactions before they agree to sign the first one, so that they can ensure they'll have enough. If we are able to estimate it, we could also display information about which accounts have enough gas for the full deployment. This could be done as a follow up issue.

kayagokalp:
This will need some more love, to be more specific needs two more integration tests:

1. Try and fail to obtain ownership of a proxy with different account
2. Deploy with proxy, make a call,  update proxy's target and confirm that the re-routing routes to the updated contract and the call result changes.

I am currently querying how to check the result of a proxy contract call using the sdk.

kayagokalp:
I guess something that is changing the generated bytecode is merged. Probably an optimization or something which caused contract id to change. Updating the test now

forc-plugins/forc-client/src/util/proxy_contract-abi.json
JoshuaBatty, line 1: Feels a bit strange having this json file floating within a sub directory of a rust project.... Perhaps a dedicated folder for this would be better?

kayagokalp: Sounds good to me! Thanks for pointing it out, fixing it now.

kayagokalp: fixed in [8628280](https://github.com/FuelLabs/sway/pull/6069/commits/862828027ae1a662d665aa8b21e5cc21db71b36a) ✅ 

Also as a side-note I want to do a separate pass over forc-client for general clean-up shortly after we finalize critical stuff before it gets less manageable

forc-plugins/forc-client/src/util/pkg.rs
kayagokalp: I think we have lots of locations already merged that might benefit from this in forc-client. I am planning to do a clean-up pass once we have critical stuff out the door (https://github.com/FuelLabs/sway/pull/6069#discussion_r1673428941), maybe we can add this to every location during that pass?

kayagokalp: mentioned in #6259

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -206,18 +263,89 @@ pub async fn deploy(command: cmd::Deploy) -> Result<Vec<DeployedContract>> {
...
+                            proxy_contract,
+                            deployed_contract_id,
+                        )
+                        .await?;
+                        Some(proxy_contract)
+                    } else {
+                        // Deploy a new proxy contract.
+                        let deployed_proxy_contract = deploy_new_proxy(
+                            &pkg,
+                            &mut owner_account_address,
```
sdankel, line 319: It looks like `owner_account_address` never gets set to anything other than the default.

sdankel:
Another thing I noticed is that when the proxy contract compiles, it has a bunch of warnings from dependency libraries. It would be nice to fix the root cause of these warnings or to silence them for a user trying out the proxy feature.

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -1,26 +1,76 @@
...
+
+Please provide the index of account to use for signing: 0
+Do you agree to sign this transaction with fuel12pls73y9hnqdqthvduy2x44x48zt8s50pkerf32kq26f2afeqdwq6rj9ar? [y/N]: y
+
+
+Contract impl-contract Deployed!
+
+Network: https://testnet.fuel.network
+Contract ID: 0x94b712901f04332682d14c998a5fc5a078ed15321438f46d58d0383200cde43d
+Deployed in block 00114d4d
```
sdankel, line 43: We should update the examples base on the new output of forc-deploy

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -219,18 +249,63 @@ pub async fn deploy(command: cmd::Deploy) -> Result<Vec<DeployedContract>> {
...
+                enabled: true,
+                address: None,
+            }) => {
+                // Deploy a new proxy contract.
+                let deployed_proxy_contract = deploy_new_proxy(
+                    pkg,
+                    &deployed_contract_id,
+                    &build_opts,
+                    &command,
+                    salt,
```
sdankel, line 285: Do we need to calculate the salt again inside of `deploy_new_proxy` rather than reusing the salt from the main contract?

forc-plugins/forc-client/src/util/tx.rs
```diff
@@ -244,6 +249,34 @@ pub(crate) async fn select_secret_key(
...
+    abigen!(Contract(
+        name = "ProxyContract",
+        abi = "forc-plugins/forc-client/abi/proxy_contract-abi.json"
+    ));
+
+    let wallet = WalletUnlocked::new_from_private_key(secret_key, Some(provider.clone()));
+
+    let proxy_contract = ProxyContract::new(proxy_contract_id, wallet);
+
+    // TODO: what happens if the call fails? Does 'FuelCallResponse' is returned as Err() in that case?
```
sdankel, line 267: We should have a test for this. Is there an "invalid" contract ID we could use to trigger a failure?

docs/book/src/forc/plugins/forc_client/index.md
sdankel: Done in https://github.com/FuelLabs/sway/pull/6069/commits/38b3d434d274d4ca980f3e02a80269068ebc2100

forc-plugins/forc-client/src/op/deploy.rs
kayagokalp: Hmm I don't think salt is something we calculate. I believe it is either provided by the user or we generate it randomly, basically something (32 byte num) we use to differentiate different contracts. It is not based on contract itself. We can regenerate a new random salt for proxy but i am not sure what it would add

forc-plugins/forc-client/src/util/tx.rs
kayagokalp: That is a good point! `async fn test_proxy_contract_re_routes_call()` test in the `tests/deploy.rs` is testing the happy path already. We also have failing test that tests this but specifically tests if the owner is not right. So i think we have both happy and failing tests covered. The problem is non owner test is not calling into this function but does the same thing separetly so we should convert it to call into this function to make sure code coverage itself is also detecting it

forc-plugins/forc-client/src/op/deploy.rs
sdankel: Okay, that makes sense! At first I thought it was using something from the `built_pkg` to determine the salt, but it looks like it's only using the package name for `contract_salt_map`. No worries then :)

forc-plugins/forc-client/src/util/tx.rs
sdankel: That's perfect! Done in https://github.com/FuelLabs/sway/pull/6069/commits/17b2bd70e1bd71030545308cffa0d3b15a285cb7

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -1,8 +1,145 @@
...
+
+use std::execution::run_external;
+use standards::src5::{{AccessError, SRC5, State}};
+use standards::src14::SRC14;
+
+/// The owner of this contract at deployment.
+const INITIAL_OWNER: Identity = Identity::Address(Address::from({addr}));
+
+// use sha256("storage_SRC14") as base to avoid collisions
+#[namespace(SRC14)]
```
sdankel, line 40: This syntax is deprecated. There's a new syntax for namespaces: https://github.com/FuelLabs/sway/pull/6064

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -1,8 +1,145 @@
...
+/// Forc.toml for the default proxy contract that 'generate_proxy_contract_src()' returns.
+pub const PROXY_CONTRACT_FORC_TOML: &str = r#"
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "proxy_contract"
+
+[dependencies]
+standards = { git = "https://github.com/FuelLabs/sway-standards/", tag = "v0.5.0" }
```
sdankel, line 23: There's a new version, 0.5.1, with an additional method in the ABI

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -219,18 +256,65 @@ pub async fn deploy(command: cmd::Deploy) -> Result<Vec<DeployedContract>> {
...
+                    &provider,
+                    &signing_key,
+                )
+                .await?;
+
+                println!("--==-- deployed proxy contract {deployed_proxy_contract:?}");
+
```
sdankel, line 299:
```suggestion
```

forc-plugins/forc-client/src/util/tx.rs
```diff
@@ -244,6 +256,34 @@ pub(crate) async fn select_secret_key(
...
```
sdankel, line 274: There's a test for it now :) 

```suggestion
```

sdankel (commented):
This looks good! The prebuilt proxy contract is a great idea. I would like to rename the `abi` folder to `proxy_abi` and add a README file to that folder that briefly explains how this works, links to the commit of the contract code, and says which version of forc and build options were used to build it.

There are a couple of tests failing too.

kayagokalp:
> This looks good! The prebuilt proxy contract is a great idea. I would like to rename the abi folder to proxy_abi and add a README file to that folder that briefly explains how this works, links to the commit of the contract code, and says which version of forc and build options were used to build it.
> 
> There are a couple of tests failing too.

I removed the duplicate test, as it was already being tested as part of re_routes test, and also renamed the abi folder. Added a new readme. Should be ready now.

sdankel (approved):

IGI-111 (approved):

bitzoic (commented):
~Minor nits~ 1 min too late

docs/book/src/forc/plugins/forc_client/index.md
```diff
@@ -1,26 +1,62 @@
...
-## Initializing the wallet and adding accounts
+We offer multiple ways to sign the transaction:
 
-If you don't have an initialized wallet or any account for your wallet you won't be able to sign transactions.
+  1. Sign the transaction via your local wallet using `forc-client` which integrates with our CLI wallet, `forc-wallet`.
+  2. Use the default signer to deploy to a local node
+  3. Use `forc-wallet` to manually sign transactions, and copy the signed transaction back to `forc-client`.
 
-To create a wallet you can use `forc wallet new`. It will ask you to choose a password to encrypt your wallet. After the initialization is done you will have your mnemonic phrase.
+The easiest and recommended way to interact with deployed networks such as our testnets is option 1, using `forc-client` to sign your transactions which reads your default `forc-wallet` vault. For interacting with local node, we recommend using the second option, which leads `forc-client` to sign transactions with the a private key that comes pre-funded in local environments.
```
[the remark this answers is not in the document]
kayagokalp: Oh looks like it got auto merged, I'll address these with the follow-up, thanks!

sdankel:
> ~Minor nits~ 1 min too late

https://github.com/FuelLabs/sway/pull/6403

PR #6098, opened 2024-06-07
Use immediates instead of data section when possible

vaivaswatha:
Loading / initialising constants from the data section is unnecessary (takes up a data section slot) when the constant can instead, just be moved to a register using `MOVi`, if the constant can fit as an immediate value to the instruction.


IGI-111 (approved):

woolworthgralt:
nice

IGI-111 (approved):

xunilrj (approved):

Issue #5789, opened 2024-03-27
Change `forc doc --manifest-path` to `forc doc --path`


bitzoic:
We currently have `forc doc --manifest-path` for generating docs from the inline documentation when specifying the directory to run forc doc. 

This should be changed to `forc doc --path` to match `forc build --path` and `forc fmt --path`. Using the same options across all forc commands will improve dev experience.

PR #6109, opened 2024-06-11
Change forc doc --manifest-path to forc doc --path

alexanderblv:
## Description
Closes https://github.com/FuelLabs/sway/issues/5789

## Checklist


kayagokalp (changes requested):
Hello, thanks for the PR. It seems like you just updated docs and readme, not the command itself.

https://github.com/FuelLabs/sway/blob/24d03f4e9fe13672961bbe696861a99508fa521a/forc-plugins/forc-doc/src/cli.rs#L26

alexanderblv:
> Hello, thanks for the PR. It seems like you just updated docs and readme, not the command itself.
> 
> https://github.com/FuelLabs/sway/blob/24d03f4e9fe13672961bbe696861a99508fa521a/forc-plugins/forc-doc/src/cli.rs#L26

I understand correctly that I have to update `manifest_path` in all sway files?

IGI-111 (dismissed):

sdankel (changes requested):
I would prefer to make `--path` a shorthand for `--manifest-path` (keeping both as valid CLI flags) so as not to break any existing users.

alexanderblv:
> I would prefer to make `--path` a shorthand for `--manifest-path` (keeping both as valid CLI flags) so as not to break any existing users.

So you want to keep both run as valid?
`run: forc doc --manifest-path ./sway-lib-std`
`run: forc doc --path ./sway-lib-std`

I'm not sure if the duplicates will lead to good code performance, but if you think so, then I'll do it.

sdankel (changes requested):
> > I would prefer to make `--path` a shorthand for `--manifest-path` (keeping both as valid CLI flags) so as not to break any existing users.
> 
> So you want to keep both run as valid? `run: forc doc --manifest-path ./sway-lib-std` `run: forc doc --path ./sway-lib-std`
> 
> I'm not sure if the duplicates will lead to good code performance, but if you think so, then I'll do it.

It shouldn't affect the code performance at all. You can use clap's [alias](https://docs.rs/clap/latest/clap/struct.Arg.html#method.alias):

```
    #[clap(long, alias="manifest-path")]
    pub path: Option<String>,
```

Please fix the compilation errors. You need to update `manifest_path` in the other places it is used.

.github/workflows/ci.yml
```diff
@@ -233,6 +233,7 @@ jobs:
         run: |
           cargo install --locked --debug --path ./forc-plugins/forc-doc
       - name: Build sway-lib-std docs
+        run: forc doc --path ./sway-lib-std
         run: forc doc --manifest-path ./sway-lib-std
```
kayagokalp, line 237:
```suggestion
        run: |
          forc doc --path ./sway-lib-std
          forc doc --manifest-path ./sway-lib-std
```

.github/workflows/gh-pages.yml
```diff
@@ -50,6 +50,7 @@ jobs:
         run: mdbook build docs/reference
 
       - name: Build Sway std library
+        run: forc doc --path ./sway-lib-std
```
kayagokalp, line 53:
```suggestion
        run: |
          forc doc --path ./sway-lib-std
          forc doc --manifest-path ./sway-lib-std
```

kayagokalp (changes requested):
Current ci.yml file is invalid, I suggested the correct version of it. But also I think we should not test whole sdk-harness twice just to see that a flag is working with different alias to preserve some valuable CI time. Let's keep the addition to the code to support it for users but remove it from ci. I think @sdankel originally wanted you to add the support to the code not to CI but pinging her in any case.

.github/workflows/gh-pages.yml
```diff
@@ -50,6 +50,7 @@ jobs:
...
         run: forc doc --manifest-path ./sway-lib-std
```
kayagokalp, line 54:
```suggestion
        run: |
          forc doc --path ./sway-lib-std
          forc doc --manifest-path ./sway-lib-std
```

kayagokalp (changes requested):

forc-plugins/forc-doc/src/cli.rs
```diff
@@ -24,6 +24,7 @@ pub struct Command {
     /// file in the current directory or any parent directory.
     #[clap(long)]
     pub manifest_path: Option<String>,
+    pub path: Option<String>,
```
kayagokalp, line 27: We can probably use, @sdankel's suggestion here:

```suggestion
    #[clap(long, alias="manifest-path")]
    pub path: Option<String>,
```

kayagokalp, line 27: We can probably use, @sdankel's suggestion here:

```suggestion
    #[clap(long, alias="manifest-path")]
    pub path: Option<String>,
```

kayagokalp (changes requested):

sdankel:
I absolutely agree with  @kayagokalp , we should not run forc-doc twice in CI. This code still doesn't compile, so I'm going to close this PR. This isn't really close to what we're looking for and I want to leave the opportunity for another contributor to pick up this task.

Issue #6118, opened 2024-06-13
Compiler panics and no DCA warning when a configurable is never used


ironcev:
The configurable `MAX_SUPPLY` is never used and optimized away but there is no dead code warning generated and that confuses the developers.

Additionally, the code compiles without warnings or errors on Sway v0.60.0 but the compiler panics on the current master:
```
thread 'main' panicked at sway-core/src/semantic_analysis/ast_node/declaration/configurable.rs:106:18:
called `Result::unwrap()` on an `Err` value: ErrorEmitted { _priv: () }


```

```Sway
contract;

abi Test {
    fn test() -> bool;
}

configurable {
    MAX_SUPPLY: u64 = 3,
}

impl Test for Contract {
    fn test() -> bool {
        true
    }
}
```

Dr-Maxwell:
Possible Solution
Prevent Optimization of Constants:
Ensure constants like MAX_SUPPLY are not optimized away by using them in a way that the compiler recognizes as necessary. For example:
const MAX_SUPPLY: u64 = 1000;

fn main() {
    println!("MAX_SUPPLY is: {}", MAX_SUPPLY); // Dummy usage to prevent optimization
    // Your application logic
}
Enable Dead Code Warnings:
Ensure dead code warnings are enabled in your development environment to catch unused constants and other dead code. Add the following to Cargo.toml:
[profile.dev]
rustflags = ["-Awarnings"]
Handle Errors Properly:
Avoid using unwrap() on Result types where an Err could occur. Instead, handle errors using match or the ? operator for proper error propagation and handling.

i hope this helps

PR #6121, opened 2024-06-13
better error and dca for configurables

xunilrj:
## Description

This PR closes https://github.com/FuelLabs/sway/issues/6118.

If for some reason an older `core` ends up being used, we now warn with a more user-friendly error.

```
error: Configurables need a function named "abi_decode_in_place" to be in scope.
  --> /home/xunilrj/github/sway/test/src/e2e_vm_tests/test_programs/should_pass/language/configurable_consts/src/main.sw:42:5
   |
...
42 |     ARRAY_BOOL: [bool; 3] = [true, false, true],
   |     ^^^^^^^^^^
   |
   = help: Certify you are accessing a compatible version of the "core" library.
____
```

We also use a better span when warning about unused configurables.

```
warning
  --> /home/xunilrj/github/sway/test/src/e2e_vm_tests/test_programs/should_pass/language/configurable_consts/src/main.sw:47:5
   |
45 | 
46 | 
47 |     NOT_USED: u8 = 1,
   |     -------- This declaration is never used.
48 |     #[allow(dead_code)]
49 |     NOT_USED_BUT_ALLOWED: u8 = 2,
   |
____

  Compiled script "configurable_consts" with 1 warning.
```

## Checklist


sway-error/src/error.rs
```diff
@@ -2430,6 +2433,18 @@ impl ToDiagnostic for CompileError {
                         },
                 }
             },
+            ConfigurableMissingAbiDecodeInPlace { span } => Diagnostic {
+                reason: Some(Reason::new(code(1), "Configurables need a function named \"abi_decode_in_place\" to be in scope.".to_string())),
```
ironcev, line 2437: Great to see expressive diagnostics being used :smile:
According to the guidelines, the "Reason" does not end with fullstop.
```suggestion
                reason: Some(Reason::new(code(1), "Configurables need a function named \"abi_decode_in_place\" to be in scope".to_string())),
```

sway-error/src/error.rs
```diff
@@ -2430,6 +2433,18 @@ impl ToDiagnostic for CompileError {
...
+                issue: Issue::error(
+                    source_engine,
+                    span.clone(),
+                    String::new()
+                ),
+                hints: vec![],
+                help: vec![
+                    "Certify you are accessing a compatible version of the \"core\" library.".into()
```
sdankel, line 2445: nit

```suggestion
                    "Verify that you are accessing a compatible version of the \"core\" library.".into()
```

ironcev, line 2445: Definitely much much better and more helpful then panicking, but since "abi_decode_in_place" is actually never seen by a developer I think they will still be confused and not know how to certify that they are using a compatible version or where exactly to look.

Providing a bit more detailed info, where exactly to look, will definitely help to solve the issue quicker and easier when it happens. E.g:
```suggestion
                    "Function \"abi_decode_in_place\" is defined in the standard library module \"core::codec\".".into(),
                    "Check that you are using a version of the \"core\" standard library that contains this function.".into(),
```

ironcev (changes requested):

sway-error/src/error.rs
xunilrj: Done.

sway-error/src/error.rs
```diff
@@ -2430,6 +2433,18 @@ impl ToDiagnostic for CompileError {
...
+            ConfigurableMissingAbiDecodeInPlace { span } => Diagnostic {
+                reason: Some(Reason::new(code(1), "Configurables need a function named \"abi_decode_in_place\" to be in scope".to_string())),
+                issue: Issue::error(
+                    source_engine,
+                    span.clone(),
+                    String::new()
+                ),
+                hints: vec![],
+                help: vec![
+                    "Function \"abi_decode_in_place\" is defined in the standard library module \"core::codec\".".into()
```
ironcev, line 2445: Hmm, having just this first sentence in the help without the second proposed one that suggests to verify if the function is actually there, this help actually becomes confusing because it states that the function _is defined_ in the "code::codec" and the error is there very likely because it is _not_ defined.

There is another one issue. The error is currently emitted on every configurable because the span points to an individual configurable. It should be just a single error pointing to either the whole `configurable` block or to the text configurable, because it is not specific to a particular configurable.

Together with the above help message it results in actually a confusing experience.

Coming to an optimally crafted diagnostic is never easy :smile: and usually takes some polishing. After looking once again at the overall formulation I propose the following changes.

We leave the old-style single line formulation in line 967 as it is.

Reason should answer why is this an issue. The fact that "Configurables need a function named \"abi_decode_in_place\" to be in scope" is not an issue that's how it should be :smile: 

So, to be consistent with the guidelines, proposal is to reformulate the `Reason` as:  Function \"abi_decode_in_place\" is not in scope

`Issue::error` should actually point at the `configurable` block or even better at the span of the "configurable" and further explain: Configurables need a function named \"abi_decode_in_place\" to be in scope. // With the fullstop this time :smile:

Note that we can still emit the error for every configurable. That's fine because the deduplication will remove duplicates and leave only a single error message. 

And finally in the help we have both sentences, to guide where to look:
"Function \"abi_decode_in_place\" is usually defined in the standard library module \"core::codec\".".into(),
"Verify that you are using a version of the \"core\" standard library that contains this function.".into(),

Note also the added word "usually" because we allow developers to define their own encoding/decoding mechanisms by implementing functions in scope.

ironcev (changes requested):

sway-error/src/error.rs
xunilrj: > There is another one issue. The error is currently emitted on every configurable because the span points to an individual configurable. It should be just a single error pointing to either the whole configurable block or to the text configurable, because it is not specific to a particular configurable.

Not sure I understood this point. What is not being used is the configurable, not the block. Why highlighting the whole block would be better?

image

ironcev: The warning is perfectly fine, of course :smile:  I was talking about the _error_ which is not related to a particular configurable and will be rendered on every configurable. Currently for e.g.:
```Sway
configurable {
    MAX_SUPPLY: u64 = 3,
    OTHER: u64 = 3,
    MORE: u64 = 3,
}
```
rendered errors will be quite overwhelming and confusing:

Configurables need a function named abi_decode_in_place to be in scope

The impression those message give is, that there is something problematic with an individual configurable, but they are all perfectly fine and there is nothing to change about them. It is actually a piece of "infrastructure" that is missing. That's why I think having a single error (duplicates automatically removed via deduplication) pointing to the block will better communicate the issue.

xunilrj: Ah. In this case it makes sense.

xunilrj:
As discussed, the error now points to the configurable block:

image

sway-error/src/error.rs
xunilrj: I will use the keyword span, as the whole block renders very strangely. 
image

ironcev: Yes unfortunately this issue with code snippets is a know one, #5499, but still open. Using keyword is also fine.

ironcev (dismissed):

ironcev (approved):

IGI-111 (approved):

JoshuaBatty (approved):

PR #6132, opened 2024-06-14
Grammar

1griffy:
## Description
Cost of execution isn't a resource

### DIFF
--- a/docs/book/src/index.md
+++ b/docs/book/src/index.md
@@ -30,7 +30,7 @@ Whilst Rust is a great systems programming language (and Sway itself is written 
 
 Rust shines because it can use zero-cost abstractions and its sophisticated borrow-checker memory model to achieve impressive runtime performance for complex programs without a garbage collector.
 
-On a blockchain, cost of execution and deployment is scarce.
+On a blockchain, both deployment of a contract and its execution are expensive.
 Memory usage is low and execution time is short.
 This makes complex memory management in general much too expensive to be worthwhile and Rust's borrow checker a burden with no upside.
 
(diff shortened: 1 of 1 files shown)


docs/book/src/index.md
```diff
@@ -30,7 +30,7 @@ Whilst Rust is a great systems programming language (and Sway itself is written
 
 Rust shines because it can use zero-cost abstractions and its sophisticated borrow-checker memory model to achieve impressive runtime performance for complex programs without a garbage collector.
 
-On a blockchain, cost of execution and deployment is the scarce resource.
+On a blockchain, cost of execution and deployment is scarce.
```
sdankel, line 33: This suggestion doesn't really make sense either. The cost isn't scarce, it's just costly.

```suggestion
On a blockchain, execution and deployment are limited by transaction costs.
```

docs/book/src/index.md
```diff
@@ -30,7 +30,7 @@ Whilst Rust is a great systems programming language (and Sway itself is written
...
 Memory usage is low and execution time is short.
```
sdankel, line 34:
```suggestion
Memory is an expensive resource.
```

IGI-111: You're losing key information and this isn't what I mean here.

docs/book/src/index.md
IGI-111: There may be a better term than cost to analogize with space and time, which are resources.

docs/book/src/index.md
1griffy: Updated

IGI-111:
Yeah, I'm just going to close this, the new formulation isn't really that much clearer and it's doing other random changes now. We don't have the time to spend on insubstantial documentation changes unless they're no-brainers like bulk typo fixes.

PR #6136, opened 2024-06-17
Resolve memory overflow when `b256::TryFrom<Bytes>` is not 32 bytes

### LINKED ISSUE #6129: `b256::try_from` should check against length is not 32 rather than greater than
The `TryFrom` implementation for `Bytes` on `b256`currently checks whether the `Bytes` is greater than 32 bytes. It should either check not equal to or pad additional zeros.

bitzoic:
## Description

Currently, when calling the `TryFrom` implementation for `b256` from `Bytes`, if the `Bytes` has less than 32 bytes a memory overflow error will occur. If there is less than 32 bytes, `None` is now returned.

Example:
```sway
    let mut bytes = Bytes::with_capacity(32);
    let mut i = 0;
    while i < 32 {
        // 0x33 is 51 in decimal
        bytes.push(51u8);
        i += 1;
    }
    let res = b256::try_from(bytes);
    assert(res.unrwap() == 0x3333333333333333333333333333333333333333333333333333333333333333);
```

Closes #6129 

## Checklist


test/src/in_language_tests/test_programs/primitive_conversions_b256_inline_tests/src/main.sw
```diff
@@ -11,10 +11,10 @@ fn b256_try_from_bytes() {
         initial_bytes.push(51u8);
         i += 1;
     }
-    let res = b256::try_from(initial_bytes);
-    let expected = 0x3333333333333333333333333333333333333333333333333333333333333333;
-
-    assert(res.unwrap() == expected);
+    let res1 = b256::try_from(initial_bytes);
+    let expected1 = 0x3333333333333333333333333333333333333333333333333333333333333333;
+    assert(res1.is_some());
```
SwayStar123, line 16: doesnt the unwrap imply that res1 is Some?

SwayStar123 (approved):

IGI-111 (changes requested):
I don't think left padding is the correct behavior. Converting less than 32 bytes to a `b256` should fail.

You may argue this for `u256` though I still think it's dubious to assume user intent there. But for `b256` it's categorically erroneous.

test/src/in_language_tests/test_programs/primitive_conversions_b256_inline_tests/src/main.sw
bitzoic: It is better to fail intentionally than fail on unwrapping `None`. Fail fast and fail early.

IGI-111:
LGTM but can you update the title and PR message to reflect the changes?

K1-R1 (approved):

IGI-111 (approved):

PR #6142, opened 2024-06-19
Fix `Bytes`, `Vec` and `String` buffer ownership

xunilrj:
## Description

This PR fixes a problem with `std::Bytes`, `std::Vec` and `std::String` buffer ownership. It also fixes a problem with buffer overflow when encoding huge types.

## Buffer Ownership

Currently, types like `Bytes`, `Vec` and `String` do not guarantee ownership of their buffer. That means that we can very easily alias the underlying buffer and change its mutability by simply creating a new instance.

For example:

```sway
let some_slice = ...;
let buffer = Bytes::from(some_slice);
let mut buffer2 = buffer.clone();
// Now I can change `some_slice`
```

## Type Inference bug

I also had to fix a small problem with the type inference of method applications. The problem is that sometimes the type check does not return an error on the first pass, but the type is still not concrete. 

For example `x.get(0) == Some(2)`, becomes `eq(x.get(0), Some(2))`.

After the first pass, `x.get(0)` is correctly inferred to be `Option<u8>`; but `Some(2)` is only typed as `Option<Numeric>`. This happens because the first pass starts with `TypeInfo::Unknown`. We can use the argument types here because they may still be non-concrete types like "Self".

What the fix is doing is that it checks if the inferred type `is_concrete`, assuming that `Numeric` is not. This will make "Some(2)" be type-checked again at the second pass, after monomorphization and be correctly typed as "Option<u8>".

## IR Verification errors

This PR also improves the error message for IR verification. Now the error is shown inside the printed IR.

image

## Script to update contract-ids (optional)

At `test/update-contract-ids.sh` there is a small script that automatically updates contract ids. Unfortunately, given the indirect nature of contract calls, it is impossible for the script to know which contract it needs to compile. So we need to specify the path to the contract. 

We also need to pass the compiler flags, because in some cases we use `debug` profile, and in others we use `release`.

```sway
const FUEL_COIN_CONTRACT_ID = 0x1a88d0982d216958d18378b6784614b75868a542dc05f8cc85cf3da44268c76c; // AUTO-CONTRACT-ID ../../test_contracts/test_fuel_coin_contract --release
```

In the example above, the path and flags are appended in the bash script. I failed trying to inject commands, so I think the append is safe.

I can remove this from the PR, if we find that this is not the solution we want at the moment.

## Checklist


IGI-111 (approved):

JoshuaBatty (dismissed):

bitzoic (approved):

IGI-111 (dismissed):

IGI-111:
LGTM though it would be good to document the ownership convention somewhere.

JoshuaBatty (approved):

IGI-111 (dismissed):

JoshuaBatty (dismissed):

JoshuaBatty (approved):

sway-core/src/ir_generation/function.rs
```diff
@@ -1717,6 +1717,224 @@ impl<'eng> FnCompiler<'eng> {
...
+
+                    s.current_block
+                        .append(context)
+                        .branch(merge_block, vec![new_ptr, new_cap]);
+
+                    // dont need realloc block
+                    s.current_block = false_block_begin;
+                    s.current_block
+                        .append(context)
+                        .branch(merge_block, vec![ptr, cap]);
```
jjcnn, line 1837: Is there any reason to have a `false_block`? AFAICT you can jump straight to the merge branch if the buffer doesn't need to grow.

sway-core/src/type_system/id.rs
```diff
@@ -456,17 +461,26 @@ impl TypeId {
...
+                    | TypeInfo::TypeParam(..)
+                    | TypeInfo::Numeric
+            ),
+            IncludeNumeric::No => matches!(
                 x,
                 TypeInfo::UnknownGeneric { .. }
                     | TypeInfo::Custom { .. }
                     | TypeInfo::Placeholder(..)
                     | TypeInfo::TraitType { .. }
                     | TypeInfo::TypeParam(..)
```
jjcnn, line 482:
```suggestion
        !nested_types.into_iter().any(|x| match x {
                TypeInfo::UnknownGeneric { .. }
                    | TypeInfo::Custom { .. }
                    | TypeInfo::Placeholder(..)
                    | TypeInfo::TraitType { .. }
                    | TypeInfo::TypeParam(..) => true,
                    TypeInfo::Numeric => matches!(include_numeric, IncludeNumeric::Yes)
            ),
```

sway-core/src/ir_generation/function.rs
xunilrj: No. I will create another PR later to fix this.

JoshuaBatty (approved):

IGI-111 (approved):

PR #6166, opened 2024-06-21
Refactor `StorageVec`'s `load_vec()` and `store_vec()` to handle types of less than 8 bytes

### LINKED ISSUE #6127: `StorageVec`'s `load_vec()` and `store_vec` do not properly handle element types with size < 8 bytes.
When calling `load_vec()` and `store_vec()` with elements of size less than 8 bytes, these are not properly handled.

bitzoic:
## Description

Sizes of less than 8 bytes such as `u8` are no longer padded to the size of 1 word. This means that the current storage API and heap are no longer 1:1 for `StorageVec` and `Vec`. As such, any conversions between these two were unhandled. 

In this PR, elements in a `Vec` of less than 8 bytes are padded to the size of a word before being stored in a `StroageVec` and vice versa. 

Note: This increases gas costs by iterating over every element for some cases. Until the storage API packs elements of less than 8 bytes into storage, this changes from an O(1) to an O(n) function.

Closes #6127 

## Checklist


sway-lib-std/src/storage/storage_vec.sw
```diff
@@ -809,7 +809,32 @@ impl<V> StorageKey<StorageVec<V>> {
...
+                        .copy_bytes_to(ptr.add_uint_offset(8 * i), size_V_bytes);
+                    i += 1;
+                }
+
+                raw_slice::from_parts::<V>(ptr, number_of_words)
+            },
+            false => {
+                vec.as_raw_slice()
+            }
+        };
```
IGI-111, line 836: no sense using match in place of if

sway-lib-std/src/storage/storage_vec.sw
```diff
@@ -865,14 +890,49 @@ impl<V> StorageKey<StorageVec<V>> {
...
+                            },
+                        )
+                    },
+                    false => {
+                        Vec::from(
+                            asm(ptr: (ptr, bytes)) {
+                                ptr: raw_slice
+                            },
+                        )
+                    }
```
IGI-111, line 934: Here too.

sway-lib-std/src/storage/storage_vec.sw
bitzoic: `match` is used to return two different `raw_slice`s in this case

sway-lib-std/src/storage/storage_vec.sw
bitzoic: In this case an `if` statement is more appropriate 👍

sway-lib-std/src/storage/storage_vec.sw
IGI-111: So?

Two branches of an if-else can return different values of the same type. These are equivalent:

```sway
match a {
    true => 1,
    false => 2,
}

if a { 1 } else { 2 }
```

sway-lib-std/src/storage/storage_vec.sw
```diff
@@ -865,14 +887,46 @@ impl<V> StorageKey<StorageVec<V>> {
             0 => Vec::new(),
             len => {
                 // Get the number of storage slots needed based on the size.
-                let bytes = len * __size_of::<V>();
+                let size_V_bytes = __size_of::<V>();
+                let bytes = match size_V_bytes < 8 {
```
IGI-111, line 891: Yet one more here.

IGI-111: couldn't you just do this:

```sway
let bytes = len * 8.max(size_V_bytes);
```

sway-lib-std/src/storage/storage_vec.sw
```diff
@@ -865,14 +887,44 @@ impl<V> StorageKey<StorageVec<V>> {
...
             len => {
                 // Get the number of storage slots needed based on the size.
-                let bytes = len * __size_of::<V>();
+                let size_V_bytes = __size_of::<V>();
+                let bytes = if size_V_bytes < 8 {
+                    // Len * size_of_word
+                    len * 8
+                } else {
+                    len * size_V_bytes
+                };
```
IGI-111, line 896: I don't think it's a requirement but It may be a good opportunity to add max and min to core similar to Rust's: https://doc.rust-lang.org/std/cmp/index.html

IGI-111 (dismissed):

sway-lib-std/src/storage/storage_vec.sw
bitzoic: https://github.com/FuelLabs/sway/issues/6169

sway-lib-std/src/storage/storage_vec.sw
```diff
@@ -809,7 +809,29 @@ impl<V> StorageKey<StorageVec<V>> {
     /// ```
     #[storage(write)]
     pub fn store_vec(self, vec: Vec<V>) {
-        let slice = vec.as_raw_slice();
+        let size_V_bytes = __size_of::<V>();
+
+        // Handle cases where elements are less than the size of word as pad to the size of a word
```
K1-R1, line 814:
```suggestion
        // Handle cases where elements are less than the size of word and pad to the size of a word
```

test/src/sdk-harness/test_projects/storage_vec_to_vec/src/main.sw
```diff
@@ -366,3 +395,73 @@ fn test_pop_struct() {
...
+    assert(returned_vec.get(0).unwrap() == 5u8);
+    assert(returned_vec.get(1).unwrap() == 7u8);
+    assert(returned_vec.get(2).unwrap() == 9u8);
+    assert(returned_vec.get(3).unwrap() == 11u8);
+}
+
+#[test]
+fn test_push_u8() {
+    let vec_abi = abi(VecToVecStorageTest, CONTRACT_ID);
+    // let mut test_vec = Vec::<u8>::new();
```
K1-R1, line 422: Can the commented out lines be removed or a TODO added?

K1-R1 (approved):

SwayStar123 (approved):

IGI-111 (approved):

PR #6195, opened 2024-06-27
Fix fn dedup miscompilation leading to `MemoryOverflow` on `Vec`s

xunilrj:
## Description

This PR fixes a miscompilation bug caused by our dedup cache just before IR generation, we started to notice this bug with some `MemoryOverflow` reverts on `Vec` tests using 0.61.  

These reverts happen because `Vec::with_capacity` allocates 16 bytes per item, but later one `Vec::push` pushes 24 bytes per item; causing some writes outside the allocated buffer.

This is happening because, before we start IR generation, we try to coalesce identical functions using the following hash key: https://github.com/FuelLabs/sway/blob/master/sway-core/src/ir_generation.rs#L75

The example below shows that two different bodies end up colliding at the same key.

```
Implementing for Some(Vec<[u64; 2]>)
Not parts of the key
    Key: 16648822319765008093
    Body hash: 14066144736883705909
Parts of the key
    Span: Span { src (ptr): 0x5f096722d090, source_id: Some(SourceId { id: 2097176 }), start: 4886, end: 5032, as_str(): "pub fn with_capacity(capacity: u64) -> Self {\n        Self {\n            buf: RawVec::with_capacity(capacity),\n            len: 0,\n        }\n    }" }
    Parameters:
        u64
    Type parameters:

Implementing for Some(Vec<SomeEnum<u32>>)
Not parts of the key
    Key: 16648822319765008093
    Body hash: 17149806863110053630
Parts of the key
    Span: Span { src (ptr): 0x5f096722d090, source_id: Some(SourceId { id: 2097176 }), start: 4886, end: 5032, as_str(): "pub fn with_capacity(capacity: u64) -> Self {\n        Self {\n            buf: RawVec::with_capacity(capacity),\n            len: 0,\n        }\n    }" }
    Parameters:
        u64
    Type parameters:
```

The difference here is that in one case `RawVec` is inferred to be `RawVec<Vec<[u64; 2]>>` and in the other `RawVec<Vec<SomeEnum<u32>>`. The crux of the issue is that this difference does not appear in the key.

The question is why this broke `with_capacity` and not `push`. And the answer in that `push` has the self and the generic parameters to differentiate them, whilst `with_capacity` does not.

```
Implementing for Some(Vec<SomeEnum<u32>>)
Not parts of the key
    Key: 8342657797980723970
    Body hash: 328023847492220083
Parts of the key
    Span: Span { src (ptr): 0x5f096722d090, source_id: Some(SourceId { id: 2097176 }), start: 5490, end: 5964, as_str(): "pub fn push(ref mut self, value: T) {\n        // If there is insufficient capacity, grow the buffer.\n        if self.len == self.buf.capacity() {\n            self.buf.grow();\n        };\n\n        // Get a pointer to the end of the buffer, where the new element will\n        // be inserted.\n        let end = self.buf.ptr().add::<T>(self.len);\n\n        // Write `value` at pointer `end`\n        end.write::<T>(value);\n\n        // Increment length.\n        self.len += 1;\n    }" }
    Parameters:
        Vec<SomeEnum<u32>>
        SomeEnum<u32>
    Type parameters:

Implementing for Some(Vec<SomeStruct<u32>>)
Not parts of the key
    Key: 15738668646378064265
    Body hash: 15996600191005707730
Parts of the key
    Span: Span { src (ptr): 0x5f096722d090, source_id: Some(SourceId { id: 2097176 }), start: 5490, end: 5964, as_str(): "pub fn push(ref mut self, value: T) {\n        // If there is insufficient capacity, grow the buffer.\n        if self.len == self.buf.capacity() {\n            self.buf.grow();\n        };\n\n        // Get a pointer to the end of the buffer, where the new element will\n        // be inserted.\n        let end = self.buf.ptr().add::<T>(self.len);\n\n        // Write `value` at pointer `end`\n        end.write::<T>(value);\n\n        // Increment length.\n        self.len += 1;\n    }" }
    Parameters:
        Vec<SomeStruct<u32>>
        SomeStruct<u32>
    Type parameters:
```

Funnily enough, we also have the inverse problem. Sometimes functions with the same body do not fall at the same key, because their `TypeId` are different, and `HashWithEngines` actually look at `TypeInfo` variants. This is probably increasing code size and/or throwing more stuff at the optimizer.

```
Implementing for Some(Vec<SomeEnum<u32>>)
Not parts of the key
    Key: 12091758953823056676
    Body hash: 328023847492220083
Parts of the key
    Span: Span { src (ptr): 0x5f096722d090, source_id: Some(SourceId { id: 2097176 }), start: 5490, end: 5964, as_str(): "pub fn push(ref mut self, value: T) {\n        // If there is insufficient capacity, grow the buffer.\n        if self.len == self.buf.capacity() {\n            self.buf.grow();\n        };\n\n        // Get a pointer to the end of the buffer, where the new element will\n        // be inserted.\n        let end = self.buf.ptr().add::<T>(self.len);\n\n        // Write `value` at pointer `end`\n        end.write::<T>(value);\n\n        // Increment length.\n        self.len += 1;\n    }" }
    Parameters:
        Vec<SomeEnum<u32>>
        SomeEnum<u32>
    Type parameters:


Implementing for Some(Vec<SomeEnum<u32>>)
Not parts of the key
    Key: 8342657797980723970
    Body hash: 328023847492220083
Parts of the key
    Span: Span { src (ptr): 0x5f096722d090, source_id: Some(SourceId { id: 2097176 }), start: 5490, end: 5964, as_str(): "pub fn push(ref mut self, value: T) {\n        // If there is insufficient capacity, grow the buffer.\n        if self.len == self.buf.capacity() {\n            self.buf.grow();\n        };\n\n        // Get a pointer to the end of the buffer, where the new element will\n        // be inserted.\n        let end = self.buf.ptr().add::<T>(self.len);\n\n        // Write `value` at pointer `end`\n        end.write::<T>(value);\n\n        // Increment length.\n        self.len += 1;\n    }" }
    Parameters:
        Vec<SomeEnum<u32>>
        SomeEnum<u32>
    Type parameters:
```

My solution for this problem is to use the body hash as the cache key. We can try to be less risky and include the body hash and keep the other items.

Below there are two graphs showing that compilation time and binary size are not affected too much by this change.

Compilation Time

image

Binary Size

image

## Checklist


IGI-111 (approved):
Nice catch!

PR #6197, opened 2024-06-28
Add a simple simplify-cfg optimization for assembly

### LINKED ISSUE #6174: `while true {}` loop and `if true` branching result in Internal Compiler Error
Compiling the following script on the newest `master`:

```Sway
script;

#[inline(never)]
fn main() {
    while true {
    }
}
```

will result in:
```
error: Internal compiler error: Program erroneously uses uninitialized virtual registers: $r32
```

Note that the `main` must not be inlined into `__entry`. If inlined, the error will not happen.

Similarly, the `should_pass/language/break_and_continue_block_ret` test will start failing with the same ICE if we mark its `main` function as `#[inline(never)]`.

Similarly, the `should_pass/language/abort_control_flow_good` test will start failing in the debug mode where `if` is not optimized away with the same ICE if we mark its `main` function as `#[inline(never)]`.

vaivaswatha:
## Description
When compiling code such as

```
fn main() {
     while true {}
}
```
we would end up with unreachable code in the assembly:
```
.program:
.2                                      ; --- start of function: main_0 ---
pusha .2                                ; save all regs
move $$locbase $sp                      ; save locals base register for main_0
cfei i0                                 ; allocate 0 bytes for locals and 0 slots for call arguments.
move $r28 $$reta                        ; save reta
move $r29 $$retv                        ; save retv
.6
ji  .7
.7
ji  .7
.3
cfsi i0                                 ; free 0 bytes for locals and 0 slots for extra call arguments.
move $$reta $r28                        ; restore reta
popa .2                                 ; restore all regs
jmp $$reta                              ; return from call

;; --- Data ---
.data:
```

Here the label `.3` and what follows is unreachable because the program gets stuck in an infinite loops with label `.7`. This means that the instruction `move $r28 $$reta` is dead and hence removed by ASM DCE.

Our [verifier](https://github.com/FuelLabs/sway/blob/4a6cf38143b398a568d9587f79bfa46c15f48459/sway-core/src/asm_generation/fuel/abstract_instruction_set.rs#L131) on the other hand will see a use of `$r28` in the unreachable code. But with the definition having been eliminated, it will now flag an error.

The verifier is unaware of reachability. Making the verifier more accurate is probably an overkill. So we workaround this by adding an unreachable code eliminator optimization in the assembly stage. While it has its own benefits (of removing unreachable code), it'll also make the verifier happy as-is. The algorithm is linear in the size of the code.

Closes #6174.


ironcev (changes requested):
Awesome! Super simple solution :smile:

Can you please also just enable the two disabled tests:
should_pass/language/break_and_continue_block_ret
should_pass/language/abort_control_flow_good

ironcev (approved):

PR #6235, opened 2024-07-05
chore: remove snapshot, as accounts are funded by default

kayagokalp:
## Description

Removes local snapshot we were using to fund our accounts as it is now done by default as of https://github.com/FuelLabs/fuel-core/pull/1894. By default `--default-signer` is signing it with the account funded by fuel-core out of the box.


JoshuaBatty (approved):
Nice

IGI-111 (approved):

PR #6237, opened 2024-07-06
LDC multi-chunk contract loader

vaivaswatha:
With the target contract broken down into chunks (with sizes that are word aligned), we can load it in pieces and then execute.

On my local test, I split `run_external_target` contract binary into `part-1.bin` with size 2080 bytes and `part-2.bin` with size 2072 bytes. (Use command `split -b 2080 ...` on Linux for convenience).

With the below commands, the test `run_external_can_proxy_call` passed.
```
$cd test/src/sdk-harness
$forc build --release
$cargo test -- run_external_can_proxy_call --ignored --show-output
```


kayagokalp:
I did some hacking to see if i can generate asm for automatically generating loader: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5ff55f234b0444e94599c5e505020ce3 With this simple asm generator I was able to replicate this test with 3 chunks.

kayagokalp:
I have a question though, is it fine to cut the contract wihtout any consideration. Meaning that is there any conditions that should be met for each chunk. Or any special case for the first or last chunk? Or can i replicate the split functionality?

vaivaswatha:
> I have a question though, is it fine to cut the contract wihtout any consideration. Meaning that is there any conditions that should be met for each chunk. Or any special case for the first or last chunk? Or can i replicate the split functionality?

The VM appends each LDC'd chunk to a word boundary. So if your chunks aren't word aligned (i.e., size not divisible by 8), then it gets padded. The padding makes it bad for our purposes since the next chunk should be immediately after the currently LDC'd one.

So word aligning each chunk (except possibly for the last one) is necessary. I don't think there are any other constraints.

kayagokalp:
I am closing this PR as this was mainly to showcase I believe and this feature is already in. If i am mistaken please re-open :)

PR #6250, opened 2024-07-09
feat: Implement ldc based automatic contract chunking 

kayagokalp:
## Description

This PR adds chunk deployment based on LDC.  Contracts larger than 100 kb size, is split into chunks and chunks are deployed as `blobs`. Out of these blobs we create a loader contract, which loads all the blobs using `LDC` opcode.

One important thing is that this feature works nicely with the proxy feature introduced in #6069, so a large contract, with proxy can be deployed directly. Large contract will be split into chunks, chunks will get deployed, loader will get get generated and deployed, after all these a proxy contract is deployed and pointed to the loader contract deployed.

Simple chunked deploy, chunked deployment re routing the call, chunked deployment behind a proxy re routes the call is tested.


kayagokalp:
based on #6069

forc-plugins/forc-client/src/cmd/deploy.rs
```diff
@@ -81,4 +81,8 @@ pub struct Command {
     /// Disable the "new encoding" feature
     #[clap(long)]
     pub no_encoding_v1: bool,
+
+    /// Maximum contrat size, after which forc-deploy automatically splits the contract into chunks.
+    #[clap(long)]
+    pub maximum_contract_size: Option<usize>,
```
JoshuaBatty, line 87: Any reason why we would want to make this configurable rather than just using the `MAX_CONTRACT_SIZE` const?

JoshuaBatty: If there isn't a valid reason I think we should remove it

kayagokalp: Sure, I made it like this as for different networks targeted the size limit for the contract might be different. But I don't think this would be a hugely useful as I am imagining we will have a set contract size limit that won't change in near future. Currently I am using it for testing but i can work around it no problems. Maybe we should look for a way to query the size limit from node. That way the value can be adjusted according to the target :)

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -38,10 +41,17 @@ use std::{
 use sway_core::language::parsed::TreeType;
 use sway_core::BuildTarget;
 
+/// Maximum contract size allowed to be in a single contract. If the target
+/// contract size is bigger than this amount, forc-deploy will automatically
+/// starts dividing the contract and deploy them in chunks automatically.
+/// The value is in bytes.
+const MAX_CONTRACT_SIZE: usize = 1_000;
```
kayagokalp, line 48: We need to make this `100_000`.

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -85,3 +102,366 @@ pub(crate) fn built_pkgs(path: &Path, build_opts: &BuildOpts) -> Result<Vec<Arc<
...
+    let types = create_type_map(abi);
+
+    // Contract header
+    contract.push_str("contract;\n\n");
+
+    // Configurables
+    contract.push_str("configurable {\n");
+    for i in 1..=num_chunks {
+        contract.push_str(&format!(
+            "    TARGET_{}: ContractId = ContractId::from({}),\n",
```
sdankel, line 121: We should have a limit on the number of chunks allowed. Otherwise, it's possible that the loader contract could exceed the allowable size of a contract.

forc-plugins/forc-client/test/data/big_contract/src/main.sw
```diff
@@ -0,0 +1,14 @@
+contract;
+
+abi MyContract {
+    fn test_function() -> bool;
+}
+
+impl MyContract for Contract {
+    fn test_function() -> bool {
```
sdankel, line 8: It would be good to add more methods to make this a more complex ABI. It's worth having a test that calls different methods in the same chunked contract.

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -85,3 +102,366 @@ pub(crate) fn built_pkgs(path: &Path, build_opts: &BuildOpts) -> Result<Vec<Arc<
...
+    func.push_str("        cfs cur_stack_size;\n");
+
+    // Do the loads
+    func.push_str("        ldc load_target1 zero length1;\n");
+    for i in 2..=num_targets {
+        func.push_str(&format!(
+            "        ldc load_target{}_heap zero length{};\n",
+            i, i
+        ));
+    }
```
sdankel, line 261: nit:

```suggestion
    for i in 1..=num_targets {
        func.push_str(&format!(
            "        ldc load_target{}_heap zero length{};\n",
            i, i
        ));
    }
```

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -85,3 +102,366 @@ pub(crate) fn built_pkgs(path: &Path, build_opts: &BuildOpts) -> Result<Vec<Arc<
...
+/// - The project folder with a correct structure if it is not already there.
+/// - Loader contract source code, that is created specifically for number of
+///   chunks in hand.
+pub(crate) fn create_chunk_loader_contract(
+    abi: &ProgramABI,
+    chunk_contract_ids: &[String],
+    num_chunks: usize,
+    pkg_name: &str,
+) -> Result<PathBuf> {
+    // Create the proxy contract folder.
```
sdankel, line 286: Should we throw an error if `num_chunks` is less than 2? This would also be a good place to check for the upper limit on chunks.

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -85,3 +102,366 @@ pub(crate) fn built_pkgs(path: &Path, build_opts: &BuildOpts) -> Result<Vec<Arc<
...
+        );
+    }
+    let mut chunks = Vec::new();
+
+    for (id, chunk) in bytecode.chunks(chunk_size).enumerate() {
+        let chunk = chunk.to_vec();
+        let size = chunk.len();
+        let contract_chunk = ContractChunk::new(id, size, chunk);
+        chunks.push(contract_chunk);
+    }
```
sdankel, line 389: nit:

```suggestion
    let chunks: Vec<ContractChunk> = bytecode
        .chunks(chunk_size)
        .enumerate()
        .map(|(id, chunk)| ContractChunk::new(id, chunk.len(), chunk.to_vec()))
        .collect();
```

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -85,3 +102,366 @@ pub(crate) fn built_pkgs(path: &Path, build_opts: &BuildOpts) -> Result<Vec<Arc<
 
     Ok(built_pkgs)
 }
+
+pub fn generate_chunk_loader_contract_src(
```
sdankel, line 106: I'd prefer to move all of these functions relating to contract chunking into their own module, `util/contract_chunk.rs`

forc-plugins/forc-client/tests/deploy.rs
```diff
@@ -405,3 +410,157 @@ fn test_deploy_interactive_wrong_password() -> Result<(), rexpect::error::Error>
...
+    let (mut node, port) = run_node();
+    let tmp_dir = tempdir().unwrap();
+    let project_dir = test_data_path().join("big_contract");
+    copy_dir(&project_dir, tmp_dir.path()).unwrap();
+    patch_manifest_file_with_path_std(tmp_dir.path()).unwrap();
+
+    let pkg = Pkg {
+        path: Some(tmp_dir.path().display().to_string()),
+        ..Default::default()
+    };
```
sdankel, line 458: This piece of code is repeated in almost every test, and should be refactored into a helper function like:

```rust
fn create_tmp_pkg(project_name: &str) -> Pkg {
    let tmp_dir = tempdir().expect("temp dir");
    let project_dir = test_data_path().join(project_name);
    copy_dir(&project_dir, tmp_dir.path()).expect("copy dir");
    patch_manifest_file_with_path_std(tmp_dir.path()).expect("patch manifest");

    Pkg {
        path: Some(tmp_dir.path().display().to_string()),
        ..Default::default()
    }
}
```

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -122,6 +132,70 @@ fn validate_and_parse_salts<'a>(
...
+    let deployed_id = deploy_pkg(command, &loader_contract, salt, provider, signing_key).await?;
+
+    println_action_green(
+        "Finished",
+        &format!("deploying loader contract for {pkg_name} {contract_url}{deployed_id}"),
+    );
+
+    Ok((deployed_id, deployed_contracts))
+}
+
```
sdankel, line 198: We need to update the logic that displays the number of transactions to include a transaction for each chunk.

image

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -85,3 +102,366 @@ pub(crate) fn built_pkgs(path: &Path, build_opts: &BuildOpts) -> Result<Vec<Arc<
...
+    // Do the loads
+    func.push_str("        ldc load_target1 zero length1;\n");
+    for i in 2..=num_targets {
+        func.push_str(&format!(
+            "        ldc load_target{}_heap zero length{};\n",
+            i, i
+        ));
+    }
+
+    // Set up jump
```
sdankel, line 263: How does the loader contract jump to the correct method in the chunks?

sdankel (commented):
Awesome work 👍

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -122,6 +132,70 @@ fn validate_and_parse_salts<'a>(
...
+        let chunk_id = contract_chunk.id();
+        let deployed_chunk = contract_chunk.deploy(provider, signing_key).await?;
+        let chunk_contract_id = deployed_chunk.contract_id();
+        println_action_green(
+            "Finished",
+            &format!("deploying chunk {chunk_id} for {pkg_name}: {chunk_contract_id}"),
+        );
+        deployed_contracts.push(deployed_chunk);
+    }
+    println_action_green("Deployed", &format!("{pkg_name} chunks"));
```
sdankel, line 163: nit: it would be nice to time the whole operation and show "Deploy X chunks in X ms"

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -122,6 +132,70 @@ fn validate_and_parse_salts<'a>(
...
 
+async fn deploy_chunked(
+    command: &cmd::Deploy,
+    compiled: &BuiltPackage,
+    salt: Salt,
+    signing_key: &SecretKey,
+    provider: &Provider,
+    pkg_name: &str,
+) -> anyhow::Result<(ContractId, Vec<ContractId>)> {
+    println_action_green("Splitting", &format!("{pkg_name} into chunks"));
```
sdankel, line 143: Rather than showing this message, we should update the `Confirming transactions` message to communicate this. It could look something like this:

`Confirming transactions [deploy big_contract + 5 contract chunks]`

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -122,6 +132,70 @@ fn validate_and_parse_salts<'a>(
...
+    let contract_chunks = split_into_chunks(compiled.bytecode.bytes.clone(), MAX_CONTRACT_SIZE)?;
+    let mut deployed_contracts = vec![];
+    println_action_green("Deploying", &format!("{pkg_name} chunks"));
```
sdankel, line 146: nit: consistency

```suggestion
    println_action_green("Deploying", &format!("contract {pkg_name} chunks"));
```

forc-plugins/forc-client/tests/deploy.rs
```diff
@@ -405,3 +410,157 @@ fn test_deploy_interactive_wrong_password() -> Result<(), rexpect::error::Error>
...
+        .iter()
+        .cloned()
+        .map(|chunk_id| chunk_id.into())
+        .collect();
+
+    // result should be true.
+    let result = instance
+        .methods()
+        .test_function()
+        .with_contract_ids(&chunks)
```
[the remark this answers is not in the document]
sdankel: We should be using the new [blob mode](https://github.com/FuelLabs/fuel-vm/pull/780/files) for `ldc` so that the contract IDs don't have to be provided as inputs to every transaction.

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -85,3 +102,366 @@ pub(crate) fn built_pkgs(path: &Path, build_opts: &BuildOpts) -> Result<Vec<Arc<
...
+    for i in 1..=num_targets {
+        if i > 1 {
+            func.push_str(", ");
+        }
+        func.push_str(&format!("load_target{}: ContractId", i));
+    }
+    func.push_str(") -> ! {\n");
+
+    // Generate assembly
+    func.push_str("    asm(\n");
```
sdankel, line 218: Could you please add a lot more detailed comments about what this assembly code is doing? It might also be nice to add the comment in the generated sway code.

forc-plugins/forc-client/src/util/pkg.rs
```diff
@@ -85,3 +102,366 @@ pub(crate) fn built_pkgs(path: &Path, build_opts: &BuildOpts) -> Result<Vec<Arc<
...
+    chunk_contract_ids: &[String],
+) -> String {
+    let mut contract = String::new();
+    let types = create_type_map(abi);
+
+    // Contract header
+    contract.push_str("contract;\n\n");
+
+    // Configurables
+    contract.push_str("configurable {\n");
```
sdankel, line 118: Do we need to use configurable here? These Ids should never change in the loader contract, right? I think we should just use regular constants.

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -122,13 +132,71 @@ fn validate_and_parse_salts<'a>(
     Ok(contract_salt_map)
 }
 
+fn resolve_storage_slots(
```
JoshuaBatty, line 135: Can we add a doc comment for this fn?

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -122,13 +132,71 @@ fn validate_and_parse_salts<'a>(
...
+            let storage_slots: Vec<StorageSlot> = serde_json::from_str(&storage_slots_file)?;
+            storage_slots
+        } else {
+            compiled.storage_slots.clone()
+        };
+    storage_slots.sort();
+    Ok(storage_slots)
+}
+
+async fn deploy_chunked(
```
JoshuaBatty, line 151: also a doc comment here also?

forc-plugins/forc-client/test/data/big_contract/src/main.sw
kayagokalp: Resolving this as we no longer use abi to generate the loader, we rely on sdk to create a loader with asm directly.

forc-plugins/forc-client/tests/deploy.rs
kayagokalp: Added this to our clean-up list #6259 to get this in asap, I will clean bunch of stuff, after we get through this release

forc-plugins/forc-client/src/op/deploy.rs
kayagokalp: We don't have the `Deployed` prompt right now. We are following the same pattern as normal deplyoment in terms of the output.

Also deployment in chunks is much faster now since we removed unnecessary steps, such as cloning the chunks before each deployment to create contracts out of them as we have blobs and sdk support for blobs now. 

If you feel like this is still valuable, we can add it to the finished prompt though

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -122,13 +132,77 @@ fn validate_and_parse_salts<'a>(
...
+            compiled.storage_slots.clone()
+        };
+    storage_slots.sort();
+    Ok(storage_slots)
+}
+
+/// Creates blobs from the contract to deploy contracts that are larger than
+/// `MAX_CONTRACT_SIZE`. Created blobs are deployed, and a loader contract is
+/// generated such that it loads all the deployed blobs, and provides the user
+/// a single contract (loader contract that loads the blobs) make call into.
```
JoshuaBatty, line 156:
```suggestion
/// a single contract (loader contract that loads the blobs) to call into.
```

JoshuaBatty (commented):
Nice work Kaya looks great. Just wondering if we can add a small section to the forc client documentation explaining that this process is going on automatically behind the scenes for large contracts.

forc-plugins/forc-client/test/data/big_contract/src/main.sw
sdankel: I think this is important. I'm not sure I understand why we wouldn't want this test.

alfiedotwtf (dismissed):

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -38,10 +41,17 @@ use std::{
 use sway_core::language::parsed::TreeType;
 use sway_core::BuildTarget;
 
+/// Maximum contract size allowed for a single contract. If the target
+/// contract size is bigger than this amount, forc-deploy will automatically
+/// starts dividing the contract and deploy them in chunks automatically.
+/// The value is in bytes.
+const MAX_CONTRACT_SIZE: usize = 100_000;
```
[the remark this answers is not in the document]
IGI-111: I don't think that's reasonable, this is meant to allow contracts that are already in practice larger than a megabyte.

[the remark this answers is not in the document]
kayagokalp: I think that is a nice idea, but for the specific use case of this feature we should be using the max contract size of ~100kb for now. Given that the average sway compiler output is almost bigger than 1460 bytes, we would end up with lots of blobs while the client is capable of handling bigger contracts.

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -122,13 +132,77 @@ fn validate_and_parse_salts<'a>(
...
+) -> Result<Vec<fuel_tx::StorageSlot>> {
+    let mut storage_slots =
+        if let Some(storage_slot_override_file) = &command.override_storage_slots {
+            let storage_slots_file = std::fs::read_to_string(storage_slot_override_file)?;
+            let storage_slots: Vec<StorageSlot> = serde_json::from_str(&storage_slots_file)?;
+            storage_slots
+        } else {
+            compiled.storage_slots.clone()
+        };
+    storage_slots.sort();
```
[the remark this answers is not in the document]
kayagokalp: That is a very valid question, I remember that we had to supply a sorted storage slot list to the client, for some time and it looks like we never removed it. I need to do some digging to see whether it is still the case.

For this PR I don't think we should touch it, and remove it in a follow up once we confirm that the sorting is no longer required

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -122,13 +132,77 @@ fn validate_and_parse_salts<'a>(
...
+    salt: Salt,
+    signing_key: &SecretKey,
+    provider: &Provider,
+    pkg_name: &str,
+) -> anyhow::Result<ContractId> {
+    println_action_green("Deploying", &format!("contract {pkg_name} chunks"));
+
+    let storage_slots = resolve_storage_slots(command, compiled)?;
+    let chain_info = provider.chain_info().await?;
+    let target = Target::from_str(&chain_info.name).unwrap_or(Target::testnet());
```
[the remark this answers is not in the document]
sdankel: There should always be a `chain_info`, so this code should be unreachable. Testnet is a reasonable fallback, but local would also be reasonable. We can revisit this.

sdankel (approved):

Voxelot (approved):

calldelegation (approved):

PR #6271, opened 2024-07-15
Make module name mandatory and private

jjcnn:
## Description

Partial fix of #5498 .

So far the `name` field in `namespace::Module` has been public and of type `Option<Ident>`. This makes little sense, since the name of a module should always have a value, and should retain that value once it is set. This PR fixes this problem.

The `visibility` and `span` fields have also been made private. `visibility` can be read-only, but there is a single instance in which the `span` field must be set later, so I have added a setter for that field too.

## Checklist


forc-pkg/src/pkg.rs
```diff
@@ -2732,7 +2733,8 @@ pub fn check(
                     .program_id(engines)
                     .read(engines, |m| m.clone());
                 //                module.name = Some(Ident::new_no_span(pkg.name.clone()));
-                module.span = Some(
+                //                module.span = Some(
```
JoshuaBatty, line 2736: Do we need this commented line in here?

forc-pkg/src/pkg.rs
```diff
@@ -2477,9 +2476,9 @@ pub fn build(
         }
 
         if let TreeType::Library = compiled.tree_type {
-            compiled.root_module.write(&engines, |root_module| {
-                root_module.name = Some(Ident::new_no_span(pkg.name.clone()));
-            });
+            //            compiled.root_module.write(&engines, |root_module| {
```
IGI-111, line 2479: Same here, should these be removed instead of commented out?

IGI-111 (approved):

JoshuaBatty (approved):

PR #6297, opened 2024-07-25
cache pid lock files to reduce IO operations

JoshuaBatty:
## Description
Implements a new `PidLockedFiles` type for efficient file locking. Before this we were doing file IO on every did change key press event. We now cache the results for efficiency.


sdankel (approved):
Looks good!

PR #6304, opened 2024-07-27
Correct code_size function implementation

SwayStar123:
## Description
The current code_size function returns a pointer, rather than the value, this pr corrects this mistake.

## Checklist


test/src/sdk-harness/test_projects/call_frames/mod.rs
```diff
@@ -36,7 +36,7 @@ async fn can_get_id_contract_id_this() {
 async fn can_get_code_size() {
     let (instance, _id) = get_call_frames_instance().await;
     let result = instance.methods().get_code_size().call().await.unwrap();
-    assert!(is_within_range(result.value));
+    assert!(10000 > result.value && result.value > 1000);
```
IGI-111, line 39: Where do those magic numbers come from? This should be explained by a comment at least.

SwayStar123: Well they didnt really come from anywhere. the is_within_range function was just checking if the number was between 0 and (1024 * 1024 * 64), which is not a very useful test. 
I changed it to between 1000 and 10000 so that future optimizations dont break this test.
Also the pointer is decoded as 11000 something so this range also excludes that.

I will add a comment

bitzoic (approved):

K1-R1 (approved):

PR #6308, opened 2024-07-29
Handle case when `Bytes` appends to `self`

bitzoic:
## Description

When calling `self.append(self)` for `Bytes`, we experienced undefined behavior as the logic would clear `self`. Now if this case is encountered, nothing is appended or cleared and `self` is returned.

## Checklist

### DIFF
--- a/sway-lib-std/src/bytes.sw
+++ b/sway-lib-std/src/bytes.sw
@@ -700,6 +700,10 @@ impl Bytes {
             return;
         };
 
+        if self.buf.ptr == other.buf.ptr {
+            return
+        };
+
         let both_len = self.len + other_len;
         let other_start = self.len;
 
--- a/test/src/in_language_tests/test_programs/bytes_inline_tests/src/main.sw
+++ b/test/src/in_language_tests/test_programs/bytes_inline_tests/src/main.sw
@@ -767,6 +767,35 @@ fn bytes_append_to_empty() {
     };
 }
 
+
+#[test()]
+fn bytes_append_self() {
+    let (mut bytes, a, b, c) = setup();
+    assert(bytes.len() == 3);
+    assert(bytes.get(0).unwrap() == a);
+    assert(bytes.get(1).unwrap() == b);
+    assert(bytes.get(2).unwrap() == c);
+
+    let length = bytes.len();
+    let capacity = bytes.capacity();
+
+    bytes.append(bytes);
+
+    assert(bytes.len() == length);
+    assert(bytes.capacity() == capacity);
+    let values = [a, b, c];
+    let mut i = 0;
+    while i < 3 {
+        assert(bytes.get(i).unwrap() == values[i]);
+        i += 1;
+    };
+
+    let mut empty_bytes = Bytes::new();
+    empty_bytes.append(empty_bytes);
+    assert(empty_bytes.len() == 0);
+    assert(empty_bytes.capacity() == 0);
+}
+
 #[test()]
 fn bytes_eq() {
     let (mut bytes, _a, _b, _c) = setup();


SwayStar123:
This is not the behaviour id expect personally. Id expect this to duplicate the array.
eg psuedocode

```
let a = [1,2,3]
let b = a.append(a)
assert( b == [1,2,3,1,2,3])
```

IGI-111:
I think there's ambiguity here since appending should both add the elements to the slice it's called on *and* consume the slice argument it's called with. The combination of these two does lead to the identity function, but it is a surprising result.

Rust gets around this problem by making it simply illegal to append onto oneself since that would be a double mutable borrow.

C++ is the more similar memory model to us and this:
```cpp
#include <string>

int main() {

  std::string str = "abc";
  str.append(str);
  printf("%s\n", str.c_str());

  return 0;
}
```

Does print `abcabc`.

I'm with @SwayStar123 here that doubling the bytes is the more intuitive result. Though it bares mention in the docs.

bitzoic:
> I think there's ambiguity here since appending should both add the elements to the slice it's called on _and_ consume the slice argument it's called with. The combination of these two does lead to the identity function, but it is a surprising result.
> 
> Rust gets around this problem by making it simply illegal to append onto oneself since that would be a double mutable borrow.
> 
> C++ is the more similar memory model to us and this:
> 
> ```c++
> #include <string>
> 
> int main() {
> 
>   std::string str = "abc";
>   str.append(str);
>   printf("%s\n", str.c_str());
> 
>   return 0;
> }
> ```
> 
> Does print `abcabc`.
> 
> I'm with @SwayStar123 here that doubling the bytes is the more intuitive result. Though it bares mention in the docs. And we should be clear about what the exact semantics are here.

I think I would prefer Rust's approach here and make it illegal.

K1-R1:
I agree with @bitzoic that we should follow Rust's approach for `append`. The addition of a `concat` method could be valuable as a follow up PR

K1-R1 (approved):

IGI-111:
Making it illegal is not reasonable without a borrow checker. We can't reliably check that you are trying to mutate the same memory twice at once without heavy annotations.

Sway is like C++ in that way.

We could just panic, but that seems a lot less coherent that picking a correct output for the reflexive call and sticking to it.

bitzoic:
> We can't reliably check that you are trying to mutate the same memory twice at once without heavy annotations.

This is what line 696 does.

https://github.com/FuelLabs/sway/blob/2776aff6e2bfb28728b1babe45bc86f5faa76da1/sway-lib-std/src/bytes.sw#L696

> We could just panic, but that seems a lot less coherent that picking a correct output for the reflexive call and sticking to it.

This is what line 697 does

https://github.com/FuelLabs/sway/blob/2776aff6e2bfb28728b1babe45bc86f5faa76da1/sway-lib-std/src/bytes.sw#L697

sway-lib-std/src/bytes.sw
```diff
@@ -688,6 +693,10 @@ impl Bytes {
     /// }
     /// ```
     pub fn append(ref mut self, ref mut other: self) {
+        if __addr_of::<Self>(self) == __addr_of::<Self>(other) {
```
IGI-111, line 696: Pointer equality is not semantic equality. This check is Undefined Behavior.

The compiler is perfectly allowed to split your struct into multiple ones or unify it with another of the same value so this will not have any coherent.

IGI-111 (changes requested):

sway-lib-std/src/bytes.sw
bitzoic: This check is to ensure the two `Bytes` structs of `self` and `other` do not point to the same struct and works as intended.

`bytes.ptr() == bytes.ptr()` would not work as when `new()` is called, we alloc size of 0 which simply [returns the top of the heap](https://docs.fuel.network/docs/specs/fuel-vm/instruction-set/#aloc-allocate-memory).

This passes:
```sway
assert(alloc(0) == alloc(0));
```

With this check we do not encounter this problem:

```sway
let bytes_1 = Bytes::new();
let bytes_2 = Bytes::new();

// This fails, they have the same pointer
assert(bytes_1.ptr() != bytes_2.ptr());

let bytes_3 = Bytes::with_capacity(1);
let bytes_4 = Bytes::with_capacity(1);

// This passes
assert(bytes_3.ptr() != bytes_4.ptr());
```

IGI-111: You don't understand. Just because this works most of the time does not mean it would work all the time or is even legal Sway.

You can't assume that just because the pointers of two structs are equal, these are the same struct or that these are not different structs. The compiler is allowed to break that assumption.

The code of append with that check would behave differently depending on what particular aliasing optimizations are done or not, and that's illegal.

xunilrj: What would happen with

```sway
fn id<T>(v: T) -> T { v }

let mut b= Bytes::with_capacity(1);
// fill b with some data
b.append(id(b));
````

I have the impression that `self` and `other` inside the `append` function will not have the same address, but they will still point to the same buffer.

This happen because `sway` still do not move semantics.

But even before all that, we do need to define the semantics of this method better. We can make this method always work, copying `other`'s buffer into a temporary buffer first.

Wasteful, but will always work. This c++ `std::memmove()` semantics (https://en.cppreference.com/w/cpp/string/byte/memmove)

> The objects may overlap: copying takes place as if the characters were copied to a temporary character array and then the characters were copied from the array to dest.

To optimize this we would need to make sure buffers do not intersect:  would need to check if both "buffers" (from `ptr`  until `ptr + len`) do no intersect;

Avoiding the temp buffer in this case.

> bytes.ptr() == bytes.ptr() would not work as when new() is called, we alloc size of 0 which simply returns the top of the heap.

This could easily be fixed moving the `ptr` checking after the "size optimization", right? If `other.len() == 0` we do not need to do anything.

Comment (IGI-111) [reply sway-lib-std/src/bytes.sw:696]: Frankly, I'd rather we sidestep the problem and avoid having to deal with implementing move semantics for now.

File: sway-lib-std/src/bytes.sw [hunk 2]
```diff
@@ -718,8 +723,10 @@ impl Bytes {
         // set capacity and length
         self.len = both_len;
 
-        // clear `other`
-        other.clear();
+        // clear `other`, if self == other then we do not want to clear self
+        if self.ptr() != other.ptr() {
```
IGI-111, line 727: This is still UB. You can't use pointer equality to check unicity.

IGI-111 (changes requested):

sway-lib-std/src/bytes.sw
bitzoic: If one cannot compare `self` and `other`, how do you suggest not clearing in this care or reverting in my original implementation? This is absolutely required to do.

IGI-111: What I'm suggesting is that we do not clear `other` and not consume the appended values as Rust does. That way the algorithm doesn't have to branch on something that it can't branch on.

Where the pointer comes from doesn't matter, this would still result in different behavior is the compiler aliases things differently which is illegal.

IGI-111 (approved):

K1-R1 (approved):

PR #6450, opened 2024-08-22
Add the `indoc` crate to `swayfmt` tests for better readability. 

JoshuaBatty:
## Description
We recently started using indoc to improve readability in fuelup. This PR just adds uses the `indoc!` macro to all the existing tests.


alfiedotwtf (approved):
Nice. TIL about `indoc`.

This PR reminded me to check that old `rustfmt` vertical alignment issue. Looks like [struct_field_align_threshold](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#struct_field_align_threshold) exists now (but off by default for now) 🤗

sdankel (commented):
With indoc, we wouldn’t be able to test if the first line of a file was indented incorrectly, since it uses the first non-space character as the starting column. I'm not sure if it's worth adding it.

JoshuaBatty:
> With indoc, we wouldn’t be able to test if the first line of a file was indented incorrectly, since it uses the first non-space character as the starting column. I'm not sure if it's worth adding it.

That's a fair point. Since `indoc` is used per test then you could always choose not to use `indoc` for a test that was testing incorrect spacing on the first line. I think the improvement in readability it offers for the rest of the tests are worth using though.

sdankel (approved):

kayagokalp (approved):

IGI-111 (approved):

PR #6461, opened 2024-08-23
Fixes Numeric type propagation

### LINKED ISSUE #6371: Type Propagation With Numerics
CS-FSSA-005

Variables in Sway are strongly typed i.e., there cannot be any implicit changes in their type. Consider the
following example:

```sway
script;
trait MyTrait{
 fn foo(ref mut self);
}
impl MyTrait for u64 {
 fn foo(ref mut self) {
 let x: &mut u64 = &mut self;
 log("64");
 }
}
impl MyTrait for u32 {
 fn foo(ref mut self) {
 let x: &mut u32 = &mut self;
 log("32");
 }
}
impl u64 {
 fn baz(self) {
 log("64");
 }
}
impl u32 {
 fn baz(self) {
 log("32");
 }
}
fn bar<T>(ref mut x: T) where T: MyTrait {
 x.foo();
}
fn main() {
 let mut a = 0;
 bar(a);
 let x: &u32 = &a;
 a.baz();
}
```

In the example, we define MyTrait which we then implement for u32 and u64. Moreover, we implement
function baz for these types. In main() we declare variable a. Since a is assigned a literal with no type
ascription, it is initially typed as numeric. We execute bar(a). The type checker at this point will
resolve to `bar::<u64>` and eventually calls foo() for u64 which logs 64. This means that a should be
typed as u64. However, we later assign a reference of a to x which is explicitly typed as &u32 which
coerces a to u32. We later, call a.baz() which calls foo() for u32 and logs 32. This script is
successfully compiled and executed as described, implying a type-propagation failure.

esdrubal:
## Description

To fix the Numeric type propagation we now do the type checking of code blocks twice, first to collect all the unification and second to unify the types of a previous namespace with the variable declarations types of the current namespace.

This change incurs a performance drop in the compilation time. The benchmark `compile` takes 25% more time to run.

This can be improved significantly by using LexicalScopes instead of cloning the whole namespace on each scope.

Fixes #6371

## Checklist


IGI-111 (approved):

JoshuaBatty:
25% performance regression is pretty hard pill to swallow. Is there an open issue for removing the namespace cloning and using LexicalScopes like you mention @esdrubal ? If not, would you mind opening one and linking this PR to the issue so this regression can at least just be temporary?

IGI-111 (approved):

IGI-111 (approved):

PR #6473, opened 2024-08-28
feat: add forc-node command for easily bootstrapping a node

kayagokalp:
## Description

WIP at the moment, I have couple of code duplications to remove but we can start playing with this.

## Forc Node

Forc node is a new plugin for easily bootstrapping fuel-core instances with sensible defaults. We currently support:

1. Ignition
2. Testnet
3. Local

## Ignition

This configuration automatically syncs with ignition network, basically run `forc node ignition` and the created node instance will start syncing. 

The defaults are taken from: https://docs.fuel.network/guides/running-a-node/running-a-mainnet-node/

```console
> ./target/debug/forc-node ignition
✔ Do you have a keypair in hand? · yes
✔ Peer Id: · 16Uiu2HAm6U2QNh35FAoob8LNS9zA6tz4BSrsyCmJxgpEzmYRNgrk
✔ Secret: · ********
? Ethereum RPC (Sepolia) Endpoint: › https://eth-mainnet.g.alchemy.com/v2/....
2024-11-28T06:06:36.369956Z  INFO fuel_core_bin::cli::run: 315: `[Importer, P2P, Producer, TxPool, GraphQL]` metrics are enabled
2024-11-28T06:06:36.370965Z  INFO fuel_core_bin::cli::run: 345: Block production disabled
2024-11-28T06:06:36.370973Z  WARN fuel_core_bin::cli::run: 405: The coinbase recipient `ContractId` is not set!
2024-11-28T06:06:36.370981Z  INFO fuel_core_bin::cli::run: 608: Fuel Core version v0.40.0
...
```

## Testnet

This is for connecting to the latest testnet. Similar to mainnet only thing required is to run `forc node testnet` to start a node created for syncing with testnet.

The defaults are taken from: https://docs.fuel.network/guides/running-a-node/running-a-testnet-node/

```console
> ./target/debug/forc-node testnet
✔ Do you have a keypair in hand? · yes
✔ Peer Id: · 16Uiu2HAm6U2QNh35FAoob8LNS9zA6tz4BSrsyCmJxgpEzmYRNgrk
✔ Secret: · ********
? Ethereum RPC (Sepolia) Endpoint: › https://eth-sepolia.g.alchemy.com/v2/....
2024-11-28T06:06:36.369956Z  INFO fuel_core_bin::cli::run: 315: `[Importer, P2P, Producer, TxPool, GraphQL]` metrics are enabled
2024-11-28T06:06:36.370965Z  INFO fuel_core_bin::cli::run: 345: Block production disabled
2024-11-28T06:06:36.370973Z  WARN fuel_core_bin::cli::run: 405: The coinbase recipient `ContractId` is not set!
2024-11-28T06:06:36.370981Z  INFO fuel_core_bin::cli::run: 608: Fuel Core version v0.40.0
...
```

## Local

This is a in-memory instance mostly suited for local developments with instant block production and debug mode enabled in the fuel-core level.
```console
> ./target/debug/forc-node local
2024-11-28T06:04:44.654891Z  INFO fuel_core_bin::cli::run: 315: `[Importer, P2P, Producer, TxPool, GraphQL]` metrics are enabled
2024-11-28T06:04:44.655762Z  INFO fuel_core_bin::cli::run::relayer: 56: Relayer service disabled
2024-11-28T06:04:44.655769Z  INFO fuel_core_bin::cli::run::p2p: 259: P2P service disabled
2024-11-28T06:04:44.655797Z  INFO fuel_core_bin::cli::run: 343: Block production mode: Instant
...
```

## Dry run

It is possible that users of this new plugin might not want to run the node without seeing the parameters send to it. So forc-node has a dry run mode that can be enabled `forc-node --dry-run <ignition|testnet|local>`  which will instead of running the node print the command that would achieve the same result:

```console
> ./target/debug/forc-node --dry-run local
fuel-core run --debug --snapshot /Users/kayagokalp/.forc/chainspecs/local --db-type in-memory --poa-instant true
```

## TODO

- [] Add docs to sway book plugins section
- [] Add CI check to compare cargo fuel-core version with min supported


forc-plugins/forc-node/src/consts.rs
```diff
@@ -0,0 +1,17 @@
+/// Minimum fuel-core version supported.
+pub const MIN_FUEL_CORE_VERSION: &str = "0.40.0";
+
+pub const TESTNET_RESERVED_NODE: &str = "/dns4/p2p-testnet.fuel.network/tcp/30333/p2p/16Uiu2HAmDxoChB7AheKNvCVpD4PHJwuDGn8rifMBEHmEynGHvHrf";
```
mchristopher, line 4: We should use the same setup for testnet that we do for mainnet. Primarily - testnet should use bootstrap nodes the same as mainnet does.

It should also use `/dnsaddr/testnet.fuel.network.` instead of a specific node like above.

forc-plugins/forc-node/src/consts.rs
```diff
@@ -0,0 +1,17 @@
...
+pub const TESTNET_SERVICE_NAME: &str = "fuel-sepolia-testnet-node";
+pub const TESTNET_SYNC_HEADER_BATCH_SIZE: u32 = 100;
+pub const TESTNET_RELAYER_LISTENING_CONTRACT: &str = "0x01855B78C1f8868DE70e84507ec735983bf262dA";
+pub const TESTNET_RELAYER_DA_DEPLOY_HEIGHT: u32 = 5827607;
+pub const TESTNET_RELAYER_LOG_PAGE_SIZE: u32 = 500;
+pub const TESTNET_SYNC_BLOCK_STREAM_BUFFER_SIZE: u32 = 30;
+
+pub const MAINNET_BOOTSTRAP_NODE: &str = "/dnsaddr/mainnet.fuel.network";
```
mchristopher, line 12: We should add a `.` to the end of this hostname to avoid any potential DNS resolution issues on the client side. So the value should be `/dnsaddr/mainnet.fuel.network.`

mchristopher (commented):
Small comments.

Also, can we remove the wasm bytecodes & benchmark JSON files? Those are not needed at runtime (it will pull updates from the live network if needed), and I don't like committing or shipping binary blobs in this repo if we can avoid it.

forc-plugins/forc-node/src/consts.rs
kayagokalp: Sounds good to me! Fixing it now, we might want to update the this docs as well as I was following this one: https://docs.fuel.network/guides/running-a-node/running-a-testnet-node/

forc-plugins/forc-node/src/chain_config.rs
```diff
@@ -0,0 +1,76 @@
+//! This module creates copies of chain configs on the disk so that `forc-node`
+//! will always have a pinned instance of a chain config for given `Mode`.
```
mchristopher, line 2: Do we actually want to do this? It makes sense for local, but I'm not sure user copies are a good idea for testnet & mainnet.

If/when we need to update the network config (gas cost, change privileged addresses, etc), we will publish an update to the `chain_config.json` and/or the `state_config.json` files. All nodes must have matching values in those files in order to connect to the network.

I can't think of a use-case for users having different values for those files on deployed networks, and I think we could cause confusion during network upgrades if we allow users to have local copies of them ("why won't my node connect?!?"). Though for running the "local" network, local copies make a ton of sense IMO.

forc-plugins/forc-node/src/util.rs
```diff
@@ -0,0 +1,147 @@
...
+/// If not, bails out with a help message about how to generate a keypair.
+pub(crate) fn ask_user_keypair() -> Result<KeyPair> {
+    let has_keypair = ask_user_yes_no_question("Do you have a keypair in hand?")?;
+    if has_keypair {
+        // ask the keypair
+        let peer_id = ask_user_string("Peer Id:")?;
+        let secret = ask_user_discreetly("Secret:")?;
+        Ok(KeyPair { peer_id, secret })
+    } else {
+        bail!("Please create a keypair with `fuel-core-keygen new --key-type peering`");
```
mchristopher, line 99: Can we generate a keypair instead of requiring the user to manually do this? For the vast majority of users, just taking care of automatically setting up the keypair would be the desired behavior.

forc-plugins/forc-node/src/chain_config.rs
kayagokalp: I realize the wording of the comment misses an aspect of what happens here so first a little clarification:

This module basically writes the embedded (I embed the chain config folders into the forc-node binary) chain config that we have for forc-node to the disk as the fuel-core cli expects a path to the config.


re: Network Updates:

In my understanding currently users needs to go and fetch the updated chain config from github, and point to that while running the node if the network config changes. So this is already happening(?) I feel like this automates that workflow. 

Currently if there is an update to the network config, they need to fetch the updated network config from github manually. With this they just need to update forc-node version, which will start using the updated network config files as I was planning to cut a new release of forc-node as we change the network params.

Other option is to check online to see if the config is changed, this enables updates without needing to update the forc-node. I felt like explicit changes with the versioning would be more cleaner but if added ux improvement has more upside than the downside of shadowing the actual changes happening in the configurations we can do that.

kayagokalp: My first thought was that we needed to have chain config by default with forc node because If we do not provide them out of the box, then running testnet/mainnet nodes through forc-node will still be involved. (They need to first fetch the config)

mchristopher: I guess I'm a bit unclear on this code - does it create a local copy that the user is expected to modify? Or is it just writing files to disk and ensuring they're up-to-date?

We definitely need to provide chain config files to users, otherwise this tool won't be much use. But for the networks that we run (testnet & mainnet), users will get no where modifying the chain config files, and IMO it is not an option we should even give them. If users want to start modifying things, that's the point where they can just run a node directly and figure it out themselves.

I just want to make sure we're not creating copies of these files on disk with no plans to ensure they're up-to-date. We need to make sure for testnet & mainnet that the most recent chain config is used on launch, especially if an update has happened since the last time the user ran the tool.

kayagokalp: Oh no this is not for the users to modify actually, this is for unpacking what we embed in to the binary as a folder to the disk so that fuel-core can read from. 

It writes to the disk so that we are sure, the chain config read by the fuel-core is the one provided via the forc-node. Once we update the embedded config for forc-node and cut a new release, the new forc-node version will be making sure that the new config will be used.

So current flow to update config is:

1. Update network config in sway repo
2. Cut a new release
3. Point the (TESTNET/MAINNET) channel to the new release of forc-node

if users are using the latest version of the fuelup channel they will have the up-to date network config.

An alternative to this is to check online the central network config repo to see if there is an update and if so fetching it. This removes the need for cutting a new release of forc-node each time we update the network config. But hides the fact that network config is changing.

We can the the 2nd option too if that sounds preferable.

forc-plugins/forc-node/src/chain_config.rs
JoshuaBatty, line 1: Can we add some doc comments to the pub types and functions here?

forc-plugins/forc-node/src/chain_config.rs
```diff
@@ -0,0 +1,551 @@
...
+    fn from(value: ChainConfig) -> Self {
+        let user_forc_dir = user_forc_directory().join(CONFIG_FOLDER);
+
+        match value {
+            ChainConfig::Local => user_forc_dir.join(LOCAL_CONFIG_FOLDER_NAME),
+            ChainConfig::Testnet => user_forc_dir.join(TESTNET_CONFIG_FOLDER_NAME),
+            ChainConfig::Ignition => user_forc_dir.join(IGNITION_CONFIG_FOLDER_NAME),
+        }
+    }
+}
```
JoshuaBatty, line 45: bit cleaner?
```suggestion
impl From<ChainConfig> for PathBuf {
    fn from(value: ChainConfig) -> Self {
        let folder_name = match value {
            ChainConfig::Local => LOCAL_CONFIG_FOLDER_NAME,
            ChainConfig::Testnet => TESTNET_CONFIG_FOLDER_NAME,
            ChainConfig::Ignition => IGNITION_CONFIG_FOLDER_NAME,
        };
        user_forc_directory().join(CONFIG_FOLDER).join(folder_name)
    }
}
```

forc-plugins/forc-node/src/chain_config.rs
kayagokalp: I implemented a online checker that checks for updates in testnet and mainnet.  For local config, if it doesn't exists we fetch one, but after that we do not check for updates and let the user change it as they see fit and we do not override it in any way.

forc-plugins/forc-node/src/chain_config.rs
kayagokalp: Added in [f03b598](https://github.com/FuelLabs/sway/pull/6473/commits/f03b598a0aeefbc0b03a41c682ff15efc9b4e8d1)

forc-plugins/forc-node/src/local/cmd.rs
```diff
@@ -0,0 +1,10 @@
+use clap::Parser;
+use std::path::PathBuf;
+
+#[derive(Parser, Debug, Clone)]
+pub struct LocalCmd {
```
zees-dev, line 5: Can we also have DB path for local node; it's good to be able to persist local node setup across restarts/runs

zees-dev:
opinion: I've seen `dry-run` mostly used for making transactions or some state-changing/persisting action; I personally think that the default command run when starting up the node should be printed in the `--help` output instead.

forc-plugins/forc-node/src/run_opts.rs
```diff
@@ -0,0 +1,155 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            DbType::InMemory => write!(f, "in-memory"),
+            DbType::RocksDb => write!(f, "rocks-db"),
+        }
+    }
+}
+
+impl RunOpts {
+    pub fn generate_params(self) -> Vec<String> {
```
zees-dev, line 78: nit: Could impl trait; so can `into()` since `Vec` is a common type.

```suggestion
impl From<RunOpts> for Vec<String> {
    fn from(opts: RunOpts) -> Self {
```

forc-plugins/forc-node/src/util.rs
```diff
@@ -0,0 +1,166 @@
...
+            DbConfig::Testnet => user_db_dir.join("ignition-testnet"),
+            DbConfig::Ignition => user_db_dir.join("ignition"),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct KeyPair {
+    pub peer_id: String,
+    pub secret: String,
```
zees-dev, line 33: nit: may not want to expose publicly (taking inspiration from `WalletUnlocked`)
```suggestion
    secret: String,
```

forc-plugins/forc-node/src/local/cmd.rs
zees-dev: This probably implies that `in-memory` is the default option (you may need to use an enum here so we can override to custom path)

forc-plugins/forc-node/src/main.rs
```diff
@@ -0,0 +1,43 @@
...
+    init_tracing_subscriber(Default::default());
+
+    let command = cmd::ForcNodeCmd::parse();
+
+    let mut handle = match op::run(command).await {
+        Ok(handler) => handler,
+        Err(err) => {
+            forc_result_bail!(err);
+        }
+    };
```
zees-dev, line 29: nit:

```suggestion
    let mut handle = op::run(command).await?;
```

zees-dev:
Is it possible to make the `forc-node` a bit more extensible; i.e. user can provide additional CLI args - which would simply be appended onto the underlying `fuel-core run` command.
I.e. something like `forc-node local -- <additional CLI flags>` - where everything after the `--` would simply be appended to the underlying `fuel-core run` command.

This way, if a user wants to run the node with some additional/custom CLI args, they could simply append those args (without us having to make a PR for it for the time being).

forc-plugins/forc-node/src/chain_config.rs
```diff
@@ -0,0 +1,574 @@
...
+            config_vault,
+        }
+    }
+
+    fn get_base_url(&self) -> &str {
+        #[cfg(not(test))]
+        return "https://api.github.com";
+
+        #[cfg(test)]
+        return &self.base_url;
```
zees-dev, line 113: not sure if we need feature gate here since `base_url` is set in `new` for test env. 
```suggestion
        return &self.base_url;
```

zees-dev, line 113: not sure if we need feature gate here since `base_url` is set in `new` for test env. 
```suggestion
        return &self.base_url;
```

forc-plugins/forc-node/src/chain_config.rs
```diff
@@ -0,0 +1,574 @@
...
+    /// utilizing the github content abi.
+    pub async fn check_fetch_required(&self, conf: &ChainConfig) -> anyhow::Result<bool> {
+        if *conf == ChainConfig::Local {
+            return Ok(false);
+        }
+
+        let local_files = match self.check_local_files(conf)? {
+            Some(files) => files,
+            None => return Ok(true), // No local files, need to fetch
+        };
```
zees-dev, line 202: nit: could warn?
```suggestion
        // If no local files exist (None), we need to fetch. Otherwise use existing files.
        let local_files = self.check_local_files(conf)?.unwrap_or_else(|| {
            println_warning("No local configuration files found for {}", conf);
            return Ok(true);
        });
```

forc-plugins/forc-node/src/chain_config.rs
```diff
@@ -0,0 +1,574 @@
...
+/// the configuration exists or not, and if it does we do not alter with it.
+/// If the chain config is missing, we are unpacking the one we embedded into
+/// forc-node.
+pub async fn check_and_update_chain_config(conf: ChainConfig) -> anyhow::Result<()> {
+    let fetcher = ConfigFetcher::new();
+    // If chain config is local we will only check if it exists.
+    // If it does not exists we will unpack the one embedded into forc-node.
+    // Otherwise we will continue with what we have in the path without
+    // overriding it.
+    if conf == ChainConfig::Local {
```
zees-dev, line 306: better to use `match` when checking enum variants

forc-plugins/forc-node/src/chain_config.rs
```diff
@@ -0,0 +1,574 @@
...
+            if update {
+                fetcher.download_config(&conf).await?;
+            } else {
+                bail!(
+                    "Missing local network configuration, create one at {}",
+                    local_conf_dir.display()
+                );
+            }
+        }
+    } else {
```
zees-dev, line 326: nit: could return `Ok(())` above a,d remove the `else {` to reduce nesting

Cargo.toml
```diff
@@ -8,7 +8,7 @@ members = [
     "forc-plugins/forc-debug",
     "forc-plugins/forc-doc",
     "forc-plugins/forc-fmt",
-    "forc-plugins/forc-lsp",
+    "forc-plugins/forc-lsp", "forc-plugins/forc-node",
```
zees-dev, line 11: nit: on next line

forc-plugins/forc-node/src/testnet/op.rs
```diff
@@ -0,0 +1,104 @@
...
+    } else {
+        ask_user_keypair()?
+    };
+
+    let relayer = if let Some(relayer) = cmd.relayer {
+        relayer
+    } else {
+        ask_user_string("Ethereum RPC (Sepolia) Endpoint:")?
+    };
+
```
zees-dev, line 38: nit:
```suggestion
    let relayer = cmd.relayer.unwrap_or_else(|| {
        ask_user_string("Ethereum RPC (Sepolia) Endpoint:").expect("Failed to get RPC endpoint")
    });
```

forc-plugins/forc-node/src/testnet/op.rs
```diff
@@ -0,0 +1,104 @@
...
+            HumanReadableCommand::from(fuel_core_command)
+        ));
+        Ok(None)
+    } else {
+        // Spawn the process with proper error handling
+        let handle = fuel_core_command
+            .spawn()
+            .with_context(|| "Failed to spawn fuel-core process:".to_string())?;
+        Ok(Some(handle))
+    }
```
zees-dev, line 65: nit:
```suggestion
            return Ok(None);
    }
    // Spawn the process with proper error handling
    let handle = fuel_core_command
        .spawn()
        .with_context(|| "Failed to spawn fuel-core process:".to_string())?;
    Ok(Some(handle))
```

zees-dev (commented):
Great work; have a few suggestions above^; otherwise looking good! 🚀

forc-plugins/forc-node/src/chain_config.rs
kayagokalp: I am not interested in other variants.

forc-plugins/forc-node/src/chain_config.rs
kayagokalp: I already do a warnings in the upper level function that calls this so adding here will be duplicate

forc-plugins/forc-node/src/run_opts.rs
kayagokalp: This is not a canonical transformation. One would not expect this to be converted to vec string as params out of this context so I specifically created an explicit function for it.

forc-plugins/forc-node/src/local/cmd.rs
kayagokalp: That is a good point! I think we should enable local node with persisting db but that shouldn't be the default. Maybe we can do it with a flag?

forc-plugins/forc-node/src/chain_config.rs
kayagokalp: Nice one!, cleaned this

forc-plugins/forc-node/src/chain_config.rs
kayagokalp: I broke down that area to remove this nesting further, thanks!

forc-plugins/forc-node/src/local/cmd.rs
zees-dev: Whatever is easiest/maintainable, don't really mind; thought an enum made sense since you'd either have 1 or the other option.
Another option is to use `Either` type (if we're certain there'd only ever be these 2 options); but an additional flag is fine too!

forc-plugins/forc-node/src/cmd.rs
```diff
@@ -0,0 +1,24 @@
+use crate::{ignition::cmd::IgnitionCmd, local::cmd::LocalCmd, testnet::cmd::TestnetCmd};
+use clap::{Parser, Subcommand};
+
+#[derive(Debug, Parser)]
+#[clap(name = "forc node", version)]
+/// Forc node is a wrapper around fuel-core with sensible defaults to provide
+/// easy way of bootstrapping a node for local development, testnet or mainnet.
+pub struct ForcNodeCmd {
+    /// Instead of directly running the fuel-core instance print the command.
```
sdankel, line 9:
```suggestion
    /// Print the fuel-core command without running it.
```

forc-plugins/forc-node/src/cmd.rs
```diff
@@ -0,0 +1,24 @@
...
+    #[arg(long)]
+    pub dry_run: bool,
+    #[command(subcommand)]
+    pub mode: Mode,
+}
+
+#[derive(Subcommand, Debug)]
+pub enum Mode {
+    /// Start a local node for development purposes.
```
sdankel, line 18:
```suggestion
    /// Starts a local node for development purposes.
```

forc-plugins/forc-node/src/ignition/cmd.rs
```diff
@@ -0,0 +1,30 @@
+use crate::{
+    consts::{DEFAULT_PEERING_PORT, DEFAULT_PORT, MAINNET_BOOTSTRAP_NODE},
+    util::DbConfig,
+};
+use clap::Parser;
+use std::{net::IpAddr, path::PathBuf};
+
+#[derive(Parser, Debug, Clone)]
+pub struct IgnitionCmd {
```
sdankel, line 9: nit: it looks like the options are the same for `TestnetCmd`, it's just the defaults that differ. We could have a common struct for the options that are the same and use `flatten`

forc-plugins/forc-node/src/consts.rs
```diff
@@ -0,0 +1,34 @@
+/// Minimum fuel-core version supported.
+pub const MIN_FUEL_CORE_VERSION: &str = "0.40.0";
```
sdankel, line 2: Will this ever need to be updated?

forc-plugins/forc-node/src/chain_config.rs
```diff
@@ -0,0 +1,577 @@
...
+        }
+    }
+
+    fn get_base_url(&self) -> &str {
+        return &self.base_url;
+    }
+
+    fn build_api_endpoint(&self, folder_name: &str) -> String {
+        format!(
+            "{}/repos/FuelLabs/{}/contents/{}",
```
sdankel, line 111: If we're always fetching from master, would this mean forc-node might be using a newer version of the chain config than what's actually deployed?

forc-plugins/forc-node/src/run_opts.rs
```diff
@@ -0,0 +1,155 @@
+use std::{fmt::Display, path::PathBuf};
+
+/// Possible parameters to set while integrating with `fuel-core run`.
+#[derive(Debug, Default)]
+pub struct RunOpts {
```
sdankel, line 5: Did you consider using the [run::Command](https://github.com/FuelLabs/fuel-core/blob/856d5da980efce99a1b63cdd8e3f08bb95e42d60/bin/fuel-core/src/cli/run.rs#L102) directly from fuel-core and call [exec](https://github.com/FuelLabs/fuel-core/blob/856d5da980efce99a1b63cdd8e3f08bb95e42d60/bin/fuel-core/src/cli/run.rs#L654) to avoid calling the fuel-core CLI? 

I could imagine there being confusion if the user has the latest version of forc-node but an older version of fuel-core and expecting forc-node to use the latest version. It could also be confusing the other way around though. Just curious to hear your thoughts on this.

With the current approach, how would we know if a new version of fuel-core came out that was incompatible with the CLI args passed below (say, an arg name changed)?

forc-plugins/forc-node/src/chain_config.rs
```diff
@@ -0,0 +1,577 @@
...
+        println_warning(&format!(
+            "Local node configuration files are missing at {}",
+            local_conf_dir.display()
+        ));
+        // Ask user if they want to update the chain config.
+        let update = ask_user_yes_no_question("Would you like to download network configuration?")?;
+        if update {
+            fetcher.download_config(&ChainConfig::Local).await?;
+        } else {
+            bail!(
```
sdankel, line 300: For some reason these errors aren't being formatted properly.
image

forc-plugins/forc-node/src/util.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::process::Command;
+
+    #[test]
+    fn test_basic_command() {
+        let mut command = Command::new("fuel-core");
```
sdankel, line 129: I'm not seeing where fuel-core gets installed for these tests. In CI, do they always run with the latest version of fuel-core?

sdankel (commented):
Great work on this! 🏆

forc-plugins/forc-node/src/local/op.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    let run_opts = RunOpts::from(cmd);
+    let params = run_opts.generate_params();
+
+    let mut fuel_core_command = Command::new("fuel-core");
+    fuel_core_command.arg("run");
+    fuel_core_command.args(params.as_slice());
+
+    if dry_run {
+        println_green(&format!(
```
sdankel, line 25: It'd be nice to print the command even when it's not a dry run.

forc-plugins/forc-node/src/testnet/cmd.rs
```diff
@@ -0,0 +1,30 @@
...
+    #[clap(long = "ip", default_value = "0.0.0.0")]
+    pub ip: IpAddr,
+    #[clap(long = "port", default_value_t = DEFAULT_PORT)]
+    pub port: u16,
+    #[clap(long = "peering-port", default_value_t = DEFAULT_PEERING_PORT)]
+    pub peering_port: u16,
+    #[clap(long = "db-path", default_value = default_testnet_db_path().into_os_string())]
+    pub db_path: PathBuf,
+    #[clap(long = "bootstrap-node", default_value_t = TESTNET_BOOTSTRAP_NODE.to_string())]
+    pub bootstrap_node: String,
```
sdankel, line 25: nit: these are the same as the default long arg, so it's not necessary to specify them.

```suggestion
    #[clap(long)]
    pub peer_id: Option<String>,
    #[clap(long)]
    pub secret: Option<String>,
    #[clap(long)]
    pub relayer: Option<String>,
    #[clap(long, default_value = "0.0.0.0")]
    pub ip: IpAddr,
    #[clap(long, default_value_t = DEFAULT_PORT)]
    pub port: u16,
    #[clap(long, default_value_t = DEFAULT_PEERING_PORT)]
    pub peering_port: u16,
    #[clap(long, default_value = default_testnet_db_path().into_os_string())]
    pub db_path: PathBuf,
    #[clap(long, default_value_t = TESTNET_BOOTSTRAP_NODE.to_string())]
    pub bootstrap_node: String,
```

forc-plugins/forc-node/src/ignition/cmd.rs
```diff
@@ -0,0 +1,30 @@
...
+    #[clap(long = "ip", default_value = "0.0.0.0")]
+    pub ip: IpAddr,
+    #[clap(long = "port", default_value_t = DEFAULT_PORT)]
+    pub port: u16,
+    #[clap(long = "peering-port", default_value_t = DEFAULT_PEERING_PORT)]
+    pub peering_port: u16,
+    #[clap(long = "db-path", default_value = default_ignition_db_path().into_os_string())]
+    pub db_path: PathBuf,
+    #[clap(long = "bootstrap-node", default_value_t = MAINNET_BOOTSTRAP_NODE.to_string())]
+    pub bootstrap_node: String,
```
sdankel, line 25: nit: these are the same as the default long arg, so it's not necessary to specify them.

```suggestion
    #[clap(long)]
    pub peer_id: Option<String>,
    #[clap(long)]
    pub secret: Option<String>,
    #[clap(long)]
    pub relayer: Option<String>,
    #[clap(long, default_value = "0.0.0.0")]
    pub ip: IpAddr,
    #[clap(long, default_value_t = DEFAULT_PORT)]
    pub port: u16,
    #[clap(long, default_value_t = DEFAULT_PEERING_PORT)]
    pub peering_port: u16,
    #[clap(long, default_value = default_ignition_db_path().into_os_string())]
    pub db_path: PathBuf,
    #[clap(long, default_value_t = MAINNET_BOOTSTRAP_NODE.to_string())]
    pub bootstrap_node: String,
```

forc-plugins/forc-node/src/consts.rs
kayagokalp: This can be kept same or updated given the breaking updates to fuel-core run cli interface. Until we have a breaking change there this can stay same, but this gives us a chance to control the environment forc-node is run. Because it is possible that the user is not using fuelup and might be maintaining their own versions

forc-plugins/forc-node/src/util.rs
kayagokalp: This is contained in forc-node, while WalletUnlocked is exposed to other crates and thus encapsulation is needed. Here I am saving some extra getter code as this is simply a short lived, pass through type.

forc-plugins/forc-node/src/util.rs
kayagokalp: These do not need fuel-core itself as these tests only tests the pretty printer. Adding integration tests are tricky and only possible for local node case (as others basically needs to sync with rest of the network > 24h to run a single test 😄) I'll try to come-up with the integration tests for happy paths for local shortly

forc-plugins/forc-node/src/local/op.rs
kayagokalp: Hmm that is an interesting idea! I think we should do it. Adding it.

kayagokalp:
I'll move forc-node tests to a sepearate run because it needs fuel-core installed. That's why test workspace stuff is failing but the code itself is ready to be reviewed! @FuelLabs/tooling

sdankel:
@kayagokalp one test is still failing:

```
---- op::call::tests::contract_call_with_abi stdout ----
thread 'op::call::tests::contract_call_with_abi' panicked at forc-plugins/forc-client/src/op/call/mod.rs:499:9:
assertion `left == right` failed
  left: "[42, 42, 0, 4718592, 131072, 65536, 0, 0, 0, 0]"
 right: "[42, 42, 0, 4718592, 65536, 65536, 0, 0, 0, 0]"

```

zees-dev:
> @kayagokalp one test is still failing:
> 
> ```
> ---- op::call::tests::contract_call_with_abi stdout ----
> thread 'op::call::tests::contract_call_with_abi' panicked at forc-plugins/forc-client/src/op/call/mod.rs:499:9:
> assertion `left == right` failed
>   left: "[42, 42, 0, 4718592, 131072, 65536, 0, 0, 0, 0]"
>  right: "[42, 42, 0, 4718592, 65536, 65536, 0, 0, 0, 0]"
> ```

That's a test I wrote. Feel free to comment out for now - or only confirm the first 3 bytes (42, 42, 0). 
The rest is remaining memory for which there isn't a guarantee on what values are read; ideally a revert should occur but it doesn't.

docs/book/src/forc/plugins/forc_node.md
zees-dev, line 1: Would be good to have docs for this?
Or will this be a seperate PR?

forc-plugins/forc-node/src/util.rs
```diff
@@ -0,0 +1,225 @@
...
+/// ```
+pub struct HumanReadableCommand<'a>(&'a Command);
+
+impl From<DbConfig> for PathBuf {
+    fn from(value: DbConfig) -> Self {
+        let user_db_dir = user_forc_directory().join(DB_FOLDER);
+        match value {
+            DbConfig::Local => user_db_dir.join("local"),
+            DbConfig::Testnet => user_db_dir.join("ignition-testnet"),
+            DbConfig::Ignition => user_db_dir.join("ignition"),
```
zees-dev, line 55: nit: can make these constants

forc-plugins/forc-node/src/run_opts.rs
```diff
@@ -0,0 +1,155 @@
...
+    pub(crate) relayer: Option<String>,
+    /// The IP address to bind the GraphQL service to.
+    pub(crate) ip: Option<std::net::IpAddr>,
+    /// The port to bind the GraphQL service to.
+    pub(crate) port: Option<u16>,
+    /// p2p network's TCP port.
+    pub(crate) peering_port: Option<u16>,
+    /// The path to the database, only relevant if the db type is not
+    /// "in-memory".
+    pub(crate) db_path: Option<PathBuf>,
```
zees-dev, line 28: Maybe struct should be embedded in the `DbType` enum? i.e. `DbType::RocksDB(Option<Pathbuf>)`
^ unless this is not supported by clap.

forc-plugins/forc-node/src/util.rs
zees-dev, line 1: a lot of these utils seem like good generic utils; maybe we can move them up 1 level; or to another crate - most of these are to do with IO.
Could be a future PR though.

forc-plugins/forc-node/src/run_opts.rs
kayagokalp: Not sure if this is supported

forc-plugins/forc-node/src/util.rs
kayagokalp: Not sure where we would use them again. If we end up needing sure we can move it at that time

docs/book/src/forc/plugins/forc_node.md
kayagokalp: As we have been discussing it the tooling syncs, docs are getting ready by @calldelegation in a separate run.

forc-plugins/forc-node/src/chain_config.rs
kayagokalp: Oh looks like i missed this comment but basically the repo is always up to date with what is deployed. They update the chain config repo after updating the deployed network environment and i think they are also planning to set up this as a CI as well. 

Basically the users are already being tunneled to this repo and download the latest version on master right now and that is the version we expect the overall network to be on. So this should be fine.

forc-plugins/forc-node/src/ignition/op.rs
```diff
@@ -0,0 +1,106 @@
...
+        }
+    } else {
+        ask_user_keypair()?
+    };
+
+    let relayer = if let Some(relayer) = cmd.connection_settings.relayer {
+        relayer
+    } else {
+        ask_user_string("Ethereum RPC (mainnet) Endpoint:")?
+    };
```
zees-dev, line 39: nit:
```suggestion
    let relayer = cmd.connection_settings.relayer.unwrap_or_else(|| ask_user_string("Ethereum RPC (mainnet) Endpoint:").unwrap());
```

forc-plugins/forc-node/src/ignition/op.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    if dry_run {
+        Ok(None)
+    } else {
+        // Spawn the process with proper error handling
+        let handle = fuel_core_command
+            .spawn()
+            .with_context(|| "Failed to spawn fuel-core process:".to_string())?;
+        Ok(Some(handle))
+    }
```
zees-dev, line 68: nit: can remove nesting (if-else):

```suggestion
    if dry_run {
        return Ok(None);
    }
    // Spawn the process with proper error handling
    let handle = fuel_core_command
        .spawn()
        .with_context(|| "Failed to spawn fuel-core process:".to_string())?;
    Ok(Some(handle))
```

forc-plugins/forc-node/src/local/op.rs
```diff
@@ -0,0 +1,60 @@
...
+    let mut fuel_core_command = Command::new("fuel-core");
+    fuel_core_command.arg("run");
+    fuel_core_command.args(params.as_slice());
+
+    println_green(&format!(
+        "{}",
+        HumanReadableCommand::from(&fuel_core_command)
+    ));
+
+    if dry_run {
```
zees-dev, line 29: nit: same if-else nit could apply here

forc-plugins/forc-node/src/ignition/op.rs
kayagokalp: There is a semantic difference betweent he original piece and suggestion. 

Suggestion panics. Original one propagetes the error. If you want to ask for a single line/functional style matching something like:

```rust
let relayer = cmd.connection_settings.relayer
    .map_or_else(
        || ask_user_string("Ethereum RPC (mainnet) Endpoint:"),
        Ok
    )?;
```
is needed because you need to be returning a result from a closure, and unfortunately rust is not very good at this as it has a rigid type system and no implicit return type coercion.

forc-plugins/forc-node/src/chain_config.rs
```diff
@@ -0,0 +1,579 @@
...
+
+        let response = self
+            .client
+            .get(&api_endpoint)
+            .header("User-Agent", "forc-node")
+            .send()
+            .await?;
+
+        if !response.status().is_success() {
+            bail!("failed to fetch updates from github")
```
zees-dev, line 142: nit: could provide details on why this failed (e.g. could know at glance if gh is down or something):

```suggestion
            bail!("Failed to fetch updates from GitHub: status {}, body: {}", 
                response.status(),
                response.text().await?)
```

zees-dev (approved):
Great work; LGTM! 🚀 🚀

forc-plugins/forc-node/src/ignition/op.rs
zees-dev: we can preserve error like so i think - but i havent followed through whether it is the same error type:
```rust
let relayer = cmd.connection_settings.relayer.unwrap_or_else(|| ask_user_string("Ethereum RPC (mainnet) Endpoint:"))?;
```
again; a nit - not required.

sdankel (approved):

forc-plugins/forc-node/src/chain_config.rs
kayagokalp: I'll add this later on not to dismiss reviews

JoshuaBatty (approved):

PR #6493, opened 2024-09-03
Fixes impl self duplicated missing error.

### LINKED ISSUE #6381: Trait Functions With the Same Name Get Disambiguated Arbitrarily
CS-FSSA-015

A type can implement two different traits that define the same function. A UFCS-style call syntax is then
needed to explicitly disambiguate because calls like x.foo() will raise a compiler error.
Such an error, however, is not raised if the type is wrapped in a generic struct that has two generic impl
blocks, one bound to the first trait, one to the other, as in the following code snippet:

```sway
script;
trait Cat {
 fn speak(self) -> u64;
}
trait Dog {
 fn speak(self) -> u64;
}
struct S<T> {
 x: T,
}
impl<T> S<T>
where
 T: Cat,
{
 fn foo(self) -> u64 {
 self.x.speak()
 }
}
impl<T> S<T>
where
 T: Dog,
{
 fn foo(self) -> u64 {
 self.x.speak()
 }
}
impl Dog for u64 {
 fn speak(self) -> u64 {
 log("2");
 2
 }
}
impl Cat for u64 {
 fn speak(self) -> u64 {
 log("1");
 1
 }
}
fn main() {
 let s = S::<u64> { x: 1 };
 s.foo();
}
```

The compilation succeeds and the execution output shows that it is the first generic impl (the one bound
to Cat) that prevails.
To the contrary, Rust refuses the compilation of an analogous snippet.

esdrubal:
## Description

While inserting trait implementations into the trait map we now compare the implementing type_id without matching UnkownGenerics trait constraints. Matching the trait constraints was causing a missing error.

Fixes #6381

## Checklist

### DIFF
--- a/sway-lib-std/src/bytes.sw
+++ b/sway-lib-std/src/bytes.sw
@@ -653,7 +653,12 @@ impl Bytes {
         (left_bytes, right_bytes)
     }
 
-    /// Moves all elements of `other` into `self`, leaving `other` empty.
+    /// Copies all elements of `other` into `self`
+    ///
+    /// # Additional Information
+    ///
+    /// NOTE: Appending `self` to itself will duplicate the `Bytes`. i.e. [0, 1, 2] => [0, 1, 2, 0, 1, 2]
+    /// This function differs from the rust `append` function in that it does not clear the `other` `Bytes`
     ///
     /// # Arguments
     ///
@@ -696,7 +701,6 @@ impl Bytes {
         // optimization for when starting with empty bytes and appending to it
         if self.len == 0 {
             self = other;
-            other.clear();
             return;
         };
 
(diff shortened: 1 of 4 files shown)


test/src/e2e_vm_tests/test_programs/should_fail/impl_self_duplicated/test.toml
```diff
@@ -0,0 +1,6 @@
+category = "fail"
+validate_abi = false
+expected_warnings = 3
+
+# check: fn foo(self) -> u64 {
+# nextln: $()Duplicate definitions for the method "foo" for type "S<T>".
```
jjcnn, line 6: This means that the error is triggered when `impl<T> S<T> where T : Dog { ... }` is typechecked.

As far as I can tell Rust doesn't throw the error until `s.foo()` is typechecked in line 41 of the test program.

I'm happy with the current behavior, but it's just worth being aware that we are diverging from Rust here.

Issue #5845, opened 2024-04-12
Collect versions under workspace manifest file and re-use it in the member packages


kayagokalp:
Currently version is specified separately for each member pkg we have in the sway repo, we should use it from workspace to make it easier to bump versions while releasing

krishnabandewar:
Can I take this issue?

kayagokalp:
Hi there thanks for the interest but i think linked PR already did the work on this and this should have been closed.

#6501

kayagokalp:
Sorry meant to link #6563

Issue #6179, opened 2024-06-24
Update all deps to be the latest version across the whole repo


JoshuaBatty:
Also, it might be worth us migrating to defining dependencies used across multiple workspace members in a `[workspace.dependencies]` section. Eg

```toml
[workspace.dependencies]
clap = "4.5"
dashmap = "6.0"
```

Then in the member projections `Cargo.toml` we can bring them in like

```toml
[dependencies]
clap = { workspace = true }
dashmap = { workspace = true }
```

alfiedotwtf:
Most dependencies have now been updated. What's left have been moved into #6536

PR #6501, opened 2024-09-05
Update all Cargo.toml files to use workspace dependency versions

alfiedotwtf:
## Description

This PR relates to #6179 which aims to simplify individual workspace member dependency versioning  by deferring upstream to the workspace dependencies itself.

I've also extended #6179 as it currently only relates to external dependencies - I've noticed that whenever Sway has a version bump, the diffs manually update all of the many `Cargo.toml` files to the new globally shared version. We can also use workspaces here to defer each `Cargo.toml` package version to the workspace package version as well.

This PR is in 3 parts:

1. Update all `Cargo.toml` external dependencies to use the shared versions defined in the workspace dependencies
2. Update all `Cargo.toml` package version to defer to the workspace package version.
3. Update all external dependencies to their latest versions

If an individual workspace member needs to break away from the shared workspace package version, we can still do that:

- Update the individual `Cargo.toml` to the new version (as normal)
- Update all references within the whole repository to state this new version

Notes:

- `[dev-dependencies]` was not factored out since `[workspace.dev-dependencies]` is currently not supported by `Cargo.toml`, meaning all dev dependencies would have to be brought into `[workspace.dependencies]` which I assume is not what we would want to do.
- `[target.'cfg(not(target_os = "macos"))'.dependencies]` was not factored out since these crates should only be brought in if the condition is met, but again as the only way to do this would be to move it to `[workspace.dependencies]`, I would advise to keep them separated.

## Checklist

- [✅] I have linked to any relevant issues.
- [✅] I have commented my code, particularly in hard-to-understand areas.
- [␆] I have updated the documentation where relevant (API docs, the reference, and the Sway book).
   - [␆] If my change requires substantial documentation changes, I have [requested support from the DevRel team](https://github.com/FuelLabs/devrel-requests/issues/new/choose)
- [␆] I have added tests that prove my fix is effective or that my feature works.
- [␆] I have added (or requested a maintainer to add) the necessary `Breaking*` or `New Feature` labels where relevant.
- [✅] I have done my best to ensure that my PR adheres to [the Fuel Labs Code Review Standards](https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/external-contributors.md).
- [✅] I have requested a review from the relevant team or maintainers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -245,6 +245,9 @@ jobs:
           toolchain: ${{ env.RUST_VERSION }}
           targets: wasm32-unknown-unknown
       - uses: Swatinem/rust-cache@v2
+      - name: Install clang
+        if: ${{ env.ACT }}
+        run:  dpkg -l clang || apt-get update && apt-get install -y clang
       - name: Install Forc
         run: cargo install --locked --debug --path ./forc
       - name: Initialize test project
@@ -423,32 +426,40 @@ jobs:
       contents: write
     steps:
       - uses: actions/checkout@v3
-      - name: Install toolchain
+      - name: Install Rust toolchain
+        if: ${{ !env.ACT }}
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ env.RUST_VERSION }}
-      - uses: Swatinem/rust-cache@v2
+      - name: Enable Rust caching
+        if: ${{ !env.ACT }}
+        uses: Swatinem/rust-cache@v2
       - name: Install Forc
+        if: ${{ !env.ACT }}
         run: cargo install --locked --debug --path ./forc
       - name: Run benchmarks
+        if: ${{ !env.ACT }}
         run: ./benchmark.sh
       - name: Checkout benchmark data
-        if: github.event_name != 'push'
+        if: ${{ !env.ACT && github.event_name != 'push' }}
         uses: actions/checkout@v3
         with:
           repository: FuelLabs/sway-performance-data
           path: performance-data
+          token: ${{ secrets.CI_ACCOUNT_PAT }}
       - name: Checkout benchmark data
-        if: github.event_name == 'push'
+        if: ${{ !env.ACT && github.event_name == 'push' }}
         uses: actions/checkout@v3
         with:
           repository: FuelLabs/sway-performance-data
           path: performance-data
           token: ${{ secrets.CI_ACCOUNT_PAT }}
       - name: Prepare benchmarks data for commit
-        if: github.event_name == 'push'
+        if: ${{ !env.ACT && github.event_name == 'push' }}
         run: ./benchmark.sh --prepare-for-commit
-      - uses: EndBug/add-and-commit@v9
+      - name: Commit benchmark data
+        if: ${{ !env.ACT }}
+        uses: EndBug/add-and-commit@v9
         with:
           cwd: "./performance-data"
           message: "Updated benchmark data"
@@ -542,7 +553,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Setup Rust and cargo-nextest
-        uses: moonrepo/setup-rust@v0
+        uses: moonrepo/setup-rust@v1
         with:
           channel: stable
           cache-target: release
@@ -588,7 +599,8 @@ jobs:
       - name: Install cargo-udeps
         run: cargo install --locked cargo-udeps
       - name: Check Unused Deps
-        run: cargo udeps --locked --all-targets
+        # cargo udeps will fail if the cache is empty on first go, so try one more time
+        run: cargo udeps --locked --all-targets || cargo udeps --locked --all-targets
 
   notify-slack-on-failure:
     needs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,8 +15,8 @@ concurrency:
 env:
   CARGO_TERM_COLOR: always
   REGISTRY: ghcr.io
-  RUST_VERSION: 1.79.0
-  NIGHTLY_RUST_VERSION: nightly-2024-07-16
+  RUST_VERSION: 1.81.0
+  NIGHTLY_RUST_VERSION: nightly-2024-09-11
 
 jobs:
   get-fuel-core-version:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -560,7 +560,7 @@ jobs:
         with:
           toolchain: ${{ env.RUST_VERSION }}
       - uses: Swatinem/rust-cache@v2
-      - name: Run forc tests separately 
+      - name: Run forc tests separately
         env:
           RUST_BACKTRACE: full
         run: cargo test --locked --release -p forc -- --nocapture
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,10 +32,63 @@ exclude = [
     "forc-test/test_data"
 ]
 
+[workspace.package]
+edition = "2021"
+version = "0.63.5"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+homepage = "https://fuel.network/"
+license = "Apache-2.0"
+repository = "https://github.com/FuelLabs/sway"
+
 [workspace.dependencies]
+#
+# Internal dependencies in order to propagate `workspace.version`
+#
+
+forc = { path = "forc/" }
+forc-pkg = { path = "forc-pkg/" }
+forc-test = { path = "forc-test/" }
+forc-tracing = { path = "forc-tracing/" }
+forc-util = { path = "forc-util/" }
+
+# Forc plugins
+forc-plugins = { path = "forc-plugins/" }
+forc-client = { path = "forc-plugins/forc-client/" }
+forc-crypto = { path = "forc-plugins/forc-crypto/" }
+forc-debug = { path = "forc-plugins/forc-debug/" }
+forc-doc = { path = "forc-plugins/forc-doc/" }
+forc-fmt = { path = "forc-plugins/forc-fmt/" }
+forc-lsp = { path = "forc-plugins/forc-lsp/" }
+forc-tx = { path = "forc-plugins/forc-tx/" }
+
+sway-ast = { path = "sway-ast/" }
+sway-core = { path = "sway-core/" }
+sway-error = { path = "sway-error/" }
+sway-lsp = { path = "sway-lsp/" }
+sway-parse = { path = "sway-parse/" }
+sway-types = { path = "sway-types/" }
+sway-utils = { path = "sway-utils/" }
+swayfmt = { path = "swayfmt/" }
+
+# Sway IR
+sway-ir = { path = "sway-ir/" }
+sway-ir-macros = { path = "sway-ir/sway-ir-macros" }
+
+#
+# External Fuel dependencies
+#
+
+# Dependencies from the `fuel-abi-types` repository:
+fuel-abi-types = "0.7.0"
+
 # Dependencies from the `fuel-core` repository:
 fuel-core-client = { version = "0.35.0", default-features = false }
 fuel-core-types = { version = "0.35.0", default-features = false }
+
+# Dependencies from the `fuels-rs` repository:
+fuels = "0.66.4"
+fuels-core = "0.66.4"
+fuels-accounts = "0.66.4"
 
 # Dependencies from the `fuel-vm` repository:
 fuel-asm = "0.56.0"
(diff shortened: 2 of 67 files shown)


.github/workflows/ci.yml
```diff
@@ -477,14 +477,11 @@ jobs:
...
 
-      - name: Install cargo-deps
-        run: cargo install cargo-deps
-
       # We require this check to avoid cyclic dependencies between 'fuels' and 'forc-pkg'.
       # Detailed explanation is found in the echo below.
       - name: Check 'forc-pkg' dependencies for 'fuels' crates
         run: |
-          deps=$(cargo deps --manifest-path forc-pkg/Cargo.toml)
+          deps=$(cargo tree --manifest-path forc-pkg/Cargo.toml)
```
sdankel, line 484: Good catch, it looks like cargo-deps does not support workspaces so this wasn't even running.

Voxelot:
nice 😎 
Screenshot from 2024-09-06 11-01-55

Cargo.toml
```diff
@@ -32,32 +32,202 @@ exclude = [
     "forc-test/test_data"
 ]
 
+[workspace.package]
+edition = "2021"
+version = "0.63.4"
```
sdankel, line 37: Just a not: although currently every package in the sway repo is on the same version, there have been talks of doing more granular releases where only the packages that changed get a version bump & release.

We can always change it back later, and for now this will be useful.

Cargo.toml
```diff
@@ -32,32 +32,202 @@ exclude = [
...
+tai64 = { version = "4.0", features = ["serde"] }
+taplo = "0.7"
+tar = "0.4.38"
+tempfile = "3"
+term-table = "1.3"
+termion = "2.0.1"
+textwrap = "0.16.0"
+thiserror = "1.0.30"
+tikv-jemallocator = "0.5"
+tokio = { version = "1.12", features = [ "fs", "io-std", "io-util", "macros", "net", "process", "rt", "rt-multi-thread", "sync", "time"] }
```
sdankel, line 217: I think we do not want to specify any features in this file.

If I understand correctly, this will include all of these features for every package that uses tokio since [inherited dependency features are additive](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace)

Cargo.toml
```diff
@@ -32,32 +32,202 @@ exclude = [
...
+tower = { version = "0.4.12", default-features = false, features = ["util"] }
+tower-lsp = { version = "0.20", features = ["proposed"] }
+tracing = "0.1"
+tracing-subscriber = { version = "0.3", features = [ "ansi", "env-filter", "json" ] }
+uint = "0.9"
+unicode-bidi = "0.3.13"
+unicode-xid = "0.2.2"
+url = { version = "2.2", features = ["serde"] }
+urlencoding = "2.1.2"
+uwuify = "^0.2"
```
sdankel, line 229: From the docs I couldn't tell if this is still optional for its inheritors. If there's a way to test it, great, otherwise I suggest we leave out optional dependencies from the workspace.

Cargo.toml
```diff
@@ -32,32 +32,202 @@ exclude = [
...
+edition = "2021"
+version = "0.63.4"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+homepage = "https://fuel.network/"
+license = "Apache-2.0"
+repository = "https://github.com/FuelLabs/sway"
+
 [workspace.dependencies]
+#
+# Internal dependencies in order to propagate `workspace.version`
```
sdankel, line 45: This is really nice, no more relative paths in `Cargo.toml`s 💯

sdankel (commented):
Looks good, just a few comments.

sdankel:
Closes https://github.com/FuelLabs/sway/issues/5845

Cargo.toml
JoshuaBatty: Good catch, agree we are going to want to change this back once each package can have their own release cycles.

Cargo.toml
JoshuaBatty: agree we should enable features at the member level, it's pretty hard to reason about why all these features are enabled when defined in the workspace manifest.

JoshuaBatty:
Awesome to see this. so much cleaner.

Cargo.toml
alfiedotwtf: Sorry, I read that wrong. I was under the assumption the workspace lists all the features to allow, and each member choses which to enable from that list. Ok cool, removed.

Cargo.toml
alfiedotwtf: Yep, sounds good. As I said in the description, probably best that for now every default to the workspace version and then if/when things depart, they do so individually. It may end up and only a handful diverge at one time, so best to centralise everything at the moment.

Either way, this now gives us the best of both worlds :)

Cargo.toml
alfiedotwtf: Cool. As above, `features` removed from the workspace `Cargo.toml`.

.github/workflows/ci.yml
alfiedotwtf: Thanks for that suggestion. It worked as planned:

https://github.com/FuelLabs/sway/pull/6515#issuecomment-2337406478

Cargo.toml
sdankel: What I meant is, perhaps optional dependencies should not be workspace dependencies since we don't know if they are actually treated as optional when they're part of the workspace.

alfiedotwtf: I [https://doc.rust-lang.org/cargo/reference/features.html#the-default-feature](found in the docs) that by default "all features are disables unless explicitly enabled", so I think we should be ok with this latest change.

... "explicitly enabled" here means the crate has overridden the empty default features list, which seams reasonable given that the crate author believes these features are the minimal set to get going.

sway-core/src/asm_generation/fuel/analyses.rs
```diff
@@ -17,9 +17,9 @@ use crate::asm_lang::{ControlFlowOp, Label, Op, VirtualRegister};
 /// Two tables are generated: `live_in` and `live_out`. Each row in the tables corresponds to an
 /// instruction in the program.
 /// * A virtual register is in the `live_out` table for a given instruction if it is live on any
-/// of that node's out-edges
+///   of that node's out-edges
 /// * A virtual register is in the `live_in` table for a given instruction if it is live on any
-/// of that node's in-edges
+///   of that node's in-edges
```
JoshuaBatty, line 22: looks like there's a ton of these new spaces throughout the comments, just wondering how they got there?

JoshuaBatty: ok out of curiosity which lint is this? Would love to understand what the reason behind it is. Seems a bit off at first glance.

alfiedotwtf:
Just forced-pushed to remove a comment that I accidentally left in while testing

sway-core/src/asm_generation/fuel/analyses.rs
alfiedotwtf: Ah, sorry my mistake... I just reverse those changes locally and it still passes tests. I must have changed these all to silence Clippy:

```
warning: doc list item without indentation
  --> sway-core/src/asm_generation/fuel/analyses.rs:20:5
   |
20 | /// of that node's out-edges
   |     ^
   |
   = help: if this is supposed to be its own paragraph, add a blank line
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_lazy_continuation
   = note: `#[warn(clippy::doc_lazy_continuation)]` on by default
help: indent this line
   |
20 | ///   of that node's out-edges
   |     ++
```

Before this change, clippy was full of this same message which was a distraction.

I guess we could `allow(clippy::doc_lazy_continuation)` instead, but this change looks cleaner to be in the editor because you can see at a glance that it's not just more paragraph text.

sway-core/src/semantic_analysis/type_check_context.rs
```diff
@@ -1201,6 +1201,7 @@ impl<'a> TypeCheckContext<'a> {
             }
 
             if !maybe_method_decl_refs.is_empty() {
+                #[allow(clippy::mutable_key_type)]
```
JoshuaBatty, line 1204: Maybe we shouldn't enforce this clippy lint since we need to disable it in quite a few places...

sway-lsp/Cargo.toml
```diff
@@ -65,8 +68,8 @@ pretty_assertions = "1.4.0"
 rand = "0.8"
 regex = "^1.10.2"
 sway-lsp-test-utils = { path = "tests/utils" }
-tikv-jemallocator = "0.5"
-tower = { version = "0.4.12", default-features = false, features = ["util"] }
+tikv-jemallocator = "0.6.0"
```
JoshuaBatty, line 71: I personally prefer using "x.y" instead of "x.y.z" for dependency versions, as it automatically incorporates bug fixes and security updates within the x.y.* range. What are your thoughts about adopting this approach?

JoshuaBatty (commented):
Nice work Alfie, just a couple of small nits.

.github/workflows/ci.yml
```diff
@@ -423,32 +426,40 @@ jobs:
       contents: write
     steps:
       - uses: actions/checkout@v3
-      - name: Install toolchain
+      - name: Install Rust toolchain
+        if: ${{ !env.ACT }}
```
sdankel, line 430: nit: I'd prefer to do the act-related stuff in a separate PR since it's not related to updating the dependencies

sway-utils/src/performance.rs
```diff
@@ -29,9 +29,9 @@ macro_rules! time_expr {
                 if cfg.metrics_outfile.is_some() {
                     #[cfg(not(target_os = "macos"))]
                     let memory_usage = {
-                        use sysinfo::{System, SystemExt};
+                        use sysinfo::{MemoryRefreshKind, System};
                         let mut sys = System::new();
-                        sys.refresh_system();
+                        sys.refresh_memory_specifics(MemoryRefreshKind::new().with_ram());
```
sdankel, line 34: Why is this changing? It looks like [refresh_memory](https://docs.rs/sysinfo/latest/sysinfo/struct.System.html#method.refresh_memory) is still available in the latest version.

Cargo.toml
```diff
@@ -32,32 +32,202 @@ exclude = [
...
+assert_matches = "1.5.0"
+assert-json-diff = "2.0"
+async-trait = "0.1.58"
+atty = "0.2.14"
+byte-unit = "5.1.4"
+bytecount = "0.6"
+bytes = "1.3.0"
+chrono = { version = "0.4", default-features = false }
+cid = "0.11"
+clap = "4.5.4"
```
sdankel, line 119: image

Cargo.toml
```diff
@@ -32,32 +32,202 @@ exclude = [
...
+include_dir = "0.7.3"
+indexmap = "2.0.0"
+indoc = "2.0"
+insta = "1.39.0"
+ipfs-api-backend-hyper = "0.6"
+itertools = "0.13.0"
+lazy_static = "1.4"
+libp2p-identity = "0.2.4"
+libtest-mimic = "0.7.3"
+lsp-types = "0.94"
```
sdankel, line 161: image

Cargo.toml
```diff
@@ -32,32 +32,202 @@ exclude = [
...
+mdbook = { version = "0.4", default-features = false }
+miden = "0.3.0"
+miden-core = "0.10.5"
+minifier = "0.3.0"
+notify = "6.1.1"
+notify-debouncer-mini = "0.4.1"
+num-bigint = "0.4.3"
+num-traits = "0.2.16"
+object = "0.36.4"
+once_cell = "1.18.0"
```
sdankel, line 171: image

Cargo.toml
```diff
@@ -32,32 +32,202 @@ exclude = [
...
+revm = "14.0.1"
+ropey = "1.5"
+rpassword = "7.2"
+rustc-hash = "2.0.0"
+semver = "1.0"
+serde = "1.0"
+serde_ignored = "0.1.9"
+serde_json = "1.0.91"
+serde_with = "3.3.0"
+serde_yaml = "0.9.27"
```
sdankel, line 198: latest is  "0.9.34"

Cargo.toml
```diff
@@ -32,32 +32,202 @@ exclude = [
...
+strsim = "0.11.1"
+strum = "0.26.3"
+syn = "2.0.77"
+tai64 = "4.0"
+taplo = "0.13.2"
+tar = "0.4.38"
+tempfile = "3"
+term-table = "1.3"
+termion = "4.0.2"
+textwrap = "0.16.0"
```
sdankel, line 214: image

Cargo.toml
```diff
@@ -32,32 +32,202 @@ exclude = [
...
+thiserror = "1.0.30"
+tikv-jemallocator = "0.6.0"
+tokio = "1.12"
+toml = "0.8"
+toml_edit = "0.22.20"
+tower = { version = "0.5.1", default-features = false }
+tower-lsp = "0.20"
+tracing = "0.1"
+tracing-subscriber = "0.3"
+uint = "0.9"
```
sdankel, line 224: image

sway-lsp/tests/lib.rs
```diff
@@ -1638,7 +1638,7 @@ fn hover_docs_for_functions_vscode() {
         req_uri: &uri,
         req_line: 20,
         req_char: 14,
-        documentation: vec!["```sway\npub fn bar(p: Point) -> Point\n```\n---\n A function declaration with struct as a function parameter\n\n---\nGo to [Point](command:sway.goToLocation?%5B%7B%22range%22%3A%7B%22end%22%3A%7B%22character%22%3A1%2C%22line%22%3A5%7D%2C%22start%22%3A%7B%22character%22%3A0%2C%22line%22%3A2%7D%7D%2C%22uri%22%3A%22file","sway%2Fsway-lsp%2Ftests%2Ffixtures%2Ftokens%2Ffunctions%2Fsrc%2Fmain.sw%22%7D%5D \"functions::Point\")"],
+        documentation: vec!["```sway\npub fn bar(p: Point) -> Point\n```\n---\n A function declaration with struct as a function parameter\n\n---\nGo to [Point](command:sway.goToLocation?%5B%7B%22range%22%3A%7B%22end%22%3A%7B%22character%22%3A1%2C%22line%22%3A5%7D%2C%22start%22%3A%7B%22character%22%3A0%2C%22line%22%3A2%7D%7D%2C%22uri%22%3A%22file","sway-lsp%2Ftests%2Ffixtures%2Ftokens%2Ffunctions%2Fsrc%2Fmain.sw%22%7D%5D \"functions::Point\")"],
```
sdankel, line 1641: It looks like the `sway/` is getting stripped off the beginning of the file path, which is used for the hover links. I don't think this should change.

sway-lsp/tests/lib.rs
```diff
@@ -1784,7 +1784,7 @@ fn hover_docs_for_self_keywords_vscode() {
...
+        hover.documentation = vec!["```sway\nstruct MyStruct\n```\n---\n\n---\n[2 implementations](command:sway.peekLocations?%5B%7B%22locations%22%3A%5B%7B%22range%22%3A%7B%22end%22%3A%7B%22character%22%3A1%2C%22line%22%3A4%7D%2C%22start%22%3A%7B%22character%22%3A0%2C%22line%22%3A2%7D%7D%2C%22uri%22%3A%22file","sway-lsp%2Ftests%2Ffixtures%2Fcompletion%2Fsrc%2Fmain.sw%22%7D%2C%7B%22range%22%3A%7B%22end%22%3A%7B%22character%22%3A1%2C%22line%22%3A14%7D%2C%22start%22%3A%7B%22character%22%3A0%2C%22line%22%3A6%7D%7D%2C%22uri%22%3A%22file","sway-lsp%2Ftests%2Ffixtures%2Fcompletion%2Fsrc%2Fmain.sw%22%7D%5D%7D%5D \"Go to implementations\")"];
```
sdankel, line 1787: Same comment here

sdankel (commented):
Looks good overall, just a few nits. I highlighted some other dependencies that have newer versions since it looks like you're aiming to upgrade everything in this PR now.

Cargo.toml
JoshuaBatty: It's ok to leave this one at 0.94.

Cargo.toml
JoshuaBatty: Yeah i think we should just be setting it to use 0.9 and leave the minor version part off. Same everywhere really.

Cargo.toml
```diff
@@ -32,32 +32,202 @@ exclude = [
...
+# External dependencies
+#
 
-[workspace.package]
-edition = "2021"
-authors = ["Fuel Labs <contact@fuel.sh>"]
-homepage = "https://fuel.network/"
-license = "Apache-2.0"
-repository = "https://github.com/FuelLabs/sway"
+annotate-snippets = "0.10.1"
```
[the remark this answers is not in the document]
alfiedotwtf: Ah sorry, I mentioned this to @JoshuaBatty but didn't say it here in the comments...

The following 3 crates will take a lot of work before we can upgrade cleanly:

- annotate-snippets
- lsp-types
- uint

Thanks for pointing out the others. "Dependi" wasn't working within Docker so I used "Sparse Crates" instead, but it looks like it's not fully up to date. But now that I'm working on bare metal again, Dependi is showing a lot of red. Thanks!

sway-lsp/tests/lib.rs
alfiedotwtf: I'm not sure `sway` belongs there since that's the name of `repo-root/..` as I only found this out because I do git checkouts via worktrees:

    git worktrees add path/to/sway/<issue-number>/ origin/master

And so when the tests ran, I was getting ".../sway/**_6179_**/sway-lsp/..." which is why I think the `sway/` prefix is hardcoded but shouldn't be there.

sway-utils/src/performance.rs
alfiedotwtf: The `refresh_system()` method was removed in a newer `sysinfo` version and so I looked for a suitable replacement. Although `refresh_memory()` exists, the line following the refresh 
call is `sys.used_memory()`, returning the total `ram` used. Seeing as total `ram` doesn't include `swap` values, I figured that we could possibly save cycles by skipping swap calculations that we never end up reading.

Looking at `sysinfo` internals, Linux provides a nice interface via `/proc/meminfo` which is read once as a whole, but all other platforms end up doing separate system calls to retrieve swap info if swap is set to refresh.

sway-lsp/tests/lib.rs
sdankel: Fair enough. I tested it and the links seem to be working fine

kayagokalp (commented):
Looks like i forgot to post my review 🤦 anyways it seems like the generated bytecode for the proxy contract changed after this one. I wan't to take a closer look but it is not nice (also not sure how this is possible) that dependency updates are causing bytecode changes in the compiler. 

Do we know why?

forc-plugins/forc-client/tests/deploy.rs
```diff
@@ -341,7 +341,7 @@ async fn test_simple_deploy() {
     node.kill().unwrap();
     let expected = vec![DeployedContract {
         id: ContractId::from_str(
-            "50fe882cbef5f3da6da82509a66b7e5e0a64a40d70164861c01c908a332198ae",
```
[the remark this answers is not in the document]
alfiedotwtf: I've turned this PR back into draft, and am pulling out all the changes so that we can see which crates can be updated with no code changes, and which ones end up doing changes like these.

Unfortunately it's a bit time consuming because you end up having to wait for the whole `ci.yml` to finish before the next 'edit -> test -> debug` cycle

sway-core/src/asm_generation/fuel/analyses.rs
alfiedotwtf: Ahh!! It turns out it was an error after all - `clippy::doc_lazy_continuation` was introduced in `1.80.0` and when I tried bumping Rust to `1.81.0`, the `cargo-clippy` GitHub Action workflow started failing - for some reason they made it an `error` rather than a `warning` 🫠

alfiedotwtf: Nice! So apart from that Clippy error, upgrading Rust to `1.81.0` has no other regressions.

PR #6507, opened 2024-09-06
fix code snippet for old diagnostics

xunilrj:
## Description

This PR solves a small bug when calculating the code snippets of some diagnostics.

image

## Checklist


test/src/e2e_vm_tests/test_programs/should_fail/type_check_analyze_errors/stdout.snap
```diff
@@ -152,18 +152,20 @@ error
...
-9 |     Vec::<u16>::new().push(_a);
-  |
+  --> test/src/e2e_vm_tests/test_programs/should_fail/type_check_analyze_errors/src/main.sw:8:14
+   |
+ 6 | 
+ 7 |     // u16
+ 8 |     let _a = 0x10000;
+   |              ^^^^^^^ Literal would overflow because its value does not fit into "u16"
+ 9 |     Vec::<u16>::new().push(_a);
+10 |
```
sdankel, line 167: It seems odd to have this extra blank line in the diagnostic

IGI-111 (dismissed):

forc-util/src/lib.rs
```diff
@@ -716,30 +716,40 @@ fn construct_window<'a>(
...
     debug_assert!(total_lines_in_input >= total_lines_of_highlight);
 
+    // these two are 0-based
     let mut current_line = 0;
     let mut lines_to_start_of_snippet = 0;
+
     let mut calculated_start_ix = None;
     let mut calculated_end_ix = None;
     let mut pos = 0;
+
```
alfiedotwtf, line 726: Since you initially `- 1` and then `+ 1` in multiple places, would it be simpler to leave the original variables and then just change the comparators i.e:

    - if current_line + NUM_LINES_BUFFER >= start.line && calculated_start_ix.is_none() {
    + if current_line + NUM_LINES_BUFFER > start.line && calculated_start_ix.is_none() {

and 

    - if current_line >= end.line + NUM_LINES_BUFFER && calculated_end_ix.is_none() {
    + if current_line > end.line + NUM_LINES_BUFFER && calculated_end_ix.is_none() {

xunilrj: I simplified doing the inverse: starting the counter with 1, instead of zero.

alfiedotwtf: I don't think starting at `1` is correct - as a contrived example if `input` is an empty string, at the end of the function `master` currently does:

    start.line = 0;

Whereas your PR will be:

    start.line = 1;

As `start` is a reference to the function arguments, this will be used external to the call.

alfiedotwtf:
Just thinking, maybe a few simple tests that do the happy path and also bounds checks might be worth it so we don't introduce unwanted behaviour?

forc-util/src/lib.rs
xunilrj: Which is correct, given that `start` and `end` use 1-based indexing.
An empty code would call

```
construct_window(
        &mut LineCol { line: 1, col: 1 },
        LineCol { line: 1, col: 1 },
        &mut start,
        &mut end,
        "",
    )
```

And this should return 

```
[forc-util/src/lib.rs:772:5] start = 0
[forc-util/src/lib.rs:772:5] end = 0
[forc-util/src/lib.rs:772:5] s = LineCol {
    line: 1,
    col: 1,
}
[forc-util/src/lib.rs:772:5] r = ""
```

I think you meant the inverse, right? I am including more tests, nonetheless.

xunilrj: In this case, I think master is incorrect. `start.line` uses 1-based index, and should never be zero. There is certainly a guard somewhere making `line = 0` work somehow.

xunilrj:
I have changed the algorithm to something easier to understand, more performant, given that we iterate the string only once, and I have also written more tests.

IGI-111 (approved):

JoshuaBatty (approved):

JoshuaBatty (approved):

forc-util/src/lib.rs
```diff
@@ -716,43 +716,104 @@ fn construct_window<'a>(
...
         }
-        pos += character.len_utf8();
+        // has first and no last
+        (Some((first_char_line, first_char_offset)), None) => {
+            (first_char_line, first_char_offset, input.len())
+        }
+        // others
+        _ => (current_line, input.len(), input.len()),
+    };
+
```
alfiedotwtf, line 755: I'm not sure about others, and although the code looks cleaner, I'm not sure we want to iterate over the whole string multiple times vs the old code that loops once but does multiple things during each iteration.

xunilrj: The current code iterate only once for everything.
The older code apart from iterating once to find the window, was also iterating two more times inside the "window", given that `chars` in `O(n)`.

```
*start_ix = input[calculated_start_ix..(calculated_start_ix + start_ix_bytes)]
        .chars()
        .count();
    *end_ix = input[calculated_start_ix..(calculated_start_ix + end_ix_bytes)]
        .chars()
        .count();
```

IGI-111 (approved):

PR #6511, opened 2024-09-07
fix array element type algorithm to avoid selecting never

xunilrj:
## Description

This PR fixes the array element type selection algorithm to avoid selecting `Never`.

## Checklist


test/src/e2e_vm_tests/test_programs/should_fail/array_wrong_elements_types/src/main.sw
```diff
@@ -0,0 +1,45 @@
...
+    let _ = [Vec::new(), vec::<u8>(), vec::<u16>()];
+    
+    // unexpected Option<u8>
+    let a = [None, Some(1), Some(1u8)];
+    let _b: Option<u16> = a[1];
+
+    // unexpected u8
+    let a = [8, 256u16, 8u8];
+    let b: u32 = a[2];
+}
```
jjcnn, line 45: I suggest adding a test with an empty array too:

```
let _a : [u8 ; 0] = [];
```

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -1863,44 +1862,32 @@ impl ty::TyExpression {
...
-                        }
-                        _ => Continue(Some(last)),
-                    }
-                }
-            })
-            .into_inner();
-        let elem_type = elem_type.unwrap_or_else(|| typed_contents[0].return_type);
+        // if the element type is still unknown, and all elements are Never,
+        // fallback to unit
+        let initial_type = if matches!(initial_type, TypeInfo::Unknown) {
```
jjcnn, line 1875: Is there a problem with using `Unknown` instead of `unit` as the result in this case?

If the type is still `Unknown` after unification, then we should throw an error because the type could not be determined. Otherwise it seems to me that using `Unknown` here solves the entire issue.

test/src/e2e_vm_tests/test_programs/should_fail/array_wrong_elements_types/src/main.sw
```diff
@@ -0,0 +1,45 @@
...
+    // unexpected u16
+    let _ = [1u8, return, 1u16];
+
+    // unexpected u16
+    let _ = [1, return, 1u8, 1u16];
+
+    // unexpected str
+    let _ = [1, "", 1u16];
+    let a = [1, 2, "hello"];
+    let _ = [1, return, "", 1u16];
```
jjcnn, line 33: On principle I would also add

```
let _ = [1, "", return, 1u16];
```

It ought to do the same as the case in line 19, but we treat `Never` as a special case, so it might be worth it just in case.

jjcnn (changes requested):

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
xunilrj: But giving an error in this case does not seem wrong to me. We just need to decide which we prefer.

test/src/e2e_vm_tests/test_programs/should_fail/array_wrong_elements_types/src/main.sw
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
IGI-111: I don't think it matters much so long as we guarantee that the unannotated type of `[]` is `[();0]`.

I suppose in the broadest sense, the programmer's intent with a program like this is unambiguous and so it's not really erroneous. It's just weird, but it's fully deterministic and since the elements aren't used at all, any zero sized type will do.

xunilrj: I think that in this case ("arrays without typing context and all elements are never") we have four options:
1 - infer the array element as never; so something like `[!;1]`;
2 - infer the whole array as never;
3 - decay into a zero-sized type, in this case, `unit`;
4 - give an error.

`1` and `2` create this complicated case of how to deal with  "generic types with never". 
I think `4` is totally reasonable in this case.
But I implemented `3` just because it is what Rust does.

IGI-111 (approved):

PR #6514, opened 2024-09-09
Add Caller addresses function

SwayStar123:
## Description
Adds a new `caller_addresses` function to return a list of all owners of input coins and messages to a transaction

## Checklist


sway-lib-std/src/auth.sw
```diff
@@ -192,6 +200,60 @@ pub fn caller_address() -> Result<Address, AuthError> {
...
+///     assert(addresses.get(0).unwrap() == some_address);
+/// }
+/// ```
+pub fn caller_addresses() -> Vec<Address> {
+    let inputs = input_count().as_u64();
+    let mut addresses = Vec::new();
+    let mut iter = 0;
+
+    while iter < inputs {
+        let type_of_input = input_type(iter);
```
ironcev, line 231: Why not simple `input_type` as a variable name?

sway-lib-std/src/auth.sw
```diff
@@ -192,6 +200,60 @@ pub fn caller_address() -> Result<Address, AuthError> {
...
+        match type_of_input {
+            Some(Input::Coin) | Some(Input::Message) => {
+                // If input type is Coin or Message, get the owner address.
+                let owner_of_input = match type_of_input {
```
ironcev, line 235: Same here, `input_owner`.

sway-lib-std/src/auth.sw
```diff
@@ -192,6 +200,60 @@ pub fn caller_address() -> Result<Address, AuthError> {
...
+/// }
+/// ```
+pub fn caller_addresses() -> Vec<Address> {
+    let inputs = input_count().as_u64();
+    let mut addresses = Vec::new();
+    let mut iter = 0;
+
+    while iter < inputs {
+        let type_of_input = input_type(iter);
+        match type_of_input {
```
ironcev, line 232: Why the outer match? Shouldn't it work exactly the same if we have only the inner match :thinking:
```
let input_owner = match input_type {
   ...
}

if let Some(address) = input_owner {
    addresses.push(address);
}

iter +=1;
```

sway-lib-std/src/auth.sw
```diff
@@ -192,6 +200,60 @@ pub fn caller_address() -> Result<Address, AuthError> {
...
+            },
+            _ => {
+            // Input type is neither Coin nor Message, continue looping.
+}
+        }
+
+        iter += 1;
+    }
+
+    addresses // Return the collected addresses.
```
ironcev, line 254: I find the comment not being necessary. It somehow states the obvious and the proposal is to remove it.
```suggestion
    addresses
```

ironcev (changes requested):

sway-lib-std/src/auth.sw
```diff
@@ -192,6 +200,60 @@ pub fn caller_address() -> Result<Address, AuthError> {
...
+/// ```sway
+/// use std::auth::caller_addresses;
+///
+/// fn foo(some_address: Address) {
+///     let addresses = caller_addresses();
+///
+///     assert(addresses.get(0).unwrap() == some_address);
+/// }
+/// ```
+pub fn caller_addresses() -> Vec<Address> {
```
bitzoic, line 225: Shouldn't this return a `Result` like `caller_address()`?

I am also unsure how I feel about `caller_addresses()` as it's so close and makes it hard to differentiate between `caller_address()`.

test/src/sdk-harness/test_projects/auth/mod.rs
```diff
@@ -135,6 +135,446 @@ async fn input_message_msg_sender_from_contract() {
...
+    let auth_instance = AuthContract::new(id_1.clone(), wallet4.clone());
+    
+    let result = auth_instance
+        .methods()
+        .returns_caller_addresses()
+        .call()
+        .await
+        .unwrap();
+    
+    dbg!(&result.value);
```
bitzoic, line 363: This should be removed

test/src/sdk-harness/test_projects/auth/mod.rs
```diff
@@ -135,6 +135,446 @@ async fn input_message_msg_sender_from_contract() {
...
```
bitzoic, line 220: Needs to be removed

test/src/sdk-harness/test_projects/auth/mod.rs
```diff
@@ -135,6 +135,446 @@ async fn input_message_msg_sender_from_contract() {
...
+    
+    let provider = wallet1.provider().unwrap();
+    let tx = tb.build(provider.clone()).await.unwrap();
+
+    // Send and verify
+    let tx_id = provider.send_transaction(tx).await.unwrap();
+    let tx_status = provider.tx_status(&tx_id).await.unwrap();
+    let result = call_handler.get_response_from(tx_status).unwrap();
+
+    dbg!(&result.value);
```
bitzoic, line 424: Remove

test/src/sdk-harness/test_projects/auth/mod.rs
```diff
@@ -135,6 +135,446 @@ async fn input_message_msg_sender_from_contract() {
...
```
bitzoic, line 571: Remove

sway-lib-std/src/auth.sw
SwayStar123: There is always atleast a single sender, so there is no need for a Result.

caller_addresses has a Result because it only returns a value if all the senders are the same, no such problem here.

I dont think the name being similar is a real problem as the return types are completely different, accidentally using one for another is unlikely

ironcev: Also IMO the similarity of the names shouldn't be an issue.

ironcev (approved):

xunilrj (approved):

sway-lib-std/src/auth.sw
IGI-111: It's more than one character of difference and a different type, I think confusion is unlikely.

I'm more concerned that `caller_address` doesn't have a name that conveys that it's only returning a unique one like `single_caller` and `all_callers` or something.

SwayStar123:
Sorry, there was a merge conflict

bitzoic (approved):

ironcev (approved):

PR #6516, opened 2024-09-09
Implement hierarchical trait map and namespace lookups.

tritao:
## Description

This PR implements a hierarchical-based trait map and namespace lookup system while type checking.

Previously we added the concept of lexical concepts, which store all the names in a tree-based hierarchy.
But currently we still rely on a single "global" namespace, which is maintained by the type check context as we go down the tree. Thus, so far, all trait and namespace lookup have relied on looking up in that single namespace.

The main idea here is to be able to just rely on the names existing in the lexical scopes, and walk up those lexical scope chains as we need to lookup any name or trait.

The logic is split into these two commits:

[Implement hierarchical trait map lookups.](https://github.com/FuelLabs/sway/pull/6516/commits/4043fb5bcb5157971e03752de57ea86fa8fc03b6)

[Implement hierarchical namespace lookups.](https://github.com/FuelLabs/sway/pull/6516/commits/486df45623eb80f186c41451159c2959277bd0a2)

This PR still does not remove that cloning, which will be done in a separate future PR.

## Checklist

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
+++ b/sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
@@ -2416,31 +2416,26 @@ impl ty::TyExpression {
         base_name: &Ident,
         projections: &[ty::ProjectionKind],
     ) -> Result<(TypeId, TypeId), ErrorEmitted> {
-        let mut lexical_scope_opt = Some(module.current_lexical_scope());
-        while let Some(lexical_scope) = lexical_scope_opt {
-            let result = Self::find_subfield_type_helper(
+        let ret = module.walk_scope_chain(|lexical_scope| {
+            Self::find_subfield_type_helper(
                 lexical_scope,
                 handler,
                 engines,
                 namespace,
                 base_name,
                 projections,
-            )?;
-            if let Some(result) = result {
-                return Ok(result);
-            }
-            if let Some(parent_scope_id) = lexical_scope.parent {
-                lexical_scope_opt = module.get_lexical_scope(parent_scope_id);
-            } else {
-                lexical_scope_opt = None;
-            }
-        }
+            )
+        })?;
 
-        // Symbol not found
-        Err(handler.emit_err(CompileError::UnknownVariable {
-            var_name: base_name.clone(),
-            span: base_name.span(),
-        }))
+        if let Some(ret) = ret {
+            Ok(ret)
+        } else {
+            // Symbol not found
+            Err(handler.emit_err(CompileError::UnknownVariable {
+                var_name: base_name.clone(),
+                span: base_name.span(),
+            }))
+        }
     }
 
     /// Returns a tuple where the first element is the [TypeId] of the actual expression, and
--- a/sway-core/src/semantic_analysis/namespace/module.rs
+++ b/sway-core/src/semantic_analysis/namespace/module.rs
@@ -299,26 +299,40 @@ impl Module {
         self.current_lexical_scope_id = parent_scope_id.unwrap_or(0);
     }
 
-    pub fn get_items_for_type(
+    pub fn walk_scope_chain<T>(
         &self,
-        engines: &Engines,
-        type_id: TypeId,
-    ) -> Vec<ResolvedTraitImplItem> {
+        mut f: impl FnMut(&LexicalScope) -> Result<Option<T>, ErrorEmitted>,
+    ) -> Result<Option<T>, ErrorEmitted> {
         let mut lexical_scope_opt = Some(self.current_lexical_scope());
-        let mut vec = vec![];
         while let Some(lexical_scope) = lexical_scope_opt {
-            vec.extend(
-                lexical_scope
-                    .items
-                    .implemented_traits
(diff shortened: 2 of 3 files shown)


sway-core/src/semantic_analysis/namespace/module.rs
```diff
@@ -285,6 +298,75 @@ impl Module {
...
+
+    pub fn get_methods_for_type(
+        &self,
+        engines: &Engines,
+        type_id: TypeId,
+    ) -> Vec<ResolvedFunctionDecl> {
+        self.get_items_for_type(engines, type_id)
+            .into_iter()
+            .filter_map(|item| match item {
+                ResolvedTraitImplItem::Parsed(_) => todo!(),
```
sdankel, line 361: nit: I see that this was just moved from another file, but I think we should replace the `todo!` with something else, like `None`

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -2429,6 +2431,188 @@ impl ty::TyExpression {
...
+            )?;
+            if let Some(result) = result {
+                return Ok(result);
+            }
+            if let Some(parent_scope_id) = lexical_scope.parent {
+                lexical_scope_opt = module.get_lexical_scope(parent_scope_id);
+            } else {
+                lexical_scope_opt = None;
+            }
+        }
```
jjcnn, line 2460: The pattern 

```
let mut lexical_scope_opt = ...;
while let Some(lexical_scope) = lexical_scope_opt {
  ...
  if let Some(parent_scope_id) = lexical_scope.parent {
    lexical_scope_opt = module.get_lexical_scope(parent_scope_id);
  } else {
    lexical_scope_opt = None;
  }
}
```
is repeated in a couple of places, and seems to capture a structural aspect of hierarchical lexical scopes. Is there a way to abstract this pattern out (ideally so that it is located in `lexical_scope.rs`), so that searching through the scopes is kept separate from what we are searching for?

sway-core/src/semantic_analysis/namespace/module.rs
```diff
@@ -285,6 +298,75 @@ impl Module {
...
+                    .items
+                    .implemented_traits
+                    .get_items_for_type(engines, type_id),
+            );
+            if let Some(parent_scope_id) = lexical_scope.parent {
+                lexical_scope_opt = self.get_lexical_scope(parent_scope_id);
+            } else {
+                lexical_scope_opt = None;
+            }
+        }
```
jjcnn, line 321: The pattern occurs again here.

sway-core/src/semantic_analysis/namespace/module.rs
```diff
@@ -285,6 +298,75 @@ impl Module {
...
+                .resolve_symbol(handler, engines, symbol)?;
+            if let Some(result) = result {
+                return Ok(result);
+            }
+            if let Some(parent_scope_id) = lexical_scope.parent {
+                lexical_scope_opt = self.get_lexical_scope(parent_scope_id);
+            } else {
+                lexical_scope_opt = None;
+            }
+        }
```
jjcnn, line 344: Same pattern again.

sway-core/src/semantic_analysis/namespace/trait_map.rs
```diff
@@ -1352,42 +1354,67 @@ impl TraitMap {
...
+                lexical_scope
+                    .items
+                    .implemented_traits
+                    .get_implemented_traits(type_id, engines),
+            );
+            if let Some(parent_scope_id) = lexical_scope.parent {
+                lexical_scope_opt = module.get_lexical_scope(parent_scope_id);
+            } else {
+                lexical_scope_opt = None;
+            }
```
jjcnn, line 1406: Same pattern again.

jjcnn (changes requested):
Overall it looks good, but I think there's a common pattern that can be abstracted out, and there's a nagging feeling at the back of my mind that you've introduced a shadowing issue. Let me know what you think - I might just be overthinking things.

sway-core/src/semantic_analysis/namespace/trait_map.rs
```diff
@@ -1352,42 +1354,67 @@ impl TraitMap {
...
+    ) -> BTreeSet<(Ident, TypeId)> {
+        let mut lexical_scope_opt = Some(module.current_lexical_scope());
+        let mut all_impld_traits: BTreeSet<(Ident, TypeId)> = Default::default();
+        while let Some(lexical_scope) = lexical_scope_opt {
+            all_impld_traits.extend(
+                lexical_scope
+                    .items
+                    .implemented_traits
+                    .get_implemented_traits(type_id, engines),
+            );
```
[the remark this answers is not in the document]
jjcnn: I suppose my question really is: Is there a case where this algorithm will throw an ambiguity error, but where shadowing actually resolves the ambiguity? I'm not sure, because I'm not sure a trait implementation can shadow another trait implementation.

jjcnn (dismissed):

jjcnn (approved):

sdankel (approved):

PR #6522, opened 2024-09-10
Add configurables section offset in the preamble

vaivaswatha:
The preamble now contains 8 bytes of offset to the configurables section. If there is no configurable const in the data-section, then the value of this integer will be equal to the size of the binary itself.

This also means that we now sort the data-section to have all the non-configurables first, and then the configurables.

The preamble final asm looks like this (the offset to configurables is 0'd out here in the example):
```
;; ASM: Final program
;; Program kind: Script
.program:
move $$tmp $pc
jmpf $zero i10
DATA_SECTION_OFFSET[0..32]
DATA_SECTION_OFFSET[32..64]
CONFIGURABLES_OFFSET[0..32]
CONFIGURABLES_OFFSET[32..64]
lw   $$ds $$tmp i1
add  $$ds $$ds $$tmp
```
The preamble bytecode looks like this:

```
0x00000000 MOVE R60 $pc                                    ;; [26, 240, 48, 0]
0x00000004 JMPF $zero 0xa                                  ;; [116, 0, 0, 10]
0x00000008                                                 ;; [0, 0, 0, 0, 0, 0, 2, 40]
0x00000010                                                 ;; [0, 0, 0, 0, 0, 0, 0, 0]
0x00000030 LW R63 R60 0x1                                  ;; [93, 255, 192, 1]
0x00000034 ADD R63 R63 R60                                 ;; [16, 255, 255, 0]
...
```


forc-test/src/execute.rs
```diff
@@ -246,18 +246,27 @@ impl TestExecutor {
...
 ///
 /// ```ignore
-/// [0] ji   i4                       ; Jumps to the data section setup.
-/// [1] noop
-/// [2] DATA_SECTION_OFFSET[0..32]
-/// [3] DATA_SECTION_OFFSET[32..64]
-/// [4] lw   $ds $is 1                ; The data section setup, i.e. where the first ji lands.
-/// [5] add  $$ds $$ds $is
-/// [6] <first-entry-point>           ; This is where we want to jump from to our test code!
+/// [ 0] ji   i4                       ; Jumps to the data section setup.
```
xunilrj, line 249: This is actually `ji i(4 + 16) (data section offset + metadata)` right?

sway-core/src/asm_lang/allocated_ops.rs
```diff
@@ -751,9 +763,10 @@ impl AllocatedOp {
 fn addr_of(
     dest: &AllocatedRegister,
     data_id: &DataId,
-    data_section: &mut DataSection,
+    data_section: &DataSection,
 ) -> Vec<fuel_asm::Instruction> {
     let offset_bytes = data_section.data_id_to_offset(data_id) as u64;
+    dbg!(data_id, offset_bytes);
```
xunilrj, line 769: dbg leftover.

sdankel:
This looks good! I'm wondering if it would make sense to move the existing data section offset inside of the metadata as well since conceptually it is also metadata.

kayagokalp:
> This looks good! I'm wondering if it would make sense to move the existing data section offset inside of the metadata as well since conceptually it is also metadata.

I think this would be a big breaking change, SDK and tooling currently reads this data offset while trying to:

1. Strip data section for predicate and script deployments (sdk)
2. Create shifted abi after the predicate and script is deployed (tooling)

So if we end up moving the location of the `DATA_SECTION_OFFSET[0..32]` `DATA_SECTION_OFFSET[32..64]` it may be wiser to do so after we cut a release of sway repo supporting fuel-core v0.39.0. Which is blocked waiting a SDK release at the moment.

IGI-111:
Two things here:

1. Changing the convention for the data section offset is entirely out of scope, and frankly not even something I think we should consider. It breaks too many things and there's really no benefit to having it in a slightly different place in the binary that's still conventional.
2. We only really need the option to add more metadata in the future, by default this should be empty aside from the configurables offset

vaivaswatha:
@IGI-111 I've renamed "Metadata" to "Configurable Offset" and it's just 64b now, instead of 256b.

IGI-111 (dismissed):

forc-test/src/execute.rs
```diff
@@ -246,18 +246,21 @@ impl TestExecutor {
...
-/// [3] DATA_SECTION_OFFSET[32..64]
-/// [4] lw   $ds $is 1                ; The data section setup, i.e. where the first ji lands.
-/// [5] add  $$ds $$ds $is
-/// [6] <first-entry-point>           ; This is where we want to jump from to our test code!
+/// [ 0] ji   i(4 + 2)                 ; Jumps to the data section setup.
+/// [ 1] noop
+/// [ 2] DATA_SECTION_OFFSET[0..32]
+/// [ 3] DATA_SECTION_OFFSET[32..64]
+/// [ 4] METADATA (0-32)
+/// [ 5] METADATA (32-64)
```
sdankel, line 254: nit: consistency

```suggestion
/// [ 4] METADATA[0..32]
/// [ 5] METADATA[32..64]
```

forc/src/cli/commands/parse_bytecode.rs
```diff
@@ -59,6 +59,14 @@ pub(crate) fn exec(command: Command) -> ForcResult<()> {
                     parsed_raw
                 )
             }
+            Err(fuel_asm::InvalidOpcode) if word_ix == 4 || word_ix == 5 => {
+                let parsed_raw = u32::from_be_bytes([raw[0], raw[1], raw[2], raw[3]]);
+                format!(
+                    "configurables offset {} ({})",
```
sdankel, line 65: Should this be referencing "metadata"?

I'm a little confused as to whether metadata is anything other than the configurables offset. If it's only that, and only can be used for that, we should simplify the code by just calling it that everywhere instead of metadata.

sway-core/src/asm_lang/virtual_ops.rs
```diff
@@ -1743,6 +1750,7 @@ impl VirtualOp {
             /* Non-VM Instructions */
             BLOB(imm) => AllocatedOpcode::BLOB(imm.clone()),
             DataSectionOffsetPlaceholder => AllocatedOpcode::DataSectionOffsetPlaceholder,
+            Metadata => AllocatedOpcode::ConfigurablesOffsetPlaceholder,
```
sdankel, line 1753: Should this be
```suggestion
            ConfigurablesOffsetPlaceholder => AllocatedOpcode::ConfigurablesOffsetPlaceholder,
```

sway-core/src/asm_generation/fuel/programs/abstract.rs
```diff
@@ -195,12 +209,18 @@ impl AbstractProgram {
                     comment: "data section offset".into(),
                     owning_span: None,
                 },
+                // word 3 -- full word u64 placeholder
+                AllocatedAbstractOp {
+                    opcode: Either::Right(ControlFlowOp::ConfigurablesOffsetPlaceholder),
+                    comment: "metadata".into(),
```
sdankel, line 215:
```suggestion
                    comment: "configurables offset".into(),
```

sway-core/src/asm_lang/mod.rs
```diff
@@ -1248,6 +1248,7 @@ impl fmt::Display for VirtualOp {
             /* Non-VM Instructions */
             BLOB(a) => write!(fmtr, "blob {a}"),
             DataSectionOffsetPlaceholder => write!(fmtr, "data section offset placeholder"),
+            Metadata => write!(fmtr, "Metadata"),
```
sdankel, line 1251:
```suggestion
            ConfigurablesOffsetPlaceholder => write!(fmtr, "configurables offset placeholder"),
```

sway-core/src/asm_lang/virtual_ops.rs
```diff
@@ -226,6 +226,7 @@ pub(crate) enum VirtualOp {
 
     /* Non-VM Instructions */
     BLOB(VirtualImmediate24),
+    Metadata,
```
sdankel, line 229:
```suggestion
    ConfigurablesOffsetPlaceholder,
```

sway-core/src/asm_lang/virtual_ops.rs
```diff
@@ -347,6 +348,7 @@ impl VirtualOp {
             /* Non-VM Instructions */
             BLOB(_imm) => vec![],
             DataSectionOffsetPlaceholder => vec![],
+            Metadata => vec![],
```
sdankel, line 351:
```suggestion
            ConfigurablesOffsetPlaceholder => vec![],
```

sway-core/src/asm_lang/virtual_ops.rs
```diff
@@ -465,6 +467,7 @@ impl VirtualOp {
             // Virtual OPs
             | BLOB(_)
             | DataSectionOffsetPlaceholder
+            | Metadata
```
sdankel, line 470:
```suggestion
            | ConfigurablesOffsetPlaceholder
```

sway-core/src/asm_lang/virtual_ops.rs
```diff
@@ -571,6 +574,7 @@ impl VirtualOp {
             | GTF(_, _, _)
             | BLOB(_)
             | DataSectionOffsetPlaceholder
+            | Metadata
```
sdankel, line 577:
```suggestion
            | ConfigurablesOffsetPlaceholder
```

sdankel (commented):
From [this comment](https://github.com/FuelLabs/sway/pull/6522#issuecomment-2431342301) I think there are some things that need to be renamed from metadata to configurables offset. Let me know if I'm not understanding correctly

forc/src/cli/commands/parse_bytecode.rs
vaivaswatha: > Should this be referencing "metadata"?
> 
> I'm a little confused as to whether metadata is anything other than the configurables offset. If it's only that, and only can be used for that, we should simplify the code by just calling it that everywhere instead of metadata.

> we should simplify the code by just calling it that everywhere instead of metadata

Yes, that was my intention. What you've rightly pointed out in the review comments are places I missed making this changes. I'm about to fix that. Thank you.

sway-core/src/asm_generation/fuel/programs/abstract.rs
```diff
@@ -195,12 +209,18 @@ impl AbstractProgram {
...
+                // word 3 -- full word u64 placeholder
+                AllocatedAbstractOp {
+                    opcode: Either::Right(ControlFlowOp::ConfigurablesOffsetPlaceholder),
+                    comment: "configurables offset".into(),
+                    owning_span: None,
+                },
                 AllocatedAbstractOp {
                     opcode: Either::Right(ControlFlowOp::Label(label)),
-                    comment: "end of metadata".into(),
+                    comment: "end of configurables offset".into(),
```
sdankel, line 220: It looks like this used to signal the end of the data section. Maybe this should still be `end of metadata` since now it signifies the end of both?

sway-core/src/asm_lang/allocated_ops.rs
```diff
@@ -525,6 +526,7 @@ impl fmt::Display for AllocatedOpcode {
 
             /* Non-VM Instructions */
             BLOB(a) => write!(fmtr, "blob {a}"),
+            ConfigurablesOffsetPlaceholder => write!(fmtr, "Metadata"),
```
sdankel, line 529: Should the placeholder look more like DataSectionOffsetPlaceholder below?

sway-core/src/asm_generation/fuel/programs/abstract.rs
IGI-111: We may want to add more metadata later, let's not write something that may be invalidated later.

IGI-111 (dismissed):

sway-core/src/asm_generation/fuel/programs/abstract.rs
vaivaswatha: This is just a comment indicating the end of where the current offset section ends. It doesn't invalidate anything we may do later.

IGI-111 (approved):

sdankel (dismissed):

IGI-111 (approved):

sdankel (approved):

PR #6535, opened 2024-09-12
feat: add a build script to forc-client to keep proxy abi in sync with file

### LINKED ISSUE #6527: Create a build.rs script for forc-client to auto update proxy abi from file to source code
We have an interesting problem, `abigen!` macro from sdk needs either a file location, or a raw abi string. Using file location fails becuase cargo flattens all packages before releasing them and changing the location we hold our abi file in. Using raw str is troublesome and error prone. We discussed using `include_str!` macro from std but then the problem is the order of macro expansions. We cannot ensure that `nclude_str!` is expanded before `abigen!`.

We can create a `build.rs` file that updates the contents of the proxy abi raw str embedded in the code to come to a middle ground, where we use raw str for abigen macro but do not need to update it manually.

kayagokalp:
## Description
closes #6527


JoshuaBatty (approved):
Would be great to come up with a nicer solution for #6527 at some point but this is fine for now

forc-plugins/forc-client/build.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    // Prepare the replacement string for the `abigen!` macro
+    let escaped_json = minified_json.replace('\\', "\\\\").replace('"', "\\\"");
+    let new_abigen = format!(
+        "abigen!(Contract(name = \"ProxyContract\", abi = \"{}\",));",
+        escaped_json
+    );
+
+    // Use a regular expression to find and replace the `abigen!` macro
+    let re = regex::Regex::new(r#"abigen!\(Contract\(name = "ProxyContract", abi = ".*?",\)\);"#)
```
alfiedotwtf, line 57: Since that's a non-greedy capture, it might need to be anchored to the end of the string, otherwise `",\)\);` appearing inside the abi source will end the match early:

    let re = regex::Regex::new(r#"abigen!\(Contract\(name = "ProxyContract", abi = ".*?",\)\);$"#)

alfiedotwtf (approved):

forc-plugins/forc-client/src/util/tx.rs
```diff
@@ -262,734 +262,7 @@ pub async fn update_proxy_contract_target(
...
...1`\"]},{\"name\":\"doc-comment\",\"arguments\":[\" * Writes: `1`\"]},{\"name\":\"storage\",\"arguments\":[\"write\"]}]}],\"loggedTypes\":[{\"logId\":\"4571204900286667806\",\"concreteTypeId\":\"3f702ea3351c9c1ece2b84048006c8034a24cbc2bad2e740d0412b4172951d3d\"},{\"logId\":\"2151606668983994881\",\"concreteTypeId\":\"1ddc0adda1270a016c08ffd614f29f599b4725407c8954c8b960bdf651a9a6c8\"},{\"logId\":\"2161305517876418151\",\"concreteTypeId\":\"1dfe7feadc1d9667a4351761230f948744068a090fe91b1bc6763a90ed5d3893\"},{\"logId\":\"4354576968059844266\",\"concreteTypeId\":\"3c6e90ae504df6aad8b34a93ba77dc62623e00b777eecacfa034a8ac6e890c74\"},{\"logId\":\"10870989709723147660\",\"concreteTypeId\":\"96dd838b44f99d8ccae2a7948137ab6256c48ca4abc6168abc880de07fba7247\"},{\"logId\":\"10098701174489624218\",\"concreteTypeId\":\"8c25cb3686462e9a86d2883c5688a22fe738b0bbc85f458d2d2b5f3f667c6d5a\"}],\"messagesTypes\":[],\"configurables\":[{\"name\":\"INITIAL_TARGET\",\"concreteTypeId\":\"0d79387ad3bacdc3b7aad9da3a96f4ce60d9a1b6002df254069ad95a3931d5c8\",\"offset\":13368},{\"name\":\"INITIAL_OWNER\",\"concreteTypeId\":\"192bc7098e2fe60635a9918afb563e4e5419d386da2bdbf0d716b4bc8549802c\",\"offset\":13320}]}",));
```
sdankel, line 265: Maybe we should keep it un-minified so it's easier to review the changes?

kayagokalp: Hmm, just to make sure we are on same page, the build script we added now will overwrite any value we put into this slot (by reading it from the file). So we won't be ever reading this line, as it will be taken from the file and put into this slot. (it get's overwritten before building the project). I did try to minify this because since the file is the only source of truth now and it is formatted, we won't ever need to read inside the source code also making source code readable adds 700 lines of mess in the file 😢 

I'll add a comment to mention that the content will be overwritten automatically, if you still think it is better to have it readable I can also do that at the same time :)

sdankel: A comment would be good for sure! I thought we would still see this get updated on PRs. If that's not the case, then a PR comment is fine. And in that case it could also be an empty string, right?

kayagokalp: Yep it could be an empty string but then it will show as a diff for different people building the workspace, as after building it the script won't be deleting what it already added

sdankel: Ahh that makes sense.

sdankel (approved):

forc-plugins/forc-client/build.rs
kayagokalp: That is a good catch! Only valid way `",\)\);` can appear in abi is probably as a doc comment as in any other place that would be non-valid abi. And luckly if that happens this will be a compile error, as the abi will not be valid (validity of the abi is checked during comp time right now).

I tried to anchor to the end but failed to detect the abi that way, probably my regex wizardy is not enough for this 😄 Also the version you gave cannot match the abi as well. So I think this is something we can omit for now at is _nearly_ impossible that this will be creating a problem  for us and given that this is a very short term solution,  So I created #6547

alfiedotwtf: Sorry, just saw this message now.

Ah nice, if it's impossible to slip through, then this is probably good enough (though let's hope nothing changes in the future where comments are allowable for debugging etc). Though yes let's have #6547 for future thinking, because the first thing I thought when seeing the regex was "SQL injection"-like attacks - and if I'm thinking that I'm sure automated offensive code scanners are looking for that too 🫣

PR #6541, opened 2024-09-12
Ensure the return path analysis traverses the entire CFG

jjcnn:
## Description

The return path analysis ensures that if a method must return a value then every path through a method returns a value. The analysis traverses the CFG of each method in a breadth-first manner from the (unique) entry point to the (unique) exit point. If there is ever a node in the graph that does not have outgoing edges, then an error is thrown.

So far the condition in the outer loop has been `!rovers.is_empty() && rovers[0] != exit_point`, where `rovers` is the current set of nodes to be analyzed. The requirement `rovers[0] != exit_point` potentially causes part of the CFG to not be traversed, but appears to have been added because of cases like the following:

```
fn main() -> u64 {
    impl core::ops::Eq for X {
            asm(r1: self, r2: other, r3) {
                eq r3 r2 r1;
                r3: bool
            }
        }
    }
    if X::Y(true) == X::Y(true) {
        a
    } else {
        a
    }
}
```

The CFG for `main` contains the CFG for `eq` as a subgraph, but since the exit point for `eq` does not have a path leading to the exit point of `main` the analysis throws an error when the condition is removed. (It is not clear to me why this is, but based on my testing it happens consistently).

This is clearly incorrect behavior since the edge leading into the `eq` CFG does not represent control flow.

To solve this problem I have introduced a condition in the inner loop in which a node is ignored if it has been visited before, or if it represents a method declaration that is different from the entry point of the method currently under analysis.

Additionally I have changed the representation of `visited` to be a `HashSet` instead of a vector. This should improve performance slightly.

I have not been able to find a test program that causes the original algorithm to allow illegal programs to pass, but the new algorithm seems more obviously correct.

## Checklist

### DIFF
--- a/sway-core/src/control_flow_analysis/analyze_return_paths.rs
+++ b/sway-core/src/control_flow_analysis/analyze_return_paths.rs
@@ -83,7 +83,7 @@ impl<'cfg> ControlFlowGraph<'cfg> {
         let mut rovers = vec![entry_point];
         let mut visited = vec![];
         let mut errors = vec![];
-        while !rovers.is_empty() && rovers[0] != exit_point {
+        while !rovers.is_empty() {
             rovers.retain(|idx| *idx != exit_point);
             let mut next_rovers = vec![];
             let mut last_discovered_span;
@@ -122,7 +122,13 @@ impl<'cfg> ControlFlowGraph<'cfg> {
                 }
                 next_rovers.append(&mut neighbors);
             }
-            next_rovers.retain(|idx| !visited.contains(idx));
+            next_rovers.retain(|idx| {
+                !visited.contains(idx)
+                    && !matches!(
+                        self.graph[*idx],
+                        ControlFlowGraphNode::MethodDeclaration { .. }
+                    )
+            });
             rovers = next_rovers;
         }
 
--- a/sway-core/src/control_flow_analysis/analyze_return_paths.rs
+++ b/sway-core/src/control_flow_analysis/analyze_return_paths.rs
@@ -10,6 +10,7 @@ use crate::{
     type_system::*,
     Engines,
 };
+use im::hashset::HashSet;
 use petgraph::prelude::NodeIndex;
 use sway_error::error::CompileError;
 use sway_types::{ident::Ident, span::Span, IdentUnique};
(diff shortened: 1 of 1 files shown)


IGI-111 (commented):
Can we test this to validate some previously unhandled state at least?

jjcnn:
> Can we test this to validate some previously unhandled state at least?

I haven't been able to find an example where the old algorithm gave an incorrect result. I can keep looking if you want?

When I removed the `rovers[0] != exit_point` check I got an error in this test case (as well as in a number of others):

https://github.com/FuelLabs/sway/tree/master/test/src/e2e_vm_tests/test_programs/should_pass/language/enum_in_fn_decl

That's the one I used to debug my algorithm.

jjcnn:
@IGI-111 : I managed to fix the linearization problem, and the traversal algorithm is much cleaner now. Take a look at your convenience.

sway-core/src/control_flow_analysis/analyze_return_paths.rs
```diff
@@ -133,7 +135,7 @@ impl<'cfg> ControlFlowGraph<'cfg> {
 /// The resulting edges from connecting a node to the graph.
 enum NodeConnection {
     /// This represents a node that steps on to the next node.
-    NextStep(Vec<NodeIndex>),
+    NextStep(Option<NodeIndex>),
```
IGI-111, line 138: I'm not quite sure I understand from the code why this needs to become an Option, can you explain that?

jjcnn: Two cases where we need it:

- Error nodes:
```
190:        ty::TyAstNodeContent::Error(_, _) => Ok(NodeConnection::NextStep(None)),
```

Looking at the code I notice we're not being entirely consistent wrt. how we handle error nodes - in some places we return `NextStep(None)` and in other cases we return `NextStep(leaf_opt)`.

- Declaration and side effect nodes:

```
186:       ty::TyAstNodeContent::SideEffect(_) => Ok(NodeConnection::NextStep(leaf_opt)),
```
(`leaf_opt` is initially `None`, and won't be `Some` until we've seen the first node)

This basically means that declaration and side effect nodes are skipped when constructing the linearized graph.

I can introduce a third variant to the `NodeConnection` enum, representing a node that we're skipping, if you prefer?

IGI-111: I don't think that'll be necessary. This isn't exactly clean code but at least those reasons make sense.

IGI-111 (approved):

JoshuaBatty (approved):

PR #6555, opened 2024-09-17
Fix `function_cache` garbage collection bug 

### LINKED ISSUE #5260: Optimize how ofter we call garbage collection in the language server
We are currently calling this method on every keystroke. We should try and find another metric that calls this less often.

JoshuaBatty:
## Description
Fixes a bug that would crash the language server when a function node was part of the AST.

#5967 added a function cache to the `QueryEngine`. Garbage collection needed to be called on this otherwise it returns references to types that have been cleared from the other engines during garbage collection. 

In the future, any other nodes that we decide to cache need to have GC applied to them or we will end up crashing the language server on the first key-pressed event. 

I've also removed the gc_frequency setting and now run GC on each keystroke. Without doing this, the spans stored in the TokenMap and what was returned from the compiler were falling out of sync. This ensures that everything is always up to date and the correct spans are used. closes #5260 

Finally, I've added a `launch.json` that allows for attaching `lldb` to a live running `forc-lsp` process. This was a pretty cruical step for tracking down this bug so would be nice to have it easily accessible for future debugging sessions. 

## Checklist

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,23 @@
+// Configuration for debugging the Sway Language Server (forc-lsp)
+// Usage instructions:
+// 1. Ensure you've built forc-lsp with debug symbols:
+//    cargo build -p forc-lsp
+// 2. Install the debug version:
+//    cargo install --path ./forc-plugins/forc-lsp --debug
+// 3. Open your Sway project in a separate VSCode window (this starts forc-lsp)
+// 4. In the forc-lsp project window, set breakpoints in the code
+// 5. Go to Run and Debug view, select "Attach to forc-lsp", and start debugging
+// 6. When prompted, select the forc-lsp process
+// 7. Debug forc-lsp as it responds to actions in your Sway project
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "lldb",
+            "request": "attach",
+            "name": "Attach to forc-lsp",
+            "pid": "${command:pickProcess}",
+            "program": "${env:HOME}/.cargo/bin/forc-lsp"
+        }
+    ]
+}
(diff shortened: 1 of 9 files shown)


alfiedotwtf (approved):

alfiedotwtf (approved):

.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
...
+// 1. Ensure you've built forc-lsp with debug symbols:
+//    cargo build -p forc-lsp
+// 2. Install the debug version:
+//    cargo install --path ./forc-plugins/forc-lsp --debug
+// 3. Open your Sway project in a separate VSCode window (this starts forc-lsp)
+// 4. In the forc-lsp project window, set breakpoints in the code
+// 5. Go to Run and Debug view, select "Attach to forc-lsp", and start debugging
+// 6. When prompted, select the forc-lsp process
+// 7. Debug forc-lsp as it responds to actions in your Sway project
+{
```
mchristopher, line 12: JSON comments are something that VS code specifically supports, but I'm unclear if the comments need to be inside the first `{` or not. If we've tested this and it works, all good, but may need to move these inside the first bracket.

mchristopher (approved):
Small thing to confirm, otherwise looks great

JoshuaBatty:
Thanks, i'm going to merge this PR and then follow up with the suggestions for the `launch.json` file in a follow up PR.

.vscode/launch.json
JoshuaBatty: Thanks i've opened #6557 to look into this in a follow up PR

.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
...
+    "version": "0.2.0",
+    "configurations": [
```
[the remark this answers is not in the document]
JoshuaBatty: Thanks i've opened #6557 to look into this in a follow up PR

kayagokalp (approved):

PR #6565, opened 2024-09-18
Profiling feature

GeorgiosDelkos:
## Description

This pull request adds the profile feature in the Sway compiler.

It communicates with an external `forc-perf` executable in order to signal the beginning and end of different compilation phases for profiling purposes.

## Checklist

### DIFF
--- a/forc-pkg/Cargo.toml
+++ b/forc-pkg/Cargo.toml
@@ -41,9 +41,13 @@ tracing = "0.1"
 url = { version = "2.2", features = ["serde"] }
 vec1 = "1.8.0"
 walkdir = "2"
+fuel-asm = { version = "0.57.0", features = ["serde"] }
+sway-ir = { version = "0.63.5", path = "../sway-ir" }
+sysinfo = { version = "0.31.4", optional = true }
 
 [dev-dependencies]
 regex = "^1.10.2"
 
-[target.'cfg(not(target_os = "macos"))'.dependencies]
-sysinfo = "0.29"
+[features]
+default = ["dep:sysinfo"]
+profile = []
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -1785,6 +1785,7 @@ pub fn compile(
 
     // First, compile to an AST. We'll update the namespace and check for JSON ABI output.
     let ast_res = time_expr!(
+        pkg.name,
         "compile to ast",
         "compile_to_ast",
         sway_core::compile_to_ast(
@@ -1823,6 +1824,7 @@ pub fn compile(
     }
 
     let asm_res = time_expr!(
+        pkg.name,
         "compile ast to asm",
         "compile_ast_to_asm",
         sway_core::ast_to_asm(&handler, engines, &programs, &sway_build_config),
@@ -1837,6 +1839,7 @@ pub fn compile(
     let mut program_abi = match pkg.target {
         BuildTarget::Fuel => {
             let program_abi_res = time_expr!(
+                pkg.name,
                 "generate JSON ABI program",
                 "generate_json_abi",
                 fuel_abi::generate_program_abi(
@@ -1875,6 +1878,7 @@ pub fn compile(
             };
 
             let abi = time_expr!(
+                pkg.name,
                 "generate JSON ABI program",
                 "generate_json_abi",
                 evm_abi::generate_abi_program(typed_program, engines),
@@ -1899,15 +1903,16 @@ pub fn compile(
         .map(|finalized_entry| PkgEntry::from_finalized_entry(finalized_entry, engines))
         .collect::<anyhow::Result<_>>()?;
 
-    let asm = match asm_res {
+    let mut asm = match asm_res {
         Err(_) => return fail(handler),
         Ok(asm) => asm,
     };
 
     let bc_res = time_expr!(
+        pkg.name,
         "compile asm to bytecode",
         "compile_asm_to_bytecode",
-        sway_core::asm_to_bytecode(&handler, asm, source_map, engines.se(), &sway_build_config),
+        sway_core::asm_to_bytecode(&handler, &mut asm, source_map, engines.se(), &sway_build_config),
         Some(sway_build_config.clone()),
         metrics
     );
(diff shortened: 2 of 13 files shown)


forc-pkg/Cargo.toml
```diff
@@ -41,9 +41,13 @@ tracing = "0.1"
...
 vec1 = "1.8.0"
 walkdir = "2"
+fuel-asm = { version = "0.57.0", features = ["serde"] }
+sway-ir = { version = "0.63.5", path = "../sway-ir" }
+sysinfo = { version = "0.31.4", optional = true }
 
 [dev-dependencies]
 regex = "^1.10.2"
 
-[target.'cfg(not(target_os = "macos"))'.dependencies]
```
JoshuaBatty, line 48: I'm not convinced we can include `sysinfo` when on mac. We had it like this due to `sysinfo` causing compatibility issues with `fuel.nix` https://github.com/FuelLabs/fuel.nix/issues/64.

I'm happy for it to be included as long as we can confirm the build issue that was occurring in that issue has been resolved.

JoshuaBatty (changes requested):

forc-pkg/src/pkg.rs
```diff
@@ -1957,6 +1962,54 @@ pub fn compile(
...
+        bytes.truncate(data_offset as usize);
+
+        let mut data_section_used = 0u64;
+
+        fn calculate_entry_size(entry: &Entry) -> u64 {
+            let padding = match &entry.padding {
+                sway_ir::Padding::Left { target_size } => target_size,
+                sway_ir::Padding::Right { target_size } => target_size,
+            };
+
```
alfiedotwtf, line 1982: To future-proof, `std::mem::size_of_val` can be used here instead of magic values

forc-pkg/Cargo.toml
alfiedotwtf: I'm thinking we could probably narrow this further since it's only Apple Intel by `target_arch`?

camden-smallwood: Thanks for pointing this out, it has been reverted to the previous implementation.

forc-pkg/src/pkg.rs
```diff
@@ -1983,9 +1983,10 @@ pub fn compile(
             };
 
             let value_size = match &entry.value {
-                sway_core::asm_generation::Datum::Byte(x) => std::mem::size_of_val(x),
-                sway_core::asm_generation::Datum::Word(x) => std::mem::size_of_val(x),
-                sway_core::asm_generation::Datum::ByteArray(byte_array) => byte_array.len() as u64,
+                sway_core::asm_generation::Datum::Byte(x)
+                | sway_core::asm_generation::Datum::Word(x)
+                | sway_core::asm_generation::Datum::ByteArray(x) => std::mem::size_of_val(x) as u64,
+
```
JoshuaBatty, line 1989: can we remove this empty new line? Sorry for my OCD here

forc-pkg/src/pkg.rs
```diff
@@ -1957,6 +1962,56 @@ pub fn compile(
         warnings,
         metrics,
     };
+
+    #[cfg(feature = "profile")]
+    {
```
JoshuaBatty, line 1967: Can we move the code between the { } braces into a function and then call it here?

forc-pkg/src/pkg.rs
camden-smallwood: Fixed.

forc-pkg/src/pkg.rs
alfiedotwtf: Nice, thanks.

Sorry just one last thing - seeing that `sum()` will panic on overflow, it might be worth replacing it with something safer:

```
   - sway_core::asm_generation::Datum::Collection(entries) => entries.iter().map(calculate_entry_size).sum(),
   + sway_core::asm_generation::Datum::Collection(entries) => entries.iter().try_fold(0u64, |acc, &v| acc.checked_add(u64::from(v)));
```

And then deal with a possible error for the `println!()` report.

camden-smallwood: We have taken a look at the data flow for the datum entries, and we have determined that it is the responsibility of the compiler to produce valid data section slot size values that sum within the range of a `u64`. We feel that if this value does overflow, there is an issue in the data section generation code, and a panic might actually be preferred. If the team feels that simply printing a warning would suffice, we can opt for this instead.

sway-core/src/asm_generation/mod.rs
```diff
@@ -10,7 +10,9 @@ pub mod fuel;
 pub mod instruction_set;
 
 mod finalized_asm;
-pub use finalized_asm::{CompiledBytecode, FinalizedAsm, FinalizedEntry};
+pub use finalized_asm::*;
+
```
sdankel, line 14: nit

```suggestion
```

forc-pkg/Cargo.toml
```diff
@@ -41,9 +41,14 @@ tracing = "0.1"
 url = { version = "2.2", features = ["serde"] }
 vec1 = "1.8.0"
 walkdir = "2"
+fuel-asm = { version = "0.57.0", features = ["serde"] }
+sway-ir = { version = "0.63.6", path = "../sway-ir" }
```
sdankel, line 45: Should these dependencies only be added for the profile feature?

forc-pkg/Cargo.toml
```diff
@@ -41,9 +41,14 @@ tracing = "0.1"
...
 
 [dev-dependencies]
 regex = "^1.10.2"
 
 [target.'cfg(not(target_os = "macos"))'.dependencies]
 sysinfo = "0.29"
+
+[features]
+profile = []
```
sdankel, line 54: nit: I'd prefer to call it `profiler`

forc-pkg/src/pkg.rs
```diff
@@ -1957,9 +1962,62 @@ pub fn compile(
...
+    let asm_information = sway_core::asm_generation::AsmInformation {
+        bytecode_size: bytes.len() as _,
+        data_section: sway_core::asm_generation::DataSectionInformation { 
+            size: data_section_size, 
+            used: data_section_used, 
+            value_pairs: compiled_asm.0.data_section.value_pairs.clone(),
+        }
+    };
+
+    println!("/forc-perf info {}", serde_json::to_string(&asm_information).unwrap());
```
sdankel, line 2018: > It communicates with an external forc-perf executable in order to signal the beginning and end of different compilation phases for profiling purposes.

So it communicates with forc-perf over stdout? Could you add to the PR description links to forc-perf documentation, and how to test this locally?

I'm wondering why `forc-perf` doesn't just invoke `forc_pkg::build` and have the performance information returned from the function? Or a separate wrapper function like `forc_pkg::build_with_profiler`

sdankel (changes requested):
Is the purpose of this profiler to profile sway code, or the rust code of the compiler itself? The name `forc perf` suggests the former, but it looks like this is attempting to do the latter.

 If it's only for ensuring the compiler doesn't regress in speed, we already have a free tool for that, codspeed. If that's the goal, we should just write benchmarks for more compiler functions using cargo bench and have the reporting (flamegraphs, etc) done through codspeed.

camden-smallwood:
> Is the purpose of this profiler to profile sway code, or the rust code of the compiler itself? The name `forc perf` suggests the former, but it looks like this is attempting to do the latter.
> 
> Sorry if I'm missing the point here, I don't have much context on this project. Is there something that this tool does that we don't get from criterion + codspeed?
> 
> I could see the value here being that we can run it against multiple sway projects, but the maintenance of maintaining this tooling does not seem worth it. We already have a [benchmark](https://codspeed.io/FuelLabs/sway/benchmarks/sway-lsp/benches/lsp_benchmarks/compile.rs::benches::benchmarks::compile) for the `compile` function which includes most if not all of the functions captured by `time_expr`. It would be good to add more benchmarks of the `compile` function over a diverse set of sway projects.

Please speak with Jibril and Nick about this, as we worked on this engagement in an external agreement with them.

You can find the forc-perf repository here: https://github.com/ourovoros-io/forc-perf

We opted to use an external profiling process so that the process-specific resource consumption is not affected by the resource consumption of the profiler itself. We added these changes in order to be able to delineate the individual phases.

forc-pkg/Cargo.toml
camden-smallwood: Thanks for pointing these out. The`fuel-asm` crate is no longer being used and has been removed. The `sway_ir` crate is indeed only being used for the profiling functionality, so we have made this an optional dependency that only loads when using the `profiler` feature.

IGI-111:
>Is the purpose of this profiler to profile sway code, or the rust code of the compiler itself? 

Both, or rather it's meant to help us track the performance of the compiler both in compilation time and runtime performance.

>Is there something that this tool does that we don't get from criterion + codspeed?

This has been in development for longer than we've added these, but the point is to get fine grained information about the compiler execution, including memory usage in various passes, statistics of instruction usage, etc.

Codspeed is nice but it doesn't really have the level of customization we're looking for for compiler benchmarks.

>I could see the value here being that we can run it against multiple sway projects

This is precisely the goal, part of the point of this is to set up a representative sample of Sway to benchmark on a standardized machine so we can track compiler performance over time.

camden-smallwood:
> @GeorgiosDelkos @camden-smallwood Still some issues with this, `fmt` check, a snapshot test failing and clippy.

These should be fixed now in the most recent commit.

camden-smallwood:
> > > @GeorgiosDelkos @camden-smallwood Still some issues with this, `fmt` check, a snapshot test failing and clippy.
> > 
> > 
> > These should be fixed now in the most recent commit.
> 
> Still seems to be failing on `---- should_pass/forc/help ----`.

I have attempted to remedy this again, but I must say that this particular test is extremely pedantic. All it's doing is checking the output of `forc build -h`, but the format that is in the snapshot is non-standard. If it fails again, I will have to request that someone from the Fuel team fix it instead, because the expected format does not match the output of `forc build -h` on my system.

camden-smallwood:
After looking at this further and speaking with Jibril, I have determined that there are 2 issues with this test:
1. The output from the help command is formatted based on the size of the current terminal.
2. The options are generated using a derive macro on a Rust ABI struct. This means the layout of the struct is not stable or deterministic, which can lead to different field ordering, and in turn different output text.

I believe the help test needs to be refactored to be more deterministic, because it's quite difficult to reproduce outside of the CI.

Issue #6560, opened 2024-09-18
forc-client should support signing transactions using a KMS


kayagokalp:
blocked by #6559 

We need to implement a KMS signer for forc-client, specifically we can start with AWS KMS solution. This will allow us sign transactions using through KMS rather than local keys for more "shared" and "secure" deployment workflows.

kayagokalp:
The rough design is the following:

1. Check cli options to see if the user wants to use a local signer or a KMS one (at this stage if the user provided an arn the signer is a aws kms right away)
2-a. If it is local one, we will guide them through all the same workflow
3-a. If it is aws KMS one, we will use the AWS KMS arn, and basically use the upcoming `AwsKmsSigner ` which implements `Signer` from rust-sdk, instead of `WalletUnlocked` which we use by default right now in forc-client

PR #6577, opened 2024-09-22
refactor: sign transacitons with ForcClientAccounts instead of raw sdk wallets

### LINKED ISSUE #6559: forc-client should support alternative signers
Currently the only signer type that forc-client supports is local signing keys. This can be provided in different forms:

1. manually by exposing the private key as a cli parameter
2. automatically by the integration with forc-wallet, via dialoguer prompts

That was fine until now but to support more complicated signers, we should be generic over the signer type. Luckily the sdk design is nice in the sense that they have a generic `Signer` trait that we can use instead of complete `WalletUnlocked` instances.

kayagokalp:
## Description
closes #6559 in preparation for #6560. Basically rather than using sdk wallet directly we are adding a level of indirection so that we can support different signers such as kms solutions that will come after this one.


forc-plugins/forc-client/src/op/run/mod.rs
```diff
@@ -116,31 +145,29 @@ pub async fn run_pkg(
...
-            command.default_signer,
-            command.signing_key,
-            wallet_mode,
-        )
+    let tx_policies = tx_policies_from_cmd(command);
+    let mut tb = call
+        .transaction_builder(tx_policies, VariableOutputPolicy::EstimateMinimum, &account)
         .await?;
 
+    account.add_witnesses(&mut tb)?;
```
alfiedotwtf, line 165: Will `add_witnesses()` need to be added to `ForcClientAccount`? Because AFAIC the default implementation is a NOP.

kayagokalp:
closing this in favor of #6578

Issue #5376, opened 2023-12-08
Flaky std test `can_get_input_message_sender`


kayagokalp:
The `tx_fields::inputs::success::message::can_get_input_message_sender` test [failed](https://github.com/FuelLabs/sway/actions/runs/7131028521/attempts/1?pr=5371) on the first run but after re-run it [succeeded](https://github.com/FuelLabs/sway/actions/runs/7131028521?pr=5371). 

reported by: @sdankel

sdankel:
This one is also flaky: `tx_fields::tx::can_get_inputs_count`

[failed](https://github.com/FuelLabs/sway/actions/runs/7202029803/job/19619365458?pr=5390) and then [passed](https://github.com/FuelLabs/sway/actions/runs/7202029803/job/19619672460?pr=5390) on rerun

sdankel:
@bitzoic Does your team own these tests? Would you mind taking a look?

sdankel:
Closing since this hasn't happened in a while. If it happens again we can try using nextest: https://github.com/FuelLabs/sway/pull/6521

kayagokalp:
https://github.com/FuelLabs/sway/actions/runs/10991028594/job/30512545184?pr=6578 looks like `tx_fields::tx::can_get_inputs_count` can still fail. I think i know why it happens, in couple of places it is probably the same reason as #6509

kayagokalp:
https://github.com/FuelLabs/sway/actions/runs/10991076805/job/30512687485?pr=6580 - this one as well

PR #6578, opened 2024-09-22
feat: AWS kms signer support for forc-client

### LINKED ISSUE #6559: forc-client should support alternative signers

### LINKED ISSUE #6560: forc-client should support signing transactions using a KMS

kayagokalp:
## Description
closes #6560.
closes #6559.

This PR adds AWS kms signer support. Which is enabled if `--aws-kms-signer <ARN>` is present. If that is the case, an aws client is created which is used to create a AwsSigner. 

With this PR, we are able to supply a signing arn and with that we can sign a transaction. This enables being able to revoke signing authority for aws kms managed keys. So we can control deployments and owner/target changes for a proxy contract. Effectively enabling multi-sig like proxy contracts via aws kms.


Cargo.toml
```diff
@@ -114,6 +114,8 @@ anyhow = "1.0"
 assert-json-diff = "2.0"
 async-trait = "0.1"
 atty = "0.2"
+aws-config = "1.5.6"
+aws-sdk-kms = "1.44"
```
alfiedotwtf, line 118: If this works for X.Y, we might want to drop the Z

kayagokalp: Oh, missed that one thanks!

kayagokalp: [23d2040](https://github.com/FuelLabs/sway/pull/6578/commits/23d2040861f2336b076bed001828850674da00e3)

kayagokalp:
Looks like we hit a flaky test (hopefully, as it was not failing in previous commit and now it is failing): https://github.com/FuelLabs/sway/issues/5376#issuecomment-2367636955

forc-plugins/forc-client/src/util/aws.rs
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    /// Sign a digest with this signer's key
+    pub async fn sign_message(
+        &self,
+        message: Message,
+    ) -> std::result::Result<fuel_crypto::Signature, anyhow::Error> {
+        self.sign_message_with_key(self.key_id.clone(), message)
+            .await
+    }
```
alfiedotwtf, line 214: If we're going to call `sign_mesage()` many times, it might be worth just calling `sign_with_kms()` directly to avoid that `clone()`.

alfiedotwtf: Not for this change though, it's fine

alfiedotwtf (dismissed):

forc-plugins/forc-client/src/util/aws.rs
kayagokalp: Yeah I am not really happy overall how this turned out tbh, once we are in clear, will come-back to this area. So I am adding this to never ending list of things to refactor :)

forc-plugins/forc-client/src/util/account.rs
```diff
@@ -0,0 +1,79 @@
...
+use fuels::{
+    prelude::*,
+    types::{coin_type_id::CoinTypeId, input::Input},
+};
+use fuels_accounts::{wallet::WalletUnlocked, Account};
+
+use super::aws::AwsSigner;
+
+#[derive(Clone, Debug)]
+pub enum ForcClientAccount {
```
JoshuaBatty, line 12: Can we add a doc comment for this enum?

JoshuaBatty (commented):
Thanks Kaya mostly looks good. Just wondering if you can first add doc comments to all public types and functions before we merge this.

Voxelot:
Is this entirely additive or does it include breaking behavior?

kayagokalp:
> Is this entirely additive or does it include breaking behavior?

This is entirely additive, `forc-run` changes just a refactor and does not change any behaviour.

forc-plugins/forc-client/src/util/account.rs
kayagokalp: Sure! did that in e9984f1d3311e5012bb304de393def7fe41c4365

forc-plugins/forc-client/src/util/account.rs
```diff
@@ -0,0 +1,85 @@
...
+    types::{coin_type_id::CoinTypeId, input::Input},
+};
+use fuels_accounts::{wallet::WalletUnlocked, Account};
+
+use super::aws::AwsSigner;
+
+#[derive(Clone, Debug)]
+/// Set of different signers available to be used with `forc-client` operations.
+pub enum ForcClientAccount {
+    /// Local signer, where the private key owned locally. This can be
```
kayagokalp, line 14:
```suggestion
    /// Local signer where the private key owned locally. This can be
```

forc-plugins/forc-client/src/util/account.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// generated through forc-wallet integration or manually by providing
```
kayagokalp, line 15:
```suggestion
    /// generated through `forc-wallet` integration or manually by providing
```

forc-plugins/forc-client/src/util/account.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// a private-key.
+    Wallet(WalletUnlocked),
+    /// A KMS Signer specifcally using AWS KMS service. The signing key is
+    /// is managed by another entity.
```
kayagokalp, line 19:
```suggestion
    /// is managed by another entity for KMS signers. Messages are
    /// signed by the KMS entity. Signed transactions are retrieved
    /// and submitted to the node by `forc-client`.
```

forc-plugins/forc-client/src/util/account.rs
```diff
@@ -0,0 +1,85 @@
...
+use super::aws::AwsSigner;
+
+#[derive(Clone, Debug)]
+/// Set of different signers available to be used with `forc-client` operations.
+pub enum ForcClientAccount {
+    /// Local signer, where the private key owned locally. This can be
+    /// generated through forc-wallet integration or manually by providing
+    /// a private-key.
+    Wallet(WalletUnlocked),
+    /// A KMS Signer specifcally using AWS KMS service. The signing key is
```
kayagokalp, line 18:
```suggestion
    /// A KMS Signer specifically using AWS KMS service. The signing key is
```

forc-plugins/forc-client/src/util/account.rs
```diff
@@ -9,8 +9,16 @@ use fuels_accounts::{wallet::WalletUnlocked, Account};
...
 
 #[derive(Clone, Debug)]
+/// Set of different signers available to be used with `forc-client` operations.
 pub enum ForcClientAccount {
+    /// Local signer where the private key owned locally. This can be
+    /// generated through `forc-wallet` integration or manually by providing
+    /// a private-key.
     Wallet(WalletUnlocked),
+    /// A KMS Signer specifically using AWS KMS service. The signing key is
+    /// is managed by another entity for KMS signers. Messages are
```
alfiedotwtf, line 19: typo: is is

forc-plugins/forc-client/src/util/aws.rs
```diff
@@ -0,0 +1,269 @@
...
+}
+
+impl AwsConfig {
+    /// Load configuration from environment variables.
+    /// For more details see: https://docs.rs/aws-config/latest/aws_config/
+    pub async fn from_env() -> Self {
+        let loader = aws_config::defaults(BehaviorVersion::latest())
+            .credentials_provider(DefaultCredentialsChain::builder().build().await);
+
+        let loader = match std::env::var("E2E_TEST_AWS_ENDPOINT") {
```
sdankel, line 31: Is this the right endpoint for production?

alfiedotwtf (dismissed):
Apart from "is is", it's good to go

forc-plugins/forc-client/src/util/aws.rs
```diff
@@ -0,0 +1,269 @@
...
+        .ok_or(anyhow::anyhow!("public key not found"))?;
+    Ok(raw.clone().into_inner())
+}
+
+async fn sign_with_kms(
+    client: &aws_sdk_kms::Client,
+    key_id: &str,
+    public_key_bytes: &[u8],
+    message: Message,
+) -> anyhow::Result<fuel_crypto::Signature> {
```
sdankel, line 124: I think it'd be cleaner to make these part of the AwsSigner impl

forc-plugins/forc-client/src/op/run/mod.rs
```diff
@@ -116,31 +144,28 @@ pub async fn run_pkg(
...
-
-    let provider = Provider::connect(node_url.clone()).await?;
-
-    let script_gas_limit = if compiled.bytecode.bytes.is_empty() {
-        0
-    } else if let Some(script_gas_limit) = command.gas.script_gas_limit {
-        script_gas_limit
-    // Dry run tx and get `gas_used`
-    } else {
-        get_script_gas_used(tb.clone().finalize_without_signature_inner(), &provider).await?
```
sdankel, line 131: Why is this no longer needed?

forc-plugins/forc-client/src/util/tx.rs
```diff
@@ -280,158 +265,11 @@ pub async fn update_proxy_contract_target(
...
-        self
-    }
-}
-
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::collections::BTreeMap;
-    use std::collections::HashMap;
+    use fuels::types::bech32::Bech32Address;
```
sdankel, line 271: Should we update the tests to remove Bech32 usage?

forc-plugins/forc-client/src/util/account.rs
```diff
@@ -0,0 +1,87 @@
...
+use super::aws::AwsSigner;
+
+#[derive(Clone, Debug)]
+/// Set of different signers available to be used with `forc-client` operations.
+pub enum ForcClientAccount {
+    /// Local signer where the private key owned locally. This can be
+    /// generated through `forc-wallet` integration or manually by providing
+    /// a private-key.
+    Wallet(WalletUnlocked),
+    /// A KMS Signer specifically using AWS KMS service. The signing key is
```
kayagokalp, line 18:
```suggestion
    /// A KMS Signer specifically using AWS KMS service. The signing key
```

forc-plugins/forc-client/src/util/aws.rs
```diff
@@ -0,0 +1,269 @@
...
+    };
+
+    // Insert the recovery id into the signature
+    debug_assert!(
+        !recovery_id.is_x_reduced(),
+        "reduced-x form coordinates are caught by the if-else chain above"
+    );
+    let v = recovery_id.is_y_odd() as u8;
+    let mut signature = <[u8; 64]>::from(sig.to_bytes());
+    signature[32] = (v << 7) | (signature[32] & 0x7f);
```
sdankel, line 179: What is this doing? It would be great to have more detailed comments here.

sdankel (approved):
utACK

forc-plugins/forc-client/src/op/run/mod.rs
kayagokalp: Because we no longer need to maintain our own signing logic for `WalletUnlocked` as the signer is abstracted as `ForcClientAccount` which implements a `Signer` and `Account` from SDK. The refactor that moves the Wallet into `ForcClientAccount` makes the custom signing logic obsolete. As a byproduct of that, we no longer need to maintain this custom signing logic for scripts. We can simply use SDK to do the signing as `ForcClientAccount` is a valid signer/account from the perspective of the SDK. 

So the refactor we did makes it possible to remove all custom signing logic we have, by moving to use SDK for signatures completly.

forc-plugins/forc-client/src/util/aws.rs
kayagokalp: This is placed here so that AwsConfig can be configured for tests in the future. It doesn't effect anything unless it is set. And by default it is not set in CI/locally. So this is here to make this thing a little bit more testable.

JoshuaBatty (approved):
Approving for now but it would be good to address Sophie's feedback in a follow up PR.

forc-plugins/forc-client/src/util/aws.rs
JoshuaBatty: agree some comments here would be very beneficial

Issue #6226, opened 2024-07-05
Reduce LSP Memory Footprint: Tracking issue.


JoshuaBatty:
* [ ] https://github.com/FuelLabs/sway/issues/5856

PR #6585, opened 2024-09-25
Optimise TokenMap to only hold either a typed or parsed AST token

JoshuaBatty:
## Description

This PR optimizes the `TokenMap` structure by modifying the `Token` type to store either a parsed or typed AST token, but not both simultaneously. Parsed tokens are now only accessible if a typed token does not exist. Previously, both parsed and typed variants were stored which was inefficient in terms of memory usage.

Key changes:
- Replaced separate `parsed` and `typed` fields in `Token` with a single `ast_node` field using an enum `TokenAstNode`
- Added helper methods `as_parsed()` and `as_typed()` to safely access the token variants
- Updated all token creation and access code throughout the codebase to use the new structure
- Refactored the attributes handling code to properly check both parsed and typed variants

Works towards the memory optimization goals in #6226.

## Checklist


alfiedotwtf (dismissed):

sway-lsp/src/utils/debug.rs
```diff
@@ -39,7 +39,7 @@ where
     tokens
         .filter_map(|entry| {
             let (ident, token) = entry.pair();
-            if token.typed.is_some() {
+            if token.as_typed().is_some() {
                 Some(warning_from_ident(ident))
             } else {
                 None
```
alfiedotwtf, line 45: If it's not impossible because of data we might have thrown away by now, will we ever need to go the other way (even just during development)?

alfiedotwtf (approved):

sway-lsp/src/utils/attributes.rs
```diff
@@ -1,47 +1,94 @@
...
-                let decl = engines.pe().get_storage(decl_id);
-                callback(&decl.attributes);
+            TypedAstToken::TypedTraitFn(trait_fn) => {
+                callback(&trait_fn.attributes);
             }
-            Declaration::AbiDeclaration(decl_id) => {
-                let decl = engines.pe().get_abi(decl_id);
+            TypedAstToken::TypedEnumVariant(variant) => {
+                callback(&variant.attributes);
+            }
```
sdankel, line 52: I think these are missing as they also have `attributes` maps:
- TypedAstToken::TypedConfigurableDeclaration
- TypedAstToken::TypedTraitTypeDeclaration
- TypedAstToken::TypedTypeAliasDeclaration

sway-lsp/src/utils/attributes.rs
```diff
@@ -1,47 +1,94 @@
...
+                    callback(&struct_decl.attributes);
+                }
+                ty::TyDecl::StorageDecl(ty::StorageDecl { decl_id, .. }) => {
+                    let storage_decl = engines.de().get_storage(decl_id);
+                    callback(&storage_decl.attributes);
+                }
+                ty::TyDecl::AbiDecl(ty::AbiDecl { decl_id, .. }) => {
+                    let abi_decl = engines.de().get_abi(decl_id);
+                    callback(&abi_decl.attributes);
+                }
```
sdankel, line 32: I think there are a few missing here:

- ty::TyDecl::FunctionDecl
- ty::TyDecl::ConstantDecl

kayagokalp (commented):
While reading this a question popped in my head, I wonder what is the average number of tokens at a given time in the token map. 

If this is sufficiently high, maybe using unions instead of enums in appropriate places will help reduce the memory usage even more at the cost of some code quality (we will need to track the variant ourselves in the code explicitly). I think the enum variant is 8 byte per Token. So the total number of tokens allocated should be very large to see meaningful gains from this change >100k. Also it is possible that rust compiler already optimizes this and compresses some of the variant effects. So the actual effect of the enum might not be directly 8 bytes.

sway-lsp/src/utils/attributes.rs
JoshuaBatty: these are handled separately outside of the `TypedDeclaration` in this section. Good call though

sway-lsp/src/utils/attributes.rs
JoshuaBatty: Nice catch, thanks!

sway-lsp/src/utils/attributes.rs
```diff
@@ -1,47 +1,94 @@
 #![allow(dead_code)]
-use crate::core::token::{AstToken, Token};
-use sway_core::{language::parsed::Declaration, transform, Engines};
+use crate::core::token::{ParsedAstToken, Token, TokenAstNode, TypedAstToken};
+use sway_core::{
+    language::{parsed::Declaration, ty},
+    transform, Engines,
+};
```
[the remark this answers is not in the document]
JoshuaBatty: cool, will add a comment

sway-lsp/src/utils/debug.rs
JoshuaBatty: we generally don't need to go back down the compiler chain once we get to typed tokens. We can always retrieve each stage, i.e lexed, parsed, and typed in vscode by using the command palette with is handy for development.

JoshuaBatty:
> While reading this a question popped in my head, I wonder what is the average number of tokens at a given time in the token map.
> 
> If this is sufficiently high, maybe using unions instead of enums in appropriate places will help reduce the memory usage even more at the cost of some code quality (we will need to track the variant ourselves in the code explicitly). I think the enum variant is 8 byte per Token. So the total number of tokens allocated should be very large to see meaningful gains from this change >100k. Also it is possible that rust compiler already optimizes this and compresses some of the variant effects. So the actual effect of the enum might not be directly 8 bytes.

That's an interesting point, thanks for raising it. I just did a few quick tests to see how big the token map gets on various projects.

sway repo counter example: 354
FUSD protocol-manager-contract: 1378
FUSD libraries project: 2876

Looks like we are sub 10,000 tokens even for the largest sway projects I have for stress testing, so we should be fine for now. Thanks for the suggestion though :)

sway-lsp/src/utils/attributes.rs
sdankel: From what I remember, sometimes the function declaration tokens will come as `TypedDeclaration`s and sometimes they'll be `ty::TyDecl::FunctionDecl`. There was a bug in autocomplete where it wouldn't work for every function because I was only handling one of the decl types. This might've changed though!

sdankel (approved):

alfiedotwtf (approved):

sway-lsp/src/utils/attributes.rs
JoshuaBatty: Thanks Sophie. Im going to rework how we handle this part going forward to remove this confusion.

PR #6586, opened 2024-09-25
migrate encoding v1 to sway-features

xunilrj:
## Description

This PR starts the implementation of https://github.com/FuelLabs/sway-rfcs/blob/master/rfcs/0013-changes-lifecycle.md. 

## CLI flags

Now, all `CLI`s, including our tests, support two new arguments

```
...
      --experimental <EXPERIMENTAL>
          Comma separated list of all experimental features that will be enabled [possible values: new_encoding, storage_domains]
      --no-experimental <NO_EXPERIMENTAL>
          Comma separated list of all experimental features that will be disabled [possible values: new_encoding, storage_domains]
...
```

Experimental features can also be enabled inside `Forc.toml` for workspaces and individual projects.

```
experimental = { encoding-v1 = true }
```

And can be enabled using environment variables:

```
FORC_NO_EXPERIMENTAL=feature_a,feature_b FORC_EXPERIMENTAL=feature_c forc build ...
```

These configurations are applied in a very specific order to allow them to be overwritten. The order from the weakest to the strongest is:
1 - Workspace `TOML`
2 - Project `TOML`
3 - CLI
4 - Environment variables.

The rationale is:
1 - We want an easy way to enable and/or disable a feature for the whole workspace;
2 - But we want to allow projects to overwrite these features, when needed.

These two are the suggested approach to configure experimental features, but we also want to allow someone to easily turn on or off features to test something in particular. For that one can use the CLI flags; and in the specific case one does not have access to the CLI, for example, when one of the `SDK` is compiling the code, one can still completely overwrite the `TOML` files using environment variables.

We are also applying the "no experimental" first. This is to allow the use case of exactly controlling which features will be enabled and disabled. In this case one can use the "meta-feature" `all` to disable everything and enable only the desired features. For example:

```
FORC_NO_EXPERIMENTAL=all FORC_EXPERIMENTAL=new_encoding your_command....
```

## Test for "-h"

This PR also introduces snapshot tests for "-h" for all forc commands at https://github.com/FuelLabs/sway/pull/6586/files#diff-20a5e677d2ae9266a2247739b6a473d2ad0c627955187d668822e7d194f8e940

## Multiple test.toml

This PR also enables the use of multiple `test.toml`. Now each `test.toml` will be a different test, allowing the same code to be compiled and asserted differently.

For example, `main_args_empty` has two files: the normal `test.toml` and a new `test.encoding_v1.toml`. One has `new_encoding` enabled and the other disabled.

When a `test.toml` file has the `experimental` field, we do not use fields with `_new_encoding` suffix anymore, making these files simpler:

test.toml:

```toml
category = "run"

# (1336, 1)
script_data = "0000000000000538 0000000000000001"
expected_result = { action = "return", value = 1337 }

validate_abi = true

experimental = { encoding_v1 = false }
```

test.encoding_v1.toml
```
script_data = "0000000000000538 0000000000000001"
expected_result = { action = "return_data", value = "0000000000000539" }

experimental = { new_encoding = true }
```

Test tomls with a suffix use `test.toml` as a base. So we do not need to repeat every field in them.

Now when we run a test, we will see two tests instead of just one:

```
...
Testing should_pass/language/main_args/main_args_empty (test.encoding_v1.toml) ... ok
Testing should_pass/language/main_args/main_args_empty (test.toml) ... ok
...
```

## Conditional compilation

It is also possible to use conditional compilation using these experimental features. Examples can be found inside `sway-lib-std`.

```sway
#[cfg(experimental_new_encoding = true)]
pub fn assert_eq<T>(v1: T, v2: T) {
...
}
``` 

## Checklist


test/src/e2e_vm_tests/test_programs/should_pass/language/main_args/main_args_empty/test.encoding_v1.toml
```diff
@@ -0,0 +1,4 @@
+script_data = "0000000000000538 0000000000000001"
+expected_result = { action = "return_data", value = "0000000000000539" }
+
+experimental = { new_encoding = true }
```
ironcev, line 4: This is actually a proposal for the file name. Right now it is `test.encoding_v1.toml`. As I understood, we can name the file arbitrarily, and also enable arbitrary experimental features in it which is great and offers perfect flexibility.

But for the most regular case where a single experimental feature will have its own `test.<feature>.toml`, for the sake of consistency I would propose to always use the experimental feature name in the file name. This will save unnecessary guessing.

It this case, it would mean having `test.new_encoding.toml`. (Or we rename the feature flag to `encoding_v1` which would actually be my preference, `new` is somehow too generic.)

ironcev (commented):
Great too see rich infrastructure for feature flags getting implemented! Just one quick comment after glancing over the PR description. Overall, the described infrastructure and intended usage looks great! :heart:

test/src/e2e_vm_tests/test_programs/should_pass/language/main_args/main_args_empty/test.encoding_v1.toml
xunilrj: I agree. I will rename it.

> It this case, it would mean having test.new_encoding.toml. (Or we rename the feature flag to encoding_v1 which would actually be my preference, new is somehow too generic.)

I agree. This is named `encoding_v1` because I forgot to rename it. The issue was that I started with feature named `encoding_v1`, but this demanded changing `sway-std-lib` conditional compilations flags.

And this caused the issue of the current version of the compiler would not compile older `sway-std-lib` anymore.

I think this is too much just for the renaming.

sway-features/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+repository.workspace = true
+
+[dependencies]
+clap = { workspace = true, features = ["derive"] }
+paste = "1.0.15"
```
ironcev, line 13: We already have `paste` on the workspace level.
```suggestion
paste.workspace = true
```

sway-features/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+
+                $(
+                pub fn [<with_ $name:snake>](mut self, enabled: bool) -> Self {
+                    self.[<$name:snake>] = enabled;
+                    self
+                }
+                )*
+            }
+
+            // enum ExperimentalFeatureError {
```
ironcev, line 96: Should we have this enum? If not, let's rather remove the commented code.

ironcev (changes requested):

sway-features/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+                                self.[<$name:snake>] = enabled;
+                                Ok(())
+                            },
+                        )*
+                        "" => Ok(()),
+                        _ => Err(Error::UnknownFeature(feature.to_string())),
+                    }
+                }
+
+                pub fn enable_feature(&mut self, feature: Features, enabled: bool) {
```
ironcev, line 69: When seeing it for the first time in the usage code, the function name created a wrong expectation of doing enabling only. It was definitely confusing. Since it is. unlike `set_enabled`, always used with explicit `true` or `false` the proposal is to have two explicit methods `enable_feature` and `disable_feature`. It will also make callers code more readable and  explicit.

forc-pkg/src/pkg.rs
```diff
@@ -2529,6 +2553,24 @@ pub fn check(
         let pkg = &plan.graph[node];
         let manifest = &plan.manifest_map()[&pkg.id()];
 
+        // Setup experimental features
```
ironcev, line 2556: This is quite some copy-paste. Can we have a single function doing it? Actually, the comment already sounds like a perfect function name: `setup_experimental_features` :smile:

forc-plugins/forc-doc/src/lib.rs
```diff
@@ -108,7 +107,8 @@ pub fn compile_html(
         tests_enabled,
         &engines,
         None,
-        experimental,
+        &[],
```
ironcev, line 110: Just a thought, something we didn't discuss. We will not have documentation for experimental features available. Not an issue IMO.

sway-core/src/build_config.rs
```diff
@@ -193,7 +193,7 @@ pub struct BuildConfig {
     pub(crate) optimization_level: OptLevel,
     pub time_phases: bool,
     pub metrics_outfile: Option<String>,
-    pub experimental: ExperimentalFlags,
+    // pub experimental: ExperimentalFeatures,
```
ironcev, line 196:
```suggestion
```

sway-core/src/build_config.rs
```diff
@@ -314,12 +311,12 @@ impl BuildConfig {
         }
     }
 
-    pub fn with_experimental(self, experimental: ExperimentalFlags) -> Self {
-        Self {
-            experimental,
-            ..self
-        }
-    }
+    // pub fn with_experimental(self, experimental: ExperimentalFeatures) -> Self {
```
ironcev, line 314: Same question going through my head :smile: I am curious why not a part of `BuildConfig`. But if it isn't lets delete the commented code.

ironcev (changes requested):

forc-pkg/src/pkg.rs
ironcev: Also, thinking of encapsulation, shouldn't that function be in the `sway_features::ExperimentalFeatures`? In that case, all the existing functions would actually be private and the tools would just be calling `ExperimentalFeatures::setup(manifest, experimental, no_experimental)` instead of `default()` and not bother about the internal logic, ordering etc. as they must now.

sway-features/src/lib.rs
```diff
@@ -0,0 +1,231 @@
+use clap::{Parser, ValueEnum};
+
+macro_rules! features {
+    (count; ) => {0};
+    (count; $name:ident $($rest:tt)?) => {1 + features!{count; $($rest)*}};
+    ($($name:ident = $enabled:literal, $url:literal),* $(,)?) => {
+        paste::paste! {
+            pub const CFG: [&str; features!{count; $($name)*}] = [
```
ironcev, line 8: We can have `pub const CFG: &[&str] = &["f1", "f2"];` here. In that case we don't need the recursive macro call just to calculate the size of the array. It simplifies the logic of the macro.

sway-features/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+#[derive(Debug)]
+pub enum Error {
+    ParseError(String),
+    UnknownFeature(String),
+}
+
+impl std::fmt::Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Error::ParseError(_) => f.write_str("ParserError"),
```
ironcev, line 146: Can we have some more user friendly errors here?
```suggestion
            Error::ParseError(feature) => f.write_fmt(format_args!("Experimental feature \"{feature}\" cannot be parsed.")),
```

sway-features/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+            Error::UnknownFeature(feature) => {
+                f.write_fmt(format_args!("UnknownFeature: {feature}"))
```
ironcev, line 148:
```suggestion
                f.write_fmt(format_args!("Unknown experimental feature: \"{feature}\"."))
```

sway-lsp/src/core/session.rs
```diff
@@ -309,7 +308,8 @@ pub fn compile(
         true,
         engines,
         retrigger_compilation,
-        experimental,
+        &[],
```
ironcev, line 311: Does this mean LSP will not be supporting experimental features, or I am understanding the code wrongly? What about the experimental features that change the parsing, add new language elements etc like `&self`?

sway-lsp/src/handlers/notification.rs
```diff
@@ -28,6 +28,7 @@ pub async fn handle_did_open_text_document(
     // Otherwise, don't recompile the project when a new file in the project is opened
     // as the workspace is already compiled.
     if session.token_map().is_empty() {
+        dbg!();
```
ironcev, line 31: A few `dbg!`s to remove from this file :smile:

ironcev (changes requested):

forc-pkg/src/manifest/build_profile.rs
```diff
@@ -65,9 +57,7 @@ impl BuildProfile {
             error_on_warnings: false,
             reverse_results: false,
             optimization_level: OptLevel::Opt0,
-            experimental: ExperimentalFlags {
-                new_encoding: false,
-            },
+            // experimental: ExperimentalFeatures::default(),
```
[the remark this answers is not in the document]
xunilrj: Done.

forc-pkg/src/manifest/build_profile.rs
```diff
@@ -88,9 +78,7 @@ impl BuildProfile {
             error_on_warnings: false,
             reverse_results: false,
             optimization_level: OptLevel::Opt1,
-            experimental: ExperimentalFlags {
-                new_encoding: false,
-            },
+            // experimental: ExperimentalFeatures::default(),
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-features/Cargo.toml
xunilrj: Done.

sway-features/src/lib.rs
xunilrj: Done.

sway-features/src/lib.rs
xunilrj: I changed some of the names. Hope is better now.

forc-pkg/src/pkg.rs
xunilrj: Done.

sway-core/src/build_config.rs
xunilrj: Done.

sway-core/src/build_config.rs
xunilrj: Done.

sway-features/src/lib.rs
xunilrj: That is actually a great idea! Done.

sway-features/src/lib.rs
xunilrj: Done.

sway-features/src/lib.rs
xunilrj: Done.

sway-lsp/src/core/session.rs
xunilrj: > Does this mean LSP will not be supporting experimental features

sway-lsp/src/server_state.rs
```diff
@@ -126,16 +126,17 @@ impl ServerState {
...
         let rx = self.cb_rx.clone();
         let last_compilation_state = self.last_compilation_state.clone();
-        let experimental = sway_core::ExperimentalFlags {
-            new_encoding: false,
-        };
         std::thread::spawn(move || {
             while let Ok(msg) = rx.recv() {
                 match msg {
                     TaskMessage::CompilationContext(ctx) => {
+                        dbg!();
```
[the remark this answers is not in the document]
xunilrj: Done.

forc-plugins/forc-doc/src/lib.rs
xunilrj: Not sure I understood.

sway-lsp/src/handlers/notification.rs
xunilrj: Done.

jjcnn (commented):
> These configurations are applied in a very specific order to allow them to be overwritten. The order from the weakest to the strongest is:
> 1 - Workspace TOML
> 2 - Project TOML
> 3 - CLI
> 4 - Environment variables.

Are you sure you want environment variables overwriting CLI and not the other way around? It feels like that could lead to some confusion for the user when he's passing a CLI argument that gets ignored because of an environment variable he didn't realize had been set.

Another thing is the use of both `experimental` and `no-experimental`. What happens if you give contradictory settings? And what happens if you fail to give a setting for a feature - presumably we fall back on a default setting, but wouldn't it then be better to only have the `experimental` option, and use the default setting if no user-defined setting is given?

xunilrj:
> Are you sure you want environment variables overwriting CLI and not the other way around? 

If CLI dominates, developers without access to `forc` cannot overwrite these features. Example: you compile through SDK or an other tool. In this case, you don't necessarily have access to the CLI and may not be able to control which features you want on or off.

Another use case is to overwrite them in a CI environment easily.

What we may do, is print a warning, that some features are being overwritten by environment variables.

> Another thing is the use of both `experimental` and `no-experimental`. What happens if you give contradictory settings? 

"Contradictory settings" is actually a feature that we want eventually. If you want to test a specific set of settings you do `FORC_NO_EXPERIMENTAL=all FORC_EXPERIMENTAL=feature_a`. So you can precisely control which features are on and off.

> And what happens if you fail to give a setting for a feature - presumably we fall back on a default setting, but wouldn't it then be better to only have the `experimental` option, and use the default setting if no user-defined setting is given?

Yes, we fallback to default. But without "NO_EXPERIMENTAL", if an experimental feature is enabled by default, there is no way to turn it off.

forc-plugins/forc-doc/src/lib.rs
ironcev: Hmm, after giving it a second thought I think I was too optimistic. Essentially, `forc doc` needs to be able to compile a project to generate the documentation. If it cannot compile it, it will generate the docs up to the point it can compile.

Assuming we have a project that internally uses experimental features, with the current implementation that always passes `&[]` for `experimental` and `no-experimental` that project will never compile. Means the docs will either not be generated or be only partially generated.

The solution is to pass the `(no-)experimental` flags to the `forc doc` as well. Looking at the `forc doc --help` I see it already has the `--no-encoding-v1` flag which makes sense. That should then be removed and replaced with `(no)-experimental`.

ironcev: Hmm, another thought. `forc doc` will still pick the flags if they are declared in `toml`/env. Only the command line ones will not be available. This will very likely be fine for usual case. Perhaps the optimistic look is ok and we don't necessarily extend the CLI or do it in another step.

But removing  the `--no-encoding-v1` still makes sense. And adding flags then perhaps shouldn't be much effort.

sway-features/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+        paste::paste! {
+            pub const CFG: &[&str] = &[
+                $(
+                    stringify!([<experimental_ $name:snake>]),
+                )*
+            ];
+
+            #[derive(Copy, Clone, Debug, ValueEnum)]
+            #[value(rename_all = "snake")]
+            pub enum Features {
```
ironcev, line 16: A small detail that suddenly became obvious when I looked at the new `set_enabled(&mut self, feature: Features, enabled: bool)`. The `feature: Feature*s*` part. `Features` should actually be `Feature` (singular) to follow the Rust convention and the one we currently have (`TypeInfo`, `TyExpression`, ...).

sway-lsp/tests/lib.rs
```diff
@@ -1531,67 +1531,67 @@ fn go_to_definition_for_storage() {
...
-            req_line: 26,
-            req_char: 23,
-            def_line: 9,
-            def_start_char: 4,
-            def_end_char: 5,
-            def_path: "sway-lsp/tests/fixtures/tokens/storage/src/main.sw",
-        };
-        // storage.var1.z.x
-        lsp::definition_check(&server, &go_to).await;
+        // let mut go_to = GotoDefinition {
```
ironcev, line 1534: Just a reminder that these tests are still commented out.

test/src/e2e_vm_tests/mod.rs
```diff
@@ -282,7 +285,7 @@ impl TestContext {
     ) -> Result<ContractId> {
         let key = DeployedContractKey {
             contract_path: contract_path.clone(),
-            new_encoding: run_config.experimental.new_encoding,
+            new_encoding: true, // TODO
```
ironcev, line 288: Also just a reminder :smile:

test/src/e2e_vm_tests/mod.rs
```diff
@@ -936,22 +934,78 @@ fn parse_test_toml(path: &Path, run_config: &RunConfig) -> Result<TestDescriptio
...
+            let base_test_toml = path.parent().unwrap().join("test.toml");
+            let base_test_toml = std::fs::read_to_string(base_test_toml)?;
+            let base_toml_content = base_test_toml.parse::<toml::Value>()?;
+            match base_toml_content {
+                toml::Value::Table(map) => {
+                    for (k, v) in map {
+                        let _ = toml_content_map.entry(&k).or_insert(v);
+                    }
+                }
+                _ => todo!(),
```
ironcev, line 956: Shouldn't this be unreachable if we sort the result of `glob::glob` by path? In that case, the base `test.toml` will always be parsed before any feature tomls and will fail at that point if it cannot be parsed. In that case we know that here it will always be available an we can simply `expect` and perhaps skip the `match` completely.

test/src/e2e_vm_tests/mod.rs
```diff
@@ -936,22 +934,78 @@ fn parse_test_toml(path: &Path, run_config: &RunConfig) -> Result<TestDescriptio
...
+            }
+        };
+    }
+
+    let mut run_config = run_config.clone();
+
+    // To keep the current test.toml compatible we check if a field named "experimental" exists
+    // or not. If it does not, we keep the current behaviour.
+    // If it does, we ignore the experimental flags from the CLI and use the one from the toml file.
+    // TODO: this backwards compatilibty can be removed after all tests migrate to new version
```
ironcev, line 966:
```suggestion
    // TODO: this backwards compatibility can be removed after all tests migrate to the new version
```

ironcev (changes requested):

test/src/e2e_vm_tests/mod.rs
xunilrj: Done.

test/src/e2e_vm_tests/mod.rs
xunilrj: Done.

sway-lsp/tests/lib.rs
xunilrj: Done.

sway-features/src/lib.rs
xunilrj: Done.

forc-plugins/forc-doc/src/lib.rs
xunilrj: Done.

xunilrj:
`forc doc` help is adapting itself to the terminal size, which means it gives different results locally and inside Github CI. I will try to fix it later.

ironcev (approved):
Looking forward to get this PR merged and start incorporating experimental flags in two features I work on right now :smile: 

Just one additional remark. The support for E2E tests is great, but we will down the road also need to decide what to do with, e.g. SDK tests or Sway language tests. The former run precompiled Sway binaries which will for the time being be compiled without experimental features. The latter just run `forc test` so again, without experimental features. I expect once we start getting first PRs with experimental features, we can discuss the approach to bringing these tests to run with experimental features as well.

test/src/e2e_vm_tests/test_programs/should_pass/language/main_args/main_args_empty/test.encoding_v1.toml
ironcev: Just to ask here, since you've mentioned you will rename the file, did you forget or rather decided in the end to keep the current name :smile:

IGI-111 (approved):

PR #6605, opened 2024-10-04
Implement Serde for AstNode types

JoshuaBatty:
## Description
This PR extends Serde support across all AstNode-related types, building upon the partial implementation from #4193. Key points:

* Adds both `Serialize` and `Deserialize` trait implementations to all relevant types.
* Prepares for upcoming language server features that will serialize/deserialize types between keystrokes.
* Important: The type interface in the compiler is not yet stable. This implementation is intended for temporary serialization/deserialization work, not for persistent storage.

## Checklist


JoshuaBatty:
> > * Prepares for upcoming language server features that will serialize/deserialize types between keystrokes.
> 
> Could you maybe expand a bit on how this will work?

Hey @tritao sorry for the delay, just getting back to this work now. So the motivation for this was noticing really high RAM usage in the language server now. For example, just running the counter example we can see 800mb of heap memory is never deallocated after the first compilation pass. 

On digging it seems this is due to the new module caching where we hold onto the types themselves, thus never allow `Drop` to be called.

```rust
#[derive(Clone, Debug)]
pub struct TypedModuleInfo {
    pub module: Arc<TyModule>,
    pub version: Option<u64>,
}
```

My motivation was to Serialize these types using [postcard](https://crates.io/crates/postcard) (same that is used in fuel-core) and then store the Serialize version in the module cache. We then need to Deserialize back into the original type on each keystroke which will incur some performance overhead but I haven't got any measurements yet, was going to tackle that in an up coming PR. WDYT, are you happy with the approach or do you think there is a better solution?

sway-ast/src/priv_prelude.rs
```diff
@@ -35,7 +35,7 @@ pub use {
     },
     extension_trait::extension_trait,
     num_bigint::BigUint,
-    serde::Serialize,
+    serde::{Deserialize, Serialize},
```
zees-dev, line 38: since theres many serialization formats; and serde is quite common; unsure if we need to re-export this

alfiedotwtf (approved):

zees-dev (approved):
LGTM 👍

IGI-111 (approved):

sway-ast/src/priv_prelude.rs
JoshuaBatty: Good point, I guess this PR just re-exports `Deserialize` and `Serialize` was already exported. Should be fairly easy to remove this in the future if `sway-ast` needs to support multiple serialisation methods.

Issue #6572, opened 2024-09-20
Trait: "method dummy and was not properly replaced" error in 0.63.6


Lukasz2891:
### Related Component

compiler

### Problem

Let have a library (named redstone) with the lib file:

lib.sw
```rust
library;

use std::{bytes::Bytes};

pub trait FromBytesConvertible {
    fn size() -> u64;
    fn _from_be_bytes(bytes: Bytes) -> Self;
}

pub trait FromBytes {
    fn from_bytes(bytes: Bytes) -> Self;
}

impl<T> FromBytes for T
where
    T: FromBytesConvertible,
{
    fn from_bytes(bytes: Bytes) -> Self {
        assert(bytes.len() <= Self::size());

        let mut bytes = bytes;

        while (bytes.len() < Self::size()) {
            bytes.insert(0, 0u8);
        }

        Self::_from_be_bytes(bytes)
    }
}


pub struct DataPoint {}
pub struct DataPackage {}
pub struct Payload {}

pub fn make_data_package(bytes: Bytes) -> (DataPackage, u64) {
    let mut data_points = Vec::new();
    data_points.push(DataPoint::from_bytes(bytes));

    return (DataPackage {}, 0);
}

impl FromBytes for DataPoint {
    fn from_bytes(bytes: Bytes) -> Self {
        Self {
        }
    }
}

impl Payload {
    pub fn from_bytes(bytes: Bytes) -> Self {
        let mut data_packages = Vec::new();
        let (data_package, _) = make_data_package(bytes);
        data_packages.push(data_package);

        Self {}
    }
}

pub fn process_input(bytes: Bytes) -> (Vec<u256>, u64) {
    let payload = Payload::from_bytes(bytes);

    (Vec::new(), 0)
}
```

and the contract:

```rust
contract;

use std::{
    bytes::Bytes,
    vec::Vec,
};
use redstone::process_input;

abi RedStoneCore {
    fn get_prices(feed_ids: Vec<u256>, payload: Bytes) -> (Vec<u256>, u64);
}

impl RedStoneCore for Contract {
    fn get_prices(feed_ids: Vec<u256>, payload_bytes: Bytes) -> (Vec<u256>, u64) {
        process_input(payload_bytes)
    }
}
```

The library compiles properly, but the contract using the library does not. It leads to an error:

```
error
 --> /[Redacted]/redstone-fuel-sdk/src/lib.sw:7:8
  |
5 | 
6 |     fn size() -> u64;
[...]

PR #6606, opened 2024-10-04
Fixes find_method_for_type bail out with 1st match

### LINKED ISSUE #5086: Trait implementations with a generic override specific cases
When a trait implementation exists that takes a generic, this will override all other cases. Implementations for specific types should instead override generic cases.

As it currently stands, it is impossible to add an additional `From` implementation for `Vec` due to this error.

In the following example, `MyStruct` has 2 `From` implementations. One uses the generic i.e. `MyStruct<T>` and converts to a `b256`, while the other is only implemented under a `MyStruct<u64>`. While the `into()` methods work, it is currently impossible to use the `from()` function.

```rust
library;

struct MyStruct<T> {
    val: u64
}

// Implements on MyStruct<T>
impl<T> From<b256> for MyStruct<T> {
    fn from(num: b256) -> Self {
        MyStruct { val: 0 }
    }

    fn into(self) -> b256 {
        ZERO_B256
    }
}

// Implements on MyStruct<u64>
impl From<u64> for MyStruct<u64> {
    fn from(num: u64) -> Self {
        MyStruct { val: 0 }
    }

    fn into(self) -> u64 {
        0
    }
}

#[test]
fn test_trait() {
    let my_struct: MyStruct<u64> = MyStruct { val: 1 };
    let my_b256 = ZERO_B256;
    let my_u64 = 1;

    // Works as anticipated
    let into_b256: b256 = my_struct.into();
    let into_u64: u64 = my_struct.into();
    // This does not work for `from_u64`
    let from_b256: MyStruct<b256> = MyStruct::from(my_b256);
    let from_u64: MyStruct<u64> = MyStruct::from(my_u64);
}
```

Screenshot 2023-09-04 at 1 50 03 PM

esdrubal:
## Description

When find_method_for_type could not match the parameters and return type with the available methods it would return the first match by name.

This was not a deterministic and secure behavior so now we strictly match the parameter and return types.

Fixes #5086

find_method_for_type now if we have: impl<T> FromBytes for T
and: impl FromBytes for DataPoint, we pick the second implementation.

Fixes https://github.com/FuelLabs/sway/issues/6572

## Checklist

### DIFF
--- a/forc-pkg/src/pkg.rs
+++ b/forc-pkg/src/pkg.rs
@@ -46,7 +46,7 @@ use sway_core::{
     transform::AttributeKind,
     write_dwarf, BuildTarget, Engines, FinalizedEntry, LspConfig,
 };
-use sway_core::{PrintAsm, PrintIr};
+use sway_core::{set_bytecode_configurables_offset, PrintAsm, PrintIr};
 use sway_error::{error::CompileError, handler::Handler, warning::CompileWarning};
 use sway_features::ExperimentalFeatures;
 use sway_types::constants::{CORE, PRELUDE, STD};
@@ -1914,7 +1914,7 @@ pub fn compile(
 
     let errored = handler.has_errors() || (handler.has_warnings() && profile.error_on_warnings);
 
-    let compiled = match bc_res {
+    let mut compiled = match bc_res {
         Ok(compiled) if !errored => compiled,
         _ => return fail(handler),
     };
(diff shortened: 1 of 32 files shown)


IGI-111 (approved):

jjcnn (approved):
One minor comment, but otherwise it looks good.

Some of the methods in type_check_context.rs are getting a bit long - I wonder if we should put some time into refactoring them?

PR #6611, opened 2024-10-06
feat: generate a loader abi of loaders generated for predicates and scripts

kayagokalp:
## Description

Thanks a lot to @hal3e for helping debug this.

With this PR, there will be an additional abi in the `out` folder for predicates and scripts that are deployed as loaders following the `{package_name}-loader-abi.json` naming convention.

This PR adds `loader-abi.json` file generation capability to forc. To do so forc needs to have the old (original) and new (loader) data offsets of the script/predicate. After getting it the difference should be applied to all configurable slots to find out the new locations of the configurable slots inside the loader binary. 

Basically enables backwards compatibility with the older sdk releases. Since forc can now generate a loader abi with correct configurable offsets, it can be used to load the script as it is, (without the newer loader stuff) and everything will be still working.


forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -356,6 +360,96 @@ pub async fn deploy(command: cmd::Deploy) -> Result<Vec<DeployedPackage>> {
...
+                REG_ADDRESS_OF_DATA_AFTER_CODE,
+                WORD_SIZE as u16,
+            ),
+            // load the data section of the executable
+            op::ldc(REG_ADDRESS_OF_DATA_AFTER_CODE, 0, REG_GENERAL_USE, 2),
+            // Jump into the memory where the contract is loaded.
+            // What follows is called _jmp_mem by the sway compiler.
+            // Subtract the address contained in IS because jmp will add it back.
+            op::sub(
+                REG_START_OF_LOADED_CODE,
```
kayagokalp, line 416: looks like this is a fixed size. I think we can also remove this and embed the size, if this solution works.

hal3e: Fixed for now but if we change it in the future in the SDK would it impact your offset calculation ?

kayagokalp: I think ideally in a future relase, sdk can expose this offsets so that we can re-use it without calculating it again

JoshuaBatty (approved):
🫡

hal3e (approved):

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -356,6 +360,103 @@ pub async fn deploy(command: cmd::Deploy) -> Result<Vec<DeployedPackage>> {
...
+            op::divi(REG_START_OF_LOADED_CODE, REG_START_OF_LOADED_CODE, 4),
+            // Jump to the start of the contract we loaded.
+            op::jmp(REG_START_OF_LOADED_CODE),
+        ]
+    };
+
+    let offset = extract_data_offset(binary)?;
+
+    if binary.len() < offset {
+        anyhow::bail!("data sectio offset is out of bounds");
```
alfiedotwtf, line 436: Nit: sectio

(This can be fixed later)

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -387,6 +488,29 @@ pub async fn deploy_executables(
...
+                println_action_green("Generating", "loader abi for the uploaded executable.");
+                let json_abi_path = out_dir.join(format!("{pkg_name}-loader-abi.json"));
+                let original_data_section = extract_data_offset(&pkg.bytecode.bytes).unwrap();
+                let offset_shift = original_data_section - loader_data_section_offset;
+                // if there are configurables in the abi we need to shift them by `offset_shift`.
+                let configurables = fuel_abi.configurables.clone().map(|configs| {
+                    configs
+                        .into_iter()
+                        .map(|config| Configurable {
+                            offset: config.offset - offset_shift as u64,
```
alfiedotwtf, line 504: Does this subtraction and the one above need to be checked just in case we get an invalid value and end up underflowing etc?

alfiedotwtf (approved):

forc-plugins/forc-client/src/op/deploy.rs
Voxelot: Maybe make an issue to fix this later. The merge is blocked on unresolved comments and I see you already approved, so Im going to resolve this now.

Issue #6597, opened 2024-09-30
Introduce `Cmp` library


bitzoic:
A number of external projects have implemented their own `min()` and `max()` `Cmp` functions. These should be included in the std-lib for the `u64`, `u8`, `u16`, `u32`, `U128` and `u256` types.

SwayStar123:
Duplicate of #2351 and #6169

Issue #2351, opened 2022-07-20
std::cmp Functionality for ordering and comparison


SwayStar123:
There is already Eq, and Ord present in the lib core, however it would be nice to have a full cmp module in the standard library

https://doc.rust-lang.org/std/cmp/index.html

Here is the rust std::cmp module for reference

jtriley-eth:
Is anything blocking this or can we move forward with it?

otrho:
Feel free to try.  Our generics support should be up to it, I think.

SwayStar123:
Needs where clauses, I don't think those are supported yet

otrho:
I thought they were, @emilyaherbert has added a bunch of stuff to the type system lately.  Do we have super traits and where clauses?

emilyaherbert:
> Do we have super traits and where clauses?

Supertraits yes, but to implement this the way you'll want to @SwayStar123, you'll need to wait for where clauses on impl blocks #3731. We currently only have where clauses for free functions.

PR #6635, opened 2024-10-15
feat: add `TotalOrd` library to `core::ops` for `uint` types

EdwardJES:
## Description
This PR addresses https://github.com/FuelLabs/sway/issues/6597. 

A `cmp` trait has been added that exposes:
- `min(self, Self) -> Self`
- `max(self, Self) -> Self`

## Checklist


IGI-111 (changes requested):
This is nice, but I'd much rather we get something like `Ord`+`PartialOrd`, which is only slightly more code to spin up.

EdwardJES:
> This is nice, but I'd much rather we get something like `Ord`+`PartialOrd`, which is only slightly more code to spin up.

Thanks for the feedback, @IGI-111. Just before I make this change. I'd just like to run my understanding by you. Extend this current trait (`cmp` (maybe this name is not appropriate given `Ord` and `PartialOrd` have `cmp` in their methods)) to have functionality from `Ord` and `PartialOrd`, and introduce the `Ordering` enum. 

I'm a little confused as to where the `PartialOrd.ge, `PartialOrd.le` would be written as the `Ord` trait in the `core` lib implements `lt, gt` already.

Lastly, I want to make sure I'm writing the API so that the full `cmp` module can be implemented later: https://github.com/FuelLabs/sway/issues/2351. Does this PR build towards this issue correctly?

I would appreciate any help on the above, thank you 🙏

IGI-111:
Ah yes, it's unfortunate that we already added an `Ord` that doesn't actually map onto Rust's `Ord`.

What I would prefer is if we can achieve that same level of functionality and have two different traits for total and partial orders. But I suppose that's what your existing code is already trying to achieve with `Cmp` mapping to `Ord` and `Ord` mapping to `PartialOrd`.

I think we can just merge `Cmp` as is or with a name change (`TotalOrd`?) and implement the `Ordering` functionality later in a breaking change that actually restores the expected names for people who are coming from Rust.

So really, forget what I said and maybe change the name of the trait to `TotalOrd` and we're good to go for this.

EdwardJES:
Thanks for the feedback @IGI-111, I've renamed occurrences of `cmp` to `TotalOrd`. 

Agree that there is a nice DX gain from maintaining a nice mapping between Rust. Let me know if you'd like me to capture that future change an issue.

sway-lib-std/src/lib.sw
```diff
@@ -45,3 +45,4 @@ pub mod low_level_call;
 pub mod array_conversions;
 pub mod bytes_conversions;
 pub mod clone;
+pub mod totalord;
```
IGI-111, line 48: One more thing.
The trait and impl should be in `core::ops` instead of `std::totalord`

EdwardJES:
Cool, makes sense @IGI-111. Have migrated the trait to `core::ops` now. Thank you.

EdwardJES:
Will get CI passing in the next day or so.

bitzoic (approved):
This is great! Thank you for the good use of the SRC-2 standard 😁

sway-lib-core/src/ops.sw
```diff
@@ -1215,6 +1215,122 @@ impl Shift for u16 {
     }
 }
 
+/// Trait to compare values of the same type.
+pub trait TotalOrd {
```
sdankel, line 1219: I wonder if this should be in a new module, like `cmp.sw`. In rust, `Ord` and `PartialOrd` are in `std::cmp`. Should we do the same and put this in the std library instead of core?

sdankel: Never mind, just saw https://github.com/FuelLabs/sway/pull/6635#discussion_r1807539128

sdankel (approved):

IGI-111 (approved):

PR #6641, opened 2024-10-15
improve asm copy propagation for MOVE instruction

xunilrj:
## Description

This PR improve value propagation optimization in two ways:

1 - First, when the optimizer knows that both source and destination registers of the `MOVE`  instruct have the same value and version; it will remove the instruction as it is useless;
2 - Otherwise, we propagate the value saying that destination now have the same value and version of the source register for other optimizations.

## Checklist

### DIFF
--- a/sway-core/src/asm_generation/fuel/optimizations.rs
+++ b/sway-core/src/asm_generation/fuel/optimizations.rs
@@ -27,7 +27,7 @@ impl AbstractInstructionSet {
         }
 
         // What does a register contain?
-        #[derive(Debug)]
+        #[derive(Debug, PartialEq, Eq)]
         enum RegContents {
             Constant(u64),
             BaseOffset(VRegDef, u64),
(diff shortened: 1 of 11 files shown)


sway-core/src/asm_generation/fuel/optimizations.rs
```diff
@@ -57,145 +57,184 @@ impl AbstractInstructionSet {
...
-                    VirtualOp::MUL(dest, opd1, opd2) => {
-                        match (reg_contents.get(opd1), reg_contents.get(opd2)) {
-                            (Some(RegContents::Constant(c1)), Some(RegContents::Constant(c2))) => {
-                                reg_contents.insert(dest.clone(), RegContents::Constant(c1 * c2));
+                either::Either::Left(op) => {
+                    match op {
+                        VirtualOp::ADD(dest, opd1, opd2) => {
+                            // We don't look for the first operand being a constant and the second
+                            // one a base register. Such patterns must be canonicalised prior.
+                            if let Some(&RegContents::Constant(c2)) = reg_contents.get(opd2) {
```
vaivaswatha, line 106: Am I correct in understanding that this is a bug fix? (unrelated to MOVs). We're missing an optimization opportunity here

sway-core/src/asm_generation/fuel/optimizations.rs
```diff
@@ -57,145 +57,184 @@ impl AbstractInstructionSet {
...
+                                let new_sw =
+                                    VirtualOp::SW(base_reg.reg.clone(), src.clone(), new_imm);
+                                // Replace the SW with a new one in-place.
+                                *op = new_sw;
+                            }
+                            _ => (),
+                        },
+                        VirtualOp::MOVE(dest, src) => {
+                            let ver = get_def_version(&latest_version, src);
+                            if let Some(RegContents::BaseOffset(src, 0)) = reg_contents.get(src) {
```
vaivaswatha, line 193: are you trying to catch something like
```
a = b
b = a
```
?? Not sure I understand what is it that you're trying to optimize, and why the existing copy propagation in the assembly generator doesn't apply.

sway-core/src/asm_generation/fuel/optimizations.rs
xunilrj: This should be the same code, but with a `if let` to accommodate that this runs inside a `retaint_mut` now.

sway-core/src/asm_generation/fuel/optimizations.rs
xunilrj: For example, from the `configurable_const` test:
image

xunilrj: On `a = b`, this match arm does two things:
1 - first, it indeed checks if `a` and `b` are already the same, and it removes the instruction;
2 - but it also propagates that `b` and `a` are the same for future optimizations.

sway-core/src/asm_generation/fuel/optimizations.rs
vaivaswatha: ah got it. I read that as an `if let` statement both before and after your change. But it's `let else` before and `if let` now.

vaivaswatha (dismissed):

vaivaswatha:
Do you think we can have numbers before and after this change? maybe for just one or two important contracts if not our testsuite.

vaivaswatha (approved):

IGI-111 (approved):

kayagokalp (approved):

PR #6651, opened 2024-10-17
[asmgen] generate code for stack argument access differently

### LINKED ISSUE #6646: Codegen panics on too many arguments on large binaries
Values of `stack_offset_bytes` superior to 4096 trigger "Too many arguments, can not handle" panic even with low argument count.

To reproduce, attempt to compile https://github.com/Props-Labs/props-fuels/tree/53d535a8e39d61021f424d55ce44abc5d58bfc8b/packages/sway-programs/Props721Collection-contract

You should get the following error:

```
thread 'main' panicked at sway-core/src/asm_generation/fuel/functions.rs:124:34:
Too many arguments, cannot handle.: Immediate12TooLarge { val: 4670, span: Span { src (ptr): 0x5585f71f33f0, source_id: None, start: 0, end: 0, as_str(): "" } }
```

vaivaswatha:
To save excess arguments on the stack, previously we would
  1. store the argument using SW with an immediate offset basing from locals-base register. This limits the offset to 12 bits.
  2. Store the offset to the first excess argument to the last ARG register so that callee can access these.

This PR reverses the order. It stores the offset to the last ARG register first. Then the store of argument to the stack slot can be made without requiring large immediate offsets, basing just on the last ARG register.

Fixes #6646

### DIFF
--- a/sway-core/src/asm_generation/fuel/functions.rs
+++ b/sway-core/src/asm_generation/fuel/functions.rs
@@ -84,49 +84,9 @@ impl<'ir, 'eng> FuelAsmBuilder<'ir, 'eng> {
                 ));
             }
         } else {
-            // Put NUM_ARG_REGISTERS - 1 arguments into arg registers and rest into the stack.
-            for (idx, arg_val) in args.iter().enumerate() {
-                let arg_reg = self.value_to_register(arg_val)?;
-                // Except for the last arg register, the others hold an argument.
-                if idx < compiler_constants::NUM_ARG_REGISTERS as usize - 1 {
-                    self.cur_bytecode.push(Op::register_move(
-                        VirtualRegister::Constant(ConstantRegister::ARG_REGS[idx]),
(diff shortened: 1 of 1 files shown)


IGI-111 (dismissed):

sdankel (commented):
Should we add the [test program](https://github.com/Props-Labs/props-fuels/tree/53d535a8e39d61021f424d55ce44abc5d58bfc8b/packages/sway-programs/Props721Collection-contract) from the original issue to our test suite?

vaivaswatha:
> Should we add the [test program](https://github.com/Props-Labs/props-fuels/tree/53d535a8e39d61021f424d55ce44abc5d58bfc8b/packages/sway-programs/Props721Collection-contract) from the original issue to our test suite?

Done.

sdankel (approved):
lgtm 👍

IGI-111 (approved):

PR #6654, opened 2024-10-17
fix: resolve UB in _add, __sub, _mul for u8, u16, u32

### LINKED ISSUE #6653: u16 and u32 subtract methods don't handle underflows when enabled

### LINKED ISSUE #6666: u16 and u32 add, subtract and multiply methods can create UB

K1-R1:
## Description

handle u32 and u16 subtract underflow manually. 
Mirror u8 impl's use of u64 conversion for u16 and u32.
Add additional u8 test.
Closes #6653. 
Closes #6666.

## Checklist


SwayStar123 (dismissed):

sway-lib-core/src/ops.sw
```diff
@@ -173,17 +173,38 @@ impl Subtract for u64 {
...
 impl Subtract for u32 {
     fn subtract(self, other: Self) -> Self {
-        __sub(self, other)
+        // any non-64-bit value is compiled to a u64 value under-the-hood
+        // constants (like Self::max() below) are also automatically promoted to u64
+        let res = __sub(self, other);
+        // integer underflow
+        if __gt(res, Self::max()) {
+            if panic_on_overflow_is_enabled() {
+                __revert(0)
```
ironcev, line 184: Are there any guidelines accepted on choosing revert codes in the `core`/`std`? For asserts we have specific codes and I must say they are helpful (yes, I already recognize them visually :smile:). IMO having a specif code here would also be helpful.

sway-lib-core/src/ops.sw
```diff
@@ -173,17 +173,38 @@ impl Subtract for u64 {
...
     }
 }
 
 impl Subtract for u16 {
     fn subtract(self, other: Self) -> Self {
-        __sub(self, other)
+        let res = __sub(self, other);
+        if __gt(res, Self::max()) {
+            if panic_on_overflow_is_enabled() {
+                __revert(0)
```
ironcev, line 201: Same question here.

sway-lib-core/src/ops.sw
K1-R1: It's a good point; in this case I chose to match the revert(0) used in the similar add methods:https://github.com/FuelLabs/sway/blob/0b3111771711104d4c52f6dfdf2cdc1702ea2202/sway-lib-core/src/ops.sw#L65
AFAIK there aren't any accepted guidelines for revert codes here. But this is something we should probably do as a follow up and then apply to both core/std

K1-R1: For now, unless there are guidelines I'm unaware I think we should get these patches through ASAP.

sway-lib-core/src/ops.sw
```diff
@@ -173,17 +173,38 @@ impl Subtract for u64 {
...
 }
 
-// unlike addition, underflowing subtraction does not need special treatment
-// because VM handles underflow
 impl Subtract for u32 {
     fn subtract(self, other: Self) -> Self {
-        __sub(self, other)
+        // any non-64-bit value is compiled to a u64 value under-the-hood
+        // constants (like Self::max() below) are also automatically promoted to u64
+        let res = __sub(self, other);
```
xunilrj, line 180: I think doing `__sub` like this is an "undefined behaviour" because it can generate an "invalid u32". We can decide to fix this in the compiler somehow. 

Given that you are already doing a comparison, why not just compare `self` to `other` directly?

```sway
if __gt(other, self) {
     ...
} else {
    // SAFE: self is greater than other
    __sub(self, other)
}
```

K1-R1:
Additional logic in addition, subtract and multiplication ops has been added. But this breaks the `const_inits` tests as in the image below.
Screenshot 2024-10-28 at 18 36 59

Are there any objections to removing the erroring line in that specific test and marking the PR as breaking?

K1-R1:
Blocked by https://github.com/FuelLabs/sway/issues/6874

bitzoic (approved):

IGI-111 (approved):

PR #6656, opened 2024-10-17
ci: move to buildjet

kayagokalp:
## Description

Moves CI runners to buildjet from github.

## Benchmarks

|         | before | after |
|---------|--------|-------|
| total   | 6823   | 4815  |
| slowest | 1004   | 717   |

Overall, %30 faster CI times, ~5 minutes removed for each PR in slowest run (e2e testing).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ env:
 
 jobs:
   get-fuel-core-version:
-    runs-on: ubuntu-latest
+    runs-on: buildjet-4vcpu-ubuntu-2204
     outputs:
       fuel_core_version: ${{steps.get_fuel_core_ver.outputs.version}}
     steps:
@@ -40,7 +40,7 @@ jobs:
           echo "version=$version" >> "$GITHUB_OUTPUT"
 
   build-sway-lib-core:
-    runs-on: ubuntu-latest
+    runs-on: buildjet-4vcpu-ubuntu-2204
     steps:
       - uses: actions/checkout@v3
       - name: Install toolchain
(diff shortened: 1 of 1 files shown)


IGI-111:
As I recall, we kept the standard runners on because the buildjet ones are more expensive. I think we should have a discussion about cost here.

kayagokalp:
> As I recall, we kept the standard runners on because the buildjet ones are more expensive. I think we should have a discussion about cost here.

Sure, the calculation below was my reasoning of opening this PR:

As far as I can see, buildjet runner I used here costs [$0.008 / min](https://buildjet.com/for-github-actions/docs/about/pricing)

And cheapest x86 github runner is also [$0.008/min](https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates-for-standard-runners).
image

That being said I think we have some CI minutes included in our github enterprise subscription, so not sure which makes more sense taking that account

cc @Voxelot

byucesoy:
Hey folks,

Sorry for jumping in, but I saw this thread and wanted to chime in. Did you see Ubicloud runners? They are fully compatible with Github runners and [10x cheaper](https://www.ubicloud.com/use-cases/github-actions). You can for example pick a bigger runner, which would be faster (if the bottleneck is on CPU, memory or disk IO) and cheaper at the same time.

Disclaimer: I'm Ubicloud employee. Feel free to ask any questions you might have.

sdankel (approved):
🚀

JoshuaBatty (approved):

Voxelot:
The buildjet to github runner price comparison isn't apples to apples. While the prices are the same, the buildjet runner is 2x the machine. On the whole build-jet is at least 2x cheaper than github when comparing similar runner specs.

We have a lot of minutes included in the enterprise plan, but those are only for the 2-core runners. I value the productivity of the team's CI cycles more than getting freebie minutes on slow machines.

PR #6660, opened 2024-10-21
Namespace module cleanups

tritao:
## Description

This PR represents another round of cleanups, this time mostly around `Namespace` and `Root`.

## Checklist


sway-core/src/semantic_analysis/namespace/module.rs
```diff
@@ -309,3 +369,62 @@ fn module_not_found(path: &[Ident]) -> CompileError {
...
 }
+
+/// Returns true if the module given by the `absolute_module_path` is external
+/// to the current package. External modules are imported in the `Forc.toml` file.
+pub fn module_is_external(module: &Module, absolute_module_path: &ModulePath) -> bool {
+    assert!(!absolute_module_path.is_empty(), "Absolute module path must have at least one element, because it always contains the package name.");
+    let root_name = module.name();
+    root_name != &absolute_module_path[0]
+}
+
```
jjcnn, line 380: The PR I'm working on moves this method to `Namespace`, because I don't think it makes sense neither in `Root` nor in `Module`.

The problem is that a module can change from internal to external during compilation. For instance, `core::codec` is internal while we're compiling `core`, but once we start compiling `std` it becomes external. In other words, "Is this module external?" is not a meaningful question to ask about a `Module`, because the answer depends on the context in which the question is asked.

The only reason the current solution and the solution here works, is the weird module structure we use where external modules are treated as submodules of each module in the program. I'm changing that structure, and then we do need `module_is_external` to take the context into account.

I'm ok with the change here, but just be aware that the method will be moved again soon.

jjcnn (changes requested):
I'm good with the changes to `type_resolve.rs`, but I would really like if the code in `namespace.rs` and `root.rs` wasn't moved around too much. I'm making a lot of changes in those files, so it's super complicated to keep track of my changes when code is moved to somewhere else.

jjcnn (approved):

JoshuaBatty (approved):

PR #6661, opened 2024-10-22
forc-crypto: vanity address generation

zees-dev:
## Description
Support for Bech32 case-insensitive vanity address generation via `forc-crypto` CLI tool.

Usage:
```sh
forc-crypto vanity --help
```

CLI options:
```
Generate a vanity address

Usage: forc-crypto vanity [OPTIONS]

Options:
      --starts-with <HEX_STRING>  Desired hex string prefix for the address
      --ends-with <HEX_STRING>    Desired hex string suffix for the address
      --regex <PATTERN>           Desired regex pattern to match the entire address (case-insensitive)
      --timeout <SECONDS>         Timeout in seconds for address generation
      --mnemonic                  Return mnemonic with address (default false)
      --save-path <PATH>          Path to save the generated vanity address to
  -h, --help                      Print help
  -V, --version                   Print version

Generate vanity addresses for the Fuel blockchain
```

## Example outputs
```sh
# > forc-crypto vanity --starts-with 0 --ends-with F

Starting to generate vanity address...
Successfully found vanity address in 0.012 seconds.
Address: 0d3c399e756dee9f7312215882f92685fdae25449bc74f33c31063000d68afdf
PrivateKey: cec536d4f5aae64685856ad36818777ba0aed349bdef848b487d6ebb1cc2a0a4
```

```sh
# > forc-crypto vanity --starts-with 0 --ends-with F --mnemonic

Starting to generate vanity address...
Successfully found vanity address in 0.141 seconds.
Address: 030b8484305c9f3af7b662d9fdd88dc75bdceb29f42d0f5ea5f72d3dfaf9380f
Mnemonic: found broccoli trap left thought attack quality smooth patrol enrich fault flavor legend amused monitor shoulder legend blast elbow custom dirt cotton tackle much
PrivateKey: 269ee26dd810a4a2df72969ed8941fe92ddc0bab236715535b867448c3ffa25e
```

## Relevant issues
- https://github.com/FuelLabs/sway/issues/6664


kayagokalp (changes requested):
Hey @zees-dev, Nice job on your first PR!

I will be leaving a second review that actually checks the code but to start with can you:

1. run `cargo clippy`
2. format the `forc-crypto` `Cargo.toml` file, basically the order of dependencies needs to be alphabetical
3. Ensure that root Cargo.lock file is free of conflicts.

kayagokalp:
Also if you use closes #6684. Github automatically marks the relevant issue  to be closed after this PR is merged.

forc-plugins/forc-crypto/src/keys/vanity.rs
```diff
@@ -0,0 +1,440 @@
...
+    pub ends_with: Option<String>,
+
+    /// Return mnemonic with address (default false)
+    #[arg(long)]
+    pub mnemonic: bool,
+
+    /// Path to save the generated vanity address to.
+    #[arg(long, value_hint = clap::ValueHint::FilePath, value_name = "PATH")]
+    pub save_path: Option<PathBuf>,
+}
```
JoshuaBatty, line 38: perhaps a timeout option could be nice? 

```rust
#[arg(long, value_name = "SECONDS")]
pub timeout: Option<u64>,
```

forc-plugins/forc-crypto/src/keys/vanity.rs
```diff
@@ -0,0 +1,440 @@
...
+
+impl VanityMatcher for RegexMatcher {
+    fn is_match(&self, addr: &Address) -> bool {
+        let addr = hex::encode(addr.as_ref());
+        self.left.is_match(&addr) && self.right.is_match(&addr)
+    }
+}
+
+fn parse_pattern(pattern: &str, is_start: bool) -> Result<Either<Vec<u8>, Regex>> {
+    if let Ok(decoded) = hex::decode(pattern) {
```
JoshuaBatty, line 275: Perhaps we could validate the the pattern length is no longer than 64

```rust
if pattern.len() > 64 { 
    return Err(anyhow!("Pattern too long: max 64 characters"));
}
````

File: forc-plugins/forc-crypto/src/keys/vanity.rs [hunk 3]
```diff
@@ -0,0 +1,440 @@
...
+    wallet_generator(use_mnemonic)
+        .find_any(|result| {
+            if let Ok((addr, _, _)) = result {
+                matcher.is_match(addr)
+            } else {
+                false
+            }
+        })
+        .ok_or_else(|| anyhow!("Failed to generate a matching address"))?
+}
```
JoshuaBatty, line 146: What are your thoughts about having some user feedback? something like...

```rust
pub fn find_vanity_address<T: VanityMatcher>(
    matcher: T, 
    use_mnemonic: bool,
) -> Result<(Address, SecretKey, Option<String>)> {
    let start = Instant::now();
    let mut attempts = AtomicUsize::new(0); // Need atomic for parallel iteration
    
    let result = wallet_generator(use_mnemonic)
        .for_each(|_| {
            let current = attempts.fetch_add(1, Ordering::Relaxed);
            if current % 10000 == 0 { // Pick a nice number that makes sense here, just going with 10000 
                let elapsed = start.elapsed().as_secs_f64();
                let rate = current as f64 / elapsed;
                println!("Tried {} addresses ({:.2} addresses/sec)...", current, rate);
            }
        })
        .find_any(|result| {
            if let Ok((addr, _, _)) = result {
                matcher.is_match(addr)
            } else {
                false
            }
        })
        .ok_or_else(|| anyhow!("Failed to generate a matching address"))?;

    Ok(result?)
}
```

JoshuaBatty (commented):
Just a couple of suggestions. Great to see the benchmarking and tests added for this.

forc-plugins/forc-crypto/src/keys/vanity.rs
zees-dev: this can be a good idea; the `solana-keygen` seems to do this.

forc-plugins/forc-crypto/src/keys/vanity.rs
```diff
@@ -0,0 +1,440 @@
...
+        Ok(Either::Left(decoded))
+    } else {
+        // Check if the pattern contains only valid hex characters
+        if !pattern
+            .chars()
+            .all(|c| c.is_ascii_hexdigit() || c == '^' || c == '$')
+        {
+            return Err(anyhow!("Invalid hex pattern: {}", pattern));
+        }
+
```
alfiedotwtf, line 288: From here we know it's a regex. Is there any need to touch it futher? i.e if the user is dropping down to a regex should we just let them match for whatever they're looking for and from here just pass it through?

forc-plugins/forc-crypto/src/keys/vanity.rs
```diff
@@ -0,0 +1,440 @@
...
+    if let Some(suffix) = ends_with {
+        match parse_pattern(&suffix, false)? {
+            Either::Left(right) => right_exact_hex = Some(right),
+            Either::Right(re) => right_regex = Some(re),
+        }
+    }
+
+    println!("Starting to generate vanity address...");
+    let start_time = Instant::now();
+
```
alfiedotwtf, line 68: If the user is using a regex, should we let them match whatever they're looking for rather than only applying it to the start or end? i.e should this instead handle:

- left
- right
- left and right
- regex

The reason I'm asking is that if they're using a regex, they could put the left or right or both IN the regex.

alfiedotwtf:
Nice!

Quick question - is there a big difference in timings between random vs mnemonic? I'm thinking it would always be a good idea to provide the mnemonic by default and if they're close in timing we should only provide that pathway... but if the performance of random is way faster, then yeah let's keep both.

zees-dev:
> Nice!
> 
> Quick question - is there a big difference in timings between random vs mnemonic? I'm thinking it would always be a good idea to provide the mnemonic by default and if they're close in timing we should only provide that pathway... but if the performance of random is way faster, then yeah let's keep both.

There is a difference in timing; quite significant actually; here is benchmark:
```
#> cargo bench -p forc-crypto

     Running benches/bench_main.rs (target/release/deps/bench_main-77704bc04ac1c0dc)
Gnuplot not found, using plotters backend
Vanity Address Generation/Vanity Address (first byte 0xff)
                        time:   [656.07 µs 689.08 µs 731.09 µs]
                        change: [-6.1118% -0.1898% +7.1988%] (p = 0.96 > 0.05)
                        No change in performance detected.
Found 8 outliers among 100 measurements (8.00%)
  2 (2.00%) low mild
  1 (1.00%) high mild
  5 (5.00%) high severe
Benchmarking Vanity Address Generation/Vanity Address with Mnemonic (first byte 0xff): Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.6s, or reduce sample count to 80.
Vanity Address Generation/Vanity Address with Mnemonic (first byte 0xff)
                        time:   [46.216 ms 55.127 ms 64.789 ms]
                        change: [-0.4164% +22.910% +51.685%] (p = 0.07 > 0.05)
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
  ```

File: forc-plugins/forc-crypto/src/keys/vanity.rs (hunk 5 above)
Comment (zees-dev) [reply forc-plugins/forc-crypto/src/keys/vanity.rs:68]: That's correct. However, in most cases, users typically choose a prefix and/or suffix based on characters, as this is what wallet UIs commonly display.
For a better user experience, I believe it's best to offer prefix/suffix options, since users are less likely to remember regex syntax offhand.

File: forc-plugins/forc-crypto/src/keys/vanity.rs (hunk 3 above)
Comment (zees-dev) [reply forc-plugins/forc-crypto/src/keys/vanity.rs]: addressed: https://github.com/FuelLabs/sway/pull/6661/commits/e493891bc22b376d7175321e1367b2342587896f

File: forc-plugins/forc-crypto/src/keys/vanity.rs (hunk 2 above)
Comment (zees-dev) [reply forc-plugins/forc-crypto/src/keys/vanity.rs]: addressed: https://github.com/FuelLabs/sway/pull/6661/commits/b16c926bada977a31c774bbd9197f969f3e8c516

File: forc-plugins/forc-crypto/src/keys/vanity.rs (hunk 1 above)
Comment (zees-dev) [reply forc-plugins/forc-crypto/src/keys/vanity.rs]: addressed: https://github.com/FuelLabs/sway/pull/6661/commits/723db3ee288d309baab22f151b656246c955bb27

File: forc-plugins/forc-crypto/src/keys/vanity.rs [hunk 6]
```diff
@@ -0,0 +1,551 @@
...
+    }
+
+    if let Ok(decoded) = hex::decode(pattern) {
+        Ok(Either::Left(decoded))
+    } else {
+        // Check if the pattern contains only valid hex characters
+        if !pattern
+            .chars()
+            .all(|c| c.is_ascii_hexdigit() || c == '^' || c == '$')
+        {
```
sdankel, line 345: We could put this in a clap validator function for the arguments, so the error comes from clap, as well as the length validation.

forc-plugins/forc-crypto/src/keys/vanity.rs
```diff
@@ -0,0 +1,551 @@
...
+pub enum VanityAddressError {
+    Timeout(u64),
+    GenerationFailed,
+    InvalidPattern(String),
+}
+
+impl fmt::Display for VanityAddressError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Timeout(secs) => write!(
```
sdankel, line 372: We use the `tracing` library in forc and forc plugins to make it easier to silence output or provide verbose output in the CLIs. These `write!`s can be replaced with `tracing::error!`

forc-plugins/forc-crypto/src/keys/vanity.rs
```diff
@@ -0,0 +1,551 @@
...
+        let start = Instant::now();
+        let attempts = std::sync::atomic::AtomicUsize::new(0); // atomic for parallel iteration
+        wallet_generator(use_mnemonic)
+            .find_any(|result| {
+                let current = attempts.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+                if current != 0 && current % breakpoint == 0 {
+                    let elapsed = start.elapsed().as_secs_f64();
+                    let rate = current as f64 / elapsed;
+                    println!(
+                        "└─ tried {} addresses ({:.2} addresses/sec)...",
```
sdankel, line 164: If you want to make it look extra cool, you could use [indicatif](https://github.com/console-rs/indicatif/tree/main) and make this statement update the address count and rate update on the same line, similar to [this example](https://github.com/console-rs/indicatif/blob/main/examples/yarnish.rs#L84)

forc-plugins/forc-crypto/src/keys/vanity.rs
```diff
@@ -0,0 +1,551 @@
...
+        .into());
+    }
+
+    if let Ok(decoded) = hex::decode(pattern) {
+        Ok(Either::Left(decoded))
+    } else {
+        // Check if the pattern contains only valid hex characters
+        if !pattern
+            .chars()
+            .all(|c| c.is_ascii_hexdigit() || c == '^' || c == '$')
```
sdankel, line 344: Since we aren't supporting other regex functionality like globs, perhaps we should only support string matching (without '^' or '$') and update the help text?

forc-plugins/forc-crypto/src/keys/vanity.rs
```diff
@@ -0,0 +1,551 @@
...
+#[derive(Debug, clap::Parser)]
+#[clap(
+    version,
+    about = "Generate a vanity address",
+    after_help = "Generate vanity addresses for the Fuel blockchain"
+)]
+pub struct Arg {
+    /// Prefix regex pattern or hex string.
+    #[arg(long, value_name = "PATTERN", required_unless_present = "ends_with")]
+    pub starts_with: Option<String>,
```
sdankel, line 30: Related to my other comment, since we aren't really supporting regex I think this should be:

```suggestion
    /// Desired hex sting prefix.
    #[arg(long, value_name = "HEX_STRING", required_unless_present = "ends_with")]
    pub starts_with: Option<String>,
```

And the same for ends_with.

forc-plugins/forc-crypto/src/lib.rs
```diff
@@ -0,0 +1,36 @@
+pub mod address;
+pub mod args;
+pub mod keccak256;
+pub mod keys;
+pub mod sha256;
+
+pub(crate) fn help() -> &'static str {
```
sdankel, line 7: The other forc plugins in this repo use the `forc_util::cli_examples` macro. It might be good to use that here for consistency.

forc-plugins/forc-crypto/src/keys/vanity.rs
```diff
@@ -0,0 +1,551 @@
...
+        (None, Some(re), None, None) => {
+            let matcher = SingleRegexMatcher { re };
+            find_vanity_address_with_timeout(matcher, mnemonic, timeout)
+        }
+        (None, None, None, Some(re)) => {
+            let matcher = SingleRegexMatcher { re };
+            find_vanity_address_with_timeout(matcher, mnemonic, timeout)
+        }
+        _ => {
+            return Err(VanityAddressError::InvalidPattern(
```
sdankel, line 113: nit: I think this should be a separate error, `InvalidPatternCombination`, since it is different and it will make the error output a little cleaner.

forc-plugins/forc-crypto/src/keys/vanity.rs
JoshuaBatty: Oh nice, indicatif looks really cool!

forc-plugins/forc-crypto/src/keys/vanity.rs
zees-dev: somewhat addressed by the refactor: https://github.com/FuelLabs/sway/pull/6661/commits/59ab719069be67c3a2ef209e632c2d5039996746

forc-plugins/forc-crypto/src/keys/vanity.rs
zees-dev: addressed: https://github.com/FuelLabs/sway/pull/6661/commits/59ab719069be67c3a2ef209e632c2d5039996746

forc-plugins/forc-crypto/src/keys/vanity.rs
zees-dev: addressed: https://github.com/FuelLabs/sway/pull/6661/commits/6746af46c48b9873e9a649fee075e7147e6069fd

forc-plugins/forc-crypto/src/lib.rs
zees-dev: cannot find an example specifically for top-level help; also this code remains unchanged as it was migrated from `main.rs`.

forc-plugins/forc-crypto/src/keys/vanity.rs
zees-dev: addressed: https://github.com/FuelLabs/sway/pull/6661/commits/59ab719069be67c3a2ef209e632c2d5039996746

forc-plugins/forc-crypto/src/keys/vanity.rs
zees-dev: addressed: https://github.com/FuelLabs/sway/pull/6661/commits/80ca7a2ce4a87e0c8a6038ff69defc3e07af117d

forc-plugins/forc-crypto/src/keys/vanity.rs
zees-dev: removed custom error as no longer needed as it did not see much usage; reverting back to anyhow: https://github.com/FuelLabs/sway/pull/6661/commits/80ca7a2ce4a87e0c8a6038ff69defc3e07af117d

forc-plugins/forc-crypto/src/keys/vanity.rs
zees-dev: removed custom error as no longer needed as it did not see much usage; reverting back to anyhow: https://github.com/FuelLabs/sway/commit/80ca7a2ce4a87e0c8a6038ff69defc3e07af117d

forc-plugins/forc-crypto/src/keys/vanity.rs
zees-dev: had a look; cant think of a great UI; also this dependency is not currently in the project.
Maybe a future enhancement.

Cargo.lock
```diff
@@ -2765,19 +2765,24 @@ dependencies = [
  "async-trait",
  "atty",
  "clap 4.5.20",
+ "criterion",
```
sdankel, line 2768: Should this be a dev dependency?

forc-plugins/forc-crypto/src/keys/vanity.rs
```diff
@@ -0,0 +1,469 @@
...
+}
+
+fn validate_regex_pattern(s: &str) -> Result<String, String> {
+    if s.len() > 128 {
+        return Err("Regex pattern too long: max 128 characters".to_string());
+    }
+
+    if s.chars()
+        .any(|c| c.is_ascii() && c.to_ascii_lowercase() >= 'h' && c.to_ascii_lowercase() <= 'z')
+    {
```
alfiedotwtf, line 39: I don't think this is needed since you're going to return error if `Regex::new()` fails. This will also fail with character classes e.g `--regex "[[:alpha:]]+"`

alfiedotwtf (dismissed):

forc-plugins/forc-crypto/src/keys/vanity.rs
zees-dev: addressed: https://github.com/FuelLabs/sway/pull/6661/commits/5a850d2f485b61a3ef3b1febd1432a9813005253

Cargo.lock
zees-dev: yeap this is declared as dev dependency in `forc-crypto` `Cargo.toml`

sdankel (approved):
Looks good! :shipit:

kayagokalp (dismissed):
Looks good to me!

kayagokalp:
I think only thing left in CI is to fix the tests generated by `forc_util::cli_examples!` macro.

kayagokalp (approved):

JoshuaBatty (approved):
Great work!

PR #6672, opened 2024-10-24
chore: bump rust version to 1.82

kayagokalp:
## Description
Bumps sway version used in CI to v1.82

### DIFF
--- a/sway-core/src/asm_generation/fuel/programs/abstract.rs
+++ b/sway-core/src/asm_generation/fuel/programs/abstract.rs
@@ -164,13 +164,13 @@ impl AbstractProgram {
     /// Right now, it looks like this:
     ///
     /// WORD OP
-    /// 1    MOV $scratch $pc
-    /// -    JMPF $zero i2
-    /// 2    DATA_START (0-32) (in bytes, offset from $is)
-    /// -    DATA_START (32-64)
-    /// 3    LW $ds $scratch 1
-    /// -    ADD $ds $ds $scratch
-    /// 4    .program_start:
+    /// [1]    MOV $scratch $pc
+    /// [-]    JMPF $zero i2
+    /// [2]    DATA_START (0-32) (in bytes, offset from $is)
+    /// [-]    DATA_START (32-64)
+    /// [3]    LW $ds $scratch 1
+    /// [-]    ADD $ds $ds $scratch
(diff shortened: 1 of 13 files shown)


kayagokalp:
We have couple of warnings around using keys with interior mutability. 

This is bad because ideally `HashMap` key shouldn't be changing so that's why we have the lint in place. 

But the lint has known false positives for types that has interior mutability but doesn't access that interior mutable type for its `Hash` or `Ord` implementation.

We need to check whether it is the case for our case and if that is the case we should deny the warning. 

https://rust-lang.github.io/rust-clippy/master/index.html#mutable_key_type

JoshuaBatty (approved):

alfiedotwtf:
> We have couple of warnings around using keys with interior mutability.

When I originally did #6501 I ran into this and had 99% of the changes, but those were dropped when other priorities came around. I'll put it on my list as something to do for a background task.

sway-core/src/asm_generation/fuel/programs/abstract.rs
```diff
@@ -164,13 +164,13 @@ impl AbstractProgram {
...
-    /// 3    LW $ds $scratch 1
-    /// -    ADD $ds $ds $scratch
-    /// 4    .program_start:
+    /// [1]    MOV $scratch $pc
+    /// [-]    JMPF $zero i2
+    /// [2]    DATA_START (0-32) (in bytes, offset from $is)
+    /// [-]    DATA_START (32-64)
+    /// [3]    LW $ds $scratch 1
+    /// [-]    ADD $ds $ds $scratch
+    /// [4]    .program_start:
```
alfiedotwtf, line 173: I couldn't get this to render, so I tried the following which seemed to workm quite nicely:

```
    /// |OFFSET|WORD OP                                      |
    /// |------|---------------------------------------------|
    /// |1     |MOV $scratch $pc                             |
    /// |1.5   |JMPF $zero i2                                |
    /// |2     |DATA_START (0-32) (in bytes, offset from $is)|
    /// |      |DATA_START (32-64)                           |
    /// |3     |LW $ds $scratch 1                            |
    /// |3.5   |ADD $ds $ds $scratch                         |
    /// |4     |.program_start:                              |
```

alfiedotwtf (approved):

PR #6674, opened 2024-10-25
feat: add function for generating bytecode identifier

sdankel:
## Description

Depends on https://github.com/FuelLabs/sway/pull/6522
Related https://github.com/FuelLabs/forc.pub/issues/16

Adds a function, `get_bytecode_id` that generates a sha256 hash of the bytecode with the configurables section of the bytecode removed. This will be used for indexing and lookups of the corresponding ABIs for contracts, predicates, and scripts in the package registry (forc.pub). 

## Checklist


forc-util/src/bytecode.rs
```diff
@@ -0,0 +1,145 @@
...
+    let instructions = {
+        let buffer = Box::leak(buffer.into_boxed_slice()); // Leak the buffer to extend its lifetime
+        fuel_asm::from_bytes(buffer.iter().cloned()).zip(buffer.chunks(fuel_asm::Instruction::SIZE))
+    };
+
+    Ok(instructions)
+}
+
+// TODO: use constants for configurables section offset
+
```
sdankel, line 37:
```suggestion
```

forc-util/src/bytecode.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+/// Gets the bytecode ID from a bytecode file. The bytecode ID is the hash of the bytecode after removing the
+/// condigurables section, if any.
+pub fn get_bytecode_id<P>(path: P) -> anyhow::Result<String>
+where
+    P: AsRef<Path> + Clone,
+{
+    let mut instructions = parse_bytecode_to_instructions(path.clone())?;
+
```
alfiedotwtf, line 43: 4 vs 6:

```
// Collect the first four instructions...
let mut first_six_instructions...
```

forc-util/src/bytecode.rs
```diff
@@ -0,0 +1,143 @@
...
+    // Collect the first four instructions into a temporary vector
+    let mut first_six_instructions = Vec::with_capacity(6);
+    for _ in 0..6 {
+        if let Some(instruction) = instructions.next() {
+            first_six_instructions.push(instruction);
+        } else {
+            return Err(anyhow!("Incomplete bytecode"));
+        }
+    }
```
alfiedotwtf, line 52: While perfectly fine, not sure if something like the following is tidier:

```
    let first_six_instructions = instructions.iter().take(6).copied().collect();
    ensure!(first_six_instructions.len() == 6, "Incomplete bytecode");
```

forc-util/src/bytecode.rs
```diff
@@ -0,0 +1,143 @@
...
+    let (hi_instr, hi_raw) = &first_six_instructions[5];
+
+    if let Err(fuel_asm::InvalidOpcode) = lo_instr {
+        if let Err(fuel_asm::InvalidOpcode) = hi_instr {
+            // Now assemble the configurables offset.
+            let configurables_offset = usize::from_be_bytes([
+                low_raw[0], low_raw[1], low_raw[2], low_raw[3], hi_raw[0], hi_raw[1], hi_raw[2],
+                hi_raw[3],
+            ]);
+
```
alfiedotwtf, line 64: Should we actually be looking at 4 and 5? Is there a chance someone could craft valid instructions at offsite 4 and 5? Maybe strictly stick to 0-4. Then these `if` can be removed too.

alfiedotwtf (dismissed):

forc-util/src/bytecode.rs
sdankel: `copied` doesn't work because of the static lifetime on the raw chunks. Plus, I think it's preferable to do it without copying the iterator.

forc-util/src/bytecode.rs
alfiedotwtf: Ah, ok cool. That makes sense then.

alfiedotwtf (dismissed):

forc-util/src/bytecode.rs
```diff
@@ -0,0 +1,143 @@
...
+{
+    let mut f = File::open(path.clone())
+        .map_err(|_| anyhow!("{}: file not found", path.as_ref().to_string_lossy()))?;
+    let metadata = fs::metadata(path.clone())
+        .map_err(|_| anyhow!("{}: file not found", path.as_ref().to_string_lossy()))?;
+    let mut buffer = vec![0; metadata.len() as usize];
+    f.read_exact(&mut buffer).expect("buffer overflow");
+
+    let instructions = {
+        let buffer = Box::leak(buffer.into_boxed_slice()); // Leak the buffer to extend its lifetime
```
JoshuaBatty, line 29: do we need to return a reference to bytes? or can we return `Vec<u8>` and remove the memory leak?

forc-util/src/bytecode.rs
```diff
@@ -0,0 +1,143 @@
...
+/// condigurables section, if any.
+pub fn get_bytecode_id<P>(path: P) -> anyhow::Result<String>
+where
+    P: AsRef<Path> + Clone,
+{
+    let mut instructions = parse_bytecode_to_instructions(path.clone())?;
+
+    // Collect the first six instructions into a temporary vector
+    let mut first_six_instructions = Vec::with_capacity(6);
+    for _ in 0..6 {
```
JoshuaBatty, line 46: should we consider using constants for some of these values?

```rust
const CONFIGURABLES_HEADER_SIZE: usize = 6;
```

forc-util/src/bytecode.rs
```diff
@@ -0,0 +1,143 @@
...
+
+            // Hash the first six instructions
+            let mut hasher = Sha256::new();
+            for (_, raw) in first_six_instructions {
+                hasher.update(raw);
+            }
+
+            // Continue hashing the remaining instructions up to the configurables section offset.
+            instructions
+                .take(configurables_offset / 4 - 6) // Minus 6 because we already hashed the first six
```
JoshuaBatty, line 73: same here, maybe some named constants would be better that numbers

vaivaswatha (dismissed):

forc-util/src/bytecode.rs
sdankel: Good catch. When I first read the docs for Box::leak I misunderstood it and thought that it would drop the inner object as long as it's in a Box, but I guess that only happens if the inner Box is created with [Box::from_raw](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.from_raw). 

Even if I return a Vec<u8>, there's still a problem when `zip`ing the buffer, since buffer doesn't live long enough. I'll make the zip happen outside of the function to avoid using Box::leak.

sdankel: I ended up writing a custom iterator that takes a `BufReader`. This should be more efficient than reading the whole binary into a buffer at once, and avoids the problem where the buffer was not owned by the iterator.

sdankel:
This is currently blocked because the related compiler change was reverted.

vaivaswatha (approved):

JoshuaBatty (approved):
Nice! Appreciate all the comments done here as well.

zees-dev:
Great work!

Do you think it would be worthwhile having the source `.sw` files for the contracts aswell as the binary files?
I can potentially imagine a scenario where we could have compiler changes, and hence the output binaries may not be regeneratable.
Having the source alongside would potentially help resolve such future test failures i'd imagine.

sdankel:
> Great work!
> 
> Do you think it would be worthwhile having the source `.sw` files for the contracts aswell as the binary files? I can potentially imagine a scenario where we could have compiler changes, and hence the output binaries may not be regeneratable. Having the source alongside would potentially help resolve such future test failures i'd imagine.

Thanks for bringing this up. I thought about having the tests generate the binary files from the source, but thought that it would be annoying to have to update the bytecode IDs in the tests whenever the compiler changes.

I figured that if the tests aren't regenerating the binary from source, there was no need to include the source, since the source projects are already checked in under `examples/`

When we need to regenerate the binary files in the future, it would just be a matter of building those two examples with `forc build` and `forc build --release` and updating the bytecode IDs in the tests. Let me know what you think.

zees-dev:
> I figured that if the tests aren't regenerating the binary from source, there was no need to include the source, since the source projects are already checked in under `examples/`

Didn't know the the source is already in examples; in that case no need for another source.
Agree that compilation would still be issue regardless - if compiler changes.

Just thinking from the point of someone else having to fix the tests if they break.
A comment in the test(s) themselves could help (point to examples).
Nevertheless will leave this upto you.

sdankel:
> Just thinking from the point of someone else having to fix the tests if they break. A comment in the test(s) themselves could help (point to examples). Nevertheless will leave this upto you.

Good idea. I'll add a comment in a separate PR since this one is pretty much ready to go.

Issue #6259, opened 2024-07-11
[Tracking] forc-client techincal debts


kayagokalp:
forc-client is a very vibrant section of our codebase, and a lot of changes land there regularly, and it is getting expanded as the time pass. I think it is time for us to have a nice cleaning done, so that we have stronger foundations in place.

1. Let's try to de-duplicate efforst between sdk and forc-client as much as possible, now that sdk has a estimation flow in place we can remove our code and use theirs
2. Folder/file structure can be improved a little: forc-client is expanding all the time and some files can be split or could use same new names.
3. As part of #6069  we have a nice improvement in testability of the crate, we should continue this effort and extend our test suite.
4. We should start following the code coverage for forc-client.

PR #6680, opened 2024-10-31
Forc-deploy import wallet support if non-existent wallet

zees-dev:
## Description

- Adds support for importing a wallet on `forc-deploy`
- Resolves https://github.com/FuelLabs/forc-wallet/issues/139

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,7 +101,7 @@ fuel-tx = "0.58"
 fuel-vm = "0.58"
 
 # Dependencies from the `forc-wallet` repository:
-forc-wallet = "0.11"
+forc-wallet = "0.11.1"
 
 #
 # External dependencies
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,7 +101,7 @@ fuel-tx = "0.58"
 fuel-vm = "0.58"
 
 # Dependencies from the `forc-wallet` repository:
-forc-wallet = "0.11.1"
+forc-wallet = "0.11"
 
 #
 # External dependencies
--- a/forc-plugins/forc-client/src/util/tx.rs
+++ b/forc-plugins/forc-client/src/util/tx.rs
@@ -11,6 +11,7 @@ use forc_wallet::{
         collect_accounts_with_verification, AccountBalances, AccountVerification, AccountsMap,
     },
     new::{new_wallet_cli, New},
+    import::{import_wallet_cli, Import},
     utils::default_wallet_path,
 };
 use fuel_crypto::SecretKey;
@@ -45,6 +46,15 @@ fn ask_user_yes_no_question(question: &str) -> Result<bool> {
     Ok(answer)
 }
 
+fn ask_user_with_options(question: &str, options: &[&str], default: usize) -> Result<usize> {
+    let selection = Select::with_theme(&ColorfulTheme::default())
+        .with_prompt(question)
+        .items(options)
+        .default(default)
+        .interact()?;
+    Ok(selection)
+}
+
 fn collect_user_accounts(
     wallet_path: &Path,
     password: &str,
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -101,7 +101,7 @@ fuel-tx = "0.58"
 fuel-vm = "0.58"
 
 # Dependencies from the `forc-wallet` repository:
-forc-wallet = "0.11"
+forc-wallet = "0.11.1"
```
alfiedotwtf, line 105: We're currently using only the first two numbers in SemVer.

forc-plugins/forc-client/src/util/tx.rs
```diff
@@ -71,20 +81,29 @@ pub(crate) fn prompt_forc_wallet_password() -> Result<String> {
...
+        let wallet_options = ask_user_with_options(
+            &question,
+            &["Create new wallet", "Import existing wallet"],
+            0,
+        )?;
+        match wallet_options {
+            0 => {
+                new_wallet_cli(wallet_path, New { force: false, cache_accounts: None })?;
+                println!("Wallet created successfully.");
+                // Derive first account for the fresh wallet we created.
```
alfiedotwtf, line 95: Not sure if it's worth it to change or keep, up to you for taste, but these 3 lines here and in the next branch arm can be done after the `match` since the `bail` won't do them.

alfiedotwtf (dismissed):

forc-plugins/forc-client/src/util/tx.rs
zees-dev: addressed: https://github.com/FuelLabs/sway/pull/6680/commits/9c042b7be80452fadf2aa45d27dccacc18923ac5

Cargo.toml
kayagokalp: This is a common problem we have right now with only using "X.Y" versions. But here is the solution:

use `cargo update -p forc-wallet` :)

More explanation of the problem for more context:

## The Problem

Once pinned to "x.y" there is no way to incorporate minor bumps without running `cargo update` or first pinning to "x.y.z" and building the workspace and reverting back to "x.y". This is because `cargo` is not checking for a minor versions at each `cargo build` operation. 

This wouldn't be a problem if we were running cargo update in each release, but we are only running cargo update on each major bump of sway repo releases.


## Solution

So we need a way to only update specific dependencies, luckily cargo-update has a `-p` parameter to use with this. You can specifically only bump a single dependency if used like:

`cargo update -p forc-wallet`


## A way to improve

I think we can write a script like update_fuel_dependencies.sh which will essentially run `cargo update -p <fuel_dep_name>`, this can further be added to a git hook. Since we all adhere to semver no minor bump should break any code 🤞

alfiedotwtf (approved):

sdankel (changes requested):
The prompt should come before asking for the wallet password.

image

zees-dev:
> The prompt about creating/importing a wallet should come before asking for the wallet password. Otherwise, once the tests are passing, LGTM 😃

The current behaviour performs the password check which is much more upstream in the code; changing that would imply a re-factor which may have knock-on effects.
I think this would be worthwhile addressing as a seperate issue.
If we have agreement; I can create an issue for that.

kayagokalp:
> > The prompt about creating/importing a wallet should come before asking for the wallet password. Otherwise, once the tests are passing, LGTM 😃
> 
> The current behaviour performs the password check which is much more upstream in the code; changing that would imply a re-factor which may have knock-on effects. I think this would be worthwhile addressing as a seperate issue. If we have agreement; I can create an issue for that.

I think we should discuss the necessary refactor to do this, because it feels a little confusing as it is. If i don't have an wallet it is highly likely that I don't have necessary context to understand what password I need to enter once that is prompted. So it might be worth doing the refactor before merging this. 

If you can come up with some kind of a description of what you need to be changed exactly I am more than happy to help out! I want to clean forc deploy paths for some time now even have a tracking issue for it. #6259

zees-dev:
> > > The prompt about creating/importing a wallet should come before asking for the wallet password. Otherwise, once the tests are passing, LGTM 😃
> > 
> > 
> > The current behaviour performs the password check which is much more upstream in the code; changing that would imply a re-factor which may have knock-on effects. I think this would be worthwhile addressing as a seperate issue. If we have agreement; I can create an issue for that.
> 
> I think we should discuss the necessary refactor to do this, because it feels a little confusing as it is. If i don't have a wallet it is highly likely that I don't have necessary context to understand what password I need to enter once that is prompted. So it might be worth doing the refactor before merging this.
> 
> If you can come up with some kind of a description of what you need to be changed exactly I am more than happy to help out! I want to clean forc deploy paths for some time now even have a tracking issue for it. #6259

I believe this should be done in another PR as the scope could become much larger.
In this PR we just want to give option for user to import seed phrase; without changing/refactoring the upstream flow.

kayagokalp:
> > > > The prompt about creating/importing a wallet should come before asking for the wallet password. Otherwise, once the tests are passing, LGTM 😃
> > > 
> > > 
> > > The current behaviour performs the password check which is much more upstream in the code; changing that would imply a re-factor which may have knock-on effects. I think this would be worthwhile addressing as a seperate issue. If we have agreement; I can create an issue for that.
> > 
> > 
> > I think we should discuss the necessary refactor to do this, because it feels a little confusing as it is. If i don't have a wallet it is highly likely that I don't have necessary context to understand what password I need to enter once that is prompted. So it might be worth doing the refactor before merging this.
> > If you can come up with some kind of a description of what you need to be changed exactly I am more than happy to help out! I want to clean forc deploy paths for some time now even have a tracking issue for it. #6259
> 
> I believe this should be done in another PR as the scope could become much larger. In this PR we just want to give option for user to import seed phrase; without changing/refactoring the upstream flow.

I don't think we should merge this as it is because as it is this will be confusing, my suggestion would be to keep this ready to go and first fix the upstream and after that using the already fixed upstream in this PR

kayagokalp (changes requested):
It turns out this situation is present for creating a new wallet as well. I opened #6704, once merged it will also fix the problem we are having here. If we can merge that and rebase this one wallet password won't be prompted.

kayagokalp (approved):

PR #6684, opened 2024-11-01
chore: add a script to bump all fuel maintained dependencies

kayagokalp:
## Description

Adds a bash script which iterates over all fuel owned/maintained crates we depend on and only runs cargo update for them. 

This is especially useful for incorporating minor bumps which happens much more often.

### DIFF
--- a/update_fuel_dependencies.sh
+++ b/update_fuel_dependencies.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# Run this script to only bump fuel maintained dependencies. 
+#
+# We currently pin dependencies using "X.Y" in the root Cargo.toml file.
+# Since `cargo build` does not check for minor version bumps at each invocation
+# it is hard to move between new versions, especially for minor bumps which
+# happens much often. Use this script to keep every fuel owned dependency up to
+# date.
+
+# Define the list of fuel maintained crates
+crates=(
+    "fuel-abi-types"
+    "fuel-core-client"
+    "fuel-core-types"
+    "fuels"
+    "fuels-core"
+    "fuels-accounts"
+    "fuel-asm"
+    "fuel-crypto"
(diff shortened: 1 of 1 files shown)


kayagokalp:
We can also add `--patch-only` to ensure we only update to latest patch release instead of latest available one.

update_fuel_dependencies.sh
```diff
@@ -0,0 +1,31 @@
...
+    "fuel-tx"
+    "fuel-vm"
+    "forc-wallet"
+)
+
+# Run `cargo update -p <crate_name>` for each fuel owned crate. 
+for crate in "${crates[@]}"; do
+    echo "Updating package: $crate"
+    cargo update -p "$crate"
+done
```
alfiedotwtf, line 31: Weird... `cargo update` doesn't seem to document `-p` but it's the same as `--package <SPEC>`, but having a look `--package` isn't documented either in `cargo` 🙃

alfiedotwtf (approved):

JoshuaBatty (approved):

update_fuel_dependencies.sh
sdankel: It looks like `cargo update -p "$crate"` is equivalent to `cargo update "$crate"`

kayagokalp: Should we prefer using non -p version? I don't think it is important but it is a little weird that -p is not documented in the original documentation. I found out about this param in a mirror doc of cargo update hosted by some third party. 

I am not sure where cargo-update is hosted. I see a repo named cargo-update under non-rust lang organization 🤷 

I think this will be fine, and possibly not documented due to lack of documentation efforts in that area. Resolving this for now feel free to ping me if anyone thinks non -p version should be preferred here 🙌

PR #6693, opened 2024-11-04
refac: Use offset from SDK for executable loaders

### LINKED ISSUE #6683: Use offset from SDK for executable loaders instead of re-calculating it
## Description

Loader offset is exposed by sdk as of https://github.com/FuelLabs/fuels-rs/pull/1522 currently we are recalculating this value in forc-deploy because at the time of the implementation this was not exposed by sdk. This is bad because if anything changes around that area, we need to update it manually. We should use the exposed value from sdk instead of recalculating.

https://github.com/FuelLabs/fuels-rs/pull/1522

### Context
-------
https://github.com/FuelLabs/sway/blob/b6bbbf8bf5939eaf2c965b4b94732d43ef656500/forc-plugins/forc-client/src/op/deploy.rs#L357-L360

## Solution

Basically remove the function showed in context, and get the exposed value [from SDK](https://github.com/FuelLabs/fuels-rs/pull/1522) directly.

Abeeujah:
## Description
Uses the offset exposed by the SDK instead of recalculating it and having to manually update the code each time the implementation changes.

closes #6683 

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2072,12 +2072,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "56254986775e3233ffa9c4d7d3faaf6d36a2c09d30b20687e9f88bc8bafc16c8"
 
 [[package]]
-name = "difference"
-version = "2.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "524cbf6897b527295dff137cec09ecf3a05f4fddffd7dfcd1585403449e74198"
-
-[[package]]
 name = "digest"
 version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -7807,7 +7801,6 @@ name = "swayfmt"
 version = "0.66.4"
 dependencies = [
  "anyhow",
- "difference",
  "forc-tracing 0.66.4",
  "indoc",
  "paste",
@@ -7815,6 +7808,7 @@ dependencies = [
  "ropey",
  "serde",
  "serde_ignored",
+ "similar",
  "sway-ast",
  "sway-core",
  "sway-error",
--- a/swayfmt/Cargo.toml
+++ b/swayfmt/Cargo.toml
@@ -25,7 +25,7 @@ thiserror.workspace = true
 toml = { workspace = true, features = ["parse"] }
 
 [dev-dependencies]
-difference = "2.0.0"
 paste = "1.0"
 prettydiff = "0.6"
+similar = "2.0"
 test-macros = { path = "test_macros" }
--- a/swayfmt/test_macros/src/lib.rs
+++ b/swayfmt/test_macros/src/lib.rs
@@ -119,18 +119,23 @@ macro_rules! fmt_test_inner {
 #[macro_export]
 macro_rules! assert_eq_pretty {
     ($got:expr, $expected:expr) => {
-        let got = &$got;
-        let expected = &$expected;
+        let got = &$got[..];
+        let expected = &$expected[..];
+
         if got != expected {
-            use difference::{Changeset, Difference};
-            let changeset = Changeset::new(expected, got, "\n");
-            for diff in changeset.diffs {
-                match diff {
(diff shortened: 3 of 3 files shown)


Abeeujah:
Hi @kayagokalp @Br1ght0ne @Salka1988 @hal3e please can you review?

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -482,28 +382,25 @@ pub async fn deploy_executables(
...
+                    .into_iter()
+                    .map(|config| Configurable {
+                        offset: config.offset - offset_shift as u64,
+                        ..config.clone()
+                    })
+                    .collect()
+            });
+            fuel_abi.configurables = configurables;
+            let json_string = serde_json::to_string_pretty(&fuel_abi)?;
+            std::fs::write(json_abi_path, json_string)?;
```
kayagokalp, line 403: Looks good to me, but I believe we need a test for this. Especially because we are not sure what is going to happen if the binary does not have a `data` section. In earlier implementation it was obvious that the function to calculate offset returns `None` in that case but now I am not sure what is going to happen. 

So if you are up for it, let's add a test which get's its oracle values (like the calculated shifted values with the previous implementation) and in a next commit let's apply this change and see if the calculated values are same.

kayagokalp, line 403: Looks good to me, but I believe we need a test for this. Especially because we are not sure what is going to happen if the binary does not have a data section. In earlier implementation it was obvious that the function to calculate offset returns None in that case but now I am not sure what is going to happen.

So if you are up for it, let's add a test which get's its oracle values (like the calculated shifted values with the previous implementation) and in a next commit let's apply this change and see if the calculated values are same.

kayagokalp (changes requested):
Looks great! Thanks for the contribution. Can we add a test maybe? I left the details in the comment 🚀

forc-plugins/forc-client/src/op/deploy.rs
Abeeujah: Hi @kayagokalp I did look at the SDK before I implemented this change, the function implementation is guaranteed to always return an offset, and cases where it wouldn't the program crashes here's a snippet from `Executable<Loader>` impl block
```rust
pub fn data_offset_in_code(&self) -> usize {
        self.code_with_offset().1
    }

    fn code_with_offset(&self) -> (Vec<u8>, usize) {
        let mut code = self.state.code.clone();

        self.state.configurables.update_constants_in(&mut code);

        let blob_id = self.blob().id();

        transform_into_configurable_loader(code, &blob_id)
            .expect("checked before turning into a Executable<Loader>")
    }
```

But if you'd want a test after looking at this, what should I be testing for? that the value isn't zero? within a specific range? I'd like to hear your POV

kayagokalp: Hey again, it is possible that a binary does not contain any data section, that happens if the sway code does not have any constants and no configurable section. So if SDK is expecting this to be checked before calling the function, we might need to check this explicitly. 

For testing we can use a example test script or predicate's binary from `/test/data`. And (possibly) extract the shift calculation to a separate function so that we can write tests against it.

Basically call this new calculate shift function and check if the calculated shift is being calculated correctly for two types of sway scripts:

1. A script that has a single main function and nothing else.
2. The script like https://github.com/FuelLabs/sway/blob/master/forc-plugins/forc-client/test/data/deployed_script/src/main.sw which needs this shift.

I can help with testing if you need it feel free to ping me

Abeeujah: Yeah, I'd need some help with testing it, How do I ping you? I tried on twitter, 2 days ago and you haven't responded

kayagokalp: I will be adding the tests to this PR directly to better streamline the process

sdankel:
@Abeeujah  Thank you for your contribution, but this PR is superseded by https://github.com/FuelLabs/sway/pull/6991 since in future versions we will no longer use the data section offset.

PR #6704, opened 2024-11-08
fix: forc-deploy asks for password before checking if the wallet exists

### LINKED ISSUE #6703: forc-deploy asks password for non-existing wallet
## Problem

Currently if the user does not have a wallet, forc-deploy offers them to create a new one. We will also offer them to import if they have one already created with #6680. But even there is no wallet at the default path the ask for password is invoked before checking this situation so we end up asking for a password for a wallet that does not exists. 

```bash
...
  Confirming transactions [deploy final-contract]
             Network: https://testnet.fuel.network
             Wallet: /Users/kayagokalp/.fuel/wallets/.wallet
✔ Wallet password · ****
? Could not find a wallet at "/Users/kayagokalp/.fuel/wallets/.wallet", would you like to create a new one? [y/N]:  (y/n) ›
```

## Fix

This is a basic logic error and the order of invoking password prompt and checking/creating a new wallet should be switched.

kayagokalp:
## Description
closes #6703 and unblocks #6680.

Forc-deploy was asking for a password without checking if a wallet exists or not and the check for existence was happening after the password prompt. This meant we were asking for a password even when the wallet does not exists.

Below examples are taken in a system where the wallet does not exists at the default path (a fresh system):
### Before

```bash
...
  Confirming transactions [deploy final-contract]
             Network: https://testnet.fuel.network
             Wallet: /Users/kayagokalp/.fuel/wallets/.wallet
✔ Wallet password · ****
? Could not find a wallet at "/Users/kayagokalp/.fuel/wallets/.wallet", would you like to create a new one? [y/N]:  (y/n) ›
```

### After

```bash
...
  Confirming transactions [deploy final-contract]
             Network: https://testnet.fuel.network
? Could not find a wallet at "/Users/kayagokalp/.fuel/wallets/.wallet", would you like to create a new one? [y/N]:  (y/n) ›
```

### DIFF
--- a/forc-plugins/forc-client/src/op/deploy.rs
+++ b/forc-plugins/forc-client/src/op/deploy.rs
@@ -7,8 +7,8 @@ use crate::{
         pkg::{built_pkgs, create_proxy_contract, update_proxy_address_in_manifest},
         target::Target,
         tx::{
-            prompt_forc_wallet_password, select_account, update_proxy_contract_target,
-            SignerSelectionMode,
+            check_and_create_wallet_at_default_path, prompt_forc_wallet_password, select_account,
+            update_proxy_contract_target, SignerSelectionMode,
         },
     },
 };
@@ -1002,6 +1002,11 @@ async fn setup_deployment_account(
     } else if let Some(arn) = &command.aws_kms_signer {
         SignerSelectionMode::AwsSigner(arn.clone())
     } else {
+        // Check if we have a wallet in the default path
+        // If there is one we will ask for the password
+        // If not we will ask the user to either create a new one or import one
+        let wallet_path = default_wallet_path();
+        check_and_create_wallet_at_default_path(&wallet_path)?;
         println_action_green("", &format!("Wallet: {}", default_wallet_path().display()));
         let password = prompt_forc_wallet_password()?;
         SignerSelectionMode::ForcWallet(password)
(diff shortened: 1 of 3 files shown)


forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -1002,6 +1002,11 @@ async fn setup_deployment_account(
     } else if let Some(arn) = &command.aws_kms_signer {
         SignerSelectionMode::AwsSigner(arn.clone())
     } else {
+        // Check if we have a wallet in the default path
+        // If there is one we will ask for the password
+        // If not we will ask the user to either create a new one or import one
```
kayagokalp, line 1007:
```suggestion
        // If not we will ask the user to either create a new one
```

zees-dev (approved):
Lgtm 👍

sdankel (dismissed):
lgtm, thanks!

kayagokalp:
Updated and it should be good now :)

The failing test assumed it was testing wrong password but it was not. Due to the same bug that is being solved with this PR, cli was asking a password for a wallet that does not exists. Updated the test, as the relevant section in the code also states we need to be able to override the default path for tests to test the happy path as well. I'll try to look into it next week 👌

kayagokalp:
Thinking about this I am curious what happens if you run cargo -t locally after this change if you a have a wallet locally will the test fail 🤔 We should probably think about overriding the default path for tests so that the tests ensure reproducability

zees-dev (approved):

sdankel (approved):

Issue #2180, opened 2022-06-30
Supporting external / 3rd-party plugin configuration in `Forc.toml` with a `metadata` table


mitchmindtree:
## Motivation

The idea of configuring external tools and forc plugins via the `Forc.toml` has been raised a few times recently, the `fuel-indexer` being the most recent (https://github.com/FuelLabs/fuel-indexer/issues/89).

The goal is to allow for the most streamlined, smooth, batteries-included UX possible. Allowing for a single source of configuration could be a nice step towards this kind of experience.

That said, we also want to be careful not to unnecessarily conflate the role of `forc` itself with 3rd-party tools that are not always necessary for all projects. It's infeasible for `forc` to be aware of all possible use-cases of these 3rd-party tools, and it would be a never-ending can of worms if `forc` were to require constantly accepting PRs to its own manifest format to accommodate them, particularly as the ecosystem expands.

`forc`'s approach to supporting _plugins_ was inspired by the desire to solve extensibility in a similar manner, but only at the command level and not the configuration level.

## Proposal - Support a `metadata` table in `Forc.toml`

I propose we follow `cargo`'s lead and introduce the concept of [a `metadata` table](https://doc.rust-lang.org/cargo/reference/manifest.html#the-metadata-table). The `metadata` table would be a dedicated table supported by `forc` that can be used to host custom configuration options for external tools.

It would explicitly *not* be `forc`'s purview to parse or process any of the content under the `metadata` table itself. Instead, the idea would be to allow for 3rd-parties to use this table to host arbitrary configuration specifically for their external tools or plugins.

One of the major benefits of having a dedicated `[metadata]` table for this is that we can still lint the manifest and warn the user about unrecognised fields. Without such a check, unrecognised fields (e.g. caused by typos) have the potential to be difficult to spot, particularly as `forc` often supports default values in the case that the intended field was not present.

## Conventions for `[metadata]`

If we are to support such a table, we should set some clear conventions for plugin developers from the beginning and document them under the forc plugins chapter.

### Dedicated sub-tables

External tools should always use a dedicated sub-table under the `metadata` table. This is in order to clearly namespace their associated configuration and to avoid conflicts with other tools. E.g. always prefer:

```toml
[metadata.foo]
opt-1 = "bar"
opt-2 = "baz"
```

and never the following:

```toml
[metadata]
foo-opt-1 = "bar"
foo-opt-2 = "baz"
```

[...]

PR #6728, opened 2024-11-17
`Forc.toml` metadata support

zees-dev:
# Description
Adds support for `metadata` as `[project.metadata]` or `[workspace.metadata]` in `Forc.toml` of value `toml::Value`

- Addresses: https://github.com/FuelLabs/rfcs/blob/22bf3936a37b60bb496232c1b52bf59e204a0697/text/rfcs/0006-metadata-in-forc-manifest.md#reference-level-explanation
- Resolves: https://github.com/FuelLabs/sway/issues/2180

## Linked examples of usage
- https://github.com/FuelLabs/example-forc-plugins/pull/1

### DIFF
--- a/docs/book/src/forc/manifest_reference.md
+++ b/docs/book/src/forc/manifest_reference.md
@@ -11,6 +11,7 @@ The `Forc.toml` (the _manifest_ file) is a compulsory file for each package and 
     * For the recommended way of selecting an entry point of large libraries please take a look at: [Libraries](./../sway-program-types/libraries.md)
   * `implicit-std` -  Controls whether provided `std` version (with the current `forc` version) will get added as a dependency _implicitly_. _Unless you know what you are doing, leave this as default._
   * `forc-version` - The minimum forc version required for this project to work properly.
+  * `metadata` - Metadata for the project; can be used by tools which would like to store package configuration in `Forc.toml`.
 
 * [`[dependencies]`](#the-dependencies-section) — Defines the dependencies.
 * `[network]` — Defines a network for forc to interact with.
@@ -41,6 +42,9 @@ entry = "main.sw"
 organization = "Fuel_Labs"
 license = "Apache-2.0"
 name = "wallet_contract"
+
+[project.metadata]
+indexing = { namespace = "counter-contract", schema_path = "out/release/counter-contract-abi.json" }
 ```
 
 ## The `[dependencies]` section
--- a/docs/book/src/forc/manifest_reference.md
+++ b/docs/book/src/forc/manifest_reference.md
@@ -55,14 +55,14 @@ The `[project.metadata]` section provides a dedicated space for external tools a
 
 Metadata can be defined at two levels:
 
-Workspace level - defined in the workspace's root `Forc.toml`:
+Workspace level - defined in the workspace\'s root `Forc.toml`:
 
 ```toml
 [workspace.metadata]
 my_tool = { shared_setting = "value" }
 ```
 
-Project level - defined in individual project's `Forc.toml`:
+Project level - defined in individual project\'s `Forc.toml`:
 
 ```toml
 [project.metadata.any_name_here]
--- a/docs/book/src/forc/manifest_reference.md
+++ b/docs/book/src/forc/manifest_reference.md
@@ -66,6 +66,7 @@ setting2 = "value"
 ```
 
 Example from an indexing tool:
+
 ```toml
 [project.metadata.indexing]
 namespace = "counter-contract"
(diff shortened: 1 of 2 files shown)


forc-pkg/src/manifest/mod.rs
```diff
@@ -166,7 +166,7 @@ impl TryInto<WorkspaceManifestFile> for ManifestFile {
 type PatchMap = BTreeMap<String, Dependency>;
 
 /// A [PackageManifest] that was deserialized from a file at a particular path.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq)]
```
kayagokalp, line 169: Is there a specific reason why we are removing `Eq`? If not can we put them back, this makes the change breaking as it is, since this removes an implementation from a public type. 

Not in the sense that we need to bump sway repo to next minor but in the sense that forc-pkg as a library gets a breaking change in it's semver. But since we cannot version indivual crates currently, we would not be able to relay this information with the versioning.
 
If it was intentional it is fine but otherwise I think we should get it back

kayagokalp (commented):
Left a nit but otherwise LGTM 💯

JoshuaBatty:
Nice one. It would be nice to add some documentation about this feature so developers of plugins are aware that this is now possible. I imagine you could put something together using the motivations already written in the RFC and #2180. 

Adding it to the manifest reference probably makes the most sense https://docs.fuel.network/docs/forc/manifest_reference/

alfiedotwtf (dismissed):

forc-pkg/src/manifest/mod.rs
zees-dev: It was intentional since the `toml::Value` does not derive the `Eq` trait impl.
Don't think we need a manual implementation for this.
Also took a look at the `cargo` implementation for the `metadata`; the struct also doesnt derive `Eq` probably due to this reason:
https://github.com/rust-lang/cargo/blob/010f171ca49ab4444806ff9568249d6a818c4103/crates/cargo-util-schemas/src/manifest/mod.rs#L120

docs/book/src/forc/manifest_reference.md
```diff
@@ -41,8 +42,64 @@ entry = "main.sw"
 organization = "Fuel_Labs"
 license = "Apache-2.0"
 name = "wallet_contract"
+
+[project.metadata]
+indexing = { namespace = "counter-contract", schema_path = "out/release/counter-contract-abi.json" }
 ```
 
+### Metadata Section in Forc.toml
```
sdankel, line 50: Putting this in backticks should satiate the spellchecker 

```suggestion
### Metadata Section in `Forc.toml`
```

Otherwise you can add "toml" to `spell-check-custom-words.txt`

forc-pkg/src/manifest/mod.rs
```diff
@@ -1330,4 +1332,287 @@ mod tests {
...
+        "#;
+
+        let manifest: WorkspaceManifest = toml::from_str(toml_str).unwrap();
+        assert!(manifest.workspace.metadata.is_some());
+        let metadata = manifest.workspace.metadata.unwrap();
+        assert!(metadata.as_table().unwrap().is_empty());
+    }
+
+    #[test]
+    fn test_workspace_complex_metadata() {
```
sdankel, line 1566: ❤️

sdankel (dismissed):
LGTM, happy to approve once CI is passing

docs/book/src/forc/manifest_reference.md
```diff
@@ -41,8 +42,64 @@ entry = "main.sw"
...
+* Choose clear, descriptive metadata key names
+* Document the exact metadata key name your tool expects
+* Don't require `Forc.toml` if tool can function without it
+* Consider using TOML format for dedicated config files
+
+Implementation Notes
+
+* The metadata section is optional
+* Forc does not parse metadata contents
+* Plugin developers handle their own configuration parsing
```
sdankel, line 89: nit: It would be good to link to the example of metadata parsing here.

JoshuaBatty: Nice, it would also be useful to describe workspace and project level metadata and how these are achieved.

zees-dev: addressed: https://github.com/FuelLabs/sway/pull/6728/commits/3f05c0628257ba7ee5e71698eb2c0e2772ae3d6e

sdankel (approved):
🚀

JoshuaBatty (approved):
Nice work on this.

PR #6775, opened 2024-12-05
docs: add sway memory model

matt-user:
## Description
Developers get confused by Sway's memory model as it is different from Rust's.  So I added this to the docs.

closes https://app.asana.com/0/1207924201336629/1208429018056446

## Checklist

- x ] I have done my best to ensure that my PR adheres to [the Fuel Labs Code Review Standards](https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/external-contributors.md).


docs/book/src/reference/rust_differences.md
```diff
@@ -34,3 +34,9 @@ struct MyStruct {
     field_two: bool,
 }
 ```
+
+## Borrow Checker
```
sdankel, line 38: nit
```suggestion
## Memory Allocation
```

docs/book/src/reference/rust_differences.md
```diff
@@ -34,3 +34,9 @@ struct MyStruct {
...
+
+In Rust, the borrow checker implements Rust's [ownership system](https://doc.rust-lang.org/1.8.0/book/ownership.html)
+
+In Sway, there is no borrow checker.  This means there is no concept of ownership, borrowing, or lifetimes.  Instead, objects are copied and moved similar to C++.  Also Sway does not have any destructors nor `Drop` traits.  This means allocated memory lives for the entire transaction and is not deallocated until the end of the transaction.  A transaction can contain 65 MB of memory.
```
sdankel, line 42: I'm not sure about this one. Does the VM allocate 65 MB for every tx or is it a dynamic amount based on the program bytecode? cc @FuelLabs/client 

```suggestion
In Sway, there is no borrow checker.  This means there is no concept of ownership, borrowing, or lifetimes.  Instead, objects are copied and moved similar to C++.  Also Sway does not have any destructors nor `Drop` traits.  This means allocated memory lives for the entire transaction and is not deallocated until the end of the transaction.  A transaction may allocate up to 65 MB of memory.
```

docs/book/src/reference/rust_differences.md
```diff
@@ -34,3 +34,9 @@ struct MyStruct {
     field_two: bool,
 }
 ```
+
+## Memory Allocation
+
+In Rust, the borrow checker implements Rust's [ownership system](https://doc.rust-lang.org/1.8.0/book/ownership.html)
+
+In Sway, there is no borrow checker.  This means there is no concept of ownership, borrowing, or lifetimes.  Instead, objects are copied and moved similar to C++.  Also Sway does not have any destructors nor `Drop` traits.  This means allocated memory lives for the entire transaction and is not deallocated until the end of the transaction.  A transaction can contain 65 MB of memory.
```
bitzoic, line 42:
```suggestion
In Sway, there is no borrow checker. This means there is no concept of ownership, borrowing, or lifetimes. Instead, objects are copied and moved similar to C++. Also Sway does not have any destructors nor `Drop` traits. This means allocated memory lives for the entire transaction and is not deallocated until the end of the transaction. A transaction can contain 65 MB of memory.
```
Was this done on purpose?

IGI-111, line 42: Do we have a source for the 65 MB number and can that be linked to the spec documentation if that's where it's from? Since that may change in the future.

docs/book/src/reference/rust_differences.md
xgreenx: We have 64MB of pre-allocated memory, which we reuse for each transaction. Each transaction starts with stack and heap size 0. Each time the stack or heap is expanded, we fill expanded areas with zeros. Only initialized memory is accessible, so each memory slice will be filled with zeroes first before it can be accessed.

docs/book/src/reference/rust_differences.md
matt-user: Yeah, that's just how I write.

matt-user: cc @FuelLabs/client

sdankel (approved):
lgtm

docs/book/src/reference/rust_differences.md
matt-user: I couldn't find it in the specs, but I linked to where it is in our code

xunilrj (approved):

PR #6782, opened 2024-12-10
Show error when contract method names conflict

xunilrj:
## Description

Today, contract call uses the method name to identify which contract method to call. That means that a contract cannot have multiple methods with the same name.

That can happen when a contract implements traits:

image

A lot of the lifetime changes were caused by clippy.

## Checklist


sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
```diff
@@ -520,9 +521,20 @@ where
...
+        let mut names = BTreeMap::new();
+
+        // generate code
         for r in contract_fns {
             let decl = engines.de().get(r);
 
+            // Contract methods must be unique
+            if let Some(old) = names.insert(decl.name.as_str().to_string(), decl.name.span()) {
+                return Err(handler.emit_err(
+                    CompileError::MultipleContractsMethodsWithTheSameName { span: old },
```
sdankel, line 534: In rust, the compiler highlights one of the duplicate definitions and the error message indicates the location of the other definition. We should consider doing something similar so it's clear where to find the other definition.

image

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
```diff
@@ -520,9 +521,20 @@ where
...
+        // used to check for name collisions
+        let mut names = BTreeMap::new();
+
+        // generate code
         for r in contract_fns {
             let decl = engines.de().get(r);
 
+            // Contract methods must be unique
+            if let Some(old) = names.insert(decl.name.as_str().to_string(), decl.name.span()) {
+                return Err(handler.emit_err(
```
sdankel, line 533: It looks like this will exit if it finds one instance of duplicates. It would be a better UX if we find all of the instances of duplicates and report each of them, so that the user can see all of the errors. That is, if there are two methods named `foo` and three methods named `bar`, they should see 2 errors: one for `foo` and one for `bar`.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
xunilrj: Done.

sway-core/src/asm_lang/virtual_register.rs
```diff
@@ -76,6 +76,32 @@ pub enum ConstantRegister {
     FuncArg5,
 }
 
+impl ConstantRegister {
+    pub(crate) fn parse_register_name(raw: &str) -> Option<ConstantRegister> {
```
zees-dev, line 80: nit: Maybe look into implementing the `impl FromStr` trait instead.
This would introduce the `parse` function on the enum - which probably is the more idiomatic rust approach in this case.
It does however return a result instead (which I guess can be converted to an option using `.ok` or something)

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
```diff
@@ -520,9 +521,22 @@ where
...
 
+            let name = decl.name.as_str();
+            if !contract_methods.contains_key(name) {
+                contract_methods.insert(name.to_string(), vec![]);
+            }
+            contract_methods
+                .get_mut(name)
+                .unwrap()
+                .push(decl.name.span());
+
```
zees-dev, line 539: nit: removal of unwrap
```suggestion
        let name = decl.name.as_str();
        contract_methods
            .entry(name.to_string())
            .or_default()
            .push(function_declaration.name.span());
```

sway-core/src/asm_lang/virtual_register.rs
xunilrj: If it is no a problem I will put this into another PR.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl.rs
xunilrj: This allocates a `String` when it is not needed. That is why I went with `contains_key`.

sway-error/src/error.rs
```diff
@@ -2778,7 +2781,19 @@ impl ToDiagnostic for CompileError {
...
+                reason: Some(Reason::new(code(1), "Multiple contracts methods with the same name.".into())),
+                issue: Issue::error(
+                    source_engine,
+                    spans[0].clone(),
+                    "This is the first method".into()
+                ),
+                hints: spans.iter().skip(1).map(|span| {
+                    Hint::error(source_engine, span.clone(), "This is the duplicated method.".into())
+                }).collect(),
+                help: vec!["Contract names must be unique, even when implementing multiple ABIs.".into()],
```
sdankel, line 2794: nit:
```suggestion
                help: vec!["Contract method names must be unique, even when implementing multiple ABIs.".into()],
```

sdankel (approved):
LGTM, just one nit

sdankel:
Actually, I noticed that the new errors aren't being passed to LSP. You can test this by installing `forc-plugins/forc-lsp` and loading the test file in VSCode. The new errors do not show up in the editor. I'm still digging into it.

xunilrj:
This happens because the LSP still compiles everything using encoding v0, and this check only runs for encoding v1. See `https://github.com/FuelLabs/sway/blob/master/sway-lsp/src/core/session.rs#L323`.

It used to be the case that LSP was not compatible with encoding v1. I am not sure this is still true.

IGI-111 (dismissed):

xunilrj:
When I enable the LSP to use encoding V1 (just removing from the `no_experimental` in `session.rs`), I see the diagnostic inside `traverse`. But `typed` at line 395, is `Err(...)`, which I think is short-circuiting the processing of all the others diagnostics, and nothing is returned.

zees-dev (dismissed):

sway-error/src/error.rs
xunilrj: Done.

xunilrj:
I managed to improve what the compiler sends to the LSP server, and now the error is being displayed correctly.

image

xunilrj:
I have rollbacked part of the solution to enable encoding v1 in the LSP. The issue is a problem with the garbage collector. However, I am pushing the solution to improve how the LSP deals with compiler errors.

This means that when the compiler fails, it returns as much information as possible to the LSP server. Which is also using as much information as it can to enrich data for tokens.

In general, LSP should be able to show more diagnostics now. But unfortunately, it will not show the errors of this PR.

sdankel (dismissed):
🚀

sway-lsp/src/capabilities/runnable.rs
```diff
@@ -15,7 +15,7 @@ pub struct RunnableTestFn {
     /// The location in the file where the runnable button should be displayed
     pub range: Range,
     /// The program kind of the current file.
-    pub tree_type: TreeType,
+    //pub tree_type: TreeType,
```
JoshuaBatty, line 18: can probably remove this if it's not used any more

sway-lsp/src/core/session.rs
```diff
@@ -626,25 +633,25 @@ fn create_runnables(
             let path = source_engine.get_path(source_id);
             let runnable = Box::new(RunnableTestFn {
                 range: token::get_range_from_span(&span.clone()),
-                tree_type: typed_program.kind.tree_type(),
+                //tree_type: typed_program.kind.tree_type(),
```
JoshuaBatty, line 636: same here

sway-lsp/tests/fixtures/auto_import/Forc.toml
```diff
@@ -6,4 +6,4 @@ name = "auto_import"
 implicit-std = false
 
 [dependencies]
-std = { git = "https://github.com/FuelLabs/sway", tag = "v0.48.1" }
+std = { path = "../../../../sway-lib-std" }
```
JoshuaBatty, line 9:
```suggestion
std = { git = "https://github.com/FuelLabs/sway", tag = "v0.66.5" }
```

We ended up pinning these examples so it wouldn't require updates when std-lib had changes. Would it be possible to just pin it to the latest master?

sway-lsp/tests/fixtures/diagnostics/dead_code/Forc.toml
```diff
@@ -3,7 +3,7 @@ authors = ["Fuel Labs <contact@fuel.sh>"]
 entry = "main.sw"
 license = "Apache-2.0"
 name = "dead_code"
-implicit-std = false
 
 [dependencies]
-std = { git = "https://github.com/FuelLabs/sway", tag = "v0.48.1" }
```
JoshuaBatty, line 9: same here

sway-lsp/tests/fixtures/runnables/Forc.toml
```diff
@@ -3,7 +3,7 @@ authors = ["Fuel Labs <contact@fuel.sh>"]
 entry = "main.sw"
 license = "Apache-2.0"
 name = "script_multi_test"
-implicit-std = false
 
 [dependencies]
-std = { git = "https://github.com/FuelLabs/sway", tag = "v0.48.1" }
```
JoshuaBatty, line 9: and here

JoshuaBatty (commented):
Nice work @xunilrj , just a couple of small nits.

sway-lsp/src/capabilities/runnable.rs
xunilrj: Done.

sway-lsp/src/core/session.rs
xunilrj: Done.

sway-lsp/tests/fixtures/runnables/Forc.toml
xunilrj: Done.

sway-lsp/tests/fixtures/diagnostics/dead_code/Forc.toml
xunilrj: Done.

sway-lsp/tests/fixtures/auto_import/Forc.toml
xunilrj: Done.

JoshuaBatty (approved):

IGI-111 (approved):

Issue #6772, opened 2024-12-04
Associated constant visibility is broken


tritao:
```
library;

pub struct I8 {
    underlying: u8,
}

impl I8 {
    const MIN: Self = Self {
        underlying: u8::min(),
    };
}

trait Foo {
    fn foo();
}

impl Foo for I8 {
    fn foo() {
        let x = Self::MIN;
        log(x);
    }
}
```

https://github.com/SwayStar123/private_associated_const_repro

PR #6785, opened 2024-12-11
Rework and fix associated const visibility for impl items

tritao:
## Description

This PR fixes associated constant visiblity checking for impl items.

The behaviour is that consts declared in traits are always considered public, whereas if they are declared anywhere else they are private unless they have the pub modifier.

```sway
library;

pub struct I8 {
    underlying: u8,
}

impl I8 {
    const MIN: Self = Self {
        underlying: u8::min(),
    };
}

trait Foo {
    fn foo();
}

impl Foo for I8 {
    fn foo() {
        let x = Self::MIN;
        log(x);
    }
}
```

Fixes https://github.com/FuelLabs/sway/issues/6772.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -221,15 +221,16 @@ jobs:
         run: cargo install --locked --debug --path ./forc
       - name: Install Forc plugins
         run: |
-          cargo install --locked --debug --path ./forc-plugins/forc-fmt
-          cargo install --locked --debug --path ./forc-plugins/forc-lsp
           cargo install --locked --debug --path ./forc-plugins/forc-client
+          cargo install --locked --debug --path ./forc-plugins/forc-crypto
           cargo install --locked --debug --path ./forc-plugins/forc-debug
           cargo install --locked --debug --path ./forc-plugins/forc-doc
-          cargo install --locked --debug --path ./forc-plugins/forc-tx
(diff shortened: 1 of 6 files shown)


sway-core/src/semantic_analysis/symbol_resolve_context.rs
```diff
@@ -43,6 +43,8 @@ pub struct SymbolResolveContext<'a> {
     ///
     /// This is `Disallow` everywhere except while checking type parameters bounds in struct instantiation.
     generic_shadowing_mode: GenericShadowingMode,
+    /// Visibility checking level for the current scope.
+    visibility_level: Visibility,
```
jjcnn, line 47: I don't understand this comment, and I'm finding it difficult to understand what `visibility_level` is supposed to represent. Could you elaborate a bit (ideally in this comment, or in the PR description)?

jjcnn (changes requested):

jjcnn (changes requested):
I'm still confused by what `visibility_level` means.

It also took me a while to figure out what problem this PR is supposed to solve in the first place. The PR deccription just links to an issue, but that issue only consists of an example program with no description of what the compiler does incorrectly for that program.

sdankel (approved):

JoshuaBatty (approved):

PR #6789, opened 2024-12-13
Demonomorphize functions

vaivaswatha:


ironcev:
Closing, as the benchmarks showed regressions and we decided not to proceed with this optimizations. It turned out that the overhead of passing the variable size and handling it increases the gas cost.

Issue #6725, opened 2024-11-14
CLI tool for querying smart contract state


zees-dev:
# Proposal: CLI tool for querying smart contract state

## Problem

Currently querying smart contract state requires multiple manual steps:
- Retrieving the contract's ABI
- Writing custom Rust/TypeScript scripts
- Setting up development environment
- Managing dependencies

This process is time-consuming and creates unnecessary friction for developers.

## Proposed Solution
A CLI tool that simplifies smart contract state queries:

```sh
forc call <contract-ID> <function-name/sig> [--args key=value]
```

## Key Features

- Automatic ABI fetching from verified contracts
- Interactive mode for exploring available functions
- Output formatting options (JSON, table, raw)

## Benefits

- Lower barrier to entry for quickly validating contract state
- Faster debugging and monitoring
- No need for custom script maintenance

## Example usage

<details>
<summary>Proposed CLI</summary>

```sh
# > forc call --help
Perform a call on a contract without publishing a transaction

Usage: forc call [OPTIONS] [TO] [SIG] [ARGS]... [COMMAND]

Commands:

Arguments:
    [TO]
    The destination of the transaction

    [SIG]
    The signature of the function to call

    [ARGS]...
    The arguments of the function to call

Options:
    --data <DATA>
        Data for the transaction

    -b, --block <BLOCK>
        The block height to query at.
        
        Can also be the tags earliest, finalized, safe, latest, or pending.

    -h, --help
        Print help (see a summary with '-h')

Display options:
    -j, --json
        Print the decoded output as JSON

Transaction options:
    --value <VALUE>
        Ether to send in the transaction, either specified in wei, or as a string with a unit type.
        Examples: 1ether, 10gwei, 0.01ether

Fuel options:
    -r, --rpc-url <URL>
        The RPC endpoint

        [env: FUEL_RPC_URL=]
        [default: "http://127.0.0.1:4000/v1/graphql"]

Wallet options - raw:
    -f, --from <ADDRESS>
        The sender account

        [env: FUEL_FROM=]

    --private-key <RAW_PRIVATE_KEY>
        Use the provided private key

    --mnemonic <MNEMONIC>
        Use the mnemonic phrase of mnemonic file at the specified path

    --mnemonic-passphrase <PASSPHRASE>
        Use a BIP39 passphrase for the mnemonic

    --mnemonic-derivation-path <PATH>
        The wallet derivation path.

        Works with both --mnemonic-path and hardware wallets.

[...]

PR #6791, opened 2024-12-17
plugin: forc-call

zees-dev:
## Description
The following PR introduces a new forc plugin; `forc-call`.

This plugin allows users to call functions on deployed contracts using the `forc call` command.
This is ideal for quickly querying the state of a deployed contract.

In this first implementation; the contract ABI is required (as a path to a local JSON file or a URL to a remote JSON file).

This is inspired by the [`cast call`](https://book.getfoundry.sh/reference/cast/cast-call) tool; which is a popular tool for interacting with deployed contracts on Ethereum.
The implementation is based on the following Github issue: https://github.com/FuelLabs/sway/issues/6725

In the current implementation, you can query a contract state using the `forc call` command by providing the target contract ID, it's respective ABI file, and the function name (selector) and arguments.

<details>
  <summary>Forc Call CLI</summary>
  
  ```sh
  forc call --help
  ```

  ```
Call a contract function

Usage: forc call [OPTIONS] <CONTRACT_ID> <FUNCTION> [ARGS]...

Arguments:
  <CONTRACT_ID>
          The contract ID to call

  <FUNCTION>
          The function signature to call. When ABI is provided, this should be a selector (e.g. "transfer") When no ABI is provided, this should be the full function signature (e.g. "transfer(address,u64)")

  [ARGS]...
          Arguments to pass into main function with forc run

Options:
      --abi <ABI>
          Optional path or URI to a JSON ABI file

      --node-url <NODE_URL>
          The URL of the Fuel node to which we're submitting the transaction. If unspecified, checks the manifest's `network` table, then falls back to `http://127.0.0.1:4000`
          
          You can also use `--target`, `--testnet`, or `--mainnet` to specify the Fuel node.
          
          [env: FUEL_NODE_URL=]

      --target <TARGET>
          Use preset configurations for deploying to a specific target.
          
          You can also use `--node-url`, `--testnet`, or `--mainnet` to specify the Fuel node.
          
          Possible values are: [local, testnet, mainnet]

      --testnet
          Use preset configuration for testnet.
          
          You can also use `--node-url`, `--target`, or `--mainnet` to specify the Fuel node.

      --mainnet
          Use preset configuration for mainnet.
          
          You can also use `--node-url`, `--target`, or `--testnet` to specify the Fuel node.

      --signing-key <SIGNING_KEY>
          Derive an account from a secret key to make the call
          
          [env: SIGNING_KEY=]

      --wallet
          Use forc-wallet to make the call

      --amount <AMOUNT>
          Amount of native assets to forward with the call
          
          [default: 0]

      --asset-id <ASSET_ID>
          Asset ID to forward with the call

      --gas-forwarded <GAS_FORWARDED>
          Amount of gas to forward with the call

      --mode <MODE>
          The execution mode to use for the call; defaults to dry-run; possible values: dry-run, simulate, live
          
          [default: dry-run]

      --gas-price <PRICE>
          Gas price for the transaction

      --script-gas-limit <SCRIPT_GAS_LIMIT>
          Gas limit for the transaction

      --max-fee <MAX_FEE>
          Max fee for the transaction

      --tip <TIP>
          The tip for the transaction

      --external-contracts <EXTERNAL_CONTRACTS>
          The external contract addresses to use for the call If none are provided, the call will automatically extract contract addresses from the function arguments and use them for the call as external contracts

  -h, --help
          Print help (see a summary with '-h')

  -V, --version
          Print version
  ```
</details>

### Example usage

```sh
forc call 0xe18de7c7c8c61a1c706dccb3533caa00ba5c11b5230da4428582abf1b6831b4d --abi ./out/debug/counter-contract-abi.json add 1 2
```

- where `0xe18de7c7c8c61a1c706dccb3533caa00ba5c11b5230da4428582abf1b6831b4d` is the contract ID
- where `./out/debug/counter-contract-abi.json` is the path to the ABI file
- where `add` is the function name (selector)
- where `1 2` are the arguments to the function

^ the sway code for the add function could be:

```sway
contract;
abi MyContract {
    fn add(a: u64, b: u64) -> u64;
}
    fn add(a: u64, b: u64) -> u64 {
        a + b
    }
}
```

## Implementation details

1. The provided ABI file downloaded (unless local path is provided)
2. The ABI is parsed into internal representation
3. The provided function selector e.g. `add` is matched with the extracted functions from the ABI
4. The provided arguments are parsed into the appropriate types which match the extracted function's inputs
5. The function selector and args are then converted into the `Token` enum, which is then ABI encoded as part of the `ContractCall` struct
6. The `ContractCall` struct is then used to make a request to the node to call the function
7. The response is then decoded into the appropriate type (based on matched function's output type)

^ In the implementation, we don't use the `abigen!` macro since this is a compile time parser of the ABI file; instead we make use of the lower level encoding and decoding primitives and functions from the [Rust SDK](https://github.com/FuelLabs/fuels-rs).

### Example 1

The example contract above with `add` function has been deployed on testnet - with ABI file available [here](https://pastebin.com/raw/XY3awY3T).
The add function can be called via the CLI:

```sh
cargo run -p forc-client --bin call -- \
  --testnet \
  --abi https://pastebin.com/raw/XY3awY3T \
  0xe18de7c7c8c61a1c706dccb3533caa00ba5c11b5230da4428582abf1b6831b4d \
  add 1 2
```

### Example 2 - get `owner` of Mira DEX contract

```sh
cargo run -p forc-client --bin call -- \
  --testnet \
  --abi https://raw.githubusercontent.com/mira-amm/mira-v1-periphery/refs/heads/main/fixtures/mira-amm/mira_amm_contract-abi.json \
  0xd5a716d967a9137222219657d7877bd8c79c64e1edb5de9f2901c98ebe74da80 \
  owner
```

Note: Testnet contract address [here](https://docs.mira.ly/developer-guides/developer-overview#testnet-deployment)

## Encoding of primitive types

When passing in function arguments, the following types are encoded as follows:

| Types                                         | Example input                                                        | Notes                                                                                          |
|-----------------------------------------------|----------------------------------------------------------------------|------------------------------------------------------------------------------------------------|
| bool                                          | `true` or `false`                                                    |                                                                                                |
| u8, u16, u32, u64, u128, u256                 | `42`                                                                 |                                                                                                |
| b256                                          | `0x0000000000000000000000000000000000000000000000000000000000000042` or `0000000000000000000000000000000000000000000000000000000000000042` | `0x` prefix is optional |
| bytes, RawSlice                               | `0x42` or `42`                                                       | `0x` prefix is optional                                                                                               |
| String, StringSlice, StringArray (Fixed-size) | `"abc"`                                                              |                                                                                                |
| Tuple                                         | `(42, true)`                                                         | The types in tuple can be different                                                                                               |
| Array (Fixed-size), Vector (Dynamic)          | `[42, 128]`                                                          | The types in array or vector must be the same; i.e. you cannot have `[42, true]`              |
| Struct                                        | `{42, 128}`                                                          | Since structs are packed encoded, the attribute names are not encoded; i.e. `{42, 128}`; this could represent the following `struct Polygon { x: u64, y: u64 }` |
| Enum                                          | `(Active: true)` or `(1: true)`       | Enums are key-val pairs with keys as being variant name (case-sensitive) or variant index (starting from 0) and values as being the variant value; this could represent the following `enum MyEnum { Inactive, Active(bool) }` |

<details>
  <summary>Encoding cheat-sheet</summary>

A few of the common types are encoded as follows:

| Types                                      | Encoding Description                                                      | Example                                                                                          |
|--------------------------------------------|--------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|
| bool, u8                                   | Encoded as a single byte. `bool`: 0x00 (false) or 0x01 (true); `u8` is the byte itself. | `bool(true) = 0x01`, `u8(42) = 0x2A`                                                     |
| u16                                        | 2-byte, big-endian                                                        | `u16(42) = 0x002A`                                                                             |
| u32                                        | 4-byte, big-endian                                                        | `u32(42) = 0x0000002A`                                                                        |
| u64                                        | 8-byte, big-endian                                                        | `u64(42) = 0x000000000000002A`                                                               |
| u128                                       | 16-byte, big-endian                                                       | `u128(42) = 0x0000000000000000000000000000002A`                                              |
| u256, b256                                 | 32-byte value. For u256: big-endian integer; For b256: raw 32 bytes       | `u256(42) = 32-bytes ending with ...0000002A`, `b256(...) = exactly the 32-byte array`       |
| Tuples, Arrays, Structs (Fixed-size)       | Concatenate the encodings of each element/field with no extra padding     | `(u8(1), bool(true)) = 0x01 0x01`; `[u16;2]: [42,100] = 0x002A0064`; `struct {u8,u8}: 0x0102` |
| Enums                                      | `u64` variant index + encoded variant data with no extra padding          | `MySumType::X(42): 0x0000000000000000 000000000000002A`                                       |
| Bytes, String, RawSlice, Vector (Dynamic)  | `u64` length prefix + raw data, no padding                                | `"abc" = length=3: 0x0000000000000003 0x61 0x62 0x63`                                         |

^ This is based on the docs here: https://docs.fuel.network/docs/specs/abi/argument-encoding
</details>

## Future improvements

1. Support for function signature based calls without ABI
2. Support for raw calldata input
3. Function selector completion - given ABI file

## Checklist


forc-plugins/forc-client/src/op/call.rs
```diff
@@ -0,0 +1,2221 @@
...
+                file.write_all(&response)?;
+                (path, true)
+            }
+        };
+
+        let abi_str = std::fs::read_to_string(file_path.clone()).expect("Failed to read ABI file");
+        if is_temp_file {
+            std::fs::remove_file(file_path).expect("Failed to remove ABI file");
+        }
+        let parsed_abi = UnifiedProgramABI::from_json_abi(&abi_str)?;
```
JoshuaBatty, line 71: should we add validation to ensure ABI files are well-formed and compatible before progressing?

```rust
// Validate minimum required fields
if parsed_abi.functions.is_empty() {
    bail!("Invalid ABI: No functions found in ABI file");
}

...
```

forc-plugins/forc-client/src/cmd/call.rs
```diff
@@ -0,0 +1,141 @@
...
+use fuels_core::types::ContractId;
+use std::{path::PathBuf, str::FromStr};
+use url::Url;
+
+pub use forc::cli::shared::{BuildOutput, BuildProfile, Minify, Pkg, Print};
+pub use forc_tx::{Gas, Maturity};
+
+forc_util::cli_examples! {
+    super::Command {
+        [ Call a contract with function parameters => "forc call <CONTRACT_ID> <FUNCTION_SIGNATURE> <ARGS>" ]
```
JoshuaBatty, line 14: i wonder if being a bit more explicit on what these calls look like might be easier for people to reference? I could go either but maybe the above could be 

```
[ Call a simple contract function => "forc call 0x123..abc add 1 2" ]
```

thoughts?

forc-plugins/forc-client/src/op/call.rs
```diff
@@ -0,0 +1,2221 @@
...
+                    .dry_run(tx)
+                    .await
+                    .expect("Failed to dry run transaction")
+            }
+            cmd::call::ExecutionMode::Simulate => {
+                forc_tracing::println_warning(&format!(
+                    "Simulating transaction with wallet... {}",
+                    wallet.address().hash()
+                ));
+                let tx_policies = gas.clone().map(Into::into).unwrap_or(TxPolicies::default());
```
JoshuaBatty, line 168:
```suggestion
                let tx_policies = gas.as_ref().map(Into::into).unwrap_or_else(TxPolicies::default);
```

Can probably remove the clone by using as_ref

forc-plugins/forc-client/src/cmd/call.rs
zees-dev: agreed; makes sense 👍

forc-plugins/forc-client/src/op/call.rs
zees-dev: I was thinking we may want to support calling a contract without any functions - where you may want to only specify an amount.
In the case of solidity, this calls the default `fallback` or `receive` function - allowing the contract to receive funds.

I was unsure whether sway has an equivalent default function - or an approach where the contract can receive funds without calling an explicit function.

zees-dev:
Note: Currently CI fails since there is reliance on new release from [rust sdk](https://github.com/FuelLabs/fuels-rs) - which contains the relevant changes:
- https://github.com/FuelLabs/fuels-rs/pull/1553
- https://github.com/FuelLabs/fuels-rs/pull/1554

forc-plugins/forc-client/src/op/call.rs
Voxelot: This is indeed possible, rather than using a `CALL` instruction, you can use a [`TR` or transfer instruction](https://docs.fuel.network/docs/specs/fuel-vm/instruction-set/#tr-transfer-coins-to-contract) to send funds into a contract.

https://docs.fuel.network/docs/nightly/fuels-ts/wallets/wallet-transferring/#transferring-assets-to-contracts

The sway also has a notion of default functions which are used by proxy contracts to pass-through calls and args to an implementation contract, but its not necessary to use this default function to receive funds.

forc-plugins/forc-client/src/cmd/call.rs
```diff
@@ -0,0 +1,176 @@
...
+use fuel_crypto::SecretKey;
+use fuels::programs::calls::CallParameters;
+use fuels_core::types::{AssetId, ContractId};
+use std::{path::PathBuf, str::FromStr};
+use url::Url;
+
+pub use forc::cli::shared::{BuildOutput, BuildProfile, Minify, Pkg, Print};
+pub use forc_tx::{Gas, Maturity};
+
+forc_util::cli_examples! {
```
sdankel, line 13: For some reason these examples aren't showing up when I do `forc call --help`. Here's what I did to get it to build if you want the changes: https://github.com/FuelLabs/sway/pull/6831

Installed with `cargo install --path forc-plugins/forc-client`

forc-plugins/forc-client/src/op/call.rs
```diff
@@ -0,0 +1,2363 @@
...
+                    &format!("{}/tx/0x{}", explorer_url, tx_hash),
+                );
+            }
+        }
+
+        return Ok(result);
+    }
+
+    let cmd::call::FuncType::Signature(function_signature) = function else {
+        bail!("Function must be a signature if no ABI is provided");
```
sdankel, line 258: nit: It would be good to give an example here. 
```suggestion
        bail!("Function must be a signature if no ABI is provided, e.g. "transfer(address,u64)"
```

I wasn't able to get it to work without providing the ABI. These commands I tried all gave me panics:
- `forc call --testnet 0xd8b98e8ceebf08e2fbc4bf0599e8eae239608636b9aab76b5348c2174d197848 "initialize_counter(1)"`
- `forc call --testnet 0xd8b98e8ceebf08e2fbc4bf0599e8eae239608636b9aab76b5348c2174d197848 "initialize_counter(1, u64)"`
- `forc call --testnet 0xd8b98e8ceebf08e2fbc4bf0599e8eae239608636b9aab76b5348c2174d197848 "initialize_counter(value,u64) 1"`
- `forc call --testnet 0xd8b98e8ceebf08e2fbc4bf0599e8eae239608636b9aab76b5348c2174d197848 "initialize_counter(value,u64)" 1`
- `forc call --testnet 0xd8b98e8ceebf08e2fbc4bf0599e8eae239608636b9aab76b5348c2174d197848 "initialize_counter(value:u64)" 1`

This worked however: `forc call --testnet --abi examples/counter/out/debug/counter-abi.json 0xd8b98e8ceebf08e2fbc4bf0599e8eae239608636b9aab76b5348c2174d197848 "initialize_counter" 1 `

forc-plugins/forc-client/src/op/call.rs
```diff
@@ -0,0 +1,2363 @@
...
+    } = cmd;
+    let node_url = get_node_url(&node, &None)?;
+    let provider: Provider = Provider::connect(node_url.clone()).await?;
+
+    let wallet = get_wallet(caller, provider).await?;
+
+    if let Some(abi) = abi {
+        // If ABI is provided, ensure function signature is just the selector
+        let cmd::call::FuncType::Selector(selector) = function else {
+            bail!("Function must be a selector");
```
sdankel, line 58: It would be good to provide an example here:
```suggestion
            bail!("Function must be a selector, e.g \"transfer\"");
```

Even better would be if we could try to correctly print the function name that they entered.

sdankel (commented):
Impressive work! It's still rough around the edges but I'm excited to use this.

forc-plugins/forc-client/src/cmd/call.rs
zees-dev: thanks for this!

forc-plugins/forc-client/Cargo.toml
```diff
@@ -36,26 +37,32 @@ futures.workspace = true
...
+url.workspace = true
 
 [dev-dependencies]
 portpicker = "0.1.1"
 pretty_assertions = "1.4.1"
 rexpect = "0.5"
-tempfile = "3"
 
 [build-dependencies]
 regex = "1.5.4"
```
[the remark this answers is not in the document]
zees-dev: addressed by you here: https://github.com/FuelLabs/sway/pull/6791#discussion_r1914400582 - thanks

forc-plugins/forc-client/src/util/tx.rs
```diff
@@ -275,6 +275,49 @@ pub(crate) async fn select_account(
     }
 }
 
+pub(crate) async fn select_local_wallet_account(
```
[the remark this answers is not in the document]
zees-dev: this is a big function; in fact i propose we incorporate this function in the function above - this way it can be re-used in both places.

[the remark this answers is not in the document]
zees-dev: addressed: https://github.com/FuelLabs/sway/pull/6791/commits/014e9a7257af47c446aed9258b26ba5346e026e7

forc-plugins/forc-client/src/util/node_url.rs
```diff
@@ -33,6 +33,21 @@ pub fn get_node_url(
     Ok(node_url)
 }
 
+/// Returns the URL for explorer
+pub fn get_explorer_url(node_target: &NodeTarget) -> Option<String> {
```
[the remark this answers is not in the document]
zees-dev: https://github.com/FuelLabs/sway/pull/6791/commits/67e22a5d0198210f85dfb4d5d8f92146d5ae9912

forc-plugins/forc-client/src/op/call.rs
```diff
@@ -0,0 +1,2363 @@
+use crate::{
```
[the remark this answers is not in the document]
zees-dev: addressed:
- https://github.com/FuelLabs/sway/pull/6791/commits/f8bbfb20cf53a8ade5906d54d2cd41980ed5cec8
- https://github.com/FuelLabs/sway/pull/6791/commits/9363181619c3e1914a0e97fbf44d6a904d5a6576

forc-plugins/forc-client/src/op/call.rs
zees-dev: https://github.com/FuelLabs/sway/pull/6791/commits/931b825065a49a75bae67e56bbd61138e8bc992d

forc-plugins/forc-client/src/op/call.rs
sdankel: Got it, we should just make sure the CLI isn't panicking in any scenario

forc-plugins/forc-client/src/cmd/call.rs
```diff
@@ -0,0 +1,176 @@
...
+/// Call a contract function.
+#[derive(Debug, Default, Parser, Clone)]
+#[clap(bin_name = "forc call", version)]
+pub struct Command {
+    /// The contract ID to call.
+    pub contract_id: ContractId,
+
+    /// Optional path or URI to a JSON ABI file.
+    #[clap(long, value_parser = parse_abi_path)]
+    pub abi: Option<Either<PathBuf, Url>>,
```
sdankel, line 131:
```suggestion
    /// Path or URI to a JSON ABI file.
    #[clap(long, value_parser = parse_abi_path)]
    pub abi: Either<PathBuf, Url>,
```

forc-plugins/forc-client/src/cmd/call.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    /// The function signature to call.
+    /// When ABI is provided, this should be a selector (e.g. "transfer")
+    /// When no ABI is provided, this should be the full function signature (e.g. "transfer(address,u64)")
```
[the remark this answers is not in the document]
zees-dev: https://github.com/FuelLabs/sway/pull/6791/commits/ce1d709506fc84ff9629803b7adfd32a673b86c5

forc-plugins/forc-client/src/cmd/call.rs
zees-dev: https://github.com/FuelLabs/sway/pull/6791/commits/ce1d709506fc84ff9629803b7adfd32a673b86c5

forc-plugins/forc-client/src/cmd/call.rs
```diff
@@ -0,0 +1,241 @@
...
+use std::{path::PathBuf, str::FromStr};
+use url::Url;
+
+pub use forc::cli::shared::{BuildOutput, BuildProfile, Minify, Pkg, Print};
+pub use forc_tx::{Gas, Maturity};
+
+#[derive(Debug, Clone)]
+pub enum FuncType {
+    Selector(String),
+    // TODO: add support for function signatures - without ABI files
```
JoshuaBatty, line 16: Have we got a github issue for this that we can mention here?

forc-plugins/forc-client/src/cmd/call.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+impl FromStr for FuncType {
+    type Err = String;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let s = s.trim().replace(' ', "");
+        if s.is_empty() {
+            return Err("Function signature cannot be empty".to_string());
+        }
+        // Check if function signature is a valid selector (alphanumeric and underscore support)
```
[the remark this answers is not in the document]
zees-dev: interesting; maybe ill just remove this requirement since we must also match the function in the abi too - which should enforce the correct naming conventions

forc-plugins/forc-client/src/cmd/call.rs
zees-dev: No; i will add

forc-plugins/forc-client/src/cmd/call.rs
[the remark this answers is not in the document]
zees-dev: https://github.com/FuelLabs/sway/pull/6791/commits/330596d0f2cd451c1410f3f35abe1f26c5404641

forc-plugins/forc-client/src/cmd/call.rs
zees-dev: https://github.com/FuelLabs/sway/pull/6791/commits/e7e0c41d88d6e45698ce37e2b04cd74dc7673392

alfiedotwtf:
Apart from my comments, looks good. I'll wait for comments before hitting approve though

docs/book/src/forc/plugins/forc_client/forc_call.md
```diff
@@ -0,0 +1,307 @@
...
+
+  --node-url <NODE_URL>
+    The URL of the Fuel node to which we're submitting the transaction. If unspecified, checks the manifest's `network` table, then falls back to `http://127.0.0.1:4000`
+    
+    You can also use `--target`, `--testnet`, or `--mainnet` to specify the Fuel node.
+    
+    [env: FUEL_NODE_URL=]
+
+  --target <TARGET>
+    Use preset configurations for deploying to a specific target.
```
JoshuaBatty, line 80: what is forc call deploying?

forc-plugins/forc-client/src/op/call/missing_contracts.rs
```diff
@@ -0,0 +1,68 @@
+use anyhow::{bail, Result};
+use fuel_tx::{PanicReason, Receipt};
+use fuels::programs::calls::{traits::TransactionTuner, ContractCall};
+use fuels_accounts::{provider::Provider, wallet::WalletUnlocked};
+use fuels_core::types::{
+    bech32::Bech32ContractId, transaction::TxPolicies, transaction_builders::VariableOutputPolicy,
+};
+
+pub async fn get_missing_contracts(
```
JoshuaBatty, line 9: would be nice to add doc comments for all `pub` functions ...

docs/book/src/forc/plugins/forc_client/forc_call.md
zees-dev: nothing; it was using existing struct - which had original use in forc-deploy i believe.
Have updated the docs there and in the book:
https://github.com/FuelLabs/sway/pull/6791/commits/66451e764ad93b27fb6d712fbc625369e5ea74b8

forc-plugins/forc-client/src/op/call/mod.rs
```diff
@@ -0,0 +1,713 @@
...
+        external_contracts,
+        output,
+    } = cmd;
+    let node_url = get_node_url(&node, &None)?;
+    let provider: Provider = Provider::connect(node_url).await?;
+
+    let wallet = get_wallet(caller, provider).await?;
+
+    let cmd::call::FuncType::Selector(selector) = function;
+    let abi_str = match abi {
```
JoshuaBatty, line 64: would also be cool to try and fetch the verified ABI from the registry if only the contract id was given with no ABI file... obviously this isn't quite ready yet but would be a nice addition. Perhaps leaving a comment or linking to an issue to integrate this would be good.

forc-plugins/forc-client/src/op/call/parser.rs
```diff
@@ -0,0 +1,1557 @@
...
+                ]),
+                enum_variants,
+            )
+                .into(),
+        );
+        let output = token_to_string(&token).unwrap();
+        assert_eq!(output, "(Input:{(Active:true), [(Pending:42)]})");
+
+        // enum (complex example 2) - 2 variants, one with a struct that contains another struct with 2 fields, another with a struct
+        // enum GenericEnum<GenericStruct<u32>> {
```
JoshuaBatty, line 1497: do we need this commented section still?

zees-dev: i think it's useful to keep it there for debugging purposes to know what we are dealing with; however can remove if we want

JoshuaBatty: No worries, it's fine. If it's useful then leave it. Was just checking.

forc-plugins/forc-client/src/op/call/missing_contracts.rs
zees-dev: addressed:
- https://github.com/FuelLabs/sway/pull/6791/commits/870e4aa537f3d86e302667beab0b8294d3430a55
- https://github.com/FuelLabs/sway/pull/6791/commits/f99b8ab8dab24f44bf7dca761840004509dac86a

forc-plugins/forc-client/src/op/call/mod.rs
zees-dev: https://github.com/FuelLabs/sway/pull/6791/commits/46e5007e434824916b964b26de81e6714e008047

JoshuaBatty (approved):
great work with this

alfiedotwtf (changes requested):

forc-plugins/forc-client/src/op/call/mod.rs
```diff
@@ -0,0 +1,713 @@
...
+            "[42, 42, 42, 42, 42, 42, 42, 42, 42, 42]"
+        );
+
+        // test_array - fails if different types
+        let cmd = get_contract_call_cmd(id, &wallet, "test_array", vec!["[42, true]"]);
+        assert_eq!(
+            call(cmd).await.unwrap_err().to_string(),
+            "failed to parse u64 value: true"
+        );
+
```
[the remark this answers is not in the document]
zees-dev: this does fail

forc-plugins/forc-client/src/op/call/parser.rs
```diff
@@ -0,0 +1,1557 @@
...
+                .map(token_to_string)
+                .collect::<Result<Vec<String>>>()?
+                .join(", ");
+            Ok(format!("{{{inner}}}"))
+        }
+        Token::Enum(selector) => {
+            let (discriminant, value, enum_variants) = &**selector;
+            let (name, _ty) = enum_variants
+                .select_variant(*discriminant)
+                .expect("failed to select variant");
```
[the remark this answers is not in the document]
zees-dev: I dont think its required yet; maybe if people as for it.
I want to enable this once we get it right; was getting false positives in testing

forc-plugins/forc-client/src/op/call/parser.rs
```diff
@@ -0,0 +1,1557 @@
...
+                    parts.push(current.trim().to_string());
+                    current = String::new();
+                }
+            }
+            (_, _, _) => {
+                escaped = false;
+                current.push(c);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
zees-dev: Dont think so

forc-plugins/forc-client/src/util/node_url.rs
```diff
@@ -6,21 +6,24 @@ use crate::NodeTarget;
 use super::target::Target;
 
 /// Returns the URL to use for connecting to Fuel Core node.
```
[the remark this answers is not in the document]
zees-dev: Yea i think there was an issue with dependencies and due to this the migration was not straightforward; hence added a todo to address later.

docs/book/src/forc/plugins/forc_client/forc_call.md
```diff
@@ -0,0 +1,307 @@
...
+Arguments:
+  <CONTRACT_ID>
+    The contract ID to call
+
+  <FUNCTION>
+    The function signature to call. When ABI is provided, this should be a selector (e.g. "transfer") When no ABI is provided, this should be the full function signature (e.g. "transfer(address,u64)")
+
+  [ARGS]...
+    Arguments to pass into main function with forc run
+
```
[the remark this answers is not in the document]
zees-dev: not a bad idea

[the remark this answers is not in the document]
zees-dev: https://github.com/FuelLabs/sway/pull/6791/commits/0eb02256c50760a5e4eb345bc1bdd30198aeb470

forc-plugins/forc-client/src/op/call/mod.rs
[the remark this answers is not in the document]
alfiedotwtf: If it fails, shouldn't this unwrap the error then?

(just in case there's confusion, I'm talking about 493 `get_contract_call_cmd(id, &wallet, "test_array", vec!["[42, 42]"]);` which looking at it is returning 10 items?))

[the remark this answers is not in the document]
zees-dev: Im assuming you're talking about this test:
```rust
        // test_array - succeeds if length not matched!?
        let cmd = get_contract_call_cmd(id, &wallet, "test_array", vec!["[42, 42]"]);
        assert_eq!(
            call(cmd).await.unwrap(),
            "[42, 42, 0, 4718592, 65536, 65536, 0, 0, 0, 0]"
        );
```

Yes this should fail since the sway function is like so:
```sway
    fn test_array(a: [u64; 10]) -> [u64; 10] {
        a
    }
```

Ideally the function/compiler should throw an error if you pass in an incorrectly length slice (in the test we pass in a slice with 2 items; while the function requires an input slice of 10 items) - however this behaviour seems to be possible at sway level (regardless of this PR).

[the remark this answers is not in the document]
alfiedotwtf: Discussed further on slack.

alfiedotwtf (approved):

forc-plugins/forc-client/src/cmd/call.rs
```diff
@@ -0,0 +1,237 @@
...
+    /// Select the caller to use for the call
+    #[clap(flatten)]
+    pub caller: Caller,
+
+    /// Call parameters to use for the call
+    #[clap(flatten)]
+    pub call_parameters: CallParametersOpts,
+
+    /// The execution mode to use for the call; defaults to dry-run; possible values: dry-run, simulate, live
+    #[clap(long, default_value = "dry-run")]
```
sdankel, line 210: Is it common in other tools to default to dryrun? A user gave feedback on sway playground asking to [default to live](https://github.com/FuelLabs/sway-playground/issues/110) since it currently defaults to dry-run.

sdankel (changes requested):
I'm getting the following error:

```
$ forc call --testnet --abi examples/counter/out/debug/counter-abi.json 0xd8b98e8ceebf08e2fbc4bf0599e8eae239608636b9aab76b5348c2174d197848 initialize_counter 1

error: io: Response errors; Unknown field "indexation" on type "NodeInfo".
```

Does this command work for you? Could be I'm misreading the docs and the request is malformed. If that's the case we should have a more helpful error message.

forc-plugins/forc-client/src/cmd/call.rs
zees-dev: Yeap this is common. 
Most people generally want to make queries - to query state. 
Querying state itself with live mode will cost user funds - we definitely don't want that.
Also if the user does not have funds, then they can't use live mode at all and hence can't perform any operation. 
Due to this dry-run, or simulation modes should be the default imo. 

Users who want to perform state changing operations that costs funds should be explicit (imo).

sdankel (approved):
🚀

JoshuaBatty (approved):

Issue #6669, opened 2024-10-23
Introduce `IsTypeParameter::Yes/No` in the definition of `TypeInfo::UnknownGeneric`


ironcev:
As discussed in this PR comment: https://github.com/FuelLabs/sway/pull/6613#discussion_r1811150004

ironcev:
Closed as per discussion in #6798.

PR #6798, opened 2024-12-20
Introduce IsTypeParameter enum for TypeInfo::UnknownGeneric

amiremohamadi:
## Description
Closes https://github.com/FuelLabs/sway/issues/6669

## Checklist

### DIFF
--- a/sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs
+++ b/sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs
@@ -867,7 +867,7 @@ fn type_check_trait_implementation(
                 // the source file in which the trait is implemented for a type. The `trait_name` points
                 // to the name in the `impl <trait_name> for ...` and is thus a good candidate.
                 let self_type_id = type_engine
-                    .new_unknown_generic_self(trait_name.span(), false)
+                    .new_unknown_generic_self(trait_name.span(), IsTypeParameter::No)
                     .0;
                 if let Some(type_arg) = type_decl.ty.clone() {
                     trait_type_mapping.extend(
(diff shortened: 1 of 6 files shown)


ironcev (changes requested):
Thanks for your contribution @amiremohamadi! Now after seeing the change I actually have a few concerns and would like to hear yours and the opinion of @tritao.

The readability of the function calls has definitely increased. But the change in the struct itself, which I proposed as an expected extension, looks very odd. Essentially, we are mimicking `bool` there and just make things more complex.

Also, the amount of additional code that is written in `impl IsTypeParameter` and needed just for conversion to `bool` is, IMO,  a high price to pay for a simple and local readability increase. This can perhaps be simplified by implementing `From` trait, but still, this even more emphasized that we need `bool` and are mimicking it here which just creates unnecessary complexity.

Another concern is, is this a pattern we actually want to have consistently? There are plenty of structs having `bool` parameters which are obviously `bool` (`is_...`, `has_...`, ...). I assume we want to keep them as `bool`s. Then why changing this one?

Also, there are plenty of functions we have, getting `bool`s as parameters, where the meaning of a single `true` or `false` parameter can be understood only by looking at the function signature. Note that in practice, when writing code in an editor, this is never needed, because the name of the parameter will be displayed in the editor, explaining the meaning. The full advantage of a better readability is there only in cases of reading the code without the support of a modern code editor.

My current thinking is, that we would be better of, if we abandon the attempt altogether. I apologize for not thinking about it in more detail before. When proposed, it looked as a very straightforward step.

@amiremohamadi @tritao What do you think?

ironcev:
My feeling is that, even if we do the change only on the function signatures and not on the structs, it is still "optimizing" this single use case which it happened we have here, where we passe `true` and `false` constants to the functions.

In hindsight, and I apologize once again for not thinking about it more upfront, @amiremohamadi if you also agree, I would skip this change and close both the issue and the PR.

amiremohamadi:
@ironcev i agree too. thanks for being welcoming!

Issue #6805, opened 2025-01-05
Properly handle `Punctuated` without trailing comma everywhere in `swayfmt`


ironcev:
`Punctuated<T, P>` consist of two parts, the `value_separator_pairs` describing all the values followed by the separator, and an optional `final_value_opt` describing the final value not followed by a separator.

Lexical elements that are defined using `Punctuated`, like, e.g., structs, enums, configurables, etc. are not all properly formatted if their elements do not end in a separator. In that case, the `final_value_opt` will be `Some` and will be formatted, but not always properly. In particular, if the formatted element has field alignment defined (`FieldAlignment::AlignFields`), the alignment will be ignored for the last element. There are other issues as well, e.g., the closing curly bracket will not be properly positioned.

Issue #6802, opened 2024-12-30
Properly handle annotations everywhere in `swayfmt`


ironcev:
There are cases in which `swayfmt` does not handle annotations at all, but just strips them away from the formatted code. One example would be stripping them from the last enum variant. Considering the upcoming language development, e.g. the [ABI errors](https://github.com/FuelLabs/sway-rfcs/blob/master/rfcs/0014-abi-errors.md) or derives, and the more prominent use of the `cfg` attribute for experimental features, having proper support for annotations everywhere in the `swayfmt` will become a necessity.

PR #6806, opened 2025-01-05
Refactor `swayfmt` to support arbitrary lexed trees

### LINKED ISSUE #6779: Refactor `swayfmt` to generate code from arbitrary lexed trees, not necessarily backed by source code
In the (near) future, we will want to develop tools that will analyze and transform Sway source code. Two typical examples are:
- the `forc migrate` tool that will migrate existing source code to new versions of Sway, assisting at breaking-changes.
- Sway's equivalent of the [Rust's Clippy tool](https://doc.rust-lang.org/clippy/). 

Such tools usually operates on the AST, traversing or pattern matching it during the analysis, and transforming it to produce the changed source code. 

Sway trees (lexed, parsed, and typed) are not full-fidelity trees in the sense that they do not carry the full source information (e.g. trivia like white-spaces and comments). Also, manipulating the trees "manually" in a program, does not update the spans already contained in those trees.

Supporting full-fidelity trees would require a significant change in the compiler. This issue describes a simpler solution for generating changed code from altered trees.

Although the lexed tree is the least convenient for analysis and tree manipulation, it allows us to reuse `swayfmt` for code generation, assuming the `swayfmt` implementation gets changed, as proposed below.

To get the idea about the kind of changes, e.g., the `forc migrate` tool will be doing in the lexed tree, here are a few concrete examples.

References will require changing `self` function parameters to `&self`: `fn method(self)` -> `fn method(&self)`.
References will require changing `ref mut` function parameters to `&mut`: `fn fun(ref mut x: MyType)` -> `fn method(x: &mut MyType)`.
Supporting `PartialEq` and `Eq` will require inserting trait implementations, e.g., `impl Eq for MyType {}`.

In short, we can have new elements in the tree entered, elements removed, renamed, etc.

The proposal for obtaining the altered source code from the altered lexed tree is to reuse the `swayfmt`. It requires adapting _the implementation_ of the `swayfmt` to not necessarily parse the code from files, but to take any lexed tree as input, no matter how the tree is created, or if there is actual existing source code behind it.

The issue with getting formatted code generated by `swayfmt` from an arbitrary lexed tree lays in the current implementation of the `swayfmt` that expects that every element of the lexed tree has a valid `Span` that points to the valid text in the source code, that contains the textual representation of that lexical element. This is done also in the cases of keywords and e.g. punctuation where the text is actually known. Essentially, when rendering the formatted output, the text is always taken from the spans, means from the original source code.

E.g., referencing like `&mut <type>` is formatted this way:

```
write!(
    formatted_code,
    "{}{}",
    ampersand_token.span().as_str(),  // <<<--- Taking the string "&" from the source code, over `span()`.
    if let Some(mut_token) = mut_token {
        format!("{} ", mut_token.span().as_str())  // <<<--- Taking the string "mut" from the source code, over `span()`.
    } else {
        "".to_string()
    },
)?;
```

E.g., `Ident`s are formatted this way (in the example below, `self` is `Ident`):

```
write!(formatted_code, "{}", self.span().as_str())?  // <<<--- Taking the ident from the source code, over `span()`.
```

The problem is, that in the case of the altered trees, the existing spans can become obsolete or completely meaningless (e.g., `ref mut` is removed from the tree), and the new elements will not have their textual representation in the source file at all (e.g., the inserted `&mut`).

The proposed solution to the problem is fairly straightforward. Instead of taking the strings from the spans/source code, for punctuations, keywords, etc. we can just use their string values, and for `Ident`s the `as_str()` method, which will use the overwritten ident, if available.

Essentially, we would have something like:

```
write!(
    formatted_code,
    "{}{}",
    AmpersandToken::AS_STR,  // AS_STR is an str constant `&`;
    if let Some(mut_token) = mut_token {
        format!("{} ", MutToken::AS_STR) // <<<--- AS_STR is an str constant `mut`;
    } else {
        "".to_string()
    },
)?;

write!(formatted_code, "{}", self.as_str())?
```

I've played with this proposal quite some, and tried out various combinations of altered lexed trees and I can confirm that everything what is needed by the `forc migrate` tool can we achieved if we do the above change in the implementation of the `swayfmt`. Once available, this approach of altering trees and getting source code out of them can be used in all tools that want to structurally change the code. E.g., our own Clippy equivalent one day.

To produce altered trees that will play well with such changed `swayfmt`, one will still need to provide "suitable faked spans". But this will also be straightforward, and we will provide rules for obtaining such spans. Also, altering lexed trees can we supported with an API that will encapsulate the rules.

The proposed change shouldn't affect how the `swayfmt` operates, and all the existing clients, like `forc-fmt`, will remain unchanged.

ironcev:
## Description

This PR refactors `swayfmt` to be able generate code from arbitrary lexed trees. Arbitrary means a lexed tree that can be fully or partially created in-memory by manipulating the tree structure in code. It does not need to necessarily be backed by source code. This is needed to be able to reuse `swayfmt` for generating source code from tools that analyze and modify Sway code, as explained in #6779.

This PR makes the `swayfmt` independent of spans backed by the source code, by doing the following changes:
- Keywords are rendered by using theirs `AS_STR` associated constants.
- The `Token` trait is extended with the `AS_STR` associated constant, and tokens are rendered by using that constant.
- `Ident`s are rendered by using theirs `as_str()` methods. This method takes the textual implementation from `Ident::name_override_opt` if it is provided, and ignores the span in that case.
- `Literal`s are rendered based on the literal value and not their spans. The exception are numeric literals that do not have empty spans. Those are considered to be backed by source code and are rendered as written in code, e.g., `1_000_000u64`.

The PR also fixes some existing bugs in `swayfmt`:
- `use path::{single_import,};` will be formatted as `use path::single_import;`
- `use path::{a,b,};` will be formatted as `use path::{a, b};`
- partial addresses #6802 by rendering annotations for final values.
- partial addresses #6805 by properly rendering final values, but still not using the expected field alignment.

The PR also removes the `path` parameter from the `parse_file`. It was used only to provide an unnecessary dummy `source_id` which was always pointing to the package root file.

Closes #6779.

## Checklist

### DIFF
--- a/sway-ast/src/keywords.rs
+++ b/sway-ast/src/keywords.rs
@@ -98,10 +98,13 @@ pub trait Token: Spanned + Sized {
 
     /// Punctuations that will not follow the token.
     const NOT_FOLLOWED_BY: &'static [PunctKind];
+
+    /// What the string representation of the token is when lexing.
+    const AS_STR: &'static str;
 }
 
 macro_rules! define_token (
-    ($ty_name:ident, $description:literal, [$($punct_kinds:ident),*], [$($not_followed_by:ident),*]) => {
+    ($ty_name:ident, $description:literal, $as_str:literal, [$($punct_kinds:ident),*], [$($not_followed_by:ident),*]) => {
         #[derive(Clone, Debug, Serialize)]
         pub struct $ty_name {
             span: Span,
@@ -132,6 +135,7 @@ macro_rules! define_token (
 
             const PUNCT_KINDS: &'static [PunctKind] = &[$(PunctKind::$punct_kinds,)*];
             const NOT_FOLLOWED_BY: &'static [PunctKind] = &[$(PunctKind::$not_followed_by,)*];
+            const AS_STR: &'static str = $as_str;
         }
 
         impl From<$ty_name> for Ident {
@@ -142,93 +146,121 @@ macro_rules! define_token (
     };
 );
 
-define_token!(SemicolonToken, "a semicolon", [Semicolon], []);
+define_token!(SemicolonToken, "a semicolon", ";", [Semicolon], []);
 define_token!(
     ForwardSlashToken,
     "a forward slash",
+    "/",
     [ForwardSlash],
     [Equals]
 );
 define_token!(
     DoubleColonToken,
     "a double colon (::)",
+    "::",
     [Colon, Colon],
     [Colon]
 );
-define_token!(StarToken, "an asterisk (*)", [Star], [Equals]);
-define_token!(DoubleStarToken, "`**`", [Star, Star], []);
-define_token!(CommaToken, "a comma", [Comma], []);
-define_token!(ColonToken, "a colon", [Colon], [Colon]);
+define_token!(StarToken, "an asterisk (*)", "*", [Star], [Equals]);
+define_token!(DoubleStarToken, "`**`", "**", [Star, Star], []);
+define_token!(CommaToken, "a comma", ",", [Comma], []);
+define_token!(ColonToken, "a colon", ":", [Colon], [Colon]);
 define_token!(
     RightArrowToken,
     "`->`",
+    "->",
     [Sub, GreaterThan],
     [GreaterThan, Equals]
 );
-define_token!(LessThanToken, "`<`", [LessThan], [LessThan, Equals]);
+define_token!(LessThanToken, "`<`", "<", [LessThan], [LessThan, Equals]);
 define_token!(
     GreaterThanToken,
     "`>`",
+    ">",
     [GreaterThan],
     [GreaterThan, Equals]
 );
-define_token!(OpenAngleBracketToken, "`<`", [LessThan], []);
-define_token!(CloseAngleBracketToken, "`>`", [GreaterThan], []);
-define_token!(EqToken, "`=`", [Equals], [GreaterThan, Equals]);
-define_token!(AddEqToken, "`+=`", [Add, Equals], []);
-define_token!(SubEqToken, "`-=`", [Sub, Equals], []);
-define_token!(StarEqToken, "`*=`", [Star, Equals], []);
-define_token!(DivEqToken, "`/=`", [ForwardSlash, Equals], []);
-define_token!(ShlEqToken, "`<<=`", [LessThan, LessThan, Equals], []);
-define_token!(ShrEqToken, "`>>=`", [GreaterThan, GreaterThan, Equals], []);
+define_token!(OpenAngleBracketToken, "`<`", "<", [LessThan], []);
+define_token!(CloseAngleBracketToken, "`>`", ">", [GreaterThan], []);
+define_token!(EqToken, "`=`", "=", [Equals], [GreaterThan, Equals]);
+define_token!(AddEqToken, "`+=`", "+=", [Add, Equals], []);
+define_token!(SubEqToken, "`-=`", "-=", [Sub, Equals], []);
+define_token!(StarEqToken, "`*=`", "*=", [Star, Equals], []);
+define_token!(DivEqToken, "`/=`", "/=", [ForwardSlash, Equals], []);
+define_token!(ShlEqToken, "`<<=`", "<<=", [LessThan, LessThan, Equals], []);
+define_token!(
+    ShrEqToken,
+    "`>>=`",
+    ">>=",
+    [GreaterThan, GreaterThan, Equals],
+    []
+);
 define_token!(
     FatRightArrowToken,
     "`=>`",
+    "=>",
     [Equals, GreaterThan],
     [GreaterThan, Equals]
 );
-define_token!(DotToken, "`.`", [Dot], []);
-define_token!(DoubleDotToken, "`..`", [Dot, Dot], [Dot]);
-define_token!(BangToken, "`!`", [Bang], [Equals]);
-define_token!(PercentToken, "`%`", [Percent], []);
-define_token!(AddToken, "`+`", [Add], [Equals]);
-define_token!(SubToken, "`-`", [Sub], [Equals]);
+define_token!(DotToken, "`.`", ".", [Dot], []);
+define_token!(DoubleDotToken, "`..`", "..", [Dot, Dot], [Dot]);
+define_token!(BangToken, "`!`", "!", [Bang], [Equals]);
+define_token!(PercentToken, "`%`", "%", [Percent], []);
+define_token!(AddToken, "`+`", "+", [Add], [Equals]);
+define_token!(SubToken, "`-`", "-", [Sub], [Equals]);
 define_token!(
     ShrToken,
     "`>>`",
+    ">>",
     [GreaterThan, GreaterThan],
     [GreaterThan, Equals]
 );
-define_token!(ShlToken, "`<<`", [LessThan, LessThan], [LessThan, Equals]);
-define_token!(AmpersandToken, "`&`", [Ampersand], [Ampersand]);
-define_token!(CaretToken, "`^`", [Caret], []);
-define_token!(PipeToken, "`|`", [Pipe], [Pipe]);
+define_token!(
+    ShlToken,
+    "`<<`",
+    "<<",
+    [LessThan, LessThan],
+    [LessThan, Equals]
+);
+define_token!(AmpersandToken, "`&`", "&", [Ampersand], [Ampersand]);
+define_token!(CaretToken, "`^`", "^", [Caret], []);
+define_token!(PipeToken, "`|`", "|", [Pipe], [Pipe]);
 define_token!(
     DoubleEqToken,
     "`==`",
+    "==",
     [Equals, Equals],
     [Equals, GreaterThan]
 );
-define_token!(BangEqToken, "`!=`", [Bang, Equals], [Equals, GreaterThan]);
+define_token!(
+    BangEqToken,
+    "`!=`",
+    "!=",
+    [Bang, Equals],
+    [Equals, GreaterThan]
+);
 define_token!(
     GreaterThanEqToken,
     "`>=`",
+    ">=",
     [GreaterThan, Equals],
     [Equals, GreaterThan]
 );
 define_token!(
     LessThanEqToken,
     "`<=`",
+    "<=",
     [LessThan, Equals],
     [Equals, GreaterThan]
 );
 define_token!(
     DoubleAmpersandToken,
     "`&&`",
+    "&&",
     [Ampersand, Ampersand],
     [Ampersand]
 );
-define_token!(DoublePipeToken, "`||`", [Pipe, Pipe], [Pipe]);
-define_token!(UnderscoreToken, "`_`", [Underscore], [Underscore]);
-define_token!(HashToken, "`#`", [Sharp], []);
-define_token!(HashBangToken, "`#!`", [Sharp, Bang], []);
+define_token!(DoublePipeToken, "`||`", "||", [Pipe, Pipe], [Pipe]);
+define_token!(UnderscoreToken, "`_`", "_", [Underscore], [Underscore]);
+define_token!(HashToken, "`#`", "#", [Sharp], []);
+define_token!(HashBangToken, "`#!`", "#!", [Sharp, Bang], []);
--- a/swayfmt/src/items/item_configurable/mod.rs
+++ b/swayfmt/src/items/item_configurable/mod.rs
@@ -31,7 +31,7 @@ impl Format for ItemConfigurable {
                 .with_code_line_from(LineStyle::Multiline, ExprKind::default()),
             |formatter| -> Result<(), FormatterError> {
                 // Add configurable token
-                write!(formatted_code, "{}", ConfigurableToken::AS_STR,)?;
+                write!(formatted_code, "{}", ConfigurableToken::AS_STR)?;
                 let fields = self.fields.get();
 
                 // Handle opening brace
@@ -91,9 +91,9 @@ impl Format for ItemConfigurable {
                                 }
                             }
                             // Add `:`, `ty` & `CommaToken`
-                            write!(formatted_code, " {} ", ColonToken::AS_STR,)?;
+                            write!(formatted_code, " {} ", ColonToken::AS_STR)?;
                             configurable_field.ty.format(formatted_code, formatter)?;
-                            write!(formatted_code, " {} ", EqToken::AS_STR,)?;
+                            write!(formatted_code, " {} ", EqToken::AS_STR)?;
                             configurable_field
                                 .initializer
                                 .format(formatted_code, formatter)?;
--- a/swayfmt/src/items/item_configurable/mod.rs
+++ b/swayfmt/src/items/item_configurable/mod.rs
@@ -10,7 +10,10 @@ use crate::{
     },
 };
 use std::fmt::Write;
-use sway_ast::{keywords::Token, ConfigurableField, ItemConfigurable};
+use sway_ast::{
+    keywords::{ColonToken, ConfigurableToken, EqToken, Keyword, Token},
+    CommaToken, ConfigurableField, ItemConfigurable,
+};
 use sway_types::{ast::Delimiter, Spanned};
 
 #[cfg(test)]
@@ -28,11 +31,7 @@ impl Format for ItemConfigurable {
                 .with_code_line_from(LineStyle::Multiline, ExprKind::default()),
             |formatter| -> Result<(), FormatterError> {
                 // Add configurable token
-                write!(
-                    formatted_code,
-                    "{}",
-                    self.configurable_token.span().as_str()
-                )?;
+                write!(formatted_code, "{}", ConfigurableToken::AS_STR,)?;
                 let fields = self.fields.get();
 
                 // Handle opening brace
@@ -44,19 +43,18 @@ impl Format for ItemConfigurable {
                 match formatter.config.structures.field_alignment {
                     FieldAlignment::AlignFields(configurable_field_align_threshold) => {
                         writeln!(formatted_code)?;
-                        let value_pairs = &fields
+                        let configurable_fields = &fields
                             .value_separator_pairs
                             .iter()
-                            // TODO: Handle annotations instead of stripping them
-                            .map(|(configurable_field, comma_token)| {
-                                (&configurable_field.value, comma_token)
-                            })
+                            // TODO: Handle annotations instead of stripping them.
+                            //       See: https://github.com/FuelLabs/sway/issues/6802
+                            .map(|(configurable_field, _comma_token)| &configurable_field.value)
                             .collect::<Vec<_>>();
                         // In first iteration we are going to be collecting the lengths of the
                         // struct fields.
-                        let field_length: Vec<usize> = value_pairs
+                        let field_length: Vec<usize> = configurable_fields
                             .iter()
-                            .map(|(configurable_field, _)| configurable_field.name.as_str().len())
+                            .map(|configurable_field| configurable_field.name.as_str().len())
                             .collect();
 
                         // Find the maximum length in the `field_length` vector that is still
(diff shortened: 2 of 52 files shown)


swayfmt/src/items/item_configurable/mod.rs
```diff
@@ -28,11 +31,7 @@ impl Format for ItemConfigurable {
                 .with_code_line_from(LineStyle::Multiline, ExprKind::default()),
             |formatter| -> Result<(), FormatterError> {
                 // Add configurable token
-                write!(
-                    formatted_code,
-                    "{}",
-                    self.configurable_token.span().as_str()
-                )?;
+                write!(formatted_code, "{}", ConfigurableToken::AS_STR,)?;
```
JoshuaBatty, line 34:
```suggestion
                write!(formatted_code, "{}", ConfigurableToken::AS_STR)?;
```

swayfmt/src/items/item_configurable/mod.rs
```diff
@@ -94,24 +91,17 @@ impl Format for ItemConfigurable {
                                 }
                             }
                             // Add `:`, `ty` & `CommaToken`
-                            write!(
-                                formatted_code,
-                                " {} ",
-                                configurable_field.colon_token.ident().as_str(),
-                            )?;
+                            write!(formatted_code, " {} ", ColonToken::AS_STR,)?;
```
JoshuaBatty, line 94:
```suggestion
                            write!(formatted_code, " {} ", ColonToken::AS_STR)?;
```

swayfmt/src/items/item_configurable/mod.rs
```diff
@@ -94,24 +91,17 @@ impl Format for ItemConfigurable {
...
                             configurable_field.ty.format(formatted_code, formatter)?;
-                            write!(
-                                formatted_code,
-                                " {} ",
-                                configurable_field.eq_token.ident().as_str()
-                            )?;
+                            write!(formatted_code, " {} ", EqToken::AS_STR,)?;
```
JoshuaBatty, line 96:
```suggestion
                            write!(formatted_code, " {} ", EqToken::AS_STR)?;
```

swayfmt/src/utils/language/expr/conditional.rs
```diff
@@ -286,11 +290,7 @@ impl Format for MatchBranch {
         formatter: &mut Formatter,
     ) -> Result<(), FormatterError> {
         self.pattern.format(formatted_code, formatter)?;
-        write!(
-            formatted_code,
-            " {} ",
-            self.fat_right_arrow_token.span().as_str()
-        )?;
+        write!(formatted_code, " {} ", FatRightArrowToken::AS_STR,)?;
```
JoshuaBatty, line 293:
```suggestion
        write!(formatted_code, " {} ", FatRightArrowToken::AS_STR)?;
```

swayfmt/src/items/item_enum/mod.rs
```diff
@@ -96,23 +96,13 @@ impl Format for ItemEnum {
                                 }
                             }
                             // Add `:`, ty & `CommaToken`
-                            write!(
-                                formatted_code,
-                                " {} ",
-                                type_field.colon_token.span().as_str(),
-                            )?;
+                            write!(formatted_code, " {} ", ColonToken::AS_STR,)?;
```
JoshuaBatty, line 99:
```suggestion
                            write!(formatted_code, " {} ", ColonToken::AS_STR)?;
```

swayfmt/src/items/item_storage/mod.rs
```diff
@@ -159,33 +163,25 @@ impl Format for ItemStorage {
                                     }
                                 }
                                 // Add `:`, `ty` & `CommaToken`
-                                write!(
-                                    formatted_code,
-                                    " {} ",
-                                    storage_field.colon_token.ident().as_str(),
-                                )?;
+                                write!(formatted_code, " {} ", ColonToken::AS_STR,)?;
```
JoshuaBatty, line 166:
```suggestion
                                write!(formatted_code, " {} ", ColonToken::AS_STR)?;
```

swayfmt/src/items/item_storage/mod.rs
```diff
@@ -159,33 +163,25 @@ impl Format for ItemStorage {
...
                                 storage_field.ty.format(formatted_code, formatter)?;
-                                write!(
-                                    formatted_code,
-                                    " {} ",
-                                    storage_field.eq_token.ident().as_str()
-                                )?;
+                                write!(formatted_code, " {} ", EqToken::AS_STR,)?;
```
JoshuaBatty, line 168:
```suggestion
                                write!(formatted_code, " {} ", EqToken::AS_STR)?;
```

swayfmt/src/items/item_struct/mod.rs
```diff
@@ -52,82 +55,75 @@ impl Format for ItemStruct {
...
                                     required_alignment -= 1;
                                 }
                             }
                             // Add `:`, ty & `CommaToken`
-                            write!(
-                                formatted_code,
-                                " {} ",
-                                type_field.colon_token.span().as_str(),
-                            )?;
+                            write!(formatted_code, " {} ", ColonToken::AS_STR,)?;
```
JoshuaBatty, line 120:
```suggestion
                            write!(formatted_code, " {} ", ColonToken::AS_STR)?;
```

swayfmt/src/items/item_use/mod.rs
```diff
@@ -208,19 +235,15 @@ fn format_use_stmt(
...
+    if item_use.root_import.is_some() {
+        write!(formatted_code, "{}", DoubleColonToken::AS_STR)?;
     }
     item_use.tree.format(formatted_code, formatter)?;
-    write!(
-        formatted_code,
-        "{}",
-        item_use.semicolon_token.span().as_str()
-    )?;
+    write!(formatted_code, "{}", SemicolonToken::AS_STR,)?;
```
JoshuaBatty, line 246:
```suggestion
    write!(formatted_code, "{}", SemicolonToken::AS_STR)?;
```

JoshuaBatty:
Nice one @ironcev great work with this. Being able to format in memory structures is a great addition. Just a couple of code nits with trailing `,` in the macros.

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
     "a double colon (::)",
+    "::",
     [Colon, Colon],
     [Colon]
 );
-define_token!(StarToken, "an asterisk (*)", [Star], [Equals]);
-define_token!(DoubleStarToken, "`**`", [Star, Star], []);
-define_token!(CommaToken, "a comma", [Comma], []);
-define_token!(ColonToken, "a colon", [Colon], [Colon]);
+define_token!(StarToken, "an asterisk (*)", "*", [Star], [Equals]);
```
alfiedotwtf, line 164: Should this be:

> define_token!(StarToken, "an asterisk (*)", "*", [Star], [Equals, Star]);

My thinking is that ti would need the `Star` otherwise `DoubleStarToken` would never match?

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
     [GreaterThan, Equals]
 );
-define_token!(LessThanToken, "`<`", [LessThan], [LessThan, Equals]);
+define_token!(LessThanToken, "`<`", "<", [LessThan], [LessThan, Equals]);
 define_token!(
     GreaterThanToken,
     "`>`",
+    ">",
     [GreaterThan],
     [GreaterThan, Equals]
```
alfiedotwtf, line 181: Do we need to change to the following to:

    [GreateThan, Equals, Colon, LeftParens]

This would allow calls to an implementation function on a generic, and turbofish i.e

    MyTrait::<u8>::add(42)
    my_generic::<u8>(42)

Though... it doesn't look like we currently have `LeftParens` and `RightParens` defined

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
+define_token!(StarToken, "an asterisk (*)", "*", [Star], [Equals]);
+define_token!(DoubleStarToken, "`**`", "**", [Star, Star], []);
+define_token!(CommaToken, "a comma", ",", [Comma], []);
+define_token!(ColonToken, "a colon", ":", [Colon], [Colon]);
 define_token!(
     RightArrowToken,
     "`->`",
+    "->",
     [Sub, GreaterThan],
     [GreaterThan, Equals]
```
alfiedotwtf, line 173: Should this be `[]` since there are no `->>` and `->=` tokens?

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
-define_token!(OpenAngleBracketToken, "`<`", [LessThan], []);
-define_token!(CloseAngleBracketToken, "`>`", [GreaterThan], []);
-define_token!(EqToken, "`=`", [Equals], [GreaterThan, Equals]);
-define_token!(AddEqToken, "`+=`", [Add, Equals], []);
-define_token!(SubEqToken, "`-=`", [Sub, Equals], []);
-define_token!(StarEqToken, "`*=`", [Star, Equals], []);
-define_token!(DivEqToken, "`/=`", [ForwardSlash, Equals], []);
-define_token!(ShlEqToken, "`<<=`", [LessThan, LessThan, Equals], []);
-define_token!(ShrEqToken, "`>>=`", [GreaterThan, GreaterThan, Equals], []);
+define_token!(OpenAngleBracketToken, "`<`", "<", [LessThan], []);
```
alfiedotwtf, line 183: Maybe add the others here too:

    define_token!(OpenAngleBracketToken, "`<`", "<", [LessThan], [LessThan, Equals]);

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
+define_token!(CloseAngleBracketToken, "`>`", ">", [GreaterThan], []);
```
alfiedotwtf, line 184: And here:

    define_token!(CloseAngleBracketToken, "`>`", ">", [GreaterThan], [GreaterThan, Equals]);

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
+    ">>=",
+    [GreaterThan, GreaterThan, Equals],
+    []
+);
 define_token!(
     FatRightArrowToken,
     "`=>`",
+    "=>",
     [Equals, GreaterThan],
     [GreaterThan, Equals]
```
alfiedotwtf, line 203: Should this be [] since there are no `=>>` and `=>=` tokens?

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
 );
-define_token!(DotToken, "`.`", [Dot], []);
-define_token!(DoubleDotToken, "`..`", [Dot, Dot], [Dot]);
-define_token!(BangToken, "`!`", [Bang], [Equals]);
-define_token!(PercentToken, "`%`", [Percent], []);
-define_token!(AddToken, "`+`", [Add], [Equals]);
-define_token!(SubToken, "`-`", [Sub], [Equals]);
+define_token!(DotToken, "`.`", ".", [Dot], []);
```
alfiedotwtf, line 205: Should this be changedd so that `DoubleDotToken` can match?

    define_token!(DotToken, "`.`", ".", [Dot], [Dot]);

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
+define_token!(DoubleDotToken, "`..`", "..", [Dot, Dot], [Dot]);
```
alfiedotwtf, line 206: And then make this: 

    define_token!(DoubleDotToken, "`..`", "..", [Dot, Dot], []);

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
+define_token!(BangToken, "`!`", "!", [Bang], [Equals]);
+define_token!(PercentToken, "`%`", "%", [Percent], []);
+define_token!(AddToken, "`+`", "+", [Add], [Equals]);
+define_token!(SubToken, "`-`", "-", [Sub], [Equals]);
```
alfiedotwtf, line 210: Should this be `define_token!(SubToken, "`-`", "-", [Sub], [GreaterThan]);` so that `RightArrowToken` can match?

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
 define_token!(
     ShrToken,
     "`>>`",
+    ">>",
     [GreaterThan, GreaterThan],
     [GreaterThan, Equals]
```
alfiedotwtf, line 216: Could this be `[]` since there's no `>>>` and `>>=` tokens?

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
+);
+define_token!(AmpersandToken, "`&`", "&", [Ampersand], [Ampersand]);
+define_token!(CaretToken, "`^`", "^", [Caret], []);
+define_token!(PipeToken, "`|`", "|", [Pipe], [Pipe]);
 define_token!(
     DoubleEqToken,
     "`==`",
+    "==",
     [Equals, Equals],
     [Equals, GreaterThan]
```
alfiedotwtf, line 233: And `[]` here as there'ss no `===` and `==>` tokens?

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
 );
-define_token!(BangEqToken, "`!=`", [Bang, Equals], [Equals, GreaterThan]);
+define_token!(
+    BangEqToken,
+    "`!=`",
+    "!=",
+    [Bang, Equals],
+    [Equals, GreaterThan]
```
alfiedotwtf, line 240: And `[]` here as no `!==` and `!=>` tokens?

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
+);
 define_token!(
     GreaterThanEqToken,
     "`>=`",
+    ">=",
     [GreaterThan, Equals],
     [Equals, GreaterThan]
```
alfiedotwtf, line 247: And `[]` here as no `>==` and `>=>` tokens?

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
 );
 define_token!(
     LessThanEqToken,
     "`<=`",
+    "<=",
     [LessThan, Equals],
     [Equals, GreaterThan]
```
alfiedotwtf, line 254: And `[]` here as no `<==` and `<=<` tokens?

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
 );
 define_token!(
     DoubleAmpersandToken,
     "`&&`",
+    "&&",
     [Ampersand, Ampersand],
     [Ampersand]
```
alfiedotwtf, line 261: And `[]` here as no `&&&$` token?

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
 );
-define_token!(DoublePipeToken, "`||`", [Pipe, Pipe], [Pipe]);
-define_token!(UnderscoreToken, "`_`", [Underscore], [Underscore]);
-define_token!(HashToken, "`#`", [Sharp], []);
-define_token!(HashBangToken, "`#!`", [Sharp, Bang], []);
+define_token!(DoublePipeToken, "`||`", "||", [Pipe, Pipe], [Pipe]);
+define_token!(UnderscoreToken, "`_`", "_", [Underscore], [Underscore]);
```
alfiedotwtf, line 264: And `[]` here as no `__` token?

sway-ast/src/keywords.rs
```diff
@@ -142,93 +146,121 @@ macro_rules! define_token (
...
+define_token!(HashToken, "`#`", "#", [Sharp], []);
```
alfiedotwtf, line 265: Should this be changes to the following so that `HashBangToken` can match?

    define_token!(HashToken, "`#`", "#", [Sharp], [Bang]);

swayfmt/src/items/item_configurable/mod.rs
ironcev: Good catch! I've run a search for `,)` and found a several more occurrences. All are removed now.

ironcev:
@alfiedotwtf Thanks for such a detailed review of token definitions! To keep this PR focused, I propose to have these potential changes as a separate PR, and combine it with the redefiniton of the `&` token which we already know we need to do to support references to references, e.g., `&&&T`.

I've opened a separate issue for that: #6808. Is this ok for you?

swayfmt/src/items/item_trait_type.rs
```diff
@@ -4,7 +4,10 @@ use crate::{
     utils::map::byte_span::{ByteSpan, LeafSpans},
 };
 use std::fmt::Write;
-use sway_ast::{keywords::Token, TraitType};
+use sway_ast::{
+    keywords::{EqToken, Keyword, SemicolonToken, Token, TypeToken},
+    TraitType,
+};
 use sway_types::Spanned;
```
alfiedotwtf, line 12: Meta question: there's a lot of `write!()` calls and so was wondering about performance... 

Here's write:

    struct EventCapture {
        pub output: Arc<Mutex<String>>,
    }

    impl Write for EventCapture {
        fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
            let mut lock = self.output.lock().unwrap();
            lock.push_str(&String::from_utf8_lossy(buf));
            Ok(buf.len())
        }
    }

And inside `push_str()`, we have:

    pub fn push_str(&mut self, string: &str) {
        self.vec.extend_from_slice(string.as_bytes())
    }

Reading the docs, it seems that `extend_from_slice()` will clone /**the entire contents** of all previous values/ for every push before adding the new value i.e running in `Θ((n*(n+1))/2))`.

I think we should change `EventCapture` to something like this:

    struct EventCapture {
        pub output: Arc<Mutex<Vec<String>>>,
    }

    impl Write for EventCapture {
        fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
            let mut lock = self.output.lock().unwrap();
            lock.push(String::from_utf8_lossy(buf));
            Ok(buf.len())
        }
    }

If I'm thinking straight, this would make `write!()` run at `Θ(1)`.

ironcev: Looking at the methods, the proposed `String::push()` cannot push a `String` onto `String`, just a single `char`. And in case of a multi-byte `char` it will internally call `extend_from_slice`.

> it seems that extend_from_slice() will clone /the entire contents of all previous values for every push before adding the new value

I am not quite sure if I understand this line. `extend_from_slice` will clone the values from the slice to append.

alfiedotwtf:
> I've opened a separate issue for that: https://github.com/FuelLabs/sway/issues/6808. Is this ok for you?

My comments may not have been correct, but creating a new issue so we don't derail this PR is a good idea :)

alfiedotwtf:
I've got no other suggestions to the PR. With my `write!()` comments, let's ignore it for this PR so I'll approve now, but I think it could be worth testing out that possible big optimisation win.

alfiedotwtf (approved):

sway-ast/src/keywords.rs
ironcev: As agreed, all token definitions will be checked in #6808.

bitzoic (approved):
No more commas!

IGI-111 (approved):

PR #6816, opened 2025-01-06
Cleanup stale lock files in the .lsp-locks directory

JoshuaBatty:
## Description
I noticed there were quite a few old stale lock files in the `.lsp-locks` directory. We now check that the PID is active for each file in this directory, if it's not active then it's removed. 

## Checklist


forc-util/src/fs_locking.rs
```diff
@@ -201,4 +236,32 @@ mod test {
...
+        // Verify that only the invalid lock file was cleaned up
+        assert_eq!(cleaned_paths.len(), 1);
+        assert_eq!(cleaned_paths[0], lock_path);
+
+        // Verify file system state
+        assert!(test_lock.0.exists(), "Active lock file should still exist");
+        assert!(!lock_path.exists(), "Lock file should be removed");
+
+        // Cleanup after test
+        test_lock.release().expect("Failed to release test lock");
```
zees-dev, line 265: nit: not a biggie, maybe we should use tempfile to ensure file/dir cleanup (automatically done when var goes out of scope - even if a test fails mid-way)

forc-util/src/fs_locking.rs
```diff
@@ -136,11 +139,43 @@ impl PidFileLocking {
         fs.flush()?;
         Ok(())
     }
+
+    /// Cleans up all stale lock files in the .lsp-locks directory
+    /// Returns a vector of paths that were cleaned up
```
zees-dev, line 144: nit:
```suggestion
/// Cleans up stale lock files in the .lsp-locks directory.
/// A lock file is considered stale if:
/// - It has a .lock extension
/// - Either its contents cannot be parsed as a valid PID
/// - Or the PID it contains is no longer active
/// 
/// Returns paths of the lock files that were removed.
```

zees-dev (approved):
Left a few optional minor improvements (maybe opinionated here); nothing major - feel free to use them 😄 
LGTM 👍

forc-util/src/fs_locking.rs
```diff
@@ -136,11 +139,43 @@ impl PidFileLocking {
...
+        for entry in entries {
+            let entry = entry?;
+            let path = entry.path();
+            if let Some(ext) = path.extension().and_then(|ext| ext.to_str()) {
+                if ext == "lock" {
+                    if let Ok(mut file) = File::open(&path) {
+                        let mut contents = String::new();
+                        if file.read_to_string(&mut contents).is_ok() {
+                            if let Ok(pid) = contents.trim().parse::<usize>() {
+                                if !Self::is_pid_active(pid) {
```
alfiedotwtf, line 159: I've stopped using PID files because I've been burned by this before - the call to `is_pid_active()` can still return true on servers with long uptimes that are burning through processes, which causes PIDs to wrap. On some systems the PID space is as small as 32768 but on my beast it's:

    > cat /proc/sys/kernel/pid_max 
    4194304

Instead, I now use advisory locking (the following is code from `forc-telemetry`), which also has the nice advantage in that it's shorter code as you don't need to read the file:

    self.logfile_lock = match Flock::lock(logfile, FlockArg::LockExclusiveNonblock) {
            Ok(logfile_lock) => Some(logfile_lock),
            Err((_, Errno::EWOULDBLOCK)) => {
                // Silently exit as another Collector is already running
                exit(0);
            }
    }

forc-util/src/fs_locking.rs
```diff
@@ -136,11 +139,43 @@ impl PidFileLocking {
...
+                                    remove_file(&path)?;
+                                    cleaned_paths.push(path);
+                                }
+                            } else {
```
alfiedotwtf, line 163: I don't think the `else` should do clean up as there's a race condition - the other process could be in the middle of writing it's PID into the lockfile itself 😅

kayagokalp:
This made me check the forc locks and same thing happens there too, without any active forc process I have more than 50 locks in `~/.forc/.locks` 🤯

sdankel (approved):

PR #6823, opened 2025-01-09
Add splice method to Bytes

SwayStar123:
## Description
Adds a splice method to bytes that lets you select start and end indices, and returns the range of bytes, removing them from the original bytes. 

## Checklist


K1-R1:
I like the addition of splice, but I think we should match Rust's impl https://doc.rust-lang.org/std/vec/struct.Vec.html#method.splice by taking a second optional param replace_with

sway-lib-std/src/bytes.sw
```diff
@@ -719,6 +719,98 @@ impl Bytes {
...
+                ptr: new_ptr,
+                cap: new_len,
+            };
+        }
+
+        let tail_len = self.len - end;
+        if splice_len != replace_with.len() && tail_len > 0 {
+            let src = self.buf.ptr().add_uint_offset(end);
+            let dst = self.buf.ptr().add_uint_offset(start + replace_with.len());
+            // Use temporary buffer to safely handle overlapping memory
```
xunilrj, line 798: Do you need this temp if you read [end..] from the original buffer? I think you don´t, and you could save this temp copy.

test/src/in_language_tests/test_programs/bytes_inline_tests/src/splice.sw
```diff
@@ -0,0 +1,186 @@
...
+
+    // bytes => [7, 9]
+    assert(bytes.len() == 2);
+    assert(bytes.get(0).unwrap() == b);
+    assert(bytes.get(1).unwrap() == c);
+
+    // spliced => [5]
+    assert(spliced.len() == 1);
+    assert(spliced.get(0).unwrap() == a);
+}
```
jjcnn, line 186: This case additonally tests one of the  situations where the end of the original buffer overlaps with the end of the new buffer, i.e., the case where you need the temporary buffer.

You should probably add a test for the other case, i.e, 

```
// bytes = [5, 7, 9]
// Remove [0..1] => [5], replace with [10, 12]
```

jjcnn (changes requested):

jjcnn (approved):

K1-R1 (approved):

K1-R1 (approved):

jjcnn (approved):

PR #6834, opened 2025-01-16
[asmgen] bugfixes in `const_indexing_aggregates_function`

### LINKED ISSUE #6810: Unhandled Bailout During AbstractInstructi
From https://reports.immunefi.com/fuel-network-or-attackathon/attackathon-_-fuel-network-32438-smart-contract-low-unhandled-bailout-during-abstractinstructionset

const_indexing_aggregates_function() process VirtualOp::LW without handling the case if the offset in addr_reg is not aligned to 8, resulting in the old constant being used and leading to incorrect program behavior.

## Vulnerability Details
While processing a VirtualOp::LW with non-aligned address offset, const_indexing_aggregates_function() does not clear register information in reg_contents and setup a new definition with record_new_def(). This means the old state of the register stored in reg_contents will still be used, causing incorrect value tracking after the LW instruction. The incorrect tracking may then cause incorrect immediate replacement in other instructions.

```
VirtualOp::LW(dest, addr_reg, imm) => match reg_contents.get(addr_reg) {
    Some(RegContents::BaseOffset(base_reg, offset))
        if get_def_version(&latest_version, &base_reg.reg) == base_reg.ver
            && ((offset / 8) + imm.value as u64)
                < compiler_constants::TWELVE_BITS =>
    {
        // bail if LW cannot read where this memory is
        if offset % 8 == 0 {
            let new_imm = VirtualImmediate12::new_unchecked(
                (offset / 8) + imm.value as u64,
                "Immediate offset too big for LW",
            );
            let new_lw =
                VirtualOp::LW(dest.clone(), base_reg.reg.clone(), new_imm);
            // The register defined is no more useful for us. Forget anything from its past.
            reg_contents.remove(dest);
            record_new_def(&mut latest_version, dest);
            // Replace the LW with a new one in-place.
            *op = new_lw;
        }
    }
    _ => {
        reg_contents.remove(dest);
        record_new_def(&mut latest_version, dest);
    }
},
```
## Impact Details
As usual, it is hard to come up with a precise impact estimation of incorrect code generation because it depends on what code the user writes. The best case scenario would be contracts that run into those bugs getting bricked, and the worst case scenario would be that incorrect program behaviors lead to loss of funds.

## References
https://github.com/FuelLabs/sway/blob/7b56ec734d4a4fda550313d448f7f20dba818b59/sway-core/src/asm_generation/fuel/optimizations.rs#L155

## Proof of concept
This test would fail because addi a a i15 incorrectly uses the reg values before lw a a i0

```
#[test]
fn incorrect_bailout() -> u64 {
    let a = asm(a, b) {
        movi a i32;		// a = 32
        aloc a;			// hp = buf[0;32]

        movi a i1;		// a = 1
        addi b hp i23;	// b = &buf[23]
        sb b a i0;		// buf[23] = a = 1
        addi a hp i9;	// a = &buf[9]
        addi b hp i1;	// b = &buf[1]
        sb b a i7;		// buf[1:9] = a = &buf[9] avoid using sw, which is buggy itself
        srli a a i8;
        sb b a i6;
        srli a a i8;
        sb b a i5;
        srli a a i8;
        sb b a i4;
        srli a a i8;
        sb b a i3;
        srli a a i8;
        sb b a i2;
        srli a a i8;
        sb b a i1;
        srli a a i8;
        sb b a i0;

        addi a hp i1;	// a = &buf[1]
        lw a a i0;		// a = buf[1:9] = &buf[9]
        addi a a i15;	// expected : a = &buf[24]              real : a = &buf[16]
        lw a a i0;		// expected : a = buf[24:32] = 0        real : a = bug[16:24] = 1
        a: u64
    };
    assert(a == 0);
    a
}
```

### LINKED ISSUE #6812: Missing Alignment Check During AbstractIns
From https://reports.immunefi.com/fuel-network-or-attackathon/attackathon-_-fuel-network-32439-smart-contract-low-missing-alignment-check-during-abstractinstructi

const_indexing_aggregates_function process VirtualOp::SW without checking the offset in addr_reg is aligned to 8, cause the wrong constant being calculated and leads to incorrect program behavior.

## Vulnerability Details
const_indexing_aggregates_function is an optimization pass of AbstractInstructionSet, it is used to propagate constants in the function. During the handling of VirtualOp::SW instruction, the function does not properly validate the offset of BaseOffset. Since the imm argument of VirtualOp::SW representing an offset equals to imm * 8, the conversion (divide by 8) here will truncate offset which isn't aligned to 8 bytes, causing SW instruction to write to incorrect address.

```
VirtualOp::SW(addr_reg, src, imm) => match reg_contents.get(addr_reg) {
    Some(RegContents::BaseOffset(base_reg, offset))
        if get_def_version(&latest_version, &base_reg.reg) == base_reg.ver
            && ((offset / 8) + imm.value as u64)
                < compiler_constants::TWELVE_BITS =>
    {
        let new_imm = VirtualImmediate12::new_unchecked(
            (offset / 8) + imm.value as u64,
            "Immediate offset too big for SW",
        );
        let new_sw = VirtualOp::SW(base_reg.reg.clone(), src.clone(), new_imm);
        // Replace the SW with a new one in-place.
        *op = new_sw;
    }
    _ => (),
},
```

## Impact Details
As usual, it is hard to come up with a precise impact estimation of incorrect code generation because it depends on what code the user writes. The best case scenario would be contracts that run into those bugs getting bricked, and the worst case scenario would be that incorrect program behaviors lead to loss of funds.

## References
https://github.com/FuelLabs/sway/blob/7b56ec734d4a4fda550313d448f7f20dba818b59/sway-core/src/asm_generation/fuel/optimizations.rs#L169

## Proof of Concept
This test would fail because buf[16] is not overwritten by sw b a i1.

```
#[test]
fn sw_missing_alignment_check() -> u64 {
    let a = asm(a, b) {
        movi a i24;     // a = 24
        aloc a;         // hp = buf[0;24]

        movi a i1;      // a = 1
        sb hp a i16;    // buf[16] = a = 1

        movi a i0;      // a = 0
        addi b hp i1;   // b = &buf[1]
        sw b a i1;      // expected : buf[9:17] = a = 0     real : buf[8:16] = a = 0

        lb a hp i16;    // a = &buf[16]
        a: u64
    };

    assert(a == 0);
    a
}
```

### LINKED ISSUE #6813: Unhandled Side Effect During AbstractInstr
From https://reports.immunefi.com/fuel-network-or-attackathon/attackathon-_-fuel-network-32453-smart-contract-low-unhandled-side-effect-during-abstractinstruction

const_indexing_aggregates_function doesn't clear tracked registers modified through side effects.

## Vulnerability Details
If an unknow instruction is encountered while constant folding, the tracked values of registers modified by it must be cleared. However, the current implementation only clears states for def_registers. This means that any Constant registers, such as hp, will not have it's state removed properly. This may result in incorrect tracking and cause incorrect immediate replacement in other instructions.

```
_ => {
    // For every Op that we don't know about,
    // forget everything we know about its def registers.
    for def_reg in op.def_registers() {
        reg_contents.remove(def_reg);
        record_new_def(&mut latest_version, def_reg);
    }
}
```

## Impact Details
As usual, it is hard to come up with a precise impact estimation of incorrect code generation because it depends on what code the user writes. The best case scenario would be contracts that run into those bugs getting bricked, and the worst case scenario would be that incorrect program behaviors lead to loss of funds.

## References
https://github.com/FuelLabs/sway/blob/9a52d74f4de90fb601d53d3210c57a714527c643/sway-core/src/asm_generation/fuel/optimizations.rs#L178

## Proof of Concept
This test would fail because lw a a i0 incorrectly uses the old hp values before the second aloc b

```
#[test]
fn side_effect_register_not_cleared() -> u64 {
    let a = asm(a, b) {
        movi b i16;     // b = 16
        aloc b;         // buf1 = [0;16]
        movi b i0;      // b = 0
        sw hp b i0;     // buf1[0:8] = b = 0
        movi a i0;      // a = 0
        add a hp a;     // a = &buf1
        movi b i16;     // b = 16
        aloc b;         // buf2 = [0;16]
        movi b i1;      // b = 1
        sw hp b i0;     // buf2[0:8] = b = 1
        lw a a i0;      // expected : a = buf1[0:8] = 0         real : a = buf2[0:8] = 1
        a
    };
    assert(a == 0);
    a
}
```

vaivaswatha:
## Description

Fixes #6810, fixes #6812 and fixes #6813


IGI-111 (approved):

jjcnn (approved):
LGTM

PR #6842, opened 2025-01-18
CompileError::ContractIdConstantNotAConstDecl

### LINKED ISSUE #6841: Eliminate CompileError::ContractIdConstantNotAConstDecl
The error `CompileError::ContractIdConstantNotAConstDecl` is only used in `contract_helpers.rs`, and is only thrown if the typecheck of the `CONTRACT_ID` declaration fails. Since this declaration is generated by the compiler, this typecheck should never fail, and if it does it should be an internal compiler error instead.

mo-hak:
## Description
In contract_helpers.rs, replaced the ContractIdConstantNotAConstDecl error with an Internal compiler error since this represents an impossible state that should never occur in normal operation.
In error.rs, removed the now unused ContractIdConstantNotAConstDecl error variant and its corresponding span implementation.
The changes make sense because:
The error was only used in one place where it represented an impossible state
If that state is ever reached, it indicates a bug in the compiler itself rather than user error
Using an internal compiler error will make it clearer that this is a compiler bug that needs to be fixed if it ever occurs
fix #6841 

## Checklist


jjcnn (dismissed):

sway-core/src/semantic_analysis/namespace/contract_helpers.rs
```diff
@@ -105,9 +105,10 @@ fn bind_contract_id_in_root_module(
...
         TyAstNodeContent::Declaration(_) => Ok(namespace.root()),
         _ => Err(
-            handler.emit_err(CompileError::ContractIdConstantNotAConstDecl {
-                span: const_item_span,
-            }),
+            handler.emit_err(CompileError::Internal(
+                "Contract ID declaration did not typecheck to a declaration, which should be impossible",
+                const_item_span,
+            )),
         ),
```
IGI-111, line 112: This should probably be an if-let.

mo-hak: Ah yes, you are right. Let me modify and commit it !

jjcnn (approved):

IGI-111 (approved):

Issue #5892, opened 2024-04-23
enforce impl coherence


IGI-111:
Collect all impls in packages and enforce a strict set of coherence rules on them (likely the same as Rust's orphan rules).

Blocked on https://github.com/FuelLabs/sway/issues/5379 as we need to resolve all the relevant symbols and names.

PR #6844, opened 2025-01-20
Improved trait coherence checking

tritao:
## Description

This PR refines trait coherence checking to guarantee that there is only one valid implementation for each trait. It introduces improved orphan rules and more rigorous overlap checks to prevent conflicting trait implementations. With these changes, the system now more reliably enforces coherence, ensuring that traits are implemented uniquely, and
reducing potential ambiguities.

This approach mirrors Rust’s design by enforcing two distinct safety and coherence checks on trait implementations:

**1. Orphan Rules Check**
The orphan rules require that for any trait implementation, either the trait or the type must be defined within the current package. This restriction prevents external packages from implementing foreign traits for foreign types, which could otherwise lead to conflicting implementations and ambiguities across different parts of a codebase. Essentially, it helps maintain clear ownership and boundaries of trait implementations.

**2. Overlap Impl Check**
The overlap impl check ensures that no two trait implementations can apply to the same type in an ambiguous manner. If two implementations could potentially match the same type, it would be unclear which one should be used, leading to coherence issues. By enforcing that implementations do not overlap, the system guarantees that trait resolution is unambiguous and predictable.

Together, these checks promote a robust and maintainable system by ensuring that trait implementations are both locally controlled (orphan rules) and non-conflicting (overlap check).

The test suite has been updated to comply with the new rules. However, there is one current limitation regarding arrays. For arrays, the coherence checks have been relaxed to avoid the need for numerous concrete implementations in the standard library for traits like `Eq` and `PartialEq`. This decision was made because support for const generics is expected soon, which will allow these traits to be implemented more cleanly.

Closes issue https://github.com/FuelLabs/sway/issues/5892.

## Checklist


IGI-111 (approved):
LGTM except for the LSP test issues

forc-util/src/fs_locking.rs
```diff
@@ -247,6 +247,7 @@ mod test {
     }
 
     #[test]
+    #[ignore]
```
JoshuaBatty, line 250:
```suggestion
```

Can we remove this as this test was addressed in #7055

kayagokalp:
Is this a breaking change?

Voxelot:
> Is this a breaking change?

Hard to say, as code that relied on the previous behavior would likely qualify as UB, as it wouldn't be predictable what conflicting implementation of a trait would take precendence.

JoshuaBatty (dismissed):

kayagokalp:
> > Is this a breaking change?
> 
> Hard to say - code that relied on the previous behavior would likely qualify as UB as it wouldn't be predictable what conflicting implementation of a trait would take precendence.

Yeah, I was confused because of that. Should we be conservative here and mark this as breaking? It is possible that some projects that were building is going to be failing with this, if i interpret things correctly.

IGI-111 (approved):
Incoherent implementations were always illegal if not enforced so this is a bug fix and not a breaking change.

JoshuaBatty (approved):

kayagokalp (approved):

PR #6847, opened 2025-01-21
Draft: consider importing submodules into namespace using glob imports

### LINKED ISSUE #6843: Consider importing submodules into namespace when using glob imports
Currently, the glob import `*` imports only items, but not the submodules. This is actually consistent with the fact that we currently cannot import a submodule. E.g., `use core::ops;` is an invalid import that will result in `Could not find symbol "ops" in this scope.` error.

Compared to Rust, this is a limitation, and we can consider enhancing it.

Here is an example:

```sway
library;

use core::*; // <<<--- `ops` and `storage` submodules not imported here.

struct S { }

impl ops::Eq for S {
//   ^^^ Could not find symbol "ops" in this scope.     <<-- ERROR.
        true
    }
}

#[inline(never)]
pub fn play() {
    let _ = core::storage::StorageKey::<u64>::new(b256::zero(), 0, b256::zero());
    let _ = storage::StorageKey::<u64>::new(b256::zero(), 0, b256::zero());
    //      ^^^^^^^ Module "impl_eq::use_core_star::storage" could not be found.   <<-- ERROR.
}
```

amiremohamadi:
## Description
Fix #6843

## Checklist

### DIFF
--- a/sway-core/src/semantic_analysis/namespace/root.rs
+++ b/sway-core/src/semantic_analysis/namespace/root.rs
@@ -181,7 +181,7 @@ impl Root {
     ) -> Result<(), ErrorEmitted> {
         self.check_module_privacy(handler, engines, src, dst)?;
 
-        let src_mod = self.module.lookup_submodule(handler, engines, src)?;
+        let src_mod = self.module.lookup_submodule(handler, engines, src)?.clone();
 
         let mut decls_and_item_imports = vec![];
 
@@ -258,6 +258,7 @@ impl Root {
         }
 
         let implemented_traits = src_mod.root_items().implemented_traits.clone();
+
         let dst_mod = self.module.lookup_submodule_mut(handler, engines, dst)?;
         dst_mod
             .current_items_mut()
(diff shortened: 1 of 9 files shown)


amiremohamadi:
i'm still testing the edge cases but feel free to do a review!
cc: @ironcev

jjcnn (changes requested):
Thank you for your contribution, but I think this approach needs to be rethought.

I have just spent a long time trying to eliminate unnecessary cloning of modules, so I would very much appreciate of you could limit cloning of modules to an absolute minimum. 

In particular, please don't clone an entire module including all its submodules when only the public submodules are visible.

More importantly, it would be much more beneficial if you could avoid cloning modules altogether and use references instead.

This is not an easy task to solve - it requires some fairly deep design considerations, both wrt. semantics and wrt. avoiding performance degredation.

jjcnn:
@amiremohamadi : I have just added a corner case to #6843 that shows why we will need a complete rewrite of how we represent submodules in order to solve this issue.

amiremohamadi:
Ah right. It makes sense

PR #6867, opened 2025-01-28
Removes TraitMap::insert_for_type.

### LINKED ISSUE #5002: Traits map can grow significantly
In the example below each `d.contains(Data2 { x: 42, y: 42 })` makes the traits map grow which might slow down compilation time.
We should probably deduplicate the map entries.

```
contract;

trait Eq {
    fn eq(self, other: Self) -> bool;
}

impl Eq for u64 {
        __eq(self, other)
    }
}

struct Data2<T, K> {
  x: T,
  y: K
}

impl<T,K> Data2<T,K> {
  fn contains(self, other: Self) -> bool where T: Eq, K: Eq  {
    true
  }
}

fn main() -> u64 {
  let d = Data2 { x: 42, y: 42 };
  d.contains(Data2 { x: 42, y: 42 });
  d.contains(Data2 { x: 42, y: 42 });
  d.contains(Data2 { x: 42, y: 42 });
/// ... and a bunch of d.contains(Data2 { x: 42, y: 42 });

  42
}
```

### LINKED ISSUE #6858: Disambiguate the associated function `Into` implementation not recognized and not automatically selected
https://github.com/FuelLabs/sway/pull/6857 introduces `Into<Bytes>` for a number of stack types. However, if the following syntax is used the compiler will not automatically recognize which `Into` implementation is the correct one.

```sway
let my_asset = AssetId::zero();
let my_bytes: Bytes = my_asset.into();
```

Image

As a result, the following is required as a work around.

```sway
let my_asset = AssetId::zero();
let my_bytes: Bytes = <AssetId as Into<Bytes>>::into(my_asset);
```

The compiler should automatically determine which `Into` implementation is appropriate.

esdrubal:
## Description

Some use cases require the compiler to call insert_implementation_for_type on every call of find_method_for_type.
This would double the compile time of a simple script. To avoid this we removed the TraitMap::insert_for_type. The TraitMap now only stores the original implementations and uses a less restrictive unify_check to get those that match the concrete implementations.

This significantly reduces the TraitMap's size and speeds up the lookup times. On the other hand, it also introduces a type substitution on every find_method_for_type.

A future PR will address the performance of doing type substitutions.

Fixes #5002.
Fixes #6858

## Checklist


jjcnn (commented):
I'm not 100% I understand what's going on, but it looks ok.

The two comments I've made are mostly something that looks odd - I don't know if they're actual mistakes. Maybe it would be worth adding comments in those places to avoid future confusion.

jjcnn (approved):
LGTM

JoshuaBatty:
nice 10% speed increase :)

IGI-111 (approved):

PR #6902, opened 2025-02-07
fix: remove invalid unpacking logic for IPFS pinned pkg sources

### LINKED ISSUE #6721: Forc IPFS package fetching capabilities does not work with Pinata pinned packages
## Problem

We cannot use forc to fetch packages pinned by pinata. Pinata is a widely used pinning service making this limitation an important blocker in terms of reaching a working package registry and reduces the overall usability of IPFS support forc-pkg currently offers.

We are also planning to use pinata for package registry to make things a little bit easier. So it is for best if we solve this before going forward.

kayagokalp:
## Description
closes #6721.

There were couple of issues:

1. Archive unpacking was getting the bytes from `text` field of response to the IPFS gateway which was adding additional stuff that messes with unpacking routine.
2. Folder name was invalid so forc was unable to find the packages in the `ipfs` cache.

Both issues are addressed and this unblocks forc to fetch ipfs fetched packages once again. Next up on our package registry push is to enable alternative sources and set the precedence for them.

Also added some tests around extracting logic, which inserts the contents under a directory named after `CID` as this is how `forc` is looking for these sources in the first place (related to point 2 above)


JoshuaBatty (approved):
LGTM

forc-pkg/src/source/ipfs.rs
```diff
@@ -130,6 +131,24 @@ impl fmt::Display for Pinned {
...
+    fn extract_archive<R: std::io::Read>(&self, reader: R, dst: &Path) -> Result<()> {
+        let dst_dir = dst.join(self.0.to_string());
+        std::fs::create_dir_all(&dst_dir)?;
+        let mut archive = Archive::new(reader);
+
+        for entry in archive.entries()? {
+            let mut entry = entry?;
+            let path = entry.path()?;
+            let new_path = dst_dir.join(path);
+            // Create parent directories if they don't exist
```
zees-dev, line 143: Since the extracted content may not necessarily be trusted; could add security check to prevent path traversal attack(s):
```suggestion
            // Prevent path traversal attacks by checking if path is absolute or contains ..
            if path.is_absolute() || path.components().any(|c| c == Component::ParentDir) {
                return Err(anyhow::anyhow!("Invalid path in archive: contains absolute path or parent directory references"));
            }
```

zees-dev (approved):
Raised a potential issue; don't want it to be a blocker though - hence approving.
LGTM 👍

forc-pkg/src/source/ipfs.rs
kayagokalp: Ahh that is a good catch! Looks like [`Entry::unpack`](https://docs.rs/tar/0.4.38/tar/struct.Entry.html#method.unpack) does not have the built-in checks as `Archive::unpack`.

Instead there is [`Entry::unpack_in`](https://docs.rs/tar/0.4.38/tar/struct.Entry.html#method.unpack_in) which includes this checks to prevent zip slip happening.

kayagokalp: [5cc8325](https://github.com/FuelLabs/sway/commit/5cc83256f16e1d7d3e02d62d75e5034655e73494) adds a test to demonstrate this is possible. Once CI finalizes and shows that the test is effective, I'll enable the fix

kayagokalp: [ed29a64](https://github.com/FuelLabs/sway/pull/6902/commits/ed29a64fd7ad0237b9d052f319a4398e58d7099f) fixes this.

zees-dev (approved):
LGTM 👍

JoshuaBatty (approved):

PR #6905, opened 2025-02-09
Add support for fetching verfied ABI from registry for forc call

### LINKED ISSUE #6893: Add support for fetching verified ABI from registry for forc call
`forc call` can currently utilize an ABI provided as local path or from URL.
We want to update this behaviour such that the ABI can also be provided from verified ABI registry (when it's ready).
- Also, the default behaviour should be this; i.e. if no ABI is provided explicitly in `forc call`, we should try fetch the ABI from registry.

Dhanraj30:
## Description
Issue: Closes #6893
Add support for fetching ABI from registry in forc call

##Changes:

Modified call.rs to default to "registry" if no ABI is provided.
Updated mod.rs to fetch the ABI from https://registry.fuel.network/abi/{contract_id}.
Added error handling for failed fetch requests.

Behavior:

If ABI is explicitly provided via path or URL, it will be used.
If no ABI is provided, forc call will attempt to fetch from the registry.
If fetching fails, an error message is returned.

### DIFF
--- a/forc-plugins/forc-client/src/cmd/call.rs
+++ b/forc-plugins/forc-client/src/cmd/call.rs
@@ -184,7 +184,7 @@ pub struct Command {
     pub contract_id: ContractId,
 
     /// Path or URI to a JSON ABI file.
-    #[clap(long, value_parser = parse_abi_path)]
+    #[clap(long, value_parser = parse_abi_path, default_value = "registry")]
     pub abi: Either<PathBuf, Url>,
 
     /// The function selector to call.
@@ -226,6 +226,9 @@ pub struct Command {
 }
 
 fn parse_abi_path(s: &str) -> Result<Either<PathBuf, Url>, String> {
+    if s == "registry" {
+        return Ok(Either::Right(Url::parse("registry://default").unwrap()));
+    }
     if let Ok(url) = Url::parse(s) {
         match url.scheme() {
             "http" | "https" | "ipfs" => Ok(Either::Right(url)),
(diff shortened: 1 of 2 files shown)


forc-plugins/forc-client/src/cmd/call.rs
```diff
@@ -226,6 +226,9 @@ pub struct Command {
 }
 
 fn parse_abi_path(s: &str) -> Result<Either<PathBuf, Url>, String> {
```
zees-dev, line 228: This function may need a re-design; i.e. we most likely will need to read the `contract_id` val

forc-plugins/forc-client/src/op/call/mod.rs
```diff
@@ -67,8 +67,12 @@ pub async fn call(cmd: cmd::Call) -> anyhow::Result<String> {
         // ↳ gh issue: https://github.com/FuelLabs/sway/issues/6893
         Either::Left(path) => std::fs::read_to_string(&path)?,
         Either::Right(url) => {
-            let response = reqwest::get(url).await?.bytes().await?;
-            String::from_utf8(response.to_vec())?
+            if url.as_str() == "registry://default" {
```
zees-dev, line 70: if the first comment is addressed; we should have the full URL at this point.

zees-dev:
Thanks for taking a quick look at this; unfortunately we don't have a canonical registry URL at this point - hence cannot merge this PR - even if the approach is sound.
I reckon we need to contract the full URL in the `forc-plugins/forc-client/src/cmd/call.rs` file.
^ Additionally, we should only default if `mainnet` is selected node target; since verified contract IDs and ABIs will most likely only be relevant to mainnet.

IGI-111:
Closing since this can't be merged for now.

PR #6908, opened 2025-02-09
fix clippy warnings

rex4539:


jjcnn (dismissed):

rex4539:
Fixed one more warning.

Now there is just one left but this one is tricky.

sway-core/src/semantic_analysis/ast_node/declaration/impl_trait.rs
```diff
@@ -707,12 +707,7 @@ impl TyImplSelfOrTrait {
             );
 
             // Do a topological sort to compute an ordered list of nodes (by function calls).
-            let sorted = match petgraph::algo::toposort(&sub_graph, None) {
-                Ok(value) => Some(value),
-                // If the dependency graph contains cycles, then this means there are recursive
```
xunilrj, line 712: This comment was lost.

sdankel:
I'm curious why these warnings weren't caught in CI since we run with `cargo clippy --all-features --all-targets -- -D warnings`. These rules should be enforced in CI otherwise more will be introduced.

rex4539:
Coming from the future :)
Running nightly here.

kayagokalp:
I think we should be doing this once we go the version that includes this clippy lints in sway repo. (It is also possible that some of them will not land etc)

JoshuaBatty:
agree with Kaya here. Closing for now.

Issue #6351, opened 2024-08-01
Implement IR v2.0


ironcev:
## Motivation

Our current IR implementation has several hindering issues, most of them coming from how we deal with pointers and aggregates in the IR. The purpose of this tracking issue is to explain the current problems and their impact, list the agreed solution steps, and track their implementation.

## Terminology

To disambiguate the term pointer, in the remaining part of this description we will consistently use the following terminology:
- `Pointer` for the IR `TypeContent::Pointer`
- raw pointer for the Sway `raw_ptr` type
- reference for the Sway `&T` type
- pointer as a synonym for both raw pointers and references.

## Current Issues

- `Pointer` is conceptually seen as an equivalent for an aggregate type. It is assumed that whenever we encounter a `Pointer` in IR, it represents an aggregate. This assumption makes it impossible to formulate references and raw pointers as `Pointer`s. (They are currently modeled as `u64` which leads to various other issues explained below.) It also represents a semantic issue, because `Pointer`s are not treated as values. E.g., in the `compile_expression_to_value` _every_ pointer will be turned into a `load`, assuming that the `Pointer` itself cannot be a value, and that the value must be behind the `Pointer`.
- `ptr_to_int` does not check if the input argument is a `Pointer`. _Anything_ can be turned into `u64` and, later on, turned back into a `Pointer` by using `int_to_ptr`. This is a _severe_ issue that led to a serious misuse, in which we pair `ptr_to_int` + `int_to_ptr` to obtain `Pointer`s to SSA registers! Having "`Pointers`" to SSA registers is semantically wrong and makes the generated IR being semantically wrong for the most of the optimization pipeline, until it reaches demotion passes, which promote those SSA registers into memory making the previously obtained `Pointer`s valid. Having semantically invalid IR all the way down to the final demotion steps is an issue in itself, but it also harms optimizations that usually bail out when encountering `ptr_to_int` instructions.
[...]

PR #6821, opened 2025-01-08
Implement `Iterator` for `StorageVec`

ironcev:
## Description

This PR implements `Iterator` for `StorageVec`.

Iterating over a `StorageVec` via `for` loop should now be the preferred option. Compared to the traversal via `while` loop and incrementing index, the `for` loop is:
- more performant. `for` loop eliminates the redundant boundary checks done in every `get()` call in the `while` loop.
- more idiomatic and convenient.

Closes #6796.

## Checklist

- [x] I have linked to any relevant issues.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation where relevant (API docs, the reference, and the Sway book).
   - [ ] If my change requires substantial documentation changes, I have [requested support from the DevRel team](https://github.com/FuelLabs/devrel-requests/issues/new/choose)
- [x] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added (or requested a maintainer to add) the necessary `Breaking*` or `New Feature` labels where relevant.
- [x] I have done my best to ensure that my PR adheres to [the Fuel Labs Code Review Standards](https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/external-contributors.md).
- [x] I have requested a review from the relevant team or maintainers.

PR #6928, opened 2025-02-14
Introduce `GlobalVar` akin to `LocalVar` and introduce `get_global` instruction similar to `get_local`.

vaivaswatha:
2nd Step as part of #6351, in continuation to #6896


sway-core/src/ir_generation/const_eval.rs
```diff
@@ -118,10 +118,15 @@ pub(crate) fn compile_const_decl(
...
-            .get_global_constant(env.context, &call_path.as_vec_string()),
+            .get_global_variable(env.context, &call_path.as_vec_string()),
         env.module_ns,
     ) {
-        (Some(const_val), _) => Ok(Some(const_val)),
+        (Some(global_var), _) => {
+            if let Some(constant) = global_var.get_initializer(env.context) {
+                return Ok(Some(Value::new_constant(env.context, *constant)));
+            }
+            Ok(None)
```
ironcev, line 128: Can this actually happen? Shouldn't it actually be an ICE? `GlobalVar`s are modeled as variables, means they do not necessarily need to have an initializer, that is fine. My question is the exact semantics of getting an uninitialized `GlobalVar` in the context of `const_eval`. We are compiling a **const** decl and we are getting a `global_var` from the `env.module`. It means it exists, but it doesn't have an initializer.

Doesn't this points to a bug somewhere before in the compilation chain and should be an ICE? Or am I missing some details?

sway-ir/src/parser.rs
```diff
@@ -43,11 +43,12 @@ mod ir_builder {
                 }
 
             rule script_or_predicate() -> IrAstModule
-                = kind:module_kind() "{" _ configs:init_config()* _ fn_decls:fn_decl()* "}" _
+                = kind:module_kind() "{" _ configs:init_config()* _ global_consts:global_const()* _ fn_decls:fn_decl()* "}" _
```
ironcev, line 46: If we start mixing the terms "global variables/vars" and "global constants/const" it `will become very confusing :smile: 

The proposal is to stay consistent and use the term "global variable" everywhere including here in the parse. Thus, to rename:
- `global_consts` -> `global_vars`
- `global_const` -> `global_var`
- `IrAstGlobalConst` -> `IrAstGlobalVar`
- `build_global_consts_map` -> `build_global_vars_map`

sway-ir/src/parser.rs
```diff
@@ -982,6 +1010,7 @@ mod ir_builder {
         let mut builder = IrBuilder {
             module,
             configs_map: build_configs_map(&mut ctx, &module, ir_ast_mod.configs, &md_map),
+            global_map: build_global_consts_map(&mut ctx, &module, ir_ast_mod.global_consts),
```
ironcev, line 1013: Same here.
```suggestion
            globals_map: build_global_vars_map(&mut ctx, &module, ir_ast_mod.global_vars),
```

sway-ir/src/parser.rs
```diff
@@ -998,6 +1027,7 @@ mod ir_builder {
     struct IrBuilder {
         module: Module,
         configs_map: BTreeMap<String, String>,
+        global_map: BTreeMap<Vec<String>, GlobalVar>,
```
ironcev, line 1030:
```suggestion
        globals_map: BTreeMap<Vec<String>, GlobalVar>,
```

ironcev (changes requested):

vaivaswatha:
@ironcev I've addressed your comments.

sway-ir/src/variable.rs
```diff
@@ -67,3 +67,64 @@ impl LocalVar {
...
+#[doc(hidden)]
+#[derive(Clone, DebugWithContext)]
+pub struct GlobalVarContent {
+    pub ptr_ty: Type,
+    pub initializer: Option<Constant>,
+    pub mutable: bool,
+}
+
+impl GlobalVar {
+    /// Return a new Global of a specific type with an optional [`Constant`] initializer.  If a
```
ironcev, line 85: Just for the consistency with other comments, to use either local or [GlobalVar].
```suggestion
    /// Return a new global of a specific type with an optional [`Constant`] initializer.  If a
```

sway-ir/src/variable.rs
```diff
@@ -67,3 +67,64 @@ impl LocalVar {
...
+    /// Global is marked as mutable then it is guaranteed to be on the stack rather than in
```
ironcev, line 86:
```suggestion
    /// global is marked as mutable then it is guaranteed to be on the stack rather than in
```

test/src/e2e_vm_tests/test_programs/should_pass/language/references/addr_of_consts/src/main.sw
```diff
@@ -0,0 +1,40 @@
+script;
```
ironcev, line 1: Considering the importance of this change, the proposal is to strengthen the tests in several directions. I'll explain in detail what I see it important, but of course, I leave the final decision about the effort put into testing this feature to your best judgment.

The proposal is also to rename the test to `referencing_global_constants` to follow the already established pattern of naming the `language/references` tests.

## Strengthen the test
This test was surely failing before because we were returning a new address every time. But it is still weak to prove the global constants have unique and stable addresses no matter from where taken. E.g., if we introduce the simplest mutation possible, where taking an address of _any_ global constant or its part always returns the same address, this test will still pass.

## Extend the test across different call points accessing the same global constant
The initial issue was that the address taken from a global constant was different when taken in different functions. To prove that's not the case and that the address is exactly the same everywhere the proposal is to have a constant in a library module and test taking its address in two functions in its module and two functions in an other module. They all must be the same.

## Extend the test to include associated constants
We still have issues there with uniquely identifying associated constants, but up to that, we should also prove that the feature work when taking addresses of associated constants.

## Extend the test to cover constants of different types
All of the `language/reference` test put effort in testing referencing different types. This payed out because a lot of existing bug were revealed. We often had issues with `b256`, `u256` and especially zero sized types. Also when inlining was involved. Note that this is not a case with pointers related features only. E.g., when implementing `Iterator` for `StorageVec` there were [issues with certain types](https://github.com/FuelLabs/sway/pull/6821/files#diff-21d9ff5bb0b0fab9194b938bc67a0fa02e8d8231b78a1239caab09d4b68341f0R126). That's the reason I see extending test over different types as beneficial, especially on the long term. E.g., at least to test zero-sized types.

`referencing_local_vars_and_values` is an example of such a test that, but of course it is not possible to write a similar generic test for a constant.

ironcev (changes requested):
Just one nitpick and a proposal to consider strengthening and extending the test coverage.

test/src/e2e_vm_tests/test_programs/should_pass/language/references/addr_of_consts/src/main.sw
vaivaswatha: I'll start extending the tests as you describe here, let me see how far I can take it. Thanks Igor. This is a useful input.

sway-ir/src/optimize/mem2reg.rs
```diff
@@ -148,19 +151,63 @@ pub fn compute_livein(
...
+                op: InstOp::GetGlobal(global_var),
+                ..
+            }) = context.values[ptr.0].value
+            {
+                if !global_var.is_mutable(context)
+                    && is_promotable_type(context, global_var.get_inner_type(context))
+                {
+                    let constant = *global_var
+                        .get_initializer(context)
+                        .expect("We're dealing with an uninitialized value");
```
ironcev, line 173: The proposal is to document here why the `GlobalVar` must be initialized here, because it doesn't need to be.
```suggestion
                        .expect("`global_var` is not mutable so it must be initialized");
```

sway-ir/src/optimize/mem2reg.rs
```diff
@@ -148,19 +151,63 @@ pub fn compute_livein(
...
 /// without offsets, and the result of such a `get_local` is used only in a load
 /// or a store.
 pub fn promote_to_registers(
     context: &mut Context,
     analyses: &AnalysisResults,
     function: Function,
 ) -> Result<bool, IrError> {
+    let promoted_globals = promote_globals(context, &function)?;
+
     let safe_locals = filter_usable_locals(context, &function);
```
ironcev, line 206: Similar to `mem_cpy_opt`, this will be more readable if we extract promoting of locals to its own function:
```Rust
let mut modified = false;
modified |= promote_globals(context, &function)?;
modified |= promote_locals(context, &function)?;

Ok(modified)
```

sway-ir/src/optimize/mem2reg.rs
```diff
@@ -148,19 +151,63 @@ pub fn compute_livein(
     result
 }
 
-/// Promote local values that are accessed via load/store to SSA registers.
-/// We promote only locals of non-copy type, whose every use is in a `get_local`
+// Promote loads of globals constants to SSA registers
+fn promote_globals(context: &mut Context, function: &Function) -> Result<bool, IrError> {
```
ironcev, line 155: Can we have an IR test for this?

sway-ir/src/optimize/dce.rs
```diff
@@ -35,7 +35,7 @@ pub fn create_fn_dce_pass() -> Pass {
         name: FN_DCE_NAME,
         descr: "Dead function elimination",
         deps: vec![],
-        runner: ScopedPass::ModulePass(PassMutability::Transform(fn_dce)),
+        runner: ScopedPass::ModulePass(PassMutability::Transform(globals_dce)),
```
ironcev, line 38: Can we have an IR test covering the elimination of globals?

sway-ir/src/variable.rs
```diff
@@ -67,3 +67,64 @@ impl LocalVar {
...
+/// [`Context`].
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, DebugWithContext)]
+pub struct GlobalVar(#[in_context(global_vars)] pub slotmap::DefaultKey);
+
+#[doc(hidden)]
+#[derive(Clone, DebugWithContext)]
+pub struct GlobalVarContent {
+    pub ptr_ty: Type,
+    pub initializer: Option<Constant>,
+    pub mutable: bool,
```
ironcev, line 81: `mutable` is currently nowhere used. It's just a marker. Considering that we've decided to support `GlobalVar`s in the IR, and not `GlobalConst`s we should also check the semantics around mutability.

In the verifier:
- a non-mutable `GlobalVar` must be initialized.
- only mutable `GlobalVar` can be stored to.

In the printer and parser we also need support for mutability and optional initilizer, but this I've already commented in respected files.

Sadly, also for the mutability of `LocalVar`s we don't have any checks, it's just a marker.

Maybe to have a separate PR to enforce mutability semantics for both `Local` and `Global` vars? But for the printer and parser at least I see it as a part of this PR.

sway-core/src/asm_generation/fuel/fuel_asm_builder.rs
```diff
@@ -1249,6 +1250,40 @@ impl<'ir, 'eng> FuelAsmBuilder<'ir, 'eng> {
         Ok(())
     }
 
+    fn compile_get_global(
```
ironcev, line 1253: `compile_get_global` does not cover the full IR semantics of `GlobalVar`s that do not need to be constants. As commented in the IR code, if a `GlobalVar` is mutable we do not put it into the data section.

sway-core/src/asm_generation/fuel/fuel_asm_builder.rs
```diff
@@ -1249,6 +1250,40 @@ impl<'ir, 'eng> FuelAsmBuilder<'ir, 'eng> {
...
+        instr_val: &Value,
+        local_var: &GlobalVar,
+    ) -> Result<(), CompileError> {
+        let span = self
+            .md_mgr
+            .val_to_span(self.context, *instr_val)
+            .unwrap_or_else(Span::dummy);
+        let Some(constant) = local_var.get_initializer(self.context) else {
+            return Err(CompileError::Internal(
+                "Global variable must have an initializer.",
```
ironcev, line 1264: As also tackled in other comments, if we model the `GlobalVar` with the same semantics as the local vars which we aim for, then this doesn't need to be true. We can have `mut` global vars that are not necessarily initialized. And as mentioned above, those should not be in the data section.

sway-ir/src/optimize/mem2reg.rs
```diff
@@ -148,19 +151,63 @@ pub fn compute_livein(
     result
 }
 
-/// Promote local values that are accessed via load/store to SSA registers.
-/// We promote only locals of non-copy type, whose every use is in a `get_local`
```
ironcev, line 152: Not related to this PR, but shouldn't this comment actually be other way around, we promote only _copy types_. The `is_copy_type`s functions see copy types as those that fit into a register. As the existing comments around the `TypeInfo::is_copy_type` implies, I am not happy neither with the choice of the name nor how those functions are now used. But sticking to that terminology, the promotable types are actually "copy types" and not "non-copy types". BDW, the `memcpyopt::is_copy_type` is actually identical to `is_promotable_type`.

ironcev (changes requested):
In the IR we opted for `GlobalVar`s instead of more restrictive `GlobalConst`s. I support this decision, because it decouples IR strictly from Sway and allows languages with global variables.

The large part of this review comments revolves around the points where we still do not implement `GlobalVar`s as variables but expect them to be constants.

In the first review round I've pointed that `const_eval` must expect initialized global var, otherwise we have an ICE somewhere before in the compilation. This is true, but for Sway and `const_eval`. If we consider not only IR produced by Sway compilation and focus only on the IR, `GlobalVar`s are vars and not necessarily consts. They do not need to be initialized, (the `initializer` is and `Option` with a reason), they can be written to, etc.

We need to support that semantics everywhere which is currently not the case:
- the IR `printer` and `parser` don't have support for `mut`.
- `fuel_asm_builder` expects `GlobalVar`s to be constants that binds it to Sway semantics and not to the IR semantics.
- IR verification does not check `store`ing to non-mutable vars in general.

sway-core/src/asm_generation/fuel/fuel_asm_builder.rs
vaivaswatha: Makes sense. What I'll do here to check if a `GlobalVar` is mutable, and if so, leave that as a `todo!()`. The remaining code will then be correct as it's only dealing with global consts.

sway-ir/src/variable.rs
vaivaswatha: > a non-mutable GlobalVar must be initialized.

I'll add a check for this.

> only mutable GlobalVar can be stored to.

This isn't easy without an analysis (we will need to use something like `get_sym` to get to the symbol being stored to in a `Store`. I'll skip this part for later.

vaivaswatha: Ok, turns out, when we create locals, at many places, we do this: `new_local_var(context, temp_arg_name, value_type, None, false)`, where we aren't providing an initializer, but marking it immutable. So my verification check added for locals fails. I'll remove it for now, to fix the locals issue separately.

vaivaswatha:
@ironcev except for some annoying typos (which I don't know what to do about), CI seems to work. Can you please re-review?

ironcev:
@vaivaswatha Regarding the typos, adding `uage` to `.typos.toml` should fix that part. And the `event.typ` in the LSP definitely looks like a typo it should be `event.type`, or? I assume fixing that identifier in LSP code in this PR is fine.

sway-ir/src/parser.rs
```diff
@@ -59,16 +60,27 @@ mod ir_builder {
...
                         kind: crate::module::Kind::Contract,
                         configs,
+                        global_vars,
                         fn_decls,
                         metadata
                     }
                 }
 
+            rule global_var() -> IrAstGlobalVar
+                = m:("mut" _)? _ "global" _ name:path() _ ":" _ ty:ast_ty() _ "=" _ init:(op_const())? {
```
ironcev, line 75: For locals we have `local mut` and for globals now `mut global`. We should unify this in the same pattern. Personally, I am more for the `mut <local/global>` as done here, but more important is to have a consistent pattern.

ironcev, line 75: Even if we don't have an initializer, the rule still makes the `=` being mandatory which is odd. I am here also for having it the same as with local variables. A single optional init rule that contains also the `=`.

sway-ir/src/optimize/mem2reg.rs
```diff
@@ -148,19 +151,72 @@ pub fn compute_livein(
...
-/// without offsets, and the result of such a `get_local` is used only in a load
-/// or a store.
+/// Promote loads of globals constants to SSA registers
+/// We promote only non-mutable globals of non-copy types
```
ironcev, line 155:
```suggestion
/// We promote only non-mutable globals of copy types
```

sway-ir/src/optimize/mem2reg.rs
```diff
@@ -148,19 +151,72 @@ pub fn compute_livein(
...
     analyses: &AnalysisResults,
     function: Function,
+) -> Result<bool, IrError> {
+    let mut modified = false;
+    modified |= promote_globals(context, &function)?;
+    modified |= promote_locals(context, analyses, function)?;
+    Ok(modified)
+}
+
+/// Promote locals to registers. We promote only locals of copy type,
```
ironcev, line 207:
```suggestion
/// Promote locals to registers. We promote only locals of copy types,
```

ironcev (changes requested):

ironcev (approved):

ironcev (approved):

vaivaswatha:
Need another approval (review) on this. @IGI-111 @FuelLabs/sway-compiler .

IGI-111 (approved):

PR #6941, opened 2025-02-19
Implements deref in reassignment for projections.

### LINKED ISSUE #6397: Dereferencing Arrays Is Not Possible
In Sway, developers can create a mutable reference to a variable using something like the following
snippet:
```sway
let mut x = 1;
let y = &mut x;
*y = 1
```
As expected this updates the value of x. However, a mutable reference is not really usable in case of an
array due to limitations in the parsing. In particular, the parser prohibts the use of parentheses on the
left-hand-side of an assignment. The following snippet does not compile:
```sway
let mut a = [1, 2];
let mut p = &mut a;
(*p)[1] = 1;
```

esdrubal:
## Description

Implements dereferencing as LHS in reassignments for projections.

With this PR we will support:
 - `(*array)[0] = 1`
 - `(*tuple).0 = 1`
 - `(*struct).a = 1`

This also includes support for nested references.

Fixes #6397
There is some progress in #5063

## Checklist

### DIFF
--- a/sway-core/src/semantic_analysis/namespace/lexical_scope.rs
+++ b/sway-core/src/semantic_analysis/namespace/lexical_scope.rs
@@ -43,6 +43,7 @@ pub(super) type SymbolUniqueMap = HashMap<IdentUnique, ResolvedDeclaration>;
 
 type SourceIdent = Ident;
 
+pub(super) type PreludeSynonyms = HashMap<Ident, (ModulePathBuf, ResolvedDeclaration)>;
 pub(super) type GlobSynonyms =
     HashMap<Ident, Vec<(ModulePathBuf, ResolvedDeclaration, Visibility)>>;
 pub(super) type ItemSynonyms = HashMap<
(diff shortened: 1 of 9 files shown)


sway-core/src/language/ty/expression/expression.rs
```diff
@@ -633,8 +633,21 @@ impl TyExpression {
             TyExpressionVariant::Break => {}
             TyExpressionVariant::Continue => {}
             TyExpressionVariant::Reassignment(reass) => {
-                if let TyReassignmentTarget::Deref(expr) = &reass.lhs {
-                    expr.check_deprecated(engines, handler, allow_deprecated);
+                if let TyReassignmentTarget::DerefAccess { exp, indices } = &reass.lhs {
+                    exp.check_deprecated(engines, handler, allow_deprecated);
+                    for indice in indices {
+                        match indice {
+                            ProjectionKind::StructField { name: _ } => {}
```
IGI-111, line 640: We should probably consider that a field access for deprecation purposes, right?

sway-core/src/semantic_analysis/ast_node/expression/typed_expression.rs
```diff
@@ -2673,41 +2712,56 @@ impl ty::TyExpression {
...
+                                }
+                            }
+
+                            // `index_span` does not contain the enclosing square brackets.
+                            // Which means, if this array index access is the last one before the
+                            // erroneous expression, the `full_span_for_error` will be missing the
+                            // closing `]`. We can live with this small glitch so far. To fix it,
+                            // we would need to bring the full span of the index all the way from
+                            // the parsing stage. An effort that doesn't pay off at the moment.
+                            // TODO: Include the closing square bracket into the error span.
```
jjcnn, line 2755: Do we have an issue for this TODO? If not, we should create one just to keep track of it. In any case, we should add a link to the issue in the comment, so that we can track it that way.

jjcnn (changes requested):
I'd like to see a should_fail test as well, just to make sure that typechecking fails if reassigning an incorrectly typed value.

Otherwise it looks good - just a small nit regarding an old TODO that was left in the code, and that we ought to have a tracking issue for.

jjcnn (approved):

sdankel (approved):

Issue #6935, opened 2025-02-17
Forc call help for callable functions


zees-dev:
# Description
Since the ABI must be provided for forc call; it would be great if forc call `--help` could print out all the callable functions of the contract.
This would make it much easier for the user to know what functions are available on the callable contract.
^ All this data is available in the ABI; running the help would also need to parse/process the ABI (if provided).

## Additional Improvements
It would be even better if we can provide an example of how to call the function - with each listed function.
For this, we'd need to parse the input params from the ABI - for the respective function; and construct a a string-parsable input for the function (assuming function accepts args).

## Example
Running `forc call --abi <..> <CONTRACT-ID> --help` should parse the ABI; read the functions, and give examples on how to call them; below the current help output.

These should be something which can simply be copy-pasted by the user themselves for execution.
E.g.
```sh
> forc call --abi ./out/debug/counter-contract-abi.json 0xe18de7c7c8c61a1c706dccb3533caa00ba5c11b5230da4428582abf1b6831b4d --help
```

```
# current help output omitted..

Contract Functions:

add(u32,u32)
> forc call --abi ./out/debug/counter-contract-abi.json 0xe18de7c7c8c61a1c706dccb3533caa00ba5c11b5230da4428582abf1b6831b4d add 1 2

sub(u32,u32)
> forc call --abi ./out/debug/counter-contract-abi.json 0xe18de7c7c8c61a1c706dccb3533caa00ba5c11b5230da4428582abf1b6831b4d sub 1 2
```

Doesn't have to be perfect; i can imagine some complexity with complex structs/enums/generics - but something along these lines is what's expected.

zees-dev:
addressed by https://github.com/FuelLabs/sway/pull/7018

PR #6950, opened 2025-02-22
forc call help for callable functions

Dhanraj30:
## Description

This PR enhances the `forc call --help` command to print out all callable functions of the contract from the provided ABI and provide examples of how to call each function. This makes it easier for users to know what functions are available on the callable contract. This update includes:
- Parsing the ABI to retrieve callable functions.
- Printing the list of functions available in the ABI.
- Providing examples of how to call each function, including the required parameters.
- Fixes issue #6935.

### DIFF
--- a/forc-plugins/forc-client/src/bin/call.rs
+++ b/forc-plugins/forc-client/src/bin/call.rs
@@ -5,8 +5,16 @@ use forc_tracing::{init_tracing_subscriber, println_error};
 async fn main() {
     init_tracing_subscriber(Default::default());
     let command = forc_client::cmd::Call::parse();
-    if let Err(err) = forc_client::op::call(command).await {
-        println_error(&format!("{}", err));
-        std::process::exit(1);
+    // Check if the print_functions flag is set
+    if command.print_functions {
+        if let Err(err) = forc_client::op::print_callable_functions(command).await {
+            println_error(&format!("{}", err));
+            std::process::exit(1);
+        }
+    } else {
+        if let Err(err) = forc_client::op::call(command).await {
+            println_error(&format!("{}", err));
+            std::process::exit(1);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


zees-dev:
Thanks for this; however I was envisioning a different approach here.
For example; running `forc call --help` should print out the current help.
Running `forc call --abi <..> <CONTRACT-ID> --help` should parse the ABI; read the functions, and give examples on how to call them; below the current help output.
These should be something which can simply be copy-pasted by the user themselves for execution.
E.g.
```
> forc call --abi ./out/debug/counter-contract-abi.json 0xe18de7c7c8c61a1c706dccb3533caa00ba5c11b5230da4428582abf1b6831b4d --help
```

```
# current help output omitted..

Contract Functions:

add(u32,u32)
> forc call --abi ./out/debug/counter-contract-abi.json 0xe18de7c7c8c61a1c706dccb3533caa00ba5c11b5230da4428582abf1b6831b4d add 1 2

sub(u32,u32)
> forc call --abi ./out/debug/counter-contract-abi.json 0xe18de7c7c8c61a1c706dccb3533caa00ba5c11b5230da4428582abf1b6831b4d sub 1 2
```

Doesn't have to be perfect; i can imagine some complexity with complex structs/enums/generics - but something along these lines is what's expected.

zees-dev:
> I like the idea, but `--help` is the wrong flag to use, it's much too standardized to just print usage for people to notice you can use it for more than that.
> 
> This should probably be some specific flag to list contract methods.
> 
> In fact maybe this is the sort of output you should get if you run forc call with an abi and a contract but no method. And have the tool suggest what you could call on this contract.

I have seen the help flag used in other parts of the tooling which displays examples on how the tool can be called.

However no hard opinions here and I can see the other side of the argument (non-standardized output).

zees-dev:
@Dhanraj30 - Im encountering compile-time errors running code in this branch.

zees-dev:
This will be resolved by https://github.com/FuelLabs/sway/pull/6950

PR #6953, opened 2025-02-24
Implement Iteration for Bytes

SwayStar123:
## Description
Adds an iterator to allow for loops over bytes

## Checklist


sway-lib-std/src/bytes.sw
```diff
@@ -871,6 +872,86 @@ impl Bytes {
 
         self.len = new_len;
     }
+
+    /// Returns an [Iterator] to iterate over this `Bytes`.
+    ///
+    /// # Returns
+    ///
+    /// * [BytesIter<V>] - The struct which can be iterated over.
```
bitzoic, line 880:
```suggestion
    /// * [BytesIter] - The struct which can be iterated over.
```
Shouldn't this just be `BytesIter`? Bytes are not generic

sway-lib-std/src/bytes.sw
```diff
@@ -871,6 +872,86 @@ impl Bytes {
...
+    ///
+    /// # Examples
+    ///
+    /// ```sway
+    /// fn foo() {
+    ///     let mut bytes = Bytes::new();
+    ///     bytes.push(5);
```
bitzoic, line 887:
```suggestion
    ///     bytes.push(5u8);
```
All of these should be `u8`

sway-lib-std/src/bytes.sw
```diff
@@ -871,6 +872,86 @@ impl Bytes {
...
+    ///     bytes.push(10);
+    ///     bytes.push(15);
+    ///
+    ///     // Get the iterator
+    ///     let iter = bytes.iter();
+    ///
+    ///     assert_eq(5, iter.next().unwrap());
```
bitzoic, line 894:
```suggestion
    ///     assert_eq(5u8, iter.next().unwrap());
```

sway-lib-std/src/bytes.sw
```diff
@@ -871,6 +872,86 @@ impl Bytes {
...
+    ///     assert_eq(10, iter.next().unwrap());
+    ///     assert_eq(15, iter.next().unwrap());
+    ///
+    ///     for elem in bytes.iter() {
+    ///         log(elem);
+    ///     }
+    /// }
+    ///
+    /// # Undefined Behavior
```
bitzoic, line 903:
```suggestion
    /// # Undefined Behavior
```
This should be addressed, maybe with a clone

sway-lib-std/src/bytes.sw
```diff
@@ -871,6 +872,86 @@ impl Bytes {
...
+    ///     }
+    /// }
+    /// ```
+    pub fn iter(self) -> BytesIter {
+        // WARNING: Be aware of caveats of this implementation
+        //          if you take it as an example for implementing
+        //          `Iterator` for other types.
+        //
+        //          Due to the Sway's copy semantics, the `values` will
+        //          actually contain **a copy of the original vector
```
bitzoic, line 926: Sentences shouldn't start on a newline so forc doc can handle it correctly

sway-lib-std/src/bytes.sw
```diff
@@ -1091,3 +1172,33 @@ impl AbiDecode for Bytes {
...
+        //         But since modifying the vector during iteration is
+        //         considered undefined behavior, this implementation,
+        //         that always checks against the length at the time
+        //         the iterator got created is perfectly valid.
+        if self.index >= self.values.len() {
+            return None
+        }
+
+        self.index += 1;
+        self.values.get(self.index - 1)
```
bitzoic, line 1202:
```suggestion
        self.values.get(self.index)
        self.index += 1;
```

sway-lib-std/src/bytes.sw
SwayStar123: These are internal comments not docs. I took these from the vector implementation

sway-lib-std/src/bytes.sw
ironcev: What exactly do you mean by a clone? Like cloning the `bytes` before iteration and iterating over the clone? This has a serious performance impact and still, iteration over the clone will again be undefined if the clone gets modified during the iteration. This undefined behavior is, unfortunately, inherent to iteration.

Or maybe I got a completely wrong idea of what you've meant by clone? :smile:

ironcev (approved):

sway-lib-std/src/bytes.sw
bitzoic: > This has a serious performance impact
Yes, but I think we should prioritize security over performance.

>  if the clone gets modified during the iteration
We should be able to set it up such that the clone never gets modified. The underlying bytes of `BytesIter` are private and we can prevent any modifications.

ironcev: The security considerations are fully understandable, but at this point I share @IGI-111's view. Different languages tackle it differently. In C++ it is an undefined behavior, in C# and Java a runtime exception, ... In Rust it's a compile time error, and that not because iterators are treated in any special way, but just because of the borrowing rules.

The issue with any approach that would go C#/Java way is the one @IGI-111 mentions. I'am also 100% sure that would happen - devs would stop using `for` and switch back to `while`.

As with other issues coming from the fact that Sway doesn't have move and borrowing semantics, I am for "solving" this with semantic analysis like, e.g., [sway-analyzer](https://github.com/ourovoros-io/sway-analyzer). It will never be 100% bullet-proof but we could cover realistic invalid usage that might incidentally happen.

bitzoic (approved):

PR #6955, opened 2025-02-24
chore: avoid duplicate crates

rex4539:


zees-dev:
A lot of these are dev-dependencies from member crates - to be used in tests.
Im thinking whether it would be worthwhile having them in a dedicated `dev-dependencies` section within the root cargo.toml.
This way we can know at glance which crates are actually used in the project/binaries and which ones are only used for testing (hence more easy to upgrade/replace).

sdankel:
Taking it over here: https://github.com/FuelLabs/sway/pull/6966

PR #6947, opened 2025-02-21
ECAL opcode support

zees-dev:
## Description
The ECAL opcode is supported in `fuel-vm` to extend the capabilities of the VM.
Users can implement rust code to interop with the VM using the ECAL opcode.
This PR introduces support for the opcode in the sway repo - such that sway assembly can be successfully compiled with the ECAL opcode.

Examples of how to utilize the `ECAL` opcode can be seen here: https://github.com/FuelLabs/fuel-vm/blob/9478991db1ea8799fab47ebe1c195946a2698a7f/fuel-vm/examples/external.rs#L51

- Closes https://github.com/FuelLabs/sway/issues/5239
- Added an e2e test to validate compilation of ecall opcode in sway assembly.

## Checklist

- [x] I have linked to any relevant issues.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation where relevant (API docs, the reference, and the Sway book).
   - [ ] If my change requires substantial documentation changes, I have [requested support from the DevRel team](https://github.com/FuelLabs/devrel-requests/issues/new/choose)
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added (or requested a maintainer to add) the necessary `Breaking*` or `New Feature` labels where relevant.
- [ ] I have done my best to ensure that my PR adheres to [the Fuel Labs Code Review Standards](https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/external-contributors.md).
- [ ] I have requested a review from the relevant team or maintainers.

PR #6957, opened 2025-02-24
Add `bldd` case to `Op::parse_opcode()`

ironcev:
## Description

The `bldd` opcode was not covered in the `Op::parse_opcode()` which resulted in `Unknown opcode: "bldd"` error.

The remaining part of the `bldd` opcode integration chain is already implemented in #6254.

## Checklist


zees-dev (commented):
LGTM 👍 

However you please add a simple test case to validate compilation of sway code which uses this assembly instruction.
E.g. maybe something like [this](https://github.com/FuelLabs/sway/blob/fccc8a928f81bdefca4ef36edb764eb84bf685c0/test/src/e2e_vm_tests/test_programs/should_pass/ecall/ecall_basic/src/main.sw#L3) (introduced in recent [PR](https://github.com/FuelLabs/sway/pull/6947) for another op-code).

zees-dev (approved):
Not sure why the ecal tests need to be updated; I'm assuming I probably did them incorrectly on the first pass; thanks for fixing.

LGTM 👍

ironcev:
> Not sure why the ecal tests need to be updated;

@zees-dev No worries :smile: And thanks for asking for tests to be written! To your question, we have [guidelines for writing E2E tests](https://github.com/FuelLabs/sway/tree/master/test/src/e2e_vm_tests#writing-end-to-end-vm-tests) but no lints that would warn developers when not following them. Essentially, if a test could be a `library` and can avoid `core` dependency it executes faster as when `core` needs to be compiled. This is important, considering the amount of tests we have. And also in this case, the test's category was `unit_tests_pass` although there were no unit tests in it. We should have lints for such cases, it is easy to overlook it when writing test.

PR #6933, opened 2025-02-14
Vec implementation using slices

xunilrj:
## Description

This PR is a POC of using slices to implement `Vec<T>`.

## Checklist

- [ ] I have linked to any relevant issues.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation where relevant (API docs, the reference, and the Sway book).
   - [ ] If my change requires substantial documentation changes, I have [requested support from the DevRel team](https://github.com/FuelLabs/devrel-requests/issues/new/choose)
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added (or requested a maintainer to add) the necessary `Breaking*` or `New Feature` labels where relevant.
- [ ] I have done my best to ensure that my PR adheres to [the Fuel Labs Code Review Standards](https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/external-contributors.md).
- [ ] I have requested a review from the relevant team or maintainers.

Issue #5110, opened 2023-09-12
Improving string slice


xunilrj:
The PR https://github.com/FuelLabs/sway/pull/4996 introduced string slices, but they are still very limited, and the ergonomics of needing `__to_str_array` to convert from literals to string arrays is horrible.

This issue will track everything that is necessary to improve the string slices:

- [ ] allow string slices in constants;
- [ ] allow string slices in storage; 
- [ ] allow string slices in configurables;
- [ ] allow string slices as main arguments;
- [ ] allow string slices as main return;
- [ ] allow slicing of string slices
- [ ] better interop with `String`

And the last step would be to remove string arrays entirely from the language. Which means 
- [ ] removing all string arrays intrinsics

PR #6965, opened 2025-02-26
fix extract_type_parameters regarding primitive types

xunilrj:
## Description

This PR is part of https://github.com/FuelLabs/sway/issues/5110. It fixes a problem that improves the usability of slices.

When a method is returned from the "method resolution algorithm", if any generic is involved, it is returned as its generic form. So for example, if a method `len` is called on `Vec<u8>`, we return something like `fn len(self: Vec<T>)`. 

Later we monomorphize `T` into `u8`. To do that we run a "pattern match" to realize that we need to replace `T` with `u8`.

In this case, the bug was that we were not considering that some primitive types have "implicit type parameters".

This was found whilst creating more tests for https://github.com/FuelLabs/sway/pull/6933, because I was not able to call the `len` method defined for slices as:

```sway
impl<T> &[T] {
    pub fn len(self) -> u64 {
        ...
    }
}
``` 

## Checklist


IGI-111 (dismissed):
Congrats on figuring this one out :detective:

test/src/e2e_vm_tests/test_programs/should_pass/language/method_on_primitives/.gitignore
```diff
@@ -0,0 +1,2 @@
+out
```
ironcev, line 1: This `.gitignore` file is not needed and should be deleted. (We need a lint here...)

test/src/e2e_vm_tests/test_programs/should_pass/language/method_on_primitives/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+license = "Apache-2.0"
+name = "method_on_primitives"
+entry = "main.sw"
+
+[dependencies]
+std = { path = "../../../../../../../sway-lib-std" }
```
ironcev, line 8: We don't need the full blown `std` here. The lightweight reduced `swayl-lib-std-assert` is enough.

ironcev (changes requested):

test/src/e2e_vm_tests/test_programs/should_pass/language/method_on_primitives/Forc.toml
ironcev: Hmmm, actually I don't see the need for any dependencies at all. We are not testing here if the `len()` method is properly implemented, but just if it is properly resolved. Thus, turning the test into `library;` without any dependencies and simply compiling
```sway
let _ = __slice(&[1u8], 0, 1).len();
```
should already prove the fix is valid.

The proposal is to remove the `assert` and all the dependencies.

test/src/e2e_vm_tests/test_programs/should_pass/language/method_on_primitives/.gitignore
xunilrj: Done.

test/src/e2e_vm_tests/test_programs/should_pass/language/method_on_primitives/Forc.toml
xunilrj: Done.

ironcev (approved):

JoshuaBatty (approved):

PR #6967, opened 2025-02-27
IR: Taking function arg address and representing references as pointers

vaivaswatha:
## Description

This is the third PR to address #6351. It allows safely taking addresses of function arguments, and represents references (`&`) as pointers in the IR.

Note: The entire function `type_correction` (in `ir_generation.rs`) and this [edit](https://github.com/FuelLabs/sway/pull/6967/commits/2de7ab9ec19e979a1cc37eafcbf964b1c319ff18#diff-3b0502f41aeb3c45ddbdb3cc590658d73d2bd7283671eb6d789d7e77a7dd2596R670) in `compile.rs`, to handle return values, must be removed once we fix and finalize the syntax of references. Until then these two hacks need to stay. Ideally I wanted to put them both in one place (which is the `type_correction` pass), but the return type fix needs to happen, unfortunately, during IR gen.


sway-core/src/ir_generation.rs
```diff
@@ -216,10 +216,267 @@ pub fn compile_program<'a>(
...
+                        stored_val,
+                    } => {
+                        let dst_ty = dst_val_ptr.get_type(ctx).unwrap();
+                        let stored_ty = stored_val.get_type(ctx).unwrap();
+                        if let Some(dst_pointee_ty) = dst_ty.get_pointee_type(ctx) {
+                            // The destination is a pointer type. We need to see if it's a double pointer.
+                            if let Some(dst_pointee_pointee_ty) =
+                                dst_pointee_ty.get_pointee_type(ctx)
+                            {
+                                // We have a double pointer. If just loading once solves our problem, we do that.
```
IGI-111, line 346: Can we have more depth than 2 to these pointers and could it blow this up?

vaivaswatha: It's possible. This pass only covers what we cover and test today. The general solution is to fix the language syntax for references. Once we do that this pass will go away wholly.

sway-core/src/ir_generation.rs
```diff
@@ -216,10 +216,267 @@ pub fn compile_program<'a>(
...
+                                        expected_ty: base_pointee_ty,
+                                        use_instr: instr,
+                                        use_idx: indices.len(),
+                                    });
+                                }
+                            }
+                        } else {
+                            // The base is not a pointer type. If a pointer to base_ty works for us, do that.
+                            let elem_ptr_ty = *elem_ptr_ty;
+                            let indices = indices.clone();
```
ironcev: Nit: at first sight, this looked like an unnecessary clone. I know this code is a temporary one, but while it's here, having a comment to avoid confusion might be good :smile:
```suggestion
                            let indices = indices.clone(); // Cloning needed because of mutable and immutable borrow of `ctx`.
```

sway-core/src/ir_generation/convert.rs
```diff
@@ -173,7 +182,14 @@ fn convert_resolved_type_info(
                 )?;
                 Type::get_typed_slice(context, elem_ir_type)
             } else {
-                Type::get_uint64(context)
+                let elem_ir_type = convert_resolved_type_id(
```
ironcev: `elem_ir_type` is a confusing name. It is fine for typed slices (above) where it is a type of a slice element, but for plain references is confusing. Proposal is `referenced_ir_type` (although it is also a bit inaccurate, I guess the correct one would be `referenced_type_ir_type` but that's IMO too wordy :smile:).

sway-core/src/ir_generation/function.rs
```diff
@@ -38,38 +42,57 @@ use sway_types::{
...
+
+    fn get_type(&self, context: &Context) -> Option<Type> {
+        self.value().get_type(context)
+    }
+
+    fn get_constant(&self, context: &Context) -> Option<Constant> {
+        self.value().get_constant(context).cloned()
+    }
+
+    fn unwrap_memory(self) -> Value {
```
ironcev: Nit: I would go here with `expect_memory`, mostly to be consistent with, e.g., `ResolvedDeclaration::expected_typed` and similar methods in other existing types, where we have the same pattern of mostly having two enum variants (`InMemory` and `InRegister` in this case), and _expecting_ one of them.

sway-core/src/ir_generation/function.rs
```diff
@@ -38,38 +42,57 @@ use sway_types::{
...
+        match self {
+            CompiledValue::InMemory(value) => value,
+            CompiledValue::InRegister(_) => panic!("Expected InMemory, got InRegister"),
+        }
+    }
+
+    fn unwrap_register(self) -> Value {
```
ironcev: Same here, `expect_register`.

sway-core/src/ir_generation/function.rs
```diff
@@ -112,22 +137,43 @@ fn to_constant(_s: &mut FnCompiler<'_>, context: &mut Context, value: u64) -> Va
     Value::new_constant(context, c)
 }
 
-fn save_to_local_return_ptr(
+/// Store a register value to a new local variable and return CompiledValue::InMemory(_).
```
ironcev: Just to extend the comment for full clarity.
```suggestion
/// Store the in-register `value` to a new local variable and return [CompiledValue::InMemory], or return the same `value` if it is already in memory.
```

sway-core/src/ir_generation/function.rs
```diff
@@ -112,22 +137,43 @@ fn to_constant(_s: &mut FnCompiler<'_>, context: &mut Context, value: u64) -> Va
...
+                    CompileError::InternalOwned(ir_error.to_string(), Span::dummy())
+                })?;
+            let local_var_ptr = s.current_block.append(context).get_local(local_var);
+            let _ = s.current_block.append(context).store(local_var_ptr, val);
+            Ok(CompiledValue::InMemory(local_var_ptr))
+        }
+    }
+}
+
+/// If a value is in memory, load it into a register and return the register value.
```
ironcev: Same here, for completenes.
```suggestion
/// If a `value` is in memory, load it into a register and return the register value.
/// If it is already in a register, return the same `value`.
```

sway-core/src/ir_generation/function.rs
```diff
@@ -193,13 +240,46 @@ impl<'a> FnCompiler<'a> {
...
+            let local_name = self.lexical_map.insert(arg_name.as_str().to_owned());
+            let local_var = self.function.new_unique_local_var(
+                context,
+                local_name.clone(),
+                arg_value.get_type(context).unwrap(),
+                None,
+                false,
+            );
+            if self.ref_mut_args.contains(&arg_name) {
+                self.ref_mut_args.insert(local_name.clone());
```
ironcev: Cloning not needed here.
```suggestion
                self.ref_mut_args.insert(local_name);
```

sway-core/src/ir_generation/function.rs
```diff
@@ -193,13 +240,46 @@ impl<'a> FnCompiler<'a> {
...
+            }
+            let local_val = entry.append(context).get_local(local_var);
+            entry.append(context).store(local_val, arg_value);
+        }
+        match self.compile_code_block(context, md_mgr, ast_block)?.value {
+            // Final value must always be a value, not a pointer.
```
ironcev: IMO a bit confusing comment. The final value can be a pointer (of type IrPtr), but the `CompiledValue` must not be `InMemory`.
```suggestion
            // Final value must always be in a register, not in memory.
```

sway-core/src/ir_generation/function.rs
```diff
@@ -326,30 +409,23 @@ impl<'a> FnCompiler<'a> {
         }
     }
 
-    fn compile_expression_to_value(
+    // Compiled expression, and if the compiled result is in memory, load it to a register.
```
ironcev:
```suggestion
    // Compile expression, and if the compiled result is in memory, load it to a register.
```

sway-core/src/ir_generation/function.rs
```diff
@@ -2276,41 +2440,50 @@ impl<'a> FnCompiler<'a> {
...
+            };
+            let elem_ir_ty = convert_resolved_type_id(
+                te,
+                de,
+                context,
+                elem_ty,
+                &first_argument_expr.span.clone(),
+            )?;
+            let return_type = Type::new_ptr(context, elem_ir_ty);
+            let ptr_to_first_element = self.current_block.append(context).asm_block(
```
ironcev, line 2479: I am thinking about the `__slice` type in IR in general and specifically if we can remove the `asm` block here and replace it with `get_elem_ptr` and `load`. We know `asm`s harm optimizations so I am thinking, if we can remove them, better so.

Essentially, in the `is_slice` case, the `first_argument_value` is the pointer to the `{ ptr T, 64 }` slice struct. The generated `asm` block just reads the address from the first part of the struct and converts it to `ptr T`. Shouldn't this be the same as:

```
first_argument_value = <however we get the pointer to the slice>
v_zero = const u64 0
v_ptr_to_ptr_part = get_elem_ptr first_argument_value, ptr ptr T, v_zero
ptr_to_first_element = load v_ptr_to_ptr_part // This should have type `ptr T`. 
```

The above code would assume we can `get_elem_ptr` from an IR `__slice[T]` (`TypeContent::TypedSlice`) which I think we should be able to.

E.g., in the following example:
```sway
let my_slice: &__slice[u64] = __slice(&[1, 2, 3], 0, 3);
let _: &u64 = __elem_at(my_slice, 0);
```

the resulting IR would be:
```
v22 = get_local ptr __slice[u64], my_slice, !18
v23 = asm(ptr: v22, ptr_out) -> ptr u64 ptr_out {
    lw     ptr_out ptr i0
}
```

Instead of that, we would have:
```
v22 = get_local ptr __slice[u64], my_slice, !18
v23 = const u64 0
v24 = get_elem_ptr v22, ptr ptr 64, v23
v25 = load v24 // This is the `ptr_to_first_element` and will have type `ptr u64`.
```

sway-core/src/ir_generation/function.rs
```diff
@@ -2356,11 +2529,11 @@ impl<'a> FnCompiler<'a> {
...
         let ptr_out_arg = AsmArg {
             name: Ident::new_no_span("ptr_out".into()),
-            initializer: Some(ptr),
+            initializer: None,
         };
 
-        let return_type = Type::get_uint64(context);
-        let ptr = self.current_block.append(context).asm_block(
+        let return_type = Type::new_ptr(context, elem_ir_type);
+        let ptr_out = self.current_block.append(context).asm_block(
```
ironcev, line 2537: Same thinking here. Can we replace the `asm` block with regular operations? Essentially, this resulting `asm` can never be optimized away in the IR, even if the common case of `idx` being a const 0 or 1:

```
asm(idx: v24, elem_ir_type_size: v25, ptr: v23, offset_temp, ptr_out) -> ptr u64 ptr_out {
    mul    offset_temp idx elem_ir_type_size
    add    ptr_out ptr offset_temp
}
```
I didn't check it, but I expect that the `idx` being 0 or 1 is optimized later on in ASM (not sure, though). But still, if optimized in IR, it potentially opens other IR optimizations.

Essentially, what the `asm` code does is similar to what the `__ptr_add` intrinsic does so I would expect the similar `asm`-free code here as well.

sway-core/src/ir_generation/function.rs
```diff
@@ -2615,55 +2801,57 @@ impl<'a> FnCompiler<'a> {
...
 
-        // TODO: (REFERENCES) Do we need to convert to `u64` here? Can we use `Ptr` directly? Investigate.
-        let int_ty = Type::get_uint64(context);
-        let val = self
-            .current_block
-            .append(context)
-            .ptr_to_int(value, int_ty)
-            .add_metadatum(context, span_md_idx);
-        Ok(TerminatorValue::new(val, context))
+        // Taking a reference is just converting the pointer into a "register" value.
```
ironcev, line 2823: Yessss :smile: :muscle:

sway-core/src/ir_generation/function.rs
```diff
@@ -4551,36 +4806,13 @@ impl<'a> FnCompiler<'a> {
                 } => {
                     // Take the optional initialiser, map it to an Option<Result<TerminatorValue>>,
                     // transpose that to Result<Option<TerminatorValue>> and map that to an AsmArg.
-                    //
-                    // Here we need to compile based on the Sway 'copy-type' vs 'ref-type' since
-                    // ASM args aren't explicitly typed, and if we send in a temporary it might
-                    // be mutated and then discarded.  It *must* be a ptr to *the* ref-type value,
-                    // *or* the value of the copy-type value.
                     let init_expr = initializer.as_ref().unwrap();
                     // I'm not sure if a register declaration can diverge, but check just to be safe
```
ironcev: Not part of this PR, but let's remove this comment. Yes, a register initialization can diverge, it can be an arbitrary expression.
```suggestion
```

sway-ir/src/instruction.rs
```diff
@@ -480,6 +484,378 @@ impl InstOp {
         }
     }
 
+    /// Set the operand at the given index to the provided value.
+    /// The indices are in the same order as returned by `get_operands`.
+    pub fn set_operand(&mut self, replacement: Value, idx: usize) {
```
ironcev, line 489: Just a remark, this was quite some work to write this function just for the `type_correction` pass. I hope it will find other usages too...

sway-ir/src/instruction.rs
```diff
@@ -480,6 +484,378 @@ impl InstOp {
...
+        match self {
+            InstOp::AsmBlock(_, args) => {
```
ironcev, line 491: In all other cases we panic on invalid index. Can we make it panic here as well?

sway-ir/src/optimize/mem2reg.rs
```diff
@@ -66,12 +67,19 @@ fn filter_usable_locals(context: &mut Context, function: &Function) -> HashSet<S
...
+            ValueDatum::Instruction(Instruction {
+                op:
+                    InstOp::Store {
+                        dst_val_ptr: _,
+                        stored_val,
+                    },
                 ..
             }) => {
-                // We understand load and store, so no problem.
+                // Make sure that a local ('s address) isn't stored.
```
ironcev: Spend some time pondering about this one until it became obvious :smile: Perhaps to extend the comment to make the case more clear.
```suggestion
                // Make sure that a local's address isn't stored.
                // E.g., in cases like `let r = &some_local;`.
```

sway-ir/src/verify.rs
```diff
@@ -548,11 +548,14 @@ impl InstructionVerifier<'_, '_> {
...
-            | BinaryOpKind::Div
-            | BinaryOpKind::Mod => {
+            BinaryOpKind::Add => {
+                if !(arg1_ty.eq(self.context, &arg2_ty) && arg1_ty.is_uint(self.context)
+                    || arg1_ty.is_ptr(self.context) && arg2_ty.is_uint64(self.context))
+                {
+                    return Err(IrError::VerifyBinaryOpIncorrectArgType);
+                }
+            }
+            BinaryOpKind::Sub | BinaryOpKind::Mul | BinaryOpKind::Div | BinaryOpKind::Mod => {
```
ironcev, line 558: Shouldn't the same logic as above be valid for `Sub`? We can also subtract from pointers. In fact, we have the `__ptr_add` and `__ptr_sub` intrinsics that use the `Add` and `Sub` respectively. There is only one test for `__ptr_sub` and I assume that one is passing because we still treat `raw_ptr` as `u64` so the `Sub` ends up having `u64` and `u64` as arguments. But once we represent `raw_ptr` with `IrType::Ptr` as we do references in this PR, I assume the `__ptr_sub` will fail because the `Sub` as is will not accept `Ptr` as the first parameter.

test/src/e2e_vm_tests/test_programs/should_pass/language/intrinsics/dbg/stdout.snap
```diff
@@ -3,15 +3,8 @@ source: test/src/snapshot/mod.rs
 assertion_line: 162
 ---
 > forc build --path test/src/e2e_vm_tests/test_programs/should_pass/language/intrinsics/dbg --asm final | sub ecal
-ecal $r0 $r2 $r3 $$retv       ; ecal id fd buf count
+ecal $r4 $r0 $r1 $r3          ; ecal id fd buf count
 ecal $r2 $r1 $zero $zero      ; ecal id fd zero zero
```
ironcev, line 7: Hmmmm, this looks pretty strange that only one of the `ecal`s "survived"? What is the explanation for this? I would expect `ecal`s not to be affected by changes in this PR :thinking:

test/src/e2e_vm_tests/test_programs/should_pass/language/intrinsics/dbg_release/stdout.snap
```diff
@@ -2,14 +2,8 @@
 source: test/src/snapshot/mod.rs
 ---
 > forc build --path test/src/e2e_vm_tests/test_programs/should_pass/language/intrinsics/dbg_release --asm final | sub ecal
-ecal $r3 $r0 $r1 $r2          ; ecal id fd buf count
-ecal $r3 $r0 $r1 $r2          ; ecal id fd buf count
-ecal $r3 $r0 $r1 $r2          ; ecal id fd buf count
-ecal $r3 $r0 $r1 $r2          ; ecal id fd buf count
-ecal $r0 $r2 $r3 $$retv       ; ecal id fd buf count
+ecal $r4 $r0 $r1 $r3          ; ecal id fd buf count
```
ironcev, line 5: Same question here.

test/src/e2e_vm_tests/test_programs/should_pass/language/references/referencing_function_parameters/test.toml
```diff
@@ -1,5 +1,5 @@
 # TODO Enable this test once https://github.com/FuelLabs/sway/issues/6898 is fixed.
```
ironcev: We can delete this comment now when the test is enabled and green.

test/src/ir_generation/tests/impl_self_reassignment.sw
```diff
@@ -23,12 +23,12 @@ fn main() {
 // check: call $(f_method=$ID)($a_var)
 
 // check: fn $f_method(self $MD: ptr { u64 }) -> ()
-// nextln: entry(self: ptr { u64 }):
+// check: entry(self: ptr { u64 }):
```
ironcev, line 26: Didn't take a look at the actual output of this test, but I would expect that this first line must be changed from `nextln` to `check`, but afterwards others can stay `nextln`. Is it so? In that case I would prefer leaving `nextln` for more strictness.

test/src/ir_generation/tests/predicate_data_mult_args.sw
```diff
@@ -15,10 +15,10 @@ fn main(x: u64, y: u64) -> bool {
...
 // check: gtf  $(r1=$REG) $r2 i512
 // nextln: jnzf $r1 $$zero $IMM
-// nextln: jmpf $$zero $IMM
-// nextln: movi $(r0=$REG) i2
-// nextln: eq   $r1 $r1 $r0
-// nextln: xori $r1 $r1 i1
-// nextln: jnzf $r1 $$zero $IMM
-// nextln: jmpf $$zero $IMM
-// nextln: ret  $$zero
+// check: jmpf $$zero $IMM
```
ironcev, line 18: Same here.

test/src/ir_generation/tests/predicate_data_single_arg.sw
```diff
@@ -15,11 +15,10 @@ fn main(x: u64) -> bool {
...
 // nextln: jnzf $r1 $$zero $IMM
-// nextln: jmpf $$zero $IMM
-// nextln: movi $(r0=$REG) i2
-// nextln: eq   $r1 $r1 $r0
-// nextln: xori $r1 $r1 i1
-// nextln: jnzf $r1 $$zero $IMM
-// nextln: jmpf $$zero $IMM
-// nextln: ret  $$zero
-
+// check: jmpf $$zero $IMM
```
ironcev, line 18: Same here.

test/src/ir_generation/tests/revert.sw
```diff
@@ -13,11 +13,11 @@ fn main() {
 // check: script {
 
 // check: fn main() -> ()
-// nextln: entry():
-// nextln: $(value=$VAL) = const u64 42
-// nextln: call revert_0($value)
+// check: entry():
```
ironcev, line 16: Same here.

test/src/ir_generation/tests/revert.sw
```diff
@@ -13,11 +13,11 @@ fn main() {
...
+// check: $(value=$VAL) = const u64 42
+// check: call revert_0($value)
 
 // check: fn revert_0($(foo=$ID)
-// nextln: entry($foo: u64):
-// nextln: revert $foo
-// nextln: } 
+// check: entry($foo: u64):
```
ironcev, line 21: Same here.

test/src/ir_generation/tests/trait.sw
```diff
@@ -42,11 +42,17 @@ fn main() -> bool {
 // check: ret bool $res
 
 // check: fn $pred_or(self $MD: { bool }, other $MD: { bool }) -> bool
-// check: $(self_pred=$VAL) = call $ID(self)
+// check: store self to $VAL
```
ironcev, line 45: Same here.

test/src/ir_generation/tests/trait.sw
```diff
@@ -42,11 +42,17 @@ fn main() -> bool {
...
+// check: store other to $VAL
+// check: $(self_ptr=$VAL) = get_local ptr { bool }, self_
+// check: $(self_val=$VAL) = load $self_ptr
+// check: $(self_pred=$VAL) = call $ID($self_val)
 // check: cbr $self_pred, $(block1=$ID)($self_pred), $(block0=$ID)()
 
 // check: $block0():
-// check: $(other_pred=$VAL) = call $ID(other)
+// check: $(other_ptr=$VAL) = get_local ptr { bool }, other_
```
ironcev, line 53: Same here.

ironcev (changes requested):
Great job @vaivaswatha! :muscle: :rocket:

To my comments, those regarding replacing `asm` code generated for slices, can IMO be addressed in a separate PR.

Looking forward to get this PR merged and to continue the work on references!

test/src/e2e_vm_tests/test_programs/should_pass/language/intrinsics/dbg/stdout.snap
vaivaswatha: The `ecal` changes are all because of inlining differences that arose with the PR. With inlining we end up with multiple `ecal`s (previously) and now, with that function not being inlined, we have just the one.

test/src/ir_generation/tests/impl_self_reassignment.sw
vaivaswatha: These changes were necessary because we now have temporaries in memory  which means other instructions (load, IIRC) come in between.

nevertheless these tests are quite flaky and IMO not serving much of a purpose at all :-(.

vaivaswatha:
> Great job @vaivaswatha! 💪 🚀

Thank you @IGI-111 . This was a tough one !

> To my comments, those regarding replacing `asm` code generated for slices, can IMO be addressed in a separate PR.

Yes, I agree with you about the comments, and that it's better done as a subsequent PR. Those ASM blocks were sources of multiple bugs that took me quite a while to get right, each time with me wondering why we have it like that.

> Looking forward to get this PR merged and to continue the work on references!

Indeed !

I'll work on the other comments that you have and update the PR soon. From a first look, I agree with them all. Thank you.

sway-ir/src/verify.rs
vaivaswatha: You are correct.

But should I make this change now is the question. Perhaps it's better to have a case possible (as you describe the `raw_ptr` case) where it'll fail, and then we make the change? We'll have a unit test too then. What do you think? I don't have a strong opinion though, so can make the change now.

vaivaswatha:
@ironcev I've responded to all your comments (either made the changes or responded to the comment). Let me know if I missed any or if any other change needs to be done.

Thank you again @ironcev !

sway-ir/src/verify.rs
ironcev: Representing `raw_ptr` as IR pointer will anyhow be one of the next steps, so I agree, the change can wait until then.

sway-core/src/ir_generation/function.rs
ironcev: As discussed, we can address this in a separate PR. Here is the issue for that #7227.

sway-core/src/ir_generation/function.rs
ironcev: As discussed, we can address this in a separate PR. Here is the issue for that #7227.

ironcev (approved):

IGI-111 (approved):

JoshuaBatty (approved):

Issue #6942, opened 2025-02-19
Fully implement `#[deprecated]` attribute


ironcev:
Currently, only structs can be marked as `#[deprecated]` and the warning for using deprecated structs will not be emitted for all expressions in which a struct can be used.

Since we do not emit any error or warning for using `#[deprecated]` aside of stucts, this leads to a misleading usage, where items are annotated as `#[depricated]` but that annotation does not have any effect.

E.g., some examples of using `#[deprecated]` without any effect in the standard library:
- `std::constants::ZERO_B256`
- `std::ecr::ec_recover`
- `std::evm::ecr::ec_recover_evm_address`

Warning for usages of `#[deprecated]` outside of struct declarations will be done in #6880.

This issues calls for step-wise implementation of `#[deprecated]` in the following directions:
- [ ] allowing it on all elements where it makes sense.
- [ ] checking for usages of deprecated items in all places.
- [ ] supporting deprecating parent elements and propagating the deprecation to all child elements. E.g., if we deprecated the whole module by annotating the module kind as `#[deprecated]` this should deprecated all the items in that module. Same is with deprecating, e.g., `impl`s.

PR #6986, opened 2025-03-04
Defining, parsing, and checking `#[attribute]`s

### LINKED ISSUE #6880: Forbid using attributes where there are not supposed to be used
Currently, all attributes can annotate any lexical element on which attributes are allowed, e.g., an item, regardless of annotation making sense or not:

```Sway
#[fallback]
#[test]
#[storage(read)]
enum E { }
```

Proposal is to provide a common infrastructure:
- to declaratively define attribute targets,
- and for checking violations.

### LINKED ISSUE #6913: Check attribute multiplicity
Currently, attributes that should be applied only once per item can be applied an arbitrary number of times. E.g., this compiles without any errors or warnings:

```
#[inline(always)]
#[inline(never)]
#[inline(always)]
#[inline(never)]
pub fn inlined_or_not() {
}

#[test]
#[test]
#[test]
fn test() {}
```

### LINKED ISSUE #6914: Check attribute arguments for number of occurrences and values
Currently, attribute arguments are not checked neither for the number of occurrences nor for the values. E.g., this compiles without any errors or warnings:

```
#[storage(read, read, read, read)]
#[storage(read = false)]
fn storage_fn() {}
```

### LINKED ISSUE #6915: Fix wrong "Expected an item after doc comment." error message
The "Expected an item after doc comment." error message appears also for regular attributes. E.g.:

```
#[allow(dead_code)]
                   ^ Expected an item after doc comment.
```

### LINKED ISSUE #6916: Emit not implemented error for inner attributes (`#!`)
Inner attributes are currently not supported but there is no error emitted. In fact, they are applied like being outer attributes. E.g., this:
```
#![allow(dead_code)]
fn fun() {}
```
has the same effect as:
```
#[allow(dead_code)]
fn fun() {}
```

### LINKED ISSUE #6917: Fix "Expected an item" error when mixing doc-comments and other attributes
Doc-comments are attributes and can be arbitrarily mixed with other attributes. E.g., the following example should compile but it produces an "Expected an item" error:
```
/// This is the first line of the comment.
#[allow(dead_code)]
/// This is the second line.
^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Expected an item.
pub fn fun() {}
```

The resulting documentation string in `forc doc` and LSP must be the joint of all the doc comments.

### LINKED ISSUE #6918: Consistently check invalid attribute arguments
Checks for attribute arguments are scattered thought the compiler code and, if done at all, the resulting errors and warnings are not consistent.

E.g., arguments of the `#[inline]` attribute are not checked at all. This compiles without any warnings or errors:
```
#[inline(sometimes)]
pub fn inlined_or_not() { }
```

E.g., the arguments of the `#[cfg]` attribute are checked both as warnings and as errors:

Image

E.g., the arguments of the `#[storage]` attribute are checked only as errors, but with different error message:

Image

### LINKED ISSUE #6931: `cfg` attribute should remove the entire item from type checking, including its annotations
By definition, a `cfg` attribute that cfg-evals to false effectively removes its annotated item from the AST and further type checking. E.g., this compiles when `new_encoding` is active, because the whole function gets removed from the type checking:
```sway
#[cfg(experimental_new_encoding = false)]
fn not_type_checked() -> bool {
    non_existing + non_existing
}
```
Since the item is removed, its annotations must be removed as well. E.g., we shouldn't get warnings or errors for invalid annotations. This is currently not the case. E.g., this code:
```sway
#[allow(something)]        // <<<<---- Emits warning here, although `not_type_checked` does not exist.
#[cfg(experimental_new_encoding = false)]
fn not_type_checked() -> bool {
    non_existing + non_existing
}
```
will emit a warning, although the `not_type_checked` does not exist in the further compilation.

### LINKED ISSUE #6981: Wrong error message on `#[cfg(program_type)]` argument when the argument value is not specified
If the `program_type` argument of the `cfg` attribute does not have a value specified, the error message wrongly states that it excepts a value for the **target** argument.

E.g.:

```
library;

#[cfg(program_type)]
      ^^^^^^^^^^^^ Expected a value for the **target** argument
pub struct S { }
```

### LINKED ISSUE #6983: `dead_code` not applied if it is not the first argument of the last `#[allow]` attribute
E.g., in this case, the `dead_code` attribute will be ignored and the warning will be emitted:
```
#[allow(dead_code)]
#[allow(deprecated)]
fn not_used() { }
```

Also in this case:
```
#[allow(deprecated, dead_code)]
fn not_used() { }
```

### LINKED ISSUE #6984: Remove deprecated warnings from `<autogenerated>` code
#6944 extended deprecation checks across additional expressions. This now causes `<autogenerated>` code for `AbiEncode/Decode` to generate warnings:

```
warning
 --> <autogenerated>:4:21
  |
2 |                 #[allow(dead_code)]
3 |                 fn abi_decode(ref mut buffer: BufferReader) -> Self {
4 |                     Self { a: buffer.decode::<u64>(), }
  |                     ---- deprecated struct: Use "NonDeprecatedStruct" instead.
5 |                 }
6 |             }
  |
____

warning
 --> <autogenerated>:4:39
  |
2 |                 #[allow(dead_code)]
3 |                 fn abi_encode(self, buffer: Buffer) -> Buffer {
4 |                     let buffer = self.deprecated_field.abi_encode(buffer);
  |                                       ---------------- deprecated struct field
```

### LINKED ISSUE #6985: `cfg_eval` exits prematurely if it encounters an `experimental_...` argument
`cfg_eval` must return false if any of the `cfg` attributes return false. Currently `cfg_eval` exists immediately if it encounters an `experimental_...` argument and returns the value of that particular evaluation.

E.g., in this example, the `fun` function exists and can be called in `main`, although the `program_type` is not a `library`.

```sway
script;

#[cfg(experimental_new_encoding = true)]
#[cfg(program_type = "library")]
fn fun() -> bool {
     true
}

fn main() {
    let _ = fun();
}
```

If the order of the two `cfg` attributes changes or the `#[cfg(experimental_new_encoding = true)]` gets removed, the `cfg_eval` will properly evaluate to `false` and the `fun` will not be available in `main`.

ironcev:
## Description

This PR reimplements how `#[attribute]`s are parsed, defined, and checked. It fixes the following issues we had with attributes:
- attributes were in general not checked for their expected and allowed usage.
- when checked, checks were fixing particular reported issues and were scattered through various compilation phases: lexing, parsing, and tree conversion.
- when checked, reported errors in some cases had bugs and were overalapping with other errors and warnings.
- attribute handling was not centralized, but rahter scattered throught the whole code base, and mostly done at use sites.

For concrete examples of these issues, see the list of closed issues below.

The PR:
- encapsulates the attributes handling within a single abstraction: `Attributes`.
- assigns the following properties to every attribute:
  - _target_: what kind of elements an attribute can annotate.
  - _multiplicity_: if more then one attribute of the same kind can be applied to an element.
  - _arguments multiplicity_: a range defining how many arguments an attribute can or must have.
  - _arguments value expectation_: if attribute arguments are expected to have values.
- validates those properties in a single place, during the `convert_parse_tree`. Note that this means that modules with attribute issues will now consistently not reach the type checking phase. This was previously the case for some of the issues with attributes where custom checks where done during the type checking phase. #6987 proposes to move those checks after the tree conversion phase, allowing the continuation of compilation.
- adds the `AttributeKind::Unknow` to preserve unknown attributes in the typed tree.
- removes redundant code related to attributes: specific warnings and errors, specialized parsing, repetitive and error-prone at use site checks, etc.
- removes the unused `Doc` attribute.

The PR also introduces a new pattern in emitting errors. The `attr_decls_to_attributes` function will always return `Attributes` even if they have errors, because:
- we expect the compilation to continue even in the case of errors.
- we expect those errors to be ignored if a valid `#[cfg]` attribute among those evaluates to false.
- we expect them to be emitted with eventual other errors, or alone, at the end of the tree conversion of the annotated element.

For more details, see the comment on `attr_decls_to_attributes` and `cfg_eval`.

- Closes #6880
- Closes #6913
- Closes #6914
- Closes #6915
- Closes #6916
- Closes #6917
- Closes #6918
- Closes #6931
- Closes #6981
- Closes #6983
- Closes #6984
- Closes #6985

## Breaking changes

Strictly seen, this PR can cause breaking changes, but only in the case of invalid existing attribute usage. We treat those breaking changes as bug fixes in the compiler and, thus, do not have them behind a feature flag.

E.g., this kind of code was possible before, but will now emit and error:
```sway
#[storage(read, write, read, write)]
struct Struct {}
```

## Checklist


sway-core/src/language/ty/declaration/declaration.rs
```diff
@@ -780,7 +780,7 @@ impl TyDecl {
         }
     }
 
-    /// Friendly type name string used for various reporting,
+    /// Friendly type name string used for various reportings,
```
jjcnn, line 783:
```suggestion
    /// Friendly type name string used for various reporting,
```

`reporting` is a verb, not a noun.

sway-core/src/language/ty/expression/expression.rs
```diff
@@ -597,11 +597,10 @@ impl TyExpression {
             TyExpressionVariant::FunctionParameter => {}
             TyExpressionVariant::MatchExp {
                 desugared,
-                //scrutinees,
+                // TODO: We must also check scrutinees.
```
jjcnn, line 600: Why is this left as a TODO? Isn't it just a single recursive call?

sway-core/src/transform/attribute.rs
```diff
@@ -1,35 +1,65 @@
-//! Each item may have a list of attributes, each with a name (the key to the hashmap) and a list of
-//! zero or more args.  Attributes may be specified more than once in which case we use the union of
-//! their args.
+//! Each item may have a list of attributes, each with a name  and a list of zero or more args.
```
jjcnn, line 1:
```suggestion
//! Each item may have a list of attributes, each with a name and a list of zero or more args.
```

sway-core/src/transform/attribute.rs
```diff
@@ -69,76 +350,610 @@ pub enum AttributeKind {
...
+            Storage => false,
+            Inline => false,
+            Test => false,
+            Payable => false,
+            Allow => true,
+            Cfg => true,
+            Deprecated => true,
+            Fallback => false,
+        }
+    }
```
jjcnn, line 609: It seems like this function is equivalent to `self.can_annotate_item_kind(ItemKind::Struct) || self.can_annotate_item_kind(ItemKind::Enum)`.

If this is the case, then this leaves us with two parallel implementations of the same semantics, which I don't think is a good idea, and so replacing the function body with two calls to `can_annotate_item_kind` would be better.

A similar concern applies to the following 6 methods.

jjcnn (changes requested):
Quite a bit of spring cleaning going on here - can't complain about that. :-)

Two things need to be addressed: 
- The handling of match scrutinees in `check_deprecated`
- What seems to be a series of duplicated implementations of the same semantics in `Attribute`

I've also added some suggestions for changes in comments. There's one instance of some commented-out code that should be removed, but the rest are just nits.

Finally, I wonder if it's worth grouping the tests in separate subdirectories of `should_fail` and `should_pass/language`, e.g., `should_fail/attributes` and `should_pass/language/attributes`, respectively?

sway-error/src/error.rs
```diff
@@ -2763,32 +2760,155 @@ impl ToDiagnostic for CompileError {
...
+                            hints
+                        },
+                        help: vec![],
                     },
                     _ => Diagnostic {
-                                // TODO: Temporary we use self here to achieve backward compatibility.
-                                //       In general, self must not be used and will not be used once we
-                                //       switch to our own #[error] macro. All the values for the formatting
-                                //       of a diagnostic must come from the enum variant parameters.
+                                // TODO: Temporary we use `self` here to achieve backward compatibility.
```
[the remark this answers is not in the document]
ironcev: Done.

sway-error/src/handler.rs
```diff
@@ -60,15 +60,20 @@ impl Handler {
...
         }
+        // let had_errors = scoped_handler.has_errors();
+
+        // self.append(scoped_handler);
+
+        // if had_errors {
+        //     Err(ErrorEmitted { _priv: () })
+        // } else {
+        //     closure_res
+        // }
```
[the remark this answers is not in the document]
ironcev: Done.

sway-core/src/language/ty/declaration/declaration.rs
ironcev: "To report/reporting" is a verb for sure, but it is a noun as well. E.g., Cambridge Dictionary [lists "reporting" first as a noun](https://dictionary.cambridge.org/dictionary/english/reporting). Also, a search for the exact phrase ["various reportings"](https://www.google.com/search?q=%22various+reportings%22) gives quite some results. That's why I would leave it as is.

sway-core/src/language/ty/expression/expression.rs
ironcev: Unfortunatelly it isn't that straightforward. I thought of doing it, but it is a bit of an effort. We have ~10 different scrutinees, each needs a different treatment, some are recursive... The whole deprication needs additional work and we have a separate issue for that, #6942. There are several TODOs in this file and I've linked them now to that issue.

sway-core/src/transform/attribute.rs
ironcev: Done.

sway-core/src/transform/attribute.rs
ironcev: Good observation! For the current attributes it can be seen like that, but it's actually not the case in general.

For example, the upcoming `#[error]` attribute will be valid only on enum variants. Means, the formula `self.can_annotate_item_kind(ItemKind::Struct) || self.can_annotate_item_kind(ItemKind::Enum)` will not be valid. In this particular case it would actually be `!self.can_annotate_item_kind(ItemKind::Struct) && self.can_annotate_item_kind(ItemKind::Enum)`.

This applies to all the `can_annotate` methods. Each one represents a possible "annotateable" element. It is up to each individual attribute to define if it can annotate that element or not. I was thinking of having a `define_attribute` macro that describes each individual attribute. I still think it's a good idea but wouldn't invest time into it now. This approach would clearly emphasize that each attribute has it's own definition and that looking for similarities between attributes or their annotated elements is by definition not meaningful.

ironcev:
> Quite a bit of spring cleaning going on here - can't complain about that. :-)

Yes, it felt exactly like spring cleaning, throwing all those lines of code out :smile:

> Finally, I wonder if it's worth grouping the tests in separate subdirectories of should_fail and should_pass/language, e.g., should_fail/attributes and should_pass/language/attributes, respectively?

IMO yes. I would like to have the same structure in both tests, but I see it also as a cleanup PR on its own with a low prio. So far I kept following the different "conventions" existing in those two types of tests.

jjcnn (dismissed):
My review comments have been addressed, so LGTM.

sdankel (approved):
The tooling changes look good, thanks!

jjcnn (dismissed):

sdankel (dismissed):

IGI-111 (dismissed):

IGI-111 (approved):

calldelegation (approved):

JoshuaBatty (approved):

PR #6996, opened 2025-03-05
feat: forc-test predicate logging

zees-dev:
## Description

This PR adds support for logging in predicates via the use of the `ECAL` opcode.
In the `ECAL` implementation we simply replicate the functionality of the `LOGD` instruction - but for predicates.

### Notes:
- To make this work we needed to create and propagate the `enable_predicate_logs` attribute in the `BuildProfile`, `BuildConfig`, and `Context` structs; this needed to be propagated down deep into the `ir_generation` so that the `__log` compiler intrinsic could branch based on the `enable_predicate_logs` attribute and the predicate program `Kind`.
- With this change; the `forc test --logs` and `forc test --raw-logs` commands will now output logs for predicates.
- `forc build` will compile the predicate (which contains logs) in `debug` mode successfully
- `forc build` will **NOT** compile the predicate (which contains logs) in `release` mode
- scripts, contracts, libraries should remain unaffected by this change

### Predicate
```sway
predicate;

pub struct CallParams {
    pub coins: u64,
    pub gas: u64,
}

fn main() -> bool {
    let call_params = CallParams {
        coins: 7,
        gas: 8,
    };
    log(call_params);
    false
}

#[test]
fn test_meaning_of_life() {
    log("test_meaning_of_life");
    assert(main() == false);
}
```

### `> forc test --logs`

Screenshot 2025-03-06 at 11 02 55 AM

### `> forc test --raw-logs`

Screenshot 2025-03-06 at 11 03 34 AM

## Checklist


forc-pkg/src/manifest/build_profile.rs
```diff
@@ -57,6 +59,7 @@ impl BuildProfile {
             profile: false,
             metrics_outfile: None,
             include_tests: false,
+            log_generation: false,
```
kayagokalp, line 62:
```suggestion
            log_generation: true,
```

forc-pkg/src/pkg.rs
```diff
@@ -309,6 +309,10 @@ pub struct BuildOpts {
     pub error_on_warnings: bool,
     /// Include all test functions within the build.
     pub tests: bool,
+    /// Whether or not to generate output logs for forc test.
```
kayagokalp, line 312:
```suggestion
    /// Whether or not to generate output logs for forc test.
```

I think we can improve the wording here because this flag does not alter the behviour of logs for contracts etc right? 

We should probably say that this flag is enabling logs in predicates or something like that maybe?

kayagokalp (commented):
Nice one! This looks good to me in general we should probably make `log_generation` flag true in debug build profile by default which i left as a comment as well.

Also we should consider renaming the this flag in my opinion. Because logs are already being generated for other program kinds other than predicates without this flag. `enable_predicate_logs` or something like this might be more on point (while being a little verbose, maybe we can find a common ground between this and enable_logs)

forc-test/src/ecal.rs
```diff
@@ -0,0 +1,130 @@
...
+/// # Output Format
+/// - Successfully decoded data is printed to console
+/// - Decoding failures are printed in red with additional debug information
+///
+/// # Dependencies
+/// The struct requires a program ABI to properly decode log data according to the
+/// types defined in the Sway program.
+#[derive(Debug, Clone)]
+pub struct PredicateLoggingEcal {
+    pub program_abi: Arc<ProgramABI>,
```
kayagokalp, line 37: Are we going to be using this for something? If i am not mistaken this is unused at the moment

zees-dev: A remnant from an initial implementation; addressed: https://github.com/FuelLabs/sway/pull/6996/commits/ca7530326e052a286eba4b9cd217d3ba6a4bf487

forc-pkg/src/pkg.rs
zees-dev: addressed: https://github.com/FuelLabs/sway/pull/6996/commits/54eb60dee3120de982f85a91a2f4e6dabd4cfa6b
Feel free to suggest improvement(s).

forc-pkg/src/manifest/build_profile.rs
zees-dev: https://github.com/FuelLabs/sway/pull/6996/commits/18d49489bb6899d3fbb19b361d546a01d4ffe3fe

forc-plugins/forc-client/src/op/deploy.rs
```diff
@@ -484,7 +484,7 @@ pub async fn deploy_executables(
         if let Some(loader_data_section_offset) =
             loader_data_offset(&pkg.bytecode.bytes, &BlobId::default())?
         {
-            if let ProgramABI::Fuel(mut fuel_abi) = pkg.program_abi.clone() {
+            if let ProgramABI::Fuel(mut fuel_abi) = pkg.program_abi.as_ref().clone() {
```
alfiedotwtf, line 487: Just checking... couldn't this be left as `pkg.program_abi.clone()`?

zees-dev:
> Nice one! This looks good to me in general we should probably make `log_generation` flag true in debug build profile by default which i left as a comment as well.
> 
> Also we should consider renaming the this flag in my opinion. Because logs are already being generated for other program kinds other than predicates without this flag. And users might end up mixing this with `--logs` in forc-test flags. `enable_predicate_logs` or something like this might be more on point (while being a little verbose, maybe we can find a common ground between this and enable_logs)

Agreed; renamed: https://github.com/FuelLabs/sway/pull/6996/commits/3ec072fa2667d6e2f339a3d452990778cdb886ec

forc-pkg/src/manifest/build_profile.rs
kayagokalp: We might want to update PR description:

| forc build will still continue to fail for predicates which contain logs (current behaviour)

I think we build in debug mode (not exactly sure) by default so it will build

forc-plugins/forc-client/src/op/deploy.rs
zees-dev: Does not compile without `.as_ref()`; the behaviour remains the same; the inner data is still copied over (as per original implementation).
A larger refactor would be required to actually utilize the arc count :/

kayagokalp:
I think this should be a breaking change? by default logs in predicates were a compile error but if the user does not pass in `--release` it will compile fine now. 

What do you think @FuelLabs/sway-compiler? Are you guys fine with `enable_predicate_logs` being on by default in `debug` builds but not in `release` builds?

sway-core/src/build_config.rs
```diff
@@ -312,6 +314,18 @@ impl BuildConfig {
         }
     }
 
+    /// Whether or not to generate output logs  for forc test.
```
alfiedotwtf, line 317: nit:

```
- output logs  for forc
+ output logs for forc
```

alfiedotwtf (changes requested):
Apart from my changes and question, looks good.

sway-core/src/build_config.rs
zees-dev: https://github.com/FuelLabs/sway/pull/6996/commits/f4f514bd7a84c7e189701038c978dab5b81baec6

alfiedotwtf (dismissed):

forc-plugins/forc-client/src/op/deploy.rs
kayagokalp: Unresolving this to make ensure you see this one, if you still feel like this is ok feel free to resolve:

I think this makes this a little bit anti-pattern as a reader of this code i see that an Arc is introduced my initial thought is that we do this remove deep copies of abi.  So while reviewing i thouhgt even though it is not related tot he feature implemented here it is ok not to rever this. But this can be dangerous in the later run if we end-up needing to profile things. We had instances where big clones hurt the memory usage heavily and this change might make the initial diagnosis for something like that a little hard to detect.

zees-dev: I think the `Arc` is quite helpful to have in the struct for any downstream references; there are more upside(s) of keeping it there - rather than having to re-introduce it if anything else downstream needs to clone it.
Instead i'll create a gh issue and reference it in the code.

zees-dev: addressed: https://github.com/FuelLabs/sway/pull/6996/commits/42b6e67d3c02352a9fc8bd0749048e3b8979e723
issue: https://github.com/FuelLabs/sway/issues/6997

alfiedotwtf (approved):

kayagokalp (changes requested):
I think we need the docs updated before merging with the build profile flag (enable predicate logging) and general limitations around predicates and logging such as: 

1) https://docs.fuel.network/guides/intro-to-predicates/debugging-with-scripts/#debugging-with-scripts
2) https://docs.fuel.network/docs/forc/manifest_reference/#the-build-profile-section

For first one we might need to contact @calldelegation to see where these docs are hosted.

IGI-111:
> forc build will NOT compile the predicate (which contains logs) in release mode

What does this mean, that `forc build --release` on a predicate will omit the logs or that it will fail to compile?

IGI-111:
> I think this should be a breaking change? by default logs in predicates were a compile error but if the user does not pass in `--release` it will compile fine now.
> 
> What do you think @FuelLabs/sway-compiler? Are you guys fine with `enable_predicate_logs` being on by default in `debug` builds but not in `release` builds?

I don't see a way for this to break existing released code or codebases. The compiler explicitly does not guarantee that it will compile to the same output across versions (otherwise we couldn't do optimizations).

As far as I understand, all that will change from merging this, from a user standpoint, is that a predicate compiled in debug mode will now be different and contain logging.

If this also changed release output, I'd be more careful because it might break some production assumptions about reproducibility, but changing debug behavior in this way seems totally fine.

zees-dev:
> > forc build will NOT compile the predicate (which contains logs) in release mode
> 
> What does this mean, that `forc build --release` on a predicate will omit the logs or that it will fail to compile?
> 
> Also will setting the build profile flag override this behavior or not?

Yeap release mode will throw typical error (`Using intrinsic "log" in a predicate is not allowed.`).
You can explicitly override this though (even in release mode):
```toml
[build-profile.release]
enable-predicate-logs = true
```

zees-dev:
> I think we need the docs updated before merging with the build profile flag (enable predicate logging) and general limitations around predicates and logging such as:
> 
> 1. https://docs.fuel.network/guides/intro-to-predicates/debugging-with-scripts/#debugging-with-scripts
> 2. https://docs.fuel.network/docs/forc/manifest_reference/#the-build-profile-section
> 
> For first one we might need to contact @calldelegation to see where these docs are hosted.

Address 2: https://github.com/FuelLabs/sway/pull/6996/commits/5c80e7b0406ac22e495cc64ef00f042b3a5f3de7
As for 1; I propose we can address that later - ensure that the docs are updated before next release - so that it doesnt block this from being merged.

zees-dev:
> Should we have a test for a project that has logging triggered from a sway unit test?

addressed: https://github.com/FuelLabs/sway/pull/6996/commits/64d5b695b8d65f28fd8219c9af8746173906515e

sdankel (approved):
LGTM, nice work!

kayagokalp (approved):
Nice one!

JoshuaBatty (approved):
:shipit:

IGI-111 (changes requested):
> Yeap release mode will fail to compile (throws typical error - `Using intrinsic "log" in a predicate is not allowed.`). You can explicitly override this though (even in release mode):
> 
> ```toml
> [build-profile.release]
> enable-predicate-logs = true
> ```
Okay, this needs to be clearer in my opinion, we should have the error direct the user to a solution (either removing the logs or turning on this option) and making it clear what's happening here.

At the least the error should be updated to tell the truth (that it's allowed under certain circumstances).

IGI-111:
I also think we should discuss eliding log predicate invocation in release builds by default instead of erroring out. I'm not sure that's the best behavior, but we should consider it versus having the user modify their source code when building for release. That would likely prove annoying.

kayagokalp:
I think this is superseded by #7015

Issue #6994, opened 2025-03-05
Implement `TryFrom<Bytes>` for `b256`


ironcev:
This is a tracking issue for _Implement `TryFrom<Bytes>` for `b256`_.
The experimental feature flag for the issue is `try_from_bytes_for_b256`.

## Description

Currently, `b256` implements `From<Bytes>` which can cause data losses. We want to remove that implementation and replace it with `TryFrom<Bytes>` which will return `None` if the provided `Bytes` cannot be safely converted to `b256`.

### Replace calls to `from(bytes)/bytes.into()` with `try_from(bytes)/bytes.try_into()`

E.g., calls to `from`:
```Sway
let result = b256::from(some_bytes);
```
will become:
```Sway
let result = b256::try_from(some_bytes).unwrap();
```

E.g., calls to `into`:
```Sway
let result: b256 = some_bytes.into();
```
will become:
```Sway
let result: b256 = some_bytes.try_into().unwrap();
```

Calling `unwrap()` assumes that the previous usage was always semantically valid. It is recommended to inspect all the previous usages of `from/into` to ensure `unwrap`ing is always safe, or to check for the `Option` result of `try_from/try_into`.

PR #7005, opened 2025-03-09
Implement trees visitors and migrations for `TryFrom<Bytes>` for `b256`

ironcev:
## Description

This PR:
- extends infrastructure for writing migrations that modify individual expressions and do not need access to a broader scope. It does that by providing lexed and typed tree visitors. Visiting lexed and typed tree is implemented to the level needed to write the migration for `impl TryFrom<Bytes> for b256`.
- implements the migration for `impl TryFrom<Bytes> for b256` as explained in #6994.

Adding `#[cfg(experimental_try_from_bytes_for_b256 = true)]` to the `impl TryFrom<Bytes> for b256` revealed that there existed already an `impl TryFrom<Bytes> for b256` in `std::primitive_conversions::b256`. Due the lack of trait coherence there was no any errors when it was re-implemented in `std::bytes`.

This PR marks the existing impl in `std::primitive_conversions::b256` with `#[cfg(experimental_try_from_bytes_for_b256 = false)]` so that it will be removed in the next breaking release, in favor of the new impl in `std::bytes`.

## Checklist


IGI-111 (dismissed):

xunilrj (dismissed):

sway-lib-std/src/primitive_conversions/b256.sw
```diff
@@ -6,6 +6,10 @@ use ::option::Option::{self, *};
 use ::u128::U128;
 use ::b512::B512;
 
+// TODO: Delete this implementation in favor of the one in `std::bytes`.
+//       Due the fact that we still don't have trait coherence, these
+//       two impls coexist for now.
+#[cfg(experimental_try_from_bytes_for_b256 = false)]
```
jjcnn, line 12: The implementations are semantically equivalent AFAICT. It's a shame we can't solve this by replacing one of the implementations with a `pub use` without also exporting a lot of other stuff.

jjcnn (approved):

forc-plugins/forc-migrate/src/migrations/try_from_bytes_for_b256.rs
```diff
@@ -0,0 +1,115 @@
...
+            // Note that neither the implementing for type not the trait are a
+            // part of the `from` function call path. All associated `from` functions
+            // in the `std::bytes` will have the same call path.
+            // We will filter further below to target exactly the `<From<Bytes> for b256>::from`.
+            let from_call_path = CallPath::fullpath(&["std", "bytes", "from"]);
+
+            let Some(implementing_for_type_id) = ty_fn_call_info.fn_decl.implementing_for_typeid
+            else {
+                return Ok(InvalidateTypedElement::No);
+            };
```
zees-dev, line 74: nit: can move above `let from_call_path` for earlier return in the else-case

zees-dev (approved):
LGTM 👍 

Cant see any major issues - other than a bunch of todo's which would need to be addressed at some point.
Would advise an issue is created for these - so that they won't be forgotten (unless the next immediate task entails working on these).

ironcev:
@zees-dev Thanks for the review! We strive to put into migrations only the effort that is immediately needed. Some of those TODOs will very likely stay open until there is an actual migration that requires that path to be covered. Otherwise I am also always for having GitHub issue, but in this particular case not having the TODOs implemented is a part of the approach and investment into migrations infrastructure.

JoshuaBatty (approved):

PR #7015, opened 2025-03-11
`Debug` trait and its auto implementation

xunilrj:
## Description

This PR implements the `__dbg(...)` intrinsic, which is very similar to Rust `dbg!(...)` macro.

Up until now, it has being the norm to use `__log` to debug values. Given that this is NOT the first use case for log, we have always found some issues with it: log does not work on predicates, log does not show "private" fields like `Vec::capacity` and others.

To solve these problems `__dbg` is being introduced:
1 - it will work on all program types, including predicates;
2 - it also prints the file name, line and column;
3 - All types will have an automatic implementation of Debug if possible, which can still be customized.
4 - Even `raw_ptr` and other non "loggable" types, have `Debug` impls.
5 - `__dbg` will be completely stripped in the release build by default. It can be turned on again if needed.

So this:

```
// Aggregates
let _ = __dbg((1u64, 2u64));
let _ = __dbg([1u64, 2u64]);

// Structs and Enums
let _ = __dbg(S { });
let _ = __dbg(E::None);
let _ = __dbg(E::Some(S { }));
```

will generate this:

```
[src/main.sw:19:13] = (1, 2)
[src/main.sw:20:13] = [1, 2]
[src/main.sw:23:13] = S { }
[src/main.sw:24:13] = None
[src/main.sw:25:13] = E(S { })
```

How does this work?

`__dbg(value)` intrinsic is desugared into `{ let f = Formatter{}; f.print_str(...); let value = value; value.fmt(f); value }`.

`Formatter` is similar to Rust's one. The difference is that we still do not support string formatting, so the `Formatter` has a lot of  `print_*` functions.

And each `print` function calls a "syscall". This `syscall` uses `ecal` under the hood and it follows unix write syscall schema.

```sway
// ssize_t write(int fd, const void buf[.count], size_t count);
fn syscall_write(fd: u64, buf: raw_ptr, count: u64) {
    asm(id: 1000, fd: fd, buf: buf, count: count) {
        ecal id fd buf count;
    }
}
```

For that to work, the VM interpreter must have its `EcalState` setup and interpret syscall number 1000 as `write`. This PR does this for `forc test` and our `e2e test suite`.

Each test in `forc test` will capture these calls and only print to the terminal when requested with the `--log` flag.

## Garbage Collector and auto generated

Before, we were associating all auto-generated code with a pseudo file called "&lt;autogenerated&gt;.sw" that was never garbage collected.

This generated a problem inside the LSP when the `auto_impl.rs` ran a second time because of a collision in the "shareable type" map. When we try to solve this collision, choosing to keep the old value or to insert the new, the type inside the map points to already collected types and the compiler panics. This is a known problem.

The workaround for this is to break the auto-generated code into multiple files. Now they are named "main.autogenerated.sw", for example. We create one pseudo-file for each real file that needs one.

When we garbage collect one file, `main.sw`, for example, we also collect its associated auto-generated file.

## Checklist


forc-test/src/execute.rs
```diff
@@ -23,10 +21,85 @@ use vm::interpreter::{InterpreterParams, MemoryInstance};
...
+        b: RegId,
+        c: RegId,
+        d: RegId,
+    ) -> fuel_vm::error::SimpleResult<()>
+    where
+        M: fuel_vm::prelude::Memory,
+    {
+        let regs = vm.registers();
+        match regs[a.to_u8() as usize] {
+            1000 => {
```
Voxelot, line 76: Should `1000` be put into some kind of consts file? This seems like a pretty important magic number that people should know about.

xunilrj: Good idea.

forc-test/src/execute.rs
```diff
@@ -23,10 +21,85 @@ use vm::interpreter::{InterpreterParams, MemoryInstance};
...
+                let addr = regs[c.to_u8() as usize];
+                let count = regs[d.to_u8() as usize];
```
zees-dev, line 78: nit: naming; len or length - since we are reading a length from offset/address

forc-test/src/execute.rs
```diff
@@ -23,10 +21,85 @@ use vm::interpreter::{InterpreterParams, MemoryInstance};
...
+            }
+            CapturedEcal::Unknown { ra, rb, rc, rd } => {
+                println!("Unknown ecal: {} {} {} {}", ra, rb, rc, rd);
+            }
+        }
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct EcalState {
```
zees-dev, line 53: Note: Could use some doc-comments for this - since we seem to be writing to a Vec here (its not a simple ecal impl)

forc-test/src/lib.rs
```diff
@@ -74,6 +75,8 @@ pub struct TestResult {
     pub logs: Vec<fuel_tx::Receipt>,
     /// Gas used while executing this test.
     pub gas_used: u64,
+    /// EcalState of the execution
+    pub ecal: Box<EcalState>,
```
zees-dev, line 79: maybe can use reference or arc here?

forc-test/src/execute.rs
```diff
@@ -23,10 +21,85 @@ use vm::interpreter::{InterpreterParams, MemoryInstance};
 use vm::state::DebugEval;
 use vm::state::ProgramState;
 
+#[derive(Debug, Clone)]
+pub enum CapturedEcal {
+    Write { fd: u64, bytes: Vec<u8> },
+    Unknown { ra: u64, rb: u64, rc: u64, rd: u64 },
+}
+
+impl CapturedEcal {
```
zees-dev, line 30: maybe we can move this to another file?
i.e. `ecal.rs` or something?

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/abi_encoding.rs
```diff
@@ -208,30 +208,36 @@ where
...
         let abi_decode_code = self.generate_abi_decode_code(
             struct_decl.name(),
             &struct_decl.type_parameters,
             abi_decode_body?,
         );
-        let abi_decode_node =
-            self.parse_impl_trait_to_ty_ast_node(engines, program_id, &abi_decode_code);
+        let abi_decode_node = self.parse_impl_trait_to_ty_ast_node(
+            engines,
+            struct_decl.span().source_id().cloned(),
```
zees-dev, line 232: nit: may not need cloned here?

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
```diff
@@ -1994,6 +1995,199 @@ fn expr_func_app_to_expression_kind(
...
+        //      f.print_str(":");
+        //      f.print_str(<current col>);
+        //      f.print_str("] ");
+        //      f.print_str(<current expr>);
+        //      f.print_str(" = ");
+        //      let arg = arg;
+        //      arg.fmt(f);
+        //      arg
+        // }"
+        Some(Intrinsic::Dbg)
```
zees-dev, line 2013: nice; makes much more sense now

sway-error/src/handler.rs
```diff
@@ -28,6 +28,7 @@ impl Handler {
 
     /// Emit the error `err`.
     pub fn emit_err(&self, err: CompileError) -> ErrorEmitted {
+        // eprintln!("{:?} {}", &err, std::backtrace::Backtrace::force_capture());
```
zees-dev, line 31: nit: remove before merging

forc-test/src/execute.rs
xunilrj: It's named `count` because this is the name of the arg in ` vm.memory().read(...)`.

forc-test/src/execute.rs
xunilrj: Done.

forc-test/src/lib.rs
xunilrj: This is inside `TestResult`, which means the `EcalState` is not shared between tests.

forc-test/src/execute.rs
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/abi_encoding.rs
xunilrj: Done.

sway-error/src/handler.rs
xunilrj: Done.

JoshuaBatty:
This is looking great @xunilrj really excited to see this land. Any ideas what is cause the performance hit and if it can be rectified?

xunilrj:
> This is looking great @xunilrj really excited to see this land. Any ideas what is cause the performance hit and if it can be rectified?

This is expected because we are generating auto-impl for all strucs/enums for the new `Debug` trait.

test/src/snapshot/mod.rs
```diff
@@ -77,40 +77,60 @@ pub(super) async fn run(filter_regex: Option<&regex::Regex>) -> Result<()> {
...
+                                    if line.contains(arg) {
+                                        new_output.push_str(line);
+                                        new_output.push('\n');
+                                    }
+                                }
+                                last_output = Some(new_output);
+                            }
+                            continue;
+                        } else {
+                            panic!("Not supported. Possible commands: forc")
```
zees-dev, line 110: grep seems to be supported to no?

sway-lib-std/src/debug.sw
zees-dev, line 1: nice; single file which contains full functionality for debug in sway

docs/book/src/debugging/index.md
zees-dev, line 1: nice, very clear

forc-test/src/ecal.rs
zees-dev, line 1: would be great to have unit test for the ecal - specifically specification of VM opcode - which would translate to debugging of a variable.
Im thinking something along the lines of this test (initial PR which tried to implement logging in predicates via ecal): https://github.com/FuelLabs/sway/blob/4ae3ab16db23159f463802586316fdc87707abf7/forc-test/src/ecal.rs#L67

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/debug.rs
```diff
@@ -0,0 +1,147 @@
...
+            _ => None,
+        }
+    }
+
+    // Auto implements Debug for structs and returns their `AstNode`s.
+    fn debug_auto_impl_struct(
+        &mut self,
+        engines: &Engines,
+        decl: &TyDecl,
+    ) -> Option<Option<TyAstNode>> {
```
IGI-111, line 35: It doesn't look like the first Option is of any use here, is it just a usage convenience?

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/debug.rs
```diff
@@ -0,0 +1,147 @@
...
+            decl.name().as_str()
+        )
+    }
+
+    // Auto implements Debug for enums and returns their `AstNode`s.
+    fn debug_auto_impl_enum(
+        &mut self,
+        engines: &Engines,
+        decl: &TyDecl,
+    ) -> Option<Option<TyAstNode>> {
```
IGI-111, line 97: Same here, why the double Option?

IGI-111 (commented):
I think this generally looks good and is the superior approach to debugging.

However we'll need two more things in the future to make this convenient:
* some kind of standard for Sway syscalls to avoid collision problems conventionally
* a nicer syntax than `__dbg(x)` that still preserves the location information

But these can be addressed later without issue.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/debug.rs
xunilrj: You are right. Done.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/debug.rs
xunilrj: Done.

test/src/snapshot/mod.rs
xunilrj: `grep` is indeed supported. Well.... a very limited version of it, to simplify the snapshots.

forc-test/src/ecal.rs
xunilrj: Good idea! Done.

docs/book/src/reference/compiler_intrinsics.md
```diff
@@ -372,3 +372,18 @@ Runtime bound checks are not generated, and must be done manually when and where
...
 - `item` is a reference to an array or a reference to a slice;
 - when `index` is a literal, it must be smaller than `item` length;
+
+---
+
+```sway
+__dbg(value: T) -> T
+```
+
+**Description:** Automatically calls the `Debug` trait on the passed `value`, with file, line and column information. The value is returned to allow be used inside any expression.
```
ironcev, line 382:
```suggestion
**Description:** Automatically calls the `Debug` trait on the passed `value`, with file, line and column information. The value is returned to allow to be used inside of any expression.
```

docs/book/src/reference/compiler_intrinsics.md
```diff
@@ -372,3 +372,18 @@ Runtime bound checks are not generated, and must be done manually when and where
...
+
+Only generate the expected output in `Debug` mode, when in `Release` behaves as identity function, just returning the passed value.
```
ironcev, line 384:
```suggestion
Only generates the expected output in `Debug` mode. When in `Release`, it behaves as identity function, just returning the passed value.
```

docs/book/src/reference/compiler_intrinsics.md
```diff
@@ -372,3 +372,18 @@ Runtime bound checks are not generated, and must be done manually when and where
...
+To enable the output generation in "Release", must be enabled in the project `Forc.toml`
```
ironcev, line 385: This formulation sounds confusing to me. What exactly must be enabled and how? Having an example here how to enable it in the `Forc.toml` would IMO be very beneficial.

```suggestion
To enable the output generation in `Release` mode, must be enabled in the project's `Forc.toml`.
```

forc-plugins/forc-doc/src/lib.rs
```diff
@@ -109,6 +109,7 @@ pub fn compile_html(
         None,
         &build_instructions.experimental.experimental,
         &build_instructions.experimental.no_experimental,
+        sway_core::DbgGeneration::Full,
```
ironcev, line 112: I understand that in the end for the documentation generation it doesn't make any difference, but still the `DbgGeneration::None` looks like intuitive first choice here. Why build with debug info on when it's not needed?

forc-plugins/forc-migrate/src/cli/shared.rs
```diff
@@ -108,6 +108,7 @@ pub(crate) fn compile_package<'a>(
         None,
         &build_instructions.experimental.experimental,
         &build_instructions.experimental.no_experimental,
+        sway_core::DbgGeneration::Full,
```
ironcev, line 111: Same reasoning here for the `forc migrate` tool.

ironcev (changes requested):
Great feature looking forward to get it supported! :rocket:

Below, a few smaller points I've noticed when starting to walk through the implementation.

docs/book/src/reference/compiler_intrinsics.md
xunilrj: Done.

docs/book/src/reference/compiler_intrinsics.md
xunilrj: Done.

docs/book/src/reference/compiler_intrinsics.md
xunilrj: I replaced this paragraph.

forc-pkg/src/manifest/mod.rs
```diff
@@ -213,6 +213,7 @@ pub struct Project {
     #[serde(default)]
     pub experimental: HashMap<String, bool>,
     pub metadata: Option<toml::Value>,
+    pub dbg_release: Option<bool>,
```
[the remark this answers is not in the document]
xunilrj: Went with "force_dbg_in_release". Emphasizes something is being forced.

forc-plugins/forc-doc/src/lib.rs
xunilrj: If we use `DbgGeneration::None`, it is possible that `forc doc` will compile, but later on `forc build` will not, because one expression being debugged is of a type that does not implement `Debug`.

In this case, I think it makes more sense to pass the full generation.

forc-plugins/forc-migrate/src/cli/shared.rs
xunilrj: See above.

forc-plugins/forc-doc/src/lib.rs
ironcev: I still didn't go through the all of the implementation, but isn't the type check something we need no matter if we emit `__dbg(something)` to `ecal` or to `something` just passing the something through. That's the internal implementation of `__dbg()`, but on the signature side, it should always expect `T: Debug`.

docs/book/src/reference/compiler_intrinsics.md
```diff
@@ -372,3 +372,37 @@ Runtime bound checks are not generated, and must be done manually when and where
 
 - `item` is a reference to an array or a reference to a slice;
 - when `index` is a literal, it must be smaller than `item` length;
+
+---
+
+```sway
+__dbg(value: T) -> T
```
ironcev, line 379: As addition to the discussion on typechecking the `__dbg` argument, shouldn't the signature be:
```
__dbg<T>(value: T) -> T where T: Debug
```
(Noticed now that `<T>` is missing in the docs :smile:)

xunilrj: Done. I also fixed other missing places.

ironcev (changes requested):

docs/book/src/debugging/index.md
```diff
@@ -11,3 +11,48 @@ node to exercise your Sway code. Instruction-by-instruction debugging is availab
 
 - [Debugging with CLI](./debugging_with_cli.md)
```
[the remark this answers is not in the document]
xunilrj: Done.

forc-test/src/ecal.rs
```diff
@@ -0,0 +1,161 @@
...
+        }
+    }
+}
+
+/// Handle VM `ecal` as syscalls.
+///
+/// The application of the syscalls can be turned off,
+/// guaranteeing total isolation from the outside world.
+///
+/// Capture of the syscalls can be turned on, allowing the
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-ast/src/intrinsics.rs
```diff
@@ -45,6 +45,7 @@ pub enum Intrinsic {
     Slice, // let ref_to_slice = __slice::<T: array or ref_to_slice>(item: T, inclusive_start_index, exclusive_end_index)
     ElemAt, // let elem: &T = __elem_at::<T: array or ref_to_slice>(item: T, index)
     Transmute, // let dst: B = __transmute::<A, B>(src)
+    Dbg,   // __dbg(1, 2)
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/lib.rs
```diff
@@ -97,7 +98,7 @@ pub fn parse(
     experimental: ExperimentalFeatures,
 ) -> Result<(lexed::LexedProgram, parsed::ParseProgram), ErrorEmitted> {
     match config {
-        None => parse_in_memory(handler, engines, input, experimental),
+        None => parse_in_memory(handler, engines, input, experimental, DbgGeneration::None),
         // When a `BuildConfig` is given,
         // the module source may declare `dep`s that must be parsed from other files.
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/debug.rs
```diff
@@ -0,0 +1,139 @@
...
+        match decl {
+            TyDecl::StructDecl(_) => self.debug_auto_impl_struct(engines, decl),
+            TyDecl::EnumDecl(_) => self.debug_auto_impl_enum(engines, decl),
+            _ => None,
+        }
+    }
+
+    // Auto implements Debug for structs and returns their `AstNode`s.
+    fn debug_auto_impl_struct(&mut self, engines: &Engines, decl: &TyDecl) -> Option<TyAstNode> {
+        if self.ctx.namespace.current_package_name().as_str() == "core" {
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/abi_encoding.rs
```diff
@@ -190,7 +190,7 @@ where
     }
 
     // Auto implements AbiEncode and AbiDecode for structs and returns their `AstNode`s.
-    fn auto_impl_struct(
+    fn abi_encode_and_decode_auto_impl_struct(
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/abi_encoding.rs
```diff
@@ -208,30 +208,36 @@ where
...
+            struct_decl.span().source_id(),
+            &abi_decode_code,
+            crate::build_config::DbgGeneration::None,
+        );
 
         Some((abi_encode_node.ok(), abi_decode_node.ok()))
     }
 
-    fn auto_impl_enum(
+    fn abi_encode_and_decode_auto_impl_enum(
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/debug.rs
```diff
@@ -0,0 +1,139 @@
...
+
+use crate::{
+    decl_engine::DeclEngineGet,
+    language::ty::{self, TyAstNode, TyDecl, TyEnumDecl, TyStructDecl},
+    Engines, TypeParameter,
+};
+
+use super::abi_encoding::AbiEncodingAutoImplInfo;
+
+pub type DebugAutoImplContext<'a, 'b> = super::AutoImplContext<'a, 'b, AbiEncodingAutoImplInfo>;
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/debug.rs
```diff
@@ -0,0 +1,139 @@
...
+        body: String,
+    ) -> String {
+        let type_parameters_declaration =
+            self.generate_type_parameters_declaration_code(type_parameters);
+        let type_parameters_constraints =
+            self.generate_type_parameters_constraints_code(type_parameters, "Debug");
+
+        let name = name.as_str();
+
+        format!("#[allow(dead_code)] impl{type_parameters_declaration} Debug for {name}{type_parameters_declaration}{type_parameters_constraints} {{
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/debug.rs
```diff
@@ -0,0 +1,139 @@
...
+            #[allow(dead_code)]
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/debug.rs
```diff
@@ -0,0 +1,139 @@
...
+
+        format!(
+            "_f.debug_struct(\"{}\"){fields}.finish();",
+            decl.name().as_str()
+        )
+    }
+
+    // Auto implements Debug for enums and returns their `AstNode`s.
+    fn debug_auto_impl_enum(&mut self, engines: &Engines, decl: &TyDecl) -> Option<TyAstNode> {
+        if self.ctx.namespace.current_package_name().as_str() == "core" {
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/debug.rs
```diff
@@ -0,0 +1,139 @@
...
+where
+    'a: 'b,
+{
+    pub fn generate_debug_impl(
+        &mut self,
+        engines: &Engines,
+        decl: &ty::TyDecl,
+    ) -> Option<TyAstNode> {
+        match decl {
+            TyDecl::StructDecl(_) => self.debug_auto_impl_struct(engines, decl),
```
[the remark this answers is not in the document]
xunilrj: Done.

sway-core/src/engine_threading.rs
```diff
@@ -52,11 +52,26 @@ impl Engines {
...
+    ///
+    /// It will also clear the associated autogenerated file for the `source_id` parameter.
     pub fn clear_module(&mut self, source_id: &sway_types::SourceId) {
         self.type_engine.clear_module(source_id);
         self.decl_engine.clear_module(source_id);
         self.parsed_decl_engine.clear_module(source_id);
         self.query_engine.clear_module(source_id);
+
+        // Check if `source_id` has an associated autogenerated file
+        // and clear it
```
JoshuaBatty, line 64: should we add this logic to the `clear_program` function above. I don't think it's getting used anywhere from memory at the moment but if we are keeping it around it probably makes sense to add this logic there as well.

sway-lsp/src/core/session.rs
```diff
@@ -195,6 +196,7 @@ impl Session {
         position: Position,
     ) -> Option<GotoDefinitionResponse> {
         let _p = tracing::trace_span!("token_definition_response").entered();
+        dbg!(uri.as_str(), position);
```
JoshuaBatty, line 199: can we remove this `dbg!`

sway-lsp/tests/lib.rs
```diff
@@ -1239,6 +1239,7 @@ fn go_to_definition_for_functions() {
 }
 
 #[test]
+#[ignore]
```
JoshuaBatty, line 1242: Can we link this issue here https://github.com/FuelLabs/sway/issues/7025

sway-core/src/engine_threading.rs
xunilrj: I don´t think we need it because the auto-generated pseudo-file is associated with the same "program_id" as the original "source_id". Thus, clearing the program will clear them both.

sway-lsp/src/core/session.rs
xunilrj: Done.

sway-lsp/tests/lib.rs
xunilrj: Done.

sway-core/src/engine_threading.rs
JoshuaBatty: ok cool thanks for clarifying.

docs/book/src/reference/compiler_intrinsics.md
```diff
@@ -372,3 +374,37 @@ Runtime bound checks are not generated, and must be done manually when and where
...
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+license = "Apache-2.0"
+entry = "main.sw"
+name = "some-project"
+force-dbg-in-release = true
+```
+
+It is strongly suggested to always remove this flag before publishing binaries as it will not have any effect when running
+on real nodes and it only increases gas usage.
```
zees-dev, line 406: Wondering whether we should even charge gas for debug since this doesnt affect real nodes.
I can imagine scenario's where one has a ton of debug's in code and also wants to gas-golf; but then they'll need to remove/comment out all these lines.

forc-test/src/ecal.rs
```diff
@@ -0,0 +1,161 @@
...
+                let s = std::str::from_utf8(bytes.as_slice()).unwrap();
+
+                use std::io::Write;
+                use std::os::fd::FromRawFd;
+
+                let mut f = unsafe { std::fs::File::from_raw_fd(*fd as i32) };
+                write!(&mut f, "{}", s).unwrap();
+
+                // Dont close the fd
+                std::mem::forget(f);
```
zees-dev, line 28: wondering if there is another non-unsafe way to do this.
Using AI i came up with this - although unsure if it suffices.

```rust
// Try to create a Rust representation of the file descriptor
let result = std::io::stdout().flush();

// Convert bytes to string if possible, or handle the error gracefully
if let Ok(s) = std::str::from_utf8(bytes.as_slice()) {
    // Use the appropriate standard I/O handle based on the fd value
    match *fd as i32 {
        0 => { /* stdin - not writable */ },
        1 => {
            // Standard output
            use std::io::Write;
            if let Err(e) = std::io::stdout().write_all(s.as_bytes()) {
                eprintln!("Failed to write to stdout: {}", e);
            }
            let _ = std::io::stdout().flush();
        },
        2 => {
            // Standard error
            use std::io::Write;
            if let Err(e) = std::io::stderr().write_all(s.as_bytes()) {
                eprintln!("Failed to write to stderr: {}", e);
            }
            let _ = std::io::stderr().flush();
        },
        custom_fd => {
            // For other file descriptors, we can use a more controlled approach
            // Note: This approach doesn't directly use the raw fd, but instead
            // would require proper file opening before this function is called
            eprintln!("Writing to custom file descriptor {} is not supported in the safe implementation", custom_fd);
            // A proper implementation might keep track of open files elsewhere
        }
    }
} else {
    eprintln!("Failed to interpret bytes as UTF-8");
}
```

zees-dev: It is much more code than the unsafe alternative though :/

sway-lib-std/src/debug.sw
```diff
@@ -0,0 +1,2576 @@
...
+            i -= 1;
+            value = value / 10;
+        }
+
+        syscall_write(STDERR, __addr_of(digits).add::<u8>(i), 80 - i);
+    }
+
+    pub fn debug_struct(self, name: str) -> DebugStruct {
+        self.print_str(name);
+        self.print_str(" { ");
```
zees-dev, line 127: I probably dont understand this; but don't we need closing braces for these complex types?

sway-types/src/source_engine.rs
```diff
@@ -91,8 +95,18 @@ impl SourceEngine {
...
+    /// Return the associated autogenerated pseudo file for the passed `source_id.
+    /// Example: main.autogenerated.sw for main.sw
+    pub fn get_associated_autogenerated_source_id(&self, source_id: &SourceId) -> SourceId {
+        let path = self.get_path(source_id);
+        let file_name = PathBuf::from_str(path.file_name().unwrap().to_str().unwrap()).unwrap();
+        let path = path.with_file_name(format!(
+            "{}.{}.{}",
+            file_name.file_stem().unwrap().to_str().unwrap(),
+            Self::AUTOGENERATED_PATH,
+            file_name.extension().unwrap().to_str().unwrap()
```
zees-dev, line 107: assuming these unwraps will not throw errors

docs/book/src/reference/compiler_intrinsics.md
ironcev: Actually `__dbg` will always be removed in release unless `force_dbg_in_release` is set. And for that flag, I expect it to be set only once either in project or workspace `Forc.toml` file. So the removal is actually removal of that one line in `Forc.toml` file.

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
```diff
@@ -2044,6 +2045,199 @@ fn expr_func_app_to_expression_kind(
...
+                whole_block_span: Span::dummy(),
+            };
+
+            return Ok(ExpressionKind::CodeBlock(block));
+        }
+        // ... and in release becomes "arg"
+        Some(Intrinsic::Dbg)
+            if !context.is_dbg_generation_full() && last.is_none() && !is_relative_to_root =>
+        {
+            return Ok(arguments[0].kind.clone());
```
ironcev, line 2239: Here we should also check for `arguments.len() != 1` and emit error. Otherwise, if we build empty call `__dbg()` in release the compiler will panic.

Since the two match arms differ only in `[!]context.is_dbg_generation_full()` and both need the check for arguments length, it will be simpler to have only one match arm, the check only on top of it and then the `if context.is_dbg_generation_full()` condition in the body.

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
```diff
@@ -2044,6 +2045,199 @@ fn expr_func_app_to_expression_kind(
...
+        //      let mut f = Formatter { };
+        //      f.print_str("[{current_file}:{current_line}:{current_col}] = ");
+        //      let arg = arg;
+        //      arg.fmt(f);
+        //      arg
+        // }"
+        Some(Intrinsic::Dbg)
+            if context.is_dbg_generation_full() && last.is_none() && !is_relative_to_root =>
+        {
+            if arguments.len() > 1 {
```
ironcev, line 2058: The check should actually be `arguments.len() != 1`. With this check, if we have only `__dbg()` call without arguments the compiler will panic below in the `arguments[0].clone()` call.

An interesting question is if we actually want to support empty `__dbg()` call like in Rust that will print only the file, line, and column information? :thinking: What do you think about that?

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
```diff
@@ -2044,6 +2045,199 @@ fn expr_func_app_to_expression_kind(
...
+                    })),
+                    span: span.clone(),
+                },
+                is_mutable: true,
+            });
+
+            let current_file = span
+                .source_id()
+                .map(|x| engines.se().get_path(x))
+                .unwrap_or_default();
```
ironcev, line 2100: This looks dangerous to me. What is `PathBuf::default()`? I understand that in all the `unwrap()`s in this implementation we expect things to exist. Why not only `unwrap()` here as well? It should never fail and if it does failing is the right thing compared to have some default file that is not related to the `source_id` and then failing in the `unwrap()` below.

test/src/e2e_vm_tests/test_programs/should_pass/language/intrinsics/dbg-release/Forc.lock
```diff
@@ -0,0 +1,8 @@
+[[package]]
+name = "dbg-release"
```
ironcev, line 2: Test names are following the `snake_case` convention. To stick to it and make the test same as other, it should be `dbg_release`.

test/src/reduced_std_libs.rs
```diff
@@ -1,4 +1,4 @@
-//! This module contains functions for creating reduced versions of the `std` library.
+// This module contains functions for creating reduced versions of the `std` library.
```
ironcev, line 1: This was a module doc-comment actually :smile:

ironcev (changes requested):

docs/book/src/reference/compiler_intrinsics.md
zees-dev: Ah makes sense yep; the fact escaped me that this line is stripped out entirely; so obviously there will be no gas used for this.

sway-lib-std/src/debug.sw
ironcev: The closing braces are done in the `finish` methods of the corresponding `DebugStruct/List/Tuple` structs. This is a builder pattern where `finish` does the last step including closing braces.

forc-test/src/ecal.rs
xunilrj: They are not equivalent as it only supports stdout/stderr. For all others, it needs the unsafe block.

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
xunilrj: Good catch. I am also creating one test for empty `__dbg()`.

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
xunilrj: > An interesting question is if we actually want to support empty __dbg() call like in Rust that will print only the file, line, and column information? 🤔 What do you think about that?

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
xunilrj: Because we need to support `Span::dummy()`, and in this case it is better to not print the file name than panicking the compiler.

test/src/reduced_std_libs.rs
xunilrj: Ah... that was needed because I was `include!(...)` the `reduced_std_libs` inside the snapshot tests. We don´t need that anymore.

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
xunilrj: The idea was to support `Span::dummy()`. I am treating it more explicitly now.

sway-types/src/source_engine.rs
xunilrj: That is a good point. I am making this function fallible by returning an `Option`.

forc-test/src/ecal.rs
```diff
@@ -0,0 +1,161 @@
...
+    let _ = client.transact(tx);
+
+    // Verify
+    let t: Transactor<MemoryInstance, MemoryStorage, Script, EcalSyscallHandler> = client.into();
+    let syscalls = t.interpreter().ecal_state().captured.clone();
+
+    assert_eq!(syscalls.len(), 1);
+    assert!(
+        matches!(&syscalls[0], Syscall::Write { fd: 1, bytes } if std::str::from_utf8(bytes).unwrap() == test_input)
+    );
```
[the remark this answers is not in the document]
xunilrj: I could break this into multiple assets, but not sure how this will be simpler:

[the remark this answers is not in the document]
ironcev: Yes I am here somehow also in favor of the original `assert`, although I agree it is quite a long one.

ironcev (dismissed):

forc-test/src/ecal.rs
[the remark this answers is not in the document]
zees-dev: maybe the following?
```rust
// Check the syscall type and fd
assert!(matches!(&syscalls[0], Syscall::Write { fd: 1, .. }));

// Check the syscall content
let Syscall::Write { bytes, .. } = &syscalls[0] else {
    panic!("Expected Syscall::Write");
}
assert_eq!(std::str::from_utf8(bytes).unwrap(), test_input);
```

sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
```diff
@@ -2094,26 +2094,31 @@ fn expr_func_app_to_expression_kind(
...
+
+                // find the manifest path of the current span
+                let program_id = engines
+                    .se()
+                    .get_program_id_from_manifest_path(&current_file)
+                    .unwrap();
+                let manifest_path = engines
+                    .se()
+                    .get_manifest_path_from_program_id(&program_id)
+                    .unwrap();
```
zees-dev, line 2111: Have more unwraps here

zees-dev (approved):
Great work on this PR!
This is really helpful functionality - and a much better approach IMO than the original (changing behaviour of the  `__log` intrinsic).

Once this is merged in, I will close the other: https://github.com/FuelLabs/sway/pull/6996

JoshuaBatty (changes requested):
Sorry for un-approving here but just wanted to check in with the degradation of the performance reported by codspeed. A 34% degrade in performance is pretty big. Is this unavoidable?

xunilrj:
I created this PR to alleviate the performance problems. https://github.com/FuelLabs/sway/pull/7080

xunilrj:
This is the performance status after https://github.com/FuelLabs/sway/pull/7080

Before, the `compile` bench was going from `4.9s` to `6.4s`
image 

Now is going from `3.6s` to `4.8s`
image

We still have the performance hit because we have a lot of more `impls` inside the `TraitMap` caused by all the auto `impls`, but it is faster than before.
I think we can "regain" this performance hit in future PRs.

ironcev:
> I think we can "regain" this performance hit in future PRs.

I agree.

xunilrj:
>Can we just add a feature flag to disable this debug generation altogether, and that way the LSP can just disable this feature until it's more performant?

We can, but if someone uses the `__dbg()` intrinsics, the LSP will not compile the project. Because the type will not implement `Debug`, which will confuse everyone.

xunilrj:
`__dbg` was more like an example. It will happen to any function that demands a `Debug` implementation. The code will not be compiled inside the LSP, but it will be using the CLI.
I will disable it for the moment, and we will accept this as a "temporary" inconvenience.

I will also look to see if there is anything else we can do for performance.

ironcev (changes requested):

sway-core/src/semantic_analysis/module.rs
```diff
@@ -604,7 +604,7 @@ impl ty::TyModule {
             }
 
             // Auto impl debug traits only if they are not explicitly implemented
-            if auto_impl_debug_traits {
+            if ctx.experimental.dbg_intrinsic && auto_impl_debug_traits {
```
ironcev, line 607: I have a very mixed feelings of having a feature flag that is only partially implemented here.

If we were originally for having `dbg_intrinsic` behind a feature flag that would mean:
- emitting `FeatureIsDisabled` error on `__dbg` usages.
- having the `Debug` trait and related structs (`Format`, `DebugStruct`, ...) and impls behind the flag etc.

Now we still have the penalty of always implementing script-generated `Debug` impls and adding them to the trait map even if the feature is disabled. I guess only this will still contribute to the compilation time due to huge number of types that impl `Debug`.

I understand that we just want a quick fix for the LSP, but as I said, very mixed feelings.

sway-features/src/lib.rs
```diff
@@ -173,6 +173,8 @@ features! {
     "https://github.com/FuelLabs/sway/issues/6765",
     const_generics = false,
     "https://github.com/FuelLabs/sway/issues/6860",
+    dbg_intrinsic = true,
+    "https://github.com/FuelLabs/sway/pull/7015"
```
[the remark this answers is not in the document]
ironcev: Resolved, because the flag is reverted.

sway-core/src/semantic_analysis/module.rs
ironcev: Resolved, because the feature flag is reverted.

ironcev (approved):

Voxelot:
FYI @JoshuaBatty getting the selective compilation for the dbg traits working had a lot of potential footguns, and given the optimizations that @xunilrj already did in #7080 we are thinking the performance hit from this PR may not be perceivable enough on the LSP to prevent merging it.

If you give the green light for this minor hit then we can merge this pr and finish the release.

JoshuaBatty (approved):
Thanks for looking into the performance improvements Daniel. Very much appreciated!

JoshuaBatty:
Disabling auto-merge to wait for #7093 to merge first as it's a prequel to this work.

xunilrj:
Before my optimisations/Now:
traverse:	230.9 ms/116.6 ms
code_action: 28.3 ms/8.4 ms

Voxelot (approved):

sway-core/src/semantic_analysis/ast_node/declaration/auto_impl/mod.rs
```diff
@@ -45,19 +47,23 @@ where
...
     /// The resulted [Parse] has source id set to autogenerated source id
     /// within the program represented by the `program_id`.
-    fn parse<T>(engines: &Engines, program_id: Option<ProgramId>, input: &str) -> T
+    fn parse<T>(engines: &Engines, original_source_id: Option<&SourceId>, src: &str) -> T
     where
         T: Parse,
     {
         // Uncomment this to see what is being generated
-        // println!("{}", input);
+        // println!("{}", src);
```
zees-dev, line 55: nit: it maybe worthwhile putting such useful prints/debugging behind a feature flag - so one does not need to comment out/in for PRs.
Maybe in another PR.

test/src/e2e_vm_tests/test_programs/should_fail/dbg_wrong_args_count/snapshot.toml
zees-dev, line 1: do we need this empty file?

zees-dev (approved):
Great work here; much needed functionality; LGTM 🚀

JoshuaBatty (approved):

Issue #6954, opened 2025-02-24
Rename U128.as_64 to try_as_u64 and deprecate the old function


SwayStar123:
The function name is inconsistent with other downcasts, it needs to be updated to try_as_u64, and the old function needs to be deprecated and then later removed.

anushkasomani:
I would like to give this a go. Can I be assigned on this issue

anushkasomani:
@SwayStar123 could you please provide the file directory in which i need to change. I'm unable to find it

SwayStar123:
sway-lib-std/src/u128.sw

FriendlyLifeguard:
Hello, could I work on this?

bernardev254:
Hello, this serves my intentions to take on the task of renaming U128.as_64 to try_as_u64 and deprecating the old function. This change will align the function name with other downcasts, making our codebase more consistent and easier to understand.

I'll ensure that the deprecation process is handled smoothly, updating any internal usages and providing clear guidance for external users to transition to the new function.

pheobeayo:
Let me handle this issue!

saimeunt:
Following up on my work on #7042 and #7043, I'd like to continue fixing issues on Sway, please assign me.

Zeegaths:
Could I take on this issue?

Sundayabel222:
Can I start working on this?

emarc99:
Can I give this a go? Relevant experience with Dojo, Cairo and Rust.

Sendi0011:
i will like to work on this.
ETA 24 hours

PR #7026, opened 2025-03-19
as_u64 updated

anushkasomani:
issue #6954 
deprecated as_u64 , replaced with try_as_u64 , added tests

### DIFF
--- a/sway-lib-std/src/u128.sw
+++ b/sway-lib-std/src/u128.sw
@@ -2,7 +2,7 @@
 library;
 
 use ::assert::assert;
-use ::convert::{From, Into, TryFrom};
+use ::convert::{From, Into};
 use ::flags::{
     disable_panic_on_overflow,
     panic_on_overflow_enabled,
@@ -13,9 +13,6 @@ use ::registers::{flags, overflow};
 use ::math::*;
 use ::result::Result::{self, *};
 use ::option::Option::{self, None, Some};
-use ::revert::revert;
-use ::ops::*;
-use ::codec::*;
 
 /// The 128-bit unsigned integer type.
 ///
@@ -48,7 +45,7 @@ impl From<u8> for U128 {
     fn from(val: u8) -> Self {
         Self {
             upper: 0,
-            lower: val.as_u64(),
+            lower: val.try_as_u64().expect("Conversion should never fail for u8"),
         }
     }
 }
@@ -72,7 +69,7 @@ impl From<u16> for U128 {
     fn from(val: u16) -> Self {
         Self {
             upper: 0,
-            lower: val.as_u64(),
+            lower: val.try_as_u64().expect("Conversion should never fail for u16"),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


SwayStar123:
Please remove all the unnecessary and irrelevant changes. Seems to be caused by some kind of git merge mistake.

anushkasomani:
> Please remove all the unnecessary and irrelevant changes. Seems to be caused by some kind of git merge mistake.

i made only the required changes in sway-lib-std/src/u128.sw and test/src/in_language_tests/test_programs/u128_inline_tests/src/main.sw

sway-lib-std/src/u128.sw
```diff
@@ -934,101 +925,3 @@ impl TotalOrd for U128 {
...
-    ///    let b256_value = b256::from(u128_value);
-    /// }
-    /// ```
-    fn from(num: U128) -> Self {
-        let input = (0u64, 0u64, num.upper(), num.lower());
-        asm(input: input) {
-            input: b256
-        }
-    }
-}
```
IGI-111, line 1034: Why remove all this functionality?

sway-lib-std/src/u128.sw
```diff
@@ -147,14 +144,22 @@ impl From<U128> for (u64, u64) {
...
 
-impl PartialEq for U128 {
+#[cfg(experimental_partial_eq = false)]
+impl core::ops::Eq for U128 {
     fn eq(self, other: Self) -> bool {
         self.lower == other.lower && self.upper == other.upper
     }
 }
-impl Eq for U128 {}
+#[cfg(experimental_partial_eq = true)]
```
IGI-111, line 153: This feature has been merged already, none of this is going to work.

sway-lib-std/src/u128.sw
```diff
@@ -766,14 +778,7 @@ impl Power for U128 {
...
         while exp & 1 == 0 {
             match u128_checked_mul(value, value) {
-                None => {
-                    if panic_on_overflow_enabled() {
-                        revert(0);
-                    } else {
-                        // Return zero on overflow as per the Fuel VM Specifications
-                        return U128::zero()
-                    }
-                },
```
IGI-111, line 776: You can't just delete the panic on overflow logic.

sway-lib-std/src/u128.sw
```diff
@@ -13,9 +13,6 @@ use ::registers::{flags, overflow};
 use ::math::*;
 use ::result::Result::{self, *};
 use ::option::Option::{self, None, Some};
-use ::revert::revert;
-use ::ops::*;
-use ::codec::*;
```
IGI-111, line 18: Why?

IGI-111 (changes requested):
Yeah this has a bunch of changes that have nothing to do with the nominal updating of this method. You probably are trying to merge from an older rev.

sway-lib-std/src/u128.sw
anushkasomani: actually , i was using the old file for the update. that's why these functions were not updated . check pr #7029 , its the final pr , it should work

PR #7029, opened 2025-03-22
try_as_64 updated

anushkasomani:
issue #6954


sway-lib-std/src/u128.sw
```diff
@@ -312,23 +312,32 @@ impl U128 {
     ///
     /// fn foo() {
     ///     let zero_u128 = U128::from(0, 0);
-    ///     let zero_u64 = zero_u128.as_u64().unwrap();
+    ///     let zero_u64 = zero_u128.try_as_u64().unwrap();
```
IGI-111, line 315: Instead of updating this doc-comment which still related to `as_u64` you should add an updated version to `try_as_u64`.

Also there's a TODO at the top of this comment that should be removed.

sway-lib-std/src/u128.sw
```diff
@@ -48,7 +48,7 @@ impl From<u8> for U128 {
     fn from(val: u8) -> Self {
         Self {
             upper: 0,
-            lower: val.as_u64(),
+            lower: val.try_as_u64(),
```
IGI-111, line 51: This is not the method you edited and shouldn't change.

sway-lib-std/src/u128.sw
```diff
@@ -72,7 +72,7 @@ impl From<u16> for U128 {
     fn from(val: u16) -> Self {
         Self {
             upper: 0,
-            lower: val.as_u64(),
+            lower: val.try_as_u64(),
```
IGI-111, line 75: Neither is this.

sway-lib-std/src/u128.sw
```diff
@@ -96,7 +96,7 @@ impl From<u32> for U128 {
     fn from(val: u32) -> Self {
         Self {
             upper: 0,
-            lower: val.as_u64(),
+            lower: val.try_as_u64(),
```
IGI-111, line 99: Neither is this.

IGI-111 (changes requested):

sway-lib-std/src/u128.sw
anushkasomani: you mean it should be lower: val.try_as_u64().expect("Conversion should never fail for u8") instead?

IGI-111: `val` here is a `u8` so this is calling `u8::try_as_u64` (which doesn't and needn't exist since conversions from byte to `u64` can't fail), whereas the issue is about `U128::try_as_u64`

anushkasomani: oh, so i dont need to replace as_u64 with try_as_u64 for u32::as_u64 as well right?

Issue #6903, opened 2025-02-07
Introduce deterministic resolver for alternative sources for dependencies in forc-pkg


kayagokalp:
Currently syntactically we can insert alternative sources for each dependency in the `Forc.toml` file.

```TOML
std = { ipfs = "bafybeighr56cwvhje2hxc5gtdiqgxia4slmcbcjvjdflqkmjxukvatqley", git = "https://github.com/fuellabs/sway" }
```

But `forc-pkg`'s behaviour is undefined at the moment. We should be resolving these alternatives sources in specific order for ensuring correct  behaviour for building. 

For example if a source has a path source, git source and a registry source, we should not be using the registry version. Because there might be local changes. The registry version is there to enable to be published to the registry. 

To reduce further ambiguity cargo does include the following sentence to their docs:

```
It is possible to specify both a registry version and a git or path location
```

It is clear that a user can specify a registry version and git or registry version and path. This way the disambiguity between the precedence of path and git is removed.

The summary of this propsed change is that:

1. Disallow git, path, ipfs to be declared at the same time
2. But they can be matched with registry version

kayagokalp:
this is closed by #7038

Issue #6742, opened 2024-11-19
[forc-pkg] [package-manager] Support pinning forc crates with semver


kayagokalp:
Once we have the basics of package manager sorted out, we should support pinning dependencies with name and semver version. Just like how cargo supports declaring packages with crate name and versions.

```TOML
[dependencies]
advanced_math = "1.0" // the infra for this is in, but we require the package manager to be functional first.
custom = { git = "https://github.com/FuelLabs/sway/" } // this already works
```

Yn2o:
Would it be useful to allow pinning to a specific commit hash as well, similar to git dependencies?

PR #7038, opened 2025-03-25
feat: registry source resolution and builds

### LINKED ISSUE #3752: forc: Package Hosting and Registry Design (using IPFS and the Fuel Network)

kayagokalp:
# Description
closes #3752, #6903, #6742.

This PR adds registry fetch and build capabilities to sway repo and it is highly coupled with the corresponding [forc.pub PR](https://github.com/FuelLabs/forc.pub/pull/48) that enables the indexation to happen which this PR consumes.

The registry source resolution is done through an index file which is committed to a central repository by forc.pub. This index file is fetched by forc using a common file location calculation shared by forc.pub and forc. After the index file is fetched for can understand and pin the dependency to its IPFS CID. IPFS CID fetching is already implemented in forc and this PR re-purposes much of it to actually get the dependency downloaded.

The rest of the build process is the same as other source types, namely: `git`, `ipfs` and `path`. 

## Future Work and Limitations

Here are some limitations we have as of this PR, which will be done as a follow up to this one:

1. #7060
2. #7061

## How to test this e2e

1. Accept my invitation request to https://github.com/kayagokalp/dummy-forc.pub-index from github
2. Generate a PAT, with enough permissions to be able to push code
3. Fetch `forc.pub` and come to `kayagokalp/index-file`  branch
4. `cp .env .env.local`
5.Place IPFS, GITHUB client you fetched from the vault into .env.local and also place your GITHUB_TOKEN/PAT there.
6.  run `./scripts/start_local_db.sh`  (make sure docker daemon is running)
7. go to `app/` folder and run `npm i`  then `npm start`
8. Step 7 will open a browser if not go to `localhost:3000` login with github
9. Click on your profile and generate an api key
10. Go to the package you want to publish and run `../target/debug/forc-publish --registry-url http://0.0.0.0:8080/ --token <TOKEN>`  (I published `std` so `forc-publish` location was like this from `sway-lib-std` folder)
11. Generate a new `forc` project and depend on this new dependency by version

```TOML
[project]
authors = ["kayagokalp"]
entry = "lib.sw"
license = "Apache-2.0"
name = "testing"
implicit-std = false

[dependencies]
std = "0.12.0"
```


forc-pkg/src/manifest/mod.rs
```diff
@@ -322,13 +323,27 @@ impl DependencyDetails {
...
+            bail!("Both version and git details provided for dependency");
+        }
+
+        if version.is_some() && ipfs.is_some() {
+            bail!("Both version and ipfs details provided for dependency");
+        }
+
+        if version.is_none() && namespace.is_some() {
+            bail!("Namespace can only be specified for sources with version");
+        }
```
zees-dev, line 346: should probably add tests for scenarios where these errors are hit

forc-pkg/src/source/reg/mod.rs
```diff
@@ -0,0 +1,523 @@
...
+
+impl GithubRegistryResolver {
+    /// Default github organization name that holds the registry git repo.
+    const DEFAULT_GITHUB_ORG: &str = "kayagokalp";
+    /// Default name of the repository that holds the registry git repo.
+    const DEFAULT_REPO_NAME: &str = "dummy-forc.pub-index";
+    /// Default chunking size of the repository that holds registry git repo.
+    pub const DEFAULT_CHUNKING_SIZE: usize = 2;
+    /// Default branch name for the repository repo.
+    const DEFAULT_BRANCH_NAME: &str = "master";
```
zees-dev, line 89: nit: branch defaulted by github is `"main"`

forc-pkg/src/source/reg/mod.rs
```diff
@@ -0,0 +1,523 @@
...
+        // registry+<package_name>?v<version>#<cid>!<namespace>
+        let s = s.trim();
+
+        // Check for "registry+" at the start.
+        let prefix_plus = format!("{}+", Self::PREFIX);
+        if s.find(&prefix_plus).is_some_and(|loc| loc != 0) {
+            return Err(PinnedParseError::Prefix);
+        }
+
+        let s = &s[prefix_plus.len()..];
```
zees-dev, line 189: nit: could rename; `without_prefix` or something

forc-pkg/src/source/reg/mod.rs
```diff
@@ -0,0 +1,523 @@
...
+where
+    F: FnOnce(IndexFile) -> Fut,
+    Fut: std::future::Future<Output = anyhow::Result<O>>,
+{
+    let tmp_dir = tmp_registry_package_dir(fetch_id, pkg_name, &source.version, &source.namespace);
+    if tmp_dir.exists() {
+        let _ = std::fs::remove_dir_all(&tmp_dir);
+    }
+
+    // TODO: remove the clone
```
zees-dev, line 444: needs to be addressed?

forc-pkg/src/source/reg/mod.rs
```diff
@@ -0,0 +1,523 @@
...
+    let contents = index_response.text().await?;
+    let index_file: IndexFile = serde_json::from_str(&contents).with_context(|| {
+        format!(
+            "Unable to deserialize a github registry lookup response. Body was: \"{}\"",
+            contents
+        )
+    })?;
+
+    let res = f(index_file).await?;
+    let _ = std::fs::remove_dir_all(&tmp_dir);
```
zees-dev, line 471: if any error occurs between the time the dir is created and this line; the tmp_dir will still remain (although it would get cleaned up in a future run).
Can use tmpfile; or tmpdir (`tempfile` dependency) - as that would still cleanup the dir if errors occurs in between

forc-pkg/src/source/mod.rs
```diff
@@ -145,37 +145,65 @@ pub struct DisplayCompiling<'a, T> {
...
+        }
+    }
+
+    /// Construct a source from version information collected from manifest file.
+    fn with_version_dependency(
+        pkg_name: &str,
+        version: &str,
+        namespace: &reg::file_location::Namespace,
+    ) -> Result<Self> {
+        // TODO: semver parsing does not work for versions like: 0.1 or ^0.1.
```
zees-dev, line 175: interesting; i wonder if we can ensure upstream that versions should always be semver compliant

forc-pkg/src/source/reg/mod.rs
kayagokalp: Default branch name we use in fuel labs repos are master, as far as i can see. 

Also this default is for pointing to the forc.pub-index repo which uses master.Although for now it uses my dummy repo for testing

forc-pkg/src/source/mod.rs
kayagokalp: At first i thought there was no out of box option for this but in a recent discussion, we found out that there is a [VersionReq](https://docs.rs/semver/1.0.26/semver/struct.VersionReq.html) I will switch to that in the follow up PR which will also add the logic for handling the caret and wildcard type of declarations. So i am keeping the parsing step like this as the logic to select correct version is still going to come in the next PR.

docs/book/src/forc/dependencies.md
```diff
@@ -33,6 +33,21 @@ The following example adds a dependency with an `ipfs` source.
 custom_lib = { ipfs = "QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG" }
 ```
 
+For `registry` sources, `forc` will first resolve the source declared by its name and version. This is done using the forc.pub-index repo (located at `https://github.com/FuelLabs/forc.pub-index`). The package name and version is used to convert the declaration to an IPFS CID. The resolved IPFS CID is then used by either a local IPFS node or a public gateway to actually fetch the package.
```
sdankel, line 36: > The resolved IPFS CID is then used by either a local IPFS node or a public gateway to actually fetch the package.

It should be using only the forc-pub IPFS gateway, not local or public gateways.

forc-pkg/src/manifest/mod.rs
sdankel, line 346: We can do this in the command definition using clap's `conflicts_with`

forc-pkg/src/source/reg/mod.rs
kayagokalp: That is a good catch, i recall you or someone else made the same comment earlier in different fetch function for sources.

I will address all of them together (if you see `git` and `ipfs` sources they also have this happening): I opened #7049, I will add this one too before merging that PR.

forc-pkg/src/manifest/mod.rs
kayagokalp: oh nice one! I will take a look at that

docs/book/src/forc/dependencies.md
kayagokalp: Hmm I agree that by default we should use forc-pub gateway instead of the public gateway we are using right now. 

But it should be still configurable as it is right now in my opinion. Because If we are forcing people to get it from forc-pub IPFS node only, than why we keep packages in IPFS in the first place? I think that makes it equivalent to keeping everything in a S3 only with extra complexity.

forc-pkg/src/source/reg/mod.rs
kayagokalp: I was thinking to but upon looking at this in more detail, I think i will leave it as it is because it is a single string clone and to remove it we need to change the Source declaration to have a lifetime param and that will effect all the possible source types whcih will be a fairly large change to remove single string clone. We can do that if we ever end up without anything to do :)

forc-pkg/src/source/reg/file_location.rs
```diff
@@ -0,0 +1,188 @@
...
+
+/// Calculates the exact file location from the root of the namespace repo.
+/// If the configuration includes a namespace, it will be the first part of
+/// the path followed by chunks.
+pub fn location_from_root(chunk_size: usize, namespace: &Namespace, name: &str) -> PathBuf {
+    let mut path = PathBuf::new();
+
+    // Add domain to path if namespace is 'Domain'
+    // otherwise skip.
+    if let Namespace::Domain(domain) = namespace {
```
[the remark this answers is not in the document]
kayagokalp: nice one!

forc-pkg/src/manifest/mod.rs
kayagokalp: I came to this comment to apply but realized these values are coming from a TOML file not CLI so not so sure about how to use clap's conflict with groups to enforce this

docs/book/src/forc/dependencies.md
kayagokalp: I made it so that if the user is trying to rely on a public gateway we override it with fuel owned one. But if they are relying on a local node explicitly, we keep that as it is

kayagokalp:
I think I addresed all of the comments besides one which i left a comment as I did not understand it fully. I think we should be ready for a re-review

forc-pkg/src/source/reg/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+    if tmp_dir.exists() {
+        let _ = std::fs::remove_dir_all(&tmp_dir);
+    }
+
+    let github_resolver = GithubRegistryResolver::with_default_github(source.namespace.clone());
+
+    let path = format!(
+        "{}",
+        location_from_root(github_resolver.chunk_size, &source.namespace, pkg_name).display()
+    );
```
JoshuaBatty, line 458:
```suggestion
    let path = location_from_root(github_resolver.chunk_size, &source.namespace, pkg_name).display().to_string();
```

forc-pkg/src/source/mod.rs
JoshuaBatty: Can we leave a link to the github issue for this here?

forc-pkg/src/source/mod.rs
```diff
@@ -60,6 +60,16 @@ impl Default for IPFSNode {
     }
 }
 
+impl IPFSNode {
+    pub fn fuel() -> Self {
```
JoshuaBatty, line 64:
```
/// Returns an IPFSNode configured to use the Fuel-operated IPFS gateway.
```

forc-pkg/src/source/mod.rs
```diff
@@ -60,6 +60,16 @@ impl Default for IPFSNode {
...
+        Self::WithUrl(DEFAULT_REGISTRY_IPFS_GATEWAY_URL.to_string())
+    }
+
```
JoshuaBatty, line 67:
```
/// Returns an IPFSNode configured to use the public IPFS gateway.
```

forc-pkg/src/source/reg/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+    let index_repo_owner = github_resolver.repo_org();
+    let index_repo_name = github_resolver.repo_name();
+    let reference = format!("refs/heads/{}", github_resolver.branch_name());
+    let github_endpoint = format!(
+        "https://raw.githubusercontent.com/{index_repo_owner}/{index_repo_name}/{reference}/{path}"
+    );
+    let client = reqwest::Client::new();
```
JoshuaBatty, line 465: should we add a timeout here to prevent it hanging?

kayagokalp: Sure, done in [a834871](https://github.com/FuelLabs/sway/pull/7038/commits/a834871f5a097a293ea5bc053229d0ab8eccf566)

forc-pkg/src/source/mod.rs
kayagokalp: done in [a834871](https://github.com/FuelLabs/sway/pull/7038/commits/a834871f5a097a293ea5bc053229d0ab8eccf566)

forc-pkg/src/source/mod.rs
kayagokalp: done in [a834871](https://github.com/FuelLabs/sway/pull/7038/commits/a834871f5a097a293ea5bc053229d0ab8eccf566)

forc-pkg/src/source/reg/mod.rs
kayagokalp: done in [a834871](https://github.com/FuelLabs/sway/pull/7038/commits/a834871f5a097a293ea5bc053229d0ab8eccf566)

forc-pkg/src/source/mod.rs
kayagokalp: Done ine [a834871](https://github.com/FuelLabs/sway/pull/7038/commits/a834871f5a097a293ea5bc053229d0ab8eccf566)

forc-pkg/src/source/reg/mod.rs
```diff
@@ -0,0 +1,545 @@
...
+        };
+        Ok((pinned, path))
+    }
+}
+
+impl source::Fetch for Pinned {
+    fn fetch(&self, ctx: source::PinCtx, path: &Path) -> anyhow::Result<PackageManifestFile> {
+        // Co-ordinate access to the registry checkout directory using an advisory file lock.
+        let mut lock = forc_util::path_lock(path)?;
+        // TODO: Here we assume that if the local path already exists, that it contains the
```
JoshuaBatty, line 343: Might be useful to open an issue for this and link it here.

JoshuaBatty:
Nice work, i'm pretty happy with where this is at. Happy to approve once the locations are updated from your dummy repo to the fuel labs registry repo.

kayagokalp:
PR is now pointing to the official fuel labs repos, and can be merged as approved @FuelLabs/tooling

sdankel (approved):
LGTM! 🚀

forc-pkg/src/source/reg/index_file.rs
```diff
@@ -0,0 +1,391 @@
...
+    source_cid: String,
+    /// IPFS CID of this specific package's abi. This is pinned by
+    /// forc.pub at the time of package publishing and thus will be
+    /// available all the time if this exists in the first place, i.e the
+    /// package is a contract.
+    abi_cid: Option<String>,
+    /// Dependencies of the current package entry. Can be consumed to enable
+    /// parallel fetching by the consumers of this index, mainly forc.
+    dependencies: Vec<PackageDependencyIdentifier>,
+}
```
sdankel, line 61: nit: should we also add `yanked: bool` here?

JoshuaBatty: Yeah that's a good idea. Happy for this to be in a follow up PR

JoshuaBatty (approved):

forc-pkg/src/source/reg/index_file.rs
kayagokalp: Merging this one not to dismiss reviews adding it in the follow-up right away, thanks for catching this :) #7077

forc-pkg/src/source/reg/mod.rs
```diff
@@ -0,0 +1,546 @@
...
+        )
+    }
+}
+
+impl FromStr for Pinned {
+    type Err = PinnedParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // registry+<package_name>?v<version>#<cid>!<namespace>
+        let s = s.trim();
```
zees-dev, line 185: nit: string length check?

PR #7042, opened 2025-03-26
Fix #6727

### LINKED ISSUE #6727: Sway hex literals have their suffix counted as a digit
### Related Component

compiler

### Problem

As the title states, hex literals have their optional suffix counted as part of their digits. This causes the following error, and related errors for types aside from b256:

```
  --> /Users/camden/Source/charcoal/out/ds_auth/src/main.sw:69:75
   |
67 |     } else if src == storage.owner.read() {
68 |         return true;
69 |     } else if storage.authority.read() == Identity::Address(Address::from(0x0000000000000000000000000000000000000000000000000000000000000000b256)) {
   |                                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ hex literals must have 1..16 or 64 digits
70 |         return false;
71 |     } else {
```

If we count the zeroes, logically we can deduce that there is in fact exactly 64 digits in the following:
```
0x0000000000000000000000000000000000000000000000000000000000000000b256
```

But there is no attempt made to strip the `b256` before counting.

### Steps

1. Create a sway project.
2. Create a b256 variable using the `0x0000000000000000000000000000000000000000000000000000000000000000b256` constant.
3. Run `forc build`

### Possible Solution(s)

Maybe strip the prefix before this point:
https://github.com/FuelLabs/sway/blob/73e1a7a80301d341864ea75de4124bf98d52a3ce/sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs#L3388-L3402

### Notes

Should be an easy fix, but it should be done ASAP.

### Installed components

```console
camden@Camdens-MacBook-Air sway % fuelup check
latest - Up to date
fuelup - Up to date
camden@Camdens-MacBook-Air sway % fuelup show
Default host: aarch64-apple-darwin
fuelup home: /Users/camden/.fuelup

Installed toolchains
--------------------
latest-aarch64-apple-darwin (default)

active toolchain
----------------
latest-aarch64-apple-darwin (default)
  forc : 0.66.4
    - forc-client
      - forc-deploy : 0.66.4
      - forc-run : 0.66.4
    - forc-crypto : 0.66.4
    - forc-debug : 0.66.4
    - forc-doc : 0.66.4
    - forc-fmt : 0.66.4
    - forc-lsp : 0.66.4
    - forc-tx : 0.66.4
    - forc-wallet : 0.11.1
  fuel-core : 0.40.0
  fuel-core-keygen : 0.40.0

fuels versions
--------------
forc : 0.66.9
forc-wallet : 0.66.9
```

saimeunt:
## Description

This PR fixes a bug where hex literals have their optional suffix counted as part of their digits.
The root cause of the bug it that the lexer wrongfully interpret the suffix as additional hex digits as "0xb256" happens to be a valid hex string.
This won't occur with other numeric literal constants such as "0x0u256" as the lexer will stop upon encountering "u" which is not a valid hex digit, the lexer will then extract the type suffix to help decoding the numeric literal which the suffix is stripped from before being further processed.
As far as binary b256 literals are concerned, the suffixed version will correctly stop interpreting characters as bits until the suffix is found, but it will fail with a lexer "InvalidIntSuffix" error because b256 is an unknown suffix in the first place (see https://github.com/FuelLabs/sway/blob/master/sway-parse/src/token.rs#L750-L764).
The suffix-free version is not impacted and correctly compiles (already was before this fix).

The initial proposed fix is to patch the literal_to_literal function which is responsible for transforming a sway_ast literal into a sway language literal: the suffix is removed from the hex literal only if it's preceded by 64 hex digits to form a correct b256 literal. The hex digits are parsed again because we can't use the parsed BigUint provided by the lexer as it may include the optional suffix wrongfully interpreted as "0xb256".

This fix feels like a band-aid and a deeper fix should be implemented at the lexer level to properly parse b256 literals.

Here's a test summarizing the bug and validating the fix:

```sway
#[test]
fn test_b256_literal_suffix() {
    // 64 zeros followed by the b256 suffix
    // the lexer will wrongfully interpret this as a 68 long hex string with the suffix adding 2 extra bytes
    // in the literal_to_literal function we have to strip the suffix and reparse it to get the correct b256
    let foo = 0x0000000000000000000000000000000000000000000000000000000000000000b256;
    assert(foo == b256::zero());
    // 60 zeroes followed by 4 hex digits "b256"
    // the trailing b256 is not a suffix but the last 2 bytes of an hex literal ending with "0xb256": it must be 
    // preserved
    let bar = 0x000000000000000000000000000000000000000000000000000000000000b256;
    assert(bar.as_u256() == 0xb256);
    // this will not compile at the moment because the lexer will thow an InvalidIntSuffix error as b256 is not a
    // known suffix
    // let foo_bin = 0b0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000b256;
    // assert(foo_bin == b256::zero());
    // this is parsed correctly
    let bar_bin = 0b0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001011001001010110;
    assert(bar_bin.as_u256() == 0xb256);
}
```

Closes #6727 

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7926,6 +7926,7 @@ dependencies = [
  "indexmap 2.7.1",
  "itertools 0.13.0",
  "lazy_static",
+ "num-bigint",
  "object",
  "parking_lot",
  "paste",
--- a/sway-core/Cargo.toml
+++ b/sway-core/Cargo.toml
@@ -25,6 +25,7 @@ im.workspace = true
 indexmap = { workspace = true, features = ["serde"] }
 itertools.workspace = true
 lazy_static.workspace = true
+num-bigint.workspace = true
 object = { workspace = true, features = ["write"] }
 parking_lot.workspace = true
 paste.workspace = true
--- a/sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
+++ b/sway-core/src/transform/to_parsed_lang/convert_parse_tree.rs
@@ -13,6 +13,7 @@ use crate::{
 };
 use either::Either;
 use itertools::Itertools;
+use num_bigint::BigUint;
 use sway_ast::{
     assignable::ElementAccess,
     expr::{LoopControlFlow, ReassignmentOp, ReassignmentOpVariant},
@@ -3528,12 +3529,27 @@ fn literal_to_literal(
             } = lit_int;
             match ty_opt {
                 None => {
-                    let orig_str = span.as_str();
-                    if let Some(hex_digits) = orig_str.strip_prefix("0x") {
-                        let num_digits = hex_digits.chars().filter(|c| *c != '_').count();
+                    let mut stripped_str = span.as_str();
+                    if let Some(stripped) = stripped_str.strip_suffix("b256") {
+                        for (prefix, required_digits) in [("0x", 64), ("0b", 256)] {
+                            if let Some(digits) = stripped.strip_prefix(prefix) {
+                                let num_digits = digits.chars().filter(|c| *c != '_').count();
+                                if num_digits == required_digits {
+                                    stripped_str = stripped;
+                                    break;
+                                }
+                            }
+                        }
+                    }
+                    if let Some(hex_digits) = stripped_str.strip_prefix("0x") {
+                        let hex_digits_filtered = hex_digits.replace("_", "");
+                        let num_digits = hex_digits_filtered.len();
                         match num_digits {
                             1..=16 => Literal::Numeric(u64::try_from(parsed).unwrap()),
                             64 => {
+                                let parsed =
+                                    BigUint::parse_bytes(hex_digits_filtered.as_bytes(), 16)
+                                        .unwrap();
                                 let bytes = parsed.to_bytes_be();
                                 let mut full_bytes = [0u8; 32];
                                 full_bytes[(32 - bytes.len())..].copy_from_slice(&bytes);
@@ -3544,11 +3560,15 @@ fn literal_to_literal(
                                 return Err(handler.emit_err(error.into()));
                             }
                         }
-                    } else if let Some(bin_digits) = orig_str.strip_prefix("0b") {
-                        let num_digits = bin_digits.chars().filter(|c| *c != '_').count();
+                    } else if let Some(bin_digits) = stripped_str.strip_prefix("0b") {
+                        let bin_digits_filtered = bin_digits.replace("_", "");
+                        let num_digits = bin_digits_filtered.len();
                         match num_digits {
                             1..=64 => Literal::Numeric(u64::try_from(parsed).unwrap()),
                             256 => {
+                                let parsed =
+                                    BigUint::parse_bytes(bin_digits_filtered.as_bytes(), 2)
+                                        .unwrap();
                                 let bytes = parsed.to_bytes_be();
                                 let mut full_bytes = [0u8; 32];
                                 full_bytes[(32 - bytes.len())..].copy_from_slice(&bytes);


IGI-111:
That test should be added to the codebase.

IGI-111 (changes requested):
The whole logic seems fine but this is breaking separation of concerns.

This kind of handling should happen in `sway-parse` if at all possible not in the parse tree conversion.

saimeunt:
@IGI-111 Thanks for your review, it helped me getting in the right direction.

I removed my fix from the `literal_to_literal` function and fixed the parser to correctly handle b256 literals having their optional suffix set.

The fix in the parser is to introduce a new variant of the `LitIntType` to support the b256 prefix.
https://github.com/FuelLabs/sway/pull/7042/files#diff-2f9f0abeb9bc800234c4324920a706eb238b46e0e33de7752f64a0f9f9360d8cR101

Then in the parge_digits function we introduce a limit over the maximum digits that should be parsed (eg. no more than 64 hex digits should be parsed when radix is 16) so we don't end up mistaking the b256 prefix for overflowing hex data.
https://github.com/FuelLabs/sway/pull/7042/files#diff-2f9f0abeb9bc800234c4324920a706eb238b46e0e33de7752f64a0f9f9360d8cR101

I have added a test validating this fix and made sure all other tests are passing.

sway-ast/src/literal.rs
```diff
@@ -42,6 +42,7 @@ pub enum LitIntType {
     I16,
     I32,
     I64,
+    B256,
```
IGI-111, line 45: There is some extra logic right above this with a boolean for b256 values that ought to be pulled out now that `b256` no longer has its `LitInt::ty_opt` set to `None`.

We should probably consider removing the `Option` if there are no valid uses of the `ty_opt` field that don't have a valid `LitIntType` but it's likely we still want to be able to parse literals that are mistyped (for error handling purposes).

IGI-111 (changes requested):
Appreciate the work on this, but considering the logic further here, using the existing integer facilities of the parser may not actually be the best way to get good behavior here.

What the original issue complains about is that the following:
`0x0000000000000000000000000000000000000000000000000000000000000000b256`
is not accepted despite being unambiguous (since it's 64 bytes followed by "b256").

However, just accepting any integer literal with a hex suffix isn't acceptable because this:

`0x01b256` is inherently ambiguous, the compiler can't know if you intended a shortform hex literal for `00 b2 56` or if you meant `01` as a `b256` type. Given the critical uses of `b256` types in Sway logic, that's unacceptable.

Hence the original introduction of the rule that all `b256` literals should be longform, since that makes them unambiguous.

I think what is probably best here to keep this property and change the special handling of the parser for b256 types to strip a potential, unambiguous, suffix; as was suggested in the original issue.

You're on the right track in the sense that you're looking at the right parts of the code, but this still needs some changes.

test/src/e2e_vm_tests/test_programs/should_pass/language/b256_literals/src/main.sw
```diff
@@ -0,0 +1,19 @@
+script;
+
```
IGI-111, line 2: One more thing is that this test should have more diverse sets of values to make sure that things like endianess works correctly.

PR #7045, opened 2025-03-29
Add debug_tuple macro to forc-util

KevinMomanyi:
## Description

I have automated the tuplet in forc-fmt thats takes too long on long method call chains


kayagokalp:
This does not make sense, in terms of what is being implemented also why as well. Possibly generated automatically using an LLM.

PR #7046, opened 2025-03-30
fix: prevent RVRT opcode for predicates

### LINKED ISSUE #5986: Statically forbid usage of revert in predicates
Using reverts in predicates produces a panic as they are illegal. Using them in that context should produce an error at compile time.

FriendlyLifeguard:
## Description

Closes #5986 

## Checklist

### DIFF
--- a/test/src/e2e_vm_tests/test_programs/should_fail/predicate_invalid_opcodes/src/main.sw
+++ b/test/src/e2e_vm_tests/test_programs/should_fail/predicate_invalid_opcodes/src/main.sw
@@ -75,6 +75,9 @@ fn main() -> bool {
     smo r1 r2 r3 r4;
   }
 
+  // cannot test rvrt op code because control flow instructions
+  // are not allowed in assembly blocks
+
   // cannot test storage opcodes due to needing to annotate main
   // with #[storage(read, write)] which is not allowed for predicates
   /*


IGI-111 (changes requested):
You can't just remove the panic and assume everything will be fine, the point of the issue is that we should check that RVRT isn't introduced earlier in the compilation process so that the error is better than a mere panic. It's not to remove safety checks that should remain there.

If we merged this, the compiler would start producing illegal binaries, which is unacceptable.

FriendlyLifeguard:
@IGI-111 I found a workaround to test `RVRT` opcode without using its opcode and instead use `revert`. The test does pass and it still detects the `rvrt` opcode.

xunilrj:
I am not so sure this is the right way here.

I think that most uses of `revert` in predicates come from `unwrap()`, for both `Option<T>`, `Result<T, E>`.
Without these, one would need to use `match` on all options and results, and somehow propagate early terminations, which would kill their ergonomy.

Maybe we need a way to generate "early termination" of predicates instead of generating "reverts", for example, maybe we should generate `RET false` or `RETD false`. 

That way we would keep the current semantics of `ùnwrap` a None would fail the predicate; and we would keep the easy ergonomy of simple early terminations.

IGI-111:
Closing this since it doesn't look like we're going to come to a resolution of this issue with this code.

FriendlyLifeguard:
Hi @IGI-111 and @xunilrj bare with me as this is my first time contributing to Fuel. I thought of a possible solution of tinkering with the compiler code to detect predicate only programs and for every unwrap function that results in a None value, it would output `ret false` instead of normal `revert opcode` output. I'm currently navigating throughout the codebase to see where I can implement this. Could I be reassigned to work on this issue?

PR #7049, opened 2025-03-31
forc-pkg: add guards to ensure proper fetch cleanups

### LINKED ISSUE #7048: forc-pkg: tmp package clean-up during source fetch might fail and leave artifacts behind
Raised by @zees-dev, currently we are manually cleaning the temporary fetch file/folders and any failure in intermediate steps might leave fetch directory dirty.

This is related to:

1. Ipfs source fetchs
2. GIt source fetchs
3. And once #7038 is merged it also effects registry sources.

kayagokalp:
## Description
closes #7048.

Adds a `scopeguard` to ensure cleanup happens while cleaning up lookup tmp files in git sources. Basically whether we return normally or panic due to an error being unwrapped in the user provided function, we do ensure that clean-up is called removing the possiblity to have dormant fetch artficats in between runs.

This will also cover registry sources once #7038 is merged I will add the fix to that place as well, so this PR should be left in draft until then.

The reason we are not using the `tempfile` is simply the location constraint. As far as i can see `tempfile` enforces the folders to be created at `/tmp` but forc has some or can have some logic to re-use this pins. So this way we keep everything in the folder they were already (`~/.forc`) and keep this change as small as possible (otherwise we should probably clean old dir to ensure people coming from old versions do not have unused stuff because of us)


Cargo.toml
```diff
@@ -211,6 +211,7 @@ ropey = "1.5"
 rpassword = "7.2"
 rustc-hash = "1.1"
 rustyline = "15.0"
+scopeguard = "1.2.0"
```
sdankel, line 214: nit

```suggestion
scopeguard = "1.2"
```

sdankel (approved):

zees-dev (approved):

JoshuaBatty (approved):

PR #7054, opened 2025-04-01
Enforce type annotations for constants

### LINKED ISSUE #5758: Enforce type annotations for constants
We should ensure all `const` declarations have explicitly named types and provide a helpful error message when they do not.

```sway
const OFFSET_TOKEN_ADDRESS = 32; // forbidden
const OFFSET_TOKEN_ADDRESS: u64 = 32; // allowed
```

saimeunt:
## Description

This PR introduces a breaking change that requires all const declarations to have a proper type annotation.

This has been implemented by throwing a new parser error when detecting a constant declaration that lacks proper type annotation, and changing the `ty_opt` member of `ItemConst` to a non-optional type as it's no longer possible to omit it.

The `ConvertParseTreeError::ConstantRequiresTypeAscription` has been removed as it's no longer allowed to omit the type of a const and relying on the initialization expression to deduce it.

A new test has been added to make sure the parser errors when a const type is not explicitly stated.

The test raising the `ConstantRequiresTypeAscription` error has been modified to check for uninitialized associated const declarations instead.

A regexp has been used to find all occurrences of incomplete const declarations and try to patch them all.

Some other parts of Sway may be affected such as eg. sway-by-example: https://github.com/FuelLabs/sway-by-example-lib/blob/main/examples/constants/src/main.sw#L7

Closes #5758 

## Checklist


IGI-111 (commented):
This all looks great and probably would work, but since it's a breaking change, it needs to be behind a feature flag.

Instead of doing that, which is a pretty significant amount of work especially if you're not familiar with the infra for those, I think a workable alternative so we can get this merged is to maintain the current behavior but reliably produce a warning (not an error) when a type is missing.

We'll easily be able to turn the warning into an error at a later date.

saimeunt:
> This all looks great and probably would work, but since it's a breaking change, it needs to be behind a feature flag.
> 
> Instead of doing that, which is a pretty significant amount of work especially if you're not familiar with the infra for those, I think a workable alternative so we can get this merged is to maintain the current behavior but reliably produce a warning (not an error) when a type is missing.
> 
> We'll easily be able to turn the warning into an error at a later date.

@IGI-111 Thank you for your review and your guidelines, I've slightly rework the PR to change the previously introduced parser error into a warning.

Issue #5955, opened 2024-05-02
Create `#[abi_name(name)]` annotation


IGI-111:
Add support for a new annotation that allows you to change the name of a struct or enum in the type definitions that gets hashed as an event hash and a type id in the ABI spec.

PR #7057, opened 2025-04-02
Add `#[abi_name(name = "foo")]` attribute to rename ABI items.

tritao:
## Description

Add `#[abi_name(name = "foo")]` attribute to rename enum and struct ABI items.

Here is example of how it can be used:

```sway
contract;

#[abi_name(name = "RenamedMyStruct")]
struct MyStruct {}

#[abi_name(name = "RenamedMyEnum")]
enum MyEnum {
  A: ()
}

abi MyAbi {
    fn my_struct() -> MyStruct;
    fn my_enum() -> MyEnum;
}

impl MyAbi for Contract {
  fn my_struct() -> MyStruct { MyStruct{} }
  fn my_enum() -> MyEnum { MyEnum::A }
}
```

Closes https://github.com/FuelLabs/sway/issues/5955.

## Remarks

The checking for ABI names is being done in the ABI generation phase, which made it a requirement to change the interface of ABI checking to return Results everywhere and passing Handler all the way down, so emitted errors can be passed down to the driver for reporting.

## Checklist


IGI-111 (approved):

sway-core/src/abi_generation/abi_str.rs
```diff
@@ -208,6 +211,19 @@ impl TypeInfo {
...
 
+fn handle_abi_name_attribute(call_path: &CallPath, attributes: &transform::Attributes) -> CallPath {
+    let mut abi_call_path = call_path.clone();
+    if let Some(abi_name_attr) = attributes.abi_name() {
+        let name = abi_name_attr.args.first().unwrap();
+        let literal = match name.value.as_ref().unwrap() {
+            Literal::String(lit_string) => lit_string,
+            _ => unreachable!(),
+        };
+        abi_call_path.suffix = BaseIdent::new(literal.span.clone().trim_quotes());
```
ironcev, line 222: A general question to renaming ABI elements. What if we rename it to something that already exist and is also serialized into ABI? Shouldn't we provide any checks here?

sway-types/src/span.rs
```diff
@@ -188,6 +188,27 @@ impl Span {
         }
     }
 
+    pub fn trim_quotes(self) -> Span {
```
ironcev, line 191: This will not be needed if the provided infrastructure is used (`AttributeArg::get_string`).

ironcev (changes requested):

ironcev (changes requested):
The attribute should be added to the _Attributes_ chapter in the documentation.

I've changed that documentation page fully in #7058 because it was seriously outdated. So, whoever gets the PR merged later will have some slight merge conflicts to solve on that page :smile:

sway-core/src/abi_generation/abi_str.rs
```diff
@@ -208,6 +211,19 @@ impl TypeInfo {
     }
 }
 
+fn handle_abi_name_attribute(call_path: &CallPath, attributes: &transform::Attributes) -> CallPath {
```
ironcev, line 214: Maybe `get_abi_call_path`? Sounds to me more like the intention, whereas the "handling of attributes" is an implementation detail.

test/src/e2e_vm_tests/test_programs/should_pass/language/abi_name_attribute/.gitignore
```diff
@@ -0,0 +1,2 @@
+out
```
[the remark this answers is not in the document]
ironcev: Lint finally added in #7059.

sway-core/src/abi_generation/abi_str.rs
ironcev: I gave this a bit of a thought. Giving names is a serious business. Aside from renaming that can clash with an existing item what about invalid names? E.g.
- `name = ""`
- `name = "this !s n0t an identif1er"`
- `name = "__this_also_not"`
- `name = "this::looks::like::a::path"`
- `name = "will this \" break ABI generation or be encoded?"`

What will SDKs do in these cases? How do they handle invalid names, if at all? I wouldn't be surprised if expecting a valid name is a part of an unwritten contract for ABI consistency.

I expect strong checks here for consistency. Honestly, I would go that far and give warnings for non-idiomatic names.

sway-types/src/span.rs
ironcev: Having a `get_string_span` is a fine addition, but in this case, we just need an `Ident` so simply using `Ident::new_no_span(name)` seams like the simplest option. Note that neither the span created with `new_no_span` nor the one with the span from the `name` argument will have the same `src` like the original name and the rest of the prefixes. That's why I would go for the simplest version here and just use `new_no_span`.

ironcev: > With the Span::trim_quotes solution, aren't we keeping the src accurate to the original

Yes we do, but that original span is not used anywhere later on. The ABI generation is looking only at the value in the `Ident` and we will never have, e.g., an error consuming the final, changed `CallPath`. Actually, having a consumer that would try to get the span of the whole path anyhow wouldn't work, because the prefixes and the suffix are in different files.

So for this particular case I agree, `new_no_span` is the simplest option.

xunilrj (dismissed):

docs/book/src/reference/attributes.md
```diff
@@ -102,3 +103,34 @@ The `#[test]` attribute marks a function to be executed as a test.
 The `#[test(should_revert)]` attribute marks a function to be executed as a test that should revert.
 
 More details in [Unit Testing](../testing/unit-testing.md).
+
+## ABI Name
+
+The `#[abi_name]` attribute allows to specify the ABI name for an item.
+This means that when a contract ABI JSON file is generated, the name that is output is the one specified
+by the attribute. This can be useful to allow renaming items, while allowing for keeping backwards
+compatibility at the contract ABI level.
```
IGI-111, line 112: I think this needs a larger example. With at least concrete mention of how this changes the ABI output.

docs/book/src/reference/attributes.md
```diff
@@ -15,6 +16,62 @@ Below is the list of attributes supported by the Sway compiler, ordered alphabet
...
+      "concreteTypeId": "d31db280ac133d726851d8003bd2f06ec2d3fc76a46f1007d13914088fbd0791",
+      "type": "struct MyStruct"
+    }
+  ],
+  ...
+}
+```
+
+We get the same JSON ABI output both before and after renaming the types, due to attributing them with
+`#[abi_name(name = ...)]`, which forces them to be generated with the previous Sway name.
```
ironcev, line 72:
```suggestion
`#[abi_name(name = ...)]`, which forces them to be generated with their previous Sway names.
```

sway-core/src/abi_generation/fuel_abi.rs
```diff
@@ -33,6 +38,42 @@ impl AbiContext<'_> {
...
 impl TypeId {
+    fn get_abi_name_and_span_from_type_id(
+        engines: &Engines,
+        type_id: TypeId,
+    ) -> Result<Option<(String, Span)>, ErrorEmitted> {
+        let handler = Handler::default();
+        match *engines.te().get(type_id) {
+            TypeInfo::Enum(decl_id) => {
+                let enum_decl = engines.de().get_enum(&decl_id);
+                if let Some(abi_name_attr) = enum_decl.attributes.abi_name() {
```
ironcev, line 49: Why not `match` on `Option` here instead of `if else`?

sway-core/src/abi_generation/fuel_abi.rs
```diff
@@ -33,6 +38,42 @@ impl AbiContext<'_> {
...
+                        .unwrap()
+                        .get_string(&handler, abi_name_attr)?;
+                    Ok(Some((name.clone(), abi_name_attr.span.clone())))
+                } else {
+                    Ok(None)
+                }
+            }
+            TypeInfo::Struct(decl_id) => {
+                let struct_decl = engines.de().get_struct(&decl_id);
+                if let Some(abi_name_attr) = struct_decl.attributes.abi_name() {
```
ironcev, line 62: Same here.

test/src/e2e_vm_tests/test_programs/should_pass/language/attributes_abi_name/Forc.toml
```diff
@@ -0,0 +1,9 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "lib.sw"
+license = "Apache-2.0"
+name = "attributes_abi_name"
+
+[dependencies]
+std = { path = "../../../../../../../sway-lib-std" }
```
ironcev, line 8: We don't need the full-blown `std` here. We can use the reduced `sway-lib-std-core`.

test/src/e2e_vm_tests/test_programs/should_fail/attributes_invalid_abi_names/stdout.snap
```diff
@@ -0,0 +1,63 @@
...
+    Building test/src/e2e_vm_tests/test_programs/should_fail/attributes_invalid_abi_names
+   Compiling library std (sway-lib-std)
+   Compiling contract attributes_invalid_abi_names (test/src/e2e_vm_tests/test_programs/should_fail/attributes_invalid_abi_names)
+error
+ --> test/src/e2e_vm_tests/test_programs/should_fail/attributes_invalid_abi_names/src/lib.sw:6:1
+  |
+4 | struct MyStruct {}
+5 | 
+6 | #[abi_name(name = "SameName")]
+  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Duplicated name found for renamed ABI type.
```
ironcev, line 16: We can heavily benefit here from expressive diagnostics, by pointing to the source of the other name. Because this name can be a name of the type which is not in the same file, or even of an another one `#[abi_name]` that is also not in the same file.

The proposal is:
- to use expressive diagnostics to point to the other name.
- to extend the test and cover the cases where the duplicated name is a name of another type, or another `#[abi_name]` in the current or some other file.

Another suggestion is to underline only the name itself, "SameName" in this example, because that's where the issue actually is. The rest of the attribute usage is perfectly fine and underlining it gives a misleading impression that the usage of the attribute is wrong.

test/src/e2e_vm_tests/test_programs/should_fail/attributes_invalid_abi_names/stdout.snap
```diff
@@ -0,0 +1,63 @@
...
+11 | 
+   |
+____
+
+error
+  --> test/src/e2e_vm_tests/test_programs/should_fail/attributes_invalid_abi_names/src/lib.sw:12:1
+   |
+10 | struct MyStruct2 {}
+11 | 
+12 | #[abi_name(name = "")]
```
ironcev, line 41: Here also, underlining only the "" and not the whole attribute better points to the actual problem.

Personally, I would have expressing diagnostics here as well, explaining that the name must be a valid Sway identifier, and in this case with addition 'and cannot be empty'.

tritao:
The PR has been updated with a significantly more robust approach for name checking, it now can find naming conflicts for more cases, namely existing types previously defined in the ABI interface surface, both in the main and imported modules. The diagnostics have also been improved, with support for expressive diagnostics and more detailed error messages.

IGI-111 (dismissed):

docs/book/src/reference/attributes.md
```diff
@@ -15,6 +16,62 @@ Below is the list of attributes supported by the Sway compiler, ordered alphabet
...
+## ABI Name
+
+The `#[abi_name]` attribute allows to specify the ABI name for an item.
+This means that when an ABI JSON file is generated, the name that is output is the one specified
+by the attribute. This can be useful to allow renaming items, while allowing for keeping backwards
+compatibility at the contract ABI level.
+
+> **Note**: At the moment, only enum and struct types support the attribute.
+
+In the example that follows, we originally had `MyStruct` and `MyEnum` types, which we, later on, renamed to `MyStruct` and `MyEnum` in code. To keep the backward compatibility of the ABI, we annotate the types with the `#[abi_name]` attribute and give them the original names:
```
ironcev:
```suggestion
In the example that follows, we originally had `MyStruct` and `MyEnum` types, which we, later on, renamed to `RenamedMyStruct` and `RenamedMyEnum` in code. To keep the backward compatibility of the ABI, we annotate the types with the `#[abi_name]` attribute and give them the original names:
```

sway-core/src/abi_generation/fuel_abi.rs
```diff
@@ -16,11 +20,27 @@ use crate::{
...
+#[derive(Clone, Debug)]
+pub enum AbiNameDiagnosticSpan {
+    Attribute(Span),
+    Type(Span),
+}
+
+impl AbiNameDiagnosticSpan {
+    pub fn span(self) -> Span {
+        match self {
+            Self::Attribute(span) => span,
```
ironcev: Nitpick!
```suggestion
            Self::Attribute(span) | Self::Type(span) => span,
```

sway-core/src/abi_generation/fuel_abi.rs
```diff
@@ -34,7 +54,79 @@ impl AbiContext<'_> {
     }
 }
 
+pub fn extract_abi_name_inner(span: &Span) -> Option<Span> {
+    let text = &span.src().text;
+    let full_attr: &str = &text[span.start()..span.end()];
```
ironcev: Aren't these two lines the same as `span.as_str()`?
```suggestion
    let full_attr: = span.as_str();
```

sway-core/src/abi_generation/fuel_abi.rs
```diff
@@ -51,7 +144,67 @@ impl TypeId {
...
+        let mut err: Option<ErrorEmitted> = None;
+
+        // Right now ABI renaming is only supported for enum and struct types.
+        let should_check_name = matches!(
+            *engines.te().get(resolved_type_id),
+            TypeInfo::Enum(_) | TypeInfo::Struct(_)
         );
+
+        if should_check_name {
+            let mut has_abi_name_attribute = false;
```
ironcev: Nitpick. Having a `mut` boolean here feels a bit over-complicated. One trivial possibility is to "calculate" `has_abi_name_attribute` after the `let (name, attribute_span)` or assign it within that same let: `let (has_abi_name, name, attr_span)`.

sway-core/src/abi_generation/fuel_abi.rs
```diff
@@ -51,7 +144,67 @@ impl TypeId {
...
+            let (name, attribute_span) =
+                match Self::get_abi_name_and_span_from_type_id(engines, resolved_type_id)? {
+                    Some(res) => {
+                        has_abi_name_attribute = true;
+                        res
+                    }
+                    None => (String::new(), Span::dummy()),
+                };
+
```
ironcev, line 162: E.g., here: `let has_abi_name_attribute = !(name.is_empty() || attribute_span.is_dummy());`

sway-error/src/error.rs
```diff
@@ -3112,6 +3122,32 @@ impl ToDiagnostic for CompileError {
                     help
                 },
             },
+            ABIDuplicateName { span, other_span: other, is_attribute } => Diagnostic {
+                reason: Some(Reason::new(code(1), "Duplicated name found for renamed ABI type.".into())),
```
ironcev: According to the guidelines, `Reason`s shouldn't end up in punctuation (we have messages that violates that, I know...)
```suggestion
                reason: Some(Reason::new(code(1), "Duplicated name found for renamed ABI type".into())),
```

sway-error/src/error.rs
```diff
@@ -3112,6 +3122,32 @@ impl ToDiagnostic for CompileError {
...
+                issue: Issue::error(
+                    source_engine,
+                    span.clone(),
+                    String::new()
+                ),
+                hints: vec![
+                    Hint::help(
+                        source_engine,
+                        other.clone(),
+                        format!("This is the existing {} with conflicting name", if *is_attribute { "attribute" } else { "type" }),
```
ironcev: According to the guidelines, hints shoul end up in punctuation (we have messages that violates that, I know...).
```suggestion
                        format!("This is the existing {} with conflicting name.", if *is_attribute { "attribute" } else { "type" }),
```

sway-error/src/error.rs
```diff
@@ -3112,6 +3122,32 @@ impl ToDiagnostic for CompileError {
...
+                    )
+                ],
+                help: vec![],
+            },
+            ABIInvalidName { span, name } => Diagnostic {
+                reason: Some(Reason::new(code(1), "Invalid name found for renamed ABI type.".into())),
```
ironcev, line 3142:
```suggestion
                reason: Some(Reason::new(code(1), "Invalid name found for renamed ABI type".into())),
```

sway-error/src/error.rs
```diff
@@ -3112,6 +3122,32 @@ impl ToDiagnostic for CompileError {
...
+                issue: Issue::error(
+                    source_engine,
+                    span.clone(),
+                    String::new()
+                ),
+                hints: vec![],
+                help: vec![format!("The name must be a valid Sway identifier{}", if name.is_empty() { " and cannot be empty" } else { "" })],
```
ironcev:
```suggestion
                help: vec![format!("The name must be a valid Sway identifier{}.", if name.is_empty() { " and cannot be empty" } else { "" })],
```

ironcev (changes requested):
One small mistake in the docs definitely worth fixing, otherwise only nitpicks :smile:

Regarding the comments about diagnostics guidelines, we should add those guidelines to the repo (currently they are only in the RFC) and polish existing diagnostics that violate them, but that's a low prio so far.

sway-parse/src/token.rs
```diff
@@ -105,6 +105,29 @@ pub fn lex(
     lex_commented(handler, src, start, end, &source_id).map(|stream| stream.strip_comments())
 }
 
+pub fn is_valid_identifier_or_path(s: &str) -> bool {
```
ironcev, line 121: If I interpret the last line of the implementation correctly, these strings will be consider valid:
- `_:`
- `:::`
- `a:b:c:::`
- `a::1234::b`
- etc.

And these will not, because `:` is not a `XID_Start`, although a valid Sway import path can start with it:
- `::some_module::in_the_same::package`

I assume such path, relative to the current package, cannot be used in the `#[abi_name]` attribute, but this is a general function in the `token.rs` so it should IMO accept all supported paths. Additional check can be done in the ABI name handling.

Proposal is to add Rust unit tests for this function.

test/src/e2e_vm_tests/test_programs/should_fail/attributes_invalid_abi_names/Forc.toml
```diff
@@ -0,0 +1,9 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "lib.sw"
+license = "Apache-2.0"
+name = "attributes_invalid_abi_names"
+
+[dependencies]
+std = { path = "../../../../../../sway-lib-std" }
```
ironcev, line 8: A reduced lib `sway-lib-std-core` can be used here.

test/src/e2e_vm_tests/test_programs/should_fail/attributes_invalid_abi_names/src/lib.sw
```diff
@@ -0,0 +1,70 @@
+contract;
```
ironcev: This is a contract, so it feels very strange that the entry point is not `main.sw` but `lib.sw`. Proposal is to rename this file to `main.sw`.

test/src/e2e_vm_tests/test_programs/should_fail/attributes_invalid_abi_names/Forc.toml
```diff
@@ -0,0 +1,9 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "lib.sw"
```
ironcev: Here as well.
```suggestion
entry = "main.sw"
```

test/src/e2e_vm_tests/test_programs/should_fail/attributes_invalid_args_values_types/src/abi_name_attr.sw
```diff
@@ -0,0 +1,4 @@
+library;
+
+#[abi_name(name = true)]
+pub fn not_ok() {}
```
ironcev: This must be either a struct or an enum, otherwise the test is not testing that the `name` argument value is of the wrong type, because the argument itself is misplaced.
```suggestion
pub struct NotOk {}
```

test/src/e2e_vm_tests/test_programs/should_fail/attributes_invalid_args_values_types/stdout.snap
```diff
@@ -45,6 +45,16 @@ error: Attribute argument value has a wrong type
   |
 ____
 
+error: Attribute cannot annotate item
+ --> test/src/e2e_vm_tests/test_programs/should_fail/attributes_invalid_args_values_types/src/abi_name_attr.sw:3:3
+  |
+...
+3 | #[abi_name(name = true)]
```
ironcev: As explained above, the test should show error that `true` is not a valid value, but it is showing that the function cannot be annotated with `#[abi_name]`.

ironcev (changes requested):

sway-parse/src/token.rs
ironcev: One more remark here, it looks like the current implementation will also accept these as valid identifiers:
- `__` (two underscores)
- `__this_is_not_a_valid_identifier`

In the regular, non compiler-generated, Sway, identifiers starting with two underscores are not allowed.

test/src/e2e_vm_tests/test_programs/should_fail/attributes_invalid_abi_names/src/lib.sw
tritao: Good spot, maybe something to consider checking in the future.

PR #7062, opened 2025-04-04
move codspeed benchmarks to buildjet

Voxelot:
## Description
The default ubuntu runners are too limited to run the codspeed benchmarks, this PR moves the benchmarks over to buildjet

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -35,7 +35,7 @@ jobs:
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
       - uses: taiki-e/install-action@v2
         with:
-          tool: cargo-binstall
+          tool: cargo-codspeed
 
       - name: Build the benchmark target(s)
         run: cargo codspeed build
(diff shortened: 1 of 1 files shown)


Voxelot:
Successfully ran the benchmarks on buildjet here: https://github.com/FuelLabs/sway/actions/runs/14256077382/job/39959538901

also reran the job to ensure the caching setup works properly

.github/workflows/benchmark.yml
```diff
@@ -5,11 +5,12 @@ on:
...
-    paths: 
-       - forc-plugins/forc-lsp/** 
-       - sway-lsp/** 
-       - forc-pkg/** 
-       - sway-core/**
+#    paths:
+#       - forc-plugins/forc-lsp/**
+#       - sway-lsp/**
+#       - forc-pkg/**
+#       - sway-core/**
```
JoshuaBatty: we probably want to add these conditions back in as these are the only workspace members that have an impact on the the LSP benchmarks that are run by codspeed

JoshuaBatty: haha just as I was typing this :)

Voxelot: This was just temporary so I could see the job actually run and verify it all worked correctly on buildjet

JoshuaBatty (approved):
🙏 thanks

sdankel (approved):

Issue #6765, opened 2024-11-29
Implement ABI Errors RFC


IGI-111:
Implement RFC from https://github.com/FuelLabs/sway-rfcs/pull/43

PR #7073, opened 2025-04-08
Implement `panic` expression

### LINKED ISSUE #7072: Confusing error message when calling `__log` with invalid number of arguments
E.g., this call:
```sway
script;

fn main() {
    __log(1, 2);
}
```
will result in the following error without any span provided:
```
error: Function "encode" expects 1 argument but you provided 2.
```

ironcev:
## Description

This PR introduces the `panic` expression to the language, as defined in the [ABI Errors RFC](https://github.com/FuelLabs/sway-rfcs/blob/master/rfcs/0014-abi-errors.md#panic).

The `panic` expression can be used without arguments, or accept an argument that implements the `std::marker::Error` trait. The `Error` trait is implemented by the compiler for the unit `()`, string slices, and `#[error_type]` enums.

Using the `panic` expression without arguments gives the symetry with the `return` expression and acts in the same way as having unit as an argument.

```
panic;
panic ();
panic "This is some error.";
panic Errors::SomeError(42);
```

Panicking without an argument or with unit as argument is discouraged to use. In the upcoming PR that finalizes the ABI errors feature, we will emit a warning if the `panic` is used without arguments or with unit as argument.

`panic` expression is available in all program kinds. In predicates it currently compiles only to revert. Once `__dbg` intrinsic is implemented, we can consider compiling to it in predicates. In the upcoming PR, the `error_codes` entry in the ABI JSON will be available for all program kinds.

The dead code analysis for the `panic` expression is implemented in the straightforward way, following the current approach of connecting reverts to the exit node. This will be revisted in a separate PR, together with the open TODOs in the DCA implementation of `Return`. Essentially, we want reverting/panicking to connect to program exit and implicit returns to the exit node.

Additionally, the PR:
- extends `forc test` CLI attributes with `--revert-codes` that prints revert codes even if tests are successful but revert.
- updates outdated "Logs Inside Tests" chapter in the documentation on unit testing.
- extends the snapshot testing infrastructure to support `forc test` in snapshot tests.
- fixes #7072.

Partially addresses #6765.

## Breaking Change

`panic` is a new reserved keyword. Once the `error_type` feature becomes integrated, compiling any existing code containing a "panic" as an identifier will result in an "Identifiers cannot be a reserved keyword." error.

In this PR, `panic` keyword is hidden behind the `error_type` feature flag. This prevents existing code from breaking, unless opted-in.

Note that, although being behind a feature flag, `panic` cannot be used in conditional compilation, means in combination with the `#[cfg(experimental_error_type = true/false)]`. The reason is that `cfg` evaluation happens after parsing, and we can either parse `panic` as a keyword or identifier during the parsing, because we cannot distinguish ambiguous cases like `panic;` or just `panic`.

This limitation means, that introducing `panic` to `std` can be done only once the `error_type` feature is fully integrated. In practice, this is not a serious limitation, because introducing `panic` in `std` would anyhow, due to effort and design decisions, happen after the feature is integrated.

We will provide a migration for this breaking change that will migrate the existing "panic" identifiers to "r#panic".

## Checklist


sdankel (approved):
The tooling changes look good to me 👍

test/src/snapshot/mod.rs
```diff
@@ -199,5 +199,17 @@ fn clean_output(output: &str) -> String {
         .unwrap();
     let result = r.replace(&result, "$1???");
 
+    // Remove individual test duration time and gas cost
+    let r = Regex::new("(test .+ )(\\(.*? gas\\))").unwrap();
```
xunilrj, line 203: Why removing gas usage? I think it is interesting to see if a change increase or decrease gas usage across the board.

ironcev: Agree and done.

Although, gas cost is always zero in tests when I run them no matter which tests. I just thought displaying gas is still not implemented but now when you say I must ask myself am I doing something wrong :thinking: I am talking about simply running `forc test`. The gas is always zero.

JoshuaBatty: I noticed this as well. @kayagokalp did some digging and tracked down this PR https://github.com/FuelLabs/sway/pull/6731 was when tests no longer reported any gas usage. @xunilrj do you have any idea why that PR may have caused this?

xunilrj: No. I will take a look.

ironcev:
Putting the PR temporarily into draft and marking it as blocked. This is just to avoid an accidental merge that would cause a temporary breaking change on the `master` - the `panic` keyword is not behind the feature flag.

Putting it behind the feature flag requires extending the feature flag infrastructure to the parsing stage. This will be done in a separate PR. Until that PR is merged this one is blocked.

docs/book/src/testing/unit-testing.md
```diff
@@ -71,9 +71,16 @@ fn test_meaning_of_life() {
 
 Tests with `#[test(should_revert)]` are considered to be passing if they are reverting.
 
+Revert codes are not shown by default in the passing test that have `should_revert`. To see the revert codes, use the `--revert-codes` flag, `forc test --revert-codes`:
```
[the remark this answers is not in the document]
ironcev: I've also noticed and fixed a few unnecessary "the"s, my usual mistakes :smile:

test/src/e2e_vm_tests/test_programs/should_pass/language/panic_expression/panicking_lib/src/lib.sw
```diff
@@ -0,0 +1,145 @@
+library;
+
+#[error_type]
+pub enum ErrorEnum {
+    #[error(m = "Error A.")]
```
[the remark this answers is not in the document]
ironcev: Yes, currently, we support only attributes of form `#[attribute(arg = value)`. Extending that support now would be an effort we rather postpone for later.

BDW, once we decide to do it, a thing to consider. In Rust, I am not aware of any standard attributes that use the `#[attr("value")]` form. The usual form id `#[attr = "value"]`. Custom attributes can accept any kind of tokens, and there we have `#[attr("value")]` cases, `#[error]` from `thiserror::Error` being a typical example. But since we do not support macros for the moment, it is worth consider of staying at `#[attr = "value"]` as a built in support.

forc-plugins/forc-migrate/src/visiting/mod.rs
```diff
@@ -634,6 +634,9 @@ impl __ProgramVisitor {
                     Self::visit_expr(ctx, lexed_returned.__as_ref(), ty_returned, visitor, output)?;
                 }
             }
+            Expr::Panic { expr_opt: _, .. } => {
+                // TODO: Implement visiting `panic`.
```
xunilrj, line 638: Would be nice to keep track of this.

xunilrj (approved):

forc-plugins/forc-migrate/src/visiting/mod.rs
ironcev: For `forc migrate` tool, there a lot of expressions that are not covered. They are all marked as TODO. The plan is not to invest the effort upfront into covering them but rather as a part of writing concrete migrations, if a need is there. That's why I didn't see it as necessary to open an issue for it. It is a known ongoing effort.

calldelegation (approved):

kayagokalp (dismissed):
We seem to have a tiny conflict happy to re-approve once it is fixed as well :)

bitzoic (dismissed):

JoshuaBatty (approved):

PR #7081, opened 2025-04-11
feat: yanking support for registry packages

### LINKED ISSUE #7077: forc-registry:  handle yanked published packages
We should also be able to yank published packages, and this logic should be enabled in forc to ensure no-yanked package is actually being used.

kayagokalp:
## Description

closes #7077.

Adds support for yanked packages. Forc will handle yanked packages by not accepting the version, and listing other possible versions.

Also adds other possible versions in the error message if the version requested is something that does not exists.

### DIFF
--- a/forc-pkg/src/source/reg/mod.rs
+++ b/forc-pkg/src/source/reg/mod.rs
@@ -390,6 +390,44 @@ impl From<Pinned> for source::Pinned {
     }
 }
 
+/// Resolve a CID from index file and pinned package. Basically goes through
+/// the index file to find corresponding entry described by the pinned instance.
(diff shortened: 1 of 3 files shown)


forc-pkg/src/source/reg/mod.rs
```diff
@@ -543,4 +570,110 @@ mod tests {
 
         assert_eq!(pinned, expected_pinned)
     }
+
+    #[test]
+    fn test_resolve_to_cid() {
```
zees-dev, line 575: nice test

zees-dev (approved):

JoshuaBatty (approved):
nice one

PR #7085, opened 2025-04-14
Subroutine calls using the new JAL instruction

Dentosal:
## Description

This PR contains an initial implementtion of subroutine calls using the in-progress [jump-and-link instruction `JAL`](https://github.com/FuelLabs/fuel-specs/pull/630). It substantially reduces the function call overhead: the old code used 4 instructions per call, while the new version uses 1-3 depending on the distance to the called function.

### Future optimizations

* Reorder functions, so those that call each other are adjacent
* Use absolute or IS-relative jumps where it makes sense, see https://github.com/FuelLabs/sway/issues/7267

## Checklist


sway-core/src/asm_lang/virtual_ops.rs
```diff
@@ -916,6 +922,7 @@ impl VirtualOp {
             JNE(_r1, _r2, _r3) => vec![],
             JNEI(_r1, _r2, _i) => vec![],
             JNZI(_r1, _i) => vec![],
+            JAL(r1, _r2, _i) => vec![r1],
```
xunilrj, line 925: Is this correct? I was expecting `JAL` to only write to the `$pc` register. In some examples below the first register is `$zero`, and we do not write to it.

sway-core/src/asm_lang/virtual_ops.rs
```diff
@@ -791,6 +796,7 @@ impl VirtualOp {
             JNE(r1, r2, r3) => vec![r1, r2, r3],
             JNEI(r1, r2, _i) => vec![r1, r2],
             JNZI(r1, _i) => vec![r1],
+            JAL(_r1, r2, _i) => vec![r2],
```
xunilrj, line 799: Is this correct? I was expecting `JAL` to read from both registers.

sway-core/src/asm_lang/virtual_ops.rs
```diff
@@ -664,6 +668,7 @@ impl VirtualOp {
...
```
xunilrj, line 671: Is this correct? I was expecting JAL to read from both registers.

sway-core/src/asm_lang/mod.rs
```diff
@@ -1419,20 +1386,7 @@ impl<Reg: Clone + Eq + Ord + Hash> ControlFlowOp<Reg> {
...
-            | Comment
-            | Jump { .. }
-            | DataSectionOffsetPlaceholder
-            | ConfigurablesOffsetPlaceholder
-            | PushAll(_)
-            | PopAll(_) => vec![],
-        })
-        .into_iter()
-        .collect()
+        BTreeSet::new()
```
xunilrj, line 1389: Shouldn' t this be like the `fn` above? Just removing `SaveRetAddr`?

sway-core/src/asm_lang/allocated_ops.rs
```diff
@@ -477,7 +479,7 @@ impl fmt::Display for AllocatedInstruction {
             WQMM(a, b, c, d) => write!(fmtr, "wqmm {a} {b} {c} {d}"),
 
             /* Control Flow Instructions */
-            JMP(a) => write!(fmtr, "jmp {a}"),
+            JMP(a) => write!(fmtr, "jmp  {a}"),
```
xunilrj, line 482: Nit. Extra space here.

sway-core/src/asm_lang/allocated_ops.rs
```diff
@@ -486,6 +488,7 @@ impl fmt::Display for AllocatedInstruction {
             JMPF(a, b) => write!(fmtr, "jmpf {a} {b}"),
             JNZB(a, b, c) => write!(fmtr, "jnzb {a} {b} {c}"),
             JNZF(a, b, c) => write!(fmtr, "jnzf {a} {b} {c}"),
+            JAL(a, b, c) => write!(fmtr, "jal  {a} {b} {c}"),
```
xunilrj, line 491: Nit two spaces before `{a}`.

sway-core/src/asm_lang/allocated_ops.rs
```diff
@@ -355,6 +356,7 @@ impl AllocatedInstruction {
             JMPF(_r1, _i) => vec![],
             JNZB(_r1, _r2, _i) => vec![],
             JNZF(_r1, _r2, _i) => vec![],
+            JAL(r1, _r2, _i) => vec![r1],
```
xunilrj, line 359: I was expecting `JAL` to not write `r1`. Is this correct?

sway-core/src/asm_lang/allocated_ops.rs
Dentosal: This is on purpose. The output format is neatly aligned with all opcodes taking 4 characters. You can see other branches here already doing that.

sway-core/src/asm_lang/allocated_ops.rs
Dentosal: This is on purpose. The output format is neatly aligned with all opcodes taking 4 characters. You can see other branches here already doing that.

sway-core/src/asm_lang/mod.rs
Dentosal: Possibly. I'm not sure if we still want to keep this so the interface looks same for all these variants.

sway-core/src/asm_lang/virtual_ops.rs
Dentosal: This is correct. The instruction only writes to r1 and only reads from r2.

sway-core/src/asm_lang/virtual_ops.rs
Dentosal: This is correct. We write the return address to r1. The [specification](https://github.com/FuelLabs/fuel-specs/blob/5682a258ca97b4deb2c2a9f3eb1fc6fb5ef86115/src/fuel-vm/instruction-set.md#jal-jump-and-link) clearly says that when JAL when writes to $zero, it discards the value.

sway-core/src/asm_lang/mod.rs
Dentosal: Upon reading this again, this is correct. Since all remaining variants produce no items, returning an empty set directly is exactly what should happen here.

xunilrj (approved):

PR #7088, opened 2025-04-14
Refactor jump label realization

Dentosal:
## Description

Simplifies jump instruction generation, moving the actual instruction selection to a single location. Removes lots of error-prone duplicate logic around far jump rewriting. This theoretically has a slight size penalty in jump-heavy contracts, but in practice this should be neglible. For instance `mira-v1-core` isn't affected at all. This PR is makes it easier to implement further optimizations, especially those around the new JAL instruction and u16/u32 integer loading opcodes. In particular, data section packing should be a lot easier after this.

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works: just a refactor, existing tests suffice


sway-core/src/asm_generation/fuel/allocated_abstract_instruction_set.rs
```diff
@@ -477,9 +412,17 @@ impl AllocatedAbstractInstructionSet {
             // Another special case for the blob opcode, used for testing.
             Either::Left(AllocatedOpcode::BLOB(ref count)) => count.value() as u64,
 
-            // These ops will end up being exactly one op, so the cur_offset goes up one.
-            Either::Right(Jump(..) | JumpIfNotZero(..) | Call(..) | LoadLabel(..))
-            | Either::Left(_) => 1,
+            // This is a concrete op, size is fixed
+            Either::Left(_) => 1,
+
+            //
```
Voxelot, line 418:
```suggestion
            // what goes here i wonder :) ?
```

sway-core/src/asm_lang/mod.rs
```diff
@@ -1279,19 +1305,22 @@ pub(crate) enum ControlFlowOp<Reg> {
...
-    Call(Label),
+    Jump {
+        /// Target label
+        to: Label,
+        /// Jump type
+        type_: JumpType<Reg>,
+        /// Force far jump, reserving space for an extra opcode to load target address.
+        /// If this is false but the target is far, [`mark_far_jumps`] pass will enable this.
+        /// Also, this will be enabled for self-jumps, as they require a noop to be inserted before them.
+        force_far: bool,
```
vaivaswatha, line 1316: Introducing a global (i.e., accessible everywhere the instruction can be) flag seems kind of ugly. Can you not just keep a `FxHashSet` of jumps requiring `force_far`? This can be created in `realize_labels`, and you pass it to `resolve_labels` as a mutable parameter  (or alternatively `resolve_labels` computes and returns it) and then directly used within the function when calling `compile_jump`.

sway-core/src/asm_generation/fuel/optimizations.rs
```diff
@@ -344,7 +344,7 @@ impl AbstractInstructionSet {
             reachables[op_idx] = true;
             let op = &ops[op_idx];
             for s in &op.successors(op_idx, ops, &label_to_index) {
-                if !reachables[*s] {
+                if !reachables[*s] && !worklist.contains(s) {
```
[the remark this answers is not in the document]
Dentosal: I don't understand how that would help. In any case, #7109 already did this change. We can refactor this later if it becomes an issue.

sway-core/src/asm_lang/mod.rs
Dentosal: Done in 75df97093

vaivaswatha (approved):

PR #7099, opened 2025-04-22
ci: move build binaries step to ubuntu-22.04

kayagokalp:
## Description

`build-and-release-forc-binaries` step failed in the last release because it was using an old OS version (20.04). The details can be seen [here](https://github.com/FuelLabs/sway/actions/runs/14605196507/job/40975381937) and [here](https://github.com/actions/runner-images/issues/11101).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -944,12 +944,12 @@ jobs:
     strategy:
       matrix:
         job:
-          - os: ubuntu-20.04
+          - os: ubuntu-22.04
             platform: linux
             target: x86_64-unknown-linux-gnu
             arch: amd64
             svm_target_platform: linux-amd64
-          - os: ubuntu-20.04
+          - os: ubuntu-22.04
             platform: linux
             target: aarch64-unknown-linux-gnu
             arch: arm64


JoshuaBatty (approved):

.github/workflows/ci.yml
```diff
@@ -944,12 +944,12 @@ jobs:
     strategy:
       matrix:
         job:
-          - os: ubuntu-20.04
+          - os: ubuntu-22.04
```
zees-dev, line 947: current/latest ubuntu LTS seems to be `24.04`; shall we not use that?

zees-dev: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

zees-dev (approved):

.github/workflows/ci.yml
zees-dev: approving since don't this to be a blocker; can re-approve if you update aswell

kayagokalp: We will move it off this runner anyways, let's merge this one and we will be using `ubuntu-latest` once the migration is done: https://github.com/FuelLabs/sway/issues/7102

PR #7109, opened 2025-04-23
codegen optimization: symbolic fuel-vm interpretation

Dentosal:
## Description

This PR adds an initial support to symbolic fuel-vm interpretation to improve codegen. The symbolic execution is designed to be extended in later PRs. In it's current form, it provides three distinct improvements:

1. If a temporary (allocated) register contains a value already held in another register, we can use that other register instead. This reduces register pressure, and removes unnecessary `mov` ops.
2. If a register already has the value we're setting it to, we can eliminate the instruction doing so.
3. For conditional jump instructions, if the value of the condition is known, one of the branches can be elminated. In such cases, the instruction is either removed or replaced with an unconditional jump. Dead code elimination pass can then potentially remove the now-unreachable branch.

In addition to these changes, this PR takes the optimization level into account for asm generation. For debug builds, we only run the optimization passes once. But for release builds we run the optimization passes multiple times, until we either cannot eliminate any more operations, or until a fixed upper limit of rounds. For instance, this means that if the new pass added in this PR manages to eliminate a branch, and then dead code elimination removes that code, it could allow the interpretation pass to do even more work.

To account for new optimization passes, I also organized them to mulple files, so that the ones which create private types are scoped well and thus easier to read.

## Impact

project | size before | size after | size reduction
-|-|-|-
mira-v1-core | 89.384 KB | 83.480 KB | 6.6%

## Checklist

- [x] I have updated the documentation where relevant (API docs, the reference, and the Sway book). No changes required!
- [x] I have added tests that prove my fix is effective or that my feature works. Just updated pre-existing ones!


sway-core/src/asm_generation/fuel/optimizations/symbolic_interpretation.rs
```diff
@@ -0,0 +1,232 @@
...
+impl AbstractInstructionSet {
+    /// Symbolically interpret code and propagate known register values.
+    pub(crate) fn interpret_propagate(mut self) -> AbstractInstructionSet {
+        if self.ops.is_empty() {
+            return self;
+        }
+
+        // The set of labels that are jump targets
+        // todo: build proper control flow graph instead
+        let jump_target_labels: FxHashSet<Label> = self
```
vaivaswatha, line 118: We could track `FxHashSet<Label, u32>`, i.e., how many times `label` is a target (or in other words, how many predecessors a label has).

With that ... [see next]

sway-core/src/asm_generation/fuel/optimizations/symbolic_interpretation.rs
```diff
@@ -0,0 +1,232 @@
...
+                }
+            }
+
+            // Some instructions can be further simplified with the known values.
+            match &mut op.opcode {
+                // Conditional jumps can be simplified if we know the value of the register.
+                Either::Right(ControlFlowOp::JumpIfNotZero(reg, lab)) => {
+                    if let Some(con) = known_values.resolve(reg).and_then(|r| r.value()) {
+                        if con == 0 {
+                            op.opcode = Either::Left(VirtualOp::NOOP);
```
vaivaswatha, line 152: Decrement `jump_target_label[lab]` here, and if it turns `0`, delete the key.

Dentosal: That's true. This is already handled by doing multiple passes, but if done like this, we could get more done in a single pass.

Dentosal: Done in da598c443

vaivaswatha (dismissed):

forc-plugins/forc-debug/tests/server_integration.rs
```diff
@@ -315,10 +315,10 @@ fn test_sourcemap_build() {
     // Verify essential source locations are mapped correctly
     let key_locations = [
         // Main function and its contents
-        (3, 4, "main function parameters"), // Should have 4 instructions
+        (3, 2, "main function parameters"), // Should have 4 instructions
```
JoshuaBatty, line 318:
```suggestion
        (3, 2, "main function parameters"), // Should have 2 instructions
```

forc-plugins/forc-debug/tests/server_integration.rs
```diff
@@ -315,10 +315,10 @@ fn test_sourcemap_build() {
...
         (4, 2, "addition operation"),       // Should have 2 instructions (add operation)
         // Helper function and its contents
-        (11, 4, "helper function parameters"), // Should have 4 instructions
+        (11, 2, "helper function parameters"), // Should have 4 instructions
```
JoshuaBatty, line 321:
```suggestion
        (11, 2, "helper function parameters"), // Should have 2 instructions
```

vaivaswatha (approved):

vaivaswatha (approved):

JoshuaBatty (approved):

PR #7110, opened 2025-04-24
chore: bump fuel-core to 0.43.1 and fuels to 0.72

kayagokalp:
## Description

Dependency bump to get fuels 0.72 and fuel-core 0.43.1 support.

### DIFF
--- a/forc-plugins/forc-client/src/op/call/missing_contracts.rs
+++ b/forc-plugins/forc-client/src/op/call/missing_contracts.rs
@@ -45,7 +45,7 @@ pub async fn get_missing_contracts(
         {
             Ok(_) => return Ok(call.external_contracts),
             Err(fuels_core::types::errors::Error::Transaction(
-                fuels::types::errors::transaction::Reason::Reverted { receipts, .. },
+                fuels::types::errors::transaction::Reason::Failure { receipts, .. },
             )) => {
                 let contract_id = find_id_of_missing_contract(&receipts)?;
                 call.external_contracts.push(contract_id);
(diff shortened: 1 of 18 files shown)


forc-plugins/forc-client/src/op/call/missing_contracts.rs
```diff
@@ -45,7 +45,7 @@ pub async fn get_missing_contracts(
         {
             Ok(_) => return Ok(call.external_contracts),
             Err(fuels_core::types::errors::Error::Transaction(
-                fuels::types::errors::transaction::Reason::Reverted { receipts, .. },
+                fuels::types::errors::transaction::Reason::Failure { receipts, .. },
             )) => {
                 let contract_id = find_id_of_missing_contract(&receipts)?;
```
zees-dev, line 50: nit: we may also be able to use the function from the rust sdk now: https://github.com/FuelLabs/fuels-rs/blob/dbc896f97765f1db5a30ac1d51b0ba69baa448aa/packages/fuels-programs/src/calls/utils.rs#L325
and with that, we could remove the implementation below.

zees-dev: could be another PR though; since its unrelated to the bump

kayagokalp: I think it is better to keep this one focused only to bump. We can create an issue for it.

zees-dev (approved):

JoshuaBatty (approved):

PR #7124, opened 2025-04-27
Add support for experimental features in the parsing phase

ironcev:
## Description

This PR brings support for experimental features to the parsing phase. This is needed so that we can support use cases like, e.g., adding new language keywords behind a feature flag. 

A concrete example would be the `panic` keyword added in #7073. The parsing of the `panic` token differs if it is interpreted as a keyword (when the `error_type` experimental feature is active) or as a regular identifier (when `error_type` is off).

Because `ExperimentalFeatures` are needed in the `Parse::parse(parser: &mut Parser)` methods, the `Parser` got extended with the `experimental` field.

The consequence of bringing the `ExperimentalFeatures` to the parsing phase is, that `swayfmt` and `forc fmt` can also require an experimental feature to be set, to properly parse a project. In those tools, `ExperimentalFeatures` are passed as a field on the `Formatter`.  

Note that this PR does not put the provided experimental flags into use. The first concrete usage will be in #7073.

This PR is a prerequisite for #7073. 

## Checklist


forc-plugins/forc-doc/src/lib.rs
```diff
@@ -174,15 +175,23 @@ fn build_docs(
...
         manifest,
         pkg_manifest,
     } = program_info;
 
+    let experimental = ExperimentalFeatures::new(
+        &pkg_manifest.project.experimental,
+        &experimental.experimental,
+        &experimental.no_experimental,
+    )
+    .map_err(|err| anyhow::anyhow!("{err}"))?;
```
zees-dev, line 193: ooks like weird design; cant exactly pinpoint what/why

ironcev: To calculate the final set of active experimental features, we need all three pieces of information. CLI input overrides the one from the package manifest file, and the CLI can specify both `--experimental` and `--no-experimental`. I don't quite see an obvious much better design that this that would provide all the three without introducing new types for grouping. The CLI input is already grouped within `CliFields` but that structure is meant to be used by `Clap`. Making `ExperimentalFeatures::new` depending on `CliFields` don't look ideal to me.

Do you see any better approach? I am open for suggestions here. 

@xunilrj As the original author, perhaps you can share your thoughts?

zees-dev: I can't foresee a better design either,   just wanted to highlight the point. Can ignore the comment in this case, should not be a blocker.

zees-dev (approved):

PR #7127, opened 2025-04-29
Upgrade CI and fix Clippy issues with Rust 1.86.0

tritao:
## Description

As title says.


JoshuaBatty (approved):

Cargo.lock
```diff
@@ -1918,7 +1918,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "18e4fdb82bd54a12e42fb58a800dcae6b9e13982238ce2296dc3570b92148e1f"
 dependencies = [
  "data-encoding",
- "syn 2.0.99",
+ "syn 1.0.109",
```
zees-dev, line 1921: why this decrement; is the lock file supposed to be checked in?

kayagokalp:
The current CI blockage is simply what i also had locally so reported to upstream : https://github.com/rust-lang/rust-clippy/issues/14685

In the mean time we can work around the false positive and rewrite 

```rust
for b in &mut bytes {
    *b = !*b;
}
```

to:

```rust
bytes.iter_mut().for_each(|b| *b = !*b);
```

kayagokalp (approved):

IGI-111 (approved):

Issue #7157, opened 2025-05-06
Compiler panics with "Function does not have CFEI instruction for locals"


ironcev:
TODO: Add detailed description.

```
thread 'main' panicked at sway-core/src/asm_generation/fuel/register_allocator.rs:729:31:
Function does not have CFEI instruction for locals
```

PR #7159, opened 2025-05-06
Run `forc-unit-tests` on CI in `release` and `debug` mode

ironcev:
## Description

This PR:
- changes the configuration of Forc and Forc plugins used in CI steps from `debug` to `release`.
- changes the `forc-unit-tests` CI step, to run the tests both in `debug` and `release` mode.

By switching to `release` builds, we ensure that we test the versions that are actually released, unlike the `debug` version. As a consequence, execution time of CI steps gets shortened. The drawback is potentially not catching issues detectable only in the `debug` mode, like, e.g., overflows. We will add a separate CI workflow, running nightly or on tagged versions only, that will use `debug` builds as explained in #7168.

Currently, we are running `forc-unit-tests` only in the `debug` mode. Compiler panics in `in_language_tests`, described in #7157, that happens only in `release` mode, shows that it is beneficial to run the unit test both in `debug` and `release` mode. Similar is with compiler panics described in #7161.

The PR also removes the explicit `forc build` part from those tests and leaves only the `forc test` part. The reasoning behind that change is, that `forc test` also builds the projects, indeed with tests included, but, since we currently don't support `#[cfg(test)]`, otherwise unchanged. Running `forc build` in addition effectively doubles the test execution time, and the actual benefit in testing, compared to running only `forc test` is, in best case, negligible. 

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,7 +127,7 @@ jobs:
         with:
           cache-provider: "buildjet"
       - name: Install Forc
-        run: cargo install --locked --debug --path ./forc
+        run: cargo install --locked --path ./forc
       - name: Build sway-lib-std
         run: forc build --path sway-lib-std
 
@@ -199,7 +199,7 @@ jobs:
         with:
           cache-provider: "buildjet"
       - name: Install forc-fmt
-        run: cargo install --locked --debug --path ./forc-plugins/forc-fmt
+        run: cargo install --locked --path ./forc-plugins/forc-fmt
       - name: Run the formatter against all sway projects and fail if any of them panic
         run: scripts/formatter/forc-fmt-check-panic.sh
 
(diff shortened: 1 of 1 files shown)


sdankel (approved):

.github/workflows/ci.yml
zees-dev, line 1: rather than the tests running sequentially, since these steps are indepedent; might be worthwhile running them in parallel?
Which I think can be achieved by running it in another job.

ironcev: Good point @zees-dev! It definitely makes sense, but I was curious about the overhead of the initial setup (toolchain, forc, ...) compared to the test execution time. Looking at a [concrete testrun](https://github.com/FuelLabs/sway/actions/runs/14860857245/job/41725095034?pr=7159) gives some interesting insights:
- the overall setup takes around a minute
- running `in_language_tests` in `debug` mode takes around a minute
- running `in_language_tests` in `release` mode takes around 8 minutes (!!!)

The last line is indicative and points out how huge the overhead of running IR optimizations can be when running a debug version of `forc`. That Rust dev builds are ~10 times or more slower is nothing surprising.

Which brings me again to the question, why don't we use the Forc `release` build on CI?

Aside form the obvious benefit of getting (much!) faster test execution, what bugs me even more with the current setup is the fact that we don't test the version that we actually release. As discussed, testing with `release` version of Forc on every PR can be combined with a nightly test that would use the `debug` version.

In short, if we stay at the current approach of using Forc `debug` for tests, splitting this job in two makes sense. If we start using the `release` version, the test execution time will be comparable to setup time and splitting the job will not pay off.

CC: @FuelLabs/tooling @Voxelot

zees-dev: Agree with the points made here. 
Especially the fact that the release version of the CI should be tested/run as the default choice since that's what reaches end-users.

kayagokalp (approved):
🚀

JoshuaBatty (approved):

.github/workflows/ci.yml
ironcev: As discussed, the used Forc and plugin versions are changed from `debug` to `release`. I was thinking of leaving the `debug` version of `forc-fmt` to be used in `forc-fmt-check-panic`, considering the stress-test nature of that test. On the other hand, since we want to introduce a separate workflow that will use `debug` builds (#7168), consistently using `release` also here looks fine to me.

JoshuaBatty (approved):

.github/workflows/ci.yml
```diff
@@ -239,22 +239,22 @@ jobs:
             target/
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
       - name: Install Forc
-        run: cargo install --locked --debug --path ./forc
+        run: cargo install --locked --path ./forc
       - name: Install Forc plugins
```
zees-dev, line 243: nit: coulduse a more programmatic approach to get the plugins; i.e. loop through all dir names in ./forc-plugins

ironcev: It makes sense to me, but I am thinking, this PR is already doing two things, and also, there might be a reason to have the plugins explicitly listed. So, I've rather created a separate issue for this change: #7169.

Issue #7111, opened 2025-04-24
Refactor LSP: Centralize Workspace-Level Resources


JoshuaBatty:
Currently, `sway-lsp` creates separate sessions with duplicated state (like `Engines` and `TokenMap`) for each member within a workspace, leading to inefficiency. This issue tracks the effort to refactor the language server to properly handle Sway workspaces by leveraging the existing workspace support within `forc-pkg`. 

Resources like `token_map`, `engines`, `diagnostics`, `runnables`, and `build_plan_cache` should be moved from existing within uniqe `Session` instances per workspace member, into the main `ServerState` so they can be shared amongst all workspace members. This refactoring will significantly improve performance and reduce memory usage for workspace users and will be implemented incrementally across several pull requests.

PR #7166, opened 2025-05-12
Refactor: Move `SyncWorkspace` from `Session` into `ServerState`

JoshuaBatty:
## Description
* Moves `SyncWorkspace` ownership from individual `Session` instances to a single, global instance within `ServerState`.
* `SyncWorkspace` is now initialized once when the LSP server starts, creating a single temporary workspace clone.

Also removes the `resync` function. This was causing the server and the IDE to fall out of sync with each other. For example, if you have 2 panels open and then save a file in one of the tabs but have unsaved changes in the other tab then those changes would not be reflected in the temp memory folder. This meant everything would be out of sync. We already are syncing everything on each keystroke per file so this fn was not necessary in the first place.

This PR is a precursor to https://github.com/FuelLabs/sway/pull/7139
part of https://github.com/FuelLabs/sway/issues/7111

## Checklist

### DIFF
--- a/sway-lsp/src/core/sync.rs
+++ b/sway-lsp/src/core/sync.rs
@@ -37,7 +37,7 @@ impl SyncWorkspace {
     }
 
     /// Clean up the temp directory that was created once the server closes down.
-    pub(crate) fn remove_temp_dir(&self) {
+    pub fn remove_temp_dir(&self) {
         if let Ok(dir) = self.temp_dir() {
             // The `temp_path` we store is `random_dir/project_name`.
             // So, we need to remove `random_dir` by getting the parent directory.
@@ -56,7 +56,7 @@ impl SyncWorkspace {
         }
     }
 
-    pub(crate) fn create_temp_dir_from_workspace(
+    pub fn create_temp_dir_from_workspace(
         &self,
         actual_workspace_root: &Path,
     ) -> Result<(), LanguageServerError> {
@@ -110,7 +110,7 @@ impl SyncWorkspace {
         Ok(())
     }
 
-    pub(crate) fn clone_manifest_dir_to_temp(&self) -> Result<(), DirectoryError> {
+    pub fn clone_manifest_dir_to_temp(&self) -> Result<(), DirectoryError> {
         copy_dir_contents(self.manifest_dir()?, self.temp_dir()?)
             .map_err(|_| DirectoryError::CopyContentsFailed)?;
 
@@ -118,7 +118,7 @@ impl SyncWorkspace {
     }
 
     /// Convert the Url path from the client to point to the same file in our temp folder
-    pub(crate) fn workspace_to_temp_url(&self, uri: &Url) -> Result<Url, DirectoryError> {
+    pub fn workspace_to_temp_url(&self, uri: &Url) -> Result<Url, DirectoryError> {
         convert_url(uri, &self.temp_dir()?, &self.manifest_dir()?)
     }
 
@@ -200,14 +200,14 @@ impl SyncWorkspace {
         }
     }
 
-    pub(crate) fn member_path(&self, temp_uri: &Url) -> Option<PathBuf> {
+    pub fn member_path(&self, temp_uri: &Url) -> Option<PathBuf> {
         let p = self.member_manifest_path(temp_uri)?;
         let dir = p.parent()?;
         Some(dir.to_path_buf())
     }
 
     /// Read the Forc.toml and convert relative paths to absolute. Save into our temp directory.
-    pub(crate) fn sync_manifest(&self) -> Result<(), LanguageServerError> {
+    pub fn sync_manifest(&self) -> Result<(), LanguageServerError> {
         let actual_manifest_dir = self.manifest_dir()?;
         let temp_manifest_dir = self.temp_dir()?;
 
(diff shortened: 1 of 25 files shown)


sway-lsp/src/handlers/request.rs
```diff
@@ -83,8 +84,9 @@ pub async fn handle_goto_definition(
         .await
     {
         Ok((uri, session)) => {
+            let sync = state.sync_workspace.get().unwrap();
```
zees-dev, line 87: nit: remove unwrap

sway-lsp/src/handlers/request.rs
```diff
@@ -127,12 +129,14 @@ pub async fn handle_hover(
     {
         Ok((uri, session)) => {
             let position = params.text_document_position_params.position;
+            let sync = state.sync_workspace.get().unwrap();
```
zees-dev, line 132: nit: unwrap here too

sway-lsp/src/core/sync.rs
```diff
@@ -23,88 +22,101 @@ pub enum Directory {
...
+                temp_workspace_base,
+                io_err
+            );
+            DirectoryError::CanonicalizeFailed
+        })?;
 
         self.directories
-            .insert(Directory::Manifest, manifest_dir.to_path_buf());
-        self.directories.insert(Directory::Temp, temp_path);
+            .insert(Directory::Manifest, actual_workspace_root.to_path_buf());
```
zees-dev, line 95: nit: canonical paths ensures prefix matching works reliably across OS-specific path representations and symlinks
```suggestion
            .insert(
                Directory::Manifest,
                actual_workspace_root.canonicalize().map_err(|_| DirectoryError::CanonicalizeFailed)?
            );
```

sway-lsp/src/core/sync.rs
```diff
@@ -23,88 +22,101 @@ pub enum Directory {
...
+    /// Clean up the temp directory that was created once the server closes down.
+    pub fn remove_temp_dir(&self) {
         if let Ok(dir) = self.temp_dir() {
-            dir.parent().map(fs::remove_dir);
+            // The `temp_path` we store is `random_dir/project_name`.
+            // So, we need to remove `random_dir` by getting the parent directory.
+            if let Some(parent_dir) = dir.parent() {
+                if parent_dir
+                    .to_string_lossy()
+                    .contains(SyncWorkspace::LSP_TEMP_PREFIX)
```
zees-dev, line 45: This string-based check may produce false positives/negatives if another directory in the path happens to contain the prefix as a substring.
Prefer testing only the directory’s own name.
i.e.
```suggestion
                if parent_dir
                    .file_name()
                    .is_some_and(|name| name.to_string_lossy().starts_with(SyncWorkspace::LSP_TEMP_PREFIX))
```
This is safer and avoids accidental deletion of unrelated directories.

sway-lsp/src/server.rs
```diff
@@ -31,6 +31,12 @@ impl LanguageServer for ServerState {
         if let Err(err) = self.register_forc_toml_watcher().await {
             tracing::error!("Failed to register Forc.toml file watcher: {}", err);
         }
+        // Populate documents from temp dir
+        if let Some(sw) = self.sync_workspace.get() {
+            if let Ok(temp_dir) = sw.temp_dir() {
+                let _ = self.documents.store_sway_files_from_temp(temp_dir).await;
```
zees-dev, line 37: nit: handle error
```suggestion
                if let Err(err) = self.documents.store_sway_files_from_temp(temp_dir).await {
                    tracing::warn!("Failed to populate documents from temp dir: {}", err);
                }
```

sway-lsp/tests/lib.rs
```diff
@@ -113,12 +114,12 @@ macro_rules! run_async {
...
     ($entry_point:expr, $capability:expr) => {{
-        let server = ServerState::default();
-        let uri = open(&server, $entry_point).await;
+        let (mut service, _) = LspService::new(ServerState::new);
+        let uri = init_and_open(&mut service, $entry_point).await;
 
         // Call the specific LSP capability function that was passed in.
-        let _ = $capability(&server, &uri).await;
-        let _ = server.shutdown_server();
+        let _ = $capability(&mut service.inner(), &uri).await;
```
zees-dev, line 121: nit: mut not needed
```suggestion
        let _ = $capability(&service.inner(), &uri).await;
```

sway-lsp/src/handlers/request.rs
zees-dev: i see this is common for the arc; in this case ignore comment above

sdankel (approved):
I agree with @zees-dev 's comments, otherwise LGTM. Nice performance improvements!

sway-lsp/src/core/sync.rs
JoshuaBatty: [964f7db](https://github.com/FuelLabs/sway/pull/7166/commits/964f7db2aaf796992abe3a9e9bce9a5f37cbda36)

sway-lsp/src/handlers/request.rs
JoshuaBatty: [f8099e5](https://github.com/FuelLabs/sway/pull/7166/commits/f8099e56dc90bac4e9673117cae178c3bb9f8b50)

zees-dev (approved):

PR #7172, opened 2025-05-13
feat: checksum manifest file in forc-pkg

kayagokalp:
## Description

This is the first of possibly many PRs to get package level caching. I decided to split this into pieces after trying to get everything done at one go.

This one basically adds checksum implementation for workspace and package level manifest files. Taking the following into account:

1. The dependencies and their versions (without taking the actual string representation in to account to prevent things like different checksums for forc.tomls with different order of dependencies)
2. The state of the all `.sw` files in the `src` folder

To give an example how to use this in the LSP:

1. Create checksum of the pkgs open in the LSP, hold it in the memory
2. If it is a single package this is simply a checksum string value, to decide whether to build it again or not, we can check the checksum again.
3. If it is a workspace we might want to hold a `ChecksumGraph` in the forc-pkg side. This will enable us to pin-point which exact package is different. 

Once we are able to serialize the compiled packages, and the engines to the disk, this can be used a primitive form of incremental building.

## TODO

- [ ] Create `ChecksumGraph` on forc-pkg side, and expose a function for consumers of forc-pkg to create this `ChecksumGraph`

### DIFF
--- a/forc-pkg/src/manifest/mod.rs
+++ b/forc-pkg/src/manifest/mod.rs
@@ -56,6 +56,8 @@ pub trait GenericManifestFile {
 
     /// Returns a mapping of member member names to package manifest files.
     fn member_manifests(&self) -> Result<MemberManifestFiles>;
+    /// Returns the checksum of the source described by this `ManifestFile`.
+    fn checksum(&self) -> Result<String>;
 }
 
 #[derive(Clone, Debug)]
@@ -136,6 +138,16 @@ impl GenericManifestFile for ManifestFile {
             ManifestFile::Package(pkg_manifest) => pkg_manifest.lock_path(),
             ManifestFile::Workspace(workspace_manifest) => workspace_manifest.lock_path(),
         }
+    }
+
+    /// Returns the checksum of the source described by this `ManifestFile`.
+    fn checksum(&self) -> Result<String> {
+        // We basically want to read Forc.toml, and anything under src/ folder.
+        // given that it ends with a .sw file.
+        //
+        // If this is a workspace manifest file, then we also need to read the
+        // top level Forc.toml and do the logic above for each member manifest.
+        todo!()
     }
 }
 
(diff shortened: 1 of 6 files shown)


forc-pkg/src/cache.rs
JoshuaBatty, line 1: do we need to check in an empty file?

forc-pkg/src/manifest/mod.rs
```diff
@@ -1148,11 +1186,500 @@ pub fn find_dir_within(dir: &Path, pkg_name: &str) -> Option<PathBuf> {
...
+
+            if path.is_file() && path.extension() == Some(std::ffi::OsStr::new("sw")) {
+                // Calculate relative path from base_path for consistent naming
+                let rel_path = path
+                    .strip_prefix(base_path)
+                    .unwrap_or(path)
+                    .to_string_lossy()
+                    .to_string();
+
+                println!("adding {:?}", rel_path);
```
zees-dev, line 1213: do we need this for prod?
If so; should we use the tracing libs print function(s)?

forc-pkg/src/manifest/mod.rs
```diff
@@ -1148,11 +1186,500 @@ pub fn find_dir_within(dir: &Path, pkg_name: &str) -> Option<PathBuf> {
...
+                // Read file contents
+                let content = std::fs::read(path).map_err(|e| {
```
zees-dev, line 1215: nit: Consider streaming each file into the hasher - instead of doing `std::fs::read`.

This skips the full-file `Vec<u8>` allocation and the extra copy into the hasher’s 64-byte buffer.
It barely matters for small projects, but on large workspaces the heap grows with total source size, so streaming can cut both memory and copy time.
Minimal example of streaming bytes from file:
```
fn hash_file(hasher: &mut Sha256, path: &std::path::Path) -> io::Result<()> {
    let mut reader = BufReader::new(File::open(path)?); // handle to the file; not read into heap
    let mut buf = [0u8; 8 * 1024];           // reusable 8 KiB stack buffer
    loop {
        let n = reader.read(&mut buf)?;
        if n == 0 { break; }
        hasher.update(&buf[..n]);            // stream chunk directly
    }
    Ok(())
}
```

Note: Can be another PR however (GH issue).

zees-dev (approved):
LGTM; thorough set of tests to validate functionality!

zees-dev (commented):
reverting approval until minor comments (non-nits) are addressed*

kmk142789 (approved):

PR #7204, opened 2025-05-29
feat: implicitly depend on std from registry, not via git

### LINKED ISSUE #7040: Publish std to forc.pub and add the std by registry by default instead of using git
Once #7038 is merged we will have a forc.pub <-> forc integration working and no longer need to rely on git for distributing the `std`.

This implies couple of stuff:

1. We need a way to publish forc packages to the registry using CI. Ideally we will have a nice CI job step that does this.
2. We need to decide on versioning. Are we going to version forc std package, starting from the version of forc and sway or is it going to start on `0.1` etc. 

Once these 2 preliminary work is done, we can switch to the registry system completely.

1st step is just a missing implementation that can be done in <1 day. But the 2nd one might take some time and it would be nice to start the discussion. Eager to hear your thoughts on this matter @FuelLabs/tooling

kayagokalp:
## Description

closes #7040. With this PR we are marking the registry supported *completed* we now have std deployed to forc.pub and forc using that std implicitly if no tag, branch, or specific git tag is supplied in env variables (this is needed by downstream, such ash the sdk for their CI etc, so we are keeping the old behavour for their dev flow).


forc-pkg/src/manifest/mod.rs
```diff
@@ -900,40 +900,49 @@ fn implicit_std_dep() -> Dependency {
...
-        git: std::env::var("FORC_IMPLICIT_STD_GIT")
-            .ok()
-            .or_else(|| Some(SWAY_GIT_REPO_URL.to_string())),
-        tag,
-        branch,
-        ..Default::default()
-    };
+    let git_target = std::env::var("FORC_IMPLICIT_STD_GIT").ok();
+
+    // If any onf the git based std variables is set, we select the git version
```
zees-dev, line 907: nit:
```suggestion
    // If any of the git based std variables is set, we select the git version
```

zees-dev (approved):
LGTM 👍

PR #7210, opened 2025-06-01
Keep track of interface surface status for entries in trait map.

tritao:
## Description

This adds bookkeeping information to the trait map so we can avoid emitting diagnostics for interface surface trait impls .

What happens is that during type checking, we first check the interface surface as a first pass, and only then do a second pass for the whole implementation, and this causes some issues with emitting diagnostics that should not be emitted.

## Checklist


tritao:
> LGTM but ideally we should add a Rust test for this to make sure we don't reintroduce the behavior.

I've added an assert now.

sway-core/src/semantic_analysis/ast_node/declaration/trait.rs
```diff
@@ -627,7 +630,9 @@ impl TyTraitDecl {
             Some(self.span()),
             IsImplSelf::No,
             IsExtendingExistingImpl::No,
+            IsImplInterfaceSurface::Yes,
         );
+        assert!(!interface_handler.has_errors());
```
JoshuaBatty: This assert! will crash the compiler (and LSP) in production if interface_handler.has_errors() returns true. Can we avoid this?

IGI-111: Yeah we should probably produce an ICE instead. But it ought to fail immediately, if this is present the output can't be guaranteed to be correct.

tritao: > This assert! will crash the compiler (and LSP) in production if interface_handler.has_errors() returns true. Can we avoid this?

Good point, would a `debug_assert` be acceptable here?

tritao: Maybe I can provide some context here, I think a `debug_assert` would be preferable here as the mere presence of an error in the handler does not necessarily indicate an internal problem that might lead to compilation failure, and using an ICE would abort of the compilation. But let me know and I can update this to use the ICE approach.

IGI-111: It's a simple question for me: if it's possible that this leads to semantically invalid binaries, it should be an ICE, if not it can be a debug assert for our debugging benefit.

As long as the error is sure to be caught by something else, this can be a recoverable error.

tritao: > It's a simple question for me: if it's possible that this leads to semantically invalid binaries, it should be an ICE, if not it can be a debug assert for our debugging benefit.

I was missing a not in my comment above, this does not lead to invalid binaries, hence why I don't think an ICE is necessary here.

IGI-111: Yeah alright then.

JoshuaBatty (approved):

